PR #29, opened 2019-11-30
Make trace and metrics features optional

### LINKED ISSUE #4: Split Jaeger exporter into its own crate
This will remove the main crate's dependency on Jaeger.

### LINKED ISSUE #5: Split Prometheus exporter into its own crate
This will remove the main crate's dependency on Prometheus.

jtescher:
Allow `trace` and `metrics` sdk features to be used optionally. This also allows the API to be used without including any SDK features for other potential SDK implementations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,12 +12,17 @@ license = "Apache-2.0"
 edition = "2018"
 
 [dependencies]
-crossbeam-channel = "0.3.9"
+crossbeam-channel = { version = "0.3.9", optional = true }
 lazy_static = "1.4.0"
-prometheus = "0.7.0"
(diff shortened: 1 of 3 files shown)


iredelmeier:
This seems like it still keeps the SDK coupled to Jaeger and Prometheus, where the idea (at the spec level) is that they would each be pulled in for exporters only. Happy to take a deeper look and/or lend a hand later! (Currently multitasking a bit 😅)

jtescher:
Yeah this is more of a separate issue where you may be using metrics or tracing individually. Can remove the part where it resolves #4 and #5 if we want to leave those open until the decoupling has happened.

jtescher:
Ok rebased

PR #30, opened 2019-12-01
Add ExportResult as specified in the specification

kichristensen:
The specification, https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md, specifies that the export should return the result of an export


jtescher:
It is currently ignored so this is safe to add

PR #33, opened 2019-12-02
Initial CODEOWNERS file

bogdandrutu:


CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+ 
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+
+# For anything not explicitly taken by someone else:
+*  @iredelmeier
```
iredelmeier: I believe we can configure this to use the team instead, so:

```suggestion
*  @open-telemetry/rust-approvers 
```

Issue #23, opened 2019-11-28
Store span kind


jtescher:
Currently span kind can only be assigned by setting the `span.kind` attribute. Span kind should be explicitly assignable when creating spans.

iredelmeier:
[From 12/03/19 SIG meeting]

This is likely blocked for now by #4

jtescher:
This is still blocked by #13

PR #41, opened 2019-12-04
Add separate OpenTelemetry Jaeger crate

### LINKED ISSUE #4: Split Jaeger exporter into its own crate

### LINKED ISSUE #38: Swap out rustracing_jaeger for jaeger exporter
Currently the span exporter infrastructure is bypassed because of [rustracing_jaeger](https://github.com/sile/rustracing_jaeger)'s implementation which reports spans on drop by sending them over a `crossbeam-channel`.

We should consider switching to use the [thrift](https://crates.io/crates/thrift) crate directly and generate a client as that would provide the majority of the implementation that the exporter requires. This could look very similar to the [go jaeger exporter implementation](https://github.com/open-telemetry/opentelemetry-go/tree/master/exporter/trace/jaeger).

This could be a separate crate as suggested in #4.

jtescher:
This patch adds an `opentelemetry-jaeger` crate to this workspace that includes a thrift generated client which implements the `SpanExporter` trait.

~~This implementation is a not yet functional, it is intended to raise design questions and show other architectural components that are required.~~

Design:

* ~~The API was mostly based on the [go jaeger exporter](https://github.com/open-telemetry/opentelemetry-go/tree/master/exporter/trace/jaeger), are there changes to make things more rust appropriate?~~ 
* ~~Should this interface be async?~~
* ~~For the collector uploader, should a [hyper client](https://github.com/hyperium/hyper/blob/master/src/client/mod.rs) be the base?~~ (using reqwest blocking client for now)

Missing components:

* ~~The SDK currently has no span processors~~ (addressed by #44)
* ~~`Provider`, `Tracer`, and `Span` need to be de-coupled from `rustracing_jaeger`~~ (addressed by #44)
* ~~Span internals are not exposed so they cannot be converted to thrift spans.~~ addressed by #44)
* ~~#30 needs to be merged~~
* ~~Requires #23~~ (addressed by #44)

Resolves #38 
Resolves #4


jtescher:
Rebased and appeased clippy

PR #44, opened 2019-12-07
Remove rustracing_jaeger

jtescher:
This patch removes `rustracing_jaeger` and replaces its implementation
with OpenTelemetry native logic.

In order to accomplish this it implements the following API types:

* `Event`: timestamped messages, stored with a configurable capacity.
* `Link`: Links to other spans, placeholder currently.
* `Sampler`: Interface for implementing sampling logic.
* `SamplingResult`: Records the result of sampling logic
* `SamplingDecision`: Enum that represents the type of decision reached.
* `SpanKind`: Enum that describes the relationship between the Span,
its parents and its children.
* `SpanStatus`: Enum that represents the status of a finished Span.
* `SpanProcessor`: An interface for processing finished spans.

Also the following exporter types:

* `SpanData`: Which can be used by exporters as standard input.

And finally the following SDK types:

* `EvictedQueue`: A queue with capacity that maintains a count of
dropped items.
* `Sampler`: SDK provided enum of default sampling strategies.
* `SimpleSpanProcessor`: A simple processor strategy that simply
forwards to the underlying exporter.

### Known unresolved design questions:

* The `Provider` <-> `Tracer` <- `Span` references here can probably be greatly improved, could also have a completely different structure, just wanted to get something out so we have a place to start discussing. Currently:
    * `Span` now has `Arc<SpanInner>` which has a clone of `Tracer`, `SpanInner` is reported to processors on `Drop`.
    * `Tracer` has a clone of `Provider`
    * `Provider` has a `Arc<ProviderInner>` which has a `HashMap<&'static str, Tracer>`
* How should the `Provider` builder accept `Exporter`s which can be simple or batched?
    * Currently `with_simple_exporter` is shown as a possible path.
* When working with `SpanData` as a `Processor` or `Exporter` should references or cloned values be passed instead of the current `Arc<SpanData>` impl?
* How should batch `SpanProcessor`s work? (probably not required for this PR but will need to answer eventually)
* Should there be multiple `SpanExporter` interfaces (single/batch, sync/async, etc) or is one that is always batch sufficient for now?

This also breaks the examples which currently rely on the jaeger impl, will need to have #41 before they are working again.


jtescher:
Note: The examples are broken currently, will be fixed once https://github.com/open-telemetry/opentelemetry-rust/pull/41 is merged.

PR #48, opened 2019-12-19
Add basic Jaeger exporter docs

jtescher:
Add crate level usage docs for the Jaeger exporter


bigkraig:
It looks like the opentelemetry-jaeger crate isn't being published to make these examples work

jtescher:
@bigkraig correct it has not yet been published. For now you can see them locally with `$ cargo doc --open`, PRs like this one are a step toward getting a `0.1.0` release 👍 .

PR #43, opened 2019-12-05
Add async tokio client example

jtescher:
Example of simple tokio client using async/await. This uses the current (and somewhat awkward) `mark_span_as_active`/`mark_span_as_inactive` API for demo and API improvement purposes.

Resulting traces:
image

Resolves #35

PR #49, opened 2019-12-21
Add Futures compatibility API

jtescher:
This adds an `Instrument` trait which allows futures to be extended with an `instrument` method, inspired by [tracing-futures](https://github.com/tokio-rs/tracing/tree/master/tracing-futures). This allows a span to be correctly associated with a given task, and fixes the issues highlighted by #43.

Using this API requires moving the span into the future it is instrumenting, so a `clone_span` method on `Tracer` was added to allow a given span to be associated with many futures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2018"
 
 [dependencies]
 lazy_static = "1.4.0"
+pin-project = { version = "0.4.6", optional = true }
 prometheus = { version = "0.7.0", optional = true }
 rand = { version = "0.7.2", optional = true }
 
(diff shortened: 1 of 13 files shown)


jtescher:
@iredelmeier yeah I had the example in https://github.com/open-telemetry/opentelemetry-rust/pull/43#issuecomment-563497213, but with them split across PRs it's a little harder to follow. One potential API improvement could be to add a `tracer` method on spans, in which case the tracer and span wouldn't both be necessary in this API, but it would constrain all impls where spans would always need to track their tracer. Thoughts?

PR #50, opened 2019-12-21
Adding actix-web example

knabben:
Initial actix-web middleware and parent span testing.

image

Missing metrics and move init_trace to a reusable module.

Run with: cargo run --example actix
Test with: curl localhost:8088/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ thrift = "0.13.0"
 futures = "0.3.1"
 tokio = { version = "0.2.10", features = ["full"] }
 
-
 [features]
 default = ["metrics", "trace"]
 trace = ["rand", "pin-project"]
@@ -36,4 +35,5 @@ serialize = ["serde", "bincode"]
 [workspace]
 members = [
     "opentelemetry-jaeger",
+    "examples/actix",
 ]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -20,6 +20,9 @@ rand = { version = "0.7.2", optional = true }
 hyper = "0.12.0"
 opentelemetry-jaeger = { path = "opentelemetry-jaeger", version = "0.1.0" }
 thrift = "0.13.0"
+futures = "0.1.25"
```
iredelmeier: Mind moving these deps so that they're only added for the example? (I can edit this PR or add a PR to your branch if you want.)

knabben: please feel free to add commits on this one.

jtescher:
Should this include the [server semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#http-server-semantic-conventions)? Have an example at https://github.com/OutThereLabs/actix-web-opentelemetry/ which includes many of them. Does add some complexity though, could be fine to have this as a simpler example as well to show usage more than full spec.

Cargo.toml
iredelmeier: Thanks!

iredelmeier:
@jtescher I think that would be a good addition but isn't necessary, given that this is an example (as opposed to a plugin that someone would import directly).

Issue #16, opened 2019-11-27
Add stdout trace exporter


iredelmeier:
A stdout exporter (or something that can exporter to a [`Write`r](https://doc.rust-lang.org/std/io/trait.Write.html) can be really valuable, especially for dev purposes.

The [go stdout trace exporter](https://github.com/open-telemetry/opentelemetry-go/tree/e6d725626d4629220a2de0112570adf80d50be21/exporter/trace/stdout) might be useful inspiration.

zoidbergwill:
I took a go at this, but it's gonna take a bit of work for us to better separate Jaeger from the core tracer, then this'll be easier to tackle.

Since the [`sdk::trace::Tracer`](https://github.com/open-telemetry/opentelemetry-rust/blob/master/src/sdk/trace/tracer.rs#L20) relies on a `Jaeger::Tracer`

zoidbergwill:
I think the same will be true for https://github.com/open-telemetry/opentelemetry-rust/issues/17 and the Prometheus stuff.

jtescher:
Clarifying the references between Provider <-> Tracer <-> Span, and how finished spans will get to a processor/exporter would be useful as well. Currently [rustracing_jaeger](https://github.com/sile/rustracing_jaeger) relies on crossbeam-channel and [the examples use a separate thread which listens to messges sent by spans on drop](https://github.com/sile/rustracing_jaeger/blob/master/examples/http_hello_server.rs#L48-L56). 

We should likely have a separate issue where we discuss the best architecture to move forward with as we fill out the current implementation gaps necessary to support additional exporters.

jtescher:
This is now unblocked after #44 landed 🎉

jtescher:
Resolved in #51

PR #51, opened 2019-12-22
Stdout exporter draft

knabben:
A draft for the stdout exporter, maybe it's a good choice to use serde for deserialization.

Related with #16

### DIFF
--- a/src/exporter/trace/stdout/mod.rs
+++ b/src/exporter/trace/stdout/mod.rs
@@ -3,12 +3,20 @@ use std::any;
 use std::sync::Arc;
 
 #[derive(Debug)]
-/// Exporter
-pub struct Exporter {}
-
-#[derive(Debug)]
 /// Builder
 pub struct Builder {}
+
+impl Builder {
+    /// Create a new builder
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    /// Build a new exporter
+    pub fn init() -> Exporter {
+        Exporter {}
+    }
+}
 
 impl Default for Builder {
     /// Return the default Exporter Builder.
@@ -17,19 +25,9 @@ impl Default for Builder {
     }
 }
 
-impl Builder {
-    /// Create a new exporter from the builder
-    pub fn init(self) -> Exporter {
-        Exporter {}
-    }
-}
-
-impl Exporter {
-    /// Create a new exporter builder.
-    pub fn builder() -> Builder {
-        Builder::default()
-    }
-}
+#[derive(Debug, Default)]
+/// Exporter
+pub struct Exporter {}
 
 impl trace::SpanExporter for Exporter {
     /// Export spans to stdout
--- a/src/exporter/trace/stdout/mod.rs
+++ b/src/exporter/trace/stdout/mod.rs
@@ -1,0 +1,49 @@
+use crate::exporter::trace;
+use std::any;
+use std::sync::Arc;
+
+#[derive(Debug)]
+/// Exporter
+pub struct Exporter {}
+
+#[derive(Debug)]
+/// Builder
+pub struct Builder {}
+
+impl Default for Builder {
+    /// Return the default Exporter Builder.
+    fn default() -> Self {
+        Builder {}
+    }
+}
+
+impl Builder {
+    /// Create a new exporter from the builder
+    pub fn init(self) -> Exporter {
+        Exporter {}
+    }
+}
+
+impl Exporter {
+    /// Create a new exporter builder.
+    pub fn builder() -> Builder {
+        Builder::default()
+    }
+}
+
+impl trace::SpanExporter for Exporter {
+    /// Export spans to stdout
+    fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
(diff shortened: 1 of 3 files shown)


src/exporter/trace/stdout/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+    fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
+        println!("{:?}", batch);
+
+        trace::ExportResult::Success
+    }
+
+    /// Ignored for now.
+    fn shutdown(&self) {}
+
+    /// Allows `Exporter` to be downcast from trait object.
```
MikeGoldsmith: Do we need this?

knabben: error[E0046]: not all trait items implemented, missing: `as_any` <- is required from the trait

knabben: error[E0046]: not all trait items implemented, missing: `as_any`, is required from the trait.

MikeGoldsmith: Thanks - I checked and could see it was part of other implementations.

src/exporter/trace/stdout/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Create a new exporter builder.
+    pub fn builder() -> Builder {
+        Builder::default()
+    }
+}
+
+impl trace::SpanExporter for Exporter {
+    /// Export spans to stdout
+    fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
+        println!("{:?}", batch);
```
iredelmeier: @knabben and others: thoughts on instead writing to a `Write`-r, rather than using `println!`? The `Builder` could then default to [`Stdout`](https://doc.rust-lang.org/std/io/struct.Stdout.html).

src/exporter/trace/stdout/mod.rs
```diff
@@ -0,0 +1,49 @@
+use crate::exporter::trace;
+use std::any;
+use std::sync::Arc;
+
+#[derive(Debug)]
+/// Exporter
+pub struct Exporter {}
+
+#[derive(Debug)]
```
iredelmeier:
```suggestion
#[derive(Debug, Default)]
pub struct Exporter {}

impl Builder {
  pub fn new() -> Self {
    Self {}
  }
  
  pub fn build() -> Exporter {
    Exporter {}
  }
}
```

src/exporter/trace/stdout/mod.rs
```diff
@@ -0,0 +1,49 @@
+use crate::exporter::trace;
+use std::any;
+use std::sync::Arc;
+
+#[derive(Debug)]
+/// Exporter
```
iredelmeier: Rust conveniently adds doc scaffolding out-of-the-box for all public code, so doc comments specifying the names of structs (and functions, etc.) are unnecessary.

src/exporter/trace/stdout/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+impl Builder {
+    /// Create a new exporter from the builder
+    pub fn init(self) -> Exporter {
+        Exporter {}
+    }
+}
+
+impl Exporter {
+    /// Create a new exporter builder.
+    pub fn builder() -> Builder {
```
iredelmeier: Why can `Exporter`s create `Builder`s? That direction seems counter-intuitive to me.

src/exporter/trace/stdout/mod.rs
knabben: I'm replacing build to init(), so we can have the same methods as the jaeger export in the Builder

src/exporter/trace/stdout/mod.rs
knabben: https://github.com/open-telemetry/opentelemetry-rust/blob/master/src/lib.rs#L205 - This does not allow me to pass the lint without the comments.

src/exporter/trace/stdout/mod.rs
knabben: Indeed, I get the skeleton from jaeger exporter, but the usage is a little different, cleaning up as you suggested, thanks :)

src/exporter/trace/stdout/mod.rs
knabben: Makes sense, I was thinking is use some JSON serialization as well, since now we have the dump of the object.

MikeGoldsmith:
Hey @knabben - I've just reviewing open PRs and can see we missed your recent change. Thanks for making updates regarding feedback.

The only open discussion is regarding writing to Builder` instead of `println`. How do you feel about this now?

@jtescher any comments from yourself?

jtescher:
Other than stdout looks good to me 👍

knabben:
I think @iredelmeier had some other modifications to do here, waiting for her on this one.

MikeGoldsmith:
I think we should go with what we have here, and we can create an follow up change if needed. We don't want this PR to totally stall waiting for feedback.

Issue #58, opened 2020-01-20
Add serialize attributes for SpanData


jodh-intel:
I'd like to be able to serialise and deserialise the public `SpanData` [1]. Since all of its fields are public I *could* do that manually or even use the magic "remote derive" feature of serde [2]. However, from a safety and usability perspective adding the standard `#[derive(Serialize, Deserialize)]` to `SpanData` would allow standard serde semantics.

---

[1] - https://github.com/kata-containers/kata-containers/issues/60
[2] - https://serde.rs/remote-derive.html

jodh-intel:
btw, do you have any thoughts on when you plan to update the crate version (since `SpanData` isn't available in the latest `0.1.5` release on crates.io)?

MikeGoldsmith:
Hi @jodh-intel 

We don't have a set date for our next release to crates.io, though it has been a while and there have been a lot of changes since the last release (including moving github org).

jodh-intel:
Thanks @MikeGoldsmith - I've tried using the latest git repo version but it got messy due to other deps (and their particular versioned deps) so would be great to be able to consume direct from crates.io ideally.

@jtescher - do you think adding these serialisation options is reasonable?

jtescher:
@jodh-intel yeah `SpanData` being serialize/deserialize seems reasonable to me. Want to do the PR? (always good to have more committers 😃)

As far as a release, once the current open PR's land we can likely get a release ready. The current issues around how current span tracking works with futures is the real meat of the outstanding work to be done.

jodh-intel:
Thanks @jtescher - here you go: https://github.com/open-telemetry/opentelemetry-rust/pull/59 :smile:

PR #59, opened 2020-01-27
SpanData: Add serde serialise support

jodh-intel:
Added `Serialize` and `Deserialize` attributes to allow `SpanData` to be serialized and deserialized using the `serde` crate.

Also added a basic unit test for `SpanData` serialization and deserialization.

Fixes: #58.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,8 @@ lazy_static = "1.4.0"
 pin-project = { version = "0.4.6", optional = true }
 prometheus = { version = "0.7.0", optional = true }
 rand = { version = "0.7.2", optional = true }
+serde = { version = "1.0.104", features = ["derive"], optional = true }
+bincode = { version = "1.2.1", optional = true }
 
 [dev-dependencies]
 hyper = "0.12.0"
@@ -28,6 +30,7 @@ tokio = { version = "0.2.10", features = ["full"] }
 default = ["metrics", "trace"]
 trace = ["rand", "pin-project"]
 metrics = ["prometheus"]
+serialize = ["serde", "bincode"]
 
 [workspace]
 members = [
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -3,3 +3,4 @@
 set -eu
 
 cargo test --all "$@"
+cargo test --all "$@" --features="default serialize"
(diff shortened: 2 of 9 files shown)


jodh-intel:
Let's see if this re-triggers the cla/linuxfoundation check...

I signed it.

jodh-intel:
nope ;( A force-push also didn't help.

jodh-intel:
I signed it

jtescher:
@jodh-intel looks good. You can rebase now to make CI pass. Also do you think it makes sense to have this as a `serde` feature to avoid version incompatibilities?

jodh-intel:
@jtescher - good point - branch updated.

jodh-intel:
Branch updated again to add an explicit test for the new feature.

btw, is your CI testing the `cargo build` "`release`" profile?

Cargo.toml
```diff
@@ -16,18 +16,21 @@ lazy_static = "1.4.0"
...
 rand = { version = "0.7.2", optional = true }
+serde = { version = "1.0.104", features = ["derive"], optional = true }
 
 [dev-dependencies]
 hyper = "0.12.0"
 opentelemetry-jaeger = { path = "opentelemetry-jaeger", version = "0.1.0" }
 thrift = "0.13.0"
 futures = "0.3.1"
 tokio = { version = "0.2.10", features = ["full"] }
+bincode = "1.2.1"
```
jtescher: Should probably mark this as optional as well and include in the serialize feature

jtescher:
@jodh-intel I didn't set up the CI, but you can see how it is executed in https://github.com/open-telemetry/opentelemetry-rust/blob/master/.github/workflows/ci.yml. Could consider running tests with full feature enabled/disabled matrix at some point

Cargo.toml
jodh-intel: @jtescher - Done. I had thought about this but it appeared to be the only way to resolve the issue: it is a test dependency (only) so naturally lives under `dev-dependencies`. But since it's only required for the `serialize` feature and since `dev-dependencies` cannot be optional, I assumed it would have to stay there. However, it seems that making it a dependency of the `serialize` feature as you suggest dtrt (`bincode` isn't linked into the resulting binaries regardless of whether `serialize` is enabled or not ;)

PR #75, opened 2020-03-11
Introduce Zipkin exporter

Monkeyanator:
This PR adds an experimental Zipkin exporter with configurable collector endpoint. Wanted to put up a draft PR to get a critique on the high-level implementation direction before I delved further into it (below implementation exports spans correctly into the Zipkin collector!)

As I understand it, [to ingest spans into a Zipkin collector](https://github.com/openzipkin/zipkin/tree/master/zipkin-server#collector-1), you can either: 1) post JSON to their HTTP API at /api/v2, 2) post Protobuf to their HTTP API with the `Content-Type: x-protobuf` header, or 3) gRPC service running on same port as HTTP API (experimental, off by default). This WIP implements only POST to the collector's HTTP API, but in the future will be extensible with the `UploaderFormat` being configurable.

There are definitely rough edges around the implementation below. Obvious work remaining:

- [x] ~~Add Protobuf export format support~~ pushing back, scoping this PR to the basic HTTP exporter
- [x] Get rid of the `.unwrap()`s from the `derive_builder` builders (either manually writing Builders or using [rust typed builder](https://github.com/idanarye/rust-typed-builder)) update: ended up switching to rust-typed-builder (which poses its own limitations with not being able to pass around the builder)
- [x] Finish adding fields on the Zipkin model structs (there are a few fields missing from the Span model)
- [x] Improve the clunkiness around the `LocalEndpoint` configuration of the `Exporter`
- [x] Document usage in `lib.rs`

Adding for reference, implementation of Python [OC span to Zipkin span conversion](https://github.com/census-instrumentation/opencensus-python/blob/bb02e5bc7c298743a0c86088ed960f650e5c171a/contrib/opencensus-ext-zipkin/opencensus/ext/zipkin/trace_exporter/__init__.py#L124).

### DIFF
--- a/opentelemetry-zipkin/Cargo.toml
+++ b/opentelemetry-zipkin/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "opentelemetry-zipkin"
+version = "0.0.1"
+authors = ["OpenTelemetry Authors <cncf-opentelemetry-contributors@lists.cncf.io>"]
+description = "Zipkin exporter for OpenTelemetry"
+homepage = "https://github.com/open-telemetry/opentelemetry-rust"
+repository = "https://github.com/open-telemetry/opentelemetry-rust"
+readme = "README.md"
+categories = ["development-tools::debugging"]
+keywords = ["opentelemetry", "zipkin"]
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+opentelemetry = "0.2.0"
+reqwest = { version = "0.10.4", features = ["blocking"] }
+serde_json = "1.0"
+serde = { version = "1.0.104", features = ["derive"] }
+typed-builder = "0.5.1"
--- a/opentelemetry-zipkin/src/lib.rs
+++ b/opentelemetry-zipkin/src/lib.rs
@@ -1,0 +1,248 @@
+//! # OpenTelemetry Zipkin Exporter
+//!
+//! Collects OpenTelemetry spans and reports them to a given Zipkin
+//! `collector` endpoint. See the [Zipkin Docs](https://zipkin.io/) for details
+//! and deployment information.
+//!
+//! ### Zipkin collector example
+//!
+//! This example expects a Zipkin collector running on `localhost:9411`.
+//!
+//! ```rust,no_run
+//! use opentelemetry::{api::Key, global, sdk};
+//! use opentelemetry_zipkin::ExporterConfig;
+//! use std::net::{SocketAddr, IpAddr, Ipv4Addr};
+//!
+//! fn init_tracer() {
+//!     let exporter = opentelemetry_zipkin::Exporter::from_config(
+//!        ExporterConfig::builder()
+//!            .with_service_name("opentelemetry-backend".to_owned())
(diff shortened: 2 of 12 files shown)


jtescher:
👍 Great start! yeah there isn't a real clear opinion on which builder pattern to use in this codebase yet.

jtescher:
Are you looking at [Tonic](https://github.com/hyperium/tonic) for the gRPC client?

Monkeyanator:
Just looked into it briefly and Tonic seems like a good choice. Should we shoot to get gRPC support in the initial release of this exporter, or get the basic HTTP exporter finished and then merge it in later (since the gRPC endpoint on Zipkin is currently experimental)?

jtescher:
I think getting the basics working first is a good approach. Providing some level of zipkin support would be nice.

Monkeyanator:
Addressed most of the TODOs and graduated from Draft status, should be ready for a closer look.

opentelemetry-zipkin/src/uploader.rs
```diff
@@ -0,0 +1,57 @@
...
+impl Uploader {
+    pub(crate) fn new(collector_endpoint: String, format: UploaderFormat) -> Self {
+        Uploader {
+            format,
+            client: reqwest::blocking::Client::new(),
+            collector_endpoint: format!("http://{}{}", collector_endpoint, API_V2_COLLECTOR_ROUTE),
+        }
+    }
+
+    /// Upload a `ListofSpans` to the designated Zipkin collector
```
jtescher: small typo `ListOfSpans`

opentelemetry-zipkin/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "opentelemetry-zipkin"
+version = "0.0.1"
+authors = ["OpenTelemetry Authors <cncf-opentelemetry-contributors@lists.cncf.io>"]
+description = "Zipkin exporter for OpenTelemetry"
+homepage = "https://github.com/open-telemetry/opentelemetry-rust"
+repository = "https://github.com/open-telemetry/opentelemetry-rust"
+readme = "README.md"
+categories = ["development-tools::debugging"]
+keywords = ["opentelemetry", "zipkin"]
+license = "Apache-2.0"
```
jtescher: Want to copy the LICENSE file from the root into this crate as well?

opentelemetry-zipkin/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+                uploader::UploaderFormat::HTTP,
+            )),
+        }
+    }
+}
+
+impl trace::SpanExporter for Exporter {
+    /// Export spans to Zipkin collector.
+    fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
+        match self.uploader.lock() {
```
jtescher: I don't think you need this Mutex (it is needed by the jaeger exporter because the thrift api requires mutable references, but reqwest does not) the functions in uploader could just take `&self`.

jtescher:
Want to also add the other root files here as well? `CODEOWNERS`, `README.md`, `CHANGELOG.md`, etc.

jtescher:
Thanks!

PR #81, opened 2020-03-19
Add base64 conversion traits

### LINKED ISSUE #54: Add a Base64 decoder
Often span IDs between applications are encoded into base64, currently you need to use an external crate i.e. https://crates.io/crates/base64 it would be more user friendly

xd009642:
I'll add some tests and more doc comments later just wanted to make sure this was in the ballpark of what was expected. 

Fixes #54

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ license = "Apache-2.0"
 edition = "2018"
 
 [dependencies]
+base64 = "0.12"
 lazy_static = "1.4.0"
 pin-project = { version = "0.4.6", optional = true }
 prometheus = { version = "0.7.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ license = "Apache-2.0"
 edition = "2018"
 
 [dependencies]
-base64 = "0.12"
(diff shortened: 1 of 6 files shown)


jtescher:
@xd009642 yeah this direction looks good, maybe have this be an optional feature like serde is currently? Nice to keep the dependency list short for people who don't need all the features.

xd009642:
I can do that once I finish work today :+1: base64 is a small dependency though with no subdependencies so should I at least make it a default feature?

xd009642:
also I signed the cla/linuxfoundation so I guess that will switch to pass on the next commit I push?

jtescher:
Sounds good! Normally I think default feature could make sense, however the binary format is in flux / may be removed so probably best to have this be opt-in for now.

xd009642:
@jtescher I've moved base64 to it's own feature and also decided to do a blanket implementation for anything that implements `BinaryFormat` filled in the `to_base64` test replicating the binary format `to_bytes` test. Just need to fill in some more tests for the `from_base64`

xd009642:
I signed it

xd009642:
I signed it

xd009642:
Done :smile:

PR #86, opened 2020-03-30
Add batch span processor

### LINKED ISSUE #77: Add batch span processor
See the [batching processor](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#batching-processor) specs for details.

jtescher:
This processor can be configured with an executor to batch and upload spans asynchronously when they end. If a library like `tokio` or `async-std` is present, the batch span processor builder can accept their respective `spawn` and `interval` functions to have batching performed in those contexts.

Example:

```rust
use futures::{stream};
use opentelemetry::{api, sdk, global};
use std::time::Duration;

async fn main() {
    let exporter = api::NoopSpanExporter {};

    // Use the builder and tokio's `spawn` and `time::interval` functions
    let batch = sdk::BatchSpanProcessor::builder(exporter, tokio::spawn, tokio::time::interval)
        .with_max_queue_size(4096)
        .build();

    // Then use the `with_batch_exporter` method to have the provider export
    // spans in batches.
    let provider = sdk::Provider::builder()
        .with_batch_exporter(batch)
        .build();

    global::set_provider(provider);
}
```

Resolves #77


src/sdk/trace/span_processor.rs
```diff
@@ -66,3 +129,215 @@ impl api::SpanProcessor for SimpleSpanProcessor {
...
+            spawn,
+            interval,
+            config: Default::default(),
+        }
+    }
+}
+
+/// Batch span processor configuration
+#[derive(Debug)]
+pub struct BatchConfig {
```
MikeGoldsmith, line 266: Other implementations also have a _block_ property that won't drop spans and prevents the app from closing until all the spans have been exported.

This can be done with a subsequent PR if preferred.

jtescher: Yeah this could be useful, will add an issue to track 👍

PR #87, opened 2020-03-30
Call shutdown as needed

Xaeroxe:
Currently the shutdown method is simply never called. This presents issues for data integrity in exporters, so this PR adds a call to the shutdown methods as needed.


jtescher:
Looks great, thanks! just have to sign the CLA and we can get this merged

Xaeroxe:
I signed it

PR #91, opened 2020-04-05
Clean up derive usage

jtescher:
This patch improves ergonomics of using this crate by:

* Deriving `Clone` for `B3Propagator`, `SamplingResult`, and `SpanBuilder`.
* ~~Deriving `Copy` for `SamplingDecision`, `SpanKind`, `StatusCode`, `ExportResult` and `Sampler`~~.
* Ensuring that impls of `Span` are `Send` and `Sync` when used in `global` module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,14 +17,14 @@ and metrics from your application. You can analyze them using [Prometheus], [Jae
 ## Quick Start
 
 ```rust
-use opentelemetry::{api::{Provider, TracerGenerics}, global, sdk};
+use opentelemetry::{api::TracerGenerics, global, sdk};
 
 fn main() {
     global::set_provider(sdk::Provider::default());
 
-    global::trace_provider().get_tracer("component-a").with_span("foo", |_span| {
-        global::trace_provider().get_tracer("component-b").with_span("bar", |_span| {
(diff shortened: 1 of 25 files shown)


hawkw:
FWIW, I'd kind of advise against deriving `Copy` in many cases where it is not absolutely necessary for performance reasons. It poses a forward-compatibility hazard, since any potential future change to a type which adds data that is not trivially copyable would necessitate the removal of the `Copy` implementation, and removing a trait implementation is a breaking change. Just something to think about as APIs continue to stabilize.

jtescher:
@hawkw good point, forward compatibility is likely more important that ergonomics in these cases.

hawkw:
@jtescher of course, it's a bit of a judgement call: it depends on whether or not there's the possibility of those types changing in such a way that they can no longer implement `Copy`. If you're confident that a type will _never_ change in such a way, or that it's unlikely enough that it's okay to have to make a breaking change, adding a `Copy` impl is probably okay. 

Since several of the types that now derive `Copy` are public `enum`s, any change to them would be a breaking change _anyway_ (since enum matching is exhaustive). If there are already enums in a public API, adding `Copy` impls to them isn't making the forward-compatibility hazard any _worse_...

PR #92, opened 2020-04-06
Add id generator

jtescher:
Allow span and trace id generation to be customized. The default impl uses the `rand` crate's thread local rng.

### DIFF
--- a/src/api/trace/id_generator.rs
+++ b/src/api/trace/id_generator.rs
@@ -1,0 +1,14 @@
+//! # OpenTelemetry Id Generator Interface
+use crate::api;
+use std::fmt;
+
+/// Interface for generating IDs
+pub trait IdGenerator: Send + Sync + fmt::Debug {
+    /// Generate a new `TraceId`
+    fn new_trace_id(&self) -> api::TraceId;
+
+    /// Generate a new `SpanId`
+    fn new_span_id(&self) -> api::SpanId {
+        api::SpanId::from_u64(0)
+    }
+}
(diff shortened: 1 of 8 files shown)


src/api/trace/id_generator.rs
```diff
@@ -0,0 +1,14 @@
...
+use std::fmt;
+
+/// Interface for generating IDs
+pub trait IdGenerator: Send + Sync + fmt::Debug {
+    /// Generate a new `TraceId`
+    fn new_trace_id(&self) -> api::TraceId;
+
+    /// Generate a new `SpanId`
+    fn new_span_id(&self) -> api::SpanId {
+        api::SpanId::from_u64(0)
```
hawkw, line 12: Take it or leave it, but IMO a default implementation that would assign every span the ID 0 seems kinda dangerous...

src/sdk/trace/id_generator.rs
```diff
@@ -0,0 +1,25 @@
+//! Id Generator
+use crate::api;
+use rand::{rngs, Rng};
+use std::cell::RefCell;
+
+/// Generates Trace and Span ids
+#[derive(Clone, Debug, Default)]
+pub struct IdGenerator {}
```
hawkw, line 8: Users can currently construct this with `IdGenerator {}`. However, that means that if fields are added to this type in the future, any code that constructs it will break.

For forward-compatibility reasons, you may want to give this a private field, like so:
```suggestion
pub struct IdGenerator {
    _private: (),
}
```
This would mean users would have to use `IdGenerator::default()` or `IdGenerator::new()` (if one was added). If additional fields are added in the future, code that uses `default` will still work.

src/api/trace/id_generator.rs
jtescher: Yes thanks for catching that! It was left here by mistake, working on too many things at once...

PR #95, opened 2020-04-09
Add global `tracer` method

jtescher:
This adds `global::tracer(name)` as a simpler way of expressing `global::trace_provider().get_tracer(name)`.


jtescher:
@MikeGoldsmith yeah good to show both, the docs for global module show both usages and I have the basic stdout example showing the spec derived version now 👍

PR #96, opened 2020-04-11
Add Resource API

jtescher:
Allow providers to be configured with information about the entity for which telemetry is recorded. See the [Resources Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/bdf39cb935a6f48badc687daeb78a66dc352761f/specification/overview.md#resources) for more information.

### DIFF
--- a/src/sdk/trace/config.rs
+++ b/src/sdk/trace/config.rs
@@ -17,6 +17,8 @@ pub struct Config {
     pub max_attributes_per_span: u32,
     /// The max links that can be added to a `Span`.
     pub max_links_per_span: u32,
+    /// Contains attributes representing an entity that produces telemetry.
+    pub resource: sdk::Resource,
 }
 
 impl Default for Config {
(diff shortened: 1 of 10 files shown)


src/sdk/trace/config.rs
```diff
@@ -17,6 +17,8 @@ pub struct Config {
     pub max_attributes_per_span: u32,
     /// The max links that can be added to a `Span`.
     pub max_links_per_span: u32,
+    /// Contains attributes representing an entity that produces telemetry.
+    pub resource: sdk::Resource,
```
jtescher, line 21: Might actually be better for this to be `Arc<sdk::Resource>`

jtescher: Could save a lot of cloning, especially if there are many attributes associated with the resource. But it would impact the `Serialize`/`Deserialize` impl as you can't deserialize an `Arc<T>`.

@MikeGoldsmith / @jodh-intel thoughts on this?

jtescher: Nvm, serde's `rc` feature actually works fine in this case.

PR #104, opened 2020-04-30
Fix Traceparent to traceparent

tensorchen:
https://w3c.github.io/trace-context/#header-name

Header name: traceparent

In order to increase interoperability across multiple protocols and encourage successful integration, by default vendors SHOULD keep the header name lowercase. The header name is a single word without any delimiters, for example, a hyphen (-).

Vendors MUST expect the header name in any case (upper, lower, mixed), and SHOULD send the header name in lowercase.


jtescher:
Thanks!

PR #116, opened 2020-05-19
Use ureq for collector instead of reqwest blocking client

glademiller:


jtescher:
great start! looks like there a a few lint issues and signing the CLA left

glademiller:
I don't know that I have the bandwidth right now to get this all the way
across the line so would love if someone would take this and run with it.
If not I'm sure at some indeterminate moment in the future I could try to
get it across the line.

LukeMathWalker:
I can pick this up :+1:

glademiller:
@LukeMathWalker thanks for running with this I'll close this PR

PR #120, opened 2020-05-21
Update rust CI targets

jtescher:
Add beta channel to test matrix and mark nightly as optional.


iredelmeier:
@jtescher unless it's changed since I last heard, only the technical committee are able to update check requirement :/

Thoughts on making the checks succeed, rather than skipping them, if any necessary components are unavailable?

jtescher:
That is what is intended in this change, the `continue-on-error` setup should allow nightly to fail without failing the whole workflow.

However the old checks from before are still being applied to this pr as well as the new ones and will now not run so they are stuck on expected?

iredelmeier:
Ah, I see what you mean. When I saw that the nightly checks were stuck for this PR, I figured that was because "continue-on-error" was getting misinterpreted as something other than `success`.

jtescher:
@iredelmeier the checks have been removed, want to review? 🙏

PR #122, opened 2020-05-28
Add example with tracing_opentelemetry and tonic

swilcox3:
This example shows a more advanced setup and integration with `tracing`.  I think this would be helpful to show a more realistic production environment.  
One concern I have is that the example Cargo.toml has to rely on the crates.io version of `opentelemetry`, because `tracing-opentelemetry` relies on that crates.io version as well.  Maybe this example should be in the `tracing-opentelemetry` repository, but I couldn't find that.


jtescher:
Thanks! Probably good idea to show a `tracing-opentelemetry` example in this repo. (The `tracing-opentelemetry` source is now in the `tracing` repo https://github.com/tokio-rs/tracing/tree/master/tracing-opentelemetry). But the example there would have a similar issue as  tonic relies on the crates.io release of tracing.

This looks good, still waiting on an admin to fix the builds or rust nightly to ship with rustfmt.

PR #128, opened 2020-06-02
Update sampler types

### LINKED ISSUE #113: Update sampler types
See https://github.com/open-telemetry/opentelemetry-specification/pull/609 for changes.

siyopao:
I tried working on this. I think I understood what needed to be done, but please let me know if I made a mistake. I didn't update the `tracing-grpc` example as it has some dependency conflicts with `tracing-opentelemetry`.

Resolves #113.

### DIFF
--- a/src/sdk/trace/sampler.rs
+++ b/src/sdk/trace/sampler.rs
@@ -8,8 +8,8 @@ pub enum Sampler {
     AlwaysOn,
     /// Never sample the trace
     AlwaysOff,
-    /// Sample if the parent span is sampled
-    ParentOrElse,
+    /// Respects the parent span's sampling decision or delegates a delegate sampler for root spans.
+    ParentOrElse(Box<Sampler>),
     /// Sample a given fraction of traces. Fractions >= 1 will always sample. If the parent span is
     /// sampled, then it's child spans will automatically be sampled. Fractions < 0 are treated as
     /// zero, but spans may still be sampled if their parent is.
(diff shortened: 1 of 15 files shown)


src/sdk/trace/sampler.rs
```diff
@@ -5,11 +5,11 @@ use crate::api;
...
 pub enum Sampler {
     /// Always sample the trace
-    Always,
+    AlwaysOn,
     /// Never sample the trace
-    Never,
+    AlwaysOff,
     /// Sample if the parent span is sampled
-    Parent,
+    ParentOrElse,
```
jtescher, line 12: From looking at the spec this should be a composite sampler, so likely something like `ParentOrElse(Box<Sampler>)`, and the logic should reflect the inner sampler when there is no parent.

siyopao: Thank you for your comment. Sorry, yes, I reread the spec and I now think I am following it 😄  Please let me know if my changes are on the right track. Thanks!

src/sdk/trace/sampler.rs
```diff
@@ -28,17 +28,29 @@ impl api::Sampler for Sampler {
...
-                }
-            }
+            Sampler::AlwaysOff => api::SamplingDecision::NotRecord,
+            // The parent decision if sampled; otherwise the decision of delegate_sampler
+            Sampler::ParentOrElse(delegate_sampler) => parent_context.map_or(
+                delegate_sampler
+                    .should_sample(
+                        parent_context,
+                        trace_id,
+                        _name,
```
jtescher, line 40: can we remove these underscores now that the variables are being used?

siyopao: 😬 Fixed

PR #140, opened 2020-06-15
Do not add `span.kind` tag if it has been set as an attribute

### LINKED ISSUE #137: opentelemetry-jaeger: Do no add `span.kind` tag if it has been set as an attribute
Currently if the `span.kind` attribute is set, opentelemetry-jaeger will add a second one when it submits the jaeger data.

siyopao:
Here is my first attempt at resolving #137.  Should I also be checking for `status.code` and `status.message`?

For `into_zipkin_span()`, if I find `span.kind` in the attributes I leave it in there and set the `kind` field in the zipkin `Span` to be `None`. I don't know if I should have extracted it from the attribute list and set the `kind` field. It is always a bit scary to go from a string to an enum. Also in this function, as was done previously, I first convert to `api::KeyValue` and then to a `HashMap`. I assume this is for data validation as the key and value have to satisfy the `Into<Key>` and `Into<Value>` trait bounds.

Thank you.

Closes #137.

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -381,13 +381,13 @@ fn links_to_references(links: &sdk::EvictedQueue<api::Link>) -> Option<Vec<jaege
 }
 
 fn build_tags(span_data: &Arc<trace::SpanData>) -> Option<Vec<jaeger::Tag>> {
-    let mut user_specified = HashSet::new();
+    let mut user_overrides = UserOverrides::default();
     // TODO determine if namespacing is required to avoid collisions with set attributes
     let mut tags = span_data
         .attributes
         .iter()
         .map(|(k, v)| {
-            user_specified.insert(k.as_str());
+            user_overrides.record_attr(k.as_str());
             api::KeyValue::new(k.clone(), v.clone()).into()
         })
         .chain(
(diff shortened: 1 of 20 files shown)


opentelemetry-jaeger/src/lib.rs
```diff
@@ -382,11 +382,15 @@ fn links_to_references(links: &sdk::EvictedQueue<api::Link>) -> Option<Vec<jaege
 fn build_tags(span_data: &Arc<trace::SpanData>) -> Option<Vec<jaeger::Tag>> {
     let mut tags = Vec::with_capacity(span_data.attributes.len() + span_data.resource.len() + 4);
     let mut user_specified_error = false;
+    let mut user_specified_span_kind = false;
```
jtescher, line 385: Actually looking over this it may make sense to pull these bools out into a struct as we should probably do the same check for status code and status message.

opentelemetry-jaeger/src/lib.rs
```diff
@@ -382,11 +382,15 @@ fn links_to_references(links: &sdk::EvictedQueue<api::Link>) -> Option<Vec<jaege
...
     for (key, value) in span_data.attributes.iter() {
         tags.push(api::KeyValue::new(key.clone(), value.clone()).into());
         if key == &api::Key::new("error") {
             user_specified_error = true;
         }
+        if key == &api::Key::new("span.kind") {
```
jtescher, line 391: Then instead of checking many times here, the struct could have a method that accepts the key's value as a str

opentelemetry-jaeger/src/lib.rs
siyopao: Does it make sense to store the keys in a `HashSet` as `str`s then before inserting checking whether the key is in the set?

opentelemetry-jaeger/src/lib.rs
siyopao: If we do it this way does this mean creating a struct containing as fields the string values of the keys we want with default value `false`, then using the method to set the value `true` if the str matches one of our keys?

opentelemetry-jaeger/src/lib.rs
siyopao: I tried something. Please let me know if it is similar to what you had in mind. Also `cargo fmt` ended up formatting some other files so I just pushed those changes in to. Thank you.

jtescher:
@siyopao great start! Do you think it would be more effecient to store the fixed set of bools for keys like `span.kind` rather than storing all attr keys in a set? What do you think of something like:

```rust
const SPAN_KIND: &str = "span.kind";

struct UserOverrides {
    span_kind: bool,
    ...
}

impl UserOverrides {
    fn record_attr(&mut self, attr: &str) {
        match attr {
            SPAN_KIND => self.span_kind = true,
            ...
            _ => (),
        }
    }
}
```

opentelemetry-jaeger/src/lib.rs
siyopao: Now that I see your explanation this makes so much more sense. Sorry about that. I think I was misunderstanding yesterday. I have made the requested change. Thank you.

opentelemetry-jaeger/src/lib.rs
```diff
@@ -380,40 +380,86 @@ fn links_to_references(links: &sdk::EvictedQueue<api::Link>) -> Option<Vec<jaege
...
+const STATUS_MESSAGE: &str = "status.message";
+
+struct UserOverrides {
+    error: bool,
+    span_kind: bool,
+    status_code: bool,
+    status_message: bool,
+}
+
+impl Default for UserOverrides {
```
jtescher, line 440: I think this is the same as `#[derive(Default)]` for all false?

siyopao: Good point. Thank you. Fixed.

opentelemetry-jaeger/src/thrift/agent.rs
```diff
@@ -8,7 +8,6 @@
```
jtescher, line 8: Hm this file should have rustfmt skipped via `#![cfg_attr(rustfmt, rustfmt_skip)]`, can probably remove this change as this is all generated code from thrift. Harder to diff if the thrift compiler output changes from version to version if we change their formatting.

jtescher:
Just one last little change, looking good!

opentelemetry-jaeger/src/thrift/agent.rs
siyopao: Sure. Reverted. I think if you want to get `rustfmt` to skip a file you can add `#[rustfmt::skip]` in `lib.rs` where the module is declared. I found the answer on [stackoverflow](https://stackoverflow.com/questions/59247458/is-there-a-stable-way-to-tell-rustfmt-to-skip-an-entire-file#:~:text=1%20Answer&text=Currently%20Rustfmt%20traverses%20the%20mod,dont_format_this_file.rs%20will%20be%20skipped.) and tried it out. It seems to work. Although it does say it may not work with rustfmt 2...

PR #141, opened 2020-06-17
Implement full metrics API

### LINKED ISSUE #5: Split Prometheus exporter into its own crate

### LINKED ISSUE #17: Add stdout metric exporter
A stdout exporter (or something that can exporter to a [`Write`r](https://doc.rust-lang.org/std/io/trait.Write.html) can be really valuable, especially for dev purposes.

The [go stdout metric exporter](https://github.com/open-telemetry/opentelemetry-go/tree/e6d725626d4629220a2de0112570adf80d50be21/exporter/metric/stdout) might be useful inspiration.

### LINKED ISSUE #22: Metric option configuration
Current `MetricOptions` implementation does not work for [with_monotonic](https://github.com/open-telemetry/opentelemetry-rust/blob/master/src/api/metrics/mod.rs#L137). The issue is the options do not know which instrument they are being created for.

We could consider moving to returning builders instead of accepting all options here?

### LINKED ISSUE #24: Support named meters using global module
Named meters should be available to different parts of the application.

See also https://github.com/open-telemetry/opentelemetry-go/pull/258

### LINKED ISSUE #135: Not possible to instantiate metrics via api::Meter
Hi!

I started looking at this crate to provide an abstraction over metrics backends in a program. The idea is that a piece of code would e.g increment a counter via an interface rather than, say, a Prometheus counter directly.

It doesn't seem it will be possible to accomplish this with this crate's metrics interface.

My assumption is that this is one of the goals of this crate - this might be a wrong assumption, in which case these issues might not be valid to begin with.

The first issue seems to be the `Meter` trait's use of self-referencing associated types which rust (as of v1.43.0) doesn't handle very well.

As an example, I might want code that talks via the `Meter` interface to the sdk implementation

```rust
    let meter = opentelemetry::sdk::Meter::new("meter1");

    let _: Box<dyn opentelemetry::api::Meter<
        LabelSet=_,
        I64Counter=_,
        F64Counter=_,
        I64Gauge=_,
        F64Gauge=_,
        I64Measure=_,
        F64Measure=_
    >> = Box::new(meter); // Does not compile.
```

This fails with:
```
 type I64Counter: Counter<i64, Self::LabelSet>;
    |                        ---------------------------- the trait cannot be made into an object because it uses `Self` as a type parameter in this
```
The second, related issue is that I won't be able to say, write a function, which at runtime, chooses and returns (say based on config) an implementation of the `Meter` trait since rust doesn't support abstracting over trait implementations that have different associated types.

For example:
```rust
  // No such values would be suitable for `???` since they differ for Meter implementations
  fn get_arbitrary_meter() -> impl Meter<???> {
    if true {
      opentelemetry::sdk::Meter::new("meter1")
    } else {
      someOtherMeterImpl{}
    }
  }
```

### LINKED ISSUE #136: sdk::Meter crashes the program if not configured properly
Hi!

Here are two cases where the sdk panics

1. using duplicate keys when creating an instrument e.g 
```rust
    let meter = sdk::Meter::new("meter1");
    let foo_key = api::Key::new("fookey");
    let bar_key = api::Key::new("fookey");
    let counter = meter.new_f64_counter("c",
      api::MetricOptions::default().with_keys(vec![foo_key.clone(), bar_key.clone()])); // panics

```

2. providing labels to `acquire_handle` that don't match those passed to the instrument e.g
```rust
    let meter = sdk::Meter::new("meter1");
    let foo_key = api::Key::new("fookey");
    let bar_key = api::Key::new("barkey");
    let counter = meter.new_f64_counter("c",
      api::MetricOptions::default().with_keys(vec![foo_key.clone(), bar_key.clone()]));

    let labels = meter.labels(vec![foo_key.string("foovalue")]);
    let _ = counter.acquire_handle(&labels); // panics

    // thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InconsistentCardinality(2, 1)'
```

In both cases, it would be friendlier with an api that makes it explicit that these operations can fail and leave the error handling up to the caller - For example `1` could return a `Result` instead while `2` an `Option` since its analogous to fetching something (a handle) that may or maynot exist.
The current behavior makes an implicit decision hidden from the user and the error message from `2` about `InconsistentCardinality` makes the problem cryptic to the user on failure.

### LINKED ISSUE #171: Add global error handler
See the spec change [here](https://github.com/open-telemetry/opentelemetry-specification/pull/757)

jtescher:
This updates the metrics API to the [latest specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md).

Closes #5, closes #17, closes #22, closes #24, fixes #135, fixes #136, resolves #171


jtescher:
This is a large PR as it fully replaces the current (non-compliant) metrics implementation with a new compliant version. (It is very similar to the go metrics impl, which should help reviewing) cc @MikeGoldsmith, @TommyCpp

src/exporter/metrics/stdout.rs
```diff
@@ -0,0 +1,401 @@
...
+    do_not_print_time: bool,
+    /// Quantiles are the desired aggregation quantiles for distribution summaries,
+    /// used when the configured aggregator supports quantiles.
+    ///
+    /// Note: this exporter is meant as a demonstration; a real exporter may wish to
+    /// configure quantiles on a per-metric basis.
+    quantiles: Vec<f64>,
+    /// Encodes the labels.
+    label_encoder: Box<dyn Encoder + Send + Sync>,
+    /// An optional user-defined funtion to format a given export batch.
```
TommyCpp, line 60:
```suggestion
    /// An optional user-defined function to format a given export batch.
```

src/sdk/metrics/mod.rs
```diff
@@ -1,210 +1,564 @@
...
+#[derive(Debug)]
+struct AccumulatorCore {
+    /// A concurrent map of current sync instrument state.
+    current: dashmap::DashMap<MapKey, Arc<Record>>,
+    /// A collection of async instrument state
+    async_instruments: Mutex<AsyncInstrumentState>,
+
+    /// The current epoch number. It is incremented in `collect`.
+    current_epoch: AtomicNumber,
+    /// THe configured processor.
```
TommyCpp, line 122:
```suggestion
    /// The configured processor.
```

src/api/metrics/number.rs
```diff
@@ -0,0 +1,227 @@
...
+impl From<u64> for Number {
+    fn from(u: u64) -> Self {
+        Number(u)
+    }
+}
+
+/// A descriptor for the encoded data type of a `Number`
+#[derive(Clone, Debug, PartialEq, Hash)]
+pub enum NumberKind {
+    /// An Number that stores `i64` values.
```
TommyCpp, line 182:
```suggestion
    /// A Number that stores `i64` values.
```

src/api/metrics/number.rs
```diff
@@ -0,0 +1,227 @@
...
+    I64,
+    /// An Number that stores `f64` values.
```
TommyCpp, line 184:
```suggestion
    /// A Number that stores `f64` values.
```

src/api/metrics/number.rs
```diff
@@ -0,0 +1,227 @@
...
+    F64,
+    /// An Number that stores `u64` values.
```
TommyCpp, line 186:
```suggestion
    /// A Number that stores `u64` values.
```

opentelemetry-prometheus/src/sanitize.rs
```diff
@@ -0,0 +1,60 @@
+/// sanitize returns a string that is truncated to 100 characters if it's too
+/// long, and replaces non-alphanumeric characters to underscores.
+pub(crate) fn sanitize<T: Into<String>>(s: T) -> String {
```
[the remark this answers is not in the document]
jtescher: Hm you're right this probably should actually truncate.

src/sdk/metrics/controllers/pull.rs
```diff
@@ -0,0 +1,146 @@
...
+impl PullController {
+    /// The provider for this controller
+    pub fn provider(&self) -> registry::RegistryMeterProvider {
+        self.provider.clone()
+    }
+
+    /// Collects all metrics if the last collected at time is past the current period
+    pub fn collect(&self) -> Result<()> {
+        if self
+            .last_collect
```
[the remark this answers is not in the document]
jtescher: Yeah you're right, this needs to take `&mut self` too in order to update. Good catch.

jtescher:
@TommyCpp thanks for the review, if you want to become an approver ([requirements](https://github.com/open-telemetry/community/blob/master/community-membership.md#requirements)) I'm happy to sponsor you, just have [open an issue](https://github.com/open-telemetry/community/issues/new?template=membership.md&title=REQUEST%3A%20New%20membership%20for%20%3Cyour-GH-handle%3E) here to get added.

TommyCpp:
> @TommyCpp thanks for the review, if you want to become an approver ([requirements](https://github.com/open-telemetry/community/blob/master/community-membership.md#requirements)) I'm happy to sponsor you, just have [open an issue](https://github.com/open-telemetry/community/issues/new?template=membership.md&title=REQUEST%3A%20New%20membership%20for%20%3Cyour-GH-handle%3E) here to get added.

That would be awesome! Thanks! But it seems I need to 2 sponsors before opening the PR. Any suggestions on that?

jtescher:
@TommyCpp this project could probably use a few more approvers, I'm sure @MikeGoldsmith or @iredelmeier would sponsor you as well 👍

TommyCpp:
Sorry I didn't see the review request yeasterday. Great work! @jtescher

PR #146, opened 2020-07-07
Allow array values for attributes

### LINKED ISSUE #142: Allow array values for attributes
Add array variant to `Value` enum, see [the spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/368) for details.

siyopao:
The spec says that the string value of the array should look like a JSON string, so I attempted that here. And, as I mentioned earlier, this will allow for non-homogeneous arrays so some there is something to do there. Please let me know if any other changes required. Thank you!

Closes #142.

### DIFF
--- a/src/api/correlation/propagation.rs
+++ b/src/api/correlation/propagation.rs
@@ -200,7 +200,7 @@ mod tests {
             (vec![KeyValue::new("key1", "val1"), KeyValue::new("key2", "val2")], vec!["key1=val1", "key2=val2"]),
 	    // "two values with escaped chars"
             (vec![KeyValue::new("key1", "val1,val2"), KeyValue::new("key2", "val3=4")], vec!["key1=val1%2Cval2", "key2=val3%3D4"]),
-	    // "values of non-string types"
+	    // "values of non-string non-array types"
             (
                 vec![
                     KeyValue::new("key1", true),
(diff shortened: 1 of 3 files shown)


siyopao:
Ah, I may need some help with Jaeger. I am not quite sure how to edit those files.

jtescher:
You are correct Jaeger does not support arrays. Looks like the go implementation simply ignores arrays when exporting to jaeger https://github.com/open-telemetry/opentelemetry-go/blob/master/exporters/trace/jaeger/jaeger.go#L281-L328

jtescher:
@siyopao maybe just serialize as Jaeger string tag for now?

siyopao:
Hi @jtescher. Sure, let me try that. I need to figure out how Thrift works, but will try soon. Thanks.

jtescher:
@siyopao Sounds good. Can look at https://github.com/open-telemetry/opentelemetry-rust/blob/master/opentelemetry-jaeger/src/lib.rs#L296-L304 to see how string values are converted currently.

siyopao:
Thank you for the pointer @jtescher.

src/api/correlation/propagation.rs
```diff
@@ -215,6 +215,19 @@ mod tests {
...
+        // "values of array types"
+            (
+                vec![
+                    KeyValue::new("key1", Value::Array(vec![Value::Bool(true), Value::Bool(false)])),
+                    KeyValue::new("key2", Value::Array(vec![Value::String("val1".to_string()), Value::String("val2".to_string())])),
+                    KeyValue::new("key3", Value::Array(vec![Value::I64(123), Value::I64(456)])),
+                ],
+                vec![
+                        "key1=[true%2Cfalse]",
+                        "key2=[val1%2Cval2]",
```
jtescher, line 227: Think the spec wants these to be json encoded values, so strings should probably be quoted here even though it complicates things a little.

jtescher:
@siyopao looks great, just one last tweak and we should be good!

src/api/correlation/propagation.rs
siyopao: Oops, sorry about that. Fixed.

PR #148, opened 2020-07-16
[b3_propagator] Follow the b3 spec

### LINKED ISSUE #147: Make B3 propagator fully support spec
The existing implementation of the B3 propagator does not fully implement the HTTP B3 specification. 

Full B3 Specification: https://github.com/openzipkin/b3-propagation

See similar PR in go impl for needed changes: https://github.com/open-telemetry/opentelemetry-go/pull/882

TommyCpp:
Update B3 propagator to follow [spec](https://github.com/openzipkin/b3-propagation). Also added some test cases.

Detail changes includes:
- Add `B3Encoding` enum as parameter of `B3Propagator`.
- Add `TRACE_FLAG_NOT_SAMPLED `, `TRACE_FLAG_SAMPLED`, `TRACE_FLAG_DEFERRED`, `TRACE_FLAG_DEBUG` flag in span context.
- Change the keys in single header and multiple headers.
- Allow only certain length of lower case string as value of `span id` and `trace id` in b3 headers
- Support send sample decision along using b3 headers.

One difference between Rust and Go implementation:

As per spec, when the sampled decision is debug, it implies sampled. So When `X-B3-Flag` is 1, we will set span context to be `sampled` and `debug`. But in Go implementation, we set the span context to be `debug` and `deferred`.  The spec isn't really clear about it. On the one hand, it says `debug` implies sampled. On the other hand, it says if `X-B3-Flag` is 1, then no `X-B3-Sampled` should be sent and when `X-B3-Sampled` is not present, it means the sample decision is deferred. 

close #147

### DIFF
--- a/src/api/trace/b3_propagator.rs
+++ b/src/api/trace/b3_propagator.rs
@@ -3,7 +3,7 @@
 //! The `B3Propagator` facilitates `SpanContext` propagation using
 //! B3 Headers. This propagator supports both version of B3 headers,
 //!  1. Single Header:
-//!    X-B3: {trace_id}-{span_id}-{sampling_state}-{parent_span_id}
+//!    b3: {trace_id}-{span_id}-{sampling_state}-{parent_span_id}
 //!  2. Multiple Headers:
 //!    X-B3-TraceId: {trace_id}
 //!    X-B3-ParentSpanId: {parent_span_id}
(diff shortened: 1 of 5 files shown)


src/api/trace/b3_propagator.rs
```diff
@@ -3,62 +3,114 @@
...
 pub struct B3Propagator {
-    single_header: bool,
+    inject_encoding: B3Encoding,
 }
 
 impl B3Propagator {
     /// Create a new `HttpB3Propagator`.
-    pub fn new(single_header: bool) -> Self {
-        B3Propagator { single_header }
+    pub fn new(encoding: B3Encoding) -> Self {
```
jtescher, line 62: Think we should simplify this? we could impl `Default` and have `new` accept no parameters and default to `B3MultipleHeader`? could then have a `B3Propagator::with_encoding` that accepts an encoding to specify.

src/api/trace/b3_propagator.rs
```diff
@@ -3,62 +3,114 @@
...
+static B3_TRACE_ID_HEADER: &str = "x-b3-traceid";
+static B3_SPAN_ID_HEADER: &str = "x-b3-spanid";
+static B3_SAMPLED_HEADER: &str = "x-b3-sampled";
+static B3_PARENT_SPAN_ID_HEADER: &str = "x-b3-parentspanid";
+
+/// B3Encoding is a bitmask to represent B3 encoding type
+#[derive(Clone, Debug, Copy)]
+pub enum B3Encoding {
+    /// B3Unspecified is an unspecified B3 encoding
+    B3Unspecified = 0,
```
jtescher, line 34: Probably don't need the double `B3` prefixes for these. E.g. this one could just be `B3Encoding::Unspecified`.

src/api/trace/b3_propagator.rs
```diff
@@ -3,62 +3,114 @@
...
 
-static B3_SINGLE_HEADER: &str = "X-B3";
-static B3_DEBUG_FLAG_HEADER: &str = "X-B3-Flags";
-static B3_TRACE_ID_HEADER: &str = "X-B3-TraceId";
-static B3_SPAN_ID_HEADER: &str = "X-B3-SpanId";
-static B3_SAMPLED_HEADER: &str = "X-B3-Sampled";
-static B3_PARENT_SPAN_ID_HEADER: &str = "X-B3-ParentSpanId";
+impl B3Encoding {
+    /// support determines if current encoding supports the `e`
+    pub fn support(&self, e: B3Encoding) -> bool {
```
jtescher, line 49: small nit: could have e be `other: &Self` here to signal that it is comparing same types? Not necessary but seems more standard sometimes for rust APIs.

src/api/trace/b3_propagator.rs
```diff
@@ -3,62 +3,114 @@
...
+/// we leave it to be lower case since we cannot tell what kind of protocol will be used.
+/// Go implementation also uses lower case.
+static B3_DEBUG_FLAG_HEADER: &str = "x-b3-flags";
+static B3_TRACE_ID_HEADER: &str = "x-b3-traceid";
+static B3_SPAN_ID_HEADER: &str = "x-b3-spanid";
+static B3_SAMPLED_HEADER: &str = "x-b3-sampled";
+static B3_PARENT_SPAN_ID_HEADER: &str = "x-b3-parentspanid";
+
+/// B3Encoding is a bitmask to represent B3 encoding type
+#[derive(Clone, Debug, Copy)]
```
jtescher, line 31: Small nit: we have not been marking enums as `Copy` as it then becomes a breaking change to remove it later.

src/api/trace/b3_propagator.rs
```diff
@@ -3,62 +3,128 @@
...
+//!    b3: {trace_id}-{span_id}-{sampling_state}-{parent_span_id}
 //!  2. Multiple Headers:
 //!    X-B3-TraceId: {trace_id}
 //!    X-B3-ParentSpanId: {parent_span_id}
 //!    X-B3-SpanId: {span_id}
 //!    X-B3-Sampled: {sampling_state}
 //!    X-B3-Flags: {debug_flag}
 //!
-//! If `single_header` is set to `true` then `X-B3` header is used to inject
+//! If `inject_encoding` is set to `B3Encoding::B3SingleHeader` then `b3` header is used to inject
```
jtescher, line 14:
```suggestion
//! If `inject_encoding` is set to `B3Encoding::SingleHeader` then `b3` header is used to inject
```

jtescher:
@TommyCpp looks good, just one last comment

PR #152, opened 2020-07-22
[span] Add record_exception method.

### LINKED ISSUE #151: Add trace exception tracking method
A `record_exception` method should be added to the [Span API](https://github.com/open-telemetry/opentelemetry-rust/blob/master/src/api/trace/span.rs#L25-L125). See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/697) for details.

TommyCpp:
See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md for details.

closes #151

### DIFF
--- a/src/api/trace/span.rs
+++ b/src/api/trace/span.rs
@@ -47,10 +47,11 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
     /// The semantic conventions for Errors are described in ["Semantic Conventions for Exceptions"](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md)
     ///
     /// For now we will not set `exception.stacktrace` attribute since the `Error::backtrace`
-    /// method is still in nightly. User can provide stacktrace by using `record_exception_with_stacktrace`
-    /// method.
+    /// method is still in nightly. Users can provide a stacktrace by using the
+    /// `record_exception_with_stacktrace` method.
     ///
-    /// User can custom exception message by overriding `fmt::Display` trait's `fmt` method for error.
+    /// Users can custom the exception message by overriding the `fmt::Display` trait's `fmt` method
+    /// for the error.
     fn record_exception(&self, err: &dyn Error) {
         let attributes = vec![api::KeyValue::new(
             "exception.message",
@@ -64,12 +65,10 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
     ///
     /// See `Span:record_exception` method for more details.
     fn record_exception_with_stacktrace(&self, err: &dyn Error, stacktrace: String) {
-        let mut attributes = vec![api::KeyValue::new(
+        let attributes = vec![api::KeyValue::new(
             "exception.message",
             err.to_string(),
-        )];
-
-        attributes.push(api::KeyValue::new("exception.stacktrace", stacktrace));
+        ), api::KeyValue::new("exception.stacktrace", stacktrace)];
 
         self.add_event("exception".to_string(), attributes);
     }
--- a/src/api/trace/span.rs
+++ b/src/api/trace/span.rs
@@ -18,6 +18,7 @@
 use crate::api;
 #[cfg(feature = "serialize")]
 use serde::{Deserialize, Serialize};
+use std::error::Error;
 use std::fmt;
 use std::time::SystemTime;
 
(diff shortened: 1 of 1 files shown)


src/api/trace/span.rs
```diff
@@ -38,6 +39,42 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
...
+    fn record_exception(
+        &self,
+        err: &dyn Error,
+        error_type: Option<String>,
+        stacktrace: Option<String>,
+    ) {
+        // build up attributes
+        let mut attributes = vec![api::KeyValue::new(
+            "exception.message",
+            format!("{:?}", err),
```
jtescher, line 61: I think the `Display` output of `Error` may be what people expect here rather than the `Debug` impl. Could just go with `err.to_string()`?

src/api/trace/span.rs
```diff
@@ -38,6 +39,42 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
...
+    ///
+    /// The semantic conventions for Errors are described in ["Semantic Conventions for Exceptions"](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md)
+    ///
+    /// If user doesn't provide `error_type` or `stacktrace`, the corresponding attributes will not be set.
+    ///
+    /// User can custom exception message by overriding error's `fmt` method.
+    fn record_exception(
+        &self,
+        err: &dyn Error,
+        error_type: Option<String>,
```
jtescher, line 55: The error type name is often difficult for end users to produce, and the backtrace could eventually be pulled from  the err directly, maybe the initially we should just accept `err: &dyn Error,`? if the nightly feature does not stablize we could always add a `record_exception_with_backtrace` method later if people want the context.

src/api/trace/span.rs
```diff
@@ -38,6 +39,42 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
...
+        stacktrace: Option<String>,
+    ) {
+        // build up attributes
+        let mut attributes = vec![api::KeyValue::new(
+            "exception.message",
+            format!("{:?}", err),
+        )];
+
+        // if user provide stacktrace, then we set it.
+        // In the future, we can use `Error::stacktrace()` method if it's stabilized.
```
jtescher, line 65: Probably want `Error::backtrace` in this comment.

src/api/trace/span.rs
TommyCpp: Yeah `Display` makes more sense.

src/api/trace/span.rs
TommyCpp: OK. Just thought add it since the spec provide this attribute. But it's kind of hard to get the type of error in runtime for Rust.

src/api/trace/span.rs
TommyCpp: Sorry kind of a typo.

TommyCpp:
Thanks for the help! @jtescher

jtescher:
Hm it seems like it would be nice to capture the backtrace from just the `err` if we are on a version of rust that supports it, but the hacks required to determine that at build time are not pretty. e.g. [the anyhow crate workaround](https://github.com/dtolnay/anyhow/blob/2aa0cc1aba82a1ea513ba9e94d0077fd4d62479d/build.rs).

Another option is to use the [backtrace crate](https://docs.rs/backtrace/0.3.50/backtrace/struct.Backtrace.html) to capture the backtrace, but it would be a breaking change to remove it once support lands in std which seems not ideal.

The simplest solution is to just not support backtraces yet (update the PR to `record_exception(&self, err: &dyn Error)`) and wait for the feature to stablize, maybe just have a comment that the user can add the event manually if they want to include backtrace information.

TommyCpp:
> Hm it seems like it would be nice to capture the backtrace from just the `err` if we are on a version of rust that supports it, but the hacks required to determine that at build time are not pretty. e.g. [the anyhow crate workaround](https://github.com/dtolnay/anyhow/blob/2aa0cc1aba82a1ea513ba9e94d0077fd4d62479d/build.rs).
> 
> Another option is to use the [backtrace crate](https://docs.rs/backtrace/0.3.50/backtrace/struct.Backtrace.html) to capture the backtrace, but it would be a breaking change to remove it once support lands in std which seems not ideal.
> 
> The simplest solution is to just not support backtraces yet (update the PR to `record_exception(&self, err: &dyn Error)`) and wait for the feature to stablize, maybe just have a comment that the user can add the event manually if they want to include backtrace information.

Or maybe we provide a `record_exception_with_backtrace(&self, err: &dyn Error,  backtrace:Option<String>)`  for user who want to manually inject backtrace.

I would image if user's application is compiled in `nightly`. It would be fairly easily to extract backtrace from Err and inject it into api.

jtescher:
We could offer both methods. It may still be useful even once the feature stabilizes as a way of overriding the backtrace.

jtescher:
Maybe `record_exception(&self, err: &dyn Error)` and `record_exception_with_backtrace(&self, err: &dyn Error, backtrace: String)` (doesn't need to be optional string as you only need to use this method if you are overriding the backtrace)

src/api/trace/span.rs
```diff
@@ -38,6 +39,41 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
...
 
+    /// Convenience method to record an exception/error as an `Event`
+    ///
+    /// An exception SHOULD be recorded as an Event on the span during which it occurred.
+    /// The name of the event MUST be "exception".
+    ///
+    /// The semantic conventions for Errors are described in ["Semantic Conventions for Exceptions"](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md)
+    ///
+    /// For now we will not set `exception.stacktrace` attribute since the `Error::backtrace`
+    /// method is still in nightly. User can provide stacktrace by using `record_exception_with_stacktrace`
```
jtescher, line 50:
```suggestion
    /// method is still in nightly. Users can provide a stacktrace by using the `record_exception_with_stacktrace`
```

src/api/trace/span.rs
```diff
@@ -38,6 +39,41 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
...
+    /// method.
+    ///
+    /// User can custom exception message by overriding `fmt::Display` trait's `fmt` method for error.
```
jtescher, line 53:
```suggestion
    /// Users can customize the exception message by overriding the `fmt::Display` trait's `fmt` method for the error.
```

src/api/trace/span.rs
```diff
@@ -38,6 +39,41 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
...
+    /// Convenience method to record a exception/error as an `Event` with custom stacktrace
+    ///
+    /// See `Span:record_exception` method for more details.
+    fn record_exception_with_stacktrace(&self, err: &dyn Error, stacktrace: String) {
+        let mut attributes = vec![api::KeyValue::new(
+            "exception.message",
+            err.to_string(),
+        )];
+
+        attributes.push(api::KeyValue::new("exception.stacktrace", stacktrace));
```
jtescher, line 72: This should be included in the `vec!` macro call. Otherwise the attributes vec will have an initial capacity of one, and the call here to `push` will have to extend the capacity up to 4. If you include both `vec![k1, k2]`, then it will be a single heap allocation and the vec's capacity will be 2.

TommyCpp: Sorry I should have considered that.

PR #155, opened 2020-07-24
fix(zipkin): add with_collector_endpoint into config builder

bnjjj:
Missing method for config builder in zipkin


jtescher:
@bnjjj looks like the CLA bot came back, can merge once that is signed 👍

bnjjj:
Done

PR #164, opened 2020-08-01
Separate Carrier into Injector and Extractor

### LINKED ISSUE #105: Separate `Carrier` into `Injector` and `Extractor`
See https://github.com/open-telemetry/opentelemetry-specification/pull/578 for spec changes

TommyCpp:
Although it feels like we should separate `HttpTextFormat` instead of `Carrier`
close #105


rodrigopandini:
Releated: #177

PR #174, opened 2020-08-12
Add resource detector

### LINKED ISSUE #167: Add resource detector
Add resource detectors to configure resources based on common environments and environmental variables to make it easier to include the [resource semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions).

TommyCpp:
Fix #167 
See https://github.com/open-telemetry/oteps/pull/111 for proposal
Also fixes the bug in test mod of `Resource`


src/sdk/env.rs
```diff
@@ -0,0 +1,88 @@
+//! EnvResourceDetector
+//!
+//! Implementation of ResourceDetector in extract information from environment.
```
jtescher, line 3:
```suggestion
//! Implementation of `ResourceDetector` to extract a `Resource` from environment variables.
```

src/sdk/env.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::sdk::resource::ResourceDetector;
+use crate::sdk::Resource;
+use crate::api::{KeyValue, Key, Value};
+use std::env;
+use std::time::Duration;
+
+const OTEL_RESOURCE_ATTRIBUTES: &str = "OTEL_RESOURCE_ATTRIBUTES";
+
+/// Resource detector implements ResourceDetector and is used to extractor general SDK configuration from environment
+/// See https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#telemetry-sdk for semantic conventions.
```
jtescher, line 13: Can we wrap these to 80 characters? rustfmt currently doesn't do this automatically, but the style guide for comments in rust seems to be 80 currently https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md#comments

src/sdk/env.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Debug)]
+pub struct EnvResourceDetector {}
```
jtescher, line 15: we've been using `_private` here so we could add more without breaking changes (forces people to use the `new` method) e.g. https://github.com/open-telemetry/opentelemetry-rust/blob/e0bc3648110b38d338dfc84180b085185c1b517a/src/api/correlation/propagation.rs#L17-L19

src/sdk/env.rs
```diff
@@ -0,0 +1,88 @@
...
+impl Default for EnvResourceDetector {
+    fn default() -> Self {
+        EnvResourceDetector {}
+    }
+}
+
+/// Extract key value pairs and construct a resource from resources string like key1=value1,key2=value2,...
+fn construct_otel_resources(s: String) -> Resource {
+    let mut key_values = vec![];
+    for entries in s.split(',') {
```
jtescher, line 42: `split_terminator` here may be preferable as it will have the behavior people probably expect form trailing commas

src/sdk/resource.rs
```diff
@@ -17,10 +17,12 @@ use crate::api;
 #[cfg(feature = "serialize")]
 use serde::{Deserialize, Serialize};
 use std::collections::{btree_map, btree_map::Entry, BTreeMap};
+use crate::api::KeyValue;
+use std::time::Duration;
 
 /// Describes an entity about which identifying information and metadata is exposed.
 ///
-/// Items are sorted by key, and are only overwritten if the value is an empty string.
+/// Items are sorted by keys, and are only overwritten if the value is an empty string.
```
jtescher, line 25: Sorted by keys is also a little awkward, maybe "sorted by their key"?

src/sdk/env.rs
TommyCpp: Fixed.

src/sdk/env.rs
TommyCpp: Sure thing. Don't think I can break the url but will wrap other comments.

src/sdk/env.rs
TommyCpp: Good point!

src/sdk/env.rs
TommyCpp: Sounds reasonable. Thanks!

jtescher:
@TommyCpp I think you can merge your own PRs once approved now 👍

TommyCpp:
@jtescher I tried but I don't think I can. Maybe you will have to be maintainer to do that?

jtescher:
Hm maybe, the permissions in the otel org seem a little overly complicated to me. I'll just merge this for now 🤷‍♂️

PR #176, opened 2020-08-14
Update OTLP exporter metadata

jtescher:
Fix invalid crate metadata (max 5 keywords) and add placeholder README


jtescher:
@TommyCpp sure sounds good, can add.

PR #177, opened 2020-08-15
Update tracing-grpc example

rodrigopandini:
Refactor Carrier to use Extractor and Injector in latest versions for tracing-grpc example.


jtescher:
Hey @rodrigopandini thanks for updating this example. Couple of bits of context for options here for additional improvements:

1. There is now tonic support (currently for 0.2.x but just a version bump for 0.3.x) in the `opentelemetry` crate itself via the `tonic` feature.
2. This particular example relies on `tracing-opentelemetry` which requires a published version so it cannot point at the latest master version (so still stuck on 0.2 support until we update and release a new version).

With those in mind, it may be best to change the tonic version back to 0.2 in this example while updating all the other dependencies to their latest and include the `tonic` feature in the opentelemetry dependency. This will allow you to get rid of the injector and extractor impls entirely as they are now included.

If you want to, you could also update the tonic dependency to 0.3 in the core opentelemetry crate and update the `grpc` example to use this which would allow you to remove the injector and extractor from that example as  well. If not this can be done by someone as a follow up PR.

jtescher:
@rodrigopandini this can be merged once you sign the CLA (comment above for quick link to sign).

jtescher (approved):

jtescher:
Closing this in favor of #258

PR #178, opened 2020-08-16
[propagator] Add get_fields function for HttpTextFormat trait

TommyCpp:
See https://github.com/open-telemetry/opentelemetry-go/blob/45b6a22cbba2fb3015c50ad095ff61f51b81aece/api/propagation/propagation.go#L82 for go implementation.

See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#fields for spec.


src/api/context/propagation/text_propagator.rs
```diff
@@ -41,4 +41,9 @@ pub trait HttpTextFormat: Debug {
     /// [`Context`]: ../../struct.Context.html
     /// [`Extractor`]: ../trait.Extractor.html
     fn extract_with_context(&self, cx: &Context, extractor: &dyn api::Extractor) -> Context;
+
+    /// Returns list of fields used by [`HttpTextFormat`]
+    ///
+    /// [`HttpTextFormat`]: ./trait.HttpTextFormat.html
+    fn get_fields(&self) -> Vec<String>;
```
jtescher, line 48: It looks like propagator fields are generally static, I wonder if we could avoid the allocations here. Trickier with the composite propagator though. Ideally something like `[&'static str]` would be nice 🤔

TommyCpp: I actually thought about it. But since it's API, I thought we should allow user tweak it as much as possible. 

Like if user are trying to acquire the header name from some kind of configuration services. You can't really make it a `&'static str`

jtescher: Hm true. Although they would likely configure the propagator once when constructed right? could do `&[String]` maybe. That would work from static, e.g. `static FIELDS: [String; 1] = [TRACEPARENT_HEADER.to_string()];` and could work if composite propagator just stores `fields: Vec<String>`.

In that case could also make this an iterator that you could construct from `&[String]`. Something like:

```rust
/// An iterator over fields of a text propagator
#[derive(Debug)]
pub struct FieldIter<'a>(slice::Iter<'a, String>);

impl<'a> FieldIter<'a> {
    /// Create a new `FieldIter` from a slice of propagator fields
    pub fn new(fields: &'a [String]) -> Self {
        FieldIter(fields.iter())
    }
}

impl<'a> Iterator for FieldIter<'a> {
    type Item = &'a String;
    fn next(&mut self) -> Option<Self::Item> {
        self.0.next()
    }
}
```

TommyCpp: Yes `&[String]` sounds like a good idea. I would image you will have to put the header names somewhere in your propagator. I am gonna give it a try. Thanks for the advice!

src/api/context/propagation/text_propagator.rs
```diff
@@ -42,8 +43,28 @@ pub trait HttpTextFormat: Debug {
...
+        FieldIter(fields.iter())
+    }
+}
+
+impl<'a> Iterator for FieldIter<'a> {
+    type Item = &'a str;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next().map(|field| field.as_str())
+    }
+}
```
jtescher: if you run `cargo fmt` it should remove these line ending notices

src/api/correlation/propagation.rs
```diff
@@ -85,8 +88,8 @@ impl api::HttpTextFormat for CorrelationContextPropagator {
         }
     }
 
-    fn get_fields(&self) -> Vec<String> {
-        return vec![CORRELATION_CONTEXT_HEADER.into()]
+    fn get_field_iter(&self) -> FieldIter {
+        FieldIter::new(CORRELATION_CONTEXT_FIELDS.deref().as_ref())
```
jtescher, line 92: You should be able to get rid of the `Deref` calls here. Just `CORRELATION_CONTEXT_FIELDS.as_ref()` should work.

src/api/trace/b3_propagator.rs
```diff
@@ -28,6 +30,12 @@ static B3_SPAN_ID_HEADER: &str = "x-b3-spanid";
 static B3_SAMPLED_HEADER: &str = "x-b3-sampled";
 static B3_PARENT_SPAN_ID_HEADER: &str = "x-b3-parentspanid";
 
+lazy_static::lazy_static! {
+static ref B3_SINGLE_FIELDS: [String; 1] = [B3_SINGLE_HEADER.to_string()];
+static ref B3_MULTI_FIELDS: [String; 4] = [B3_TRACE_ID_HEADER.to_string(), B3_SPAN_ID_HEADER.to_string(), B3_SAMPLED_HEADER.to_string(), B3_DEBUG_FLAG_HEADER.to_string()];
+static ref B3_SINGLE_MULTI_FIELDS: [String; 5] = [B3_SINGLE_HEADER.to_string(), B3_TRACE_ID_HEADER.to_string(), B3_SPAN_ID_HEADER.to_string(), B3_SAMPLED_HEADER.to_string(), B3_DEBUG_FLAG_HEADER.to_string()];
```
jtescher, line 36: Maybe `B3_SINGLE_AND_MULTI_FIELDS`? or `B3_ALL_FIELDS`? SINGLE_MULTI just sounds a little contradictory.

jtescher: Also these static refs could be indented in

TommyCpp: Sounds reasonable. Will change those.

src/api/correlation/propagation.rs
TommyCpp: hmmm interesting. I tried `&CONRRELATION_CONTEXT_FIELDS` and it didn't work, but `as_ref()` works. I guess `as_ref()` is different.

jtescher: Yeah `as_ref` does additional conversion magic that `&` doesn't.

TommyCpp:
Gonna rebase it after being reviewed

PR #180, opened 2020-08-18
Semantic conventions crate

### LINKED ISSUE #162: Add semantic convention constants module
Add semantic conventions for [tracing](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions), [resources](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions), and [metrics](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/metrics/semantic_conventions).

frigus02:
This adds a new crate `opentelemetry-semantic-conventions`, which contains constants for the attributes defined in:

- https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions
- https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions

In order to add `const` `Key`s, this also adds a new function `from_static_str` to the `Key` implementation in the main crate.

Closes #162


frigus02:
For future reference, I used the following bash script plus a bit of manual editing to convert the spec markdown files to Rust code:

```sh
#!/bin/bash

md_to_rs() {
	while read line; do
		if [[ "$line" =~ ^\|\ \`?([a-z0-9._]+)\`?\ +\|\ ([^|]+)\ +\| ]]; then
			name="${BASH_REMATCH[1]}"
			const_name=$(echo "${name^^}" | sed "s/\\./_/g")
			doc="${BASH_REMATCH[2]}"
			echo ""
			echo "/// $doc"
			echo "pub const $const_name: Key = Key::from_static_str(\"$name\");"
		fi
	done <<<"$1"
}

md_url_to_rs() {
	md_to_rs "$(curl -sSf "$1")"
}

#md_url_to_rs "https://raw.githubusercontent.com/open-telemetry/opentelemetry-specification/master/specification/trace/semantic_conventions/span-general.md"
# repeat with all other files here

#md_url_to_rs "https://raw.githubusercontent.com/open-telemetry/opentelemetry-specification/master/specification/resource/semantic_conventions/README.md"
# repeat with all other files here
```

Given that it doesn't "just work", I'm not sure it's worth committing. But if anyone needs to update this in the future, they might want to use this to get started.

frigus02:
All attributes are there now, it's compiling and the documentation looks good to me. So I'm going to mark this ready for review.

I quickly scanned through the examples and exporters to see if there are any attributes that should now uses these constants but I didn't spot any. I'm not sure if we want to have another example using any of constants since there is nothing special about them. But happy to add that tomorrow if we think it would be useful.

opentelemetry-semantic-conventions/src/trace.rs
```diff
@@ -0,0 +1,339 @@
...
+
+/// String containing the execution id of the function. E.g.
+/// `af9d5aa4-a685-4c5f-a22b-444f80b3cc28`
+pub const FAAS_EXECUTION: Key = Key::from_static_str("faas.execution");
+
+/// A boolean indicating that the serverless function is executed for the first
+/// time (aka cold start).
+pub const FAAS_COLDSTART: Key = Key::from_static_str("faas.coldstart");
+
+/// The name of the source on which the operation was perfomed. For example, in
```
TommyCpp, line 168:
```suggestion
/// The name of the source on which the operation was performed. For example, in
```

opentelemetry-semantic-conventions/src/trace.rs
```diff
@@ -0,0 +1,339 @@
...
+pub const EXCEPTION_MESSAGE: Key = Key::from_static_str("exception.message");
+
+/// A stacktrace as a string in the natural representation for the language
+/// runtime. The representation is to be determined and documented by each
+/// language SIG. E.g. `"Exception in thread \"main\" java.lang.RuntimeException:
+/// Test exception\n at
+/// com.example.GenerateTrace.methodB(GenerateTrace.java:13)\n at
+/// com.example.GenerateTrace.methodA(GenerateTrace.java:9)\n at
+/// com.example.GenerateTrace.main(GenerateTrace.java:5)"`.
+pub const EXCEPTION_STACKTRACE: Key = Key::from_static_str("exception.stacktrace");
```
TommyCpp, line 153: Probably best to replace the example with Rust's stacktrace

opentelemetry-semantic-conventions/src/trace.rs
```diff
@@ -0,0 +1,339 @@
...
+/// time (aka cold start).
+pub const FAAS_COLDSTART: Key = Key::from_static_str("faas.coldstart");
+
+/// The name of the source on which the operation was perfomed. For example, in
+/// Cloud Storage or S3 corresponds to the bucket name, and in Cosmos DB to the
+/// database name.
+pub const FAAS_DOCUMENT_COLLECTION: Key = Key::from_static_str("faas.document.collection");
+
+/// Describes the type of the operation that was performed on the data.<br /> It
+/// SHOULD be one of the following strings: "insert", "edit", "delete".
```
TommyCpp, line 174:
```suggestion
/// Describes the type of the operation that was performed on the data. 
/// 
/// It SHOULD be one of the following strings: "insert", "edit", "delete".
```

opentelemetry-semantic-conventions/src/resource.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// A namespace for `service.name`.
+/// A string value having a meaning that helps to distinguish a group of
+/// services, for example the team name that owns a group of services.
+/// `service.name` is expected to be unique within the same namespace. The field
+/// is optional. If `service.namespace` is not specified in the Resource then
+/// `service.name` is expected to be unique for all services that have no
+/// explicit namespace defined (so the empty/unspecified namespace is simply one
+/// more valid namespace). Zero-length namespace string is assumed equal to
+/// unspecified namespace.
```
TommyCpp, line 42: This would result in a single line comment. Probably best to break it into multiple lines because of the length.

opentelemetry-semantic-conventions/src/resource.rs
```diff
@@ -0,0 +1,242 @@
...
+/// `service.namespace,service.name,service.id` triplet MUST be globally
+/// unique). The ID helps to distinguish instances of the same service that
+/// exist at the same time (e.g. instances of a horizontally scaled service). It
+/// is preferable for the ID to be persistent and stay the same for the lifetime
+/// of the service instance, however it is acceptable that the ID is ephemeral
+/// and changes during important lifetime events for the service (e.g. service
+/// restarts). If the service has no inherent unique ID that can be used as the
+/// value of this attribute it is recommended to generate a random Version 1 or
+/// Version 4 RFC 4122 UUID (services aiming for reproducible UUIDs may also use
+/// Version 5, see RFC 4122 for more recommendations).
```
TommyCpp, line 57: Same. Probably best to break it into multiple lines.

frigus02:
Thanks for a thorough review, @TommyCpp. I admit, I didn't actually read all the doc comments properly.

I think I addressed all your comments. I added some documentation to the generation script and added it to the `opentelemetry-semantic-conventions/` folder. Let me know if you can think of a better place.

opentelemetry-semantic-conventions/generate-consts-from-spec.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
jtescher, line 1: Maybe put this in a `scripts` folder? Could be done as a follow up pr as well.

PR #186, opened 2020-08-21
Add Jaeger propagator

TommyCpp:
See [specification change](https://github.com/open-telemetry/opentelemetry-specification/pull/770)


opentelemetry-contrib/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! # OpenTelemetry Contrib
+//!
+//! This is a library for extensions that is not part of core api. But still may be found useful by
+//! some users.
+//!
+//! Typically, those include vendor specific propagators
+
+pub mod trace_propagator;
```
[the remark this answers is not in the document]
jtescher: Yeah figure starting simple could be best, can always break out if it gets complex

opentelemetry-contrib/src/trace_propagator/b3_propagator.rs
```diff
@@ -80,6 +80,7 @@ impl B3Propagator {
     }
 
     /// Create a new `HttpB3Propagator` that uses `encoding` as encoding method
+    #[allow(unused)]
```
jtescher, line 83: shouldn't need this one either

TommyCpp: Sorry I missed it. will delete it.

PR #193, opened 2020-08-23
Ensure extractors are case insensitive

jtescher:


jtescher:
Yeah I think we can update the docs to be more in sync with the latest spec in a future PR as GA for tracing is close 👍

PR #199, opened 2020-08-27
Add end_with_timestamp method for trace span.

### LINKED ISSUE #14: End time should be configurable
Per the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/e6cd52d034c981ca4fcf896ccd7464e783589c0f/specification/api-tracing.md#end).

TommyCpp:
closes #14 

On a side note, maybe we could create a mod for test utils. We have `TestSpan` in four different files, which could be put into one place.

### DIFF
--- a/src/api/trace/span.rs
+++ b/src/api/trace/span.rs
@@ -152,8 +152,15 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
     /// Calls to `end` a Span MUST not have any effects on child `Span`s as they may
     /// still be running and can be ended later.
     ///
-    ///This API MUST be non-blocking.
+    /// This API MUST be non-blocking.
     fn end(&self);
+
+    /// Finishes the `Span` with given timestamp
+    ///
+    /// For more details, refer to [`Span::end`]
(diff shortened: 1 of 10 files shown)


src/api/trace/span.rs
```diff
@@ -152,8 +152,15 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
     /// Calls to `end` a Span MUST not have any effects on child `Span`s as they may
     /// still be running and can be ended later.
     ///
-    ///This API MUST be non-blocking.
+    /// This API MUST be non-blocking.
     fn end(&self);
```
jtescher, line 156: It may be useful to implement this method by default in terms of `end_with_timestamp` now to avoid requiring redundant impls for both methods.

jtescher:
@TommyCpp looks like the clippy rules got updated today so this is no longer passing 😐

TommyCpp:
> @TommyCpp looks like the clippy rules got updated today so this is no longer passing

That's interesting. I will update it later today.

PR #200, opened 2020-08-28
Add ID methods for hex and byte array formatting

### LINKED ISSUE #194: Add methods to trace/span id to format as hex and byte array
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/836) for details.

awiede:
Fixes #194 

Add methods to SpanId and TraceId to format as lower Hex and big endian byte array.

### DIFF
--- a/src/api/trace/span_context.rs
+++ b/src/api/trace/span_context.rs
@@ -56,6 +56,16 @@ impl TraceId {
     pub fn to_byte_array(self) -> [u8; 16] {
         self.0.to_be_bytes()
     }
+
+    /// Construct a new TraceId from Hexadecimal String
+    pub fn from_hex(hex: &str) -> Self {
+        TraceId(u128::from_str_radix(hex, 16).unwrap_or(0))
+    }
+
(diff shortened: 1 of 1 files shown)


awiede:
Wondering if I should add methods to build a `TraceId` and a `SpanId` from hex/byte array? The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#retrieving-the-traceid-and-spanid) only specifies retrieval methods, but I think it could be useful to add `from_*` methods.

jtescher:
@awiede I'm not against having from impls, do you have a use case in mind? seems like common usage may not be creating any ids generally

awiede:
@jtescher I came across this because I just implemented an AWS X-Ray propagator which had to parse a hex trace ID. It's not a hug deal to delegate the hex --> decimal conversion to the propagator code, but I'd assume that wouldn't be specific to just that vendor.

awiede:
@jtescher - Added here.

Issue #195, opened 2020-08-25
Add `keys` method to extractors 


jtescher:
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/825) for details.

siyopao:
I would like to attempt this if that is okay?

PR #202, opened 2020-08-30
AWS X-Ray Trace Context Propagator

awiede:
Create Trace Context Propagator for AWS X-Ray. Collect Root TraceId, Parent SegmentId and Sampling decision in new SpanContext. Will add to `/examples` later.

### DIFF
--- a/opentelemetry-contrib/src/trace_propagator/aws_xray_propagator.rs
+++ b/opentelemetry-contrib/src/trace_propagator/aws_xray_propagator.rs
@@ -1,0 +1,280 @@
+use opentelemetry::api::{
+    Context, Extractor, FieldIter, Injector, SpanContext, SpanId, TextMapFormat, TraceContextExt,
+    TraceId, TRACE_FLAG_DEFERRED, TRACE_FLAG_NOT_SAMPLED, TRACE_FLAG_SAMPLED,
+};
+use std::collections::HashMap;
+
+const AWS_XRAY_TRACE_HEADER: &str = "x-amzn-trace-id";
+const AWS_XRAY_VERSION_KEY: &str = "1";
+const HEADER_PARENT_KEY: &str = "Parent";
+const HEADER_ROOT_KEY: &str = "Root";
+const HEADER_SAMPLED_KEY: &str = "Sampled";
+
+const SAMPLED: &str = "1";
+const NOT_SAMPLED: &str = "0";
+const REQUESTED_SAMPLE_DECISION: &str = "?";
+
+lazy_static::lazy_static! {
+    static ref AWS_XRAY_HEADER_FIELD: [String; 1] = [AWS_XRAY_TRACE_HEADER.to_string()];
+}
+
+/// # AWS X-Ray Trace Propagator
+///
+/// Extracts and injects values to/from the `x-amzn-trace-id` header. Converting between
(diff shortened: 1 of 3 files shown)


opentelemetry-contrib/src/trace_propagator/aws_xray_propagator.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+impl TextMapFormat for XrayTraceContextPropagator {
+    fn inject_context(&self, cx: &Context, injector: &mut dyn Injector) {
+        let span_context: SpanContext = cx.span().span_context();
+        if span_context.is_valid() {
+            let xray_trace_id: XrayTraceId = span_context.trace_id().into();
+            let mut components: Vec<String> =
+                vec![format!("{}={}", HEADER_ROOT_KEY, xray_trace_id.0)];
```
TommyCpp, line 98: Could probably use `vec![...; 3]` to avoid possible reallocation

opentelemetry-contrib/src/trace_propagator/aws_xray_propagator.rs
```diff
@@ -0,0 +1,280 @@
...
+///
+/// [otel-spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spancontext
+/// [xray-trace-id]: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
+/// [xray-header]: https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader
+#[derive(Clone, Debug, Default)]
+pub struct XrayTraceContextPropagator {
+    _private: (),
+}
+
+impl XrayTraceContextPropagator {
```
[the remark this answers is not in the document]
awiede: ✔️

opentelemetry-contrib/src/trace_propagator/aws_xray_propagator.rs
awiede: So specifying `vec![...; 3]` put in three copies of the first element, refactored to use `Vec::with_capacity` instead.

awiede:
Thanks for the 👀  @TommyCpp! Made the suggested updates

opentelemetry-contrib/src/trace_propagator/aws_xray_propagator.rs
jtescher: As the formatted string is always the same, you could do one allocation `injector.set(AWS_XRAY_TRACE_HEADER, format!("{}={};{}={};{}={}"))` to do one string allocation rather than 4 strings and a vec?

awiede: Good call @jtescher, updated. Originally I had this iterating through all k/v pairs to handle arbitrary key/value pairs in the header. The [X-Ray SDK](https://github.com/aws/aws-xray-sdk-go/blob/master/header/header.go#L74) lets you pass any additional k/v pairs through, but there isn't a way to capture anything other than traceId, parent ID, and decision in the current `SpanContext` I don't think? Is that part of the thought with #195?

jtescher: Yep #195 should get support for that.

PR #207, opened 2020-09-02
Add InstrumentationLibrary for instrumentation library information

### LINKED ISSUE #204: Add instrumentation library version information
As in [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracerprovider-operations). We should provide an optional param to allow users to set instrumentation library version when we `get_tracer` from `TracerProvider/Provider`

TommyCpp:
fix #204 

As per spec, we should provide an optional param for version in `TraceProvider::get_tracer`. I think we should stick with spec and add second param here.

But it feels like most of the time people won't use it. So we maybe able to provide `global::tracer` and `global::tracer_with_version` methods so that users won't need to type `None` everytime when they retrive a tracer.
Not sure if it's desired, so I added second param into `global::tracer` for now.

### DIFF
--- a/benches/trace.rs
+++ b/benches/trace.rs
@@ -74,7 +74,7 @@ fn trace_benchmark_group<F: Fn(&sdk::Tracer)>(c: &mut Criterion, name: &str, f: 
                 ..Default::default()
             })
             .build()
-            .get_tracer("always-sample");
+            .get_tracer("always-sample", None);
 
         b.iter(|| f(&always_sample));
     });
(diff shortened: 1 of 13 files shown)


jtescher:
I think a second method maybe be more ergonomic, many people don't even like having to specify the first argument as brought up in https://github.com/open-telemetry/opentelemetry-specification/issues/586.

TommyCpp:
> I think a second method may be more ergonomic, many people don't even like having to specify the first argument as brought up in [open-telemetry/opentelemetry-specification#586](https://github.com/open-telemetry/opentelemetry-specification/issues/586).
 
Make sense. Will add another method in `global` mod. But do we want to add a second method in `TracerProvider` as well?

jtescher:
Hm seems less critical as people are less likely to use `TracerProvider` directly anymore. Maybe better to go with single method there to encourage library developers to specify their versions 👍

jtescher (approved):
Looks good

PR #209, opened 2020-09-03
Add keys method to extractors

### LINKED ISSUE #195: Add `keys` method to extractors

siyopao:
First attempt at #195. I could probably use some help with the documentation though 🙇  

Closes #195.

### DIFF
--- a/src/api/context/propagation/mod.rs
+++ b/src/api/context/propagation/mod.rs
@@ -163,6 +163,7 @@
 //!
 use crate::api;
 use std::collections::HashMap;
+use std::iter::Iterator;
 
 pub mod composite_propagator;
 pub mod text_propagator;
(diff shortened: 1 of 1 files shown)


jtescher:
@siyopao great start! What do you think about having `keys` return an iterator over `&str`s instead of a `Vec`? Could help people avoid the additional allocation.

siyopao:
Thanks for the comment @jtescher. Good idea! I've updated to return an iterator.

jtescher:
@siyopao Sorry I hadn't looked into this closely enough, I don't think we can have this be an iterator without additional allocations as the lifetimes for something like `&dyn Iterator<Item=&str>` can't be easily resolved. @TommyCpp any clever ideas here? If not then the original `Vec<String>` is probably the better return type here.

TommyCpp:
> @siyopao Sorry I hadn't looked into this closely enough, I don't think we can have this be an iterator without additional allocations as the lifetimes for something like `&dyn Iterator<Item=&str>` can't be easily resolved.

I agree. There is no guarantee how extractors will store those keys. So we may as well use `Vec`

siyopao:
No worries. Reverted that commit.

jtescher (approved):
Ok looks good 👍

PR #210, opened 2020-09-03
Add otlp exporter pipeline

jtescher:
This change adds a pipeline builder to `opentelemetry-otlp`. It simplifies recommended configuration and adds documentation to help people get started quickly. Simple example:

```rust
use opentelemetry::api::Tracer;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let tracer = opentelemetry_otlp::new_pipeline().install()?;

    tracer.in_span("doing_work", |cx| {
        // Traced app logic here...
    });

    Ok(())
}
```

A builder style pattern is also added to the sdk trace `Config` to allow for a consistent configuration in the full kitchen sink example added to the docs as well.

The `tokio` and `async-std` optional features are moved to `opentelemetry` from `opentelemetry-jaeger` to avoid duplicating the logic in `opentelemetry-otlp`. This allows trace provider builders to have a `with_exporter` that can be a batch exporter or simple exporter depending on the specified feature.

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -1,8 +1,206 @@
-//! # OpenTelemetry OTLP Exporter
+//! The OTLP Exporter supports exporting trace and metric data in the OTLP format to the OpenTelemetry collector. The OpenTelemetry Collector offers a
+//! vendor-agnostic implementation on how to receive, process, and export
+//! telemetry data. In addition, it removes the need to run, operate, and
+//! maintain multiple agents/collectors in order to support open-source
+//! telemetry data formats (e.g. Jaeger, Prometheus, etc.) sending to multiple
+//! open-source or commercial back-ends.
 //!
-//! The OpenTelemetry OTLP Exporter supports exporting of trace and metric data in the OTLP format.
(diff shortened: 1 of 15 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -1,8 +1,206 @@
-//! # OpenTelemetry OTLP Exporter
+//! The OTLP Exporter supports exporting trace and metric data in the OTLP format to the OpenTelemetry collector. The OpenTelemetry Collector offers a
```
TommyCpp, line 1: This line looks longer than 80 chars. may need to be wrapped.
```suggestion
//! The OTLP Exporter supports exporting trace and metric data in the 
//! OTLP format to the OpenTelemetry collector. The OpenTelemetry Collector offers a
```

TommyCpp (approved):
Looks great! Good to see users can now get the tracer up and running more easily. Just one small nit in comments.

opentelemetry-otlp/src/lib.rs
jtescher: ah yeah good catch

Issue #205, opened 2020-09-01
Specify that resources map to jaeger process tags


jtescher:
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/891) for details.

PR #215, opened 2020-09-08
resources map to jaeger process tags

danrusei:
#205 
I hope my understanding about what is requested here is correct, otherwise I'll modify based on the feedback.

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -211,8 +211,25 @@ impl trace::SpanExporter for Exporter {
     fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
         match self.uploader.lock() {
             Ok(mut uploader) => {
-                let jaeger_spans = batch.into_iter().map(Into::into).collect();
-                uploader.upload(jaeger::Batch::new(self.process.clone(), jaeger_spans))
+                let mut jaeger_spans: Vec<jaeger::Span> = Vec::with_capacity(batch.len());
+                let mut process_tags: Vec<jaeger::Tag> = Vec::new();
+
+                for (idx, span) in batch.into_iter().enumerate() {
+                    if idx == 0 {
+                        process_tags.extend(build_process_tags(&span));
+                    }
+                    jaeger_spans.push(span.into());
+                }
+                let mut process = self.process.clone();
+                let tags: Vec<jaeger::Tag> = match process.tags {
+                    Some(mut item) => {
+                        item.extend(process_tags);
+                        item
+                    }
+                    None => process_tags,
+                };
+                process.tags = Some(tags);
+                uploader.upload(jaeger::Batch::new(process, jaeger_spans))
             }
             Err(_) => trace::ExportResult::FailedNotRetryable,
         }
@@ -455,7 +472,7 @@ impl Into<jaeger::Span> for Arc<trace::SpanData> {
                 .duration_since(self.start_time)
                 .unwrap_or_else(|_| Duration::from_secs(0))
                 .as_micros() as i64,
-            tags: build_tags(&self),
+            tags: build_span_tags(&self),
             logs: events_to_logs(&self.message_events),
         }
     }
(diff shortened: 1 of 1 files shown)


opentelemetry-jaeger/src/lib.rs
```diff
@@ -211,8 +211,23 @@ impl trace::SpanExporter for Exporter {
     fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
         match self.uploader.lock() {
             Ok(mut uploader) => {
-                let jaeger_spans = batch.into_iter().map(Into::into).collect();
-                uploader.upload(jaeger::Batch::new(self.process.clone(), jaeger_spans))
+                let mut jaeger_spans: Vec<jaeger::Span> = Vec::new();
+                let mut process_tags: Vec<jaeger::Tag> = Vec::new();
+
+                for span in batch.iter() {
```
TommyCpp, line 217: Could just use the `batch.into_iter()` and switching the following two line like 
```rust
process_tags.extend(build_process_tags(&span));
jaeger_spans.push(span.into());
```

opentelemetry-jaeger/src/lib.rs
```diff
@@ -431,7 +446,7 @@ impl Into<jaeger::Log> for api::Event {
     }
 }
 
-impl Into<jaeger::Span> for Arc<trace::SpanData> {
+impl Into<jaeger::Span> for &Arc<trace::SpanData> {
```
TommyCpp, line 449: Probably no need to change this if we could just use `batch.into_iter()` above.

TommyCpp (changes requested):
Good start, thanks! have some suggestions

opentelemetry-jaeger/src/lib.rs
```diff
@@ -211,8 +211,23 @@ impl trace::SpanExporter for Exporter {
     fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
         match self.uploader.lock() {
             Ok(mut uploader) => {
-                let jaeger_spans = batch.into_iter().map(Into::into).collect();
-                uploader.upload(jaeger::Batch::new(self.process.clone(), jaeger_spans))
+                let mut jaeger_spans: Vec<jaeger::Span> = Vec::new();
```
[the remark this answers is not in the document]
danrusei: Thank you, I'm updating the PR with the suggested changes. But do we know the capacity of the process_tags ?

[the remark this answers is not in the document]
TommyCpp: Ha sorry I was looking at the wrong variable. We don't know the capacity of the process tags.

TommyCpp (approved):
Sorry for the late reply. LGTM. Thanks!

opentelemetry-jaeger/src/lib.rs
```diff
@@ -211,8 +211,23 @@ impl trace::SpanExporter for Exporter {
     fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult {
         match self.uploader.lock() {
             Ok(mut uploader) => {
-                let jaeger_spans = batch.into_iter().map(Into::into).collect();
-                uploader.upload(jaeger::Batch::new(self.process.clone(), jaeger_spans))
+                let mut jaeger_spans: Vec<jaeger::Span> = Vec::with_capacity(batch.len());
+                let mut process_tags: Vec<jaeger::Tag> = Vec::new();
+
+                for span in batch.into_iter() {
+                    process_tags.extend(build_process_tags(&span));
```
jtescher, line 218: Will this duplicate these tags for every span?

jtescher:
@danrusei generally looks good, just want to make sure this won't add duplicate process tags when exporting batches of spans

danrusei:
> @danrusei generally looks good, just want to make sure this won't add duplicate process tags when exporting batches of spans

thank you, would this work ?

    for tag in build_process_tags(&span) {
                        if !process_tags.contains(&tag) {
                            process_tags.push(tag)
                        }
                    }

jtescher:
@danrusei That should work but it may do a lot of work checking each time. The resource _should_ be the same for all spans, so you could do something like `for (idx, tag) in tags.enumerate() { if idx == 0 { add tags} ... }` to only add them the first time?

jtescher:
Looks good @danrusei, just have to rebase and make sure tests pass

jtescher (approved):
Looks good thanks @danrusei

PR #217, opened 2020-09-12
Add TraceState to SpanContext

### LINKED ISSUE #211: Add TraceState operations
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/905) for details.

awiede:
Resolves #211 

Adds `TraceState` to `SpanContext` and gives `TraceState` new methods per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracestate).

### DIFF
--- a/src/api/trace/span_context.rs
+++ b/src/api/trace/span_context.rs
@@ -28,7 +28,7 @@ pub const TRACE_FLAG_DEFERRED: u8 = 0x02;
 pub const TRACE_FLAG_DEBUG: u8 = 0x04;
 
 lazy_static::lazy_static! {
-    static ref TRACE_STATE_KEY: regex::Regex = regex::Regex::new(r#"^[a-zA-Z0-9]+[a-zA-Z0-9|\\_|\\-|\\*|\\/]*$"#).unwrap();
+    static ref TRACE_STATE_KEY: regex::Regex = regex::Regex::new(r#"^[a-z0-9]+[a-z0-9|\\_|\\-|\\*|\\/]*$"#).unwrap();
 }
 
 /// TraceId is an 16-byte value which uniquely identifies a given trace
--- a/src/api/trace/span_context.rs
+++ b/src/api/trace/span_context.rs
@@ -27,6 +27,10 @@ pub const TRACE_FLAG_DEFERRED: u8 = 0x02;
 /// TRACE_FLAGS_DEBUG is a bitmask with the debug bit set.
 pub const TRACE_FLAG_DEBUG: u8 = 0x04;
 
+lazy_static::lazy_static! {
+    static ref TRACE_STATE_KEY: regex::Regex = regex::Regex::new(r#"^[a-zA-Z0-9]+[a-zA-Z0-9|\\_|\\-|\\*|\\/]*$"#).unwrap();
+}
+
 /// TraceId is an 16-byte value which uniquely identifies a given trace
 /// The actual `u128` value is wrapped in a tuple struct in order to leverage the newtype pattern
 #[cfg_attr(feature = "serialize", derive(Deserialize, Serialize))]
@@ -127,6 +131,28 @@ pub struct TraceState {
 }
 
 impl TraceState {
+    /// Validates that the given `TraceState` list-member key is valid per the [W3 Spec]['spec'].
+    ///
+    /// ['spec']: https://www.w3.org/TR/trace-context/#key
+    fn valid_key(key: &str) -> bool {
+        if key.len() > 256 {
+            return false;
+        }
+
+        TRACE_STATE_KEY.is_match(key)
+    }
+
+    /// Validates that the given `TraceState` list-member value is valid per the [W3 Spec]['spec'].
+    ///
+    /// ['spec']: https://www.w3.org/TR/trace-context/#value
+    fn valid_value(value: &str) -> bool {
+        if value.len() > 256 {
+            return false;
+        }
+
+        !(value.contains(',') || value.contains('='))
+    }
+
     /// Creates a new `TraceState` from the given key-value collection.
     ///
     /// # Examples
@@ -168,6 +194,10 @@ impl TraceState {
     /// given key, this updates the value and updates the value's position. Returns the updated
     /// `TraceState`.
     pub fn insert(self, key: String, value: String) -> TraceState {
+        if !TraceState::valid_key(key.as_str()) || !TraceState::valid_value(value.as_str()) {
+            return self;
+        }
+
         let mut trace_state = self.delete(key.clone());
 
         trace_state.ordered_data.push_front((key, value));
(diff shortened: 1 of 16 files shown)


src/api/trace/span_context.rs
```diff
@@ -111,6 +114,110 @@ impl SpanId {
...
+/// of key-value pairs. TraceState allows multiple tracing systems to
+/// participate in the same trace.
+///
+/// Please review the [W3C specification] for details on this field.
+///
+/// [W3C specification]: https://www.w3.org/TR/trace-context/#tracestate-header
+#[cfg_attr(feature = "serialize", derive(Deserialize, Serialize))]
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct TraceState {
+    data: HashMap<String, String>,
```
jtescher, line 127: If this is an ordered map, could it simply be a `BTreeMap`?

src/api/trace/span_context.rs
```diff
@@ -111,6 +114,110 @@ impl SpanId {
...
+    ordered_keys: VecDeque<String>,
+}
+
+impl TraceState {
+    /// Creates a new `TraceState` from the given key-value collection.
+    pub fn from_key_value(trace_state: Vec<KeyValue>) -> Self {
+        let mut data: HashMap<String, String> = HashMap::with_capacity(trace_state.len());
```
jtescher, line 134: If you were to go with `BTreeMap` this could be simplified to something like:

```rust
pub fn new<T: IntoIterator<Item = (String, String)>>(trace_state: T) -> Self {
    TraceState {
        data: trace_state.into_iter().collect(),
    }
}
```

src/api/trace/span_context.rs
```diff
@@ -111,6 +114,110 @@ impl SpanId {
...
+        for kv in trace_state {
+            data.insert(kv.key.clone().into(), kv.value.into());
+            ordered_keys.push_back(kv.key.into());
+        }
+
+        TraceState { data, ordered_keys }
+    }
+
+    /// Retrieves a value for a given key from the `TraceState` if it exists.
+    pub fn get(&self, key: String) -> Option<String> {
```
jtescher, line 146: This could probably return `Option<&str>`, let the caller `clone` if they want owned data

src/api/trace/span_context.rs
```diff
@@ -111,6 +114,110 @@ impl SpanId {
...
+
+    /// Creates a new `TraceState` header string, with the given key/value delimiter and entry delimiter.
+    pub fn header_delimited(&self, entry_delimiter: &str, list_delimiter: &str) -> String {
+        let mut ordered: Vec<String> = Vec::with_capacity(self.ordered_keys.len());
+        for key in self.ordered_keys.clone() {
+            ordered.push(format!(
+                "{}{}{}",
+                key,
+                entry_delimiter,
+                self.data.get(key.as_str()).cloned().unwrap()
```
jtescher, line 189: could do `as_ref().unwrap()` here to avoid the clone.

If you go with `BTreeMap` this also could be simplified to:

```rust
for (key, value) in self.data.iter() {
    ordered.push(format!("{}{}{}", key, entry_delimiter, value))
}
```

src/api/trace/span_context.rs
```diff
@@ -178,6 +299,11 @@ impl SpanContext {
     pub fn is_sampled(&self) -> bool {
         (self.trace_flags & TRACE_FLAG_SAMPLED) == TRACE_FLAG_SAMPLED
     }
+
+    /// Returns the context's `TraceState`.
+    pub fn trace_state(&self) -> TraceState {
```
jtescher, line 304: This could also return `&TraceState` to let the caller clone.

jtescher (changes requested):
Great start @awiede! added a few ideas for potential allocation improvements

src/api/trace/span_context.rs
awiede: The problem is that the sorting is by insertion order, not key. So we could just keep everything in a `Vec` but then you have to iterate through everything on `get`. Rust doesn't have something like [Boost multi index](https://www.boost.org/doc/libs/1_62_0/libs/multi_index/doc/index.html), which would be ideal. There's an upper bound on how long trace state can be in the spec, so maybe it's not a big deal to just make this a wrapper around a `VecDeque` and `get` can just be `O(n)` where `n` isn't that big.

awiede:
@jtescher - I _think_ I addressed all of your comments. RE: using a `BTreeMap`, I don't think that will work since we aren't sorting on the keys, but instead on the insertion order. Given that this should be a relatively small data set, I updated to just use an ordered collection of tuples. `TraceState::get` is now `O(n)` instead `O(constant)`, _but_ given the limitations on the size of `tracestate` itself (I think the spec says 32 elements), I don't know if that's a big deal.

src/api/trace/span_context.rs
```diff
@@ -111,6 +113,116 @@ impl SpanId {
...
+        TraceState {
+            ordered_data: trace_state
+                .into_iter()
+                .map(|(key, value)| (key.to_string(), value.to_string()))
+                .collect(),
+        }
+    }
+
+    /// Retrieves a value for a given key from the `TraceState` if it exists.
+    pub fn get(&self, key: String) -> Option<&str> {
```
jtescher, line 157: This could probably take an `&str` to avoid callers having to clone.

src/api/trace/span_context.rs
```diff
@@ -111,6 +113,116 @@ impl SpanId {
...
+                None
+            }
+        })
+    }
+
+    /// Inserts the given key-value pair into the `TraceState`. If a value already exists for the
+    /// given key, this updates the value and updates the value's position. Returns the updated
+    /// `TraceState`.
+    pub fn insert(self, key: String, value: String) -> TraceState {
+        let mut trace_state = self.delete(key.clone());
```
jtescher, line 171: I just re-read the spec and it looks like:

> All mutating operations MUST return a new TraceState with the modifications applied. 

So we'll have to clone the trace state instead of modifying in place looks like.

Also 

> If invalid value is passed the operation MUST NOT return TraceState containing invalid data

So looks like these methods have to return `Result` types and do some validating

src/api/trace/trace_context_propagator.rs
```diff
@@ -171,20 +152,21 @@ impl api::TextMapFormat for TraceContextPropagator {
...
 mod tests {
     use super::*;
-    use crate::api::{Extractor, TextMapFormat};
+    use crate::api::{Extractor, Injector, TextMapFormat};
     use std::collections::HashMap;
+    use std::str::FromStr;
 
     #[rustfmt::skip]
-    fn extract_data() -> Vec<(&'static str, api::SpanContext)> {
+    fn  extract_data() -> Vec<(&'static str, &'static str, api::SpanContext)> {
```
jtescher, line 160: this first whitespace was probably not intended.

jtescher (changes requested):
@awiede yeah that makes sense I hadn't read the spec all that closely 😬, looked it over and noticed a few other changes we may want to do

src/api/trace/trace_context_propagator.rs
awiede: Ooops, serves me right for skipping formatting on this function 😞

src/api/trace/span_context.rs
awiede: So `insert` and `delete` are `FnOnce` functions, I can have them return a `Result<TraceState, ()>`, but that would mean on error the original valid `TraceState` has been moved/lost. I could do `Result<TraceState, TraceState>` which would return a clone of the unmodified input which I guess is better than silently ignoring an operation with an invalid key/value?

jtescher: I think they have to be `Fn` functions and take `&self` to be spec compliant, unless I'm misunderstanding the `All mutating operations MUST return a new TraceState with the modifications applied.` part.

jtescher: Seems like trace state is meant to be immutable so you have copies with the changes instead of mutating the underlying values

awiede:
@jtescher updated to `Fn` for `insert` and `delete`. Spec technically says that keys should be lowercase which leads to some annoying conflicts with the AWS propagator. I think it's fine for now since the AWS reserved keys only capitalize the first character, custom camel cased keys might get hairy.

jtescher (approved):
Looks good thanks @awiede

PR #218, opened 2020-09-15
DDSketch 

TommyCpp:
Probably need to add more comments or include some of my notes on algorithm and Go implementation. But want to open this PR to make sure I am in right track.


jtescher:
I'm not very familiar with the ddsketch paper, but the implementation looks good so far to me 👍

jtescher (approved):
LGTM nice work @TommyCpp!

PR #222, opened 2020-09-20
Add datadog exporter

### LINKED ISSUE #216: Datadog Exporter
Datadog uses a custom trace export format defined [here](https://docs.datadoghq.com/api/v1/tracing/#send-traces). In addition to the JSON format documented there, official implementations use msgpack - see [here](https://github.com/DataDog/dd-trace-go/blob/v1/ddtrace/tracer/payload.go).

I have found the beginnings of a datadog exporter for the opentelemetry collector [here](https://github.com/DataDog/opentelemetry-collector-contrib/tree/master/exporter/datadogexporter) but it would be preferable for Rust applications to be able to export directly to DataDog.

I'd be happy to submit an implementation if this would be of interest

tustvold:
Closes #216 

Unfortunately this turned into a bit of a treasure hunt as the Datadog docs aren't particularly helpful, and Datadog has a slightly different design to OpenTelemetry requiring some fudgery. I've tried to document and provide links to where I sourced information from.

This implementation also uses the v0.5 trace ingestion API included in the latest datadog release. I hadn't realised quite how recent the API change was until I had already written most of the code, but the v0.3 is significantly simpler and so could be fairly easily added in future if there is demand for it.

To implement this exporter I made two changes to the actual OpenTelemetry "core"

- Add the ability to get values from EvictedHashMap by key
- Expose the InstrumentationLibrary in SpanData, the Golang OpenTelemetry implementation already does this so hopefully this isn't too controversial - see [here](https://github.com/open-telemetry/opentelemetry-go/blob/a12224a454135a5d7ec17831b6b39cf9723c0cdb/sdk/export/trace/trace.go#L79)

### DIFF
--- a/examples/datadog/README.md
+++ b/examples/datadog/README.md
@@ -1,0 +1,15 @@
+# Datadog Exporter Example
+
+Sends spans to a datadog-agent collector.
+
+## Usage 
+
+First run version 7.22.0 or above of the datadog-agent locally as described [here](https://docs.datadoghq.com/agent/)
+
+Then run the example to report spans:
+
+```shell
+$ cargo run
+```
+
+Traces should appear in the datadog APM dashboard
--- a/src/exporter/trace/mod.rs
+++ b/src/exporter/trace/mod.rs
@@ -85,6 +85,9 @@ pub struct SpanData {
     pub status_message: String,
     /// Resource contains attributes representing an entity that produced this span.
     pub resource: Arc<sdk::Resource>,
+    /// Instrumentation library that produced this span
+    #[cfg_attr(feature = "serialize", serde(skip))]
+    pub instrumentation_lib: sdk::InstrumentationLibrary,
 }
 
 #[cfg(feature = "serialize")]
@@ -136,6 +139,7 @@ mod tests {
             status_code,
             status_message,
             resource,
+            instrumentation_lib: sdk::InstrumentationLibrary::new("", None),
         };
 
         let encoded: Vec<u8> = bincode::serialize(&span_data).unwrap();
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ members = [
     "examples/aws-xray",
     "examples/basic",
     "examples/basic-otlp",
+    "examples/datadog",
     "examples/grpc",
     "examples/http",
     "examples/zipkin",
--- a/benches/ddsketch.rs
+++ b/benches/ddsketch.rs
@@ -1,12 +1,8 @@
-use criterion::{
-    criterion_group, criterion_main, Criterion
-};
+use criterion::{criterion_group, criterion_main, Criterion};
 use opentelemetry::api::metrics::{InstrumentKind, Number, NumberKind};
 use opentelemetry::sdk::export::metrics::Aggregator;
 use opentelemetry::{
-    api::{
-        metrics::Descriptor,
-    },
+    api::metrics::Descriptor,
     sdk::{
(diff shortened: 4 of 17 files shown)


jtescher (changes requested):
Great start @tustvold! The changes to expose methods on the exported data seem reasonable here.

I think from a location perspective this the exporter may make sense as part of the contrib crate (or in the [data dog](https://github.com/DataDog?q=opentelemetry&type=&language=) org maybe?). As far as I know the otel preference is to have vendor specific exporters in their own repos or in contrib.

tustvold:
Moved into opentelemetry-contrib behind a feature flag, let me know if this isn't what you meant.

opentelemetry-contrib/src/datadog/model.rs
```diff
@@ -0,0 +1,202 @@
...
+// 		 6: 0                    // Start
+// 		 7: 0                    // Duration
+// 		 8: 0                    // Error
+// 		 9: map[uint32]uint32{}  // Meta (empty map)
+// 		10: map[uint32]float64{} // Metrics (empty map)
+// 		11: 0                    // Type is ""
+//
+// 		The dictionary in this case would be []string{""}, having only the empty string at index 0.
+//
+pub(crate) fn encode_v05(
```
TommyCpp, line 75: Might be better to make this an enum if we are expecting multiple version. But also OK if do it in a separate PR.

TommyCpp (commented):
Nice work! Small suggestions that may make adding other versions easier.

opentelemetry-contrib/src/datadog/model.rs
tustvold: Added - and implemented the v0.3 API whilst I was at it

src/exporter/trace/mod.rs
```diff
@@ -85,6 +85,9 @@ pub struct SpanData {
     pub status_message: String,
     /// Resource contains attributes representing an entity that produced this span.
     pub resource: Arc<sdk::Resource>,
+    /// Instrumentation library that produced this span
+    #[cfg_attr(feature = "serialize", serde(skip))]
```
jtescher, line 89: Why skip this?

jtescher (commented):
Looks good to me, couple of small nits, @TommyCpp did you have more comments?

TommyCpp (approved):
@jtescher Nope. LGTM, Thanks!

jtescher:
@tustvold couple of comments, could address them in a follow up PR as well if needed.

src/exporter/trace/mod.rs
tustvold: InstrumentationLibrary uses borrowed strings which causes issues for serde Deserialize.

Additionally avoids any unintentional side effects of this change

examples/datadog/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Usage 
+
+First run version 7.22.0 or above of the datadog-agent locally as described [here](https://docs.datadoghq.com/agent/)
+
+Then run the example to report spans:
+
+```shell
+$ cargo run
+
```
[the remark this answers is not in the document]
tustvold: Fixed

src/exporter/trace/mod.rs
jtescher: ok, can revisit if this causes issues for anyone who wants to depend on the serialized version at some point. Seems fine for now as the binary formats are still in flux.

jtescher (approved):

Issue #156, opened 2020-07-24
Simplify common configurations


jtescher:
There is currently a lot of boilerplate to set up a tracing / metrics pipeline as seen in the examples `init_tracer` functions. We should consider adding a high level pipeline configuration builder to support simpler sdk config.

Pipeline config needed:

- [x] `opentelemetry-jaeger` #189
- [x] `opentelemetry-otlp` #210
- [x] `opentelemetry-prometheus` #141
- [x] `opentelemetry-zipkin`  #214
- [x] `opentelemetry` (stdout)

PR #224, opened 2020-09-22
Pipeline builder for stdout trace exporter

frigus02:
Part of #156

Implement the pipeline builder API for the stdout trace exporter, similar to the other exporters. With this a stdout trace pipeline can now be installed like this:

```rust
let tracer = opentelemetry::exporter::trace::stdout::new_pipeline().install();
```

I made 2 small changes along the way, which seemed to make sense to me. Happy to change in either direction, though. These are:

- Defer requirement of all additional writer traits to the `PipelineBuilder::install` function. Before `Debug` was required for the builder and `Send + 'static` were required only for the exporter. Now the builder is happy with only `Write` and requires `Debug + Send + 'static` for installation. I think I don't see any downsides here. We could also require all of these traits for the builder already. Not sure what's better.

- Defer wrapping writer in Mutex to when the exporter is created, which seems to be the point at which it's needed. Does this have any downsides?

Additional note:

This makes the pipeline the only way to construct a stdout exporter. This is in line with other exporters, but I'm not sure it's desired. Should we make `Exporter::new` public, so users can still create an exporter and setup a pipeline manually?

Update: `Exporter::new` is now public.

### DIFF
--- a/examples/aws-xray/src/client.rs
+++ b/examples/aws-xray/src/client.rs
@@ -4,21 +4,17 @@ use opentelemetry::{api, exporter::trace::stdout, global, sdk};
 use opentelemetry_contrib::{XrayIdGenerator, XrayTraceContextPropagator};
 
 fn init_tracer() {
-    // Create stdout exporter to be able to retrieve the collected spans.
-    let exporter = stdout::Builder::default().init();
-
+    // Install stdout exporter pipeline to be able to retrieve the collected spans.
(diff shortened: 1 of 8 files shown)


jtescher (approved):
This looks good @frigus02, I think making `Exporter::new` public is probably a good idea, can do that here or in a follow up PR

frigus02:
Sounds good. I just pushed a commit to make `Exporter::new` public and add docs for it.

Issue #168, opened 2020-08-04
Add configuration options from environment variables


jtescher:
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/666) and [the followup](https://github.com/open-telemetry/opentelemetry-specification/pull/758) for details.

ozerovandrei:
Note that OTLP Exporter spec was updated: [exporter.md](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)

PR #228, opened 2020-09-24
Add config options from env for BatchSpanProcessor

TommyCpp:
part of #168

### DIFF
--- a/src/sdk/trace/span_processor.rs
+++ b/src/sdk/trace/span_processor.rs
@@ -179,7 +179,7 @@ impl api::SpanProcessor for BatchSpanProcessor {
 #[allow(missing_debug_implementations)]
 pub struct BatchSpanProcessorWorker {
     exporter: Box<dyn exporter::trace::SpanExporter>,
-    messages: Pin<Box<dyn Stream<Item = BatchMessage> + Send>>,
+    messages: Pin<Box<dyn Stream<Item=BatchMessage> + Send>>,
     config: BatchConfig,
     buffer: Vec<Arc<exporter::trace::SpanData>>,
 }
@@ -243,10 +243,10 @@ impl BatchSpanProcessor {
         interval: I,
         config: BatchConfig,
     ) -> Self
-    where
-        S: Fn(BatchSpanProcessorWorker) -> SO,
-        I: Fn(time::Duration) -> IS,
-        IS: Stream<Item = ISI> + Send + 'static,
+        where
+            S: Fn(BatchSpanProcessorWorker) -> SO,
+            I: Fn(time::Duration) -> IS,
+            IS: Stream<Item=ISI> + Send + 'static,
     {
         let (message_sender, message_receiver) = mpsc::channel(config.max_queue_size);
         let ticker = interval(config.scheduled_delay).map(|_| BatchMessage::Tick);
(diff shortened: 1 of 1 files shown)


src/sdk/trace/span_processor.rs
```diff
@@ -269,6 +283,56 @@ impl BatchSpanProcessor {
...
+    ) -> BatchSpanProcessorBuilder<E, S, I>
+        where
+            E: exporter::trace::SpanExporter,
+            S: Fn(BatchSpanProcessorWorker) -> SO,
+            I: Fn(time::Duration) -> IO,
+    {
+        let mut config = BatchConfig::default();
+        let schedule_delay = u64::from_str(
+            std::env::var(OTEL_BSP_SCHEDULE_DELAY_MILLIS)
+                .unwrap_or_else(|_| "".to_string())
```
jtescher, line 301: Would these be simpler as:

```rust
std::env::var(OTEL_BSP_SCHEDULE_DELAY_MILLIS)
    .and_then(|delay| u64::from_str(&delay).ok())
    .unwrap_or(OTEL_BSP_SCHEDULE_DELAY_MILLIS_DEFAULT);
```

At a quick glance it's not clear that `"".to_string()` isn't 0, or some other case.

src/sdk/trace/span_processor.rs
```diff
@@ -269,6 +283,56 @@ impl BatchSpanProcessor {
             config: Default::default(),
         }
     }
+
+    /// Create a new batch processor builder and set the config value based on environment variables.
+    pub fn from_env<E, S, SO, I, IO>(
```
jtescher, line 288: Do you prefer the explicit `from_env` here? Has advantages of being clear when it is looking from env, alternative could be to have the batch config be options and fall back to checking env variables for any unset values sorta merging both constructors?

jtescher (commented):
Great start @TommyCpp, had a couple of questions but looking good

src/sdk/trace/span_processor.rs
TommyCpp: I guess that depends on whether we should assume users always want to apply environment variable configuration. Personally I'd like know exactly where does the configuration comes from. But we could use environment variable to fill unset values in configs too.

jtescher: Yeah I see the pros and cons of both (e.g. the jaeger exporter has an explicit `from_env`). I think it is fine for now to get feedback from the community on which is preferred.

src/sdk/trace/span_processor.rs
```diff
@@ -269,6 +283,57 @@ impl BatchSpanProcessor {
...
+    ) -> BatchSpanProcessorBuilder<E, S, I>
+        where
+            E: exporter::trace::SpanExporter,
+            S: Fn(BatchSpanProcessorWorker) -> SO,
+            I: Fn(time::Duration) -> IO,
+    {
+        let mut config = BatchConfig::default();
+        let schedule_delay = std::env::var(OTEL_BSP_SCHEDULE_DELAY_MILLIS)
+            .and_then(|delay| u64::from_str(&delay).or(Ok(OTEL_BSP_SCHEDULE_DELAY_MILLIS_DEFAULT)))
+            .unwrap();
```
jtescher, line 307: This way will panic if the var is not set I think? Probably best to go with a version that doesn't have a straight `.unwrap` in it, those tend to make people nervous when they look at the code.

jtescher (changes requested):

src/sdk/trace/span_processor.rs
TommyCpp: Yeah you are right. Should have tested it with no env vars set I guess. Fixed it.

jtescher (approved):
Thanks @TommyCpp

PR #229, opened 2020-09-24
Add pipeline uninstall mechanism

### LINKED ISSUE #225: TraceProvider::with_exporter immediately drops BatchSpanProcessor
While testing #224, I came across an issue.

## Steps to reproduce

1. Checkout the PR (at the point of writing https://github.com/open-telemetry/opentelemetry-rust/pull/224/commits/ffa96df67008202b784b64d0986e255ffd26cac1).
1. Open `examples/stdout.rs` and change:
   ```diff
   + #[tokio::main]
   + async fn main() {
   - fn main() {
   ```
1. Run `cargo run --example stdout` and observe the span being printed to the console.
1. Run `cargo run --example stdout --features tokio` and observe that nothing is printed. With the tokio feature enabled it used the BatchSpanProcessor setup by TraceProvider::with_exporter.

## Analysis

I reproduced it on master as well and with a few print statements found that the `BatchSpanProcessorWorker` is immediately dropped after calling with_exporter.

See my experiments here: https://github.com/open-telemetry/opentelemetry-rust/commit/13fc775f92967eced4ba3581129d6091112ed8f3

The reason is that with_exporter uses either [`tokio::task::spawn_blocking`](https://docs.rs/tokio/0.2.22/tokio/task/fn.spawn_blocking.html) or [`async_std::task::spawn_blocking`](https://docs.rs/async-std/1.6.4/async_std/task/fn.spawn_blocking.html) (see [sdk::TraceProvider](https://github.com/open-telemetry/opentelemetry-rust/blob/0fa4e7d506cb52520607fa5da70d0efa15e1f6cb/src/sdk/trace/provider.rs)). Both of those expect a blocking function, not a function returning a future. This means the BatchSpanProcessorWorker is never polled.

## Possible solution

I assume the reason for using `spawn_blocking` is that the exporters currently run blocking, so it's good to run on a special thread for blocking work. The solution for the problem then is most likely to execute the future using block_on:

```rust
let spawn = |future| {
    tokio::task::spawn_blocking(|| {
        futures::executor::block_on(future);
        // or tokio::runtime::Handle::current().block_on(future)
        // or async_std::task::block_on(future)
    })
};
let batch = sdk::BatchSpanProcessor::builder(exporter, spawn, async_std::stream::interval);
```

If I do this, however, the program never exits. You can test this with the above linked example. The BatchSpanProcessorWorker just keeps running because it does never receive a shotdown message.

I'm not sure why, yet. I think the provider should be dropped and then notify all registered span processors. But that does not seem to happen.

If I manually call shutdown for all span processors, the worker stops properly, exports the remaining spans, and then the program exits.

I'm happy to work on a fix, but I would like to hear your thoughts first. What's the best way to proceed here?

jtescher:
In order to address the issues raised in #225, the following changes are proposed:

1) Fix the current bug where futures are being dropped instead of properly awaited by updating:

```diff
- let spawn = |future| tokio::task::spawn_blocking(|| future);
+ let spawn = |future| tokio::task::spawn_blocking(|| futures::executor::block_on(future));
```

2) Because `spawn_blocking` will now block the program from exiting until the future completes, add a drop guard to pipeline install methods which uninstalls the global tracer provider (and in the otlp case in the future also the meter provider) so the exporters properly shut down, causing the future to resolve.

    Alternatives considered here could be: 

    a) let users unset the global(s) themselves (but they would likely forget, causing the same blocking behavior)
    b) return a method that could be called instead of a drop guard (similar problems)
    c) Current impl uses custom uninstall objects per exporter to make the struct `opentelemetry_jaeger::Uninstall` rather than a more generic `global::TracerProviderGuard` or `global::UninstallPipeline`, alternative name suggestions could be helpful if they improve clarity.
    c) Return a pipeline object that contains a `Tracer`, and could be the drop guard itself, maybe cleaner API but less explicit, **possibly a better choice**.

3) Because `Tracer`s currently retain strong references to their `TracerProvider`, and shutdown is called on the inner's drop, it is still never called. To fix this, update the `Tracer` to instead contain a `Weak<TracerProviderInner>` reference, which can be upgraded when needed. This does force other logic like span creation to consider the case where the `TracerProvider` has been dropped on shutdown, but making this case explicit seems beneficial in most cases.

4) Additional clarity/consistency tweaks:

    * Rename `global::set_provider` to `global::set_tracer_provider`
    * Rename `global::trace_provider` to `global::tracer_provider`
    * Rename `ProviderInner` to `TracerProviderInner`

Fixes #225

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ and metrics from your application. You can analyze them using [Prometheus], [Jae
 use opentelemetry::{api::TracerGenerics, global, sdk};
 
 fn main() {
-    global::set_tracer_provider(sdk::Provider::default());
+    let _guard = global::set_tracer_provider(sdk::Provider::default());
 
     global::tracer("component-a").in_span("foo", |_context| {
         global::tracer("component-b").in_span("bar", |_context| {
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ and metrics from your application. You can analyze them using [Prometheus], [Jae
 use opentelemetry::{api::TracerGenerics, global, sdk};
 
 fn main() {
-    global::set_provider(sdk::Provider::default());
+    global::set_tracer_provider(sdk::Provider::default());
 
     global::tracer("component-a").in_span("foo", |_context| {
         global::tracer("component-b").in_span("bar", |_context| {
--- a/examples/aws-xray/src/client.rs
+++ b/examples/aws-xray/src/client.rs
@@ -20,7 +20,7 @@ fn init_tracer() -> (sdk::Tracer, stdout::Uninstall) {
 
 #[tokio::main]
 async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    let _guard = init_tracer();
+    let (_tracer, _guard) = init_tracer();
 
     let client = Client::new();
     let span = global::tracer("example/client").start("say hello");
--- a/examples/aws-xray/src/client.rs
+++ b/examples/aws-xray/src/client.rs
@@ -3,7 +3,9 @@ use opentelemetry::api::{Context, TraceContextExt, Tracer};
 use opentelemetry::{api, exporter::trace::stdout, global, sdk};
 use opentelemetry_contrib::{XrayIdGenerator, XrayTraceContextPropagator};
 
-fn init_tracer() {
+fn init_tracer() -> (sdk::Tracer, stdout::Uninstall) {
+    global::set_text_map_propagator(XrayTraceContextPropagator::new());
+
     // Install stdout exporter pipeline to be able to retrieve the collected spans.
     // For the demonstration, use `Sampler::AlwaysOn` sampler to sample all traces. In a production
     // application, use `Sampler::ParentBased` or `Sampler::TraceIdRatioBased` with a desired ratio.
@@ -13,14 +15,12 @@ fn init_tracer() {
             id_generator: Box::new(XrayIdGenerator::default()),
             ..Default::default()
         })
-        .install();
-
-    global::set_text_map_propagator(XrayTraceContextPropagator::new());
+        .install()
 }
 
 #[tokio::main]
 async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    init_tracer();
+    let _guard = init_tracer();
 
     let client = Client::new();
     let span = global::tracer("example/client").start("say hello");
--- a/examples/actix-http/src/main.rs
+++ b/examples/actix-http/src/main.rs
@@ -6,7 +6,7 @@ use opentelemetry::api::{Key, TraceContextExt, Tracer};
 use opentelemetry::{global, sdk};
 use std::error::Error;
 
-fn init_tracer() -> Result<sdk::Tracer, Box<dyn Error>> {
+fn init_tracer() -> Result<(sdk::Tracer, opentelemetry_jaeger::Uninstall), Box<dyn Error>> {
     opentelemetry_jaeger::new_pipeline()
         .with_collector_endpoint("http://127.0.0.1:14268/api/traces")
         .with_service_name("trace-http-demo")
@@ -25,7 +25,7 @@ async fn index() -> &'static str {
 async fn main() -> std::io::Result<()> {
     std::env::set_var("RUST_LOG", "debug");
     env_logger::init();
-    let tracer = init_tracer().expect("Failed to initialise tracer.");
+    let (tracer, _uninstall) = init_tracer().expect("Failed to initialise tracer.");
 
     HttpServer::new(move || {
         let tracer = tracer.clone();
--- a/examples/actix-udp/src/main.rs
+++ b/examples/actix-udp/src/main.rs
@@ -6,7 +6,7 @@ use opentelemetry::api::{Key, TraceContextExt, Tracer};
 use opentelemetry::{global, sdk};
 use std::error::Error;
 
-fn init_tracer() -> Result<sdk::Tracer, Box<dyn Error>> {
+fn init_tracer() -> Result<(sdk::Tracer, opentelemetry_jaeger::Uninstall), Box<dyn Error>> {
     opentelemetry_jaeger::new_pipeline()
         .with_agent_endpoint("localhost:6831")
         .with_service_name("trace-udp-demo")
@@ -25,7 +25,7 @@ async fn index() -> &'static str {
 async fn main() -> std::io::Result<()> {
     std::env::set_var("RUST_LOG", "debug");
     env_logger::init();
-    init_tracer().expect("Failed to initialise tracer.");
+    let _uninstall = init_tracer().expect("Failed to initialise tracer.");
 
     HttpServer::new(|| {
         App::new()
--- a/examples/async/src/main.rs
+++ b/examples/async/src/main.rs
@@ -50,7 +50,7 @@ async fn run(addr: &SocketAddr) -> io::Result<usize> {
     write(&mut stream).with_context(cx).await
 }
 
-fn init_tracer() -> Result<sdk::Tracer, Box<dyn Error>> {
+fn init_tracer() -> Result<(sdk::Tracer, opentelemetry_jaeger::Uninstall), Box<dyn Error>> {
     opentelemetry_jaeger::new_pipeline()
         .with_service_name("trace-demo")
         .install()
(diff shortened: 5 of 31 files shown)


jtescher:
cc @frigus02

examples/aws-xray/src/client.rs
```diff
@@ -13,14 +15,12 @@ fn init_tracer() {
...
-        .install();
-
-    global::set_text_map_propagator(XrayTraceContextPropagator::new());
+        .install()
 }
 
 #[tokio::main]
 async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    init_tracer();
+    let _guard = init_tracer();
```
frigus02, line 23: Is it worth destructuring here to make it explicit which of the parts is the guard? It's a similar case for the other examples. Or maybe we shouldn't worry about it too much.

```suggestion
    let (_tracer, _guard) = init_tracer();
```

README.md
```diff
@@ -20,7 +20,7 @@ and metrics from your application. You can analyze them using [Prometheus], [Jae
 use opentelemetry::{api::TracerGenerics, global, sdk};
 
 fn main() {
-    global::set_provider(sdk::Provider::default());
+    global::set_tracer_provider(sdk::Provider::default());
```
frigus02, line 23: I think we need to hold the guard here, otherwise this example will not work as expected.
```suggestion
    let _guard = global::set_tracer_provider(sdk::Provider::default());
```

frigus02 (commented):
Nice. Thanks for working on this. I left 2 tiny comments, but it's probably better to focus on the overall direction.

Regarding the `Uninstall` struct returned from the exporters:
- I like the name. "Uninstall" is quite clear to me. It doesn't immedately tell me it's going to happen on drop, which something like `PipelineGuard` might do. But I can't say if that's actually a better name. I think wouldn't mind returning `global::TracerProviderGuard` directly, either. But really, `Uninstall` seems good 🙂.

The additional tweaks make sense to me. 👍

I'm not entirely convinced using weak references in `Tracer`s is a good idea. I'm still thinking about it but these are my thoughts:
- What if I don't install a global tracer provider? This could be an option when using `tracing-opentelemetry` or maybe if I'm writing a performance critical application? In any case, if I understand it correctly I would have to keep at least 1 strong reference to the provider around, which is not obvious.
- In general I like the idea of calling shutdown on drop. I think this means whenever there is a span or tracer somewhere, there is definitelty still a reference to the provider, so the pipeline is still running.
- Weak references are only needed because we keep references to named tracers. I might be missing something but I don't understand the value this brings at the moment. It doesn't save memory because we're cloning anyway. And there is no logic happening on tracer initialization, so we don't save CPU, either. Would we loose anything if we remove the named tracers HashMap? 🤔

jtescher:
@frigus02 great point, probably better to just drop the map. The locking is likely a higher cost than initializing a new tracer (even if they were reduced to Arc strong increments), so this may be a good idea in either case.

I'm not totally convinced that uninstall is the best name, but on the other hand I'm not sure how familiar the rust community is generally with drop guards, this option seems a little more oriented to devs who want things to just work.

The pipeline object _may_ have the benefit of being less of a breaking change to update things, e.g. not totally clear for OTLP if people would prefer one pipeline initialization vs separate metrics and trace pipelines 🤷‍♂️. But generally I like the simplicity of the tuple destructure. The current impl may optimize slightly more for the getting started example readability, which has benifits as that may be how most devs will interact with otel in general.

TommyCpp:
Checked with the current version of spec about named tracer. 
> It is unspecified whether or under which conditions the same or different Tracer instances are returned from this functions.

Didn't really say we need to return the same tracer instance given the same instrument library info. We just need to make sure they can pick up the configuration change, which could be done by keeping a reference of `TracerProvider` in `Tracer`. So it should be good to drop the map

frigus02:
> I'm not totally convinced that uninstall is the best name

Yeah, it might be better to be consistent with other Rust APIs and use something like a `PipelineGuard`.

I also like the idea of a `Pipeline` object because of it's extensibility. But I'm not sure if the drop behaviour is obvious in that case. Althought this might be fine if we add a big note in the docs 🙂

jtescher:
@frigus02 removed the provider's tracer cache 👍 

> Weak references are only needed because we keep references to named tracers. 

The other reason to switch to weak maps is because it is reasonable for other systems to store tracers or objects with references to tracers in thread locals or other effectively `'static` ways (e.g. [tracing](https://github.com/tokio-rs/tracing/tree/master/tracing-opentelemetry) stores a tracer in the otel layer impl).

frigus02:
Yeah, that's true. Maybe we need to document that traders will stop working if their provider is not stored in any strong reference.

jtescher:
A larger alternative would instead be to switch exporters to be async, allowing `spawn` instead of `spawn_blocking` which may remove the need to uninstall? Not sure if there are other reasons even in that case where we would want an explicit uninstall.

frigus02:
Thinking about it again, I think I would prefer if `tracing-opentelemetry` would be responsible to provide a way to release the `Tracer` reference, similar to the `Uninstall` guard in this PR. It feels like a slightly more intuitive API without weak references. But that might well be subjective. 🙂

> A larger alternative would instead be to switch exporters to be async, allowing spawn instead of spawn_blocking which may remove the need to uninstall? Not sure if there are other reasons even in that case where we would want an explicit uninstall.

Async exporters could be nice. I wonder if that eliminates the need for a controlled shutdown, though. We would still need to notify the batch processor to export the last spans and we want to wait for this to finish before exiting the program.

We actually still have this issue with `async-std` with the implementation in this PR. It seems to be a `tokio` thing to automatically wait for blocking tasks to finish.

I played a bit with this recently. This makes the BatchSpanProcessor wait for its worker to finish on shutdown: https://github.com/frigus02/opentelemetry-rust/commit/5dbe15da14269af9d701e519e05dc0173b90cce3. This works for async-std and tokio now and I think it should also work for async exporters later. If that looks alright, I can submit this as a follow up PR. Or feel free to copy-paste it here.

jtescher:
@frigus02 good point, I applied your patch with a slight change to generics to allow `tokio::spawn` to be used directly as we aren't actually interested in the results 👍

jtescher:
> Thinking about it again, I think I would prefer if `tracing-opentelemetry` would be responsible to provide a way to release the Tracer reference, similar to the Uninstall guard in this PR. It feels like a slightly more intuitive API without weak references. But that might well be subjective. 🙂

Seems like the two architectural choices here are drop guards vs explicit method calls, and weak refs vs other mechanism for storing shutdown state and making subsequent spans no-ops?

I think I'd lean toward drop guards as they are trickier to use incorrectly (although assigning to `_` is one way people will likely run into issues), I don't have as strong feelings about weak refs vs some other no-op state mechanism as they seem like internal implementation details. Did you have other alternatives or tradeoffs in mind?

frigus02:
> I think I'd lean toward drop guards as they are trickier to use incorrectly

I think I agree.

> I don't have as strong feelings about weak refs vs some other no-op state mechanism as they seem like internal implementation details.

My only concern with Tracer having a weak reference to the TraceProvider is that code like this would not work anymore, because the provider would be dropped after `init_tracer`:

```rust
fn init_tracer() -> sdk::Tracer {
    let provider = sdk::TracerProvider::builder().with_exporter(...).build();
    provider.get_tracer("", None)
}

fn main() {
    let tracer = init_tracer();
    tracer.in_span(...);
}
```

That could also be solved with documentation. I probably prefer the way it currently works. But that might just be because I'm used to it 🙂.

frigus02 (approved):
Again, thanks for working on this!

I mentioned the downside of having to keep a TracerProvider around, which we may want to document. And everything else looks great 🎉 

There are 2 READMEs, which we should update as well:

- opentelemetry-jaeger/README.md
- opentelemetry-otlp/README.md

jtescher:
@frigus02 yeah seems like that is a casualty of the drop guard choice, with an explicit call to shutdown that example could work, but with guards it is tricky (in that example it would need _some_ method of shutting down if the batch exporter was used in the builder)

frigus02:
Hm, if the Tracer owns a clone of the TracerProvider it should work, I think. When the last Tracer is dropped, the TracerProviderInner would drop as well and can shutdown the span processor. 🤔

TommyCpp (approved):
Not too familar with async, but the rest changes looks good. Nice job.

frigus02 (approved):

jtescher:
Ok can merge for now to keep things going. If there are clearer name ideas or feedback about how to improve this we can add in subsequent PR's 👍

jtescher:
@frigus02  if you want to be a member of the otel org ([requirements](https://github.com/open-telemetry/community/blob/master/community-membership.md#requirements)) I'm happy to sponsor you, just have [open an issue here](https://github.com/open-telemetry/community/issues/new?template=membership.md&title=REQUEST%3A%20New%20membership%20for%20%3Cyour-GH-handle%3E) to get added.

frigus02:
Yeah, why not. The responsibilities are reasonable. Sounds like I need one more sponsor. @TommyCpp would you be up for that? 🙂

TommyCpp:
@frigus02 More than happy to :smile:! Just added me as sponsor in the issue.

PR #232, opened 2020-09-27
Switch to async exporters

jtescher:
This change updates the `SpanExporter interface to export asynchronously. The primary API change is:

```diff
- fn export(&self, batch: Vec<Arc<trace::SpanData>>) -> trace::ExportResult;
+ async fn export(&self, batch: &[Arc<trace::SpanData>]) -> trace::ExportResult;
```

Implementation notes:

* Switch from `Vec<_>` to `&[_]` in exporters to avoid additional allocation when batch exporting.
* Currently uses [async_trait](https://github.com/dtolnay/async-trait), could change method signature to avoid the dependency if preferable.
* `BatchSpanProcessorWorker`'s `Future` impl now awaits sub-futures so replaced with async block for greater clarity.
* The [thrift](https://docs.rs/thrift) ~~and [grpcio](https://docs.rs/grpcio)~~ crates are not currently async, so some work still necessary to make jaeger and otlp exporters async (they currently call blocking code in a future, which is not great)


TommyCpp (commented):
I think grpcio generally support async(I could be wrong since I didn't use the async part before). But didn't find thrift async crate. May need to figure out how to make that work.

opentelemetry-otlp/src/span.rs
```diff
@@ -166,7 +166,7 @@ impl SpanExporter for Exporter {
             call_options = call_options.headers(metadata_builder.build());
         }
 
-        match self.trace_exporter.export_opt(&request, call_options) {
+        match self.trace_exporter.export_async_opt(&request, call_options) {
```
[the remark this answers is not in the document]
jtescher: oh interesting, didn't notice that, great!

jtescher:
@TommyCpp yeah may have to just use the thrift crate to write to a buffer synchronously and then send async

TommyCpp (approved):
LGTM :+1:

jtescher:
Ok going to see if making the jaeger exporter is a quick change, if not can do in a follow up PR

src/sdk/trace/span_processor.rs
```diff
@@ -268,12 +206,34 @@ impl BatchSpanProcessor {
...
+                        for batch in spans.chunks(config.max_export_batch_size) {
+                            exporter.export(batch).await;
+                        }
+                    }
+                    // Stream has terminated or processor is shutdown, return to finish execution.
+                    BatchMessage::Shutdown => {
+                        for batch in spans.chunks(config.max_export_batch_size) {
+                            exporter.export(batch).await;
+                        }
+                        break;
```
frigus02, line 232: Do we need to call `exporter.shutdown()` here?

src/sdk/trace/span_processor.rs
```diff
@@ -268,12 +206,34 @@ impl BatchSpanProcessor {
...
+                match message {
+                    // Span has finished, add to buffer of pending spans.
+                    BatchMessage::ExportSpan(span) => {
+                        if spans.len() < config.max_queue_size {
+                            spans.push(span);
+                        }
+                    }
+                    // Span batch interval time reached, export current spans.
+                    BatchMessage::Tick => {
+                        for batch in spans.chunks(config.max_export_batch_size) {
```
frigus02, line 223: Does this still remove exported spans from `spans`?

frigus02 (commented):
Looks really nice 👍

jtescher:
@frigus02 good catch, still toying with the jaeger exporter but will fix those as well 🙏

jtescher:
Ok switched jaeger collector client to use the `isahc` http client as it is executor agnostic, and manually maintain a buffer for the UDP agent and flush async. Could consider making the zipkin exporter executor agnostic in a future PR, but at least all the exporters are now non-blocking 🎉 .

I also removed the max packet size option for the agent exporter as I don't think you can actually configure the packet size in any of the implementations.

frigus02 (approved):
I'm not familiar with the Jaeger exporter, but as far as I understand these changes look nice.

Is it worth documenting, that users might want to enable the tokio or async-std feature when they're using the Jaeger agent?

Issue #223, opened 2020-09-21
Export `otel.status_code` tag in Zipkin exporter


jtescher:
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/967) for details (may not be exported currently).

danrusei:
can't the otel.status_code & otel.status_description be added to tags (into_zipkin_span function) ?

tags.insert("otel.status_code".into(), span_data.status_code.to_string());
tags.insert("otel.status_description".into(), span_data.status_message.clone());

Display has to be implemented for StatusCode enum.

jtescher:
@danrusei yeah should be easy to add, I don't think that anything else needs to format status codes that way, I believe the other exports use the numeric values. Could just add a function from `StatusCode`  -> `&'static str` that formats the way zipkin expects maybe?

jtescher:
Resolved in #236

PR #236, opened 2020-10-01
add otel.status_code and otel.status_message tags for Zipkin exporter

danrusei:
to address #223

### DIFF
--- a/opentelemetry-zipkin/src/model/mod.rs
+++ b/opentelemetry-zipkin/src/model/mod.rs
@@ -25,6 +25,29 @@ impl Into<annotation::Annotation> for api::Event {
     }
 }
 
+/// Converts StatusCode to str
+fn from_statuscode_to_str(status_code: api::StatusCode) -> &'static str {
+    match status_code {
+        api::StatusCode::OK => "OK",
+        api::StatusCode::Canceled => "CANCELLED",
+        api::StatusCode::Unknown => "UNKNOWN",
+        api::StatusCode::InvalidArgument => "INVALID_ARGUMENT",
+        api::StatusCode::DeadlineExceeded => "DEADLINE_EXCEEDED",
+        api::StatusCode::NotFound => "NOT_FOUND",
+        api::StatusCode::AlreadyExists => "ALREADY_EXISTS",
+        api::StatusCode::PermissionDenied => "PERMISSION_DENIED",
+        api::StatusCode::ResourceExhausted => "RESOURSE_EXHAUSTED",
+        api::StatusCode::FailedPrecondition => "FAILED_PRECONDITION",
+        api::StatusCode::Aborted => "ABORTED",
+        api::StatusCode::OutOfRange => "OUT_OF_RANGE",
+        api::StatusCode::Unimplemented => "UNINPLEMENTED",
+        api::StatusCode::Internal => "INTERNAL",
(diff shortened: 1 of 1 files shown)


opentelemetry-zipkin/src/model/mod.rs
```diff
@@ -25,6 +25,29 @@ impl Into<annotation::Annotation> for api::Event {
...
+        api::StatusCode::PermissionDenied => "PermissionDenied",
+        api::StatusCode::ResourceExhausted => "ResourceExhausted",
+        api::StatusCode::FailedPrecondition => "FailedPrecondition",
+        api::StatusCode::Aborted => "Aborted",
+        api::StatusCode::OutOfRange => "OutOfRange",
+        api::StatusCode::Unimplemented => "Unimplemented",
+        api::StatusCode::Internal => "InternalError",
+        api::StatusCode::Unavailable => "Unavailable",
+        api::StatusCode::DataLoss => "DataLoss",
+        api::StatusCode::Unauthenticated => "Unauthenticated",
```
frigus02, line 47: The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/zipkin.md) mentions that the status code should follow the [gRPC spec](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md), which formats them in uppercase. I assume that's what we should do as well, but I'm not too familiar with Zipkin. What do you think?

frigus02 (commented):
Thanks for the PR. This looks great.

opentelemetry-zipkin/src/model/mod.rs
danrusei: Good point. The GO implementation is using [codes](https://pkg.go.dev/google.golang.org/grpc/codes). The differences I noticed are on "UnknownError" vs "Unknown"  and " InternalError" vs "Internal".

frigus02: I just had a look at the other SDKs:

- The JavaScript SDK uses uppercase status code names. See [Zipkin exporter](https://github.com/open-telemetry/opentelemetry-js/blob/f731c65201ef49d91d8ec14ad329d9e485e2bcfb/packages/opentelemetry-exporter-zipkin/src/transform.ts#L81) and [CanonicalCode declaration](https://github.com/open-telemetry/opentelemetry-js/blob/f731c65201ef49d91d8ec14ad329d9e485e2bcfb/packages/opentelemetry-api/src/trace/status.ts).
- The Java SDK uses uppercase status code names, but they already implemented the status change away from gRPC status code equivalents to just unset, ok, error. See [Zipkin exporter](https://github.com/open-telemetry/opentelemetry-java/blob/64c2b8cc65a023bdf6f1e09bfc73bbe20168b627/exporters/zipkin/src/main/java/io/opentelemetry/exporters/zipkin/ZipkinSpanExporter.java#L157) and [CanonicalCode declaration](https://github.com/open-telemetry/opentelemetry-java/blob/64c2b8cc65a023bdf6f1e09bfc73bbe20168b627/api/src/main/java/io/opentelemetry/trace/Status.java#L45).
- And the Go SDK uses numbers?

danrusei: I guess you are right, this is not a long lasting change as the spec has changed. GO impl. has an open PR to migrate to new codes.

frigus02 (approved):

PR #238, opened 2020-10-02
Refactor TraceState for fewer allocations

jtescher:
Currently `TraceState` is a `VecDeque` because the spec has inserts moving the associated key to the front of the list. This is unfortunate as `VecDeque` will _always_ allocate (see https://github.com/rust-lang/rust/issues/68990).

This change addresses the issue by instead having `TraceState` store an `Option<VecDeque<_>>`, allowing the common case where there is no tracestate content to not heap allocate.


frigus02 (approved):
👍

PR #239, opened 2020-10-02
Ensure context guard is !Send

jtescher:
Currently you can hold a `ContextGuard` across await points accidentally via something like:

```rust
async {
    let cx = Context::current();
    let _g = cx.attach();
    some_future.await;
    // ...
}
```

This may cause inconsistent and buggy behavior as `ContextGuard`s rely on thread locals, this change causes the compiler to catch this case by marking the guard as `!Send` though a phantom type that is `!Send`.


frigus02 (approved):
Oh nice. I remember wondering about this a while back. Looks good 👍

PR #240, opened 2020-10-03
Ensure trace noop structs use `new`

jtescher:
Add private fields to trace module's noop structs to be consistent with this crate's conventions around backwards compatibility.


frigus02 (approved):
👍

TommyCpp (commented):
Looks good :+1:. This should work, but is there a reason why we don't use `#[non_exhaustive]` in this case?

jtescher:
@TommyCpp we don't want to allow matching at all, rather than merely non exhaustive matching, but this is more about the constructor specifically. By default constructors via `NoopSpan {}` are public, this change makes users go through `NoopSpan::new()` instead so if we want to add some state or otherwise change the internals this makes it not a breaking change.

PR #241, opened 2020-10-03
Use last value aggregator for value observers

jtescher:
* Switch to last value aggregator for value observers per https://github.com/open-telemetry/opentelemetry-specification/pull/984
* Add ddsketch to simple selector options
* Extract ddsketch config for use in selectors
* Add missing `Count` bound on `Histogram` aggregator trait.

### DIFF
--- a/benches/ddsketch.rs
+++ b/benches/ddsketch.rs
@@ -5,7 +5,7 @@ use opentelemetry::{
     api::metrics::Descriptor,
     sdk::{
         export::metrics::Quantile,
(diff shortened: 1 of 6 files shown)


jtescher:
@TommyCpp hope this `DDSkechConfig` makes sense for use by selectors, basically the selector will need to have a collection of these values to use for each aggregator so it seemed clearer to have this be its own struct.

TommyCpp (approved):
@jtescher Yeah it looks like a nice fix, Thanks!

PR #243, opened 2020-10-04
[exporter] Export instrument library information in Jaeger and Zipkin.

TommyCpp:
According to spec https://github.com/open-telemetry/opentelemetry-specification/pull/800 and https://github.com/open-telemetry/opentelemetry-specification/pull/801, We MUST export instrumentation library information for Jaeger and Zipkin exporter.

Also update Jaeger exporter so that the instrumentation version will be set to crate version.

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -142,6 +142,7 @@
 //! ```
 #![deny(missing_docs, unreachable_pub, missing_debug_implementations)]
 #![cfg_attr(test, deny(warnings))]
+
 mod agent;
 #[cfg(feature = "collector_client")]
 mod collector;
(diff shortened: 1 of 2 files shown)


jtescher (approved):
These look good, thanks @TommyCpp. Think it may be good to have an option to disable this in the pipeline builder as well, the values are constant and storage-conscious people may not want to have these on all recorded spans. Can do as a follow up pr later if you prefer.

TommyCpp:
@jtescher That's a good idea! I thought about it but the only concern I have was that spec seems pretty insist to include those in tags. Maybe this worth an issue in specifications?

jtescher:
@TommyCpp I think as long as this is the default it's fine to include an opt-out. For example if you are just using otel to send data to jaeger there is no need to include these if you don't want to as jaeger will not treat these values any differently.

jtescher:
Ok I'll merge this for now, can do the opt-out in a follow up

PR #244, opened 2020-10-05
Move propagators and ID generators

### LINKED ISSUE #235: Consider moving the propagators to from contrib to SDK
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/1043) and the [comment that explains it](https://github.com/open-telemetry/opentelemetry-specification/issues/637#issuecomment-644846847) for details.

awiede:
Resolves #235 

* Moves default `trace_context_propagator` implementation to `sdk`
* Moves all trace propagators in `contrib` to `sdk`
* Moves `id_generator` from `contrbi` to `sdk`


jtescher:
Looks good but let's wait for the discussion in https://github.com/open-telemetry/opentelemetry-specification/issues/1047 to settle before we go either way

awiede:
O whoops. Didn't realize it was still an open discussion...

jtescher:
@awiede actually it _does_ seem likely that b3 for example should be moved. Think we should just merge this and potentially move some of them back if that is the decision?

awiede:
That works. If we do have to move them back I'd like to nest them under a trace mod and export like in #255.

jtescher (approved):

jtescher:
@awiede looks good, just have to update the imports 😬

awiede:
Yikes! That was a nasty rebase. I think I got it all fixed now.

PR #245, opened 2020-10-05
Flush full batches instead of silently dropping

samschlegel:
Currently if we have an export rate higher than the buffer can hold for the given interval, we just lose spans which doesn't seem great. This changes the worker logic to flush when the buffer is full.

I believe it will probably have some conflicts with master due to the async changes.

### DIFF
--- a/src/sdk/trace/span_processor.rs
+++ b/src/sdk/trace/span_processor.rs
@@ -131,7 +131,7 @@ impl api::SpanProcessor for SimpleSpanProcessor {
 }
 
 /// A [`SpanProcessor`] that asynchronously buffers finished spans and reports
-/// them at a preconfigured interval.
+/// them when the batch is full or times out.
 ///
 /// [`SpanProcessor`]: ../../../api/trace/span_processor/trait.SpanProcessor.html
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


jtescher:
@samschlegel looks good, should be easy to rebase, or just apply your changes [here](https://github.com/open-telemetry/opentelemetry-rust/blob/master/src/sdk/trace/span_processor.rs#L215-L217) now.

jtescher:
There may be other questions around backpressure here if the intent is to have spans _never_ be dropped for load shedding purposes. If spans are being produced faster than the exporter can send, where should they buffer and should they buffer until OOM or should there be a cap elsewhere at which point spans start dropping? Currently for example you could set your `max_queue_size` to be very high and you shouldn't drop any spans, but if it _did_ happen to hit that high limit then it would start dropping rather than OOM and crash.

samschlegel:
Yeah I was thinking that as well right after I posted this. Perhaps what would be better instead is to add metrics so that we can track the buffer size and how many spans have been dropped. As is it's pretty silent, and in our services we hit that default queue size pretty quickly (which which works out to ~410 spans/sec) and I imagine others probably would as well.

jtescher:
@samschlegel yeah definitely folks interested in having more metrics awareness generally in tracing https://github.com/open-telemetry/opentelemetry-specification/issues/381 for example. There is also now a [global error handler](https://github.com/open-telemetry/opentelemetry-rust/blob/master/src/global/error_handler.rs#L14), it's currently used for errors in metrics but could (and should) be extended to trace errors, this could be a reasonable way to get a hook do do app-specific behavior.

samschlegel:
[open-telemetry/opentelemetry-specification#381](https://github.com/open-telemetry/opentelemetry-specification/issues/381) seems to be more about deriving metrics from traces, not adding traces to the internals of the SDK side of things? What I'd want is the span_processor to export internal metrics for the queue length that we could hookup through normal opentelemetry metrics means to export to prometheus.

The current metrics I'm thinking about are `current_queue_length`, `max_queue_length`, and `spans_dropped`. Can use the first two as an early warning/alerting system, and the third for knowing how much we've lost

jtescher:
@samschlegel that seems reasonable, but likely outside the scope of this PR. Should we close this an open an issue for span processor metrics?

PR #255, opened 2020-10-06
Trace module exports

### LINKED ISSUE #249: Remove `metrics` feature from default list
As tracing gets closer to GA it makes sense to signal that metrics (and logs in the future) are still unstable.

### LINKED ISSUE #254: Move trace exports to api/trace and sdk/trace level
Currently trace module types are flattened into the `api` and `sdk` modules directly. This may be confusing as we add metrics and log types, so consider moving them down a level into api/trace and sdk/trace.

awiede:
Resolves #254 Resolves #249 

* Moves trace exports from `api` and `sdk` to `api::trace` and `sdk::trace`
* Removes `metrics` from the default features

### DIFF
--- a/benches/trace.rs
+++ b/benches/trace.rs
@@ -4,7 +4,7 @@ use opentelemetry::{
         trace::{Span, Tracer, TracerProvider},
         Key,
     },
-    sdk,
+    sdk::trace as sdktrace,
 };
 
 fn criterion_benchmark(c: &mut Criterion) {
@@ -67,13 +67,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     });
 }
 
-fn trace_benchmark_group<F: Fn(&sdk::trace::Tracer)>(c: &mut Criterion, name: &str, f: F) {
+fn trace_benchmark_group<F: Fn(&sdktrace::Tracer)>(c: &mut Criterion, name: &str, f: F) {
     let mut group = c.benchmark_group(name);
 
     group.bench_function("always-sample", |b| {
-        let always_sample = sdk::trace::TracerProvider::builder()
-            .with_config(sdk::trace::Config {
-                default_sampler: Box::new(sdk::trace::Sampler::AlwaysOn),
+        let always_sample = sdktrace::TracerProvider::builder()
+            .with_config(sdktrace::Config {
+                default_sampler: Box::new(sdktrace::Sampler::AlwaysOn),
                 ..Default::default()
             })
             .build()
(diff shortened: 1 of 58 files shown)


src/api/mod.rs
```diff
@@ -32,22 +32,3 @@ pub use context::propagation::{
     text_propagator::TextMapFormat, Extractor, Injector,
 };
 pub use context::Context;
-
```
awiede, line 35: So `baggage` and `context` are feature flagged to trace but not under the `trace` module. Is there a strong opinion about whether these should be reorganized under the `trace` mod?

jtescher: Probably makes sense to keep the directory structure in line with https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-layout.md#api-package, but if features aren't used outside of trace then feature gating with `trace` probably makes sense. I think context is used by metrics though, so it should probably not be.

awiede: Ok - I'll leave `baggage` and `context` alone.

benches/trace.rs
```diff
@@ -64,13 +67,13 @@ fn criterion_benchmark(c: &mut Criterion) {
     });
 }
 
-fn trace_benchmark_group<F: Fn(&sdk::Tracer)>(c: &mut Criterion, name: &str, f: F) {
+fn trace_benchmark_group<F: Fn(&sdk::trace::Tracer)>(c: &mut Criterion, name: &str, f: F) {
```
jtescher, line 70: These are kinda unergonomic now with the double module prefix, but not sure how to avoid that as they share names with the traits. Any ideas here? or is this the best we can do currently.

jtescher: Looks like go uses the equivalent of:

```rust
use opentelemetry::sdk::trace as sdktrace;
// ...
let tracer: sdktrace::Tracer;

// or
 
use opentelemetry::sdk::trace::Tracer as SdkTracer;
// ...
let tracer: SdkTracer;
```

Alternatively we could have a convention like our examples always import `api::Tracer as _` and `Tracer` is always in reference to sdk, or vice versa.

awiede: I mean, we _could_ `pub use` it in `src/sdk/mod.rs`, but then we go back to the inconsistency. It's weird, but if we flatten this into `sdk`, then do we do the same for any type that is prefixed with `Trace*`? And then we're back where we started. 🤷

awiede: Yeah, I think example aliases are probably good for now.

awiede: @jtescher updated examples, doc tests and tracing bench to use the alias.

jtescher: Ok cool, the aliases are not breaking changes so looks good to me even if we want to experiment here

jtescher (approved):

PR #257, opened 2020-10-07
Prepare for 0.9.0 release

jtescher:


awiede (approved):

TommyCpp:
The only concern I have is the `opentelemetry-zipkin` is currently broken with anything other than tokio. So not sure if we want to update that too

jtescher:
Yeah just getting this ready, should probably have the current PR's merged first, I can keep this updated as they come in.

PR #259, opened 2020-10-07
Allow users to choose http client in zipkin and datadog exporter

### LINKED ISSUE #248: Expoters uses reqwest panic without tokio
As [the issue](https://github.com/seanmonstar/reqwest/issues/719) points out, reqwest relys on tokio to work. Currently, any expoters with reqwest as http client will panic with following error without tokio (For example, the datadog and zipkin expoters example are broken).
```
there is no reactor running, must be called from the context of Tokio runtime
```
Possible solution could be move away from reqwest or try to implement async and sync uploader with non-blocking and blocking client (But I guess this doesn't really solve the problem to running those expoters with `async-std`)

TommyCpp:
Reqwest is built on tokio. Without tokio, it will just not work. So move to isahc so that we could run in sync, tokio, and other async runtime.

Fixes #248


opentelemetry-contrib/src/datadog/mod.rs
```diff
@@ -149,11 +146,9 @@ impl Default for DatadogPipelineBuilder {
 impl DatadogPipelineBuilder {
     /// Create `ExporterConfig` struct from current `ExporterConfigBuilder`
     pub fn install(mut self) -> Result<(sdk::trace::Tracer, Uninstall), Box<dyn Error>> {
-        let exporter = DatadogExporter::new(
-            self.service_name.clone(),
-            self.agent_endpoint.parse()?,
-            self.version,
-        );
+        let endpoint = self.agent_endpoint + self.version.path();
```
TommyCpp, line 149: The isahc's URL parser cannot modify the `Uri`, its `Uri builder` doesn't really support parse from string. We could introduce some other url parser to help validate and modify the endpoint but that feels like a little bit overhead.

jtescher: I think concatenating here is fine, only happens once, and the URI is still parsed below so it's a valid URI after install

opentelemetry-contrib/src/datadog/mod.rs
```diff
@@ -200,13 +195,15 @@ impl trace::SpanExporter for DatadogExporter {
             Err(_) => return trace::ExportResult::FailedNotRetryable,
         };
 
-        let resp = self
-            .client
-            .post(self.request_url.clone())
-            .header(CONTENT_TYPE, self.version.content_type())
+        let req = match Request::builder()
```
frigus02, line 198: I haven't tried it, yet, but I feel like this is missing the method and url. Should this be something like this?
```rust
let req = match Request::post(self.request_url.clone())
    .header("content-type", self.version.content_type())
    .body(data)
```

frigus02 (commented):
Looks good overall 👍

opentelemetry-contrib/src/datadog/mod.rs
TommyCpp: Good catch! I guess I forget it somehow. Fixed.

frigus02 (approved):

TommyCpp:
Also wonder if there is a way for us to test datadog exporter without datadog account/subscription. Or maybe we should have some mock datadog receivers.

MikailBag:
Does it mean `libcurl` will be required when building and running?

frigus02:
Good question. I think the `static-curl` feature is on by default in `isahc`, which should mean `libcurl` is not required for build nor run. It should to build from a bundled version and link libcurl statically. 🤔

MikailBag:
But C++ compiler is still required, isn't it?
Also, AFAIU, libcurl will probably spawn some background threads etc.
Did you consider using e.g. implicit Tokio runtime by default instead?
I mean roughly:
```rust
// Lazy from `once-cell` or `lazy-static` crate
static RUNTIME: Lazy<tokio::runtime::Runtime> = Lazy::new(||{
    // build a runtime here
})

// ...
impl DatadogExporter {
    pub fn new() {
        let rt_handle = tokio::runtime::Handle::try_current().unwrap_or_else(|| RUNTIME.handle());
        rt_handle.enter(|| {
              // here runtime is definitely created and available
        })
    }
}
```

I should add I don't have strong motivation for either approach (to be honest I'd like user to manually construct runtime because only user can decide on its settings), I just want to share my thoughts.

hawkw:
> if the user is using tide or otherwise in the async-std ecosystem, they should ideally be able to run exporters without having to spin up a tokio runtime. isahc seems like the most maintained executor agnostic http client crate currently.

Right, I think what @MikailBag was pointing out is that `libcurl` _will_ still spin up an async threadpool in the background, it's just written in C and used only by `libcurl`. 

In practice, this is the same as just bundling a `tokio` runtime for `reqwest` to use, except that it also means that a C compiler and libcurl are now dependencies. In some ways, this is  probably _worse_ than using an HTTP client that's tightly coupled to any Rust async runtime — it won't use the user's async threadpool regardless of whether it's `tokio`, `async-std`, `bastion`, or something else. 

I think the only truly runtime-agnostic approach would be some way to allow the user to pass in an arbitrary HTTP client (maybe by implementing some trait for the various existing HTTP client libraries?). This might take more work than just switching implementations, but it would be the only way to ensure that the user can run the HTTP client on the runtime of their choosing, rather than creating a separate runtime for it.

jtescher:
@hawkw yeah that makes sense, shouldn't be too difficult to accept a client if we want to go down that path as the usage here is relatively modest (single post request)

frigus02:
Yep, makes sense. I wonder if something like https://github.com/http-rs/http-client would be a quick win, since it can be configured to use different client implementations.

TommyCpp:
Try to introduce the http_client traits into zipkin. If it looks good, will continue with datadog exporter.

opentelemetry-zipkin/src/lib.rs
```diff
@@ -92,29 +92,35 @@ mod uploader;
...
+use std::ops::Deref;
 
 /// Default Zipkin collector endpoint
 const DEFAULT_COLLECTOR_ENDPOINT: &str = "http://127.0.0.1:9411/api/v2/spans";
 
 /// Default service name if no service is configured.
 const DEFAULT_SERVICE_NAME: &str = "OpenTelemetry";
 
+lazy_static::lazy_static! {
+    static ref DEFAULT_NOOP_HTTP_CLIENT: NoopHttpClient = NoopHttpClient::default();
```
jtescher, line 109: Having the default behavior be a blocking reqwest client may be preferable here. I think having a no-op would lead to a lot of users being confused why their spans are not showing up in zipkin (although it does mean a default reqwest dependency)

opentelemetry-zipkin/src/lib.rs
```diff
@@ -167,6 +175,12 @@ impl ZipkinPipelineBuilder {
         self
     }
 
+    /// Assign client implementation
+    pub fn with_client<T: http_client::HttpClient + Send + Sync>(mut self, client: &'static T) -> Self {
```
jtescher, line 179: I think an owned value would be simpler here than static refs. E.g.

```rust
pub fn with_client<T: http_client::HttpClient + Send + Sync + 'static>(mut self, client: T) -> Self {
    self.client = Box::new(client);
}
```

jtescher (changes requested):
Good start @TommyCpp couple of ideas

opentelemetry-zipkin/src/lib.rs
jtescher: Sorry forgot this is now an async api, if you're going with http_client this should probably be hyper as they impl it already.

TommyCpp: They do have a http client called H1 comes with the `http_client`. I also tried how easy it would be to implement `http_client::Client` for reqwest http clients. It actually takes quite some time even only accpet post requests.

jtescher:
Taking a step back, not sure that https://github.com/http-rs/http-client is necessary as it depends on async-std (through http-types). 

Would a simpler solution be a small trait that could easily be impled for arbitrary clients? Something like https://github.com/jtescher/opentelemetry-rust-1/commit/a59efe9000aaa50cc1aac38cbb2696fbb0c8efb7

The above commit adds support for reqwest and surf (could add others) and basically switches things to use:

```rust
#[async_trait]
pub trait HttpClient {
    async fn send(&self, request: http::Request<String>) -> Result<ExportResult, Box<dyn std::error::Error>>;
}
```

frigus02:
Yeah, I was wondering if a custom trait might be better.

`http-client` can be nice because of all the clients it implements (`h1_client` with async-std, `hyper_client` with tokio). We could enable one of the clients using a feature flag, which means users don't necessarily write code. But that does come at the cost of an additional dependency.

I'm happy either way.

examples/zipkin/src/main.rs
```diff
@@ -10,6 +10,7 @@ fn bar() {
     span.end()
 }
 
+
```
jtescher, line 13: Probably need to run cargo fmt

opentelemetry-otlp/Cargo.toml
```diff
@@ -21,7 +21,7 @@ async-trait = "0.1"
 futures = "0.3.5"
 grpcio = "0.6"
 opentelemetry = { version = "0.8.0", default-features = false, features = ["trace"], path = ".." }
-protobuf = "~2"
+protobuf = "2.18.0"
```
jtescher, line 24: Could probably have this be `2.18` to match the constraints we use elsewhere.

opentelemetry-zipkin/Cargo.toml
```diff
@@ -15,10 +15,19 @@ keywords = ["opentelemetry", "zipkin", "tracing", "async"]
...
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+reqwest = { version = "0.10", features = ["blocking"], optional = true }
+surf = { version = "2.0", optional = true }
 typed-builder = "0.5"
+lazy_static = "1.4.0"
+http = "0.2"
+
+[dev-dependencies]
+isahc = "0.9"
```
jtescher, line 33: newline fmt

opentelemetry-zipkin/src/lib.rs
```diff
@@ -46,6 +46,13 @@
 //! [`tokio`]: https://tokio.rs
 //! [`async-std`]: https://async.rs
 //!
+//! ## Bring your own http clients
+//! Users can choose appropriate http clients to align with their runtime. By default, the opentelemetry-zipkin will use blocking reqwest http client. Other than that, opentelemtry-zipkin also provide support for async reqwest http client surf http client.
```
jtescher, line 50: nit: can probably wrap these to 80 chars

opentelemetry-zipkin/src/lib.rs
```diff
@@ -140,6 +177,7 @@ impl Default for ZipkinPipelineBuilder {
             service_addr: None,
             collector_endpoint: DEFAULT_COLLECTOR_ENDPOINT.to_string(),
             trace_config: None,
+            client: Box::new(reqwest::blocking::Client::new()),
```
jtescher, line 180: May want to use the optional version that I had and use the features to set defaults if present, currently this will not compile without the `reqwest` feature.

opentelemetry-zipkin/src/lib.rs
```diff
@@ -187,7 +231,7 @@ impl ZipkinPipelineBuilder {
 }
 
 #[async_trait]
-impl trace::SpanExporter for Exporter {
+impl<'a> trace::SpanExporter for Exporter {
```
jtescher, line 234: don't need this lifetime anymore

jtescher (changes requested):

opentelemetry-zipkin/src/lib.rs
TommyCpp: I think I made the `reqwest` as default feature to make sure builder will always have some kind of http client. But think it again, it still can be broken when user override the default features. So it make sense to have an option here. Will fix it.

jtescher:
@TommyCpp zipkin looking good, want to do the same for datadog?

TommyCpp:
> @TommyCpp zipkin looking good, want to do the same for datadog?

Absolutely. But I think their implementations will roughly be the same. So I am trying to figure out if we should put the http client trait in main crate so that users don't need to implement the same thing twice.

jtescher:
@TommyCpp hm yeah maybe in exporter::trace alongside `SpanExporter`?

TommyCpp:
I think I accidently override the change I made when merging the upstream changes :disappointed:. Will revisit it tonight.

jtescher:
@TommyCpp yeah was going to comment on the dependency issues, let me know when you're ready for another look.

opentelemetry-contrib/src/datadog/mod.rs
```diff
@@ -177,6 +249,15 @@ impl DatadogPipelineBuilder {
         self
     }
 
+    /// Choose the http client used by uploader
+    pub fn with_http_client<T: HttpClient + 'static>(
```
cpcloud, line 253: This method would have to go away if you go down the generic route.

cpcloud (commented):
Just had a couple passing thoughts after tinkering with this PR.

jtescher:
@cpcloud adding a generic is certainly a possibility, I think the constraints are to provide the most ergonomic initialization ideally without tying to a specific executor or http client. Little less worried about heap allocations here as they would typically occur only once when the process starts, but the more difficult question for datadog is how to best provide the default impl while allowing the user to bring their own, i.e. what type does `new_pipeline` return if no default features are selected?

jtescher:
@TommyCpp I'm not sure the default logic here works for datadog as the `reqwest-blocking` feature is _always_ present given it is a dependency of the `datadog` feature itself. Could drop the default support and make users chose a feature or provide a client themselves (alternatively could move the logic around and assume the feature is always there but ignore it if others are there, but that would still make reqwest a required dependency which some users may not prefer).

opentelemetry-contrib/src/datadog/mod.rs
```diff
@@ -100,19 +140,24 @@ const DEFAULT_SERVICE_NAME: &str = "OpenTelemetry";
...
-        let mut request_url = agent_endpoint;
-        request_url.set_path(version.path());
+    fn new(
+        service_name: String,
+        agent_endpoint: Uri,
+        version: ApiVersion,
+        client: Box<dyn HttpClient>,
+    ) -> Self {
+        let request_url = agent_endpoint;
+        // request_url.set_path(version.path());
```
jtescher, line 157: Is this path not needed anymore?

TommyCpp: I think we already set path to version path [here](https://github.com/open-telemetry/opentelemetry-rust/pull/259/files#diff-3a90e7a53d74def8021c417e499b39296a43bca32ef38d94e0131676b2f8a804R213-R220). But yeah I should probably just use the param

opentelemetry-contrib/src/datadog/mod.rs
```diff
@@ -100,19 +140,24 @@ const DEFAULT_SERVICE_NAME: &str = "OpenTelemetry";
 /// Datadog span exporter
 #[derive(Debug)]
 pub struct DatadogExporter {
-    client: reqwest::Client,
-    request_url: Url,
+    client: Box<dyn HttpClient>,
```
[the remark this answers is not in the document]
TommyCpp: This looks really nice, thanks for the effort. The reason why we choose to use conditional compilation in the first place is we want to eliminate the dependencies as much as possible and let user enable dependencies only necessary. Most of the http client in rust relay on some kind of native library, for example, `reqwest` relay on certain version of `OpenSSL` on Linux. 

If we use `reqwest` as default. Then I believe users will have to install all required library of `reqwest` to compile. So we want to offer users a chance to opt out and choose other create/library.

jtescher (approved):
LGTM

jtescher:
@TommyCpp want to push an empty commit? looks like the CLA bot got stuck

jtescher:
@TommyCpp @cpcloud Ok will merge this to unblock the 0.9.0 release. We can continue to work on the usability and perf in subsequent PRs as breaking changes in exporters are less consequential 👍

PR #261, opened 2020-10-10
Update OTLP proto

jtescher:


TommyCpp:
Rust updated to [1.47.0](https://blog.rust-lang.org/2020/10/08/Rust-1.47.html), which seems to bring new warning items that's why CIs are breaking.

TommyCpp (approved):
Rust updated to [1.47.0](https://blog.rust-lang.org/2020/10/08/Rust-1.47.html), which seems to bring new warning items that's why CIs are breaking. Otherwise LGTM

PR #262, opened 2020-10-10
Fix warnings based on updated clippy

TommyCpp:
Rust recently released the new version, which includes some new warnings. This PR fixed those warnings


jtescher (approved):
Looks good!

PR #264, opened 2020-10-11
Send owned span data to exporters

jtescher:
As most exporters copy the majority of the exported span data, switch API to send owned data to exporters.

The export method now becomes:

```diff
- async fn export(&self, batch: &[Arc<trace::SpanData>]) -> trace::ExportResult
+ async fn export(&self, batch: Vec<trace::SpanData>) -> trace::ExportResult 
```

Additional small tweaks / bug fixes encountered during the refactor:

* Add missing `tokio` feature to async example's jaeger dependency
* Switch to custom `TBufferChannel` impl in jaeger exporter because the default buffer size is fixed and unknown at compile time.
* add trace_state impl to otlp exporter
* Expose dropped count on `EvictedQueue` and add to otlp exporter

### DIFF
--- a/src/sdk/trace/span.rs
+++ b/src/sdk/trace/span.rs
@@ -154,9 +154,9 @@ impl Drop for SpanInner {
                             data.end_time = SystemTime::now();
                         }
                     }
-                    let last_processor_idx = provider.span_processors().len().saturating_sub(1);
-                    for (idx, processor) in provider.span_processors().iter().enumerate() {
-                        let span_data = if idx == last_processor_idx {
+                    let mut processors = provider.span_processors().iter().peekable();
+                    while let Some(processor) = processors.next() {
+                        let span_data = if processors.peek().is_none() {
                             // last loop or single processor/exporter, move data
                             span_data.take()
                         } else {
--- a/src/sdk/trace/span.rs
+++ b/src/sdk/trace/span.rs
@@ -23,7 +23,7 @@ pub struct Span {
 /// Inner data, processed and exported on drop
 #[derive(Debug)]
 struct SpanInner {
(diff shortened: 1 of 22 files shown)


src/sdk/trace/span.rs
```diff
@@ -142,14 +146,27 @@ impl Drop for SpanInner {
...
                 if let Some(provider) = self.tracer.provider() {
-                    for processor in provider.span_processors() {
-                        processor.on_end(exportable_span.clone())
+                    // Set end time if unset or invalid
+                    if let Some(data) = span_data.as_mut() {
+                        if data.end_time <= data.start_time {
+                            data.end_time = SystemTime::now();
+                        }
+                    }
+                    let last_processor_idx = provider.span_processors().len().saturating_sub(1);
```
jtescher, line 157: Seems like there should be a simpler way to do this, basically in the single processor/exporter case this should just move the span data, if there are multiple it should clone all but the last iteration.

TommyCpp: Maybe use `peek()` to determine whether it's the last element? Something like 
```rust
                    let mut it = provider.span_processors().iter().peekable();
                    while let Some(processor) = it.next(){
                        let span_data = if it.peek().is_some() {
                            // not last
                            span_data.clone()
                        } else {
                            // last
                            span_data.take()
                        };
                        if let Some(span_data) = span_data {
                            processor.on_end(span_data);
                        }
                    }
```

jtescher: Yeah that's probably cleaner, meant more the pattern of "move on last iteration, clone on others" but seems like this is the simplest way of accomplishing that.

TommyCpp (approved):
LGTM:+1:

PR #266, opened 2020-10-12
Move propagators to sdk::propagation module

jtescher:
Currently the propagation modules are mixed in a confusing way (e.g. baggage propagator is in api, but trace context is in sdk). This change standardizes the propagator locations by having propagation formats in `api::propagation` (only text map currently), keeps global propagation utilities in `global::propagation`, and moves propagator implementations to `sdk::propagation`.

High level changes:

* Move propagators from `sdk::trace::propagator` to `sdk::propagation`
* Rename `TextMapFormat` to `TextMapPropagator` to match the spec.
* Rename W3CPropagator back to `TraceContextPropagator` to match the spec and because the baggage propagator is also a w3c propagator.
* Remove `TextMapFormat` from examples because they should be setting/getting the global so they interact properly with middleware.


src/sdk/propagation/mod.rs
```diff
@@ -0,0 +1,12 @@
+//! OpenTelemetry Propagators
+pub mod aws;
+mod b3;
+mod baggage;
+mod composite;
+pub mod jaeger;
```
awiede, line 6: Why not export the `jaeger` and `aws` propagators like the others?

jtescher: ah yeah had a feature flag at one point but remembered we're not doing that, can fix

awiede (approved):
LGTM

PR #275, opened 2020-10-16
Update semantic conventions

jtescher:
The regex is not a little more complicated as the semantic conventions can now have links, as and typically have a `type` as the second column.

### DIFF
--- a/opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
+++ b/opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
@@ -20,7 +20,8 @@ md_to_rs() {
 		# Regex explanation:
 		# Find markdown tables (lines starting with "|") and take the first 2 columns
 		# - The 1st column is the attribute name consisting of a-z0-9._
-		# - The 2nd column is the description
+		# - The 2nd column is (usually) the type
+		# - The 3rd column is (usually) the description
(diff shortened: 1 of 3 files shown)


opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
```diff
@@ -21,10 +21,10 @@ md_to_rs() {
 		# Find markdown tables (lines starting with "|") and take the first 2 columns
 		# - The 1st column is the attribute name consisting of a-z0-9._
 		# - The 2nd column is the description
-		if [[ "$line" =~ ^\|\ \`?([a-z0-9._]+)\`?\ +\|\ ([^|]+) ]]; then
+		if [[ "$line" =~ ^\|\ \[?\`?([a-z0-9._]+)\`?\]?(:?\([^|]+\))?\ +\|\ ([^|]+)\ +\|\ ([^|]+) ]]; then
```
frigus02, line 24: I think we should either update the comment about the regex in the line above or remove it, if it's not helpful. 🙂

frigus02 (approved):

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
jtescher: Yeah good idea

PR #278, opened 2020-10-17
Fix single threaded runtime tokio feature bug

### LINKED ISSUE #277: Batch span exporter does not shut down properly with single threaded tokio runtime
This blocking call will stop progress on the future it is waiting on if the `tokio` feature is used with a single threaded runtime, which results in the process not shutting down. https://github.com/open-telemetry/opentelemetry-rust/blob/3417e22547aacf285e53c56a5419df06e29249fe/src/sdk/trace/span_processor.rs#L139

jtescher:
* Add `tokio` feature to actix example which uses the single threaded tokio runtime
* switch from `tokio::spawn` back to `|fut| tokio::task::spawn_blocking(|| futures::executor::block_on(fut))` to fix the processor shutdown bug in single threaded runtimes
* Fix `async-std` import (we should add more tests that exercise the full feature matrix)
* Only block waiting on the pan processor future if sending shutdown was successful.

Fixes #277

### DIFF
--- a/examples/actix-http/Cargo.toml
+++ b/examples/actix-http/Cargo.toml
@@ -4,8 +4,8 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-opentelemetry = { path = "../../" }
-opentelemetry-jaeger = { path = "../../opentelemetry-jaeger", features = ["collector_client"] }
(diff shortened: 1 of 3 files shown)


frigus02 (approved):
Nice catch. Changes make sense to me

PR #280, opened 2020-10-17
Allow metrics instruments to be cloned

jtescher:
Allow metrics to be cloned and synchronized (e.g. for use in actix-web workers)


frigus02 (approved):
I've not used metrics, yet, so I'm not sure why this is needed. I also don't see that it harms anything. So 👍 🙂

PR #283, opened 2020-10-18
Small development improvements

cpcloud:
1. Add `*/target` to `.gitignore` to prevent tracking of workspace cargo output
1. Add `rustfmt.toml` to allow `rustfmt` to automatically work with `async fn`s.


rustfmt.toml
```diff
@@ -0,0 +1 @@
+edition = "2018"
```
frigus02, line 1: I've never needed to use that, so had to check when/why it's necessary.

> Rustfmt is able to pick up the edition used by reading the Cargo.toml file if executed through the Cargo's formatting tool `cargo fmt`. Otherwise, the edition needs to be specified in your config file.

So it's useful if you're running rustfmt directly. Sounds good to me.

frigus02 (approved):

rustfmt.toml
cpcloud: I have my editor set up to format on save, and the plugin I'm using invokes `rustfmt` directly, so this comes up :)

PR #285, opened 2020-10-18
Update mapping otel events to Jaeger logs attributes

zoidbergwill:
I'm pretty new to Rust, so I struggled with this a bunch more than I probably should have, but I managed to get it compiling and working. I tested it by fiddling with the basic example adding some `event` and `dropped_attributes_count` attributes, and they seemed to all do the right things. It would be amazing to have tests for them though.


jtescher:
@zoidbergwill I think your confusion makes sense, I'm reading over the spec now and noticing that it mentions `dropped_attributes_count` but does not actually mention it anywhere in the trace spec so the source is a little confusing. If it is the span's dropped attribute count it probably makes more sense as a tag than a log message, and if it is the event's dropped attribute count then the spec probably needs to have limits on event attribute length. May be best to not worry about implementing that one yet.

Setting the log's `event` tag value from the event's name or an attribute's value seems more straightforward. I think you're approach will have the correct result, but you can make some efficiency improvements by allocating fewer strings. E.g. instead of converting all the otel event attributes into jaeger tags first and then extracting the name and removing / adding tags, you could probably note while building up your vec of jaeger tags whether the `event` attribute was set, and only override it if it has not yet been assigned. Something along the lines of:

```rust
let mut event_set_via_attribute = false;
let mut fields: std::vec::Vec<jaeger::Tag> = self
    .attributes
    .into_iter()
    .map(|attr| {
        if attr.key.as_str() == "event" {
            event_set_via_attribute = true;
        };
        attr.into()
    })
    .collect::<Vec<_>>();

// Use event name as log event key, unless there is an event attribute
if !event_set_via_attribute {
    fields.push(api::Key::new("event").string(self.name).into());
}
```

zoidbergwill:
@jtescher Thanks for the quick review!

I dropped all the changes about the `dropped_attributes_count`, because I agree, it's weird that there is no another mention of it in the specification repo.

I also updated checking for the `event` attribute to happen during that first `.map`, that simplifies the code a bunch, and is also better for performance.

opentelemetry-jaeger/src/lib.rs
```diff
@@ -436,12 +436,21 @@ impl Into<jaeger::Log> for api::trace::Event {
             .duration_since(SystemTime::UNIX_EPOCH)
             .unwrap_or_else(|_| Duration::from_secs(0))
             .as_micros() as i64;
-        let mut fields = self
+        let mut event_set_via_attribute = false;
+        let mut fields: std::vec::Vec<jaeger::Tag> = self
```
jtescher, line 440: Small nit: this doesn't actually need the explicit type anymore

```suggestion
        let mut fields = self
```

jtescher (commented):
Looks good, is this ready to remove draft status?

frigus02 (approved):

Issue #260, opened 2020-10-09
Handle baggage metadata as opaque type


jtescher:
See  https://github.com/open-telemetry/opentelemetry-specification/pull/1066 for details.

TommyCpp:
Gonna take it if no one else is currently working on it 😄

PR #287, opened 2020-10-18
[propagation] Add support for baggage metadata/prop.

TommyCpp:
W3 baggage spec allows users to pass properties along with baggage. This commit introduces metadata into `KeyValueMetadata` type. Propagators that supports baggage metadata could use it to propagate data.

For example, user can use `insert_with_metadata` to add an entry into the baggage.
```
let mut cc = Baggage::new();
let _ = cc.insert_with_metadata("my-name", "my-value", "test");
```

Or if the propagation the users uses doesn't support metadata. Users can also use `insert`
```
let mut cc = Baggage::new();
let _ = cc.insert_with("my-name", "my-value");
```

Internally metadata is just an opaque string wrapper with no semantic meanings.
We will assign a metadata for every entry in baggage.

Note that since the Metadata currently has no semantic meanings. We will just take the metadata string from header without any validation and process.

Closes https://github.com/open-telemetry/opentelemetry-rust/issues/260


src/api/core.rs
```diff
@@ -202,6 +202,30 @@ fn format_value_array_as_string(v: &[Value]) -> String {
...
+pub struct Metadata(String);
+
+impl Metadata {
+    /// Return underlying string
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl Into<Metadata> for String {
```
frigus02, line 217: Should we instead implement `From<String> for Metadata`?

src/api/baggage.rs
```diff
@@ -112,35 +162,53 @@ impl FromIterator<(Key, Value)> for Baggage {
...
-    fn current_with_baggage<T: IntoIterator<Item = KeyValue>>(baggage: T) -> Self;
+    fn with_cleared_baggage(&self) -> Self;
 
+    /// Returns a reference to this context's baggage, or the default
+    /// empty baggage if none has been set.
+    fn baggage(&self) -> &Baggage;
+}
+
+/// Method to retrieve key value pairs, possible along with the metadata.
+pub trait AddBaggage<I> {
```
frigus02, line 205: I think you could get away without the extra `AddBaggage` trait and merge everything into the `BaggageExt` trait using `I: Into<KeyValueMetadata>`. That works for both iterators over `KeyValue` and `KeyValueMetadata`. Longer example:

```rust
pub trait BaggageExt {
    fn with_cleared_baggage(&self) -> Self;
    fn baggage(&self) -> &Baggage;
    fn with_baggage<T: IntoIterator<Item = I>, I: Into<KeyValueMetadata>>(&self, baggage: T) -> Self;
    fn current_with_baggage<T: IntoIterator<Item = I>, I: Into<KeyValueMetadata>>(baggage: T) -> Self;
}

impl BaggageExt for Context {
    // ...

    fn with_baggage<T: IntoIterator<Item = I>, I: Into<KeyValueMetadata>>(&self, baggage: T) -> Self {
        let merged: Baggage = self
            .baggage()
            .iter()
            .map(|(key, (value, metadata))| {
                KeyValueMetadata::new(key.clone(), value.clone(), metadata.clone())
            })
            .chain(baggage.into_iter().map(|kvs| kvs.into()))
            .collect();

        self.with_value(merged)
    }

    fn current_with_baggage<T: IntoIterator<Item = I>, I: Into<KeyValueMetadata>>(baggage: T) -> Self {
        Context::current().with_baggage(baggage)
    }
}
```

src/api/baggage.rs
```diff
@@ -150,46 +218,69 @@ pub trait BaggageExt {
...
+}
+
+impl AddBaggage<KeyValueMetadata> for Context {
+    fn with_baggage<T: IntoIterator<Item = KeyValueMetadata>>(&self, kvm: T) -> Self {
         let merged: Baggage = self
             .baggage()
             .iter()
-            .map(|(key, value)| KeyValue::new(key.clone(), value.clone()))
-            .chain(kvs.into_iter())
+            .map(|(key, (value, metadata))| {
```
frigus02, line 255: I don't particularly like the nested tuple from the iterator here. From the perspective or `KeyValue` it makes sense, because it maps a key to a (value, metadata) pair. But using this here feels weird. I may prefer this to be a flat tuple with 3 items (key, value, metadata). But that might just be me and I'm not super sure about this, either.

frigus02 (commented):
The logic looks good. Just left a few suggestions about the API.

src/api/baggage.rs
TommyCpp: Good advice! It looks more elegant.

src/api/baggage.rs
TommyCpp: I think have `key, (value, metadata)` give some clearity around which part is the key and which part is the value. A ternary `a,b,c` could be group in either `(a,b),c` or `a, (b,c)`. But it should be a easy change so we can also make it we'd like to

src/api/core.rs
```diff
@@ -202,6 +202,30 @@ fn format_value_array_as_string(v: &[Value]) -> String {
     )
 }
 
+/// `Metadata` uses by `KeyValue` pairs to represent related metadata.
+#[cfg_attr(feature = "serialize", derive(Deserialize, Serialize))]
+#[derive(Clone, Debug, PartialOrd, PartialEq, Default)]
+pub struct Metadata(String);
```
jtescher, line 208: As this is only used in baggage, maybe it would add clarity to name this `BagggageMetadata` and move it to the baggage module?

src/api/core.rs
```diff
@@ -226,6 +250,44 @@ impl KeyValue {
     }
 }
 
+/// `KeyValue` pairs with metadata
+#[cfg_attr(feature = "serialize", derive(Deserialize, Serialize))]
+#[derive(Clone, Debug, PartialEq)]
+pub struct KeyValueMetadata {
```
jtescher, line 256: This could probably also be in baggage

src/sdk/propagation/baggage.rs
```diff
@@ -13,7 +13,7 @@
 //! # Examples
 //!
 //! ```
-//! use opentelemetry::api::{BaggageExt, Key, propagation::TextMapPropagator};
+//! use opentelemetry::api::{Key, propagation::TextMapPropagator, BaggageExt};
```
jtescher, line 16: I think this can be reverted (rustfmt would have them sorted alphabetically)

src/api/core.rs
```diff
@@ -202,6 +202,30 @@ fn format_value_array_as_string(v: &[Value]) -> String {
...
+impl Metadata {
+    /// Return underlying string
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl From<String> for Metadata {
+    fn from(s: String) -> Metadata {
+        Metadata(s)
```
jtescher, line 219: Should trim here I guess?

> Leading and trailing OWS is allowed but MUST be trimmed when converting the header into a data structure.

src/api/core.rs
```diff
@@ -202,6 +202,30 @@ fn format_value_array_as_string(v: &[Value]) -> String {
...
+    }
+}
+
+impl From<&str> for Metadata {
+    fn from(s: &str) -> Self {
+        Metadata(s.to_string())
```
jtescher, line 225: Same

jtescher (changes requested):

src/api/baggage.rs
```diff
@@ -56,12 +73,44 @@ impl Baggage {
...
+        key: K,
+        value: V,
+        metadata: S,
+    ) -> Option<(Value, Metadata)>
+    where
+        K: Into<Key>,
+        V: Into<Value>,
+        S: Into<Metadata>,
+    {
+        self.inner
```
jtescher, line 107: `insert_with_metadata` and `insert` should both probably check https://w3c.github.io/baggage/#limits as well

src/api/core.rs
TommyCpp: Yeah I think that make sense. Will make the changes.

src/sdk/propagation/baggage.rs
TommyCpp: Must be something I changed after running rustfmt. Should really have enabled auto format on save 😞

jtescher: Oh rustfmt can't reformat code in doc comments yet unfortunately (also I think it would actually have modules first, so maybe move Key to the end instead?)

src/api/core.rs
TommyCpp: Hmm maybe we should add the trim logic in `TextMapPropagator`.  Also maybe not store the separator(`;`) in metadata. The reason is I think Baggage should be format agnostic, which means we shouldn't introduce format conversion logic inside metadata, those should be in propagator of that format.

src/api/baggage.rs
TommyCpp: Good catch. Will add those checks

src/api/baggage.rs
```diff
@@ -1,22 +1,30 @@
...
+const MAX_KEY_VALUE_PAIRS: usize = 180;
+const MAX_BYTES_FOR_ONE_PAIR: usize = 4096;
+const MAX_LEN_OF_ALL_PAIRS: usize = 8192;
+
 /// A set of name/value pairs describing user-defined properties across systems.
 #[derive(Debug, Default)]
 pub struct Baggage {
-    inner: HashMap<Key, Value>,
+    inner: HashMap<Key, (Value, BaggageMetadata)>,
+    len: usize,
```
jtescher, line 19: May want to clarify that this is not the baggage len, maybe something like `kv_content_len` or some other name to disambiguate

src/sdk/propagation/baggage.rs
```diff
@@ -76,10 +76,16 @@ impl TextMapPropagator for BaggagePropagator {
         if !baggage.is_empty() {
             let header_value = baggage
                 .iter()
-                .map(|(name, value)| {
+                .map(|(name, (value, metadata))| {
+                    let metadata_str = if metadata.as_str().trim().is_empty() {
+                        "".to_string()
```
jtescher, line 81: this can probably have `metadata_str` be a str instead of string, don't need either string conversion or the as str below

TommyCpp: Hmm I think in `else` caluse. The `format!(";{}"...)` cannot be stored as `&str` since it will be freed at the end clause if `metadata_str` doesn't own it.

jtescher: oh right you need the prefix. What about something like:

```rust
let metadata_str = metadata.as_str().trim();
let prefix = if metadata_str.is_empty() { "" } else { ";" };
utf8_percent_encode(name.as_str().trim(), FRAGMENT)
    .chain(iter::once("="))
    .chain(utf8_percent_encode(String::from(value).trim(), FRAGMENT))
    .chain(iter::once(prefix))
    .chain(iter::once(metadata_str))
    .collect()
})
```

TommyCpp: Great idea :+1:

src/api/baggage.rs
```diff
@@ -33,6 +41,22 @@ impl Baggage {
...
+    ///
+    /// # Examples
+    /// ```
+    /// use opentelemetry::api::{Baggage, Value};
+    ///
+    /// let mut cc = Baggage::new();
+    /// let _ = cc.insert("my-name", "my-value");
+    ///
+    /// // By default, the metadata is empty
+    /// assert_eq!(cc.get_with_metadata("my-name"), Some(&(Value::String("my-value".to_string()), "".into())))
```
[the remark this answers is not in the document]
TommyCpp: Here we use `into` because we need a `BaggageMetadata`. But I change it to `BaggageMetadata::from` to give more clearity.

src/api/baggage.rs
```diff
@@ -193,3 +332,121 @@ impl BaggageExt for Context {
         self.get::<Baggage>().unwrap_or_else(|| &DEFAULT_BAGGAGE)
```
jtescher, line 332: Unrelated, but this could probably be:

```suggestion
        self.get::<Baggage>().unwrap_or(&DEFAULT_BAGGAGE)
```

src/api/baggage.rs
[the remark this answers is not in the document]
jtescher: oh I see right

jtescher:
@frigus02 any other thoughts or is this good to merge?

src/api/baggage.rs
```diff
@@ -169,27 +299,154 @@ pub trait BaggageExt {
...
+    ) -> Self {
+        let mut merged: Baggage = self
             .baggage()
             .iter()
-            .map(|(key, value)| KeyValue::new(key.clone(), value.clone()))
-            .chain(kvs.into_iter())
+            .map(|(key, (value, metadata))| {
+                KeyValueMetadata::new(key.clone(), value.clone(), metadata.clone())
+            })
+            // .chain(baggage.into_iter().map(|kv| kv.into()))
```
frigus02, line 314: You can probably remove this commented line.

frigus02 (approved):
Looks great. Found one line of commented code. 🙂

jtescher (approved):
Thanks @TommyCpp

PR #288, opened 2020-10-19
[exporter] Allow user choose whether to export instrumentation library

TommyCpp:
Sometimes it's too much overhead for users to export every instrumentation library of every span. Follow up on #243

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -307,7 +307,7 @@ impl PipelineBuilder {
     }
 
     /// Config whether to export information of instrumentation library.
-    pub fn export_instrumentation_library(self, export: bool) -> Self {
+    pub fn with_instrumentation_library_tags(self, export: bool) -> Self {
         PipelineBuilder {
             export_instrument_library: export,
             ..self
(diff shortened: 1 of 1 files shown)


opentelemetry-jaeger/src/lib.rs
```diff
@@ -299,6 +306,14 @@ impl PipelineBuilder {
         }
     }
 
+    /// Config whether to export information of instrumentation library.
+    pub fn export_instrumentation_library(self, export: bool) -> Self {
```
frigus02, line 310: I suggest we start the function with `with_` to follow the convention. Maybe something like `with_instrumentation_library_tags`.

opentelemetry-jaeger/src/lib.rs
```diff
@@ -519,6 +500,53 @@ fn links_to_references(
     }
 }
 
+impl Into<jaeger::Span> for trace::SpanData {
```
frigus02, line 503: Do we still need this implementation? I feel like we may as well remove it otherwise.

frigus02 (approved):
Two small suggestions, but I think this looks good as it is

opentelemetry-jaeger/src/lib.rs
TommyCpp: Yeah make sense. Will remove it.

PR #289, opened 2020-10-20
Update crate lint configurations

jtescher:
Add useful lints to each crate, we are following all of them besides the `rust_2018_idioms`'s use of implicit lifetimes, which we probably should be doing.


frigus02 (approved):
This is nice. Agree about the implicit lifetimes.

Some thoughts about the list of lints. In rustc 1.47 (not sure if any of those changed in recent versions):

- `bad_style` is a deprecated alias for `nonstandard_style` (according to [rustc docs](https://doc.rust-lang.org/rustc/lints/groups.html))
- `dead_code`, `path_statements`, `unused_allocation` and `unused_parens` are part of `unused`. It probably fine to specify them explicitly. Just wanted to mention in case you want to keep the list shorter.

jtescher:
@frigus02 good catch, I cleaned it up a bit. Makes more sense now?

frigus02 (approved):
Looks great 🙂

Issue #252, opened 2020-10-06
Extract common test utils


jtescher:
Structs like `TestSpan` are currently re-defined in multiple test modules. Would be nice to collect into common testing utilities.

drexler:
@jtescher if you haven't started this, I'll like to take this on.

PR #294, opened 2020-10-22
Extract common test utilities

drexler:
Reference Issue: #252


TommyCpp:
Maybe we could rename `src/testing/util.rs` into `src/testing/trace.rs` or something like that so that we could separate the utils of trace and metrics?

jtescher:
@TommyCpp did you have other thoughts here or is this ready?

TommyCpp:
@jtescher Yep this looks good to me.

jtescher (approved):

PR #296, opened 2020-10-23
Add MSRV

jtescher:
* Support rust 1.42.x for core crate and trace exporters (metrics exporters are still unstable and `protobuf` does not support 1.42)
* Make small inconsequential syntax changes to support backwards compatibility
* Ensure dependencies support 1.42 and remove unused default features

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-std = { version = "1.6", features = ["unstable"], optional = true }
+async-std = { version = "1.6", features = ["unstable"], default-features = false, optional = true  }
 async-trait = { version = "0.1", optional = true }
 base64 = { version = "0.13", optional = true }
 bincode = { version = "1.2", optional = true }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -20,25 +20,25 @@ edition = "2018"
...
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
 http = { version = "0.2", optional = true }
 thiserror = { version = "1.0", optional = true }
-tokio = { version = "0.2", features = ["rt-core", "blocking", "time", "stream"], optional = true }
-tonic = { version = "0.3", optional = true }
-reqwest = { version = "0.10", features = ["blocking"], optional = true }
-surf = { version = "2.0", optional = true }
+tokio = { version = "0.2", default-features = false, features = ["rt-core", "blocking", "time", "stream"], optional = true }
+tonic = { version = "0.3", default-features = false, optional = true }
+reqwest = { version = "0.10", default-features = false, features = ["blocking"], optional = true }
```
frigus02, line 40: I wonder if we should specify the "blocking" here or not. If users use the Zipkin exporter and:
- enable the "reqwest-blocking-client" feature, it enables "opentelemetry/reqwest" and "reqwest/blocking"
- enable the "reqwest-client" feature, it enables only "opentelemetry/reqwest". But does this now also enable the blocking feature because of this definition here?

Anyway, this is actually not related to this PR. I can check that later on.

frigus02 (approved):
Looks really good. Well done with the documentation 👍

Cargo.toml
TommyCpp: I think it's here because we implemented the blocking and async client when `opentelemetry` enables `reqwest` feature. But we could definitely separate them into two different features.

jtescher: yeah can open an issue if there are concerns here 👍

frigus02: Oh yeah. I get it now. Doesn't seem super important, to be honest.

PR #297, opened 2020-10-23
Add feature requirements to docs

jtescher:
* Add `rustdoc-args = ["--cfg", "docsrs"]` and `#[cfg_attr(docsrs, doc(cfg(feature = "<feature-name>")))]` to have features show up on docs.rs
* Move module docs in private modues to their primary structs (`Context`, `Tracer`, etc) as they were currently hidden.
* allow `private_doc_tests` as we currently use private modules to conform to spec layout without forcing users to go through layers of modules.
* Fix some broken doc links


frigus02 (approved):
I haven't tested it, yet, but this looks great.

You could add the doc feature attribute on the different implementations for `HttpClient` as well.

frigus02:
Not sure what the exact command is to test this locally. I tried this, which should be quite close:

```
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
```

Looks nice:

image

PR #308, opened 2020-10-26
Make timestamp parameter in end_with_timestamp optional

morigs:
According to [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#end) the parameter should be optional and `end_with_timestamp` must use current timestamp as default value.
This PR suggests a possible implementation.
Library UX suffers a little, so I'm unsure what is better: spec compliance or shorter way to call a method.

### DIFF
--- a/src/api/trace/noop.rs
+++ b/src/api/trace/noop.rs
@@ -103,7 +103,7 @@ impl trace::Span for NoopSpan {
     }
 
     /// Ignores `Span` endings
-    fn end_with_timestamp(&self, _timestamp: SystemTime) {
+    fn end_with_timestamp(&self, _timestamp: Option<SystemTime>) {
         // Ignored
     }
 }
--- a/src/api/trace/span.rs
+++ b/src/api/trace/span.rs
@@ -154,7 +154,7 @@ pub trait Span: fmt::Debug + 'static + Send + Sync {
     ///
     /// This API MUST be non-blocking.
     fn end(&self) {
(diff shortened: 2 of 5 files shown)


frigus02:
The Span trait provides a function end() with a default implementation, which calls end_with_timestamp(SystemTime::now()). This felt like a nicer API design in Rust where we don't have default parameters to functions.

morigs:
@frigus02 Unfortunately no 😞 
`end` really looks more pretty. It's sad that Rust doesn't support optional parameters.
I you're sure that separate method is more convenient than `Option` feel free to close PR

morigs:
At least this question is clarified now and will be in Github history 😃

frigus02:
This was originally added in #199. I personally like the `end` and `end_with_timestamp` methods. What do you think @TommyCpp?

TommyCpp:
I guess the problem with `Option` is you still need to pass a `None` if you want to use the default value(current time in this case), which doesn't feels like a good ergonomic choice. The other consideration is I think most of the time users would just want to close with current time. So it feels more user friendly if they could just do it. I think there is only few case where you want to end with specific timestamp. Like when you parse some span json files. So my intuition is to cover most of the use case with simple `end` but gives users a `end_with_timestamp` just in case.

frigus02:
That's what I think, too. Let's close this then. But thanks anyway for brining this up @morigs. It's always good to hear about those issues. 👍

PR #309, opened 2020-10-26
Add basic tests for SDK Span

morigs:
This PR adds tests for Span implementation.
The purpose of these changes is to make it easier to make changes and ensure compliance with the specification.
Please let me know if some tests are unnecessary or some are missing


jtescher (approved):
Looks good, thanks @morigs

PR #310, opened 2020-10-27
[trace] Move mark_as_active_span and get_active_span into context

### LINKED ISSUE #273: Consider updating trace context extensions
Consider changing `TraceContextExt` to `TracingContextUtilities`. see the spec change here https://github.com/open-telemetry/opentelemetry-specification/pull/1063/

TommyCpp:
Resolve #273 
Spec gives a few choices on where those two functions should be put. Since Rust is not an OOP language. We should just expose the functions in trace module. Leave those two functions in context since they are delegations of context operations.

I feel like `TraceContextExt` is more consistent with rust world so I didn't really rename it into `TracingContextUtilities`. But could definitely make the change if we think `TracingContextUtilities` is more intutive.

### DIFF
--- a/src/api/trace/tracer.rs
+++ b/src/api/trace/tracer.rs
@@ -115,7 +115,7 @@ use std::time::SystemTime;
 /// the following example _will not_ work:
 ///
 /// ```no_run
-/// # use opentelemetry::{global, trace::{Tracer, mark_span_as_active}};///
+/// # use opentelemetry::{global, trace::{Tracer, mark_span_as_active}};
 /// # let tracer = global::tracer("foo");
 /// # let span = tracer.start("foo-span");
 /// async {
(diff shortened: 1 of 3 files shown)


src/api/trace/tracer.rs
```diff
@@ -115,12 +115,12 @@ use std::time::SystemTime;
 /// the following example _will not_ work:
 ///
 /// ```no_run
-/// # use opentelemetry::{global, trace::Tracer};
+/// # use opentelemetry::{global, trace::{Tracer, mark_span_as_active}};///
```
jtescher, line 118: nit: trailing slashes

jtescher (approved):
This seems reasonable to me, @frigus02 or others have opinions on function vs method?

frigus02 (approved):
Makes sense to me 👍

Issue #234, opened 2020-09-30
Error Flagging with Status Codes


jtescher:
See the [spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/966) for details.

drexler:
@jtescher i'll take this one on.

drexler:
Blocked on https://github.com/open-telemetry/opentelemetry-proto/pull/224. I'll push my branch once that merges.

jtescher:
Closed via #315

PR #315, opened 2020-10-29
Update span status codes to match spec

drexler:
Reference issue: #234


opentelemetry-jaeger/src/lib.rs
```diff
@@ -622,7 +622,7 @@ fn build_span_tags(
     }
 
     // Ensure error status is set
-    if status_code != StatusCode::OK && !user_overrides.error {
+    if status_code != StatusCode::Ok && !user_overrides.error {
```
jtescher, line 625: This should probably instead look for error now

jtescher (changes requested):
Looks good @drexler, just one small request

opentelemetry-jaeger/src/lib.rs
drexler: done.

jtescher (approved):
Looks good!

PR #319, opened 2020-10-30
[propagator] Ensure compliance with b3 propagator spec.

### LINKED ISSUE #314: Ensure compliance with b3 propagator spec
See the clarification of the spec here https://github.com/open-telemetry/opentelemetry-specification/pull/1045

TommyCpp:
1. Force extract to try to extract from single header format first, if it doesn't work, then try multiple header format.
2. Don't attch empty span context as remote span context.

Fix #314


src/sdk/propagation/b3.rs
```diff
@@ -269,17 +269,16 @@ impl TextMapPropagator for B3Propagator {
     /// format was retrieved OR if the retrieved data is invalid, then the current
     /// `Context` is returned.
     fn extract_with_context(&self, cx: &Context, extractor: &dyn Extractor) -> Context {
-        let span_context = if self.inject_encoding.support(&B3Encoding::SingleHeader) {
-            self.extract_single_header(extractor).unwrap_or_else(|_|
-                // if invalid single header should fallback to multiple
-                self.extract_multi_header(extractor)
-                    .unwrap_or_else(|_| SpanContext::empty_context()))
-        } else {
+        let span_context = self.extract_single_header(extractor).unwrap_or_else(|_|
```
jtescher, line 272: Could we improve this even further by removing the extra empty span context? Something like:

```rust
let extracted = self.extract_single_header(extractor).or_else(|_| {
    // if invalid single header should fallback to multiple
    self.extract_multi_header(extractor)
});

if let Some(span_context) = extracted.ok().filter(|sc| sc.is_valid()) {
    cx.with_remote_span_context(span_context)
} else {
    cx.clone()
}
```

jtescher (approved):
looks good!

frigus02 (approved):

PR #322, opened 2020-10-30
Fix reqwest feature dependency for zipkin

### LINKED ISSUE #321: Actix Zipkin exemples
I can't make actix work with opentelemetry-zipkin (https://github.com/OutThereLabs/actix-web-opentelemetry/issues/32) so i think it would help if there was some examples here.

jtescher:
Currently `opentelemetry-zipkin` will not compile with the non-blocking reqwest configuration as `reqwest` is not specified as a dependency.

Fixes #321


frigus02 (approved):

TommyCpp (approved):
May need to fix it in datadog exporter too. https://github.com/open-telemetry/opentelemetry-rust/blob/468adf9aa4768607f24b2000e9e0effc783ed345/opentelemetry-contrib/Cargo.toml#L25

Issue #317, opened 2020-10-30
api::Value is much more loosely typed than the spec


djc:
In particular, it allows heterogenous arrays and has a `Bytes` variant that doesn't seem to appear in the spec. Would it be useful to define it more like this?

```rust
enum Primitive {
    Bool(bool),
    Int(i64),
    Float(f64),
    String(String),
}

enum Array {
    Bool(Vec<Option<bool>>),
    Int(Vec<Option<i64>>),
    Float(Vec<Option<f64>>),
    String(Vec<Option<String>>),
}

enum Value {
    Primitive(Primitive),
    Array(Array),
}
```

https://github.com/open-telemetry/opentelemetry-specification/blob/9e6a3de19cb3db983b2e4d56b37ec6313fd4ffc6/specification/common/common.md#attributes

jtescher:
Additional discussion around optional array values: https://github.com/open-telemetry/opentelemetry-specification/pull/1155

djc:
After #323, #332 and #333, I think this is now in a pretty good shape.

PR #323, opened 2020-10-31
Remove support for u64 and bytes values (see #317)

djc:
The [current spec](https://github.com/open-telemetry/opentelemetry-specification/blob/9e6a3de19cb3db983b2e4d56b37ec6313fd4ffc6/specification/common/common.md#attributes) does not recognize these types.

### DIFF
--- a/src/api/core.rs
+++ b/src/api/core.rs
@@ -35,14 +35,6 @@ impl Key {
         }
     }
 
-    /// Create a `KeyValue` pair for `u64` values.
-    pub fn u64(&self, value: u64) -> KeyValue {
-        KeyValue {
-            key: self.clone(),
-            value: Value::U64(value),
-        }
-    }
-
     /// Create a `KeyValue` pair for `f64` values.
     pub fn f64(&self, value: f64) -> KeyValue {
         KeyValue {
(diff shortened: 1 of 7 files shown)


src/api/core.rs
```diff
@@ -193,10 +165,7 @@ fn format_value_array_as_string(v: &[Value]) -> String {
     format!(
         "[{}]",
         v.iter()
-            .map(|elem| match elem {
-                v @ Value::String(_) | v @ Value::Bytes(_) => format!(r#""{}""#, String::from(v)),
-                v => String::from(v),
-            })
+            .map(String::from)
```
frigus02, line 168: I think the intention was to produce a JSON formatted array, which would require strings to be quoted. I also think this didn't produce correct JSON all the time because it didn't escape special characters in string values. But maybe we want to keep the quoted strings?

frigus02 (commented):
Thanks for working on this. Changes look good to me. I do have one concern about the array formatting, though.

src/api/core.rs
djc: Ah, good point, I skipped over that part a little. Fixed up in the next push.

frigus02 (approved):

PR #320, opened 2020-10-30
Add code coverage github action

jtescher:
Pull out nightly tests into a new CI run that includes test coverage using `grcov` and `codecov`

Resolves #318

PR #324, opened 2020-10-31
Move opentelemetry crate into a directory to flatten the workspace

djc:
This makes working with cargo easier, because many commands (like cargo
check, cargo test) will apply to all crates by default, rather than
only selecting the crate in the top-level directory.


.github/workflows/ci.yml
```diff
@@ -61,4 +61,4 @@ jobs:
         toolchain: 1.42.0
         components: clippy
     - name: Run tests
-      run: cargo test --verbose --features base64_format,trace,metrics,serialize,binary_propagator,tokio,async-std,serde,http,tonic,reqwest -p opentelemetry -p opentelemetry-jaeger -p opentelemetry-zipkin
+      run: cargo test --verbose --all-features -p opentelemetry -p opentelemetry-jaeger -p opentelemetry-zipkin
```
jtescher, line 64: We can't use `--all-features` here because there are some optional features like `surf` support which do not support rust 1.42.0. The msrv tests _should_ be failing for this pr?

jtescher:
Should probably move the changelog in as well as each crate has their own

.github/workflows/ci.yml
jtescher: May need to add `override: true`

djc: What do you mean by adding `override: true`? What would that do?

I did verify that "surf" is getting enabled and the test passes, so let's just leave it like this for now?

jtescher: Does `$ cargo +1.42.0 test --all-features` work for you locally?

jtescher: (I think without `default` or `override` set to `true` it will not be using 1.42.0 in tests if I'm reading https://github.com/actions-rs/toolchain#inputs correctly)

jtescher: I added those in #320 so should be able to verify by merging master

djc:
So the new meta job is failing without a clear error message. It looks like it can no longer find `protoc`, maybe? @jtescher, any ideas?

Cargo.toml
```diff
@@ -1,59 +1,3 @@
...
-[features]
-default = ["trace"]
-base64_format = ["base64", "binary_propagator"]
-trace = ["rand", "pin-project", "async-trait", "regex", "percent-encoding"]
-metrics = ["thiserror", "dashmap", "fnv"]
-serialize = ["serde", "bincode"]
-binary_propagator = []
-
 [workspace]
 members = [
```
frigus02, line 2: Do we need to add the new `opentelemetry` folder to the workspace members?

djc: Huh, good catch!

jtescher:
@djc the `opentelemetry` crate was the default before, think adding `-p opentelemetry` should fix the tests

jtescher (approved):
LGTM, still working on things? or is this ready to be merged.

djc:
Nope, I'm happy!

Issue #265, opened 2020-10-12
Ensure `is_recording` is false after span ends


jtescher:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1011 for details.

morigs:
It seems currently span can be ended in two ways:
- when goes out of scope ([using `drop`](https://github.com/open-telemetry/opentelemetry-rust/blob/b47794e5818d7ba39115933ea7f752b8784cf025/src/sdk/trace/span.rs#L147))
- when [`end_with_timestamp`](https://github.com/open-telemetry/opentelemetry-rust/blob/b47794e5818d7ba39115933ea7f752b8784cf025/src/sdk/trace/span.rs#L138) is called

Right now [`is_recording`](https://github.com/open-telemetry/opentelemetry-rust/blob/b47794e5818d7ba39115933ea7f752b8784cf025/src/sdk/trace/span.rs#L107) handles only the first case. So I would like to suggest just to replace [this line](https://github.com/open-telemetry/opentelemetry-rust/blob/b47794e5818d7ba39115933ea7f752b8784cf025/src/sdk/trace/span.rs#L107) with something like this:
```rust
match self.with_data(|data| data.end_time.is_some()) {
  Some(ended) => !ended,
  None => false,
}
```

jtescher:
@morigs yeah additionally it may be a good idea to move the logic from drop to `end_with_timestamp` so other methods like setting attributes are no-ops as defined by the spec

morigs:
@jtescher I think it's better not to move logic from drop because this will not allow to get SpanContext after end.
As an alternative I could suggest adding additional check to every method (check `end_time` for example)

morigs:
Oh-oh, I forgot that `end_time` is always set. May be we should define some new bool (`is_ended`) and set it in `end`?

jtescher:
Actually it's probably preferable to move span context outside the mutex, would also allow span context to return a reference instead of a clone which would be an improvement. I have a branch that does this that I could clean up and push.

morigs:
I have added some tests in #309 
Some of them related to this issue

morigs:
@jtescher Hi. I'm working on moving logic from drop to `end_with_timestamp`. Two points here:
- We still must end span on drop. So I've replaced `impl Drop for SpanInner` with `impl Drop for Span` because `SpanInner` doesn't have `end` method
[...]

PR #325, opened 2020-11-01
Return span context ref

jtescher:
Currently the `Span` trait in the API has its `span_context` method return an owned `SpanContext`. In order to be more consistent with the rest of the APIs, this patch updates the trait to instead return a reference, and moves the `SpanContext` data outside of the mutex in the SDK to implement the trait. This also allows #265 to be implemented
without having to lock.

Additionally the spec now states that the `on_start` method for `SpanProcessor`s must be passed a span and the parent context so those are updated as well.


TommyCpp (approved):
Looks nice! 👍

PR #328, opened 2020-11-02
Improve performance of EvictedHashMap

jtescher:
A large portion of tracing time is typically spent processing span attributes. This patch improves performance of inserting additional span attributes at creation time by optimizing the backing evicted hash map data structure.

Changes:

* Use `entry` api on insert to avoid hashing keys twice
* Specify an initial attribute capacity when building a span to avoid growing backing hashmap capacity repeatedly
* Avoid allocating and appending extra links vec if unused
* (non-perf) add new type for `Iter` and `IntoIter` to avoid std hashmap in public api

Benchmark changes:

```
EvictedHashMap/insert 1 time:   [210.50 ns 212.56 ns 215.01 ns]
                        change: [-6.6266% -4.9711% -3.3843%] (p = 0.00 < 0.05)
                        Performance has improved.
EvictedHashMap/insert 5 time:   [828.72 ns 836.35 ns 843.99 ns]
                        change: [-21.606% -20.885% -20.169%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
EvictedHashMap/insert 10
                        time:   [1.5435 us 1.5517 us 1.5604 us]
                        change: [-26.554% -26.207% -25.837%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
EvictedHashMap/insert 20
                        time:   [2.9865 us 2.9960 us 3.0070 us]
                        change: [-28.683% -27.587% -26.707%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  2 (2.00%) high severe
```

### DIFF
--- a/opentelemetry-contrib/src/datadog/model/mod.rs
+++ b/opentelemetry-contrib/src/datadog/model/mod.rs
@@ -90,7 +90,7 @@ mod tests {
         let end_time = start_time.checked_add(Duration::from_secs(1)).unwrap();
 
         let capacity = 3;
-        let mut attributes = sdk::trace::EvictedHashMap::new(capacity);
+        let mut attributes = sdk::trace::EvictedHashMap::new(capacity, capacity as usize);
         attributes.insert(Key::new("span.type").string("web"));
 
         let message_events = sdk::trace::EvictedQueue::new(capacity);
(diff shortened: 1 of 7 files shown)


frigus02 (approved):
Nice 👍 

I noticed we no longer implement `IntoIterator` for `&EvictedQueue`, `&mut EvictedQueue` and `&mut EvictedHashMap`. Those might not be needed, so I'm not too bothered.

jtescher:
@frigus02 yeah we're not using those currently, can always add back if someone has a use for them

PR #333, opened 2020-11-02
Make sure value arrays are homogeneous (see #317)

djc:


jtescher (approved):
Looks good, we can always remove the `Option`s if the spec clarification disallows nulls

PR #334, opened 2020-11-02
[trace] Move SpanProcessor into sdk. Use ReadableSpan as parameter in on_start

### LINKED ISSUE #327: Move SpanProcessor into sdk and update the type of span parameter to sdk::Span
While applying today's change into my zpage brach. I found one issue with span parameter type being `&dyn api::Span`. 
As defined in spec, 
> span - a read/write span object for the started span.

the span should be a [read/write span object](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#additional-span-interfaces), which means it should be able to read all information added into span such as span name, start time, etc. But the current type(`&dyn api::Span`) only provides write API. So ideally we should update the type into `&sdk::Span`. 

But it's kind of confusing to have a type in api module(SpanProcessor) to depend on a type from sdk module(sdk::Span). So we might as well move SpanProcessor into sdk. The other reason is SpanProcessor is defined in sdk instead of api in spec.

TommyCpp:
The spec says the SpanProcessor's on_start method must have a parameter that can access [all information](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span) added into span. The api::trace::Span cannot access most of the information. 

This PR creates a new trait `ReadableSpan` which has a method to returns a owned exporter::trace::SpanData. This should help users if they need access any information in on_start method.

Close #327


jtescher:
I think this is a reasonable approach, basically allow reading a snapshot of state at any point through a trait. It seems [the `on_start` spec](https://github.com/open-telemetry/opentelemetry-specification/blob/323cd1d8811d3e3df08bec689f9dacd6cb782888/specification/trace/sdk.md#onstart) hints that it would be nice to be able to clone this type:

> It SHOULD be possible to keep a reference to this span object and updates to the span SHOULD be reflected in it. For example, this is useful for creating a SpanProcessor that periodically evaluates/prints information about all active span from a background thread.

But I think that would either force the method to accept an sdk span, which would make span processors not work with other spans (which may be fine?), or it would mean introducing some tricky generics. Anyone have preferences here?

TommyCpp:
>  which would make span processors not work with other spans

I think using `sdk::trace::Span` would be a easy solution. It's nice to allow user bring their own implementation. But since we are talking about sdk rather than api. If we don't allow users to do so it should be fine. 

> It seems the on_start spec hints that it would be nice to be able to clone this type

Maybe we could wrap current `ReadableSpan` with `Arc` but that seems would bring some performance overhead.

jtescher:
Yeah I think forcing `Arc` here when the default processors ignore `on_start` is probably not great. Going with `sdk::trace::Span` seems reasonable as it is `Clone`.

jtescher (approved):

PR #335, opened 2020-11-03
Add feature flag documentation to crates

jtescher:
Add feature flag descriptions to crate docs and copy README.md/CODEOWNERS into `opentelemetry` folder. The `opentelemetry` crate README may diverge from the root README, but for now they are the same.


opentelemetry-contrib/src/lib.rs
```diff
@@ -4,6 +4,16 @@
 //! some users.
 //!
 //! Typically, those include vendor specific propagators.
+//!
+//! ## Crate Feature Flags
+//!
+//! The following crate feature flags are available:
+//!
+//! * `datadog`: Adds a Datadog trace exporter.
+//! * `reqwest-blocking-client`: Export spans using the reqwest blocking http
```
TommyCpp, line 13: Should be `exports` I guess?
```suggestion
//! * `reqwest-blocking-client`: Exports spans using the reqwest blocking http
```

opentelemetry-jaeger/src/lib.rs
```diff
@@ -144,6 +144,12 @@
 //! }
 //! ```
 //!
+//! ## Crate Feature Flags
+//!
+//! The following crate feature flags are available:
+//!
+//! * `collector_client`: Export span data directly to a Jaeger collector.
```
TommyCpp, line 151: Could be helpful to mention the underlying http client
```suggestion
//! * `collector_client`: Export span data directly to a Jaeger collector using isahc http client.
```

opentelemetry-contrib/src/lib.rs
jtescher: Hm this was intended to read more as "If you use this feature it will export spans using x" rather than "the client exports spans using x" as it is a description of the feature rather than what the client does

TommyCpp (approved):

PR #343, opened 2020-11-05
Move binary propagator and base64 format to contrib.

### LINKED ISSUE #336: Consider moving base64 and binary propagators to contrib
They are not currently part of the spec, we should consider moving them before GA

TommyCpp:
* Moved them as they are not part of the spec for now and should not be in main crate when GA. (Moved files and two features from main crate to contrib)
* Moved binary propagator and base64 format into a `binary` module and reexport the public types.
* Also restructure the contrib into two submodule, one for tracing and one for metric.

closes #336


jtescher (approved):

examples/datadog/src/main.rs
```diff
@@ -17,7 +17,7 @@ fn bar() {
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
-    let (tracer, _uninstall) = opentelemetry_contrib::datadog::new_pipeline()
+    let (tracer, _uninstall) = opentelemetry_contrib::trace::exporter::datadog::new_pipeline()
```
jtescher, line 20: small nit: may want to `use` this as above? calling a function 4 namespaces deep currently 😂

TommyCpp: Yeah definitely should have imported them at the top

Issue #338, opened 2020-11-03
Consider allowing multiple attributes to be set at once


jtescher:
Currently span attributes must be set one at a time, and take a (likely uncontended) lock each time. Consider whether it is worth it to have a second method that takes an array of kv, or if the method should take a generic. Possibly not necessary, but we should decide before GA

TommyCpp:
I don't see why we can't implement one. This should improve ergonomics and  seems easy to add.  I think maybe take a generic iterator of KV?

jtescher:
Give it a shot and see if it is ergonomic, can look at the benchmarks too to see if batch inserts matter there

TommyCpp:
So I image we could also close this issue and remove it from GA checklist?

PR #346, opened 2020-11-06
Add expend_attributes method

TommyCpp:
Part of #338 
Did a POC on generic iterator of KeyValue. One problem is since we are using `dyn Span` in `global` module. We cannot really have a function with generic argument in `Span` trait. Hacked it around a little bit based on [this post](https://internals.rust-lang.org/t/generic-methods-over-object-safe-traits/6774). But still feel like we might need to use Vec in the end.

Benchmarked the speed of inserting 1000 key value pairs in one thread. Looks promising, should be more difference if running in multiple threads.
```
set attributes in batch/insert in batch                                                                            
                        time:   [60.776 us 60.903 us 61.047 us]
                        change: [-4.5324% -3.8952% -2.7253%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  3 (3.00%) high severe
set attributes in batch/insert one by one                                                                            
                        time:   [73.007 us 73.119 us 73.279 us]
                        change: [-3.7832% -3.5073% -3.2366%] (p = 0.00 < 0.05)
                        Performance has improved.

```

### DIFF
--- a/opentelemetry/benches/trace.rs
+++ b/opentelemetry/benches/trace.rs
@@ -6,20 +6,11 @@ use opentelemetry::{
 };
 
 fn criterion_benchmark(c: &mut Criterion) {
-    let mut group = c.benchmark_group("EvictedHashMap");
-    group.bench_function("insert 1", |b| {
-        b.iter(|| insert_keys(sdktrace::EvictedHashMap::new(32, 1), 1))
-    });
-    group.bench_function("insert 5", |b| {
-        b.iter(|| insert_keys(sdktrace::EvictedHashMap::new(32, 5), 5))
-    });
-    group.bench_function("insert 10", |b| {
-        b.iter(|| insert_keys(sdktrace::EvictedHashMap::new(32, 10), 10))
-    });
-    group.bench_function("insert 20", |b| {
-        b.iter(|| insert_keys(sdktrace::EvictedHashMap::new(32, 20), 20))
-    });
-    group.finish();
+    evicted_hashmap_benchmark(c);
+    let tracer = sdktrace::TracerProvider::builder()
+        .build()
+        .get_tracer("tracer", None);
+    set_attribute_in_batch(c, &tracer);
 
     trace_benchmark_group(c, "start-end-span", |tracer| tracer.start("foo").end());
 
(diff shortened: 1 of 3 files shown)


jtescher:
Yeah this is a tricky one to get the ergonomics of without introducing too many extra types or impls. Can think about it.

TommyCpp:
Switched to use `Vec<KeyValue>` as parameter type for now. Tried following method with generic iterator but don't seem to work. Open to any suggestions.

1. Add `extend_attributes` to `Span` trait. The problem with this method is we are using trait object in `global.rs`, which in return prevent us from having a type parameter. I tried adding `where Self: Sized` to avoid this problem but then only structs can call this function. In API, tracers return span in `&dyn Span` type. So user won't be able to use it.

2. Build a separate trait, for example `SpanExt` for this method. But then because we returned `&dyn Span` in tracer. We will need to implement this trait for all types that implemented `Span` trait. But If we do so we won't be able to provide a second implementation in `sdk::trace::Span`. Thus we won't gain any performance improve.

frigus02:
I wonder if we could use the [`Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) trait for this. Maybe in a way similar to this:

1. Implement `Extend` for `EvictedHashMap`. We can make this efficiently insert multiple entries by taking the size hint into account.

2. Implement something like this for `Span`:

   ```rust
   fn attributes_mut(&mut self) -> &mut EvictedHashMap
   ```

TommyCpp:
> 2. Implement something like this for `Span`:
   ```rust
   fn attributes_mut(&mut self) -> &mut EvictedHashMap
   ```

Great idea. But I guess the problem here is we can't return a `sdk::trace::EvictedHashMap` from a trait in `api`. Because it would require users to use `sdk` and block people from using their own implementation of `api`. 

Also I think normally users would get a immutable reference of a span trait object so the receiver would have to be `&self`

frigus02:
> Great idea. But I guess the problem here is we can't return a `sdk::trace::EvictedHashMap` from a trait in `api`. Because it would require users to use `sdk` and block people from using their own implementation of `api`.

Ah yes. I didn't think about that. That's really annoying.

jtescher:
Hm if inserting 1000 pairs is only 10us difference is this method necessary? wondering if we should hold off until the demand is clearer if it is not a sizable perf win. Thoughts?

frigus02:
Yeah, I think I agree. Makes sense for me to wait until we have a clear demand.

TommyCpp:
Yeah I agree that we could wait until it become a performance bottleneck for someone. It's also not in spec. Close it for now.

PR #347, opened 2020-11-06
Change ExportResult to std::result::Result

### LINKED ISSUE #331: Update ExportResult: use std::result::Result type
Update [`ExportResult`](https://github.com/open-telemetry/opentelemetry-rust/blob/866ae41bd238a293a9da25e8d457da3162df3aa9/opentelemetry/src/exporter/trace/mod.rs#L21) to match the latest version in the spec:

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#exportbatch

> ExportResult is one of:
> - **Success** - The batch has been successfully exported. For protocol exporters this typically means that the data is sent over the wire and delivered to the destination server.
> - **Failure** - exporting failed. The batch must be dropped. For example, this can happen when the batch contains bad data and cannot be serialized.

Since this only defines 2 states now (error and success), we should probably go the idiomatic Rust route and use a `std::result::Result`.

frigus02:
The spec changed and ExportResult no longer includes the hint about error being retriable. Since it now only distinguishes between success and error, it makes sense to use Rust's standard Result type.

Closes #331

### DIFF
--- a/opentelemetry-contrib/src/trace/exporter/datadog/mod.rs
+++ b/opentelemetry-contrib/src/trace/exporter/datadog/mod.rs
@@ -86,13 +86,13 @@
 //!
 //! #[async_trait]
 //! impl HttpClient for IsahcClient {
-//!   async fn send(&self, request: http::Request<Vec<u8>>) -> Result<ExportResult, Box<dyn Error + Send + Sync + 'static>> {
+//!   async fn send(&self, request: http::Request<Vec<u8>>) -> ExportResult {
 //!     let result = self.0.send_async(request).await?;
 //!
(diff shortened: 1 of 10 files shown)


jtescher (approved):
Looks good

jtescher:
Also looks like the CLA bot is stuck again..

frigus02:
Hm yeah. How do we get it unstuck? Do we just wait?

jtescher:
Not sure why it stops working, but pushing an empty commit should trigger it (when it starts working again?)

frigus02:
Cool. I push another commit tomorrow 🤞

PR #350, opened 2020-11-08
Change SpanExporter::export &self to &mut self 

frigus02:
This is a proposal. I think it would make sense to change `&self` in the `SpanExporter::export` function to a `&mut self`. The spec says 

> Export() will never be called concurrently for the same exporter instance. Export() can be called again only after the current call returns.

This change would enforce the restriction at compile time, which is nice.

While doing this change I noticed that the `SimpleSpanProcessor` doesn't adhere to this. It calls `export` in `on_end` without any checks and `on_end` can be called from multiple threads concurrently. I changed it to use a Mutex for now.

As a second step I removed the `Sync` requirement for `SpanExporter`. We can now do this without any problems, because both of it's functions take a `&mut self`. This means the stdout exporter no longer needs an internal Mutex. I'm not entirely sure about this change, though. Is it possible for the spec to have a non-breaking change, that would require us to add `Sync` back (which would be a breaking change)?

### DIFF
--- a/opentelemetry/src/exporter/trace/mod.rs
+++ b/opentelemetry/src/exporter/trace/mod.rs
@@ -50,8 +50,8 @@ pub trait SpanExporter: Send + Debug {
     /// not allowed and should return an error.
     ///
     /// This function should not block indefinitely (e.g. if it attempts to
-    /// flush the data and the destination is unavailable). Language library
-    /// authors can decide if they want to make the shutdown timeout
+    /// flush the data and the destination is unavailable). SDK authors
+    /// can decide if they want to make the shutdown timeout
     /// configurable.
     fn shutdown(&mut self) {}
 }
--- a/opentelemetry/src/exporter/trace/mod.rs
+++ b/opentelemetry/src/exporter/trace/mod.rs
@@ -23,45 +23,43 @@ pub type ExportResult = Result<(), Box<dyn std::error::Error + Send + Sync + 'st
 /// implement so that they can be plugged into OpenTelemetry SDK and support
 /// sending of telemetry data.
 ///
-/// The goals of the interface are:
-///
-/// - Minimize burden of implementation for protocol-dependent telemetry
-///  exporters. The protocol exporter is expected to be primarily a simple
-/// telemetry data encoder and transmitter.
-/// - Allow implementing helpers as composable components that use the same
-/// chainable Exporter interface. SDK authors are encouraged to implement common
-/// functionality such as queuing, batching, tagging, etc. as helpers. This
-/// functionality will be applicable regardless of what protocol exporter is used.
+/// The goal of the interface is to minimize burden of implementation for
+/// protocol-dependent telemetry exporters. The protocol exporter is expected to
+/// be primarily a simple telemetry data encoder and transmitter.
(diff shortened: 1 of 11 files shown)


opentelemetry/src/exporter/trace/mod.rs
```diff
@@ -31,45 +31,43 @@ pub enum ExportResult {
...
+    /// This function should be called only once for each `SpanExporter`
+    /// instance. After the call to `shutdown`, subsequent calls to `export` are
+    /// not allowed and should return an error.
     ///
-    /// Shutdown should not block indefinitely (e.g. if it attempts to flush the
-    /// data and the destination is unavailable). SDK authors can
-    /// decide if they want to make the shutdown timeout to be configurable.
+    /// This function should not block indefinitely (e.g. if it attempts to
+    /// flush the data and the destination is unavailable). Language library
+    /// authors can decide if they want to make the shutdown timeout
```
TommyCpp, line 62:
```suggestion
    /// flush the data and the destination is unavailable). SDK authors
    /// can decide if they want to make the shutdown timeout
```

On a side note, do we want to added a timeout for exporters or should we let the exporter implementations decide?

TommyCpp (approved):
Looks great! Just a small nit.

opentelemetry/src/exporter/trace/mod.rs
frigus02: Good question. I noticed that the spec defines a `exportTimeoutMillis` option for the batch span processor. I think we need to implement at least that ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#batching-processor)).

I think I'm not too worried about the `SimpleSpanProcessor` because it's not supposed to be super performant anyway.

jtescher:
I think this is a good idea, should be able to remove the inner mutexes from the exporters now too

e.g. https://github.com/open-telemetry/opentelemetry-rust/blob/03fe982cbf0dc4ada28f98df861b6bdb621bd556/opentelemetry-jaeger/src/agent.rs#L40-L43

frigus02:
> should be able to remove the inner mutexes from the exporters now too

I think I looked at all exporters except Jaeger. Didn't think it would contain mutexes 🤦. Will remove those, too.

jtescher (approved):
Looks good! I think the jaeger exporter buffer clone can probably be removed now too, but can be done in a subsequent pr

frigus02:
Yeah, I _think_ we should be able to remove that. But I'm not entirely sure how. We want the thrift client to write into a buffer, which we can later read. What's the pattern we would typically use in Rust? A channel?

jtescher:
Would be nice if the thrift crate had a cleaner way of writing and reading since it does not do either one in parallel, but the current API only lets you split. In that case it would ordinarily be fine to have a `Rc<RefCell<T>>`, but we need this processing to be spawned so it has to be `Send`. Even if we stay with `Arc<Mutex<T>>` we should still be able to lock, send data to the socket and then clear the vec, as opposed to clone then send data via reference to the cloned vec, but std mutexs don't play well across `await` points. This could use the async mutex types that were used before behind feature flags though.

PR #353, opened 2020-11-09
Implement Display trait for Key and Value types

djc:
This has the advantage that you can write these types into a string buffer without allocating. The split between `&Value` and `Value` `From` impls was not used efficiently in some of the other crates, and that is fixed here as well.

In addition, I believe this makes the API more idiomatic.


jtescher (approved):
Looks good, some new nightly clippy lints, but they can be addressed in subsequent PRs

djc:
Oh, I didn't see those. None of them seem to be related to the changes here, though.

PR #354, opened 2020-11-09
Fix parent based sampling in tracer

### LINKED ISSUE #351: Sampler::TraceIdRatioBased problem
I've used code from example basic

```
use opentelemetry::global;
use opentelemetry::sdk::{trace as sdktrace};
use std::error::Error;
use opentelemetry::api::trace::{Tracer, TraceContextExt};
use opentelemetry::sdk::trace::{Config, Sampler};

fn init_tracer() -> Result<
    (sdktrace::Tracer, opentelemetry_jaeger::Uninstall),
    Box<dyn Error>,
> {
    opentelemetry_jaeger::new_pipeline()
        .with_service_name("test")
        .with_trace_config(Config::default().with_default_sampler(Sampler::TraceIdRatioBased(0.25)))
        .install()
}

fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
    let _uninstall = init_tracer().unwrap();

    let tracer = global::tracer("ex.com/basic");

    tracer.in_span("operation", |cx| {
        let _span = cx.span();
        tracer.in_span("Sub operation...", |cx| {
            let _span = cx.span();
            // some sub operation
        });
    });

    Ok(())
}
```
With sampler always on everything works fine, but when I use Sampler::TraceIdRatioBased(0.25) sometimes I receive in jaeger just sub span. So span "operation" didn't sampled, but "Sub operation" sampled
image

How can I use ratio < 1.0

jtescher:
Currently the sdk `Tracer` implementation will filter out invalid / dropped parent span contexts. This causes the parent based sampler to not function properly as it will then assume that child spans are roots.

This patch resolves this issue by adding a `has_active_span` method to the trace context extension, which allows the tracer to distinguish between invalid and unset parent contexts.

Fixes #351

### DIFF
--- a/opentelemetry/src/sdk/trace/tracer.rs
+++ b/opentelemetry/src/sdk/trace/tracer.rs
@@ -172,9 +172,14 @@ impl crate::trace::Tracer for Tracer {
         let parent_span_context = builder
             .parent_context
             .as_ref()
-            .or_else(|| Some(cx.span().span_context()).filter(|cx| cx.is_valid()))
-            .or_else(|| cx.remote_span_context())
-            .filter(|cx| cx.is_valid());
+            .or_else(|| {
+                if cx.has_active_span() {
+                    Some(cx.span().span_context())
+                } else {
+                    None
+                }
+            })
+            .or_else(|| cx.remote_span_context());
         // Build context for sampling decision
         let (no_parent, trace_id, parent_span_id, remote_parent, parent_trace_flags) =
             parent_span_context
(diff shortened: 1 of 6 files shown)


opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -172,9 +172,14 @@ impl crate::trace::Tracer for Tracer {
...
-            .or_else(|| cx.remote_span_context())
-            .filter(|cx| cx.is_valid());
+            .or_else(|| {
+                if cx.has_active_span() {
+                    Some(cx.span().span_context())
+                } else {
+                    None
+                }
+            })
+            .or_else(|| cx.remote_span_context());
```
frigus02, line 182: We have very similar logic in `NoopTracer`. Do we need to change that, too?

https://github.com/open-telemetry/opentelemetry-rust/blob/bd696420f33295922ddfb62e075e19b6d9fc4487/opentelemetry/src/api/trace/noop.rs#L152-L158

frigus02 (approved):

opentelemetry/src/sdk/trace/tracer.rs
jtescher: Oh yeah good catch

frigus02 (approved):

PR #358, opened 2020-11-11
[span processor] Add force_flush method.

### LINKED ISSUE #349: Implement SpanProcessor::force_flush
Implement `force_flush` as specified here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#forceflush

> Exports all spans that have not yet been exported to the configured Exporter.
> 
> ForceFlush SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.
> 
> ForceFlush SHOULD only be called in cases where it is absolutely necessary, such as when using some FaaS providers that may suspend the process after an invocation, but before the Processor exports the completed spans.
> 
> ForceFlush SHOULD complete or abort within some timeout. ForceFlush can be implemented as a blocking API or an asynchronous API which notifies the caller via a callback or an event. Language library authors can decide if they want to make the flush timeout configurable.

TommyCpp:
See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#forceflush for more details.

closes #349

### DIFF
--- a/opentelemetry/src/sdk/trace/span_processor.rs
+++ b/opentelemetry/src/sdk/trace/span_processor.rs
@@ -126,9 +126,8 @@ impl SpanProcessor for SimpleSpanProcessor {
     }
 
     fn force_flush(&self) {
-        // Ignored since all span in Simple Processor will be exported as they ended.
+        // Ignored since all spans in Simple Processor will be exported as they ended.
     }
-
 
     fn shutdown(&mut self) {
         if let Ok(mut exporter) = self.exporter.lock() {
@@ -181,7 +180,7 @@ impl SpanProcessor for SimpleSpanProcessor {
 /// [`async-std`]: https://async.rs
 pub struct BatchSpanProcessor {
     message_sender: Mutex<mpsc::Sender<BatchMessage>>,
-    worker_handle: Option<Pin<Box<dyn Future<Output=()> + Send + Sync>>>,
+    worker_handle: Option<Pin<Box<dyn Future<Output = ()> + Send + Sync>>>,
 }
 
 impl fmt::Debug for BatchSpanProcessor {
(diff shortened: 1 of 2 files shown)


frigus02 (commented):
This looks nice. I'm unsure about this part of the spec:

> ForceFlush SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.

This sounds like, similar to shutdown, the function should wait until the spans have been exported and only return then.

I don't immediately know how to implement this, though. And it's only a "should". So I think we're spec compliant with the way you did it.

flycash (commented):
From my understanding,
```
ForceFlush SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.
...
ForceFlush SHOULD complete or abort within some timeout. 
```
I think `force_flush` need to add the `timeout` parameter, and use enum or status code as return value.

Implementations can use something like `future` to flush spans.

TommyCpp:
> ForceFlush SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.

@frigus02 Yeah I also noticed that, but I find the `shutdown` method also doesn't return any result. So feel like we should have a dedicated PR to discuss how to implement this. 

> I think force_flush need to add the timeout parameter, and use enum or status code as return value.

@flycash I agree. But I think we might as well add timeout for both `shutdown` and `force_flush` in `SimpleSpanProcessor` and `BatchSpanProcessor`. So I was going to create another PR for that.

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -500,4 +516,35 @@ mod tests {
         assert_eq!(builder.config.max_export_batch_size, 120);
         assert_eq!(builder.config.max_queue_size, 120);
     }
+
+    #[tokio::test]
+    async fn test_batch_span_processor() {
+        let (exporter, mut export_receiver, _shutdown_receiver) = new_tokio_test_exporter();
+        let mut config = BatchConfig::default();
+        config.scheduled_delay = Duration::from_secs(60 * 60 * 24); // set the tick to 24 hours so we know the span must be exported vis force_flush
```
frigus02, line 524:
```suggestion
        config.scheduled_delay = Duration::from_secs(60 * 60 * 24); // set the tick to 24 hours so we know the span must be exported via force_flush
```

frigus02 (approved):
Makes sense to me 👍

PR #312, opened 2020-10-28
Add ecosystem crates to readme

jtescher:
Update docs for crates and fix missing labels `metrics` feature flag

PR #360, opened 2020-11-11
Mention opentelemetry-stackdriver as part of the ecosystem

djc:


TommyCpp:
I think we [thought about it](https://github.com/open-telemetry/opentelemetry-rust/pull/312#discussion_r513148967) back then but decided not to add it as it didn't seem to be in active development. But if that's not the case, :+1: for adding it.

djc:
I've brought it up to date. If it goes silent again I'll fork it.

jtescher (changes requested):
Can we add this to the README as well? (currently both this crate and repo root readmes, could consider making one a symlink)

djc:
Done.

jtescher (approved):

PR #362, opened 2020-11-14
Implement timeout for force_flush and shutdown.

TommyCpp:
- Add the envionment variable support for export timeout.
- Add support of time out when exporting. 
- Added `TraceError` enum, still need to add other error type.

Note that the spec only has configuration for maximum time to export data. So the timeout is based on the export time.

Also included the update from otel-proto

### DIFF
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -26,7 +26,7 @@ async-trait = { version = "0.1", optional = true }
 bincode = { version = "1.2", optional = true }
 dashmap = { version = "4.0.0-rc6", optional = true }
 fnv = { version = "1.0", optional = true }
-futures = { version = "0.3", features = ["std"] }
+futures = "0.3"
 lazy_static = "1.4"
 percent-encoding = { version = "2.0", optional = true }
 pin-project = { version = "0.4", optional = true }
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -21,17 +21,17 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-async-std = { version = "1.6", features = ["unstable"], default-features = false, optional = true  }
+async-std = { version = "1.6", features = ["unstable"], default-features = false, optional = true }
 async-trait = { version = "0.1", optional = true }
 bincode = { version = "1.2", optional = true }
 dashmap = { version = "4.0.0-rc6", optional = true }
 fnv = { version = "1.0", optional = true }
-futures = "0.3"
+futures = { version = "0.3", features = ["std"] }
 lazy_static = "1.4"
 percent-encoding = { version = "2.0", optional = true }
 pin-project = { version = "0.4", optional = true }
 rand = { version = "0.7", default-features = false, features = ["std"], optional = true }
-regex = { version = "1.3", default-features = false, features = ["std", "perf"], optional = true}
+regex = { version = "1.3", default-features = false, features = ["std", "perf"], optional = true }
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
 http = { version = "0.2", optional = true }
 thiserror = { version = "1.0", optional = true }
@@ -44,6 +44,7 @@ surf = { version = "2.0", default-features = false, optional = true }
 criterion = "0.3.1"
 rand_distr = "0.3.0"
 tokio = { version = "0.2", features = ["full"] }
+async-std = { version = "1.6", features = ["unstable"] }
 
 [features]
 default = ["trace"]
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -48,7 +48,7 @@ async-std = { version = "1.6", features = ["unstable"] }
 
 [features]
 default = ["trace"]
-trace = ["rand", "pin-project", "async-trait", "regex", "percent-encoding"]
+trace = ["rand", "pin-project", "async-trait", "regex", "percent-encoding", "thiserror"]
 metrics = ["thiserror", "dashmap", "fnv"]
 serialize = ["serde", "bincode"]
 
--- a/opentelemetry/src/api/trace/mod.rs
+++ b/opentelemetry/src/api/trace/mod.rs
@@ -136,3 +136,14 @@ pub use self::{
     },
     tracer::{SpanBuilder, Tracer},
 };
+
+use thiserror::Error;
+
+/// Errors returned by the trace API.
+#[derive(Error, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum TraceError {
+    /// Other errors not covered by specific cases.
+    #[error("Trace error: {0}")]
+    Other(String),
+}
--- a/opentelemetry/src/api/trace/mod.rs
+++ b/opentelemetry/src/api/trace/mod.rs
@@ -110,6 +110,8 @@
 //! Please review the W3C specification for details on the [Tracestate
 //! field](https://www.w3.org/TR/trace-context/#tracestate-field).
 //!
+use thiserror::Error;
+
 mod context;
 mod event;
 mod futures;
@@ -136,8 +138,6 @@ pub use self::{
     },
     tracer::{SpanBuilder, Tracer},
 };
-
-use thiserror::Error;
 
 /// Errors returned by the trace API.
 #[derive(Error, Debug, PartialEq)]
--- a/opentelemetry/src/sdk/trace/span_processor.rs
+++ b/opentelemetry/src/sdk/trace/span_processor.rs
@@ -278,7 +278,7 @@ impl BatchSpanProcessor {
                     // Span batch interval time reached or a force flush has been invoked, export current spans.
                     BatchMessage::Flush(Some(ch)) => {
                         let mut results = Vec::with_capacity(
-                            spans.len() / config.max_export_batch_size + 1 as usize,
+                            spans.len() / config.max_export_batch_size + 1,
                         );
                         while !spans.is_empty() {
                             let batch = spans.split_off(
@@ -316,7 +316,7 @@ impl BatchSpanProcessor {
                     // Stream has terminated or processor is shutdown, return to finish execution.
                     BatchMessage::Shutdown(ch) => {
                         let mut results = Vec::with_capacity(
-                            spans.len() / config.max_export_batch_size + 1 as usize,
+                            spans.len() / config.max_export_batch_size + 1,
                         );
                         while !spans.is_empty() {
                             let batch = spans.split_off(
--- a/opentelemetry/src/sdk/trace/span_processor.rs
+++ b/opentelemetry/src/sdk/trace/span_processor.rs
@@ -76,10 +76,10 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
     /// API, therefore it should not block or throw an exception.
     fn on_end(&self, span: SpanData);
     /// Force the spans lying in the cache to be exported.
-    fn force_flush(&self) -> ExportResult;
+    fn force_flush(&self) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>;
     /// Shuts down the processor. Called when SDK is shut down. This is an
     /// opportunity for processors to do any cleanup required.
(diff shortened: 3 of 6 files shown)


opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -71,10 +76,10 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
     /// API, therefore it should not block or throw an exception.
     fn on_end(&self, span: SpanData);
     /// Force the spans lying in the cache to be exported.
-    fn force_flush(&self);
+    fn force_flush(&self) -> ExportResult;
     /// Shuts down the processor. Called when SDK is shut down. This is an
     /// opportunity for processors to do any cleanup required.
-    fn shutdown(&mut self);
+    fn shutdown(&mut self) -> ExportResult;
```
frigus02, line 82: I think I would prefer to use a different result type here. The spec mentions `ExportResult` as something the export function returns. The functions `force_flush` and `shutdown` don't have a defined result type in the spec. So maybe we just define one ourselves. I think `Result<(), Box<dyn Error>>` or even `TraceError` would be fine.

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -125,13 +130,17 @@ impl SpanProcessor for SimpleSpanProcessor {
...
+        Ok(())
     }
 
-    fn shutdown(&mut self) {
+    fn shutdown(&mut self) -> ExportResult {
         if let Ok(mut exporter) = self.exporter.lock() {
             exporter.shutdown();
+            Ok(())
+        } else {
+            Err(TraceError::Other("When force flushing the SimpleSpanProcessor, the exporter's lock has been positioned".into()).into())
```
frigus02, line 143:
```suggestion
            Err(TraceError::Other("When force flushing the SimpleSpanProcessor, the exporter's lock has been poisoned".into()).into())
```

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -202,47 +211,56 @@ impl SpanProcessor for BatchSpanProcessor {
         }
     }
 
-    fn force_flush(&self) {
-        if let Ok(mut sender) = self.message_sender.lock() {
-            let _ = sender.try_send(BatchMessage::Flush);
+    fn force_flush(&self) -> ExportResult {
+        let mut sender = self.message_sender.lock().map_err(|_| TraceError::Other("When force flushing the BatchSpanProcessor, the message sender's lock has been positioned".into()))?;
```
frigus02, line 215:
```suggestion
        let mut sender = self.message_sender.lock().map_err(|_| TraceError::Other("When force flushing the BatchSpanProcessor, the message sender's lock has been poisoned".into()))?;
```

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -202,47 +211,56 @@ impl SpanProcessor for BatchSpanProcessor {
...
-    fn shutdown(&mut self) {
-        if let Ok(mut sender) = self.message_sender.lock() {
-            // Send shutdown message to worker future
-            if sender.try_send(BatchMessage::Shutdown).is_ok() {
-                if let Some(worker_handle) = self.worker_handle.take() {
-                    // Block waiting for worker to shut down if sending was successful
-                    futures::executor::block_on(worker_handle)
-                }
+    fn shutdown(&mut self) -> ExportResult {
+        let mut sender = self.message_sender.lock().map_err(|_| TraceError::Other("When force flushing the BatchSpanProcessor, the message sender's lock has been positioned".into()))?;
```
frigus02, line 227:
```suggestion
        let mut sender = self.message_sender.lock().map_err(|_| TraceError::Other("When force flushing the BatchSpanProcessor, the message sender's lock has been poisoned".into()))?;
```

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -357,15 +419,45 @@ impl BatchSpanProcessor {
...
+where
+    D: (Fn(time::Duration) -> DS) + Send + Sync + 'static,
+    DS: Future<Output = ()> + 'static + Send + Sync,
+    E: SpanExporter + ?Sized,
+{
+    let export = exporter.export(batch);
+    let timeout = delay(time_out);
+    pin_mut!(export);
+    pin_mut!(timeout);
+    // TODO: Surface error through global error handler
```
frigus02, line 454: I think the comment may be in the wrong place. This function returns the result. I think the reporting is necessary in places where we don't do anything with the result, e.g. in the Flush/Shutdown events of the batch span processor.

frigus02 (commented):
This looks like the correct approach to me. I didn't manage to review the tests, yet. Will try to do this later today.

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -125,13 +130,17 @@ impl SpanProcessor for SimpleSpanProcessor {
...
     }
 
-    fn force_flush(&self) {
+    fn force_flush(&self) -> ExportResult {
         // Ignored since all spans in Simple Processor will be exported as they ended.
+        Ok(())
     }
 
-    fn shutdown(&mut self) {
+    fn shutdown(&mut self) -> ExportResult {
```
flycash, line 138: Do we need to call `force_flush` in `shutdown`? From my understanding, `shutdown` need to do some cleanup before shutdown.

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -547,4 +670,108 @@ mod tests {
...
+    }
+
+    // If the time_out is true, then the result suppose to ended with timeout.
+    // otherwise the exporter should be able to export within time out duration.
+    async fn timeout_test_tokio(time_out: bool) {
+        let mut config = BatchConfig::default();
+        config.max_export_timeout = time::Duration::from_secs(if time_out { 5 } else { 60 });
+        config.scheduled_delay = Duration::from_secs(60 * 60 * 24); // set the tick to 24 hours so we know the span must be exported via force_flush
+        let exporter = BlockingExporter {
+            delay_for: time::Duration::from_secs(if !time_out { 5 } else { 60 }),
```
frigus02, line 755: I think 5 seconds is quite a long time for a test. What kind of delay do we need if we're testing that it does not time out? Would the smallest unit (e.g. 1ms) be enough?

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -547,4 +670,108 @@ mod tests {
...
+        let mut runtime = tokio::runtime::Builder::new()
+            .threaded_scheduler()
+            .enable_all()
+            .build()
+            .unwrap();
+        runtime.block_on(timeout_test_tokio(true));
+        runtime.block_on(timeout_test_tokio(false));
+
+        async_std::task::block_on(timeout_test_std_async(true));
+        async_std::task::block_on(timeout_test_std_async(false));
```
frigus02, line 717: It might be nice to make 4 different tests out of those. This makes it easier to figure out what went wrong if one of them fails.

opentelemetry/src/sdk/trace/span_processor.rs
TommyCpp: Yeah `ExportResult` is actually an alias for `Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>`. Will update to make this clear.

opentelemetry/src/sdk/trace/span_processor.rs
TommyCpp: You are right. The spec says shutdown must contains force_flush. But in `SimpleSpanProcessor`'s force_flush does nothing so there is really no point to call it.

frigus02 (approved):
There is a clippy warning:
> casting integer literal to `usize` is unnecessary

Otherwise this looks good to me 👍

flycash (approved):

jtescher:
Looks like cargo fmt check is failing still

opentelemetry/src/api/trace/mod.rs
```diff
@@ -136,3 +136,14 @@ pub use self::{
     },
     tracer::{SpanBuilder, Tracer},
 };
+
+use thiserror::Error;
```
jtescher, line 140: I don't think we have use statements in the middle of files in the rest of the codebase, may be clearer to move this to the top.

opentelemetry/src/api/trace/mod.rs
```diff
@@ -136,3 +136,14 @@ pub use self::{
...
+
+/// Errors returned by the trace API.
+#[derive(Error, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum TraceError {
```
jtescher, line 145: If we are adding this, should the results below be `Result<(), TraceError>` now?

opentelemetry/Cargo.toml
```diff
@@ -21,17 +21,17 @@ all-features = true
...
 
 [dependencies]
-async-std = { version = "1.6", features = ["unstable"], default-features = false, optional = true  }
+async-std = { version = "1.6", features = ["unstable"], default-features = false, optional = true }
 async-trait = { version = "0.1", optional = true }
 bincode = { version = "1.2", optional = true }
 dashmap = { version = "4.0.0-rc6", optional = true }
 fnv = { version = "1.0", optional = true }
-futures = "0.3"
+futures = { version = "0.3", features = ["std"] }
```
jtescher, line 29: isn't `std` a default feature? did you want to have `default-features = false`? or was this for something else?

opentelemetry/src/api/trace/mod.rs
TommyCpp: Thought it would be better to include as part of unify error PR. Probably need to change `Other(String)` into `Other(dyn Error + Sync + Send + 'static)` to include all possible errors.

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -125,13 +130,17 @@ impl SpanProcessor for SimpleSpanProcessor {
...
+        Ok(())
     }
 
-    fn shutdown(&mut self) {
+    fn shutdown(&mut self) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         if let Ok(mut exporter) = self.exporter.lock() {
             exporter.shutdown();
+            Ok(())
+        } else {
+            Err(TraceError::Other("When force flushing the SimpleSpanProcessor, the exporter's lock has been poisoned".into()).into())
```
jtescher, line 143: should this comment say when shutting down rather than force flushing?

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -180,7 +189,7 @@ impl SpanProcessor for SimpleSpanProcessor {
 /// [`async-std`]: https://async.rs
 pub struct BatchSpanProcessor {
     message_sender: Mutex<mpsc::Sender<BatchMessage>>,
-    worker_handle: Option<Pin<Box<dyn Future<Output = ()> + Send + Sync>>>,
+    _worker_handle: Option<Pin<Box<dyn Future<Output = ()> + Send + Sync>>>,
```
jtescher, line 192: do we still need this reference?

opentelemetry/src/api/trace/mod.rs
jtescher: Yeah probably a `Box<dyn Error + Send + Sync>` probably makes sense, could have a constructor at some point similar to https://doc.rust-lang.org/std/io/struct.Error.html#method.new

opentelemetry/src/sdk/trace/span_processor.rs
TommyCpp: Nope I guess we don't really need it. Will remove

opentelemetry/src/sdk/trace/span_processor.rs
TommyCpp: Yeah good catch!

opentelemetry/Cargo.toml
TommyCpp: Cannot really recall why I changed that. But seems unnecessary. Will revert this part.

jtescher:
Looks good thanks @TommyCpp, if there are changes we want to do for #364 we can do them in a subsequent PR

PR #365, opened 2020-11-20
Add wasm support to jaeger

boxdot:
This PR add another feature `wasm_collector_client` to the `opentelemetry-jaeger` crate which enable a WebAssembly collector.

Note: At the moment of writing Jaeger does not support Option Method requests, that are required by CORS: https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request. A simple solution is to put Jaeger behind a proxy and handle the request in the proxy.

### DIFF
--- a/opentelemetry-jaeger/src/collector.rs
+++ b/opentelemetry-jaeger/src/collector.rs
@@ -9,7 +9,7 @@ pub(crate) struct CollectorAsyncClientHttp {
     endpoint: Uri,
     #[cfg(feature = "collector_client")]
     client: isahc::HttpClient,
-    #[cfg(feature = "wasm_collector_client")]
+    #[cfg(all(feature = "wasm_collector_client", not(feature = "collector_client")))]
     client: WasmHttpClient,
     payload_size_estimate: AtomicUsize,
 }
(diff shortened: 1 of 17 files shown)


opentelemetry-jaeger/src/collector.rs
```diff
@@ -1,88 +1,247 @@
...
 
 /// `CollectorAsyncClientHttp` implements an async version of the
 /// `TCollectorSyncClient` interface over HTTP
 #[derive(Debug)]
 pub(crate) struct CollectorAsyncClientHttp {
     endpoint: Uri,
-    client: HttpClient,
+    #[cfg(feature = "collector_client")]
+    client: isahc::HttpClient,
+    #[cfg(feature = "wasm_collector_client")]
```
jtescher, line 12: These features have to be exclusive or else `--all-features` won't compile

boxdot: Thanks. I now enable the wasm feature only if the bare feature is not set.

jtescher:
@boxdot have you looked more at the linux foundation privacy / CLA situation?

boxdot:
> @boxdot have you looked more at the linux foundation privacy / CLA situation?

Signed. 

Looks like the test coverage got a hit. I did not add any wasm client tests. Technically, the wasm client code is also not compiled in CI since the the features are exclusive and the native client is preferred by default. Tell me if you prefer to have the CI changes in this PR, otherwise I would propose to make them in a consecutive review.

jtescher (approved):
Looks good, thanks @boxdot

jtescher:
Can do further improvements in subsequent PRs, this looks like a good place to start.

PR #366, opened 2020-11-21
fix(otlp): propagate Resource from Span to proto

lucacasonato:
Previously `with_resource` had no effect, because the keys were not being propagated into the ResourceSpan.﻿

### DIFF
--- a/opentelemetry-otlp/src/transform/traces.rs
+++ b/opentelemetry-otlp/src/transform/traces.rs
@@ -56,9 +56,16 @@ impl From<Link> for Span_Link {
 
 impl From<SpanData> for ResourceSpans {
     fn from(source_span: SpanData) -> Self {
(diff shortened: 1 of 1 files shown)


opentelemetry-otlp/src/transform/traces.rs
```diff
@@ -56,9 +56,16 @@ impl From<Link> for Span_Link {
 
 impl From<SpanData> for ResourceSpans {
     fn from(source_span: SpanData) -> Self {
+        let resource_attributes: Attributes = source_span
+            .resource
+            .clone()
```
jtescher, line 61: I don't believe this clone is necessary

jtescher (changes requested):
@lucacasonato thanks! just one small comment

jtescher (approved):

PR #367, opened 2020-11-21
feat(otlp): add openssl feature flags for grpcio

lucacasonato:
`grpcio` has an `openssl` and `openssl-vendored` feature to use `openssl` instead of `boringssl`. This PR exposes these features through an `openssl` and `openssl-vendored` feature﻿ in `opentelemetry-otlp`.


jtescher (changes requested):
Looks good, can we add the feature flag docs to lib.rs as well for consistency?

TommyCpp (approved):

jtescher:
Looks like just a `cargo fmt` issue remaining

jtescher (approved):

PR #368, opened 2020-11-23
 update ShouldSample's parent parameter to be Context

### LINKED ISSUE #290: Update ShouldSample's parent parameter to be Context
See https://github.com/open-telemetry/opentelemetry-specification/pull/881 for details.

drexler:
Resolves #290

### DIFF
--- a/opentelemetry/src/sdk/trace/sampler.rs
+++ b/opentelemetry/src/sdk/trace/sampler.rs
@@ -37,8 +37,10 @@
 //! cause gaps in the distributed trace, and because of this OpenTelemetry API
 //! MUST NOT allow this combination.
 
-use crate::trace::{Link, SpanContext, SpanKind, TraceId, TraceState};
-use crate::KeyValue;
+use crate::{
+    trace::{Link, SpanKind, TraceContextExt, TraceId, TraceState},
+    Context, KeyValue,
+};
 
 /// The `ShouldSample` interface allows implementations to provide samplers
 /// which will return a sampling `SamplingResult` based on information that
@@ -48,7 +50,7 @@ pub trait ShouldSample: Send + Sync + std::fmt::Debug {
     #[allow(clippy::too_many_arguments)]
     fn should_sample(
         &self,
-        parent_context: Option<&SpanContext>,
+        parent_context: Option<&Context>,
         trace_id: TraceId,
         name: &str,
         span_kind: &SpanKind,
@@ -98,7 +100,7 @@ pub enum Sampler {
 impl ShouldSample for Sampler {
     fn should_sample(
         &self,
-        parent_context: Option<&SpanContext>,
+        parent_context: Option<&Context>,
         trace_id: TraceId,
         name: &str,
         span_kind: &SpanKind,
@@ -116,7 +118,8 @@ impl ShouldSample for Sampler {
                     .should_sample(parent_context, trace_id, name, span_kind, attributes, links)
                     .decision,
                 |ctx| {
-                    if ctx.is_sampled() {
+                    let parent_span_context = ctx.span().span_context();
+                    if parent_span_context.is_sampled() {
                         SamplingDecision::RecordAndSample
                     } else {
                         SamplingDecision::Drop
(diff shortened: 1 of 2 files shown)


opentelemetry/src/sdk/trace/sampler.rs
```diff
@@ -116,7 +118,8 @@ impl ShouldSample for Sampler {
                     .should_sample(parent_context, trace_id, name, span_kind, attributes, links)
                     .decision,
                 |ctx| {
-                    if ctx.is_sampled() {
+                    let parent_span_context = ctx.span().span_context();
+                    if parent_span_context.is_valid() && parent_span_context.is_sampled() {
```
TommyCpp, line 122: > The Span's SpanContext may be invalid to indicate a root span

I think we should allow invalid span context here

drexler: You're right. I re-read the Built-in Samplers specifications and that seems implied. Thanks.

jtescher:
So looking further into the discussion in https://github.com/open-telemetry/opentelemetry-specification/pull/881, it looks like the intent may be to have access to the parent span itself from the context. May have to think about this more as currently `Extractor`s product a context with a remote span context, but not a remote span. The current patch will technically be in accordance with the api in the spec, but it is effectively still only exposing the span context.

TommyCpp:
> currently Extractors product a context with a remote span context, but not a remote span

I think it would be wired to build a `sdk::Span` for a remote span because in that case it's hard to assign a tracer to it. Maybe just [set the remote span context in context](https://github.com/open-telemetry/opentelemetry-rust/blob/master/opentelemetry/src/api/trace/context.rs#L83-L85) and pass it into Sampler?

jtescher:
@TommyCpp yeah that seems similar to what we currently have right? Not sure the best way to apply the spec change if they _do_ want to have access to full spans, also an option to simply not conform to the spec for this language.

TommyCpp:
I think at API level, passing `Context` gives the potential to determine whether to sample based on all available information in `Context`. That make sense because the `Extractor` can put anything into the `Context`. 

But for SDK I guess we are only use the `SpanContext` to determine whether to sample. That could come from either remote span or local parent span. Could we have some logic that check for `Span` in `Context` first, and if it's not available, we could use `RemoteSpanContext`?

jtescher:
Any preferences for the API here @awiede / @frigus02?

frigus02:
Sorry, personal life is a bit busy at the moment. I don't manage to do as much on the project. I hope that changes in a week or two.

On topic: I think you're right. The purpose of this change seems to be to have the entire parent span available.

It sounds like this requires additional changes anyway, like always accepting a `Context` to specify the parent. With that in mind, I think I would be happy with the PR. I would hope that this API can stay.

I didn't yet think about if it's feasible to implement these changes, though. So I'm okay with going in another direction, too.

TommyCpp (approved):
I don't think we need to change API even if we want to do a subsequent PR. So I guess we can merge it as it is.

opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -77,8 +77,14 @@ impl Tracer {
     ) -> Option<(u8, Vec<KeyValue>, TraceState)> {
         let provider = self.provider()?;
         let sampler = &provider.config().default_sampler;
+
+        let ctx = parent_context.map(|span_context| {
+            let span = Span::new(span_context.clone(), None, self.clone());
+            Context::new().with_span(span)
```
jtescher, line 83: If you add another parameter to this private method to pass through the `cx: &Context` from the caller, then at least the correct other context variables should be present. E.g. `tracer.build_with_context(builder, some_non_current_context)`

```suggestion
            cx.with_span(span)
```

jtescher (changes requested):
I think this can be merged with one small adjustment below, even if the intention of the spec is not followed, then as @TommyCpp mentioned it will at least not be a future breaking change. The performance changes look negligible from the trace benches so no real downside.

drexler:
Thanks for the review and feedback. I'll take a look wrap this up come the weekend.

jtescher (approved):
Looks good, thanks! we can work on the best ways of getting access to what is intended in the spec in subsequent PRs

Issue #286, opened 2020-10-18
Unify Errors


jtescher:
Currently the metrics module uses a `MetricsError` type while the trace module was created before the spec allowed for error results and therefore does not have any result types. Exporters are using variations of `std::error::Error`. We should consider introducing an `opentelemetry::Error` type and update the exporters to use it. `global::handle_error` should also accept this type, and the tracing module should call this function internally instead of dropping errors.

cpcloud:
Would it be appropriate to use `thiserror` library-wide?

jtescher:
@cpcloud yeah I think that probably makes the most sense

jtescher:
@cpcloud it could also be useful to re-evalutate the proper level of granularity, i.e. as a consumer of these APIs, what level of error information is worth responding to? Could for example go down to effectively two enum values, `Trace` and `Metrics`, and have data within those be opaque, or have some categories within those like an initialization error vs a runtime error, etc

cpcloud:
The way I would want to handle errors from this library (and related workspace crates) is:

1. `opentelemetry::Error` enum with as-specific-as-possible variants for all errors that are not specific to a framework (what are effectively the workspace crates)
1. `<workspace crate>::Error` enum with as-specific-as-possible variants for all errors pertaining to the particular framework (e.g., `opentelemetry_jaeger::Error`, etc)

That way, I can easily choose (as long as my crate is using `thiserror`!) to propagate errors up, or handle specific variants.

jtescher:
I'd be happy to review a PR to that effect, are there good example crates that define errors well with respect to granularity that you've found?

jtescher:
https://github.com/open-telemetry/opentelemetry-rust/issues/263 would likely be a part of this

danburkert:
Perhaps at the risk of piling on here, I'm finding `opentelemetry-jaeger` to be difficult to work with because the error type it returns `Box<dyn std::error::Error + ...>` doesn't impl `std::error::Error`.  If this isn't clear, see this compile failure: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f14fe95aeec58829b465ddad594f1e41

frigus02:
[...]

PR #371, opened 2020-11-26
Revisit error handling in trace

TommyCpp:
#### Current approaches
Currently, most of the errors in `trace` have been returned via `Box<dyn std::error::Error + Send + Sync + 'static>`. This approach gives users good flexibility to return basically any errors. But it totally depends on user to return meaningful messages in errors to help end user to understand the issue. 

#### Proposed approaches
As discussed in #286, we decided to use `thiserror` library wide to allow multiple level of errors throughout the whole process to send span. 

1. We define a `TraceError` enum to represent any possible errors from `trace` module. It provides some predefined variantion, and we should add more variantions as needed. It also provides two generic variantions `Other` and `Boxed`. `Other` could hold any string and `Boxed` could hold any `Box<dyn std::error::Error + Send + Sync + 'static>`. It allows the user to adapt their currently implementation easily. 

2. We define a `ExporterError` trait, which contains one method to return the name of the exporter. Exporter implementations are engouraged to define their own Error type and implement ExporterError. This allow any exporter implementations to pass their custom errors back to SDK. This should also help narrow down which exporter causes the error when errors being returned.

3. Change function signature to return `TraceError` or `ExporterError` when possible. And use global error handler to handle the rest of errors

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -165,17 +165,17 @@
 //! supported compiler version is not considered a semver breaking change as
 //! long as doing so complies with this policy.
 #![warn(
-future_incompatible,
-missing_debug_implementations,
-missing_docs,
-nonstandard_style,
-rust_2018_idioms,
-unreachable_pub,
-unused
+    future_incompatible,
+    missing_debug_implementations,
+    missing_docs,
+    nonstandard_style,
+    rust_2018_idioms,
+    unreachable_pub,
+    unused
 )]
 #![cfg_attr(docsrs, feature(doc_cfg), deny(broken_intra_doc_links))]
 #![doc(
-html_logo_url = "https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/master/assets/logo.svg"
+    html_logo_url = "https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/master/assets/logo.svg"
 )]
 #![cfg_attr(test, deny(warnings))]
 
@@ -194,6 +194,7 @@ use agent::AgentAsyncClientUDP;
 use async_trait::async_trait;
 #[cfg(feature = "collector_client")]
 use collector::CollectorAsyncClientHttp;
+use opentelemetry::exporter::trace::ExportError;
 use opentelemetry::{
     exporter::trace,
     global, sdk,
@@ -205,7 +206,6 @@ use std::{
     time::{Duration, SystemTime},
 };
 use uploader::BatchUploader;
-use opentelemetry::exporter::trace::ExportError;
 
 /// Default service name if no service is configured.
 const DEFAULT_SERVICE_NAME: &str = "OpenTelemetry";
@@ -357,8 +357,8 @@ impl PipelineBuilder {
     #[cfg(feature = "collector_client")]
     #[cfg_attr(docsrs, doc(cfg(feature = "collector_client")))]
     pub fn with_collector_endpoint<T>(self, collector_endpoint: T) -> Self
-        where
-            http::Uri: core::convert::TryFrom<T>,
+    where
+        http::Uri: core::convert::TryFrom<T>,
     {
         PipelineBuilder {
             collector_endpoint: core::convert::TryFrom::try_from(collector_endpoint).ok(),
@@ -393,7 +393,7 @@ impl PipelineBuilder {
     }
 
     /// Assign the process service tags.
-    pub fn with_tags<T: IntoIterator<Item=KeyValue>>(mut self, tags: T) -> Self {
+    pub fn with_tags<T: IntoIterator<Item = KeyValue>>(mut self, tags: T) -> Self {
         self.process.tags = tags.into_iter().collect();
         self
     }
@@ -409,7 +409,8 @@ impl PipelineBuilder {
     /// Install a Jaeger pipeline with the recommended defaults.
     pub fn install(
         self,
-    ) -> Result<(sdk::trace::Tracer, Uninstall), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    ) -> Result<(sdk::trace::Tracer, Uninstall), Box<dyn std::error::Error + Send + Sync + 'static>>
+    {
         let tracer_provider = self.build()?;
         let tracer =
             tracer_provider.get_tracer("opentelemetry-jaeger", Some(env!("CARGO_PKG_VERSION")));
@@ -422,7 +423,8 @@ impl PipelineBuilder {
     /// Build a configured `sdk::trace::TracerProvider` with the recommended defaults.
     pub fn build(
         mut self,
-    ) -> Result<sdk::trace::TracerProvider, Box<dyn std::error::Error + Send + Sync + 'static>> {
+    ) -> Result<sdk::trace::TracerProvider, Box<dyn std::error::Error + Send + Sync + 'static>>
+    {
         let config = self.config.take();
         let exporter = self.init_exporter()?;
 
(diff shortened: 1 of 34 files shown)


opentelemetry/src/api/trace/mod.rs
```diff
@@ -138,12 +139,50 @@ pub use self::{
...
+    #[error("Exporting failed with {0}")]
+    ExportFailed(Box<dyn ExportError>),
+
+    /// Export failed to finish after certain period and processor stopped the export.
+    #[error("Exporting timed out after {} seconds", .0.as_secs())]
+    ExportTimedOut(time::Duration),
+
+    /// Other errors propagated from trace SDK that weren't covered above
+    #[error(transparent)]
+    Boxed(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
```
jtescher, line 166: Could we combine this with `Other`? could look at `std::io::Error` or other crates to see best way of having catchall error type

opentelemetry/src/exporter/trace/mod.rs
```diff
@@ -10,28 +11,25 @@ use http::Request;
...
-/// Timed out when exporting spans to remote
-#[derive(Debug, Default)]
-pub struct ExportTimedOutError {}
+pub type ExportResult = Result<(), TraceError>;
 
-impl Display for ExportTimedOutError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str("export timed out")
+/// Marker trait for errors returned by exporters
+pub trait ExportError: std::error::Error + Send + Sync + 'static {
```
jtescher, line 26: I wonder if this should be moved to the parent module, i.e. should metrics / trace / log export errors be basically the same, or do they all need different export errors?

opentelemetry/src/exporter/trace/mod.rs
```diff
@@ -10,28 +11,25 @@ use http::Request;
...
+    /// The name of exporter that returned this error
+    fn exporter_name(&self) -> &'static str {
```
jtescher, line 28: Is there a reason that this has a default impl? are there unnamed exporters?

opentelemetry/src/exporter/trace/mod.rs
```diff
@@ -81,6 +79,65 @@ pub trait HttpClient: Debug + Send + Sync {
     async fn send(&self, request: Request<Vec<u8>>) -> ExportResult;
 }
 
+/// Error when sending http requests visa HttpClient implementation
+#[cfg(feature = "http")]
+#[cfg_attr(docsrs, doc(cfg(feature = "http")))]
+#[derive(Debug)]
+pub enum HttpClientError {
```
jtescher, line 86: Enum seems a bit awkward here as it is for 3rd party errors.

opentelemetry/src/exporter/trace/stdout.rs
```diff
@@ -130,16 +131,44 @@ where
...
-/// Uninstalls the stdout pipeline on drop.
+/// Uninstalls the stdout pipeline on dropping.
 #[derive(Debug)]
 pub struct Uninstall(global::TracerProviderGuard);
+
+/// Stdout exporter's error
+#[derive(Debug)]
+struct Error(std::io::Error);
+
+impl Display for Error {
```
jtescher, line 156: could likely use thiserror to remove all these impls too

jtescher (commented):
Nice start, couple of thoughts as the draft progresses

opentelemetry/src/exporter/trace/mod.rs
TommyCpp: It's more for the http client error from our HttpClient implementation, which is most like not realted to exporter itself. It's also difficult to get expoter name from httpclient.

opentelemetry-jaeger/src/lib.rs
```diff
@@ -431,7 +430,8 @@ impl PipelineBuilder {
     /// Build a configured `sdk::trace::TracerProvider` with the recommended defaults.
     pub fn build(
         mut self,
-    ) -> Result<sdk::trace::TracerProvider, Box<dyn Error + Send + Sync + 'static>> {
+    ) -> Result<sdk::trace::TracerProvider, Box<dyn std::error::Error + Send + Sync + 'static>>
```
jtescher, line 433: These could be `TraceError` too right? Would be nice to get the box dyn errors out of the api where possible

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -294,11 +292,13 @@ impl BatchSpanProcessor {
                                     &delay,
                                     batch,
                                 )
-                                .await,
+                                    .await,
```
jtescher, line 295: I think these moves are unintentional, doesn't seem like cargo fmt is working on this part? odd.

jtescher (commented):
Looking good! only a couple of small comments, also looks like the msrv tests are now broken by an async std dependency, maybe we could feature flag those tests so they only run if async-std feature is used?

jtescher:
Or maybe just remove async-std form msrv job in https://github.com/open-telemetry/opentelemetry-rust/blob/ac8e4e580a1c2403007fa1cab9767feaa30a4857/.github/workflows/ci.yml#L55 if that fixes it

opentelemetry-jaeger/src/lib.rs
```diff
@@ -447,7 +444,9 @@ impl PipelineBuilder {
     /// Initialize a new exporter.
     ///
     /// This is useful if you are manually constructing a pipeline.
-    pub fn init_exporter(self) -> Result<Exporter, Box<dyn Error + Send + Sync + 'static>> {
+    pub fn init_exporter(
+        self,
+    ) -> Result<Exporter, Box<dyn std::error::Error + Send + Sync + 'static>> {
```
jtescher, line 449: sorry can probably update this error as well

jtescher: hard to catch them all when reviewing 😬

TommyCpp: 😝 Sorry I should have caught those. It's just too many std errors there. Hard to find out all of them. I _think_ I get them all in zillion and jaeger this time. Let me know if I missed some

jtescher (approved):
small nits left but looking good, ready whenever out of draft state

opentelemetry/src/sdk/trace/span_processor.rs
jtescher: Couple more of these left, I wonder why cargo fmt doesn't rewrite these..

opentelemetry/Cargo.toml
```diff
@@ -22,6 +22,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 async-std = { version = "1.6", features = ["unstable"], default-features = false, optional = true }
+async-io = { version = "~1.2.0", optional = true }
```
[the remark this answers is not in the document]
TommyCpp: I tried to remove `async-std` but it didn't seem to resolve the problem. Could add some comments on this dependency to explain thing a little bit.

opentelemetry/src/sdk/trace/span_processor.rs
TommyCpp: I think I got all of them. It's stange that `cargo fmt` didn't correct those.

opentelemetry/Cargo.toml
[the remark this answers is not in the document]
jtescher: `$ cargo +1.42.0 test  --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio,serde,http,tonic,reqwest` works for me if I remove the `async-std` feature from the msrv ci yaml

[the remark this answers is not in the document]
TommyCpp: Try delete `Cargo.lock` maybe? Sometime it cached the 1.2.0 version of async-io, which works with Rust 1.42.0.
I tried to delete the async-std feature in [this commit](https://github.com/open-telemetry/opentelemetry-rust/pull/371/commits/118044481268bd7124a4213b6fdef7f727d06ab7) and the CI still managed to compile a 1.3.0 version of async-io.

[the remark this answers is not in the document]
jtescher: ah right, you have to add  `#[cfg(feature = "async-std")]` to the async-std specific tests in `span_processor.rs` and remove `async-std` from the dev dependencies as well. Then it should work without this.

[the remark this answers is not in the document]
TommyCpp: :+1: Updated. But according to MSRV policy in README. We might consider upgrade msrv to 1.43. I think this should be done via a separate PR for document purpose.

jtescher (approved):
Looks good!

PR #372, opened 2020-11-28
Add warning about assigning the uninstaller to '_'

arlyon:
It is common rust convention to use _ to signify an unused variable which, to new users who don't understand the inner workings, `_uninstall` is.

When integrating the library I spent an hour or so wondering why I was getting no logs until I realised that using `_` drops the value immediately after calling and _not_ when the `_` goes out of scope. Obviously this is my fault, but to prevent people from dealing with the same thing I thought a warning front-and-center would be nice.

In the long run an alternative to the `must_use` lint (`must_assign` or similar) may be a nice alternative.

If anyone has some suggestions on where to put this in the code itself rather than just the readme I'm happy to include the warning in the appropriate places.


frigus02 (approved):

README.md
```diff
@@ -37,6 +37,8 @@ use opentelemetry::{exporter::trace::stdout, trace::Tracer};
 
 fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
     // Create a new instrumentation pipeline
+    // Note: using `_` instead of assigning to a variable will immediately
+    // drop the tracer and uninstall it so make sure to properly assign it.
```
jtescher, line 41: May be clearer to have this say that it will drop the uninstall guard and make the tracer a no-op as the tracer won't be dropped unless you assign it to `_` as well.

jtescher:
Would be nice if a must assign was available here, but a good step would be to find all of the exporter's `Uninstall` structs and add `#[must_use]` to them. At least this would prevent calling install without assigning it to anything, even if assigning them to `_` still is a footgun.

README.md
arlyon: Ah yes, that is maybe not crystal clear.

jtescher (approved):
Looks great, thanks!

PR #373, opened 2020-11-29
Add actix-http-tracing example

jtescher:
Add an example for actix / tracing integration as this configuration seems to be common


TommyCpp (approved):
LGTM. Just need to fix MSRV problem.

PR #377, opened 2020-12-09
Corrects span StatusCode ordering

willroberts:
When using `opentelemetry` and `opentelemetry-contrib` to report traces to Datadog, I noticed that all spans are marked as errors by default. I was able to correct this by manually calling `span.set_status(StatusCode::Ok, ...)`, but I have since learned that the expectation (which the docs support) is that the default should be `Ok` instead.

This change updates the ordering to make `Ok` the default, and updates tests in the Datadog encoder to reflect this change.


jtescher (approved):
Good catch, thanks!

TommyCpp (approved):

willroberts:
I've completed the CLA! Merge at your convenience :)

PR #378, opened 2020-12-10
Switch Jaeger collectors to use http client trait.

### LINKED ISSUE #337: Switch jaeger collector client to use `HttpClient` trait
Should use the same pattern as Zipkin/Datadog exporters

TommyCpp:
closes #337 

Added 4 default collector http client implementations(`surf_collector_client`, `reqwest_collector_client`, `reqwest_blocking_collector_client`, `isahc_collector_client`). Note that with any of those enabled. The user can still override the http client setting with `with_http_client`

If the user chooses none of above. They will have to call `with_http_client` to set their own. Otherwise, it will throw a `No http client error`.

Upgrade Note:
- Change the feature from `collector_client` to `isahc_collector_client`.

### DIFF
--- a/opentelemetry-jaeger/Cargo.toml
+++ b/opentelemetry-jaeger/Cargo.toml
@@ -38,6 +38,7 @@ lazy_static = "1.4"
 
 [dev-dependencies]
 opentelemetry = { version = "0.10", default-features = false, features = ["trace", "testing"], path = "../opentelemetry" }
+futures = "0.3"
 
 [dependencies.web-sys]
 version = "0.3.4"
--- a/opentelemetry-jaeger/Cargo.toml
+++ b/opentelemetry-jaeger/Cargo.toml
@@ -35,6 +35,9 @@ wasm-bindgen = { version = "0.2", optional = true }
 wasm-bindgen-futures = { version = "0.4.18", optional = true }
 thiserror = "1.0"
 lazy_static = "1.4"
+reqwest = { version = "0.10", default-features = false, optional = true }
+headers = { version = "0.3.2", optional = true }
+surf = { version = "2.0", optional = true }
 
 [dev-dependencies]
 opentelemetry = { version = "0.10", default-features = false, features = ["trace", "testing"], path = "../opentelemetry" }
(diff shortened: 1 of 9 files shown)


opentelemetry-jaeger/Cargo.toml
```diff
@@ -54,7 +58,11 @@ optional = true
 
 [features]
 default = []
-collector_client = ["isahc", "http"]
+collector_client = ["http", "opentelemetry/http"]
+isahc_collector_client = ["isahc", "collector_client"]
```
jtescher, line 62: I don't think we actually need to support isahc? in the other exporters it's just part of the example for "bring your own" right?

jtescher (approved):
Looks great, just one comment

opentelemetry-jaeger/Cargo.toml
TommyCpp: Yes. But if we remove it, it would cause a breaking change for existing user. I think might as well just added a default implementation for `isahc`. So ideally existing users will only have to change the feature from `collector_client` to `isahc_collector_client` and everything should work just fine.

jtescher: Hm none of the isahc types were in the public api right? shouldn't codebases that were using it via the `collector_client` feature still work the same way now? There was never a way of passing the client manually so it shouldn't break any apps as far as I can see.

That being said we could also keep it and maybe mark as deprecated? Just seems like one we may not want to be maintaining over time

TommyCpp: Yep. So after merging this PR. The `collector_client` feature will contains no default http client. Users will have to call `with_http_client` explicitly to provide one otherwise it will throw a `NoHttpClient` error. That could break all of the current codebase.

Could mark `isahc_collector_client` to be deprecate and remove it later.

jtescher: Oh right, the ol' no reasonable default here, sounds fine then

opentelemetry-jaeger/README.md
```diff
@@ -109,15 +109,23 @@ example expects a Jaeger collector running on `http://localhost:14268`.
 
 ```toml
 [dependencies]
-opentelemetry-jaeger = { version = "..", features = ["collector_client"] }
+opentelemetry-jaeger = { version = "..", features = ["collector_client", "isahc"] }
```
jtescher, line 112: This should probably show the version that doesn't require manually setting the client

```suggestion
opentelemetry-jaeger = { version = "..", features = ["isahc_collector_client"] }
```

jtescher:
Looks good thanks @TommyCpp

PR #382, opened 2020-12-12
Zipkin - exclude status code if unset

### LINKED ISSUE #355: Zipkin: Do not send status unless ok or error
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/1186

drexler:
Fixes #355

### DIFF
--- a/opentelemetry-zipkin/src/exporter/model/mod.rs
+++ b/opentelemetry-zipkin/src/exporter/model/mod.rs
@@ -34,12 +34,13 @@ impl Into<annotation::Annotation> for Event {
     }
 }
 
-/// Converts StatusCode to str
-fn from_statuscode_to_str(status_code: StatusCode) -> &'static str {
+/// Converts StatusCode to Option<&'static str>
+/// `Unset` status code is unused.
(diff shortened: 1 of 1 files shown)


TommyCpp (changes requested):
Would be nice to indicate the `UNSET` should never be used here. 
https://github.com/open-telemetry/opentelemetry-rust/blob/d157ecffb26b34d0170671558a1dfe349b238646/opentelemetry-zipkin/src/exporter/model/mod.rs#L38-L44

drexler:
> Would be nice to indicate the `UNSET` should never be used here.
> 
> https://github.com/open-telemetry/opentelemetry-rust/blob/d157ecffb26b34d0170671558a1dfe349b238646/opentelemetry-zipkin/src/exporter/model/mod.rs#L38-L44

I had intentionally left that with the thinking that the function might be useful in future if there was a need for a stringified version of the `StatusCode` elsewhere. Anyway, this update should be more flexible. Thanks for the review.

TommyCpp (approved):
LGTM. Thanks！

PR #385, opened 2020-12-15
Add non monotonic counter support to prometheus

jtescher:
* Export non-monotonic and last value instruments as prometheus gauges
* Replace `PassThrough` export kind with `Stateless` export kind selector, allowing support of both cumulative and delta export kinds.


TommyCpp (approved):
Looks good 👍

PR #392, opened 2020-12-18
feat: add instrumentation version for instrument config.

TommyCpp:
Working on the otel metrics exporter. Think we should include instrumentation version in metrics as we did in tracing.


jtescher (approved):
looks good

PR #393, opened 2020-12-19
fix: use InstrumentationLibrary in metrics.

TommyCpp:
Currently we don't use InstrumentationLibrary to store the instrumentation library information. This should help unify things a little bit. My only concern is whether there will be any cases that users cannot use static string as argument(need to fetch intrumentation information somewhere else). So far I haven't thought any, so leave it as it is for now.


jtescher:
Yeah I think requiring the static lifetime is fine here

TommyCpp:
> Yeah I think requiring the static lifetime is fine here

👍 We can also switch to the owned string if someone brings a valid use case and that shouldn't break existing APIs.

jtescher (approved):

PR #395, opened 2020-12-23
fix: Counter adds the delta from last collection.

### LINKED ISSUE #394: prometheus: metrics keep increasing on their own
Hi folks! I'm experiencing a weird behaviour with the `opentelemetry-rust` crate and Actix Web.

Here's what's happening:

Whenever a metric is kicked off, it starts increasing on its own without any other interaction with the system. E.g. A counter increasing on every HTTP request keeps ticking up even if no requests are coming in.

I prepared a little [repro project](https://github.com/matteojoliveau/actix-prom-bug) with 3 examples: one using [actix-web-opentelemetry](https://docs.rs/actix-web-opentelemetry) (basically just a wrapper around `opentelemetry-prometheus` with an Actix middleware to record HTTP metrics and the `/metrics` endpoint implementation), one using `opentelemetry-prometheus` directly to expose the `/metrics` endpoint, and one using the [prometheus](https://docs.rs/prometheus) crate on its own,

The pure-prometheus example works fine, the other two present the described behaviour.

I'm having a hard time figuring out why this is happening, and if I'm either doing something wrong or if the problem is elsewhere.

Anyone has any idea?

TommyCpp:
Fix #394 

During each collection interval. The processor will maintain two places to store the values. `cumulative` and `current`. `current` won't get clear until the next time instrument gets updated. So if the instrument hasn't updated since the last collection. We should skip any change made to the `cumulative` part.

### DIFF
--- a/opentelemetry/src/sdk/metrics/processors/basic.rs
+++ b/opentelemetry/src/sdk/metrics/processors/basic.rs
@@ -215,9 +215,10 @@ impl Checkpointer for BasicLockedProcessor<'_> {
                 // If this processor does not require memory, stale, stateless
                 // entries can be removed. This implies that they were not updated
                 // over the previous full collection interval.
-                if stale && stateless && has_memory {
+                if stale && stateless && !has_memory {
                     return false;
                 }
+                return true;
             }
 
             // Update Aggregator state to support exporting either a


jtescher (approved):
Good catch

PR #396, opened 2020-12-26
Prepare for v0.11.0 release

jtescher:


TommyCpp (approved):
LGTM. But it looks like there are some warnings on deprecated methods.

jtescher:
Interesting, yeah looks like there are some new nightly clippy lints

jtescher:
I also think I may have a way to clean up the sampler's context vs span context change a bit. Will see if I can get a PR for that quickly too.

PR #398, opened 2020-12-27
fix: HistogramAggregator returns sum when asked for count

TommyCpp:
Find this bug when testing the otlp metric exporter. We still might need to add more tests around metrics. But I think this should fix the problem.


jtescher (approved):
Good catch. Yeah need lots of tests before this can stabilize.

PR #399, opened 2020-12-27
Store parent context in span builder

jtescher:
The span builder currently holds a parent `SpanContext`, and `Context` information is passed separately when starting a span directly or via a builder. This poses a few problems, first you can assign a context which contains an active span _as well as_ a builder with a parent span context, which can lead to confusion, and secondly we have to construct a wrapper context when sampling even in cases when the passed in context already contains an active span.

This patch resolves this by changing the span builder's parent context to instead store a `Context` directly. This allows the builder methods to have a single context to look at when searching for parent span contexts, and allows the wrapper context to only be created for sampling when it is facilitating a remote parent.

### DIFF
--- a/opentelemetry/src/sdk/trace/tracer.rs
+++ b/opentelemetry/src/sdk/trace/tracer.rs
@@ -67,26 +67,44 @@ impl Tracer {
     #[allow(clippy::too_many_arguments)]
     fn make_sampling_decision(
         &self,
-        parent_context: Option<&SpanContext>,
+        parent_cx: Option<&Context>,
+        parent_span_context: Option<&SpanContext>,
         trace_id: TraceId,
         name: &str,
         span_kind: &SpanKind,
         attributes: &[KeyValue],
         links: &[Link],
-        ctx: &Context,
     ) -> Option<(u8, Vec<KeyValue>, TraceState)> {
         let provider = self.provider()?;
         let sampler = &provider.config().default_sampler;
 
-        let ctx = parent_context.map(|span_context| {
-            let span = Span::new(span_context.clone(), None, self.clone());
-            ctx.with_span(span)
-        });
+        // Sampler needs to be able to access remote span context as parent span
+        let sampling_result = if parent_cx.map_or(false, |cx| {
+            !cx.has_active_span() && cx.remote_span_context().is_some()
+        }) {
+            let parent_cx = parent_cx.map(|cx| {
+                let span = Span::new(
+                    parent_span_context
+                        .cloned()
+                        .unwrap_or_else(SpanContext::empty_context),
+                    None,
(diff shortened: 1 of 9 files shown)


opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -67,26 +67,44 @@ impl Tracer {
...
-
-        let sampling_result =
-            sampler.should_sample(ctx.as_ref(), trace_id, name, span_kind, attributes, links);
+        // Sampler needs to be able to access remote span context as parent span
+        let sampling_result = if parent_cx.map_or(false, |cx| {
+            !cx.has_active_span() && cx.remote_span_context().is_some()
+        }) {
+            let parent_cx = parent_cx.map(|cx| {
+                let span = Span::new(
+                    parent_span_context
```
TommyCpp, line 87: Maybe use `cx.remote_span_context()` would be more intuitive? I think if `!cx.has_active_span() && cx.remote_span_context().is_some()` is true. Parent span context should be equal to the remote span context in parent context.

opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -218,13 +234,13 @@ impl crate::trace::Tracer for Tracer {
             self.process_sampling_result(sampling_result, parent_span_context)
         } else if no_parent || remote_parent {
```
TommyCpp, line 235: Maybe we can separate those two situations. If there is no parent, we build a parent context with the invalid span context. If there is a remote parent, we build a temporary span with the remote parent. Then we pass that context into `make_sampling_decision` or just use the sampler to make the sample decision.

TommyCpp (commented):
I feel a little confusing that `parent_span_context` is kind of part of the `parent_ctx`, and it feels like we pass the duplicate element into the `make_sample_decision` function. But overall I think it looks good.

opentelemetry/src/sdk/trace/tracer.rs
jtescher: hm yeah the paths are a little mixed. Trying to not to duplicate the span context extraction, and not clone the context unnecessarily. Might be a clearer way.

jtescher: Simplified it a bit, now there is only one param for sampling

TommyCpp (approved):
LGTM 👍

opentelemetry/src/sdk/trace/tracer.rs
jtescher: yeah this could have been a path too, was trying to do it without having to check before hand to know which one of these cases it was. But as this is all internal we should be able to keep iterating on this in the future without breaking changes

PR #401, opened 2020-12-28
Use default context if unset for builders

jtescher:
The previous change to the builder actually introduced an issue if a builder was used _without_ explicitly specifying its parent context. This is resolved here by defaulting to the current context if unset.

If you want to create a new root span, you can still explicitly set the builder's parent context to be a new context.


jtescher:
@TommyCpp ran through all the examples, and some external crates like `tracing-opentelemetry` and `actix-web-opentelemetry` and this was the only issue I found, so should be good to release after this fix 😅

TommyCpp (approved):
Sorry didn't get a chance to look at it earlier. LGTM :+1:. Good catch!

PR #402, opened 2020-12-29
feat: otel metric exporter.

TommyCpp:
-  Add support for OTLP metrics exporter with `tokio` and `tonic` only.
-  Add transform from record to metric proto types. Need to support `grpcio` in the future.
-  Add `OtlpMetricExporterBuilder` pipeline. Need to figure out a way to combine it with the tracing pipeline. The current blocker is the type parameter from the push controller. Ideally, we gonna need to rewrite the whole pipeline so that when users enable `tracing` only, there will be no type parameter, when users enable the `metrics`, the pipeline will have type parameters.
- Testing has been hard as the OTLP's structure is nested and unordered in general. Need to figure out a way to compare `ResourceMetrics`. 

Probably gonna address the remaining work in other PRs. Thought this is a big enough PR to be reviewed.


examples/basic-otlp/src/main.rs
```diff
@@ -92,5 +97,8 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         });
     });
 
+    // wait for 1 seconds so that we could see metrics being pushed via OTLP every 10 seconds.
+    tokio::time::delay_for(Duration::from_secs(60)).await;
```
jtescher, line 101: Are these not pushed on drop?

jtescher (approved):

examples/basic-otlp/src/main.rs
TommyCpp: Currently, the metric export in `Exporter` trait is a sync function, so I have to use another tokio task to actually send the request. We could add a reply channel in the request and block on the reply channel to do force push/push or drop.
But no we don't have it now. Though this should come after we add timeout logic.

Issue #397, opened 2020-12-27
`actix-http-tracing` example not working as expected


gnieto:
I was trying the `actix-http-tracing` example on this repo and I found that the second request to `/metrics` do not respond.

Steps to reproduce:

1- cargo run
2- `curl http://localhost:8080/users/a`
3- `curl http://localhost:8080/metrics` (this request responds as expected, with the Prometheus metrics)
4- `curl http://localhost:8080/metrics` (this request do not respond)

CPU goes up to 100% after step 4.

My Rust version is `rustc 1.48.0-nightly (f68e08933 2020-09-19)`

TommyCpp:
Thanks for reporting this issue. Although I tried the example but failed to reproduce the problem. I have the following responses.
```shell
> rustc -V
rustc 1.48.0-nightly (f68e08933 2020-09-19)
> curl http://localhost:8080/users/a
Hello a%                                                                                               
> curl http://localhost:8080/metrics
# HELP http_requests_duration HTTP request duration per route
# TYPE http_requests_duration histogram
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="0.5"} 1
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="0.9"} 1
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="0.99"} 1
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="+Inf"} 1
http_requests_duration_sum{method="GET",route="/users/{username}",status="200"} 0.001045649
http_requests_duration_count{method="GET",route="/users/{username}",status="200"} 1
# HELP http_requests_total HTTP requests per route
# TYPE http_requests_total counter
http_requests_total{method="GET",route="/users/{username}",status="200"} 1
> curl http://localhost:8080/metrics
# HELP http_requests_duration HTTP request duration per route
# TYPE http_requests_duration histogram
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="0.5"} 2
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="0.9"} 2
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="0.99"} 2
http_requests_duration_bucket{method="GET",route="/users/{username}",status="200",le="+Inf"} 2
http_requests_duration_sum{method="GET",route="/users/{username}",status="200"} 0.002091298
[...]

PR #407, opened 2020-12-31
Fix dropped metrics instruments

jtescher:
Currently metrics can deadlock when dropping the last reference to an instrument. This patch addresses this by switching to `https://docs.rs/dashmap/4.0.1/dashmap/struct.DashMap.html#method.retain`.

Related #397

### DIFF
--- a/opentelemetry-prometheus/tests/integration_test.rs
+++ b/opentelemetry-prometheus/tests/integration_test.rs
@@ -7,6 +7,33 @@ use opentelemetry_prometheus::PrometheusExporter;
 use prometheus::{Encoder, TextEncoder};
 
 #[test]
+fn free_unused_instruments() {
+    let exporter = opentelemetry_prometheus::exporter()
+        .with_default_histogram_boundaries(vec![-0.5, 1.0])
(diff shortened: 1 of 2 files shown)


opentelemetry-prometheus/tests/integration_test.rs
```diff
@@ -6,6 +6,33 @@ use opentelemetry::{
 use opentelemetry_prometheus::PrometheusExporter;
 use prometheus::{Encoder, TextEncoder};
 
+#[test]
+fn free_unused_instruments() {
```
TommyCpp, line 10: Hmm for some reason I run this test against `master` and it works. Not sure if I missed something.

jtescher: This behavior seems to have changed between dashmap version `4.0.0-rc6` and dashmap version `4.0.0` which was released this week. If you cargo update you should see the issue.

TommyCpp: Ha make sense.  LGTM

TommyCpp (approved):

PR #412, opened 2021-01-04
opentelemetry: dissolve api module in favor of top-level modules

djc:
This makes the source tree less nested and makes the internal API and
filesystem more similar to the public API exposed by the crate.


jtescher:
This looks good to me. Anyone have reservations or other structural ideas we should consider? It deviates from the recommended otel project layout, but I think that's fine.

djc:
To me the `api` vs `sdk` split feels very unnatural and confusing, I think they should probably just be collapsed, with top-level `metrics` and `trace` modules that contain all the respective stuff plus any other stuff that's orthogonal to trace vs metrics also on the top level.

TommyCpp:
>  It deviates from the recommended otel project layout

I think there are some other implementations not following the recommended layout for different reasons. Golang for example. I think as long as we separate sdk and api types and allow users to bring their own SDK, we should be good.

jtescher (approved):

Issue #403, opened 2020-12-29
Move integration code out of the main opentelemetry crate


djc:
Into separate integration crates. This makes for more smaller crates, but makes sure we don't have dependencies in the core that we shouldn't have (which causes to have to upgrade in lockstep with their releases).

djc:
Things that probably qualify:

* async-std
* bincode?
* regex?
* tokio
* tonic
* reqwest
* surf

jtescher:
👍 could also consider http and the testing utils as candidates

TommyCpp:
For proto definition. I'd suggest we move it to a separate it into separate crate(`opentelemetry-proto` maybe)? And support `grpcio` or `tonic` there.

djc:
If we need to support grpcio *and* tonic, those should probably be two separate crates IMO.

djc:
I spent the morning looking at this. The `Injector` and `Extractor` interfaces are tricky, because coherence requires that implementations for foreign types live in the same crate as the trait definitions. I did some experiments with using propagator types that implemented `Extractor<T>` (with `T` something like `http::HeaderMap`), but this clashes with the design of the global propagator, which cannot easily abstract over a particular type of propagator.

I did remove the regex dependency (#411, in favor of a small function that validates the trace state key) and moved bincode to dev-dependencies (#413, along with some other dependency management).

djc:
#415 moves the dependencies on http, reqwest and surf into a separate opentelemetry-http crate. At this point I'm inclined to leave tokio and async-std in, as they're pretty stable anyway (especially now that tokio has gone 1.0).

jtescher:
@djc is the work related to this issue complete? or should we keep this issue open

TommyCpp:
I want to mention that we need to update the doc in `opentelemetry/lib.rs`. Especially the `Crate Feature Flags` part.
https://github.com/open-telemetry/opentelemetry-rust/blob/f79586b4caae79d26abd3e6ba69fa230c7c68f0e/opentelemetry/src/lib.rs#L36-L54

djc:
@TommyCpp thanks for taking care of the documentation update.

I think this ticket is complete, closing.

TommyCpp:
@jtescher @djc I don't think we published the `opentelemetry-aws` and `opentelemetry-datadog`. Should we do that?

jtescher:
@TommyCpp yeah I think we could probably cut a new release, want to prepare the changelogs?

PR #414, opened 2021-01-04
opentelemetry: remove tonic dependency

djc:
See #403. This just moves the code into an example -- I'm not sure what else would be relying on these impls?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         override: true
     - name: Run tests
       run: cargo --version &&
-        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio,serde,http,tonic,reqwest,testing &&
+        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio,serde,http,reqwest,testing &&
         cargo test --manifest-path=opentelemetry-jaeger/Cargo.toml &&
         cargo test --manifest-path=opentelemetry-zipkin/Cargo.toml
   meta:
--- a/examples/grpc/Cargo.toml
+++ b/examples/grpc/Cargo.toml
@@ -16,7 +16,7 @@ http = "0.2"
 tonic = "0.3"
 prost = "0.6"
 tokio = { version = "0.2", features = ["full"] }
-opentelemetry = { path = "../../opentelemetry", features = ["tonic"] }
+opentelemetry = { path = "../../opentelemetry" }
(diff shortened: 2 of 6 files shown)


jtescher (approved):
Think this was it. Was added for integration convenience to avoid tonic users having to copy this into all their apps I believe, but likely better as external crate or just copy/paste as locking the version here is probably makes the whole thing not worth it.

djc:
Good. Shall I get rid of `http` too, then?

jtescher:
Sounds good 👍

TommyCpp:
IMHO we still should provide the tonic integration as the external crates if we can so that it can be used out of the box. Usually, Users would likely want to spend less time on telemetry compared with business logic, which means they should be able to do it with less configuration.

djc:
It's not clear how that would work with the current definition of the `Extractor`/`Injector` traits, though. (Also, I think those names could be improved -- they're more like `Injectable` and `Extractable`.)

jtescher:
@djc yeah the names come from the propagators api spec https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#injectors-and-extractors-as-separate-interfaces, we don't need to have the names match exactly if we can add clarity. The rust naming convention for single method traits seems to be more aligned with the method name like `ToString::to_string` or `Clone::clone`, we currently follow that other places like `ShouldSample::should_sample`, could consider doing that for inject and extract.

djc:
I think my point on naming is a little different and potentially orthogonal to what you're saying. The problem that had me a little confused at first is that IIRC Rust traits that are named for a verb can actually "do the thing" (so you can `read()` on `Read` for example), whereas with `Injector` and `Extractor` the injecting/extracting is done "to the thing" that implements the traits. This is where I thought it might make sense to change the interface, but ran aground on the notion of a global propagator (not sure exactly how/why that gets used): if we defined an `Inject` interface that takes a type param for the type it can inject into, we'd be able to solve the coherence issues.

jtescher:
Yeah the global propagator is tricky to express. E.g. if an actix-web or rocket middlware crate wants to be able to extract headers for a given app, then it needs a way to find the globally configured propagator so it knows which headers to extract (b3, jaeger, w3c trace context, some combination, etc). Thus the current global method for getting a reference to an impl of the `TextMapPropagator` trait which can inject via a `dyn Injector` and extract via a `dyn Extractor` (which to your point may be easier to understand as `Injectable` and `Extractable`).

Having the global propagator generically accessible for instrumentation anywhere in the app seems like the challenge for the type system currently.

TommyCpp (approved):

PR #415, opened 2021-01-05
Move HTTP-related code into a separate crate

djc:
See #403.

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -161,6 +161,54 @@ impl HttpClient for surf::Client {
     }
 }
 
+#[cfg(feature = "isahc")]
+impl ExportError for IsahcError {
+    fn exporter_name(&self) -> &'static str {
+        "isahc"
+    }
+}
+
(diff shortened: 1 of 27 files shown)


jtescher (approved):
LGTM

jtescher:
Can add the usual crate files license, readme, etc as well or in a follow up pr

opentelemetry-http/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+    #[derive(Debug)]
+    struct SurfError(surf::Error);
+
+    impl ExportError for SurfError {
+        fn exporter_name(&self) -> &'static str {
+            "surf"
+        }
+    }
+
+    #[cfg(feature = "surf")]
```
TommyCpp, line 133: Do we still need this even the mod is guarded by a `#[cfg(feature = "surf")]`?

djc: We don't, good catch! Fixed it in the latest version.

TommyCpp (approved):
Nice work! Thanks a lot

PR #418, opened 2021-01-10
feat: add the function to assert ResourceMetrics

TommyCpp:
This should give us more accurate test result on `test_sink`. Also move all util functions before test functions.


jtescher (approved):
LGTM

PR #421, opened 2021-01-12
Upgrade Tokio to v1

### LINKED ISSUE #420: Upgrade to Tokio 1.0
rust-opentelemetry currently depends on Tokio 0.2. [Tokio 1.0 released on December 22](https://tokio.rs/blog/2020-12-tokio-1-0) and some of the ecosystem has started moving over to it ([hyper](https://github.com/hyperium/hyper/blob/master/CHANGELOG.md#v0140-2020-12-23), [reqwest](https://github.com/seanmonstar/reqwest/blob/HEAD/CHANGELOG.md#v0110), [rusoto](https://github.com/rusoto/rusoto/blob/master/CHANGELOG.md#0460---2021-01-05), etc.).

mraerino:
*Fixes #420*

I also upgraded these dependencies:
- reqwest
- hyper
- tonic
- prost

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
     - uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: 1.42.0
+        toolchain: 1.46.0
         override: true
     - name: Run tests
       run: cargo --version &&
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         override: true
     - name: Run tests
       run: cargo --version &&
-        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio,serde,http,reqwest,testing &&
+        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio-support,serde,http,reqwest,testing &&
         cargo test --manifest-path=opentelemetry-jaeger/Cargo.toml &&
         cargo test --manifest-path=opentelemetry-zipkin/Cargo.toml
   meta:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
         override: true
     - name: Run tests
       run: cargo --version &&
-        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio-support,tokio,serde,testing &&
+        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio-support,serde,testing &&
         cargo test --manifest-path=opentelemetry-jaeger/Cargo.toml &&
         cargo test --manifest-path=opentelemetry-zipkin/Cargo.toml
   meta:
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ analysis in order to understand your software's performance and behavior. You
 can export and analyze them using [Prometheus], [Jaeger], and other
 observability tools.
 
-*Compiler support: [requires `rustc` 1.42+][msrv]*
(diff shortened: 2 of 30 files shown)


mraerino:
the minimum support rust version for tokio is 1.45: https://github.com/tokio-rs/tokio#supported-rust-versions

do you think we could raise the version here as well? 1.42 seems pretty old

TommyCpp:
> do you think we could raise the version here as well? 1.42 seems pretty old

I think we can. We promised three versions before current stable version, which is 1.49. So we can bump our minimum supported version to 1.45 or 1.46

mraerino:
This is ready for a review now.
I expect this to need a major version bump since a feature flag changed and people can't update automatically via semver since they need to upgrade their tokio version too

jtescher:
@mraerino thanks for working on this, as far as MSRV is concerned we could go either way, for example the `async-std` feature currently doesn't support 1.42 either. Maybe we should just remove the tokio support from the "supported" features for 1.42 and the associated ci workflow and bump it in a subsequent PR? Nice to support folks on older versions in the more stable core even if opt-in features can't support them.

Also pre-1.0 minor changes often include breaking changes, so not as much of a problem there until we hit GA

mraerino:
> Maybe we should just remove the tokio support from the "supported" features for 1.42 and the associated ci workflow and bump it in a subsequent PR?

tokio is used in the tests even if zero features are enabled. i don't think what you propose is possible with the current CI flow

djc:
> Maybe we should just remove the tokio support from the "supported" features for 1.42 and the associated ci workflow and bump it in a subsequent PR? Nice to support folks on older versions in the more stable core even if opt-in features can't support them.

Alternatives:

* Remove tokio (and probably also async-std) from the core crate (as discussed in #403)
* Keep an opentelemetry branch alive for a bit longer with 0.11.x maintenance releases

The reason I hadn't done the upgrade here is there are also some tonic dependencies, and tonic support for tokio 1 has not yet been released. So if we want to prevent spawning multiple runtimes in the background, this PR should (a) wait until tonic 0.4 has been released -- judging from chat activity, this should come pretty soon, or (b) change crates depending on tonic to depend on a released version of opentelemetry, sticking to tokio 0.2 for now.

mraerino:
thanks for pointing out tonic! totally did not see that.
the PR is merged so once they release it we can move forward: https://github.com/hyperium/tonic/pull/530

mraerino:
i made https://github.com/open-telemetry/opentelemetry-rust/pull/424 to test the otlp exporter closer to reality. With that test, CI would be red right now since tonic fails to run in a Tokio v1 executor.

djc:
FYI, tonic 0.4 has been released.

mraerino:
yup, i saw. merging master and upgrading tonic

mraerino:
can you retry the CI workflow? it's a ratelimit failure.

.github/workflows/ci.yml
```diff
@@ -48,11 +48,11 @@ jobs:
     - uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: 1.42.0
+        toolchain: 1.46.0
         override: true
     - name: Run tests
       run: cargo --version &&
-        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio,serde,testing &&
+        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio-support,tokio,serde,testing &&
```
djc, line 55: `tokio` is implied by `tokio-support`, so we don't need both here:

```suggestion
        cargo test --verbose --manifest-path=opentelemetry/Cargo.toml --features trace,metrics,serialize,tokio-support,serde,testing &&
```

djc (commented):
Other than the nit below, this is looking good to me!

djc (approved):

mraerino:
@djc @jtescher is there a timeline for releasing this?

jtescher:
ASAP

PR #422, opened 2021-01-12
Adds Send + Sync to Async instruments (metrics)

markdingram:
- #280 added this to sync instruments previously


jtescher (approved):

TommyCpp (commented):
There are some places in sdk uses `Arc<dyn AsyncInstrumentCore + Send + Sync>` now. Could simplify it to `Arc<dyn AsyncInstrumentCore>`
For example, https://github.com/open-telemetry/opentelemetry-rust/blob/37b64b45e1c0e63053b088345826060142326c5f/opentelemetry/src/sdk/metrics/mod.rs#L81-L88

jtescher:
@TommyCpp good catch, yeah those should get cleaned up.

markdingram:
> @TommyCpp good catch, yeah those should get cleaned up.

Happy to do this!

PR #424, opened 2021-01-13
Test otlp exporter in integration scenario

mraerino:
This is contributing a test that hardens the suite to detect issues when using the otlp exporter in a real-world scenario.

It uses a small mock server that is able to receive the trace exports via plaintext http gRPC. The alternative would have been to spin up the original otlp-collector in docker, but that introduces more dependencies and is harder to assert on.

(This could have detected breakage in #421 which i missed.)

### DIFF
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -30,7 +30,6 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-async-stream = { version = "0.3.0", optional = true }
 async-trait = "0.1"
 futures = "0.3"
 grpcio = { version = "0.7", optional = true }
@@ -54,7 +53,7 @@ tls = ["tonic/tls"]
 tls-roots = ["tls", "tonic/tls-roots"]
 openssl = ["grpcio/openssl"]
 openssl-vendored = ["grpcio/openssl-vendored"]
-integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "async-stream", "opentelemetry/trace"]
+integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "opentelemetry/trace"]
 
 [build-dependencies]
 protobuf-codegen = { version = "2.16", optional = true }
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -15,15 +15,22 @@ keywords = ["opentelemetry", "otlp", "logging", "tracing", "metrics"]
 license = "Apache-2.0"
 edition = "2018"
 build = "build.rs"
+autotests = false
 
 [lib]
 doctest = false
+
+[[test]]
+name = "smoke"
+path = "tests/smoke.rs"
+required-features = ["integration-testing"]
 
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
+async-stream = { version = "0.3.0", optional = true }
 async-trait = "0.1"
 futures = "0.3"
 grpcio = { version = "0.7", optional = true }
@@ -47,6 +54,7 @@ tls = ["tonic/tls"]
 tls-roots = ["tls", "tonic/tls-roots"]
 openssl = ["grpcio/openssl"]
 openssl-vendored = ["grpcio/openssl-vendored"]
+integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "async-stream", "opentelemetry/trace"]
 
 [build-dependencies]
 protobuf-codegen = { version = "2.16", optional = true }
--- a/opentelemetry-otlp/src/proto.rs
+++ b/opentelemetry-otlp/src/proto.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "tonic")]
-pub(crate) mod collector {
+pub mod collector {
     pub mod metrics {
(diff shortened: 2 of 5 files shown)


mraerino:
looks like the `meta` CI task is broken?

TommyCpp:
I think we ran too many Github actions in a short amount of time that we have API exceed limit. Could try rerun it later

opentelemetry-otlp/Cargo.toml
```diff
@@ -15,15 +15,22 @@ keywords = ["opentelemetry", "otlp", "logging", "tracing", "metrics"]
...
+name = "smoke"
+path = "tests/smoke.rs"
+required-features = ["integration-testing"]
+
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
+async-stream = { version = "0.3.0", optional = true }
```
djc, line 33: I think you ended up using tokio-stream for #421, is there a reason for not using that here?

opentelemetry-otlp/src/proto.rs
```diff
@@ -1,5 +1,5 @@
 #[cfg(feature = "tonic")]
-pub(crate) mod collector {
+pub mod collector {
```
djc, line 2: An alternative could be just to have the integration in test in `src/tests.rs`, do you think that could work? It would prevent having to have the `integraion-testing` feature and all the conditionally public APIs.

djc (commented):
Will be great to have this covered in CI, thanks for working on it!

opentelemetry-otlp/Cargo.toml
mraerino: good point. the TcpListenerStream was only recently added so i missed it. since this PR is supposed to work with tokio 0.2 too i cannot add it yet though. will do once we are on tokio 1

opentelemetry-otlp/src/proto.rs
mraerino: yeah, i've been thinking about this a bit.

the part that i really don't like is the change i need to apply to the build script. I would love to just have my own build script for the integration test which would make all of this go away, since i would be able to generate the grpc server myself.

since we don't want to build the gRPC server when not in the integration-test i don't really see a good way to get rid of the feature flag though.

mraerino:
could someone trigger a CI rerun?

opentelemetry-otlp/src/proto.rs
djc: We could just have a separate crate for the integration test, right?

opentelemetry-otlp/Cargo.toml
mraerino: i was able to get rid of the dependency since `TcpListener` is a stream on tokio 0.2 - i will use tokio-stream in the tokio v1 PR once this is merged.

opentelemetry-otlp/src/proto.rs
djc: Couldn't we use `cfg(test)` to conditionally build the gRPC server only for test scenarios?

mraerino: oh, good point.

any ideas how a separate crate would integrate with the cargo test runner?

djc: Just have a separate crate with an integration test in it?

mraerino: i explored the unit test idea, but looks like `cfg!(test)` does not work in `build.rs` - there are some open cargo issues about this.

mraerino:
i have no idea what's wrong with the `meta` flow 🤔

opentelemetry-otlp/src/proto.rs
mraerino: is the separate crate what you'd prefer over my current solution?

djc: Actually a separate crate would probably not solve it, it'd still need the internal APIs exposed to the test crate. So that probably doesn't make sense, then.

opentelemetry-otlp/Cargo.toml
```diff
@@ -15,10 +15,16 @@ keywords = ["opentelemetry", "otlp", "logging", "tracing", "metrics"]
...
 build = "build.rs"
+autotests = false
 
 [lib]
 doctest = false
 
+[[test]]
+name = "smoke"
+path = "tests/smoke.rs"
+required-features = ["integration-testing"]
```
djc, line 26: Instead of this, we could just add `#![cfg(feature = "integration-testing")]` to the test, any thoughts on which is better?

mraerino: i guess if you are ok with the conditionally public proto mod, i'd leave it like this. this is where integration tests belong.

opentelemetry-otlp/src/proto.rs
mraerino: it would work, b/c i would have a separate build script using the same proto files. i don't think it would need any special modules

djc:
Okay, I think I'm satisfied this is pretty much the best we can do. One more thing left, though: I do think this actually enables the new test in CI?

mraerino:
i think it did run: https://github.com/open-telemetry/opentelemetry-rust/pull/424/checks?check_run_id=1701227622#step:5:1507

the test scripts runs tests on all crates with `--all-features` i guess?

djc:
Ah, fair enough.

djc (approved):

PR #425, opened 2021-01-13
Missing metrics creator fns

markdingram:
Looks a bit messy, first commit is a consistent reordering of the existing creator fns.
 
Second commit adds a few that were missing.

### DIFF
--- a/opentelemetry/src/metrics/meter.rs
+++ b/opentelemetry/src/metrics/meter.rs
@@ -76,7 +76,7 @@ impl Meter {
         UpDownCounterBuilder::new(self, name.into(), NumberKind::F64)
     }
 
-    /// Creates a new `ValueRecorderBuilder` for `f64` values with the given name.
+    /// Creates a new floating point `ValueRecorderBuilder` for `f64` values with the given name.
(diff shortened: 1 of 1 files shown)


opentelemetry/src/metrics/meter.rs
```diff
@@ -120,41 +99,80 @@ impl Meter {
         )
     }
 
-    /// Creates a new integral `SumObserverBuilder` instrument with the given name
-    /// and callback
-    pub fn u64_sum_observer<T, F>(&self, name: T, callback: F) -> SumObserverBuilder<'_, u64>
+    /// Creates a new floating point `UpDownSumObserverBuilder` instrument with the
```
TommyCpp, line 102: Emm, I think we should probably change it like below as `UpDownSumObserver` is an instrument but `UpDownSumObserverBuilder` technically isn't.
```suggestion
    /// Creates a `UpDownSumObserverBuilder` for the floating point `UpdownSumObserver` instrument with the
```

djc (approved):

jtescher:
Code changes look good, can either address docs first or in follow up PR

opentelemetry/src/metrics/meter.rs
markdingram: I did a pass through all the comments for consistency.

PR #428, opened 2021-01-15
fix: with_collector_endpoint is error prune.

### LINKED ISSUE #426: PipelineBuilder::with_collector_endpoint API is error prone
When passing a malformed URI to [`PipelineBuilder::with_collector_endpoint`](https://github.com/open-telemetry/opentelemetry-rust/blob/b2ab7996c6dcc5f0df6ee4185570d156f7c969ff/opentelemetry-jaeger/src/exporter/mod.rs#L211), the failure to parse the URI is silently discarded and trace collection is not enabled.  This is easy to mess up, for instance by passing `127.0.0.1:14268/api/traces` instead of `http://127.0.0.1:14268/api/traces`.

TommyCpp:
Fix #426.

Now we will store the result of try_from function until we init the uploader. If the result is error, we wrap it into TraceError and return.


jtescher (approved):
Looks good @TommyCpp

PR #429, opened 2021-01-17
metrics: implement batch observer

jtescher:
This patch adds an implementation of the metrics batch observer. The API is not identical to the example in [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#batch-observer) as it is inconvenient to register metrics after they are moved into the observer callback in rust. Instead the `Meter::batch_observer` method accepts a closure in which instruments can be registered before being moved into the callback.

```rust
meter.batch_observer(|batch| {
    let inst = batch.u64_sum_observer("example").init();

    move |result| {
        result.observe(&[KeyValue::new("a", "1")], inst.observation(42)]);
    }
});
```

### DIFF
--- a/opentelemetry/src/metrics/meter.rs
+++ b/opentelemetry/src/metrics/meter.rs
@@ -1,7 +1,7 @@
 use crate::sdk::InstrumentationLibrary;
 use crate::{
     metrics::{
-        sdk_api, AsyncRunner, BatchObserver, BatchObserverCallback, CounterBuilder, Descriptor,
+        sdk_api, AsyncRunner, BatchObserver, BatchObserverResult, CounterBuilder, Descriptor,
         Measurement, NumberKind, ObserverResult, Result, SumObserverBuilder, UpDownCounterBuilder,
         UpDownSumObserverBuilder, ValueObserverBuilder, ValueRecorderBuilder,
     },
@@ -93,7 +93,7 @@ impl Meter {
         SumObserverBuilder::new(
             self,
             name.into(),
-            AsyncRunner::F64(Box::new(callback)),
+            Some(AsyncRunner::F64(Box::new(callback))),
             NumberKind::F64,
         )
     }
@@ -111,7 +111,7 @@ impl Meter {
         UpDownSumObserverBuilder::new(
             self,
             name.into(),
-            AsyncRunner::F64(Box::new(callback)),
+            Some(AsyncRunner::F64(Box::new(callback))),
             NumberKind::F64,
         )
     }
@@ -125,7 +125,7 @@ impl Meter {
         ValueObserverBuilder::new(
             self,
             name.into(),
-            AsyncRunner::F64(Box::new(callback)),
+            Some(AsyncRunner::F64(Box::new(callback))),
             NumberKind::F64,
         )
     }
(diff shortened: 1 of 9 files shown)


opentelemetry/src/metrics/meter.rs
```diff
@@ -265,15 +265,75 @@ impl Meter {
...
+    /// # }
+    /// ```
+    pub fn build_batch_observer<B, F>(&self, builder: B) -> Result<()>
+    where
+        B: Fn(BatchObserver<'_>) -> Result<F>,
+        F: Fn(BatchObserverResult) + Send + Sync + 'static,
+    {
+        let observer = builder(BatchObserver::new(self))?;
+        self.core
+            .new_batch_observer(AsyncRunner::Batch(Box::new(observer)))
```
djc, line 302: I'd probably just inline the call to `builder()` here, so you don't have to name the intermediate.

opentelemetry/src/metrics/meter.rs
```diff
@@ -265,15 +265,75 @@ impl Meter {
...
+    /// });
+    /// ```
+    pub fn batch_observer<B, F>(&self, builder: B)
+    where
+        B: Fn(BatchObserver<'_>) -> F,
+        F: Fn(BatchObserverResult) + Send + Sync + 'static,
+    {
+        let observer = builder(BatchObserver::new(self));
+        self.core
+            .new_batch_observer(AsyncRunner::Batch(Box::new(observer)))
```
djc, line 335: Same here.

opentelemetry/src/sdk/metrics/mod.rs
```diff
@@ -79,9 +79,14 @@ struct MapKey {
...
 #[derive(Default, Debug)]
 struct AsyncInstrumentState {
-    /// runners maintains the set of runners in the order they were
-    /// registered.
-    runners: Vec<(AsyncRunner, Arc<dyn sdk_api::AsyncInstrumentCore>)>,
+    /// The set of runners in the order they were registered.
+    ///
+    /// Runners and instruments are 1:1 except batch observers, which have a
+    /// single runner for all their instruments.
+    runners: Vec<(AsyncRunner, Option<Arc<dyn sdk_api::AsyncInstrumentCore>>)>,
```
djc, line 86: This seems complex enough that it might be nice to define a separate type for it.

djc (commented):
I don't really know the semantics of this code, so just some style remarks.

The nested closure patterns seems a little ugly, but I can't think of a good way around it right now.

TommyCpp (approved):
:+1: LGTM

jtescher:
> The nested closure patterns seems a little ugly, but I can't think of a good way around it right now.

@djc yeah it's not ideal, not sure that there is better syntax for this type of setup currently though. Can iterate on this API if someone comes up with a cleaner way of expressing this as metrics still are not stable.

opentelemetry/src/metrics/meter.rs
jtescher: Are you suggesting this specifically?
```rust
self.core
    .new_batch_observer(AsyncRunner::Batch(Box::new(builder(BatchObserver::new(
        self,
    ))?)))
```

opentelemetry/src/sdk/metrics/mod.rs
jtescher: extracted type alias for this

opentelemetry/src/metrics/meter.rs
djc: Yeah. Maybe the way rustfmt treats it's not actually an improvement in this case...

djc (approved):

Issue #276, opened 2020-10-16
Add `with_unit` option to metric instrument builders


jtescher:
Currently there is no good way to associate units with instruments, this method should be added to all the metric instrument builders.

jtescher:
Resolved in #431

PR #431, opened 2021-01-18
Add with_unit method for metrics API

ozerovandrei:
Add `with_unit` method to `CounterBuilder`, `SumObserverBuilder`,
`UpDownSumObserverBuilder`, `ValueObserverBuilder` structs.

Add `set_unit` method to `Descriptor` struct.

For #276

### DIFF
--- a/opentelemetry/src/metrics/descriptor.rs
+++ b/opentelemetry/src/metrics/descriptor.rs
@@ -1,6 +1,5 @@
 use crate::metrics::{InstrumentConfig, InstrumentKind, NumberKind};
 use crate::sdk::InstrumentationLibrary;
-use crate::Unit;
 use fnv::FnvHasher;
 use std::hash::{Hash, Hasher};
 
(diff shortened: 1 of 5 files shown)


TommyCpp (approved):
LGTM 👍 Thanks!

opentelemetry/src/metrics/descriptor.rs
```diff
@@ -73,6 +74,11 @@ impl Descriptor {
         self.config.unit.as_ref().map(|unit| unit.as_ref())
     }
 
+    /// Assign a new unit.
+    pub fn set_unit(&mut self, unit: Unit) {
```
jtescher, line 78: Hm we probably don't want this in the public api. Maybe marking the descriptor properties as `pub(crate)` would be simpler?

ozerovandrei: I removed this method and made the `config` property `pub(crate)` to access it directly in builders.

jtescher (approved):
Looks good thanks @ozerovandrei

PR #432, opened 2021-01-18
metrics: remove extra instrument builders

jtescher:
This reverts some of the extra instruments introduced in #425. Specifically `Meter::u64_up_down_counter` and `Meter::u64_up_down_sum_observer` which are not necessary because of their type (e.g. a monotonically increasing  `up_down_counter` is simply a `counter`).


djc:
Should we keep the alphabetic order introduced at the same time? Seems like a nice touch.

jtescher:
> Should we keep the alphabetic order introduced at the same time? Seems like a nice touch.

Hard to say, mixes up the instruments so the builders are no longer grouped which may make the "which type is supported by which instrument" question less clear. The current order has the advantage that it matches the instrument order of the documentation.

djc (approved):
Fair enough.

PR #436, opened 2021-01-22
opentelemetry: improve the 0.12 changelog

djc:
The api module (not a crate) was already private, so that entry doesn't seem to be useful for most users. Additionally, this missed the removal of the http-related types and implementations, which probably is relevant to a larger audience.


TommyCpp (approved):
Good catch! Sorry I missed it earlier

PR #440, opened 2021-01-27
[datadog] Add DatadogPropagator

samschlegel:
This adds support for using the Datadog propagation format, pulled from `datadog-trace-go`. We're already using this in production and it seems to be working great!

### DIFF
--- a/opentelemetry-contrib/src/trace/propagator/datadog.rs
+++ b/opentelemetry-contrib/src/trace/propagator/datadog.rs
@@ -1,8 +1,8 @@
 use opentelemetry::{
     propagation::{text_map_propagator::FieldIter, Extractor, Injector, TextMapPropagator},
     trace::{
-        SpanContext, SpanId, TraceContextExt, TraceId, TraceState, TRACE_FLAG_NOT_SAMPLED,
-        TRACE_FLAG_SAMPLED,
+        SpanContext, SpanId, TraceContextExt, TraceId, TraceState, TRACE_FLAG_DEFERRED,
+        TRACE_FLAG_NOT_SAMPLED, TRACE_FLAG_SAMPLED,
     },
     Context,
 };
@@ -10,8 +10,7 @@ use opentelemetry::{
 const DATADOG_TRACE_ID_HEADER: &str = "x-datadog-trace-id";
 const DATADOG_PARENT_ID_HEADER: &str = "x-datadog-parent-id";
 const DATADOG_SAMPLING_PRIORITY_HEADER: &str = "x-datadog-sampling-priority";
-// TODO: Implement origin propagation once the Context API is stable
-// const HTTP_HEADER_ORIGIN: &'static str = "x-datadog-origin";
+
 lazy_static::lazy_static! {
     static ref DATADOG_HEADER_FIELDS: [String; 3] = [
         DATADOG_TRACE_ID_HEADER.to_string(),
@@ -26,11 +25,12 @@ enum SamplingPriority {
     AutoKeep = 1,
     UserKeep = 2,
 }
+
+#[derive(Debug)]
 enum ExtractError {
     TraceId,
     SpanId,
     SamplingPriority,
-    SpanContext,
 }
 
 /// Extracts and injects `SpanContext`s into `Extractor`s or `Injector`s using Datadog's header format.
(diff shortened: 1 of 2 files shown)


opentelemetry-contrib/src/trace/propagator/datadog.rs
```diff
@@ -0,0 +1,154 @@
...
+        TRACE_FLAG_SAMPLED,
+    },
+    Context,
+};
+
+const DATADOG_TRACE_ID_HEADER: &'static str = "x-datadog-trace-id";
+const DATADOG_PARENT_ID_HEADER: &'static str = "x-datadog-parent-id";
+const DATADOG_SAMPLING_PRIORITY_HEADER: &'static str = "x-datadog-sampling-priority";
+// TODO: Implement origin propagation once the Context API is stable
+// const HTTP_HEADER_ORIGIN: &'static str = "x-datadog-origin";
```
samschlegel, line 14: I need to check if this is stable now (or what it's for for that matter), and how this would integrate with OpenTelemetry.

Might be worth just pulling this out into an issue to track for later.

jtescher (commented):
Looks good, just a few clippy warnings to clean up

opentelemetry-contrib/src/trace/propagator/datadog.rs
jtescher: Yeah could open an issue for this one

opentelemetry-contrib/src/trace/propagator/datadog.rs
```diff
@@ -0,0 +1,154 @@
...
+        };
+
+        let trace_state = TraceState::default();
+
+        let span_context = SpanContext::new(trace_id, span_id, sampled, true, trace_state);
+
+        if span_context.is_valid() {
+            Ok(span_context)
+        } else {
+            Err(ExtractError::InvalidSpanContext)
```
TommyCpp, line 112: Could use[ global error handler](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/global/error_handler.rs) to handle those errors as we cannot propagate it from `TextMapPropagator`

TommyCpp (approved):
It looks good, could add some unit tests in this PR or do it in another one. Thanks for working on this

opentelemetry-contrib/src/trace/propagator/datadog.rs
samschlegel: Do you meant use the Error enum in there instead?

samschlegel: Or I guess at least impl Into<error_handler::Error>, and then manually call `handle_error`?

TommyCpp: Yeah we usually wrap the error into TraceError so users can choose how to handle them

samschlegel: I'm a bit hesitant to throw this into `handle_error` as I feel like this could be *VERY* high volume, and wouldn't want to have this just spam stderr for every trace that runs into this error. Exporter errors at least have the benefit of typically only happening once per batch interval where this would happen on every RPC call.

But I do think having this log *somehow* would be good instead of just silently eating the error. Not sure what the best move is here.

TommyCpp: Good point. And currently, we don't have many logs in place, might need to look into that. I think we could leave it as it is and add structured logs once we have it

samschlegel:
Added tests! Will leave the error handling as is so not to spam stderr in the case of malformed headers.

samschlegel:
It looks like CI didn't run any of the tests on that last commit though? Maybe it's just slow?

jtescher:
Looks like both passed. Just need to merge in main unless @TommyCpp has other comments

TommyCpp:
> Looks like both passed. Just need to merge in main unless @TommyCpp has other comments

Nope it all LGTM. Although I think the last commit didn't trigger any check

jtescher:
Oh true was only the CLA check that passed. Looks like github actions are experiencing an outage currently https://www.githubstatus.com/incidents/tf9v5jjmq2lg

samschlegel:
@jtescher looks like it's all run and passed now 😄

jtescher:
ok want to rebase or merge changes from main? (otel repos are set up to require up to date branches to merge)

PR #443, opened 2021-01-28
[datadog] group exported spans by trace_id

samschlegel:
While not technically required, not doing this causes issues with the datadog-trace-agent rate limits, and will lead to dropped spans when the span trace exceeds the agent-configured trace rate limit.

This also set the X-Datadog-Trace-Count header which is used by the agent to enforce this rate limit quickly, to report to the exporter that the rate limit was exceeded, and to update internal metrics for tracking trace reporting statistics  https://github.com/DataDog/datadog-agent/blob/master/pkg/trace/api/api.go#L442

TODO: 
- [x] Add some tests that actually test this grouping beyond more than one span in the export

### DIFF
--- a/opentelemetry-contrib/Cargo.toml
+++ b/opentelemetry-contrib/Cargo.toml
@@ -22,7 +22,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 default = []
 base64_format = ["base64", "binary_propagator"]
 binary_propagator = []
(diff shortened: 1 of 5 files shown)


samschlegel:
Not sure where this codecov reduction is coming from 🤔 `codecov/patch` succeeds so maybe just classic flaky code coverage

jtescher:
Yeah the codecov config needs to be updated. Too touchy currently.

jtescher (approved):
LTMT

PR #444, opened 2021-01-28
Remove tracer provider guard.

### LINKED ISSUE #364: Batch span processor shutdown deadlock on non-futures runtime
This line https://github.com/open-telemetry/opentelemetry-rust/blob/310df359a0b58b4377b9f8cf6ee3512cd15573db/opentelemetry/src/sdk/trace/span_processor.rs#L217 deadlocks a `BatchSpanProcessor` if it was spawn on a different runtime than the `futures` crate executor.

Test to reproduce: https://github.com/boxdot/opentelemetry-rust/commit/63dd30c4600ef3b0e03ab126167406f09af1eaa1

There is another place with `block_on` src/sdk/trace/span_processor.rs:124.

A proper fix would be to use `.await`, however both functions (`on_end`/`shutdown`) in the `SpanProcessor` are not marked as async. Another possibility to fix it would be to provide a custom await-function along with `spawn` function. However, this does not work with `SimpleSpanProcessor`. And at last, we could just shutdown and not wait. The processor will send the data away in any case and shutdown, since the sender is gone. 

I personally prefer the last solution. It is also the only one which is not API-breaking. If we decide on a solution, I can provide a fix.

P.S.: Thank you very much for this project! I am very happy that an implementation exists.

### LINKED ISSUE #427: PipelineBuilder::install & Uninstall APIs are error prone
When `Uninstall` drops it disables trace collection, which forces callers to use `mem::forget` in instances where global trace collection is enabled in a function other than `main`.  The [primary example](https://github.com/tokio-rs/tracing/blob/master/examples/examples/opentelemetry.rs) I've found of using opentelemetry-jaeger does not point this out, and it can be very difficult to debug why traces are not being collected when basing code off that example and moving it into a new function.

TommyCpp:
Resolve #427 
Resolve #364 

## Problem statement

The problem we have is how to allow users to send remaining spans before exits. The current approach works for multiple thread tokio runtime, but not really working for the current thread tokio runtime.

Also currently we allow the users to use multiple tracer providers throughout the application, which I believe is not a common situation. Most people should have one global tracer provider.

## Proposed solution

We currently have the `set_tracer_provider` function to set the tracer provider. It currently will return a guard, the drop of the guard will trigger the shutdown of the tracer provider. 

the function instead of returning the guard, we return the replaced tracer provider. If users need to use multiple tracer providers. They can manage those tracer providers on their own.

In order to shut down tracer provider properly. We could have a `shut_down_provider` function that will block in place until the current tracer provider shuts down. The problem then becomes that how do we spawn the background task so that it will not block forever when shuts down.

### Different situations

The objective is if users `shut_down_provider` before exit, the remaining spans in `BatchSpanProcessor` would sending all remaining spans. If not, then all remaining spans in it will be dropped.

Tried a few different set up to see if we can do it. 

| Runtime | How to spawn worker task in `BatchProcessor` | Call `shut_down_provider` before exit? | Result | 
| -----------|----------------|------------------|------|
| Single thread tokio | `tokio::spawn` | no | exit successfully | 
| Multiple thread tokio | `tokio::spawn` | no | exit successfully | 
| Single thread tokio | `tokio::spawn` | yes | hang forever |
| Multiple thread tokio | `tokio::spawn` | yes | send spans and exit |
| Single thread tokio | spawn_blocking | no | hang forever | 
| Multiple thread tokio | spawn_blocking | no | hang forever | 
| Single thread tokio | spawn_blocking | yes | send spans and exit  |
| Multiple thread tokio | spawn_blocking | yes | send spans and exit |

\* spawn_blocking = `|fut| tokio::spawn_blocking(|| futures::executor::block_on(fut))`

So my theory on why `opentelemetry` hang forever when using the single thread tokio:

when `shutdown` `BatchSpanProcessor`. We need to wait for the worker task to finish. But by blocking here, we prevent the worker task from being polled. If we don't block and wait here. When the `shutdown` finishes, the runtime will go ahead and drop the worker task, so we can't guarantee the spans in `BatchSpanProcessor` can be sent.
https://github.com/open-telemetry/opentelemetry-rust/blob/f06f38b5c33ed28c5eda1459b3efa0ff8c67e961/opentelemetry/src/sdk/trace/span_processor.rs#L232-L240

I have been proking around and tried random stuff on it but none of them seem to work. So would love some advice here as I believe some of you might have more experience on how tokio works.

### Final solution
- `set_tracer_provider` function should return the replaced tracer provider.
- `shut_down_tracer_provider` function will be used to shut down the tracer provider.
- `tokio::spawn` will be used to spawn background tasks in BatchProcessor if the runtime has multiple threads.
- If tokio runtime only have one thread. We will create a new thread and a new runtime to spawn background tasks.

### DIFF
--- a/opentelemetry/src/global/trace.rs
+++ b/opentelemetry/src/global/trace.rs
@@ -381,10 +381,10 @@ mod tests {
     #[test]
     #[ignore]
     fn test_set_tracer_provider() {
-        let _guard = set_tracer_provider(TestTracerProvider::new("global one"));
+        let _ = set_tracer_provider(TestTracerProvider::new("global one"));
 
         {
-            let _guard = set_tracer_provider(TestTracerProvider::new("inner one"));
+            let _ = set_tracer_provider(TestTracerProvider::new("inner one"));
             assert!(format!("{:?}", tracer_provider()).contains("inner one"));
         }
 
@@ -394,7 +394,7 @@ mod tests {
     #[test]
     #[ignore]
     fn test_set_tracer_provider_in_another_thread() {
-        let _guard = set_tracer_provider(TestTracerProvider::new("global one"));
+        let _ = set_tracer_provider(TestTracerProvider::new("global one"));
 
         let handle = thread::spawn(move || {
             assert!(format!("{:?}", tracer_provider()).contains("global one"));
@@ -409,7 +409,7 @@ mod tests {
     #[ignore]
     fn test_set_tracer_provider_in_another_function() {
         let setup = || {
-            let _guard = set_tracer_provider(TestTracerProvider::new("global one"));
+            let _ = set_tracer_provider(TestTracerProvider::new("global one"));
             assert!(format!("{:?}", tracer_provider()).contains("global one"))
         };
 
--- a/opentelemetry/src/global/trace.rs
+++ b/opentelemetry/src/global/trace.rs
@@ -276,7 +276,7 @@ where
 
 /// Shut down the current tracer provider. This will invoke shut_down method on all span processors.
 /// span processors should export remaining spans before return
-pub fn shut_down_provider() {
+pub fn shutdown_tracer_provider() {
     let mut tracer_provider = GLOBAL_TRACER_PROVIDER
         .write()
         .expect("GLOBAL_TRACER_PROVIDER RwLock poisoned");
@@ -462,7 +462,7 @@ mod tests {
 
             tracer.in_span("test", |_cx| {});
 
-            shut_down_provider();
+            shutdown_tracer_provider();
 
             assert!(assert_writer.len() > 0);
         }
@@ -495,7 +495,7 @@ mod tests {
         #[ignore]
         async fn test_set_provider_single_thread_tokio_shutdown() {
             let _ = test_set_provider_in_tokio().await;
-            // shut_down_provider(); Will block forever
+            // shutdown_tracer_provider(); Will block forever
         }
 
         // Test if the multiple thread tokio runtime could exit successfully when not force flushing spans
@@ -511,7 +511,7 @@ mod tests {
         #[ignore]
         async fn test_set_provider_multiple_thread_tokio_shutdown() {
             let assert_writer = test_set_provider_in_tokio().await;
-            shut_down_provider();
+            shutdown_tracer_provider();
             assert!(assert_writer.len() > 0);
         }
 
--- a/opentelemetry/src/global/trace.rs
+++ b/opentelemetry/src/global/trace.rs
@@ -294,12 +294,12 @@ pub fn shutdown_tracer_provider() {
 mod tests {
     use super::*;
     use crate::trace::NoopTracer;
-    #[cfg(any(feature = "tokio-support", feature = "tokio-rt-current-thread"))]
+    #[cfg(any(feature = "rt-tokio", feature = "rt-tokio-current-thread"))]
     use crate::trace::Tracer;
     use std::fmt::Debug;
-    #[cfg(any(feature = "tokio-support", feature = "tokio-rt-current-thread"))]
+    #[cfg(any(feature = "rt-tokio", feature = "rt-tokio-current-thread"))]
     use std::io::Write;
-    #[cfg(any(feature = "tokio-support", feature = "tokio-rt-current-thread"))]
+    #[cfg(any(feature = "rt-tokio", feature = "rt-tokio-current-thread"))]
     use std::sync::Mutex;
     use std::thread;
     use std::thread::sleep;
@@ -308,7 +308,7 @@ mod tests {
     macro_rules! cfg_tokio {
         ($($item:item)*) => {
             $(
-                #[cfg(any(feature = "tokio-support", feature = "tokio-rt-current-thread"))]
+                #[cfg(any(feature = "rt-tokio", feature = "rt-tokio-current-thread"))]
                 $item
             )*
         }
@@ -488,7 +488,7 @@ mod tests {
     // Test if the multiple thread tokio runtime could exit successfully when not force flushing spans
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     #[ignore]
-    #[cfg(feature = "tokio-support")]
+    #[cfg(feature = "rt-tokio")]
     async fn test_set_provider_multiple_thread_tokio() {
         let assert_writer = test_set_provider_in_tokio().await;
         assert_eq!(assert_writer.len(), 0);
@@ -497,7 +497,7 @@ mod tests {
     // Test if the multiple thread tokio runtime could exit successfully when force flushing spans
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     #[ignore]
-    #[cfg(feature = "tokio-support")]
+    #[cfg(feature = "rt-tokio")]
     async fn test_set_provider_multiple_thread_tokio_shutdown() {
         let assert_writer = test_set_provider_in_tokio().await;
         shutdown_tracer_provider();
@@ -508,7 +508,7 @@ mod tests {
     // Expected to see the spans being exported to buffer
     #[tokio::test]
     #[ignore]
-    #[cfg(feature = "tokio-rt-current-thread")]
+    #[cfg(feature = "rt-tokio-current-thread")]
     async fn test_set_provider_single_thread_tokio_with_simple_processor() {
         let assert_writer = AssertWriter::new();
         let _ = set_tracer_provider(build_tracer_provider(false, assert_writer.clone()));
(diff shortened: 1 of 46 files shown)


TommyCpp:
So for single thread runtime. One solution that would do the trick is spawn a new thread and a new runtime on it to run the batch processor. That solve the deadlock issue. For example
```rust
      let spawn = |box_future: BoxFuture<'static, ()>| {
            thread::spawn(move || {
                let rt = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
                rt.block_on(box_future);
            });
        };
```

djc:
That seems like a fine solution!

jtescher:
May want to consider if any of these potentially deadlock if used in tests which execute in parallel.

TommyCpp:
> May want to consider if any of these potentially deadlock if used in tests which execute in parallel.

I don't think it will cause deadlock as we guard the global tracer provider with a lock. But if two tests try to modify the global tracer provider at the same time, the result will likely be non-deterministic. Thus, I run tests in `global/trace.rs` sequentially.

TommyCpp:
Still gonna need some tests, examples, and docs. But the function changes should be good to review.

opentelemetry/src/global/trace.rs
```diff
@@ -285,9 +268,276 @@ where
...
+        let (sender1, sender2) = (sender.clone(), sender);
+        let _handle1 = thread::spawn(move || {
+            sleep(Duration::from_secs(1));
+            let _guard = set_tracer_provider(TestTracerProvider::new("thread 1"));
+            sleep(Duration::from_secs(2));
+            let _ = sender1.send(format!("thread 1: {:?}", tracer_provider()));
+        });
+        let _handle2 = thread::spawn(move || {
+            sleep(Duration::from_secs(2));
+            let _guard = set_tracer_provider(TestTracerProvider::new("thread 2"));
```
frigus02, line 434: This no longer returns a guard. Should we rename the variable to sth like `_previous` or just `_`?

opentelemetry/src/global/trace.rs
```diff
@@ -285,9 +268,276 @@ where
...
+    #[cfg(feature = "tokio-rt-current-thread")]
+    async fn test_set_provider_single_thread_tokio_with_simple_processor() {
+        use crate::sdk::trace::TracerProvider;
+
+        let assert_writer = AssertWriter::new();
+
+        let exporter =
+            crate::sdk::export::trace::stdout::Exporter::new(assert_writer.clone(), true);
+        let provider = TracerProvider::builder()
+            .with_simple_exporter(exporter)
```
frigus02, line 512: I like those tests. I found them mostly easy to read and understand what they test.

I wonder if we could move the tracer provider initialization here into a similar function as `test_set_provider_in_tokio`. It might seem unnecessary because we would only use it once. But it took me a bit to figure out why this test looks different than the others. The test name indicates that. But having a shorter function body might make it even easier.

frigus02 (commented):
This is great!

I used `opentelemetry` recently for a small CLI and ran into 2 issues:

- Using the Tokio current thread runtime
- Waiting for remaining spans to export when also using `std::process::exit(1);` to specify an exit code

I'm beginning to think that I would prefer an async shutdown function that I could call explicitly. It seems to be the hardest thing to accidentally misuse. I had worked about the above mentioned issues by doing something like the following. When I tried to use this PR I added the shutdown function and it didn't work:

```rust
async fn main() -> i32 {
    // install_pipeline ... create spans ...
}
fn main() {
    let exit_code = tokio::runtime::Builder::new_multi_thread()
        .enable_all()
        .build()
        .unwrap()
        .block_on(async_main);
    // opentelemetry::global::shutdown_tracer_provider(); <-- works without this line on master. spans don't export when I add this line using this PR
    std::process::exit(exit_code);
}
```

The error I got was "send failed because receiver is gone" (using the Jaeger exporter). I think this makes sense because I tried to shutdown the pipeline after shutting down the Tokio runtime. If `shutdown_tracer_provider` was async, I wouldn't have tried this code because I couldn't have awaited the future.

That said, though, this PR already seems like a huge improvement. Both of the following _much simpler_ versions of my main function worked just fine based on this PR.

```rust
#[tokio::main]
async fn main() {
    let exit_code = async_main().await;
    pipeline::shutdown_pipeline();
    std::process::exit(exit_code);
}
```

```rust
#[tokio::main(flavor = "current_thread")]
async fn main() {
    let exit_code = async_main().await;
    pipeline::shutdown_pipeline();
    std::process::exit(exit_code);
}
```

So to me this is definitely a nice improvement. And having the tests to make sure we don't regress is nice as well. 👍

opentelemetry/src/lib.rs
```diff
@@ -44,22 +46,31 @@
...
-//! [tonic]: https://crates.io/crates/tonic
-//! [`HttpClient`]: crate::sdk::export::trace::HttpClient
-//! [reqwest]: https://crates.io/crates/reqwest
-//! [surf]: https://crates.io/crates/surf
+//!
+//! ## Working with runtimes
+//!
+//! Opentelemetry API & SDK supports different runtimes. Usually, when working with async runtime. To
+//! improve the performance, we recommend to use batch span processors where the spans will be sent in
+//! batch, reducing the number of requests and resource needed.
```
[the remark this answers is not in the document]
TommyCpp: Good catch. I guess I was interpreted by something when I wrote this :man_facepalming:

scripts/test.sh
```diff
@@ -6,6 +6,8 @@ cargo test --all "$@"
 
 # See https://github.com/rust-lang/cargo/issues/5364
 cargo test --manifest-path=opentelemetry/Cargo.toml --no-default-features
+# Run global tracer provider test in single thread
+cargo test --manifest-path=opentelemetry/Cargo.toml --features=tokio-support -- --ignored --test-threads=1
```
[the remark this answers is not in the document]
TommyCpp: Yeah we do need that. Good point!

jtescher (approved):
Thanks for all the work on this @TommyCpp, looks good to me. Users may forget to call shutdown and miss a few last unexported spans, but that is preferable to the current situation where shutdown can happen silently and no spans are exported at all.

jtescher:
@djc / @frigus02 any other thoughts on this one? global function name and behavior / new tokio feature flag all look good?

opentelemetry/src/global/trace.rs
```diff
@@ -285,9 +268,275 @@ where
...
+
+        fn get_tracer(&self, _name: &'static str, _version: Option<&'static str>) -> Self::Tracer {
+            NoopTracer::default()
+        }
+    }
+
+    #[test]
+    #[ignore]
+    fn test_set_tracer_provider() {
+        let _guard = set_tracer_provider(TestTracerProvider::new("global one"));
```
frigus02, line 384: There are a few more instances of the `_guard` variable here in in the following lines/tests. Not sure if you want to replace them with `_previous` or just `_`.

frigus02:
The global function name seems fine to me.

I think I would love to unify the naming on the different runtime features. Maybe something alone the lines of `rt-tokio`, `rt-tokio-current-thread` and `rt-async-std`. That would more clearly indicate to me that they are mutually exclusive.

We could possibly even throw compile errors if more than one runtime feature is activated. I saw this pattern in another project. Not sure if this requires any special features (translated to our use case but untested):

```rust
#[cfg(not(any(feature = "rt-tokio", feature = "rt-tokio-current-thread", feature = "rt-tokio-async-std")))]
#[cfg(all(feature = rt-tokio", feature = "rt-tokio-current-thread"))]
#[cfg(all(feature = rt-tokio", feature = "rt-async-std"))]
#[cfg(all(feature = rt-tokio-current-thread", feature = "rt-async-std"))]
compile_error!(
    "exactly one of the features ['rt-tokio', 'rt-tokio-current-thread', 'rt-async-std'] must be enabled"
);
```

But I'm happy to discuss both of those things in separate PRs.

frigus02 (approved):

djc:
> I think I would love to unify the naming on the different runtime features. Maybe something alone the lines of `rt-tokio`, `rt-tokio-current-thread` and `rt-async-std`. That would more clearly indicate to me that they are mutually exclusive.

Sounds good to me.

opentelemetry/src/global/trace.rs
TommyCpp: Oops, I think I forget to clean the remaining `_guard`. Should be fixed now.

TommyCpp:
> We could possibly even throw compile errors if more than one runtime feature is activated. I saw this pattern in another project. Not sure if this requires any special features (translated to our use case but untested):
> 
> ```rust
> #[cfg(not(any(feature = "rt-tokio", feature = "rt-tokio-current-thread", feature = "rt-tokio-async-std")))]
> #[cfg(all(feature = rt-tokio", feature = "rt-tokio-current-thread"))]
> #[cfg(all(feature = rt-tokio", feature = "rt-async-std"))]
> #[cfg(all(feature = rt-tokio-current-thread", feature = "rt-async-std"))]
> compile_error!(
>     "exactly one of the features ['rt-tokio', 'rt-tokio-current-thread', 'rt-async-std'] must be enabled"
> );
> ```

I think the current approach when multiple conflicting features are enabled is to choose one of them. For example, enable `tokio-support` and `async-std` at the same time is equal to enable the `tokio-support`. 

I'd vote to change this behavior but it seems we should address this in a different PR.

frigus02:
> I'd vote to change this behavior but it seems we should address this in a different PR.

Definitely agree, this should be a different PR.

frigus02 (approved):

TommyCpp:
This is good to merge I guess?

Issue #268, opened 2020-10-14
Update trace config max defaults


jtescher:
See the spec change here for details https://github.com/open-telemetry/opentelemetry-specification/pull/942

morigs:
Do I understand correctly that the task is reduced to replacing [these constants](https://github.com/open-telemetry/opentelemetry-rust/blob/b47794e5818d7ba39115933ea7f752b8784cf025/src/sdk/trace/config.rs#L80-L82) with the values of environment variables (with a default value of 1000 of course)?

jtescher:
@morigs the conventions so far in this workspace have been to have constants (that need to be updated here) and the struct or builder to have a `from_env` method that would use the env vars

morigs:
Ok. So `default()` should always use values of 1000.
But where `from_env` should be located? Is it ok to be right inside [`impl Config`](https://github.com/open-telemetry/opentelemetry-rust/blob/b47794e5818d7ba39115933ea7f752b8784cf025/src/sdk/trace/config.rs#L30)?
So usage would be `Config::default().from_env()`?
This will require adding `.from_env ()` to all places where config was created, because environment variables must always be respected (imho).

jtescher:
@morigs I personally prefer that env vars should always be used when available as well. This project went with the explicit version both because of conventions set by the golang impl (e.g. [`FromEnv`](https://github.com/open-telemetry/opentelemetry-go/blob/1f7c2206522accd0a176ed8d93183f1168d70939/sdk/resource/env.go#L36), [`WithProcessFromEnv`](https://github.com/open-telemetry/opentelemetry-go/blob/d6dd84f6fa740ddfcf028dab0d79c07513b072cb/exporters/trace/jaeger/env.go#L94), etc) as that is the most widely used distribution, as well as the ambiguity when using struct constructor with defaults. 

E.g. `Config { max_events_per_span: 100, ..Default::default() }` wouldn't pick up the variables.

It does have the somewhat substantial downside that consumers of the API may be confused when they set env vars that do not have the effect they expect. I would be open to re-evaluating the decision to use explicit from env if you want to open an issue for that (e.g. we could make all fields private across the projects to force going through constructors that would pick up the environment)

morigs:
> consumers of the API may be confused when they set env vars that do not have the effect they expect
Totally agreed
> if you want to open an issue for that (e.g. we could make all fields private across the projects to force going through constructors that would pick up the environment)
[...]

PR #445, opened 2021-01-29
Create BatchSpanProcessor from env in TracerProvider

samschlegel:
This allows for configuring BSP when just `with_exporter` is used, instead of requiring users to create the BSP themselves when they want to pull from the environment.

It's possible that maybe this isn't the best route to go, and should instead just have users of TracerProvider expose with_batch_exporter themselves and have people pass `with_env` up, but figured I'd throw this together as it was the simplest thing to get what we needed and start the conversation 😄

### DIFF
--- a/opentelemetry/src/sdk/trace/provider.rs
+++ b/opentelemetry/src/sdk/trace/provider.rs
@@ -116,11 +116,11 @@ impl Builder {
         // drop. We cannot assume we are in a multi-threaded tokio runtime here, so use
         // `spawn_blocking` to avoid blocking the main thread.
         let spawn = |fut| tokio::task::spawn_blocking(|| futures::executor::block_on(fut));
-        let batch = sdk::trace::BatchSpanProcessor::builder(
+        let batch = sdk::trace::BatchSpanProcessor::from_env(
             exporter,
             spawn,
-            tokio::time::sleep,
             crate::util::tokio_interval_stream,
+            tokio::time::sleep,
         );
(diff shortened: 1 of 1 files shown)


opentelemetry/src/sdk/trace/provider.rs
```diff
@@ -116,11 +116,11 @@ impl Builder {
         // drop. We cannot assume we are in a multi-threaded tokio runtime here, so use
         // `spawn_blocking` to avoid blocking the main thread.
         let spawn = |fut| tokio::task::spawn_blocking(|| futures::executor::block_on(fut));
-        let batch = sdk::trace::BatchSpanProcessor::builder(
+        let batch = sdk::trace::BatchSpanProcessor::from_env(
             exporter,
             spawn,
-            tokio::time::sleep,
             crate::util::tokio_interval_stream,
+            tokio::time::sleep,
```
samschlegel, line 123: Oddly, from_env switches these two last arguments :(

jtescher: yeah that looks like an oversight, want to create a PR to fix that?

jtescher (approved):
Looks good

jtescher:
We discussed the `from_env` situation a bit in https://github.com/open-telemetry/opentelemetry-rust/issues/268, ideally env vars are picked up by default but haven't gotten around to updating yet

TommyCpp (approved):

PR #446, opened 2021-01-31
Move datadog and aws out of contrib crate

TommyCpp:
Also added opentelemetry-jaeger, opentelemetry-zipkin, opentelemetry-datadog, opentelemetry-aws into code coverage check.


opentelemetry-aws/README.md
```diff
@@ -0,0 +1,32 @@
+![OpenTelemetry — An observability framework for cloud-native software.][splash]
+
+[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
+
+# OpenTelemetry AWS
+
+Community supported vendor integrations for applications instrumented with [`OpenTelemetry`].
+
+[![Crates.io: opentelemetry-aws](https://img.shields.io/crates/v/opentelemetry-aws.svg)](https://crates.io/crates/opentelemetry-contrib)
```
jtescher, line 9:
```suggestion
[![Crates.io: opentelemetry-aws](https://img.shields.io/crates/v/opentelemetry-aws.svg)](https://crates.io/crates/opentelemetry-aws)
```

opentelemetry-aws/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "opentelemetry-aws"
+version = "0.1.0"
+authors = ["OpenTelemetry Authors <cncf-opentelemetry-contributors@lists.cncf.io>"]
+description = "Rust contrib repo for OpenTelemetry"
```
jtescher, line 5:
```suggestion
description = "AWS exporters and propagators for OpenTelemetry"
```

opentelemetry-datadog/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "opentelemetry-datadog"
+version = "0.1.0"
+authors = ["OpenTelemetry Authors <cncf-opentelemetry-contributors@lists.cncf.io>"]
+description = "Rust contrib repo for OpenTelemetry"
```
jtescher, line 5:
```suggestion
description = "Datadog exporters and propagators for OpenTelemetry"
```

opentelemetry-datadog/README.md
```diff
@@ -0,0 +1,90 @@
+![OpenTelemetry — An observability framework for cloud-native software.][splash]
+
+[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
+
+# OpenTelemetry Datadog
+
+Community supported vendor integrations for applications instrumented with [`OpenTelemetry`].
+
+[![Crates.io: opentelemetry-datadog](https://img.shields.io/crates/v/opentelemetry-datadog.svg)](https://crates.io/crates/opentelemetry-contrib)
```
jtescher, line 9:
```suggestion
[![Crates.io: opentelemetry-datadog](https://img.shields.io/crates/v/opentelemetry-datadog.svg)](https://crates.io/crates/opentelemetry-datadog)
```

jtescher (changes requested):
Great! couple of places to update in the new crates

opentelemetry-aws/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! # Opentelemetry AWS Integration
+//!
+//! This crate provides integration with different telemetry platform.
+//!
+//!
+
+#[cfg(feature = "trace")]
+pub mod trace;
```
djc, line 8: I'm not sure if it really makes sense to us this nested API layout `crate::trace::xray::propagator::XrayPropagator` over just `crate::XrayPropagator` or maybe `crate::trace::XrayPropagator`. Looks like overkill to me. Also, probably makes sense to use inline modules for this rather than files.

opentelemetry-jaeger/src/exporter/mod.rs
```diff
@@ -642,6 +642,7 @@ const OTEL_STATUS_CODE: &str = "otel.status_code";
 const OTEL_STATUS_DESCRIPTION: &str = "otel.status_description";
 
 #[derive(Default)]
+#[allow(unused)]
```
djc, line 645: Why is this code still necessary? Should we just remove it?

opentelemetry-datadog/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+//!
+//!     Ok(())
+//! }
+//! ```
+
+mod exporter;
+mod propagator;
+
+pub use exporter::{new_pipeline, ApiVersion, DatadogExporter, DatadogPipelineBuilder, Error};
+pub use propagator::DatadogPropagator;
```
djc, line 123: Same here, I'd remove some layers and/or use inline modules unless there's actually substantial code in a module.

djc (commented):
This looks much better to me than the previous situation, thanks for working on this!

opentelemetry-aws/src/lib.rs
TommyCpp: I think the public API is already the `opentelemetry_aws::trace::XrayPropagator`. The inline module sounds good. Fixed that

opentelemetry-jaeger/src/exporter/mod.rs
```diff
@@ -605,8 +605,8 @@ fn build_span_tags(
     }
 
     if status_code != StatusCode::Unset {
-        // Ensure error status is set
-        if status_code == StatusCode::Error {
+        // Ensure error status is set or user manually set it to error
+        if status_code == StatusCode::Error || user_overrides.error {
```
jtescher, line 609: I think these should be the opposite: if user_overrides.error is set, then we don't need to set it here as they have already done that. 

This is for the case where the user previously sets one of the fields manually that we are setting for all spans.

TommyCpp: Spec is a little vogue here(On the one hand the error tag MUST be set to true, on the other hand, it doesn't have to override the previous value). But I agree users should be able to override it
> When Span Status is set to ERROR, an error span tag MUST be added with the Boolean value of true. The added error tag MAY override any previous value.

jtescher (approved):
LGTM. @djc any other comments on your end?

opentelemetry-aws/src/lib.rs
djc: What do you mean? We don't really need to keep the public API the same, right?

TommyCpp: @djc No we don't. But I think it make sense to keep the `propagator` mod because we might add `XrayExporter` in the future. So the current API layout is `crate::trace::propagator::XrayPropagator`

djc: I don't see why we couldn't just have `XrayExporter` and `XrayPropagator` in `crate::trace` (or even just in the root). Is your goal here to maintain some kind of symmetry to how the opentelemetry crate (and spec) are laid out? For things that aren't really covered by the spec, I would value the simplicity of just having the simple types in the crate root over mirroring the core crate (which clearly has to somehow organize since it contains much more stuff). For imports, all users will have to type out (and remember) the path to the actual type. Seems like a bad cost/benefit ratio to me.

TommyCpp: So I think there are two thing here. One is how we organize the modules, the other is how our public API looks like. Those two can be different because we can re-export types in modules. And as long as the public API is the same, we can change how we organize the modules without breaking changes.

For public API, I think `opentelemetry_aws::trace::XrayPropagator` makes the most sense as I hope we will add cloudwatch(metrics) support some point in the future. And separate the tracing and metrics API is consistent with the main create. 

For module structure. I tend to encapsulate the related stuff into modules as it will help developer to understand those structures are belong to the same component. For example, by putting `XrayTraceId ` and `XrayPropagator` into the same module. It gives the clarity  that `XrayTraceId ` is something that related to `XrayPropagator`.

In current structure, we have the following file/module structure
```
- root
--- trace
----- propagator
------- XrayPropagator, XrayTraceId
```
But because I re-export the `XrayPropagator` in `trace`
https://github.com/open-telemetry/opentelemetry-rust/blob/ea53b6dc648971077f15457fd30855bd420d8ef8/opentelemetry-aws/src/trace/mod.rs#L323

Users can just import it as `opentelemetry_aws::trace::XrayPropagator`

djc: @jtescher any thoughts? Curious about your opinion on these things.

jtescher: Hm good question. I think we should go with whichever is most intuitive here but it would certainly help if all the crates in this repo had a consistent set of guidelines to drive that intuition. Seems like the usage so far is fairly clear for cases that only impl a single signal (i.e. trace or metrics only) like otel jaeger, otel prometheus, etc where we don't namespace, and it is also clearer for very complex cases like the core otel crate that have many types in those namespaces. The question seems to be around crates with multiple signals and few types like otlp (not currently namespaced) and aws.

I don't think the current otlp experience is confusing if you are trying to understand the crate from reading https://docs.rs/opentelemetry-otlp/ so I think I'd lean toward aws following the same non-namespaced approach for now. The general recommendation in this case could be to namespace only once a certain level of complexity has been introduced where the mixed types become confusing.

TommyCpp: :+1: looks like there is a consensus, will update accordingly.

djc (approved):

PR #448, opened 2021-02-02
opentelemetry-jaeger: fixed setting agent endpoint to a host via env.

### LINKED ISSUE #447: Setting agent endpoint via env does not support hostnames
According to the [environment variables specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md), the `OTEL_EXPORTER_JAEGER_AGENT_HOST` environment variable should support a hostname like 'localhost', however opentelemetry-jaeger parses the host/port using `std::net::SocketAddr` which only allows an IP address.

This one had me confused for a bit because it fails silently and acts as if those variables are not set. Using `.with_agent_endpoint("localhost:123")` works as expected.

fourbytes:
Closes #447

### DIFF
--- a/opentelemetry-jaeger/src/exporter/env.rs
+++ b/opentelemetry-jaeger/src/exporter/env.rs
@@ -1,6 +1,5 @@
 use crate::PipelineBuilder;
 use std::env;
-use std::net;
 
 /// The name under which Jaeger will group reported spans.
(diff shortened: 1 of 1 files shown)


djc:
Looks good to me!

jtescher (approved):

PR #450, opened 2021-02-06
Configure tls and add example for grpcio

jaysonsantos:
This PR adds a new function to otlp's builder called `with_tls` in order to configure empty credentials on grpcio so it talks with https and not in a decrypted manner.
I think that for the future this with_tls could be used on tonic as well so it can configure the host for TLS in case the server you are talking to uses SNI.
Let me know if you have questions or something has o change there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,25 @@ jobs:
         args: --all -- --check
     - name: Lint
       run: ./scripts/lint.sh
+  non-default-examples:
+    strategy:
+      matrix:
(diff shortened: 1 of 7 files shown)


TommyCpp:
Might need to remove the example from the workspace and ask CI to check it separately.

jaysonsantos:
@TommyCpp 
> Might need to remove the example from the workspace and ask CI to check it separately.

should I do it only on this example or for all of them?

jtescher:
Looks reasonable, thanks

jtescher (approved):

PR #452, opened 2021-02-08
Remove non-existent feature flags from docs.

boxdot:


djc:
Oops, must have missed those in #415. Thanks!

djc (approved):

PR #453, opened 2021-02-08
Fix clippy warnings and deny them in CI

djc:


jtescher:
Looks like some of the tests were failing here. Might be fixed after rebase?

TommyCpp:
I think it's because `clippy::from_over_into` is still in nightly. So when running in stale/beta channel it will fail with not found error.

.github/workflows/ci.yml
```diff
@@ -78,8 +77,3 @@ jobs:
         RUSTDOCFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Clink-dead-code -Coverflow-checks=off -Cpanic=abort -Zpanic_abort_tests'
     - uses: actions-rs/grcov@v0.1
     - uses: codecov/codecov-action@v1
-    - run: cargo fmt -- --check
```
TommyCpp, line 81: I think we should check the format in a different workflow instead of ignoring it. It helps since sometimes ppl forget to format the code before push(I did that several times :grimacing:)

djc: You're right, of course. I usually keep it in the lint job, so I somehow expected it to be there. Added it there now.

jtescher (approved):
Looks good

PR #455, opened 2021-02-12
fix: outdated examples in opentelemetry-jaeger.

TommyCpp:
For some reasons we didn't update those examples in README last time.


jtescher (approved):
Good call @TommyCpp

PR #457, opened 2021-02-15
Add jaeger max packet size constraint

### LINKED ISSUE #441: Jaeger exporter max UDP packet size config
The jaeger agent exporter needs to cap the UDP packet max length to `65000` to be in sync with the jaeger server impl and avoid span data loss. Needs to send and start a new batch with remaining spans if over that limit.

jtescher:
Currently jaeger will silently ignore batches of spans where the resulting thrift payload packet sizes is over the max 65000
bytes. This patch brings the exporter in line with similar otel exporters by ensuring that the payload is less than the specified limit.

Resolves #441

### DIFF
--- a/opentelemetry-jaeger/README.md
+++ b/opentelemetry-jaeger/README.md
@@ -167,7 +167,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         .with_agent_endpoint("localhost:6831")
         .with_service_name("my_app")
         .with_tags(vec![KeyValue::new("process_key", "process_value")])
-	.with_max_packet_size(65_000)
+        .with_max_packet_size(9_216)
         .with_trace_config(
             trace::config()
                 .with_default_sampler(Sampler::AlwaysOn)
--- a/opentelemetry-jaeger/README.md
+++ b/opentelemetry-jaeger/README.md
@@ -167,6 +167,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         .with_agent_endpoint("localhost:6831")
         .with_service_name("my_app")
         .with_tags(vec![KeyValue::new("process_key", "process_value")])
+	.with_max_packet_size(65_000)
         .with_trace_config(
             trace::config()
                 .with_default_sampler(Sampler::AlwaysOn)
--- a/opentelemetry-jaeger/src/exporter/agent.rs
+++ b/opentelemetry-jaeger/src/exporter/agent.rs
@@ -11,6 +11,9 @@ use thrift::{
     transport::{ReadHalf, TIoChannel, WriteHalf},
 };
 
+/// The max size of UDP packet we want to send, synced with jaeger-agent
+const UDP_PACKET_MAX_LENGTH: usize = 65_000;
+
 struct BufferClient {
     buffer: ReadHalf<TBufferChannel>,
     client: agent::AgentSyncClient<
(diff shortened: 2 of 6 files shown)


opentelemetry-jaeger/README.md
```diff
@@ -167,6 +167,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         .with_agent_endpoint("localhost:6831")
         .with_service_name("my_app")
         .with_tags(vec![KeyValue::new("process_key", "process_value")])
+	.with_max_packet_size(65_000)
```
djc, line 170: The indentation seems off here.

opentelemetry/src/sdk/trace/provider.rs
```diff
@@ -25,7 +27,11 @@ pub(crate) struct TracerProviderInner {
 impl Drop for TracerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
-            let _result = processor.shutdown();
+            let result = processor.shutdown();
+
+            if let Err(err) = result {
```
djc, line 32: Maybe just write `if let Err(err) = processor.shutdown() {`?

opentelemetry-jaeger/README.md
djc, line 170: This default maximum only supports IPv6 and probably only on LANs. Can we assume the Jaeger agent is (almost) always addressed via IPv6, over a LAN?

opentelemetry/src/sdk/trace/provider.rs
jtescher: Yeah for single method calls this is probably simpler. For some of the chained calls it can get a little unreadable but for this one here the more concise variant is probably better 👍

opentelemetry-jaeger/README.md
jtescher: Good point, may be better to have a lower max in the examples 9216 byes is also common

opentelemetry-jaeger/src/exporter/agent.rs
```diff
@@ -40,12 +43,17 @@ pub(crate) struct AgentAsyncClientUDP {
...
 
 impl AgentAsyncClientUDP {
     /// Create a new UDP agent client
-    pub(crate) fn new<T: ToSocketAddrs>(host_port: T) -> thrift::Result<Self> {
-        let (buffer, write) = TBufferChannel::with_capacity(512).split()?;
+    pub(crate) fn new<T: ToSocketAddrs>(
+        host_port: T,
+        max_packet_size: Option<usize>,
+    ) -> thrift::Result<Self> {
+        let max_packet_size = max_packet_size.unwrap_or(UDP_PACKET_MAX_LENGTH);
```
djc, line 55: Sorry, but I still think this isn't quite right. Here, you default `max_packet_size` to `UDP_PACKET_MAX_LENGTH` which is still 65000, but the `UdpSocket::bind()` below binds on `0.0.0.0`, which is an IPv4 address. I think we can do better on this?

jtescher: What did you have in mind?

TommyCpp: Maybe try to connect to localhost via IPv6 and if that's doesn't work, then try to connect to localhost via IPv4? We could set the default max packet size based on the protocol

djc: Or just be conservative and restrict `max_packet_size` to 9216 by default since it can be configured anyway? (Maybe also rename `UDP_PACKET_MAX_LENGTH` to `DEFAULT_MAX_PACKET_SIZE`?)

jtescher: Hm looking around seems the other languages do 65,000 as well. e.g. [go](https://github.com/open-telemetry/opentelemetry-go/blob/9524ac095466565e01746600da3d83f27220a5da/exporters/trace/jaeger/agent.go#L30), [ruby](https://github.com/open-telemetry/opentelemetry-ruby/blob/dc109c7d851539864961a94d1510d67a10bbbac9/exporter/jaeger/README.md#how-can-i-configure-the-jaeger-exporter), [python](https://github.com/open-telemetry/opentelemetry-python/blob/18761d78e30a4293abd65a897129b57fbf0c0a15/exporter/opentelemetry-exporter-jaeger/src/opentelemetry/exporter/jaeger/send/thrift.py#L25), etc. Probably to be in sync with the [jaeger default](https://github.com/jaegertracing/jaeger/blob/5062366157fff65a052dc6ac6779e7621054248d/cmd/agent/app/servers/thriftudp/transport.go#L28-L31).

djc: To me that just looks like everybody is getting it wrong -- but I'm fine if you want to keep it to stay aligned.

jtescher: I'm ok with reducing if that is a common case, I know that macos will have the lower limit by default. Anyone have a good resource on different os defaults? If non-mac systems default to 65k we could just have a cfg on the const specifically for that.

TommyCpp: I think it will confuse ppl if the Rust implementation behaviors different from other implementations. So I incline to keep our limit here the same as others.

jtescher: Works for me if we want to merge this as is. Happy to update in a subsequent PR if we find a better heuristic.

djc (approved):

PR #462, opened 2021-02-26
Sanitize Prometheus labels names

frigus02:
OpenTelemetry recommends to use metric labels with dots (e.g. "http.server.duration"). Prometheus doesn't allow dots in labels, though. We already sanitise metric names, but don't do so for label names.

This adds the same sanitization for label names and adds an integration test.

---

In a separate commit I also refactored the metric name and label generation code a bit, so that:

- Sanitization happens when we convert from an OpenTelemetry type to a Prometheus type
- We don't store label keys in an additional Vec, because we didn't use them.

This seemed to make sense to me. But I'm happy to remove the last commit if you don't agree 🙂


jtescher (approved):
LGTM thanks

PR #464, opened 2021-02-28
Update GRPC examples

markdingram:
Copied in the Injector / Extractor impls from the grpc example to tracing grpc so there's a bit of code duplication.. not sure it warrants a shared crate though.


djc:
LGTM.

djc (approved):

PR #465, opened 2021-03-02
chore: fix documentation

TommyCpp:
Also notice that we never publish `opentelemetry-datadog` and `opentelemetry-aws` crates.

### DIFF
--- a/opentelemetry-prometheus/examples/hyper.rs
+++ b/opentelemetry-prometheus/examples/hyper.rs
@@ -27,23 +27,23 @@ async fn serve_req(
 ) -> Result<Response<Body>, hyper::Error> {
     let request_start = SystemTime::now();
 
-    state.http_counter.add(1);
-    state
-        .http_req_histogram
-        .record(request_start.elapsed().map_or(0.0, |d| d.as_secs_f64()));
-
     let mut buffer = vec![];
     let encoder = TextEncoder::new();
     let metric_families = state.exporter.registry().gather();
-    let _test = metric_families.len();
     encoder.encode(&metric_families, &mut buffer).unwrap();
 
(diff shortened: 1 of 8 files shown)


opentelemetry-prometheus/examples/hyper.rs
```diff
@@ -27,24 +27,24 @@ async fn serve_req(
...
 
+    state.http_counter.add(1);
+    state
+        .http_req_histogram
+        .record(request_start.elapsed().map_or(0.0, |d| d.as_secs_f64()));
+
     let mut buffer = vec![];
     let encoder = TextEncoder::new();
     let metric_families = state.exporter.registry().gather();
+    let _test = metric_families.len();
```
jtescher, line 38: I see you're looking at the first render metrics, I think were this way to mirror https://github.com/tikv/rust-prometheus/blob/master/examples/example_hyper.rs#L42-L55 for folks coming from experience with that crate. I don't feel super strongly about sticking to what they had, just want to be explicit about the change.

jtescher: Probably fine if we go with the re-ordered direction, but this test is probably not needed?

TommyCpp: Sorry this shouldn't be part of this PR. I was investigating an interesting finding on this example and was tweaking around to see what could be the problem. 

The finding is the `serve_req` function is called twice for each incoming request. As a result, the `request_total` metric increases by 2 every time. 

But I think it's on the `hyper` part, as I see [their example](https://github.com/hyperium/hyper/blob/master/examples/state.rs) has the same problem

jtescher (approved):

PR #466, opened 2021-03-03
chore: convert to intra-doc links

Folyd:
This PR converts all legacy-style doc links to intra-doc links.

### DIFF
--- a/examples/grpc/Cargo.toml
+++ b/examples/grpc/Cargo.toml
@@ -6,10 +6,12 @@ edition = "2018"
 [[bin]] # Bin to run the gRPC server
 name = "grpc-server"
 path = "src/server.rs"
+doc = false
 
 [[bin]] # Bin to run the gRPC client
 name = "grpc-client"
 path = "src/client.rs"
+doc = false
 
 [dependencies]
 http = "0.2"
--- a/examples/http/Cargo.toml
+++ b/examples/http/Cargo.toml
@@ -6,10 +6,12 @@ edition = "2018"
 [[bin]] # Bin to run the http server
 name = "http-server"
 path = "src/server.rs"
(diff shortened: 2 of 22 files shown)


djc (approved):
This looks great, thanks!

The only thing I was thinking about is that I don't love the `super::super` (although clearly that's kind of how it was before). How do you feel about replacing those with `crate::`-rooted paths?

Folyd:
> This looks great, thanks!
> 
> The only thing I was thinking about is that I don't love the `super::super` (although clearly that's kind of how it was before). How do you feel about replacing those with `crate::`-rooted paths?

Thanks for reviewing. This has been fixed with the new patch. 🍺

djc:
Nice, thanks!

TommyCpp (commented):
> ```
> $ cargo doc --no-deps     
> warning: unresolved link to `GlobalProvider`
>   --> opentelemetry/src/global/trace.rs:68:16
>    |
> 68 | /// Wraps the [`GlobalProvider`]'s [`Tracer`] so it can be used generically by
>    |                ^^^^^^^^^^^^^^^^ no item named `GlobalProvider` in scope
>    |
>    = note: `#[warn(broken_intra_doc_links)]` on by default
>    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
> ```

Thanks for the great work! I think this should be `crate::global::GlobalTracerProvider`. We rename it from `GlobalProvider` a while ago but didn't seem to update the doc.

Folyd:
> I think this should be crate::global::GlobalTracerProvider. We rename it from GlobalProvider a while ago but didn't seem to update the doc.

@TommyCpp This has been resolved. Thanks a lot.

PR #467, opened 2021-03-04
feat: clean up lint.sh, use API to set grpc layer

TommyCpp:
- As per https://github.com/open-telemetry/opentelemetry-rust/pull/444#issuecomment-785684996.
We should try to throw compile errors when users set two exclusive features instead of default to use one of them.
- Added a `precommit.sh` to run `fmt`, `clippy` and `test`
- Clean up the `lint.sh` a little bit. Haven't worked with Bash for a while so please let me know if something doesn't look right.


djc:
IMO making features non-additive is frowned upon in the ecosystem. Perhaps we could do it as part of the tests to flush out issues in our test suite, but I don't think we should intentionally start doing this for our published libraries.

Cargo.toml
```diff
@@ -25,3 +25,28 @@ members = [
     "examples/zipkin",
 ]
 exclude = ["examples/external-otlp-grpcio-async-std"]
+default-members = [
+    "opentelemetry",
+    "opentelemetry-contrib",
+    "opentelemetry-datadog",
+    "opentelemetry-http",
+    "opentelemetry-jaeger",
+#    "opentelemetry-otlp",
```
djc, line 34: Maybe add a comment on why you commented this out?

frigus02:
> IMO making features non-additive is frowned upon in the ecosystem. Perhaps we could do it as part of the tests to flush out issues in our test suite, but I don't think we should intentionally start doing this for our published libraries.

I'm not super happy with this, either. If I understand correctly, the features are already mutually exclusive, though. At the moment we just silently ignore one of the features if the other one is enabled. I would prefer a helpful error message instead.

That said, I'm still undecided if it would be better to configure these things at runtime.

---

The scripts look fine to me. I think a trailing newline might be nice. I think that's common for shell scripts, too.

TommyCpp:
I think the real problems I would like to solve is how do we tell user that featureA and featureB is exclusive and should only enable one of them. Some of our crates have complex features. For example, in `opentelemetry-otlp` we have following features. 

```
async = ["default"]
trace = ["opentelemetry/trace"]
metrics = ["opentelemetry/metrics"]
default = ["tonic", "tonic-build", "prost", "tokio"]
grpc-sys = ["grpcio", "protobuf", "protobuf-codegen", "protoc-grpcio"]
tls = ["tonic/tls"]
tls-roots = ["tls", "tonic/tls-roots"]
openssl = ["grpcio/openssl"]
openssl-vendored = ["grpcio/openssl-vendored"]
integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "opentelemetry/trace"]
```

This basically comes down to four choices. 
- Is it `trace` or `metrics`. 
- If it's `trace`, for grpc layer, is it `tonic` or `grpcio`. 
- Is it tls/ssl or not encrypted 
- Different choice on how to encrypted. 

But I don't think it's trivial to understand how to choose between those features without some knowledge of the implementation.

djc:
`trace` or `metrics` doesn't have to be an either/or choice, or does it?

I made a change in hyper-rustls which removed the compiler error on feature conflict in favor of having the API be explicit about which feature to prefer. https://github.com/ctz/hyper-rustls/commit/fc60b8c0c2ab15638c5b3bcc731f78be73ddaee3 I think this is probably the best way to do it if possible, though I don't quite understand why the grpc vs tonic features (for example) necessarily/naturally conflict.

TommyCpp:
> `trace` or `metrics` doesn't have to be an either/or choice, or does it?

Sorry those two are not mutually exclusive. 🤦‍♂️ 

But thanks for the example! I think it could be a good approach. 

> though I don't quite understand why the grpc vs tonic features (for example) necessarily/naturally conflict.

I think it's mostly because we have different to build the client based on different grpc layer. But I will try to poke around to see if we can use the API to separate those two features.

TommyCpp:
Still need to add docs, change tests and fix small nits. But my general idea for this crate is to use an enum to represent different grpc layers. 

The reason for not using trait is that the trace exporter here has two tasks, 
- convert the otel spans into otlp format
-  send it via grpc. 

In our current implementation, `tonic` and `grpcio` take different approaches in both of those tasks. It's hard to abstract the behavior as it involves a lot of sturcts like `Span`s, `Link`s, etc. 

If users want to bring their own grpc layer, they can just implement a new tracer exporter with SpanExporter trait and install it.

frigus02:
I'm honestly still on the fence between cargo features and setting things up at runtime. Since cargo doesn't have a way of describing that features are mutually exclusive, I'm leaning slightly towards the setup in Rust code at runtime.

This looks quite nice to me. Didn't see anything worth commenting on. 👍 

Do you want to update docs/tests/etc in this PR?

opentelemetry-otlp/src/lib.rs
```diff
@@ -119,27 +119,22 @@
...
+#[cfg(feature = "grpc-sys")]
 use std::collections::HashMap;
 
 use std::str::FromStr;
 use std::time::Duration;
 
-#[cfg(all(feature = "tonic", not(feature = "integration-testing")))]
+#[cfg(
+not(feature = "integration-testing")
+)]
```
djc, line 130: This looks a little odd?

opentelemetry-otlp/src/lib.rs
```diff
@@ -119,27 +119,22 @@
...
-#[allow(clippy::all, unreachable_pub)]
-mod proto;
-
-#[cfg(all(
-    feature = "grpc-sys",
-    not(feature = "tonic"),
-    not(feature = "integration-testing")
-))]
-#[allow(clippy::all, unreachable_pub, dead_code)]
+#[allow(warnings)]
```
djc, line 132: Why are we allowing all the warnings here? Maybe add a comment about this?

scripts/lint.sh
```diff
@@ -2,45 +2,33 @@
...
+  cargo_feature opentelemetry-otlp "trace,grpc-sys,openssl"
+  cargo_feature opentelemetry-otlp "trace,grpc-sys,openssl-vendored"
+
+  cargo_feature opentelemetry-jaeger "surf_collector_client"
+  cargo_feature opentelemetry-jaeger "isahc_collector_client"
+  cargo_feature opentelemetry-jaeger "reqwest_blocking_collector_client"
+  cargo_feature opentelemetry-jaeger "reqwest_collector_client"
+  cargo_feature opentelemetry-jaeger "collector_client"
+
+fi
```
djc, line 34: Nit: add a newline?

djc (commented):
I think the API-based approach works out pretty well, I especially like the builder setup.

opentelemetry-otlp/src/lib.rs
TommyCpp: I think the idea is to hide the proto mod unless we are running integrationt testing.

opentelemetry-otlp/src/lib.rs
TommyCpp: Mostly because the proto mod only contains generated code. We should change it.

jtescher (approved):
LGTM @djc is this good to merge or do you have additional feedback?

djc:
LGTM. @TommyCpp, good work!

djc (approved):

PR #469, opened 2021-03-05
For discussion: set async runtime in code instead of cargo feature

frigus02:
Hi all. I played around last week with setting the async runtime in code instead of using a cargo feature. I thought this make things easier to understand and work with. Did a bit of tidying up today and thought I would present this for discussion.

With this:

- The `rt-*` cargo features can be enabled at the same time. The only downside of enabling more features is a longer build time.
- Users have to pick a runtime by providing an object implemeting a `Runtime` trait. Without that, we fall back to the simple span processor.
- We include implementations for Tokio, Tokio with the current thread flavor and async-std.
- Users can bring their own implementation if needed.

Example:

```rust
let tracer = opentelemetry_jaeger::new_pipeline()
    .with_runtime(opentelemetry::runtime::Tokio)
    .install()?;
```

Runtime docs are able to show all features:

grafik

Some notes:

- The runtime trait adds generics to all pipeline builders, which is a bit annoying.
- We could instead use a trait object, but that would add dynamic dispatch.
- We could also use a runtime enum (`enum Runtime { Tokio, AsyncStd }`), but that would mean users could not bring their own runtimes.

What do you think? Is that worth exploring further?


opentelemetry/src/runtime.rs
```diff
@@ -0,0 +1,87 @@
+//! Provides an abstraction of several runtimes, which allows OpenTelemetry to work with any
+//! current or future runtime. There are currently builtin implementations for Tokio and async-std.
+
+use futures::{future::BoxFuture, stream::BoxStream};
+use std::time::Duration;
+
+/// A runtime in an abstraction of a (possibly async runtime) like tokio or async-std. It allows
+/// OpenTelemetry to work with any current and future runtime implementation.
+pub trait Runtime: Clone {
+    /// Create a [Stream][futures::Stream], which returns a new item every
```
djc, line 14: Did you consider using associated types for the return types here, or are you aiming to keep this object safe? If the latter, might be useful to make that explicit in a comment.

frigus02: Yeah, I was undecided about object safety. If that's not needed we can use associated types instead, which would remove some mapping in the functions.

djc: I'd maybe start with a smaller take that just adds the `Runtime` trait, its impls, and replaces the outer layer of methods that take the interminable number of parameters with it, and use associated types (equivalent to what's there now, just bundled up in a trait).

djc: (But on the whole, I really like this approach and think it's definitely better than what we have now -- even though I haven't reviewed it in full detail just yet.)

TommyCpp:
Nice work! Personally, I don't think there is much need to bring users' own runtime. Also, given the time we spent on fixing the tokio-current-thread runtime, it may not be a trivial job to integrate a runtime with the batch span processor.

djc:
There are other runtimes outside of async-std, smol and Tokio. For example, it doesn't seem unreasonable that Fuchsia might want to use opentelemetry for something.

opentelemetry/src/runtime.rs
jtescher: Agree, this is a nice improvement! @frigus02 any issues you're still working through? or close to having this ready to review?

frigus02: Sorry the slow progress. The last week was quite busy for me. I plan on on working on this tomorrow and get it ready for review.

I'll try to split this into 2 things:

1. The runtime trait but without public API changes.
2. The changes from cargo features to Rust code.

This way things should be easier to review and we can also talk about the changes separately.

frigus02: Alright. I split this PR into two: #480 and #481

PR #471, opened 2021-03-08
Switch to Cow for event names

jtescher:
Avoid additional unnecessary allocations when working with static event names.


TommyCpp (approved):
Seems like a good idea to me. May want to check and see how much performance improvement we can get from it?

jtescher:
@TommyCpp more important for many events, but no real downsides aside from a small amount of complexity as far as I can see

djc (approved):

PR #474, opened 2021-03-08
Allow opentelemetry-zipkin to be configured with rustls

### LINKED ISSUE #472: Rustls Support for Reqwest client
Is there any desire to support Rustls either by default or with a feature flag for the reqwest client used in these crates? It would be nice to not have to link OpenSSL.

djc:
Fixes #472. @alechenthorne, does this work for you?


jtescher (approved):
Looks good to me if this is the preferred config

alechenthorne:
This solves my issue! Thanks!

PR #476, opened 2021-03-10
otlp: mention service.name in kitchen sink cfg doc - closes #475

### LINKED ISSUE #475: opentelemetry-otlp how to name service: OTLPResourceNoAttributes
Currently, all my traces show up as top level `OTLPResourceNoAttributes`:

trace viewer screenshot

Is there a way to inject the service name with `opentelemetry-otlp` like in [opentelemetry-jaeger](https://docs.rs/opentelemetry-jaeger/0.11.0/opentelemetry_jaeger/struct.PipelineBuilder.html#method.with_service_name)?

clux:
adds a possibly slightly more useful key/value combination for docs.

it's probably a common thing to want to set, even if it's not a thing that otel cares about.

### DIFF
--- a/opentelemetry-otlp/README.md
+++ b/opentelemetry-otlp/README.md
@@ -131,7 +131,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
                 .with_max_events_per_span(64)
                 .with_max_attributes_per_span(16)
                 .with_max_events_per_span(16)
-                 with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
+                 .with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
         )
         .install()?;
 
--- a/opentelemetry-otlp/README.md
+++ b/opentelemetry-otlp/README.md
@@ -131,7 +131,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
                 .with_max_events_per_span(64)
                 .with_max_attributes_per_span(16)
                 .with_max_events_per_span(16)
-                .with_resource(Resource::new(vec![KeyValue::new("key", "value")])),
+                 with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
         )
         .install()?;
 
--- a/opentelemetry-otlp/README.md
+++ b/opentelemetry-otlp/README.md
@@ -131,7 +131,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
                 .with_max_events_per_span(64)
                 .with_max_attributes_per_span(16)
                 .with_max_events_per_span(16)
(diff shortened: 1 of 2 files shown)


TommyCpp (approved):

TommyCpp (changes requested):
May need to do the same change in README?

clux:
Done.

Btw would you accept PR for adding `OtlpPipelineBuilder::with_service_name` like in [opentelemetry-jaeger](https://docs.rs/opentelemetry-jaeger/0.11.0/opentelemetry_jaeger/struct.PipelineBuilder.html#method.with_service_name)?

TommyCpp:
> Btw would you accept PR for adding `OtlpPipelineBuilder::with_service_name` like in [opentelemetry-jaeger](https://docs.rs/opentelemetry-jaeger/0.11.0/opentelemetry_jaeger/struct.PipelineBuilder.html#method.with_service_name)?
> Might cater to the common use case a bit more ergonomically.

Hmm, I guess the problem here is OTLP still doesn't have the `service.name` concept, `with_service_name` would be confusing if users are using a backend that doesn't recognize the `service.name` resource. But maybe we should make a note in the doc to explain if the backend recognizes a Jaeger-style service name, we can set it via resource.

TommyCpp (approved):

opentelemetry-otlp/README.md
```diff
@@ -131,7 +131,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
                 .with_max_events_per_span(64)
                 .with_max_attributes_per_span(16)
                 .with_max_events_per_span(16)
-                .with_resource(Resource::new(vec![KeyValue::new("key", "value")])),
+                 with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
```
jtescher, line 134:
```suggestion
                 .with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
```

jtescher (changes requested):

opentelemetry-otlp/README.md
clux: ah, whoops :|

opentelemetry-otlp/README.md
```diff
@@ -131,7 +131,7 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
                 .with_max_events_per_span(64)
                 .with_max_attributes_per_span(16)
                 .with_max_events_per_span(16)
-                .with_resource(Resource::new(vec![KeyValue::new("key", "value")])),
+                 .with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
```
clux, line 134: # #
```suggestion
                .with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
```

jtescher (approved):

PR #480, opened 2021-03-14
Abstraction trait for different runtimes

frigus02:
Extracted from #469 

This contains only the new `Runtime` trait, which replaces the individual function arguments to the batch span processor.

grafik

I realized that `BatchSpanProcessor::builder` is a public function, so the `Runtime` trait and implementations need to be public as well. This makes this a breaking change. I don't expect many projects to use `BatchSpanProcessor::builder` directly, though. Most will create a pipeline from one of the exporter crates, which call `TraceProvider::with_exporter` internally. And that stayed the same.

The `Runtime` trait now uses associated types. This means we don't require additional boxing into `BoxFuture` or `BoxStream`. But as far as I understand this means `Runtime` is no longer object-safe. That should be fine.

There is one exception. I didn't manage to remove `BoxFuture` from the async-std implementation of `delay`. It doesn't have an explicit return type that I could reference. Does anyone know if (and how) it's possible to work around this?


opentelemetry/src/runtime.rs
```diff
@@ -0,0 +1,122 @@
...
+        // BatchSpanProcessor to export spans in a background task, meanwhile we need to block the
+        // shutdown function so that the runtime will not finish the blocked task and kill any
+        // remaining tasks. But there is only one thread to run task, so it's a deadlock
+        //
+        // Thus, we spawn the background task in a separate thread.
+        std::thread::spawn(move || {
+            let rt = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .unwrap();
```
TommyCpp, line 89: I was reviewing this part the other day and thought I probably should have added a `expect` here just in case the build is failed. Could we add something like this?

frigus02: Yes, that sounds a good idea. 👍  I added this change in this PR.

TommyCpp (approved):
Nice! LGTM

PR #481, opened 2021-03-14
Configure async runtime in Rust code instead of cargo feature

frigus02:
Extracted from #469.

Configure the async runtime (Tokio, async-std, ...) in Rust using `install_batch(runtime)` on the trace pipeline builders instead of using cargo features.

The main advantages of using Rust code is that cargo doesn't have a way of describing mutually exclusive features at the moment. We currently fall back to Tokio if multiple runtime features are enabled. This can be confusing, especially if a runtime feature is enabled by another crate.

Choosing the runtime in Rust code means the entire trace pipeline configuration is in one place.

With this:

- The `rt-*` cargo features can be enabled at the same time. The only downside of enabling more features is a longer build time.
- Users have to pick a runtime by providing an object implemeting a `Runtime` trait or use `install_simple()` to use the simple span processor.

Examples:

```rust
let tracer = opentelemetry_jaeger::new_pipeline()
    .install_simple()?;
```

```rust
let tracer = opentelemetry_jaeger::new_pipeline()
    .install_batch(opentelemetry::runtime::Tokio)?;
```


opentelemetry-datadog/src/exporter/mod.rs
```diff
@@ -48,21 +49,22 @@ impl DatadogExporter {
 }
 
 /// Create a new Datadog exporter pipeline builder.
-pub fn new_pipeline() -> DatadogPipelineBuilder {
+pub fn new_pipeline() -> DatadogPipelineBuilder<()> {
```
djc, line 52: Hmm, the `impl Runtime for ()` seems like it makes it a run-time error to not configure a runtime, rather than a compile-time error. I feel like that could be done differently, maybe by using some type state here?

jtescher: Would be nice if we could default to the `futures` executor or otherwise have a default that "worked" even if it wasn't as performant. Unfortunately we'd need a solution for interval/delay.

frigus02: Yep, that's true. I made it fallback to the simple span processor when the pipeline is `()`, but you could still get a runtime error in case you manually create a batch span processor using the `()` runtime.

The other solution is let the user specifically choose simple vs batch processing. If been thinking of a nice API. This is the easiest I could come up with:

```rust
impl PipelineBuilder {
    fn install() {
        let mut builder = sdk::trace::TracerProvider::builder().with_simple_exporter(exporter);
        ...
    }
    fn install_with_batch_processing<R: Runtime>(runtime: R) {
        let mut builder = sdk::trace::TracerProvider::builder().with_exporter(exporter, runtime);
        ...
    }
}
```

---

But even the simple span processor needs async support. It uses the `futures` executor, which seems like a good default. It would be amazing if we could use that here, too. We could possibly build our own `interval` using `delay`. But we need a solution for `delay`. Is there some kind of minimal crate for that?

frigus02: Hm. I tried to read up a bit on the subject. The easiest way to implement `delay` is possibly by using `std::thread::spawn()` and `std::thread::sleep()`. But I suspect that this is not very efficient.

We could also use [`futures-timer`](https://crates.io/crates/futures-timer) for `delay` and implement `interval` ourselves. There is even [`futures-ticker`](https://crates.io/crates/futures-ticker), which seems to do exactly that.

TommyCpp: I won't worry much about performance much because users should not use batch processor without async runtime IMHO.

But I'd rather we don't introduce dependency because of this.

djc: I'm not sure, I'd rather have something that fails to compile by default unless I make an explicit choice than getting something by default that is inconspicuously slow.

frigus02: This argument makes sense to me. Lets stick with 3 runtimes then and remove `()`. Instead find another way to choose the simple span processor.

I played with `Option<R>` but that doesn't work with a generic runtime because even if the user choose `None` there has to be a type for the runtime. The alternative I can see is 2 different function names, e.g. `install_simple()` and `install_batch(runtime)`. Doesn't look too bad to me, I think. Might want to sleep over it, though. 🙂

djc: Yeah, that approach sounds good to me.

jtescher: Works for me. We can also work on docs/examples more clearly showing perf differences between simple and batch choices.

frigus02: I made the changes in a separate commit for now. It compiles but I should probably check if all the examples are actually still working.

Also feel free to bikeshed the function names `install_simple()` and `install_batch(runtime)`. 🙂

frigus02: That's true. I wonder what information we want to give here. I thought it would be nice to change the name away from just `install` to trigger a compile error for everyone upgrading. Because most people will have to change to `install_batch`.

Hm, or we rename `install_batch` to just `install` to indicate that it's the preferred way to install a pipeline.

We can still rename `install_simple`. I'm trying to think of a better name.

djc: Yeah, if batch is the default it makes sense to me to rename `install_batch()` to just `install()`.

djc: Some ideas to get the blood flowing: `install_slow()`, `install_single_thread()`, `install_per_trace()`.

jtescher: Yeah the intuition for "simple" isn't great, on the other hand the benefit is that it matches the official docs / spec / other languages https://github.com/open-telemetry/opentelemetry-specification/blob/3210212d4081ac542c698fd15cc725014feea6d1/specification/trace/sdk.md#built-in-span-processors

May be some leeway in the description as well for perf improvements:

> This is an implementation of `SpanProcessor` which passes finished spans and passes the export-friendly span data representation to the configured `SpanExporter`, as soon as they are finished.

E.g. (odd grammar aside) that description could still match a channel send to another thread as long as they are still sent to the exporter immediately rather than queued / batched?

TommyCpp: maybe `install_blocking()`? Indicate when span ends, we will block the current thread until the span exported

djc: `blocking` seems pretty good.

frigus02: > that description could still match a channel send to another thread as long as they are still sent to the exporter immediately rather than queued / batched

That's a good point. We could consider implementing this at some point in the future. We could also let the user configure the async runtime for the simple span processor in the future. That feels like too much for me to think about right now, though.

I dug a bit in the go implementation. They seem to call the simple span processor "sync":

https://github.com/open-telemetry/opentelemetry-go/blob/345f264a137ed7162c30d14dd4739b5b72f76537/sdk/trace/provider.go#L258-L268

jtescher: The spec intention reads "batching vs send now" to me more than "sync vs async" or "blocking vs non-blocking", but I could be wrong there. Hard to walk the line between the name being general enough that it doesn't lock in a specific implementation, and also intention revealing enough to not be meaningless or confusing.

frigus02: I read that as "batching vs send now", too. Could try something like:

- `.install_batch(runtime::Tokio)`
- `.install_immediate()`, `.install_one_by_one()`, `.install_single()`, `.install_individual()`

Or maybe `install_simple` is fine. It might be easy enough to find out the differences here. We could expand the documentation of the function a little bit and mention the difference.

frigus02: There is a discussion in the spec repo about this exact issue, which tends towards "single" although there are other suggestions as well: https://github.com/open-telemetry/opentelemetry-specification/issues/858

Since this is still ongoing I propose to stick with "simple" here for now.

TommyCpp:
Might need an empty commit to nudge the EasyCLA

TommyCpp (approved):
Nice job 👍

jtescher (approved):
LGTM @djc additional comments?

djc (approved):
Nope, LGTM!

frigus02:
I'm going to test the examples now to make sure they all still work

frigus02:
Result:
- No examples broken because of this PR
- Some examples had issues. I added commits to fix them in this PR
- I didn't manage to fix the grpc example. The client span doesn't seem to be sent.

From my side this is ready to be merged.

PR #483, opened 2021-03-17
fix: otel-otlp will try to create another tokio runtime inside one if async feature was not enabled.

TommyCpp:
- The bug was introduced in #467. I believe the async feature was there to spin up a tokio runtime if the exporter was not already in one. It's mostly because the tonic cannot work without a tokio runtime.
- Renamed the async feature to be less confusing.
- Cleaned up all the clippy problems in the metrics module and add some links into the doc.
- Add a with_timeout method in `opentelemetry/sdk/metrics/controller/push.rs` push controller to config the timeout of each push.

### DIFF
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -45,13 +45,15 @@ chrono = "0.4"
 tokio-stream = { version = "0.1", features = ["net"] }
 
 [features]
-async = ["default"]
 trace = ["opentelemetry/trace"]
 metrics = ["opentelemetry/metrics", "default"]
+
 default = ["tonic", "tonic-build", "prost", "tokio"]
-grpc-sys = ["grpcio", "protobuf", "protobuf-codegen", "protoc-grpcio"]
+tonic-sync = ["default"]
 tls = ["tonic/tls"]
 tls-roots = ["tls", "tonic/tls-roots"]
+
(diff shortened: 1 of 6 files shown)


frigus02:
Looks good to me.

I have to admit, the feature name "tokio-sync" is still not super obvious to me. But I can't think of a better name, either. Should we explain this particular feature in the readme?

djc:
This also seems to be fixing up quite a few formatting issues. We have rustfmt set up in CI, right? Shouldn't it be catching those.

I also find the `tokio-sync` name confusing. What are the semantics exactly? Maybe we can brainstorm a better name?

TommyCpp:
> We have rustfmt set up in CI, right? Shouldn't it be catching those.

I believe rustfmt doesn't really format the document so that could be the reason. 

I think the feature whose name in question is `tonic-sync`, which helps the user to run a tonic client when the application itself is not in a tokio environment by creating a tokio runtime internally.

djc:
I feel like we maybe just shouldn't provide that option. IIRC the error you get when spawning without a running runtime is pretty decent these days, and I feel like the cost/benefit ratio of providing this feature isn't so great.

TommyCpp:
Yeah, that's a fair point. I would think most of the users will use tonic inside a tokio runtime.

frigus02 (approved):

djc (approved):

PR #485, opened 2021-03-17
bump prometheus dependency

jan-xyz:
This is a bump to the latest version of the prometheus library.

```
# Changelog

## 0.12.0

 - Improvement: Fix format string in panic!() calls (#391)

 - Improvement: Replace regex with memchr (#385)

 - Improvement: Update reqwest requirement from ^0.10 to ^0.11 (#379)
```
https://github.com/tikv/rust-prometheus/compare/v0.11.0...v0.12.0


djc (approved):
Thanks!

TommyCpp (approved):

PR #487, opened 2021-03-18
Merge opentelemetry-stackdriver into this repo.

Xaeroxe:
Per @djc this PR merges the history of https://github.com/vivint-smarthome/opentelemetry-stackdriver into this repository.

I, the author of this crate, will soon have no further reason to work on this as I'm changing employers. When I informed @djc of this they suggested that we transfer this crate to the open-telemetry organization. That seems like a good idea to me, so this PR aims to accomplish that. If this PR is accepted I'll transfer the ownership on crates.io accordingly.


Xaeroxe:
@jasonahills @luisholanda can you guys agree to the CLA here?

djc:
Note, I've been a fairly substantial contributor to this project for the last few months -- I would be continuing my maintenance on this crate as part of the otel repo organization.

There might be an issue here because opentelemetry-stackdriver depends on tracing-opentelemetry, which depends on opentelemetry but does not live in this repository (and generally gets updated a bit slower than stuff in this repository).

Xaeroxe:
That's true @djc but that's only a dependency because of the example. I've deleted the example for now as it's non-trivial to get it working from this location. So, we can put off that problem.

djc:
Oh, good call!

Xaeroxe:
I have no idea why CI is failing at this point. I can't reproduce that problem locally, and I'm not getting any useful info from the CI logs.

Xaeroxe:
It's worth noting the code coverage test is the only one using a nightly toolchain.

opentelemetry-stackdriver/src/lib.rs
```diff
@@ -0,0 +1,356 @@
+/*
+   Copyright 2020 Vivint Smarthome
```
jtescher, line 2: Probably have to update copyright here too

jtescher (changes requested):
@Xaeroxe the coverage tests are a previous issue, don't worry about those. I think this org has an issue with MIT license, maybe remove and just keep the Apache 2?

Xaeroxe:
@jtescher Done! Thanks for your review!

jtescher (approved):
LGTM

TommyCpp (approved):
LGTM. We may need to edit the README file to move the `opentelemetry-stackdriver` into projects maintained by opentelemetry but we can also do that in a different PR.

PR #489, opened 2021-03-24
Fix jaeger span error reporting and spec compliance

jtescher:
* Fixes logic in error reporting where OK statuses are incorrectly marked as errors.
* Removes old status code and status message


frigus02 (approved):
Good catch 👍

asaaki:
Awesome fix. Any chance to get a bugfix release for the otel jaeger crate soon?

jtescher:
@asaaki released a patch version 👍

PR #490, opened 2021-03-24
opentelemetry-datadog: document panic in `new_pipeline`

little-dude:
This is already documented at the crate level, but it isn't clear
_where_ the panic can occur. This makes it more explicit.

Edit: Thanks a lot for this crate it is awesome :)
Edit2: I find the CLA thing pretty invasive. Do I really have to give my address to contribute code?!


TommyCpp:
Looks like something to do `Runtime`. Which runtime are you using?

PR #491, opened 2021-03-25
Remove unnecessary clone in `Key` type

Folyd:
I'm quite sure we should consume the `Key`'s ownership to construct a new `KeyValue`.
If the user did need to reuse the `Key` to create multiple `KeyValue`, they can just clone it later. 

This change can slightly improve the performance I think. However, I haven't done a benchmark.


djc (approved):
Makes sense to me!

PR #493, opened 2021-03-25
Update Semantic Conventions

frigus02:
Semantic conventions were a bit outdated. They also exist as YAML files now and there is a code generation tool. This PR:

- Changes the generation script to use the `otel/semconvgen` Docker image and the semantic conventions YAML files.
- Updates the `const`s based on version v1.1.0 of the spec.

This means we loose the manual changes we made to some of the constants. Possibly the most notable one is the Rust backtrace example. But using the generation tool is so much easier that I think it's worth it.

Let me now what you think.

### DIFF
--- a/opentelemetry-semantic-conventions/.gitignore
+++ b/opentelemetry-semantic-conventions/.gitignore
@@ -1,0 +1,1 @@
+opentelemetry-specification/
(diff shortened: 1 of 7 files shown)


jtescher (approved):
Looks like an improvement! thanks @frigus02

PR #495, opened 2021-03-26
fix: all clippy errors introduced in rust 1.51.0

TommyCpp:
* Fix all clippy errors
* Update protobuf version


djc (approved):

opentelemetry-aws/src/lib.rs
```diff
@@ -216,9 +216,9 @@ pub mod trace {
     #[derive(Clone, Debug, PartialEq)]
     struct XrayTraceId(String);
 
-    impl Into<Result<TraceId, ()>> for XrayTraceId {
-        fn into(self) -> Result<TraceId, ()> {
-            let parts: Vec<&str> = self.0.split_terminator('-').collect();
+    impl From<XrayTraceId> for Result<TraceId, ()> {
```
frigus02, line 219: I think `TryFrom` may be even more idiomatic here, i.e.:

```rust
use std::convert::TryFrom;

impl TryFrom<XrayTraceId> for TraceId {
    type Error = ();

    fn try_from(id: XrayTraceId) -> Result<Self, Self::Error> {
        // ...
    }
}
```

Probably out of scope for this PR, though.

opentelemetry/src/sdk/metrics/aggregators/ddsketch.rs
```diff
@@ -49,32 +49,32 @@ pub fn ddsketch(config: &DDSketchConfig, kind: NumberKind) -> DDSKetchAggregator
 ///
 /// In order to support both negative and positive inputs, DDSketchAggregator has two DDSketch store within itself to store the negative and positive inputs.
 #[derive(Debug)]
-pub struct DDSKetchAggregator {
+pub struct DdSKetchAggregator {
```
frigus02, line 52: Should the "K" be lowercase here?

TommyCpp: :+1: Good catch. Wonder why clippy didn't find it

frigus02 (approved):

PR #501, opened 2021-03-29
Remove #[must_use] from set_tracer_provider

LukeMathWalker:
I believe `#[must_use]` here is a legacy leftover from the previous version of `opentelemetry`, where `set_tracer_provider` returned a guard.
I have updated the documentation as well to clarify what is being returned.


jtescher (approved):
Good catch

PR #502, opened 2021-03-29
Simple span processor perf

jtescher:
Currently the simple span processor his _highly_ sensitive to any latency in its underlying exporter as it will block the current thread for the full export duration when each span ends.

This patch addresses this by moving span exporting to a separate thread, and communicating via channels.

Spans are still exported as soon as they end, and shutdown will wait for all spans to be successfully exported, preserving the simple span processor semantics of immediate exports, and never dropping data.

### DIFF
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -34,6 +34,7 @@ serde = { version = "1.0", features = ["derive", "rc"], optional = true }
 thiserror = "1"
 tokio = { version = "1.0", default-features = false, features = ["rt", "time"], optional = true }
 tokio-stream = { version = "0.1", optional = true }
+crossbeam-channel = { version = "0.5", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3"
@@ -47,7 +48,7 @@ tokio-stream = "0.1"
 
 [features]
 default = ["trace"]
-trace = ["rand", "pin-project", "async-trait", "percent-encoding"]
+trace = ["crossbeam-channel", "rand", "pin-project", "async-trait", "percent-encoding"]
 metrics = ["dashmap", "fnv"]
 serialize = ["serde"]
 testing = ["trace", "metrics", "rt-async-std", "rt-tokio", "rt-tokio-current-thread"]
--- a/opentelemetry/src/sdk/trace/span_processor.rs
+++ b/opentelemetry/src/sdk/trace/span_processor.rs
@@ -155,7 +155,7 @@ impl SpanProcessor for SimpleSpanProcessor {
     }
 
     fn shutdown(&mut self) -> TraceResult<()> {
-        if let Ok(_) = self.sender.send(None) {
+        if self.sender.send(None).is_ok() {
             if let Err(err) = self.shutdown.recv() {
                 global::handle_error(TraceError::from(format!(
                     "error shutting down span processor: {:?}",
--- a/opentelemetry/src/sdk/trace/span_processor.rs
+++ b/opentelemetry/src/sdk/trace/span_processor.rs
@@ -43,8 +43,7 @@ use crate::{
     Context,
 };
 use futures::{channel::mpsc, channel::oneshot, executor, future::Either, pin_mut, StreamExt};
-use std::env;
-use std::{fmt, str::FromStr, sync::Mutex, time::Duration};
+use std::{env, fmt, str::FromStr, sync::Mutex, thread, time::Duration};
 
 /// Delay interval between two consecutive exports.
 const OTEL_BSP_SCHEDULE_DELAY: &str = "OTEL_BSP_SCHEDULE_DELAY";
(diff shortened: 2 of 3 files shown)


frigus02:
The implementation looks good.

I wonder if this is a good idea, though.

- At the moment the simple span processor is highly sensitive to any latency in its underlying exporter. But it is easily predictable.
- If I have a program using the simple span processor, where the program logic is faster than the exporter, then with the current implementation my program will run slower. But it should run just fine. With the change it can possibly build up a very large backlog of spans, which consumes memory. The bigger the difference between span exporter latency and business logic speed the smaller the overall benefit of making the requests in another thread.

That said, I'm really not sure either way. There are certainly benefits in exporting in a different thread. I wouldn't object to merging this.

jtescher:
Yeah I'm open to alternatives as well, I think there are always going to be drawbacks to the simple strategy, the benefit in this approach is that it should be better for cases where the simple strategy makes sense (sending one-by-one, never dropping spans). If you are producing faster than you can export, you probably should be using the batch span processor to do load shedding. 

The main alternative seems to be to close this and keep the existing behavior which avoids the OOM by simply blocking the progress of the current thread, but it seems to conflict a bit with https://github.com/open-telemetry/opentelemetry-specification/pull/1555, which adds a non-blocking requirement to ending spans.

frigus02:
I don't know of good alternatives. And you're right. The spec suggests that we shouldn't block, which the simple span processor currently does.

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -138,15 +155,16 @@ impl SpanProcessor for SimpleSpanProcessor {
...
-        if let Ok(mut exporter) = self.exporter.lock() {
-            exporter.shutdown();
-            Ok(())
-        } else {
-            Err(TraceError::Other(
-                "When shutting down the SimpleSpanProcessor, the exporter's lock has been poisoned"
-                    .into(),
-            ))
+        if self.sender.send(None).is_ok() {
+            if let Err(err) = self.shutdown.recv() {
```
frigus02, line 159: What's the advantage of the shutdown channel? Could we store the `JoinHandle` of the thread and join here? Is it that `JoinHandle::join` takes `self` and so we would have to put it inside an `Option` or something similar?

opentelemetry/Cargo.toml
```diff
@@ -34,6 +34,7 @@ serde = { version = "1.0", features = ["derive", "rc"], optional = true }
 thiserror = "1"
 tokio = { version = "1.0", default-features = false, features = ["rt", "time"], optional = true }
 tokio-stream = { version = "0.1", optional = true }
+crossbeam-channel = { version = "0.5", optional = true }
```
frigus02, line 37: Do we use crossbeam because the `crossbeam_channel::Sender` is `Sync`?

frigus02 (approved):

opentelemetry/src/sdk/trace/span_processor.rs
```diff
@@ -104,13 +103,37 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
...
-    pub(crate) fn new(exporter: Box<dyn SpanExporter>) -> Self {
+    pub(crate) fn new(mut exporter: Box<dyn SpanExporter>) -> Self {
+        let (span_tx, span_rx) = crossbeam_channel::unbounded();
+        let (shutdown_tx, shutdown_rx) = crossbeam_channel::bounded(0);
+
+        let _ = thread::Builder::new()
+            .name("opentelemetry-exporter".to_string())
+            .spawn(move || {
+                while let Ok(Some(span)) = span_rx.recv() {
+                    if let Err(err) = executor::block_on(exporter.export(vec![span])) {
```
TommyCpp, line 119: Should we consider adding a timeout around here to prevent the exporter block for too long?

TommyCpp:
Overall LGTM. I think we may also consider supporting user without async runtime to use batch span processor somehow, may be helpful as I think batch span processor generally have a better performance

opentelemetry/src/sdk/trace/span_processor.rs
jtescher: Had join handle originally, could probably go back to that for simplicity

opentelemetry/Cargo.toml
jtescher: More for performance, could impl with std channel as well and have crossbeam be an opt-in feature for people worried about the additional dependency maybe?

opentelemetry/src/sdk/trace/span_processor.rs
jtescher: I was considering that, but the spec doesn't have any config for it and if you are creating faster than you are exporting to the point where it would be a big issue you should probably just use the batch processor as your app may OOM or do other bad things as well with the simple processor.

TommyCpp: To that point we should probably warn users that using simple span processor may cause OOM.

jtescher: Yeah I think a docs PR followup on the difference between both and when to choose one vs the other would be a good idea 👍

opentelemetry/Cargo.toml
frigus02: I think it's fine. We could possibly replace some uses of `futures::channel::mpsc` with crossbeam as well later on if there are performance benefits.

opentelemetry/src/sdk/trace/span_processor.rs
frigus02: Probably doesn't matter. I was just curious 🙂

TommyCpp (approved):

PR #505, opened 2021-03-30
Local span perf improvements

jtescher:
Currently span data recording performance is capped because all changes are synchronized behind a mutex. This patch changes the span API to take mutable references instead, and only requires the use of mutexes when setting values through a shared reference to the active span in the context API.

<details><summary>results: 19-47% performance improvement</summary>

```
start-end-span/always-sample
                        time:   [321.64 ns 323.85 ns 326.40 ns]
                        change: [-48.064% -47.628% -47.169%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) high mild
  2 (2.00%) high severe
start-end-span/never-sample
                        time:   [219.28 ns 220.25 ns 221.29 ns]
                        change: [-34.537% -33.419% -32.409%] (p = 0.00 < 0.05)
                        Performance has improved.
  8 (8.00%) high mild
  2 (2.00%) high severe

                        time:   [895.10 ns 904.83 ns 916.50 ns]
                        change: [-31.319% -30.690% -30.038%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  3 (3.00%) high severe
                        time:   [249.18 ns 250.59 ns 252.33 ns]
                        change: [-30.774% -29.948% -29.168%] (p = 0.00 < 0.05)
                        Performance has improved.
  8 (8.00%) high mild
  5 (5.00%) high severe

                        time:   [1.5984 us 1.6098 us 1.6226 us]
                        change: [-27.169% -25.705% -24.416%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  3 (3.00%) high severe
                        time:   [287.27 ns 288.77 ns 290.62 ns]
                        change: [-24.716% -23.535% -22.466%] (p = 0.00 < 0.05)
                        Performance has improved.

                        time:   [1.2813 us 1.2911 us 1.3018 us]
                        change: [-25.382% -24.399% -23.483%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  2 (2.00%) high severe
                        time:   [263.15 ns 264.19 ns 265.32 ns]
                        change: [-31.910% -30.959% -30.085%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  5 (5.00%) high severe

                        time:   [2.2835 us 2.2974 us 2.3119 us]
                        change: [-20.158% -19.254% -18.254%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  2 (2.00%) high severe
                        time:   [312.36 ns 313.86 ns 315.61 ns]
                        change: [-27.570% -26.865% -26.059%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  4 (4.00%) high severe
```

</details>

### DIFF
--- a/opentelemetry/src/sdk/export/trace/mod.rs
+++ b/opentelemetry/src/sdk/export/trace/mod.rs
@@ -1,16 +1,14 @@
 //! Trace exporters
-use std::fmt::Debug;
-use std::sync::Arc;
-use std::time::SystemTime;
-
-use async_trait::async_trait;
-#[cfg(feature = "serialize")]
-use serde::{Deserialize, Serialize};
-
 use crate::{
     sdk,
     trace::{Event, Link, SpanContext, SpanId, SpanKind, StatusCode, TraceError},
 };
+use async_trait::async_trait;
+#[cfg(feature = "serialize")]
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::time::SystemTime;
 
 pub mod stdout;
 
--- a/opentelemetry/src/sdk/export/trace/mod.rs
+++ b/opentelemetry/src/sdk/export/trace/mod.rs
@@ -1,14 +1,16 @@
 //! Trace exporters
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::time::SystemTime;
+
+use async_trait::async_trait;
+#[cfg(feature = "serialize")]
+use serde::{Deserialize, Serialize};
+
 use crate::{
     sdk,
     trace::{Event, Link, SpanContext, SpanId, SpanKind, StatusCode, TraceError},
 };
-use async_trait::async_trait;
-#[cfg(feature = "serialize")]
-use serde::{Deserialize, Serialize};
-use std::fmt::Debug;
-use std::sync::Arc;
-use std::time::SystemTime;
 
 pub mod stdout;
 
--- a/opentelemetry/src/sdk/trace/tracer.rs
+++ b/opentelemetry/src/sdk/trace/tracer.rs
@@ -11,7 +11,7 @@ use crate::sdk::{
     trace::{
         provider::{TracerProvider, TracerProviderInner},
         span::{Span, SpanData},
-        EvictedHashMap, EvictedQueue, SamplingDecision, SamplingResult,
+        Config, EvictedHashMap, EvictedQueue, SamplingDecision, SamplingResult,
     },
     InstrumentationLibrary,
 };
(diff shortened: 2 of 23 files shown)


opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -192,11 +190,13 @@ impl crate::trace::Tracer for Tracer {
                 _ => cx,
             }
         };
+        let span = parent_cx.span();
         let parent_span_context = if parent_cx.has_active_span() {
-            Some(parent_cx.span().span_context())
+            Some(span.span_context())
```
frigus02, line 195: The order seemed odd to me here. How can we get a span before checking if there is one? After looking at the implementation it makes sense now.

Out of curiosity: I think the following isn't the same, is it? Because I could write something like `Context::current_with_span(NoopSpan::new())`, in which case we should use explicitly provided invalid parent context.

```rust
let span = parent_cx.span();
let span.span_context().is_valid() {
    Some(span.span_context())
} else {
    None
}
```

frigus02 (approved):
This looks like a good improvement. Nice work!

opentelemetry/src/sdk/export/trace/mod.rs
```diff
@@ -1,16 +1,14 @@
...
 use crate::{
     sdk,
     trace::{Event, Link, SpanContext, SpanId, SpanKind, StatusCode, TraceError},
 };
+use async_trait::async_trait;
+#[cfg(feature = "serialize")]
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::time::SystemTime;
```
djc, line 11: I usually prefer to separate std, third-party, and crate-local imports, and I think that's a common way of doing things. Is grouping everything the common style here?

jtescher: oh this was actually not intentional, but it does seem like a standard grouping formatting in rustfmt would be nice.

opentelemetry/src/sdk/trace/tracer.rs
jtescher: could check the spec but I think that is allowed? would be a way to make a new root span I think?

opentelemetry/src/sdk/export/trace/mod.rs
jtescher: reverted

opentelemetry/src/sdk/trace/tracer.rs
frigus02: Yeah, I think so. Oh well 😄

djc (approved):

PR #507, opened 2021-04-01
fix: otlp docs cannot compile.

### LINKED ISSUE #496: no docs.rs build for opentelemetry-otlp 0.6.0
See https://docs.rs/crate/opentelemetry-otlp/0.6.0

TommyCpp:
The bug was introduced since we are compiling grpc files using both tonic and grpcio.

The build logs say:
 thread 'main' panicked at 'Error generating protobuf: Os { code: 30, kind: Other, message: "Read-only file system" }

But we already have compiled files in `proto/grpcio`. Thus ignore this error and print a warning message.

fix #496

### DIFF
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -25,6 +25,11 @@ name = "smoke"
 path = "tests/smoke.rs"
 required-features = ["integration-testing"]
 
+[[test]]
+name = "grpc_build"
+path = "tests/grpc_build.rs"
+required-features = ["grpc-sys"]
+
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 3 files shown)


frigus02 (approved):
Since we commit the generated grpcio proto files I wonder if we should `compile_grpc_protos` during build time at all. Could we instead follow the [manual generation](https://github.com/tikv/grpc-rs#option-1---manual-generation) docs? But I might be missing something and that can also be another improvement later on.

This looks like an easy fix for the issue 👍

djc:
One trick I've seen is to have a test case that checks if generated code in `src` is still up to date. That moves any build dependencies to dev dependencies which is nice by itself.

TommyCpp:
Moved the building part into test files.

Looks like there is a bug in actix-http-3.0.0-beta.4 that causes the lint failure. Not sure when it can be fixed.

frigus02 (approved):
Looks great!

PR #508, opened 2021-04-04
Remove remote span context

jtescher:
Storing a Span's SpanContext two separate ways in a Context (one for local another for remote) is unnecessary and adds complexity throughout the project when determining heredity of a Span. This moves to storing the Span directly in the Context uniformly (for both local and remote) as current Span.


frigus02 (approved):
This is nice. 👍

PR #510, opened 2021-04-05
Use current span for SDK-less context propagation

frigus02:
In the absence of an SDK the API should not record spans anymore but it should propagate the span context. In our case absence of an SDK means we're using the `NoopTracer`.

See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk

There are different ways to create new spans and they provide different ways of specifying the parent context:

- `Tracer::start`: uses `Context::current()`
- `Tracer::build`: uses the context from the span builder or falls back to the current context. BUT we did not have this fallback in the `NoopTracer`.
- `Tracer::start_with_context`: uses the specified context

This changes the `SpanBuilder` to make its `parent_context` non-optional. That way the current context fallback is done on
`SpanBuilder` creation. Tracer implementations just need to read the `parent_context` from the span builder.

---

- I hope I understood the spec correctly here 🙂 
- I'm not entirely sure if this is the best way to do it. Happy to take suggestions.

### DIFF
--- a/opentelemetry/src/sdk/trace/tracer.rs
+++ b/opentelemetry/src/sdk/trace/tracer.rs
@@ -139,10 +139,7 @@ impl crate::trace::Tracer for Tracer {
     where
         T: Into<Cow<'static, str>>,
     {
-        let mut builder = self.span_builder(name);
-        builder.parent_context = Some(cx);
-
-        self.build(builder)
+        self.build(SpanBuilder::from_name_with_context(name, cx))
     }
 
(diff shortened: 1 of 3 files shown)


TommyCpp (approved):
Nice catch. LGTM

jtescher:
This looks good, wondering if there are cases where people may want to build up span data before they have the context, but can't think of any reason currently.

PR #511, opened 2021-04-05
opentelemetry-http: return response from HttpClient

frigus02:
Make the `HttpClient` trait in opentelemetry-http return a HTTP response instead of `()`. This enables exporters using this trait to implement retry behavior based on the result of the call.

The response currently contains the status code and response body. If needed we can also add the response headers later on.

The response body is typed as `bytes::Bytes`, which seems was the easiest to implement. `reqwest` returns `Bytes` and we can't easily get to the underlying bytes without copying. The other clients return a form of `[u8]`, which is easy to turn into `Bytes`.

The only error that can be returned from the `HttpClient` is because of connection issues, timeouts or possibly infinite redirects. I can create a custom error type instead of `Box<Error>` if that seems more appropriate.

My use case is the Azure Application Insights exporter. AppInsights may accept only some of the spans in a batch and not others because of rate limiting. To understand which ones went through we need to look at the status code and response body.

### DIFF
--- a/opentelemetry-datadog/Cargo.toml
+++ b/opentelemetry-datadog/Cargo.toml
@@ -38,5 +38,7 @@ lazy_static = "1.4"
 
 [dev-dependencies]
 base64 = "0.13"
+bytes = "1"
+futures-util = "0.3"
 isahc = "0.9"
 opentelemetry = { path = "../opentelemetry", features = ["trace", "testing"] }
--- a/opentelemetry-datadog/src/exporter/mod.rs
+++ b/opentelemetry-datadog/src/exporter/mod.rs
@@ -12,7 +12,7 @@ use opentelemetry::sdk::export::trace;
 use opentelemetry::sdk::export::trace::SpanData;
 use opentelemetry::trace::TraceError;
 use opentelemetry::{global, sdk, trace::TracerProvider};
-use opentelemetry_http::HttpClient;
+use opentelemetry_http::{HttpClient, ResponseExt};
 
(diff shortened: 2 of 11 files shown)


TommyCpp (approved):
LGTM. Given that many exporter uses HTTP protocol. It may be reasonable to include a HTTP error  in export error that can handle the response of http client as needed.

frigus02:
> Given that many exporter uses HTTP protocol. It may be reasonable to include a HTTP error in export error that can handle the response of http client as needed.

Are you talking about the  `response.status().is_success()` logic? We could create a helper function inside opentelemetry-http so the for the existing exporters the code could look like this. And `handle_http_error` would create a TraceError if the HTTP status code is >= 300.

```rust
opentelemetry_http::handle_http_error(res)?;
```

TommyCpp:
> Are you talking about the response.status().is_success() logic

Yeah I think here we have the same logic(check status and return error if needed) in different exporters. May as well to have a function or a error type to handle it

> We could create a helper function inside opentelemetry-http so the for the existing exporters the code could look like this

Yeah that also make sense 👍 .

frigus02:
Cool, will do. I'm also happy to create another enum variant in `TraceError` for HTTP errors. The logic to convert a response to that error probably needs to live in opentelemetry-http because otherwise opentelemetry would have to have a dependency on `http`.

frigus02:
Alright, that's working now. 🙂

jtescher (approved):
LGTM

PR #512, opened 2021-04-06
feat: add force_push_tracer_provider function.

### LINKED ISSUE #499: Flushing spans without shutting down the instrumentation
Is there a way to trigger flushing of all in-flight spans without shutting down the instrumentation?
E.g. a flushing equivalent of [`global::shutdown_tracer_provider`](https://docs.rs/opentelemetry/0.13.0/opentelemetry/global/fn.shutdown_tracer_provider.html)?

TommyCpp:
This helps users to force push all remaining spans within span processors.

fix #499


jtescher:
This generally looks good, seems like the name should be `force_flush` rather than `force_push` to match the others though right?

jtescher (approved):

PR #515, opened 2021-04-07
feat: allow users to bring their own tonic channel.

### LINKED ISSUE #486: opentelemetry_otlp not support uds endpoint
use unix domain socket address get err : ExportFailed(InvalidUri(InvalidUri(InvalidFormat)))

TommyCpp:
Sometimes users want to use custom channel like UDS. Added a `from_tonic_channel` function to allow users to bring the custom channel.

fixes #486

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -346,9 +346,9 @@ impl TonicPipelineBuilder {
         self
     }
 
-    /// Use `channel` as tonic's transports channel.
+    /// Use `channel` as tonic's transport channel.
     /// this will override tls config and should only be used
-    /// when working with non-HTTP transport.
+    /// when working with non-HTTP transports.
     ///
     /// Users MUST make sure the [`ExporterConfig::timeout`] is
     /// the same as the channel's timeout.
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -346,7 +346,7 @@ impl TonicPipelineBuilder {
         self
     }
 
-    /// Use `channel` as tonic's transport channel.
+    /// Use `channel` as tonic's transports channel.
     /// this will override tls config and should only be used
     /// when working with non-HTTP transport.
     ///
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -295,6 +295,7 @@ impl OtlpPipelineBuilder {
             exporter_config: self.exporter_config,
             trace_config: self.trace_config,
             tonic_config: TonicConfig::default(),
+            channel: None,
         }
     }
 
@@ -327,6 +328,7 @@ pub struct TonicPipelineBuilder {
     exporter_config: ExporterConfig,
     tonic_config: TonicConfig,
     trace_config: Option<sdk::trace::Config>,
+    channel: Option<tonic::transport::Channel>,
 }
 
 #[cfg(feature = "tonic")]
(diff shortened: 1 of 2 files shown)


jtescher:
Seems like a good idea to allow users to pass a channel when using tonic, could this be added through the builder to avoid the need to synchronize timeouts and generally simplify things?

jtescher:
Hm guess no reader so maybe not simpler

TommyCpp:
> could this be added through the builder to avoid the need to synchronize timeouts and generally simplify things

Adding it to the builder may simplify things as users won't need to manually construct config structs. 

But users will still need to sync the timeouts because we need to set up the timeout in [`Endpoint`](https://docs.rs/tonic/0.4.1/tonic/transport/struct.Endpoint.html#method.timeout)s instead of [`Channel`](https://docs.rs/tonic/0.4.1/tonic/transport/struct.Channel.html)s. Thus we cannot change the timeout of the user-provided channel.

TommyCpp:
Furthermore, we may want to find a way to allow users to provide more configurations in `Endpoint` like [concurrent_limit](https://docs.rs/tonic/0.4.1/tonic/transport/struct.Endpoint.html#method.concurrency_limit), [rate_limit](https://docs.rs/tonic/0.4.1/tonic/transport/struct.Endpoint.html#method.rate_limit)

opentelemetry-otlp/src/span.rs
```diff
@@ -206,16 +206,16 @@ impl Debug for TraceExporter {
...
     pub fn new_tonic(
         config: ExporterConfig,
         tonic_config: TonicConfig,
     ) -> Result<Self, crate::Error> {
-        let endpoint = TonicChannel::from_shared(config.endpoint)?;
+        let endpoint = TonicChannel::from_shared(config.endpoint.clone())?;
 
         #[cfg(feature = "tls")]
-        let channel = match tonic_config.tls_config {
+        let channel = match tonic_config.tls_config.clone() {
```
djc, line 218: I think we could only clone here in the `Some` case?

djc:
The design constraint that the timeouts must match feel like something that should usually be expressed in the type system somehow rather than only in the documentation. Maybe the additional constructor should take an `Endpoint` rather than a `Channel` (and document that it overwrites the timeout)?

TommyCpp:
> Maybe the additional constructor should take an `Endpoint` rather than a `Channel` (and document that it overwrites the timeout)

Sometimes ppl need to pass a connector function into `Endpoint` like in  [Uds](https://github.com/hyperium/tonic/blob/master/examples/src/uds/client.rs#L21-L26) example. Thus we might still need them to pass the constructed channel.

I also thought about allowing users to pass the connector function but the type of it seems to be bind with `tower` and is not trivial.

jtescher (approved):
Looks reasonable to me, @djc did you have other comments?

opentelemetry-otlp/src/lib.rs
```diff
@@ -344,6 +346,19 @@ impl TonicPipelineBuilder {
         self
     }
 
+    /// Use `channel` as tonic's transport channel.
+    /// this will override tls config and should only be used
+    /// when working with non-HTTP transport.
```
djc, line 351: Nit: I think `transport` should be plural here?

djc (approved):
Just one nit.

opentelemetry-otlp/src/lib.rs
jtescher: @TommyCpp I think he meant the other one, first line singular, last line plural

TommyCpp: Oops sorry I misunderstand that one 😬

PR #516, opened 2021-04-07
adding otlp http transport, using proto binary

### LINKED ISSUE #477: Add OTLP span exporter using HTTP/Binary protocol
As per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp). Add a OTLP exporter using http and binary.

Also may need to think about how do we build a builder pipeline to allow users to choose transportation protocol and request body format.

danrusei:
this commit fixes #477 
This PR implements the OTLP/HTTP that  uses HTTP POST requests to send telemetry data from clients to servers.
OTLP/HTTP uses Protobuf payloads encoded either in binary format or in JSON format, ths PR addresses only the binary format.
The Protobuf schema of the messages is the same for OTLP/HTTP and OTLP/gRPC. 

This is the first iteration, looking forward for your feedback.

### DIFF
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -44,6 +44,10 @@ protobuf = { version = "2.18", optional = true }
 thiserror = "1.0"
 tonic = { version = "0.4", optional = true }
 tokio = { version = "1.0", features = ["full"], optional = true }
+opentelemetry-http = { version = "0.2", path = "../opentelemetry-http", optional = true }
+reqwest = { version = "0.11", optional = true, default-features = false }
+surf = { version = "2.0", optional = true }
+http = "0.2"
 
 [dev-dependencies]
 chrono = "0.4"
@@ -64,5 +68,12 @@ openssl = ["grpcio/openssl"]
 openssl-vendored = ["grpcio/openssl-vendored"]
 integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "opentelemetry/trace"]
 
+http-proto = ["prost-build", "prost", "opentelemetry-http"]
+reqwest-blocking-client = ["reqwest/blocking", "opentelemetry-http/reqwest"]
+reqwest-client = ["reqwest", "opentelemetry-http/reqwest"]
+reqwest-rustls = ["reqwest", "reqwest/rustls-tls-native-roots"]
+surf-client = ["surf", "opentelemetry-http/surf"]
+
 [build-dependencies]
 tonic-build = { version = "0.4", optional = true }
+prost-build = {version = "0.7", optional = true }
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -148,7 +148,7 @@
 
 use opentelemetry::{global, runtime::Runtime, sdk, trace::TracerProvider};
 
-#[cfg(feature = "grpc-sys")]
+#[cfg(any(feature = "grpc-sys", feature = "http-proto"))]
 use std::collections::HashMap;
 
 use std::str::FromStr;
@@ -187,11 +187,17 @@ pub use crate::span::TonicConfig;
 #[cfg(feature = "grpc-sys")]
 pub use crate::span::GrpcioConfig;
 
+#[cfg(feature = "http-proto")]
+pub use crate::span::HttpConfig;
+
 #[cfg(feature = "metrics")]
 pub use crate::metric::{new_metrics_pipeline, MetricsExporter, OtlpMetricPipelineBuilder};
 
 #[cfg(feature = "grpc-sys")]
 pub use crate::span::{Compression, Credentials};
+
+#[cfg(feature = "http-proto")]
+use opentelemetry_http::HttpClient;
 
 use opentelemetry::sdk::export::ExportError;
 use opentelemetry::trace::TraceError;
@@ -306,6 +312,17 @@ impl OtlpPipelineBuilder {
             exporter_config: self.exporter_config,
             trace_config: self.trace_config,
             grpcio_config: GrpcioConfig::default(),
+        }
+    }
+
+    /// Use HTTP as transport layer, return a `HttpPipelineBuilder` to config the http transport
+    /// and build the exporter
+    #[cfg(feature = "http-proto")]
+    pub fn with_http(self) -> HttpPipelineBuilder {
+        HttpPipelineBuilder {
+            exporter_config: self.exporter_config,
+            trace_config: self.trace_config,
+            http_config: HttpConfig::default(),
         }
     }
 }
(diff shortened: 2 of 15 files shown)


opentelemetry-otlp/src/proto.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(feature = "tonic")]
+#[cfg(any(feature = "tonic", feature="http-proto"))]
```
TommyCpp, line 1: It's good to reuse the existing proto-generated files. But this also makes `http-proto` to be dependent on the `tonic`. It may need `prost` but I don't think it need `tonic`

opentelemetry-otlp/src/span.rs
```diff
@@ -70,6 +82,18 @@ pub enum TraceExporter {
         /// The Grpc trace exporter
         trace_exporter: GrpcioTraceServiceClient,
     },
+    #[cfg(feature = "http-proto")]
+    /// Trace Exporter using HTTP transport
+    HTTP {
```
TommyCpp, line 87:
```suggestion
    Http {
```

I think [the naming convention](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) requires camel case for enum variants

opentelemetry-otlp/src/span.rs
```diff
@@ -336,6 +425,46 @@ impl SpanExporter for TraceExporter {
...
+            #[cfg(feature = "http-proto")]
+            TraceExporter::HTTP {
+                trace_exporter,
+                collector_endpoint,
+                headers,
+                ..
+            } => {
+                let req = TonicRequest {
+                    resource_spans: batch.into_iter().map(Into::into).collect(),
+                };
```
TommyCpp, line 438: Does this mean the `http-proto` will depends on the `tonic`? We should probably figure out how to generate proto file with `prost` only and use it for `tonic` and `http-proto`

TommyCpp (changes requested):
Overall a good start point. I like the new example as it include docker setup to run a collector.

We may also want to modify the `lint.sh` to add run `cargo check` for the new features. I notice the `cargo check --features=http-proto --no-default-features` fails.

opentelemetry-otlp/Cargo.toml
```diff
@@ -64,5 +68,11 @@ openssl = ["grpcio/openssl"]
 openssl-vendored = ["grpcio/openssl-vendored"]
 integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "opentelemetry/trace"]
 
+http-proto = ["prost", "opentelemetry-http"]
```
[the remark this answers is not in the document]
danrusei: Indeed it's difficult to select the right naming as the underlying transport is the same for both proto/binary and json, as far as I am aware, only the body of the request differs. Wouldn't it better to use a generic feature name that refers to both proto/binary and json, and maybe add a function to the HttpPipelineBuilder afterwards that allows the lib user to choose between the binary or json Protocol ?

opentelemetry-otlp/src/proto.rs
danrusei: not sure if there is a dependence on tonic, I'm using "any" for conditional compilation to make it true if any of the predicate is used.

opentelemetry-otlp/src/span.rs
danrusei: I'll change it, thanks.

opentelemetry-otlp/src/span.rs
danrusei: TonicRequest is just an alias to the ExportTraceServiceRequest (at the top of teh file), should we use another name ?
I believe in order to generate the proto rs files, the tonic-build is used, that is listed under build-dependencies and it is not available in the package itself.

opentelemetry-otlp/src/proto.rs
TommyCpp: I was more referring to the fact that we use `tonic_build` here, which is not listed as a dependency for feature `http-proto`. Is it your intention to ask users to specify `tonic_build` explicitly?

opentelemetry-otlp/src/span.rs
TommyCpp: Yeah I think it would be reasonable to list the `tonic_build` as a dependency of `http-proto` as it was marked optional.

opentelemetry-otlp/Cargo.toml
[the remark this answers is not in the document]
TommyCpp: Oh I see. Yes it make sense to use one feature for both json and binary. `http` would have been a good feature name in this case but we already have `http` dependency. I think `http-proto` would do the trick in this case as it indicate we both need http as transport layer and protobuf support to compile the necessary files/types

opentelemetry-otlp/src/span.rs
danrusei: I see your point and I don't have a strong opinion on this. The tonic-build is a slim package that includes prost-build, for consistency I would use it, but we could try to build the protos with prost-build if you feel that it is better.

danrusei: also should I remove the alias name TonicRequest and leave it with the original ExportTraceServiceRequest to eliminate any possible confusions ?

TommyCpp: You are right, I tweaked it a little bit over the weekend. Looks like the `tonic_build` is a wrapper of `prost-build` so I think we could just use the `tonic_build`. Just need to add it into the dependency.

---
> also should I remove the alias name TonicRequest and leave it with the original ExportTraceServiceRequest to eliminate any possible confusions ?

Yeah that sounds like a good idea 👍  I think those types were named as `TonicXXX` just to distinguish with grpcio generated types.

danrusei: I take my word back, I'm afraid that we should continue to refer to TonicXXX, as tonic remains a dependency as long as the transformations (like SpanData to Tonic ResourceSpans ) and others are constrained by tonic feature utilization. Not sure what would be the best approach here.

TommyCpp: It would be best if we can avoid `tonic` dependency here as we don't need it for HTTP-based transports. One way to avoid it is to use `prost-build` to build another set of type files. It will requires some changes to the `build.rs`.

Currently, the files generated by `tonic` are located in `OUT_DIR`. So in order to have generated files from `prost-build` and `tonic-build`. We should first try to build them into their respective folders.

And then we should use `include!` to include those generated files in `proto` mod.

danrusei: I agree, we can try it out. The only thing I'm a little afraid is the duplication of code in the "transform" module, as we may have to  create all the transformations (ex from SpanData to whatever resources the prost-build will create), which I expect to look similar to Tonic generated resources.

TommyCpp: I think it's fine as we already have two transform, one for grpcio and the other one for tonic. But if you feel it's too much for one PR we can merge this one with tonic dependency and use another PR to add transform.

danrusei: I have updated based on your suggestions.

opentelemetry-otlp/Cargo.toml
```diff
@@ -64,5 +68,12 @@ openssl = ["grpcio/openssl"]
 openssl-vendored = ["grpcio/openssl-vendored"]
 integration-testing = ["tonic", "tonic-build", "prost", "tokio/full", "opentelemetry/trace"]
 
+http-proto = ["prost-build", "prost", "opentelemetry-http"]
+reqwest-blocking-client = ["reqwest/blocking", "opentelemetry-http/reqwest"]
+reqwest-client = ["reqwest", "opentelemetry-http/reqwest"]
+reqwest-rustls = ["reqwest", "reqwest/rustls-tls-native-roots"]
+surf-client = ["surf", "opentelemetry-http/surf"]
```
TommyCpp, line 75: May need to include those into lint.sh to make sure it compiles.

TommyCpp (commented):
Overall looks good :+1: . Some small nits. We could add some unit tests in next PR. Thanks for working on this.

opentelemetry-otlp/Cargo.toml
danrusei: done.

opentelemetry-otlp/src/lib.rs
```diff
@@ -553,7 +657,8 @@ pub enum Protocol {
     Grpc,
     // TODO add support for other protocols
     // HttpJson,
-    // HttpProto,
+    /// HTTP protocol with binary protobuf
+    HttpProto,
```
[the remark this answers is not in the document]
danrusei: sure.

TommyCpp (approved):
LGTM 👍 Thanks for working on this 🎉

opentelemetry-otlp/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-//! The OTLP Exporter supports exporting trace and metric data in the OTLP
```
jtescher, line 1: Was this intentionally deleted?

PR #517, opened 2021-04-08
Fix coverage tests

jtescher:


TommyCpp:
I think there are still some issues with codecov. Maybe we need to update their default branch to `main`?

jtescher:
Hm something odd with gcov

```
==> Running gcov in . (disable via -X gcov)
./target/debug/deps/serde_qs-449a08a5b9a9f17e.gcno:version '408*', prefer 'A93*'
find: ‘gcov’ terminated by signal 11
```

TommyCpp:
> Hm something odd with gcov
> 
> ```
> ==> Running gcov in . (disable via -X gcov)
> ./target/debug/deps/serde_qs-449a08a5b9a9f17e.gcno:version '408*', prefer 'A93*'
> find: ‘gcov’ terminated by signal 11
> ```

It has been the error we see for a while. Checked the GitHub action repo but I didn't find anything obvious. I believe the report was uploaded to codecov according to the log. But the codecov cannot process it somehow.

jtescher:
@TommyCpp specifying the output file seems to have fixed the issue 🎉

TommyCpp (approved):
Nice fix :tada:

PR #519, opened 2021-04-14
Always export batch when limit reached

### LINKED ISSUE #468: batchexporter doesnt export spans when batch size is reached

jtescher:
Currently spans will begin to drop if more than the max queue size are generated within a given batch window. This change causes the batch to export when full, allowing more spans to be exported within a given window even if the scheduled export time has not yet been reached.

Resolves #468

### DIFF
--- a/opentelemetry/src/sdk/trace/span_processor.rs
+++ b/opentelemetry/src/sdk/trace/span_processor.rs
@@ -246,30 +246,30 @@ impl SpanProcessor for BatchSpanProcessor {
 
     fn force_flush(&self) -> TraceResult<()> {
         let mut sender = self.message_sender.lock().map_err(|_| TraceError::from("When force flushing the BatchSpanProcessor, the message sender's lock has been poisoned"))?;
-        let (res_sender, res_receiver) = oneshot::channel::<Vec<ExportResult>>();
+        let (res_sender, res_receiver) = oneshot::channel();
         sender.try_send(BatchMessage::Flush(Some(res_sender)))?;
-        for result in futures::executor::block_on(res_receiver)? {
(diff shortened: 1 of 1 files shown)


TommyCpp (commented):
Overall looks good 👍 . I wonder if it make sense to move the background job into a dedicate function and add some comment to explain how it works

jtescher:
> I wonder if it make sense to move the background job into a dedicate function and add some comment to explain how it works

Open to suggestions, its relatively simple (loop with 3 branches based on message type), so moving the closure into a function doesn't bring all that much clarity, it _might_ be easier to reason about if the three inner blocks were functions, but most of what they are doing is error handling and the function calls would add some indirection making the code a bit harder to follow, but happy to do that if others find it easier to read / reason about that way.

frigus02 (approved):

TommyCpp (approved):

jtescher:
Ok will merge this for now, we can think about cleanup with a subsequent issue or PR as it shouldn't be a breaking change.

PR #521, opened 2021-04-15
feat: add max_attributes_per_event and max_attributes_per_link.

TommyCpp:
Added `max_attributes_per_event` and `max_attributes_per_link`.

We enforced those rules in tracer or spans rather than having an `EvictedQueue` is because we cannot get the max_attributes when we create links or events. We cannot assume the users are using a tracer created by the global tracer provider as in theory users can have multiple tracer providers.

Need to add unit tests


jtescher:
Nice to add the additional limit config 👍, I wonder if there may be a better way to enforce at event creation (or if we need to change the event creation api to support limits better)

TommyCpp:
> I wonder if there may be a better way to enforce at event creation (or if we need to change the event creation api to support limits better)

I do think we should make members of `Event` private just in case we need to add fields. But in order to enforce the rules. `Event` must have some connection with the tracer provider when creating. We could use a tracer to create an event but I don't think it's in the spec and it feels wired TBH.

opentelemetry/src/sdk/trace/span.rs
```diff
@@ -78,9 +79,18 @@ impl crate::trace::Span for Span {
...
     ) {
+        let max_attributes_per_event =
+            self.tracer
+                .provider()
+                .map(|provider| provider.config().max_attributes_per_event)
+                .unwrap_or(DEFAULT_MAX_ATTRIBUTES_PER_EVENT) as usize;
         self.with_data(|data| {
+            if attributes.len() > max_attributes_per_event {
+                let _dropped: Vec<_> = attributes.drain((max_attributes_per_event)..).collect();
+            }
```
frigus02, line 92: I think you should be able to replace these 3 lines with a call to [`truncate`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate):

```suggestion
            attributes.truncate(max_attributes_per_event);
```

jtescher:
@TommyCpp looks like the current config struct was renamed to span limits in https://github.com/open-telemetry/opentelemetry-specification/pull/1416, what do you think about the idea of having spans storing their limits directly to avoid the tracer provider lookup (e.g. the [go version](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L132))

TommyCpp:
> what do you think about the idea of having spans storing their limits directly to avoid the tracer provider lookup (e.g. the [go version](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L132))

We could do that, it increases the size of a span a little but we can avoid upgrading weak pointer in the tracer. But we still need to enforce the limit when adding events or links.

opentelemetry/src/sdk/trace/span_limit.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub(crate) const DEFAULT_MAX_EVENT_PER_SPAN: u32 = 128;
+pub(crate) const DEFAULT_MAX_ATTRIBUTES_PER_SPAN: u32 = 128;
+pub(crate) const DEFAULT_MAX_LINKS_PER_SPAN: u32 = 128;
+pub(crate) const DEFAULT_MAX_ATTRIBUTES_PER_EVENT: u32 = 128;
+pub(crate) const DEFAULT_MAX_ATTRIBUTES_PER_LINK: u32 = 128;
+
+/// Span limit configuration to keep attributes, events and links to a span in a reasonable number.
+#[derive(Copy, Clone, Debug)]
+pub struct SpanLimit {
```
jtescher, line 24: I think the spec wants this pluralized

```suggestion
pub struct SpanLimits {
```

opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -253,18 +254,31 @@ impl crate::trace::Tracer for Tracer {
...
             let mut attributes =
-                EvictedHashMap::new(config.max_attributes_per_span, attribute_options.len());
+                EvictedHashMap::new(span_limit.max_attributes_per_span, attribute_options.len());
             for attribute in attribute_options {
                 attributes.insert(attribute);
             }
-            let mut links = EvictedQueue::new(config.max_links_per_span);
+            let mut links = EvictedQueue::new(span_limit.max_links_per_span);
             if let Some(link_options) = &mut link_options {
+                for link in link_options.iter_mut() {
```
jtescher, line 263: Do we still need to do this? going through `append_vec` below will call [`push_back`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/sdk/trace/evicted_queue.rs#L32-L35) which should already handle the limits configured above

TommyCpp: The `append_vec`  only enforces the limit on the number of links. Not the number of attributes of a link.

jtescher: Oh sorry I see what's happening nvm

TommyCpp: It may make sense to redesign how we create a link to enforce the limit when we creating one :thinking: in the future. Gonna have to use `EvictedQueue` to store the attributes too

opentelemetry/src/sdk/trace/span.rs
```diff
@@ -51,11 +53,19 @@ impl Span {
         span_context: SpanContext,
         data: Option<SpanData>,
         tracer: sdk::trace::Tracer,
+        span_limit: Option<SpanLimits>,
```
jtescher, line 56: Does this need to be optional? Could switch tests to use `Default::default()` instead of `None`

opentelemetry/src/sdk/trace/tracer.rs
jtescher: Yeah can rethink he we store dropped count and others in a follow up PR perhaps

opentelemetry/src/sdk/trace/span.rs
TommyCpp: If the `span_limit` is `None`. Then we will try to get the span limit from the tracer provider. If there isn't a tracer provider connected with the tracer, we will use the default. 

Sometimes we may change the `span_limit` during the span building(For example, we may record the number of dropped entities) so we may pass modified span_limit. But we want to keep the span limit the same as the tracer provider, we can pass `None`.

opentelemetry/src/sdk/trace/tracer.rs
TommyCpp: My current plan is to add fields in `SpanLimits` to allow us to store the dropped number in it. 
Also need to print the warning message when the dropped number reaches some threshold. Cannot really warn whenever we drop because the spec says don't be too noisy on this issue.

opentelemetry/src/sdk/trace/span.rs
jtescher: Hm yeah we should store them somewhere. Maybe best path is to have `SpanLimits` not be optional here and add dropped counts to `SpanData`? keeps things simple and tracks attribute drops without having to upgrade the provider.

TommyCpp: Sounds reasonable. I removed the option around `SpanLimits`. I probably would open another PR to address how to record and expose the number of dropped entities.

jtescher (approved):
Ok good place to start

PR #525, opened 2021-04-17
Remove metrics quantiles

jtescher:
Quantiles did not end up being part of the current metrics spec


djc (approved):
Can hardly say no to +21/-337!

xnorpx:
@djc @jtescher what is the recommendation to get quantiles with this removed?

PR #526, opened 2021-04-20
fix: allow users to use custom export kind selector.

TommyCpp:
Follow up on #497, which allows users to bring their own aggregator selector.

Also updated the name of examples.


jtescher (approved):
LGTM

PR #529, opened 2021-04-21
Add dropped attribute counts to events and links

jtescher:


jtescher:
@TommyCpp was looking at the proto and it looks like it needs drop counts per [event/link](https://github.com/open-telemetry/opentelemetry-proto/blob/493f9f8f3bebd6c9019df77723819522fcd0873e/opentelemetry/proto/trace/v1/trace.proto#L183). This seem like the best approach? or given the count requirements there you think there may be a better approach?

TommyCpp (approved):
This feels like the most natural way to do so. LGTM. 
Sorry, I forgot to change the `span_limit` variable name into plural.

PR #535, opened 2021-04-30
fix: update metrics proto and metric transformation.

TommyCpp:
- Removed all `IntXXX` objects as they will be removed in coming months and are currently deprecated.
- Updated Grpcio generated files.


precommit.sh
```diff
@@ -1 +1 @@
-cargo fmt --all && ./scripts/lint.sh && ./scripts/test.sh
\ No newline at end of file
+cargo update && cargo fmt --all && ./scripts/lint.sh && ./scripts/test.sh
```
djc, line 1: Why do this? (It seems like the kind of thing that probably deserves its own commit or even PR.)

djc:
Most of the changes look good to me, maybe some discussion about the precommit hook change is warranted (it doesn't seem to really fit in with this PR).

precommit.sh
TommyCpp: Sorry, I should have added some comments around it. But it basically we update dependencies here to make sure the local test and lint have the same result as the ci pipeline. CI pipeline downloads those dependencies in their latest version. 

Sometimes the different versions of dependencies can cause problems. For example, if there is a bug in the new version. Or in this case, the version of `protobuf` was actually part of the files generated by grpcio. So if you don't update the dependency locally you will pass the tests locally but fail the ci pipeline.

djc: I guess it is in line with the rest of the precommit script. Personally, I wouldn't want to use this because it seems very heavyweight for something that runs implicitly.

djc (approved):

PR #537, opened 2021-05-03
Update resource merge behaviour

frigus02:
The resource merge behaviour changed 3 months ago to be consistent with Span.SetAttribute ([commit](https://github.com/open-telemetry/opentelemetry-specification/commit/77bfc630300b639f976128c865143581a178d545)). It now specifies:

> The resulting resource MUST have all attributes that are on any of the two input resources. If a key exists on both the old and updating resource, the value of the updating resource MUST be picked (even if the updated value is empty).

This updates sdk::Resource::merge to behave as described in the spec.

---

Note: I didn't have the time to check where `sdk::Resource::merge` is used and if this change affects anything, yet.


jtescher (approved):
LGTM thanks @frigus02

PR #538, opened 2021-05-04
feat: ignore links with invalid context

TommyCpp:
- Ignore links with invalid context
- Fix wrong shutdown function name in READMEs.

### DIFF
--- a/opentelemetry/src/trace/tracer.rs
+++ b/opentelemetry/src/trace/tracer.rs
@@ -458,14 +458,10 @@ impl SpanBuilder {
     }
 
     /// Assign links
-    pub fn with_links(self, links: Vec<Link>) -> Self {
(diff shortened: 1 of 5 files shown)


opentelemetry/src/trace/tracer.rs
```diff
@@ -460,7 +460,12 @@ impl SpanBuilder {
     /// Assign links
     pub fn with_links(self, links: Vec<Link>) -> Self {
         SpanBuilder {
-            links: Some(links),
+            links: Some(
+                links
+                    .into_iter()
+                    .filter(|l| l.span_context().is_valid())
+                    .collect(),
```
jtescher, line 467: Maybe use `Vec::retain` instead of allocating a new one here?

jtescher (approved):
Looks good!

PR #543, opened 2021-05-08
fix: lint issues in 1.52.0

TommyCpp:
* Also bumped the MSRV according to the README

### DIFF
--- a/opentelemetry-otlp/tests/grpc_build.rs
+++ b/opentelemetry-otlp/tests/grpc_build.rs
@@ -40,17 +40,17 @@ fn build_grpc() {
 }
 
 fn build_content_map() -> HashMap<PathBuf, String> {
-    let mut map = HashMap::new();
-    let dict =
-        std::fs::read_dir("src/proto/grpcio").expect("cannot open dict of generated grpc files");
-
-    for entry in dict.into_iter().flatten() {
-        map.insert(
-            entry.path(),
-            std::fs::read_to_string(entry.path()).unwrap_or_else(|_| {
-                panic!("cannot read from file {}", entry.path().to_string_lossy())
-            }),
-        );
-    }
-    map
+    std::fs::read_dir("src/proto/grpcio")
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/tests/grpc_build.rs
```diff
@@ -43,15 +43,14 @@ fn build_content_map() -> HashMap<PathBuf, String> {
...
-        if let Ok(entry) = entry {
-            map.insert(
-                entry.path(),
-                std::fs::read_to_string(entry.path()).unwrap_or_else(|_| {
-                    panic!("cannot read from file {}", entry.path().to_string_lossy())
-                }),
-            );
-        }
+
+    for entry in dict.into_iter().flatten() {
```
djc, line 47: Maybe rewrite this to collect() while you're at it? Potentially more efficient to allocate the correct size immediately.

djc (approved):

opentelemetry-otlp/tests/grpc_build.rs
TommyCpp: Sounds good to me. But I thought the compiler could allocate the correct size with for loop also?

opentelemetry-otlp/tests/grpc_build.rs
```diff
@@ -44,13 +44,17 @@ fn build_content_map() -> HashMap<PathBuf, String> {
...
-            entry.path(),
-            std::fs::read_to_string(entry.path()).unwrap_or_else(|_| {
-                panic!("cannot read from file {}", entry.path().to_string_lossy())
-            }),
-        );
-    }
+    let _ = dict
+        .into_iter()
+        .flatten()
+        .map(|entry| {
```
djc, line 50: I meant you could collect into `map` directly by yielding `(entry.path(), std::fs::read_to_string())` items.

jtescher (approved):

PR #551, opened 2021-05-13
Expose the `Error` type to allow users to set custom error handlers

LukeMathWalker:
Expose the `Error` type in the `global` module to allow users to define their own error handlers.  
This is currently impossible because `Error` is private and therefore users cannot provide a function that satisfies the constraints of `set_error_handler`.

### DIFF
--- a/opentelemetry/src/global/error_handler.rs
+++ b/opentelemetry/src/global/error_handler.rs
@@ -23,7 +23,7 @@ pub enum Error {
     #[cfg(feature = "metrics")]
     #[cfg_attr(docsrs, doc(cfg(feature = "metrics")))]
     #[error(transparent)]
-    /// Failed to export metrics.
+    /// An issue raised by the metrics module.
     Metric(#[from] MetricsError),
     #[error("{0}")]
     /// Other types of failures not covered by the variants above.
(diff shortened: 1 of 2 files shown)


opentelemetry/src/global/error_handler.rs
```diff
@@ -18,12 +18,15 @@ pub enum Error {
     #[cfg(feature = "trace")]
     #[cfg_attr(docsrs, doc(cfg(feature = "trace")))]
     #[error(transparent)]
+    /// Failed to export traces.
     Trace(#[from] TraceError),
     #[cfg(feature = "metrics")]
     #[cfg_attr(docsrs, doc(cfg(feature = "metrics")))]
     #[error(transparent)]
+    /// Failed to export metrics.
```
TommyCpp, line 26: I think [`MetricsError`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/metrics/mod.rs#L44) covers more than export errors. It also covers some configuration errors so maybe rephrase it into something like `errors from metrics module`  would be more accurate.

TommyCpp (approved):
Good catch. Thanks! Just a small suggestion on the comment

opentelemetry/src/global/error_handler.rs
LukeMathWalker: Updated 👍🏼

PR #560, opened 2021-05-26
feat: allow users to use different channels based on runtime in batch span processor

### LINKED ISSUE #520: Batch span processor on_end mutex

TommyCpp:
Resolve #520.
Close #533 

Another proposed solution to remove mutex in batch span processors. 

I tried to keep the type parameter within the channel types. But I don't think I figure out a way to do so without introducing a type parameter in `Runtime`. So currently all `Sender`s and `Receiver`s from `Runtime` will have item type as `BatchMessage`. Let me know there are better way to handle this.

TODO:

- [x] Return a better error message.


jtescher (approved):
LGTM

TommyCpp:
@jtescher I realized that the implementation used types from `trace` mod while `runtime` is not part of the `trace`. So I separate the channel part from `runtime` to `sdk/trace/runtime`.

jtescher:
@TommyCpp seems good 👍

TommyCpp:
I tested a POC of the custom channel from the unbounded channel and it didn't seem to outperform the runtime version. I'd suggest we merge this PR as we have been investigating this issue for a while.

frigus02 (approved):
Agree. This looks really good. Nice work 👍

jtescher:
Great work @TommyCpp

PR #562, opened 2021-06-01
Move hyper prometheus example into something runnable

dmathieu:
This change moves the hyper-prometheus example out of the opentelemetry-prometheus, and into the global examples folder.

This is the only example located in a subpackage, making it harder to find.
That example also wasn't executable, making it harder to experiment on. By making the folder a part of the workspace, we can now run `cargo run` on it and see the behavior.

I've also changed the behavior a little bit to return the metrics only on the `/metrics` path, which is the usually expected behavior from a Prometheus exporter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ members = [
     "examples/external-otlp-tonic-tokio",
     "examples/grpc",
     "examples/http",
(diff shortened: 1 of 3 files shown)


TommyCpp (approved):
Good catch 👍 LGTM!

PR #563, opened 2021-06-05
Update lib and tracing module docs with examples

jtescher:
This updates the overview pages to be a little higher level and adds getting started sections to show simple tracing and metrics usage.


TommyCpp (approved):
LGTM 👍

PR #564, opened 2021-06-06
Move unit into metrics module

jtescher:


djc (approved):
LGTM. Given the size of the code, I'd probably just stick it in `metrics/mod.rs` rather than creating a separate file for it.

jtescher:
@djc seems reasonable 👍

PR #565, opened 2021-06-07
Update trace flags to match spec

### LINKED ISSUE #559: Please document effect of `trace_flags` in `SpanContext` (must set sampled to process trace)
I was trying to do some distributed tracing using `SpanContext`. The `trace_flags` field is just documented as:

> Returns details about the trace. Unlike TraceState values, these are present in all traces. Currently, the only option is a boolean sampled flag.

This is slightly inaccurate; there are several constants defined at https://docs.rs/opentelemetry/0.14.0/opentelemetry/trace/index.html#constants . It'd be helpful to have a link to those constants.

But also, this documentation does not explain the net effect of the flags. I spent quite a long time trying to debug why I didn't get any traces out after setting a parent context based on a remote span. Finally, I discovered that a trace with `trace_flags` set to 0 would always be ignored. Setting `trace_flags` to `opentelemetry::trace::TRACE_FLAG_SAMPLED` caused the traces to be emitted again.

Please consider adding some documentation about this, to help people building distributed trace systems.

jtescher:
Remove all trace flags except `sampled` as that is the only supported flag in the current trace spec. Removes `SpanContext::is_deferred` and `SpanContex::is_debug` as they are also no longer supported.

Resolves #559


TommyCpp (commented):
Looks good! Just a small nit: I wonder if use `!TraceFlags::SAMPLED` in tests will be better than `TraceFlags::default()` as readers don't need to figure out what's the default of `TraceFlags`?

jtescher:
@TommyCpp yeah little confusing. As `TraceFlags` is a bitset `!TraceFlags::SAMPLED` is the same as `!0x01` which is `0xfe` rather than `0x00`. So what you would want would maybe be more like `TraceFlags::default() & !TraceFlags::SAMPLED`, but that's not super readable either. Looked at the other impls and it doesn't seem like others have an unsampled constant in the public API.

frigus02 (approved):

joshtriplett:
This seems like an improvement. Would you consider adding documentation on SpanContext and on TraceFlags::SAMPLED that mentions that some tracing tools will completely ignore spans that don't have the sampled flag set?

opentelemetry/src/trace/span_context.rs
```diff
@@ -371,6 +372,9 @@ impl FromStr for TraceState {
 }
 
 /// Immutable portion of a `Span` which can be serialized and propagated.
+///
+/// Spans that do not have the `sampled` flag set in their [`TraceFlags`] will
+/// be ignored by most tracing tools.
```
joshtriplett, line 377: Perfect, thanks!

jtescher:
@TommyCpp I can think more about test clarity and do a follow up PR if a better solution emerges there.

PR #567, opened 2021-06-08
fix: TraceState cannot insert new key-value pairs.

### LINKED ISSUE #566: Can't not add a new key/value pair to `TraceState`
According to [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracestate):
- Add a new key/value pair

But the insert function of `TraceState::insert` can't not add new pair.

The error occurred at
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/trace/span_context.rs#L236

TommyCpp:
fix #566

### DIFF
--- a/opentelemetry-aws/src/lib.rs
+++ b/opentelemetry-aws/src/lib.rs
@@ -37,8 +37,11 @@
 #[cfg(feature = "trace")]
 pub mod trace {
     use opentelemetry::{
+        global::{self, Error},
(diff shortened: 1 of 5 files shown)


TommyCpp:
@jtescher While I am working on it. I also added a new Error type for `TraceStateError`. Will try to eliminate other places where the returned error type are `()` in the future PRs.

jtescher (approved):
LGTM

PR #572, opened 2021-06-13
Prepare for v0.15.0 release

jtescher:


TommyCpp (approved):
LGTM 🎉

TommyCpp (commented):
One thing that we may want to change is the current OTLP metric exporter is broken in some cases because the collector hasn't updated to the latest version. 
> For the record, the proto submodule has to be checked out to commit 8ab21ed9a to have it working.

Maybe we should downgrade first to match the collector?

jtescher:
@TommyCpp should we merge this now? could be nice to have a little more time to consider #571

TommyCpp:
Yeah, I think we can merge it 👍 . I am still looking at options to lock the commit of submodule but don't think I can get it done tonight. And I also think we should give #571 more time to consider all possibilities,

Issue #384, opened 2020-12-14
Define default for service.name


jtescher:
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/1269

denismaxim0v:
Should be closed?

ymtdzzz:
seems to be already defined?
https://github.com/open-telemetry/opentelemetry-rust/blob/879d6ff9d251db158da7a5720ff903fa6cb7c1e2/opentelemetry-sdk/src/resource/env.rs#L79-L97

ivanitskiy:
I just tried a few things with [mini-redis](https://github.com/tokio-rs/mini-redis/blob/069a8e5ee0de445ad3e642548d196ef6b4fa8bc1/src/bin/server.rs#L71). It seems to me it doesn't use `CARGO_BIN_NAME`.

steps:
1. using this docker `docker run -d --rm --name jaeger -e COLLECTOR_OTLP_ENABLED=true -p 16686:16686 -p 4317:4317 -p 4318:4318 jaegertracing/all-in-one:latest`
2. start mini-redis server `RUST_LOG=debug cargo run --bin mini-redis-server --features otel`
3. set a value `RUST_LOG=debug cargo run --bin mini-redis-cli --features otel -- set foo bar `
4. then validate in Jaeger traces:
<img width="1253" alt="image" src="https://github.com/open-telemetry/opentelemetry-rust/assets/4936227/112d49c2-a560-4e70-a624-39ee614f1710">
5. it does use env variable  **OTEL_SERVICE_NAME** `OTEL_SERVICE_NAME=whatever RUST_LOG=debug cargo run --bin mini-redis-server --features otel`
<img width="1645" alt="image" src="https://github.com/open-telemetry/opentelemetry-rust/assets/4936227/a46711d9-e676-46e7-9d0c-5be5ead16bcf">

however when i change mini-redis code to use `CARGO_BIN_NAME` via  `with_resource()`:
```git diff
diff --git a/src/bin/server.rs b/src/bin/server.rs
index 701a1f1..ad4df39 100644
--- a/src/bin/server.rs
+++ b/src/bin/server.rs
@@ -12,12 +12,12 @@ use clap::Parser;
 use tokio::net::TcpListener;
 use tokio::signal;
 
-#[cfg(feature = "otel")]
-// To be able to set the XrayPropagator
-use opentelemetry::global;
 #[cfg(feature = "otel")]
[...]

PR #579, opened 2021-06-22
small performance boost for `Resource.get`

Frefreak:
Since `Resource`'s attrs is already a btreemap, no need to iterate to lookup.
Also here is another question regarding #384: it seems this has been partially resolved by recent PRs which add resource_detectors. However if the user provided a `trace_config` created with `with_resource` which doesn't contains a `service.name` entry, the final span will still don't have a `service.name` resource attribute. Does this behavior expected and conform to the spec?

### DIFF
--- a/opentelemetry/src/sdk/resource/mod.rs
+++ b/opentelemetry/src/sdk/resource/mod.rs
@@ -138,7 +138,7 @@ impl Resource {
 
     /// Retrieve the value from resource associate with given key.
     pub fn get(&self, key: Key) -> Option<Value> {
-        self.attrs.get(&key).map(|x| x.clone())
+        self.attrs.get(&key).cloned()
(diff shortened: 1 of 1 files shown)


djc (approved):

TommyCpp (approved):
Good catch! To your question, I believe it is expected behavior. The spec wrote 
> This resource MUST be associated with a TracerProvider or MeterProvider if another resource was not explicitly specified.

Sometimes users may not want to annotate service name in application. For example, this could happen in a collector running as sidecar and the collector can pull related information of this application from service discovery or other configurations.

We also discovered that compared with spans without any resource, adding a resource with default key values could cause ~15%-20% performance overhead so we want to keep the option open to applications which are sensitive to the performance.

PR #580, opened 2021-06-23
fix(datadog): status_code must be 0 or 1

fasterthanlime:
opentelemetry's `StatusCode` enum can be 0 (Unset), 1 (Ok) or 2 (Error).

otel specification says "don't ever use Ok, just leave it unset".

So for, non-error spans, the opentelemetry-datadog exporter does the right thing: it sets the datadog trace's status code to 0.

https://github.com/open-telemetry/opentelemetry-rust/blob/269c8004ff726c9af8327c40c66448a4517b81aa/opentelemetry-datadog/src/exporter/model/v05.rs#L114

However, for errors, the opentelemetry-datadog exporter sets the status_code to 2.

This isn't explicitly discouraged by the available Datadog docs, but it causes something really interesting: the trace
shows up with a "red" left border in Datadog APM (instead of a "green" one). The "Ok/Error" filter filters those traces correctly. However, half the UI is missing: when opening the trace, there's no '!' in the left-hand corner, there's no "Errors (1)" tab, and there's no detail of the error being shown in a user-friendly way.

**tl;dr In the Datadog backend, there's some code that does "status_code != 0", and some code that does "status_code == 1". The 2 value isn't set by their Go exporter, and so the opentelemetry-rust exporter should never use it.**

### DIFF
--- a/opentelemetry-datadog/src/exporter/model/mod.rs
+++ b/opentelemetry-datadog/src/exporter/model/mod.rs
@@ -127,7 +127,7 @@ pub(crate) mod tests {
         let traces = get_traces();
         let encoded = base64::encode(ApiVersion::Version03.encode("service_name", traces)?);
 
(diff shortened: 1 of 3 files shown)


fasterthanlime:
(This was tested with the `v05` format, but I applied the fix to the `v03` format as well, because I haven't seen the error being set to 2 _anywhere_ in Datadog's own Go exporter).

jtescher (approved):
Looks good, thanks @fasterthanlime

fasterthanlime:
Thanks for merging this! (Can you ping me when a new opentelemetry-datadog version is cut?)

A bit of post-merge confirmation: the official "opentelemetry-collector datadog exporter" also only ever uses 0 and 1: https://twitter.com/Ericmustin/status/1407764010083233796

jtescher:
@fasterthanlime new patch version [0.3.1](https://crates.io/crates/opentelemetry-datadog/0.3.1) released.

PR #592, opened 2021-07-03
feat: add hash for span context & allow spans to clone and export current state at any time

TommyCpp:
Trying to break my work on zPages into smaller commits so it's easier to review.

We need hash for span context because zPages implementation needs it to move spans around between running, error, and finished buckets.

Adding an `export_data()`  function helps the zPages implementation to sample the running spans before they finish.

### DIFF
--- a/opentelemetry/src/sdk/trace/span.rs
+++ b/opentelemetry/src/sdk/trace/span.rs
@@ -70,6 +70,21 @@ impl Span {
     {
         self.data.as_mut().map(f)
     }
+
+    /// Convert information in this span into `exporter::trace::SpanData`.
+    /// This function copies all data from the current span, which will create a
+    /// overhead.
+    pub fn exported_data(&self) -> Option<crate::sdk::export::trace::SpanData> {
+        let (span_context, tracer) = (self.span_context.clone(), &self.tracer);
+        let resource = if let Some(provider) = self.tracer.provider() {
+            provider.config().resource.clone()
+        } else {
+            None
+        };
+        self.data
+            .as_ref()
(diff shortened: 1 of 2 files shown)


opentelemetry/src/sdk/trace/span.rs
```diff
@@ -70,6 +70,21 @@ impl Span {
     {
         self.data.as_mut().map(f)
     }
+
+    /// Convert information in this span into `exporter::trace::SpanData`.
+    /// This function copies all data from the current span, which will create a
+    /// overhead.
+    pub fn exported_data(&self) -> Option<crate::sdk::export::trace::SpanData> {
```
jtescher, line 77: Hm I wonder if a kind of `ReadableSpan` makes sense here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#additional-span-interfaces

TommyCpp: I guess it makes sense. But on the other hand, `SpanData` is kind of the snapshot of current span data, which can be considered as a type of `ReadableSpan`

jtescher: Yeah probably fine. It's just in the SDK anyway.

jtescher (approved):

opentelemetry/src/sdk/trace/span.rs
TommyCpp: @jtescher I have been thinking if we could have a `ReadableSpan` somehow. I guess the problem here is we cannot store an immutable reference in span processors because it will prevent mutable borrow in the future.

PR #593, opened 2021-07-04
feat: enforce span status code's order.

### LINKED ISSUE #569: Restrict order of span status updates
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/1685

TommyCpp:
fix #569

### DIFF
--- a/opentelemetry/src/trace/span.rs
+++ b/opentelemetry/src/trace/span.rs
@@ -265,7 +265,7 @@ impl StatusCode {
 
     /// Return the priority of the status code.
     /// Status code with higher priority can override the lower priority one.
-    pub fn priority(&self) -> i32 {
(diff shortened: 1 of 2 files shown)


opentelemetry/src/trace/span.rs
```diff
@@ -260,4 +262,14 @@ impl StatusCode {
             StatusCode::Error => "ERROR",
         }
     }
+
+    /// Return the priority of the status code.
+    /// Status code with higher priority can override the lower priority one.
+    pub fn priority(&self) -> i32 {
```
jtescher, line 268: Could this method be private?

TommyCpp: Probably not private, but we can make it public within the crate.

jtescher (approved):
Looks good!

Issue #269, opened 2020-10-14
Add experimental zpages


jtescher:
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/890

TommyCpp:
I think currently zPages should contain following components:
1. a custom span processor(zPage processor I guess) to collect spans. 
2. an aggregatorto collect sample span, and record related metrics (running, error, latency buckets counts, etc.)
3. a http server to serve those data to users.

I'd image we could combine aggregator and http server and let them running in backgroud where zPage processor "export" spans through a channel share with aggregator.  If those sounds like a good start. I could dig around and work on it to see if there is any catch.

jtescher:
@TommyCpp I haven't looked to much into it, but that sounds reasonable to me

jtescher:
Completed in #595

PR #595, opened 2021-07-13
[zpage] initial implementation

TommyCpp:
related #269.

Adding a `tracez` implementation to debug spans. 

It consists of three parts: a `ZPageSpanProcessor` to install with the `TracerProvider`; a `SpanAggregator` to accumulate the sample spans; a channel to send query requests to `SpanAggregator`. 

We probably can extract the protobuf related transform into a separate crate(`opentelemetry-proto` maybe?) to be shared between `opentelemetry-otlp` and `opentelemetry-zpage`. 

Another limitation is we cannot record the status of the running spans as it requires we keep an immutable reference to the span.  I don't think it's possible without introducing a lock inside the span. #602 

Note to reviewers:
- The `transform.rs` contains duplicate code from `opentelemetry-otlp` to transform between protobuf generated types and internal types.


jtescher (approved):
This looks like a great start! Should we add a note about the experimental status in the readme? seems to be suggested in https://github.com/open-telemetry/opentelemetry-specification/tree/main/experimental

PR #599, opened 2021-07-20
Set client-to-agent UDP comm based on runtime

### LINKED ISSUE #598: opentelemetry-jaeger: tokio and async-std features are mutually exclusive
The implementation of `AgentAsyncClientUdp` depends on the features enabled/disabled.

If both tokio and async-std features are enabled then the tokio implementation is used.

This is problematic when having a workspace with one application A running on the tokio runtime and another application B running on the async-std runtime and both dependent to opentelemetry-jaeger. 
Application B will panic at runtime because the agent expect to run on tokio.

A suggestion to fix this would be to have one implementation of  `AgentAsyncClientUdp` per runtime and allow to pass the runtime to the `init_exporter` and `init_uploader` functions that would choose the correct agent implementation.

LehMaxence:
Fix #598 

There are a few changes to opentelemetry-jaeger (including breaking ones):

1.  Added `rt-tokio`, `rt-tokio-current-thread`, and `rt-async-std` features that will enable the corresponding features in `opentelemetry`. These features are additive, meaning one can enable all of them if needed. These features should be preferred rather than just enabling `tokio` or `async-std`
2. `JaegerTraceRuntime` is an extension to `TraceRuntime` that provides a UDP socket and a way to communicate over it. There's an implementation for `tokio` and `async-std` that are enabled according to the enabled features discussed in 1
3. Exporter and uploader initializers need to know on which runtime they will be used. Simple exporter and uploader initializers were implemented for a synchronous comm between client and agent.  
4. Uploaders and agents have new implementations. There's a `SimpleUploader` (should be `SimpleBatchUploader`?) which uses a synchronous udp agent `AgentSyncClientUdp` and a generic `BatchUploader<R: JaegerTraceRuntime>`  that uses an asynchronous udp agent `AgentAsyncClientUdp<R: JaegerTraceRuntime>`.

### DIFF
--- a/opentelemetry-jaeger/src/exporter/mod.rs
+++ b/opentelemetry-jaeger/src/exporter/mod.rs
@@ -37,7 +37,7 @@ use std::{
     net,
     time::{Duration, SystemTime},
 };
-use uploader::{BatchUploader, SimpleUploader, Uploader};
+use uploader::{AsyncUploader, SyncUploader, Uploader};
 
 #[cfg(all(
     any(
@@ -333,7 +333,7 @@ impl PipelineBuilder {
     /// Build a configured `sdk::trace::TracerProvider` with a simple span processor.
     pub fn build_simple(mut self) -> Result<sdk::trace::TracerProvider, TraceError> {
         let (config, process) = self.build_config_and_process();
-        let exporter = self.init_simple_exporter_with_process(process)?;
+        let exporter = self.init_sync_exporter_with_process(process)?;
         let mut builder = sdk::trace::TracerProvider::builder().with_simple_exporter(exporter);
         builder = builder.with_config(config);
 
@@ -347,7 +347,7 @@ impl PipelineBuilder {
         runtime: R,
     ) -> Result<sdk::trace::TracerProvider, TraceError> {
         let (config, process) = self.build_config_and_process();
-        let exporter = self.init_exporter_with_process(process, runtime.clone())?;
+        let exporter = self.init_async_exporter_with_process(process, runtime.clone())?;
         let mut builder =
             sdk::trace::TracerProvider::builder().with_batch_exporter(exporter, runtime);
         builder = builder.with_config(config);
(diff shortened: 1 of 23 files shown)


jtescher (approved):
This looks great, nice to be more consistent with the core crate's flags

jtescher:
@LehMaxence just need to sign the CLA

LehMaxence:
> @LehMaxence just need to sign the CLA

The corporate EasyCLA workflow seems broken at the moment. I submitted a support request with EasyCLA to get this sorted out.

opentelemetry-jaeger/src/exporter/mod.rs
```diff
@@ -341,30 +342,56 @@ impl PipelineBuilder {
...
+            export_instrumentation_lib,
+            uploader,
+        })
+    }
+
     /// Initialize a new exporter.
     ///
     /// This is useful if you are manually constructing a pipeline.
-    pub fn init_exporter(mut self) -> Result<Exporter, TraceError> {
+    pub fn init_exporter<R: JaegerTraceRuntime>(
```
TommyCpp, line 380: Maybe `init_batch_exporter` is a better name here? Just so it distinguishes itself from the simple exporter. Same with `init_uploader`, maybe we should use `init_batch_uploader`?

LehMaxence: I agree the naming (and documentation) is probably not good enough.
In practice, the simple exporter/uploader is also a batch exporter/uploader. The only difference is that it makes blocking calls to emit these batches over UDP. I didn't like calling them blocking exporter/uploader so I settled for simple instead.

Let me know what you think. I can rename the asynchronous exporter/uploader init functions `init_batch_exporter ` and `init_batch_uploader` if you think that's ok.

TommyCpp: Right. Maybe `SyncUploader` and `AsyncUploader` would be more suitable for this case? Similarly, we can have `init_async_exporter` and `init_sync_exporter`.

TommyCpp (approved):
Looks good. Thanks a lot!

asaaki:
Hi there,

a tiny note on the release:
The feature flag renaming was mentioned here in this PR, but sadly in none of the CHANGELOGs of the crates.
Caused me a bit of confusion while upgrading dependencies and getting weird compile errors.
Not a big deal (for me at least, figured it out pretty quickly), but would be nice if that could be added retroactively to the changelogs for other folks maybe.

PR #600, opened 2021-07-23
Make SpanRef public

ramosbugs:
#505 improved local span performance by moving the mutex into the
`Context` (via `SynchronizedSpan`) and introduced a `SpanRef` type
returned by `TraceContextExt::span()`. However, `SpanRef` is not currently
exported via the public API, which precludes passing it around easily within
client code. This PR exports that type alongside `TraceContextExt`.


TommyCpp (approved):
Good catch! Thanks

PR #601, opened 2021-07-23
Make SpanProcessor::on_start take a mutable span

ramosbugs:
As a side effect of #505, `SpanProcessor::on_start` callbacks are no
longer able to modify the span. The [OpenTelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart) states that the
span parameter is a [read/write span object](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#additional-span-interfaces). This PR introduces a breaking
change to restore adherence to the spec.

### DIFF
--- a/opentelemetry/src/sdk/trace/provider.rs
+++ b/opentelemetry/src/sdk/trace/provider.rs
@@ -182,7 +182,7 @@ mod tests {
     }
 
     impl SpanProcessor for TestSpanProcessor {
-        fn on_start(&self, _span: &Span, _cx: &Context) {
+        fn on_start(&self, _span: &mut Span, _cx: &Context) {
             unimplemented!()
         }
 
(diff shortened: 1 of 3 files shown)


TommyCpp (approved):
👍  Looks good! Thanks. This API may need further changes to allow implementation to keep a reference to the span. As stated in spec,
> It SHOULD be possible to keep a reference to this span object and updates to the span SHOULD be reflected in it

jtescher:
@ramosbugs looks great thanks, @TommyCpp yeah looks like the spec wants that, I think it's fine for now as that's just a SHOULD at this point.

ramosbugs:
thank you both for the quick merge!

PR #603, opened 2021-07-25
fix(otlp): fix error messages.

TommyCpp:
* Use error messages instead of debug message in the global error handler.
* Add more readable error messages for tonic status errors.


jtescher (approved):
LGTM

Issue #557, opened 2021-05-21
Need for a Viable Security Scanning Tool 


xyira:
A security workflow does not currently exist as part of the CI/CD or security workflows in the Rust repository. This is necessary to satisfy the security requirements of https://github.com/open-telemetry/opentelemetry-specification/issues/1333.

CodeQL is the common security scanning tool that is used currently for all of the supported OpenTelemetry language repositories (Go, Java, Javascript, Python, C++, etc.). Unfortunately CodeQL does not offer support for Rust. Thus an alternative would need to be found. 

We evaluated different possibilities but there seems to be limited available code scanning tools for Rust that could be integrated with Github Actions. The following code scanning tools for Rust could work:

* [Rust Audit Check](https://github.com/actions-rs/audit-check). This tool however only checks for reported security vulnerabilities in dependencies

We’d like to better understand the opinions from Rust engineers on a recommended code scanning tool so that one may be added for https://github.com/open-telemetry/oteps/issues/144. 

cc @KKelvinLo @alolita

djc:
I like to use cargo-deny, which will check for vulnerable dependencies but can also help checking for inappropriate licenses and duplicate dependency versions.

TommyCpp:
Should we also consider adding a license check? I ran a quick check with `cargo deny`. The only crate that seems to be a problem is `ring`. It contains a custom LICENSE file. The template in `cargo deny` describes `ring`'s LICENSE as `MIT AND ISC AND OpenSSL`. I suppose this is compatible with our repo's license?

PR #604, opened 2021-07-25
feat: add cargo deny

TommyCpp:
related to #557 

- Add CI pipeline to check Rust security advisory.
- Fix the security advisory in the repo.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,129 +1,15 @@
-# This template contains all of the possible sections and their default values
-
-# Note that all fields that take a lint level have these possible values:
-# * deny - An error will be produced and the check will fail
-# * warn - A warning will be produced, but the check will not fail
-# * allow - No warning or error will be produced, though in some cases a note
-# will be
-
-# The values provided in this template are the default values that will be used
-# when any section or field is not specified in your own configuration
-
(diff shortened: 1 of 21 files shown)


deny.toml
```diff
@@ -0,0 +1,198 @@
+# This template contains all of the possible sections and their default values
```
djc, line 1: I'd prefer having a minimal config with only the options that are non-default.

examples/actix-http-tracing/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "actix-http-tracing"
 version = "0.1.0"
 edition = "2018"
 publish = false
+license = "Apache-2.0"
```
djc, line 6: Instead of this, I usually set `publish = false` in `Cargo.toml` and then enable ignoring private crates in `deny.toml`. But maybe it's useful to explicitly license the examples in this case?

TommyCpp: I guess that depends on the policy of opentelemetry org. But we should add `publish = false` regardless

djc (approved):

TommyCpp:
close #557

PR #605, opened 2021-07-29
feat: add method`from_env` to prometheus exporter builder

### LINKED ISSUE #242: Add prometheus exporter env vars
See https://github.com/open-telemetry/opentelemetry-specification/pull/1021 for details.

### LINKED ISSUE #293: prometheus: Add optional feature to expose metrics over http
Metrics should be able to be exposed by non-web applications. In order to stay executor agnostic, this feature should likely include a trait that can be implemented (and may include default impls) for common tokio and async-std web servers.

LanceEa:
Fixes #242 

Allows configuring the `PrometheusExporter` using environment variables defined in the semantic conventions of the Otel specification. Currently, supports setting the host and port for prometheus in the builder and will fall back to the defaults defined in the otel specification.

### DIFF
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -93,13 +93,25 @@ const DEFAULT_CACHE_PERIOD: Duration = Duration::from_secs(0);
 
 const EXPORT_KIND_SELECTOR: ExportKindSelector = ExportKindSelector::Cumulative;
 
+/// Default host used by the Prometheus Exporter when env variable not found
+const DEFAULT_EXPORTER_HOST: &str = "0.0.0.0";
+
+/// Default port used by the Prometheus Exporter when env variable not found
+const DEFAULT_EXPORTER_PORT: &str = "9464";
+
+/// The hostname for the Promtheus Exporter
+const ENV_EXPORTER_HOST: &str = "OTEL_EXPORTER_PROMETHEUS_HOST";
+
+/// The port for the Prometheus Exporter
+const ENV_EXPORTER_PORT: &str = "OTEL_EXPORTER_PROMETHEUS_PORT";
+
 /// Create a new prometheus exporter builder.
 pub fn exporter() -> ExporterBuilder {
     ExporterBuilder::default()
 }
 
 /// Configuration for the prometheus exporter.
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct ExporterBuilder {
     /// The OpenTelemetry `Resource` associated with all Meters
     /// created by the pull controller.
@@ -135,30 +147,21 @@ pub struct ExporterBuilder {
     port: Option<String>,
 }
 
-impl ExporterBuilder {
-    /// Set builder fields for host and port from the OS environment variables
-    ///
-    /// Fields set by developer takes precedent and the OS environment variables will be ignored
-    pub fn from_env() -> Self {
-        let mut builder = ExporterBuilder::default();
-
-        if let Some(host) = env::var("OTEL_EXPORTER_PROMETHEUS_HOST")
-            .ok()
-            .filter(|s| !s.is_empty())
-        {
-            builder = builder.with_host(host)
+impl Default for ExporterBuilder {
+    fn default() -> Self {
+        ExporterBuilder {
+            resource: None,
+            cache_period: None,
+            default_histogram_boundaries: None,
+            default_summary_quantiles: None,
+            registry: None,
+            host: env::var(ENV_EXPORTER_HOST).ok().filter(|s| !s.is_empty()),
+            port: env::var(ENV_EXPORTER_PORT).ok().filter(|s| !s.is_empty()),
         }
-
-        if let Some(port) = env::var("OTEL_EXPORTER_PROMETHEUS_PORT")
-            .ok()
-            .filter(|s| !s.is_empty())
-        {
-            builder = builder.with_port(port);
-        }
-
-        builder
     }
+}
 
+impl ExporterBuilder {
     /// Set the resource to be associated with all `Meter`s for this exporter
     pub fn with_resource(self, resource: Resource) -> Self {
         ExporterBuilder {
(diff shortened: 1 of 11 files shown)


LanceEa:
This is my first attempt at addressing #242. I'm new to Rust so any suggestions on better approaches are much appreciated. Also, hopefully this aligns with the direction this issues was looking to address.

I have switched it to Draft because I need to get the CLA addressed later tonight when the kids are off to bed :).

TommyCpp:
I think the CLA issue is because you didn't link your email in commit to your GitHub account.

opentelemetry-prometheus/src/lib.rs
```diff
@@ -227,6 +284,8 @@ impl PrometheusExporter {
         controller: PullController,
         default_summary_quantiles: Vec<f64>,
         default_histogram_boundaries: Vec<f64>,
+        host: String,
+        port: String,
```
TommyCpp, line 288: 6 parameters seem a lot in this case. Should we consider deprecate this method in favor of the builder path?

TommyCpp (commented):
Looks like a good start, thanks for the effort! 

It may be worth updating the `hyper-prometheus` example to demonstrate users can use `host` and `port` information to start the webserver on the desired port. 

The next step would be to try to solve #293 and use the host/port configuration in the default web server

opentelemetry-prometheus/tests/integration_test.rs
```diff
@@ -159,3 +161,31 @@ fn compare_export(exporter: &PrometheusExporter, mut expected: Vec<&'static str>
...
+    assert_eq!(exporter.host(), "prometheus-test");
+    assert_eq!(exporter.port(), "9000");
+
+    // environment variables are available and empty
+    env::set_var(otel_exporter_prometheus_host, "");
+    env::set_var(otel_exporter_prometheus_port, "");
+    let exporter = opentelemetry_prometheus::ExporterBuilder::from_env().init();
+    assert_eq!(exporter.host(), "0.0.0.0");
+    assert_eq!(exporter.port(), "9464");
+}
```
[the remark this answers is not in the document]
LanceEa: I had it there at first and then moved it into the integration test to stay consistent. I will move it back.

opentelemetry-prometheus/src/lib.rs
```diff
@@ -122,9 +123,42 @@ pub struct ExporterBuilder {
...
+        if let Some(port) = env::var("OTEL_EXPORTER_PROMETHEUS_PORT")
+            .ok()
+            .filter(|s| !s.is_empty())
+        {
+            builder = builder.with_port(port);
+        }
+
+        builder
+    }
+
```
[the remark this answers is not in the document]
LanceEa: Ahh...I didn't see that and was going off from the old comment in the issue. I will go ahead and implement it in the default. 

Also, It looks like the `OTEL_RESOURCE_ATTRIBUTES` was added to the Default implementation of Resource so I don't think I need to do anything special for the prometheus exporter.

opentelemetry-prometheus/src/lib.rs
LanceEa: I agree that it felt a little long...the builder pattern feels more consistent with the other exporters.  How does something like this sound?

```rust
#[deprecated(
    since = "v0.9.0",
    note = "Please use the ExporterBuilder to initialize a PrometheusExporter"
)]
```

LanceEa:
Thanks for the review. I will address the feedback tomorrow.

> I think the CLA issue is because you didn't link your email in commit to your GitHub account.

Yes, thanks I need to get me correct email linked.



> It may be worth updating the `hyper-prometheus` example to demonstrate users can use `host` and `port` information to start the webserver on the desired port.

I will take a look at this too and see if I can get it updated.

LanceEa:
> It may be worth updating the `hyper-prometheus` example to demonstrate users can use `host` and `port` information to start the webserver on the desired port.

It looks like a standard prometheus pull based scenario where hyper is setup with an endpoint for scraping. This to me aligns with case 2 in the issue where it is a middleware in a web framework.  In that scenario, I don't think these would get used but rather they would be used for either case 1 (push) or case 3 (non-existing web server). Granted since it is part of the public api of `PrometheusExplorer` it could be used but I guess it doesn't feel like the intended use. The spec doesn't provide any specifics about when to use so I didn't see any guidance there. 

So, I think updating or including a new example once #293 lands is probably more appropriate, thoughts?

LanceEa:
I have addressed feedback and left comments. I also signed the CLA and got my user-id added to the new commit. It looks like the initial commit is still not associated to my user id. I can squash the commit and clean-up the commit message but I didn't want to confuse the review process. Let me know how you want to proceed!

Cheers,
Lance

TommyCpp:
Yeah I think you need to do a force push or something like that. The changes looks good. Will take a closer look on them and the comments later.

TommyCpp:
> So, I think updating or including a new example once #293 lands is probably more appropriate, thoughts?

Sounds reasonable to me 👍

LanceEa:
Alright, I amended the commit with the missing author. Looks like all the CLA stuff is good to go now. I have updated to rust 1.54, fixed all the new clippy lint issues and rebased so CI should be green now. 

I have left all the individual commits in so that I didn't mess up the PR review history. Does this project squash and cleanup the commit messages before merging into master or would you rather I squash them?

opentelemetry-prometheus/src/lib.rs
```diff
@@ -122,6 +135,30 @@ pub struct ExporterBuilder {
...
+    /// The host used by the prometheus exporter
+    ///
+    /// If not set it will be defaulted to all addresses "0.0.0.0"
+    host: Option<String>,
+
+    /// The port used by the prometheus exporter
+    ///
+    /// If not set it will be defaulted to port "9464"
+    port: Option<String>,
```
TommyCpp, line 147: I wonder if it will make sense to use an unsigned int here?

TommyCpp (commented):
Awesome work! We do squash and merge on PRs but feel free to squash it now if you want to.

Just one last suggestion and we should be good to go.

opentelemetry-prometheus/src/lib.rs
LanceEa: Yeah, I think that makes sense. I did a quick look around to see how the various frameworks and standard library handle it.

`std::net::SocketAddr` treats port as `u16` - https://doc.rust-lang.org/std/net/enum.SocketAddr.html#method.port
`std::net::ToSocketAddr` trait also shows most examples treating it as `u16` - https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html

If I look at a few Web Frameworks they are `u16` becuase they ultimately accept the SocketAddr.
- https://actix.rs/ - leverages a tuple with the port treated like a u16 for conversion to SocketAdd
- https://rocket.rs/v0.5-rc/guide/configuration/#overview
- https://github.com/hyperium/tonic/blob/master/tonic/src/transport/server/mod.rs#L622
- https://docs.rs/hyper/0.14.11/hyper/server/index.html

I can make the switch but I guess the only question is that when parsing the env variable in the impl for `default` how would we handle the potential parse error? I guess it would just get ignored and we fall back to default port? The only problem with that is we are not giving the user any feedback that it occurred so would panic early be a better dev experience?

TommyCpp: That's a good question. I think a good approach is to fall back to default one and print a error message using `global::handle_error` function

LanceEa: Cool, I have gone ahead and made the change and logged an error message.

LanceEa:
I'm not sure what happened with the CLA? I haven't changed anything and I looked and all the commits are associated to the correct email address.

TommyCpp:
Oh, I think CLA is down for now. Gonna have to wait for it to come back.

TommyCpp (approved):
LGTM, good work!

PR #608, opened 2021-07-31
Fix latest clippy lints

jtescher:


TommyCpp:
I think #605 also covers the fixes. Should we just merge that one as it seems to be ready?

jtescher:
@TommyCpp fine by me, can merge that one and close this if it's redundant

Issue #613, opened 2021-08-10
precommit.sh fail on main branch due to recent protobuf version released


stefvic:
```
running 1 test
test build_grpc ... FAILED

failures:

---- build_grpc stdout ----
thread 'build_grpc' panicked at 'generated file has changed, please commit the change file and rerun the test', opentelemetry-otlp/tests/grpc_build.rs:36:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    build_grpc

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.19s

error: test failed, to rerun pass '--test grpc_build'
```

PR #614, opened 2021-08-10
Fix grpc protobuf 2.25.0 version

stefvic:
Fix https://github.com/open-telemetry/opentelemetry-rust/issues/613


.gitignore
```diff
@@ -2,3 +2,4 @@
 */target/
 **/*.rs.bk
 Cargo.lock
+.idea
```
TommyCpp, line 5: We generally encourage ppl to ignore folders like this in the global git configuration.

TommyCpp (changes requested):
Looks good overall. Thanks! Just a small nit.

.gitignore
stefvic: > We generally encourage ppl to ignore folders like this in the global git configuration.

indeed

TommyCpp (approved):

Issue #545, opened 2021-05-09
Allow option to split oversiezd jaeger UDP batches


jtescher:
Currently the Jaeger exporter will simply error if the batch size is too large. It would be better to have an option to split and export multiple sub-batches instead.

DCjanus:
From [this line](https://github.com/open-telemetry/opentelemetry-rust/blob/ae4aeb364cd7054fe43a6fee8d340b6959624f3c/opentelemetry-jaeger/src/exporter/agent.rs#L71), looks like the only way to known actual size of payload, is actual write batch to buffer.

That means, to split oversized batches, maybe we have to code like this:

```
FUNCTION EMIT( BATCH ):
    WRITE ( BUFFER )
    IF LEN ( BUFFER ) <= SIZE_LIMIT:
        WRITE_TO_SOCKET ( BUFFER )
        RETURN
    HALF_A, HALF_B = SPLIT( BATCH )
    EMIT ( HALF_A )
    EMIT ( HALF_B )
```

I'm not sure is this ok

jtescher:
@DCjanus yes that looks like the right approach

PR #619, opened 2021-08-18
feat: try split batch if payload size larger than max_package_size

DCjanus:
Close:  #545


opentelemetry-jaeger/src/exporter/agent.rs
```diff
@@ -120,30 +120,74 @@ impl<R: JaegerTraceRuntime> AgentAsyncClientUdp<R> {
...
-        self.runtime.write_to_socket(&self.conn, payload).await?;
-
         Ok(())
     }
 }
+
+fn split_batch(
+    client: &mut BufferClient,
+    mut batch: jaeger::Batch,
+    auto_split: bool,
```
TommyCpp, line 149: It feels wired to call the `split_batch` function if you don't want to split at all(auto split is false). It may be better to check it before calling the `split_batch` function. If the auto split is `false`. We could just error out if the payload is too large.

TommyCpp (commented):
Overall looks like a good solution. Thanks a lot. Could you add some unit tests around it?

opentelemetry-jaeger/src/exporter/agent.rs
DCjanus: There might be more dump code out of `split_batch`, that's why I combine auto split check in this function.

TommyCpp: That's true. But if the users set the `auto_split` to `false`, we should avoid batch clone in L153 to reduce the overhead

DCjanus: > Overall looks like a good solution. Thanks a lot. Could you add some unit tests around it?

To be honest, I don't know how to test this part code, is there any example for me?

DCjanus: > we should avoid batch clone in L153 to reduce the overhead

I've updated some code, reduced overhead if `auto_split` is false.

TommyCpp: > To be honest, I don't know how to test this part code, is there any example for me?

Hmm, I think you are right, it's tricky to simulate the `emit_batch` function here. I guess we could merge it without test and revisit it when we figure out how to test the jaeger exporter.

opentelemetry-jaeger/src/exporter/mod.rs
```diff
@@ -245,6 +247,12 @@ impl PipelineBuilder {
         self
     }
 
+    /// Config whether to auto split batch if packet size is too large.
```
TommyCpp, line 250: Small nit. Can we add some notes on how it works? For example, if one span is too large, will other smaller spans in the same batch being exported. Thanks in advance.

DCjanus: I'm not native speaker, and there might be some grammatical errors, feel free to push commit to my branch.

TommyCpp: Sorry, that was supposed to be a question. From what I understand, if one span is too large. Other spans may or may not get exported, depends on the index of the large span. Any span that comes after the large span will be discarded as we will error out on the large span.

TommyCpp (approved):
LGTM. Awesome work 🎉

PR #622, opened 2021-08-27
Implement Serialize & Deserialize for Sampler

cecton:
Hello :wave:

I noticed that a few types like Resource and KeyValue implement `Serialize` and
`Deserialize` which is very handy for me because I can integrate the different
parameter into a configuration struct.

Unfortunately it is not yet implemented for `Sampler` so I had to implement it
on my crate but I think this could be beneficial for others.

Let me know what you think

### DIFF
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -42,7 +42,6 @@ js-sys = "0.3"
 [dev-dependencies]
 bincode = "1.2"
 criterion = "0.3.1"
-serde_json = "1.0.66"
 rand_distr = "0.4.0"
 
 [features]
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -42,6 +42,7 @@ js-sys = "0.3"
 [dev-dependencies]
 bincode = "1.2"
 criterion = "0.3.1"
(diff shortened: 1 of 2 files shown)


TommyCpp (commented):
👍 Overall looks good. Nice job! The only thing we might want to examine is the current serialization/de-serialization process eliminated the `ParentBased` variant.  I don't think it would cause any issue but I think the serialization process shouldn't change the variant.

cecton:
omg what am I doing. Why do I even bother implementing them manually :exploding_head: 

Oh I know... I had in my own crate because I don't own the type. Then I copied here thinking I had too.

@TommyCpp this should be much easier to review. I'm not sure the test is useful at all. I suggest we delete it entirely.

TommyCpp (approved):

TommyCpp:
Looks good. Thanks

cecton:
Hey @TommyCpp I tried to lock the version on my side but it's actually very complicated. Because it's linked to tracing I have to fork tracing too and replace all the tracin dependencies in my project. Would it be possible for you to make a release?

cecton:
@TommyCpp never mind, I just tried with the patch syntax instead of pinning the version in the dependency and it worked like a charm. The release can wait :)

PR #626, opened 2021-09-01
Implement Serialize & Deserialize for more stuff

cecton:
Same than #622 but for:
- SpanLimits
- Protocol
- Credentials (also re-export to make it public)
- Compression (also re-export to make it public)

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -198,7 +198,7 @@ mod metric;
 mod span;
 mod transform;
 
-pub use crate::exporter::{Credentials, ExportConfig};
+pub use crate::exporter::ExportConfig;
 pub use crate::span::{OtlpTracePipeline, SpanExporter};
 
 #[cfg(feature = "metrics")]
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -214,12 +214,15 @@ pub use crate::exporter::{
 use opentelemetry::sdk::export::ExportError;
 
 #[cfg(feature = "grpc-sys")]
-pub use crate::exporter::grpcio::GrpcioExporterBuilder;
+pub use crate::exporter::grpcio::{Compression, Credentials, GrpcioExporterBuilder};
```
cecton, line 217: There are a few things to say here:

1. They should probably be prefixed with `Grpcio` as they are specific to that exporter
2. They were public but not accessible. There was no way to call `with_compression` or `with_credentials` because the types couldn't be instantiated. In the doc you can see it because the types are not clickable: 
image

TommyCpp (approved):
LGTM 👍 Thanks.

PR #637, opened 2021-09-26
Allow static str and string in span methods

jtescher:
Improve performance and ergonomics by allowing `&'static str` or `String` types to be passed to span methods.


opentelemetry/src/global/trace.rs
```diff
@@ -1,34 +1,177 @@
...
     /// Records events at a specific time in the context of a given `Span`.
     ///
     /// Note that the OpenTelemetry project documents certain ["standard event names and
     /// keys"](https://github.com/open-telemetry/opentelemetry-specification/tree/v0.5.0/specification/trace/semantic_conventions/README.md)
     /// which have prescribed semantic meanings.
-    fn add_event_with_timestamp(
+    fn add_event_with_timestamp<T>(
         &mut self,
-        name: String,
+        name: T,
```
djc, line 167: Style nit: I think `impl Into<Cow<'static, str>>` is a little easier to follow (less indirection).

djc (commented):
It's not obvious to me why the `GenericSpan`/`BoxedSpan` changes are in the same commit/PR. Maybe split that out as a separate commit to make it clearer what's going on there (or at least explain the change in the commit message/PR description)?

jtescher:
@djc Could separate them but the global changes are only motivated by the non object-safe span method changes, maybe `ObjectSafeSpan` would be a clearer name than `GenericSpan`? Also open to better patterns here if there are simpler ways of providing a global sdk-agnostic interface.

djc:
Yeah, `GenericSpan` seems very... well, generic. `ObjectSafeSpan` might be better, or `DynSpan` as it was before?

TommyCpp:
+1 for `ObjectSafeSpan`

PR #638, opened 2021-09-27
Add `set_attributes` method for Span

srikanthccv:
This adds the convenience  method `set_attributes` for set multiple attributes.

### DIFF
--- a/opentelemetry-api/src/trace/context.rs
+++ b/opentelemetry-api/src/trace/context.rs
@@ -129,6 +129,21 @@ impl SpanRef<'_> {
         self.with_inner_mut(move |inner| inner.set_attribute(attribute))
     }
 
+    /// Set multiple attributes of this span.
+    ///
+    /// Setting an attribute with the same key as an existing attribute
+    /// generally overwrites the existing attribute's value.
+    ///
+    /// Note that the OpenTelemetry project documents certain "[standard
+    /// attributes]" that have prescribed semantic meanings and are available via
+    /// the [opentelemetry_semantic_conventions] crate.
+    ///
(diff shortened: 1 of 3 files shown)


opentelemetry/src/global/trace.rs
```diff
@@ -51,6 +51,13 @@ impl trace::Span for BoxedSpan {
         self.0.set_attribute(attribute)
     }
 
+    /// Sets attributes of this span.
+    fn set_attributes(&mut self, attributes: Vec<KeyValue>) {
```
djc, line 55: This should probably take an `impl Iterator<Item = KeyValue>` instead (or would that endanger object safety for the `Span` trait?), so the caller isn't required to allocate a `Vec` for it. I'm not sure about the `set_` prefix, which to me implies overwriting any existing attributes. Maybe `update_attributes()`, and document the behavior wrt overwriting pre-existing attributes in the docstring?

jtescher: We should be able to have this be generic after #637

lonewolf3739: @djc I don't have enough expertise here, can you elaborate how caller could avoid the allocation for `Vec`?

djc: The caller might pass in some other iterator, for example something with `.map()` combinator that generates `KeyValue` from some other type. In that case, it would be unfortunate if they have to allocate a `Vec` just to pass their attributes to this method. (In the 2021 edition, which is coming in a few weeks, you could also pass in an array, I think.)

TommyCpp: @lonewolf3739 any chance you want to implement the above changes? Thanks in advance.

lonewolf3739: I will convert this to draft for now and make the necessary changes when I find time, thanks.

jtescher (approved):

PR #641, opened 2021-09-29
Fix default resource detection for tracer provider

lonewolf3739:
Spec [says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable)
>The SDK MUST extract information from the OTEL_RESOURCE_ATTRIBUTES environment variable and merge this, as the secondary resource, with any resource information provided by the user, i.e. the user provided resource information has higher priority.

### DIFF
--- a/opentelemetry/src/sdk/trace/provider.rs
+++ b/opentelemetry/src/sdk/trace/provider.rs
@@ -8,7 +8,7 @@
 //! propagators) are provided by the `TracerProvider`. `Tracer` instances do
 //! not duplicate this data to avoid that different `Tracer` instances
 //! of the `TracerProvider` have different versions of these data.
-use crate::sdk::resource::SdkProvidedResourceDetector;
+use crate::sdk::resource::{SdkProvidedResourceDetector, EnvResourceDetector};
 use crate::sdk::trace::runtime::TraceRuntime;
 use crate::sdk::Resource;
 use crate::trace::TraceResult;
@@ -145,7 +145,7 @@ impl Builder {
         let mut config = self.config;
         let sdk_provided_resource = Resource::from_detectors(
             Duration::from_secs(0),
-            vec![Box::new(SdkProvidedResourceDetector)],
+            vec![Box::new(EnvResourceDetector::new()), Box::new(SdkProvidedResourceDetector)],
         );
         config.resource = match config.resource {
             None => Some(Arc::new(sdk_provided_resource)),
@@ -175,6 +175,7 @@ mod tests {
     use crate::trace::{TraceError, TraceResult, TracerProvider};
     use crate::{Context, Key, KeyValue};
     use std::sync::Arc;
+    use std::env;
 
     #[derive(Debug)]
     struct TestSpanProcessor {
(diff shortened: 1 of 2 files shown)


opentelemetry/src/sdk/trace/provider.rs
```diff
@@ -145,17 +145,15 @@ impl Builder {
...
             None => Some(Arc::new(sdk_provided_resource)),
-            Some(resource) => {
-                if resource.is_empty() {
-                    None
-                } else {
-                    Some(resource)
-                }
-            }
+            // User provided resource information has higher priority.
+            Some(resource) => Some(Arc::new(sdk_provided_resource.merge(resource))),
```
TommyCpp, line 156: I think we can keep the `merge` function as it is and switch the `resource` and `sdk_provided_resource`. i. e `resource.merge(sdk_provided_resource)`?

lonewolf3739: I don't think we could do that since it wouldn't be spec complaint?

jtescher: Either way we should check the benchmarks to see if we still want the optimization for the case where both are empty.

lonewolf3739: Will both of them ever be empty? At least there will be SDK provided [default values](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#semantic-attributes-with-sdk-provided-default-value), right?

TommyCpp (approved):
Looks good to me. Thanks! really love the unit tests.

TommyCpp:
@jtescher Any more comments or it's good to merge?

opentelemetry/src/sdk/trace/provider.rs
jtescher: Currently the sdk allows users who are focused on the lowest overhead to out through specifying [with_no_resource](https://docs.rs/opentelemetry/0.16.0/opentelemetry/sdk/trace/struct.Config.html#method.with_no_resource).

jtescher:
@TommyCpp one last issue to resolve one way or the other

opentelemetry/src/sdk/trace/provider.rs
lonewolf3739: Just want to make sure I am understanding you clearly. When you say _both are empty_ What are you referring to here? I can see how `config.resource` can be empty, but what is there other one?

jtescher: Ah yeah just the `config.resource` check that was there before.

lonewolf3739: The current code vs something like code below?

```rust
if resource.is_empty() {
	Some(Arc::new(sdk_provided_resource))
} else {
	Some(Arc::new(sdk_provided_resource.merge(resource))),
}
```
You wanted to check the benchmarks when that resource provided empty, right?

jtescher: The current optimization is to have the resource be `None` in this case.

lonewolf3739: Hmm, that would not be conforming to the specification. I think it would be surprising for users to find this behaviour different in rust SDK if they were also using other language SDKs. If you prefer that way I can change it back to `None`.

jtescher: I think it would be simpler to keep the current optimization and have a subsequent issue/pr around removing it as `with_no_resource` would also need to be removed.

lonewolf3739: Sure, that sounds good to me.

PR #642, opened 2021-09-29
[OTLP] Changed tonic error to include the server's message

GrandChaman:
Hello!

The error message regarding a tonic failure can be misleading. Would it be OK to add the message sent back by the server so that a bit more information could be available when debugging those errors ?


djc (approved):
Makes sense to me, thanks!

TommyCpp (approved):
Looks good. Just need to sign the CLA. Thanks!

PR #655, opened 2021-10-17
otlp: enable directly constructing a SpanExporter

### LINKED ISSUE #640: Support building an HTTP SpanExporter
I'm trying to migrate from opentelemetry-rust 0.14 to 0.16. I'm using OTLP HTTP. In 0.14, I had this:

```rust
        opentelemetry_otlp::TraceExporter::new_http(
            opentelemetry_otlp::ExporterConfig {
                endpoint: "https://api.honeycomb.io/v1/traces".to_string(),
                protocol: opentelemetry_otlp::Protocol::HttpBinary,
                ..Default::default()
            },
            opentelemetry_otlp::HttpConfig {
                headers: Some(headers),
                ..Default::default()
            },
        )
```

But in 0.16, `HttpConfig` doesn't seem to be public anymore, so I can't create one to call `new_http`. I tried creating an `HttpExporterBuilder`, but once I'm done building, I don't see how to get from there to a `SpanExporter`, because `SpanExporterBuilder::build_span_exporter` is private.

What's the supported way to build a `SpanExporter` in 0.16?

Alternatively, a solution to https://github.com/open-telemetry/opentelemetry-rust/issues/540 (providing a function to write a `Vec<SpanData>` into the pipeline's channel) would make this unnecessary; the main reason I need a `SpanExporter` is to build my own pipeline by hand so that I have access to a channel sender to send additional spans.

jtescher:
Currently it is not possible to construct an `opentelemetry_otlp::SpanExporter` directly. This patch makes both the
builder and the `build_span_exporter` method public so users may construct one if necessary.

Fixes #640

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -199,7 +199,7 @@ mod span;
 mod transform;
 
 pub use crate::exporter::ExportConfig;
-pub use crate::span::{OtlpTracePipeline, SpanExporter};
+pub use crate::span::{OtlpTracePipeline, SpanExporter, SpanExporterBuilder};
 
 #[cfg(feature = "metrics")]
 pub use crate::metric::{MetricsExporter, OtlpMetricPipeline};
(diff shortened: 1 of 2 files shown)


jtescher:
This would allow construction through the current builder via something like 

```rust
SpanExporterBuilder::from(
        .tonic()
        .with_endpoint("http://localhost:4317"),
)
.build_span_exporter();
```

Alternatively (or in addition to) we could expose a method on the pipeline to build instead of install

```rust
    .tracing()
    .with_exporter(
            .tonic()
            .with_endpoint("http://localhost:4317"),
    )
    .build_span_exporter()
```

PR #661, opened 2021-10-31
limit futures deps

asonix:
This change removes the dependency on futures-io by depending directly on futures-core in opentelemetry-otlp and futures-util, futures-channel, and futures-executor in opentelemetry

### DIFF
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -51,6 +51,7 @@ http = "0.2"
 serde = { version = "1.0", features = ["derive"], optional = true }
 
 [dev-dependencies]
+futures-util = "0.3"
 tokio-stream = { version = "0.1", features = ["net"] }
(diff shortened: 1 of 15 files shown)


djc (approved):

djc:
If you can't tell if it did anything, it probably didn't. It should open a web page and lead you through installing a GitHub app and agreeing to the CLA. Did that not happen for you?

asonix:
@djc it connected the bot to my github account but didn't walk me through signing a CLA. it could be one of my many firefox extensions breaking it, so I'll try again after disabling some of those

TommyCpp (approved):
Looks good. Thanks!

asonix:
looks like opentelemetry-otlp cannot be changed this way, since futures is referenced directly in generated code due to this line in grpcio-compiler: https://github.com/tikv/grpc-rs/blob/master/compiler/src/codegen.rs#L595

It's a shame, since Future is even in the standard library now

PR #662, opened 2021-11-01
fix: detect service.name from OTEL_SERVICE_NAME and OTEL_RESOURCE_ATTRIBUTES

TommyCpp:
OTEL_SERVICE_NAME will take priority and if it's not available. We will try to detect the service.name from OTEL_RESOURCE_ATTRIBUTES. If it's also not available. We will use the default one.


jtescher (approved):
Looks good

PR #663, opened 2021-11-02
fix: Mapping between Jaeger processes and Otel process. 

### LINKED ISSUE #657: Jaeger: service.name is not with_service_name
Using:

```rust
opentelemetry::sdk::trace::BatchSpanProcessor::builder(
        opentelemetry_jaeger::new_pipeline()
            .with_service_name("mre-jaeger")
            .with_agent_endpoint(addr)
            .with_tags(vec![opentelemetry::KeyValue::new("exporter", "jaeger")])
            .init_async_exporter(opentelemetry::runtime::Tokio).expect("Jaeger Tokio async exporter"),
        opentelemetry::runtime::Tokio,
    )
    .build();
 ```

Jaeger does not show a `service.name` entry:

image

I understand the service name passed in is the service name all spans sit under.
It would be useful to be able to set this value within a span, perhaps to a different value.  However, I haven't been able to find any documentation on how to populate the `service.value` shown above.

### PR DESCRIPTION
**Pull request by TommyCpp**
fix #657 

The spec maps resource tags to process tags and `service.name` entry in resource to be the service name of the process.

####  `Unknown_service` was used as `service.name` process tags even though users provided another name via `with_service_name` function.

Service name serves a special purpose for Jaeger as it requires every span to have a service name. In the open telemetry model, `service.name` is just a resource, which should be provided by users or have a default value but users can override it to be empty. 

To address the difference between those two models. We need to answer the following questions. 

1. Should we store the service name within the exporter or store it in resource and extract when exporting?
2. What's the priority of different ways to set the service name?
3. Should we report the `service.name` as a process tags/resource for jaeger spans?

In this PR, we implemented the following process
1. We store the service name as part of the `process` field in the exporter
2. The priority of different methods are listed below from high to low
- `with_service_name` method in `PipelineBuilder`
- passing `service.name` resource from `with_trace_config` method in `PipelineBuilder`
- SDK provided service name(it can come from env vars or default `unknown_service`)
3. We append a `service.name` process tag for each Jaeger span

#### Duplicate process tags

Process tags can be set via `with_tags` function in Jaeger pipeline or by the resource within the trace config(`with_trace_config` function in Jaeger pipeline).

We didn't de-duplicate entries from those two functions.

For this problem, we should deprecate the `with_tags` method and asks users to use `with_trace_config` and store the process tags/resource only in one place.

We can store the process tags in either of the following places
- exporter's process tags
- trace config resources

From a performance standpoint, we should store the tags in the exporter's process tags. Jaeger clients only require a process instance for one batch. If we store the tags as resources, we will copy them for each span in the batch, which will be discarded later in Jaeger clients. 

However, storing tags in exporters may cause confusion when users install multiple exporters in the tracer provider. Other exporters could get the resource from the trace config while Jaeger exporter will use resources/tags stored in itself.

### DIFF
--- a/examples/actix-udp/src/main.rs
+++ b/examples/actix-udp/src/main.rs
@@ -7,8 +7,6 @@ use opentelemetry::{
     trace::{FutureExt, TraceContextExt, Tracer},
     Key,
 };
-use opentelemetry::sdk::Resource;
-use opentelemetry::sdk::trace::Config;
 
 fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
     opentelemetry_jaeger::new_pipeline()
--- a/examples/actix-udp/src/main.rs
+++ b/examples/actix-udp/src/main.rs
@@ -12,6 +12,17 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
     opentelemetry_jaeger::new_pipeline()
         .with_agent_endpoint("localhost:6831")
         .with_service_name("trace-udp-demo")
+        .with_tags(vec![
+            opentelemetry::KeyValue::new("exporter", "jaeger"),
+            opentelemetry::KeyValue::new("service.name", "my-service2"),
+            opentelemetry::KeyValue::new("service.namespace", "my-namespace2"),
+        ])
+        .with_trace_config(opentelemetry::sdk::trace::config().with_resource(
+            opentelemetry::sdk::Resource::new(vec![
+                opentelemetry::KeyValue::new("service.name", "my-service"),
+                opentelemetry::KeyValue::new("service.namespace", "my-namespace"),
+            ]),
+        ))
         .install_simple()
 }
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by TommyCpp** [comment]
Will add some more docs if we all agree on the methods.

File: examples/actix-udp/src/main.rs [hunk 1]
```diff
@@ -12,6 +12,17 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
     opentelemetry_jaeger::new_pipeline()
         .with_agent_endpoint("localhost:6831")
         .with_service_name("trace-udp-demo")
+        .with_tags(vec![
+            opentelemetry::KeyValue::new("exporter", "jaeger"),
+            opentelemetry::KeyValue::new("service.name", "my-service2"),
```
jtescher, line 17: may be confusing to have an example that uses both tags and a resource

examples/basic/src/main.rs
```diff
@@ -14,11 +15,16 @@ use std::time::Duration;
 
 fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
     opentelemetry_jaeger::new_pipeline()
-        .with_service_name("trace-demo")
+        // .with_service_name("trace-demo")
```
jtescher, line 18: should this still be here?

jtescher (changes requested):
Seems good, just a few example nits

jtescher (approved):

PR #664, opened 2021-11-03
Simplify trace core traits

jtescher:
This patch removes the `std::fmt::Debug` and `'static` requirements from the `TracerProvider`, `Tracer`, and `Span` traits as they are not generally necessary. `'static` is only required in the global module context.


djc (approved):
Makes sense to me, nice!

TommyCpp (approved):
Looks good!

PR #667, opened 2021-11-12
Add with_aggregator_selector for Prometheus

wpbrown:
Adds the `with_aggregator_selector` fn to prometheus `ExporterBuilder` so we can provide custom `AggregatorSelector`.

### DIFF
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -69,7 +69,9 @@ pub use prometheus::{Encoder, TextEncoder};
 
 use opentelemetry::global;
 use opentelemetry::sdk::{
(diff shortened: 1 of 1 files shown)


opentelemetry-prometheus/src/lib.rs
```diff
@@ -240,6 +246,17 @@ impl ExporterBuilder {
         }
     }
 
+    /// Set the aggregation selector for the prometheus exporter
+    pub fn with_aggregator_selector(
+        self,
+        aggregator_selector: Box<dyn AggregatorSelector + Send + Sync>,
```
TommyCpp, line 252: Maybe we can take a type parameter `T` and wrap it using `Box` inside the function. May be more ergonomics?

TommyCpp (approved):
LGTM. Thanks

PR #668, opened 2021-11-13
Add SharedString Value type

wpbrown:
Add ability to pass Arc<String> values to reduce string copies in case on of non-'static str.

### DIFF
--- a/opentelemetry-jaeger/src/exporter/thrift/mod.rs
+++ b/opentelemetry-jaeger/src/exporter/thrift/mod.rs
@@ -4,6 +4,7 @@
 use std::time::{Duration, SystemTime};
 
 use opentelemetry::trace::Event;
+use opentelemetry::util::take_or_else_clone;
 use opentelemetry::{Key, KeyValue, Value};
 
 pub(crate) mod agent;
@@ -60,6 +61,7 @@ impl From<KeyValue> for jaeger::Tag {
         let KeyValue { key, value } = kv;
         match value {
             Value::String(s) => jaeger::Tag::new(key.into(), jaeger::TagType::String, Some(s.into()), None, None, None, None),
+            Value::SharedString(s) => jaeger::Tag::new(key.into(), jaeger::TagType::String, Some(take_or_else_clone(s)), None, None, None, None),
             Value::F64(f) => jaeger::Tag::new(key.into(), jaeger::TagType::Double, None, Some(f.into()), None, None, None),
             Value::Bool(b) => jaeger::Tag::new(key.into(), jaeger::TagType::Bool, None, None, Some(b), None, None),
             Value::I64(i) => jaeger::Tag::new(key.into(), jaeger::TagType::Long, None, None, None, Some(i), None),
--- a/opentelemetry-otlp/src/transform/common.rs
+++ b/opentelemetry-otlp/src/transform/common.rs
@@ -1,4 +1,5 @@
 use opentelemetry::sdk::trace::EvictedHashMap;
+use opentelemetry::util::take_or_else_clone;
 use opentelemetry::{Array, Value};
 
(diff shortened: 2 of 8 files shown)


jtescher:
As this is a performance optimization, do you mind adding a benchmark to show the performance differences?

djc:
Should we consider replacing the existing `String` variant rather than adding another one?

wpbrown:
> As this is a performance optimization, do you mind adding a benchmark to show the performance differences?

I added a benchmark for KeyValue create+drop. It's significant for us because we don't have stored copies of KeyValue. Keys are static and known ahead of time, but KeyValue structs have to be newed for every metric emission. Some label values can be quite large too.

I think ideally there would have been owned and borrowed forms of KeyValue. In the common case where they just need to be hashed for lookup the borrowed form would be sufficient. In lieu of that restructure 

image

> Should we consider replacing the existing String variant rather than adding another one?

It's important to have a case for 'static str, which the current String variant covers via Cow. It might makes sense to change the current string variant to just StaticStr and direct any owned String in to SharedString. However I did a quick test of that.. and the overhead of cloning a String into Arc vs the Cow path is constant ~35%. That might be recouped later when the KeyValue gets cloned on first entry of the attribute in to a new attributeset... however on-going use of that time series each time the metric is updated will take the penalty. There maybe more value to having shared string always downstream in exports... I haven't looked that far.

TommyCpp (approved):
Sorry for the late review. I think overall it would be a nice improvement in thems of performance.

jtescher:
Hm are these in a tracing or metrics context? wondering if there are other paths to avoid cloning

TommyCpp:
> Hm are these in a tracing or metrics context?

I think it applies to both. Lots of label copying happens in spans and in metrics. 

Should we consider an approach similar to `Cow` for `Values` in the long term?

jtescher:
@TommyCpp the normal string value is already a `Cow<'static, str>` so this optimization seems like it's for values that are not fully static where `&'static str` could be used, but also not unique as they are being cloned, so trying to understand the use case more clearly.

wpbrown:
> Hm are these in a tracing or metrics context?

I'm currently only working with metrics, but I assume tracing will have similar challenges. 

> wondering if there are other paths to avoid cloning

> Should we consider an approach similar to Cow for Values in the long term?

Indeed KeyValue and Value will need borrowed forms that are generic over a lifetime (vs just 'static). The value would only be converted to an OwnedValue as needed. In the case of metrics where the Record exists already we have to clone the string to make the value to make the keyvalue to make the array of keyvalues. It then is hashed in acquire handle to look up the existing Record and thrown away.

Currently, unless you have a 'static str to work with, Value must own the String. And KeyValue must own the Value. For copy types this doesn't matter, but for String and Array it's costly at scale. 

BTW, I don't have a use case personally where this is a problem for Keys (mine are statically known), but for someone building a dynamic application (e.g. with user customizable metrics), Keys would have the same issue as Values.

I thought this change was reasonable for being backward compatible and providing acceptable performance in the short term, but for long term, introducing a borrowing API as the default in a new proper semver would be ideal. Even with a borrowing API, it will still make sense for Value to be aware of shared Strings and Arrays, so that when the SDK does need to take ownership of Values (and non 'static Keys) the caller has the option of letting it do so cheaply.

>  this optimization seems like it's for values that are not fully static where &'static str could be used, but also not unique as they are being cloned

Exactly. The values exist (at high order of magnitude), they are not known statically, and continue to be used in many other parts of the code base so they can't just be moved in to a Value.

jtescher:
We may need a more general solution for strings or allocations in general. I think we should re-assess this when we do a performance pass before 1.0 as many parts of the trace api currently take `Cow<'static, str>`, if we wanted to introduce changes we may want to be consistent where we accept what string-like types.

jtescher:
closing for now

PR #670, opened 2021-11-17
feat: using `Cow<'static, str>` for instrumentation library.

### LINKED ISSUE #666: global::tracer dynamic naming
I have a scenario where I would like to use a name for my Tracers that is not known at compile time.

Currently the only way to do that, due to the usage of `&'static str` for name and version, is to cause a memory leak with something like 
`
let s = String::from("some id");
let trace = global::tracer(format!("id-{}", Box::leak(s.into_boxed_str())));
`

I did try to go down the rabbit hole of switching out &'static str for Cow<'static, str> or String, but found that the InstrumentationLibrary requires the Copy trait, which neither Cow nor String implement.  Removing the Copy trait causes the downstream effect of the inability to move an instance of InstrumentationLibrary on this line (and similar):
`instrumentation_lib: *tracer.instrumentation_library(),`

Could this be switched to a clone instead of dereference/move so that the rest of the InstrumentationLibrary struct can be more friendly?

TommyCpp:
It allows users to pass in a `String` if needed, although it will have a performance overhead because we need to clone the string every time a span gets exported. I did a quick test and it seems the overhead is manageable. Less than 10%

Users that pass a `'static str` will not be impacted as `Cow::Borrow` just copies the pointer.

Fix #666

### DIFF
--- a/opentelemetry-datadog/src/exporter/model/v03.rs
+++ b/opentelemetry-datadog/src/exporter/model/v03.rs
@@ -41,7 +41,7 @@ pub(crate) fn encode(
             rmp::encode::write_str(&mut encoded, service_name)?;
 
             rmp::encode::write_str(&mut encoded, "name")?;
-            rmp::encode::write_str(&mut encoded, span.instrumentation_lib.name)?;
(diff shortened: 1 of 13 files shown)


jtescher (approved):
Looks good!

PR #671, opened 2021-11-17
Ignore the time/chrono related issue and examples in cargo deny

TommyCpp:
The excluded examples depended on tokio pre 1.0 indirectly, which has some issues. It doesn't jeopardise our libraries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ members = [
     "opentelemetry-stackdriver",
     "opentelemetry-zipkin",
     "opentelemetry-zpages",
-    "examples/actix-udp",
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -12,7 +12,6 @@ members = [
     "opentelemetry-stackdriver",
     "opentelemetry-zipkin",
     "opentelemetry-zpages",
-    "examples/actix-udp",
```
TommyCpp, line 15: This is duplicate with line 18

djc (commented):
I think this makes sense (in fact, I added exactly the same cargo-deny config in my work project yesterday).

PR #679, opened 2021-12-13
Add opentelemetry-dynatrace crate

diceride:
- Adds a new exporter for Dynatrace

    This exporter is inspired by `opentelemetry-otlp`. It attempts to follow the same design and naming conventions. This is also the case for feature flag names, for example.

    In comparison to `opentelemetry-otlp`, this exporter does not use autogenerated code from Protocol Buffers. The transportation layer is based on `opentelemetry-http` instead of gRPC. 

- Sorted all crate names lexicographically in the README.md files


TommyCpp:
Some of the lint issues were introduced in the new Clippy version, which I fixed recently. You can try to merge the master branch and clear them out.

opentelemetry-dynatrace/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+## v0.1.0
+
+### Added
+
+- Dynatrace metrics exporter #XXXXXXXXXXXXXXXXXXXXXXXXX
```
TommyCpp, line 7: nit: remove comments

opentelemetry-dynatrace/Cargo.toml
```diff
@@ -0,0 +1,87 @@
...
+wasm = [
+    "base64",
+    "futures-util",
+    "getrandom/js",
+    "js-sys",
+    "pin-project",
+    "wasm-bindgen",
+    "wasm-bindgen-futures",
+    "web-sys",
+]
```
TommyCpp, line 50: I don't think we have a `wasm http client`. Do we still need this feature in the crate?

opentelemetry-dynatrace/src/transform/metrics.rs
```diff
@@ -0,0 +1,1165 @@
...
+pub struct MetricLine {
+    kind: NumberKind,
+    key: MetricKey,
+    dimensions: Option<DimensionSet>,
+    min: Option<Number>,
+    max: Option<Number>,
+    sum: Option<Number>,
+    count: Option<u64>,
+    delta: Option<Number>,
+    gauge: Option<Number>,
```
TommyCpp, line 279: For the payload part. Maybe using an enum here is more intuitive and will reduce the size by a little bit.

opentelemetry-dynatrace/src/transform/metrics.rs
```diff
@@ -0,0 +1,1165 @@
...
+                            prev_char_underscore = false;
+                            Some(c)
+                        } else if !prev_char_underscore {
+                            prev_char_underscore = true;
+                            Some('_')
+                        } else {
+                            None
+                        }
+                    })
+                    .take(100)
```
TommyCpp, line 154: Maybe worth commenting why we limit on 100 here. I understand it's because dimension only supports up to 50 pairs?

opentelemetry-dynatrace/src/transform/metrics.rs
```diff
@@ -0,0 +1,1165 @@
...
+                self.sum.to_owned().unwrap().to_f64(&self.kind),
+                self.count.to_owned().unwrap(),
+            ))?;
+        } else if let Some(delta) = self.delta.to_owned() {
+            fmt.write_char(' ')?;
+            fmt.write_fmt(format_args!("count,delta={}", convert(&self.kind, delta)))?;
+        } else if let Some(gauge) = self.gauge.to_owned() {
+            fmt.write_char(' ')?;
+            fmt.write_fmt(format_args!("gauge,{}", convert(&self.kind, gauge)))?;
+        }
```
TommyCpp, line 375: I think we can simplify something like `gauge,min=80.6,max=80.6,sum=80.6,count=1` to `80.6`. Maybe it will be helpful to add it here also.

TommyCpp (changes requested):
This generally looks good. Thanks for working on this! One issue is how do we deal with different async runtimes. 

Currently in the PR one runtime will override the other one. We used to manage runtimes like this but the problem here is it becomes hard to test around because you need to explicitly disable `tokio` to use `async-std`. 

We end up asking users to provide a `Runtime` to run the collector. See #480. It may require some major updates so if you want we can address this in a following up PR. 

However, either way, we should make sure `rt-async-std` gets tested. This is currently untested because if we run the test on all features `rt-tokio` will override it.

opentelemetry-dynatrace/src/metric.rs
```diff
@@ -0,0 +1,499 @@
...
+                        .map(|sender| {
+                            let _ = sender.try_send(ClientMessage::Export(Box::new(req)));
+                        })
+                        .map_err(|_| Error::PoisonedLock("dynatrace metrics exporter"))?;
+
+                    #[cfg(all(not(feature = "rt-tokio"), feature = "rt-async-std"))]
+                    self.sender
+                        .lock()
+                        .map(|mut sender| {
+                            let _ = sender.try_send(ClientMessage::Export(req));
```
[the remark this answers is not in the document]
diceride: Oh .. fixed, very good!

opentelemetry-dynatrace/src/transform/metrics.rs
```diff
@@ -0,0 +1,1165 @@
...
+    let kind = descriptor.number_kind();
+
+    let key = if prefix.is_some() {
+        MetricKey::new(format!("{}.{}", prefix.unwrap(), descriptor.name()))
+    } else {
+        MetricKey::new(descriptor.name().to_string())
+    };
+
+    let source_key = METRICS_SOURCE;
+    let source_value = Value::String("opentelemetry".into());
```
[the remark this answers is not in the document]
diceride: This is an internal requirement to detect issues with data ingestion sources.

opentelemetry-dynatrace/src/transform/metrics.rs
```diff
@@ -0,0 +1,1165 @@
...
+            && match (self.gauge.clone(), other.gauge.clone()) {
+                (Some(a), Some(b)) => a.partial_cmp(&self.kind, &b) == Some(cmp::Ordering::Equal),
+                (None, None) => true,
+                _ => false,
+            }
+            && self.timestamp.eq(&other.timestamp)
+    }
+}
+
+/// Transform a record to [Dynatrace metrics ingestion protocol] metric line.
```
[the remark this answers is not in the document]
diceride: I removed the link

opentelemetry-dynatrace/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// The pipeline will need a exporter to complete setup. Throw this error if none is provided.
+    #[error("no exporter builder is provided, please provide one using with_exporter() method")]
+    NoExporterBuilder,
+
+    /// Invalid URI.
+    #[error("invalid URI {0}")]
+    InvalidUri(#[from] http::uri::InvalidUri),
+
+    /// Http requests failed because no http client is provided.
+    #[error("no http client, you must select one")]
```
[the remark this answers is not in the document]
diceride: Yes good point, this makes it more understandable. I updated the error message in [opentelemetry-otlp/src/lib.rs](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/lib.rs#L315) as well (in a separate commit).

opentelemetry-dynatrace/src/exporter/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl ExportConfig {
+    /// Set the address of the Dynatrace endoint.
+    pub fn with_endpoint<T: Into<String>>(mut self, endpoint: T) -> Self {
+        self.endpoint = Some(endpoint.into());
+        self
+    }
+
+    /// Set the API token for authentication.
+    pub fn with_token(mut self, token: String) -> Self {
```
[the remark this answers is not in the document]
diceride: Yes, there is no reason not to use `<T: Into<String>` here too.

opentelemetry-dynatrace/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+[`OpenTelemetry`] is a collection of tools, APIs, and SDKs used to instrument,
+generate, collect, and export telemetry data (metrics, logs, and traces) for
+analysis in order to understand your software's performance and behavior. This
+crate provides additional propagators and exporters for sending telemetry data
+to [`Dynatrace`].
+
+## Exporter features
+
+>
```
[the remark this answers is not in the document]
diceride: I removed the reference block from the markdown file.

opentelemetry-dynatrace/Cargo.toml
diceride: hmm from my understanding, `reqwest` (and other http clients) will do some magic when the compilation target is set to wasm32-unknown-unknown. Sure using the feature flag won't do it, but I expect the target audience to understand this.

https://github.com/seanmonstar/reqwest/issues/288

opentelemetry-dynatrace/src/transform/metrics.rs
diceride: Yes, I'm going to improve this in a subsequent PR. This thing doesn't look good at all and definitively needs some work.

opentelemetry-dynatrace/src/transform/metrics.rs
diceride: I added comments to all occurrences of `take` to make it better understandable.

opentelemetry-dynatrace/src/transform/metrics.rs
diceride: Yes, very good idea. I implemented this change and updated the integration test.

opentelemetry-dynatrace/src/transform/metrics.rs
TommyCpp: @diceride Can we a TODO item here? Thanks in advance

opentelemetry-dynatrace/Cargo.toml
```diff
@@ -0,0 +1,92 @@
...
+
+[[test]]
+name = "integration_tokio"
+path = "tests/http_test.rs"
+required-features = ["rt-tokio"]
+
+[[test]]
+name = "integration_async_std"
+path = "tests/http_test.rs"
+required-features = ["rt-async-std"]
```
TommyCpp, line 30: I think currently there are only tokio tests in `tests/http_test.rs`?

TommyCpp:
Some small nits but otherwise I think it looks good. Thanks for the hard work on this!

opentelemetry-dynatrace/Cargo.toml
diceride: I removed the configuration for `async-std`

opentelemetry-dynatrace/src/exporter/mod.rs
```diff
@@ -0,0 +1,53 @@
+//! OpenTelemetry Dynatrace Exporter Configuration
+#![allow(unused_attributes)]
```
TommyCpp, line 2: Any special reason why we should allow this here?

TommyCpp:
Since we don't have integration test with `async-std`. I think it will be helpful to add a lint check. It can be added in https://github.com/open-telemetry/opentelemetry-rust/blob/main/scripts/lint.sh

Something like 
```
cargo_feature opentelemetry-dynatrce "async-std,metrics" 
```
should work. Or you can add multiple for any feature combination you want check.

opentelemetry-dynatrace/src/exporter/mod.rs
diceride: I remember cargo fmt throwing some warnings in previous versions about unused variables, but I guess this is now obsolete. I removed the configuration.

diceride:
Yes, I added additional matrix tests to the lint.sh script.

There was also an issue I resolved in 3f7f197 caused by #692.

diceride:
Yes, I added additional matrix tests to the lint.sh script.

There was also an issue I resolved in 70dd5dd caused by #692.

diceride:
Yes, I added additional matrix tests to the lint.sh script.

There was also an issue I resolved in 70dd5dd mentioned in #692.

TommyCpp (approved):
LGTM. Thanks a lot 🎉

Issue #555, opened 2021-05-20
Add support for schema URL


jtescher:
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/1666

TommyCpp:
We might want to discuss if it's possible to add a new configuration into tracer API without changing the public API. This feature was added after tracing spec stabled. So I imagine they could add more configuration into it.

dgetu:
Could I take this issue?

dgetu:
I'm not exactly sure where to start with that. I've only just started getting involved with the project.

TommyCpp:
I guess the idea is to design the API so that we can add other configurations in the future without breaking the existing code. Other languages support it with optional parameters. Need to figure out what's the best choice in Rust.

dgetu:
I don't think I have enough experience with either the language or the codebase to provide anything meaningful to that discussion. I'd still like to contribute however I can.

dgetu:
I took another look at spec 1666, and I think I'm starting to see the problem. The only way I can think of to implement this without it being a breaking change is to implement new function headers. Either a replacement for an overload a la Java or a setter for a builder pattern. Is that the kind of change to the tracer API you wanted to see made to the spec? I guess I'm not sure what the difference is between "tracer API" and "public API."

TommyCpp:
We haven't declared GA yet so it's fine to make breaking change as for now. I want to explore if we can rebuild the Rust tracer API so that in the future we can add more options to it if needed. The current implementation doesn't support that because if we add another parameter to the function, it will break existing code. 

I feel like we can do it either by having a builder for the tracer or having a configuration object as the parameter(like what Golang did). So I don't think we need to change the spec API.

dgetu:
I think that having a config struct would be more clear than a builder. I can try both and see which one is more suitable.

dgetu:
I'm in the process of migrating to a single config struct parameter for `TracerProvider::get_tracer`, but I'm not sure how to go about building it. My current approach is to have the user update their own config struct using Default, but this poses a problem. When every field in the struct is populated, attempting to update it results in a [linter warning](https://rust-lang.github.io/rust-clippy/master/index.html#needless_update), which I have reservations about suppressing. 

[...]

PR #682, opened 2021-12-23
trace: split `tracer` and `versioned_tracer` methods

jtescher:
The [trace spec] requires that `TracerProvider`s MUST accept optional `version` and `schema_url` parameters. This introduces ergonomic issues as rust does not have a variadic solution outside of macros, leading to many calls with only a single relevant argument (e.g. `tracer(name, None, None)`).

This patch splits the current `TracerProvider::tracer` method into `TracerProvider::versioned_tracer` which implements the spec mandated optional fields, as well as a `TracerProvider::tracer` method that accepts only a `name` parameter as a convenience method for the above listed common case.

[trace spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.4.0/specification/trace/api.md#get-a-tracer

### DIFF
--- a/opentelemetry-datadog/src/exporter/mod.rs
+++ b/opentelemetry-datadog/src/exporter/mod.rs
@@ -169,7 +169,11 @@ impl DatadogPipelineBuilder {
             sdk::trace::TracerProvider::builder().with_simple_exporter(exporter);
         provider_builder = provider_builder.with_config(config);
         let provider = provider_builder.build();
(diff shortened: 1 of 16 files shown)


TommyCpp:
I think it's related to #555. Since we started preparing GA. One question here is what if the spec adds another parameter in the future?

jtescher:
I think as the trace spec is frozen we can assume they will not for now. If they do, then the simplest thing seems to be to either add another method or do a major version bump to support the new spec.

TommyCpp (approved):
Looks good 👍

TommyCpp:
I think we can close #555 with this PR

PR #684, opened 2021-12-29
Reduce dependency on futures crate

robjtede:
Size of dep graph:

| | before|after
|-|-|-|
opentelemetry|47|38
opentelemetry-contrib|122|113
opentelemetry-datadog|136|130
opentelemetry-http|53|44
opentelemetry-jaeger|129|120
opentelemetry-otlp|140|139
opentelemetry-prometheus|52|43
opentelemetry-aws|137|128
opentelemetry-semantic-conventions|48|39
opentelemetry-stackdriver|180|180
opentelemetry-zipkin|218|211
opentelemetry-zpages|149|149

Only `opentelemetry-otlp` could not have it's `futures` dep removed (yet). It is blocked on the release of https://github.com/tikv/grpc-rs/pull/554 in order for the codegen to reference `std`'s `Future`.

### DIFF
--- a/examples/actix-http/Cargo.toml
+++ b/examples/actix-http/Cargo.toml
@@ -8,7 +8,6 @@ publish = false
 opentelemetry = { path = "../../opentelemetry", features = ["rt-tokio"] }
 opentelemetry-jaeger = { path = "../../opentelemetry-jaeger", features = ["reqwest_collector_client", "rt-tokio"] }
 thrift = "0.13"
-futures = "0.3"
 actix-web = "4.0.0-beta.4"
 actix-service = "2.0.0-beta.5"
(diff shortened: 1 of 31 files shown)


djc:
Nice, would be happy to merge this.

djc (approved):

djc:
How do you measure size of individual crate's dep graphs?

robjtede:
It's not scientific but: `cargo tree -e=normal -p="$p" | wc -l`

robjtede:
Updated with slightly more scientific:
```sh
cargo tree -e=normal --prefix=none -p="$p" | grep -v '*' | wc -l | xargs
```

PR #688, opened 2021-12-31
Fix broken docs

jtescher:


djc:
Maybe add a CI step along these lines? Should help catch these issues in CI.

```
    - name: doc
      run: cargo doc --no-deps
      env:
        RUSTDOCFLAGS: -Dwarnings
```

djc:
Why not fix up those allow thingies?

jtescher:
They are in generated code, taking a look.

djc:
That's fair. Would be nice if we can restrict the allows to the generated code and/or add a comment that they're there because of generated code.

djc (approved):

jtescher:
That should take care of the semantic convention crate, the stackdriver issues will take some more work. Can do those in a follow up.

PR #689, opened 2022-01-01
trace: spec compliant trace and span ids

jtescher:
The spec only defines conversions between ids and hex/binary values. This updates the `TraceId` and `SpanId` APIs to be compliant by removing explicit conversions from u128/u64 (except for test convenience methods) and adding methods for converting to/from byte arrays and hex strings.

### DIFF
--- a/opentelemetry/src/trace/span_context.rs
+++ b/opentelemetry/src/trace/span_context.rs
@@ -139,12 +139,6 @@ impl From<[u8; 16]> for TraceId {
     }
 }
 
-impl From<TraceId> for [u8; 16] {
-    fn from(id: TraceId) -> [u8; 16] {
-        id.to_bytes()
-    }
-}
-
 impl fmt::Debug for TraceId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_fmt(format_args!("{:032x}", self.0))
@@ -204,12 +198,6 @@ impl SpanId {
 impl From<[u8; 8]> for SpanId {
     fn from(bytes: [u8; 8]) -> Self {
         SpanId::from_bytes(bytes)
-    }
-}
-
-impl From<SpanId> for [u8; 8] {
-    fn from(id: SpanId) -> [u8; 8] {
-        id.to_bytes()
     }
 }
 
--- a/opentelemetry/src/trace/span_context.rs
+++ b/opentelemetry/src/trace/span_context.rs
@@ -15,6 +15,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::VecDeque;
 use std::fmt;
 use std::hash::Hash;
+use std::num::ParseIntError;
 use std::ops::{BitAnd, BitOr, Not};
 use std::str::FromStr;
 use thiserror::Error;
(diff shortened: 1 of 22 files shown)


opentelemetry/src/trace/span_context.rs
```diff
@@ -94,89 +95,139 @@ impl fmt::LowerHex for TraceFlags {
...
+}
 
-    /// Convert from TraceId to Big-Endian byte array
-    pub fn to_byte_array(self) -> [u8; 16] {
-        self.0.to_be_bytes()
+impl From<TraceId> for [u8; 16] {
+    fn from(id: TraceId) -> [u8; 16] {
+        id.to_bytes()
     }
+}
```
djc, line 146: What's the motivation for having this? I think I'd maybe prefer to require callers to spell `id.to_bytes()` explicitly (because this sort of loses semantic context).

opentelemetry/src/trace/span_context.rs
```diff
@@ -94,89 +95,139 @@ impl fmt::LowerHex for TraceFlags {
...
+    /// Invalid trace id
+    pub const INVALID: TraceId = TraceId(0);
+
+    /// Create a trace id from its representation as a byte array.
+    pub const fn from_bytes(bytes: [u8; 16]) -> Self {
+        TraceId(u128::from_be_bytes(bytes))
     }
 
-    /// Convert from u128 to TraceId
-    pub fn from_u128(item: u128) -> Self {
```
djc, line 110: Hmm, does the spec mandate that we not have a conversion from `u128`? That seems a little silly. I think having an `impl From<u128> for TraceId` could make sense.

opentelemetry/src/trace/span_context.rs
```diff
@@ -94,89 +95,139 @@ impl fmt::LowerHex for TraceFlags {
...
-pub struct SpanId(u64);
+#[derive(Clone, PartialEq, Eq, Copy, Hash)]
+pub struct SpanId(pub(crate) u64);
 
 impl SpanId {
-    /// Construct a new invalid (zero-valued) SpanId
-    pub fn invalid() -> Self {
-        SpanId(0)
+    /// Invalid span id
+    pub const INVALID: SpanId = SpanId(0);
```
djc, line 175: Similar comments here compared to `TraceId`.

djc (approved):
Some minor concerns, but this mostly looks good. I like having `INVALID` as a const.

opentelemetry/src/trace/span_context.rs
jtescher: They were trying to avoid specifying an endianness I believe. I don't have strong feelings about it either way.

opentelemetry/src/trace/span_context.rs
jtescher: sounds reasonable 👍

opentelemetry/src/trace/span_context.rs
jtescher: Trying to get the api slimmed down to just what the spec mandates so we can get a 1.0 release, can always bring this back if there is demand but currently this does bring the exporters more in line with other language's implementations in terms of relying more on opaque byte arrays.

djc: Fair enough.

djc (approved):

PR #693, opened 2022-01-04
sdk: optimize span creation internals

jtescher:
* switch from `Option::map` to pattern match in sdk tracer: ~16% perf improvement on span creation.
* optimize for common case of single span processor in sdk span: ~1% perf improvement on span end

Overall improvement on span start-end benchmark: ~17%


opentelemetry/src/sdk/trace/span.rs
```diff
@@ -173,34 +173,49 @@ impl crate::trace::Span for Span {
 impl Span {
     fn ensure_ended_and_exported(&mut self, timestamp: Option<SystemTime>) {
         if let Some(mut data) = self.data.take() {
```
djc, line 175: I think while you're in here it would be good to further reduce rightward drift. You can turn the `if let Some(mut data) = self.data.take()` and `if let Some(timestamp) = timestamp` into `let foo = match bar { Some(x) => x, None => return };` expressions, which would really help.

opentelemetry/src/sdk/trace/span.rs
```diff
@@ -173,34 +173,49 @@ impl crate::trace::Span for Span {
...
+                        &self.tracer,
+                    ));
+                } else {
+                    // send to all processors
+                    let mut processors = processors.iter().peekable();
+                    let resource = provider.config().resource.clone();
+                    let mut span_data = Some((data, resource));
+                    while let Some(processor) = processors.next() {
+                        let span_data = if processors.peek().is_none() {
+                            // last loop or single processor/exporter, move data
```
djc, line 204: The `single processor/exporter` case no longer gets here, right?

opentelemetry/src/sdk/trace/span.rs
```diff
@@ -173,34 +173,49 @@ impl crate::trace::Span for Span {
...
+                }
+
+                // Ensure end time is set via explicit end or implicitly on drop
+                if let Some(timestamp) = timestamp {
+                    data.end_time = timestamp;
+                } else if data.end_time == data.start_time {
+                    data.end_time = crate::time::now();
+                }
+
+                if processors.len() == 1 {
```
djc, line 189: This might be even more efficient as a match?

```
match processors {
    [] => return,
    [processor] => // export on single processor,
    _ => // send to all processors
}
```

jtescher: good idea 👍

opentelemetry/src/sdk/trace/span.rs
jtescher: yep!

opentelemetry/src/sdk/trace/span.rs
jtescher: 👍

opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -186,9 +184,8 @@ impl crate::trace::Tracer for Tracer {
         };
 
         // Build context for sampling decision
-        let (no_parent, trace_id, parent_span_id, remote_parent, parent_trace_flags) = parent_span
-            .as_ref()
-            .map(|parent| {
+        let (no_parent, trace_id, parent_span_id, remote_parent, parent_trace_flags) =
```
djc, line 187: Maybe use `match` directly here instead of the `if let Some(_) .. { } else { }`? You could add a `.map(|parent| parent.span_context())` to the match expression to make it more concise.

It would be nice to get rid of the 5-tuple, too. Since `no_parent` and `remote_parent` are just `bool`s, maybe initialize them to the non-parent case before the match? Can probably do the same for `parent_span_id`, since initializing to the const `SpanId::INVALID` should be cheap enough.

opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -286,22 +285,34 @@ impl crate::trace::Tracer for Tracer {
...
-                events: events_queue,
-                links,
-                status_code,
-                status_message,
-            }
-        });
-
-        let span_context = SpanContext::new(trace_id, span_id, flags, false, span_trace_state);
-        let mut span = Span::new(span_context, inner, self.clone(), span_limits);
+            let span_context = SpanContext::new(trace_id, span_id, flags, false, trace_state);
```
djc, line 288: Given that the only differences here are `flags` and `trace_state`, it seems unfortunate to duplicate the calls to `SpanContext::new()` and `Span::new()` entirely.

djc (commented):
Spent some more time looking on this, more style suggestions:

opentelemetry/src/sdk/trace/tracer.rs
jtescher: sounds good, definitely nice to not have the 5-tuple 👍

opentelemetry/src/sdk/trace/tracer.rs
jtescher: Yeah this one is a bit unfortunate, but it does seem more performant than returning say `let (span_context, inner) = ...` to construct the span from the result.

jtescher: (`~407.23 ns` vs  `~389.59 ns` on my machine for `start-end-span/always-sample` bench.)

opentelemetry/src/sdk/trace/tracer.rs
```diff
@@ -184,27 +188,17 @@ impl crate::trace::Tracer for Tracer {
...
-                    true,
-                    builder
-                        .trace_id
-                        .unwrap_or_else(|| config.id_generator.new_trace_id()),
-                    SpanId::INVALID,
-                    false,
-                    TraceFlags::default(),
-                )
-            };
+        if let Some(sc) = parent_span.as_ref().map(|parent| parent.span_context()) {
```
djc, line 191: Any reason not to `let trace_id = match parent_span.as_ref() [etc] {};`?

jtescher: Don't have strong feelings either way, seems easier to read if there are either assignments to mut vars or returned data. Mixing both seemed slightly more likely to be misread.

djc (approved):

PR #701, opened 2022-01-11
Remove authors from crate metadata (see RFC 3052)

djc:


jtescher (approved):

Xuanwo:
For anyone who jumped into this PR but didn't know what's `RFC 3052` is, here is the message:

The `RFC 3052` mentioned here is a Rust RFC that intends to make the `authors` field in `Cargo.toml` optional.

RFC Link: https://rust-lang.github.io/rfcs/3052-optional-authors-field.html

PR #703, opened 2022-01-12
stackdriver: further improvements

djc:
* Avoid spawning inside the library
* Use a concrete error type
* Generate protobuf bindings in a test (avoids build script + dependencies, makes generated code visible to the IDE)


opentelemetry-stackdriver/tests/generate.rs
```diff
@@ -0,0 +1,234 @@
...
+    // Compare the old version and new version and fail the test if they're different.
+
+    let mut keys = versions[0].keys().collect::<Vec<_>>();
+    keys.extend(versions[1].keys());
+    keys.sort_unstable();
+    keys.dedup();
+
+    let _ = fs::remove_dir_all(SOURCE_DIR);
+    fs::rename(GENERATED_TEMP, SOURCE_DIR).unwrap();
+    assert!(versions[0] == versions[1]);
```
TommyCpp, line 207: I think it may be helpful to add some comments asking people to commit the changes. Just in case people updating other crates encounters this.

djc: Thanks for the review! I added a bunch of comments to make it clearer what's going on.

PR #705, opened 2022-01-17
stackdriver: avoid calling log from inside exporter

djc:


djc:
Failure is unrelated.

PR #706, opened 2022-01-17
fix: jaeger collector only works in batch mode

skyzh:
The docs is misleading for jaeger HTTP collector mode -- `install_simple` will always initialize a UDP client even if users have configured to use HTTP collector.

In `install_simple`, we called `build_simple`.

https://github.com/open-telemetry/opentelemetry-rust/blob/605d6c04dd2d3311787372bac52cd0a0c021d5ea/opentelemetry-jaeger/src/exporter/mod.rs#L324-L333

In `build_simple`, we called `init_sync_exporter_with_process`.

https://github.com/open-telemetry/opentelemetry-rust/blob/605d6c04dd2d3311787372bac52cd0a0c021d5ea/opentelemetry-jaeger/src/exporter/mod.rs#L405-L413

In that, we calls `init_sync_uploader`.

https://github.com/open-telemetry/opentelemetry-rust/blob/605d6c04dd2d3311787372bac52cd0a0c021d5ea/opentelemetry-jaeger/src/exporter/mod.rs#L439-L448

... which ignores the HTTP collector settings and will only create a UDP socket.

https://github.com/open-telemetry/opentelemetry-rust/blob/605d6c04dd2d3311787372bac52cd0a0c021d5ea/opentelemetry-jaeger/src/exporter/mod.rs#L477-L485

In the future, we should warn users or force use reqwest / isahc sync mode here. But for now, I think it's a quick fix to simply change the docs and prevent readers to mis-configure their clients.

Thanks for review!


skyzh:
cc @djc @TommyCpp would you please take a look? Thanks!

TommyCpp (changes requested):
Good catch! And yes we probably need to use sync mode in HTTP clients if that's available. 

I believe we also need to add `#[tokio::main]` on top of the main function and make it async for it to works.

skyzh:
> Good catch! And yes we probably need to use sync mode in HTTP clients if that's available.
> 
> I believe we also need to add `#[tokio::main]` on top of the main function and make it async for it to works.

Thanks! However, it seems other parts of this docs don't add `#[tokio::main]` for those examples. For example,

https://github.com/open-telemetry/opentelemetry-rust/blob/5972c5db422e3e28967120b21478a1874bc26e13/opentelemetry-jaeger/src/lib.rs#L136-L151

As those examples won't run in doctests, I think it's reasonable to ignore things like `#[tokio::main]`?

TommyCpp (approved):
LGTM. Thanks

PR #710, opened 2022-01-23
[proto] add opentelemetry-proto crates

### LINKED ISSUE #627: Publish protobuf conversions as their own crate or otherwise expose
I have a usecase where I've written a Rust lib that is used by a variety of other languages to create a single worker process. I don't want to require my users to run the OTel collector when they are running our workers, but I do have a need to propagate spans and metrics from the Rust library, to the <lang> library where they can they be exported by the user's choice of exporter.

Seemingly, the easiest way to do this is for me to essentially do what happens when exporting to an OTel collector, but rather than actually calling a collector over HTTP or gRPC, I simply pass the protobufs in memory to the <lang> side where they are unpacked, turned into that languages' typical representation (would also be nice if the SDKs had proto->lang api converters to make that easy), and then passed to the user's exporter.

As it stands though, all the proto conversion stuff is privately hidden inside https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-otlp/src/transform -- if this was it's own crate, or even publicly exposed, that would be awesome! Right now I'm resorting to just copying your transformation code in order to do this.

Thanks!

TommyCpp:
Added a separate crate with type conversion between generated types and types defined in `opentelemetry`.
- Added an `opentelemetry-proto` crate, reduced duplicate conversion codes.
- Update proto version to 0.9
- Clean up features in `opentelemetry-otlp`. Rename some of them to keep them consistent. 

fix #627


TommyCpp:
I think the coverage cannot compile because the compiler panic when compiling using nightly 1.60

jtescher (approved):
This looks reasonable to me

PR #714, opened 2022-02-02
Add logs to opentelemetry-proto crate

mbrobbel:
Please note that they are still [experimental](https://github.com/open-telemetry/opentelemetry-proto#maturity-level).


TommyCpp:
Looks good. Thanks. It maybe worth to explore how we can add log support in the future.

TommyCpp (approved):

Issue #549, opened 2021-05-11
eprintln is used for logging


kjvalencik:
Currently, non-fatal errors (for example, exporting traces) use `eprintln` for logging.

https://github.com/open-telemetry/opentelemetry-rust/blob/1cdf725c0734d0a16c29a6f693ef6fe795000bbe/opentelemetry/src/global/error_handler.rs#L50

This is very noisy, especially in cases such as development where it is an expected error.

```
OpenTelemetry trace error occurred ExportFailed(ThriftAgentError(TransportError { kind: NotOpen, message: "connection refused" }))
```

Ideally `opentelemetry` would use the [`log`](https://crates.io/crates/log) crate, allowing users to opt-in `opentelemetry` logs.

Additionally, these logs may not be visible at all where `stderr` is not connected (e.g., Android).

jtescher:
We're holding off on supporting log/tracing directly for logging purposes while the otel log spec is being worked on (and as you mentioned you can easily override the current behavior via [opentelemetry::global::set_error_handler](https://docs.rs/opentelemetry/0.14.0/opentelemetry/global/fn.set_error_handler.html) to support either logging facade yourself), so closing this for now, but we can re-open once the logging spec has progressed to a point where we have more clarity.

kjvalencik:
Thanks for the update @jtescher. That makes sense.

The error enum isn't exported so you can't do much with it expect display/format it. Is that intentional?

https://github.com/open-telemetry/opentelemetry-rust/blob/133ab27744776134eeee798a442f23a291571f6e/opentelemetry/src/global/error_handler.rs#L14-L28

Take this with a grain of salt since it wasn't too hard to find, but the docs don't mention the global error handler outside of the small line in the API docs. I needed to read the source to find where the `eprintln` were coming from and how to prevent them. It would be nice if it were more discoverable.

Thanks again! Cheers!

PR #718, opened 2022-02-05
opentelemetry-zipkin: Add Environment Variables

### LINKED ISSUE #534: Add support for OTEL_EXPORTER_ZIPKIN_TIMEOUT
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/1636

hdost:
As described in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md
* Add support for:
** OTEL_EXPORTER_ZIPKIN_ENDPOINT
** OTEL_EXPORTER_ZIPKIN_TIMEOUT
* Update Default to use

Closes #534

Signed-off-by: Harold Dost <h.dost@criteo.com>

### DIFF
--- a/opentelemetry-zipkin/src/exporter/env.rs
+++ b/opentelemetry-zipkin/src/exporter/env.rs
@@ -1,0 +1,63 @@
+use std::env;
+use std::time::Duration;
+
+/// Default Zipkin collector endpoint
+const DEFAULT_COLLECTOR_ENDPOINT: &str = "http://127.0.0.1:9411/api/v2/spans";
+
+/// HTTP endpoint for Zipkin collector.
+/// e.g. "http://localhost:9411/api/v2/spans"
+const ENV_ENDPOINT: &str = "OTEL_EXPORTER_ZIPKIN_ENDPOINT";
+
+/// Maximum time the Zipkin exporter will wait for each batch export
+const ENV_TIMEOUT: &str = "OTEL_EXPORTER_ZIPKIN_TIMEOUT";
+
+/// Default Zipkin timeout in milliseconds
+const DEFAULT_COLLECTOR_TIMEOUT: Duration = Duration::from_millis(10_000);
+
+pub(crate) fn get_timeout() -> Duration {
+    match env::var(ENV_TIMEOUT).ok().filter(|var| !var.is_empty()) {
+        Some(timeout) => match timeout.parse() {
+            Ok(timeout) => Duration::from_millis(timeout),
+            Err(e) => {
+                eprintln!("{} malformed defaulting to 10000: {}", ENV_TIMEOUT, e);
+                DEFAULT_COLLECTOR_TIMEOUT
+            }
+        },
+        None => DEFAULT_COLLECTOR_TIMEOUT,
+    }
+}
+
+pub(crate) fn get_endpoint() -> String {
+    match env::var(ENV_ENDPOINT).ok().filter(|var| !var.is_empty()) {
+        Some(endpoint) => endpoint,
+        None => DEFAULT_COLLECTOR_ENDPOINT.to_string(),
+    }
+}
+
+#[test]
+fn test_collector_defaults() {
+    // Ensure the variables are undefined.
+    env::remove_var(ENV_TIMEOUT);
+    env::remove_var(ENV_ENDPOINT);
+    assert_eq!(DEFAULT_COLLECTOR_TIMEOUT, get_timeout());
+    assert_eq!(DEFAULT_COLLECTOR_ENDPOINT, get_endpoint());
+}
+
+#[test]
+fn test_collector_bad_timeout() {
+    env::set_var(ENV_TIMEOUT, "a");
+    assert_eq!(DEFAULT_COLLECTOR_TIMEOUT, get_timeout());
+}
+
+#[test]
+fn test_collector_good_timeout() {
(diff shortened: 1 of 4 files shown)


opentelemetry-zipkin/src/exporter/env.rs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    #[cfg(any(
+        feature = "reqwest-client",
+        feature = "surf-client",
+        feature = "reqwest-blocking-client"
+    ))]
+    {
+        if let Some(timeout) = env::var(ENV_TIMEOUT).ok().filter(|var| !var.is_empty()) {
+            let timeout = duration_str::parse(&timeout)
```
TommyCpp, line 42: I don't think we need to parse duration here. [As defined in spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#duration), all duration should be an integer representing duration in ms

hdost: Indeed I missed that part.

hdost: Also apparently should only warn and fail gracefully . I'll patch for that also

opentelemetry-zipkin/src/exporter/env.rs
```diff
@@ -0,0 +1,66 @@
...
+                feature = "reqwest-client"
+            ))]
+            let client = reqwest::Client::builder().timeout(timeout).build().unwrap();
+            #[cfg(all(
+                not(feature = "reqwest-client"),
+                not(feature = "reqwest-blocking-client"),
+                feature = "surf-client"
+            ))]
+            let client: surf::Client = surf::Config::new()
+                .set_timeout(Some(timeout)).into();
```
TommyCpp, line 59: We may want to move the client initialization into this function instead of `default` and we should probably provide a default value of `10s` if the env var is not set. 

I see some of the client builders return `Result`. Generally, we want to avoid `unwrap`. Maybe we can fall back to the client without a timeout in this case. Although I don't expect it will happen

TommyCpp:
Overall looks good. Just need to add some default for timeout. We probably shouldn't rely on the default timeout in different HTTP clients as they may be different. 

Thanks in advance!

opentelemetry-zipkin/src/exporter/env.rs
TommyCpp: Thanks a lot!

hdost:
> Overall looks good. Just need to add some default for timeout. We probably shouldn't rely on the default timeout in different HTTP clients as they may be different.
> 
> Thanks in advance!

Sounds good I'll update 👍🏼

hdost:
Ok should be set assuming test pass in the pipeline. 
- [x] Warn on bad values
- [x] Default to values when bad values received
- [x]  Default values when none provided.

hdost:
And I’ll get some tests on it. I just wanted to ensure the main code was as desired.

opentelemetry-zipkin/src/exporter/mod.rs
```diff
@@ -57,21 +55,31 @@ pub struct ZipkinPipelineBuilder {
...
     fn default() -> Self {
+        let timeout = env::get_timeout();
         ZipkinPipelineBuilder {
             #[cfg(feature = "reqwest-blocking-client")]
-            client: Some(Box::new(reqwest::blocking::Client::new())),
+            client: Some(Box::new(
+                reqwest::blocking::Client::builder()
+                    .timeout(timeout)
+                    .build()
+                    .unwrap(),
```
TommyCpp, line 65: Maybe we can use `unwrap_or` to create a client without timeout if for some reason the build failed. It would be better to break ppl's application.

opentelemetry-zipkin/src/exporter/env.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// Default Zipkin timeout in milliseconds
+const DEFAULT_COLLECTOR_TIMEOUT: Duration = Duration::from_millis(10_000);
+
+pub(crate) fn get_timeout() -> Duration {
+    match env::var(ENV_TIMEOUT).ok().filter(|var| !var.is_empty()) {
+        Some(timeout) => match timeout.parse() {
+            Ok(timeout) => Duration::from_millis(timeout),
+            Err(e) => {
+                warn!("{} malformed defaulting to 10000: {}", ENV_TIMEOUT, e);
```
TommyCpp, line 23: We are currently just using `eprintln!` for output logs. The reason is in the future this crate will add support for logs so it will be tricky.  See https://github.com/open-telemetry/opentelemetry-rust/issues/549

TommyCpp (commented):
Sorry for the delay. Thanks for the change logs. Just a few comments and we should be good to merge. Thanks again for working on this.

opentelemetry-zipkin/src/exporter/mod.rs
hdost: Swapped for unwrap_or_else as unwrap_or is typically meant for an existing value. Otherwise shows a clippy warning.

opentelemetry-zipkin/src/exporter/env.rs
hdost: Done

TommyCpp (approved):
Nice! Thanks again for working on it.

PR #727, opened 2022-02-09
opentelemetry-prometheus: Remove deprecate call

hdost:
Remove depcecated call to PrometheusExporter::new() which was deprecated
in favor of using ExporterBuilder for new Exporters.


TommyCpp (approved):

TommyCpp:
Looks good. Thanks!

PR #728, opened 2022-02-09
Deprecate ValueRecorder in favor of Histogram

hdost:
The Metric API Spec is now stable and ValueRecorder was replaced with
Histogram.

### DIFF
--- a/opentelemetry/src/metrics/histogram.rs
+++ b/opentelemetry/src/metrics/histogram.rs
@@ -5,19 +5,19 @@ use crate::metrics::{
 use crate::KeyValue;
 use std::marker;
 
-/// ValueRecorder is a metric that records per-request non-additive values.
+/// Histogram is a metric that records per-request non-additive values.
 #[derive(Clone, Debug)]
-pub struct ValueRecorder<T>(SyncInstrument<T>);
+pub struct Histogram<T>(SyncInstrument<T>);
 
-impl<T> ValueRecorder<T>
+impl<T> Histogram<T>
 where
     T: Into<Number>,
 {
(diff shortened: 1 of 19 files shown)


hdost:
@TommyCpp I know the codecov are optional is it ok in this case since it's bascially just replacing a type?

TommyCpp:
> @TommyCpp I know the codecov are optional is it ok in this case since it's bascially just replacing a type?

Yeah, don't worry about it especially for -0.02% changes

opentelemetry/src/metrics/histogram.rs
```diff
@@ -0,0 +1,106 @@
...
+        self
+    }
+
+    /// Set the unit for this `Histogram`.
+    pub fn with_unit(mut self, unit: Unit) -> Self {
+        self.descriptor.config.unit = Some(unit);
+        self
+    }
+
+    /// Creates a new value recorder.
```
TommyCpp, line 89:
```suggestion
    /// Tries to create a new `Histogram`.
```

opentelemetry/src/metrics/histogram.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn try_init(self) -> Result<Histogram<T>> {
+        let instrument = self.meter.new_sync_instrument(self.descriptor)?;
+        Ok(Histogram(SyncInstrument::new(instrument)))
+    }
+
+    /// Creates a new value recorder.
```
TommyCpp, line 95:
```suggestion
    /// Creates a new `Histogram`.
```

hdost:
Only change was a rebase

TommyCpp:
Looks good. Thanks

PR #729, opened 2022-02-09
opentelemetry-jaeger: Add Timeout Environment Var

### LINKED ISSUE #528: Add support for OTEL_EXPORTER_JAEGER_TIMEOUT
See the spec change for more details https://github.com/open-telemetry/opentelemetry-specification/pull/1612

hdost:
OTEL_EXPORTER_JAEGER_TIMEOUT is now supported for all cases where a
non-custom HTTP Client is utilized.

Closes #528


hdost:
@TommyCpp anything further needed in this PR?

jtescher (commented):
Direction looks good, tests failing though

hdost:
> Direction looks good, tests failing though

Yea :/ there was a flaky test on zipkin. made it more like the ones here on jaeger

hdost:
@TommyCpp out of curiosity has there been the thought to turn off the requirement of being up to the latest before merging? I know there's benefits of both ways. 🤷🏼‍♂️

TommyCpp:
> @TommyCpp out of curiosity has there been the thought to turn off the requirement of being up to the latest before merging? I know there's benefits of both ways. 🤷🏼‍♂️ Just seems like a lot of unnecessary merge commits :/

I agree but AFAIK it's an org-level rule on protected branches. I don't think maintainers can change it in settings :disappointed:

hdost:
> > @TommyCpp out of curiosity has there been the thought to turn off the requirement of being up to the latest before merging? I know there's benefits of both ways. 🤷🏼‍♂️ Just seems like a lot of unnecessary merge commits :/
> 
> I agree but AFAIK it's an org-level rule on protected branches. I don't think maintainers can change it in settings disappointed

I'm not aware of any such possibilities. I know you can on repos, but from the Repo's that I admin I don't see anything like that.

TommyCpp:
> > > @TommyCpp out of curiosity has there been the thought to turn off the requirement of being up to the latest before merging? I know there's benefits of both ways. 🤷🏼‍♂️ Just seems like a lot of unnecessary merge commits :/
> > 
> > 
> > I agree but AFAIK it's an org-level rule on protected branches. I don't think maintainers can change it in settings disappointed
> 
> I'm not aware of any such possibilities. I know you can on repos, but from the orgs that I admin I don't see anything like that.

From [here](https://stackoverflow.com/questions/67332877/how-to-setup-this-branch-is-out-of-date-with-the-base-branch-in-github-repo) I understand the reason for such requirement is because the `main` branch is protected and someone enabled additional checks on it. I took a look at the settings of the repo on my end and I don't think I can change the repo branch settings. From what I understand the maintainers of open telemetry repo don't have full access to repo settings.

hdost:
Is the something that @bogdandrutu or someone could help with?

opentelemetry-jaeger/src/exporter/mod.rs
```diff
@@ -110,6 +115,10 @@ impl trace::SpanExporter for Exporter {
 #[derive(Debug)]
 pub struct PipelineBuilder {
     agent_endpoint: Vec<net::SocketAddr>,
+    // There are many variations in which it's read unclear which is causing not to be.
```
TommyCpp, line 118: Can we add a to-do item here to address this when we clean up the features in this creates? I think maybe splitting the `wasm` collector into a different mod would help. Also, make those `*_collector_client` features additive. Meaning they will not override each other. And then we can probably have less complex feature combinations and a better understanding of why the linter complains

TommyCpp (approved):
LGTM otherwise

opentelemetry-jaeger/src/exporter/mod.rs
hdost: Created #736

hdost:
> > > > @TommyCpp out of curiosity has there been the thought to turn off the requirement of being up to the latest before merging? I know there's benefits of both ways. 🤷🏼‍♂️ Just seems like a lot of unnecessary merge commits :/
> 
> > > 
> 
> > > 
> 
> > > I agree but AFAIK it's an org-level rule on protected branches. I don't think maintainers can change it in settings disappointed
> 
> > 
> 
> > I'm not aware of any such possibilities. I know you can on repos, but from the orgs that I admin I don't see anything like that.
> 
> 
> 
> From [here](https://stackoverflow.com/questions/67332877/how-to-setup-this-branch-is-out-of-date-with-the-base-branch-in-github-repo) I understand the reason for such requirement is because the `main` branch is protected and someone enabled additional checks on it. I took a look at the settings of the repo on my end and I don't think I can change the repo branch settings. From what I understand the maintainers of open telemetry repo don't have full access to repo settings. 
> 
> 
> 
> So it's not an org-level setting you can find in Github but rather the org has some access model so that only a handful of ppl can change the branch settings on every repo within the org.

Was able to get them to update 

https://github.com/open-telemetry/community/issues/979

TommyCpp:
@jtescher any other comments or it's good to merge?

PR #734, opened 2022-02-13
fix(exmaples): actix web should use `TokioCurrentThread` as runtime

TommyCpp:


jtescher (approved):

djc:
It would be useful if there was a bit more context in the comments and/or commit message about why this example should use the current thread runtime.

TommyCpp:
> It would be useful if there was a bit more context in the comments and/or commit message about why this example should use the current thread runtime.

Sorry I somehow missed this before merging 😞 . I will add some comments in the example to explain this. But in general actix web is running on tokio single thread scheduler so the runtime has to be TokioCurrentThread

PR #735, opened 2022-02-14
Split out `opentelemetry-api` and `opentelemetry-sdk`

### LINKED ISSUE #721: Separate SDK into its own package
Based on the Usage Guidelines in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/library-guidelines.md does it make sense to split the SDK out into its own crate? 
I feel like with the current implementation it kinda already achieves the noop and since unused code won't be in the final binary I suppose it doesn't matter all that much.

jtescher:
* Split out `opentelemetry` crate into api and sdk crates, and re-export to maintain existing compatibility.
* Move `util` module to sdk as it was internal and not documented publicly.
* Remove api doc tests for now that relied on the sdk.
* Switch default text map propagator to be a noop (allowed by the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.8.0/specification/context/api-propagators.md#global-propagators))
* Move `ExportError` and `InstrumentationLibrary` to API.
* Move `SamplingResult` and `SamplingDecision` to trace API.

Closes #721


hdost (approved):
Seems like it still got some here and there fixes but 👍🏼  very nice

opentelemetry-api/src/lib.rs
```diff
@@ -0,0 +1,68 @@
+//! OpenTelemetry provides a single set of APIs, libraries, agents, and collector
+//! services to capture distributed traces and metrics from your application. You
+//! can analyze them using [Prometheus], [Jaeger], and other observability tools.
+//!
+//! *Compiler support: [requires `rustc` 1.46+][msrv]*
```
hdost, line 5:
```suggestion
//! *Compiler support: [requires `rustc` 1.46+][msrv]*
//! [Prometheus]: https://prometheus.io
//! [Jaeger]: https://www.jaegertracing.io
//! [msrv]: #supported-rust-versions
```

opentelemetry-api/src/baggage.rs
```diff
@@ -11,39 +11,6 @@
...
-//! let cx_with_additions = cx.with_baggage(vec![Key::new("server_id").i64(42)]);
-//!
-//! // Inject baggage into http request
-//! propagator.inject_context(&cx_with_additions, &mut headers);
-//!
-//! let header_value = headers.get("baggage").expect("header is injected");
-//! assert!(header_value.contains("user_id=1"), "still contains previous name-value");
-//! assert!(header_value.contains("server_id=42"), "contains new name-value pair");
-//! # }
 //! ```
```
hdost, line 14:
```suggestion
```

hdost (changes requested):

TommyCpp (commented):
A few questions about where to put the items but overall looks good :+1: This is a good start and we can address those issues in follow-up PR as this is already a big PR.

opentelemetry-api/src/trace/mod.rs
```diff
@@ -311,8 +288,12 @@ impl Event {
...
 pub struct Link {
     span_context: SpanContext,
-    pub(crate) attributes: Vec<KeyValue>,
-    pub(crate) dropped_attributes_count: u32,
+
+    /// Attributes describing this link
+    pub attributes: Vec<KeyValue>,
+
+    /// The number of attributes that were above the limit, and thus dropped.
+    pub dropped_attributes_count: u32,
```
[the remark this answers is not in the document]
jtescher: Yeah wanted to get minimal diff first, can do cleanup PRs subsequently

TommyCpp (approved):
LGTM. I think we can merge once the two issue of comments above resolved.

PR #737, opened 2022-02-15
Pull sampling probability from env var

jbtrystram:
Pull sampling probability from env var `OTEL_TRACES_SAMPLER_ARG` in the default sdk config
I could not find an enum with the expected values strings, so I hard coded them in there, if I missed the file please let me know !

Tho whomever may review this : The spec states the following : 
> For variables accepting an enum value, if the user provides a value the SDK does not recognize, the SDK MUST generate a warning and gracefully ignore the setting.

this why I added some calls to `log` but it looks like no logger is initialized. Should I add a logger to the crate ? Thanks for your guidance

Signed-off-by: jbtrystram <jbtrystram@redhat.com>

### DIFF
--- a/opentelemetry-sdk/src/trace/config.rs
+++ b/opentelemetry-sdk/src/trace/config.rs
@@ -3,6 +3,7 @@
 //! Configuration represents the global tracing configuration, overrides
 //! can be set for the default OpenTelemetry limits and Sampler.
 use crate::trace::{span_limit::SpanLimits, Sampler, ShouldSample};
+use opentelemetry_api::global::{handle_error, Error};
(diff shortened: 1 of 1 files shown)


TommyCpp:
Hi :wave: Thanks for the contribution. We are currently avoiding using `log` directly as open telemetry includes log support so at some point this repo will implement the `log` facade. Usually, we just use `eprintln!` for warning or logging.

jbtrystram:
@TommyCpp  thanks for the tips. This should be good but i can't run the CI !

jtescher (approved):

Issue #711, opened 2022-01-24
Add `schema_url` in `Resource`


TommyCpp:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1692 for details. We will need to 

- [x] Add `schema_url` to resource, 
- [x] Make sure the `schema_url` in OTLP is assigned accordingly, this applies to all struct with `schema_url`
- [x] Set the `schema_url` as needed in exporters we maintains
- [x] Make the `schema_url` optional for meters
- [x] Add `schema_url` to tracer. #743 

[OTEP 152](https://github.com/open-telemetry/oteps/blob/main/text/0152-telemetry-schemas.md) provides more context on what is schema and why we need it.

PR #743, opened 2022-02-22
feat(otlp,sdk): add schema_url to tracer.

TommyCpp:
part of #711 

We have already introduced the `schema_url` parameters when we created the tracer. This PR pass it to the protos when exporting in OTLP format. See https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L71-L72


jtescher:
@TommyCpp should we add this to the [`InstrumentationLibrary`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-api/src/common.rs#L270) struct directly?

TommyCpp:
> @TommyCpp should we add this to the [`InstrumentationLibrary`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-api/src/common.rs#L270) struct directly?

Good point. But I think semantically it isn't used to describe the instrumentation library. Also, we will need to add it to `Resource` which doesn't have an instrumentation library so I feel it may be wired if we have `schema_url` as part of the instrumentation library in tracer but as a stand-alone attribute in `Resource`. 

Maybe we can remove the instrumentation library as a struct and just use `instrumentation_name` and `instrumentation_version`? I can open a separate PR if that sounds more elegant.

jtescher:
@TommyCpp other impls seem to add it to both `InstrumentationLibrary` as well as `Resource`, that's probably the simplest path. e.g. https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/resources/ResourceBuilder.java#L21-L22 and https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/resource/resource.go#L35-L36

opentelemetry-proto/src/transform/traces.rs
```diff
@@ -58,8 +58,13 @@ pub mod tonic {
...
                 schema_url: "".to_string(), // todo: replace with actual schema url.
                 instrumentation_library_spans: vec![InstrumentationLibrarySpans {
-                    instrumentation_library: Default::default(),
-                    schema_url: "".to_string(), // todo: replace with actual schema url.
+                    schema_url: source_span
+                        .instrumentation_lib
+                        .schema_url
+                        .as_ref()
+                        .map(ToString::to_string)
+                        .unwrap_or_else(|| "".to_string()),
```
jtescher, line 66: Nit

```suggestion
                        .unwrap_or_default(),
```

opentelemetry-proto/src/transform/traces.rs
```diff
@@ -172,7 +177,15 @@ pub mod grpcio {
                 })),
                 instrumentation_library_spans: RepeatedField::from_vec(vec![
                     InstrumentationLibrarySpans {
-                        instrumentation_library: Default::default(),
+                        schema_url: source_span
+                            .instrumentation_lib
+                            .schema_url
+                            .as_ref()
+                            .map(ToString::to_string)
+                            .unwrap_or_else(|| "".to_string()),
```
jtescher, line 185:
```suggestion
                            .unwrap_or_default(),
```

jtescher (approved):

PR #744, opened 2022-02-22
feat(stackdriver): Added mappings from OTel attributes to Google Cloud Traces

carlosroman:
When sending data to Google Cloud Traces, for OTel attributes to appear in the console correctly, the exporter must map them to a particular `project.traces` [label](https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces). The code is similar to what Google's own Golang OTel exporter does [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/trace/trace_proto.go#L234).

The PR also checks to ensure that label keys are less than 128 bytes long as GCP will reject those.

### DIFF
--- a/opentelemetry-stackdriver/src/lib.rs
+++ b/opentelemetry-stackdriver/src/lib.rs
@@ -24,13 +24,17 @@ use async_trait::async_trait;
 use futures::stream::StreamExt;
 use opentelemetry::{
     global::handle_error,
-    sdk::export::{
-        trace::{ExportResult, SpanData, SpanExporter},
-        ExportError,
+    sdk::{
+        export::{
+            trace::{ExportResult, SpanData, SpanExporter},
+            ExportError,
+        },
+        trace::EvictedHashMap,
     },
     trace::TraceError,
     Value,
 };
+use opentelemetry_semantic_conventions as semcov;
 use thiserror::Error;
 #[cfg(any(feature = "yup-authorizer", feature = "gcp_auth"))]
 use tonic::metadata::MetadataValue;
@@ -240,12 +244,6 @@ where
         let mut entries = Vec::new();
         let mut spans = Vec::with_capacity(batch.len());
         for span in batch {
-            let attribute_map = span
-                .attributes
-                .into_iter()
-                .map(|(key, value)| (key.as_str().to_owned(), value.into()))
-                .collect();
-
             let trace_id = hex::encode(span.span_context.trace_id().to_bytes());
             let span_id = hex::encode(span.span_context.span_id().to_bytes());
             let time_event = match &self.log_client {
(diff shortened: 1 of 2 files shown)


opentelemetry-stackdriver/src/lib.rs
```diff
@@ -657,3 +652,223 @@ pub enum MonitoredResource {
 
 const TRACE_APPEND: &str = "https://www.googleapis.com/auth/trace.append";
 const LOGGING_WRITE: &str = "https://www.googleapis.com/auth/logging.write";
+const HTTP_PATH_ATTRIBUTE: &str = "http.path";
+use lazy_static::lazy_static;
+
+lazy_static! {
```
djc, line 658: Unless you have actual benchmark data showing that this helps, I'd prefer to just have these as `const &str`s, with the `From` impl calling `to_owned()` on those.

opentelemetry-stackdriver/src/lib.rs
```diff
@@ -657,3 +652,223 @@ pub enum MonitoredResource {
...
+            dropped_attributes_count,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use opentelemetry::{sdk::trace, KeyValue, Value};
```
djc, line 732: I think it would be more idiomatic to import `EvictedHashMap` directly here.

opentelemetry-stackdriver/src/lib.rs
```diff
@@ -657,3 +652,223 @@ pub enum MonitoredResource {
...
+        assert_eq!(actual.attribute_map.len(), 1);
+        assert_eq!(actual.dropped_attributes_count, 0);
+        assert_attribute_map_contains(
+            &actual,
+            "/http/path",
+            AttributeValue::from(Value::String("/path/12314/?q=ddds#123".into())),
+        );
+    }
+
+    fn assert_attribute_map_contains(
```
djc, line 837: I'd prefer to just replace this with the inner `assert_eq!()` call. It doesn't seem useful to separately check whether the value is in there before asserting that the `get()` is `Some()`?

djc (commented):
Thanks for working on this, this seems useful!

I have mostly stylistic concerns below. I might do a style pass after this is merged for that, though.

opentelemetry-stackdriver/src/lib.rs
carlosroman: Give me a couple of hours and I'll push a change with `const &str`. I preferred it that way but wasn't sure as saw a lot of `const` in other exporters for fields like this. Also, the change means I can use a `match` rather than the horrible if statement I've got.

carlosroman: Ha, yeah! I just reverted my change and remembered why I wasn't using a matcher 😅

carlosroman:
> Thanks for working on this, this seems useful!
> 
> I have mostly stylistic concerns below. I might do a style pass after this is merged for that, though.

I've pushed some changes to the PR. Hopefully, they will address your feedback.

djc (approved):

djc:
Thanks!

Issue #736, opened 2022-02-15
Split `wasm` collector from jaeger standard collector. 


hdost:
As mentioned https://github.com/open-telemetry/opentelemetry-rust/pull/729#discussion_r805450667
We may want to look at separating the wasm implementation since it looks very different.

TommyCpp:
We probably still want to reuse the part where we parse the open telemetry spans to jaeger spans but we probably need to look at how to better support wasm overall. It's not urgent but I do think it's a good opportunity.

hdost:
Yea I was kinda curious if maybe there's a common crate (potentially the existing one) and then the wasm crate just has it as a dependency which does special wasm things.

TommyCpp:
We have something in the API for system clock in open telemetry create. Although I don't think it currently working. https://github.com/open-telemetry/opentelemetry-rust/issues/696 need to further investigate it.

PR #748, opened 2022-03-08
feat(jaeger): better configuration pipeline.

TommyCpp:
I think the jaeger configuration pipeline has been a little hard to understand between agent's configurations and collector's configurations. This PR:

- Separate agent pipeline and collector pipeline. it's now `new_agent_pipeline` and `new_collector_pipeline`. Note that this means if the collector exporter to start we will not fall back to agent exporter.
- Add `AgentPipeline` and `CollectorPipeline` to configure agent exporter and collector exporter.
- Add agent only configurations(`auto_batches_split`, `max_packet_size`)
- Add collector only configurations(4 build-in http clients, timeout, username, password and endpoint)
- Rename `with_agent_endpoint` to `with_endpoint` as agent pipeline and collector pipeline are separate now. Similar to collector.
- Add `Configurable` trait to include common attributes shared by agent pipeline and collector pipeline.
- Removed `with_tag` method.
- Make build-in HTTP client features(`surf_collector_client`, `isahc_collector_client`, `reqwest_collector_client`, `reqwest_collector_blocking_client`) additive. 
- Make `collector_client` and `wasm_collector_client` additive. Now users can start a wasm collector using `wasm_collector_client`. 
- Add a `ConfigError` for all error from the builder. Most of them are invalid value like string that cannot convert to URI
- Update the docs

Overall we seprate the collector configurations from agent configurations to prevent users set unrelevant configurations. For example,
```
let tracer = new_pipeline()
                 .with_agent_endpoint(...) // This confiugration will not be used if the collector is built successfully
                 .with_collector_pipeline(...)
                 .install_simple()
```

The downside is now if the collector fails to build, we no longer fall back to agent.

BREAKING CHANGES: 
As we separate the agent pipeline and collector pipeline. The user must understand which one they are using and change `new_pipeline` to `new_collector_pipeline` or `new_agent_pipeline`. 

Functions that contains `agent` or `collector` will no longer contains it as the pipeline are different. For example, `with_collector_username` will now be `with_username`.

For users using build in http client(*isahc*, *reqwest*, *reqwest-blocking* or *surf*). They need to add a new function to pipeline that explicitly point out the http client they want to use.

#736


TommyCpp:
I think I have a way to resolve the msrv error. But it will expose `HasRequiredConfig` trait. Maybe we should just bump the MSRV?

djc:
What would we need to bump it to?

TommyCpp:
Dig a little bit and looks like the error was relaxed by https://github.com/rust-lang/rust/pull/90586/. 
Since we cannot bring the MSRV to 1.59. I will implement a medication as follows:
```rust
trait PrivateTrait{}

pub trait HasRequiredConfig: PrivateTrait{
}
```
This should prevent users from implement the `HasRequiredConfig` trait

opentelemetry-jaeger/src/exporter/config/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+///                       sdk::trace::Config::default()
+///                         .with_resource(Resource::new(
+///                             vec![KeyValue::new("service.name", "my-service")])
+///                         )
+///                 );
+/// ```
+///
+/// [`CollectorPipeline`]: crate::config::collector::CollectorPipeline
+/// [`AgentPipeline`]: crate::config::agent::AgentPipeline
+pub trait Configurable: Sized {
```
jtescher, line 80: Hm is there a way to not require the end user to use this trait?

TommyCpp: We can delete this trait and just copy paste the method into the agent pipeline and collector pipeline? We should probably still keep the `HasRequiredConfig` trait to make sure the two pipeline has the common configurations we want.

jtescher: Yeah copying the methods probably has enough of a developer experience benefit that it's worth the maintenance cost

jtescher (approved):

PR #750, opened 2022-03-08
chore: reduce Tokio feature demands

### LINKED ISSUE #747: Otlp requires `full` feature from Tokio
The `opentelemetry-otlp`, `opentelemetry-aws`, `opentelemetry-prometheus`, and `opentelemetry-zpages` crates in this repository consistently require `full` features from Tokio and Hyper, even if they are not needed by those crates. Only `otlp` specifies this requirement as a non-dev-dependency.

Using `full` in libraries results in added compile times for those crates and binaries that do not require these features, and goes counter to the reason that the `tokio` and `hyper` crates expose these features. Instead, each crate should only be requesting the `tokio` or `hyper` features that it actually requires.

Based on my initial inspection:
* `otlp` only needs `sync` and `rt`, and for testing needs `macros` and `rt-multi-threaded`. This is the main one which unexpectedly bubbled up `full` to my application.
* `aws` only needs `macros` and `rt` for testing.
* `prometheus` claims to need `hyper` and `tokio` as dev dependencies, but I don't see them referred to in that module.
* `zpages` seems to only need `tokio` as a dev dependency, but should be able to get away with `macros` and `rt`.

There are a variety of examples that also declare that they need "full" options, but these may not be as much of an issue except insofar as they make it harder to determine the actual minimal set of features required.

neoeinstein:
This also makes a few small fixes to being able to compile some of the crates in isolation in testing contexts. The four primary crates altered were isolated by commenting out every other crate in the workspace `Cargo.toml` and ensuring that `cargo build` and `cargo test` succeeded. This indicates that the features specified by that crate and its dependencies were sufficient.

In doing this, I noted that some of the `testing` features were not being requested by the candidate crates, so I threaded through the appropriate feature demands to `opentelemetry`, `opentelemetry-api`, and `opentelemetry-sdk`.

I then uncommented everything and ensured that it all compiles and tests too. There I discovered that the `actix-http` example was failing. Changing the `beta` dependency versions to the release versions cleared those errors for me.

Fixes #747

### DIFF
--- a/opentelemetry-aws/Cargo.toml
+++ b/opentelemetry-aws/Cargo.toml
@@ -28,5 +28,5 @@ lazy_static = "1.4"
 [dev-dependencies]
 opentelemetry = { path = "../opentelemetry", features = ["trace", "testing"] }
 opentelemetry-http = { path = "../opentelemetry-http" }
-hyper = { version = "0.14", features = ["full"] }
-tokio = { version = "1.0", features = ["full"] }
+hyper = { version = "0.14" }
+tokio = { version = "1.0", features = ["macros", "rt"] }
--- a/examples/actix-http/Cargo.toml
+++ b/examples/actix-http/Cargo.toml
@@ -8,7 +8,7 @@ publish = false
 opentelemetry = { path = "../../opentelemetry", features = ["rt-tokio"] }
 opentelemetry-jaeger = { path = "../../opentelemetry-jaeger", features = ["reqwest_collector_client", "rt-tokio-current-thread"] }
 thrift = "0.13"
-actix-web = "4.0.0-beta.4"
-actix-service = "2.0.0-beta.5"
+actix-web = "4.0.0"
+actix-service = "2.0.0"
 env_logger = "0.8.2"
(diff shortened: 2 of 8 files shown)


opentelemetry-aws/Cargo.toml
```diff
@@ -28,5 +28,5 @@ lazy_static = "1.4"
 [dev-dependencies]
 opentelemetry = { path = "../opentelemetry", features = ["trace", "testing"] }
 opentelemetry-http = { path = "../opentelemetry-http" }
-hyper = { version = "0.14", features = ["full"] }
-tokio = { version = "1.0", features = ["full"] }
+hyper = { version = "0.14" }
```
djc, line 31: Should we explicitly do `default-features = false` here, and for tokio?

djc (approved):

opentelemetry-aws/Cargo.toml
neoeinstein: We could, but I'm not sure that it is very important to do so, as this is just for dev-dependencies, so generally doesn't pollute dependant crates.

PR #755, opened 2022-03-12
api: make `TraceStateError` priviate

jtescher:
This change simplifies the error handling in the `trace` module by making `TraceStateError` private, and exposing these errors as the more general `TraceError`.


TommyCpp (approved):
:+1: Looks good. I wonder should we add some more standard attributes to those errors to make the logs looks more structured?

jtescher:
Sure, could do in a follow up.

PR #756, opened 2022-03-12
api: rename `Span::record_exception` to `Span::record_error`

jtescher:
This change aligns the error recording methods with Rust's naming conventions which prefer the term `Error` over `Exception`.

This also removes `Span::record_exception_with_stacktrace` as Rust's backtrace functionality is not yet stable. Users who wish to record this data can use `Span::add_event` directly.

### DIFF
--- a/opentelemetry-api/src/trace/context.rs
+++ b/opentelemetry-api/src/trace/context.rs
@@ -60,17 +60,14 @@ impl SpanRef<'_> {
         self.with_inner_mut(|inner| inner.add_event(name, attributes))
     }
 
-    /// Record an exception event
-    pub fn record_exception(&self, err: &dyn Error) {
-        self.with_inner_mut(|inner| inner.record_exception(err))
-    }
-
-    /// Record an exception event with stacktrace
-    pub fn record_exception_with_stacktrace<T>(&self, err: &dyn Error, stacktrace: T)
-    where
-        T: Into<Cow<'static, str>>,
-    {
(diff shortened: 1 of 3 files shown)


opentelemetry-api/src/trace/context.rs
```diff
@@ -60,17 +60,14 @@ impl SpanRef<'_> {
...
-
-    /// Record an exception event with stacktrace
-    pub fn record_exception_with_stacktrace<T>(&self, err: &dyn Error, stacktrace: T)
-    where
-        T: Into<Cow<'static, str>>,
-    {
-        self.with_inner_mut(|inner| inner.record_exception_with_stacktrace(err, stacktrace))
+    /// Record an error as an event for this span.
+    ///
+    /// An additional call to [Span::set_status] is required if the status of the
```
djc, line 65: Instead of calling this out in the docs, why don't we just fix it?

djc (approved):

opentelemetry-api/src/trace/context.rs
jtescher: It appears they are intended to be able to be called independently, I'd guess to support recording errors that are caught or otherwise not resulting in an error state for the trace. See the [Java impl](https://github.com/open-telemetry/opentelemetry-java/blob/286383b218567ba64af8b9b5be6d6b7a81c0e829/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RecordEventsReadableSpan.java#L396-L421) and the [go impl](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L404-L422).

PR #757, opened 2022-03-12
api: unify `Event` and `Link` access patterns

jtescher:
This makes the data in span `Event`s and `Link`s `pub`, but guards against future additions by making the structs `non_exhaustive`.


djc (approved):
I like it. Was there a particular motivation for doing this now?

jtescher:
Trying to get all the api changes needed before 1.0 release

PR #758, opened 2022-03-12
api: move `SpanKind` display format impl to jaeger

jtescher:
This format implementation is jaeger specific and not defined in the otel spec.


TommyCpp:
Is there a downside if we keep this in the APIs? I feel like this can be an implementation detail to help ppl build their implementation using the APIs.

djc:
I don't think we should keep this, it seems pretty arbitrary and the Debug impl should be enough for most people.

djc (approved):

TommyCpp (approved):

PR #760, opened 2022-03-13
api: replace `StatusCode` and message with `Status`

jtescher:
The spec defines status as unset, ok, or error with a description. The current api allows for confusion and misuse by providing invalid combinations (e.g. `Ok` with a description). This change simplifies the api and removes these illegal states by moving the status to be an enum where only the `Status::Error` case accepts a message.


TommyCpp (approved):
:+1:

PR #763, opened 2022-03-18
docs: link to the otlp crate doc for kitchen sink

diezguerra:
Replace link to the kitchen sink example with the local OTLP one, as it was pointing to the Jaeger one instead since https://github.com/open-telemetry/opentelemetry-rust/pull/618 was merged.


TommyCpp (approved):
Good catch! Thanks

PR #768, opened 2022-03-22
Update opentelemetry-proto Protobuf definitions to v0.14.0

vibhavp:
Except for removing `labels` from metrics, I don't believe this adds any breaking changes.


TommyCpp (approved):
Looks nice 👍  Thanks

PR #769, opened 2022-03-26
Update sdk trace docs

jtescher:
* Only export instrumentation library once
* Update resource mod docs
* Add `TracerProvider::force_flush` docs


TommyCpp (approved):
:+1:

TommyCpp:
I think the new nightly(1.61) has regression which breaks the rings. I will ignore the coverage error and merge the PRs for now. 

https://github.com/rust-lang/rust/issues/95267

PR #770, opened 2022-03-27
feat(datadog): allow custom mapping in datadog.

TommyCpp:
There is no official mapping between opentelemetry spans and datadog spans so in some cases our default mapping may not be ideal. 

This PR introduct three functions to allow users to override the mappings for `name`, `resource` and `service_name` fields in datadog spans.

cc @hamiltop


jtescher (approved):

hamiltop (commented):
This is great! Thanks so much!

PR #775, opened 2022-04-07
feat(common): add schema_url to resource.

TommyCpp:
Part of #711.

Note that when merging two resources:
> If the old resource's Schema URL is empty then the resulting resource's Schema URL will be set to the Schema URL of the updating resource,
Else if the updating resource's Schema URL is empty then the resulting resource's Schema URL will be set to the Schema URL of the old resource,
Else if the Schema URLs of the old and updating resources are the same then that will be the Schema URL of the resulting resource,
Else this is a merging error (this is the case when the Schema URL of the old and updating resources are not empty and are different). The resulting resource is undefined, and its contents are implementation-specific.

For the last cases, merging error. The current implementation set the `schema_url` of the result to `None`.

Block on #776

### DIFF
--- a/opentelemetry-sdk/src/resource/mod.rs
+++ b/opentelemetry-sdk/src/resource/mod.rs
@@ -159,14 +159,11 @@ impl Resource {
         resource
     }
 
-    /// Return the [schema url] of the resource. If the resource does not have a schema url, return empty string.
+    /// Return the [schema url] of the resource. If the resource does not have a schema url, return `None`.
     ///
     /// [schema url]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
-    pub fn schema_url(&self) -> String {
-        self.schema_url
-            .clone()
-            .map(|s| s.to_string())
-            .unwrap_or_default()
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -109,9 +141,34 @@ impl Resource {
...
+            }
+        }
+
         resource
     }
 
+    /// Return the [schema url] of the resource. If the resource does not have a schema url, return empty string.
+    ///
+    /// [schema url]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
+    pub fn schema_url(&self) -> String {
```
jtescher, line 165: Would it make sense to have this be `Option<String>`? or maybe more optimally `Option<&str>`? schema_url being a valid URL or `None` may be less error prone than having it be a valid URL or an empty string.

TommyCpp: Sounds reasonable :+1: Updated

jtescher: Cool, could we get `&str` here? or does that cause issues

jtescher (approved):

PR #779, opened 2022-04-10
Prepare for v0.18.0 release

jtescher:
See [VERSIONING.md](https://github.com/open-telemetry/opentelemetry-rust/blob/main/VERSIONING.md) for what is being stabilized.


djc:
Hm, not sure if this is a good idea. The way Cargo version resolution, it considers pre-releases semver-compatible with other pre-releases for the same release. Thus, if you release this and there would later be a 1.0.0-beta.2 with any semver-incompatible changes in it, you would break all downstream users.

jtescher:
@djc good point, is there a better beta convention?

TommyCpp:
+1 on release using 0.x version and adding notes to communicate this is a beta 1.0 release. Also, I don't think the VERSIONING.md is clear about what we are stabilizing(should only be the `trace` feature). Maybe need to add that to the release note?

TommyCpp:
Do we want to wait for #781 as it has an API change?

djc:
> Do we want to wait for #781 as it has an API change?

I think that would make sense.

jwilm:
Any chance this will land soon? We are eagerly awaiting the new version. :smile:

jtescher:
Should be soon hopefully, just have to resolve the MSRV issues on main

jtescher:
@open-telemetry/rust-approvers should we wait for the current open PRs to land? or merge this sooner than that.

TommyCpp:
I think we should include #833 as it includes a fairly important fix for sampling IMO

djc (commented):
LGTM, 🚢!

jtescher:
@djc need a 👍 to merge 😄

TommyCpp (approved):
🛳️

PR #781, opened 2022-04-14
Support concurrent exports

jwilm:
## Problem

Applications generating significant span volume can end up dropping data due to the synchronous export step. According to the opentelemetry spec,

> This function [`export()`] will never be called concurrently for the same exporter instance. It can be called again only after the current call returns.

However, it does not place a restriction on concurrent I/O or anything of that nature. There is an [ongoing discussion] about tweaking the language to make this more clear, and it seems there is consensus on concurrent transmissions being OK with the spec.

## Approach

With that in mind, this commit makes the exporters return a future that can be spawned concurrently. Unfortunately, this means that the `export()` method can no longer be async while taking &mut self. The latter is desirable to enforce the no concurrent calls line of the spec, so the choice is made here to return a future instead with the lifetime decoupled from self. This resulted in a bit of additional verbosity, but for the most part the async code can still be shoved into an async fn for the ergonomics.

The jaeger exporter is left untouched for now as the changes here might take a couple of different approaches (decouple exporter via a channel, or a bigger refactor to fix the &mut borrows deep in the exporter logic).

## Discussion

I'm opening this PR as a draft to start some discussion with the OpenTelemetry-Rust maintainers. To kick this off, I have results to share from a particularly busy application of ours which has been dropping tons of spans due to reaching queue capacity:

image

This is a 2 day period. To the left, we have millions of spans dropped because the queue was not being flushed fast enough. To the right, there are nearly 0 occurrences of dropped spans due to this issue.

For us, this is a massive improvement.

A couple of questions for the maintainers:

1. Is this sort of feature interesting to the project? I recognize that not everyone will need this, but at least for us, it is quite important (to the point we would likely maintain a patch to keep this functionality).
2. Assuming the feature is interesting for the project, how do you feel about the overall approach?
3. Finally, I could use some input on the jaeger side if there is interest in landing this patch. This exporter is quite a bit more complex and relies heavily on &mut borrows several layers down.

## Remaining Work

- [x] Join spawned tasks on `shutdown()`
- [x] Configuration for number of spawned export tasks `export_concurrency`
- [x] Limit spawned export tasks to `export_concurrency`
- [x] Synchronous behavior when `export_concurrency: 1`
- [x] Attempt to drain queue on `force_flush`
- [x] Jaeger exporter
- [ ] Clean up TODOs

[ongoing discussion]: https://github.com/open-telemetry/opentelemetry-specification/issues/2434


TommyCpp:
Thanks for the feedback:+1:, I agree with the approach in principle. We interpreted the spec as BatchSpanProcessor must finish sending one batch span before starting another.

My concern is, like others pointed out in the spec issue, without a bound it can cause an OOM. We can add a parameter to config the number of concurrent sending tasks. If the parameter is set to 1 then disable the concurrent sending. 

Another issue is based on the current model, the `shutdown` function will return after the last batch is exported. If we are exporting multiple batches concurrently we need to make sure they all finished exporting before returning from `shutdown`. Similar to `force_flush` function(We may need to define whether flush means here. Is it starting the exporting or finishing the exporting?)

Jaeger is a little complex. I think the &mut borrow is caused by `emit_batch` method of generated thrift agent. We may need to propagate futures from uploaders. Need to take a closer look to figure out what's the best approach here. 

PS. For some reason, your link to the spec issue is not working. https://github.com/open-telemetry/opentelemetry-specification/issues/2434.

jwilm:
Thanks for the quick reply!

> My concern is, like others pointed out in the spec issue, without a bound it can cause an OOM. We can add a parameter to config the number of concurrent sending tasks. If the parameter is set to 1 then disable the concurrent sending.

Makes sense; this would be fairly straightforward to add.

> Another issue is based on the current model, the `shutdown` function will return after the last batch is exported. If we are exporting multiple batches concurrently we need to make sure they all finished exporting before returning from `shutdown`. 

The `shutdown` "joining" logic makes sense; a bit embarrassed I didn't consider that up front. Also should be relatively simple to add.

> Similar to `force_flush` function(We may need to define whether flush means here. Is it starting the exporting or finishing the exporting?)

My 2c: in the non-concurrent model it meant starting a request up to the batch size. In the concurrent model, it's still possible to have pending items in queue up to the timeout; to me it still makes sense to consider flushing the start of an export.

> Jaeger is a little complex. I think the &mut borrow is caused by `emit_batch` method of generated thrift agent. We may need to propagate futures from uploaders. Need to take a closer look to figure out what's the best approach here.

That's right. The generated thrift agent has some mutable methods for working with the io streams and some sort of frame counter. It's by no means untenable but certainly more work than the other exports (hence the draft for discussion).

> PS. For some reason, your link to the spec issue is not working. [open-telemetry/opentelemetry-specification#2434](https://github.com/open-telemetry/opentelemetry-specification/issues/2434).

Fixed, thank you.

Would it be premature for me to continue with the suggested fixes?

jtescher:
@jwilm these all sound like good improvements to me 👍

TommyCpp:
I think we can implement the proposed changes. One caveat is your branch seems stale and I'd suggest rebase on the latest to include some critical changes like we split `opentelemetry` into `opentelemetry-sdk` and `opentelemetry-api`

jwilm:
I rebased the existing work on main, and I'm starting on the proposed changes (will track on OP).

In the mean time, I would appreciate some input on the Jaeger exporter. I noticed that stackdriver exporter spawns a separate task for the actual export work, and the `SpanExporter` impl simply sends the batch via channel. Perhaps the same approach would be suitable for Jaeger?

jwilm:
I believe all the requested changes have been implemented, save for the Jaeger exporter.

Considering the implementation path there, I think a task queue system could be utilized where the Jaeger exporter will only export a single batch concurrently (regardless of BSP settings). This seems like the path of least resistance to landing this PR, but it also means that the Jaeger exporter doesn't actually gain any concurrency. Perhaps this is OK for now and Jaeger concurrency can be a future enhancement?

Separately, I took a closer look at the stackdriver exporter, and I think it could easily take advantage of the BSP concurrency management instead of rolling its own. Let me know if you'd like that done.

Waiting on your feedback before proceeding.

Thanks!

opentelemetry-jaeger/src/exporter/config/agent.rs
```diff
@@ -228,25 +228,25 @@ impl AgentPipeline {
...
-            self.transformation_config.export_instrument_library,
-            self.build_sync_agent_uploader()?,
-        );
-
-        builder = builder.with_simple_exporter(exporter);
-        builder = builder.with_config(config);
-
-        Ok(builder.build())
-    }
+    // pub fn build_simple(mut self) -> Result<TracerProvider, TraceError> {
```
jwilm, line 231: These build simple / sync methods are the last blocker for the jaeger implementation. In order to support the non-pipelined I/O resources like the Agent uploaders, the Exporter and Uploaders were decoupled via a channel, and the uploaders now run on a dedicated task. However, this meant that `Exporter::new` has to return a task to spawn as well, similar to the stackdriver exporter. This was problematic with this `build_simple`, etc. methods since there's no runtime available on which to spawn that task.

I'll need feedback on whether we can simply thread the Runtime through here and whether this design is acceptable.

jwilm: I updated the commit to thread the runtime through, but I don't have enough context in the project to know if this is the right approach.

TommyCpp: Thanks for the hard work! Sorry didn't get much time yesterday to take a look. Will try to take a look tomorrow or over the weekend. We do want to keep `install_simple` without runtime parameter just so that you can work with opentelemetry without a runtime. But I need a closer look to see it's possible to do it under the new exporter APIs.

TommyCpp:
> Considering the implementation path there, I think a task queue system could be utilized where the Jaeger exporter will only export a single batch concurrently (regardless of BSP settings). This seems like the path of least resistance to landing this PR, but it also means that the Jaeger exporter doesn't actually gain any concurrency. Perhaps this is OK for now and Jaeger concurrency can be a future enhancement?

I think the easiest way is to wrap the upload using mutex? I think the ideal solution is to replace the `Uploader` trait with two enum/trait that separates the thread-safe uploader(which supports concurrent exporting) and non-thread-safe one(which need to export sequentially(. 

> Separately, I took a closer look at the stackdriver exporter, and I think it could easily take advantage of the BSP concurrency management instead of rolling its own. Let me know if you'd like that done.

I think it makes sense. Curious what's your thoughts @djc. Also, we can open a separate PR to address this.

djc:
> Separately, I took a closer look at the stackdriver exporter, and I think it could easily take advantage of the BSP concurrency management instead of rolling its own. Let me know if you'd like that done.

Sounds good, are you actually going to be using stackdriver or are you just cleaning stuff up here? :)

opentelemetry-jaeger/src/exporter/config/agent.rs
jwilm: Is it just that method in particular, or is the goal that the opentelemetry library can be used without a runtime? If it's just that method, we can probably figure something out using the cfg(feature) attrs internally.

jwilm:
> I think the easiest way is to wrap the uploader using mutex?

I considered that, but it has some downsides including loss of ordering of uploads. The task queue seemed cleaner, but it did necessitate spawning a future somehow. I suppose another option is spawning a thread for the non-async/await uploader(s).

> I think the ideal solution is to replace the Uploader trait 

Would it be possible to expose two different exporters from Jaeger? This might open up the design space a bit more.

> Sounds good, are you actually going to be using stackdriver or are you just cleaning stuff up here? :)

Just an offer to clean up :)

TommyCpp:
> but it has some downsides including loss of ordering of uploads

I think if we allow concurrent exporting. We won't be able to guarantee the strict order of spans. i. e multiple requests in flight could finish in arbitrary order. Need to double-check how Jaeger or other backend handle this but I don't think it is a no go. I will report back what I found

opentelemetry-jaeger/src/exporter/config/agent.rs
TommyCpp: I think in general we want to allow people to use it without a runtime. As a util library, we shouldn't assume the runtime and deny certain use cases.

djc:
> Just an offer to clean up :)

Would be great, thanks!

TommyCpp:
> Need to double-check how Jaeger or other backend handle this but I don't think it is a no go. I will report back what I found

Tested around Jaeger and it seems OK for us in ingest spans out of order. I am OK if we spawn a thread or use a mutex now. As long as we don't make an API change we can work on improving the internals in the following PR.

jwilm:
I think there's a path to keeping the task based approach if the only blocker is the install_simple method not taking a runtime. Given that the core library requires some sort of async runtime, we could potentially detect that based on cargo features and insert the runtime internally rather than requiring it as an argument in the public API.

What do you think?
Thanks!

opentelemetry-zipkin/Cargo.toml
```diff
@@ -38,6 +38,7 @@ http = "0.2"
 reqwest = { version = "0.11", optional = true, default-features = false }
 surf = { version = "2.0", optional = true, default-features = false }
 thiserror = { version = "1.0"}
+futures = "0.3"
```
djc, line 41: Can we save on dependencies by using futures-util (or even futures-core) for most of these?

jwilm: With the exception of the current Jaeger exporter implementation, we should be able to get away with using only futures-util. I'll make that change.

jwilm: Ah, and maybe in some cases even futures-core (at least for Zipkin)

jwilm: Done in the latest commit.

jwilm:
This should be ready for another review. The Jaeger API is as it was before, and a thread is used internally to handle the task execution.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -63,6 +67,10 @@ const OTEL_BSP_MAX_EXPORT_BATCH_SIZE_DEFAULT: usize = 512;
 const OTEL_BSP_EXPORT_TIMEOUT: &str = "OTEL_BSP_EXPORT_TIMEOUT";
 /// Default maximum allowed time to export data.
 const OTEL_BSP_EXPORT_TIMEOUT_DEFAULT: u64 = 30_000;
+/// Default max concurrent exports for BSP
+// TODO jwilm: I omitted the OTEL_ prefix here as this is a non-standard config
+// value. Should it be kept this way or prefix with OTEL_ for consistency?
+const BSP_MAX_CONCURRENT_EXPORTS: usize = 0;
```
TommyCpp, line 73: I think it's reasonable to add `OTEL_` prefix and `_DEFAULT` postfix. 

Also may be set to 1 is more reasonable so that the behavior is consistent with the previous users. For user that encounter span loss they can opt-in the feature

scripts/test.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eu
 
-cargo test --all "$@"
+cargo test --all "$@" -- --test-threads=1
```
TommyCpp, line 5: any reason why we need the test thread to be 1 here? Does running them in parallel cause some issues?

TommyCpp (commented):
I think it looks really great and should be a great improvement :+1: . We can look into Jaeger more and optimize the jaeger exporter in a following up PR.

scripts/test.sh
jwilm: Ah yeah, I noticed there were some flakey tests due to modifying the environment and colliding with each other.

opentelemetry-sdk/src/trace/span_processor.rs
jwilm: Done.

scripts/lint.sh
```diff
@@ -29,14 +29,7 @@ if rustup component add clippy; then
...
 
-  cargo_feature opentelemetry-jaeger "surf_collector_client, surf/curl-client"
-  cargo_feature opentelemetry-jaeger "isahc_collector_client"
-  cargo_feature opentelemetry-jaeger "reqwest_blocking_collector_client"
-  cargo_feature opentelemetry-jaeger "reqwest_collector_client"
-  cargo_feature opentelemetry-jaeger "collector_client"
-  cargo_feature opentelemetry-jaeger "wasm_collector_client"
-  cargo_feature opentelemetry-jaeger "collector_client, wasm_collector_client"
-  cargo_feature opentelemetry-jaeger "default"
+  cargo_feature opentelemetry-jaeger "full"
```
TommyCpp, line 32: This is nice to have the `full` feature, but we probably want to keep those separate tests to make sure those feature can also work on their own

TommyCpp (commented):
Nice work :+1: Just one nit and fixing the lint/test I think we are good to merge

scripts/lint.sh
jwilm: Done.

TommyCpp (approved):
LGTM :tada:

TommyCpp:
Any other suggestions or do you think it's good to merge? cc @jtescher @djc

djc:
I don't have bandwidth to review this in detail, sorry.

jwilm:
Anything I can do to help land this?

jwilm:
:tada: thanks for the merge!

Are there plans to publish a new release to crates.io anytime soon? We would love to get off our git dependencies.

TommyCpp:
I hope we can release a new version of 0.18 sometime soon. You can track the process in  #779

PR #783, opened 2022-04-15
opentelemetry-otlp: upgrade to tonic 0.7

benesch:


TommyCpp:
I think you also need to update https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-proto/Cargo.toml#L50 as it provides the generated type of protos

TommyCpp (approved):
Looks good. Thanks

PR #787, opened 2022-04-25
fix(datadog): parse the endpoint to remove tailing slash.

### LINKED ISSUE #782: invalid URL generated from the Datadog agent endpoint
Context: https://github.com/apollographql/router/issues/851

If the URL passed to [with_agent_endpoint](https://docs.rs/opentelemetry-datadog/latest/opentelemetry_datadog/struct.DatadogPipelineBuilder.html#method.with_agent_endpoint) is of the format `http://localhost:8126/` (with a trailing slash), the exporter will generate an URL of the format `http://datadog.default:8126//v0.5/traces` and fail to send traces there.

The issue is that the endpoint and version path are concatenated instead of appending a proper path to an existing URL: http://datadog.default:8126//v0.5/traces

TommyCpp:
close #782

### DIFF
--- a/opentelemetry-stackdriver/Cargo.toml
+++ b/opentelemetry-stackdriver/Cargo.toml
@@ -15,7 +15,7 @@ gcp_auth = { version = "0.7", optional = true }
 hex = "0.4"
 http = "0.2"
 hyper = "0.14.2"
-hyper-rustls = { version = "0.22.1", optional = true }
+hyper-rustls = { version = "0.23", optional = true }
 opentelemetry = { version = "0.17", path = "../opentelemetry" }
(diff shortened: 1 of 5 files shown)


opentelemetry-stackdriver/Cargo.toml
```diff
@@ -15,7 +15,7 @@ gcp_auth = { version = "0.7", optional = true }
 hex = "0.4"
 http = "0.2"
 hyper = "0.14.2"
-hyper-rustls = { version = "0.22.1", optional = true }
+hyper-rustls = { version = "0.23", optional = true }
```
TommyCpp, line 18: `yup-oauth2` updated to `hyper-rustls 0.23` in 6.6

jtescher (approved):

PR #788, opened 2022-04-29
Initial implementation of the logging SDK spec

vibhavp:
This PR is a first stab at implementing the [logging SDK specification](https://github.com/open-telemetry/opentelemetry-specification/blob/76e618535d841451b310f51522996362a6c1ec8f/specification/logs/logging-library-sdk.md). Owing to the similar nature of traces and logging data, it's mostly a `s/span/log/g` on the trace SDK, but without corresponding additions to the `api` package, as the specification's intent is for the SDK to be used by logging libraries, not developers.

There is a bit of duplication, especially with respect to how the batching is implemented, and I'm curious in seeing whether that could be resolved by unifying `TraceRuntime` and `LogRuntime` into a signal-agnostic trait.


jtescher (changes requested):
This is a great step thanks for starting this, the duplication between runtimes is indeed unfortunate, we could look at unifying those in subsequent PRs. Also note the changes discussed in https://github.com/open-telemetry/opentelemetry-rust/pull/781.

opentelemetry-api/src/global/error_handler.rs
```diff
@@ -1,6 +1,8 @@
 use std::sync::PoisonError;
```
TommyCpp, line 1: I think similar to `GlobalTracerProvider`. We need a `GlobalLogEmitterProvider` here

opentelemetry-proto/src/transform/logs.rs
```diff
@@ -0,0 +1,332 @@
...
+            let record = LogRecord {
+                time_unix_nano: log_record.timestamp.map(to_nanos).unwrap_or_default(),
+                observed_time_unix_nano: log_record
+                    .observed_timestamp
+                    .map(to_nanos)
+                    .unwrap_or_default(),
+                severity_number: log_record
+                    .severity_number
+                    .map(SeverityNumber::from)
+                    .map(Into::into)
```
TommyCpp, line 108: Can we implement a conversion between the `serverity_number` type in `log_record` and the grpc generated types here? Two mappings here seem to be unnecessary.

opentelemetry-sdk/src/export/log/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use opentelemetry_api::log::LogError;
+use opentelemetry_api::{log::LogResult, InstrumentationLibrary};
+use std::{fmt::Debug, sync::Arc};
+
+pub mod stdout;
+
+/// `LogExporter` defines the interface that log exporters should implement.
+#[async_trait]
+pub trait LogExporter: Send + Debug {
+    /// Exports a batch of `ResourceLogs`.
```
TommyCpp, line 13:
```suggestion
    /// Exports a batch of `LogData`.
```
nit: I believe the `ResourceLogs` is a concept of the OTLP protocol

opentelemetry-sdk/src/export/log/stdout.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl<W> LogExporter for Exporter<W>
+where
+    W: Write + Debug + Send + 'static,
+{
+    /// Export spans to stdout
```
TommyCpp, line 131:
```suggestion
    /// Export logs to stdout
```

opentelemetry-sdk/src/log/log_emitter.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// Force flush all remaining logs in log processors and return results.
+    pub fn force_flush(&self) -> Vec<LogResult<()>> {
+        self.log_processors()
+            .iter()
+            .map(|processor| processor.force_flush())
+            .collect()
+    }
+
+    /// Shuts down this `LogEmitterProvider`, panicking on failure.
+    pub fn shutdown(&mut self) -> Vec<LogResult<()>> {
```
TommyCpp, line 78: We can implement the `shutdown` on Drop?

opentelemetry-sdk/src/log/log_processor.rs
```diff
@@ -0,0 +1,368 @@
...
+    shutdown: crossbeam_channel::Receiver<()>,
+}
+
+impl SimpleLogProcessor {
+    pub(crate) fn new(mut exporter: Box<dyn LogExporter>) -> Self {
+        let (log_tx, log_rx) = crossbeam_channel::unbounded();
+        let (shutdown_tx, shutdown_rx) = crossbeam_channel::bounded(0);
+
+        let _ = thread::Builder::new()
+            .name("opentelemetry-exporter".to_string())
```
TommyCpp, line 51: maybe worth to change it to `opentelemetry-log-exporter`

opentelemetry-sdk/src/log/log_processor.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    fn force_flush(&self) -> LogResult<()> {
+        let (res_sender, res_receiver) = oneshot::channel();
+        self.message_sender
+            .try_send(BatchMessage::Flush(Some(res_sender)))?;
+
+        futures_executor::block_on(res_receiver)
+            .map_err(|err| LogError::Other(err.into()))
+            .and_then(|identity| identity)
```
TommyCpp, line 132: Curious why we name it `identity`

opentelemetry-sdk/src/log/record.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub fn builder() -> LogRecordBuilder {
+        LogRecordBuilder::new()
+    }
+}
+
+/// TraceContext stores the trace data for logs that have an associated
+/// span.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct TraceContext {
```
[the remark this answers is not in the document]
TommyCpp: I think it refers to [W3C trace context standard](https://github.com/open-telemetry/opentelemetry-specification/blob/76e618535d841451b310f51522996362a6c1ec8f/specification/logs/data-model.md#trace-context-fields). Maybe we can merge it with the definition in `trace` mod?

TommyCpp (commented):
Thanks for working on this I think it's a good start!

opentelemetry-sdk/src/log/record.rs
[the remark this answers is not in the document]
jtescher: Yeah, just confusing as the concept is used multiple places, like the [TraceContextPropagator](https://docs.rs/opentelemetry/latest/opentelemetry/sdk/propagation/struct.TraceContextPropagator.html#) doesn't use this and the [TraceContextExt](https://docs.rs/opentelemetry/latest/opentelemetry/trace/trait.TraceContextExt.html) doesn't involve it either (maybe the ext should be renamed as well to disambiguate?)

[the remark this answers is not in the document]
TommyCpp: Maybe easier to divide the trace context field into trace id, span id, and trace flag and use the basic types. Or alternatively, those fields should only be present when `trace` is enabled. so that we can just refer the types from `trace` mod

opentelemetry-api/Cargo.toml
```diff
@@ -24,3 +24,4 @@ default = ["trace"]
 trace = ["pin-project"]
 metrics = ["fnv"]
 testing = ["trace"]
+log = []
```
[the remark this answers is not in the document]
vibhavp: Yeah, that seems a little less awkward as well. Done.

opentelemetry-proto/tests/grpc_build.rs
```diff
@@ -18,6 +18,8 @@ fn build_grpc() {
             "src/proto/opentelemetry-proto/opentelemetry/proto/collector/trace/v1/trace_service.proto",
             "src/proto/opentelemetry-proto/opentelemetry/proto/metrics/v1/metrics.proto",
             "src/proto/opentelemetry-proto/opentelemetry/proto/collector/metrics/v1/metrics_service.proto",
+	    "src/proto/opentelemetry-proto/opentelemetry/proto/logs/v1/logs.proto",
```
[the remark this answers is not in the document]
vibhavp: Whoops, fixed.

opentelemetry-sdk/src/log/record.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// The original severity string from the source
+    pub severity_text: Option<Cow<'static, str>>,
+    /// The corresponding severity value, normalized
+    pub severity_number: Option<Severity>,
+
+    /// Record body
+    pub body: Option<Any>,
+
+    /// Resource attributes for the entity that produced this record
+    pub resource: Option<BTreeMap<Cow<'static, str>, Any>>,
```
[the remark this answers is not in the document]
vibhavp: From my reading of the [log data model](https://github.com/open-telemetry/opentelemetry-specification/blob/76e618535d841451b310f51522996362a6c1ec8f/specification/logs/data-model.md#field-resource), no. The `Resource` field here seems as if it's set by the logging library, while `LogData`'s `Resource` is set by the `LogEmitter`.

opentelemetry-sdk/src/log/log_emitter.rs
vibhavp: Good idea, added an initial implementation that tries to gracefully handle errors.

opentelemetry-sdk/src/log/log_processor.rs
vibhavp: I believe it's a reference to the [identity function](https://en.wikipedia.org/wiki/Identity_function), even though the parameter is not the mapping in itself. I refactored this to just use `std::convert::identity` instead.

opentelemetry-proto/src/transform/logs.rs
vibhavp: Yup, moved the mapping code to the function(s) itself.

opentelemetry-sdk/src/log/record.rs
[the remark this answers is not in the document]
vibhavp: There is [`SpanContext`](https://docs.rs/opentelemetry/latest/opentelemetry/trace/struct.SpanContext.html#method.span_id), which stores the span ID, trace ID, and trace flags. However, it also stores an additional `TraceState` field, which is not in a Log Record. That's why I had to create this field.

opentelemetry-sdk/src/log/record.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub trace_id: TraceId,
+    /// Span Id
+    pub span_id: SpanId,
+    /// Trace flags
+    pub trace_flags: Option<TraceFlags>,
+}
+
+/// Value types for representing arbitrary values in a log record.
+#[derive(Debug, Clone)]
+pub enum Any {
```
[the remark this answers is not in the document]
vibhavp: What about `LogAny`or `AnyValue`? I initially wanted to reuse `Value`, but both the spec and the Rust implementation mandate that an `Array` be homogenous in its member types.

opentelemetry-api/src/global/error_handler.rs
vibhavp: Yes. However, that would also mean that we would have to define `opentelemetry-api` traits for a `GlobalLogEmitterProvider`, while the spec explicitly leaves out an API, as the SDK is meant to be utilized only by library developers. 

Although it doesn't sound very idiomatic, what do you think about having a `GlobalLogEmitterProvider` in `opentelemetry-sdk`, in the `logs` module?

TommyCpp: Yeah I think that works as the user doesn't suppose to interact with `GlobalLogEmitterProvider`

TmLev:
Hey there!

What's the status of this PR? Maybe some help needed?

hdost:
👋🏼  @vibhavp  thank you again for making the PR, do you mind if we rebase this on main?

cijothomas:
> 👋🏼 @vibhavp thank you again for making the PR, do you mind if we rebase this on main?

@vibhavp Are you okay if one of us take this over? (We'll likely create a new PR copying this)

jtescher:
Do we need this currently? may want to come at it through the logging ecosystems like `tracing` and `log` crates and see what their needs would be rather than implement it here first as there is no logging API.

cijothomas:
> Do we need this currently? may want to come at it through the logging ecosystems like `tracing` and `log` crates and see what their needs would be rather than implement it here first as there is no logging API.

Yes. We (at Microsoft) have several teams needing a logging solution, which integrates with the rest of OpenTelemetry ecosystem. From what I see there are other people interested in this as well. (We discussed this in the last Rust SIG meetings, and we are ready to contribute the code.)

The end-user calling logging API could be the tracing or log crates itself, as OTel does not (yet) specify an end-user callable logging API. This repo only needs the "log-bridge" (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md) part, which allows end-users to continue using log/tracing crates, but logs will be routed through OTel (and gaining correlation with tracing, consistent Resources applied, exporters etc.)

jtescher:
Got it, happy to have a bridge impl as long as it's driven by an actual use 👍

vibhavp:
Hello! Sorry for the late replies, I have updated this branch against `main`. I haven't fully tested the new code yet, though.

opentelemetry-sdk/src/export/logs/stdout.rs
```diff
@@ -0,0 +1,139 @@
+//! # Stdout Log Exporter
+//!
+//! The stdout [`LogExporter`] writes debug printed [`LogRecord`]s to its configured
```
cijothomas, line 3: stdout has been moved to own create https://github.com/open-telemetry/opentelemetry-rust/pull/1027

I think its best for logs to also follow the same approach.

opentelemetry-api/src/logs/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+use futures_channel::{mpsc::TrySendError, oneshot::Canceled};
+use std::time::Duration;
+use thiserror::Error;
+
+/// Describe the result of operations in log SDK.
+pub type LogResult<T> = Result<T, LogError>;
+
+#[derive(Error, Debug)]
+#[non_exhaustive]
+/// Errors returned by the log SDK.
```
cijothomas, line 13: I think this should be in the SDK only. Its probably a mistake that was already in traces api. See https://github.com/open-telemetry/opentelemetry-rust/issues/1042

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn emit(&self, record: LogRecord) {
+        let provider = match self.provider() {
+            Some(provider) => provider,
+            None => return,
+        };
+
+        let config = provider.config();
+        for processor in provider.log_processors() {
+            let data = LogData {
+                record: record.clone(),
```
lalitb, line 252: nit - is this necessary to clone the data, or can we pass it as reference to the processor/exporter. The clone operation is in general extensive.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -0,0 +1,259 @@
...
+            }),
+        }
+    }
+}
+
+#[derive(Debug)]
+/// The object for emitting [`LogRecord`]s.
+///
+/// [`LogRecord`]: crate::log::LogRecord
+pub struct LogEmitter {
```
cijothomas, line 216: The spec name now is `Logger`

cijothomas:
> Hello! Sorry for the late replies, I have updated this branch against `main`. I haven't fully tested the new code yet, though.

Thanks for continuing to work on this! I did a quick review and left some small comments. Also, the spec has changes some names recently - would you update this PR to reflect the new spec names? (Or a follow-up to do rename?)

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -0,0 +1,259 @@
...
+            Some(provider) => provider,
+            None => return,
+        };
+
+        let config = provider.config();
+        for processor in provider.log_processors() {
+            let data = LogData {
+                record: record.clone(),
+                resource: config.resource.clone(),
+                instrumentation: self.instrumentation_lib.clone(),
```
lalitb, line 254: Not sure if I am missing something, but is this code compiling correctly - as `LogData` struct doesn't seem to have `instrumentation` field.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -0,0 +1,298 @@
...
+    pub fn with_span_context(self, span_context: &SpanContext) -> Self {
+        Self {
+            record: LogRecord {
+                trace_context: Some(TraceContext {
+                    span_id: span_context.span_id(),
+                    trace_id: span_context.trace_id(),
+                    trace_flags: Some(span_context.trace_flags()),
+                }),
+                ..self.record
+            },
```
lalitb, line 204: Should we check if SpanContext is valid using `span_context.is_valid()` ?

opentelemetry-sdk/src/logs/runtime.rs
```diff
@@ -0,0 +1,119 @@
...
+#[cfg(feature = "rt-async-std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "rt-async-std")))]
+impl LogRuntime for AsyncStd {
+    type Receiver = async_std::channel::Receiver<BatchMessage>;
+    type Sender = async_std::channel::Sender<BatchMessage>;
+
+    fn batch_message_channel(&self, capacity: usize) -> (Self::Sender, Self::Receiver) {
+        async_std::channel::bounded(capacity)
+    }
+}
```
lalitb, line 119: Probably need to see if we can reuse RunTime trait across both Logs and Traces (as mentioned in some other comments too).

lalitb:
Thanks @vibhavp for updating the PR. Have added few comments mostly related to updating to latest specs, and avoiding code duplication. But this looks to be right direction. And would be fine if even we can merge existing change as base by keeping it behind experimental feature flag, and incrementally fixes the comments, as the PR has been open for long now. Will let existing approvers decide on that.

opentelemetry-api/src/logs/logger.rs
```diff
@@ -0,0 +1,23 @@
...
+pub trait Logger {
+    fn emit(&self, record: LogRecord);
+}
+
+pub trait LoggerProvider {
+    type Logger: Logger;
+
+    fn versioned_logger(
+        &self,
+        name: Cow<'static, str>,
```
lalitb, line 14: This can be: `name: impl Into<Cow<'static, str>>,`
this will allow us to directly pass string literal as `name` argument.

opentelemetry-api/src/logs/logger.rs
```diff
@@ -0,0 +1,23 @@
...
+        version: Option<Cow<'static, str>>,
+        schema_url: Option<Cow<'static, str>>,
+        attributes: Option<Vec<KeyValue>>,
+    ) -> Self::Logger;
+
+    fn logger(&self, name: Cow<'static, str>) -> Self::Logger {
```
lalitb, line 20: This can be: name: impl Into<Cow<'static, str>>,
this will allow us to directly pass string literal as name argument.

lalitb:
Thanks @TommyCpp for fixing lint. As discussed in the community meeting, do you think we can merge this PR, as that will allows others to incrementally contribute the improvements over this. I did some trivial testing using ostream-exporter, and the logs are printed successfully to standard output.

lalitb (approved):

cijothomas (approved):
Thanks!

cijothomas:
> Thanks @TommyCpp for fixing lint. As discussed in the community meeting, do you think we can merge this PR, as that will allows others to incrementally contribute the improvements over this. I did some trivial testing using ostream-exporter, and the logs are printed successfully to standard output.

@jtescher Could you also take another look? There is a request-changes from your earlier review.

jtescher (approved):
Still needs some cleanup and I'd like to see the `Any` type renamed as it conflicts with the type in the standard library, but we can do it follow up changes as well.

opentelemetry-sdk/src/log/record.rs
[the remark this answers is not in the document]
vibhavp: Renamed this to `AnyValue`.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::Resource;
+use async_trait::async_trait;
+use opentelemetry_api::{
+    logs::{LogError, LogRecord, LogResult},
+    InstrumentationLibrary,
+};
+use std::{borrow::Cow, fmt::Debug};
+
+/// `LogExporter` defines the interface that log exporters should implement.
+#[async_trait]
```
Robo210, line 11: I don't think traits that every 3rd-party exporter is expected to implement should be forcing non-standard things like the async-trait crate. It's hacky and unnecessary, and will probably turn into an issue in the future when/if built-in support for async functions in traits shows up.

While the SpanExporter's BoxedFuture is unpleasant, it feels like a more future-proof design and one that is less burdensome on developers.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -0,0 +1,232 @@
...
+                .collect()
+        })
+    }
+}
+
+impl Drop for LoggerProvider {
+    fn drop(&mut self) {
+        match self.try_shutdown() {
+            None => handle_error(Error::Other(
+                "canont shutdown LoggerProvider when child Loggers are still active".into(),
```
Robo210, line 103: cannot

opentelemetry-api/src/logs/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub use logger::{Logger, LoggerProvider};
+pub use noop::NoopLoggerProvider;
+pub use record::{AnyValue, LogRecord, LogRecordBuilder, Severity, TraceContext};
+
+/// Describe the result of operations in log SDK.
+pub type LogResult<T> = Result<T, LogError>;
+
+#[derive(Error, Debug)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
lalitb: Agree, better to fail early at compile time. Probably we can remove non-exhaustive attribute before releasing stable version of logs.

opentelemetry-sdk/src/export/logs/mod.rs
lalitb: Agree, exporter trait needn't be async, and let the processor decide how this exporter would be invoked. We can keep this consistent with SpanExporter - https://github.com/open-telemetry/opentelemetry-rust/blob/69fd972fa63a3cb75f61fa60ffdac210071b93f7/opentelemetry-sdk/src/export/trace.rs#L19

cijothomas:
@TommyCpp @hdost Could we merge this, given major feedbacks are already addressed, and we can iterate on this further.

TommyCpp:
Sounds good! Thank you for working on this @vibhavp and appriciate everyone that reviewed it

PR #790, opened 2022-05-01
Switch to static resource references

jtescher:
The spec suggests all spans should have an associated `Resource`. This change switches trace config and span data from `Option<Arc<Resource>>` to `Cow<'static, Resource>` and removes `Config::with_no_resource` to accommodate this requirement.


TommyCpp (approved):
Looks good :+1: Curious how does the benchmark look like

jtescher:
@TommyCpp I don't think the benchmarks should change all that much as `Arc::clone` is not all that expensive, but this is more spec compliant and a few nanoseconds faster per span 👍

Issue #791, opened 2022-05-04
metrics tag missing from datadog exporter v3


rex-remind101:
Looking at dd-trace we see that `"sampling.priority"` is added as a tag implicitly https://github.com/DataDog/dd-trace-js/blob/3419a4c09584a3109952b7c7da7d84924c36b866/packages/dd-trace/src/format.js#L98

Given that `"sampling.priority"` is represented by a number it should be exported as part of the `metrics` part of the span data as shown here https://github.com/DataDog/dd-trace-js/blob/3419a4c09584a3109952b7c7da7d84924c36b866/packages/dd-trace/src/format.js#L146

However, looking at the datadog exporter v3, there does not appear to be any mention of forwarding `"sampling.priority"` nor mention of `metrics` https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-datadog/src/exporter/model/v03.rs

Is this inserted elsewhere? If not what would be required to add it?

rex-remind101:
fyi, it appears that when the user tags `"sampling.priority"` it gets converted to x
https://github.com/DataDog/dd-trace-js/blob/c89a35f7d27beb4a60165409376e170eacb194c5/packages/dd-trace/src/priority_sampler.js#L106
and
https://github.com/DataDog/dd-trace-js/blob/7b27059fe3b9de6788c0517a2b499ff6b78da5ff/ext/tags.d.ts#L6
and
https://github.com/DataDog/dd-trace-js/blob/c89a35f7d27beb4a60165409376e170eacb194c5/packages/dd-trace/src/constants.js#L4

rex-remind101:
pr https://github.com/open-telemetry/opentelemetry-rust/pull/792

TommyCpp:
Thanks for bringing it up. I think maybe we can just drop the span when exporting instead of encode it and send out.  

Have you tested it against a datadog endpoint? I lost my access to datadog a while ago so I can't test it myself.

rex-remind101:
We need to export even in cases where spans are not sampled because the agent needs to be aware of metrics to properly calculate request/error statistics on datadog's end. The agent will then sampling on its end anyway afaiu.

rex-remind101:
Actually not sure if i'm understanding you, what do you mean by "drop the span when exporting"?

TommyCpp:
For unsampled spans, we don't have to pass it to the span exporter as per the spec. 
[...]

PR #792, opened 2022-05-11
feat(datadog): add sampling priority tag in datadog spans

rex-remind101:
encode sampling priority into metrics object of data dog exported data. data dog uses this to sample.

solves https://github.com/open-telemetry/opentelemetry-rust/issues/791

### DIFF
--- a/opentelemetry-datadog/src/exporter/model/v05.rs
+++ b/opentelemetry-datadog/src/exporter/model/v05.rs
@@ -1,4 +1,5 @@
 use crate::exporter::intern::StringInterner;
+use crate::exporter::model::SAMPLING_PRIORITY_KEY;
 use crate::exporter::{Error, ModelConfig};
 use opentelemetry::sdk::export::trace;
 use opentelemetry::sdk::export::trace::SpanData;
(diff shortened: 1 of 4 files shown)


TommyCpp:
Need to format the code

opentelemetry-datadog/src/exporter/model/v05.rs
```diff
@@ -161,7 +161,10 @@ where
                 rmp::encode::write_u32(&mut encoded, interner.intern(key.as_str()))?;
                 rmp::encode::write_u32(&mut encoded, interner.intern(value.as_str().as_ref()))?;
             }
-            rmp::encode::write_map_len(&mut encoded, 0)?;
+            rmp::encode::write_map_len(&mut encoded, 1)?;
+            // https://github.com/DataDog/dd-trace-js/blob/c89a35f7d27beb4a60165409376e170eacb194c5/packages/dd-trace/src/constants.js#L4
+            rmp::encode::write_u32(&mut encoded, interner.intern("_sampling_priority_v1"))?;
+            rmp::encode::write_f64(&mut encoded, if span.span_context.is_sampled() { 1.0 } else { 0.0 })?;
```
TommyCpp, line 167: Could we define some constant at `model/mod.rs` and reuse them across v3 and v5 exporter?

TommyCpp (changes requested):
Have one nit and need to format the code but otherwise LGTM. Thanks for the help!

rex-remind101:
amended my commit with the suggestions. attempted to use the EasyCLA but I'm receiving the following The Individual CLA option is not enabled for this project. Please to your administrator to enable the Individual CLA option for this project. so i'm not sure how to move forward on this.

TommyCpp:
> The Individual CLA option is not enabled for this project. Please to your administrator to enable the Individual CLA option for this project

Could you take a screenshot? It should be something like 
image

rex-remind101:
it worked this time

TommyCpp:
@rex-remind101 I think we need to bump the msrv to 1.56 to resolve the CI issues. Could you make the change in https://github.com/open-telemetry/opentelemetry-rust/blob/main/.github/workflows/ci.yml#L71 ?

I don't have the write access to the branch.

rex-remind101:
@TommyCpp ready!

rex-remind101:
@TommyCpp looks like there is still an error actually, could you help provide direction?

TommyCpp:
> @TommyCpp looks like there is still an error actually, could you help provide direction?

I think the problem is still the MSRV. Let me experiment with some versions and see which one we need to use.

TommyCpp (approved):
Sorry for the delay. LGTM. Thanks!

PR #797, opened 2022-05-14
feat(trace): jaeger remote sampler

### LINKED ISSUE #606: Add jaeger remote sampler to SDK configuration
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/1791

TommyCpp:
fix #606.

Add initial implementation of jaeger remote sampler


TommyCpp:
cc @open-telemetry/rust-approvers if someone could take a look :ghost:

opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -46,6 +46,13 @@ use opentelemetry_api::{
 };
 use std::convert::TryInto;
 
+mod jaeger_remote;
```
jtescher, line 49: Hm could we have this as a feature in `opentelemetry-jaeger`? Just need it to implement [`ShouldSample`](https://docs.rs/opentelemetry/latest/opentelemetry/sdk/trace/trait.ShouldSample.html) and it could be set the same way as the rest of the sampler config

jtescher (changes requested):

opentelemetry-sdk/src/trace/sampler.rs
TommyCpp: I thought about it but decided to leave it here for two reasons.

- Although the name is jaeger remote sampler, the OTLP collector also supports this
- `Sampler` contains all SDK defined sampler and jaeger remote sampler is also one of them.
- We may need to support it in [`OTEL_TRACERS_SAMPLER`](https://sourcegraph.com/github.com/open-telemetry/opentelemetry-rust@HEAD/-/blob/opentelemetry-sdk/src/trace/config.rs) env var.

jtescher: Yeah good point, I missed that this is defined in the sdk spec

jtescher: think it should be an opt-in feature? it does add a few dependencies other than that looks good

TommyCpp: Yeah I think we can make it opt-in

TommyCpp: Updated

jtescher (approved):
Looks good, nice to support this one 👍

PR #799, opened 2022-05-17
Allow O(1) get operations for SpanBuilder::attributes [breaking]

### LINKED ISSUE #794: API design: SpanBuilder::attributes
Attributes attached to a `SpanBuilder` are currently stored as a `Vec<opentelemetry::KeyValue>` - see [docs](https://docs.rs/opentelemetry/latest/opentelemetry/trace/struct.SpanBuilder.html#structfield.attributes).

This makes it impossible to access the value attached to a specific key or to verify that an attribute is present in O(1). It always take O(n), where n is the number of attributes - this can be significant.

Why was a `Vec<KeyValue>` chosen over a `HashMap<Key, Value>`?
I see it gets later processed into an `EvictedHashMap` when `SpanBuilder` gets converted into a `Span` 🤔

LukeMathWalker:
Closes #794 

It changes the type of `SpanBuilder::attributes` from `Vec<KeyValue>` to `OrderMap`. `OrderMap` is a thin wrapper around `index_map::IndexMap` that only exposes the methods that preserve the original insertion order, which is what we care about later in the pipeline when enforcing the maximum number of attributes per span.

### DIFF
--- a/opentelemetry-api/src/trace/tracer.rs
+++ b/opentelemetry-api/src/trace/tracer.rs
@@ -332,7 +332,7 @@ impl SpanBuilder {
     /// via an [`OrderMap`] instance.
     pub fn with_attributes<I>(self, attributes: I) -> Self
     where
-        I: IntoIterator<Item = KeyValue>
+        I: IntoIterator<Item = KeyValue>,
     {
         SpanBuilder {
             attributes: Some(OrderMap::from_iter(attributes.into_iter())),
--- a/opentelemetry-api/src/trace/tracer.rs
+++ b/opentelemetry-api/src/trace/tracer.rs
@@ -1,8 +1,8 @@
+use crate::trace::OrderMap;
 use crate::{
     trace::{Event, Link, Span, SpanId, SpanKind, Status, TraceContextExt, TraceId, TraceState},
     Context, Key, KeyValue, Value,
 };
-use indexmap::IndexMap;
 use std::borrow::Cow;
 use std::time::SystemTime;
 
@@ -260,7 +260,7 @@ pub struct SpanBuilder {
     pub end_time: Option<SystemTime>,
 
     /// Span attributes
-    pub attributes: Option<indexmap::IndexMap<Key, Value>>,
+    pub attributes: Option<OrderMap<Key, Value>>,
 
     /// Span events
     pub events: Option<Vec<Event>>,
@@ -326,7 +326,7 @@ impl SpanBuilder {
     }
 
     /// Assign span attributes
-    pub fn with_attributes(self, attributes: IndexMap<Key, Value>) -> Self {
+    pub fn with_attributes(self, attributes: OrderMap<Key, Value>) -> Self {
         SpanBuilder {
             attributes: Some(attributes),
             ..self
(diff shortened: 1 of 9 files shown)


TommyCpp:
👍  Nice work. Just one requests

TommyCpp:
👍  Nice work. Just one requests

LukeMathWalker:
I've addressed it in an additional commit 👍🏻 
That code example now compiles without having to be touched at all.

opentelemetry-api/src/trace/tracer.rs
```diff
@@ -325,7 +326,7 @@ impl SpanBuilder {
     }
 
     /// Assign span attributes
-    pub fn with_attributes(self, attributes: Vec<KeyValue>) -> Self {
```
TommyCpp, line 328: Thanks for the change. It looks better. But do you think we can do something like 
```rust
    pub fn with_attributes<I: IntoIterator<Item = KeyValue>>(self, attributes: I) -> Self {
        SpanBuilder {
            attributes: Some(OrderMap::from_iter(attributes)),
            ..self
        }
    }
```
This way we won't break the existing implementation and also hide the `OrderMap` from the end user, who probably don't need to know this implementation details. 

(I know this PR is a breaking change anyway but I think most of the end user will only set the attributes here so maybe we can make their upgrade easier)

maybe add another function `with_attributes_order_map` which accepts `OrderMap<Key,Value>` as parameters.

LukeMathWalker: That's a great shout - I've amended it!

TommyCpp (approved):
LGTM. Thanks a lot. This is a great improvement 🎉

jtescher:
@LukeMathWalker @TommyCpp Uping MSRV to 1.55 will likely break compatibility with `tracing`'s 1.49 MSRV for `tracing-opentelemtry`. What's the motivation for optimizing attribute get operations?

TommyCpp:
> @LukeMathWalker @TommyCpp Uping MSRV to 1.55 will likely break compatibility with `tracing`'s 1.49 MSRV for `tracing-opentelemtry`. What's the motivation for optimizing attribute get operations?

#794 has the full discussion but on the high level, lib writers wants to manipulate the attributes attached on spans and current implementation only supports do it by iterate through the attributes array, which is O(N) 

On top of that we need to preserve the relative order the insertion so that the when number of the attributes exceed the limits we can trim the last attributes. 

Combine the requirements we somehow want a OrderMap. `indexmap` provided  that function. It's not a must have as we can just implement the same function on our own.

We can try to pin the version of `indexmap` to an older version to avoid a MSRV bump. I think tower depends on `indexmap` 1.0?

There are other places in the PR uses features stabilized after 1.49 but we can probably remove them.

jtescher:
>https://github.com/open-telemetry/opentelemetry-rust/issues/794 has the full discussion but on the high level, lib writers wants to manipulate the attributes attached on spans and current implementation only supports do it by iterate through the attributes array, which is O(N)

I see. The spec has [some similar issues open](https://github.com/open-telemetry/opentelemetry-specification/issues/1857), could be worth documenting the relationship between `OrderMap` and `EvictedHashMap`, or consider merging the concepts and exposing in API crate if necessary.

Also worth considering the configuration options for other structs besides `Span` that currently take a `Vec<KeyValue>` (e.g. `Event`, `Link`) as there is now some inconsistency in how attributes are passed.

PR #802, opened 2022-05-18
consolidate basic-otlp* examples docker-compose setups

guswynn:
I found that `basic-otlp` and `basic-otlp-http` are good examples to show how to use `http` vs `grpc` (though `tonic`) as the transport, but only the `http` had a `docker-compose` setup.

With a simple change to the port setup, the `docker-compose` can be moved into what I consider the "base directory" for `basic-otlp` examples (`examples/basic-otlp/`), and the README's can be adjusted to reflect the possibility of using `docker-compose` for either!


examples/basic-otlp-http/README.md
```diff
@@ -2,9 +2,11 @@
 * Run the application locally, to run as a docker container you have to change the relative paths from the `Cargo.toml`
 * The Collector then sends the data to the appropriate backend, in this case JAEGER
 
-This demo uses `docker-compose` and by default runs against the `otel/opentelemetry-collector-dev:latest` image.
+This demo uses `docker-compose` and by default runs against the `otel/opentelemetry-collector-dev:latest` image,
+and uses `http` as the transport.
 
 ```shell
+# in the `examples/basic-otlp/` directory:
```
TommyCpp, line 9: I think it maybe worth the copy the Dockerfile/docker-compose file/collector config file in this `examples/basic-otlp-http` too. 

The reasons its kind of intuitive to switch to another folder and then switch back to run the main application. Also docker related files doesn't really change overtime so it shouldn't be too hard to keep them in sync.

guswynn: good point! i pushed a new commit that does just this

guswynn:
ci issues seem unrelated?

TommyCpp:
Yeah I think the issue is introduced by another PR and someone has already proposed the fix.

examples/basic-otlp/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM rust:1.51
+COPY . /usr/src/basic-otlp-http/
+WORKDIR /usr/src/basic-otlp-http/
+RUN cargo build --release
+RUN cargo install --path .
+CMD ["/usr/local/cargo/bin/basic-otlp-http"]
```
TommyCpp, line 6:
```suggestion
COPY . /usr/src/basic-otlp/
WORKDIR /usr/src/basic-otlp/
RUN cargo build --release
RUN cargo install --path .
CMD ["/usr/local/cargo/bin/basic-otlp"]
```

TommyCpp (approved):
Looks good. Thanks. Just one nit

PR #806, opened 2022-06-03
automatically add traces / metrics paths

### LINKED ISSUE #773: Add signal path automatically for OTLP HTTP exporter
As per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#example-1), OTLP http exporter should be able to automatically add a path based on signal.

GaryPWhite:
Resolves #773

### DIFF
--- a/opentelemetry-otlp/src/exporter/mod.rs
+++ b/opentelemetry-otlp/src/exporter/mod.rs
@@ -13,7 +13,7 @@ use std::str::FromStr;
 use std::time::Duration;
 
 /// Target to which the exporter is going to send signals, defaults to https://localhost:4317.
-/// Learn about the relationship between this constant and metrics/spans/logs at 
+/// Learn about the relationship between this constant and metrics/spans/logs at
 /// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp
 pub const OTEL_EXPORTER_OTLP_ENDPOINT: &str = "OTEL_EXPORTER_OTLP_ENDPOINT";
 /// Default target to which the exporter is going to send signals.
--- a/opentelemetry-otlp/src/exporter/mod.rs
+++ b/opentelemetry-otlp/src/exporter/mod.rs
@@ -14,7 +14,7 @@ use std::time::Duration;
 
 /// Target to which the exporter is going to send signals, defaults to https://localhost:4317.
 /// Learn about the relationship between this constant and metrics/spans/logs at
-/// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp
+/// <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp>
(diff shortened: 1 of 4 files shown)


GaryPWhite:
noting that we could probably add logs exporter -- I noticed there was documentation for log signals, but the logs.rs exporter doesn't exist :)

TommyCpp:
> noting that we could probably add logs exporter -- I noticed there was documentation for log signals, but the logs.rs exporter doesn't exist :)

Ahh yeah once we merged #788 we can implement a logs exporter

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -12,20 +12,15 @@ use crate::Protocol;
 use std::str::FromStr;
 use std::time::Duration;
 
-/// Target to which the exporter is going to send spans or metrics, defaults to https://localhost:4317.
```
TommyCpp, line 15: :+1: Like this doc change

opentelemetry-otlp/src/span.rs
```diff
@@ -63,6 +64,11 @@ use opentelemetry::{
 
 use async_trait::async_trait;
 
+/// Target to which the exporter is going to send spans, defaults to https://localhost:4317.
```
TommyCpp, line 67:
```suggestion
/// Target to which the exporter is going to send spans, defaults to https://localhost:4317/v1/traces.
```

TommyCpp (commented):
I think it generally looks good. Thanks!

One ask I have is maybe we can add some document to explain the relationship between `OTEL_EXPORTER_OTLP_END` and `OTEL_EXPORTER_OTLP_{signal}_ENDPOINT` like https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp.

And reference it from env var constants. It may be easier for user to understand what to choose

opentelemetry-otlp/src/exporter/mod.rs
GaryPWhite: thanks much!

GaryPWhite:
> > noting that we could probably add logs exporter -- I noticed there was documentation for log signals, but the logs.rs exporter doesn't exist :)
> 
> Ahh yeah once we merged #788 we can implement a logs exporter

I can make an issue or continue from #773 when it is merged! feel free to assign me or let me know here!

TommyCpp (approved):
It seems log may need sometime to land. I think we can close #773 and create a new one for add log path and related envionment variables. 

The change looks nice 👍

jtescher:
Just waiting on tests/lints now?

GaryPWhite:
sorry about that -- doesn't run without maintainers clicking it so the feedback loop is long :)

TommyCpp:
> sorry about that -- doesn't run without maintainers clicking it so the feedback loop is long :)

It only applies to the first PR(to prevent bad actors from injecting crypto code in PR to exploit CI pipeline). Should be better once it's merged

TommyCpp:
Running `./precommit.sh` can help find the lint issue

GaryPWhite:
alright let me run that and be back in a sec

TommyCpp:
Emm looks like the EasyCLA stuck again 😞

GaryPWhite:
no worries. I'll find something to fix i'm sure with the script and push a new one up

Issue #800, opened 2022-05-18
Sampling Trouble


kevincox:
I was trying to use the [`opentelemetry::sdk::trace::ShouldSample`](https://docs.rs/opentelemetry/latest/opentelemetry/sdk/trace/trait.ShouldSample.html) API and am having trouble. I figured that I would at least write it down here in case others have the same issue or this could be used to improve the library. This is likely just a documentation problem, but I really struggled.

My goal was to export traces to [Honeycomb](https://www.honeycomb.io/) with sampling. I managed to do something simple like this:

```rust
#[derive(Debug)]
struct Sampler;

impl opentelemetry::sdk::trace::ShouldSample for Sampler {
	fn should_sample(&self,
		parent_context: Option<&opentelemetry::Context>,
		trace_id: opentelemetry::trace::TraceId,
		_name: &str,
		_span_kind: &opentelemetry::trace::SpanKind,
		attributes: &[opentelemetry::KeyValue],
		_links: &[opentelemetry::trace::Link],
		_instrumentation_library: &opentelemetry::sdk::InstrumentationLibrary,
	) -> opentelemetry::sdk::trace::SamplingResult {
		let sample_rate: u128 = attributes.iter()
			.find(|kv| kv.key == opentelemetry::Key::from_static_str("sampleRate"))
			.map(|kv| match kv.value {
				opentelemetry::Value::I64(rate) => rate.try_into().unwrap(),
				ref value => panic!("Unexpected value {:?} for sample rate", value),
			})
			.unwrap_or(1)
			.max(1);

		opentelemetry::sdk::trace::SamplingResult {
			decision: if u128::from_le_bytes(trace_id.to_bytes()) % sample_rate == 0 {
				opentelemetry::sdk::trace::SamplingDecision::RecordAndSample
			} else {
				opentelemetry::sdk::trace::SamplingDecision::Drop
			},
			attributes: Vec::new(),
			trace_state: match parent_context {
				Some(ctx) => opentelemetry::trace::TraceContextExt::span(ctx).span_context().trace_state().clone(),
				None => opentelemetry::trace::TraceState::default(),
			},
		}
	}
}
```

I'm sure the code could be slightly better, for example there is definitely some bias on the sampling depending on the sampleRate used. However this had a lot of problems:

[...]

PR #807, opened 2022-06-05
sdk: Update sampling documentation

hdost:
* Move Sampling header into `Sampler`
* Enrich classes to have links in tracing module

Relates #800

Signed-off-by: Harold Dost <github@hdost.com>

### DIFF
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -2,10 +2,10 @@
 //!
 //! The tracing SDK consist of a few main structs:
 //!
-//! * The `Tracer` struct which performs all tracing operations.
-//! * The `Span` struct with is a mutable object storing information about the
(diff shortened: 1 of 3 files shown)


hdost:
@TommyCpp barring the obvious failures is this a good direction for #800?

TommyCpp:
> @TommyCpp barring the obvious failures is this a good direction for #800?

Yeah I think it's a good start on docs, Thanks! I think we should sample at every span to be consistent with other SDKs. 

This has a risk of breaking change as the current behavior will be `Parent<SubSampler>` but I think it's worth it.

jtescher:
Yeah we should follow the spec's steps here https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sdk-span-creation

TommyCpp (approved):
Looks good. Thanks!

PR #810, opened 2022-06-08
opentelemetry-jaeger: Fix clearing span context in Propagator

lsr0:
When `Propagator::extract_with_context()` is run with an Extractor empty
of trace metadata, it returns an empty (invalid) Context, clearing
TraceId, and disabling downwards propagation. As there is already an
implicit TraceId (if we're in a Span), we'd like to use it, traces have
to start somewhere.

This just aligns the mechanics with opentelemetry-sdk[1]: if span
extraction errors out we return a clone of the existing context.

[1] https://github.com/open-telemetry/opentelemetry-rust/blob/fd587c39da24a47c1f4a5b3d423409400f4c8681/opentelemetry-sdk/src/propagation/trace_context.rs#L136-L138

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -513,10 +513,9 @@ mod propagator {
         }
 
         fn extract_with_context(&self, cx: &Context, extractor: &dyn Extractor) -> Context {
-            cx.with_remote_span_context(
-                self.extract_span_context(extractor)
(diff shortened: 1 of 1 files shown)


TommyCpp (approved):
Good catch! Thanks

lsr0:
Great. Just a note in case: I don't have permissions here to merge this.

lsr0:
Sure, just rebased on main.

PR #811, opened 2022-06-13
msrv: bump from 1.49 to 1.55

TommyCpp:
- Follow up https://github.com/open-telemetry/opentelemetry-rust/pull/799. Update the MSRV in README files.
- Pin the `async-std` version to `1.8` as any version > `1.8` will requires rustc 1.59 support, which breaks our MSRV guarantee. Ideally we should revert this change once https://github.com/async-rs/async-global-executor/issues/10 resolves and the new version releases
- Revert the MSRV in ci config from 1.59 to 1.55, it was previously bumped to unblock the PR merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
     - uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: 1.59.0
+        toolchain: 1.55.0
         override: true
     - name: Run tests
       run: cargo --version &&
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ analysis in order to understand your software's performance and behavior. You
 can export and analyze them using [Prometheus], [Jaeger], and other
 observability tools.
 
-*Compiler support: [requires `rustc` 1.49+][msrv]*
+*Compiler support: [requires `rustc` 1.55+][msrv]*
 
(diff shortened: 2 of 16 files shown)


jtescher (approved):
We may still want to support 1.49 as many people integrate via `tracing` https://github.com/tokio-rs/tracing/tree/master/tracing#supported-rust-versions

jtescher:
Which changes would we need to revert to continue supporting 1.49?

TommyCpp:
> Which changes would we need to revert to continue supporting 1.49?

The bump was introduced by https://github.com/open-telemetry/opentelemetry-rust/pull/799. I think alternatively we can rewrite part of the `indexmap` to avoid the bump. 

Although Tokio ecosystem promise 6-month compatibility. 1.55 was released in Sep 2021 so it's still consistent with that promise.

jtescher:
Actually nvm https://github.com/tokio-rs/tracing/pull/1913 also moved to 1.49 so we're good to go there.

jtescher:
Nvm we could open an issue to bump tracing as well, but probably simpler to retain 1.49 support if we can't get that in.

TommyCpp:
> but probably simpler to retain 1.49 support if we can't get that in.

I will give it a try! I will merge this for now so that we can maintain at least 1.55 MSRV when building the new PRs. The current one in main branch breaks our MSRV promise.

PR #812, opened 2022-06-13
chore: Bump actix-web examples dependencies

atanunq:
Bump the dependencies of `actix-*` examples and make them compile.

However, I cannot get `actix-http` to work, while `actix-http-tracing` and `actix-udp` run just fine. The thread panics after the first request that it receives:

```
[2022-06-13T23:01:07Z INFO  actix_server::builder] Starting 4 workers
[2022-06-13T23:01:07Z INFO  actix_server::server] Actix runtime found; starting in Actix runtime
[2022-06-13T23:01:30Z INFO  actix_web::middleware::logger] 127.0.0.1 "GET / HTTP/1.1" 200 5 "-" "curl/7.83.1" 0.000652
thread '<unnamed>' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', /home/atanunq/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.19.2/src/runtime/context.rs:54:26
OpenTelemetry trace error occurred. oneshot canceled
```

 I've traced it to [this thread](https://github.com/open-telemetry/opentelemetry-rust/blob/75348915b50879b300cfe2af3e750e3a313a4bc7/opentelemetry-jaeger/src/exporter/mod.rs#L85-L87) being unable to execute any tokio tasks, including the necessary [http send](https://github.com/open-telemetry/opentelemetry-rust/blob/3806258ee5db2223395fd38daac1d9e352974a17/opentelemetry-http/src/lib.rs#L65-L72). Though, my async runtime understanding only goes so far. Could it be a misconfiguration on my side?

### DIFF
--- a/examples/actix-http-tracing/Cargo.toml
+++ b/examples/actix-http-tracing/Cargo.toml
@@ -5,11 +5,11 @@ edition = "2018"
 publish = false
 
 [dependencies]
-actix-web = "3.2"
-actix-web-opentelemetry = { version = "0.9", features = ["metrics"] }
-opentelemetry = { version = "0.11", features = ["metrics", "tokio"] }
-opentelemetry-jaeger = { version = "0.10", features = ["tokio"] }
-opentelemetry-prometheus = "0.4"
+actix-web = "4.1"
+actix-web-opentelemetry = { version = "0.12", features = ["metrics"] }
+opentelemetry = { version = "0.17", features = ["metrics", "tokio"] }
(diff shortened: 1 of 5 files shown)


TommyCpp:
I think it's a bug in `opentelemetry-jaeger`. In simple words, the `reqwest` should be running within a tokio runtime but here we didn't select the right runtime for it. I will try to push a fix for it. Thanks for the reporting

jtescher (approved):
Updating the example dependencies is still an improvement even if the original issue may have another cause 👍

PR #818, opened 2022-06-20
fix: update `dashmap` to avoid soundness hole

Daniel-Bloom-dfinity:
xacrimon/dashmap#167 has a soundness hole pre `5.1.0`.


TommyCpp (approved):
Looks good. Thanks for catching this. Just need sign the CLA

PR #819, opened 2022-06-21
Update metrics API and SDK

### LINKED ISSUE #733: Asynch Instruments: Spec Conformance
In an effort to get closer to the spec I was taking a look at the asynchronous instruments. There's an opportunity for a few different namings which might make more sense. I'd like to put it to 
UpDownCounterObserver is now [Asynchronous UpDownCounter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-updowncounter) 
Sum Observer is now [Aysynchronous Counter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter)
ValueObserver is now [Asynchronous Gauge](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-gauge) 

However even in the spec there is some different suggestions for names. Before I go down the route of making a big patch I'd like to discuss the naming. Personally I think the following would be the best names for the replacements (in order):
* `<type>_asynchronous_up_down_counter`
* `<type>_asynchronous_counter`
* `<type>_asynchronous_gauge`

Alternatives would be:
* `<type>_observable_up_down_counter`
* `<type>_observable_counter`
* `<type>_observable_gauge`

Obviously there would be CamelCase types associated, but these would be the meter methods.

Additionally, I've seen in some languages that their choosing to implement the asynchronous gauges as use gauges. In either case of the above mentioned there could be an possibility of using: `<type>_gauge`

jtescher:
This change aligns metrics with the spec, changes include:

* Rename `MeterProvider::meter` to `MeterProvider::versioned_meter` for
  consistency with `TracerProvider` trait.
* Move metrics sdk api types to `opentelemetry-sdk`
* Consolidate instrument builders into `InstrumentBuilder`
* Remove value observers and add gauges.
* Move from batch observer to registered callbacks.
* Rename `ExportKindFor` to `TemporalitySelector`
* Consolidate `PushController` and `PullController` into
  `BasicController`
* Remove `MinMaxSumCountAggregator` and `ArrayAggregator`
* Update examples and exporters for new api/sdk

Resolves #733


jtescher:
cc @hdost

hdost:
for the #733 I think this looks good, However I do wonder if we might want to also add some convenience methods, could be in a followup PR of allowing users to not need to always pass the context. I can open up a separate ticket for this. I'm sure from a performance perspective passing the Context is more performant.

hdost:
Although seeing this https://github.com/open-telemetry/opentelemetry-rust/issues/828 it's possible this, perhaps there's some futher research to be done on ensuring it's working as expected ingeneral.

TommyCpp:
Sorry I have been busy moving. Will try to take a look today

opentelemetry-api/src/metrics/meter.rs
```diff
@@ -1,369 +1,165 @@
...
-        UpDownCounterBuilder::new(self, name.into(), NumberKind::I64)
-    }
-
-    /// Creates a new floating point `UpDownCounterBuilder` for an `f64` up down counter with the given name.
-    pub fn f64_up_down_counter<T>(&self, name: T) -> UpDownCounterBuilder<'_, f64>
-    where
-        T: Into<String>,
-    {
-        UpDownCounterBuilder::new(self, name.into(), NumberKind::F64)
+    /// creates an instrument for recording increasing values.
```
TommyCpp, line 50: Maybe 
```suggestion
    /// creates an instrument builder for recording increasing values.
```
?

opentelemetry-api/src/metrics/meter.rs
```diff
@@ -1,369 +1,165 @@
...
-    /// meter.batch_observer(|batch| {
-    ///   let instrument = batch.u64_value_observer("test_instrument").init();
-    ///
-    ///   move |result: BatchObserverResult| {
-    ///     result.observe(&[KeyValue::new("my-key", "my-value")], &[instrument.observation(1)]);
-    ///   }
-    /// });
-    /// ```
-    pub fn batch_observer<B, F>(&self, builder: B)
+    /// It is only valid to call `observe` within the scope of the passed function.
```
TommyCpp, line 149: This is kind of confusing. I suppose this is saying we cannot use Sync* instrument in callback?

opentelemetry-dynatrace/src/metric.rs
```diff
@@ -122,18 +122,16 @@ where
...
 where
     AS: AggregatorSelector + Send + Sync + 'static,
-    ES: ExportKindFor + Send + Sync + Clone + 'static,
-    SP: Fn(PushControllerWorker) -> SO,
-    I: Fn(time::Duration) -> IO,
-    IO: Stream<Item = IOI> + Send + 'static,
+    TS: TemporalitySelector + Clone + Send + Sync + 'static,
+    RT: Runtime,
 {
     /// Build with resource key value pairs.
```
TommyCpp, line 131:
```suggestion
    /// Build with resource
```

opentelemetry-sdk/src/metrics/aggregators/histogram.rs
```diff
@@ -77,8 +81,17 @@ impl Histogram for HistogramAggregator {
     }
 }
 
+impl Aggregation for HistogramAggregator {
+    fn kind(&self) -> &AggregationKind {
+        &AggregationKind::SUM
```
TommyCpp, line 86: Shouldn't it be `&AggregationKind::HISTOGRAM`?

examples/dynatrace/src/main.rs
```diff
@@ -128,28 +121,22 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
     let tracer = global::tracer("ex.com/basic");
     let meter = global::meter("ex.com/basic");
 
-    let one_metric_callback =
-        |res: ObserverResult<f64>| res.observe(1.0, COMMON_ATTRIBUTES.as_ref());
-    let _ = meter
-        .f64_value_observer("ex.com.one", one_metric_callback)
+    let gauge = meter
+        .f64_observable_gauge("ex.com.one")
         .with_description("A ValueObserver set to 1.0")
```
TommyCpp, line 132:
```suggestion
        .with_description("A GaugeObserver set to 1.0")
```

opentelemetry-sdk/src/metrics/sdk_api/instrument_kind.rs
```diff
@@ -3,26 +3,26 @@
 /// | **Name** | Instrument kind | Function(argument) | Default aggregation | Notes |
 /// | ----------------------- | ----- | --------- | ------------- | --- |
 /// | **Histogram**       | Synchronous  | Record(value) | MinMaxSumCount  | Per-request, any non-additive measurement |
-/// | **ValueRecorder**       | Synchronous  | Record(value) | MinMaxSumCount  | Depreated. Use Histogram. |
+/// | **Gauge**       | Synchronous  | Record(value) | MinMaxSumCount  | Depreated. Use Histogram. |
 /// | **ValueObserver**       | Asynchronous | Observe(value) | MinMaxSumCount  | Per-interval, any non-additive measurement |
 /// | **Counter**             | Synchronous additive monotonic | Add(increment) | Sum | Per-request, part of a monotonic sum |
 /// | **UpDownCounter**       | Synchronous additive | Add(increment) | Sum | Per-request, part of a non-monotonic sum |
 /// | **SumObserver**         | Asynchronous additive monotonic | Observe(sum) | Sum | Per-interval, reporting a monotonic sum |
 /// | **UpDownSumObserver**   | Asynchronous additive | Observe(sum) | Sum | Per-interval, reporting a non-monotonic sum |
```
TommyCpp, line 11: May also need to update the name here?

TommyCpp (commented):
Overall looks good. Just some nits

opentelemetry-sdk/src/metrics/aggregators/histogram.rs
jtescher: ya good catch

opentelemetry-sdk/src/metrics/sdk_api/instrument_kind.rs
jtescher: Yeah they seemed to move away from this chart in the spec. Will remove.

opentelemetry-api/src/metrics/meter.rs
jtescher: yeah that's the intention

jtescher:
@TommyCpp / @hdost any last issues or should we merge this and iterate from here?

TommyCpp (approved):
LGTM :+1:

PR #821, opened 2022-06-22
Allow ref counted keys and values

### LINKED ISSUE #809: Allow using `Arc<str>` in key-value strings
Currently, `opentelemetry` represents strings in key-value pairs (i.e. `Key`s and `Value::String`s) using `Cow<'static, str>`. This has the advantage that when a string constant is used, it isn't necessary to allocate a `String` on the heap for it. If a string is dynamically constructed at runtime, a `String` is used instead.

One deficiency of this approach is that oftentimes, a string may be dynamically determined at runtime, but repeated many times in an application's traces. In this case, such a string is represented as a `String`, and cloning it requires allocating memory and copying all of the string's bytes into that allocation. This has an unfortunate performance cost, especially when the same string is repeated a large number of times. It would be nice if it was also possible to use an `Arc<str>` to represent strings in key-value pairs. That way, cloning a repeated string to emit it as part of a trace would only require incrementing a reference count, reducing the overhead of using the string (and decreasing overall memory use).

This came up while reviewing a change to the `tracing` crate's OpenTelemetry integration (https://github.com/tokio-rs/tracing/pull/2135#discussion_r878597135).

I think it should be possible to support `Arc<str>` as a representation for `Key`s without making a breaking change, since `Key` is a newtype whose field is private. Although the `Key::new` constructor takes `S: Into<Cow<'static, str>>`, we could add a new `from_arc` (or similar) constructor. Then, the internal representation of `Key` could be changed to something like this:
```rust
pub struct Key(KeyInner);

enum KeyInner {
    Static(&'static str),
    Owned(String),
    RefCounted(Arc<str>),
}

impl Key {
    pub fn new<S: Into<Cow<'static, str>>>(value: S) -> Self {
        Self(match value.into() {
             Cow::Owned(s) => KeyInner::Owned(s),
             Cow::Borrowed(s) => KeyInner::Static(s),
        })
    }

    pub fn from_arc(value: Arc<str>) -> Self {
         Self(KeyInner::RefCounted(value))
    }
    
    // ...
}
```

Unfortunately, though, a similar change for `Value` isn't possible without breaking API compatibility, because `Value` is a `pub enum` without `#[non_exhaustive]`.

jtescher:
This updates attribute `Key` and `Value` inner types from `Cow<'static, str>` to a new private `OtelString` enum to support static, owned, or ref counted values.

Resolves #809

### DIFF
--- a/opentelemetry-api/src/common.rs
+++ b/opentelemetry-api/src/common.rs
@@ -1,23 +1,35 @@
 use std::borrow::Cow;
-use std::fmt;
+use std::sync::Arc;
+use std::{fmt, hash};
 
 /// The key part of attribute [KeyValue] pairs.
 ///
 /// See the [attribute naming] spec for guidelines.
 ///
 /// [attribute naming]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/common/attribute-naming.md
-#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct Key(Cow<'static, str>);
+#[derive(Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Key(OtelString);
 
 impl Key {
     /// Create a new `Key`.
-    pub fn new<S: Into<Cow<'static, str>>>(value: S) -> Self {
-        Key(value.into())
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use opentelemetry_api::Key;
+    /// use std::sync::Arc;
+    ///
+    /// let key1 = Key::new("my_static_str");
+    /// let key2 = Key::new(String::from("my_owned_string"));
+    /// let key3 = Key::new(Arc::from("my_ref_counted_str"));
+    /// ```
+    pub fn new(value: impl Into<Key>) -> Self {
+        value.into()
     }
 
     /// Create a new const `Key`.
     pub const fn from_static_str(value: &'static str) -> Self {
-        Key(Cow::Borrowed(value))
+        Key(OtelString::Static(value))
     }
 
     /// Create a `KeyValue` pair for `bool` values.
@@ -44,8 +56,8 @@ impl Key {
         }
     }
 
-    /// Create a `KeyValue` pair for `String` values.
-    pub fn string<T: Into<Cow<'static, str>>>(self, value: T) -> KeyValue {
+    /// Create a `KeyValue` pair for string-like values.
+    pub fn string(self, value: impl Into<StringValue>) -> KeyValue {
         KeyValue {
             key: self,
             value: Value::String(value.into()),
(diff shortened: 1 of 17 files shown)


opentelemetry-contrib/Cargo.toml
```diff
@@ -34,7 +34,7 @@ serde_json = { version = "1", optional = true}
 futures = {version = "0.3", optional = true}
 async-trait = {version = "0.1", optional = true}
 tokio = { version = "1.0", features = ["fs", "io-util"], optional = true }
-async-std = { version = "1.6", optional = true }
+async-std = { version = "1.10", optional = true }
```
TommyCpp, line 37: Although it probably won't break the MSRV, we should probably pin this to 1.10 to be consistent with other crates?

TommyCpp (approved):
I think to pin the `indexmap` to `=1.8` should fix the MSRV. I will spend some time and see if we can remove the `indexmap` as a dependency. Some context of this change can be found in https://github.com/open-telemetry/opentelemetry-rust/pull/799

opentelemetry-api/src/common.rs
```diff
@@ -166,11 +272,75 @@ pub enum Value {
...
+            OtelString::Owned(s) => s.fmt(f),
+            OtelString::Static(s) => s.fmt(f),
+            OtelString::RefCounted(s) => s.fmt(f),
+        }
+    }
+}
+
+impl StringValue {
+    /// Returns a string slice to this value
+    pub fn as_str(&self) -> &str {
```
djc, line 302: Should probably also implement `AsRef<str>`.

opentelemetry-sdk/src/resource/process.rs
```diff
@@ -25,13 +24,10 @@ impl ResourceDetector for ProcessResourceDetector {
...
-            .map(|arg| Cow::from(arg.to_string_lossy().into_owned()))
-            .collect::<Vec<Cow<'_, str>>>();
+            .map(|arg| arg.to_string_lossy().into_owned())
+            .collect::<Vec<String>>();
         Resource::new(vec![
-            KeyValue::new(
-                "process.command_args",
-                Value::Array(Array::String(cmd_arg_val)),
-            ),
+            KeyValue::new("process.command_args", Value::Array(cmd_arg_val.into())),
```
djc, line 30: For example here we first allocate a `Vec<String>` and then `cmd_arg_val.into()` will allocate a new `Vec<StringValue>`, redoing the outer allocation.

opentelemetry-api/src/common.rs
```diff
@@ -153,9 +226,42 @@ into_array!(
     (Vec<bool>, Array::Bool),
     (Vec<i64>, Array::I64),
     (Vec<f64>, Array::F64),
-    (Vec<Cow<'static, str>>, Array::String),
+    (Vec<StringValue>, Array::String),
 );
 
+impl From<Vec<&'static str>> for Array {
+    /// Convenience method for creating a `Value` from a `&'static str`.
```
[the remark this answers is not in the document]
jtescher: Happy to remove them

opentelemetry-sdk/src/resource/process.rs
jtescher: Agreed

opentelemetry-contrib/Cargo.toml
jtescher: This was only to get CI passing. Still have to resolve the MSRV issues on main.

djc (approved):

PR #831, opened 2022-07-04
Fix span processor stream import

jtescher:
[futures_util < 0.3.17](https://docs.rs/futures-util/0.3.16/futures_util/index.html) does not export `Stream` and `StreamExt` at the root.


TommyCpp (approved):

djc:
Uh, if you follow that link it sure does? (Under Re-exports.)

jtescher:
@djc right you are! will fix to specify v0.3.17 where it was introduced instead.

PR #839, opened 2022-07-14
Perform sampling as explained in the specification.

hdost:
* Sampling should always defer to the existing sampler. The logic before
  this patch reflected something similar to what is described in the
  ParentBased sampler.

Relates #800

### DIFF
--- a/opentelemetry-sdk/src/trace/tracer.rs
+++ b/opentelemetry-sdk/src/trace/tracer.rs
@@ -152,10 +152,7 @@ impl opentelemetry_api::trace::Tracer for Tracer {
         let span_kind = builder.span_kind.take().unwrap_or(SpanKind::Internal);
         let mut attribute_options = builder.attributes.take().unwrap_or_default();
         let mut link_options = builder.links.take();
-        let mut no_parent = true;
-        let mut remote_parent = false;
         let mut parent_span_id = SpanId::INVALID;
-        let mut parent_trace_flags = TraceFlags::default();
         let trace_id;
 
         let parent_span = if parent_cx.has_active_span() {
@@ -166,10 +163,7 @@ impl opentelemetry_api::trace::Tracer for Tracer {
 
         // Build context for sampling decision
         if let Some(sc) = parent_span.as_ref().map(|parent| parent.span_context()) {
-            no_parent = false;
-            remote_parent = sc.is_remote();
             parent_span_id = sc.span_id();
-            parent_trace_flags = sc.trace_flags();
             trace_id = sc.trace_id();
         } else {
             trace_id = builder
@@ -177,14 +171,11 @@ impl opentelemetry_api::trace::Tracer for Tracer {
                 .unwrap_or_else(|| config.id_generator.new_trace_id());
         };
 
-        // There are 3 paths for sampling.
-        //
-        // * Sampling has occurred elsewhere and is already stored in the builder
-        // * There is no parent or a remote parent, in which case make decision now
-        // * There is a local parent, in which case defer to the parent's decision
+        // In order to accomodate use cases like `tracing-opentelemetry` we there is the ability
+        // to use pre-sampling. Otherwise, the standard method of sampling is followed.
         let sampling_decision = if let Some(sampling_result) = builder.sampling_result.take() {
             self.process_sampling_result(sampling_result, parent_cx)
-        } else if no_parent || remote_parent {
+        } else {
             self.make_sampling_decision(
                 parent_cx,
                 trace_id,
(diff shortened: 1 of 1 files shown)


opentelemetry-api/src/trace/tracer.rs
```diff
@@ -271,9 +271,6 @@ pub struct SpanBuilder {
 
     /// Span status
     pub status: Status,
-
-    /// Sampling result
-    pub sampling_result: Option<SamplingResult>,
```
jtescher, line 276: This is necessary to support pre-sampled tracing implementations (e.g. `tracing-opentelemetry`) which require sampling information to be known before spans are created.

hdost: That should be handled by either the `Sampler` or the `SpanProcessor` used. Either of which can handle this.

jtescher: [This](https://github.com/tokio-rs/tracing/blob/master/tracing-opentelemetry/src/tracer.rs#L66-L109) is the implementation that will be affected by this change. We can consider supporting alternate implementations to facilitate the intended behavior there but breaking support for `tracing` entirely would not be an ideal choice.

TommyCpp: I guess it doesn't hurt to provide additional ability outside the API. Integration with `tracing` has been important to opentelemetry adoption in Rust community IMO.

hdost: Yea that makes sense, so then in that case it should be simplified down to. and re add the sampling_result.

```
let sampling_decision = if let Some(sampling_result) = builder.sampling_result.take() {
    self.process_sampling_result(sampling_result, parent_cx)
} else {                                      
    self.make_sampling_decision(
        parent_cx,
        trace_id,        
        &builder.name,        
        &span_kind,        
        &attribute_options,        
        link_options.as_deref().unwrap_or(&[]),        
        provider.config(),        
        &self.instrumentation_lib,        
    )        
}    
```

jtescher: Yeah that looks good to me 👍

jtescher:
@hdost want some help making the last changes? Think this is the only blocker to releasing v0.18

hdost:
> @hdost want some help making the last changes? Think this is the only blocker to releasing v0.18

I can push out a change tomorrow 👍🏼

hdost:
Sorry again for the delay on this. All should be good now.

hdost:
> Sorry again for the delay on this. All should be good now.

Unsure why i have an issue with msrv...

djc:
Because there was a new release of the h2 library recently, and it requires 1.56. I don't think our 1.49 is tenable anymore, we should just bump to 1.56 like the rest of the ecosystem.

hdost:
> Because there was a new release of the h2 library recently, and it requires 1.56. I don't think our 1.49 is tenable anymore, we should just bump to 1.56 like the rest of the ecosystem.

Indeed 😞 
@jtescher @TommyCpp what are your thoughts?

jtescher:
> @jtescher @TommyCpp what are your thoughts?

I'm in favor of going to 1.56 and 2021 edition 👍

TommyCpp:
Agreed. I think tracing also have a PR to bump the MSRV too

hdost:
> Agreed. I think tracing also have a PR to bump the MSRV too

It appears so though it's been open for a month i feel they probably will. , should we open a separate PR for MSRV increase?

djc:
See #866.

hdost:
😢 
```
error: package `async-global-executor v2.3.0` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.56.1
```

jtescher:
@hdost if you merge `main` you'll pick up the script changes to get MSRV passing again

hdost:
Finally! 👍🏼 Things are passing

jtescher (approved):

jtescher:
@hdost if you want to become an approver ([requirements](https://github.com/open-telemetry/community/blob/master/community-membership.md#requirements)) I'm happy to sponsor you, just have [open an issue](https://github.com/open-telemetry/community/issues/new?template=membership.md&title=REQUEST%3A%20New%20membership%20for%20%3Cyour-GH-handle%3E) here to get added.

TommyCpp:
> @hdost if you want to become an approver ([requirements](https://github.com/open-telemetry/community/blob/master/community-membership.md#requirements)) I'm happy to sponsor you, just have [open an issue](https://github.com/open-telemetry/community/issues/new?template=membership.md&title=REQUEST%3A%20New%20membership%20for%20%3Cyour-GH-handle%3E) here to get added.

+1 if you need another sponsor. Again thanks for all the work on the sampling side!

PR #848, opened 2022-07-23
feat(jaeger): remove internal message queue between exporter and exporting tasks

TommyCpp:
#781 introduced an internal queue between jaeger exporter's `export` method and uploader that actually handles the exporting because Uploader's `upload` methods takes a mutable reference and hence cannot be called concurrently. 

The internal queue helped unblock #781 but has following problems:
- when sending to collectors we use HTTP clients and those clients are `Send + Sync` thus the queue is not needed.
- The queue has an arbitrary capacity of 64, if we can concurrently export more than 64 spans it can result in span dropping. 

This PR proposes that:
- Make `upload` takes immutable self so that we can call it concurrently.
- Internally add lock around agent uploader 
- Remove the internal queue inside jaeger exporter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,3 +38,6 @@ members = [
     "examples/zpages"
 ]
 exclude = ["examples/external-otlp-grpcio-async-std"]
+
+[profile.release]
+debug = true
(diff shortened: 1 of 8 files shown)


jtescher (approved):
Looks good

PR #849, opened 2022-07-25
backport: dashmap change to fix security issue

TommyCpp:
Backport dashmap change in #818 to v0.17.0


opentelemetry/Cargo.toml
```diff
@@ -22,7 +22,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 async-std = { version = "1.6", features = ["unstable"], optional = true }
 async-trait = { version = "0.1", optional = true }
-dashmap = { version = "4.0.1", optional = true }
+dashmap = { version = "=5.1.0", optional = true }
```
djc, line 25: Maybe add a comment here about why we're pinning?

examples/external-otlp-tonic-tokio/src/main.rs
```diff
@@ -47,7 +47,7 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
         .map(|(name, value)| {
             let header_name = name
                 .strip_prefix(HEADER_PREFIX)
-                .map(|h| h.replace("_", "-"))
+                .map(|h| h.replace('_', "-"))
```
djc, line 50: It's a little ugly to have this in a commit whose message is "change log".

opentelemetry-aws/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.5.1
```
jtescher, line 3: Do we need to release new versions of these dependent crates? they should still be compatible with the 0.17.1 without a change right?

TommyCpp: Yep no user facing API has changed. So probably we don't need release a new version for each crates. I will remove those

examples/external-otlp-tonic-tokio/src/main.rs
TommyCpp: Sorry I didn't clean up the commit logs. Will force push a cleaned up version

TommyCpp:
MSRV issue is caused by https://github.com/plotters-rs/plotters/issues/398.

TommyCpp:
I think with the new release we can close this out

PR #850, opened 2022-07-25
[stackdriver] fix LogEntry.trace not populated correctly

weiwei-lin:
According to https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#logentry
LogEntry.trace should have the format of "projects/{project_id}/traces/<trace_id>".


djc (approved):
Nice, thanks!

Issue #855, opened 2022-07-28
Proposal: Add support for IPv6 Jaeger agent addresses


DevinCarr:
The Jaeger exporters all bind to INADDR_ANY ([opentelemetry-jaeger/src/exporter/runtime.rs](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-jaeger/src/exporter/runtime.rs#L26)) which means that you are unable to connect to Jaeger agent collectors hosted on IPv6 address ranges, such as `[::1]:6831`.

It would be great to inspect the type of address provided as the destination address and bind to either `0.0.0.0:0` for IPv4 addresses and `[::]:0` for IPv6 addresses. This is similar to what the opentelemetry-go library does for their socket binding: https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/jaeger/agent.go#L102

djc:
Sounds good, would you be able to contribute a PR?

PR #856, opened 2022-07-29
Add IPv6 support for Jaeger agent addresses

DevinCarr:
This adds support to setup socket binding to IN6ADDR_ANY and INADDR_ANY depending on which IP address family is provided as the agent endpoint.

Addresses #855

### DIFF
--- a/opentelemetry-jaeger/src/exporter/runtime.rs
+++ b/opentelemetry-jaeger/src/exporter/runtime.rs
@@ -78,12 +78,15 @@ impl JaegerTraceRuntime for opentelemetry::runtime::AsyncStd {
 }
 
 /// Sample the first address provided to designate which IP family to bind the socket to.
-/// Returns either INADDR_ANY or IN6ADDR_ANY
-fn addrs_and_family(host_port: &impl ToSocketAddrs) -> Result<(Vec<SocketAddr>, SocketAddr), io::Error> {
+/// IP families returned be INADDR_ANY as [`Ipv4Addr::UNSPECIFIED`] or
+/// IN6ADDR_ANY as [`Ipv6Addr::UNSPECIFIED`].
+fn addrs_and_family(
+    host_port: &impl ToSocketAddrs,
+) -> Result<(Vec<SocketAddr>, SocketAddr), io::Error> {
     let addrs = host_port.to_socket_addrs()?.collect::<Vec<_>>();
     let family = match addrs.first() {
-        Some(SocketAddr::V4(_)) | None => SocketAddr::from((Ipv4Addr::new(0, 0, 0, 0), 0)),
-        Some(SocketAddr::V6(_)) => SocketAddr::from((Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0), 0)),
+        Some(SocketAddr::V4(_)) | None => SocketAddr::from((Ipv4Addr::UNSPECIFIED, 0)),
+        Some(SocketAddr::V6(_)) => SocketAddr::from((Ipv6Addr::UNSPECIFIED, 0)),
     };
     Ok((addrs, family))
 }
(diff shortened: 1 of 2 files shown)


djc:
This would be easier to review if the rename was split out into a separate commit from the functional changes.

DevinCarr:
Do you mean a separate PR or just two commits in this PR?

djc:
Just two commits would be great, thanks!

opentelemetry-jaeger/src/exporter/runtime.rs
```diff
@@ -23,7 +23,7 @@ impl JaegerTraceRuntime for opentelemetry::runtime::Tokio {
     type Socket = tokio::net::UdpSocket;
 
     fn create_socket<T: ToSocketAddrs>(&self, host_port: T) -> thrift::Result<Self::Socket> {
-        let conn = std::net::UdpSocket::bind("0.0.0.0:0")?;
+        let conn = std::net::UdpSocket::bind(bind_addr(&host_port))?;
```
djc: This solution doesn't seem great, since it calls `to_socket_addrs()` twice (potentially resolving the host name twice). Instead, I think we should use a helper function more like this:

```rust
fn addrs_and_family(host_port: &impl ToSocketAddrs) -> Result<(Vec<SocketAddr>, SocketAddr), io::Error> {
    let mut addrs = host_port.to_socket_addrs()?.collect::<Vec<_>>();
    let family = match addrs.first() {
        Some(SocketAddr::V4(_)) | None => SocketAddr::from(([0, 0, 0, 0], 0)),
        Some(SocketAddr::V6(_)) => SocketAddr::from(([0, 0, 0, 0, 0, 0, 0, 0], 0)),
    };
    Ok((addrs, family))
}
```

We can then pass `&[SocketAddr]` to `connect()`.

opentelemetry-jaeger/src/exporter/runtime.rs
```diff
@@ -23,8 +26,9 @@ impl JaegerTraceRuntime for opentelemetry::runtime::Tokio {
     type Socket = tokio::net::UdpSocket;
 
     fn create_socket<T: ToSocketAddrs>(&self, endpoint: T) -> thrift::Result<Self::Socket> {
-        let conn = std::net::UdpSocket::bind(bind_addr(&endpoint))?;
-        conn.connect(endpoint)?;
+        let (addrs, family) = addrs_and_family(&endpoint)?;
+        let conn = std::net::UdpSocket::bind(family)?;
+        conn.connect(addrs.as_slice())?;
```
djc, line 31: Style nit: should be able to just pass `&addrs` instead of calling `as_slice()` explicitly?

opentelemetry-jaeger/src/exporter/runtime.rs
```diff
@@ -71,23 +77,13 @@ impl JaegerTraceRuntime for opentelemetry::runtime::AsyncStd {
...
-                INADDR_ANY
-            } else {
-                IN6ADDR_ANY
-            }
-        },
-    )
+fn addrs_and_family(host_port: &impl ToSocketAddrs) -> Result<(Vec<SocketAddr>, SocketAddr), io::Error> {
+    let addrs = host_port.to_socket_addrs()?.collect::<Vec<_>>();
+    let family = match addrs.first() {
+        Some(SocketAddr::V4(_)) | None => SocketAddr::from((Ipv4Addr::new(0, 0, 0, 0), 0)),
```
djc, line 85: I guess we could use `Ipv4Addr::UNSPECIFIED` here instead? And same for `Ipv6Addr`, I suppose.

opentelemetry-jaeger/src/exporter/runtime.rs
DevinCarr: I did originally attempt to use `&addrs` but rust complained that:
```
the trait 'std::net::ToSocketAddrs' is not implemented for 'std::vec::Vec<std::net::SocketAddr>'
```

djc (approved):

jtescher:
Looks like just clippy lint remaining

DevinCarr:
> Looks like just clippy lint remaining

Not sure how to fix that clippy lint without ignoring it since I can't see how it is reporting the function as not used when it is used in all three methods in that file.

djc:
The lint is not from clippy but from the compiler. All of callers are guarded with various `cfg` guards, so the new function should be guarded with the union of those guards.

DevinCarr:
Ahh, that makes sense!

djc:
You'll want the same guard on the new imports.

PR #866, opened 2022-08-22
chore(common): bump MSRV to 1.56

TommyCpp:
- Bump MSRV to 1.56
- Bump the edition to 2021
- Set thr `rust-version` in `Cargo.toml`s
- Add `patch_dependencies` to lower `dashmap` and `time` version. `dashmap` latest version requires 1.59, `time` latest version requires 1.57
- Unpin `indexmap` from `1.8`


djc:
I'm confused about the dashmap change (the repository seems to be testing 1.49). In general, in the crates ecosystem bumping the MSRV without bumping the feature version seems to be the de facto standard so calling it out doesn't seem helpful. I would probably just go with 1.57 if that's what's required for time.

TommyCpp:
> I'm confused about the dashmap change (the repository seems to be testing 1.49)

Their `rust-version` in `Cargo.toml` is 1.59. 

> I would probably just go with 1.57 if that's what's required for time.

Emm I think we should try to keep MSRV as low as possible. I intend to replace the `patch_dependencies.sh` with [cargo-minimal-versions](https://crates.io/crates/cargo-minimal-versions) and run MSRV with minimal compatible version. So our MSRV checks won't break whenever one of dependencies bumps MSRV

djc:
> Emm I think we should try to keep MSRV as low as possible. I intend to replace the `patch_dependencies.sh` with [cargo-minimal-versions](https://crates.io/crates/cargo-minimal-versions) and run MSRV with minimal compatible version. So our MSRV checks won't break whenever one of dependencies bumps MSRV

I wonder how useful this form of MSRV support is. For any downstream users that actually want to keep their dependencies up to date, it doesn't seem very meaningful. If our dependencies require a higher MSRV we should just adopt their MSRV and/or push back on their MSRV bumps. Otherwise the result is that MSRV xor bug fixes, which doesn't seem great either.

TommyCpp:
> I wonder how useful this form of MSRV support is

I think one use case is users that who are using 1.56 can pin `time` and `dashmap` dependency in their application to "force" resolve dependencies that work with 1.56. If we bump the MSRV to 1.59 then they won't be able to use the create at all.

`tracing` also runs their MSRV check against the minimal version of dependencies(See [here](https://github.com/tokio-rs/tracing/blob/master/.github/workflows/CI.yml#L168)). That's why they depended on `time` and still have an MSRV at 1.49.

I am also concerned that simply keeping our MSRV equal to the `max(MSRV of all our dependencies)` will break our CI pipeline and force us to release versions more often. Also, it's possible we won't be keep up with our MSRV promise as we now have less control of MSRV.

jtescher (approved):
This looks good to me. I agree that this isn't a perfect solution but I don't have a better one at the moment.

TommyCpp:
Gonna merge it as it is. Feel free to reopen the discussion if needed.

PR #867, opened 2022-08-25
Fix XrayPropagator when no header is present

khalsah:
This corrects an issue where using `XrayPropagator` without the `x-amzn-trace-id` trace header present would cause spans not to get logged/processed.

#### Background

My use case is in a project that combines `tracing`, `tracing-opentelemetry`, `opentelemetry`, and `opentelemetry-jaeger`.

I initially had everything working, but when I added the `XrayPropagator` I noticed that my spans were no longer appearing in jaeger when I made requests directly to my service. I then switched to `TraceContextPropagator` and everything worked again.

This change simply copies the logic used in `TraceContextPropagator` into `XrayPropagator`.

I tested it in my local project and confirmed that with this change traces without a parent are still appearing in `jaeger`.

### DIFF
--- a/opentelemetry-aws/src/lib.rs
+++ b/opentelemetry-aws/src/lib.rs
@@ -203,11 +203,9 @@ pub mod trace {
         }
 
         fn extract_with_context(&self, cx: &Context, extractor: &dyn Extractor) -> Context {
-            let extracted = self
-                .extract_span_context(extractor)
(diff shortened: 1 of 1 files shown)


TommyCpp (approved):
Looks good to me. Good catch

TommyCpp:
MSRV is probably not related. We will bump to 1.56 before release.

PR #871, opened 2022-09-02
Fix span processor exporting unsampled spans

### LINKED ISSUE #870: Exporters should not be sent "is_sampled=false" spans
I have a custom `ShouldSample` implemetation. I ran into https://github.com/open-telemetry/opentelemetry-rust/issues/800 and am glad to see https://github.com/open-telemetry/opentelemetry-rust/pull/839 is almost there.

But there's a remaining problem afaict. As per the [opentelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampling):

> `Sampled` flag in `TraceFlags` on `SpanContext`. This flag is propagated via the `SpanContext` to child Spans. For more details see the [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#sampled-flag). This flag indicates that the `Span` has been `sampled` and will be exported.
>
> **[Span Exporters](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-exporter) MUST receive those spans which have `Sampled` flag set to true and they SHOULD NOT receive the ones that do not.**

The problem is that, in the `opentelemetry` crate, span exporters are simply span processors. `Span::ensure_ended_and_exported` sends export data to all processors unconditionally:

https://github.com/open-telemetry/opentelemetry-rust/blob/043e4b7523f66e79338ada84e7ab2da53251d448/opentelemetry-sdk/src/trace/span.rs#L186-L208

(..unless the span was already exported, or the provider was shut down). It doesn't check for `self.span_context().is_sampled()`, and can't tell the difference between exporters and other span processors anyway.

Here's what happens whe installing a simple exporter:

https://github.com/open-telemetry/opentelemetry-rust/blob/c92a1ad5af6a3c4098ecfc1e54134c17f3a1e708/opentelemetry-sdk/src/trace/provider.rs#L154-L160

The `SimpleSpanProcessor` unconditionally sends to its internal channel:

https://github.com/open-telemetry/opentelemetry-rust/blob/c92a1ad5af6a3c4098ecfc1e54134c17f3a1e708/opentelemetry-sdk/src/trace/span_processor.rs#L143-L147

Which unconditionally calls `exporter.export`:

https://github.com/open-telemetry/opentelemetry-rust/blob/c92a1ad5af6a3c4098ecfc1e54134c17f3a1e708/opentelemetry-sdk/src/trace/span_processor.rs#L116-L118

So, this seems like another departure from the spec (which was very surprising to me).

---

A bad fix for this is to teach both the "simple" and "batched" exporters to ignore spans where `is_sampled=false`. This is bad for a bunch of reasons:

  * Other exporter crates implement their own processors, which might not know they have to filter
  * It's really not the responsibility of exporters to do filtering
  * `build_export_data` is not free, calling it for a non-sampled span is wasteful

Another fix is to not send the span to _any_ processor if it's not sampled - ie, to return early from `ensure_ended_and_exported` if `is_sampled=false`. This also seems bad because what if there's processors that _aren't_ exporters?

Again, the spec says:

  * span processors must receive spans with `is_recording=true`, regardless of `is_sampled`
  * span exporters must only receive spans with `is_recording=true && is_sampled=true`

It seems like the actual fix is to either:

  * keep separate lists for "processors" and "exporters"
  * add an "fn is_exporter(&self) -> bool" method to `SpanProcessor` - this seems expensive for no good reason

Either way this seems like a breaking change.

jtescher:
Currently span processors will send span data to exporters even if the sampled flag is false. This patch fixes this by explicitly checking the sampled state in both the batch and simple span processors.

Fixes #870

### DIFF
--- a/opentelemetry-sdk/src/testing/trace.rs
+++ b/opentelemetry-sdk/src/testing/trace.rs
@@ -9,14 +9,22 @@ use crate::{
 use async_trait::async_trait;
 use futures_util::future::BoxFuture;
 pub use opentelemetry_api::testing::trace::TestSpan;
-use opentelemetry_api::trace::{SpanContext, SpanId, SpanKind, Status};
+use opentelemetry_api::trace::{
+    SpanContext, SpanId, SpanKind, Status, TraceFlags, TraceId, TraceState,
+};
 use std::fmt::{Display, Formatter};
 use std::sync::mpsc::{channel, Receiver, Sender};
 
 pub fn new_test_export_span_data() -> SpanData {
     let config = Config::default();
     SpanData {
-        span_context: SpanContext::empty_context(),
+        span_context: SpanContext::new(
+            TraceId::from_u128(1),
+            SpanId::from_u64(1),
+            TraceFlags::SAMPLED,
+            false,
+            TraceState::default(),
+        ),
         parent_span_id: SpanId::INVALID,
         span_kind: SpanKind::Internal,
         name: "opentelemetry".into(),
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -141,6 +141,10 @@ impl SpanProcessor for SimpleSpanProcessor {
     }
 
     fn on_end(&self, span: SpanData) {
+        if !span.span_context.is_sampled() {
+            return;
+        }
+
         if let Err(err) = self.sender.send(Some(span)) {
             global::handle_error(TraceError::from(format!("error processing span {:?}", err)));
         }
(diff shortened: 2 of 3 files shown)


TommyCpp (commented):
the logic LGTM. Maybe better if we have some unit tests to prevent regression in the future?

jtescher:
@TommyCpp added a regression test 👍

TommyCpp (approved):

TommyCpp:
Sorry took a look yesterday but forgot to approve

jtescher:
@TommyCpp All good, thanks for the review 🥳

PR #872, opened 2022-09-04
Cleanup external-otlp examples

DavidS:
* Fix comment header to use simpler language
* Fix comment header to refer to correct binary
* Coalesce use statements
* Update env_logger to 0.9.0
* ensure that both examples are formatted the same


jtescher (approved):
LGTM thanks

PR #875, opened 2022-09-12
Match API and SDK crate name underscores

jtescher:
The existing [api](https://crates.io/crates/opentelemetry_api) and [sdk](https://crates.io/crates/opentelemetry_sdk) crates that this will replace were previously registered with underscores, so this matches that to enable publishing.


TommyCpp:
I think we also need to update the `coverage` command line but otherwise LGTM

jtescher:
@TommyCpp yeah we can update it, if you 👍 we should still be able to merge this as no coverage changed.

TommyCpp (approved):

PR #880, opened 2022-09-18
fix(datadog): send resource along with attributes

### LINKED ISSUE #876: Datadog exporter doesn't send resources from `trace::Config`
In `DatadogExporter` resources (https://docs.rs/opentelemetry/latest/opentelemetry/sdk/trace/struct.Config.html#structfield.resource) are not sent to datadog. Only span attributes are sent to Datadog.

TommyCpp:
fix #876 

Added resource along with the attributes when converting span to datadog format.

One follow-up would be to adopt our model to be consistent with [datadog's agent](https://github.com/DataDog/datadog-agent/blob/main/pkg/trace/api/otlp.go), which converts the OTLP span to Datadog span.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
         command: test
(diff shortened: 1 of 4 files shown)


jtescher (approved):

lroux-at-jellysmack:
Hello, sorry for the soundness, but is there any ETA in `crates.io` for this feature ? 
Thanks.

deweyjose:
Hi all - any word on a release for this fix?

Issue #877, opened 2022-09-15
Upgrade `grpcio` to 0.10 because of `boringssl-src` dependency.


BratSinot:
Greetings!

`grpcio` 0.9 with `openssl` flag still take `boringssl-src` as dependency. It was fixed in `0.10`: https://github.com/tikv/grpc-rs/pull/558

TommyCpp:
Fixed in #881

PR #881, opened 2022-09-18
feat(proto): remove build script and generate files using unit tests.

### LINKED ISSUE #873: Consider removing `tonic-build` from opentelemetry-proto build dependencies
At the moment, opentelemetry-proto has tonic-build in its build dependencies: 

https://github.com/open-telemetry/opentelemetry-rust/blob/a57993cdf3fd84fb1f3e6f0f128db5f87a81d267/opentelemetry-proto/Cargo.toml#L64-L66

This means `.proto -> .rs` compilation happens in the downstream projects which use otl-rust. That is, downstream project: 

* should have access to `protoc` during build
* spend time compiling `tonic-build` itself, and running it on fixed otl protos
* have more deps in Cargo.lock

An alternative would be to publish `otl-proto` crate with all Rust code pre-generated. A couple of ways to go about that: 

* commit generated code to the repo, with a `#[test]` which checks that the generated code is fresh (so, move `tonic-build` from a build-dep to a dev-deps. dev deps, unilke builddeps, don't "infect" downstream consumers)
* run code generation specifically before publishing, such that the generated code gets into the `.crate` archive sumbitted to crates.io, but not into the git repo.

TommyCpp:
Fix #873 #877

- Remove the build script in `opentelemetry-proto` crate.
- Generating the protobuf using unit tests.
- Remove `build-server` and `build-client` feature flag.


opentelemetry-proto/src/proto.rs
```diff
@@ -1,65 +1,70 @@
 #[cfg(feature = "gen-tonic")]
 /// Generated files using [`tonic`](https://docs.rs/crate/grpcio) and [`prost`](https://docs.rs/crate/protobuf/latest)
 pub mod tonic {
+    macro_rules! include_tonic {
```
djc, line 4: IMO the `include_tonic!` macro doesn't carry its weight here. Instead, it would be better to put a `#[path = "proto/tonic/opentelemetry.proto.collector.logs.v1.rs"]` attribute on the `mod` declaration, which removes a layer of indirection and should make this more legible to Rust tooling (like IDEs).

opentelemetry-proto/src/proto.rs
```diff
@@ -1,66 +1,68 @@
 #[cfg(feature = "gen-tonic")]
+#[path = "proto/tonic"]
```
djc, line 2: I think we should avoid generating `#[path]` attributes for modules that are defined here (as opposed to just being declared).

TommyCpp: Not sure I understand. Are you proposing we generate the folder structure for generated files?

djc: I'm saying we should not generate `#[path = ""]` attributes.

TommyCpp: I think if we don't annotate those mods with `#[path = ""]` then compiler will try to add another directory with the same name as the module when searching for files. 

For example, `tonic::log::v1` will try to find the file in `proto/tonic/log/v1` instead of `proto/tonic`. Not sure if there is a better way to do it.

djc: Well, the better way to do it is to write out the module hierarchy to the filesystem, as we do in stackdriver, IMO. But I guess this is fine too if we actually need those empty-`path` attributes.

opentelemetry-proto/tests/grpc_build.rs
```diff
@@ -41,9 +60,27 @@ fn build_grpc() {
...
+        .out_dir(TONIC_PROTO_DIR)
+        .compile(TONIC_PROTO_FILES, TONIC_INCLUDES)
+        .expect("cannot compile protobuf using tonic");
+    let after_build = build_content_map(TONIC_PROTO_DIR);
+    // see build_grpc for why we cannot use assert! here.
+    assert!(
+        before_build == after_build,
+        "generated file has changed, please commit the change file and rerun the test"
+    );
+}
```
matklad, line 79: This is slightly suboptimal: if I am reading this correctly, the code will `touch` generated files even if they are unchanged. And the problem with `touch` would be that it'd cause `Cargo` to re-build stuff (cargo looks at mtime, not at the file hash). Ideally, you'd want a pattern like this her: 

https://github.com/rust-lang/rust-analyzer/blob/d9e462fd97fe56a63da62d11bb5c4d2132e4c74a/crates/sourcegen/src/lib.rs#L171-L194

matklad: Oh, forgot the second problem: in some CI setups, the directory with sources is read-only. The lazy update a-la `ensure_file_contents` works with that nicely -- only failing tests want to mess with src.

TommyCpp: Ah good point. I revise the PR to generate the files in temp dir first and then move to the src if needed

djc (approved):

benesch:
Oh this is wonderful! Thanks so much, @jtescher! Is this something that can be released independently of the rest of the otel crates, or do we need to wait for the v0.19 release?

jtescher:
@benesch are you depending on this crate directly? could release a `v0.2.0` of `opentelemetry-proto` by itself.

jtescher:
The dependency changes here probably require a minor version bump so I think we'll have to include this in the next `opentelemetry-otlp` release.

benesch:
No problem, just thought I'd ask. Thanks again for making the change!

PR #884, opened 2022-09-25
feat(metrics): add instrument validation to `InstrumentBuilder`.

TommyCpp:
- [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#general-characteristics) describes some general rules on instrument configurations. This PR adds those validations and throws errors when any of them fails.
- Fix document on some functions in `InstrumentBuilder`


jtescher (approved):
Looks great

PR #887, opened 2022-09-28
add public functions for AWS trace header

### LINKED ISSUE #885: [Feature Request] Expose functions in opentelemetry-aws to serialize/desiralize AWS tracing information
Context:
When I try to introduce OpenTelemetry into AWS SQS messaging, I found that opentelemetry-aws doesn't support extracting context from SQS message.
In SQS, we can extract tracing information from the message system attribute (Ref: https://docs.aws.amazon.com/xray/latest/devguide/xray-services-sqs.html).
And, we can get the attributes in the SQS message as `HashMap<String, String>` by using `aws-sdk-sqs`.
However, we cannot extract the span context using `XrayPropagator` in `opentelemetry-aws` because it only supports HTTP header, not SQS message attrtibutes.
In HTTP header, trace header is in "x-amzn-trace-id". In SQS message attribute, the trace header is in "AwsTraceHeader".

Also, we cannot use `XrayPropagator` directly to inject span context to SQS message attribute when we send SQS message with `aws-sdk-sqs`.
In `aws-sdk-sqs`, the type of message system attribute is `HashMap<MessageSystemAttributeNameForSends,MessageSystemAttributeValue>`, so this doesn't implement `Injector`.

However, the format of the trace header value is the same as the HTTP header. So most logic in this crate is reusable for SQS as well.

Proposal:
To support SQS (and other AWS support as well?), I'd like to expose some functions to extract `SpanContext` from `String`, and construct `String` from `SpanContext` as public functions.

Thanks.

garasubo:
Fix #885 

Add public functions to reuse trace header serialize/deserialize logic in this crate

### DIFF
--- a/opentelemetry-aws/src/lib.rs
+++ b/opentelemetry-aws/src/lib.rs
@@ -87,6 +87,12 @@ pub mod trace {
         _private: (),
     }
 
+    /// Extract `SpanContext` from AWS X-Ray format string
+    ///
+    /// Extract OpenTelemetry [SpanContext][otel-spec] from [X-Ray Trace format][xray-trace-id] string.
+    ///
+    /// [otel-spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#SpanContext
+    /// [xray-trace-id]: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
     pub fn span_context_from_str(value: &str) -> Option<SpanContext> {
         let parts: Vec<(&str, &str)> = value
             .split_terminator(';')
@@ -140,6 +146,12 @@ pub mod trace {
         }
     }
 
+    /// Generate AWS X-Ray format string from `SpanContext`
+    ///
+    /// Generate [X-Ray Trace format][xray-trace-id] string from OpenTelemetry [SpanContext][otel-spec]
+    ///
+    /// [xray-trace-id]: https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sendingdata.html#xray-api-traceids
+    /// [otel-spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#SpanContext
     pub fn span_context_to_string(span_context: &SpanContext) -> Option<String> {
         if !span_context.is_valid() {
             return None;
--- a/opentelemetry-aws/src/lib.rs
+++ b/opentelemetry-aws/src/lib.rs
@@ -45,7 +45,8 @@ pub mod trace {
         },
         Context,
     };
-    use std::convert::{TryFrom, TryInto};
+    use std::borrow::Cow;
+    use std::convert::TryFrom;
 
     const AWS_XRAY_TRACE_HEADER: &str = "x-amzn-trace-id";
     const AWS_XRAY_VERSION_KEY: &str = "1";
@@ -86,28 +87,23 @@ pub mod trace {
         _private: (),
     }
 
-    #[allow(clippy::result_unit_err)]
-    pub fn deserialize_span_context(value: &str) -> Result<SpanContext, ()> {
+    pub fn span_context_from_str(value: &str) -> Option<SpanContext> {
         let parts: Vec<(&str, &str)> = value
             .split_terminator(';')
             .filter_map(from_key_value_pair)
             .collect();
 
-        let mut trace_id: TraceId = TraceId::INVALID;
-        let mut parent_segment_id: SpanId = SpanId::INVALID;
+        let mut trace_id = TraceId::INVALID;
+        let mut parent_segment_id = SpanId::INVALID;
         let mut sampling_decision = TRACE_FLAG_DEFERRED;
-        let mut kv_vec: Vec<(String, String)> = Vec::with_capacity(parts.len());
+        let mut kv_vec = Vec::with_capacity(parts.len());
 
         for (key, value) in parts {
             match key {
-                HEADER_ROOT_KEY => {
-                    let converted_trace_id: Result<TraceId, ()> =
-                        XrayTraceId(value.to_string()).try_into();
-                    match converted_trace_id {
-                        Err(_) => return Err(()),
-                        Ok(parsed) => trace_id = parsed,
-                    }
-                }
+                HEADER_ROOT_KEY => match TraceId::try_from(XrayTraceId(Cow::from(value))) {
+                    Err(_) => return None,
+                    Ok(parsed) => trace_id = parsed,
+                },
                 HEADER_PARENT_KEY => {
                     parent_segment_id = SpanId::from_hex(value).unwrap_or(SpanId::INVALID)
                 }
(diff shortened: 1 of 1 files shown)


opentelemetry-aws/src/lib.rs
```diff
@@ -86,6 +86,107 @@ pub mod trace {
         _private: (),
     }
 
+    pub fn deserialize_span_context(value: &str) -> Result<SpanContext, ()> {
```
djc, line 90: `Result<.., ()>` is considered to be unidiomatic, just use `Option<..>` for this case instead of allowing the clippy linlt.

I would probably name this `span_context_from_str()`.

opentelemetry-aws/src/lib.rs
```diff
@@ -86,6 +86,107 @@ pub mod trace {
...
+        let parts: Vec<(&str, &str)> = value
+            .split_terminator(';')
+            .filter_map(from_key_value_pair)
+            .collect();
+
+        let mut trace_id: TraceId = TraceId::INVALID;
```
djc, line 96: It's usually best to avoid type annotations unless they are actually necessary, which doesn't seem to be the case in this block of bindings.

opentelemetry-aws/src/lib.rs
```diff
@@ -86,6 +86,107 @@ pub mod trace {
...
+
+        for (key, value) in parts {
+            match key {
+                HEADER_ROOT_KEY => {
+                    let converted_trace_id: Result<TraceId, ()> =
+                        XrayTraceId(value.to_string()).try_into();
+                    match converted_trace_id {
+                        Err(_) => return Err(()),
+                        Ok(parsed) => trace_id = parsed,
+                    }
```
djc, line 109:
```suggestion
                    match TraceId::from(XrayTraceId(value.to_string()) {
                        Err(_) => return Err(()),
                        Ok(parsed) => trace_id = parsed,
                    }
```

Also, it's a bit silly/sad that we currently require the allocation (in `value.to_string()`) in order to parse from `XrayTraceId` to `TraceId`. We should either extract that `TryFrom` impl into a separate method or potentially change `XrayTraceId`'s inner value to be a `Cow<'a, str>`.

opentelemetry-aws/src/lib.rs
```diff
@@ -86,6 +86,108 @@ pub mod trace {
...
+        let sampling_decision: &str =
+            if span_context.trace_flags() & TRACE_FLAG_DEFERRED == TRACE_FLAG_DEFERRED {
+                REQUESTED_SAMPLE_DECISION
+            } else if span_context.is_sampled() {
+                SAMPLED
+            } else {
+                NOT_SAMPLED
+            };
+
+        let trace_state_header: String = span_context
```
djc, line 165: Don't think we need the type annotation here.

djc (commented):
This makes sense to me, and your changes mostly looks good to me.

There are a lot of pre-existing style issues IMO, so might be nice to clean those up, too?

garasubo:
Is this failure caused by my change?
https://github.com/open-telemetry/opentelemetry-rust/actions/runs/3141646959/jobs/5104324937

I believe this change doesn't affect GRPC related part.
```
---- build_grpc stdout ----
thread 'main' panicked at 'generated file has changed, please commit the change file and rerun the test', opentelemetry-proto/tests/grpc_build.rs:38:5
error: test failed, to rerun pass '-p opentelemetry-proto --test grpc_build'
```

djc:
No, that failure is unrelated.

opentelemetry-aws/src/lib.rs
garasubo: Changed the return type, and renamed the function name

opentelemetry-aws/src/lib.rs
garasubo: Removed unnecessary type annotations

opentelemetry-aws/src/lib.rs
garasubo: Fixed. And refactored `XrayTraceId` to avoid memory allocation

opentelemetry-aws/src/lib.rs
```diff
@@ -86,6 +86,108 @@ pub mod trace {
...
+    }
+
+    pub fn serialize_span_context(span_context: &SpanContext) -> Option<String> {
+        if !span_context.is_valid() {
+            return None;
+        }
+
+        let xray_trace_id: XrayTraceId = span_context.trace_id().into();
+
+        let sampling_decision: &str =
```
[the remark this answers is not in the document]
garasubo: Fixed

opentelemetry-aws/src/lib.rs
garasubo: Fixed

opentelemetry-aws/src/lib.rs
```diff
@@ -250,15 +252,15 @@ pub mod trace {
         }
     }
 
-    impl From<TraceId> for XrayTraceId {
+    impl From<TraceId> for XrayTraceId<'_> {
```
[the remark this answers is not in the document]
garasubo: Fixed

opentelemetry-aws/src/lib.rs
```diff
@@ -86,126 +87,127 @@ pub mod trace {
...
-                    Ok(context)
-                }
-                Err(trace_state_err) => {
-                    global::handle_error(Error::Trace(TraceError::Other(Box::new(
-                        trace_state_err,
-                    ))));
-                    Err(()) //todo: assign an error type instead of using ()
-                }
-            }
+            span_context_from_str(header_value)
```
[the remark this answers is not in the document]
garasubo: Fixed

opentelemetry-aws/src/lib.rs
```diff
@@ -86,126 +87,127 @@ pub mod trace {
...
+                ))
+            }
+            Err(trace_state_err) => {
+                global::handle_error(Error::Trace(TraceError::Other(Box::new(trace_state_err))));
+                None //todo: assign an error type instead of using None
+            }
+        }
+    }
+
+    pub fn serialize_span_context(span_context: &SpanContext) -> Option<String> {
```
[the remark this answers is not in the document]
garasubo: Renamed

opentelemetry-aws/src/lib.rs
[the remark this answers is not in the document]
djc: Actually you copied the `unwrap_or()` that was there instead of using `?`. 🙂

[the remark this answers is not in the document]
garasubo: Oh, it is not intended. Fixed.

djc (approved):

TommyCpp:
I think if you merge the upstream changes the test should be fixed

opentelemetry-aws/src/lib.rs
```diff
@@ -86,126 +87,125 @@ pub mod trace {
         _private: (),
     }
 
-    impl XrayPropagator {
-        /// Creates a new `XrayTraceContextPropagator`.
-        pub fn new() -> Self {
-            XrayPropagator::default()
+    pub fn span_context_from_str(value: &str) -> Option<SpanContext> {
```
TommyCpp, line 90: Nit: can we add some documents for those functions now that they are public

garasubo: Add documents

PR #892, opened 2022-10-11
Fix `otel-collector-config.yaml` for insecure Jaeger connections

saschagrunert:
The latest exporter deployment will complain about the invalid configure ion, which now has moved to a new `tls` key.


jtescher (approved):
Looks good, thanks @saschagrunert

Issue #823, opened 2022-06-26
WithContext is not exposed through exports


exFalso:
Hi, I'm trying to use `WithContext` and I need to specify its type in a trait implementation. This is in a tonic GRPC service with a streaming API, where I need to specify the stream's type with an associated type.
However, I cannot find any way to actually import `WithContext`, I can only construct instances of with with the `with_context()` functions. Is this intended?

mkatychev:
This would also help in isntrumenting by attaching a span and context:

```rust
pub trait InstrumentWithContext: Instrument {
    fn instrument_cx(self, span: Span) -> Instrumented<WithContext<Self>>;
}
```

PR #893, opened 2022-10-11
pub use `opentelemetry_api::trace::WithContext`

mkatychev:
Resolves https://github.com/open-telemetry/opentelemetry-rust/issues/823

This scenario is specifically meant to make instrumenting async functions with contexts:

```rust
pub trait InstrumentWithContext: Instrument {
    fn instrument_cx(self, span: tracing::Span) -> Instrumented<WithContext<Self>>;
}

impl<T> InstrumentWithContext for T
where
    T: Instrument,
{
    fn instrument_cx(self, span: tracing::Span) -> Instrumented<WithContext<T>> {
        self.with_context(span.context()).instrument(span)
    }
}
```

Right now, calling `some_async_fn().instrument(tracing_span)` does not set the span's context as the `some_async_fn`'s current `Context`. 

Instead, the current flow is to chain both methods: `some_async_fn().with_context(span.context()).instrument(span).await`

This is a first step to bundling ot Contexts more consistently.

Traits involved:
* [`tracing::instrument::Instrument`](https://docs.rs/tracing/latest/tracing/instrument/trait.Instrument.html)
* [`opentelemetry::trace::FutureExt`](https://docs.rs/opentelemetry/latest/opentelemetry/trace/trait.FutureExt.html#method.with_context)

### DIFF
--- a/opentelemetry-api/src/trace/mod.rs
+++ b/opentelemetry-api/src/trace/mod.rs
@@ -175,7 +175,9 @@ mod tracer;
 mod tracer_provider;
 
 pub use self::{
-    context::{get_active_span, mark_span_as_active, FutureExt, SpanRef, TraceContextExt},
+    context::{
+        get_active_span, mark_span_as_active, FutureExt, SpanRef, TraceContextExt, WithContext,
+    },
     order_map::OrderMap,
     span::{Span, SpanKind, Status},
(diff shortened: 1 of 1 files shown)


jtescher (approved):
Looks good thanks @mkatychev

PR #895, opened 2022-10-12
Add `CollectorPipeline::build_collector_exporter`

### LINKED ISSUE #894: `build_exporter` for `opentelemetry-jaeger` `CollectorPipeline`
It looks like somewhere along the way to 0.18.0 the ability to create a standalone exporter was removed from `optentelemetry-jaeger` for Jaeger Collector.

This means that it is now impossible to use jaeger collector exporter in combination with another exporter as there is no way to build a tracer provider in the following way:

```
  let mut builder = sdk::trace::TracerProvider::builder();
  builder = builder.with_span_processor(jaeger_processor);
  builder = builder.with_span_processor(other_processor);
  builder.build()
```

Would it be possible to add `build_async_collector_exporter` to `CollectorPipeline` so that we can have this functionality back?

On a side note it looks like every exporter has a very slightly different way of building an exporter:

`otlp`->`build_span_exporter`
`zipkin`->`init_exporter`
`datadog`->`build_exporter`
`jaeger`->`build_async_agent_exporter` (but nothing for collector)

It'd be great to standardize on one name to build an exporter.

BrynCooke:
Fixes #894

### DIFF
--- a/opentelemetry-jaeger/CHANGELOG.md
+++ b/opentelemetry-jaeger/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## Main
+
+### Added
+
+- Added `CollectorPipeline::build_collector_exporter` #894
+
 
 ## v0.17.0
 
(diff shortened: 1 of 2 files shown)


jtescher (commented):
This looks quite reasonable! want to add a similar method to get an agent exporter?

BrynCooke:
I believe that the agent exporter methods are already present: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-jaeger/src/exporter/config/agent.rs#L338-L371

TommyCpp (approved):
LGTM. Thanks!

PR #896, opened 2022-10-14
Headers on exporters

### LINKED ISSUE #889: OTLP exporters should emit standard user agent string
As per the following change in the specification (1.14 releasing Oct 4):

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
> ```OTel OTLP Exporter Python/1.2.3```

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

GaryPWhite:
Resolves #889 

Adds headers to oltp exporters specifying `User-Agent` for current version of `OTel OLTP Exporter Rust/$version`.

### DIFF
--- a/opentelemetry-otlp/src/exporter/grpcio.rs
+++ b/opentelemetry-otlp/src/exporter/grpcio.rs
@@ -25,7 +25,10 @@ pub struct GrpcioConfig {
 impl Default for GrpcioConfig {
     fn default() -> Self {
         let mut headers: HashMap<String, String> = HashMap::new();
-        headers.insert("User-Agent".to_string(), format!("OTel OTLP Exporter Rust/{}", env!("CARGO_PKG_VERSION")));
+        headers.insert(
+            "User-Agent".to_string(),
+            format!("OTel OTLP Exporter Rust/{}", env!("CARGO_PKG_VERSION")),
+        );
         GrpcioConfig {
             credentials: None,
             headers: Some(headers),
--- a/opentelemetry-otlp/src/exporter/grpcio.rs
+++ b/opentelemetry-otlp/src/exporter/grpcio.rs
@@ -3,6 +3,8 @@ use crate::ExportConfig;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+use super::default_headers;
+
 /// Configuration of grpcio
 #[derive(Debug)]
 pub struct GrpcioConfig {
@@ -24,14 +26,9 @@ pub struct GrpcioConfig {
 
 impl Default for GrpcioConfig {
     fn default() -> Self {
-        let mut headers: HashMap<String, String> = HashMap::new();
-        headers.insert(
-            "User-Agent".to_string(),
-            format!("OTel OTLP Exporter Rust/{}", env!("CARGO_PKG_VERSION")),
-        );
         GrpcioConfig {
             credentials: None,
-            headers: Some(headers),
+            headers: Some(default_headers()),
             compression: None,
             use_tls: None,
             completion_queue_count: 2,
--- a/opentelemetry-otlp/src/exporter/grpcio.rs
+++ b/opentelemetry-otlp/src/exporter/grpcio.rs
@@ -25,7 +25,7 @@ pub struct GrpcioConfig {
 impl Default for GrpcioConfig {
     fn default() -> Self {
         let mut headers: HashMap<String, String> = HashMap::new();
-        headers.insert("User-Agent".to_string(), format!("OTel OLTP Exporter Rust/{}", env!("CARGO_PKG_VERSION")));
+        headers.insert("User-Agent".to_string(), format!("OTel OTLP Exporter Rust/{}", env!("CARGO_PKG_VERSION")));
         GrpcioConfig {
             credentials: None,
             headers: Some(headers),
--- a/opentelemetry-otlp/src/exporter/grpcio.rs
+++ b/opentelemetry-otlp/src/exporter/grpcio.rs
@@ -24,9 +24,11 @@ pub struct GrpcioConfig {
 
 impl Default for GrpcioConfig {
     fn default() -> Self {
+        let mut headers: HashMap<String, String> = HashMap::new();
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/src/exporter/http.rs
```diff
@@ -66,12 +66,21 @@ pub struct HttpExporterBuilder {
 
 impl Default for HttpExporterBuilder {
     fn default() -> Self {
+        let mut headers: HashMap<String, String> = HashMap::new();
+        headers.insert(
+            "User-Agent".to_string(),
+            format!("OTel OTLP Exporter Rust/{}", env!("CARGO_PKG_VERSION")),
```
TommyCpp, line 72: nit: since we are reusing those two across multiple places. Maybe a good idea to make those static string constants

opentelemetry-otlp/src/exporter/grpcio.rs
```diff
@@ -85,7 +90,9 @@ impl GrpcioExporterBuilder {
 
     /// Set additional headers to send to the collector.
     pub fn with_headers(mut self, headers: HashMap<String, String>) -> Self {
-        self.grpcio_config.headers = Some(headers);
+        let mut inst_headers = self.grpcio_config.headers.unwrap_or_default();
+        inst_headers.extend(headers.into_iter());
+        self.grpcio_config.headers = Some(inst_headers);
```
TommyCpp, line 95: nit: is it possible to use https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert here? It may looks cleaner

TommyCpp (commented):
I think the approach looks good. Just some small nit.

opentelemetry-otlp/src/exporter/http.rs
GaryPWhite: makes sense.

GaryPWhite: alright, because we have to format/read from environment variable, not gonna work as a `const`. However, I'll pull them into a function so we don't have to rewrite _this_ everywhere, just `defaultHeaders`.

opentelemetry-otlp/src/exporter/grpcio.rs
GaryPWhite: I'm not sure if that works -- since what we need is the headers provided PLUS the headers already found in `self.grpcio_config.headers` -- `get_or_insert` would only allow us to overwrite the headers, OR write over them with `headers` provided.

GaryPWhite: so we'd run into a couple situations (for future proofing)
1. We'd be writing over the headers we had already put in (User-Agent plus whatever we may include later)
2. We would get a pointer back, which we then have to do the same work to write into and put back into it's own.. reference? Getting the ref might be tricky to update that data OR write over it, I guess.

Maybe I'm missing something with your suggestion -- but it seems like `get_or_insert` would write over the non-existent headers or we'd have to extend this way anyways. What do you think?

GaryPWhite:
@TommyCpp dealing with some weird... linting...

Basically it's upset if I do include a [`..<default-rest-of-struct>` for `TonicConfig`](https://github.com/open-telemetry/opentelemetry-rust/actions/runs/3266968550/jobs/5371482501) and it's upset [if I don't include it](https://github.com/open-telemetry/opentelemetry-rust/actions/runs/3267114006/jobs/5371844082). Since _not_ including it will also fail a test, I'm looking to understand if I should:

1. find a way to silence the linter for this case, in isolation on that one file
2. work around it some other way, open to suggestions :)

opentelemetry-otlp/src/exporter/grpcio.rs
TommyCpp: Emm I assumed your intention is to override the `User-Agent` header even when the user already provided one. In that case, `get_or_insert` can be used. Note that the `extend` method can be applied to a mutable reference and it will just change the Hashmap the reference points to. Here is an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d5f0a91fd1aedeb9e6a19ab891a61279).

On second thought I think we should not override the `User-Agent` header if the user provided one. So[ `extend` ](https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-Extend%3C(K%2C%20V)%3E-for-HashMap%3CK%2C%20V%2C%20S%3E) probably won't work in this case as it will override the value. We can still get a mutable reference using `get_or_insert`, we can then check if the `User-Agent` is present and add one if it doesn't

opentelemetry-otlp/src/exporter/tonic.rs
```diff
@@ -25,13 +27,32 @@ pub struct TonicConfig {
...
 
+impl Default for TonicExporterBuilder {
+    fn default() -> Self {
+        let tonic_config = TonicConfig {
+            metadata: Some(MetadataMap::from_headers(
+                (&default_headers())
+                    .try_into()
+                    .expect("Invalid tonic headers"),
+            )),
+            ..Default::default()
```
TommyCpp, line 45:
```suggestion
        let tonic_config = TonicConfig {
            metadata: Some(MetadataMap::from_headers(
                (&default_headers())
                    .try_into()
                    .expect("Invalid tonic headers"),
            )),
            #[cfg(feature = "tls")]
            tls_config: None,
```

Should do the trick

GaryPWhite: makes sense! I tried to do something like this but obviously did it wrong, it wouldn't compile :) this compiles, so i'll run it through! Thanks!

opentelemetry-otlp/src/exporter/grpcio.rs
TommyCpp: @GaryPWhite Let me know if you think this makes sense or if maybe I misunderstand something here?

GaryPWhite: put in get_or_insert

TommyCpp (approved):
Looks good! Thanks for the work 🎉

PR #899, opened 2022-10-23
feat(trace): add TelemetryResourceDetector

TommyCpp:
https://github.com/open-telemetry/opentelemetry-demo/pull/504 needs to add `telemetry.sdk.*` info to the resources

### DIFF
--- a/opentelemetry-sdk/src/resource/telemetry.rs
+++ b/opentelemetry-sdk/src/resource/telemetry.rs
@@ -1,0 +1,27 @@
+use crate::resource::ResourceDetector;
+use crate::Resource;
+use opentelemetry_api::KeyValue;
+use std::time::Duration;
+
+/// Detect the telemetry SDK information used to capture data recorded by the instrumentation libraries.
+///
+/// It provides:
+/// - The name of the telemetry SDK(`telemetry.sdk.name`). It will be `opentelemetry` for SDK provided by opentelemetry project.
+/// - The language of the telemetry SDK(`telemetry.sdk.language`). It will be `rust` for this SDK.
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/resource/telemetry.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Note that the `telemetry.auto.version` is not provided as of now.
+///
+/// See [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#telemetry-sdk) for details.
+#[derive(Debug)]
+pub struct TelemetryResourceDetector;
+
+impl ResourceDetector for TelemetryResourceDetector {
+    fn detect(&self, _timeout: Duration) -> Resource {
+        Resource::new(vec![
+            KeyValue::new("telemetry.sdk.name", "opentelemetry"),
```
hdost, line 22:
```suggestion
            KeyValue::new("telemetry.sdk.name", "opentelemetry-sdk"),
```

hdost (changes requested):
Looks good! One minor comment.

hdost (approved):
The comment could honestly be done in a follow up.

opentelemetry-sdk/src/resource/telemetry.rs
TommyCpp: Good call out, but the spec asks all opentelemetry provided SDK to use `opentelemetry` as the value for this key

> The default OpenTelemetry SDK provided by the OpenTelemetry project MUST set telemetry.sdk.name to the value opentelemetry

PR #901, opened 2022-10-24
Tonic Interceptor

blogle:
Adds optional config for an interceptor to enable modifying each outbound metrics request.

For example - my company has a centralized opentelemetry-collector that gives our SRE team flexibility to determine where they should pipe and store metrics that feed into grafana. This central collector sits behind google IAP. Each internal service needs to supply an oauth token in the grpc headers to publish metrics to this endpoint. This PR exposes the machinery for applications to access the request object and supply such metadata.


opentelemetry-otlp/src/exporter/tonic.rs
```diff
@@ -25,11 +26,31 @@ pub struct TonicConfig {
 ///
 /// [tonic]: <https://github.com/hyperium/tonic>
 /// [channel]: tonic::transport::Channel
-#[derive(Default, Debug)]
+#[derive(Default)]
```
jtescher, line 29: Nit: The custom debug impl could be moved to `BoxInterceptor`.

opentelemetry-otlp/src/exporter/tonic.rs
```diff
@@ -25,11 +26,31 @@ pub struct TonicConfig {
...
 pub struct TonicExporterBuilder {
     pub(crate) exporter_config: ExportConfig,
     pub(crate) tonic_config: TonicConfig,
     pub(crate) channel: Option<tonic::transport::Channel>,
+    pub(crate) interceptor: Option<BoxInterceptor>,
+}
+
+pub(crate) struct BoxInterceptor(Box<dyn tonic::service::Interceptor>);
```
jtescher, line 37: This bound should be on the interceptor impl rather than assumed via unsafe.

```suggestion
pub(crate) struct BoxInterceptor(Box<dyn tonic::service::Interceptor + Send>);
```

opentelemetry-otlp/src/exporter/tonic.rs
```diff
@@ -25,11 +26,31 @@ pub struct TonicConfig {
...
+unsafe impl Send for BoxInterceptor {}
+impl tonic::service::Interceptor for BoxInterceptor {
+    fn call(&mut self, request: tonic::Request<()>) -> Result<tonic::Request<()>, tonic::Status> {
+        self.0.call(request)
+    }
+}
+
+impl Debug for TonicExporterBuilder {
```
jtescher, line 45: As mentioned above, not needed if `Debug` is implemented for `BoxInterceptor`.

jtescher (changes requested):
Thanks for getting this started, makes sense to have interceptor support, added some comments.

opentelemetry-otlp/src/metric.rs
```diff
@@ -289,6 +284,29 @@ impl MetricsExporter {
...
+    <T::ResponseBody as Body>::Error: Into<StdError> + Send,
+{
+    while let Some(msg) = receiver.recv().await {
+        match msg {
+            ExportMsg::Shutdown => {
+                break;
+            }
+            ExportMsg::Export(req) => {
+                let r = client.export(req).await;
+                println!("opentelemetry-otlp {:?}", r)
```
jtescher, line 304: Remove extra println

blogle:
bump @jtescher

Is anything else you would like to see on this PR? Getting this auth setup working is becoming a higher priority for our team, so would be great to get this integrated.

TommyCpp:
Sorry didn't get a chance to take a look over the weekend. I will take a look before cut 0.19 release

blogle:
Is this the only item going into the 0.19 release? If so what needs to be done to move this forward? We would really like to start migrating our metrics, but this has been a blocking issue for us.

hdost (approved):
Looks like you've addressed @jtescher suggestions but yes you'll need to need to sign the CLA. 
Thank you for your contribution!

TommyCpp (approved):

TommyCpp:
Probably need to address the lint issue also. But I can take a look if needed

blogle:
CLA should now be cleared up. @TommyCpp as for the lints, I believe those `pub` warnings are pre-existing as I don't think I touched visibility of anything. That said, if this is blocking the PR, I can probably make those all `pub(crate)` since `TonicExporterBuilder` is the actual public interface there.

TommyCpp:
Looks like it needs your approval cc @jtescher

TommyCpp:
I have pushed a change to fix the lint issue and make the configurations public

jtescher (approved):

PR #903, opened 2022-10-25
fix(datadog): Update msgpack accounting for sampling_priority

aengelas:
When @rex-remind101 added 3806258ee5db2223395fd38daac1d9e352974a17 he didn't update the length of the map encoded as msgpack.

This increases the numbers so that the messages are valid. I've confirmed this fixes v0.3 protocol messages.


aengelas:
I'm also happy to add some more robust tests around this, assuming you all agree this is the right fix.

djc:
This seems like a minimum viable fix. Isn't there some serde implementation for MessagePack or some other way of encoding things in the type system so that we can statically prevent this problem from happening in the future?

aengelas:
Yes, I believe there is. I can explore that, but meanwhile I believe this is broken for everyone using 0.6.0 of opentelemetry-datadog on 0.3 or 0.5 protocol versions. I'd like to merge the MVF prior to doing a refactor.

aengelas:
I'll go ahead and fix the tests.

aengelas:
Not sure why https://github.com/open-telemetry/opentelemetry-rust/actions/runs/3332558642/jobs/5513568207 is failing, but it looks like it's not required. I believe this is ready for review.

aengelas:
Let me know how you like handle rebasing, etc. I generally leave my commits separate until review, and then squash/rebase prior to merge.

TommyCpp (approved):
It looks reasonable to me. Thanks for the fix. A serde implementation would be helpful.

aengelas:
Okay, I've squashed and rebased; this should be good to go whenever tests pass; it doesn't look like I'm authorized to merge.

aengelas:
@TommyCpp Any chance you could merge this for me?

TommyCpp:
> @TommyCpp Any chance you could merge this for me?

Sorry missed it earlier. Could you merge the upstream or rebase on top of the `main` branch? I think that should clear the CI errors.

PR #907, opened 2022-11-03
handle parent_span_id in jaeger_json

themanforfree:
the `span.parent_span_id` is not handled in jaeger_json module, it will lose the parent-child relationship of spans when exporting.


djc (approved):
LGTM.

Issue #908, opened 2022-11-04
OTEL_EXPORTER_OTLP_ENDPOINT_DEFAULT is incorrect


BrynCooke:
According to the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) the otlp exporter default endpoint should not use https by default.
In addition, depending on the protocol, the port number changes.

Currently the default is set to https, but this only reveals itself to be an issue when tonic `tls-roots` is enabled.
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/mod.rs#L21

TommyCpp:
close with #912. Thanks again!

PR #912, opened 2022-11-07
Improve OTLP environment variable handling

### LINKED ISSUE #909: Support for OTEL_EXPORTER_OTLP_PROTOCOL
`OTEL_EXPORTER_OTLP_PROTOCOL` and `OTEL_EXPORTER_OTLP_*_PROTOCOL` are defined in the OT specification, see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol but aren't supported in opentelemetry-otlp.

It's not clear whether this is a missing feature, or a conscious decision by this library to make this a compile-time concept. Should this be added?

BrynCooke:
OTLP exporter default endpoint changed to http

There are a couple of places that the OTLP exporter differed from the spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md

1. OTEL_EXPORTER_OTLP_PROTOCOL was not supported.
2. The default endpoint used https rather than http.

The following has changed:
1. OTEL_EXPORTER_OTLP_PROTOCOL is supported.
2. If http-proto feature is enabled the default will be to use http-proto rather than grpc.
3. The default endpoint will be set according to the default protocol.
4. The default endpoint uses http rather than https.
5. Tests added.

Fixes #909 #908

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## Main
+
+### Changed
+
+- Improve OTLP exporter environment variable handling #911
+- OTLP exporter default endpoint changed to http #911
 
 ## v0.11.0
 
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -3,8 +3,8 @@
 
 ### Changed
 
-- Improve OTLP exporter environment variable handling #911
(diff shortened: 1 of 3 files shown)


BrynCooke:
Once https://github.com/open-telemetry/opentelemetry-rust/pull/911 is merged this PR needs to be rebased.

BrynCooke:
Rebased and CI is passing now.

TommyCpp (approved):
LGTM. Thanks!

PR #917, opened 2022-11-14
Add abbility to disable stdout / stderr output.

BratSinot:
Greetings!

Printing to stdout / stderr creates problem, when you export logs in certain format from service. So, why not to use `tracing` optionaly?

### DIFF
--- a/opentelemetry-api/Cargo.toml
+++ b/opentelemetry-api/Cargo.toml
@@ -18,6 +18,7 @@ once_cell = "1.12.0"
 pin-project-lite = { version = "0.2", optional = true }
 thiserror = "1"
 tokio-stream = { version = "0.1", optional = true }
+tracing = { version = "0.1", optional = true }
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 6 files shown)


djc:
Since opentelemetry is often used as the backend for traces, it doesn't obviously make sense to send errors from the opentelemetry crates into tracing as well (which could create a feedback loop).

BratSinot:
> Since opentelemetry is often used as the backend for traces, it doesn't obviously make sense to send errors from the opentelemetry crates into tracing as well (which could create a feedback loop).

Okey, then at least may be ability to disable output to stdout / stderr?

jtescher:
@BratSinot the intention with [global::set_error_handler](https://docs.rs/opentelemetry/latest/opentelemetry/global/fn.set_error_handler.html), is to allow you to customize this behavior. The jaeger and zipkin crates should be handling the error case through the global error handler, happy to merge a PR with that effect as that should help you silence or otherwise redirect these logs.

BratSinot:
> @BratSinot the intention with [global::set_error_handler](https://docs.rs/opentelemetry/latest/opentelemetry/global/fn.set_error_handler.html)

Thanks a lot! That what I was looking for (obviously searching not too good =) ).

PR #918, opened 2022-11-18
add response headers in response for HttpClient implementations

bnjjj:
When using `HttpClient` implementations you didn't return headers in http response.

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -138,10 +138,8 @@ pub mod surf {
             for header_name in response.header_names() {
                 for header_value in &response[header_name.to_string().as_str()] {
                     headers.append(
-                        HeaderName::from_str(&header_name.to_string())
-                            .expect("it was a header name before"),
-                        HeaderValue::from_str(header_value.as_str())
-                            .expect("it was a header value before"),
+                        HeaderName::from_str(&header_name.to_string())?,
+                        HeaderValue::from_str(header_value.as_str())?,
                     );
                 }
             }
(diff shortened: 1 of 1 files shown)


djc (commented):
LGTM. Your commit doesn't have a real email address set, maybe fix that?

bnjjj:
It is intentional to use this address. It's in your gh settings it generates you an email linked to your gh account but you can't be contacted with it. It's to avoid spams.

bnjjj:
For context https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address

opentelemetry-http/src/lib.rs
```diff
@@ -122,9 +134,24 @@ pub mod surf {
             }
 
             let mut response = self.send(request_builder).await?;
-            Ok(Response::builder()
+            let mut headers = HeaderMap::new();
+            for header_name in response.header_names() {
+                for header_value in &response[header_name.to_string().as_str()] {
+                    headers.append(
+                        HeaderName::from_str(&header_name.to_string())
+                            .expect("it was a header name before"),
```
TommyCpp, line 142: Can we convert it into `HttpError` here? Doesn't seems to right to break the application because one header cannot parse

bnjjj: Done, thanks :)

bnjjj:
@TommyCpp Do I need to care about the cargo-deny job failure ?

jtescher (approved):
Looks reasonable to me unless @djc or @TommyCpp have other comments

TommyCpp:
Nope LGTM. #919  should fix cargo deny issue so I think we can merge this

PR #919, opened 2022-11-21
chore(common): fix deny.

TommyCpp:
Most of the warnings from deny are unmaintained/deprecated create which we depend on indirectly either in dev or build

I don't believe there are any action item for us to fix them.

Also split the doc as separate task in CI. It's confusing to fail the coverage because of the doc failures

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,24 @@ jobs:
       - uses: EmbarkStudios/cargo-deny-action@v1
         with:
           command: check advisories
+  docs:
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: true
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          components: rustfmt
+          override: true
+      - uses: arduino/setup-protoc@v1
+      - name: doc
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -87,6 +87,23 @@ jobs:
       - uses: EmbarkStudios/cargo-deny-action@v1
         with:
           command: check advisories
+          log-level: error
```
TommyCpp, line 90: By default the log level is warning. Most of the warnings we get are deprecated repo/unmaintained repos that we depend on indirectly. Setting it to error can reduce noise while notifying us if there is a security issue. I am open to suggestions if we should leave it to warning vs change it to error

djc: Hmm, I think you can add yanked/deprecated to an allow list, and I think that might be better? If I understand correctly this would also silence warnings about direct dependencies that are yanked/unmaintained, which seems undesirable.

TommyCpp: I think if we allow yanked/deprecated it means we allow it for both direct dependencies and indirect dependencies. I don't think cargo deny differentiate direct dependencies vs indirect dependencies

TommyCpp:
I think we can merge it as it is and open another PR if we find a better solution to notify us there are unmaintained crates in our dependency tree while not blocking PRs from merging.

hdost (approved):

PR #921, opened 2022-11-25
Implement Display on Baggage

### LINKED ISSUE #905: Encode Baggage as string
Baggage type does not support encoding into a string compliant with the W3C specification.  See similar API method in Go implementation: https://pkg.go.dev/go.opentelemetry.io/otel@v1.10.0/baggage#Baggage.String

wperron:
Tests are taken from the [Go test cases][1]

Fixes #905

[1]: https://github.com/open-telemetry/opentelemetry-go/blob/v1.10.0/baggage/baggage_test.go#L414

### DIFF
--- a/opentelemetry-api/src/baggage.rs
+++ b/opentelemetry-api/src/baggage.rs
@@ -285,21 +285,18 @@ impl FromIterator<KeyValueMetadata> for Baggage {
 
 impl fmt::Display for Baggage {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let mut v: Vec<String> = self
-            .into_iter()
-            .map(|(k, v)| {
-                let mut val = encode(&v.0.as_str()).to_string();
-                if !v.1.as_str().is_empty() {
-                    val.push_str(format!(";{}", v.1.as_str()).as_str())
-                }
+        for (i, (k, v)) in self.into_iter().enumerate() {
+            write!(f, "{}={}", k, encode(&v.0.as_str()))?;
+            if !v.1.as_str().is_empty() {
+                write!(f, ";{}", v.1)?;
+            }
 
-                format!("{}={}", k, val)
-            })
-            .collect();
+            if i < self.len() - 1 {
+                write!(f, ",")?;
+            }
+        }
 
-        v.sort();
-        let s = v.join(",");
-        Ok(write!(f, "{}", s)?)
+        Ok(())
     }
 }
 
@@ -566,8 +563,5 @@ mod tests {
         b.insert_with_metadata("foo", StringValue::from("1"), "red;state=on");
         b.insert_with_metadata("bar", StringValue::from("2"), "yellow");
         assert_eq!("bar=2;yellow,foo=1;red;state=on", b.to_string());
-
-        // assert it implements Display
-        let _ = format!("{}", b);
     }
 }
--- a/opentelemetry-api/src/baggage.rs
+++ b/opentelemetry-api/src/baggage.rs
@@ -562,6 +562,10 @@ mod tests {
         let mut b = Baggage::default();
         b.insert_with_metadata("foo", StringValue::from("1"), "red;state=on");
         b.insert_with_metadata("bar", StringValue::from("2"), "yellow");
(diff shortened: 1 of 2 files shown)


opentelemetry-api/src/baggage.rs
```diff
@@ -280,6 +283,26 @@ impl FromIterator<KeyValueMetadata> for Baggage {
...
+                let mut val = encode(&v.0.as_str()).to_string();
+                if !v.1.as_str().is_empty() {
+                    val.push_str(format!(";{}", v.1.as_str()).as_str())
+                }
+
+                format!("{}={}", k, val)
+            })
+            .collect();
+
+        v.sort();
```
djc, line 300: Is the sorting only intended for the tests? If we didn't have to sort you could write directly into the `Formatter`, which would be substantially better for performance.

wperron: I'm not 100% to be honest, I looked at the Go tests for this, and noticed `foo` being added before `bar` when inserting in the Baggage, yet `bar` coming first in the serialized output. I assume that makes sense in Go because of the non-deterministic key ordering in maps, but I'm not sure if it really has its place here. You're probably more familiar with the spec than I am, if you think it unnecessary I can just remove it simply `write` directly to the Formatter like you say.

djc: I skimmed [a spec I found](https://www.w3.org/TR/baggage/) which doesn't seem to require any particular order. If anything it seems to recommend preserving order through mutations, so maybe storing in a `HashMap` is not the best approach? But in any case, I think it would be better not to sort here and instead makes the tests robust to random ordering.

opentelemetry-api/src/baggage.rs
```diff
@@ -494,4 +522,50 @@ mod tests {
...
+
+        // "two members with properties"
+        let mut b = Baggage::default();
+        b.insert_with_metadata("foo", StringValue::from("1"), "red;state=on");
+        b.insert_with_metadata("bar", StringValue::from("2"), "yellow");
+        let stringified = b.to_string();
+        assert!(
+            stringified == "bar=2;yellow,foo=1;red;state=on"
+                || stringified == "foo=1;red;state=on,bar=2;yellow"
+        )
```
wperron, line 569: Removing the sort makes `.to_string()` non deterministic, which is all fine and dandy but testing is a bit harder, idk, this feels a bit heavy handed but I'll let you be the judge of that 😂

opentelemetry-api/src/baggage.rs
```diff
@@ -17,7 +17,10 @@
 use crate::{Context, Key, KeyValue, Value};
 use once_cell::sync::Lazy;
 use std::collections::{hash_map, HashMap};
+use std::fmt;
 use std::iter::FromIterator;
+use urlencoding::encode;
+// use urlencoding::encode;
```
jtescher, line 23:
```suggestion
```

jtescher (changes requested):

opentelemetry-api/src/baggage.rs
djc: I'd maybe do something like `stringified.contains("bar=2;yellow") && stringified.contains("foo=1;red;state=on")` which feels more robust/obvious. But I do think shifting the complexity from library to test is probably for the best!

djc (approved):

opentelemetry-api/src/baggage.rs
wperron: Good catch thanks! 👍

wperron:
Is that failing test a known flaky test? it's passing locally and I'm not sure why it'd be failing here 🤔

CleanCut:
@wperron Hmm, `cargo test` passes for me as well. But if I run `script/test.sh` then it fails just as in CI, and it changes three files (which need to be committed and added to the PR according to the output).

wperron:
Ah makes sense, I'll get that updated sometime today 👍

TommyCpp (approved):

TommyCpp:
@jtescher Looks like we need your approval to merge. Any more comments or it's good to merge?

jtescher (approved):

PR #933, opened 2022-12-11
chore(proto): update proto generated files

TommyCpp:


jtescher (approved):

hdost:
Is there a benefit to not having derive Eq?

TommyCpp:
> Is there a benefit to not having derive Eq?

I think the choice was made by tonic authors. I'd say it's just to keep the generated code minimal and allow user to implement the trait as needed

PR #937, opened 2022-12-26
fix(otlp): metadata is not set correctly

### LINKED ISSUE #935: Bug: opentelemetry_otlp::exporter::tonic::TonicExporterBuilder::with_metadata ignores the provided metadata
I am currently tracing through a bug in which the Rust opentelemetry libraries are failing to export to Honeycomb (which has been an *extremely* painful experience to debug since there is nowhere to dump all traffic, and I have resorted to patching all of h2, tonic, and opentelemetry-rust).

I have arrived at 

```rust
    // gets them from the env
    let tracer_meta = get_tracer_headers();
    dbg!(&tracer_meta); // yes, they are present here

    let otlp_tracer = opentelemetry_otlp::new_pipeline()
        .tracing()
        .with_exporter(
                .tonic()
                .with_env()
                .with_metadata(tracer_meta),
        )
        .with_trace_config(
            opentelemetry_sdk::trace::config()
                .with_resource(Resource::new([KeyValue::new("service.name", "nix-otel")])),
        )
        .install_batch(opentelemetry_sdk::runtime::Tokio)
        .wrap_err("Failed to install `opentelemetry` tracing pipeline")?;
    let ctx = opentelemetry::Context::new();
    let mut span = otlp_tracer.start_with_context("root span", &ctx);
    span.set_attribute(KeyValue {
        key: Key::from("wtf"),
        value: Value::String("yeah".into()),
    });
    span.end();
    ctx.with_span(span);
    drop(ctx);

    if let Some(p) = otlp_tracer.provider() {
        for x in p.force_flush() {
            x.unwrap();
        }
    }
```

Then:

```
DEBUG received
  frame=Headers { stream_id: StreamId(1), flags: (0x5: END_HEADERS | END_STREAM), fields: {"server": "aw
selb/2.0", "date": "Tue, 13 Dec 2022 00:41:51 GMT", "content-type": "application/grpc", "content-length"
: "0", "grpc-status": "16", "grpc-message": "missing 'x-honeycomb-team' header"} }
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: ExportFailed(Status { code: Unauthenticated, mess
age: ", detailed error message: missing 'x-honeycomb-team' header" })
```

So I added some debug statements to opentelemetry-otlp in the tonic exporter function creator:

```
[opentelemetry-rust/opentelemetry-otlp/src/exporter/tonic.rs:69] &self.tonic_config.metadata = Some(
    MetadataMap {
        headers: {
            "user-agent": "OTel OTLP Exporter Rust/0.11.0",
        },
    },
)
```

So what's happening is that .get_or_insert is doing nothing since there's already metadata present: the default user agent!

https://github.com/open-telemetry/opentelemetry-rust/blob/5ad9dcdf914a2dc73794fe33cbb9678693381a26/opentelemetry-otlp/src/exporter/tonic.rs#L65-L71

I think this should be release blocking; it is definitely a regression since opentelemetry-otlp 0.11.0 since I copy pasted this code from nix-otel, which definitely works with this code on 0.11.0.

TommyCpp:
Fix #935.

Also update the proto & fix the lint issue raised with rust version 1.66

### DIFF
--- a/opentelemetry-datadog/src/exporter/mod.rs
+++ b/opentelemetry-datadog/src/exporter/mod.rs
@@ -369,8 +369,7 @@ fn group_into_traces(spans: Vec<SpanData>) -> Vec<Vec<SpanData>> {
     spans
         .into_iter()
         .into_group_map_by(|span_data| span_data.span_context.trace_id())
-        .into_iter()
-        .map(|(_, trace)| trace)
+        .into_values()
         .collect()
 }
 
--- a/opentelemetry-datadog/src/exporter/model/v05.rs
+++ b/opentelemetry-datadog/src/exporter/model/v05.rs
@@ -77,7 +77,7 @@ where
     let mut payload = Vec::new();
     rmp::encode::write_array_len(&mut payload, 2)?;
 
-    rmp::encode::write_array_len(&mut payload, interner.len() as u32)?;
+    rmp::encode::write_array_len(&mut payload, interner.len())?;
     for data in interner.iter() {
         rmp::encode::write_str(&mut payload, data)?;
     }
--- a/opentelemetry-otlp/src/exporter/grpcio.rs
+++ b/opentelemetry-otlp/src/exporter/grpcio.rs
@@ -111,3 +111,35 @@ impl GrpcioExporterBuilder {
         self
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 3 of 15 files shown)


jtescher (approved):
Nice

PR #946, opened 2023-01-16
fix(metrics): remove in_memory settings

### LINKED ISSUE #939: Metrics counter with delta_temporality_selector unexpected behavior
I was trying to work with OTLP counters with the delta_temporality_selector, and found an unexpected behavior. I wanted to check if I am missing some config.

I created a basic opentelemetry-prometheus exporter to capture the metrics, and created a counter. The delta aggregations works as long as there were some activity on the counter between the export requests. However if there was no counter operations between two export requests, I would expect the counter to report `0`, but instead it was reporting the value from the previous reported request.

TommyCpp:
The `memory` settings are used to control whether the processor should process the instrument that is not updated in this collection interval.  

Users can achieve the same behavior using temporality. It also causes bugs when memory is `true` but the temporality is `Delta`. In this case, I think we should delete the record if the instrument didn't update. But if `memory` is `true` it will prevent the stale record from deleting.

**Is this fix #939?**
Well, sort of. The root cause of #939 is we don't delete the stale record for `Delta` temporality because the memory is set to `true`, which should be fixed now that we remove the memory settings. However, for #939 the bigger issue is we shouldn't allow user to use `Detla` temporality with Prometheus exporter. 

> A Prometheus Exporter MUST only support [Cumulative Temporality](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#temporality).

### DIFF
--- a/examples/hyper-prometheus/src/main.rs
+++ b/examples/hyper-prometheus/src/main.rs
@@ -76,13 +76,10 @@ struct AppState {
 
 #[tokio::main]
 pub async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    let controller = controllers::basic(
-        processors::factory(
-            selectors::simple::histogram([1.0, 2.0, 5.0, 10.0, 20.0, 50.0]),
-            aggregation::cumulative_temporality_selector(),
-        )
-        .with_memory(true),
-    )
+    let controller = controllers::basic(processors::factory(
+        selectors::simple::histogram([1.0, 2.0, 5.0, 10.0, 20.0, 50.0]),
+        aggregation::cumulative_temporality_selector(),
+    ))
     .build();
 
     let exporter = opentelemetry_prometheus::exporter(controller).init();
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -15,7 +15,6 @@
 //!             selectors::simple::histogram([1.0, 2.0, 5.0, 10.0, 20.0, 50.0]),
 //!             aggregation::cumulative_temporality_selector(),
 //!         )
-//!         .with_memory(true),
 //!     )
 //!     .build();
 //!
--- a/opentelemetry-prometheus/tests/integration_test.rs
+++ b/opentelemetry-prometheus/tests/integration_test.rs
@@ -9,13 +9,10 @@ use prometheus::{Encoder, TextEncoder};
 #[test]
 fn free_unused_instruments() {
     let cx = Context::new();
-    let controller = controllers::basic(
-        processors::factory(
-            selectors::simple::histogram(vec![-0.5, 1.0]),
-            aggregation::cumulative_temporality_selector(),
-        )
-        .with_memory(true),
-    )
+    let controller = controllers::basic(processors::factory(
+        selectors::simple::histogram(vec![-0.5, 1.0]),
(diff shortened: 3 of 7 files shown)


jtescher (approved):

davidhoyt:
Will this affect gauges?

davidhoyt:
What will happen if this stops reporting counter values for deltas? e.g. say at a regular reporting interval we're pushing metrics. So we have a counter where at T0 (time 0)=0, it goes up by 30 so T1 reports T1=30, then it doesn't report because it's not changed ever. So T2=null, T3=null, etc. If I'm graphing this out and I try to use my last known value, it'd be 30. Shouldn't it at least emit one more time (T2=0) before not emitting again until the value changes?

TommyCpp:
> Will this affect gauges?

That's a good point. I don't think so but I can add some tests in this PR 

> Shouldn't it at least emit one more time (T2=0) before not emitting again until the value changes?

Note that in delta temporality, we will export the metrics when there is **any change** for the instrument even if the value is 0. So if we emit one more time after the final round of change it can be confusing. i.e `null` means "there is no change in this round", and `0` means "the change is 0 in this round". 

If your use case doesn't need to differentiate `null` and `0` I think most dashboards have the ability to convert the null to 0 for you :slightly_smiling_face:

davidhoyt:
That's my concern, that there is a hidden delta here that isn't emitted. Null, to me, isn't the same as zero. Zero is an assertion and null implies lack of information/data. Here there's a known change -- the count has reverted to zero but it isn't asserted.

TommyCpp:
I see your point. 
> Here there's a known change -- the count has reverted to zero but it isn't asserted.

For `Delta` temporaliy this is implicit. That logically the SDK "forget" everything after each collection. The `Delta` here is comapre against a initial value(for counter, 0) instead of last know value. 

So given the following data points:
`[T0, T1, 30], [T1, T2, 20]`, the total sum between T0 and T2 is 50

I think one thing worth mention is in OTLP metrics data model. The aggregated metrics are associated with a range of time instead of a timestamp. The difference of temporality is just when this range starts. For `Cumulative`, the start of the range will always be `T0`(in other words, the start time of process). For `Delta`, the start time will be the end of last collection interval. 

So in your previous example, if SDK exporting
- `[T0, T1, 30], [T1, T2, 0]`, it means the sum of the counter between T0 and T1 is 30; the sum of the counter between T1 and T2 is 0. 
- `[T0, T1, 30]`, it means the counter has not been called since T1. 
- `[T0, T1, 30], [T7, T8, 20]` means the counter has been called since T1 until sometime between T7 and T8.

davidhoyt:
Thank you for your patient/full explanation -- it was very helpful. 😄 

In your example, is there any way to force it to report 0s for T2-T7 for the counters it knows about? Or at least a way for an exporter to access the available counters (even if they haven't been changed)?

TommyCpp:
> Or at least a way for an exporter to access the available counters (even if they haven't been changed)?

Yeah in the exporter implementation can "remember" all seen instruments should it choose to. I am working on refactoring metrics APIs so that we are align with the spec. 

TBH I am a little worried that if we allow SDK to "remember" seen instruments it can easily cause memory issues. For example, if someone includes the request id as an attribute accidentally the number of instruments will grow along with number of requests even if all instruments will only have one data point.

PR #948, opened 2023-01-17
Only run ParentBased delegate sampler when there is no parent

SebastiaanYN:
The delegate sampler is always called even when it's not the root span. Replacing it with `map_or_else` causes it to only run when there's no parent span.


jtescher (approved):
Thanks

SebastiaanYN:
I believe the CI test failed because of a ratelimit. Could those tests be manually restarted?

PR #949, opened 2023-01-17
Add Using Warp API + Custom Tracing + Prom Metrics

fmetaphcode:
Hello everyone! This is my first contribution! Sorry for some mistake. Here is my simple work.

- In this example, we will work with trance events that will be sent to Jaeger.
- Will also show data scraping by Prometheus through an exclusive port for metrics service.
- However, it will be presented how to separate distinct routes for Clients and Metrics.


TommyCpp:
Thanks for working on this! The lint issue is about the format, run `cargo format` should resolve them.

djc:
I'm honestly not sure we should be adding more examples like this to the tree, especially examples that are this elaborate. So far we've seen that the examples don't get maintained very much in the future. Especially for Warp, which IMO is not very popular these days, I don't think carrying the burden of maintenance makes sense for the OpenTelemetry-rs team.

On the other hand, if there are more tightly scoped documentation changes or examples in docstrings that would have made your work easier (like, made it less necessary to add an example like this), those would be great to have!

fmetaphcode:
> Thanks for working on this! The lint issue is about the format, run `cargo format` should resolve them.
> 
> As for the `cargo deny` error, it's probably because you have some dependence that is using `time 0.1.x` crate. Could you check if you can update dependencies so that `time 0.1` will not be an indirect dependency?

Nice! I'm running lint locally to preview them =/

fmetaphcode:
Let's close this question to focus on more important things right now. Show! Excited for the next contribution.

PR #951, opened 2023-01-18
Fix formatting in ShouldSample docs

fracek:
As I was going through the docs I noticed this typo made the [hard to read](https://docs.rs/opentelemetry/latest/opentelemetry/sdk/trace/trait.ShouldSample.html).


jtescher (approved):
Thanks

PR #960, opened 2023-01-29
Update semantic conventions to spec version 1.17.0

jtescher:
Also add mark deprecated keys using `#[deprecated]`


TommyCpp (approved):
Need to update some of the docs but otherwise LGTM

PR #969, opened 2023-02-13
Update some dependencies

djc:
Here are some of the easy updates from #922. I left out:

* All of the `examples`
* Updates in -contrib (base64)
* Updates in -jaeger (thrift, tonic, prost, prost-types)
* Updates in -otlp (grpc, protobuf)
* Updates in -proto (protobuf, protobuf-codegen)

For the examples, I think we have way too many example projects to reasonably maintain over time, so I think we should prune back on how many things we want to explicitly support. It would probably help if we move examples related to a particular integration crate into that crate, so that PRs are forced to update related examples. (In many cases, it looks like the example projects mainly duplicate example code from the doctests or tests in the integration crates...) This also conveniently allows me to avoid dealing with [unrelated changes](https://github.com/open-telemetry/opentelemetry-rust/pull/922#discussion_r1037053029) suggested in review.

As for grpc and protobuf, it looks like the grpcio maintainers are [uninterested](https://github.com/tikv/grpc-rs/issues/584) in updating grpcio to use protobuf 3. Personally, I think the prost/tonic stack is more attractive in any way, so I question the need for this project to maintain two gRPC integrations in parallel (and am personally not willing to spend time on that).

The base64 crate unfortunately [made their API worse](https://github.com/marshallpierce/rust-base64/issues/205) in recent releases, and the maintainer is unwilling to improve on that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
     - uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: 1.57
+        toolchain: 1.60.0
         override: true
     - name: Patch dependencies versions # some dependencies bump MSRV without major version bump
       run: ./scripts/patch_dependencies.sh
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ analysis in order to understand your software's performance and behavior. You
 can export and analyze them using [Prometheus], [Jaeger], and other
 observability tools.
 
-*Compiler support: [requires `rustc` 1.57+][msrv]*
+*Compiler support: [requires `rustc` 1.60+][msrv]*
 
 [Prometheus]: https://prometheus.io
 [Jaeger]: https://www.jaegertracing.io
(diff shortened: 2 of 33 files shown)


hdost:
Perhaps the comments about pruning, could be an issue raised. I would agree that the surface area of the project should be reduced, and I believe I had made similar comments in the past when I mentioned making some of the code build separately. By example moving code into a separate contrib repo (might have been a different suggestion, but the core argument is the same) keep the core well maintained and allow for contribution on but keep support low on non-core components.

TommyCpp:
I took a stab to update some dependencies in some of the creates. 

I second the idea to move the example to their crates and having a contrib repo to store less core creates. I can open an issue with admins to create a new repo and assign proper access to maintainer/approvers

Looking at the [issue](https://github.com/tikv/grpc-rs/issues/584) on grpcio. It seems they are still open to adding support for proto 3? I am open to removing grpcio support but I don't think we need to figure it out now. I feel like we can open an issue to gather some ideas on if anyone is hard depending on grpcio

TommyCpp:
https://github.com/djc/opentelemetry-rust/pull/2
Bump MSRV to 1.60 as it's required by tonic 0.8

TommyCpp (approved):
LGTM. Thanks

djc:
Thank you for looking into the MSRV stuff and the other crates!

PR #974, opened 2023-02-26
feat: add otel_scope_info and scope labels

zengxilong:
PR for #930 

1. Add otel_scope_name and otel_scope_version attributes to all metrics based on the instrumentation scope name and version.
2. Add an otel_scope_info metric.
3. Add prometheus exporter config, provide a disable_scope_info flag to disable the otel_scope_info metric and otel_scope_ labels.
4. I will make a new PR to resolve collisions between TYPE, HELP, and UNIT comments.

### DIFF
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -86,7 +86,7 @@ use opentelemetry::sdk::metrics::sdk_api::Descriptor;
 #[cfg(feature = "prometheus-encoding")]
 pub use prometheus::{Encoder, TextEncoder};
 
-use opentelemetry::global;
+use opentelemetry::{global, InstrumentationLibrary, StringValue};
 use opentelemetry::sdk::{
     export::metrics::{
         aggregation::{Histogram, LastValue, Sum},
@@ -240,7 +240,7 @@ impl prometheus::core::Collector for Collector {
                     let instrument_kind = record.descriptor().instrument_kind();
 
                     let desc = get_metric_desc(record);
-                    let labels = get_metric_labels(record, controller.resource());
+                    let labels = get_metric_labels(record, controller.resource(),_library);
 
                     if let Some(hist) = agg.as_any().downcast_ref::<HistogramAggregator>() {
                         metrics.push(build_histogram(hist, number_kind, desc, labels)?);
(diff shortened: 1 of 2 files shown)


opentelemetry-prometheus/src/lib.rs
```diff
@@ -175,6 +204,22 @@ impl ExporterBuilder {
...
+#[derive(Debug, Default)]
+pub struct ExporterConfig {
+    /// disable the otel_scope_info metric and otel_scope_ labels.
+    disable_scope_info: bool,
+}
+
+impl ExporterConfig {
+    /// Set disable_scope_info for [`ExporterConfig`].
+    /// It's the flag to disable the otel_scope_info metric and otel_scope_ labels.
+    pub fn disable_scope_info(mut self) -> Self {
```
TommyCpp, line 217: I know we should default to `false` here but still may be worth to allow the user to pass a bool to change the behavior

TommyCpp (commented):
Thanks for working on this. Overall like a good start. I think maybe use something like `with_scope_info` will be more clear. And we can have the default value to be `true`

opentelemetry-prometheus/src/lib.rs
zengxilong: So does `otel_scope_info` need to be disabled by default?

hdost: So based on the spec it should be on by default https://github.com/open-telemetry/opentelemetry-specification/blob/0d9081669c0c087c092bf35fe5c5652c2b505b94/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1
but I agree that  `with_scope_info`  would be  more inline with how the interfaces are working.

zengxilong: > So based on the spec it should be on by default https://github.com/open-telemetry/opentelemetry-specification/blob/0d9081669c0c087c092bf35fe5c5652c2b505b94/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1 but I agree that `with_scope_info` would be more inline with how the interfaces are working.

I changed `disable_scope_info(mut self) -> Self` to `with_scope_info(mut self, enabled: bool) -> Self`, and let the default value be true.

hdost (approved):

Issue #915, opened 2022-11-09
opentelemetry_otlp: Allow simple configuration of `BatchConfig` for install_batch


MidasLamb:
Currently, you can quite easily install a batch processor by using `install_batch`, however, if you want to have any sort of restrictions/needs to change how the batching runs, you can't just simply pass the `BatchConfig` that is used under the hood.

A user has provided a solution on stack overflow: https://stackoverflow.com/questions/74264248/opentelemetry-rust-increasing-the-channel-size-of-batched-otlptracepipeline.

I think it would be nice for users to easily configure the batch installation as with the implementation above, by simply having a `BatchConfig` object to pass along.

wperron:
I could be wrong but my understanding is that you can (and should) use environment variables to configure the batch configuration. At least that's what [the spec implies][1] and it's what's implemented in the [`Default` implementation for `BatchConfig`][2].

[1]: https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/#batch-span-processor
[2]: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/span_processor.rs#L509

zengxilong:
[`opentelemetry-jaeger::CollectorPipeline`](https://github.com/open-telemetry/opentelemetry-rust/blob/6500d04883e904991323f22014a01e369ccd8dd4/opentelemetry-jaeger/src/exporter/config/collector/mod.rs#L92) already has this function [`with_batch_processor_config `](https://github.com/open-telemetry/opentelemetry-rust/blob/6500d04883e904991323f22014a01e369ccd8dd4/opentelemetry-jaeger/src/exporter/config/collector/mod.rs#L421) to  config `install_batch`, maybe I can add a implementation to opentelemetry_otlp.

cijothomas:
`BatchConfig` can be used to build a required BatchExportProcessor and add it via `with_span_processor`. Also ENV variables some control, which I think is good enough?

(We removed the install_batch methods previous, so this may not be applicable anymore. Please feel free to re-open/comment if we there are suggestions on improving configuration aspect.)

PR #979, opened 2023-03-01
feat: add batch config for opentelemetry otlp pipeline

zengxilong:
PR for #915

- Add a `batch_config` field to the `OtlpTracePipeline`, and add a `with_batch_config` to set the batch span processor configuration for `OtlpTracePipeline`.
- build `BatchSpanProcessor` with `batch_config` in `build_batch_with_exporter`.


TommyCpp (approved):
Looks good to me. Maybe it's worth to call out in function document that it will override the env vars.

PR #991, opened 2023-03-12
suggestion: CARGO_BIN_NAME instead of unknown_service as default serv…

### LINKED ISSUE #942: suggestion: use env!("CARGO_BIN_NAME") instead of "unknown_service" as default service name
Currently, when there is no service name defined, this library annotates everything with `service_name="unknown_service"`, which is not very useful. It could simply not include service_name, or even better, use the value of `CARGO_BIN_NAME` (defined at runtime) to extract a meaningful name.

danielschaffel:
closes #942

### DIFF
--- a/opentelemetry-sdk/src/resource/env.rs
+++ b/opentelemetry-sdk/src/resource/env.rs
@@ -9,7 +9,6 @@ use std::time::Duration;
 
 const OTEL_RESOURCE_ATTRIBUTES: &str = "OTEL_RESOURCE_ATTRIBUTES";
 const OTEL_SERVICE_NAME: &str = "OTEL_SERVICE_NAME";
-const CARGO_BIN_NAME: &str = "CARGO_BIN_NAME";
 
 /// Resource detector implements ResourceDetector and is used to extract
 /// general SDK configuration from environment.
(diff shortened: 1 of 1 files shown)


hdost (changes requested):
Looks great! can you please update the comment docs on the `SdkProvidedResourceDetector` struct so that it matches the functionality as its is no longer using doing the provided steps?

djc:
Please note that there is a difference between `env!("CARGO_BIN_NAME")` and `env::var("CARGO_BIN_NAME")`. The former gets the binary name at compile-time through a variable set by Cargo, while the former gets the value of the environment variable at run-time. IMO the former is useful for this, but the latter isn't.

danielschaffel:
From what I understand using `env!("CARGO_BIN_NAME")` would make it impossible to build and test the `opentelemetry-sdk` package. This is because it is a `lib` not a `bin` so this method will always fail and cause an error. I chose to use `option_env` and not add a test because I don't know how I would test something that only happens when building the package as part of a library.

opentelemetry-sdk/src/resource/env.rs
```diff
@@ -88,7 +88,10 @@ impl ResourceDetector for SdkProvidedResourceDetector {
                         .get(Key::new("service.name"))
                         .map(|v| v.to_string())
                         .filter(|s| !s.is_empty())
-                        .unwrap_or_else(|| "unknown_service".to_string())
+                        .unwrap_or_else(|| match option_env!("CARGO_BIN_NAME") {
+                            Some(s) => s.to_string(),
```
hdost, line 92: Also I apologize for glossing over this before but the idea is to use `env!()` which should always produce a value although https://doc.rust-lang.org/std/macro.env.html would likely serve us best just in case. 
This will grab it at build time whereas the current configuration grabs at runtime.

hdost (changes requested):
Again I apologize for missing this on the first review 😔

opentelemetry-sdk/src/resource/env.rs
danielschaffel: Sorry if I am misunderstanding what you say, but won't using `env!` in a library always fail? Where as using `option_env!` allows for it to compile as a `lib` and will resolve properly when being used as part of a `bin`. 

Please correct me if I am missing or misunderstanding something here.

hdost (approved):

opentelemetry-sdk/src/resource/env.rs
hdost: I think my review was a delayed, I hadn't seen djc's response 😓

Issue #995, opened 2023-03-13
Deprecating opentelemetry-jaeger


cijothomas:
The current project for OpenTelemetry jaeger (https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-jaeger) seems to be doing 2 things:
1. Jaeger propagator
2. Jaeger exporter.

Proposing to do:
1. Given the Jaeger exporter is slated to be deprecated (https://github.com/open-telemetry/opentelemetry-specification/pull/2858), proposing to remove the Jaeger exporter from this, and instead advice users to move to OTLP.

2. The spec recommends Jaeger propagator to be shipped as an extension package (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#propagators-distribution).
Once the exporter is removed, this package/crate can be renamed to reflect its actual purpose, which would be just the propagator. (a name change might also be needed)

cijothomas:
_We also need some documents and examples around how to switch._

Adding a PR showing that for reference, if needed in this SIG : https://github.com/open-telemetry/opentelemetry-dotnet/pull/3940/files
It also shows it was very straightforward to make the switch.

cijothomas:
> The actix examples still use the jeager, so there are no example showing how to use the recommended otlp exporters with actix or hyper.

Yes. This was discussed in today's SIG meeting. I'll be sending PR to update example to use OTLP Exporter for Jaeger.

cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/pull/3567/files Spec has removed Jaeger exporter specs. Most other languages have removed. I'll start the clean up in this repo.

cijothomas:
Tagging with help-wanted if anyone willing to contribute!

cijothomas:
The next step would be to create a separate crate for the propagator alone, "opentelemetry-propagator-jaeger", extracting the propagation part from opentelemetry-jaeger crate.

mattbodd:
@cijothomas I'd be interested in taking on this work

cijothomas:
> @cijothomas I'd be interested in taking on this work

Thanks! https://github.com/open-telemetry/opentelemetry-rust/issues/995#issuecomment-1851173398 lists the next step to be followed, and after that, we can work towards deprecating the current one. Feel free to share anyone alternate ways of handling the deprecation in a smooth way!

PR #996, opened 2023-03-14
Add warning to jaeger docs about future deprecation:

cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-rust/issues/995.
This does not deprecate anything. Instead a warning about a future deprecation is added to the docs, so users can start switch to OTLP one sooner.

Note:
https://github.com/rust-lang/rust/pull/106561 is not yet merged, so I could not find any better way to put a "Warning" thing into the doc. Open to any suggestions!


hdost (approved):
We've been talking about creating a contrib potentially even if after we decide to remove we could always move it to contrib.

PR #998, opened 2023-03-16
implement w3c trace context response propagation

### LINKED ISSUE #523: Support injecting a context in to a response
The latest W3C draft supports injecting the trace context in to an HTTP response.

https://w3c.github.io/trace-context/#trace-context-http-response-headers-format

This is the same format as the request context, except:

1. There's no `tracestate` header.
2. The header with the context is called `traceresponse`

Suggest:

1. Adding an `inject_response_context` method to `TextMapPropagator` that implements this.
2. Rename `inject_context` to `inject_request_context`
3. Keep `inject_context` as a (deprecated) wrapper around `inject_request_context` that can be removed in the next release that allows breaking changes

Quick and dirty implementation (where the common code between this and `inject_context` should obviously be factored out).

```
/// Inject the current span context for use in a response.
///
/// This follows the current W3C draft proposal for a `traceresponse`
/// header, see https://w3c.github.io/trace-context/#trace-context-http-response-headers-format
fn inject_response_context(cx: &opentelemetry::Context, injector: &mut dyn opentelemetry::propagation::Injector) {
    use opentelemetry::trace::{TraceContextExt, TRACE_FLAG_SAMPLED};

    const TRACERESPONSE_HEADER: &str = "traceresponse";
    const SUPPORTED_VERSION: u8 = 0;

    let span = cx.span();
    let span_context = span.span_context();
    if span_context.is_valid() {
        let header_value = format!(
            "{:02x}-{:032x}-{:016x}-{:02x}",
            SUPPORTED_VERSION,
            span_context.trace_id().to_u128(),
            span_context.span_id().to_u64(),
            span_context.trace_flags() & TRACE_FLAG_SAMPLED
        );
        injector.set(TRACERESPONSE_HEADER, header_value);
    }
}
```

wperron:
Adds a `TraceContextResponsePropagator` to the contrib crate implementing the [W3C Trace Context Response Header].

Fixes #523

[W3C Trace Context Response Header]: https://w3c.github.io/trace-context/#traceresponse-header

### DIFF
--- a/examples/traceresponse/README.md
+++ b/examples/traceresponse/README.md
@@ -1,0 +1,29 @@
+# HTTP Example
+
+This is a simple example using [hyper] that demonstrates tracing http request
+from client to server, and from the server back to the client using the
+[W3C Trace Context Response] header. The example shows key aspects of tracing
+such as:
+
+- Root Span (on Client)
+- Child Span (on Client)
+- Child Span from a Remote Parent (on Server)
+- SpanContext Propagation (from Client to Server)
+- SpanContext Propagation (from Server to Client)
+- Span Events
+- Span Attributes
+
+[hyper]: https://hyper.rs/
+[W3C Trace Context Response]: https://w3c.github.io/trace-context/#traceresponse-header
+
+## Usage
+
+```shell
+# Run server
+$ cargo run --bin http-server
+
+# In another tab, run client
+$ cargo run --bin http-client
+
+# The spans should be visible in stdout in the order that they were exported.
+```
--- a/examples/traceresponse/README.md
+++ b/examples/traceresponse/README.md
@@ -6,7 +6,6 @@ from client to server, and from the server back to the client using the
 such as:
 
 - Root Span (on Client)
-- Child Span (on Client)
 - Child Span from a Remote Parent (on Server)
 - SpanContext Propagation (from Client to Server)
 - SpanContext Propagation (from Server to Client)
--- a/examples/traceresponse/src/client.rs
+++ b/examples/traceresponse/src/client.rs
@@ -1,0 +1,66 @@
+use hyper::http::HeaderValue;
+use hyper::{body::Body, Client};
+use opentelemetry::global;
+use opentelemetry::propagation::TextMapPropagator;
+use opentelemetry::sdk::export::trace::stdout;
+use opentelemetry::sdk::{
+    propagation::TraceContextPropagator,
+    trace::{self, Sampler},
+};
+use opentelemetry::{
+    trace::{TraceContextExt, Tracer},
+    Context, KeyValue,
+};
+use opentelemetry_contrib::trace::propagator::trace_context_response::TraceContextResponsePropagator;
+use opentelemetry_http::{HeaderExtractor, HeaderInjector};
+
+fn init_tracer() -> impl Tracer {
+    global::set_text_map_propagator(TraceContextPropagator::new());
+    // Install stdout exporter pipeline to be able to retrieve the collected spans.
+    // For the demonstration, use `Sampler::AlwaysOn` sampler to sample all traces. In a production
+    // application, use `Sampler::ParentBased` or `Sampler::TraceIdRatioBased` with a desired ratio.
+    stdout::new_pipeline()
+        .with_trace_config(trace::config().with_sampler(Sampler::AlwaysOn))
+        .install_simple()
+}
+
+#[tokio::main]
+async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    let _tracer = init_tracer();
+
+    let client = Client::new();
+    let span = global::tracer("example/client").start("say hello");
+    let cx = Context::current_with_span(span);
+
+    let mut req = hyper::Request::builder().uri("http://127.0.0.1:3000");
+    global::get_text_map_propagator(|propagator| {
+        propagator.inject_context(&cx, &mut HeaderInjector(req.headers_mut().unwrap()))
+    });
+    let res = client.request(req.body(Body::from("Hallo!"))?).await?;
+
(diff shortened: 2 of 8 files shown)


examples/traceresponse/src/client.rs
```diff
@@ -0,0 +1,66 @@
...
+        .with_trace_config(trace::config().with_sampler(Sampler::AlwaysOn))
+        .install_simple()
+}
+
+#[tokio::main]
+async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    let _tracer = init_tracer();
+
+    let client = Client::new();
+    let span = global::tracer("example/client").start("say hello");
```
cijothomas, line 32: make the span kind = Client.

examples/traceresponse/src/server.rs
```diff
@@ -0,0 +1,58 @@
...
+        propagator.extract(&HeaderExtractor(req.headers()))
+    });
+    let _cx_guard = parent_cx.attach();
+    let mut span = global::tracer("example/server").start("hello");
+    span.add_event("handling this...", Vec::new());
+
+    let mut res = Response::new("Hello, World!".into());
+
+    let response_propagator: &dyn TextMapPropagator = &TraceContextResponsePropagator::new();
+    response_propagator.inject(&mut HeaderInjector(res.headers_mut()));
```
cijothomas, line 29: (Rust newbie comment) - this is supposed to return the traceid-spanid-traceflags, where spanid will be the spanid of the server span created here, right? I do not see this setting the right spanid here...

examples/traceresponse/src/server.rs
```diff
@@ -0,0 +1,58 @@
...
+use opentelemetry_contrib::trace::propagator::trace_context_response::TraceContextResponsePropagator;
+use opentelemetry_http::{HeaderExtractor, HeaderInjector};
+use std::{convert::Infallible, net::SocketAddr};
+
+async fn handle(req: Request<Body>) -> Result<Response<Body>, Infallible> {
+    let parent_cx = global::get_text_map_propagator(|propagator| {
+        propagator.extract(&HeaderExtractor(req.headers()))
+    });
+    let _cx_guard = parent_cx.attach();
+    let mut span = global::tracer("example/server").start("hello");
```
cijothomas, line 23: make this a server span

examples/traceresponse/src/server.rs
```diff
@@ -0,0 +1,58 @@
...
+    response_propagator.inject(&mut HeaderInjector(res.headers_mut()));
+
+    Ok(res)
+}
+
+fn init_tracer() -> impl Tracer {
+    global::set_text_map_propagator(TraceContextPropagator::new());
+
+    // Install stdout exporter pipeline to be able to retrieve the collected spans.
+    // For the demonstration, use `Sampler::AlwaysOn` sampler to sample all traces. In a production
```
cijothomas, line 38: nit: do we need to explicitly set AlwaysOn here? The default would be parent_based(ALwaysOn), which should be good enough for this example.

wperron:
@cijothomas I'll be honest, for the example, I simply copied over what's in `examples/http` and added bits to inject/extract the `traceresponse`

wperron:
I'm also not sure what's up with the CI/coverage job, it works fine locally 🤔

examples/traceresponse/src/server.rs
wperron: Doh. You're right 🤦 that's on me for going too fast, will fix 👍

examples/traceresponse/src/server.rs
wperron: IMO this is fine, it clearly illustrates that `AlwaysOn` is being used for demonstration purposes only, it ensures it always gets logged to stdout no matter what happens in the `client` example, for a demo this makes sense to me.

examples/traceresponse/README.md
```diff
@@ -0,0 +1,29 @@
+# HTTP Example
+
+This is a simple example using [hyper] that demonstrates tracing http request
+from client to server, and from the server back to the client using the
+[W3C Trace Context Response] header. The example shows key aspects of tracing
+such as:
+
+- Root Span (on Client)
+- Child Span (on Client)
```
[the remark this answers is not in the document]
wperron: Looks like there's only one span, maybe at some point the http client example had two? In any case, I've removed this line.

cijothomas:
> @cijothomas I'll be honest, for the example, I simply copied over what's in `examples/http` and added bits to inject/extract the `traceresponse`

np! I will take a look at the existing examples and propose some improvements. (I am totally new to Rust and this Repo :))

examples/traceresponse/src/server.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    let tracer = global::tracer("example/client");
+    let span = tracer
+        .span_builder("say hello")
+        .with_kind(SpanKind::Server)
+        .start(&tracer);
+
+    let cx = Context::current_with_span(span);
+
+    cx.span().add_event("handling this...", Vec::new());
```
cijothomas, line 33:
```suggestion
    span.add_event("handling this...", Vec::new());
```

wperron: This change would also require importing `opentelemetry::trace::Span` which would make `span` a `BoxedSpan` as opposed to a `SpanRef`, which also means making `span` mutable because `BoxedSpan::add_event` takes a mutable reference, and moving the definition of `cx` further down because `Context::current_with_span` moves the span, and `BoxedSpan` doesn't implement `Clone`.

I don't mind doing this change, but is it really necessary, when the purpose here is to illustrate how `traceresponse` works?

cijothomas: No need of changing. (I am Rust newbie, so my comments could be wrong!)

jtescher (approved):

PR #1002, opened 2023-03-21
fix: Fix the array encoding length of datadog version 05 exporter

glynnmccabe:
Fixes the array encoding length used for datadog version 05 api exporting. This is causing https://github.com/open-telemetry/opentelemetry-rust/issues/970


djc:
I feel this is like #903 in that we really should be using some derive-based encoding of types to send Datadog messages, instead of the raw MessagePack encoding that we're doing here.

glynnmccabe:
In this case it seems that the span array is always 12 elements long regardless of other improvements we might want, or am I missing something?

TommyCpp:
Another issue is we don't have access to Datadog backend to verify any fix so it's tricky

hdost:
> Running the datadog agent locally the current version will not accept traces with a 400 error and will accept them with the change.
> 
> There is also a test agent which also will not accept traces with a value that is not 12 - https://github.com/DataDog/dd-apm-test-agent/blob/cc71f96050a79bbe114b9e82005d92a7a8ca1e4f/ddapm_test_agent/trace.py#L304

I've tried running the example against the test agent and I get failures even with your patch.
What are you using to test?

glynnmccabe:
> > Running the datadog agent locally the current version will not accept traces with a 400 error and will accept them with the change.
> > There is also a test agent which also will not accept traces with a value that is not 12 - https://github.com/DataDog/dd-apm-test-agent/blob/cc71f96050a79bbe114b9e82005d92a7a8ca1e4f/ddapm_test_agent/trace.py#L304
> 
> I've tried running the example against the test agent and I get failures even with your patch. What are you using to test?

Datadog agent 7.43.1, although I have used the test agent previously.

Are the failures you see related to headers? The test agent expects some headers that this exporter doesn't send, but don't seem to be required by the actual agent. If so turning off the invariant checks listed here would fix that. https://github.com/DataDog/dd-apm-test-agent#trace-invariant-checks

TommyCpp (approved):
Looks good. Thanks for the fix!

PR #1006, opened 2023-03-26
chore: use tonic and prost patch releases

realtimetodie:
This follows the Cargo.toml package definition of the tonic crate that loosely couples patch releases to the prost crate, allowing users to globally pin a specific patch release to their needs if required.


realtimetodie:
@hdost Other crates might not follow the semantic versioning concept of major, minor and patch releases, but this is different for the tonic and prost crates. As I mentioned before, this change is of cosmetic nature and is safe.

djc:
Unless you audited whether we depend on any new features added between tonic 0.8.0 and 0.8.3, this change is actually not safe. Given that it's cosmetic, I think this PR is not worth having. Generally the patch version in dependency requirements should only ever increase.

(Please note that there's no pinning involved here, which Cargo only does when an explicit `=` operator is used.)

PR #1016, opened 2023-04-02
feat: upgrade to tonic 0.9

SimenB:
See https://github.com/hyperium/tonic/pull/1337

I just updated the versions and ran `cargo test`, committing the changed generated code as requested.


djc (approved):

djc:
LGTM, thanks! I wonder if we can release this as an API-compatible release in some of the crates at least, since we just did a release...

purkhusid:
Would love to see this released soon

PR #1017, opened 2023-04-02
feat(metrics): add in memory metrics exporter

### LINKED ISSUE #961: In-Memory Metrics Exporter
Hey all,

Didn't see an open issue for this so I figured I'd add one. It would be great to have an in-memory exporter per the spec [here](https://opentelemetry.io/docs/reference/specification/metrics/sdk_exporters/in-memory/) to make testing metric reporting a bit easier.

I might be able to make a contribution if this is an area where one would be welcome.

Thanks!

TommyCpp:
close #961 

- Add in memory metrics exporter, which can be used to inspect the `ResourceMetrics` exported using `get_finished_metrics`
- Implement `Clone` for all `DataPoint` with basic types


jtescher (approved):
This makes sense, think we should add this behind the testing flag as it's used for mocks? would keep the public API and compilation size down for non-testing cases.

hdost (commented):
Yes I like this, and maybe this could go into `src/metrics/testing` so that it's entirely self contained?

Or maybe we could (might be overdoing it) into it's own crate? 
perhaps a `opentelemetry-sdk-testing` crate which has all things related to testing and then we could have feature flags for tracing, metrics, logs. that way people could isolate to needing it as a dev-dependency.

TommyCpp:
Yeah `testing` seems like the right place. I will revise this

Issue #1015, opened 2023-03-31
`MeterProvider::meter` should not use `&'static str`


tqwewe:
It seems like the `MeterProvider::meter` method takes `&'static str`'s for name (as well as version and schema_url).
<https://docs.rs/opentelemetry/latest/opentelemetry/metrics/trait.MeterProvider.html>

In my case, the string I want to use is not `'static`, and this means I cannot use it to create meters.

I see that the instrument builder methods take `impl Into<String>` for the names. Maybe the meter functions could take the same, or `Cow<'static, str>` directly (since it seems like thats whats stored in the [`InstrumentationLibrary`](https://docs.rs/opentelemetry/latest/opentelemetry/struct.InstrumentationLibrary.html#structfield.name))

djc:
`Cow<'static, str>` probably makes sense, would you be able to send a PR?

PR #1018, opened 2023-04-03
feat: use `Cow<'static, str>` instead of `&'static str`

tqwewe:
Closes https://github.com/open-telemetry/opentelemetry-rust/issues/1015

Uses `Cow<'static, str>` instead of `&'static str` for some functions.

I initially tried `impl Into<Cow<'static, str>>` for the types, but it seemed to complicate things when used with the `Option` for version and schema_url.


opentelemetry-api/src/global/metrics.rs
```diff
@@ -26,9 +29,9 @@ impl fmt::Debug for GlobalMeterProvider {
 impl MeterProvider for GlobalMeterProvider {
     fn versioned_meter(
         &self,
-        name: &'static str,
-        version: Option<&'static str>,
-        schema_url: Option<&'static str>,
+        name: Cow<'static, str>,
```
lzchen, line 32: What was the reason for not using `impl Into<Cow<'static, str>>` again?

opentelemetry-api/src/global/metrics.rs
```diff
@@ -72,8 +75,8 @@ pub fn meter_provider() -> GlobalMeterProvider {
 /// If the name is an empty string, the provider will use a default name.
 ///
 /// This is a more convenient way of expressing `global::meter_provider().meter(name, None, None)`.
-pub fn meter(name: &'static str) -> Meter {
-    meter_provider().versioned_meter(name, None, None)
+pub fn meter(name: impl Into<Cow<'static, str>>) -> Meter {
+    meter_provider().versioned_meter(name.into(), None, None)
```
lzchen, line 79: Nit: Can we use ` meter_provider().meter(name.into())` to avoid the extra `None`s?

opentelemetry-api/src/global/trace.rs
```diff
@@ -349,13 +349,13 @@ impl trace::TracerProvider for GlobalTracerProvider {
     /// Create a versioned tracer using the global provider.
     fn versioned_tracer(
         &self,
-        name: impl Into<Cow<'static, str>>,
```
lzchen, line 352: Why remove this?

opentelemetry-api/src/trace/noop.rs
```diff
@@ -31,9 +31,9 @@ impl trace::TracerProvider for NoopTracerProvider {
     /// Returns a new `NoopTracer` instance.
     fn versioned_tracer(
         &self,
-        _name: impl Into<Cow<'static, str>>,
-        _version: Option<&'static str>,
-        _schema_url: Option<&'static str>,
+        _name: Cow<'static, str>,
+        _version: Option<Cow<'static, str>>,
```
lzchen, line 35: Curious as to why we are removing the `&` for `version` and `schema_url`? Are we allowing accepting of both `String` and `&str` for these fields?

opentelemetry-api/src/trace/noop.rs
```diff
@@ -31,9 +31,9 @@ impl trace::TracerProvider for NoopTracerProvider {
     /// Returns a new `NoopTracer` instance.
     fn versioned_tracer(
         &self,
-        _name: impl Into<Cow<'static, str>>,
-        _version: Option<&'static str>,
-        _schema_url: Option<&'static str>,
+        _name: Cow<'static, str>,
```
lzchen, line 34: What are the implications of using `impl Into<Cow<'static, str>>` to `Cow<'static, str>`?

opentelemetry-api/src/global/metrics.rs
tqwewe: I can use impl Into here, but not for the Option types unfortunately, as you cannot convert &str into `Option<Cow<...>>`.

So the best i can do is just the first parameter for name. Would you like me to make this change?

opentelemetry-api/src/trace/noop.rs
tqwewe: Yes previously it only allowed for static strings, but here it allows for `Cow<'static, str>`, which means it can be an owned string if the version is not static (eg might come from user input, io, etc).

hdost (approved):

Issue #815, opened 2022-06-16
Add support for instrumentation scope attributes


jtescher:
See the spec change for details https://github.com/open-telemetry/opentelemetry-specification/pull/2579

wperron:
Wouldn't mind taking a shot at this, is there an SDK that already implements this? Just took a quick look at the Go one and it doesn't look like it's been implemented there.

lzchen:
Related: https://github.com/open-telemetry/opentelemetry-rust/issues/883

PR #1021, opened 2023-04-09
Add support for instrumentation scope attributes

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/815

Note that according to the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) the uniqueness of an `InstrumentationScope` does not include the attributes.

I am new to Rust and OT-Rust so am ramping up. Please advise if `Option<Vec<KeyValue>>` is not the correct typing and there's some Rust-specific nuances related to memory management that I have overlooked :)


opentelemetry-api/src/common.rs
```diff
@@ -451,18 +451,40 @@ pub struct InstrumentationLibrary {
...
+        self.name.hash(state);
+        self.version.hash(state);
+        self.schema_url.hash(state);
+    }
 }
 
 impl InstrumentationLibrary {
     /// Create an new instrumentation library.
-    pub fn new<T>(name: T, version: Option<T>, schema_url: Option<T>) -> InstrumentationLibrary
+    pub fn new<T>(name: T, version: Option<T>, schema_url: Option<T>, attributes: Option<Vec<KeyValue>>) -> InstrumentationLibrary
```
TommyCpp, line 479: With the idea to move tracing to 1.0 in mind, I am a little worry how we handle new parameters here. Since spec doesn't consider adding new parameters a breaking change we should probably make it not a breaking change in our API.

jtescher: Yeah that's a good point. Also with potential other new additions in the future I wonder if a builder makes the most sense where libraries are generally specified. e.g. 

```rust
let tracer = TracerProvider::tracer_builder("tracer_name")
    .with_version("1.0.0")
    .with_attributes([kv])
    .build();
```

jtescher: Or something along those lines to prevent future API changes.

TommyCpp: `Option<Vec<KeyValue>>` works here but may not be the most ergonomic. Maybe `Option<V> where V: Into<Vec<KeyValue>>` so that user can pass a `Vec`, a `&[KeyValue]` or other struct implemented `Into<Vec<T>>` trait

lzchen: @TommyCpp 

I changed the function signature for `versioned_tracer` to look like this:

```rust
fn versioned_tracer(
        &self,
        name: impl Into<Cow<'static, str>>,
        version: Option<&'static str>,
        schema_url: Option<&'static str>,
        attributes: Option<impl Into<Vec<opentelemetry_api::KeyValue>>>,
    )
```
to try to allow any struct that implements `Into<Vec<T>>`. This forces me to pass in explicit typing whenever I use `versioned_tracer()` like below:

```rust
fn tracer(&self, name: impl Into<Cow<'static, str>>) -> Self::Tracer {
        self.versioned_tracer(name, None, None, None::<Vec<KeyValue>>)
    }
```
Is this the correct typing or is it unavoidable to always have to specify type information, even if the value is `None`?

TommyCpp: > Just to clarify, did you mean it would be desirable to have our api that uses InstrumentationLibrary (versioned_tracer()) accept different types of input for the attributes field

Yes that's my intention. It's similar to other parameters. For `Option<T>` struct in Rust has to know the actual type of `T` during compile time. 

I think at his point we can just use `Option<Vec<KeyValue>>` for now.  A better solution will be using builder pattern and get rid of the `Option`. Then we can ask user just provide any type that can convert to `Vec<KeyValue>`

lzchen:
@TommyCpp 

How does codecov work? Does it run on documentation testing?

examples/basic-tracer/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "basic-tracer"
+version = "0.1.0"
+edition = "2021"
```
cijothomas, line 4:
```suggestion
edition = "2023"
```
 🤣

examples/basic-tracer/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+use opentelemetry::{trace::Tracer, KeyValue};
+
+fn init_tracer() -> impl Tracer {
+    let v = vec![KeyValue::new("key", "value")];
+    stdout::new_pipeline()
+        .with_trace_config(trace::config().with_sampler(Sampler::AlwaysOn))
+        .install_with_tracer_attributes(v)
+}
+
+#[tokio::main]
```
cijothomas, line 12: nit: if this is truly intending to show a most basic tracing example, we don't need async, tokio here, right?

opentelemetry-api/src/common.rs
cijothomas: Is it possible to create "overloads" of `versioned_tracer`?

fn versioned_tracer(..)
{
versioned_tracer(..,None::<Vec<KeyValue>>)
}

fn versioned_tracer(.., attributes: Option<Vec<opentelemetry_api::KeyValue>>)

cijothomas: Just learned Rust has no overloading, so ignore my above comment.

examples/basic-tracer/Cargo.toml
lzchen: @cijothomas 
https://doc.rust-lang.org/edition-guide/rust-2021/index.html

examples/basic-tracer/src/main.rs
lzchen: Removed.

TommyCpp (approved):
LGTM. Thanks 👍

examples/basic-tracer/src/main.rs
```diff
@@ -0,0 +1,17 @@
+use opentelemetry::sdk::export::trace::stdout;
+use opentelemetry::sdk::trace::{self, Sampler};
+use opentelemetry::{trace::Tracer, KeyValue};
+
+fn init_tracer() -> impl Tracer {
+    let v = vec![KeyValue::new("key", "value")];
+    stdout::new_pipeline()
+        .with_trace_config(trace::config().with_sampler(Sampler::AlwaysOn))
+        .install_with_tracer_attributes(v)
```
cijothomas, line 9: This would conflict with https://github.com/open-telemetry/opentelemetry-rust/pull/1027/files. 
Need to merge one first and tweak the other accordingly.

lzchen: I can remove the basic example for now.

cijothomas (approved):
LGTM.
Could you add an entry to changelog file as well?
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-api/CHANGELOG.md

lzchen:
@cijothomas 

I haven't seen any other PRs add entries to the CHANGELOG.

@TommyCpp  is there a practice for this? Do we populate the CHANGELOG all at once during release or per relevant PR?

cijothomas:
> @cijothomas
> 
> I haven't seen any other PRs add entries to the CHANGELOG.
> 
> @TommyCpp is there a practice for this? Do we populate the CHANGELOG all at once during release or per relevant PR?

Irrespective of what was done in other PRs, I think all PRs affecting end users should include an entry in changelog as a good practice for any product.

TommyCpp:
> @TommyCpp is there a practice for this? Do we populate the CHANGELOG all at once during release or per relevant PR?

We do have a CHANGELOG for every create. But so far the practice has been one of us collecting all PRs before a release

TommyCpp:
Personally I'd love to require CHANGELOG entry for every PR. Makes release easier

cijothomas (approved):

cijothomas:
> Personally I'd love to require CHANGELOG entry for every PR. Makes release easier

I agree. Adding a PR template to encourage changelogs : https://github.com/open-telemetry/opentelemetry-rust/pull/1036

lzchen:
@TommyCpp 
Can this be merged?

TommyCpp (approved):

TommyCpp:
Will take a look on remaining PRs tonight

PR #1022, opened 2023-04-11
Modify Actix example to use OTLP ingestion for Jaeger

cijothomas:
Related to this : https://github.com/open-telemetry/opentelemetry-rust/issues/995
This PR modifies the Actix-Http example to use OTLP Exporter instead of Jaeger Exporter. The name of the project is still "actix-http".

There are quite a few examples with actix showing http/hyper/udp. It could be consolidated to a single one. If there is a desire to consolidate, I can propose a PR for that separately.

### DIFF
--- a/examples/actix-http/Cargo.toml
+++ b/examples/actix-http/Cargo.toml
@@ -6,8 +6,8 @@ publish = false
 
 [dependencies]
 opentelemetry = { path = "../../opentelemetry", features = ["rt-tokio"] }
-opentelemetry_sdk = { path = "../../opentelemetry-sdk"}
-opentelemetry-otlp = { path = "../../opentelemetry-otlp", features = ["tonic"] }
+opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["rt-tokio-current-thread"]}
+opentelemetry-otlp = { path = "../../opentelemetry-otlp" }
 actix-web = "4.1.0"
 actix-service = "2.0.0"
 env_logger = "0.9.0"
--- a/examples/actix-http/Cargo.toml
+++ b/examples/actix-http/Cargo.toml
@@ -6,7 +6,8 @@ publish = false
 
 [dependencies]
 opentelemetry = { path = "../../opentelemetry", features = ["rt-tokio"] }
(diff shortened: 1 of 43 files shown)


examples/actix-http/Cargo.toml
```diff
@@ -6,7 +6,8 @@ publish = false
 
 [dependencies]
 opentelemetry = { path = "../../opentelemetry", features = ["rt-tokio"] }
-opentelemetry-jaeger = { path = "../../opentelemetry-jaeger", features = ["reqwest_collector_client", "rt-tokio-current-thread"] }
+opentelemetry_sdk = { path = "../../opentelemetry-sdk"}
+opentelemetry-otlp = { path = "../../opentelemetry-otlp", features = ["tonic"] }
```
lzchen, line 10: Not sure if this is right but I don't seem to see `tonic` as a listed [feature](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/Cargo.toml#L61)?

examples/actix-http/src/main.rs
```diff
@@ -8,13 +8,21 @@ use opentelemetry::{
...
+        .tracing()
+        .with_exporter(
+            opentelemetry_otlp::new_exporter()
+                .tonic()
+                .with_endpoint("http://localhost:4317"),
+        )
+        .with_trace_config(sdktrace::config().with_resource(Resource::new(vec![
+            opentelemetry::KeyValue::new("service.name", "trace-http-demo"),
+        ])))
+        .install_batch(opentelemetry::runtime::Tokio)
```
TommyCpp, line 25: Probably should use `TokioCurrentThread` here. IIRC actix_web uses single thread tokio runtime by default

cijothomas: done

TommyCpp (approved):
LGTM. Thanks

TommyCpp:
Pushed a commit to fix the issue caused by #1018. I will open another PR to address the root case and revert the commit.

cijothomas:
> Pushed a commit to fix the issue caused by #1018. I will open another PR to address the root case and revert the commit.

Can you merge this please?

PR #1027, opened 2023-04-16
Create opentelemetry-stdout

### LINKED ISSUE #1024: Implement Console metrics exporter
Similar to [trace](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/export/trace/stdout.rs)

jtescher:
As the stdout/console exporter is generally used for testing and examples rather than production instrumentation, this change extracts it to a new crate. This has multiple benefits including:

* This crate can now include dependencies like `serde` and `serde_json` which would not be ideal as generally unused sdk dependencies.
* The representations of keys and values in the serialized formats can be changed without changing core SDK types.
* It moves code out of the SDK, making the SDK easier to maintain.
* Simplifies tests as they can use existing testing exporters

One notable change here is to prefer the standard `TracerProvider::builder().with_simple_exporter()` construction over a stdout-specific specialized pipeline. This is both because the stdout exporter is commonly used in addition to an existing exporter for debugging or other cases, which the pipeline version did not support, and because the pipeline did not meaningfully simplify setup.

Resolves #1024


hdost (approved):

README.md
```diff
@@ -39,11 +39,18 @@ observability tools.
...
+    trace::{Tracer, TracerProvider as _},
+};
 
 fn main() {
     // Create a new trace pipeline that prints to stdout
-    let tracer = stdout::new_pipeline().install_simple();
+    let provider = TracerProvider::builder()
+        .with_simple_exporter(opentelemetry_stdout::SpanExporter::default())
+        .build();
+    let tracer = provider.tracer("readme_example");
```
cijothomas, line 53: :100: this change looks good!

examples/http/src/client.rs
```diff
@@ -1,29 +1,29 @@
...
     trace::{TraceContextExt, Tracer},
     Context, KeyValue,
 };
 use opentelemetry_http::HeaderInjector;
+use opentelemetry_sdk::propagation::TraceContextPropagator;
+use opentelemetry_sdk::trace::TracerProvider;
+use opentelemetry_stdout::SpanExporter;
 
-fn init_tracer() -> impl Tracer {
+fn init_tracer() {
```
cijothomas, line 12: nit: this fn can be renamed to `init_tracer_provider`

examples/http/src/server.rs
```diff
@@ -22,20 +19,22 @@ async fn handle(req: Request<Body>) -> Result<Response<Body>, Infallible> {
...
+fn init_tracer() {
     global::set_text_map_propagator(TraceContextPropagator::new());
 
     // Install stdout exporter pipeline to be able to retrieve the collected spans.
     // For the demonstration, use `Sampler::AlwaysOn` sampler to sample all traces. In a production
     // application, use `Sampler::ParentBased` or `Sampler::TraceIdRatioBased` with a desired ratio.
-    stdout::new_pipeline()
-        .with_trace_config(trace::config().with_sampler(Sampler::AlwaysOn))
-        .install_simple()
+    let provider = TracerProvider::builder()
```
cijothomas, line 28: nits: The comment above about AlwaysOnSampler being used is no longer relevant.

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -0,0 +1,174 @@
...
+    },
+    Aggregation, InstrumentKind,
+};
+
+use crate::MetricsData;
+
+type Encoder = Box<dyn Fn(&mut dyn Write, MetricsData) -> Result<()> + Send + Sync>;
+
+/// An OpenTelemetry exporter that writes to stdout on export.
+pub struct MetricsExporter {
```
cijothomas, line 24: does the stdout exporter for Metrics have a way to make sure the periodic metric reader will be using the 10,000 msec interval instead of the 60,000 msec interval?

The spec recommends 10,000 for stdout exporter here:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/stdout.md

opentelemetry-stdout/src/common.rs
```diff
@@ -0,0 +1,150 @@
...
+        )
+    }
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct Resource {
+    attributes: Vec<KeyValue>,
+    #[serde(skip_serializing_if = "is_zero")]
+    dropped_attributes_count: u64,
```
cijothomas, line 24: Does resource need attribute_count limits?
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#exempt-entities

opentelemetry-stdout/src/trace/transform.rs
```diff
@@ -0,0 +1,235 @@
...
+
+#[derive(Debug, Clone, Copy)]
+enum SpanKind {
+    #[allow(dead_code)]
+    Unspecified = 0,
+    Internal = 1,
+    Server = 2,
+    Client = 3,
+    Producer = 4,
+    #[allow(dead_code)]
```
cijothomas, line 139: Consumer requires allow(dead_code) ?

cijothomas (approved):
LGTM. Left few comments.
Thanks for extracting stdout to own crate!

opentelemetry-stdout/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+    "development-tools::profiling",
+    "asynchronous",
+]
+keywords = ["opentelemetry", "tracing", "metrics"]
+license = "Apache-2.0"
+edition = "2021"
+rust-version = "1.60"
+
+[features]
+trace = ["opentelemetry_api/trace", "opentelemetry_sdk/trace", "futures-util"]
```
lzchen, line 19: Should we have all features enabled by default since this crate is pramarily used for testing/examples?

opentelemetry-stdout/examples/basic.rs
```diff
@@ -0,0 +1,52 @@
+//! run with `$ cargo run --example basic --all-features
```
lzchen, line 1: Nit: Can we put this example with the other examples in the top level folder?

lzchen (approved):

examples/http/src/server.rs
jtescher: Yeah can probably remove now

opentelemetry-stdout/examples/basic.rs
```diff
@@ -0,0 +1,52 @@
...
+    let tracer = tracer_provider.tracer("stdout-test");
+    let mut span = tracer.start("test_span");
+    span.set_attribute(KeyValue::new("test_key", "test_value"));
+    span.end();
+
+    let meter = meter_provider.meter("stdout-test");
+    let c = meter.u64_counter("test_events").init();
+    c.add(&cx, 1, &[KeyValue::new("test_key", "test_value")]);
+
+    meter_provider.shutdown()?;
```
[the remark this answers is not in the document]
jtescher: It shuts down on drop. I think we should consider adding a manual shutdown as well for parity though.

opentelemetry-stdout/src/metrics/exporter.rs
jtescher: Can consider this in a follow up as the reader config would have to be specified outside of the context of the exporter.

opentelemetry-stdout/src/common.rs
jtescher: It probably should get one in a follow up, the otlp spec includes this as well so we should probably record it.

opentelemetry-stdout/src/trace/transform.rs
jtescher: Oh good catch, this was from earlier but is not necessary.

opentelemetry-stdout/Cargo.toml
jtescher: Possibly, this matches the other exporters where you specify the signals you want to export, but I'm open to changing it in the future.

opentelemetry-stdout/examples/basic.rs
jtescher: I think we're trying to pare back the amount of examples we have in the top level, if anything we could remove this example.

PR #1030, opened 2023-04-17
Wait for exports on the simple span processor's ForceFlush

cschramm:
ForceFlush seems to have been left behind in #502. With those changes, the processing is not really synchronous anymore, i.e. OnEnd now only sends the span down the pipe to be processed in the separate thread as soon as possible.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush-1 says:

> In particular, if any SpanProcessor has any associated exporter, it SHOULD try to call the exporter's Export with all spans for which this was not already done and then invoke ForceFlush on it.

As the comment states, all spans previously got exported synchronounsly right away, so that no such spans existed, but now they might be anywhere between the channel and (the end of) the export call. Doing nothing in ForceFlush even violates the specification as...

> The built-in SpanProcessors MUST do so.

Awaiting all open tasks from the channel on ForceFlush fixes this.

Previous discussions regarding parts of the specification that this does not tackle in line with Shutdown:

> ForceFlush SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.

https://github.com/open-telemetry/opentelemetry-rust/pull/358#issuecomment-725449486

> ForceFlush SHOULD complete or abort within some timeout.

https://github.com/open-telemetry/opentelemetry-rust/pull/502/files#r603722431

This brings the simple processor a step closer to the batch processor with the obvious main difference of batches and the (not so obvious, also see https://github.com/open-telemetry/opentelemetry-rust/pull/502#issuecomment-809740071) difference that it works without a presumed async runtime.

### DIFF
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -92,45 +92,64 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
     fn shutdown(&mut self) -> TraceResult<()>;
 }
 
-/// A [`SpanProcessor`] that exports synchronously when spans are finished.
-///
-/// # Examples
-///
-/// Note that the simple processor exports synchronously every time a span is
-/// ended. If you find this limiting, consider the batch processor instead.
+/// A [`SpanProcessor`] that exports when spans are finished.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


jtescher (approved):
This looks good, thanks @cschramm. If you merge from main you should pick up changes to address the CI failures.

PR #1034, opened 2023-04-20
Add temporality benchmark tests for counter in metrics

lzchen:
Will leave histograms for another PR.

Also adds some sample runtimes run on local machine.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -4,7 +4,7 @@
 ### Added
 - Add instrument validation to `InstrumentBuilder` [#884](https://github.com/open-telemetry/opentelemetry-rust/pull/884).
 - Add `TelemetryResourceDetector` [#899](https://github.com/open-telemetry/opentelemetry-rust/pull/899).
-- Added `delta` temporality to metric counter benchmark tests [#899](https://github.com/open-telemetry/opentelemetry-rust/pull/899).
+- Added `delta` temporality to metric counter benchmark tests [#1034](https://github.com/open-telemetry/opentelemetry-rust/pull/1034).
 
 ### Changed
 - Update to `opentelemetry_api` v0.19.
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -4,7 +4,7 @@
 ### Added
 - Add instrument validation to `InstrumentBuilder` [#884](https://github.com/open-telemetry/opentelemetry-rust/pull/884).
 - Add `TelemetryResourceDetector` [#899](https://github.com/open-telemetry/opentelemetry-rust/pull/899).
-- Added `delta` temporality to metric counter benchmark tests [#1034](https://github.com/open-telemetry/opentelemetry-rust/pull/1034).
+- Add support for instrumentation scope attributes [#1021](https://github.com/open-telemetry/opentelemetry-rust/pull/1021).
 
 ### Changed
 - Update to `opentelemetry_api` v0.19.
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Added
 - Add instrument validation to `InstrumentBuilder` [#884](https://github.com/open-telemetry/opentelemetry-rust/pull/884).
 - Add `TelemetryResourceDetector` [#899](https://github.com/open-telemetry/opentelemetry-rust/pull/899).
+- Added `delta` temporality to metric counter benchmark tests [#899](https://github.com/open-telemetry/opentelemetry-rust/pull/899).
 
 ### Changed
 - Update to `opentelemetry_api` v0.19.
--- a/opentelemetry-sdk/benches/metric.rs
+++ b/opentelemetry-sdk/benches/metric.rs
@@ -9,7 +9,7 @@ use opentelemetry_sdk::{
     metrics::{
         data::{ResourceMetrics, Temporality},
         new_view,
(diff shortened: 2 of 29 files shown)


opentelemetry-sdk/src/metrics/reader.rs
```diff
@@ -102,6 +102,28 @@ impl TemporalitySelector for DefaultTemporalitySelector {
     }
 }
 
+/// Configure delta temporality for all [InstrumentKind]
+///
+/// [Temporality::Delta] will be used for all instrument kinds if this
+/// [TemporalitySelector] is used.
+#[derive(Clone, Default, Debug)]
+pub struct DeltaTemporalitySelector {
```
jtescher, line 110: If this is only needed for benchmarks it should probably stay in that file rather than being exposed in the public API, unless this is part of a spec.

lzchen: Moved to benchmark file.

opentelemetry-sdk/benches/metric.rs
```diff
@@ -67,13 +128,115 @@ fn bench_counter(view: Option<Box<dyn View>>) -> (Context, SharedReader, Counter
...
+            )
+        })
+    });
+    group.bench_function("AddThreeAttrDelta", |b| {
+        b.iter(|| {
+            cntr2.add(
+                &cx2,
+                1,
+                &[
+                    KeyValue::new("K", "V"),
```
cijothomas, line 163: I believe this will result in the same attributes being repeatedly used ? Suggest to use different values for the dimension. (A lot of cost in Metric SDKs would be sorting/finding the right metric stream to aggregate this to, so unless we use different dimensions, it'll be aggregating to produce a single metric point)

opentelemetry-sdk/benches/metric.rs
```diff
@@ -53,9 +53,70 @@ impl MetricReader for SharedReader {
...
+}
+
+// * Summary *
+
+// rustc 1.68.0 (2c8cc3432 2023-03-06)
+// cargo 1.68.0 (115f34552 2023-02-26), OS=Windows 11 Enterprise
+// Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz   2.59 GHz
+// 12 logical and 6 physical cores
+
+// Counter/AddNoAttrs      time:   [65.406 ns 65.535 ns 65.675 ns]
```
cijothomas, line 85: qn: what does the 3 set of numbers indicatr? Avg/Min/Max?

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Added
 - Add instrument validation to `InstrumentBuilder` [#884](https://github.com/open-telemetry/opentelemetry-rust/pull/884).
 - Add `TelemetryResourceDetector` [#899](https://github.com/open-telemetry/opentelemetry-rust/pull/899).
+- Added `delta` temporality to metric counter benchmark tests [#1034](https://github.com/open-telemetry/opentelemetry-rust/pull/1034).
```
cijothomas, line 7: my suggestion is to add changelog only for end user facing changes, and not required for test/benchmark additions.

opentelemetry-sdk/benches/metric.rs
lzchen: https://bheisler.github.io/criterion.rs/book/user_guide/command_line_output.html#time

opentelemetry-sdk/benches/metric.rs
```diff
@@ -53,9 +53,70 @@ impl MetricReader for SharedReader {
...
+// Counter/AddNoAttrsDelta time:   [65.553 ns 65.761 ns 65.981 ns]
+// Counter/AddOneAttr      time:   [341.55 ns 344.40 ns 347.58 ns]
+// Counter/AddOneAttrDelta time:   [340.11 ns 342.42 ns 344.89 ns]
+// Counter/AddThreeAttr    time:   [619.01 ns 624.16 ns 630.16 ns]
+// Counter/AddThreeAttrDelta
+//                         time:   [606.71 ns 611.45 ns 616.66 ns]
+// Counter/AddFiveAttr     time:   [3.7551 µs 3.7813 µs 3.8094 µs]
```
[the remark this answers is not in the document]
lzchen: Interesting, in Python, we never did anything specifically for perf involving number of attributes. Curious as to how you were able to achieve this specifically? Also I had to run the benchmark several times to be able to get increasing numbers like these (there were outliers every so often) so maybe we can run the test on the same machine used to run the dot net benchmarks to get a good baseline.

opentelemetry-sdk/benches/metric.rs
lzchen: Done.

opentelemetry-sdk/benches/metric.rs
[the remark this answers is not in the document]
cijothomas: From API standpoint: .NET has the above overloads for up to 3 attributes. 

counter.Add(long value, attribute)
counter.Add(long value, attribute1, attribute2)
counter.Add(long value, attribute1, attribute2, attribute3)

More than 3, we don't have more overloads :D, so one has to allocate an array and pass it. Allocating on heap is bad (.NET is garbage collected, so the goal is to avoid heap allocations as much as possible. 0 alloc is the goal for metrics!)
counter.Add(long value, attributes[]) --> passing an array (arrays are heap allocated).

From SDK, it keeps a per-thread Array for 1 attribute, 2 attributes, 3 attributes, .... 8 attributes. So up to 8 is alloc free (as it reuses the same array) from SDK. Anything more than 8, SDK has to allocate array, so perf drops.

jtescher (approved):

PR #1037, opened 2023-04-23
Improve tracer and meter creation ergonomics from #1018

jtescher:
This makes #1018 less of a breaking change by allowing existing uses that pass static strings to continue working without adding `.into` at each callsite.

It also updates the doc examples and exporters to correctly use schema URLs instead of passing `None`.


TommyCpp (approved):
Looks good! The `None::<Type>` is still annoying but I think we will eventually get rid of them when we have builder pattern in place

jtescher:
Yeah having default generic parameters on functions would be nice to avoid specifying the type of `None`, but it seems like this case should be pretty rare actually. If you are using the versioned tracer you should just include the current schema url like the other exporters do now.

PR #1045, opened 2023-04-25
Add basic stress testing framework + counter.add test

lzchen:
Discussed in the weekly Rust SIG, creates a stress testing framework under the folder `stress` to test throughput of functions in our API.

Measures the amount of times a certain function is called within a sliding window (2 seconds, can be configured in the future) to using the max number of cpus available.

Note this provides basic functionality and only prints to the console. 
Sample output:

```text
Throughput: 412998.00 requests/sec
Throughput: 404226.00 requests/sec
Throughput: 417000.00 requests/sec
Throughput: 403440.00 requests/sec
Throughput: 395148.00 requests/sec
Throughput: 404172.00 requests/sec
Throughput: 395412.00 requests/sec
Throughput: 400740.00 requests/sec
Throughput: 390132.00 requests/sec
Total requests processed: 7418976
```

### DIFF
--- a/stress/README.md
+++ b/stress/README.md
@@ -25,12 +25,12 @@ cargo run --release --bin metrics
 ```
 
 ```text
-Throughput: 4278494.00 requests/sec
-Throughput: 4157096.50 requests/sec
-Throughput: 4190185.50 requests/sec
-Throughput: 4185362.50 requests/sec
-Throughput: 4225514.00 requests/sec
-Throughput: 4220563.00 requests/sec
-Throughput: 4203673.50 requests/sec
-Throughput: 4230679.50 requests/sec
+Throughput: 4813019.00 requests/sec
+Throughput: 4805997.50 requests/sec
+Throughput: 4796934.50 requests/sec
+Throughput: 4817836.50 requests/sec
+Throughput: 4812457.00 requests/sec
+Throughput: 4773890.50 requests/sec
+Throughput: 4760047.00 requests/sec
+Throughput: 4851061.50 requests/sec
 ```
--- a/stress/README.md
+++ b/stress/README.md
@@ -24,6 +24,10 @@ e.g.
 cargo run --release --bin metrics
 ```
 
+Press (Ctrl + C) to quit the tests.
+
+Example output:
+
 ```text
 Throughput: 4813019.00 requests/sec
 Throughput: 4805997.50 requests/sec
--- a/stress/README.md
+++ b/stress/README.md
@@ -1,25 +1,36 @@
-# HTTP Example
+# OpenTelemetry Stress Tests
 
-This is a simple example using [hyper] that demonstrates tracing http request from client to server. The example
-shows key aspects of tracing such as:
+## Why would you need stress test
 
-- Root Span (on Client)
-- Child Span (on Client)
-- Child Span from a Remote Parent (on Server)
-- SpanContext Propagation (from Client to Server)
-- Span Events
-- Span Attributes
-
-[hyper]: https://hyper.rs/
+* It helps you to understand performance.
+* You can keep it running for days and nights to verify stability.
+* You can use it to generate lots of load to your backend system.
+* You can use it with other stress tools (e.g. a memory limiter) to verify how
+  your code reacts to certain resource constraints.
 
 ## Usage
 
-```shell
-# Run server
-$ cargo run --bin http-server
+Open a console, run the following command from the current folder:
 
-# In another tab, run client
-$ cargo run --bin http-client
+```sh
+cargo run --bin X
+```
 
-# The spans should be visible in stdout in the order that they were exported.
+where `X` is the specific stress test you would like to run.
+
+e.g.
+
+```sh
+cargo run --bin metrics
+```
+
+```text
+Throughput: 70878.00 requests/sec
+Throughput: 70272.00 requests/sec
+Throughput: 70326.00 requests/sec
+Throughput: 71724.00 requests/sec
+Throughput: 74382.00 requests/sec
+Throughput: 74856.00 requests/sec
+Throughput: 75666.00 requests/sec
+Total requests processed: 1035564
 ```
--- a/stress/README.md
+++ b/stress/README.md
@@ -25,12 +25,14 @@ cargo run --release --bin metrics
 ```
 
 ```text
-Throughput: 70878.00 requests/sec
-Throughput: 70272.00 requests/sec
-Throughput: 70326.00 requests/sec
-Throughput: 71724.00 requests/sec
-Throughput: 74382.00 requests/sec
-Throughput: 74856.00 requests/sec
-Throughput: 75666.00 requests/sec
-Total requests processed: 1035564
+Throughput: 412998.00 requests/sec
+Throughput: 404226.00 requests/sec
+Throughput: 417000.00 requests/sec
+Throughput: 403440.00 requests/sec
+Throughput: 395148.00 requests/sec
+Throughput: 404172.00 requests/sec
+Throughput: 395412.00 requests/sec
+Throughput: 400740.00 requests/sec
+Throughput: 390132.00 requests/sec
+Total requests processed: 7418976
 ```
--- a/stress/README.md
+++ b/stress/README.md
@@ -25,14 +25,12 @@ cargo run --release --bin metrics
 ```
 
 ```text
-Throughput: 412998.00 requests/sec
-Throughput: 404226.00 requests/sec
-Throughput: 417000.00 requests/sec
-Throughput: 403440.00 requests/sec
-Throughput: 395148.00 requests/sec
-Throughput: 404172.00 requests/sec
-Throughput: 395412.00 requests/sec
-Throughput: 400740.00 requests/sec
-Throughput: 390132.00 requests/sec
-Total requests processed: 7418976
+Throughput: 4278494.00 requests/sec
+Throughput: 4157096.50 requests/sec
+Throughput: 4190185.50 requests/sec
+Throughput: 4185362.50 requests/sec
+Throughput: 4225514.00 requests/sec
+Throughput: 4220563.00 requests/sec
+Throughput: 4203673.50 requests/sec
+Throughput: 4230679.50 requests/sec
 ```
--- a/stress/README.md
+++ b/stress/README.md
@@ -13,7 +13,7 @@
 Open a console, run the following command from the current folder:
 
 ```sh
-cargo run --bin X
+cargo run --release --bin X
 ```
 
 where `X` is the specific stress test you would like to run.
@@ -21,7 +21,7 @@ where `X` is the specific stress test you would like to run.
 e.g.
 
 ```sh
-cargo run --bin metrics
+cargo run --release --bin metrics
 ```
 
 ```text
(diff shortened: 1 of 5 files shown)


stress/README.md
```diff
@@ -0,0 +1,36 @@
...
+where `X` is the specific stress test you would like to run.
+
+e.g.
+
+```sh
+cargo run --release --bin metrics
+```
+
+```text
+Throughput: 70878.00 requests/sec
```
cijothomas, line 28: this is very low (we should be seeing millions/sec, as each call is only 100s of nanosecs from the cargo bench runs). I think the calculation or orchestration is not fully correct.

stress/src/throughput.rs
```diff
@@ -0,0 +1,48 @@
...
+        let elapsed = end_time.duration_since(start_time).as_secs();
+        if elapsed >= SLIDING_WINDOW_SIZE {
+            let throughput = count as f64 / elapsed as f64;
+            println!("Throughput: {:.2} requests/sec", throughput);
+            start_time = Instant::now();
+            count = 0;
+        }
+
+        let num_threads = num_cpus::get();
+        (0..num_threads).into_par_iter().for_each(|_| {
```
cijothomas, line 33: this should be one time operation, and not part of the loop. Each thread need to update some count variable indicating how many iterations it has ran, and there should be a main loop which checks the total count from all threads (and calculate the rate).

stress/README.md
lzchen: I forgot to change the README but the execution is actually somewhere around 400k/sec

cijothomas: I think it should be ~15 Million/sec, based on the ~65ns it takes to make s counter.Add(1) with no Attributes.

lzchen: I'm getting about 4 Million/Sec now with the new changes. I wonder if acquiring the lock to update the counts within the closure adds a bit of overhead.

stress/src/throughput.rs
lzchen: Fixed.

hdost (approved):

stress/README.md
lzchen: @jtescher @TommyCpp  @hdost 

In your experience, would obtaining a lock take that much time in `let mut counts = counts.lock().unwrap();`? If so, not sure how to implement this to allow multiple threads to modify the same vector. @cijothomas  if not possible, are we able to use the current readings as a baseline, even though it is "slower" than what we expect?

lalitb: Not sure, but will it help if we use message passing (std::sync::mpsc;), so each of the threads would write it's local count to message-queue every few ms (or lesser - maybe every increment), and main loop will keep receiving them in it's loop and adding. I assume read and writes to the queue are lockfree, so at-least it shouldn't have overhead for lock. Just a thought if this can be tried.

TommyCpp: One simple replacement is `AtomicU64` instead of a locked int

TommyCpp: > In your experience, would obtaining a lock take that much time in let mut counts = counts.lock().unwrap()

I believe that in this case, the counts variable will experience significant contention because multiple threads are attempting to lock it.

TommyCpp: Also second @lalitb we probably want to keep local variable local and throw them into queue. One choice can be https://crates.io/crates/crossbeam-queue

lzchen: @TommyCpp 
I'll give that a try.

lzchen: Attempt #1 with atomic u64:

```
Throughput: 4813019.00 requests/sec
Throughput: 4805997.50 requests/sec
Throughput: 4796934.50 requests/sec
Throughput: 4817836.50 requests/sec
Throughput: 4812457.00 requests/sec
Throughput: 4773890.50 requests/sec
Throughput: 4760047.00 requests/sec
Throughput: 4851061.50 requests/sec
```
Seems like incrementing an atomic u64 still has some overhead. Will continue experimenting...

hdost: Is it possible that we're running into a limitation in our implementation? And that we need to optimize the SDK counter? I noticed that in the hot path that we're doing Aggregate, so I'm wondering if that's not something we should consider moving. I'd have assumed that just like with our stress tester we'd want to have the actual aggregation on the alternative thread.

TommyCpp: For existing [implementation](https://github.com/open-telemetry/opentelemetry-rust/pull/1045/commits/97b179c532bdf8c790d1a775d6221e54da9470d2) I get
```
Throughput: 813044.50 requests/sec
Throughput: 824154.50 requests/sec
Throughput: 827773.50 requests/sec
Throughput: 806773.50 requests/sec
Throughput: 820415.50 requests/sec
Throughput: 813496.00 requests/sec
Throughput: 813064.50 requests/sec
Throughput: 807269.50 requests/sec
Throughput: 815698.50 requests/sec
Throughput: 803485.50 requests/sec
Throughput: 799224.00 requests/sec
Throughput: 799852.00 requests/sec
Throughput: 789417.50 requests/sec
Throughput: 785463.50 requests/sec
```

I think we can future optimize by having a vector of atomic number reference. i.e 
```rust
  let mut total_count = vec![];
    let mut total_count_refs = vec![];
    for _ in 0..num_cpus::get() {
        let count = Arc::new(AtomicU64::new(0));
        total_count.push(count.clone());
        total_count_refs.push(count);
    }

    rayon::spawn( move || {
        total_count.into_par_iter().for_each(|count| loop {
            func();
            count.fetch_add(1, Ordering::SeqCst);
            if STOP.load(Ordering::SeqCst) {
                break;
            }
        });
    });

    loop {
        let elapsed = end_time.duration_since(start_time).as_secs();
        if elapsed >= SLIDING_WINDOW_SIZE {
            let mut total_count_u64 = 0;
            for i in 0..num_cpus::get(){
                total_count_u64 += total_count_refs[i].load(Ordering::Relaxed);
            }
            let current_count = total_count_u64 - total_count_old;
            total_count_old = total_count_u64;
            let throughput = current_count as f64 / elapsed as f64;
            println!("Throughput: {:.2} requests/sec", throughput);
            start_time = Instant::now();
        }

        if STOP.load(Ordering::SeqCst) {
            break;
        }

        end_time = Instant::now();
    }
```
I can get slightly better result 
```
Throughput: 917330.50 requests/sec
Throughput: 915197.00 requests/sec
Throughput: 930210.50 requests/sec
Throughput: 919314.00 requests/sec
Throughput: 932616.50 requests/sec
Throughput: 913703.00 requests/sec
```

lalitb: > Can someone else try to run this? It could possibly be my cpu.

I see slightly better performance with existing implementation in this PR, of-course this is the release build:
```
$ cargo run --release
    Finished release [optimized] target(s) in 0.11s
     Running `/tmp/tbd2/opentelemetry-rust/target/release/metrics`
Throughput: 6086848.00 requests/sec
Throughput: 6212395.50 requests/sec
Throughput: 6218707.00 requests/sec
Throughput: 6244260.00 requests/sec
Throughput: 6220992.00 requests/sec
Throughput: 5648886.50 requests/sec
Throughput: 6265541.50 requests/sec
Throughput: 5894563.50 requests/sec
Throughput: 6267127.50 requests/sec
Throughput: 5647298.00 requests/sec
```

TommyCpp: > @TommyCpp I get similar results using your code as with the current implementation with atomicu64:
> 
> ```
> Throughput: 4542732.50 requests/sec
> Throughput: 4549933.50 requests/sec
> Throughput: 4427950.50 requests/sec
> Throughput: 4525449.50 requests/sec
> Throughput: 4521940.50 requests/sec
> Throughput: 4538587.00 requests/sec
> ```
> 
> Could we maybe merge as is and then iterate on the solution? @cijothomas wdyt

Yeah I think we can merge as it is and keep iterating

jtescher:
Running this locally looks like most of the time is spent waiting on the mutex contention when aggregating, we could consider more optimized methods for updating those sums. e.g. I can get more than 2x the current throughput by switching to a [SkipMap](https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/struct.SkipMap.html) for the sum's value map so there is likely quite a bit of headroom here.

Also the conversion into attribute sets will quickly become a bottleneck too once a test is added for that case, so we may want to consider an API change that allows the attribute sets to be precomputed.

stress/README.md
cijothomas: > @hdost We are comparing with the ~60ns it takes to run this function on average once as shown in the [microbenchmarking](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/metric.rs#L85) results. By this, (even with our current "slow" implementation) we _should_ be getting around ~15m calls/second. Is there some nuances in Rust itself or the way we are setting up the test that is adding some additional overhead that doesn't translate to simply 2 seconds / 60ns = 15m ? @cijothomas wdyt

My guess is the microbenchmarking (cargo bench) does not actually run multiple threads, but this stress test does, so microbenchmarking is not showing lock contention in its measurements, but stress test will include contention. So the stress may show much lower than the 15m/sec we anticipate based on cargo bench results.

stress/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+where `X` is the specific stress test you would like to run.
+
+e.g.
+
+```sh
+cargo run --release --bin metrics
+```
+
+```text
```
cijothomas, line 27: nit: say that this is an example output.

stress/README.md
```diff
@@ -0,0 +1,36 @@
...
+```sh
+cargo run --release --bin X
+```
+
+where `X` is the specific stress test you would like to run.
+
+e.g.
+
+```sh
+cargo run --release --bin metrics
```
cijothomas, line 24: also document how to quit (Ctrl + X) the tests

stress/src/throughput.rs
```diff
@@ -0,0 +1,51 @@
...
+    })
+    .expect("Error setting Ctrl-C handler");
+    let mut start_time = Instant::now();
+    let mut end_time = Instant::now();
+    let mut total_count_old: u64 = 0;
+    let total_count = Arc::new(AtomicU64::new(0));
+    let total_count_clone = Arc::clone(&total_count);
+
+    rayon::spawn( move || {
+        (0..num_cpus::get()).into_par_iter().for_each(|_| loop {
```
cijothomas, line 25: for a follow up PR - make the count configurable instead of defaulting to num_cpus. We can use it to measure how good we scale by trying different thread count from 1, 2, ...till num_cpus.

stress/src/throughput.rs
```diff
@@ -0,0 +1,51 @@
...
+            let throughput = current_count as f64 / elapsed as f64;
+            println!("Throughput: {:.2} requests/sec", throughput);
+            start_time = Instant::now();
+        }
+
+        if STOP.load(Ordering::SeqCst) {
+            break;
+        }
+
+        end_time = Instant::now();
```
cijothomas, line 49: this loop will eat up one whole thread reading now() continuously ?

cijothomas (approved):
looks good start. we can iterate this in follow ups.

stress/src/throughput.rs
lzchen: https://github.com/open-telemetry/opentelemetry-rust/issues/1055

stress/src/throughput.rs
lzchen: Yeah the external loop may eat up some cpu resources (most evident when running an old machine like mine). Since it probably isn't the major cause of slow throughput, we can probably iterate on this and improve in later prs.

lalitb (approved):

PR #1049, opened 2023-04-27
Add microbenching testings for histogram.add()

lzchen:
Seems like there's some notable slowdown whenever there are 5 attributes (but only 5). Please attempt to replicate with other machines.


cijothomas:
> Seems like there's some notable slowdown whenever there are 5 attributes (but only 5). Please attempt to replicate with other machines.

I don't see anything notable for 5 attributes! 
Histogram/Record3Attrs10bounds
                        time:   [252.09 ns 263.17 ns 276.32 ns]
Histogram/Record5Attrs10bounds
                        time:   [281.43 ns 291.04 ns 302.33 ns]
Histogram/Record7Attrs10bounds
                        time:   [308.35 ns 316.25 ns 325.99 ns]

lzchen:
@cijothomas 
Ahh interesting. Could be my machine then. I think we can use your test machine for running benchmarks that we will make analysis on in the future since mine is quite old.

cijothomas:
> @cijothomas Ahh interesting. Could be my machine then. I think we can use your test machine for running benchmarks that we will make analysis on in the future since mine is quite old.

We don't necessarily have to use same machine. If a PR is being sent which is improving the perf of something, the author can run the benchmarks before and after the PR from the same machine, so we can compare.

opentelemetry-sdk/benches/metric.rs
```diff
@@ -301,9 +342,60 @@ fn counters(c: &mut Criterion) {
     });
 }
 
+static MAX_BOUND: usize = 100000;
```
TommyCpp, line 345: nit: why not `constant`?

lzchen: Fixed.

TommyCpp (approved):

PR #1056, opened 2023-05-03
Append /v1/traces for http endpoints

### LINKED ISSUE #997: opentelemetry-otlp: '/v1/traces' is not appended to endpoint for http exporter
The HTTP exporter doesn't add `/v1/traces` to the 'endpoint' specified in the config:

https://github.com/open-telemetry/opentelemetry-rust/blob/592e77a8f323fe5c490de9eab458b7e40e9c437b/opentelemetry-otlp/src/span.rs#L404-L419

This seems like an oversight because this is done in other places:

https://github.com/open-telemetry/opentelemetry-rust/blob/592e77a8f323fe5c490de9eab458b7e40e9c437b/opentelemetry-otlp/src/span.rs#L319-L328

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/metric.rs#L235-L245

thorhs:
Fixes #997 

## Changes

We already append /v1/traces for grpc-tonic, start appending /v1/traces to http URLs, as well as using the
OTEL_EXPORTER_OTLP_TRACES_ENDPOINT if it is set, as specified by the standard.

To preserve backwards compatability we don't append /v1/traces to URLs passed to the new_http function if it already contains a path component.  Otherwise we would break existing code that is specifying the /v1/traces, or other specific path.


opentelemetry-otlp/src/span.rs
```diff
@@ -424,13 +424,38 @@ impl SpanExporter {
...
+        // If no path is supplied we append the standard path for traces.
+        // This means that non-standard location of the traces endpoint must
+        // be specified in full.
+        let url: Uri = match url.path() {
+            "" | "/" => {
+                let mut parts = url.into_parts();
+                parts.path_and_query = Some(http::uri::PathAndQuery::from_static("/v1/traces"));
+                parts.try_into().expect(
+                    "Should always be a valid URL, since we are constructing from a valid URL",
+                )
```
TommyCpp, line 454: Should we consider keep the query when motify the Uri? It's probably gonna be a noop for official OTEL collector but some users may have custom implementation that compatible with OTEL protocol

pksunkara:
@thorhs Are you planning to update this PR?

@TommyCpp What's remaining in this if I want to take over?

opentelemetry-otlp/src/span.rs
thorhs: Interesting point I hadn't thought about.  As I stated in the commit message I leave the URL alone if any path is specified for backwards compatibility with code that already has the full URL specified.  I assumed that was sufficient.

I don't have enough experience with OTEL in the wild to know if these are issues we need to address.  Maybe we should just always append /v1/traces to the path, and keep the query as-is?

I think the possibilities are:
* Keep it as is, if you need anything else than the default supplies, you must supply the full path to the trace endpoint.
* Append /v1/traces to all paths if /v1/traces is not a postfix of the path, and keep the query parts.

The second option might be an issue if a collector does not use the /v1/traces postfix.

thorhs:
> @thorhs Are you planning to update this PR?

I had honestly forgot about it, I am using this branch locally so it hasn't been an issue for it.  I can try to move it forward but am totally fine if you want to take over.

opentelemetry-otlp/src/span.rs
pksunkara: FWIW, both `tonic` traces implementation and metrics implementation just append the path as a string disregarding the query completely. IMHO, we should do the same here to keep the same pattern.

thorhs: Yes, I did notice that but feared someone might be relying on the current implementation, and starting to replace the path unconditionally might break existing code.  It would be great to get some input from the OTEL team on how to proceed.  I'm fine with both solutions as long as I don't need to supply the path part.

TommyCpp: Sorry for the late response. Since `url.path()` only output path information without any query information. I am concerned if users try to config something like `https://otel-collector.net?page=2` we will add a path for them. In this case I think we should follow the same pattern to leave it as it is.

thorhs: @TommyCpp - Just to clarify, should we set the path if neither path nor query are set? Or were you thinking of some other option?

hdost:
@thorhs There's been quite a few large changes recently. If you feel comfortable you could rebase/cherry pick onto main, or if you don't mind one of the maintainers can take a look at it.

pksunkara:
This can be closed since the underlying has been fixed in other PRs.

PR #1061, opened 2023-05-09
Move OrderMap module to root of otel-api crate

lalitb:
Fixes #

## Changes

OrderMap module is currently stored under `crate::opentelemetry-api::trace` hierarchy. Going ahead as part of #788, this data structure would also be used in logs bridge-api/sdk  to log attributes. As this is a generic data structure, and doesn't logically belong to Trace, moving it to root of opentelemetry-api crate. This will enable it to used in both trace and logs in much cleaner way.


TommyCpp (approved):
Looks good, thanks. Could you add a CHANGELOG?

>  this may be breaking change for existing trace users as they have to change the specified path of this module in their code

One way to mitigate it is to re-export it in traces module

lalitb:
> Looks good, thanks. Could you add a CHANGELOG?
> One way to mitigate it is to re-export it in traces module

Thanks, added changelog, and re-export in trace module.

cijothomas (approved):

Issue #1060, opened 2023-05-09
Docs for Metrics - add basic and advanced examples


cijothomas:
Opening an issue to track adding docs for Metrics.

Proposing to add 2 examples to begin with:
1. Basic instrumentation - This will use stdout exporter, and show example usage of all instrument kinds. More geared towards "learning the various OTel instruments".
2. Advanced configuration - This will use OTLP Exporter, and shows advanced configuration - Views (and its various use cases), changing Temporality.

The docs alone might not be sufficient, hence the proposal to have 2 examples. Happy to iterate based on feedback. 

Note: It was discussed in today's SIG meeting as well, in the context of filling the spec compliant matrix. The examples will demonstrate most common use cases, and not every line item from the spec-compliance matrix.

Related issue : #914

jtescher:
@cijothomas this considered resolved now?

cijothomas:
> @cijothomas this considered resolved now?

Not yet. Will need another follow up to https://github.com/open-telemetry/opentelemetry-rust/pull/1170 (it has desc. about the nest PRs.)

cijothomas:
One key thing missing today is:
Changing Temporality to Delta (as a lot of observability vendors accept delta only, but default is cumulative.)

Tagging as good-first-issue, to see if anyone wants to enhance the example to show the above.

cijothomas:
Need to address the issue about using Views with OTLP Exporter as well. Its probably involves code change, but a temporary docs change could help as well : https://github.com/open-telemetry/opentelemetry-rust/issues/1174#issuecomment-1660646295

mattbodd:
I'm interested in taking this on @cijothomas

Annosha:
@mattbodd are you working on updating the docs? If not I'd like to work on this issue. TIA @cijothomas

cijothomas:
@Annosha feel free to send short PRs with doc improvements. There are WIP PRs like https://github.com/open-telemetry/opentelemetry-rust/pull/2221 and https://github.com/open-telemetry/opentelemetry-rust/pull/2220 that affect some docs, so short, focused PRs are recommended.

Few things I suggest to make it super clear in docs (just an initial list from top of mind)
1. MeterProvider should be created once only.
2. Meters should be created once only
3. Instruments should be created once and re-used. It is okay to create instruments from multiple places (internally it'll be same instrument), but users should store the instrument and reuse.
[...]

PR #1062, opened 2023-05-11
Add basic example showing how to use various metric instruments

cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-rust/issues/1060

## Changes

Adds a example which showcases the usage of various instruments. The stdout exporter is used to display metrics.
There'll be a separate PR which shows common customizations, like changing temporality, using views, changing interval etc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ members = [
     "examples/grpc",
     "examples/http",
     "examples/hyper-prometheus",
+    "examples/metrics-basic",
     "examples/traceresponse",
     "examples/tracing-grpc",
     "examples/jaeger-remote-sampler",
--- a/examples/metrics-basic/Cargo.toml
+++ b/examples/metrics-basic/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "metrics-basic"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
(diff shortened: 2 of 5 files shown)


TommyCpp (approved):
Looks reaonable to me. I understand we have too many examples at this point but with the new metrics example I think this one will be really benificial for users. We can probably clean up some other examples to reduce maintaince burdan

cijothomas:
> Looks reaonable to me. I understand we have too many examples at this point but with the new metrics example I think this one will be really benificial for users. We can probably clean up some other examples to reduce maintaince burdan

Agree. I'll be able to help with some cleanups/consolidation of the examples.

cijothomas:
@TommyCpp Could you merge please?

gautamg795:
Does your example of recording data to a histogram here use the `ExponentialBucketHistogram` type under the hood, or is it using an explicitly bucketed histogram with some default value for the buckets?

cijothomas:
> Does your example of recording data to a histogram here use the `ExponentialBucketHistogram` type under the hood, or is it using an explicitly bucketed histogram with some default value for the buckets?

It follows the spec, which states that the default is ExplicitBucket aggregation, using the default buckets. 
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/reader.rs#L154

Issue #1065, opened 2023-05-17
Implement cardinality limits for metrics streams


lzchen:
From spec: https://github.com/open-telemetry/opentelemetry-specification/pull/2960

Implement ability to specify cardinality limits per metric stream to help protect against excessive data production from a single instrument.

cijothomas:
Can be closed with #1066

lzchen:
Can't be closed yet as the feature to configure the rate has not been implemented yet.

cijothomas:
Tracked via https://github.com/open-telemetry/opentelemetry-rust/issues/1951 closing this.

cijothomas:
Reopening this issue, and using to track enabling this feature.

cijothomas:
Take care of this: https://github.com/open-telemetry/opentelemetry-rust/issues/2878

PR #1066, opened 2023-05-19
Implement cardinality limits for metric streams

lzchen:
Part of https://github.com/open-telemetry/opentelemetry-rust/issues/1065

This PR implements a fixed cardinality limit for metric data points of the same stream to 2000. If an attempt to use an instrument to create a new metric data point for the same stream is made which would exceed the limit, the data point is added/recorded to the data point representing the overflow metric stream which is denoted by an `AttributeSet` containing only `{otel.metric.overflow: True}`.

Note there is a bit of duplication of code done on purpose to optimize for the most common scenarios, which is usually when an entry exists already within a hashmap.

### DIFF
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -30,6 +30,7 @@ url = { version = "2.2", optional = true }
 tokio = { version = "1.0", default-features = false, features = ["rt", "time"], optional = true }
 tokio-stream = { version = "0.1", optional = true }
 http = { version = "0.2", optional = true }
+lazy_static = "1.4"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -30,7 +30,6 @@ url = { version = "2.2", optional = true }
 tokio = { version = "1.0", default-features = false, features = ["rt", "time"], optional = true }
 tokio-stream = { version = "0.1", optional = true }
 http = { version = "0.2", optional = true }
-lazy_static = "1.4"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/opentelemetry-sdk/src/metrics/internal/aggregator.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregator.rs
@@ -2,6 +2,8 @@ use std::sync::Arc;
 
 use crate::{attributes::AttributeSet, metrics::data::Aggregation};
 
+const STREAM_CARDINALITY_LIMIT: u32 = 2000;
+
 /// Forms an aggregation from a collection of recorded measurements.
 pub(crate) trait Aggregator<T>: Send + Sync {
     /// Records the measurement, scoped by attr, and aggregates it into an aggregation.
@@ -14,6 +16,11 @@ pub(crate) trait Aggregator<T>: Send + Sync {
     /// Used when filtering aggregators
     fn as_precompute_aggregator(&self) -> Option<Arc<dyn PrecomputeAggregator<T>>> {
         None
+    }
+
+    /// Checks whether aggregator has hit cardinality limit for metric streams
+    fn check_stream_cardinality(&self, size: usize) -> bool {
+        size < STREAM_CARDINALITY_LIMIT as usize
     }
 }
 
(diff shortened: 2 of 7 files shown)


opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -26,10 +26,20 @@ impl<T: Number<T>> ValueMap<T> {
...
+            match values.entry(attrs) {
+                Entry::Occupied(mut occupied_entry) => {
+                    let count = occupied_entry.get_mut();
+                    *count += measurement;
+                }
+                Entry::Vacant(vacant_entry) => {
+                    if self.check_stream_cardinality(size) {
+                        vacant_entry.insert(measurement);
+                    } else {
+                        println!("Warning: Maximum metric streams exceeded. Entry not added.");
```
cijothomas, line 39: nit: This is not limiting the metric stream counts, but limiting the metric points within a given stream.

cijothomas: In actual PR, the plan is still to enter a pre-fixed AttributeSet (`otel.metric.overflow`:`true`) right?

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -26,10 +29,24 @@ impl<T: Number<T>> ValueMap<T> {
...
+                }
+                Entry::Vacant(vacant_entry) => {
+                    if self.check_stream_cardinality(size) {
+                        vacant_entry.insert(measurement);
+                    } else {
+                        values
+                            .entry(STREAM_OVERFLOW_ATTRIBUTE_SET.clone())
+                            .and_modify(|val| *val += measurement)
+                            .or_insert(measurement);
+                        println!("Warning: Maximum data points for metric stream exceeded. Entry added to overflow.");
```
cijothomas, line 46: feedback from SIG meeting: 
Use standard error handling mechanism, instead of println.

cijothomas: use the std mechanism now, @TommyCpp can explore how to enable it for Warning level internal logs.

lzchen: Done

opentelemetry-sdk/Cargo.toml
```diff
@@ -30,6 +30,7 @@ url = { version = "2.2", optional = true }
 tokio = { version = "1.0", default-features = false, features = ["rt", "time"], optional = true }
 tokio-stream = { version = "0.1", optional = true }
 http = { version = "0.2", optional = true }
+lazy_static = "1.4"
```
jtescher, line 33: nit: can we use [once_cell](https://docs.rs/once_cell/latest/once_cell/) over lazy_static? more of the standard now that it is being added to std.

opentelemetry-sdk/src/metrics/internal/aggregator.rs
```diff
@@ -15,6 +24,11 @@ pub(crate) trait Aggregator<T>: Send + Sync {
     fn as_precompute_aggregator(&self) -> Option<Arc<dyn PrecomputeAggregator<T>>> {
         None
     }
+
+    /// Checks whether aggregator has hit cardinality limit for metric streams
+    fn check_stream_cardinality(&self, size: usize) -> bool {
```
jtescher, line 29: nit: functions with `check` tend to suggest a `Result` return type, maybe have the name be something along the lines of `is_valid_stream_cardinality` or `is_under_cardinality_limit`, etc?

opentelemetry-sdk/src/metrics/internal/aggregator.rs
```diff
@@ -1,6 +1,19 @@
+use crate::{attributes::AttributeSet, metrics::data::Aggregation};
+use once_cell::sync::OnceCell;
+use opentelemetry_api::KeyValue;
 use std::sync::Arc;
 
-use crate::{attributes::AttributeSet, metrics::data::Aggregation};
+const STREAM_CARDINALITY_LIMIT: u32 = 2000;
+static STREAM_OVERFLOW_ATTRIBUTE_SET: OnceCell<AttributeSet> = OnceCell::new();
```
jtescher, line 7: Nit: could use lazy here to reduce boilerplate https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html

jtescher (approved):

PR #1067, opened 2023-05-19
Unify trace and logs runtime extensions traits.

vibhavp:
Merge the `TraceRuntime` and `LogRuntime` traits in the trace and logs module respectively into `MessageRuntime`, an extension trait to `Runtime`, generic over the type of Message/Signal type to be sent and received. All references to the older traits now become `MessageRuntime<trace::BatchMessage>` and `MessageRuntime<logs::BatchMessage>` respectively.


opentelemetry-sdk/src/runtime.rs
```diff
@@ -131,3 +132,105 @@ impl Runtime for AsyncStd {
...
+    ///
+    /// A failure can be due to either a closed receiver, or a depleted buffer.
+    fn try_send(&self, item: Self::Message) -> Result<(), TrySendError>;
+}
+
+#[cfg(any(feature = "rt-tokio", feature = "rt-tokio-current-thread"))]
+impl<T: Send> TrySend for tokio::sync::mpsc::Sender<T> {
+    type Message = T;
+
+    fn try_send(&self, item: T) -> Result<(), TrySendError> {
```
lalitb, line 177:
```suggestion
    fn try_send(&self, item: Message) -> Result<(), TrySendError> {
```

opentelemetry-sdk/src/runtime.rs
```diff
@@ -131,3 +132,105 @@ impl Runtime for AsyncStd {
...
+            tokio_stream::wrappers::ReceiverStream::new(receiver),
+        )
+    }
+}
+
+#[cfg(feature = "rt-async-std")]
+impl<T: Send> TrySend for async_std::channel::Sender<T> {
+    type Message = T;
+
+    fn try_send(&self, item: T) -> Result<(), TrySendError> {
```
lalitb, line 219:
```suggestion
    fn try_send(&self, item: Message) -> Result<(), TrySendError> {
```

opentelemetry-sdk/src/runtime.rs
vibhavp: Fixed, thanks.

opentelemetry-sdk/src/runtime.rs
vibhavp: Fixed, thanks.

opentelemetry-sdk/src/runtime.rs
```diff
@@ -131,3 +132,108 @@ impl Runtime for AsyncStd {
         Box::pin(async_std::task::sleep(duration))
     }
 }
+
+/// `MessageRuntime` is an extension to [`Runtime`]. Currently, it provides a
+/// channel that is used by the [log] and [span] batch processors.
+///
+/// [log]: crate::logs::BatchLogProcessor
+/// [span]: crate::trace::BatchSpanProcessor
+pub trait MessageRuntime<T: Debug + Send>: Runtime {
```
djc, line 141: I hadn't really noticed this before (and I note the name was equally confusing before), but I think the `*Runtime` name for this is actually pretty bad. How about naming it `RuntimeChannel`? Given the implementations for `tokio::sync::mpsc` and `async_std::channel` that seems like a much more descriptive name.

djc (approved):
This is a nice improvement!

lalitb (approved):

hdost (approved):

opentelemetry-sdk/src/runtime.rs
vibhavp: Agreed, `Message` doesn't communicate that much. That being said, `MessageRuntime` seems to be only used for the batch processor, so what about something like `BatchRuntime`?

However, if there are plans to use this trait for purposes other than batch processing, why not just rename it to `RuntimeExt`?

djc: These types aren't runtimes, they're channels. Calling it runtime is confusing at best.

Robo210: > However, if there are plans to use this trait for purposes other than batch processing, why not just rename it to RuntimeExt?

It would be nice if batch processors could have a way to spawn additional async work and use the configured runtime to drive it to completion as part of their export. I imagine this could look like a `BatchProcessorRuntimeExt`  (or perhaps just an updated `Runtime` trait itself).

```rust
pub trait Runtime: Clone + Send + Sync + 'static {
<snip>
    /// Spawn a new task or thread, which executes the given future.
    ///
    /// # Note
    ///
    /// This is mainly used to run batch span processing in the background. Note, that the function
    /// does not return a handle. OpenTelemetry will use a different way to wait for the future to
    /// finish when TracerProvider gets shutdown. At the moment this happens by blocking the
    /// current thread. This means runtime implementations need to make sure they can still execute
    /// the given future even if the main thread is blocked.
    fn spawn(&self, future: BoxFuture<'static, ()>);
```

Robo210: Making such a change is certainly out of scope of this PR. But maybe ideas for the future can help pick a good name right now.

jtescher (approved):

jtescher:
Looks good! just the one docs issue remaining

djc:
Thanks!

Issue #841, opened 2022-07-14
Proposal: Move Extra packages out of this repo into a opentelemetry-rust-contrib


hdost:
Following reading more on the IdGenerator I was also taking a look at some of the other language repos, and I wonder if there shouldn't be a `opentelemetry-rust-contrib` repo created to hold anything outside the "core".

Updated on 2023-11-02 for consistency with the repo. 

*Keep in Primary:*

- `opentelemetry`
- `opentelemetry-http`
- `opentelemetry-jaeger`
- `opentelemetry-otlp`
- `opentelemetry-prometheus`
- `opentelemetry-proto`
- `opentelemetry-sdk`
- `opentelemetry-semantic-conventions`
- `opentelemetry-stdout`
- `opentelemetry-zipkin`
- `opentelemetry-appender-log` - For `log` which is a crate maintained by Rust Lang team
- `opentelemetry-appender-tracing` - Keep for now, might be deleted or merged into `tracing-opentelemetry` later

*Move to Contrib:*

- `opentelemetry-aws`
- `opentelemetry-contrib`
- `opentelemetry-datadog`
- `opentelemetry-dynatrace`
- `opentelemetry-stackdriver`
- `opentelemetry-user-events-logs`
- `opentelemetry-user-events-metrics`
- `opentelemetry-zpages`

> Additional `IdGenerator` implementing vendor-specific protocols such as AWS X-Ray trace id generator MUST NOT be maintained or distributed as part of the Core OpenTelemetry repositories.

Source: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#id-generators

Java: https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/aws-xray

TommyCpp:
I think that makes sense. The discussion around this before was worried about it will be harder to maintain two separate repos.

We also publish a [contrib crate](https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-contrib) that basically includes anything is not in standard but not complicate enough for a separate crate.

djc:
It's not obvious to me why Jaeger, Prometheus and Zipkin are in core and the others aren't?

TommyCpp:
> It's not obvious to me why Jaeger, Prometheus and Zipkin are in core and the others aren't?

Those exporters are defined in spec as something SDK should support.

[...]

PR #1071, opened 2023-05-22
Add LogAppender for log crate

cijothomas:
Initial Version of Log Appender for the Log crate. 
~TODO: The appender is now dependent on the SDK, which should be fixed to only depend on the API~ Done with help from @tommycpp
~TODO: Decide on naming of the crate.~ The name looks good based on feedback from the folks attending SIG meeting on 5/23.

Does not support the structured log as it requires unstable features. Structured logs can be added as a follow up PR. If the overall structure is good, I'd send PR to add similar appenders for other logging libraries, specifically slog, and tracing.

Update: 
Based on discussion in this PR, this create is temporarily hosted in the main opentelemetry rust repo. This would be moved to a opentelemetry owned contrib repo (once is it created), or, even to https://github.com/rust-lang/log if the maintainers agree.


opentelemetry-appender-log/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "opentelemetry-appender-log"
```
cijothomas, line 2: From SIG meeting:
Name looks good.
Okay to accept more appenders.

djc:
It's unclear to me why there's a need for a new crate -- what will this crate contain, and why is it out of place in any of the existing crates?

vibhavp:
I had created a `log-rs` adaptor for otel logs, here: https://github.com/vibhavp/log-rs-opentelemetry. The code is similar to this crate, with a few improvements. Would it be possible to merge the two?

That being said, should this be maintained under `opentelemetry`? For instance, the `tracing` adaptor for opentelemetry-rust is at https://github.com/tokio-rs/tracing-opentelemetry, and it would be better if we followed a similar pattern, making this its own repository.

Additionally, we should offer additional flexibility to include the trace context in records created by this adaptor. `tracing` is the de-facto tracing library for Rust, which uses 64-bit span IDs which are then duct-taped to OpenTelemetry's 128-bit IDs through [`OpenTelemetrySpanExt`](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/trait.OpenTelemetrySpanExt.html#). As a result, opentelemetry's `Context` cannot be used to extract the current trace context, which is what the Logger uses.

cijothomas:
> It's unclear to me why there's a need for a new crate -- what will this crate contain, and why is it out of place in any of the existing crates?

> what will this crate contain

This is the Logging Bridge (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#log-appender--bridge) for the Log create. This is a *new* crate, because this was made possible only recently with this PR : https://github.com/open-telemetry/opentelemetry-rust/pull/788, which provided the Bridge API and the SDK side, and this is the 1st Appender/Bridge leveraging it. (I have mentioned the interest to make 2 more bridges, for slog and tracing as well in yesterday's SIG meeting.)

> why is it out of place in any of the existing crates?

I cannot see any existing components where this can be put. Do you have any particular create in mind? 
I would have preferred to make this in the contrib repo, but was told this repo is currently okay, and could be moved out once contrib repo is created (https://github.com/open-telemetry/opentelemetry-rust/issues/841#issuecomment-1504560778).

examples/logs-basic/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+            )])))
+            .with_simple_exporter(exporter).build();
+
+    // Setup Log Appender for the log crate.
+    let otel_log_appender = OpenTelemetryLogBridge::new(Level::Info, &logger_provider);
+    log::set_boxed_logger(Box::new(otel_log_appender)).unwrap();
+
+    // Emit logs using macros from the log crate.
+    // These logs gets piped through OpenTelemetry bridge and gets exported to stdout.
+    error!(target: "cijo-warn", "hello from {}. My price is {}", "apple", 2.99);
```
Robo210, line 23: Probably shouldn't put your name into the samples.

cijothomas: Totally! Will remove before I remove the "Draft" status!

examples/logs-basic/src/main.rs
```diff
@@ -0,0 +1,27 @@
...
+        .with_simple_exporter(exporter)
+        .build();
+
+    // Setup Log Appender for the log crate.
+    let otel_log_appender = OpenTelemetryLogBridge::new(Level::Info, &logger_provider);
+    log::set_boxed_logger(Box::new(otel_log_appender)).unwrap();
+
+    // Emit logs using macros from the log crate.
+    // These logs gets piped through OpenTelemetry bridge and gets exported to stdout.
+    error!(target: "my-target", "hello from {}. My price is {}", "apple", 2.99);
```
cijothomas, line 26: I can confirm that the correlation with traces work when using the OpenTelemetry Tracing API. If there is interest, I can modify this example itself to show traces as well, so users can see that logs are correlated with traces automatically.

(Need to test with tracing-opentelemetry, as a follow up next :))

opentelemetry-api/src/logs/logger.rs
```diff
@@ -17,7 +17,7 @@ pub trait Logger {
 /// Interfaces that can create [`Logger`] instances.
 pub trait LoggerProvider {
     /// The [`Logger`] type that this provider will return.
-    type Logger: Logger;
```
TommyCpp, line 20: Hmm why remove this requirement here?

examples/logs-basic/src/main.rs
```diff
@@ -0,0 +1,27 @@
+use log::{error, Level};
```
TommyCpp, line 1: Can we move this example to `opentelemetry-appender-log` folder? The examples is used to demonstrate the how to use this bridge after all and we already have too many examples in top-level example folder

opentelemetry-appender-log/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    fn flush(&self) {}
+}
+
+impl<P, L> OpenTelemetryLogBridge<P, L>
+where
+    P: LoggerProvider<Logger = L> + Send + Sync,
+    L: Logger + Send + Sync,
+{
+    pub fn new(level: Level, provider: &P) -> Self {
```
TommyCpp, line 43: nit: `level` -> `min_level`.

TommyCpp (commented):
small nits but overall looks good!

opentelemetry-appender-log/src/lib.rs
```diff
@@ -0,0 +1,61 @@
+use log::{Level, Metadata, Record};
+use opentelemetry_api::logs::{AnyValue, LogRecordBuilder, Logger, LoggerProvider, Severity};
+
+pub struct OpenTelemetryLogBridge<P, L>
+where
+    P: LoggerProvider<Logger = L> + Send + Sync,
+    L: Logger + Send + Sync,
+{
+    logger: L,
+    min_level: Level,
```
Robo210, line 10: I think in order for this to useful, the level can't be fixed at construction. Instead, the enabled check should query something like the sampler to see if the event is enabled. Otherwise systems that can change the collected level dynamically (e.g. ETW, user_events\eventheader) won't be able to function well unless the level here is always fixed at the "collect everything" value, which really defeats the whole point of having a level to begin with.

Robo210:
> The tracing-opentelemetry repo is not under CNCF/OpenTelemetry org

This doesn't seem like a good reason to duplicate the work they have already done instead of extend it over in their repo. Competing bridges will just bring about confusion and ill-will between projects, and unifying them later will be infinitely harder than not forking the community to begin with.

Plus I personally still don't understand what any of the benefits are for having the code live here vs. there. In the end everyone gets it from crates.io just the same.

djc:
> This doesn't seem like a good reason to duplicate the work they have already done instead of extend it over in their repo. Competing bridges will just bring about confusion and ill-will between projects, and unifying them later will be infinitely harder than not forking the community to begin with.
> 
> Plus I personally still don't understand what any of the benefits are for having the code live here vs. there. In the end everyone gets it from crates.io just the same.

Strong +1 from me.

opentelemetry-api/src/logs/logger.rs
cijothomas: Thanks for catching that! I was an accidental leftover I missed to remove.

examples/logs-basic/src/main.rs
cijothomas: Though I agree it doesn't look good to have too many top-level examples, I believe this one is best served in the top-level.
1. It should be okay to have a basic "hello world" style example for each of the signals (metrics, traces already exist) in the top level folder.
2. I can do some cleanup of the existing top-level examples to trim them down.

Thoughts?

cijothomas:
> > > The tracing-opentelemetry repo is not under CNCF/OpenTelemetry org
> > 
> > 
> > This doesn't seem like a good reason to duplicate the work they have already done instead of extend it over in their repo. Competing bridges will just bring about confusion and ill-will between projects, and unifying them later will be infinitely harder than not forking the community to begin with.
> > Plus I personally still don't understand what any of the benefits are for having the code live here vs. there. In the end everyone gets it from crates.io just the same.
> 
> @djc @Robo210 Repasting the below from my earlier reply.
> 
> > "The tracing-opentelemetry repo is not under CNCF/OpenTelemetry org. This topic was discussed in yesterday's SIG call too, and the consensus was to host the **tracing bridge here to begin with**. Once the components are here, we can evaluate potentially lifting moving it out elsewhere/merging with existing tracing-opentelemetry repo. (That crete already converts events as SpanEvents, but the logging bridge would convert them to OTel logs, so there are issues like duplication, context stamping etc. that need to be sorted out.)"
> 
> I do not have strong preference if the tracing bridge (for logs) should live in OpenTelemetry repo, or live along with the tracing-opentelemetry (which is currently only for traces). I will start a discussion with the maintainers of the tracing-opentelemetry repo to see if they are happy to host it there, and update back.
> 
> @djc The Logging Bridge PR just got merged and needs more polishing before it can be published and leveraged by outside repos. Would you allow this PR (and a similar appender for tracing logs) to proceed in this repo for now, while I gauge the opinions from the tracing-opentelemetry maintainers (to host it there once we have the logging bridge crates published), so that the logging work is unblocked ?

Tagging all maintainers, @jtescher @djc @TommyCpp @hdost to help unblock. (Sorry for pushing, I am trying to get this through (and similar one for tracing), so we can start testing the full logging story with OTel Rust). As mentioned in the earlier comment, happy to move the tracing appender to tracing-opentelemetry repo, once the basic version is ready, and we publish api/sdk that support logs to crates.io.

opentelemetry-appender-log/README.md
```diff
@@ -0,0 +1,23 @@
+![OpenTelemetry — An observability framework for cloud-native software.][splash]
+
+[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
+
+# OpenTelemetry Log Appender
+
+Community supported vendor integrations for applications instrumented with [`OpenTelemetry`].
+
+[![Crates.io: opentelemetry-contrib](https://img.shields.io/crates/v/opentelemetry-contrib.svg)](https://crates.io/crates/opentelemetry-contrib)
```
jtescher, line 9: These still point to contrib

jtescher (approved):
I think this is ok to move forward as long as it is clear the intent is to move it to the contrib repo once available. Could be clearer in that sense if this was simply added to the current `opentelemetry-contrib` crate for now, but doesn't seem urgent enough to block.

hdost (approved):
I think the intent here is not to step on any toes, but make some progress. Concur with @jtescher

examples/logs-basic/src/main.rs
cijothomas: Will have the example modified in a follow up PR to show correlation.

cijothomas:
@djc @Robo210 I have opened https://github.com/open-telemetry/opentelemetry-rust/issues/1080 to track the work for log-appender for tracing. I hope your concerns are addressed correctly in that issue. If not, lets use #1080 to continue that discussion.

Thanks everyone else for reviewing!

@TommyCpp Please let me know if the top level example is okay (i replied to your comment), and if yes, please merge. Thank you.

djc:
I don't want to block this, just provide guidance on the right approach to move forward. Honestly it's not clear to me why we need a bridge for the log crate (note that the crate is called "log", not "Log", so if the title is referring to that crate, that is confusing to me) when tracing already creates bridging functionality for the log crate natively.

KodrAus:
:wave: I'm currently helping maintain the `log` crate. I've been tinkering with the SDK independently and just stumbled upon this issue and thought I'd chime in.

> This would be moved to a opentelemetry owned contrib repo (once is it created), or, even to https://github.com/rust-lang/log if the maintainers agree.

A contrib repo may be a good start because `log` is currently in the Rust project's organization, so would probably need some input from the wider Libraries team before adopting anything new.

> Honestly it's not clear to me why we need a bridge for the log crate ... when tracing already creates bridging functionality for the log crate natively.

I think creating a bridge natively for `log` is reasonable even though `tracing` already also integrates with `log`. I wouldn't expect `tracing` to do that native integration indefinitely, especially as it's grown in usage. I think that was necessary to bootstrap something new, but I wouldn't think it unreasonable for them even to drop native `log` support in their `0.2.x` release. The `tracing` maintainers may have a totally different view though.

TommyCpp (approved):
Looks good!

davidbarsky:
> I think creating a bridge natively for `log` is reasonable even though `tracing` already also integrates with `log`. I wouldn't expect `tracing` to do that native integration indefinitely, especially as it's grown in usage. I think that was necessary to bootstrap something new, but I wouldn't think it unreasonable for them even to drop native `log` support in their `0.2.x` release. The `tracing` maintainers may have a totally different view though.

To be honest, I don't see a world where we drop support for `log`. Doing so will reduce the utility of both `log` and `tracing` and that would be a net-loss for the entire Rust ecosystem.

Issue #576, opened 2021-06-19
Serde support


jtescher:
Should the core `opentelemetry` crate support serde? There are standard serialization formats, it's unclear that this is necessary (or compliant with the intention of the spec).

TommyCpp:
I don't believe the spec explicitly requires to provide a serialization formats but rather it leave this questions to each propagator and exporter. I think in some case the serde could make things easier for exporter/propagator writers but usually exporters/propagators need some kind of conversion before serialization. Thus, serde is not a must here.

frigus02:
Good question. I'd say we probably don't need it. How does serde support work with backwards compatibility? If we add another field to a struct, which implements `Serialize` and `Deserialize`, would that count as a breaking change?

I like the idea of a separate crate in case there is demand for serialization and deserialization for some of the types.

jtescher:
Removed in #738

slinkydeveloper:
Hi, is there any alternative available right now for this?

slinkydeveloper:
For whoever stumbles on this, my workaround is to use `serde_with` and an intermediate struct:

```rust
#[derive(Serialize, Deserialize)]
struct SpanContextDef {
    trace_id: [u8; 16],
    span_id: [u8; 8],
    trace_flags: u8,
    is_remote: bool,
    trace_state: String,
}

// Provide a conversion to construct the remote type.
impl From<SpanContextDef> for SpanContext {
    fn from(def: SpanContextDef) -> Self {
        SpanContext::new(
            TraceId::from_bytes(def.trace_id), SpanId::from_bytes(def.span_id), TraceFlags::new(def.trace_flags), def.is_remote, TraceState::from_str(&def.trace_state).unwrap()
        )
    }
}

impl From<SpanContext> for SpanContextDef {
    fn from(ctx: SpanContext) -> Self {
        Self {
            trace_id: ctx.trace_id().to_bytes(),
            span_id: ctx.span_id().to_bytes(),
            trace_flags: ctx.trace_flags().to_u8(),
            is_remote: ctx.is_remote(),
            trace_state: ctx.trace_state().header()
        }
    }
}

// In the struct where you need SpanContext
#[serde_as(as = "FromInto<SpanContextDef>")] 
SpanContext spanContext
```

See https://docs.rs/serde_with/2.0.1/serde_with/guide/serde_as_transformations/index.html#convert-to-an-intermediate-type-using-into for more details about serde_with

[...]

Issue #437, opened 2021-01-22
Rename default branch to `main`


jtescher:
Requires a TC member to update. (cc @SergeyKanzhelev)

SergeyKanzhelev:
Done

PR #1075, opened 2023-05-23
Optional serde support for `SpanContext`

### LINKED ISSUE #1074: SpanContext should implement serde::Serialize, Deserialize
Per [Rust API Guidelines for Interoperability](https://rust-lang.github.io/api-guidelines/interoperability.html#data-structures-implement-serdes-serialize-deserialize-c-serde) I think [`SpanContext`](https://docs.rs/opentelemetry/latest/opentelemetry/trace/struct.SpanContext.html) should implement `serde::Serialize` and `Deserialize`.

I'll see about making this contribution myself, but wanted to open the issue in case there is disagreement with the approach. (From my understanding, we want to prefer idiomatic usage of the language features for the API.) This will help me and others in building domain-specific propagators ala [`.with_remote_span_context`](https://docs.rs/opentelemetry/latest/opentelemetry/trace/trait.TraceContextExt.html#tymethod.with_remote_span_context).

shaun-cox:
Fixes #1074

## Changes

Follow Rust API Guidelines in regards to `SpanContext` data structure implementing Serde's Serialize and Deserialize traits, optionally based on "serde" feature flag.


hdost (approved):
Makes sense and having behind a feature flag is good 👍🏼

jtescher:
I think this seems a little against the spirit of the propagator spec as this would suggest a simple way to bypass that API, but I don't feel strongly enough to block if others find this useful. @TommyCpp  thoughts?

djc:
> I think this seems a little against the spirit of the propagator spec as this would suggest a simple way to bypass that API, but I don't feel strongly enough to block if others find this useful. @TommyCpp thoughts?

Since we've had a few people speak up in favor of this in #576, it feels like we should accept it.

@jtescher what are the risks/issues with bypassing the propagator?

djc:
@shaun-cox what do you think about `SpanData` vs `SpanContext`? What's your use cases for specifically wanting to (de)serialize `SpanContext`?

shaun-cox:
@djc, @wperron, @jtescher thanks for the feedback and questions and patience as I attempt to learn and contribute here.

My use case is in in the context of a library that I'm wanting to instrument. This library communicates with other instances of itself over the network using a binary protocol. That binary protocol is already relying on `serde` and is performance sensitive. My aim is simply to enable distributed tracing across this library's network boundary.

My first thought was that I should limit myself to `opentelemetry-api` types, and not have my library take any dependency on `opentelemetry-sdk`, as I'm not building a generic exporter or anything, just wanting to stick to the "instrumenting" use case, but carry the necessary minimum context across my boundary.

I looked at `SpanData`, but it's not readily apparent to me how to leverage it for my use case. I'd like to avoid `TextMapPropagator` (for now) as it seems aimed at HTTP/W3C type stuff, which feels a lot more heavy-weight than what I need. IOW, I'd like to avoid allocating a temporary `HashMap<String, String>` via `TextMapPropagator` then to hand that to `serde` to serialize into my binary protocol.

I thought I was on the right track with this PR approach, after noticing that `TraceContextExt.with_remote_span_context` was the way to construct a new (correlated) `Context` at the receive side of my binary protocol after providing the deserialized `SpanContext`.  But if that is not the right approach, I'd love to be educated more...

I also noticed that `opentelemetry-api::propagation` module has some comments about `BinaryFormat`, but no implementation yet, so wasn't sure if there might be contributions needed in this area to implement the spec more?  I might be up for contributing there, if needed, and if it ties in with my use case better.

To be clear, I want to help push `opentelemetry-rust` along as I like what I see so far, and I don't just want to add a patch that suits my use case if it's not really "in line" with the general open telemetry vision/spec, which I am still learning.

Thanks!

shaun-cox:
I've also since understood that binary format propagators have been removed from the spec, and there is an [issue #437](https://github.com/open-telemetry/opentelemetry-specification/issues/437) to address adding them back at some point.

I think I understand the need for "open" binary protocols to propagate context, but I hope the community sees the interim need for context to be opaquely propagated inside "closed" binary protocols in the meantime, which is how I view my use case.

wperron:
> I think I understand the need for "open" binary protocols to propagate context, but I hope the community sees the interim need for context to be opaquely propagated inside "closed" binary protocols in the meantime, which is how I view my use case.

My issue here is that this is possible to do using the Propagator traits already. It's not clear to me how `Serialize`/`Deserialize` are supposed to interact with `Injector`/`Extractor`. I can see them being used in conjunction with one another, but for spec compliance reasons, `Injector`/`Extractor` must be the canonical way to propagate context across service boundary.

I'm not necessarily opposed to this change, I would just like to see a better use-case defined; Why do the existing traits not work for you? What are the possible drawbacks or footguns that come with this change?

djc:
> I can see them being used in conjunction with one another, but for spec compliance reasons, `Injector`/`Extractor` must be the canonical way to propagate context across service boundary.

Do we know why the spec mandates this? Leaving out useful features for spec compliance seems like a bad pattern. (And having a canonical way to do something doesn't maybe quite mandate that it's the only way?)

wperron:
It's also worth pointing out that Binary Propagators [are coming at some point](https://github.com/open-telemetry/opentelemetry-specification/issues/437) to the specification, which will solve the `inject to Map -> serialize to binary` two-step awkwardness

wperron:
> > I can see them being used in conjunction with one another, but for spec compliance reasons, `Injector`/`Extractor` must be the canonical way to propagate context across service boundary.
> 
> Do we know why the spec mandates this? Leaving out useful features for spec compliance seems like a bad pattern. (And having a canonical way to do something doesn't maybe quite mandate that it's the only way?)

I'm really not sure. I'm just an interested party here, might be worth asking in the otel-specification channel in the CNCF Slack.

Again though I'm not opposed to having `serde` in the mix, I'm sure there's tons of values in being able to serialize span data to JSON say, or whatever other format, but I do think that if we're talking about _Propagation_ specifically, we should stick to the standard.

Or, put differently, the `api` package should be as close to the standard as possible, and the `sdk` package can take some liberties. Which is why I was suggesting implementing `serde` on `SpanData`, as that's part of the `sdk` package, and imo the use-case for `serde` in the context of an Exporter is much stronger.

shaun-cox:
@wperron wrote:
> Maybe the best course of action would be an experimental BinaryPropagator in the contrib package?

Yeah, jtescher pointed out above that it's there at [`BinaryFormat`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-contrib/src/trace/propagator/binary/binary_propagator.rs#L13). I'll probably use that in the meantime, though it doesn't integrate with any `Injector/Extractor` abstractions.

All it does is serialize the `trace_id`, `span_id`, and `trace_flags` constituents of `SpanContext`, and does it using the `to_bytes()` methods on each of those.  (One could even argue/wonder why `SpanContext` doesn't just have a `to_bytes()` and `from_bytes()` methods as its constituents do, which is arguably as idiomatic for Rust as using `serde` for serialization.

I am also interpreting the following, from [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md#focus-on-capabilities-not-structure-compliance)
> It is preferable to have contributions follow the idioms of the language rather than conform to specific API names or argument patterns in the spec.

TommyCpp:
> I think this seems a little against the spirit of the propagator spec as this would suggest a simple way to bypass that API, but I don't feel strongly enough to block if others find this useful. @TommyCpp thoughts?

I think it's fine to introduce the `serde` trait behind a feature flag. It will be helpful for library developers to implement custom propagator and prevent end users to using it in accident

shaun-cox:
I also updated the PR to not actually serialize the `is_remote` field, and on deserialization, to set it to true, which honors the intention of the spec, I believe.

jtescher:
Something worth noting is that span context serialization is standardized through the propagation apis so that environments with multiple language implementations are more likely to properly propagate the context correctly. In that sense users should be discouraged from implementing their own serialization formats unless there is a compelling reason to do so (e.g. noticing that span context implements `Serialize` and propagating as json via `serde_json` would not generally be a recommended approach).

As we get closer to 1.0 for the trace signal we should be clear about what API guarantees we want to make, and how lower context readers will interpret and use the interfaces we expose.

shaun-cox:
We chatted briefly about this in the Rust SIG meeting today. I'm happy to close this and go think more on how to use binary propagators in my own protocol. We decided to leave the PR open in the meantime for more discussion/context.

shaun-cox:
While binary propagator contrib stuff wasn't applicable to my use case, I was able to work around with my own struct which is convertible from `SpanContext`, so closing this.

PR #1076, opened 2023-05-24
Remove Context from sync instruments

cijothomas:
Opening a PR which removes `Context` as part of the parameters to the measurement reporting API for Counter. ~If this is agreed, I can remove it for other sync instruments as well.~ (Updated to remove for all sync instruments) I am not sure if this was intentionally kept, as the Metrics API spec removed this long ago(?) or it never existed in the spec. The SDK could infer the current implicit context OR we could introduce a new ctr.AddWithContext() accepting explicit context if there is a need for passing non-implicit context..


hdost:
I believe @jtescher  added it last year.

jtescher:
I think the main use here would be for exemplars, but it's not being used yet. Could work that through first and see if we can delete it after that, or I suppose alternatively we could remove it now and re-add it if needed for exemplars.

lalitb:
It seems the [specs ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#synchronous-and-asynchronous-instruments) lets the language implementation decide whether to pass the Context parameter explicitly, or else use the current active context. C++ and Java provides overload methods with context parameter, Rust has this parameter as mandatory, while other languages use the currently active context.

cijothomas:
> I think the main use here would be for exemplars, but it's not being used yet. Could work that through first and see if we can delete it after that, or I suppose alternatively we could remove it now and re-add it if needed for exemplars.

Would you agree to the following plan:
1. Remove the Context parameter from instruments now. (This PR)
2. When working on Exemplar +  the capability to add attributes from Baggage (the only 2 features in Metrics that are spec-ed out which require Context), we can start with relying on the implicit context without requiring user to explicitly pass it.
3. If there is a desire to provide a different context than the implicit one, introduce a new api (counter.AddWithContext()) that accepts the context (like C++, Java)

opentelemetry-api/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Removed
+
+- Synchronous instruments no longer accepts `Context` while reporting
```
lzchen, line 7: Nit: Add pr #

lzchen:
Title should be renamed to only address sync counters

jtescher:
> Would you agree to the following plan:
> 
> 1. Remove the Context parameter from instruments now. (This PR)
> 2. When working on Exemplar +  the capability to add attributes from Baggage (the only 2 features in Metrics that are spec-ed out which require Context), we can start with relying on the implicit context without requiring user to explicitly pass it.
> 3. If there is a desire to provide a different context than the implicit one, introduce a new api (counter.AddWithContext()) that accepts the context (like C++, Java)

Yeah I think that's fine, we have both options when creating traces, I think long term having both is reasonable.

cijothomas:
@jtescher @hdost @lzchen This is rebased from main, and is ready for merge, once approved! Please take a look! @TommyCpp I'd like to ensure this is part of the upcoming release (as metrics already has breaking change, so why not have them all in one go!)

jtescher (approved):

PR #1077, opened 2023-05-24
Minor fixes for logs feature

Robo210:
Some minor changes to make the new logs feature more usable (especially when making a custom LoggerProvider), and make it a little bit more symmetrical with tracing.
- Implement Display for AnyValue, just like it is implemented for Value
  - Use json-style strings for lists and maps, just like Value does for Array
- Let LogData be constructed by things outside of the current crate by removing non_exhaustive
  - If new fields are being added, that should probably be a breaking change anyway
- Allow the EvictedHashMap iterator to be cloned

### DIFF
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -89,10 +89,6 @@ impl fmt::Display for AnyValue {
             AnyValue::Boolean(v) => v.fmt(fmt),
             AnyValue::Bytes(v) => common::display_array_str(v.as_slice(), fmt),
             AnyValue::ListAny(values) => {
-                if values.is_empty() {
-                    return Ok(());
-                }
-
                 write!(fmt, "[")?;
                 for (i, t) in values.iter().enumerate() {
                     if i > 0 {
@@ -103,10 +99,6 @@ impl fmt::Display for AnyValue {
                 write!(fmt, "]")
             }
             AnyValue::Map(map) => {
-                if map.is_empty() {
-                    return Ok(());
-                }
-
                 write!(fmt, "{{")?;
                 for (i, (k, v)) in map.into_iter().enumerate() {
                     if i > 0 {
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -1,8 +1,9 @@
 use crate::{
+    common,
     trace::{SpanContext, SpanId, TraceContextExt, TraceFlags, TraceId},
     Array, Key, OrderMap, StringValue, Value,
 };
-use std::{borrow::Cow, time::SystemTime};
+use std::{borrow::Cow, fmt, time::SystemTime};
 
 #[derive(Debug, Clone, Default)]
 #[non_exhaustive]
@@ -77,6 +78,48 @@ pub enum AnyValue {
     ListAny(Vec<AnyValue>),
     /// A map of string keys to `Any` values, arbitrarily nested.
     Map(OrderMap<Key, AnyValue>),
+}
+
+impl fmt::Display for AnyValue {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            AnyValue::Int(v) => v.fmt(fmt),
+            AnyValue::Double(v) => v.fmt(fmt),
+            AnyValue::String(v) => fmt.write_str(v.as_str()),
+            AnyValue::Boolean(v) => v.fmt(fmt),
+            AnyValue::Bytes(v) => common::display_array_str(v.as_slice(), fmt),
+            AnyValue::ListAny(values) => {
+                if values.is_empty() {
+                    return Ok(());
+                }
+
+                write!(fmt, "[")?;
+                for (i, t) in values.iter().enumerate() {
+                    if i > 0 {
+                        write!(fmt, ",")?;
+                    }
+                    write!(fmt, "{t}")?;
+                }
+                write!(fmt, "]")
+            }
+            AnyValue::Map(map) => {
+                if map.is_empty() {
+                    return Ok(());
+                }
+
+                write!(fmt, "{{")?;
+                let mut i = 0;
+                for (k, v) in map {
+                    if i > 0 {
+                        write!(fmt, ",")?;
+                    }
+                    write!(fmt, "{k}: {v}")?;
+                    i += 1;
+                }
+                write!(fmt, "}}")
+            }
+        }
+    }
 }
 
 macro_rules! impl_trivial_from {
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -108,13 +108,11 @@ impl fmt::Display for AnyValue {
                 }
 
                 write!(fmt, "{{")?;
-                let mut i = 0;
-                for (k, v) in map {
+                for (i, (k, v)) in map.into_iter().enumerate() {
                     if i > 0 {
                         write!(fmt, ",")?;
                     }
                     write!(fmt, "{k}: {v}")?;
-                    i += 1;
                 }
                 write!(fmt, "}}")
             }
--- a/opentelemetry-sdk/src/export/logs/mod.rs
+++ b/opentelemetry-sdk/src/export/logs/mod.rs
@@ -2,16 +2,19 @@
 use crate::Resource;
 use async_trait::async_trait;
 use opentelemetry_api::{
-    logs::{LogError, LogRecord, LogResult},
+    logs::{LogRecord, LogResult},
     InstrumentationLibrary,
 };
 use std::{borrow::Cow, fmt::Debug};
+
+/// Describes the result of an export.
+pub type ExportResult = LogResult<()>;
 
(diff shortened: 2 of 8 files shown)


opentelemetry-api/src/logs/record.rs
```diff
@@ -79,6 +80,48 @@ pub enum AnyValue {
...
+                for (i, t) in values.iter().enumerate() {
+                    if i > 0 {
+                        write!(fmt, ",")?;
+                    }
+                    write!(fmt, "{t}")?;
+                }
+                write!(fmt, "]")
+            },
+            AnyValue::Map(map) => {
+                if map.is_empty() {
```
djc, line 106: Same as for arrays.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -2,24 +2,26 @@
...
 use async_trait::async_trait;
 use opentelemetry_api::{
-    logs::{LogError, LogRecord, LogResult},
+    logs::{LogRecord, LogResult},
     InstrumentationLibrary,
 };
 use std::{borrow::Cow, fmt::Debug};
 
+/// Describes the result of an export.
+pub type ExportResult = LogResult<()>;
```
djc, line 11: This doesn't seem like an improvement to me. What's the motivation here?

opentelemetry-sdk/src/logs/config.rs
```diff
@@ -1,5 +1,10 @@
 use std::borrow::Cow;
 
+/// Default trace configuration
+pub fn config() -> Config {
```
djc, line 4: Why? Using `Config::default()` seems like an idiomatic convention, this is more opaque.

opentelemetry-sdk/src/trace/evicted_hash_map.rs
```diff
@@ -137,7 +137,7 @@ impl<'a> IntoIterator for &'a EvictedHashMap {
 }
 
 /// An iterator over the entries of an `EvictedHashMap`.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
djc, line 140: What's the motivation here?

opentelemetry-sdk/src/export/logs/mod.rs
Robo210: - The SpanExporter returns an ExportResult. Having these two very similar things follow the same patterns is something I see as an improvement when trying to write code that functions as both a SpanExporter and LogExporter.
  - It would be even nicer if they both could use the *same* ExportResult, but that's not something that belongs in this change.
- The LogExporter ExportResult as-is is unreferenced.
- The Exporter should probably return an ExportResult. That's what the names imply.

opentelemetry-sdk/src/logs/config.rs
Robo210: Because this is what the `tracing` module provides: https://github.com/open-telemetry/opentelemetry-rust/blob/0a893ce7dceb182ba443a944d49ac7804237bbe9/opentelemetry-sdk/src/trace/config.rs#LL12C1-L15C2

My goal is to bring logs and tracing a little bit closer so I don't have to write as much duplicate code.

opentelemetry-sdk/src/trace/evicted_hash_map.rs
Robo210: To be able to operate over a collection of attributes generically, cloning the iterator multiple times so the clones can be mutated.

```rust
pub(crate) fn get_attributes_as_json<'a, C>(attribs: C) -> String
where C: Iterator<Item = (&'a Key, &'a Value)>
{
    do_thing_with_attributes(attribs.clone());
    do_a_second_thing_with_attributes(attribs));
}
```

The underlying iterator type is Clone, and the OrderedMap iterator for Log attributes is Clone. Having the wrapper struct implement Clone shouldn't be impactful except through its absence.

opentelemetry-api/src/logs/record.rs
```diff
@@ -79,6 +80,48 @@ pub enum AnyValue {
...
+        match self {
+            AnyValue::Int(v) => v.fmt(fmt),
+            AnyValue::Double(v) => v.fmt(fmt),
+            AnyValue::String(v) => fmt.write_str(v.as_str()),
+            AnyValue::Boolean(v) => v.fmt(fmt),
+            AnyValue::Bytes(v) => common::display_array_str(v.as_slice(), fmt),
+            AnyValue::ListAny(values) => {
+                if values.is_empty() {
+                    return Ok(());
+                }
```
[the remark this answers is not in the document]
Robo210: Sure

opentelemetry-sdk/src/logs/config.rs
djc: Okay, how about we remove this on the `tracing` side?

opentelemetry-sdk/src/trace/evicted_hash_map.rs
djc: Okay, I guess. Seems pretty niche?

opentelemetry-sdk/src/export/logs/mod.rs
djc: Personally, I find `Error` and `Result` type aliases to be a net negative when reading code. IMO they make it harder to understand the actual types that are being passed around by wrapping them in a layer that's hard to pierce for humans (have to look the import or alias definition) while being almost completely transparent to the compiler (no type safe distinction). As such, I feel even stronger about having type aliases of type aliases.

On top of that, giving different types the same name seems to make the type aliases even more confusing for humans.

I understand the goal of consistency between different areas of the code, but if there are inevitable differences between certain mostly-similar components then in my mind we should actually make the differences stand out more rather than obfuscating them.

hdost: Perhaps this is something we could add to contributing.md because things like this are a bit "style guide"-esque

opentelemetry-sdk/src/logs/config.rs
Robo210: I'm not going to make such a breaking change in this PR. My goal is only to bring logs up to where tracing is, before logs gets widely depended on and changes become harder to make.

opentelemetry-sdk/src/export/logs/mod.rs
Robo210: > if there are inevitable differences between certain mostly-similar components then in my mind we should actually make the differences stand out more rather than obfuscating them.

Setting aside the particulars of this precise change, from my perspective as an outside developer writing against an unfamiliar SDK, differences within an SDK are confusing and can be frustrating. Internal consistency (be it patterns, naming, formatting, style, tone, etc.) and predictability aid in helping new users internalize the SDK and use it as a familiar tool in their toolbox, so they can focus on building what they need to rather than focus on the tools themselves.

Robo210: Also, I'd argue that since LogResult<T> is generic, but T must always be () when being returned from something implementing the exporter trait, then a type alias that always provides the correct generic type is beneficial. Every use of LogResult has the same generic type parameter though, so maybe it shouldn't be generic at all.

opentelemetry-sdk/src/logs/config.rs
```diff
@@ -1,6 +1,5 @@
-use std::borrow::Cow;
-
 use crate::Resource;
+use std::borrow::Cow;
```
Robo210, line 2: Not sure why this line got moved. Maybe cargo fmt did it?

hdost:
@Robo210 Is this something you're still actively working on , or would you like use to do another round of reviews?

Robo210:
All the comments have been addressed, and the ones that received pushback were reverted. I believe all that's left is for somebody to take another look at it.

opentelemetry-api/src/logs/record.rs
```diff
@@ -79,6 +80,38 @@ pub enum AnyValue {
     Map(OrderMap<Key, AnyValue>),
 }
 
+impl fmt::Display for AnyValue {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
```
lalitb, line 84: nit - change the variable name to say `f` instead of `fmt` for better readability. A `fmt` is also a module.

opentelemetry-api/src/logs/record.rs
```diff
@@ -79,6 +80,38 @@ pub enum AnyValue {
...
+        match self {
+            AnyValue::Int(v) => v.fmt(fmt),
+            AnyValue::Double(v) => v.fmt(fmt),
+            AnyValue::String(v) => fmt.write_str(v.as_str()),
+            AnyValue::Boolean(v) => v.fmt(fmt),
+            AnyValue::Bytes(v) => common::display_array_str(v.as_slice(), fmt),
```
lalitb, line 90: Not sure if `display_array_str` is write method to format byte vector, as it will try to display as string. We can have a similar method `display_array_byte` which print the bytes as a sequence of hexadecimal values:

`write!(f, "{:02x}", t)?;`

cijothomas:
@Robo210 There are lot of unresolved conversations. Could you respond to them and ask to resolve them?

PR #1078, opened 2023-05-25
Add js-sys as dependency for api crate when building wasm targets

davids-work:
It's used to implement now() for (non-WASI) WASM targets, but was not included as a dependency

Fixes #

## Changes
Adds js-sys as a dependency to opentelemetry_api when building for non-WASI WASM targets. Without this change, the crate will not compile for these target as `js_sys::Date::now()` is used in opentelemetry-api/src/lib.rs


jtescher (approved):

jtescher:
Thanks @davids-work!

PR #1082, opened 2023-05-26
Improve stress tests

lzchen:
The original stress tests produce output about ~4m requests/second which is significantly less than what is expected (approx 15-20m). This pr fixes a lot of points of contention that were using cpu that is unrelated to testing the functions themselves.

1. rip out rayon -> this is another layer of abstraction that may be adding more cpu usage in the heuristics of spawning threads 

Solution: Spawned threads directly instead 

2. cache lines -> make sure cores get access to their own individual count in separate cache lines so cpu usage won't be spent on having to obtain ownership of the cache lines 

Solution: Padded struct with atomicu64

3. perhaps amortize and reduce the frequency at which you poll the count 

Solution: Added batching of 1000 before adding to shared resource

4. cache contention with count calculation 

Solution: Added sleep of 5000ms after outputting count

Running this gave me this output:

```
Throughput: 18013400.00 requests/sec
Throughput: 17975800.00 requests/sec
Throughput: 17620800.00 requests/sec
Throughput: 17726600.00 requests/sec
Throughput: 17998400.00 requests/sec
Throughput: 18033400.00 requests/sec
```


stress/src/throughput.rs
```diff
@@ -1,12 +1,19 @@
...
 
 const SLIDING_WINDOW_SIZE: u64 = 2; // In seconds
 
 static STOP: AtomicBool = AtomicBool::new(false);
 
+#[repr(C)]
+#[derive(Default)]
+struct WorkerStats {
+    count: AtomicU64,
+    padding: [u64; 15],
```
cijothomas, line 14: nit: add a comment about why padding is done to help future readers

stress/src/throughput.rs
```diff
@@ -15,37 +22,65 @@ where
...
+            let elapsed = end_time.duration_since(start_time).as_secs();
+            if elapsed >= SLIDING_WINDOW_SIZE {
+                let total_count_u64: u64 = worker_stats_shared_monitor
+                    .iter()
+                    .map(|worker_stat| worker_stat.count.load(Ordering::Relaxed))
+                    .sum();
+                let current_count = total_count_u64 - total_count_old;
+                total_count_old = total_count_u64;
+                let throughput = current_count as f64 / elapsed as f64;
+                println!("Throughput: {:.2} requests/sec", throughput);
```
cijothomas, line 51:
```suggestion
                println!("Throughput: {:.2} iterations/sec", throughput);
```

stress/src/throughput.rs
```diff
@@ -15,37 +22,65 @@ where
...
+    let func_arc = Arc::new(func);
+    let mut worker_stats_vec: Vec<WorkerStats> = Vec::new();
+
+    for _ in 0..num_threads {
+        worker_stats_vec.push(WorkerStats::default());
+    }
+    let worker_stats_shared = Arc::new(worker_stats_vec);
+    let worker_stats_shared_monitor = Arc::clone(&worker_stats_shared);
+
+    let handle1 = thread::spawn(move || {
```
cijothomas, line 37: nit: rename to handle_main_thread or similar

stress/src/throughput.rs
```diff
@@ -15,37 +22,65 @@ where
...
-        if STOP.load(Ordering::SeqCst) {
-            break;
-        }
+    handles.push(handle1);
+
+    for thread_index in 0..num_threads - 1 {
+        let worker_stats_shared = Arc::clone(&worker_stats_shared);
+        let func_arc_clone = Arc::clone(&func_arc);
+        let handle = thread::spawn(move || loop {
+            for _ in 0..1000 {
```
cijothomas, line 70: its good to extract these to constants above, for better readability than magic numbers.

cijothomas (approved):
Awesome to see this!
Left few nits. Once this is merged, we can modify the actual metric emission to be more dynamic instead of repeatedly reporting measurements with same attributes.

stress/src/throughput.rs
```diff
@@ -15,37 +22,65 @@ where
...
-    let mut total_count_old: u64 = 0;
-    let total_count = Arc::new(AtomicU64::new(0));
-    let total_count_clone = Arc::clone(&total_count);
-
-    rayon::spawn(move || {
-        (0..num_cpus::get()).into_par_iter().for_each(|_| loop {
-            func();
-            total_count_clone.fetch_add(1, Ordering::SeqCst);
+    let num_threads = num_cpus::get_physical();
+    println!("Number threads: {}", num_threads);
```
lalitb, line 26: Nit - I understand this is stress test, and so would like to utilize all the cores. But as the enhancement separately, if we can also make num_threads configurable (optionally passed as command line argument, and use total available cores otherwise), it will help test how the metrics SDK scales as we increase number of concurrent requests.

lalitb (approved):

lalitb (approved):

stress/src/throughput.rs
```diff
@@ -15,37 +22,65 @@ where
...
-    let mut end_time = Instant::now();
-    let mut total_count_old: u64 = 0;
-    let total_count = Arc::new(AtomicU64::new(0));
-    let total_count_clone = Arc::clone(&total_count);
-
-    rayon::spawn(move || {
-        (0..num_cpus::get()).into_par_iter().for_each(|_| loop {
-            func();
-            total_count_clone.fetch_add(1, Ordering::SeqCst);
+    let num_threads = num_cpus::get_physical();
```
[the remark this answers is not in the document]
lzchen: When I run the test on a cpu that has more logical cores (12) than physical cores (6), it seems to greatly slow down the throughput. This is most likely due to the threads created conflicting with each other and trying to access the same memory at once, which contributed greatly to cpu usage.

stress/src/throughput.rs
lzchen: Can leave this for a seperate pr.

TommyCpp (approved):

hdost:
Awesome progress !

PR #1084, opened 2023-05-28
fix: allow span id to be less than 16 characters in jaeger propagator

### LINKED ISSUE #1047: Jaeger propagator doesn't accept `span-id` shorter than 16 characters
Sometimes we receive requests from upstream services that have `uber-trace-id` in their headers with `span-id` part of the `uber-trace-id` being shorter than 16 characters. This results in failed trace propagation due to the fact that Jaeger propagator in `opentelemetry-rust` [requires](https://github.com/open-telemetry/opentelemetry-rust/blob/3f9689dcf72a1f976bf0566b75cd396751761928/opentelemetry-jaeger/src/lib.rs#L487) `span-id`s to be exactly of length 16.

Requiring `span-id`s to be exactly of length 16 on a receiver seems to go against [Jaeger specification](https://www.jaegertracing.io/docs/1.22/client-libraries/#tracespan-identity), which states that receivers MUST accept hex-strings shorter than 16 characters and 0-pad them on the left.

TommyCpp:
Fixes #1047 

## Changes

When span id is less than 16 characters, padding 0 to left

### DIFF
--- a/opentelemetry-jaeger/src/lib.rs
+++ b/opentelemetry-jaeger/src/lib.rs
@@ -484,11 +484,18 @@ mod propagator {
 
         /// Extract span id from the header.
         fn extract_span_id(&self, span_id: &str) -> Result<SpanId, ()> {
-            if span_id.len() != 16 {
-                return Err(());
+            match span_id.len() {
+                // exact 16
+                16 => SpanId::from_hex(span_id).map_err(|_| ()),
+                // more than 16 is invalid
+                17.. => Err(()),
+                // less than 16 will result padding on left
+                _ => {
+                    let mut padded = String::with_capacity(16 - span_id.len());
+                    padded.push_str(span_id);
+                    SpanId::from_hex(&padded).map_err(|_| ())
+                }
             }
-
-            SpanId::from_hex(span_id).map_err(|_| ())
         }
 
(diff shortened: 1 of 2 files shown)


opentelemetry-jaeger/src/lib.rs
```diff
@@ -484,11 +484,18 @@ mod propagator {
...
-            if span_id.len() != 16 {
-                return Err(());
+            match span_id.len() {
+                // exact 16
+                16 => SpanId::from_hex(span_id).map_err(|_| ()),
+                // more than 16 is invalid
+                17.. => Err(()),
+                // less than 16 will result padding on left
+                _ => {
+                    let mut padded = String::with_capacity(16 - span_id.len());
```
jtescher, line 494: `with_capacity` will allocate enough room but not left pad. I think format may be easier? 

```rust
let padded = format!("{span_id:0>16}");
```

jtescher (changes requested):

opentelemetry-jaeger/src/lib.rs
TommyCpp: Good point. Let me revise it

jtescher (approved):

PR #1085, opened 2023-05-30
Add log appender for tracing

cijothomas:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-appender-tracing/examples/basic.rs
+++ b/opentelemetry-appender-tracing/examples/basic.rs
@@ -1,0 +1,25 @@
+//! run with `$ cargo run --example basic --all-features
+
+use opentelemetry_appender_tracing::{layer};
+use tracing::{info, error};
+use opentelemetry_sdk::{logs::{LoggerProvider, Config}, Resource};
+use opentelemetry_api::KeyValue;
+use tracing_subscriber::prelude::*;
+
+fn main() {
+    let exporter  = opentelemetry_stdout::LogExporter::default();
+    let provider: LoggerProvider = LoggerProvider::builder()
+    .with_config(
+        Config::default().with_resource(Resource::new(vec![KeyValue::new(
+            "service.name",
+            "log-appender-tracing-example",
+        )])),
+    )
+    .with_simple_exporter(exporter)
+    .build();
+    let layer  = layer::OpenTelemetryTracingBridge::new(&provider);
+    tracing_subscriber::registry().with(layer).init();
+
+    info!(fruit = "apple", price = 2.99, message = "hello from from price");
+    drop(provider);
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
     "opentelemetry-http",
     "opentelemetry-jaeger",
     "opentelemetry-appender-log",
+    "opentelemetry-appender-tracing",
     "opentelemetry-otlp",
     "opentelemetry-prometheus",
     "opentelemetry-proto",
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -1,0 +1,2 @@
+# Changelog
+
--- a/opentelemetry-appender-tracing/CODEOWNERS
+++ b/opentelemetry-appender-tracing/CODEOWNERS
@@ -1,0 +1,5 @@
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+
+# For anything not explicitly taken by someone else:
+*  @open-telemetry/rust-approvers
--- a/opentelemetry-appender-tracing/Cargo.toml
+++ b/opentelemetry-appender-tracing/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "opentelemetry-appender-tracing"
+version = "0.1.0"
+edition = "2021"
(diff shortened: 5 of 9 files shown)


opentelemetry-appender-tracing/examples/basic.rs
```diff
@@ -0,0 +1,25 @@
...
+            "service.name",
+            "log-appender-tracing-example",
+        )])),
+    )
+    .with_simple_exporter(exporter)
+    .build();
+    let layer  = layer::OpenTelemetryTracingBridge::new(&provider);
+    tracing_subscriber::registry().with(layer).init();
+
+    info!(fruit = "apple", price = 2.99, message = "hello from from price");
```
cijothomas, line 23: This is just an initial version.
We need to discuss and finalize
1. Move this to a new repo under tokio/tracing org OR merge with tokio/tracing-opentelemetry
2. Let users configure where the events should be sent to - either as Logs or SpanEvents.

cijothomas, line 23: [Aaron] qn. Why use tracing for logging at all?
log, slog, tracing (events  API) - 3 populate logs.

tracing-otel already converts events into SpanEvents, so why is that no sufficient already, and why we need another "bridge" to convert to Logs.?
[Cijo]
1. SpanEvents require Span, and Span's sampling affects SpanEvents.
2. OTel has limits/throttle on the number of SpanEvents per Span
3. SpanEvents are given to Exporters, only when the Span itself is ended/dropped, as opposed to Logs(events) being sent to Exporters immediately (depending on Export Processor.)

[Lalit]
tracing allows events to be created outside Span, but SpanEvents in OTel API does not allow it outside tracing API.

The log facade could be used for pure logging., but does not allow structured logs.

Tracing crate already has ability to emit Logs Always (need to explore - https://docs.rs/tracing/latest/tracing/#emitting-log-records)

event!(Level::TRACE, greeting = ?my_struct);
How do we user decide if this should go as span_event or log?
1. We need to investigate/explore this more. (there may be "Filter" that can be leverages Metadata)
2. We need end-user feedback.


Python
Log API
Otel Tracing API (Spans, SpanEvents)

Rust
Log Crate -- the answer is similar to Python/other languages
Tracing crate - that has "events" *and* "Span", so the distinction is not easy to make.
Otel Tracing API (Spans, SpanEvents)

jtescher: My vote for 1) would be to merge into tracing-opentelemetry, and 2) support whichever is configured as there may be a case where traces _and_ logs would be requested as they could be sent to different exporters.

cijothomas: There is an issue of potential duplication:

1. A user is currently using tracing for Spans and Events, and have currently configured tracing-opentelemetry, so they end up as OTel Spans and SpanEvents respectively.

2. The user now wants to use tracing for standalone Events (i.e outside of any Spans). They'll be using the Events API from tracing. And they setup the equivalent of this PR (so that Events gets sent as LogRecords). But this also causes a side-effect - the Events from 1, will now be sent both as SpanEvents and as LogRecords. This is a cost concern.

@jtescher Do you have any suggestions on fixing the above? I think an option in tracing-opentelemetry to ignore Events could work? By default, if user does 1,2, they get events as SpanEvents and LogRecords. If that is not desired, they can turn off SpanEvents from 1 or turn off LogRecords from 2.

djc: We should definitely make sure that tracing-opentelemetry is configurable about whether it should turn events into (a) SpanEvents, (b) LogRecords or (c) both.

davidbarsky: > Tracing crate already has ability to emit Logs Always (need to explore - https://docs.rs/tracing/latest/tracing/#emitting-log-records)

You shouldn't enable this feature for `opentelemetry-rust`. it's meant for libraries like hyper or tokio to smoothly transition to `tracing` in a minor semver release, allowing users to not immediately lose logging when then update their dependencies.

> How do we user decide if this should go as span_event or log? [...] We need to investigate/explore this more. (there may be "Filter" that can be leverages Metadata)

Using [`dynamic_filter_fn`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/fn.dynamic_filter_fn.html) is an option, but be sure to benchmark this since it might be a little too costly. We can always improve it on the tracing side of things though, especially since there is some low-hanging, performance-related fruit in `tracing-subscriber` that we've been meaning to improve. Y'all saying that it's a little too slow would sufficient for me to justify the positive impact of me working on fixing it.

cijothomas:
@djc @jtescher @davidbarsky @valkum I marked this PR as non-draft, to unblock progress. Once we release Log Bridge API/SDK, this will be moved to tracing-opentelemetry repo (with some configs to let user decide how to export events). We can continue the discussion about how exactly that config looks like in tracing-opentelemetry repo. I'll need some helps there, will open an issue there.

For now, this PR (and follow ups here), can help validate the Log Bridge API/SDK., help test its perf etc.

cijothomas:
@open-telemetry/rust-approvers Could you help review? 
Also see : https://github.com/open-telemetry/opentelemetry-rust/pull/1085#issuecomment-1583441064

lalitb (approved):

hdost (approved):

PR #1088, opened 2023-05-31
Simplify expression and improve perf to get parent span id

shaun-cox:
## Changes

Avoids calling `Context.has_active_span()` followed by `Context.span()`, which is two lookups,
and instead relies on checking for an invalid span context on the return value of the latter.

Code is a little easier to understand and shows about 4-6% perf improvement in start-end-span benchmarks.

My assumption here is that a span with an invalid span context could never be made active in the first place.

### DIFF
--- a/opentelemetry-sdk/src/trace/tracer.rs
+++ b/opentelemetry-sdk/src/trace/tracer.rs
@@ -150,23 +150,22 @@ impl opentelemetry_api::trace::Tracer for Tracer {
         let span_kind = builder.span_kind.take().unwrap_or(SpanKind::Internal);
         let mut attribute_options = builder.attributes.take().unwrap_or_default();
         let mut link_options = builder.links.take();
-        let mut parent_span_id = SpanId::INVALID;
-        let trace_id;
-
-        let parent_span = if parent_cx.has_active_span() {
-            Some(parent_cx.span())
-        } else {
-            None
-        };
 
         // Build context for sampling decision
-        if let Some(sc) = parent_span.as_ref().map(|parent| parent.span_context()) {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/Cargo.toml
```diff
@@ -37,7 +37,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dev-dependencies]
 bincode = "1.2"
-criterion = "0.4.0"
+criterion = "0.5.0"
```
djc, line 40: Not sure this really belongs in this PR. Also, might run into MSRV issues.

shaun-cox: Ah, sorry... This is a "workflow issue" on my end. (pushed to the wrong branch.)

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -150,23 +150,22 @@ impl opentelemetry_api::trace::Tracer for Tracer {
...
+                (
+                    parent_cx.span().span_context().span_id(),
+                    parent_cx.span().span_context().trace_id(),
+                )
+            } else {
+                (
+                    SpanId::INVALID,
+                    builder
+                        .trace_id
+                        .unwrap_or_else(|| config.id_generator.new_trace_id()),
```
jtescher, line 166: Hm I don't think this will have identical behavior to the previous implementation as children of invalid spans may be valid (have valid trace ids if not explicitly set to invalid in builder).

jtescher:
Is the perf similar if you get the parent span context once and then check if it's valid subsequently, rather than checking if there is an active span before extracting the span context?

PR #1089, opened 2023-06-01
Make TraceContextExt::span() return an Option

djc:
## Changes

The changes that @shaun-cox made in #1088 make sense to me, but I wanted to look into the core `TraceContextExt` API, which I don't think makes much sense for Rust. Instead of having a `span()` method that fabricates a `NOOP_SPAN` and yields if it there is no active span and a separate `has_active_span()` to check if there is a currently active span, it seems more idiomatic to make `span()` yield an `Option`.

### DIFF
--- a/examples/actix-http/src/main.rs
+++ b/examples/actix-http/src/main.rs
@@ -28,7 +28,9 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
 async fn index() -> &'static str {
     let tracer = global::tracer("request");
     tracer.in_span("index", |ctx| {
-        ctx.span().set_attribute(Key::new("parameter").i64(10));
+        if let Some(span) = ctx.span() {
+            span.set_attribute(Key::new("parameter").i64(10));
+        }
         "Index"
     })
 }
@@ -45,8 +47,9 @@ async fn main() -> std::io::Result<()> {
             .wrap(Logger::default())
             .wrap_fn(move |req, srv| {
                 tracer.in_span("middleware", move |cx| {
-                    cx.span()
-                        .set_attribute(Key::new("path").string(req.path().to_string()));
+                    if let Some(span) = cx.span() {
+                        span.set_attribute(Key::new("path").string(req.path().to_string()));
+                    }
                     srv.call(req).with_context(cx)
                 })
             })
--- a/examples/actix-hyper/src/main.rs
+++ b/examples/actix-hyper/src/main.rs
@@ -20,7 +20,9 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
 async fn index() -> &'static str {
     let tracer = global::tracer("request");
     tracer.in_span("index", |ctx| {
-        ctx.span().set_attribute(Key::new("parameter").i64(10));
+        if let Some(span) = ctx.span() {
+            span.set_attribute(Key::new("parameter").i64(10));
+        }
         "Index"
     })
 }
@@ -37,8 +39,10 @@ async fn main() -> std::io::Result<()> {
             .wrap(Logger::default())
             .wrap_fn(move |req, srv| {
                 tracer.in_span("middleware", move |cx| {
-                    cx.span()
-                        .set_attribute(Key::new("path").string(req.path().to_string()));
+                    if let Some(span) = cx.span() {
+                        span.set_attribute(Key::new("path").string(req.path().to_string()));
+                    }
+
                     srv.call(req).with_context(cx)
                 })
             })
--- a/examples/actix-udp/src/main.rs
+++ b/examples/actix-udp/src/main.rs
@@ -25,7 +25,10 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
 async fn index() -> &'static str {
     let tracer = global::tracer("request");
     tracer.in_span("index", |ctx| {
-        ctx.span().set_attribute(Key::new("parameter").i64(10));
+        if let Some(span) = ctx.span() {
+            span.set_attribute(Key::new("parameter").i64(10));
+        }
+
         "Index"
     })
 }
@@ -42,8 +45,10 @@ async fn main() -> std::io::Result<()> {
             .wrap_fn(|req, srv| {
                 let tracer = global::tracer("request");
                 tracer.in_span("middleware", move |cx| {
-                    cx.span()
-                        .set_attribute(Key::new("path").string(req.path().to_string()));
+                    if let Some(span) = cx.span() {
+                        span.set_attribute(Key::new("path").string(req.path().to_string()));
+                    };
+
                     srv.call(req).with_context(cx)
                 })
             })
--- a/examples/aws-xray/src/client.rs
+++ b/examples/aws-xray/src/client.rs
@@ -45,7 +45,8 @@ async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sy
 
     let res = client.request(req.body(Body::from("Hallo!"))?).await?;
 
-    cx.span().add_event(
+    // `cx` initialized with span above, so unwrapping is safe
+    cx.span().unwrap().add_event(
         "Got response!".to_string(),
         vec![KeyValue::new("status", res.status().to_string())],
     );
--- a/examples/basic-otlp-http/src/main.rs
+++ b/examples/basic-otlp-http/src/main.rs
@@ -62,7 +62,11 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     histogram.record(&cx, 5.5, COMMON_ATTRIBUTES.as_ref());
 
     tracer.in_span("operation", |cx| {
-        let span = cx.span();
+        let span = match cx.span() {
+            Some(span) => span,
+            None => return,
+        };
+
         span.add_event(
             "Nice operation!".to_string(),
             vec![Key::new("bogons").i64(100)],
@@ -70,7 +74,11 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         span.set_attribute(ANOTHER_KEY.string("yes"));
 
         tracer.in_span("Sub operation...", |cx| {
-            let span = cx.span();
+            let span = match cx.span() {
+                Some(span) => span,
+                None => return,
+            };
+
             span.set_attribute(LEMONS_KEY.string("five"));
 
             span.add_event("Sub span event", vec![]);
--- a/examples/basic-otlp/src/main.rs
+++ b/examples/basic-otlp/src/main.rs
@@ -80,7 +80,11 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     histogram.record(&cx, 5.5, COMMON_ATTRIBUTES.as_ref());
 
     tracer.in_span("operation", |cx| {
-        let span = cx.span();
+        let span = match cx.span() {
+            Some(span) => span,
+            None => return,
+        };
+
         span.add_event(
             "Nice operation!".to_string(),
             vec![Key::new("bogons").i64(100)],
@@ -88,7 +92,11 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         span.set_attribute(ANOTHER_KEY.string("yes"));
 
         tracer.in_span("Sub operation...", |cx| {
-            let span = cx.span();
+            let span = match cx.span() {
+                Some(span) => span,
+                None => return,
+            };
+
             span.set_attribute(LEMONS_KEY.string("five"));
 
             span.add_event("Sub span event", vec![]);
--- a/examples/datadog/src/main.rs
+++ b/examples/datadog/src/main.rs
@@ -24,7 +24,11 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         .install_simple()?;
 
     tracer.in_span("foo", |cx| {
-        let span = cx.span();
+        let span = match cx.span() {
+            Some(span) => span,
+            None => return,
+        };
+
         span.set_attribute(Key::new("span.type").string("web"));
         span.set_attribute(Key::new("http.url").string("http://localhost:8080/foo"));
         span.set_attribute(Key::new("http.method").string("GET"));
--- a/examples/external-otlp-tonic-tokio/src/main.rs
+++ b/examples/external-otlp-tonic-tokio/src/main.rs
@@ -85,7 +85,11 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     let tracer = tracer("ex.com/basic");
 
     tracer.in_span("operation", |cx| {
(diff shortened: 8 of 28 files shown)


djc:
I'm happy to look at the breaking tests here, but would like some consensus first that this is a good direction.

shaun-cox:
I like the direction! I'll add your remote and run some benchmarks...

jtescher:
Hm I agree that this is more idiomatic but think it needs to return something to satisfy [this requirement of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk).

> If the parent Context contains no Span, an empty non-recording Span MUST be returned instead (i.e., having a SpanContext with all-zero Span and Trace IDs, empty Tracestate, and unsampled TraceFlags).

Could possibly have another method that has this behavior to work around the spec constraint? but the confusion may make it net negative from a DX perspective.

hdost:
Maybe the trade off could be `span()` returns as it does today, but we have `try_span()` though I'm not honestly sure what value it provides 🤔 . 
Maybe a default Span is the invalid span as @jtescher listed in the spec?

davidbarsky:
(Might as well add some feedback)

If I'm reading that spec correctly, it seems to be designed around languages with pervasive nullability, where retuning a null value could take down the process. @cijothomas mentioned recently that the specification is generally flexible, and .NET has taken some liberties with their official implementation to adhere more closely to .NET's idioms. I don't see why Rust shouldn't follow that precedent, especially if the divergence from the specification is an improvement in UX. Put differently, I think adhering to the _spirit_ of the the specification rather than letter is warranted.

djc:
> If I'm reading that spec correctly, it seems to be designed around languages with pervasive nullability, where retuning a null value could take down the process. @cijothomas mentioned recently that the specification is generally flexible, and .NET has taken some liberties with their official implementation to adhere more closely to .NET's idioms. I don't see why Rust shouldn't follow that precedent, especially if the divergence from the specification is an improvement in UX. Put differently, I think adhering to the _spirit_ of the the specification rather than letter is warranted.

I was reading some part of the OpenTelemetry spec/documentation earlier that mentioned something like implementations should focus on the conceptual capabilities more than the exact API prescriptions, but of course I can't find that bit now. I do feel like the way this is written doesn't really make sense for Rust -- and I feel like providing an idiomatic API does make things a lot easier.

If you look at the code changes here, I think a lot of changes make the code simpler to follow, and there's potentially other simplifications (like dropping the `Option` inside `SynchronizedSpan`, or some validity checks that are basically trying to recover the vailidity of the `Span` after checking whether `.span()` is `Some`).

TommyCpp:
I feel like the reason why spec asks for a noop spans is to make sure the propgation works in absence of a SDK. So as long as we can make sure the following is true, we should return a `Option` from`span()` as it's more friendly for Rust users

> This means that a SpanContext that has been provided by a configured Propagator will be propagated through to any child span and ultimately also Inject, but that no new SpanContexts will be created.

cijothomas:
> I was reading some part of the OpenTelemetry spec/documentation earlier that mentioned something like implementations should focus on the conceptual capabilities more than the exact API prescriptions, but of course I can't find that bit now

It's right here (and in every language's contributing guide.): https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md#focus-on-capabilities-not-structure-compliance

djc:
@jtescher any further thoughts on this?

jtescher:
Hm so I think the spec provides default values in APIs like this mostly for convenience. E.g. as a user of this crate you don't need to check for `None` cases.

```rust
// can always call methods, values ignored by non recording spans
cx.span().add_event(my_event)

// would have to explicitly handle the none case, with map or if let some..
cx.span().map(|span| span.add_event(my_event))
```

A similar choice is made in `tracing`'s [Span::current](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#method.current) which always returns a `Span`, with a disabled one instead of `None`.

I'm not opposed to deviating from the spec, especially where it makes things more erconomic or idiomatic, I'm just not sure that is the case here. Are there other crates or otel language impls that return `None` in this case?

shaun-cox:
> ```rust
> // can always call methods, values ignored by non recording spans
> cx.span().add_event(my_event)
> 
> // would have to explicitly handle the none case, with map or if let some..
> cx.span().map(|span| span.add_event(my_event))
> ```

My concern for the first case above, and the current direction of returning no-op spans, is the performance implications. While it's nice that once can always call all span methods and the values, are ignored, it's definitely not nice that one still has to do all the work to create those ignored values in the first place. (And they are not exactly "cheap" values... one has to give up ownership of a `Vec<KeyValue>` to `add_event`.)

In my ideal world, I *want* a lazily evaluated lambda to be passed which *only* gets called back when the all the data it's going to produce is actually going to be consumed for a benefit and not just immediately dropped.

If one wants to add multiple attributes to a span (only if its recording) the current api would have us do:
```rust
cx.span().add_attribute(KeyValue::new(...));
cx.span().add_attribute(KeyValue::new(...));
...
```
Which is a real head-scratcher for me to understand in the context of the goals of instrumenting to be as close to zero-overhead as possible for the case when no sdk is being used or when 99.9% of the spans are not being sampled.

So, I strongly prefer the `Option<Span>` approach.

Another alternative is a `visit_span` method on `TraceContextExt` which takes an `Fn<&mut Span>` argument or something and only calls the callback if there is an active span...?

jtescher:
> My concern for the first case above, and the current direction of returning no-op spans, is the performance implications. While it's nice that once can always call all span methods and the values, are ignored, it's definitely not nice that one still has to do all the work to create those ignored values in the first place. (And they are not exactly "cheap" values... one has to give up ownership of a Vec<KeyValue> to add_event.)

This would be an argument for changing all of the APIs that return spans (and likely meter instruments too) and removing the noop spec impl. E.g. creating a new span would need to also return an option as the span may not be recording. Could consider doing that, but it would be a larger scope.

> If one wants to add multiple attributes to a span (only if its recording) the current api would have us do:

The spec would have you do a check to `Span::is_recording`

```rust
let span = cx.span();
if span.is_recording() {
    let attrs = expensive_call(); 
    span.set_attributes(attrs);
}
```
> Another alternative is a visit_span method on TraceContextExt which takes an Fn<&mut Span> argument or something and only calls the callback if there is an active span...?

I think that would likely be the most compliant way to accomplish what you're looking for as you'd want to both check that there is an active span _and_ that the span is recording. Would look similar to the check above with a closure vs the current `if` statement to make it also work on tracer created spans and could have a similar method on the context ext if necessary.

```rust
let span = cx.span();
span.visit_recording(|span| {
    let attrs = expensive_call(); 
    span.set_attributes(attrs);
});
```

djc:
I think where possible, having a simple `Option`-based API is more ergonomic, and more clear, than relying on closures that are called conditionally.

jtescher:
> I think where possible, having a simple Option-based API is more ergonomic, and more clear, than relying on closures that are called conditionally.

Agreed, but would still potentially need a way of distinguishing the concepts of an active vs recording. You can have all 3 possibilities: no span, active with recording, active without recording. So either this is returning `None` to only indicate that there is no active span, in which case the API becomes something along the lines of:

```rust
if let Some(recording_span) = cx.span().filter(|span| span.is_recording()) {
    let attrs = expensive_call(); 
    span.set_attributes(attrs);
}
```

Or you need a different way of propagating the current trace id as this method would no longer work:

```rust
// default instead of existing trace if span is not recording
let span_context = cx.span().map(|span| span.span_context()).unwrap_or_default() 
```

The current API requires the explicit check to `Span::is_recording` but that may still have a more consistent DX with fewer footguns.

djc:
Okay, fair point. So maybe we should yield a trivariant enum?

```rust
enum SpanState<'a> {
    Recording(&'a mut Span),
    Active(&'a Span),
    None,
}
```

jtescher:
> Okay, fair point. So maybe we should yield a trivariant enum?

I think that would be the clearest way to represent it in the type system. But do we want end users of this crate to have to handle all types at each callsite?

Still seems like the current API has the simplest mental model exposed that allows for optional perf optimizations through checking span recording state.

djc:
> I think that would be the clearest way to represent it in the type system. But do we want end users of this crate to have to handle all types at each callsite?
> 
> Still seems like the current API has the simplest mental model exposed that allows for optional perf optimizations through checking span recording state.

I mean, we can then add some helper methods on top of that type that make it easy to adapt to the caller's needs? Like `fn active(&self) -> Option<&Span>` and `fn recording(&mut self) -> Option<&mut self>`.

I think Rust APIs usually optimize not for simplicity but for accurately modeling the underlying state. Yes, this can make things a bit more verbose, but also less error-prone and often more efficient.

shaun-cox:
In the interest of moving this forward towards a decision, here's my opinion on an improvement.

1. change `TraceContextExt` to have `span()` return `Option<SpanRef<'_>>`
2. remove `TraceContextExt` `has_active_span()` function, and just have callers check `span().is_some()` instead.

To handle further checking if the span is recording, encourage:
```rust
    let span = cx.span().filter(is_recording);
```

I feel this is the idiomatic middle-ground that is still simple to use.

djc:
@shaun-cox just to clarify, how do you see that being different from the current changes in this PR? (I think it's pretty close?)

shaun-cox:
> @shaun-cox just to clarify, how do you see that being different from the current changes in this PR? (I think it's pretty close?)

@djc, yep, no different than this PR.  Just trying to vote for banking this progress rather than considering a bigger trivariant approach right now.  Thanks!

jtescher:
I personally still think the existing APIs are a better choice as I believe others will find them simpler to understand and use, as well as having the same performance optimizations available that an `Option` or other type-level approach would have, without the downsides of being less idiomatic in terms of otel spec compliance and symmetry with other otel language implementations, and not matching existing rust ecosystem implementations like `tracing`, all of which return spans from these APIs rather than another type as far as I can see.

That being said if the other @open-telemetry/rust-approvers feel that this is the best way forward I won't block it. But if an alternate API is being considered it should be consistent with the `Tracer` and `Meter` traits or have similar capabilities as it would be extra confusing if they behaved differently.

shaun-cox:
> I believe others will find them simpler to understand and use, as well as having the same performance optimizations available that an `Option` or other type-level approach would have

@jtescher , I'm really trying to see where the same performance optimizations would be available with the current API.  Could you expand a bit on it?  I'd characterize a lot of the perf wins I've recently been involved with as eliminating expensive `.clone()` on these no-op instances that otherwise wouldn't be there if those instances were held by an `Option` that was `None` instead. I admit to being biased in favor of `Option`, and it may be clouding my view into other optimizations that may be possible without it...

jtescher:
> @jtescher , I'm really trying to see where the same performance optimizations would be available with the current API. Could you expand a bit on it? I'd characterize a lot of the perf wins I've recently been involved with as eliminating expensive `.clone()` on these no-op instances that otherwise wouldn't be there if those instances were held by an `Option` that was `None` instead. I admit to being biased in favor of `Option`, and it may be clouding my view into other optimizations that may be possible without it...

Which expensive clones on the no ops are you referring to? I was responding to the perf issues you were bringing up in https://github.com/open-telemetry/opentelemetry-rust/pull/1089#issuecomment-1584990205, e.g. being able to conditionally record fields if there is overhead when not recording.

shaun-cox:
I was referring to https://github.com/open-telemetry/opentelemetry-rust/pull/1140, where avoiding cloning `Context` in order to invoke `has_active_span()` was a performance win, and how that is the same pattern here... i.e. avoid having `span()` return/clone a no-op `Span` for further manipulation. 

So instead of this:
``` rust
let span = cx.span();
if span.is_recording() {
    let attrs = expensive_call(); 
    span.set_attributes(attrs);
}
```

We could do something like this:
``` rust
cx.span()
    .filter(is_recording)
    .map(|span| {
        let attrs = expensive_call(); 
        span.set_attributes(attrs);
    })
});
```
Now there is no no-op `Span` initialized and dropped... it's just getting an optional `SpanRef` and early exiting as it proceeds down the pipeline.

Putting it together with previous optimization to avoid obtaining a cloned `Context` too:
``` rust
Context::map_current(TraceContextEx::span)
    .filter(is_recording)
    .map(|span| {
        let attrs = expensive_call(); 
        span.set_attributes(attrs);
    })
});
```

jtescher:
Hm not seeing any overhead of dropping the noop span when running benches vs an option.

<details>
<summary>benchmarks</summary>

```rust
group.bench_function(BenchmarkId::new("noop", p), |b| {
    b.iter(|| {
        black_box(Context::map_current(|cx| {
            let span = cx.span();
            drop(span)
        }))
    })
});

group.bench_function(BenchmarkId::new("option", p), |b| {
    b.iter(|| {
        black_box(Context::map_current(|cx| {
            if let Some(span) = cx.span_opt() {
                drop(span)
            }
        }))
    })
});
```

They look effectively the same in terms of active or not active perf

```
context/noop/no-active-span
                        time:   [3.9029 ns 3.9201 ns 3.9380 ns]
                        change: [-0.8448% +0.1502% +1.1875%] (p = 0.79 > 0.05)
  4 (4.00%) high mild
  2 (2.00%) high severe
context/option/no-active-span
                        time:   [3.8490 ns 3.8651 ns 3.8833 ns]
                        change: [-0.6115% +0.0975% +0.8096%] (p = 0.79 > 0.05)
  1 (1.00%) high mild

context/noop/with-active-span
                        time:   [6.8758 ns 6.9062 ns 6.9364 ns]
                        change: [-1.2357% -0.6886% -0.1323%] (p = 0.02 < 0.05)
  6 (6.00%) high mild
context/option/with-active-span
                        time:   [6.9951 ns 7.0255 ns 7.0601 ns]
                        change: [-1.8125% -0.8761% +0.0237%] (p = 0.06 > 0.05)
```

Using this commit for diff https://github.com/jtescher/opentelemetry-rust-1/commit/beabfcd0847b6fa3046ff7ca3230984dc7729fb1

</details>

Possibly something else was the source of the perf differences in https://github.com/open-telemetry/opentelemetry-rust/pull/1140?

shaun-cox:
> Possibly something else was the source of the perf differences in https://github.com/open-telemetry/opentelemetry-rust/pull/1140?

Sorry, I didn't mean to confuse that it was just no-op Span clone elimination that would explain all of the performance gains, just that an example of using Option-like optimizations ala #1140 rather than returning cloned temporaries did result in performance gains... similar to how this PR has also measured performance gains (up to 8% in some scenarios glancing at the top of this thread) by having span() return an Option rather than a no-op instance.

My original question a few replies up was trying to gain insight into how to achieve these performance optimizations without using Option, as you alluded that you think they are still possible?

jtescher:
I can't reproduce the 8% number above. When I compare the start-end-span benchmarks against main I actually see worse performance on this branch, so something else may have changed in the mean time.

hdost:
Perhaps we could talk about this at the next SIG if @jtescher can make it?

cijothomas:
Closing old, inactive PRs. Feel free to reopen when ready. Thanks.

PR #1091, opened 2023-06-04
Improve span creation performance by 10-20%

shaun-cox:
It's cheaper to generate random u128 and u64 than it is to generate [u8; 16] and [u8; 8].

Changing `TraceId::From` and `SpanId::From` to be from these primitive types, rather than arrays, improves `start-end-span` benchmark performance by 10% in the sampling case and 21% in the non-sampling case.

```text
start-end-span/always-sample
                        time:   [517.71 ns 535.68 ns 551.39 ns]
                        change: [-14.991% -10.282% -5.3887%] (p = 0.00 < 0.05)
                        Performance has improved.
start-end-span/never-sample
                        time:   [133.18 ns 133.46 ns 133.76 ns]
                        change: [-22.160% -21.965% -21.769%] (p = 0.00 < 0.05)
                        Performance has improved.
```

## Changes

- Change TraceId::From and SpanId::From to be from primitive types, rather than arrays with an assumed byte ordering.
- DatadogPropagator had to change too, but it will be faster as it's no longer converting to big-endian bytes just to call TraceId/SpanId::From.

### DIFF
--- a/opentelemetry-api/src/trace/span_context.rs
+++ b/opentelemetry-api/src/trace/span_context.rs
@@ -120,9 +120,9 @@ impl TraceId {
     }
 }
 
-impl From<[u8; 16]> for TraceId {
-    fn from(bytes: [u8; 16]) -> Self {
-        TraceId::from_bytes(bytes)
+impl From<u128> for TraceId {
+    fn from(value: u128) -> Self {
+        TraceId(value)
     }
 }
 
@@ -181,9 +181,9 @@ impl SpanId {
     }
 }
 
-impl From<[u8; 8]> for SpanId {
-    fn from(bytes: [u8; 8]) -> Self {
-        SpanId::from_bytes(bytes)
+impl From<u64> for SpanId {
+    fn from(value: u64) -> Self {
+        SpanId(value)
     }
 }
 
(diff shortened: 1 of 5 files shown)


opentelemetry-api/src/trace/span_context.rs
```diff
@@ -120,9 +120,9 @@ impl TraceId {
     }
 }
 
-impl From<[u8; 16]> for TraceId {
```
djc, line 123: Might make sense to keep the old `From` impl and introduce the new ones as an addition?

djc (approved):
Nice, that makes a lot of sense!

opentelemetry-api/src/trace/span_context.rs
shaun-cox: I don't think it makes a lot of sense to keep the old `From<[u8; 16]>` for at least three reasons:
1. there is an unspecified ambiguity about the byte-ordering of those bytes
2. there are already explicit public methods (`from_bytes` and `to_bytes`) for handling these conversions
3. primitive types like `u64` also have `from_xx_bytes` method, but do not implement `From<[u8; 8]>` presumably due to 1 above

Regarding 2 above, I'd think we should call those methods `from_be_bytes` and `to_be_bytes` to be consistent with std and give the caller a better hint that byte-order is big endian.

djc: Fair enough!

shaun-cox:
I just rebased on main, which has the fix for the clippy issue.

cijothomas (approved):
LGTM!

PR #1094, opened 2023-06-06
chore: clean up examples

TommyCpp:
## Changes

We have too many examples on top level and that's making it harder to maintain. 

This PR split the examples that belong to the specific crates into their crates. It also removed unnecessary cargo projects for simple examples. As a general rule, examples that don't requires additional dependencies will now become a single file instead of a cargo project


.github/codecov.yaml
```diff
@@ -12,9 +12,19 @@ coverage:
         threshold: 0.5%
 
 ignore:
-  - "examples"
   - "opentelemetry/src/testing" # test harnesses
   - "opentelemetry-jaeger/src/testing" # test harness
   - "opentelemetry-jaeger/src/exporter/thrift" # auto generated files
   - "opentelemetry-otlp/src/proto" # auto generated files
   - "opentelemetry-proto/src/proto" # auto generated files
+  # examples below
```
cijothomas, line 20: :100:

TommyCpp:
> although I'm not convinced that just moving the examples around will substantially reduce maintenance.

I think it give us clearity on how many examples each integration has and how many examples belong to no specific integration. It might be a longer discussion but I can delete the one I think that's not needed in this PR, or in a following PR if ppl prefer

jtescher:
I agree that the total number of examples is probably not worth the maintenance cost currently, @TommyCpp could be good to prune some of the redundant examples in this PR 👍

opentelemetry-dynatrace/examples/dynatrace/Cargo.toml
```diff
@@ -9,7 +9,7 @@ futures = "0.3"
 http = "0.2"
 lazy_static = "1.4"
 opentelemetry = { version = "0.19", default-features = false, features = ["trace"] }
-opentelemetry-dynatrace = { path = "../../opentelemetry-dynatrace" }
+opentelemetry-dynatrace = { path = "../.." }
```
cijothomas, line 12: While at this, could you also modify the opentelemetry dependency to be the latest (path) instead of pinned to "0.19".

jtescher: I don't think dynatrace has anyone maintaining it currently. We could consider dropping it.

cijothomas: That makes sense. From their docs, it looks like OTLP is natively supported, so a custom exporter is not needed to be maintained.
https://opentelemetry.io/ecosystem/vendors/

I'll just open an issue to see if any objections.

TommyCpp:
The examples we currently have:

Example|Use case|Deleted?
---|---|---
actix-http| jaeger example with OTLP exporter | Yes, duplication with OTLP examples
actix-http-tracing| tracing examples with actix | Yes, duplicate with tracing-grpc 
actix-hyper| jaeger example with http collector and batch span processor / exporter | Yes, duplication with OTLP examples
actix-udp| jaeger example with jaeger agent running on UDP protocol | No, renamed and move to jaeger create as it's unique to jaeger and requires an agent to demostate 
async| jaeger example that shows how to instrument async code | Yes, easy enough to demonstrate in documentation
aws-xray| aws propagator examples | Yes, easy enough to demonstrate in documentation
basic-otlp| OTLP examples using tokio and tonic | No
basic-otlp-http| OTLP example using HTTP | No
datadog| datadog basic examples | No
dynatrace| dynatract example | Yes as we are deprecating the create
external-otlp-grpcio-async-std| OTLP examples using grpcio and async-std | No, I think as OTLP is the main exporter we provide we should probably provide more examples to promote adaption.
external-otlp-tonic-tokio| OTLP examples using tokio and tonic | Yes, duplication with other OTLP examples
grpc| examples on passing information using context propagation | No
http| example using hyper to show how to instructment client and server | Yes, should be covered in other examples like `traceresponse`
hyper-prometheus| prometheus example using hyper | No
jaeger-remote-sampler| example that shows how to setup the remote sampler | No, it requires a docker setup to demonstrate
logs-basic| basic example for logs | No
metrics-basic| basic example for metrics | No
multiple-span-processors| how to setup multiple span processor for one tracer provider | Yes, this is unusual and probably not needed for most use case
traceresponse| client - server example that includes instrument spans and context propagation | No
tracing-grpc| example with `tracing-opentelemetry`
zipkin| zipkin examples | No
zpages| zpage example | No

hdost (approved):
Yes 🙌 love this. Keep the examples as close to the regular code as possible.

hdost:
The only one I'm partially wondering is would it be possible to include a multiple span processors example in in docs, rather than a full blown example.

djc (approved):
Sounds great!

lzchen (approved):

TommyCpp:
> The only one I'm partially wondering is would it be possible to include a multiple span processors example in in docs, rather than a full blown example.

Yeah I can look into it in follow up PRs

PR #1102, opened 2023-06-08
Update main readme with logging status

cijothomas:
## Changes

1. Add project status right in this repo so it is easy to maintain and update. (There was some [erroneous update](https://github.com/open-telemetry/opentelemetry.io/pull/2416) to the status earlier by mistake).

2. Added logging status
    a. Clarify that OTel Rust is not building its own new logging api (similar to what Otel C++ did)
    b. Provide a recommendation to use tokio/tracing for logging needs.

3. Updated link to versioning doc from this repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,14 +51,16 @@ available:
 * opentelemetry-appender-tracing // TODO: Add link once available
 * opentelemetry-appender-slog // TODO: Add link once available
 
-If you already use the logging APIs from above, continue to use them, and use the appenders above to bridge the logs to OpenTelemetry. If you are using a library not listed here, feel free to
-contribute a new appender for the same.
+If you already use the logging APIs from above, continue to use them, and use
+the appenders above to bridge the logs to OpenTelemetry. If you are using a
+library not listed here, feel free to contribute a new appender for the same.
 
 If you are starting fresh, then consider using
 [tracing](https://github.com/tokio-rs/tracing) as your logging API. It supports
 structured logging and is actively maintained.
 
-Project versioning information and stability guarantees can be found [here](VERSIONING.md).
+Project versioning information and stability guarantees can be found
+[here](VERSIONING.md).
 
 ## Getting Started
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,31 @@ observability tools.
 
 ## Project Status
 
-[OpenTelemetry Signal Status for Rust][otel-signals-rust]. Libraries adhere to [Stability Guarantees.][otel-signal-stability]
+| Signal  | Status     |
+| ------- | ---------- |
+| Logs    | Alpha*     |
+| Metrics | Alpha      |
+| Traces  | Beta       |
+
+*OpenTelemetry Rust is not introducing a new end user callable Logging API.
+Instead, it provides [Logs Bridge
+API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md),
+that allows one to write log appenders that can bridge existing logging
+libraries to the OpenTelemetry log data model. The following log appenders are
+available:
+
+* [opentelemetry-appender-log](opentelemetry-appender-log/README.md)
+* opentelemetry-appender-tracing // TODO: Add link once available
+* opentelemetry-appender-slog // TODO: Add link once available
+
+If you already use the logging APIs from above, continue to use them, and use the appenders above to bridge the logs to OpenTelemetry. If you are using a library not listed here, feel free to
+contribute a new appender for the same.
+
+If you are starting fresh, then consider using
+[tracing](https://github.com/tokio-rs/tracing) as your logging API. It supports
+structured logging and is actively maintained.
+
+Project versioning information and stability guarantees can be found [here](VERSIONING.md).
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,7 +34,33 @@ observability tools.
...
+
+* [opentelemetry-appender-log](opentelemetry-appender-log/README.md)
+* opentelemetry-appender-tracing // TODO: Add link once available
+* opentelemetry-appender-slog // TODO: Add link once available
+
+If you already use the logging APIs from above, continue to use them, and use
+the appenders above to bridge the logs to OpenTelemetry. If you are using a
+library not listed here, feel free to contribute a new appender for the same.
+
+If you are starting fresh, then consider using
```
cijothomas, line 58: I initially thought of using "we strongly recommend using tracing", but it was pointed out in the SIG meeting that, it may not be desirable to strongly recommend anything. Having no recommendation from OTel Rust also feels not so good, as I believe this community should recommended something for new projects trying to find a logging API. (Of course, end users are free to ignore this and pick anything and use the existing/new bridges, given the log bridge is totally neutral)
Ended up with this wording; happy to adjust based on feedbacks.

hdost: I think until things shake out more it's better to provide options than recommendations. Everything has a trade off, so potentially if we add recommendations they should have included caveats.

djc: @hdost I disagree. IMO it's helpful for people to the ecosystem to provide guidance as to what we think is the best-supported route out there. Of course if there are known caveats to that choice we can briefly mention those, too. We are the experts here, and avoiding giving advice to people who will likely have less expertise because the topic is too complex doesn't seem like a good way forward.

djc (approved):

README.md
cijothomas: I am re-thinking about the wording here, to make a "strong recommendation to use tracing".

1. The built-in log crate does not support structured-logging, so we can't really recommend it for new users. (The structured logging is there but as experimental/nightly feature).
2. The next popular one is slog, but their repo already [recommends ](https://github.com/slog-rs/slog#you-might-consider-using-tracing-instead) folks to consider tracing.
3. This repo already has [special treatment](https://github.com/open-telemetry/opentelemetry-rust/blob/06d17f76bd682b976a02dd109d01eb25f7c6c564/opentelemetry-sdk/src/trace/tracer.rs#L172) for tracing-opentelemetry, no reason to make it more formal by recommending tracing as the logging api.

Tagging @jtescher and @davidbarsky, maintainers of tracing, who are also active in this repo to hear more thoughts!

TommyCpp: I feel like we can just recommend build in log for simple use case and tracing for any sofisicated use case. slog seems sitting in between but since they are recommending tracing we don't need to recommend them here.

TommyCpp: Also we mentioned that if none of those popular one works you can always build your own bridge and enjoy the rest of the benifit of OTel

cijothomas: > I feel like we can just recommend build in log for simple use case and tracing for any sofisicated use case. slog seems sitting in between but since they are recommending tracing we don't need to recommend them here.

recommending log is not really needed in my view. It might encourage folks to do unstructured logging. I think it is the responsibility of OpenTelemetry group to encourage structured logging. But not a strong opinion, so I can add the case for log too here.

hdost: @djc fair enough, it's not as if it's set in stone, it's more for me about the "strong recommendation", for me it gives the reader a certain level of comfort but we should be cautious for the same reason. If we're providing a recommendation, we should explain why. 

> I am re-thinking about the wording here, to make a "strong recommendation to use tracing".

@cijothomas Indeed if we're going to provide recommendations, I'd agree it makes more sense to be a bit more nuanced.

lzchen (approved):

hdost (approved):
Given our conversation not sure there need to be changes right now.

README.md
davidbarsky: > Tagging [jtescher] and @davidbarsky, maintainers of tracing, who are also active in this repo to hear more thoughts!

Sorry, I missed this ping. That's my bad. My biased opinion is, obviously, "use `tracing`". To expand on my feeling:
- recommendations from opentelemetry can change over time, so I don't view the recommendation isn't a long-term commitment.
- For folks newer to Rust, using `tracing` will set them up for success with regard to structured logging _and_ distributed tracing (note that I think that `log`'s structured K/V values are an improvement, but them being off by default is unfortunate, but understandable, choice. Note that `tracing` can't support really support log's K/V values properly without us enabling [`valuable`](https://docs.rs/valuable/latest/valuable/) support, which... we probably should at some point.
- `tracing` also has a pretty large community of integrations (e.g., things like Perfetto, flamegraphs, Sentry, etc.). This results, I think, in a really solid developer experience that `log`—by design—doesn't try to replicate.
- I think the simplicity aspect of `log` is a compelling one. If people don't really want to think about logging, `log` + `env_logger` is a very good choice, but I think `tracing` gets pretty close with the "don't think about it" experience with `tracing_subscriber` (see, for instance, https://github.com/tokio-rs/tracing/tree/v0.1.x#in-applications). I personally feel—and seen from other users at my employer—that the benefits from using `tracing` instead of `log` occur sooner than most people might otherwise think.
- `log` can integrate with non-Rust codebases, such as C or C++, better than `tracing` can. At the moment, `tracing` can't really do that right now because it assumes a lot of the core values inside of `tracing` are `'static`, and logging libraries like `glog` do not, which can lead to dangling pointers in Rust if not careful (`tracing` _can_ integrate with non-Rust logging libraries via `log`, but this is something that we'd like to improve on in the future).

PR #1110, opened 2023-06-14
Otel user_events Log exporter.

### LINKED ISSUE #1103: Otel user_events logs exporter
[user_events](https://docs.kernel.org/trace/user_events.html) is Linux solution for user process tracing, similar to [ETW](https://learn.microsoft.com/en-us/windows/win32/etw/about-event-tracing) (Event Tracing for Windows) on Windows. It builds on top of the Linux Tracepoints, and so allows user processes to create events and trace data that can be viewed via existing tools like ftrace and perf.

This kernel feature is supported starting Linux kernel 5.18 onwards, and it enables
 - A faster path for tracing from user mode application by utilizing kernel mode memory address space. 
 - User processes can now export telemetry events only when it is useful i.e, when the registered set of tracepoint events are enabled in kernel.

 The user_events exporter will enable applications to use OpenTelemetry API to capture the telemetry events, and write to user_events subsystem. From user_events, the events can be
  - Captured by the agents running locally, and listening for specific events within user_events subsystem.
  - Or monitored in real-time using local Linux tool like [perf](https://perf.wiki.kernel.org/index.php/Main_Page) or ftrace.

The exporter could be useful for 
 - system applications
 - applications with strict performance requirements.

These type of applications are typically written in C/C++ and Rust.

OpenTelemetry C++ already provides user_events exporter in the contrib repo - https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/exporters/user_events. For Rust, this can be initially added to main repo, and later moved once contrib repo is available.

Reference:
[user_events](https://docs.kernel.org/trace/user_events.html) - User-based Event Tracing
[Example](https://github.com/torvalds/linux/tree/master/samples/user_events) - Basic example from Linux kernel repo
[eventheader_dynamic](https://crates.io/crates/eventheader_dynamic) - Rust crate to write to user_events:

``
eventheader_dynamic provides a mid-level API for generating runtime-specified events using the EventHeader convention. The events are written using the user_events system. This is intended for use as an implementation layer for a higher-level API like OpenTelemetry.
``

lalitb:
Fixes Towards #1103 

## Changes

The initial prototype for the Otel user_events exporter.
This is the stripped down version taken from - https://github.com/Robo210/opentelemetry-etw-tracelogging/tree/log/ (credit to @robo210) - removing ETW exporter, and trace exporter and adding serialization format for logs.

The serialization format is similar to that for Otel C++ - https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/exporters/user_events

Open Issue:
 - The configuration for user_events name ( i.e., provider group and provider name). As of now, it is configured as part of UserEventsExporter constructor, but can change in future.
 - The serialization format is simpler enough to use for generic purpose, but may need changes if required as separate PR.


opentelemetry-userevents-exporter/examples/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+    let exporter_config = ExporterConfig{keyword : 1};
+    let prov_group = Some("prov_group");
+    let realtime_processor = RealTimeLogProcessor::new("test12345", None , exporter_config);
+    LoggerProvider::builder()
+    .with_log_processor(realtime_processor).build()
+}
+
+fn main() {
+    let logger_provider = init_logger();
+    let prov_group = Some("test1234");
```
cijothomas, line 34: prov_group is unused here

opentelemetry-userevents-exporter/examples/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+
+use opentelemetry_userevents_exporter_log::{ExporterConfig, RealTimeLogProcessor, ProviderGroup};
+use std::time::SystemTime;
+use std::thread;
+
+fn init_logger() -> LoggerProvider {
+    let exporter_config = ExporterConfig{keyword : 1};
+    let prov_group = Some("prov_group");
+    let realtime_processor = RealTimeLogProcessor::new("test12345", None , exporter_config);
```
cijothomas, line 27: its not obvious from here that the processor creates a exporter inside it. Perhaps its best to make it explicit by creating the exporter right here, and feed it to the processor.

opentelemetry-userevents-exporter/examples/basic.rs
```diff
@@ -0,0 +1,55 @@
...
+    LoggerProvider::builder()
+    .with_log_processor(realtime_processor).build()
+}
+
+fn main() {
+    let logger_provider = init_logger();
+    let prov_group = Some("test1234");
+    let logger: opentelemetry_sdk::logs::Logger = logger_provider.logger("test");
```
cijothomas, line 35: lets use `log` crate to do the logging, so this example can use the log-appender. There is pretty good chance this example will misled users to believe *this is the way* to log in OTel Rust!

opentelemetry-userevents-exporter/src/logs/realtime_logprocessor.rs
```diff
@@ -0,0 +1,45 @@
...
+use opentelemetry_api::{InstrumentationLibrary, logs::LogRecord, logs::LogResult};
+use opentelemetry_sdk::{export::logs::LogExporter, export::logs::LogData, export::logs::ExportResult};
+
+use crate::logs::exporter::*;
+use crate::logs::exporter::ExporterConfig;
+
+use std::cell::Cell; 
+
+#[derive(Debug)]
+pub struct RealTimeLogProcessor {
```
cijothomas, line 15: maybe a better name ? I can think of `ReentrantExportProcessor` or `SynchronousExportProcessor`.
Also, given this is a public processor, lets document the need for this (or in other words, add a comment why the simple/batch built-ins are not sufficient).

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Internal/ReentrantExportProcessor.cs#L23-L28 some wordings could be copied from here.

opentelemetry-userevents-exporter/src/logs/realtime_logprocessor.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+impl opentelemetry_sdk::logs::LogProcessor for RealTimeLogProcessor {
+
+    fn emit(&self, data: LogData) {
+        self.event_exporter.export_log_data(&data);
+    }
+
+    fn force_flush(&self) -> LogResult<()> {
```
cijothomas, line 38: nit: add comment "Force_flush is no-op, as this processor does not keep anything in memory to be flushed out" /similar

opentelemetry-userevents-exporter/src/logs/exporter.rs
```diff
@@ -0,0 +1,252 @@
...
+impl UserEventsExporter {
+    pub(crate) fn new(
+        provider_name: &str,
+        provider_group: ProviderGroup,
+        exporter_config: ExporterConfig,
+    ) -> Self
+    {
+        let mut options = eventheader_dynamic::Provider::new_options();
+        options = *options.group_name(provider_name);
+        let mut provider: eventheader_dynamic::Provider = eventheader_dynamic::Provider::new(provider_name, &options);
```
cijothomas, line 54:
```suggestion
        let mut eventheader_provider: eventheader_dynamic::Provider = eventheader_dynamic::Provider::new(provider_name, &options);
```

cijothomas: nit to help to avoid confusing "provider", as Otel also has provider concept

opentelemetry-userevents-exporter/src/logs/exporter.rs
```diff
@@ -0,0 +1,252 @@
...
+                    if k.as_str() == "event_name" {
+                        event_name = match v {
+                            AnyValue::String(value) => {event_count = event_count + 1; value.as_ref()}
+                            _ => ""
+                        }
+                    }
+                }
+                cs_b_count = cs_b_count + event_count;
+                if log_data.record.body.is_some() {
+                    cs_b_count = cs_b_count + 1;
```
cijothomas, line 189: nit: these bookkeeping of counts can be done at the moment we actually serialize body, to avoid checking same if condition more than once.

opentelemetry-userevents-exporter/src/logs/exporter.rs
```diff
@@ -0,0 +1,252 @@
...
+                         &chrono::DateTime::<chrono::Utc>::from(log_data.record.timestamp.unwrap()));
+
+                        eb.add_str("time", time, FieldFormat::Default, 0);
+                    }
+                }
+                // Get Event_Id and Event_Name if present.
+                let (mut event_id, mut event_name) = (0, "");
+                let mut event_count = 0;
+                for (k, v ) in log_data.record.attributes.as_ref().unwrap().into_iter() {
+                    if k.as_str()== "event_id" {
```
cijothomas, line 174: move the magicstrings to a const, so its easy to change later.

opentelemetry-userevents-exporter/src/logs/exporter.rs
lalitb: We need this count value before doing the serialization -to store it in the PartB struct. Another approach would be to have another bool flag to know if the body really exists. That would need multiple such flags for each of the fields. It's tradeoff between memory and performance :)

opentelemetry-userevents-exporter/examples/basic.rs
lalitb: Thanks removed.

opentelemetry-userevents-exporter/src/logs/realtime_logprocessor.rs
lalitb: Good point. Renamed to ReentrantExportProcessor, and added relevant comments.

opentelemetry-userevents-exporter/src/logs/realtime_logprocessor.rs
lalitb: Thanks done.

opentelemetry-userevents-exporter/src/logs/exporter.rs
lalitb: Good point. Done.

opentelemetry-userevents-exporter/src/logs/exporter.rs
lalitb: Done. Thanks.

opentelemetry-userevents-exporter/examples/basic.rs
lalitb: Good point, as of now I have added log-appender example. But it is very limited to export unstructured text. Have kept log-bridge example with disclaimer just as it helps to test the changes better. We can remove it once tracing subscriber example is added. But I don't have any strong opinion - we can remove it if it's really causing confusion.

opentelemetry-userevents-exporter/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "opentelemetry-userevents-exporter"
```
cijothomas, line 2: we need to finalize a name. The existing exporters like zipkin don't have the word "exporter" in it.! Maybe we should do a bulk renaming to all existing crates! 

Not a blocker!

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
```diff
@@ -0,0 +1,53 @@
...
+use opentelemetry_sdk::{export::logs::LogExporter, export::logs::LogData, export::logs::ExportResult};
+
+use crate::logs::exporter::*;
+use crate::logs::exporter::ExporterConfig;
+
+use std::cell::Cell; 
+
+// This export processor exports without synchronization.
+// This is currently only used in users_event export, where we know
+// that the underlying system is safe under concurrent calls
```
cijothomas, line 16:
```suggestion
// that the underlying exporter is safe under concurrent calls
```

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
```diff
@@ -0,0 +1,53 @@
...
+use opentelemetry_api::{InstrumentationLibrary, logs::LogRecord, logs::LogResult};
+use opentelemetry_sdk::{export::logs::LogExporter, export::logs::LogData, export::logs::ExportResult};
+
+use crate::logs::exporter::*;
+use crate::logs::exporter::ExporterConfig;
+
+use std::cell::Cell; 
+
+// This export processor exports without synchronization.
+// This is currently only used in users_event export, where we know
```
cijothomas, line 15:
```suggestion
// This is currently only used in users_event exporter, where we know
```

cijothomas (approved):
LGTM for the initial version! Of course, we can iterate on it to add more features/fixes/changes before shipping stable, especially the ones that were stripped out temporarily from the original implementation.

Cargo.toml
```diff
@@ -16,6 +16,7 @@ members = [
     "opentelemetry-semantic-conventions",
     "opentelemetry-stackdriver",
     "opentelemetry-stdout",
+    "opentelemetry-userevents-exporter",
```
lzchen, line 19: Do we eventually want one or multiple crates to contain exporters to user events? If multiple, maybe rename this to "opentelemetry-user-events-logs"?

opentelemetry-userevents-exporter/Cargo.toml
lzchen: Do we eventually want one or multiple crates to contain exporters to user events? If multiple, maybe rename this to "opentelemetry-user-events-logs"?

opentelemetry-userevents-exporter/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+homepage = "https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-userevents-exporter"
+repository = "https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-userevents-exporter"
+readme = "README.md"
+rust-version = "1.68.0"
+keywords = ["opentelemetry", "log", "trace", "user_events"]
+license = "Apache-2.0"
+
+[features]
+logs = ["dep:async-trait", "opentelemetry_sdk/logs"]
+default = ["logs"]
```
lzchen, line 15: Is there a point to having a features section if the default encapsulates all features?

opentelemetry-userevents-exporter/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Changelog
+
+## Unreleased
+
```
lzchen, line 4: Nit:

```
## v0.1.0

### Added

- Initial Alpha implementation
```

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl opentelemetry_sdk::logs::LogProcessor for ReenterantLogProcessor {
+
+    fn emit(&self, data: LogData) {
+        self.event_exporter.export_log_data(&data);
+    }
+
+    // This is a no-op as this processor doesn't keep anything 
+    // in memort to be flushed out.
```
lzchen, line 43:
```suggestion
    // in memory to be flushed out.
```

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
```diff
@@ -0,0 +1,53 @@
...
+// This export processor exports without synchronization.
+// This is currently only used in users_event export, where we know
+// that the underlying system is safe under concurrent calls
+
+#[derive(Debug)]
+pub struct ReenterantLogProcessor {
+    event_exporter: UserEventsExporter,
+}
+
+impl ReenterantLogProcessor{
```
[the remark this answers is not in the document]
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_processor.rs#L45
The SimpleLogProcessor waits for one export to finish before calling the next export. So if there are 8 threads, each calling Log, then the export of logs from each thread occurs one-after-another.... For achieving higher throughput, re-entrant processor, does not wait for an export to complete before calling the next. This requires Exporter to be written in re-entrant aware way, which is the case of user_event, etw exporters. The net effect is, all 8 logs gets exported simultaneously leading to much higher throughput. 

(There are several other reasons as well. I hope to get them all documented here by the time user_events is widely available and this crate is ready for production use)

There are existing implementations for this model in OTel C++ and .NET for ETW (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Internal/ReentrantExportProcessor.cs#L23) which are already production ready for years. Most of the user_events exporting is inspired from its windows counterpart :)

lzchen (approved):

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::logs::exporter::ExporterConfig;
+
+use std::cell::Cell; 
+
+// This export processor exports without synchronization.
+// This is currently only used in users_event export, where we know
+// that the underlying system is safe under concurrent calls
+
+#[derive(Debug)]
+pub struct ReenterantLogProcessor {
```
cijothomas, line 19:
```suggestion
pub struct ReentrantLogProcessor {
```

opentelemetry-userevents-exporter/Cargo.toml
jtescher: it would allow opting out via `default-features = false` https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature

The question I guess is does this crate still make sense without that feature enabled?

opentelemetry-userevents-exporter/src/logs/exporter.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+#[async_trait]
+impl opentelemetry_sdk::export::logs::LogExporter for UserEventsExporter {
+    async fn export(&mut self, batch: Vec<opentelemetry_sdk::export::logs::LogData>) -> opentelemetry_api::logs::LogResult<()> {
+        for log_data in batch {
+            let _ = self.export_log_data(&log_data);    
+        }
+        Ok(())
+    }
```
lzchen, line 287: Does `shutdown()` also need to be implemented?

opentelemetry-userevents-exporter/Cargo.toml
lalitb: Have renamed this to `opentelemetry-userevents-logs` for now. We can later decide if to merge.

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
lalitb: Thanks fixed.

opentelemetry-userevents-exporter/src/logs/reenterant_logprocessor.rs
lalitb: Fixed.

opentelemetry-userevents-exporter/src/logs/exporter.rs
lalitb: There is nothing to cleanup in shutdown. The deregister for user_events is automatically handled when EventHeader instance goes out of scope.

cijothomas (approved):

opentelemetry-userevents-exporter/Cargo.toml
djc: I think we definitely don't want to string together "userevents" in the name, should split those up as "user-events".

opentelemetry-userevents-exporter/src/logs/exporter.rs
lzchen: Interesting. I thought the compiler would complain if we didn't implement all the functions in an inherited trait explicitly.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/export/logs/mod.rs#L16 There is default implementation (empty block), so its optional to implement it.

opentelemetry-userevents-exporter/Cargo.toml
lalitb: Yes agree, have modified to `opentelemetry-user-events-logs`

opentelemetry-user-events-logs/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "opentelemetry-user_events-logs"
```
TommyCpp, line 2:
```suggestion
name = "opentelemetry-user-events-logs"
```

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -0,0 +1,349 @@
+#![allow(unused_imports, unused_mut, unused_variables, dead_code)]
```
TommyCpp, line 1: Could you create a tracking issue to resolve those warnings in a follow up PRs. I think we should clean those up before a new release

TommyCpp (approved):
Mostly looks good to me. But I think we should clean up any TBD and warnings before release.

TommyCpp:
If there is no other comments I will merge it tomorrow

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: Agree, created a tracking issue here : #1120

djc: Why not just clean them up before merging this? I don't see much value in postponing this work out of this PR.

lalitb: Sure, have fixed the warnings now. @TommyCpp -  fyi, if this is good to merge.

PR #1116, opened 2023-06-15
Update main readme with sig meeting info

cijothomas:
1. Putting SIG meeting info in the main readme itself for easier finding. (a practice common in many other otel repos, like Python, .NET).
2. Add the image showing all contributors (almost all OTel repos do this, so why not!)
3. Added a welcoming line about "meeting is open for all to join".


hdost (approved):
Sounds great!

lzchen (approved):

Issue #1114, opened 2023-06-15
Otel user_events metrics exporter


lzchen:
Similar to https://github.com/open-telemetry/opentelemetry-rust/issues/1103, but for sending metrics.

Metrics will be encoded using [protobuf](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/metrics/v1/metrics_service.proto#L36). An api is taken from [LinuxTracepoints](https://github.com/microsoft/LinuxTracepoints/tree/main/rust) to send data to user events. 

OT C++ has an equivalent implementation [here](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/291)

PR #1117, opened 2023-06-15
Otel user_events Metrics exporter.

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1114

## Changes

The initial prototype for the Otel user_events metrics exporter. For now, we are putting this in a separate crate from https://github.com/open-telemetry/opentelemetry-rust/pull/1110.

Serialization format is protobuf and uses the exact same logic for encoding as otlp exporter. Transport logic uses linux tracepoints (https://github.com/microsoft/LinuxTracepoints/tree/main/rust) instead of grpc.


opentelemetry-user-events-metrics/examples/basic.rs
```diff
@@ -0,0 +1,58 @@
...
+        )]))
+        .with_reader(reader)
+        .build()
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let exporter = opentelemetry_user_events_metrics::MetricsExporter::new();
+    let meter_provider  = init_metrics(exporter);
+    let cx = Context::new();
```
cijothomas, line 32: the PR removing cx just merged, so you can remove this.

opentelemetry-user-events-metrics/examples/basic.rs
```diff
@@ -0,0 +1,58 @@
...
+        .with_unit(Unit::new("test_unit"))
+        .init();
+
+    c.add(&cx, 1.0, [
+        KeyValue::new("mykey1", "myvalue1"),
+        KeyValue::new("mykey3", "myvalue3"),
+    ].as_ref());
+
+    meter_provider.shutdown()?;
+    // Call this explictly since provider shutdown is not propagated to exporter for now
```
cijothomas, line 52: link to gh issue.

opentelemetry-user-events-metrics/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+autotests = false
+
+[lib]
+doctest = false
+
+[dependencies]
+opentelemetry_api = { version = "0.19", path = "../opentelemetry-api", features = ["metrics"] }
+opentelemetry_sdk = { version = "0.19", path = "../opentelemetry-sdk", features = ["metrics", "rt-tokio"] }
+tokio = { version = "1.0", features = ["full"] }
+protobuf = { version = "2.18", optional = true }
```
cijothomas, line 26: is this optional?

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+
+impl TemporalitySelector for MetricsExporter {
+    fn temporality(&self, _kind: InstrumentKind) -> Temporality {
+        Temporality::Cumulative
+    }
+}
+
+impl AggregationSelector for MetricsExporter {
+    fn aggregation(&self, _kind: InstrumentKind) -> Aggregation {
+        Aggregation::Sum
```
cijothomas, line 47: please add TODO here as these should be not just SUM...

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+                println!("Tracepoint failed to register.");
+            }
+        }
+        MetricsExporter {  }
+    }
+}
+
+impl TemporalitySelector for MetricsExporter {
+    fn temporality(&self, _kind: InstrumentKind) -> Temporality {
+        Temporality::Cumulative
```
cijothomas, line 41: please add TODO here as these should be not just Cumulative

opentelemetry-user-events-metrics/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+rust-version = "1.60"
+autotests = false
+
+[lib]
+doctest = false
+
+[dependencies]
+opentelemetry_api = { version = "0.19", path = "../opentelemetry-api", features = ["metrics"] }
+opentelemetry_sdk = { version = "0.19", path = "../opentelemetry-sdk", features = ["metrics", "rt-tokio"] }
+tokio = { version = "1.0", features = ["full"] }
```
cijothomas, line 25: do we need tokio for this exporter?

opentelemetry-user-events-metrics/src/tracepoint/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+/// check [`enabled()`] and only build the data and call `write()` if `enabled()`
+/// returns true.
+///
+/// Requires: data.len() < 65536.
+///
+/// Return value is 0 for success or an errno code for error. The return value is
+/// provided to help with debugging and should usually be ignored in release builds.
+pub fn write(data: &[u8]) -> i32 {
+    // This must stay in sync with the METRICS_EVENT_DEF string.
+    assert!(data.len() <= u16::MAX as usize);
```
cijothomas, line 34: what does assert do here if it fails assertion? Also shouldn't we be within the max supported by user_events, instead of u16::MAX?

opentelemetry-user-events-metrics/src/transform/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+                    flags: TonicDataPointFlags::default() as u32,
+                    value: Some(dp.value.into()),
+                })
+                .collect(),
+            aggregation_temporality: TonicTemporality::from(sum.temporality).into(),
+            is_monotonic: sum.is_monotonic,
+        }
+    }
+
+    fn transform_exemplar<T: Into<TonicExemplarValue> + Copy>(
```
cijothomas, line 115: nit: we can strip down un-used parts like exemplars.

opentelemetry-user-events-metrics/examples/basic.rs
```diff
@@ -0,0 +1,58 @@
...
+    let _ = exporter.shutdown();
```
lalitb, line 53: Do we need the explicit shutdown for exporter ?

opentelemetry-user-events-metrics/src/exporter/mod.rs
lalitb: This seems to be the default temporality, can we have `Delta` as default ?

opentelemetry-user-events-metrics/Cargo.toml
lzchen: Removed because we already have opentelemetry-proto dependency.

opentelemetry-user-events-metrics/src/exporter/mod.rs
lzchen: Default temporality will probably be implemented with a separate "DefaultTemporalitySelector" (e.g. [DefaultTemporalitySelector](https://github.com/open-telemetry/opentelemetry-rust/blob/a08c6406bb8eadcb8568740a377cfce913a1b66f/opentelemetry-sdk/src/metrics/reader.rs#L99) when we decide to implement this. As for now, I will leave a TODO.

opentelemetry-user-events-metrics/Cargo.toml
lzchen: removed

lzchen: Moved to [dev-dependencies].

opentelemetry-user-events-metrics/src/tracepoint/mod.rs
lzchen: This will panic if the assert fails. I could output a warning and have the write not succeed. u16::MAX was the value given to me by Doug and it is also probably dependent on the METRICS_EVENT_DEF string. Thoughts on how we should proceed?

cijothomas: we don't want panic here. if size exceeds our limit, log a warning and move on. The warning will let users know that the exporter is dropping metrics.

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::tracepoint;
+use prost::Message;
+use std::fmt::{Debug, Formatter};
+
+#[derive(Clone, Copy)]
+pub struct MetricsExporter {}
+
+impl MetricsExporter {
+    pub fn new() -> MetricsExporter {
+        // This is unsafe because if the code is used in a shared object (DLL),
```
lalitb, line 21: nit - remove DLL as it becomes windows specific. It is also unsafe in *inux  for dynamically loaded shared objects.

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        // the event MUST be unregistered before the shared object unloads.
+        unsafe {
+            let result = tracepoint::register();
```
lalitb, line 24: What is the tracepoint name associated with this user_event - is it unique for each exporter instance, and can be enabled/disabled separately ?

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+            if result != 0 {
+                println!("Tracepoint failed to register.");
```
lalitb, line 26: nit -  As discussed in the community meeting today, we should eventually use the custom error handler here. If not planned for this PR, good to add TODO here. 

Also if we plan to add TODO here, replace the `println!` with `eprintln!`, to ensure that the error messages are written to stderr, not the stdout. This holds true for other places in this PR.

opentelemetry-user-events-metrics/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+version = "0.1.0"
+description = "Metrics exporter for user events"
+homepage = "https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-user-events-metrics"
+repository = "https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-user-events-metrics"
+readme = "README.md"
+categories = [
+    "development-tools::debugging",
+    "development-tools::profiling",
+    "asynchronous",
+]
```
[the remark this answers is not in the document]
lzchen: Removed.

opentelemetry-user-events-metrics/src/exporter/mod.rs
lzchen: Fixed.

lalitb (approved):
Thanks for the PR.

lzchen:
@TommyCpp 
This is ready for review!

opentelemetry-user-events-metrics/examples/basic.rs
```diff
@@ -0,0 +1,63 @@
...
+        ]
+        .as_ref(),
+    );
+
+    meter_provider.shutdown()?;
+    // Call this explictly since provider shutdown is not propagated to exporter for now
+    // Waiting on: https://github.com/open-telemetry/opentelemetry-rust/issues/1118
+    exporter.shutdown();
+
+    thread::sleep(Duration::from_secs(5));
```
TommyCpp, line 60: Probably best to use tokio sleep here as it's in an async function

lzchen: Done.

lzchen:
@TommyCpp 

Can this be merged?

jtescher (approved):

PR #1121, opened 2023-06-20
Introduce key/value map bench

shaun-cox:
Two new benchmarks:
- `key_value_map` for evaluating different implementation choices
- `span_builder` to focus on performance analysis of `SpanBuilder`

## Overview 

There are four implementations considered:
- `EvictedHashMap`: currently used by `SpanData` to carry span attributes
- `IndexMap`: currently used by `SpanBuilder` to carry span attributes (in an `OrderMap`)
- `OneVec`: `Vec<(Key, Value>)` where no hashing or duplicate key detection takes place
- `TwoVec`: a `Vec<Key>` and a `Vec<Value>`, linked by shared indices, where no hashing or duplicate key detection takes place

There are two main operations to consider for each implementation:
- `lookup`: find two attributes in "the map"
- `populate`: populate `n` (2, 8, or 32) attributes in the map

`lookup` approximates what a sampling decision might do in consulting specific keys present in the `SpanBuilder`. Its performance looks like the following: (Note, the `Vec` based implementations are pessimized for worst-case, finding the last two attributes in the map.)

violin

`OneVec` beats `IndexMap` for 2 and 8 attributes in the map, but loses for 32 attributes.

`populate` approximates what anyone who wants to create a `Span` must pay to get attributes into the `SpanBuilder`. Its performance looks like the following:

violin

As expected, populating hash maps is a lot more expensive than populating vectors, but that's the cost of detecting duplicates.

## Learnings

Knowing that we have to both `populate` and `lookup` when creating a `Span`, it's useful to see both together:

violin

Now we see that `OneVec` beats `IndexMap` for all cases, and is more than twice as fast as `IndexMap` for 32 attributes.

So this PR is meant to do three things:
1. Give us a tool for further study.
2. Help us come to a decision or better strategy about where to handle duplicate detection of attributes. (Personally, I think the cost should be paid for by the user or in the sdk as an option, but not in the api as is currently done.)
3. Remind us all that O(1) vs. O(N) applies in the large, not necessarily for "tiny" data sets as we have here with attribute sets.  So I think #794 should be revisited in this light. e.g. if a span processor or sampling decision wants to lookup attributes, it should be considering indexing to make those lookups faster in light of how many attributes are actually present and not necessarily make everyone who creates a `SpanBuilder` that ends up producing a `Span` that is non-recording pay for indexing that will never get used.

### DIFF
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -36,6 +36,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dev-dependencies]
+indexmap = "1.8"
 criterion = { version = "0.4.0", features = ["html_reports"] }
 pprof = { version = "0.11.1", features = ["flamegraph", "criterion"] }
 
@@ -49,6 +50,14 @@ testing = ["opentelemetry_api/testing", "trace", "metrics", "logs", "rt-async-st
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
+
+[[bench]]
+name = "key_value_map"
+harness = false
+
+[[bench]]
+name = "span_builder"
+harness = false
 
 [[bench]]
 name = "trace"
--- a/opentelemetry-sdk/benches/key_value_map.rs
+++ b/opentelemetry-sdk/benches/key_value_map.rs
@@ -1,0 +1,208 @@
+use criterion::{
+    black_box, criterion_group, criterion_main, BatchSize::SmallInput, BenchmarkId, Criterion,
+};
+use indexmap::IndexMap;
+use opentelemetry_api::{Key, KeyValue, Value};
+use opentelemetry_sdk::trace::EvictedHashMap;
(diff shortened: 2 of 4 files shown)


djc:
Awesome work! It makes a lot of sense to me that, even if duplicate detection is necessary, the data sizes are such that a simple sequential span might still be the optimal way to store data.

cijothomas:
> Help us come to a decision or better strategy about where to handle duplicate detection of attributes. (Personally, I think the cost should be paid for by the user or in the sdk as an option, but not in the api as is currently done.)

I really like idea of removing duplication detection logic (i.e switch from HashMaps to OneVec/TwoVec ideas presented) from the API/SDK.  If there is strong need to dedup, then it can be added as an optional span/log processor OR in the OTLPExporter itself, so the main user thread won't pay any cost. The spec actually allows flexibility in terms on where the de-dup should occur.

https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute-collections
**_"The enforcement of uniqueness may be performed in a variety of ways as it best fits the limitations of the particular implementation."_**

(This topic was discussed in the SIG call on 06/20 as well, but after re-reading the spec I think it totally makes sense for this SIG to remove the costly de-dup.)

jtescher (approved):

PR #1126, opened 2023-06-22
Stress test for metric modified to use random attributes

cijothomas:
The existing test did not pass any attributes at all. This PR adds random attributes (~1000 timeseries, within the 2000 limit), so as to measure real world throughput.

### DIFF
--- a/stress/src/metrics.rs
+++ b/stress/src/metrics.rs
@@ -1,6 +1,7 @@
 use lazy_static::lazy_static;
-use opentelemetry_api::metrics::{Counter, MeterProvider as _};
+use opentelemetry_api::{metrics::{Counter, MeterProvider as _, Histogram}, KeyValue, Value};
 use opentelemetry_sdk::metrics::{ManualReader, MeterProvider};
+use rand::Rng;
 use std::borrow::Cow;
 
 mod throughput;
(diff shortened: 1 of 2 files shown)


stress/src/metrics.rs
```diff
@@ -20,5 +28,19 @@ fn main() {
 }
 
 fn test_counter() {
-    COUNTER.add(1, &[]);
+    let mut rng = rand::thread_rng();
+    let len = ATTRIBUTE_VALUES.len();
+    let index_first_attribute = rng.gen_range(0..len);
+    let index_second_attribute = rng.gen_range(0..len);
+    let index_third_attribute = rng.gen_range(0..len);
```
lalitb, line 35: Initializing/generating random numbers could have significant overhead, need to see if there is easy way not to include this overhead in throughput calculation. If this is not easy to fix in this PR, we can add a TODO to take care in future.

cijothomas: I think its inevitable that the cost must be paid to generate random number! An alternate could be to just call attribute[1,2,3,...6,1,2,3...] sequentially.
I'll check if this can be optimized further.

djc: You're using the `ThreadRng` here, though, which is much slower than necessary for this purpose because it is documented to be secure. Try using a `SmallRng` instead.

lzchen:
Curious, what numbers are you getting for your throughput? We should probably add an equivalent test in `benches` so we know what to expect.

stress/src/metrics.rs
lzchen: Perhaps we could also generate a sequence of three value tuples before hand (outside of this function) and then call test_counter with each tuple sequentially?

jtescher (approved):

jtescher:
@cijothomas should we merge this as is? or still working on it?

stress/src/metrics.rs
cijothomas: Moving to `SmallRng` improved throughtput in my machine from ~2-3 M/sec to ~6 M/sec.

lalitb (approved):

cijothomas:
@jtescher Ready to merge!

stress/src/metrics.rs
hdost: > Perhaps we could also generate a sequence of three value tuples before hand (outside of this function) and then call test_counter with each tuple sequentially?

I was thinking the same thing, because at some level we're testing the random generator in this stress test but if all the attributes are before it better tests the machinery.

Issue #1115, opened 2023-06-15
SpanId and TraceId improvements and their usage in SpanData


valkum:
Currently, `SpanId` and `TraceId` are simple new types around `u128`/`u64`.

The value of 0 as used as an invalid placeholder.
In some places (e.g. with span parents) we see a pattern of Option<SpanId>. This can be optimized by size if we would use `NonZeroU128`/`NonZeroU64`

The reason for this change gets clearer if we look at `sdk::SpanData`. This type holds a `span_parent_id` and is used by exporters. When this gets constructed for a root span the `parent_span_id` [gets set](https://docs.rs/opentelemetry_sdk/latest/src/opentelemetry_sdk/trace/tracer.rs.html#155) to `SpanId::INVALID`.

`opentelemetry_stackdriver` currently exports the `parant_span_id` without checkout the validity, which it should. 
A recent change in googles backend now marks spans with a parent of `0x0` as having missing parent spans.

To fix this, we can either check the validity in the exporter or change `SpanId` to` SpanId(NonZeroU64)` and use `Option<SpanId>` in `SpanData`.
The optimization holds for `Options` of new types as well, see:
```
pub struct SpanId(core::num::NonZeroU64);

fn main() {
    use std::mem::size_of;
    assert_eq!(size_of::<Option<SpanId>>(), size_of::<u64>());
}
```

I think the `Option<SpanId>` way feels more ergonomic, at least for `SpanData::parent_span_id`.

The inner type of `SpanId` and `TraceId` is not exposed publicly. `SpanData::parent_span_id` is public and thus changing this would be a breaking change.

djc:
See similar discussion in https://github.com/open-telemetry/opentelemetry-rust/pull/1089.

shaun-cox:
I think `Option<SpanId>` feels way more idiomatic for Rust than SpanID::INVALID as well.  Thanks for pointing it out!

PR #1127, opened 2023-06-22
Stackdriver improvements

valkum:
Fixes issue mentioned in #1115

## Changes

This fixes an issue with stackdriver and invalid parent span ids when a span has no parent. Currently the invalid zero `SpanId` is send and causes Cloud Trace to show missing spans. Until we improve the `SpanIds` we need to check if the ID is invalid (which currently means this is the root span) and send an empty parent span ID to cloud trace.

Also adds support for span links to `opentelemetry-stackdriver`, but it seems the current UI does not show the links.


opentelemetry-stackdriver/src/lib.rs
```diff
@@ -311,6 +311,11 @@ where
                 }
             };
 
+            let parent_span_id = match span.parent_span_id {
+                SpanId::INVALID => "".to_owned(),
+                _ => hex::encode(span.parent_span_id.to_bytes()),
+            };
```
djc, line 317: Style nit: can just inline this into the initializer below, I think?

Is there documentation for this? Should we point to it here?

opentelemetry-stackdriver/src/lib.rs
```diff
@@ -731,6 +732,24 @@ impl From<EvictedHashMap> for Attributes {
...
+        return None;
+    }
+
+    Some(Links {
+        dropped_links_count: links.dropped_count() as i32,
+        link: links
+            .iter()
+            .map(|link| Link {
+                trace_id: hex::encode(link.span_context.trace_id().to_bytes()),
+                span_id: hex::encode(link.span_context.span_id().to_bytes()),
```
djc, line 746: Can we have `INVALID` IDs here, and if so, should we treat them differently?

valkum: So, the `SpanBuilder` does check the validity of links, but if you add links by hand later, you could add links with invalid spans contexts. But I don't think this is something which should be cought here. Maybe the span builder should add an API and make the `links` field private. Enforcing the validity check in the new added API.
Note that the check above is only needed because currently SpanData uses the invalid span id to mark the span as a root span.

djc (approved):

PR #1129, opened 2023-06-23
Create tracer using a shared instrumentation library

shaun-cox:
Would like to submit this for review/comments...

I noticed two things regarding creating tracers:
1. a lot of repetition in passing around the following:
```rust
        name: Cow<'static, str>,
        version: Option<Cow<'static, str>>,
        schema_url: Option<Cow<'static, str>>,
        attributes: Option<Vec<KeyValue>>,
```
2. the inability of a user to use a shared `InstrumentationLibrary` to create tracers at multiple different places.  IOW, user can now have a lazy static `InstrumentationLibrary` and hopefully in the future could use it for tracing as well as metrics.

## Changes

Rather than have the SDK be the only place that creates the `InstrumentationLibrary` internal to its `versioned_tracer` implementation, it now can accept one from the user.

The debatable part is how I did this to minimize disruption. I added a new `library_tracer` function to the `global::TracerProvider` api trait and then default-implemented the existing `versioned_tracer` to defer to the new function.

### DIFF
--- a/opentelemetry-api/src/common.rs
+++ b/opentelemetry-api/src/common.rs
@@ -464,7 +464,7 @@ pub struct InstrumentationLibrary {
     pub schema_url: Option<Cow<'static, str>>,
 
     /// Specifies the instrumentation scope attributes to associate with emitted telemetry.
-    pub attributes: Option<Vec<KeyValue>>,
+    pub attributes: Vec<KeyValue>,
 }
 
 // Uniqueness for InstrumentationLibrary/InstrumentationScope does not depend on attributes
@@ -498,7 +498,7 @@ impl InstrumentationLibrary {
             name: name.into(),
             version: version.map(Into::into),
             schema_url: schema_url.map(Into::into),
-            attributes,
+            attributes: attributes.unwrap_or_default(),
         }
     }
 }
--- a/opentelemetry-api/src/trace/tracer_provider.rs
+++ b/opentelemetry-api/src/trace/tracer_provider.rs
@@ -1,5 +1,5 @@
-use crate::{trace::Tracer, KeyValue};
-use std::borrow::Cow;
+use crate::{trace::Tracer, InstrumentationLibrary, KeyValue};
+use std::{borrow::Cow, sync::Arc};
 
 /// Types that can create instances of [`Tracer`].
 ///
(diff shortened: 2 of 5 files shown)


opentelemetry-api/src/common.rs
```diff
@@ -464,7 +464,7 @@ pub struct InstrumentationLibrary {
     pub schema_url: Option<Cow<'static, str>>,
 
     /// Specifies the instrumentation scope attributes to associate with emitted telemetry.
-    pub attributes: Option<Vec<KeyValue>>,
```
shaun-cox, line 467: I removed `Option` here as I don't see the point in having both states (`None` and `Some(vec![])`) represent the logical case of having no attributes.

opentelemetry-api/src/common.rs
```diff
@@ -498,7 +498,7 @@ impl InstrumentationLibrary {
             name: name.into(),
             version: version.map(Into::into),
             schema_url: schema_url.map(Into::into),
-            attributes,
+            attributes: attributes.unwrap_or_default(),
```
shaun-cox, line 501: I kept the `attributes` parameter as `Option<Vec<KeyValue>>` here because it's more idiotmatic to express `None` for "I have no attributes to provide" than to require the api user to actually give up ownership of an empty vector.  IOW, `Option` better expresses "optional" parameters... but we don't need to store it as such, hence the `unwrap_or_default()`

opentelemetry-api/src/trace/tracer_provider.rs
```diff
@@ -75,5 +75,31 @@ pub trait TracerProvider {
...
+    /// ```
+    /// use opentelemetry_api::{global, trace::TracerProvider};
+    ///
+    /// let provider = global::tracer_provider();
+    ///
+    /// // tracer used in applications/binaries
+    /// let tracer = provider.tracer("my_app");
+    ///
+    /// // tracer used in libraries/crates that optionally includes version and schema url
+    /// let tracer = provider.versioned_tracer(
```
TommyCpp, line 97: shouldn't this be using `library_tracer`?

TommyCpp (approved):

djc (approved):

PR #1130, opened 2023-06-26
fix: replace `futures` crates with `futures-*` crates.

### LINKED ISSUE #1064: Use of the `futures` crate as a library dependency
Hi, I noticed that a bunch of the opentelemetry crates use the `futures` crate as a (non-dev) dependency. I think they should depend on the required `futures-foo` crates instead:

- So only the parts that are actually needed are pulled in
- To improve build pipelining (even if the final app uses all of `futures-*`, more fine-grained dependency specifications allow more crates to be compiled in parallel)

Would a PR to implement this be welcome?

TommyCpp:
fix #1064

## Changes
- Replace all `futures` crate with `futures-*` creates unless they are dev dependencies


djc (approved):

djc:
Nice!

PR #1136, opened 2023-06-29
Tracing subscriber: handle string and bool values correctly.

lalitb:
## Changes

Override default implementation of `Visit::record_str()` and `Visit::record_bool()` to handle string and bool attribute values properly, instead of relying on `Visit::record_debug()`. This improves the performance of fetching these values from tracing crate.

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -1,8 +1,6 @@
 use opentelemetry_api::{
-    logs::LoggerProvider,
-    logs::Severity,
-    logs::{LogRecord, Logger},
-    OrderMap,
+    logs::{AnyValue, LogRecord, Logger, LoggerProvider, Severity},
+    Key, OrderMap,
 };
 
 use tracing_subscriber::Layer;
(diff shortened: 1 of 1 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -27,6 +25,27 @@ impl<'a> tracing::field::Visit for EventVisitor<'a> {
         }
     }
 
+    fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
+        if let Some(ref mut map) = self.log_record.attributes {
+            map.insert(field.name().into(), value.to_owned().into());
+        } else {
+            let mut map = OrderMap::with_capacity(1);
+            let k: Key = field.name().into();
```
djc, line 33: Style nit: confused why you bind `k` and `v` here rather than inlining them as you've done in `record_bool()`. The type annotations should not be necessary here.

lalitb: I got the error without type annotations but then it works after specifying the explicit type for OrderMap. Have done the changes. Thanks for the comment.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -27,6 +25,25 @@ impl<'a> tracing::field::Visit for EventVisitor<'a> {
         }
     }
 
+    fn record_str(&mut self, field: &tracing_core::Field, value: &str) {
+        if let Some(ref mut map) = self.log_record.attributes {
+            map.insert(field.name().into(), value.to_owned().into());
+        } else {
+            let mut map: OrderMap<Key, AnyValue> = OrderMap::with_capacity(1);
+            map.insert(field.name().into(), value.to_owned().into());
```
djc, line 33: Aren't you forgetting to set `self.log_record.attributes` here?

lalitb: Good catch. Was testing with the right code, and then missed during the commit. Fixed now :)

djc (approved):

Issue #1118, opened 2023-06-15
PeriodicReader.shutdown() does not propagate to exporter.shutdown()


lzchen:
From the logic [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/periodic_reader.rs#L362) it doesn't seem the corresponding exporter is shutdown when the reader is shutdown.

lzchen:
@TommyCpp 
I'll be working on this.

PR #1138, opened 2023-06-30
Propagate periodicreader.shutdown() to the exporter

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1118

Removes `async` from PushMetricsExporter.shutdown(), similar to tracing and logging.


TommyCpp (approved):
Looks good :+1:

PR #1142, opened 2023-07-05
[Logs SDK] Modify LogRecord to use Vector instead of OrderMap for attributes

lalitb:
## Changes

This is to further validate the benchmark introduced in #1121, changing the `LogRecord` to store attributes in `Vec` instead of `OrderMap`/`IndexMap`.  The benchmark was done to emit the tokio-tracing log to user-event exporter.
As an example, to emit the error log with below attributes:

```
error!(
        event_name = "my-event-name",
        event_id = 20,
        user_name = "otel user",
        user_email = "otel@opentelemetry.io",
        login_success =  false
)
```

With `OrderMap`:
It takes around ~1000 ns to emit a single event.
With `Vector`:
It takes around ~700 ns to emit a single event.

The user_events tracepoint was disabled, so no actual export was happening. Which means most of the above overhead was coming from logs SDK (and not the exporter). 

The user_events export iterate over the attributes to serialize, and doesn't require to do any lookup on particular key. 

These changes come at below cost:
 - If the number of attributes are large, and exporter need to do direct lookup for particular attribute - the operation would be fast in OrderMap as compared to Vector. I feel this use-case is somewhat rare.
 - The duplicate key detection need to be done by application/instrumentation library before writing attributes.

The PR is raised to further discuss if these changes would be good for log signal, or any better suggestion. Another option I thought was to have SDK support both data structures, and let user select which one to use. Something like this (not tested):

```rust

enum AttributeData<K, V> {
    HashStore(HashMap<K, V>),
    VectorStore(Vec<(K, V)>),
}

 
impl<K, V> AttributeData<K, V> {
    fn iterator(&self) -> Box<dyn Iterator<Item = &(K, V)>> {
        match self {
            AttributeData::HashStore(index_map) => {
                Box::new(index_map.iter().map(|(k, v)| &(k, v)))
            }
            AttributeData::VectorStore(vector) => Box::new(vector.iter()),
        }
    }

  fn insert(&mut self, key: K, value: V) {
        match self {
            AttributeData::HashStore(index_map) => {
                index_map.insert(key, value);
            }
            AttributeData::VectorStore(vector) => {
                vector.push((key, value));
            }
        }
    }
}

pub struct LogRecord {
    // existing fields ..

    pub attributes: Option<AttributeData<Key, AnyValue>>,
}
```

### DIFF
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -342,12 +342,11 @@ impl LogRecordBuilder {
         K: Into<Key>,
         V: Into<AnyValue>,
     {
-        if let Some(ref mut map) = self.record.attributes {
-            map.push((key.into(), value.into()));
+        if let Some(ref mut vec) = self.record.attributes {
+            vec.push((key.into(), value.into()));
         } else {
-            let mut map = Vec::with_capacity(1);
-            map.push((key.into(), value.into()));
-            self.record.attributes = Some(map);
+            let vec = vec![(key.into(), value.into())];
+            self.record.attributes = Some(vec);
         }
 
         self
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -326,7 +326,8 @@ impl LogRecordBuilder {
         }
     }
 
-    /// Assign attributes, overriding previously set attributes
+    /// Assign attributes. 
+    /// The SDK doesn't carry on any deduplication on these attributes.
     pub fn with_attributes(self, attributes: Vec<(Key, AnyValue)>) -> Self {
         Self {
             record: LogRecord {
@@ -336,7 +337,8 @@ impl LogRecordBuilder {
         }
     }
 
-    /// Set a single attribute for this record
+    /// Set a single attribute for this record.
+    /// The SDK doesn't carry on any deduplication on these attributes.
     pub fn with_attribute<K, V>(mut self, key: K, value: V) -> Self
     where
         K: Into<Key>,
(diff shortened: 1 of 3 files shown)


djc (approved):
Why was an `OrderMap` chosen in the first place? Do we do any lookups/deduplication on this data?

shaun-cox:
> Why was an `OrderMap` chosen in the first place? Do we do any lookups/deduplication on this data?

I think it was a result of https://github.com/open-telemetry/opentelemetry-rust/issues/794

shaun-cox (approved):

lalitb:
> Do we do any lookups/deduplication on this data?

There are no lookups in the log SDK. The lookups could be part of any external exporters, but there shouldn't be significant difference in complexity - o(n) vs o(1) - for smaller number of attributes. 
With this change, the deduplication would now be the responsibility of the instrumented library/application. The application needs to ensure that all the attributes are unique.

opentelemetry-api/src/logs/record.rs
```diff
@@ -327,7 +327,7 @@ impl LogRecordBuilder {
     }
 
     /// Assign attributes, overriding previously set attributes
-    pub fn with_attributes(self, attributes: OrderMap<Key, AnyValue>) -> Self {
+    pub fn with_attributes(self, attributes: Vec<(Key, AnyValue)>) -> Self {
```
TommyCpp, line 330: Since user will now handle deduplication themselves. Should we call it call in comments explicitly so that for users needs unique attributes can handle it themselves

lalitb: Good point, have added the comment for `LogRecordBuilder::with_attributes` and `LogRecordBuilder::with_attribute` methods.

PR #1147, opened 2023-07-11
Export logs only if the level/logger/target is enabled.

lalitb:
This PR is for discussion in the community meeting. As of now, the pipeline to emit the logs using the tokio-tracing library with otel-sdk is as follows (using user_events exporter as an example)

`error!("login failed")` -> `Layer::on_event()` -> `create LogRecord` -> `logger.emit()` -> `create LogData` -> `processor.emit()` -> `user_events_exporter::export()`  
               -> `[Linux Kernel ring buffer]` <- `perf listener`

So the event is written to the Linux kernel ring buffer, from where the [perf ](https://man7.org/linux/man-pages/man1/perf.1.html) listener will read the events. The event is written to the kernel buffer even when there is no listener reading the ring buffer. Further, the listener can be configured only to collect (say) critical errors/warnings, and ignore debug/trace events. This means that the application will be spending ~1000 ns to emit the events which are eventually dropped if there is no listener configured to read that event.  This overhead can be avoided if it is possible for the tokio-tracing log appender to emit the events only when a listener is attached or the listener is configured to read that type of event. 
The proposal is to add a new method `exporter::event_enabled()` in Exporter. This method will return true if the exporter is configured to read that type of event. The type here could be a combination of level + logger-name. This method can then be called from tokio-tracing log appender as below:

`error("login failed")` -> `Layer::event_enabled()` -> `logger::event_enabled(level, target)` -> `processor::event_enabled(level, target, logger_name)` -> `exporter::event_enabled(level, target, logger_name)`
 
  -> False, drop the event ( this has ~40 ns of overhead)
  -> True, emit the event ( this has ~1000 ns of overhead)

So, `event_enabled` method would also be added in the logger and processor to eventually enable calling this method from the exporter. As this method is not part of otel specs, we can add it behind the feature flag as below:
```rust
fn event_enabled(&self, level : crate::logs::Severity) -> bool {
   . . .
 }
```
So to summarise - for scenarios where the exporter is only configured to receive specific types of events, we can get rid of all the unnecessary processing for the other events. These performance gains are significant for the system applications written in C/C++ and Rust.


opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -51,59 +73,36 @@ impl UserEventsExporter {
...
+        eventheader_provider.register_set(eventheader::Level::Verbose, keyword);
+        eventheader_provider.register_set(eventheader::Level::Warning, keyword);
+        eventheader_provider.register_set(eventheader::Level::Error, keyword);
+        eventheader_provider.register_set(eventheader::Level::CriticalError, keyword);
+
+        eventheader_provider.create_unregistered(true, eventheader::Level::Informational, keyword);
+        eventheader_provider.create_unregistered(true, eventheader::Level::Verbose, keyword);
+        eventheader_provider.create_unregistered(true, eventheader::Level::Warning, keyword);
+        eventheader_provider.create_unregistered(true, eventheader::Level::Error, keyword);
+        eventheader_provider.create_unregistered(true, eventheader::Level::CriticalError, keyword);
```
idigdoug, line 103: What is the `create_unregistered` for?

`create_unregistered` is a no-op if the event set is already registered, which it should be because you just called `register_set`.

`create_unregistered` is "for testing purposes".

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -17,18 +19,38 @@ thread_local! { static EBW: RefCell<EventBuilder> = RefCell::new(EventBuilder::n
...
-    pub(crate) fn get_log_event_keyword(&self) -> u64 {
-        self.keyword
+    pub(crate) fn get_log_keyword(&self, name: &str) -> Option<u64> {
+        match self.keywords_map.get(name) {
+            Some(value) => Some(*value),
+            _ => None,
+        }
+    }
+
+    pub(crate) fn get_log_keywork_or_default(&self, name: &str) -> Option<u64> {
```
idigdoug, line 43: Spelling

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -51,59 +73,36 @@ impl UserEventsExporter {
...
-        );
-        eventheader_provider.create_unregistered(
-            true,
-            eventheader::Level::CriticalError,
-            exporter_config.get_log_event_keyword(),
-        );
-
+        if exporter_config.keywords_map.len() == 0 {
+            println!(
+                "Register default keyworkd {}",
```
idigdoug, line 78: Spelling.
Is this println intended to still be here in the final PR?

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: Thanks for clarifying. Do we need to do anything explicit to ensure that the events are unregistered (i,e tracepoint names are removed) once the application shutdowns?

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: This will be removed eventually. As of now, the PR is raised to discuss various approaches.

lalitb:
Two options apart from the existing proposal were discussed in today's meeting :
- Configure the `event_enabled` callback as part of opentelemetry_sdk::Logger, similar to how IdGenerator / Samplers are configured for Tracer. This callback would result in events getting dropped, or generate non-recordable events (similar to non-recordable spans).
- @shaun-cox - You had some suggestions to create the log equivalent for `TraceContextExt`. If you could elaborate on it further here.

opentelemetry-user-events-logs/src/logs/exporter.rs
idigdoug: The event will be unregistered when the provider is dropped.

lalitb: Thanks, have removed `create_unregistered` calls here.

TommyCpp (approved):
Looks good to me. We can experiment this under the feature flag.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -14,6 +16,11 @@ pub trait LogExporter: Send + Debug {
     async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()>;
     /// Shuts down the expoter.
     fn shutdown(&mut self) {}
+    //#[cfg(feature = "logs_level_enabled")]
+    /// Chek if logs are enabled.
+    fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
```
cijothomas, line 21: the method in the API does not have the third parameter (name) ? 
`fn event_enabled(&self, level: Severity, target: &str) -> bool;`

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -342,4 +338,29 @@ impl opentelemetry_sdk::export::logs::LogExporter for UserEventsExporter {
         }
         Ok(())
     }
+
+    #[cfg(feature = "logs_level_enabled")]
+    fn event_enabled(&self, level: Severity, _target: &str, name: &str) -> bool {
+        println!("event_enabled called");
```
cijothomas, line 344: nit: remove this

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -342,4 +338,29 @@ impl opentelemetry_sdk::export::logs::LogExporter for UserEventsExporter {
...
+                Some(x) => (true, x),
+                _ => (false, 0),
+            }
+        };
+        if !found {
+            return false;
+        }
+        let es = self
+            .provider
+            .find_set(self.get_serverity_level(level), keyword);
```
cijothomas, line 359:
```suggestion
            .find_set(self.get_severity_level(level), keyword);
```

opentelemetry-sdk/src/export/logs/mod.rs
lalitb: Yes, the third parameter is part of `LogProcessor::event_enabled` and `LogExporter::event_enabled`. The third parameter is logger-name, which is added by logger.

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: Good catch. Have fixed the method name throughout.

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: removed.

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -51,59 +66,30 @@ impl UserEventsExporter {
...
-            exporter_config.get_log_event_keyword(),
-        );
-        eventheader_provider.create_unregistered(
-            true,
-            eventheader::Level::CriticalError,
-            exporter_config.get_log_event_keyword(),
-        );
-
+        if exporter_config.keywords_map.is_empty() {
+            println!(
```
cijothomas, line 70: remove println

cijothomas (approved):
LGTM.
We need to evolve the parameters for the is_enabled further based on needs. The current "target", "name" seems to be good enough to start with.

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: done.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -14,6 +16,11 @@ pub trait LogExporter: Send + Debug {
     async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()>;
     /// Shuts down the expoter.
     fn shutdown(&mut self) {}
+    //#[cfg(feature = "logs_level_enabled")]
```
jtescher, line 19: nit: comment

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -1,6 +1,8 @@
 //! Log exporters
 use crate::Resource;
 use async_trait::async_trait;
+//#[cfg(feature = "logs_level_enabled")]
```
jtescher, line 4: nit: comment

jtescher (approved):
looks reasonable

opentelemetry-sdk/src/export/logs/mod.rs
lalitb: fixed. Thanks.

lalitb:
@jtescher - Can we include this in 0.20.0 release?

PR #1149, opened 2023-07-13
Create instance of tracepointstate in user_events metric exporter

lzchen:
Previously, a shared static `TracepointState` was being used to write to user events. This obviously won't work if user had multiple user events exporters, as there might be conflicts when registering/unregistering. So each exporter will have it's own instance of tracepointstate that it uses to write.

As well, the tracepointstate will unregister [automatically](https://github.com/microsoft/LinuxTracepoints-Rust/blob/main/eventheader/src/native.rs#L618) upon dropping so no explicit unregister call needs to be made upon shutdown.


jtescher (approved):

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -2,31 +2,40 @@ use crate::transform::transform_resource_metrics;
...
         // the event MUST be unregistered before the shared object unloads.
         unsafe {
-            let result = tracepoint::register();
-            if result != 0 {
-                eprintln!("Tracepoint failed to register.");
+            let result = tracepoint::register(trace_point.as_ref());
+            // If result is -1, tracepoint might be in an erroneous state, attempt to retry
+            if result == -1 {
+                eprintln!("Tracepoint failed to register. Try restarting your application.");
+            } else if result != 0 {
```
lalitb, line 32: positive `result` is also considered failure?

lalitb (approved):

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -2,31 +2,40 @@ use crate::transform::transform_resource_metrics;
...
 impl MetricsExporter {
     pub fn new() -> MetricsExporter {
+        let trace_point = Box::pin(ehi::TracepointState::new(0));
         // This is unsafe because if the code is used in a shared object,
         // the event MUST be unregistered before the shared object unloads.
         unsafe {
-            let result = tracepoint::register();
-            if result != 0 {
-                eprintln!("Tracepoint failed to register.");
+            let result = tracepoint::register(trace_point.as_ref());
```
idigdoug, line 28: As mentioned in the comments for register, the result of register should be used for debugging only and should generally be ignored in release code.

opentelemetry-user-events-metrics/src/exporter/mod.rs
```diff
@@ -67,16 +76,16 @@ impl Debug for MetricsExporter {
...
-        if tracepoint::enabled() {
+        if self.trace_point.enabled() {
             let proto_message = transform_resource_metrics(metrics);
 
             let mut byte_array = Vec::new();
             let _encode_result = proto_message
                 .encode(&mut byte_array)
                 .map_err(|err| MetricsError::Other(err.to_string()))?;
-            let result = tracepoint::write(byte_array.as_slice());
+            let result = tracepoint::write(&self.trace_point, byte_array.as_slice());
```
idigdoug, line 86: As mentioned in the comments for write, the result of write is for debugging. It can fail for various reasons, mainly if there is nobody listening for the event. These failures should not affect the running program and are only interesting for diagnostics.

opentelemetry-user-events-metrics/src/tracepoint/mod.rs
```diff
@@ -67,19 +61,21 @@ pub fn enabled() -> bool {
...
-        .register(ffi::CStr::from_bytes_with_nul_unchecked(METRICS_EVENT_DEF))
-}
-
-/// Unregisters this tracepoint.
-///
-/// Return value is 0 for success or an errno code for error. The return value is
-/// provided to help with debugging and should usually be ignored in release builds.
-pub fn unregister() -> i32 {
-    METRICS_EVENT.unregister()
+    let result = panic::catch_unwind(|| {
```
idigdoug, line 67: This is pointless. Don't catch the panic. There are two cases:

- Either you create a new Tracepoint and then register it, in which case it will never panic.
- Or you have a bug in your code, and it will panic to help you find the bug.

If there is a bug in your code, you want it to panic so you can find and fix the bug. If your code is good then there will never be a panic. Instead of catching the panic (hiding the bug), you should fix your code so that you don't hit the panic.

opentelemetry-user-events-metrics/src/exporter/mod.rs
idigdoug: The panic catch returns -1 for panic. (Shouldn't happen.)

The actual API returns 0 for success and an errno value (positive) for error.

lzchen: Removed.

opentelemetry-user-events-metrics/src/tracepoint/mod.rs
lzchen: Removed the panic catch.

opentelemetry-user-events-metrics/src/exporter/mod.rs
lzchen: Removed.

Issue #1029, opened 2023-04-17
SdkProvidedResourceDetector fallback to env!("CARGO_BIN_NAME") not working


DSeeLP:
The environment variables cargo sets are crate specific, resulting in `CARGO_BIN_NAME` not being set for opentelemetry.
`CARGO_BIN_NAME` is only set for the actual binary crate and not for opentelemetry.

PR #1150, opened 2023-07-14
Don't use CARGO_BIN_NAME for service name

pkoenig10:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1029

## Changes

Change `SdkProvidedResourceDetector` to no longer use the `CARGO_BIN_NAME` environment variable at compile time. This crate is not a binary crate, so the value of `option_env!("CARGO_BIN_NAME")` will always be false.

This can be reproduced by creating a binary crate and confirming that the detected `service.name` is `unknown_service`.

```rust
use opentelemetry::sdk::resource::{ResourceDetector, SdkProvidedResourceDetector};
use std::time::Duration;

fn main() {
    assert_eq!(
        SdkProvidedResourceDetector
            .detect(Duration::ZERO)
            .get("service.name".into()),
        Some("unknown_service".into())
    );
}
```

This effectively reverts https://github.com/open-telemetry/opentelemetry-rust/pull/991.

### DIFF
--- a/opentelemetry-sdk/src/resource/env.rs
+++ b/opentelemetry-sdk/src/resource/env.rs
@@ -3,7 +3,7 @@
 //! Implementation of `ResourceDetector` to extract a `Resource` from environment
 //! variables.
 use crate::resource::{Resource, ResourceDetector};
-use opentelemetry_api::{Key, KeyValue};
+use opentelemetry_api::{Key, KeyValue, Value};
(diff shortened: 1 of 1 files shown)


jtescher (approved):
Yeah looks like `CARGO_BIN_NAME` would only work for the examples. Good to remove.

PR #1151, opened 2023-07-14
Fix EvictedQueue bug with zero capacity

### LINKED ISSUE #1148: SpanLimits going over limit set for max events per spans
I'm trying to set the limit on events per spans to 0, but events are still being exported, bypassing the limit set in the config. Here's a minimal repro:

```rust
/// Run with `OTEL_SPAN_EVENT_COUNT_LIMIT=0 cargo run`

use opentelemetry::{
    global,
    sdk::{self, export::trace::stdout},
    trace::{Span, Tracer},
};

fn main() {
    // Create a new trace pipeline that prints to stdout
    let tracer = stdout::new_pipeline()
        .with_trace_config(sdk::trace::config())
        .install_simple();

    {
        let mut span = tracer.start("doing work");
        span.add_event("my event", vec![]);
        span.add_event("my other event", vec![]);
    }

    // Shutdown trace pipeline
    global::shutdown_tracer_provider();
}
```

Outputs:
```
SpanData { span_context: SpanContext { trace_id: 681f5965702c75e41724caf5b33b9d1b, span_id: f682e85763f26d16, trace_flags: TraceFlags(1), is_remote: false, trace_state: TraceState(None) }, parent_span_id: 0000000000000000, span_kind: Internal, name: "doing work", start_time: SystemTime { tv_sec: 1689254989, tv_nsec: 766169000 }, end_time: SystemTime { tv_sec: 1689254989, tv_nsec: 766173000 }, attributes: EvictedHashMap { map: {}, evict_list: [], max_len: 128, dropped_count: 0 }, events: EvictedQueue { queue: Some([Event { name: "my event", timestamp: SystemTime { tv_sec: 1689254989, tv_nsec: 766172000 }, attributes: [KeyValue { key: Static("my_attribute"), value: String(Static("some_value")) }], dropped_attributes_count: 0 }]), max_len: 0, dropped_count: 1 }, links: EvictedQueue { queue: None, max_len: 128, dropped_count: 0 }, status: Unset, resource: Resource { attrs: {Static("service.name"): String(Owned("unknown_service"))}, schema_url: None }, instrumentation_lib: InstrumentationLibrary { name: "opentelemetry", version: Some("0.19.0"), schema_url: None } }
```

I thought this might be an edge case for `0` and tried setting the limit to `1` but the same issue appears. I also tried overwriting the limit in code rather than through the environment variable, no luck.

I haven't looked to see if this also applied to the other limits as well.

---

Edit: This also has a negative side effect when setting events with `SpanBuilder`, if the limit is set to 0 in that case, all the events are attached to the EvictedQueue, basically putting no limit on the number of events or links on a span.

```rust
let sb = tracer.span_builder("doing work").with_events(vec![
    Event::new("my event", SystemTime::now(), vec![], 0),
    Event::new("my second event", SystemTime::now(), vec![], 0),
    Event::new("my third event", SystemTime::now(), vec![], 0),
]);
let _ = tracer.build_with_context(sb, &Context::current());
```

This particularly affects [tracing-opentelemetry](https://crates.io/crates/tracing-opentelemetry) since that's how they create OpenTelemetry spans from their spans.

wperron:
The EvictedQueue was checking for the length _before_ inserting, and popping extra items, then doing the insertion. In the case where the capacity is set to zero, it caused the pop operation to be a no-op on the first insert, and then insert an item anyway.

This commit fixes the issue by moving the length check after the insert and popping any extra items.

Fixes #1148

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -9,6 +9,11 @@
 - Propagate shutdown calls from `PeriodicReader` to metrics exporter
   [#1138](https://github.com/open-telemetry/opentelemetry-rust/pull/1138).
 
+### Changed
+
+- Fix EvictedQueue bug when capacity is set to 0
+  [#1151](https://github.com/open-telemetry/opentelemetry-rust/pull/1151).
+
 ### Removed
 
 - Samplers no longer has access to `InstrumentationLibrary` as one of parameters
--- a/opentelemetry-sdk/src/trace/evicted_queue.rs
+++ b/opentelemetry-sdk/src/trace/evicted_queue.rs
@@ -26,11 +26,12 @@ impl<T> EvictedQueue<T> {
     /// recording dropped count if over capacity.
     pub(crate) fn push_back(&mut self, value: T) {
         let queue = self.queue.get_or_insert_with(Default::default);
-        if queue.len() as u32 == self.max_len {
+        queue.push_back(value);
+
+        if queue.len() as u32 > self.max_len {
             queue.pop_front();
             self.dropped_count += 1;
         }
-        queue.push_back(value);
     }
 
     /// Moves all the elements of other into self, leaving other empty.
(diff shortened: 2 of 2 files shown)


jtescher (approved):

lalitb (approved):

jtescher:
@wperron should get your lint issues resolved if you merge the changes currently on the main branch

wperron:
On vacation right now, will do when I get back in ~2 weeks 👍

jtescher:
All good, looks like I can merge main for you as well.

wperron:
@jtescher thanks, much appreciated!

PR #1157, opened 2023-07-18
fix: unit conversion in prom exporter

### LINKED ISSUE #927: Prometheus exporter: Unit in metric names
From the [OTel prometheus spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-metadata-1):

> The unit MUST be added as a suffix to the metric name, and SHOULD be converted to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) recommended by OpenMetrics when possible. The unit suffix comes before any type-specific suffixes.

I would recommend against converting to base units (e.g. converting `milliseconds` to `seconds`), but would recommend converting abbreviations to full words for commonly-used units (e.g. `ms` to `milliseconds`).

As an example of what metrics should look like when this is complete, a metric with name `foo`, and unit `ms` should have the resulting name: `foo_milliseconds`.

TommyCpp:
Fixes #927 

## Changes
- move the sanitizing util in Prometheus exporter into a separate file
- Add more units matching from [collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheus/normalize_name.go)
- regenerate the proto generated files

### DIFF
--- a/opentelemetry-prometheus/src/utils.rs
+++ b/opentelemetry-prometheus/src/utils.rs
@@ -20,16 +20,11 @@ pub(crate) fn get_unit_suffixes(unit: &Unit) -> Option<Cow<'static, str>> {
     // e.g
     // "test/y" => "per_year"
     // "km/s" => "kilometers_per_second"
-    if unit.as_str().contains('/') {
-        let components = unit.as_str().splitn(2, '/').collect::<Vec<&str>>();
-        if components.len() != 2 {
-            return None;
-        }
-
+    if let Some((first, second)) = unit.as_str().split_once('/') {
         return match (
-            NON_APPLICABLE_ON_PER_UNIT.contains(&components[0]),
-            get_prom_units(components[0]),
-            get_prom_per_unit(components[1]),
+            NON_APPLICABLE_ON_PER_UNIT.contains(&first),
+            get_prom_units(first),
+            get_prom_per_unit(second),
         ) {
             (true, _, Some(second_part)) | (false, None, Some(second_part)) => {
                 Some(Cow::Owned(format!("per_{}", second_part)))
--- a/opentelemetry-prometheus/src/utils.rs
+++ b/opentelemetry-prometheus/src/utils.rs
@@ -31,11 +31,13 @@ pub(crate) fn get_unit_suffixes(unit: &Unit) -> Option<Cow<'static, str>> {
             get_prom_units(components[0]),
             get_prom_per_unit(components[1]),
         ) {
-            (true, _, Some(second_part)) | (false, None, Some(second_part)) => Some(Cow::Owned(format!("per_{}", second_part))),
+            (true, _, Some(second_part)) | (false, None, Some(second_part)) => {
+                Some(Cow::Owned(format!("per_{}", second_part)))
+            }
             (false, Some(first_part), Some(second_part)) => {
                 Some(Cow::Owned(format!("{}_per_{}", first_part, second_part)))
             }
-            _ => None
+            _ => None,
         };
     }
 
--- a/opentelemetry-prometheus/src/utils.rs
+++ b/opentelemetry-prometheus/src/utils.rs
@@ -1,0 +1,199 @@
+use opentelemetry_api::metrics::Unit;
+use std::borrow::Cow;
+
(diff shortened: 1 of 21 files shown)


jtescher:
Is there a spec or list of these conversions we can link to? Also did the proto submodule get reset? should be on https://github.com/open-telemetry/opentelemetry-proto/commit/c4dfbc51f3cd4089778555a2ac5d9bc093ed2956 for 1.0.0 of the spec.

lalitb:
> Is there a spec or list of these conversions we can link to?

Maybe linking to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.81.0/pkg/translator/prometheus/normalize_name.go#L14C1-L72C10

jtescher:
> Maybe linking to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.81.0/pkg/translator/prometheus/normalize_name.go#L14C1-L72C10

Hm would be nice to have something more authoritative than a contrib repo impl if possible as renaming user's metrics seems like something we'd want to be pretty sure about the stability of. Does look like the java impl does something similar but none of the other languages have yet 🤔

TommyCpp:
Ah sorry looks like I bumped the submodule commits.

I believe the units are coming from https://ucum.org/ucum

cijothomas:
> > Maybe linking to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.81.0/pkg/translator/prometheus/normalize_name.go#L14C1-L72C10
> 
> Hm would be nice to have something more authoritative than a contrib repo impl if possible as renaming user's metrics seems like something we'd want to be pretty sure about the stability of. Does look like the java impl does something similar but none of the other languages have yet 🤔

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-metadata-1

There is some mention of this in the spec (which, btw, is still experimental, and not many languages have implemented these conversions either!)

TommyCpp:
> > > Maybe linking to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.81.0/pkg/translator/prometheus/normalize_name.go#L14C1-L72C10
> > 
> > 
> > Hm would be nice to have something more authoritative than a contrib repo impl if possible as renaming user's metrics seems like something we'd want to be pretty sure about the stability of. Does look like the java impl does something similar but none of the other languages have yet 🤔
> 
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-metadata-1
> 
> There is some mention of this in the spec (which, btw, is still experimental, and not many languages have implemented these conversions either!)

I believe this is #927 refers to. If we think the current unit implementation is OK we can close this PR. My concerns is we only convert 3 units and just discard others so provideing a wider range of units maybe helpful here. For users that don't want to include the units they can disable the `add_unit` option and add the unit to their metrics name directly

jtescher:
> I believe this is #927 refers to. If we think the current unit implementation is OK we can close this PR. My concerns is we only convert 3 units and just discard others so provideing a wider range of units maybe helpful here. For users that don't want to include the units they can disable the `add_unit` option and add the unit to their metrics name directly

Ah yeah that makes sense if the spec is on the prometheus side 👍

opentelemetry-prometheus/src/utils.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    // converting foo/bar to foo_per_bar
+    // split the string by the first '/'
+    // if the first part is empty, we just return the second part if it's a match with known per unit
+    // e.g
+    // "test/y" => "per_year"
+    // "km/s" => "kilometers_per_second"
+    if unit.as_str().contains('/') {
+        let components = unit.as_str().splitn(2, '/').collect::<Vec<&str>>();
```
jtescher, line 24: nit: you could use https://doc.rust-lang.org/std/string/struct.String.html#method.split_once here instead of allocating a new vec

jtescher (approved):

opentelemetry-prometheus/src/utils.rs
TommyCpp: Good point :+1: . Updated

PR #1158, opened 2023-07-19
prometheus: add `without_counter_suffixes` option

jtescher:
## Adds

This adds a method on the prometheus exporter config to not include counter suffixes as required in this spec change https://github.com/open-telemetry/opentelemetry-specification/pull/3590

### DIFF
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -93,7 +93,7 @@
 )]
 #![cfg_attr(test, deny(warnings))]
 
-use once_cell::sync::OnceCell;
+use once_cell::sync::{Lazy, OnceCell};
 use opentelemetry_api::{
     global,
     metrics::{MetricsError, Result, Unit},
@@ -112,6 +112,7 @@ use prometheus::{
     proto::{LabelPair, MetricFamily, MetricType},
 };
 use std::{
+    any::TypeId,
     borrow::Cow,
     collections::{BTreeMap, HashMap},
     sync::{Arc, Mutex},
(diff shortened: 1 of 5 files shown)


opentelemetry-prometheus/src/lib.rs
```diff
@@ -197,7 +199,65 @@ struct CollectorInner {
...
+static GAUGE_TYPES: Lazy<[TypeId; 3]> = Lazy::new(|| {
+    [
+        TypeId::of::<data::Gauge<i64>>(),
+        TypeId::of::<data::Gauge<u64>>(),
+        TypeId::of::<data::Gauge<f64>>(),
+    ]
+});
+
 impl Collector {
+    fn metric_type_and_name(&self, m: &data::Metric) -> Option<(MetricType, Cow<'static, str>)> {
```
TommyCpp, line 227: Can we move this later when `add_*` functions? It may save us one downcast/runtime type check. I do see this version is easier to understand and less repetitive though but wonder if the performance wise if it make sense to avoid repeating downcast when the type is `Sum`

hdost (approved):

opentelemetry-prometheus/src/lib.rs
jtescher: We could, but I don't think the downcasting has any noticeable perf impact. Could do a benchmark but threads like https://users.rust-lang.org/t/downcasting-overhead/69718/6 make it pretty clear.

jtescher: Doesn't seem to make a difference when I run quick bench, can re-assess if perf becomes an issue later.

TommyCpp: Sg. Thanks for checking!

PR #1164, opened 2023-07-22
Install logger provider

pkoenig10:
## Changes

Install the logger provider when building the log pipeline.

Currently the logger provider is not installed and is immediately dropped at the end of the `build` method, which results in logs not being emitted.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1168, opened 2023-07-27
Add OTLP LogExporter to the example

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/1164#issuecomment-1652405946

Adds Logging to the OTLP examples, along with nit cleanups.

### DIFF
--- a/opentelemetry-otlp/examples/basic-otlp/Cargo.toml
+++ b/opentelemetry-otlp/examples/basic-otlp/Cargo.toml
@@ -7,8 +7,10 @@ publish = false
 
 [dependencies]
 once_cell = "1.17"
-opentelemetry_api = { path = "../../../opentelemetry-api", features = ["metrics"] }
-opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio"] }
-opentelemetry-otlp = { path = "../../../opentelemetry-otlp", features = ["tonic", "metrics"] }
+opentelemetry_api = { path = "../../../opentelemetry-api", features = ["metrics", "logs"] }
+opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "logs"] }
+opentelemetry-otlp = { path = "../../../opentelemetry-otlp", features = ["tonic", "metrics", "logs"] }
 opentelemetry-semantic-conventions = { path = "../../../opentelemetry-semantic-conventions" }
 tokio = { version = "1.0", features = ["full"] }
+opentelemetry-appender-log = { path = "../../../opentelemetry-appender-log"}
+log = {version = "0.4.17"}
--- a/opentelemetry-otlp/examples/basic-otlp/otel-collector-config.yaml
+++ b/opentelemetry-otlp/examples/basic-otlp/otel-collector-config.yaml
@@ -1,36 +1,27 @@
+# This is a configuration file for the OpenTelemetry Collector intended to be
+# used in conjunction with the opentelemetry-otlp example.
+#
+# For more information about the OpenTelemetry Collector see:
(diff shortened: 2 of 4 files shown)


opentelemetry-otlp/examples/basic-otlp/otel-collector-config.yaml
```diff
@@ -1,36 +1,27 @@
...
     protocols:
-      http:
       grpc:
+      http:
 
 exporters:
   logging:
     loglevel: debug
 
-  jaeger:
```
cijothomas, line 11: this was cleaned up as the purpose of this example is to show how to use OTLP Exporter. The config will print the telemetry to stdout which is good enough to confirm things working. Even if there is a need to use Jaeger to visualize, Jaeger can natively understand OTLP, so no need to futher export from collector.

opentelemetry-otlp/examples/basic-otlp/otel-collector-config.yaml
```diff
@@ -1,36 +1,27 @@
...
-
-extensions:
-  health_check:
-  pprof:
-    endpoint: :1888
-  zpages:
-    endpoint: :55679
-
 service:
-  extensions: [pprof, zpages, health_check]
```
cijothomas, line 27: removed everything unnecessary to keep the example config the the minimal.

opentelemetry-otlp/examples/basic-otlp/Cargo.toml
```diff
@@ -7,8 +7,10 @@ publish = false
...
 once_cell = "1.17"
-opentelemetry_api = { path = "../../../opentelemetry-api", features = ["metrics"] }
-opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio"] }
-opentelemetry-otlp = { path = "../../../opentelemetry-otlp", features = ["tonic", "metrics"] }
+opentelemetry_api = { path = "../../../opentelemetry-api", features = ["metrics", "logs"] }
+opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "logs"] }
+opentelemetry-otlp = { path = "../../../opentelemetry-otlp", features = ["tonic", "metrics", "logs"] }
 opentelemetry-semantic-conventions = { path = "../../../opentelemetry-semantic-conventions" }
 tokio = { version = "1.0", features = ["full"] }
+opentelemetry-appender-log = { path = "../../../opentelemetry-appender-log"}
```
lalitb, line 15: We are not using `opentelemetry-appender-tracing` here because of instrumentation loopback issue ?

cijothomas: yes. Will be opening an issue to track that.

lalitb (approved):

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -86,6 +124,8 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         );
         span.set_attribute(ANOTHER_KEY.string("yes"));
 
+        error!(target: "my-target", "hello from {}. My price is {}. I am also inside a Span!", "banana", 2.99);
```
lalitb, line 127: sorry for nitpicking,- It's not really an error message. Can be made as `info!` after changing the max level, or else fine to keep it as such.

opentelemetry-otlp/examples/basic-otlp/Cargo.toml
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/1171

jtescher (approved):
Looks good 👍

PR #1175, opened 2023-08-01
Set timestamp correctly in log appenders

cijothomas:
Log-Appender modified to set Timestamp, instead of ObservedTimestamp.
Tracing-Appender modified to set Timestamp (it was not setting any before).

## Changes

### DIFF
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -31,7 +31,7 @@ where
             self.logger.emit(
                 LogRecordBuilder::new()
                     .with_severity_number(map_severity_to_otel_severity(record.level()))
-                    .with_observed_timestamp(SystemTime::now())
+                    .with_timestamp(SystemTime::now())
                     .with_severity_text(record.level().as_str())
                     .with_body(AnyValue::from(record.args().to_string()))
                     .build(),
--- a/opentelemetry-appender-log/CHANGELOG.md
+++ b/opentelemetry-appender-log/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## vNext
+
+- Populate TimeStamp on LogRecord, instead of ObservedTimeStamp
+  [#1175](https://github.com/open-telemetry/opentelemetry-rust/pull/1175).
+
 ## v0.1.0
 
 Initial crate release
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## vNext
+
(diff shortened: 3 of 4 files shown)


opentelemetry-appender-log/src/lib.rs
```diff
@@ -31,7 +31,7 @@ where
             self.logger.emit(
                 LogRecordBuilder::new()
                     .with_severity_number(map_severity_to_otel_severity(record.level()))
-                    .with_observed_timestamp(SystemTime::now())
+                    .with_timestamp(SystemTime::now())
```
lalitb, line 34: I feel the `timestamp` should be populated by the logging framework, which is `tokio-tracing` here. And not the appender. It should remain empty if the logging framework is not generating it.

cijothomas: tokio-tracing itself does not populate it as it is merely a "facade", outsourcing most of the work to "subscribers". In this case, the subscriber *is* the also the log appender, so I think this is the right place to fetch timestamp.

cijothomas: https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L842 current tracing-opentelemetry impln. (its not for logs, but just sharing to show tracing crate itself is not in charge of timestamps)

lalitb: Yeah it's tricky. Log Appender shouldn't populate timestamp (as per otel specs), but subscriber could.
And opentelemetry-appender-log crate is both. I was thinking of scenario where the tracing subscriber is chained with multiple layers, and one of the intermediate layer is enriching the timestamp which appender can use. But probably not that frequent use-case.

But I don't have strong opinion here, should be good.

lalitb (approved):

opentelemetry-appender-log/src/lib.rs
cijothomas: Yes! 
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Logs/ILogger/OpenTelemetryLogger.cs#L91 Doing this in the OpenTelemetry bridge/appender is the only way ! This is from .NET appender. Its possible there are multiple subscribers for .NET as well, but everyone is responsible to set own timestamp!

cijothomas: Btw, Log spec does not forbid populating timestamp, so we good from that perspective.

jtescher (approved):

opentelemetry-appender-log/src/lib.rs
jtescher: This seems reasonable for now, we can get feedback on the design as people start integrating it as well

PR #1178, opened 2023-08-02
[Logs API/SDK] - Populate ObservedTimestamp within the SDK

lalitb:
## Changes

As per the log data model, `Observed Timestamp` should either be specified through bridge-api (by appender),  or else the SDK should set it to current time.

Refer: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md 

> [Observed Timestamp](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-observedtimestamp). If unspecified the implementation SHOULD set it equal to the current time.

As of now, this is an optional field in LogRecord, and SDK doesn't default it to current timestamp if not already set. Modifying the code to make it mandatory field with default value.

**otlp log exporter output before this change:**

```
LogRecord #0
ObservedTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 2023-08-02 06:04:50.0677901 +0000 UTC
SeverityText: INFO
SeverityNumber: Info(9)
Body: Str(hello from banana. My price is 2.99. I am also inside a Span!)
Trace ID: 32b8f0eb91e559d03b39be3004741ffe
Span ID: f471fb669a2a7a0e
Flags: 1
```
**otlp log exporter output after this change:**
```
LogRecord #0
ObservedTimestamp: 2023-08-02 06:13:50.1720326 +0000 UTC
Timestamp: 2023-08-02 06:13:50.1720323 +0000 UTC
SeverityText: INFO
SeverityNumber: Info(9)
Body: Str(hello from banana. My price is 2.99. I am also inside a Span!)
Trace ID: f722f369b03b1d46e987b5a00720064a
Span ID: 14f94c1be9e9f160
Flags: 1
```
  
### DIFF
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -7,14 +7,14 @@ use std::{borrow::Cow, time::SystemTime};
 #[derive(Debug, Clone)]
 /// Timestamp for when the record was observed by OpenTelemetry
 /// Defaults to the current time.
-pub struct ObservedTimeStamp {
+pub struct ObservedTimestamp {
     /// Time when the record was observed by OpenTelemetry
     pub time: SystemTime,
 }
 
-impl Default for ObservedTimeStamp {
+impl Default for ObservedTimestamp {
     fn default() -> Self {
-        ObservedTimeStamp {
+        ObservedTimestamp {
             time: SystemTime::now(),
         }
     }
@@ -29,7 +29,7 @@ pub struct LogRecord {
     pub timestamp: Option<SystemTime>,
 
     /// Timestamp for when the record was observed by OpenTelemetry
-    pub observed_timestamp: ObservedTimeStamp,
+    pub observed_timestamp: ObservedTimestamp,
 
     /// Trace context for logs associated with spans
     pub trace_context: Option<TraceContext>,
@@ -277,7 +277,7 @@ impl LogRecordBuilder {
     pub fn with_observed_timestamp(self, timestamp: SystemTime) -> Self {
         Self {
             record: LogRecord {
-                observed_timestamp: ObservedTimeStamp { time: timestamp },
+                observed_timestamp: ObservedTimestamp { time: timestamp },
                 ..self.record
             },
         }
--- a/opentelemetry-api/src/logs/record.rs
+++ b/opentelemetry-api/src/logs/record.rs
@@ -4,6 +4,22 @@ use crate::{
 };
 use std::{borrow::Cow, time::SystemTime};
 
+#[derive(Debug, Clone)]
+/// Timestamp for when the record was observed by OpenTelemetry
+/// Defaults to the current time.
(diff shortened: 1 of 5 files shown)


opentelemetry-api/src/logs/record.rs
```diff
@@ -4,6 +4,22 @@ use crate::{
 };
 use std::{borrow::Cow, time::SystemTime};
 
+#[derive(Debug, Clone)]
+/// Timestamp for when the record was observed by OpenTelemetry
+/// Defaults to the current time.
+pub struct ObservedTimestamp {
+    /// Time when the record was observed by OpenTelemetry
+    pub time: SystemTime,
+}
```
TommyCpp, line 13: Hmm do we expect to add more fields in this struct? If we do we should add `#[non_exhaustive]` attribute for this struct. If we don't why not just use `SystemTime` in the LogRecord and have the default there?

lalitb: Thanks, @TommyCpp. How can I add the default for `observed_timestamp` within LogRecord? This was what I was originally trying out, but seems not doable without explicitly defining the default trait for LogRecord.

lalitb: Ok. I have added the default trait for LogRecord. This was much simpler as rest of the field are optional, and so default to `None`. Thanks for the suggestion.

opentelemetry-api/src/logs/record.rs
```diff
@@ -30,6 +30,20 @@ pub struct LogRecord {
     pub attributes: Option<Vec<(Key, AnyValue)>>,
 }
 
+impl Default for LogRecord {
+    fn default() -> Self {
+        LogRecord {
+            timestamp: None,
+            observed_timestamp: SystemTime::now(),
```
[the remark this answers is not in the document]
TommyCpp: I wonder instead of using the `SystemTime` here I wonder if it makes sense to provide a generic method/struct to get current time here. And then we can allow users to provide a `dummy clock` and avoid the cost. It also makes testing easier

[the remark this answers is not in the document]
cijothomas: Discussed offline:
1. Log-Appenders should not populate TimeStamp, instead they should only do ObservedTimeStamp (and they don't really have to do it, as they can rely on OTel API/SDK to populate it automatically). This is because, log-appenders may not be the only layers, and depending on the number/type of other layerss, it may be impossible to obtain the "true" timestamp. Unfortunately, neither of the log crates (log, tracing), providers the true timestamp. So the best option is to modify the PR (https://github.com/open-telemetry/opentelemetry-rust/pull/1175) to remove populating timestamp from the appenders.
2. with the above point, there is no longer the perf concern of reading ::now() twice.

cijothomas (approved):

TommyCpp (approved):

PR #1185, opened 2023-08-03
datadog: do not set invalid remote context

mightyguava:
## Changes

When no remote span can be extracted, the datadog propagator will set the active span to an empty span. This seems to cause child spans to use the empty span's trace ID of 0, breaking propagation of new traces. The datadog agent seems to be ok with getting these traces, adding a trace ID itself.

### DIFF
--- a/opentelemetry-datadog/CHANGELOG.md
+++ b/opentelemetry-datadog/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## v0.9.0
+## vNext
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


opentelemetry-datadog/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.9.0
```
jtescher, line 3:
```suggestion
## vNext
```
nit: leave this as next or unreleased in case this is a patch release fix

jtescher:
Thanks @mightyguava, good catch

PR #1187, opened 2023-08-05
Restructure OTLP config

jtescher:
This restructuring makes it easier to create exporters from a given client builder.

## Changes

Adding build methods for each exporter alows for much simpler configuration for users that want to use multiple exporters or configure the exporter directly (e.g. adding views to meter provider config).

example new methods for tonic builder:

```rust
// Create a span exporter you can use to when configuring tracer providers
let span_exporter = opentelemetry_otlp::new_exporter().tonic().build_span_exporter()?;

// Create a metrics exporter you can use when configuring meter providers
let metrics_exporter = opentelemetry_otlp::new_exporter()
    .tonic()
    .build_metrics_exporter(
        Box::new(DefaultAggregationSelector::new()),
        Box::new(DefaultTemporalitySelector::new()),
    )?;

// Create a log exporter you can use when configuring logger providers
let log_exporter = opentelemetry_otlp::new_exporter().tonic().build_log_exporter()?;
```

In order to add these, the client construction logic was moved internally, but the logic and external interfaces should remain the same.


TommyCpp (approved):
Looks good!

PR #1188, opened 2023-08-07
fix: allow user to set arbitrary units in promethus suffix.

### LINKED ISSUE #1173: `Unit`s don't appear in metric names -- unless you use the right unit name
I'm updating to v0.20 and ran into a bit of unexpected behavior / an issue with the API with metrics that are exported to Prometheus.

I have a histogram that uses seconds:
```rust
global::meter("example")
  .f64_histogram("my_histogram")
  .with_unit(Unit::new("seconds"))
  .init()
```

The default behavior says that the units should be included, so the metric should be `my_histogram_seconds_bucket`, etc. However, with this unit set, it appears as `my_histogram_bucket`.

In looking for the issue, I found this function, which determines whether the unit is one of the expected names and completely ignores it if not. https://github.com/open-telemetry/opentelemetry-rust/blob/dfeac078ff7853e7dc814778524b93470dfa5c9c/opentelemetry-prometheus/src/utils.rs#L6-L40

If there is a set of unit names that will be respected, these should really be represented as enum variants as opposed to having a hidden function that looks for specific strings. As a user, it's not obvious and not mentioned in the struct docs that `"s"` should be used instead of `"seconds"`.

TommyCpp:
Fixes #1173

## Changes

Before this change, if user-specific the full name of the unit. It will not be added to the suffix of the result metrics.
After this change, most of the user-specific unit will be added to the result metrics with the exception of any unit containing `/`.

### DIFF
--- a/opentelemetry-prometheus/CHANGELOG.md
+++ b/opentelemetry-prometheus/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+### Changed
+- allow custom units in prometheus suffix [#1188](https://github.com/open-telemetry/opentelemetry-rust/pull/1188)
+
 ## v0.13.0
 
 ### Added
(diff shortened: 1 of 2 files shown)


jtescher:
Changelog entry?

PR #1202, opened 2023-08-11
Add `grpcio` metrics exporter

jtescher:
## Changes

This adds missing `grpcio` support for exporting metrics in OTLP.

Instead of porting the metrics transformations to the unsupported `protobuf` v2.x impl (v3 is not supported by grpcio), this switches grpcio to use prost as well, this simplifies and unifies the proto transformations to the respective `tonic` and `grpcio` ecosystems.


opentelemetry-proto/tests/grpc_build.rs
```diff
@@ -97,6 +89,8 @@ fn ensure_files_are_same(
     after_build: HashMap<String, String>,
     target_dir: &'static str,
 ) {
+    dbg!(&before_build.keys());
+    dbg!(&after_build.keys());
```
TommyCpp, line 93: Do we still need those?

opentelemetry-otlp/src/metric.rs
```diff
@@ -80,10 +91,23 @@ impl MetricsExporterBuilder {
...
             MetricsExporterBuilder::Http(builder) => {
                 builder.build_metrics_exporter(aggregation_selector, temporality_selector)
             }
+
+            #[cfg(not(any(feature = "http-proto", feature = "grpc-sys", feature = "grpc-tonic")))]
+            MetricsExporterBuilder::Unconfigured => {
+                drop(temporality_selector);
+                drop(aggregation_selector);
+                Err(opentelemetry_api::metrics::MetricsError::Other(
+                    "no configured span exporter".into(),
```
TommyCpp, line 108:
```suggestion
                    "no configured metrics exporter, enable `http-proto`, `grpc-sys` or `grpc-tonic` feature to configure a metrics exporter".into(),
```

TommyCpp (approved):
Mostly LGTM. Small nits

opentelemetry-proto/tests/grpc_build.rs
jtescher: ah no, good catch

PR #1203, opened 2023-08-11
bump MSRV to 1.64 and indexmap to 2.0

shaun-cox:
Due to requirement of #1199, this is a separate PR to bump MSRV.
Contributing factors:
- `tokio` now has MSRV of 1.63
- we're using older `indexmap 1.18`, and `indexmap 2.0` has MSRV of 1.64

## Changes
- bump MSRV to 1.64
- bump `indexmap` to 2.0
- fix newly exposed lint about deriving `Default` instead of manually implementing

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ analysis in order to understand your software's performance and behavior. You
 can export and analyze them using [Prometheus], [Jaeger], and other
 observability tools.
 
-*Compiler support: [requires `rustc` 1.60+][msrv]*
+*Compiler support: [requires `rustc` 1.64+][msrv]*
 
 [Prometheus]: https://prometheus.io
 [Jaeger]: https://www.jaegertracing.io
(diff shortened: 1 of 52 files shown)


TommyCpp (approved):
LGTM Thanks!

README.md
```diff
@@ -25,7 +25,7 @@ analysis in order to understand your software's performance and behavior. You
 can export and analyze them using [Prometheus], [Jaeger], and other
 observability tools.
 
-*Compiler support: [requires `rustc` 1.60+][msrv]*
+*Compiler support: [requires `rustc` 1.64+][msrv]*
```
lalitb, line 28: Are there any implications of incrementing MSRV? Specifically, I'm concerned about the impact on applications that utilize our SDKs and are currently on a lower toolchain. Upgrading would mean enforcing them to use a higher compiler version. There could be features deprecated/removed in higher versions, and the applications still be using them.

Probably, I am thinking this with a C++ mindset There, we often strive to support the earliest versions across various compilers. For example, for C++11 support, we make sure that we can build successfully with the lowest supported GCC version, such as gcc-4.8. It seems that upgrades are more frequent in the Rust ecosystem. :)

jtescher: I think it's ok to update as tokio updated to 1.63+ in https://github.com/tokio-rs/tokio/pull/5887, this project is less foundational to the rust ecosystem than they are.

lalitb: Yes, that makes sense.

shaun-cox:
Will be updating CHANGELOG.md's, then readying this for review/approval...

djc (approved):

lalitb (approved):

cijothomas (approved):

djc:
Thanks for taking care of this.

PR #1204, opened 2023-08-11
Add stale PR workflow

cijothomas:
Trying to keep inactive PRs get auto-closed.
Similar to Spec repo and other repos:
 https://github.com/open-telemetry/opentelemetry-specification/blob/main/.github/workflows/stale-pr.yaml.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,21 @@
+# Syntax: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+# Github Actions Stale: https://github.com/actions/stale
+
+name: "Close stale pull requests"
+on:
+  schedule:
+    - cron: "12 3 * * *"  # arbitrary time not to DDOS GitHub
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-pr-message: 'This PR was marked stale due to lack of activity and will be closed in 7 days. Commenting or Pushing will instruct the bot to automatically remove the label. This bot runs once per day.'
+          close-pr-message: 'Closed as inactive. Feel free to reopen if this PR is still being worked on.'
+          operations-per-run: 400
+          days-before-pr-stale: 7
+          days-before-issue-stale: -1
+          days-before-pr-close: 7
+          days-before-issue-close: -1


TommyCpp (approved):

TommyCpp:
Is there a way we can mark a PR to be excluded from this flow(maybe a label for the PR?). Some PRs have long discussion and probably should remain open.

cijothomas:
> Is there a way we can mark a PR to be excluded from this flow(maybe a label for the PR?). Some PRs have long discussion and probably should remain open.

if a PR *has* discussions/commits, then it will be ignored by this workflow. It only targets really stale PRs with no new commits or comments!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-pr-message: 'This PR was marked stale due to lack of activity and will be closed in 7 days. Commenting or Pushing will instruct the bot to automatically remove the label. This bot runs once per day.'
+          close-pr-message: 'Closed as inactive. Feel free to reopen if this PR is still being worked on.'
+          operations-per-run: 400
+          days-before-pr-stale: 7
+          days-before-issue-stale: -1
+          days-before-pr-close: 7
+          days-before-issue-close: -1
```
lalitb, line 21: nit - should also add an exception not to stale the PR if labeled appropriately.
Say:
`exempt-pr-labels: "do-not-stale"`

lalitb (approved):

djc:
I'm not a fan of stuff like this. If we want to close stale PRs it doesn't seem too onerous to do so with a handwritten comment that properly takes the context into account. It's not like we have a lot of volume in this repo anyway.

lalitb:
> I'm not a fan of stuff like this. If we want to close stale PRs it doesn't seem too onerous to do so with a handwritten comment that properly takes the context into account. It's not like we have a lot of volume in this repo anyway.

Good point. @cijothomas  - Any reason not to instead have it ONLY for issues as they are in general large in volume, and difficult to manage? During triage, we can also add the appropriate label for issues not to be staled. We do the same in ote-cpp, but I think you mentioned some concerns to do it during the last community meeting.

djc:
I also think the volume for issues is not that bad and we can deal with it manually once in a while.

hdost:
> I also think the volume for issues is not that bad and we can deal with it manually once in a while.

On this point it's good to have a process before automating it. We can strive to get through PRs and issues, settle on something which works and then if it starts to grow worry about this.

cijothomas:
Given there is no pressing need for this, and the volume is manageable to deal with manually, will close this. 
We can start with manually checking with current stale PR owners, and see if they can be unblocked or considered abandoned.

cijothomas:
_I think you mentioned some concerns to do it during the last community meeting_

In Spec SIG some time back, it was decided to use stale automation only for PRs, not issues.. Don't recollect fully, but that was what I was referring to as "concerns" !

PR #1210, opened 2023-08-17
fix(otlp): endpoint urls for otlp http exporter.

TommyCpp:
Follow up on #1187 

## Changes
- removed the `with_env` method as we will always check env var before creating exporters.
- append the signal path to `OTEL_EXPORTER_OTLP_ENDPOINT` per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp)

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -144,14 +144,11 @@ impl HttpExporterBuilder {
         signal_endpoint_path: &str,
         signal_timeout_var: &str,
     ) -> Result<OtlpHttpClient, crate::Error> {
-        let endpoint = match env::var(signal_endpoint_var)
-            .ok()
-            .or(env::var(OTEL_EXPORTER_OTLP_ENDPOINT).ok())
-            .and_then(|s| s.parse().ok())
-        {
-            Some(val) => val,
-            None => format!("{}{signal_endpoint_path}", self.exporter_config.endpoint).parse()?,
-        };
+        let endpoint = resolve_endpoint(
+            signal_endpoint_var,
+            signal_endpoint_path,
+            self.exporter_config.endpoint.as_str(),
+        )?;
 
         let timeout = match env::var(signal_timeout_var)
             .ok()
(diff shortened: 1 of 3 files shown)


TommyCpp:
We currently prioritize the env vars over the endpoint provided by users using `with_endpoint`, which may be counter-intuitive. 

In order to prioritize the value provided by `with_endpoint` function we will need to make the fields in `ExportConfig` to be optional. Then we can resolve the endpoint if it's `Some`, and check env var if it's `None`, default endpoint will be used as the fall back

cijothomas:
> > We currently prioritize the env vars over the endpoint provided by users using `with_endpoint`, which may be counter-intuitive.
> 
> Doesn't seem counterintuitive to me. Prioritizing runtime environment over compile-time environment seems pretty normal?

This is a widely debated topic with no clarification from spec I think : https://github.com/open-telemetry/opentelemetry-specification/issues/1171

In OpenTelemetry .NET, configuration provided in code wins over Env_vars. Its possible some languages have designed it to prefer env variables.

TommyCpp:
> Doesn't seem counterintuitive to me. Prioritizing runtime environment over compile-time environment seems pretty normal?

Hmm I think it maybe confusing for some people if they provided a value expclitly in code but end up finding their trace in a different place because of the env var override. 

I don't have strong preference here but I think given the confusion maybe we should discuss it as a tenet, document it in crate documentaion and contribution guide, and checks all create to make sure we are consistent across our crates?

cijothomas:
> > Doesn't seem counterintuitive to me. Prioritizing runtime environment over compile-time environment seems pretty normal?
> 
> Hmm I think it maybe confusing for some people if they provided a value expclitly in code but end up finding their trace in a different place because of the env var override.
> 
> I don't have strong preference here but I think given the confusion maybe we should discuss it as a tenet, document it in crate documentaion and contribution guide, and checks all create to make sure we are consistent across our crates?

My preference is treating anything explicitly specified in code higher in priority than env vars. If users want the operators to determine endpoint, then they should not configure it in code, so operators can decide which endpoint to use.

TommyCpp:
Feel like the priority of the env vs compiling time value is a larger topic, will start a separate issue for that

PR #1212, opened 2023-08-17
Pretty print stdout in examples

cijothomas:
I tend to use this a lot, not sure if this we should make the default in examples to do this. This PR adds a comment about how to get pretty print easily.


TommyCpp (approved):
Looks good. Can merge it once we get the lint error out of the way

Issue #1211, opened 2023-08-17
Create tests for metrics api


lzchen:
We currently do not have much/any tests for metrics api surface. Creating a tracking issue here.

lzchen:
@TommyCpp @jtescher 

Can I be assigned to this?

PR #1213, opened 2023-08-17
Add doc tests to opentelemetry metrics api

lzchen:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1211


TommyCpp (approved):
👍 LGTM but I wonder do we want to keep #1211 open to track the work to add more unit tests

PR #1226, opened 2023-08-24
use opentelemetry instead of opentelemetry_api

### LINKED ISSUE #1186: What is the difference between the API and SDK?
As a user of `opentelemetry`, there are quite a few terms to learn when navigating the various packages to achieve some goal. 

One of the things I've found confusing is the separation between the `opentelemetry_api` and `opentelemetry_sdk`. As an outsider that may be missing some project-specific context, those two terms sound like they ought to be interchangeable. My general understanding of SDKs are that they're libraries or sets of libraries that help you build against or on top of APIs. I don't understand why they're separate in OpenTelemetry world and it always feels somewhat random to me what things I need to import from the API or SDK crate.

The OpenTelemetry [glossary](https://opentelemetry.io/docs/concepts/glossary/) doesn't help that much:

> Application Programming Interface. In the OpenTelemetry project, used to define how telemetry data is generated per [Data Source](https://opentelemetry.io/docs/concepts/glossary/#data-source).

> Short for Software Development Kit. Refers to a telemetry SDK that denotes a [Library](https://opentelemetry.io/docs/concepts/glossary/#library) that implement the OpenTelemetry [API](https://opentelemetry.io/docs/concepts/glossary/#api).

Could this be clarified further in the docs or (I assume not) could the crates just be combined?

shaun-cox:
Fixes #1186

## Changes

As part of #1199, @djc asked:
> At this point, should we just move all of the opentelemetry_api code back into the opentelemetry directory?

This is the first step in that process, to make reviewing easier. The change here is just to:
- change Cargo.toml dependencies from `opentelemetry_api` to `opentelemetry`
- replace `use opentelemetry_api::` with `use opentelemetry::`

A subsequent PR will complete @djc's request by renaming `opentelemetry-api` directory to `opentelemetry` and removing the latter stub.


opentelemetry-api/README.md
```diff
@@ -1,28 +0,0 @@
-![OpenTelemetry — An observability framework for cloud-native software.][splash]
```
cijothomas, line 1: it may be a good idea to to keep this file, with a note saying "its no longer maintained and users should use opentelemetry crate instead". Or maybe that can be done in crates.io description manually.

opentelemetry/README.md
```diff
@@ -58,11 +58,12 @@ See the [examples](./examples) directory for different integration patterns.
...
 
-In addition to `opentelemetry`, the [`open-telemetry/opentelemetry-rust`]
-repository contains several additional crates designed to be used with the
-`opentelemetry` ecosystem. This includes a collection of trace `SpanExporter`
-and metrics pull and push controller implementations, as well as utility and
-adapter crates to assist in propagating state and instrumenting applications.
+In addition to `opentelemetry`, which only carries the API, the
+[`open-telemetry/opentelemetry-rust`] repository contains several additional
+crates designed to be used with the `opentelemetry` ecosystem. This includes a
+collection of trace `SpanExporter` and metrics pull and push controller
```
cijothomas, line 64: nit: this is outdated. will send a follow up to address it after this PR.

cijothomas:
> commit 3
> updates CHANGELOG.md files to remove mentions of opentelemetry-api and just use opentelemetry
> (maybe this isn't appropriate, in which case I'll revert.)

mm.. I don't think we need to update past changelogs.

opentelemetry-otlp/Cargo.toml
```diff
@@ -30,7 +30,7 @@ async-trait = "0.1"
 futures-core = "0.3"
 opentelemetry-proto = { version = "0.3", path = "../opentelemetry-proto", default-features = false }
 grpcio = { version = "0.12", optional = true }
-opentelemetry_api = { version = "0.20", default-features = false, path = "../opentelemetry-api" }
+opentelemetry = { version = "0.21", default-features = false, path = "../opentelemetry" }
```
cijothomas, line 33: nit: we can add this to changelog of OTLP to say something like "Changed dependency from opentelemetry_api to opentelemetry as the latter is now the API crate".

cijothomas (approved):
LGTM. I left a couple of nits, and also do not think we need commit3 to update old changelog entries.

Thanks a lot for this, and really appreciate the way step-by-step commits is done to make reviewing much easier!

lalitb:
> > commit 3
> > updates CHANGELOG.md files to remove mentions of opentelemetry-api and just use opentelemetry
> > (maybe this isn't appropriate, in which case I'll revert.)
> 
> mm.. I don't think we need to update past changelogs.

Same thought. Do we need to update old changelogs entries ?

opentelemetry/README.md
```diff
@@ -58,11 +58,12 @@ See the [examples](./examples) directory for different integration patterns.
 
 ### Related Crates
 
-In addition to `opentelemetry`, the [`open-telemetry/opentelemetry-rust`]
-repository contains several additional crates designed to be used with the
-`opentelemetry` ecosystem. This includes a collection of trace `SpanExporter`
-and metrics pull and push controller implementations, as well as utility and
-adapter crates to assist in propagating state and instrumenting applications.
+In addition to `opentelemetry`, which only carries the API, the
```
lalitb, line 61: Just a thought, if opentelemetry/README.md should now only contain the information specific to API. Most of the content is similar to the root level README.md, and can be removed. Probably something to be addressed once this PR is merged.

lalitb (approved):
Nicely done. Thanks.

opentelemetry-otlp/Cargo.toml
shaun-cox: Done.

opentelemetry-api/README.md
TommyCpp: +1 we will need to figure out a way to prompt users to `opentelemetry` crate when they end up on the `opentelemetry_api` crate

hdost (approved):
Glad people are finally on board with this.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 ### Changed
 
 - Bump MSRV to 1.64 [#1203](https://github.com/open-telemetry/opentelemetry-rust/pull/1203)
+- Changed dependency from `opentelemetry_api` to `opentelemetry` as the latter
```
hdost, line 13: Perhaps we want to add this to some of the other crates as well.

opentelemetry/Cargo.toml
```diff
@@ -21,16 +21,25 @@ all-features = true
...
-trace = ["opentelemetry_api/trace"]
-metrics = ["opentelemetry_api/metrics"]
-logs = ["opentelemetry_api/logs"]
-logs_level_enabled = ["logs", "opentelemetry_api/logs_level_enabled"]
-testing = ["opentelemetry_api/testing"]
+trace = ["pin-project-lite"]
+metrics = []
+testing = ["trace", "metrics"]
+logs = []
+logs_level_enabled = ["logs"]
```
djc, line 45: Nit: add the final newline here?

djc (approved):
LGTM, thank you for working through this -- must have been pretty tedious.

opentelemetry-api/README.md
djc: Could submit an informational RustSec security advisory setting the status for opentelemetry_api to unmaintained.

jayvdb: I've submitted https://github.com/rustsec/advisory-db/issues/1994 and a PR

PR #1237, opened 2023-08-30
Update minimal versions

MidasLamb:
I was playing around with `cargo +nightly update -Z minimal versions` and I noticed that there were some errors in `opentelemetry-sdk`. 

I've found the minimal version that are required to build a simple test project that depends on `opentelemetry-sdk` with feature `rt-tokio`.

The commits themselves contain a link to the relevant release/changelogs.

I'm not sure if this is worthy of mentioning in the changelog here?


djc (approved):
Probably not worth mentioning in the changelog, but thanks for doing the legwork on this!

PR #1240, opened 2023-08-31
Add Sync bound on the LogExporter and SpanExporter traits

FSMaxB:
## Changes

Previously, the `opentelemetry_otlp::LogExporter` and `opentelemetry_otlp::SpanExporter` used `Box<dyn ...Expoter>` without ~~`Send` and~~ `Sync` bounds, making them impossible to use on a multi-threaded tokio runtime. This adds these ~~`Send` and~~ `Sync` bounds.

Note that this is technically a breaking change because previously the `LogExporter::new` or `SpanExpoter::new` constructors could be used with a custom non-threadsafe implementation of the `LogExporter` and `SpanExporter` traits. I can add a changelog entry if that is considered big-enough of a user-facing change.

Other note about consistency: The `MetricsClient` already has both `Send` and `Sync` bounds (through the traits themselves). 

### DIFF
--- a/opentelemetry-sdk/src/export/logs/mod.rs
+++ b/opentelemetry-sdk/src/export/logs/mod.rs
@@ -11,7 +11,7 @@ use std::{borrow::Cow, fmt::Debug};
 
 /// `LogExporter` defines the interface that log exporters should implement.
 #[async_trait]
-pub trait LogExporter: Send + Debug {
+pub trait LogExporter: Send + Sync + Debug {
     /// Exports a batch of [`LogData`].
     async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()>;
     /// Shuts down the exporter.
--- a/opentelemetry-sdk/src/export/trace.rs
+++ b/opentelemetry-sdk/src/export/trace.rs
@@ -16,7 +16,7 @@ pub type ExportResult = Result<(), TraceError>;
 /// The goal of the interface is to minimize burden of implementation for
 /// protocol-dependent telemetry exporters. The protocol exporter is expected to
 /// be primarily a simple telemetry data encoder and transmitter.
-pub trait SpanExporter: Send + Debug {
+pub trait SpanExporter: Send + Sync + Debug {
     /// Exports a batch of readable spans. Protocol exporters that will
     /// implement this function are typically expected to serialize and transmit
     /// the data to the destination.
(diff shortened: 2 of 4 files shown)


FSMaxB:
Sorry, forgot `cargo fmt`

djc:
Should we add these as super trait bounds instead?

FSMaxB:
> Should we add these as super trait bounds instead?

You mean add it to the `LogExpoter` and `SpanExporter` traits from `opentelemetry_sdk` directly? Sounds good to me if it is ok to do so.

Also I just noticed that they require `Send` already, so I can remove the explicit `Send` bounds from my PR.

FSMaxB:
> Yeah, if we have the Send bounds already it probably makes sense to add the Sync bounds to the trait too.

Will do!

FSMaxB:
Forgot to update the tests.

FSMaxB:
:thinking: Actually the issue seems to be that the `Sync` bound on `std::mpsc::Sender` was added on Rust 1.72.0, so the MSRV changes. I'll see how I can work around that, since it's just test-code that is affected.

djc:
Okay, one approach I've started taking in other crates is to only use `cargo check --lib --all-features` for MSRV jobs in CI. This avoids MSRV changes from dev-dependencies (and test code). Maybe that would need a separate PR, though.

FSMaxB:
I'm just switching from `std::mpsc` to `crossbeam_channel` for now. It's already a dependency anyways.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -11,7 +11,7 @@ use std::{borrow::Cow, fmt::Debug};
 
 /// `LogExporter` defines the interface that log exporters should implement.
 #[async_trait]
-pub trait LogExporter: Send + Debug {
+pub trait LogExporter: Send + Sync + Debug {
```
cijothomas, line 14: Please add an entry to changelog.md file.

opentelemetry-sdk/src/export/trace.rs
```diff
@@ -16,7 +16,7 @@ pub type ExportResult = Result<(), TraceError>;
 /// The goal of the interface is to minimize burden of implementation for
 /// protocol-dependent telemetry exporters. The protocol exporter is expected to
 /// be primarily a simple telemetry data encoder and transmitter.
-pub trait SpanExporter: Send + Debug {
+pub trait SpanExporter: Send + Sync + Debug {
```
cijothomas, line 19: Please add an entry to changelog.md file.

opentelemetry-sdk/src/export/logs/mod.rs
FSMaxB: Done!

opentelemetry-sdk/src/export/trace.rs
FSMaxB: Done

shaun-cox (approved):

FSMaxB:
Rebased to fix conflicts.

jtescher (approved):

PR #1246, opened 2023-09-01
Refactor LogRecordBuilder with in-place modifications

lalitb:
## Changes

Opening this PR for discussion, as I am not fully sure if the proposed approach has any drawbacks.

This PR refactors the `LogRecordBuilder` struct methods to modify the `LogRecord` object in place. All methods that previously consumed `self` and returned a new `Self` instance have been updated to operate in-place by taking a `&mut self` parameter.

So.
```rust
    pub fn with_timestamp(self, timestamp: SystemTime) -> Self {
        Self {
            record: LogRecord {
                timestamp: Some(timestamp),
                ..self.record
            },
        }
    }
```

is changed to 

```rust
    pub fn with_timestamp(&mut self, timestamp: SystemTime) -> &mut Self {
        self.record.timestamp = Some(timestamp);
        self
   }
```
The motivation behind this change 

- is to improve the performance of building LogRecord by avoiding object creation and memory allocation within these methods. The in-place modification provides a more efficient way to update LogRecordBuilder. While the existing approach creates the new LogRecord on the stack, and only does the shallow copy, there are still improved perf numbers for `opentelemetry-appender-log` with the above changes when the properties are substantial.

-  allow using `LogRecordBuilder` for `opentelemetry-appender-tracing` where method chaining is not possible. There won't be perf improvement with this change as direct updating LogRecord is always much faster. However, I didn't see performance degrade either, and this allows us to add abstraction over the internal implementation of the LogRecord structure.

The downside is it is not consistent with [SpanBuilder](https://github.com/open-telemetry/opentelemetry-rust/blob/b6e91a9baf4474571b92693ec87f65835eb45385/opentelemetry/src/trace/tracer.rs#L279) which returns and uses `Self` instant, instead of `&mut Self`. Feel free to comment if this approach doesn't align with idiomatic Rust practices.

### DIFF
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -17,12 +17,12 @@ where
     L: Logger + Send + Sync,
 {
     fn enabled(&self, _metadata: &Metadata) -> bool {
-        //#[cfg(feature = "logs_level_enabled")]
-        //return self.logger.event_enabled(
-        //    map_severity_to_otel_severity(_metadata.level()),
-        //    _metadata.target(),
-        // );
-        // #[cfg(not(feature = "logs_level_enabled"))]
+        #[cfg(feature = "logs_level_enabled")]
+        return self.logger.event_enabled(
+            map_severity_to_otel_severity(_metadata.level()),
+            _metadata.target(),
+        );
+        #[cfg(not(feature = "logs_level_enabled"))]
         true
     }
 
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -17,26 +17,26 @@ where
     L: Logger + Send + Sync,
 {
     fn enabled(&self, _metadata: &Metadata) -> bool {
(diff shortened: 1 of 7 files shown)


TommyCpp:
I think one downside of mutable reference, at least for `SpanBuilder` is when working with function that requires `SpanBuilder`. User will need to assign the `SpanBuilder` somewhere instead of creating it when calling the function

lalitb:
> I think one downside of mutable reference, at least for SpanBuilder is when working with function that requires SpanBuilder. User will need to assign the SpanBuilder somewhere instead of creating it when calling the function

Ok yes, we can't create the `SpanBuilder` inline as part of the function call with this new design. Regarding the two points mentioned for motivation of the change, I tested the performance for `opentelemetry-appender-log`  once again, and didn't see any apparent improvement this time. A small improvement earlier was probably due to different loads on the system during these tests. Would still want to validate if the existing `LogRecordBuilder` (instead of `LogRecord`) can be used in  `opentelemetry-appender-tracing` - and if so, we are good to close this PR.

PR #1250, opened 2023-09-06
Use edition 2021 resolver

shaun-cox:
Eliminates the following warning when doing cargo update:
```bash
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

For more information, refer to [Cargo docs](https://doc.rust-lang.org/cargo/reference/features.html#feature-resolver-version-2)

## Changes

- add `workspace.resolver = "2"` to virtual workspace `Cargo.toml` file

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ members = [
     "examples/tracing-grpc",
     "stress",
 ]
+resolver = "2"
 
 [profile.bench]
 # https://doc.rust-lang.org/cargo/reference/profiles.html#bench


djc (approved):

Cargo.toml
```diff
@@ -32,9 +32,10 @@ members = [
...
+resolver = "2"
 
 [profile.bench]
-# https://doc.rust-lang.org/cargo/reference/profiles.html#bench
+# https://doc.rust-lang.org/cargo/reference/profiles.html#debug
 # See function names in profiling reports.
-# 2/true is too much, 0 is not enough, 1 is just right for back traces
-debug = 1
+# full is too much, none is not enough, limited is just right for back traces
+debug = "limited"
```
djc, line 41: Looks like our MSRV doesn't yet support this.

djc (approved):

Cargo.toml
shaun-cox: That's what I get for trying to change something small but not related to the real purpose of the PR. ;)

djc: Indeed, and in the same commit no less!

PR #1259, opened 2023-09-08
Nit updates to OTLP Example readme

cijothomas:
Mostly updating the example output as the current one was quite old.


jtescher (approved):

jtescher:
Thanks @cijothomas

PR #1266, opened 2023-09-13
Re-use memory via aggregate functions

jtescher:
## Changes

This simplifies memory reuse on subsequent collection cycles.

Also small ergonomics changes:

* Move `MetricsProducer` config to builders to match other config
* Remove the need for separate `Delta*` and `Cumulative*` aggregate types
* Return error earlier if readers are shut down
* Log warning if two instruments have the same name with different casing
* Log warning if view is created with empty criteria


opentelemetry-prometheus/src/config.rs
```diff
@@ -108,17 +110,22 @@ impl ExporterBuilder {
     ///
     /// [DefaultAggregationSelector]: opentelemetry_sdk::metrics::reader::DefaultAggregationSelector
     pub fn with_aggregation_selector(mut self, agg: impl AggregationSelector + 'static) -> Self {
-        self.aggregation = Some(Box::new(agg));
+        self.reader = self.reader.with_aggregation_selector(agg);
+        self
+    }
+
+    /// Registers a an external [MetricProducer] with this reader.
```
shaun-cox, line 117: nit: an extra `a ` before `an `

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -218,52 +217,57 @@ impl Stream {
     }
 }
 
-/// the identifying properties of a stream.
+/// The identifying properties of a instrument.
```
shaun-cox, line 220: nit: maybe prefer `an instrument` to `a instrument`?

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -218,52 +217,57 @@ impl Stream {
...
 #[derive(Debug, PartialEq, Eq, Hash)]
-pub(crate) struct StreamId {
+pub(crate) struct InstId {
     /// The human-readable identifier of the stream.
```
shaun-cox, line 223: `of the stream` to `of the instrument`?

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -218,52 +217,57 @@ impl Stream {
     }
 }
 
-/// the identifying properties of a stream.
+/// The identifying properties of a instrument.
 #[derive(Debug, PartialEq, Eq, Hash)]
-pub(crate) struct StreamId {
+pub(crate) struct InstId {
```
shaun-cox, line 222: consider `InstrumentId` to `InstId`?  (Though maybe the spec calls for this name, I didn't look.)

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -218,52 +217,57 @@ impl Stream {
...
-    /// The stream uses for an instrument.
-    pub(crate) aggregation: String,
-    /// Monotonic is the monotonicity of an instruments data type. This field is
-    /// not used for all data types, so a zero value needs to be understood in the
-    /// context of Aggregation.
-    pub(crate) monotonic: bool,
-    /// Temporality is the temporality of a stream's data type. This field is
-    /// not used by some data types.
-    pub(crate) temporality: Option<Temporality>,
     /// Number is the number type of the stream.
```
shaun-cox, line 231: another `stream` reference above which should be `instrument`.  (trying to be helpful, not picky. ;))

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -218,52 +217,57 @@ impl Stream {
...
 
+impl InstId {
+    /// Instrument names are considered case-insensitive.
+    ///
+    /// Standardize the instrument name to always be lowercase so it can be compared
+    /// via hash.
+    pub(crate) fn normalize(&mut self) {
+        if self.name.chars().any(|c| c.is_ascii_uppercase()) {
+            self.name = self.name.to_lowercase().into();
+        }
```
shaun-cox, line 243: Curious why `is_ascii_uppercase` was chosen to check for need to call `to_lowercase` as opposed to [`is_uppercase`](https://doc.rust-lang.org/std/primitive.char.html#method.is_uppercase)?

It occurs to me there could be a strange behavior where a name that contains both ASCII uppercase and other Unicode uppercase characters gets all lower-cased, but a name that only contains Unicode uppercase characters does not...

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -0,0 +1,185 @@
...
+    last_value::LastValue,
+    sum::{PrecomputedSum, Sum},
+    Number,
+};
+
+const STREAM_CARDINALITY_LIMIT: u32 = 2000;
+pub(crate) static STREAM_OVERFLOW_ATTRIBUTE_SET: Lazy<AttributeSet> = Lazy::new(|| {
+    let key_values: [KeyValue; 1] = [KeyValue::new("otel.metric.overflow", "true")];
+    AttributeSet::from(&key_values[..])
+});
```
shaun-cox, line 22: I get that its convenient/simple to represent a set of one attribute as an `AttributeSet`, but internally, that is a `BTreeSet`, which seems a bit heavy weight to carry around one constant attribute.  So, I guess I'm wondering out loud, is a lazily created static `AttributeSet` the only way to represent this, or could a `const KeyValue` be used implementing whatever trait might be used to access/enumerate the set?  (This is a question for my own education, rather than any objection to the PR.)

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -50,424 +50,304 @@ impl<T: Number<T>> Aggregator<T> for ValueMap<T> {
...
+        let s_data = s_data.unwrap_or_else(|| new_agg.as_mut().expect("present if s_data is none"));
+        s_data.temporality = Temporality::Cumulative;
+        s_data.is_monotonic = self.monotonic;
 
-    fn aggregation(&self) -> Option<Box<dyn Aggregation>> {
         let values = match self.value_map.values.lock() {
-            Ok(guard) => guard,
-            Err(_) => return None,
+            Ok(v) => v,
+            Err(_) => return (0, None),
```
shaun-cox, line 160: nit: The above feels like an open-coded [`Result::ok()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok)?

shaun-cox (approved):

opentelemetry-prometheus/src/config.rs
jtescher: thanks

opentelemetry-sdk/src/metrics/instrument.rs
jtescher: thanks

opentelemetry-sdk/src/metrics/instrument.rs
jtescher: Seems reasonable 👍

opentelemetry-sdk/src/metrics/instrument.rs
jtescher: thanks

opentelemetry-sdk/src/metrics/instrument.rs
jtescher: The names are always ASCII so this is slightly more effecient. Made the lower conversion ascii too and added a reference to https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/metrics/api.md#instrument-name-syntax

opentelemetry-sdk/src/metrics/internal/aggregate.rs
jtescher: This was moved unchanged from the previous location, can certainly consider changing it but maybe in another PR as this one already has a few fiddly changes in it?

opentelemetry-sdk/src/metrics/internal/sum.rs
jtescher: It is a bit of that, but there are two separate cases as the first number represents the number of elements added to the aggregation or if it can be ignored (and the memory reused) for the 0 case, while the second is about if the previous data could be re-used in place vs creating a new aggregation to store the data for this cycle.

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// Checks whether aggregator has hit cardinality limit for metric streams
+pub(crate) fn is_under_cardinality_limit(size: usize) -> bool {
+    size < STREAM_CARDINALITY_LIMIT as usize - 1
```
TommyCpp, line 26: Why we `-1` here? Does it mean 1999 should return `false`?

TommyCpp (approved):
One nit otherwise looks good to me

opentelemetry-sdk/src/metrics/internal/aggregate.rs
lalitb: I guess this was added for #1066. Return false for 1999, as the overflow measurements are aggregated in separate metric stream. So total number of metric streams remain 2000.

lalitb (approved):

opentelemetry-sdk/src/metrics/internal/aggregate.rs
jtescher: yeah this one is also unchanged from before (aside from removing the unused `&self` param)

TommyCpp: Got it. We should probably add it to comment to make it clearer

PR #1267, opened 2023-09-13
Add Exponential histogram

jtescher:
## Changes

Adds support for exponential histograms

Based on #1266, need to merge that first.


jtescher:
@open-telemetry/rust-approvers anyone able to take a look? should be similar to other lang's exponential histograms.

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -186,6 +186,88 @@ impl<T: Copy> Clone for HistogramDataPoint<T> {
...
+    /// The range of positive value bucket counts.
+    pub positive_bucket: ExponentialBucket,
+    /// The range of negative value bucket counts.
+    pub negative_bucket: ExponentialBucket,
+
+    /// The width of the zero region.
+    ///
+    /// Where the zero region is defined as the closed interval
+    /// [-zero_threshold, zero_threshold].
+    pub zero_threshold: f64,
```
TommyCpp, line 252: Hmmm didn't see we actually use this in exponential histogram implementation?

TommyCpp:
Really nice work putting this together :+1: Have some questions but most of it LGTM

hdost (approved):
Love it 👍🏼 we can look at performance improvements after this is integrated.

opentelemetry-sdk/src/metrics/aggregation.rs
```diff
@@ -62,6 +62,32 @@ pub enum Aggregation {
...
+
+        /// Indicates whether to not record the min and max of the distribution.
+        ///
+        /// By default, these values are recorded.
+        ///
+        /// It is generally not valuable to record min and max for cumulative data
+        /// as they will represent the entire life of the instrument instead of just
+        /// the current collection cycle, you can opt out by setting this value to
+        /// `false`
+        record_min_max: bool,
```
lalitb, line 89: There are default value for these parameters as per the specs - max_size : 160, max_scale: 20. Would it be possible for application to not provide these values, and use the default one ? If not, something we can support in subsequent PR.

opentelemetry-sdk/src/metrics/data/mod.rs
lalitb, line 252: nit  - `zero_threshold` is optional as per the specs.  Should this be Option type, or `0` is considered as unset ?

lalitb (approved):
LGTM. Nicely done. Some nit comments, nothing blocking thought :)

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
```diff
@@ -0,0 +1,1704 @@
...
+            }
+            if v > self.max {
+                self.max = v;
+            }
+        }
+        if self.record_sum {
+            self.sum += v;
+        }
+
+        let abs_v = v.into_float().abs();
```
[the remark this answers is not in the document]
jtescher: Yeah the nan cases are handled, I think a pass in general could be made to make the code clearer, maybe follow up with that

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
```diff
@@ -0,0 +1,1704 @@
...
+///
+/// Each histogram is scoped by attributes and the aggregation cycle the
+/// measurements were made in.
+pub(crate) struct ExpoHistogram<T> {
+    record_sum: bool,
+    record_min_max: bool,
+    max_size: i32,
+    max_scale: i8,
+
+    values: Mutex<HashMap<AttributeSet, ExpoHistogramDataPoint<T>>>,
```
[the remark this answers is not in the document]
jtescher: This is true, as it is with the other aggregations, we could look into concurrent hashmap implementations or other methods, but getting a correct first impl seems like the best starting point. I did try a few other approaches but they didn't have performance wins that offset their complexity and edge cases yet

opentelemetry-sdk/src/metrics/data/mod.rs
jtescher: Yeah this is an optional part of the spec, it's in the protos too so worth having it in the data model so it could be used later without data model changes

PR #1268, opened 2023-09-13
Performance: Context holds SynchronizedSpan directly, not via HashMap

shaun-cox:
## Changes

Add the `span` field below:
```rust
pub struct Context {
    pub(super) span: Option<Arc<SynchronizedSpan>>,
    entries: HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>,
}
```
so that current span manipulation doesn't have to traverse the expensive HashMap and then get dispatched with virtual function calls.

Further, this keeps the HashMap empty for the common case of just having a span in context and no other types (like baggage).

I get that the opentelemetry specification is trying to be general with its view that `Context` is extensible and can hold anything, but there comes a real performance penalty for handling everything through that abstraction.

Results of `cargo bench -p opentelemetry_sdk --bench context`:

```text
context/has_active_span/in-cx/alt
                        time:   [3.5772 ns 3.5806 ns 3.5842 ns]
                        thrpt:  [279.01 Melem/s 279.28 Melem/s 279.55 Melem/s]
                 change:
                        time:   [-47.017% -46.850% -46.708%] (p = 0.00 < 0.05)
                        thrpt:  [+87.646% +88.145% +88.739%]
                        Performance has improved.
context/is_sampled/in-cx/alt
                        time:   [5.0980 ns 5.1075 ns 5.1170 ns]
                        thrpt:  [195.43 Melem/s 195.79 Melem/s 196.15 Melem/s]
                 change:
                        time:   [-41.439% -41.320% -41.208%] (p = 0.00 < 0.05)
                        thrpt:  [+70.091% +70.414% +70.762%]
                        Performance has improved.
context/is_recording/in-cx/alt
                        time:   [5.3604 ns 5.3951 ns 5.4407 ns]
                        thrpt:  [183.80 Melem/s 185.35 Melem/s 186.55 Melem/s]
                 change:
                        time:   [-41.657% -41.417% -41.156%] (p = 0.00 < 0.05)
                        thrpt:  [+69.940% +70.698% +71.399%]
                        Performance has improved.
context/has_active_span/in-cx/spec
                        time:   [12.958 ns 12.988 ns 13.022 ns]
                        thrpt:  [76.791 Melem/s 76.995 Melem/s 77.174 Melem/s]
                 change:
                        time:   [-63.146% -63.063% -62.980%] (p = 0.00 < 0.05)
                        thrpt:  [+170.12% +170.73% +171.34%]
                        Performance has improved.
[...]

PR #1270, opened 2023-09-13
Performance: By using const over static, improve noop span creation performance another 10%

shaun-cox:
Improves performance for span creation when the SDK is not configured another 10%.

(Depends on https://github.com/open-telemetry/opentelemetry-rust/pull/1268 to go first.)

```text
new_span/if_parent_sampled/no-sdk/spec
                        time:   [83.124 ns 83.224 ns 83.312 ns]
                        thrpt:  [12.003 Melem/s 12.016 Melem/s 12.030 Melem/s]
                 change:
                        time:   [-9.3988% -9.2273% -9.0668%] (p = 0.00 < 0.05)
                        thrpt:  [+9.9709% +10.165% +10.374%]
                        Performance has improved.
```

## Changes

- remove branch in `build_with_context` for the noop tracer. the current context's `span()` method already returns &NOOP_SPAN in the case when no active span is present, which already has the right span context value to use.
- changed NOOP_SPAN from a lazy `static` to a `const`
- changed other constituents of span context to use `const` default values too.

### DIFF
--- a/opentelemetry/src/trace/noop.rs
+++ b/opentelemetry/src/trace/noop.rs
@@ -5,7 +5,7 @@
 //! to have minimal resource utilization and runtime impact.
 use crate::{
     propagation::{text_map_propagator::FieldIter, Extractor, Injector, TextMapPropagator},
-    trace::{self, TraceContextExt, TraceFlags, TraceState},
+    trace::{self, TraceContextExt as _},
     Context, InstrumentationLibrary, KeyValue,
 };
 use std::{borrow::Cow, sync::Arc, time::SystemTime};
@@ -38,25 +38,11 @@ pub struct NoopSpan {
     span_context: trace::SpanContext,
 }
 
-impl Default for NoopSpan {
-    fn default() -> Self {
-        NoopSpan::new()
-    }
-}
-
 impl NoopSpan {
-    /// Creates a new `NoopSpan` instance.
-    pub fn new() -> Self {
-        NoopSpan {
-            span_context: trace::SpanContext::new(
-                trace::TraceId::INVALID,
-                trace::SpanId::INVALID,
-                TraceFlags::default(),
-                false,
-                TraceState::default(),
-            ),
-        }
-    }
+    /// The default `NoopSpan`, as a constant
+    pub const DEFAULT: NoopSpan = NoopSpan {
+        span_context: trace::SpanContext::NONE,
+    };
 }
 
 impl trace::Span for NoopSpan {
@@ -135,12 +121,8 @@ impl trace::Tracer for NoopTracer {
     /// If the span builder or the context's current span contains a valid span context, it is
     /// propagated.
     fn build_with_context(&self, _builder: trace::SpanBuilder, parent_cx: &Context) -> Self::Span {
-        if parent_cx.has_active_span() {
-            NoopSpan {
-                span_context: parent_cx.span().span_context().clone(),
-            }
-        } else {
-            NoopSpan::new()
+        NoopSpan {
+            span_context: parent_cx.span().span_context().clone(),
         }
     }
 }
@@ -178,7 +160,7 @@ impl TextMapPropagator for NoopTextMapPropagator {
 mod tests {
     use super::*;
     use crate::testing::trace::TestSpan;
-    use crate::trace::{self, Span, Tracer};
+    use crate::trace::{self, Span, TraceState, Tracer};
 
     fn valid_span_context() -> trace::SpanContext {
         trace::SpanContext::new(
--- a/opentelemetry/src/trace/span_context.rs
+++ b/opentelemetry/src/trace/span_context.rs
@@ -20,6 +20,14 @@ use thiserror::Error;
 pub struct TraceFlags(u8);
 
 impl TraceFlags {
+    /// Trace flags with the `sampled` flag set to `0`.
+    ///
+    /// Spans that are not sampled will be ignored by most tracing tools.
+    /// See the `sampled` section of the [W3C TraceContext specification] for details.
+    ///
+    /// [W3C TraceContext specification]: https://www.w3.org/TR/trace-context/#sampled-flag
+    pub const NOT_SAMPLED: TraceFlags = TraceFlags(0x00);
+
     /// Trace flags with the `sampled` flag set to `1`.
     ///
     /// Spans that are not sampled will be ignored by most tracing tools.
@@ -216,6 +224,9 @@ impl fmt::LowerHex for SpanId {
 pub struct TraceState(Option<VecDeque<(String, String)>>);
 
 impl TraceState {
+    /// The default `TraceState`, as a constant
+    pub const NONE: TraceState = TraceState(None);
+
     /// Validates that the given `TraceState` list-member key is valid per the [W3 Spec].
     ///
     /// [W3 Spec]: https://www.w3.org/TR/trace-context/#key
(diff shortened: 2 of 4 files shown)


opentelemetry/src/trace/span_context.rs
```diff
@@ -457,15 +468,18 @@ pub struct SpanContext {
...
+        trace_id: TraceId::INVALID,
+        span_id: SpanId::INVALID,
+        trace_flags: TraceFlags::NOT_SAMPLED,
+        is_remote: false,
+        trace_state: TraceState::NONE,
+    };
+
     /// Create an invalid empty span context
     pub fn empty_context() -> Self {
-        SpanContext::new(
```
shaun-cox, line 462: There are a lot of places using `empty_context()` that can be replaced with `SpanContext::NONE`.  I haven't changed them in this PR, but my plan would be to change them all in a future PR and then remove the `empty_context()` function.  Anyone see any problems with that approach?

opentelemetry/src/trace/noop.rs
```diff
@@ -38,25 +38,11 @@ pub struct NoopSpan {
...
-                trace::TraceId::INVALID,
-                trace::SpanId::INVALID,
-                TraceFlags::default(),
-                false,
-                TraceState::default(),
-            ),
-        }
-    }
+    /// The default `NoopSpan`, as a constant
+    pub const DEFAULT: NoopSpan = NoopSpan {
```
djc, line 43: I suppose we could use an inherent `const fn default()` here instead of the associated const. On the one hand, that would look more familiar, on the other hand maybe it's an uncanney valley trap?

opentelemetry/src/trace/span_context.rs
```diff
@@ -20,6 +20,14 @@ use thiserror::Error;
 pub struct TraceFlags(u8);
 
 impl TraceFlags {
+    /// Trace flags with the `sampled` flag set to `0`.
```
djc, line 23: Style suggestion: I usually like to put `const`s near the bottom instead of near the top, since they're usually relatively simple and don't change often.

opentelemetry/src/trace/noop.rs
```diff
@@ -135,13 +121,8 @@ impl trace::Tracer for NoopTracer {
...
     /// propagated.
     fn build_with_context(&self, _builder: trace::SpanBuilder, parent_cx: &Context) -> Self::Span {
-        if parent_cx.has_active_span() {
-            NoopSpan {
-                span_context: parent_cx.span().span_context().clone(),
-            }
-        } else {
-            NoopSpan::new()
-        }
+        let span_context = parent_cx.span().span_context().clone();
```
djc, line 124: Style nit: I would just inline this in the `NoopSpan` initializer.

djc (approved):
Nice!

lalitb:
Just to understand more, the performance improvement here is for the first call to `NoopTracer::build_with_span()`, when the static `NOOP_SPAN` (in existing code) is initialized. The subsequent calls to `NoopTracer::build_with_span()` will utilize this already created NOOP_SPAN, so there won't be perf improvement ?

shaun-cox:
> Just to understand more, the performance improvement here is for the first call to `NoopTracer::build_with_span()`, when the static `NOOP_SPAN` (in existing code) is initialized. As the subsequent calls to `NoopTracer::build_with_span()` will utilize this already created `NOOP_SPAN`.

Good question.  It's not so much about optimizing the first call which accesses a lazy static (to save the initialization), its more about optimizing every call to the lazy static which has to check that that initialization has been done before proceeding.  When it's a constant, there is nothing to initialize, so there is no need for a call to check that it has been initialized.

The [generated assembly for this sample code](https://rust.godbolt.org/z/j7TKW1Mrf) should make it clearer.

```
example::access_static:
        push    rax
        lea     rdi, [rip + example::STATIC_THING]
        call    qword ptr [rip + <once_cell::sync::Lazy<T,F> as core::ops::deref::Deref>::deref@GOTPCREL]
        pop     rcx
        ret

example::access_const:
        lea     rax, [rip + .L__unnamed_20]
        ret
```

shaun-cox:
Oh, and I should have set the optimization flag for the compiler for release, above.  [When I do that](https://rust.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:13,endLineNumber:15,positionColumn:13,positionLineNumber:15,selectionStartColumn:13,selectionStartLineNumber:15,startColumn:13,startLineNumber:15),source:'use+once_cell::sync::Lazy%3B%0A%0Astatic+STATIC_THING:+Lazy%3CThing%3E+%3D+Lazy::new(Thing::new)%3B%0A%0Aconst+CONST_THING:+Thing+%3D+Thing::NONE%3B%0A%0Apub+struct+Thing+%7B%0A++++pub+data:+u128,%0A%7D%0A%0Aimpl+Thing+%7B%0A++++const+NONE:+Thing+%3D+Thing+%7Bdata:+0%7D%3B%0A%0A++++fn+new()+-%3E+Self+%7B%0A++++++++Thing%7Bdata:+0%7D%0A++++%7D%0A%7D%0A%0Apub+fn+access_static()+-%3E+%26!'static+Thing+%7B%0A++++%26STATIC_THING%0A%7D%0A%0Apub+fn+access_const()+-%3E+%26!'static+Thing+%7B%0A++++%26CONST_THING%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:44.21235857267189,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:r1720,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!((name:once_cell,ver:'1120')),options:'-C+opt-level%3D3',overrides:!((name:edition,value:'2021')),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+rustc+1.72.0+(Editor+%231)',t:'0')),k:50,l:'4',m:82.7258320126783,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'rustc+1.72.0',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+rustc+1.72.0+(Compiler+%231)',t:'0')),header:(),l:'4',m:17.27416798732171,n:'0',o:'',s:0,t:'0')),k:55.787641427328104,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), the results are even more obvious. :)

```
example::access_static:
        push    rax
        lea     rax, [rip + example::STATIC_THING]
        mov     qword ptr [rsp], rax
        movzx   eax, byte ptr [rip + example::STATIC_THING+24]
        cmp     al, 2
        jne     .LBB11_4
.LBB11_1:
        movzx   eax, byte ptr [rip + example::STATIC_THING+24]
        cmp     al, 2
        jne     .LBB11_7
        cmp     qword ptr [rip + example::STATIC_THING], 0
        je      .LBB11_3
        lea     rax, [rip + example::STATIC_THING+8]
        pop     rcx
        ret
.LBB11_4:
        mov     rdi, rsp
        call    once_cell::imp::OnceCell<T>::initialize
        movzx   eax, byte ptr [rip + example::STATIC_THING+24]
        cmp     al, 2
        je      .LBB11_1
        lea     rdi, [rip + .L__unnamed_11]
        lea     rdx, [rip + .L__unnamed_12]
        mov     esi, 41
        call    qword ptr [rip + core::panicking::panic@GOTPCREL]
        ud2
.LBB11_7:
        lea     rdi, [rip + .L__unnamed_13]
        lea     rdx, [rip + .L__unnamed_14]
        mov     esi, 39
        call    qword ptr [rip + core::panicking::panic@GOTPCREL]
        ud2
.LBB11_3:
        lea     rdi, [rip + .L__unnamed_5]
        lea     rdx, [rip + .L__unnamed_15]
        mov     esi, 23
        call    qword ptr [rip + core::panicking::panic@GOTPCREL]
        ud2

example::STATIC_THING:
        .zero   8
        .zero   16
        .zero   1
        .zero   7
        .quad   example::Thing::new
```
vs
```
example::access_const:
        lea     rax, [rip + .L__unnamed_16]
        ret

.L__unnamed_16:
        .zero   16
```

lalitb:
> The [generated assembly for this sample code](https://rust.godbolt.org/z/j7TKW1Mrf) should make it clearer.

Thanks, the assembly makes it much clear.

lalitb (approved):

TommyCpp (approved):

PR #1272, opened 2023-09-18
feat: force user to add exporter builder pipeline.

TommyCpp:
Currently, when users don't configure a builder pipeline. We will throw a runtime error. This could be frustrating as user may not realize they forgot to config an exporter until runtime.

Use will not be able to call `build` or `install_*` functions until they configure an exporter with the `with_exporter` function. Existing use case should remain unimpacted 

## Changes
- This change forces the user to call `with_exporter` before building exporters or installing the pipeline and eliminates the `NoExporterBuilder` error. The downside is users cannot call `with_exporter` multiple times to override the exporter pipeline. However, I think this use case is limited and not worth supporting. 

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -213,7 +213,6 @@ pub use crate::span::{
     OtlpTracePipeline, SpanExporter, SpanExporterBuilder, OTEL_EXPORTER_OTLP_TRACES_COMPRESSION,
     OTEL_EXPORTER_OTLP_TRACES_ENDPOINT, OTEL_EXPORTER_OTLP_TRACES_TIMEOUT,
 };
-use std::marker::PhantomData;
 
 #[cfg(feature = "metrics")]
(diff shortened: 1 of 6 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -408,3 +405,9 @@ pub enum Protocol {
     /// HTTP protocol with binary protobuf
     HttpBinary,
 }
+
+#[derive(Debug, Default)]
+#[doc(hidden)]
+/// Placeholder type when no exporter pipeline has been configured in telemetry pipeline.
+/// Users shouldn't be able to use this type, hence the private PhantomData.
+pub struct NoExporterConfig(PhantomData<()>);
```
jtescher, line 413: nit: This doesn't require `PhantomData` as ` NoExporterConfig(());` also has a private constructor

jtescher (approved):
Nice, always good to move these types of errors to compile time.

Issue #1265, opened 2023-09-11
Allow use of "/" in Metrics Instrument Names, and with a 255 char limit


lalitb:
As per the recent specs change : 
 - Increase max instrument name length from 63 to 255: https://github.com/open-telemetry/opentelemetry-specification/pull/3648
 - Instrument names can have "/" character : https://github.com/open-telemetry/opentelemetry-specification/pull/3684

cijothomas:
Club it here https://github.com/open-telemetry/opentelemetry-rust/issues/1264 or do separate to quickly get this out?

lalitb:
Either is fine, as long as we can track these two specs changes :)

cijothomas:
Straightforward, so tagging with "good first issue", to see if anyone wants to try.

Issue #1219, opened 2023-08-22
Create unit tests for metrics sdk


lzchen:
Tracking issue for making doctests/unit tests for metrics sdk

PR #1274, opened 2023-09-18
Fix metric instrument name validation

lzchen:
Dash should be an underscore.

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter.rs
+++ b/opentelemetry-sdk/src/metrics/meter.rs
@@ -25,7 +25,7 @@ use crate::metrics::{
 const INSTRUMENT_NAME_MAX_LENGTH: usize = 255;
 // maximum length of instrument unit name
 const INSTRUMENT_UNIT_NAME_MAX_LENGTH: usize = 63;
-const INSTRUMENT_NAME_ALLOWED_NON_ALPHANUMERIC_CHARS: [char; 4] = ['-', '.', '-', '/'];
+const INSTRUMENT_NAME_ALLOWED_NON_ALPHANUMERIC_CHARS: [char; 4] = ['_', '.', '-', '/'];
 
 // instrument validation error strings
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -25,7 +25,7 @@ use crate::metrics::{
 const INSTRUMENT_NAME_MAX_LENGTH: usize = 255;
 // maximum length of instrument unit name
 const INSTRUMENT_UNIT_NAME_MAX_LENGTH: usize = 63;
-const INSTRUMENT_NAME_ALLOWED_NON_ALPHANUMERIC_CHARS: [char; 4] = ['-', '.', '-', '/'];
+const INSTRUMENT_NAME_ALLOWED_NON_ALPHANUMERIC_CHARS: [char; 4] = ['_', '.', '-', '/'];
```
cijothomas, line 28: https://github.com/open-telemetry/opentelemetry-rust/issues/1265 This has some unit-tests added, wondering why the unittests did not caught this?

lalitb (approved):

opentelemetry-sdk/src/metrics/meter.rs
lalitb: I think the existing unit-test doesn't catch this. Good to add one.

TommyCpp: +1 for unit tests

cijothomas (approved):

ymgyt:
sorry about the misss in the move validation, and thanks for fix

opentelemetry-sdk/src/metrics/meter.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/1219 is opened to track adding tests. Resolving this conversation and would suggest to merge this PR with the fix.

PR #1278, opened 2023-09-22
Cleanup crate documentation

### LINKED ISSUE #1256: OpenTelemetry docs should talk only about instrumenting, and not about sdk or exporters.
[OpenTelemetry](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/Cargo.toml#L35-L37) crate has dev-dependencies on the SDK/stdout crates as they are demonstrated in the [docs](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/lib.rs#L25).

Proposing to remove that dependency, and modify the doc to only show the API usage, without showing anything related to SDK/Exporters. We should say something like "For the telemetry to be exported, need to install additional crates and point to them", so users are aware that OpenTelemetry crate alone is just the facade which does nothing in the absence of further setup.

This approach is used in https://docs.rs/tracing/latest/tracing/ docs already - it just talks about the tracing facade there, and [lists the tracing-subscriber/other](https://docs.rs/tracing/latest/tracing/#related-crates) crates that are required for the traces to land somewhere.

Similar change is done in https://github.com/open-telemetry/opentelemetry-rust/pull/1249. There could be more places, but want to get general feedback on this proposal before further cleanups.

shaun-cox:
Fixes #1256

## Changes

- remove the `sdk` and `stdout` crates from `dev-dependencies` of the `api` crate.
- update `api` documentation to only depend on `api` types in the examples.

### DIFF
--- a/opentelemetry/src/lib.rs
+++ b/opentelemetry/src/lib.rs
@@ -1,11 +1,9 @@
-//! OpenTelemetry provides a single set of APIs, libraries, agents, and collector
-//! services to capture distributed traces and metrics from your application. You
-//! can analyze them using [Prometheus], [Jaeger], and other observability tools.
+//! Implements the [`API`] component of [OpenTelemetry].
 //!
 //! *Compiler support: [requires `rustc` 1.64+][msrv]*
 //!
-//! [Prometheus]: https://prometheus.io
-//! [Jaeger]: https://www.jaegertracing.io
+//! [`API`]: https://opentelemetry.io/docs/specs/otel/overview/#api
+//! [OpenTelemetry]: https://opentelemetry.io/docs/what-is-opentelemetry/
 //! [msrv]: #supported-rust-versions
 //!
 //! # Getting Started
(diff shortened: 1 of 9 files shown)


opentelemetry/src/lib.rs
```diff
@@ -109,18 +94,9 @@
...
-//!
-//! * `rt-tokio`: Spawn telemetry tasks using [tokio]'s multi-thread runtime.
-//! * `rt-tokio-current-thread`: Spawn telemetry tasks on a separate runtime so that the main runtime won't be blocked.
-//! * `rt-async-std`: Spawn telemetry tasks using [async-std]'s runtime.
-//!
-//! [tokio]: https://crates.io/crates/tokio
-//! [async-std]: https://crates.io/crates/async-std
+//! * `trace`: Includes the trace API (enabled by default).
+//! * `metrics`: Includes the unstable metrics API.
+//! * `logs`: Includes the logs API.
```
cijothomas, line 99:
```suggestion
//! * `logs`: Includes the logs bridge API.
```

djc (approved):

TommyCpp (approved):

PR #1279, opened 2023-09-28
Upgrade dependencies

shaun-cox:
See https://deps.rs/crate/opentelemetry_sdk/0.20.0#vulnerabilities

## Changes
- ran `cargo upgrade -p ordered-float@4.1 -p pprof@0.13`

### DIFF
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -19,7 +19,7 @@ futures-channel = "0.3"
 futures-executor = "0.3"
 futures-util = { version = "0.3.17", default-features = false, features = ["std", "sink", "async-await-macro"] }
 once_cell = "1.10"
-ordered-float = "3.4.0"
+ordered-float = "4.0"
 percent-encoding = { version = "2.0", optional = true }
 rand = { version = "0.8", default-features = false, features = ["std", "std_rng"], optional = true }
 regex = { version = "1.0", optional = true }
@@ -39,7 +39,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 indexmap = "2.0"
 criterion = { version = "0.5", features = ["html_reports"] }
 [target.'cfg(not(target_os = "windows"))'.dev-dependencies]
-pprof = { version = "0.12", features = ["flamegraph", "criterion"] }
+pprof = { version = "0.13", features = ["flamegraph", "criterion"] }
 
 [features]
 default = ["trace"]
--- a/opentelemetry-contrib/Cargo.toml
+++ b/opentelemetry-contrib/Cargo.toml
@@ -47,7 +47,7 @@ criterion = { version = "0.5", features = ["html_reports"] }
 futures-util = { version = "0.3", default-features = false, features = ["std"] }
 opentelemetry_sdk = { path = "../opentelemetry-sdk", features = ["trace", "testing"] }
 [target.'cfg(not(target_os = "windows"))'.dev-dependencies]
-pprof = { version = "0.12", features = ["flamegraph", "criterion"] }
+pprof = { version = "0.13", features = ["flamegraph", "criterion"] }
 
 [[bench]]
 name = "new_span"
--- a/opentelemetry-stdout/Cargo.toml
+++ b/opentelemetry-stdout/Cargo.toml
@@ -29,7 +29,7 @@ opentelemetry = { version = "0.21", path = "../opentelemetry", default_features 
 opentelemetry_sdk = { version = "0.20", path = "../opentelemetry-sdk", default_features = false }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-ordered-float = "3.4.0"
+ordered-float = "4.0"
 
 [dev-dependencies]
 opentelemetry = { version = "0.21", path = "../opentelemetry", features = ["metrics"] }


djc:
I generally don't feel that it's very useful for all the libraries to upgrade within semver-compatible ranges -- applications should sort this out with their dep graph.

shaun-cox:
> I generally don't feel that it's very useful for all the libraries to upgrade within semver-compatible ranges -- applications should sort this out with their dep graph.
> 
My thought was to keep it simple and avoid any security vulnerability warnings from `deps.rs` for our crates by just moving to the minimum version without the vulnerability.  (1.24 version of `tokio` is still about 9 months old).

djc:
IMO then this feature from deps.rs is not thought through very well, and we should ignore it. Doing updates any time a semver-compatible vulnerability fix is released in an upstream dependency doesn't scale very well.

jtescher:
I'm partial to the requirements showing minimum required for functionality rather than security at the lib level.

opentelemetry-sdk/Cargo.toml
```diff
@@ -19,7 +19,7 @@ futures-channel = "0.3"
 futures-executor = "0.3"
 futures-util = { version = "0.3.17", default-features = false, features = ["std", "sink", "async-await-macro"] }
 once_cell = "1.10"
-ordered-float = "3.4.0"
+ordered-float = "4.1"
```
djc, line 22: Do you have a reason we require 4.1 instead of just 4? Generally since we previously worked with 3.4 there should be no reason we need 4.1.

shaun-cox: No reason, other than to observe the tooling seems to favor the latest...
- if I run `cargo upgrade` on a package, it generally picks the latest version.
- if I had just added the dependency today, using `cargo add ordered-float`, it would pick the latest version.
- as I look at existing Cargo.toml entries, they predominantly do not specify just `"<major>"`, or `"<major>.0"`, but typically all specify `minor` and sometimes `patch` levels.

Honestly, I don't think I understand (or have found) the definitive guidance for how or when to upgrade versions, even after trying to read all I can find about semver.  Can you share any good references or other tooling I may be unaware of?

In the meantime, do you want it to just say `ordered-float = "4"`?

djc: > No reason, other than to observe the tooling seems to favor the latest...
> 
>     * if I run `cargo upgrade` on a package, it generally picks the latest version.

IMO running `cargo upgrade` without `--compatible=ignore` is a nuisance, it bumps all the versions without regard for compatibility (thus restricting the version range beyond what's needed). I've gotten into trouble before when an upstream library used this approach and restricted the range for a dependency of theirs that had recently done a release and then decided to yank that release -- just blocking me from upgrading my dependency for no good reason. There's some basic support for `-Z minimal-versions` and we should typically try to avoid unnecessarily restricting the minimal version to allow our downstreams some lenience.

>     * if I had just added the dependency today, using `cargo add ordered-float`, it would pick the latest version.

If you first start using a dependency, there's no guarantee that you aren't using features that are new in the very latest version. So this is warranted, but is a separate case from upgrading like we do here.

>     * as I look at existing Cargo.toml entries, they predominantly do not specify just `"<major>"`, or `"<major>.0"`, but typically all specify `minor` and sometimes `patch` levels.

I think there isn't really consensus on the approach I'm advocating here to leave off meaningless trailing zero components. However, I've seen a lot of people newer to Rust get confused about the meaning of operator-less semver requirements like we have in `Cargo.toml` (thinking that `1.0.0` behaves more like `=1.0.0`). I feel like using just `1` makes it more obvious what's going on -- it's also just more concise.

> Honestly, I don't think I understand (or have found) the definitive guidance for how or when to upgrade versions, even after trying to read all I can find about semver. Can you share any good references or other tooling I may be unaware of?

Not sure I have any good references -- someone should really collect a bunch of this knowledge in one place.

> In the meantime, do you want it to just say `ordered-float = "4"`?

That's what I would do, yes.

shaun-cox: And, FWIW, I've been playing with `-Z minimal-versions` builds, but getting failures not related to our direct dependencies being specified to low, but failures in other deeper dependencies of dependencies.  e.g.
using [`cargo minimal-versions check --workspace`](https://github.com/taiki-e/cargo-minimal-versions)
```
info: running `cargo check` on actix-udp-example (10/31)
    Checking rand_core v0.6.2
   Compiling syn v1.0.98
    Checking tokio-util v0.7.9
    Checking io-lifetimes v1.0.1
    Checking ahash v0.8.3
    Checking zstd-safe v6.0.6
   Compiling syn v2.0.0
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> /home/shaun/.cargo/registry/src/index.crates.io-6f17d22bba15001f/io-lifetimes-1.0.1/src/lib.rs:31:26
   |
31 | #![cfg_attr(rustc_attrs, feature(rustc_attrs))]
   |                          ^^^^^^^^^^^^^^^^^^^^
```

djc: Yeah, unfortunately not all of the library ecosystem is very good about doing minimal versions. Doesn't mean we can't try our best...

djc (approved):

Issue #1255, opened 2023-09-08
Implement support for exporting metrics to stackdriver


thomaseizinger:
As far as I can tell, exporting metrics via `opentelemetry-stackdriver` is not (yet) supported. Opening this issue to coordinate its implementation.

cijothomas:
Is it needed or the native OTLP capability is sufficient? https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/otlp#otlp-metrics

thomaseizinger:
Unfortunately, GCP does not support running the OpsAgent on container-optimised OS so we can't use the OTLP exporter :(

cijothomas:
Got it. Is the current crate supported by Google? I am asking because in other languages, dedicated exporters are not supported by Google (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Exporter.Stackdriver#stackdriver-exporter-for-opentelemetry-net) and is usually unmaintained...

thomaseizinger:
I am not affiliated with this crate in any way, just a user myself but I think it is not officially supported.

djc:
The crate is not supported by Google, but we as users are maintaining it. @thomaseizinger happy to review a PR!

thomaseizinger:
> @thomaseizinger happy to review a PR!

I'll see what I can do. We are first integrating structured logging. Metrics will come after that :)

Issue #1225, opened 2023-08-23
Discussion: priority on environment variable vs compiling time value


TommyCpp:
Most of our creates generally allow two ways to configure values:

- Use environment variable fetched during runtime
- User provided value during compiling time(`with_xxx`)

However, this is unclear when both of the configurations present which one should take priority.  Regardless we should align on this and use it as a tenet across all our creates.

cc @open-telemetry/rust-approvers

djc:
I find it very confusing to make the compiled-in value take priority over the run-time value (from an environment variable). Code is usually written (and compiled) to function across many environments, whereas the runtime environment is specific to a particular invocation. Clearly the runtime environment, with its better specificity, should be trusted more than the code.

Maybe we should just not provide any way to configure these values in code?

See also this recent discussion in Cargo on a related problem:

https://github.com/rust-lang/cargo/pull/12515

In this case, there is no value provided in code, but there's a clear hierarchy of specificity.

TommyCpp:
I think most of people favors the environment variables. I think we can favor env vars here and make sure it's consistent.

Please leave comments if you disgree @open-telemetry/rust-approvers

hdost:
> I think most of people favors the environment variables. I think we can favor env vars here and make sure it's consistent.
> 
> 
> 
> Please leave comments if you disgree @open-telemetry/rust-approvers 

So while this is not a deviation (more of a clarification), it might also be useful to document. https://github.com/open-telemetry/opentelemetry-rust/issues/1297

cijothomas:
Reopening to further discuss given above.

cijothomas:
This is not consistent enforced everywhere. Adding to Log SDK stable milestone, as this is important to get closure.

lalitb:
Good catch.

I believe the priority order as discussed above should be 
 - Signal specific env-variables
 - Generic env-variables
 - with_xxx compiled values.
 - default values

Quick look at [BatchConfigBuilder](https://github.com/open-telemetry/opentelemetry-rust/blob/7c9447ff6ca43003b00b5ba32a6b0849106b96a9/opentelemetry-sdk/src/logs/log_processor.rs#L742) , the priority order seems to be:
 - with_xxx compiled values
 - env-variables
 - default values

This needs to be fixed. Will raise an issue for this first.

cijothomas:
Thanks. I'd like to revisit the priority order also.

Re-stating my opinion:

[...]

PR #1290, opened 2023-10-06
Allow specifying OTLP HTTP headers from env variable

lalitb:
Fixes #

## Changes

As per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/protocol/exporter.md#specifying-headers-via-environment-variables), the custom HTTP headers for OTLP exporter can be specified through env variables - `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXPORTER_OTLP_METRICS_HEADERS`. This PR add the support.

The existing implementation supports specifying the header at compile time. And as discussed in [#1255](https://github.com/open-telemetry/opentelemetry-rust/issues/1225#issuecomment-1726014869), the env-variable takes precedence. Currently in this PR, headers specified through configuration and environment variables are merged, with the environment variable keys overriding existing header keys."

Compiler time: `k1=v1, k2=v2`
Env variable: `k1=v1_new, k3=v3`
final: `k1=v1_new, k2=v2, k3=v3`

As of now, change are only done for  HTTP, and not gRPC. The [MetadataMap ](https://docs.rs/tonic/latest/tonic/metadata/struct.MetadataMap.html) in gRPC `tonic` crate only allows to specify metadata headers at compile time (through `&'static str`), so it seems not possible to populate this structure from env-var.


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -410,4 +451,21 @@ mod tests {
             // You may also want to assert on the specific error type if applicable
         });
     }
+
+    #[test]
+    fn test_add_header_from_string() {
```
cijothomas, line 456: as a future improvement, we need to vastly expand the test coverage and make sure all priority orders are tested, and the bad input behaviors are covered as well..

cijothomas (approved):
LGTM with a changelog entry.
Also left a comment about improving test coverage for this area to adequately cover all scenarios.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -291,6 +312,26 @@ fn resolve_endpoint(
         .map_err(From::from)
 }
 
+fn add_header_from_string(input: &str, headers: &mut HashMap<HeaderName, HeaderValue>) {
+    for pair in input.split(',') {
```
jtescher, line 316: nit: you can use [split_terminator](https://doc.rust-lang.org/stable/std/primitive.str.html#method.split_terminator) here to avoid the empty last instance check

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -291,6 +312,26 @@ fn resolve_endpoint(
...
+        if pair.is_empty() {
+            continue;
+        }
+        let mut kv_iter = pair.splitn(2, '=');
+        match (kv_iter.next(), kv_iter.next()) {
```
jtescher, line 321: nit: matching on [split_once](https://doc.rust-lang.org/stable/std/primitive.str.html#method.split_once) is simpler

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -182,6 +185,12 @@ impl HttpExporterBuilder {
             })
             .collect();
 
+        if let Ok(input) =
+            env::var(signal_http_headers_var).or_else(|_| env::var(OTEL_EXPORTER_OTLP_HEADERS))
```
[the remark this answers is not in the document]
lalitb: Done.

opentelemetry-otlp/src/exporter/http/mod.rs
lalitb: Good point. Improved test coverage, with priority order and bad input.

opentelemetry-otlp/src/exporter/http/mod.rs
lalitb: Thanks done.

opentelemetry-otlp/src/exporter/http/mod.rs
lalitb: Good point. Done.

jtescher (approved):

Issue #1284, opened 2023-10-03
SpanAttribute key deduplication


cijothomas:
LogRecord has switched from [HashMap(OrderMap to be precise) to simple Vector](https://github.com/open-telemetry/opentelemetry-rust/pull/1142) to attributes to achieve significant perf gains.

Span is currently using HashMap(OrderMap), and want to gain more opinions on moving this to simple Vector as well. There was some argument for keeping HashMap to keep O(1) lookup as requested here : https://github.com/open-telemetry/opentelemetry-rust/issues/794.

The [performance benchmarks in this PR](https://github.com/open-telemetry/opentelemetry-rust/pull/1121) have presented the case for switching to Vector. The PR, which added benchmarks got merged, but there was no issue to track making the actual changes! 

This issue is opened to track changing the HashMap/OrderMap to vectors, and remove the SDK from doing the de-duplication. It is still possible to make an opt-in feature to do this (as a processor or even at exporter thread to not affect user thread), if we there is enough demand for such a feature.

Related to https://github.com/open-telemetry/opentelemetry-rust/issues/1283 as well.

cijothomas:
Related to similar issue specifically for Metrics : https://github.com/open-telemetry/opentelemetry-rust/issues/1098

cijothomas:
Closed with https://github.com/open-telemetry/opentelemetry-rust/pull/1293, with a near 50% performance improvement.

Issue #1283, opened 2023-10-03
Need for EvictedHashMap, EvictedQueue


cijothomas:
We currently maintain a [EvictedHashMap ](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/evicted_hash_map.rs)and [EvictedQueue ](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/evicted_queue.rs)structures.

The main reason, as I understand it, for the usage of the above structures are to enforce the [SpanLimits ](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/trace/sdk.md#span-limits)from the spec.

The current SDK implementation seem to be copying all attributes into these data structures : [attributes](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/tracer.rs#L205-L208), [events](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/tracer.rs#L225-L237), [links](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/tracer.rs#L212-L222).

The above has perf overheads. However, the spec does not require the kind of "eviction" behavior. i.e if user tries to add more attributes/events/links than allowed, ignore the newer one. There is no need of "evict" an existing one. We could achieve this with regular Vec/HashMap.

@TommyCpp mentioned the desire to give predictable behavior for users, if we are dropping attributes. I guess it can be achieved with modifications to the API/contract:
AddAttribute(KeyValue) - adds attribute. If size limit is reached, drop this one. 
AddAttribute(Array/IntoIterator of KeyValues) - adds attributes in the order, and drops from the moment size limit is reached. If user wants to control the ones that gets dropped, they can provide attributes in that order.

Related to this the SpanBuilder being a Struct in the API, exposing the data structures used to [store Attributes, Events and Links.](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/trace/tracer.rs#L263-L269). I think this should be a trait in the API, and the implementation (i.e SDK) should decide the data structures to store it. See similar issue for [Logs](https://github.com/open-telemetry/opentelemetry-rust/issues/1189).

Don't have an PR with some proposal or benchmark numbers yet, but opening the issue to get feedback about the general direction:
[...]

PR #1293, opened 2023-10-11
SpanAttributes modified to use Vec instead of OrderMap/EvictedHashMap

cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-rust/issues/1284 and https://github.com/open-telemetry/opentelemetry-rust/issues/1283. Only focused on Span attributes (to be extended to Events, Links post this PR).

This PR modifes Span attributes to use Vec<KeyValue> instead of OrderMap<Key, Value> in SpanBuilder( part of API), EvictedHashMap in Span (part of SDK). This is similar to what was done for [Logs](https://github.com/open-telemetry/opentelemetry-rust/pull/1142). 

Performance Implications:
The performance gains are matching what is already demonstrated in https://github.com/open-telemetry/opentelemetry-rust/pull/1121, but sharing the results from my run:

[Stress test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/traces.rs):
main branch:
Number of threads: 4
Throughput: 6,592,200 iterations/sec
Throughput: 6,504,400 iterations/sec
Throughput: 6,461,400 iterations/sec

This PR:
Number of threads: 4
Throughput: 10,274,200 iterations/sec
Throughput: 10,295,800 iterations/sec
Throughput: 10,329,000 iterations/sec

This is **~50% improvement** in throughput!

Benchmarks:
	time:   [586.89 ns 594.65 ns 603.28 ns]
	**change: [-38.279%** -36.582% -34.871%] (p = 0.00 < 0.05)
	Performance has improved.

This means we'll no longer be doing de-duplicating of span attributes. [Opened an issue](https://github.com/open-telemetry/opentelemetry-rust/issues/1300) to track user feedback on this, which is also linked from changelog for easy visibility.

Additional Details: 
~The feature to enforce attributes limit is not done now. dropped_attribute_count will be 0 hence. It can be added back as a follow up, but I'll open a separate issue to discuss some issues in implementation of that, with some proposals.~ 

~Have just done the bare minimum to show the changes and its perf implications. I expect this to be 1st of a series of PRs doing similar for Events, Links, and also bringing back the attribute limit enforcement, along with new unit tests coverage. **_I want to get an okay that this direction is agreed upon, before working further on this_**. (Note: There were few other potential ideas discussed in todays SIG call, which can further help perf, but I'll open separate issues to track them., and they can be additive changes after this). Also, there might be interest in doing de-dup on a opt-in basis, either in SDK or a custom Processor, or Exporter  - this is another candidate for follow up!~

Thanks @shaun-cox and @lalitb for prior works, benchmarks and other consultation!


shaun-cox:
I love the direction and think it's necessary to achieve the [project's mission](https://github.com/open-telemetry/community/blob/main/mission-vision-values.md#we-value-performance).

cijothomas:
I made a quick attempt to fix [this ](https://github.com/open-telemetry/opentelemetry-rust/issues/1209) issue for Spans on top of this branch, and the throughput reaches 16m/sec. Overall that'd be more than doubling the perf (6M/sec to 16M/sec). Of course I haven't thought through the full implications, so will wait for green light on this PR's direction before further exploring that...

jtescher:
@cijothomas I think this makes sense, thanks for working on it. Retaining first n rather than last n seems fine so the perf win here is well worth it. We probably still want to keep track of the dropped counts where possible, could store vec and droppped count, or have a struct that stored both?

cijothomas:
> We probably still want to keep track of the dropped counts where possible, could store vec and droppped count, or have a struct that stored both

I can bring it back. I left some comments in here about that feature overall : https://github.com/open-telemetry/opentelemetry-rust/issues/1283#issuecomment-1757922468. I'd pursue that separately, to avoid distracting this PR.

Given there is +1s to the direction of this PR from other maintainer/approvers, I'll continue working on it.
Will mark it draft now, to polish it further and to add back the feature of tracking dropped_count. Once done, will undraf the PR to request full review! Thanks.

TommyCpp:
I think it's the right direction. We should be explicit in docs that attributions can duplicate so we don't need to have deduplication or sort logic for them. Thanks for working on it

jtescher:
could also see where perf is with hashmap to keep de-duplication but change to last-write-wins, would be worth knowing the cost of each

cijothomas:
> could also see where perf is with hashmap to keep de-duplication but change to last-write-wins, would be worth knowing the cost of each

Yes thats what I tried first!. From my stress tests:
6.2 M/sec main branch

7.5 M/sec with regular HashMap (so sdk does dedup, but no eviction).

10 M/Sec with vec, where we removed de-duplication as well. 

(I do see a future where users will ask for de-duplication in the sdk level. We can do that as an opt-in SpanProcessor or even an Exporter feature, so as to no penalize every user.)

cijothomas:
@open-telemetry/rust-approvers Please review!

TommyCpp (approved):

lalitb (approved):

opentelemetry-datadog/src/exporter/model/mod.rs
```diff
@@ -281,18 +276,19 @@ pub(crate) mod tests {
...
             &unified_tags,
         )?);
 
-        assert_eq!(encoded.as_str(), "kp6jd2VirHNlcnZpY2VfbmFtZaljb21wb25lbnSocmVzb3VyY2WpaG9zdC5uYW\
-        1lpHRlc3Snc2VydmljZax0ZXN0LXNlcnZpY2WjZW52qHRlc3QtZW52p3ZlcnNpb26sdGVzdC12ZXJzaW9uqXNwYW4udH\
-        lwZbVfc2FtcGxpbmdfcHJpb3JpdHlfdjGRkZzOAAAAAc4AAAACzgAAAAPPAAAAAAAAAAfPAAAAAAAAAGPPAAAAAAAAAA\
-        HTAAAAAAAAAADTAAAAADuaygDSAAAAAIXOAAAABM4AAAAFzgAAAAbOAAAAB84AAAAIzgAAAAnOAAAACs4AAAALzgAAAA\
-        zOAAAAAIHOAAAADcsAAAAAAAAAAM4AAAAA");
+        // TODO: Need someone to generate the expected result or instructions to do so.
+        // assert_eq!(encoded.as_str(), "kp6jd2VirHNlcnZpY2VfbmFtZaljb21wb25lbnSocmVzb3VyY2WpaG9zdC5uYW\
```
shaun-cox, line 287: Should we address this before merge?

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -120,7 +120,7 @@ impl Tracer {
     }
 }
 
-static EMPTY_ATTRIBUTES: Lazy<OrderMap<Key, Value>> = Lazy::new(Default::default);
+static EMPTY_ATTRIBUTES: Lazy<Vec<KeyValue>> = Lazy::new(Default::default);
```
shaun-cox, line 123: I wonder if we even need this as `static` seeing as how it's pretty cheap to create an empty `Vec`.  In fact, `Vec::new` is a `const fn`.

shaun-cox (approved):

opentelemetry-datadog/src/exporter/model/mod.rs
cijothomas: I don't think so. Opened issue to tackle this : https://github.com/open-telemetry/opentelemetry-rust/issues/1311

I'll re-bring the topic of creating contrib repo and move more components there, to make it easier to make changes in the main repo.

TommyCpp:
Discussed with @cijothomas and we will follow up with the nits. Merging this for now

opentelemetry-sdk/src/trace/tracer.rs
cijothomas: makes sense. Tacking this in the next PR continuing this direction!

Issue #1171, opened 2023-07-28
Need a way to suppress telemetry from SDKs own operation


cijothomas:
There is no mechanism today to suppress telemetry arising from SDK's own operations. For example, OTLP Exporter uses `tonic`, which is instrumented with `tracing` crate. When one enables tracing-appender, along with OTLP Exporter, all the logs emitted by `tonic` inside OTLPExporter, gets routed to OpenTelemetry! This is a never ending "live loop", as every export cause logs to be produced, which causes export to occur again, which further creates logs again and so on.

There was never a spec written for this (https://github.com/open-telemetry/opentelemetry-specification/issues/530), but languages like Python, .NET had provided solutions for this.

Opening an issue to track adding a capability to "Suppress instrumentation" in OTel Rust SDK, and have the official exporters leverage it.

cijothomas:
Closing as duplicate

Issue #1262, opened 2023-09-11
`new_view` should escape name before it is converted to a regex.


BrynCooke:
When using `new_view` it does a bunch of pre-processing on the name to turn it into a regex.
This includes:
* stripping `^` and `$`
* Replacing `?` with `.`
* Replacing `*` with `.*`

As metrics are likely namespaced with a `.` this means that if you want to create a view that matches a particular prefix you have to game the api by inserting `\\.` in your pattern (we have a_b_* metrics that we want to exclude and `a.b.*` metrics that we want to include due to legacy constraints).

Maybe a dedicated glob library should be used or just stick with regex without the partial globbing support?

jtescher:
I think glob matching could make sense as the spec doesn't call for more than wildcards, do you have time to submit a PR?

lalitb:
@BrynCooke do you plan to work on it, or else I can pick it as it will help us unblock CI - https://github.com/open-telemetry/opentelemetry-rust/pull/1294#issuecomment-1761734757

lalitb:
#1301 fixes it.

PR #1294, opened 2023-10-12
Add OTLP HTTP Log Exporter example using tracing appender.

lalitb:
## Changes
- Update the existing OTLP HTTP exporter to export logs instrumented through tokio library, using tracing appender. This change could have suffered from infinite live-loop issue described in #1171, but the Hyper HTTP library emits tracing/debug logs only if the tracing feature (which is disabled by default) is enabled for this library, and Cargo.toml for this example doesn't enable it. Also, there are no errors reported from this library (which would also get routed to OpenTelemetry appender) if the steps in examples are executed properly. So only the logs, traces and metrics geerated from the example code are exported to the collector.

- Also for simplicity removed all-in-one jaeger exporter from the collector configuration, and instead used Debug Exporter which dumps all events to console.

- The infinite live-issue discussed in #1171 need to be investigated separately, one option could be to filter all such non-required events using `event_enabled` function by checking the target/module.


opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -4,10 +4,33 @@ use opentelemetry::{
...
+use tracing::info;
+use tracing_subscriber::prelude::*;
+
+fn init_logs() -> Result<sdklogs::Logger, opentelemetry::logs::LogError> {
+    opentelemetry_otlp::new_pipeline()
+        .logging()
+        .with_log_config(
+            sdklogs::Config::default().with_resource(resource::Resource::new(vec![KeyValue::new(
+                opentelemetry_semantic_conventions::resource::SERVICE_NAME,
+                "basic-otlp-logging-example",
```
shaun-cox, line 24: Consider passing `env!("CARGO_BIN_NAME")` instead of hard-coding the example name, as that demonstrates best practice and works if someone copies the example and renames their binary.

shaun-cox (changes requested):
Thanks for doing this.  Just one nit which would be nice to address before merging...

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
lalitb: Good point, using hardcoded name now only if the env-variable is not set:
```rust
    let service_name = std::env::var("CARGO_BIN_NAME")
        .unwrap_or_else(|_| String::from("basic-otlp-http-logging-example"));
```

lalitb:
The CI is failing with:

``
error: package `regex-syntax v0.8.1` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.64.0
Error: Process completed with exit code 101.
``

Does it mean we need to bump up msrv to 1.65.0 ?

opentelemetry-otlp/examples/basic-otlp-http/README.md
```diff
@@ -1,20 +1,17 @@
...
 docker-compose up
-or
-docker-compose up -d
 ```
 
 In another terminal run the application `cargo run`
 
-Use the browser to see the trace:
-- Jaeger at http://0.0.0.0:16686
+The docker-compose terminal will display traces, metrics.
```
cijothomas, line 14:
```suggestion
The docker-compose terminal will display logs, metrics and traces.
```

opentelemetry-otlp/examples/basic-otlp-http/docker-compose.yaml
```diff
@@ -21,17 +13,7 @@ services:
       - "4317:4317"   # OTLP gRPC receiver
```
cijothomas, line 13: we don't need gRPC as this is only for http.. maybe better to combine both to a single example, as switching from http to gRPC is a small change in exporter.

lalitb: Yes, this PR is meant to add the log exporter in existing OTLP HTTP example. We can consolidate both the examples into single as separate cleanup PR.

lalitb: Have removed the unwanted ports from both the examples now.

cijothomas (approved):

lalitb:
> The CI is failing with:
> 
> `` error: package `regex-syntax v0.8.1` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.64.0 Error: Process completed with exit code 101. ``
> 
> Does it mean we should bump up msrv to 1.65.0 ?

@jtescher @TommyCpp  Seems the PRs are blocked on this error. Should we raise a PR to bump the msrv version ?

jtescher:
@lalitb We could do that, an alternative would be to swap the view regex impl to use a glob impl instead as suggested in https://github.com/open-telemetry/opentelemetry-rust/issues/1262, which would drop the need for the dependency entirely

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -52,12 +52,14 @@ fn init_metrics() -> metrics::Result<MeterProvider> {
 }
 
 fn init_logs() -> Result<opentelemetry_sdk::logs::Logger, LogError> {
+    let service_name = std::env::var("CARGO_BIN_NAME")
```
shaun-cox, line 55: sorry for the confusion earlier, but I meant use the standard [`env!`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) to get this.  e.g. just
``` rust
let service_name = env!("CARGO_BIN_NAME");
```

shaun-cox (changes requested):

opentelemetry-otlp/examples/basic-otlp/src/main.rs
lalitb: Yes my bad, this variable will not be available at run-time. Fixed it accordingly.

shaun-cox (approved):

PR #1295, opened 2023-10-12
Fix Resource dedup logic

BrynCooke:
## Changes

Previously Resource would have last k,v wins when supplied a list of k,v even if it was empty.

This change brings the logic in line with the documentation on Resource::new and also applies the same to `from_detectors`.

## Design
Suspect that the implementation was just an oversight as the existing doc comment seems to describe the desired behaviour.

Users will be able to specify a list of detectors in priority order and expect them to have first non empty value wins.

### DIFF
--- a/opentelemetry-sdk/src/resource/mod.rs
+++ b/opentelemetry-sdk/src/resource/mod.rs
@@ -385,9 +385,6 @@ mod tests {
 
     #[test]
     fn detect_resource_merge() {
-        env::set_var("OTEL_RESOURCE_ATTRIBUTES", "key=value, k = v , a= x, a=z");
-        env::set_var("irrelevant".to_uppercase(), "20200810");
-
         let resource1 = Resource::new(vec![
             KeyValue::new("a", ""),
             KeyValue::new("b", "2"),
--- a/opentelemetry-sdk/src/resource/mod.rs
+++ b/opentelemetry-sdk/src/resource/mod.rs
@@ -66,7 +66,7 @@ impl Resource {
         }
     }
 
-    /// Insert and entry if it does not exist or is an empty string
+    /// Insert an entry if it does not exist or is an empty string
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -65,6 +66,19 @@ impl Resource {
         }
     }
 
+    /// Insert and entry if it does not exist or is an empty string
+    fn insert(&mut self, key: Key, value: Value) {
+        match self.attrs.entry(key) {
+            Entry::Occupied(mut e) if e.get().to_string().is_empty() => {
```
lalitb, line 72: nit perf improvement:

```suggestion
            Entry::Occupied(mut e) if e.get().as_str().is_empty() => {
```

lalitb (approved):

hdost:
@BrynCooke thank you for your contribution. Unfortunately looking at the specs we should not even have an insert https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md

BrynCooke:
@hdost The `insert` method is private, it's not part of the public API, just to keep things dry.

hdost:
> @hdost The `insert` method is private, it's not part of the public API, just to keep things dry.

Indeed, however I do have one other issue as this works counter to the spec for merge.

BrynCooke:
Makes sense, I've instead raise an issue for discussion https://github.com/open-telemetry/opentelemetry-rust/issues/1298 that hopefully captures the issues around the current behaviour. I'll leave the PR for you to close though.

lalitb:
> Indeed, however I do have one other issue as this works counter to the spec for merge.

Good point.

BrynCooke:
Also from the spec:
| Resource detector packages MAY detect resource information from multiple possible sources and merge the result using the Merge operation described above.

So definitely agree that this PR doesn't fit the bill.

Issue #1296, opened 2023-10-13
Should opentelemetry-stdout conform to a strict serialization format?


erichulburd:
Currently, `opentelemetry-stdout` writes data in a Rust SDK-specific format: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-stdout/src/trace/transform.rs#L57 . Note, these structs do not implement `Deserialize`. This makes it hard to do anything with the data.

I'm currently using opentelemetry-stdout for making assertions on trace data exported from Rust. I propose `opentelemetry-stdout` write data in OTLP messages in JSON. This makes the data written to stdout machine readable - even across languages (which is especially relevant to work enabling tracing across Python-Rust FFI: https://github.com/rigetti/pyo3-opentelemetry/pull/3).

The snag I hit is that the trace and span ids written to stdout (or file) are not properly encoded to the OTel + Proto spec. Specifically, according to the Proto language specification, `bytes` fields within JSON should be encoded as:

>  a string using standard base64 encoding with paddings. Either standard or URL-safe base64 encoding with/without paddings are accepted.

See https://protobuf.dev/programming-guides/proto3/#json

_My reading of with / without paddings are the trailing "="._

Trace id and span id are encoded as hex in this library: https://github.com/open-telemetry/opentelemetry-rust/blob/6713143b59659dc509b7815404ebb57ad41cfe3a/opentelemetry-stdout/src/trace/transform.rs#L96

This isn't an issue for _every value_ , but there are some pathological examples where this produces invalid base64:

* trace id: 15769111199087022768103192234192075546
* encoded as: "bdd05355c559cbb7c36ee676b58fb1a"
* this should be "0bdd05355c559cbb7c36ee676b58fb1a"

cijothomas:
stdout exporter is not meant for machine readable. [Its just for humans to read, and learn how things would work.](https://github.com/open-telemetry/opentelemetry-rust/pull/1192#issuecomment-1669994637). I was planning to add some warnings about this on the doc, but just haven't got to it yet.

I would not recommend betting on stdout format for any automation testing. Instead, rely on InMemoryExporter for ensuring the apps are correctly instrumented with Otel. 
If you want to validate by using external process, then it may be better off setting up a OTLPEndpoint and validate the data comes in the expected format. OTLP output will be guaranteed to be stable.

jtescher:
[...]

PR #1299, opened 2023-10-13
Use display for standard trace and span id formats

jtescher:
As brought up in #1296 not all of the implementations of span and trace id formatting properly pad as required by the spec.

## Changes

Have consistent padding and hex casing by using the `Display` impl of trace and span ids which is in accordance with the [spec].

[spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.26.0/specification/trace/api.md#retrieving-the-traceid-and-spanid


shaun-cox:
Thanks for making it all consistent.  Might it be worth adding some method on `SpanContext` itself to perform this canonical formatting, and have most of the places that were touched here just invoke that?

djc (approved):

lalitb (approved):

TommyCpp (approved):

PR #1303, opened 2023-10-17
Add files to expose the currently used external types.

hdost:
To get ready for release candidate, we should keep track of these as it affects how stable we can consider being.

## Changes

Catalogs currently used External Types.

### DIFF
--- a/opentelemetry-sdk/allowed-external-types.toml
+++ b/opentelemetry-sdk/allowed-external-types.toml
@@ -1,0 +1,18 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+# This is used with cargo-check-external-types to reduce the surface area of downstream crates from
+# the public API. Ideally this can have a few exceptions as possible.
+allowed_external_types = [
+    "opentelemetry::*",
+    "async_channel::Receiver",
+    "async_channel::Sender",
+    "async_std::stream::interval::Interval",
+    "futures_channel::oneshot::Sender", # TODO: This is a pre-1.0 crate, we can't easily stabilize with this in the public API
+    "futures_core::future::BoxFuture", # TODO: This is a pre-1.0 crate, we can't easily stabilize with this in the public API
+    "futures_core::stream::Stream", # TODO: This is a pre-1.0 crate, we can't easily stabilize with this in the public API
+    "opentelemetry_http::HttpClient", # TODO: We probably shouldn't be depending on another SDK not in the API from the SDK.
+    "tokio_stream::wrappers::interval::IntervalStream", # TODO: This is a pre-1.0 crate, we can't easily stabilize with this in the public API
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/allowed-external-types.toml
```diff
@@ -0,0 +1,18 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+# This is used with cargo-check-external-types to reduce the surface area of downstream crates from
+# the public API. Ideally this can have a few exceptions as possible.
+allowed_external_types = [
```
cijothomas, line 5: do we expect to run CI checks for this? Or this is for manually checking

opentelemetry/allowed-external-types.toml
```diff
@@ -0,0 +1,19 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+# This is used with cargo-check-external-types to reduce the surface area of downstream crates from
+# the public API. Ideally this can have a few exceptions as possible.
+allowed_external_types = [
+    "indexmap::map::core::Entry",
```
cijothomas, line 6: i hope we can get to zero for opentelemtry api!

TommyCpp (approved):
Ah I was wondering how to guard our public API. Thanks for it!

opentelemetry/allowed-external-types.toml
lalitb: Agree, that would be an ideal scenario. In Otel-C++, there is no external dependency for API + SDK + Memory Exporter + OStream Exporter.

lalitb (approved):

opentelemetry-sdk/allowed-external-types.toml
hdost: Yes ideally in the future we can have check-external-types running on each crate. API,SDKs, Exporters, etc.

djc: I think we should not merge this without these files actually being used in CI...

TommyCpp: +1 let's run the current setup as CI?

TommyCpp (changes requested):
Block this PR for now as we get feedback to add the current setup as CI

hdost:
> Block this PR for now as we get feedback to add the current setup as CI

https://github.com/open-telemetry/opentelemetry-rust/actions/runs/6618694903/job/17977650680 << Added to CI, and verified that it will fail if we don't have something in the list.
We can always add more crates, but for now I figured we focus on API and SDK.

TommyCpp:
Perfect! Thanks for setting this up. I agree this is be enough for now

TommyCpp (approved):

cijothomas (approved):

PR #1305, opened 2023-10-17
Add tracing tests using inmemoryexporter

cijothomas:
Adding module level tests, which tests the entire tracing api/sdk pipeline. These are not unit tests, but not integration tests either. 
I think this is a good way to test things, and improve coverage. This is pretty close to testing what an end user experiences (minus the actual exporter part, which can be covered by spinning up collectors in the future.)

This was discussed in todays SIG meeting, and opening this PR to get early feedback. If all likes this direction, then I can send follow up PRs greatly improving coverage, and extend it to metrics and logs as well. (Or create issues with "help-wanted" tag inviting more folks to contribute.)

This uses in_memory_exporter which were recently added. If we ourselves don't use it, who else will 🤣 !

### DIFF
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -47,7 +47,7 @@ mod tests {
     };
 
     #[test]
-    fn tracing_in_span_test() {
+    fn tracing_in_span() {
         // Arrange
         let exporter = InMemorySpanExporterBuilder::new().build();
         let provider = TracerProvider::builder()
@@ -71,7 +71,7 @@ mod tests {
     }
 
     #[test]
-    fn tracing_tracer_start_test() {
+    fn tracing_tracer_start() {
         // Arrange
         let exporter = InMemorySpanExporterBuilder::new().build();
         let provider = TracerProvider::builder()
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -39,9 +39,12 @@ mod runtime_tests;
 
 #[cfg(all(test, feature = "testing"))]
 mod tests {
-    use crate::testing::trace::InMemorySpanExporterBuilder;
-    use opentelemetry::{trace::{Tracer, Span, TracerProvider as _}, KeyValue};
     use super::*;
+    use crate::testing::trace::InMemorySpanExporterBuilder;
+    use opentelemetry::{
+        trace::{Span, Tracer, TracerProvider as _},
+        KeyValue,
+    };
 
     #[test]
     fn tracing_in_span_test() {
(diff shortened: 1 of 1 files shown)


cijothomas:
Tagging @hdost @TommyCpp @lalitb with whom this was discussed in the SIG call today! Open for feedback from all!

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -36,3 +36,62 @@ pub use sampler::{JaegerRemoteSampler, JaegerRemoteSamplerBuilder};
...
+        let provider = TracerProvider::builder()
+            .with_span_processor(SimpleSpanProcessor::new(Box::new(exporter.clone())))
+            .build();
+
+        // Act
+        let tracer = provider.tracer("test_tracer");
+        let mut span = tracer.start("span_name");
+        span.set_attribute(KeyValue::new("key1", "value1"));
+        drop(span);
+        provider.force_flush();
```
TommyCpp, line 86: This part 
```
let exporter = InMemorySpanExporterBuilder::new().build();
        let provider = TracerProvider::builder()
            .with_span_processor(SimpleSpanProcessor::new(Box::new(exporter.clone())))
            .build();
```
seems to be a common pattern tests will use can we extract it as `setup` function?

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -36,3 +36,62 @@ pub use sampler::{JaegerRemoteSampler, JaegerRemoteSamplerBuilder};
...
+            .get_finished_spans()
+            .expect("Spans are expected to be exported.");
+        assert_eq!(exported_spans.len(), 1);
+        let span = &exported_spans[0];
+        assert_eq!(span.name, "span_name");
+        assert_eq!(span.instrumentation_lib.name, "test_tracer");
+    }
+
+    #[test]
+    fn tracing_tracer_start_test() {
```
TommyCpp, line 74: I don't think we need to add `_test` suffix as we already have `#[test]`?

TommyCpp (approved):
Small nit for simplify the test setup.

opentelemetry-sdk/src/trace/mod.rs
cijothomas: I'll defer it now. There are quite a number of tests coming, and once they all land, I can extract duplicates to helper method. I expect even setup would be different for many, so might need more customization than just a simple setup() function.

lalitb (approved):

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -36,3 +36,62 @@ pub use sampler::{JaegerRemoteSampler, JaegerRemoteSamplerBuilder};
...
+            .get_finished_spans()
+            .expect("Spans are expected to be exported.");
+        assert_eq!(exported_spans.len(), 1);
+        let span = &exported_spans[0];
+        assert_eq!(span.name, "span_name");
+        assert_eq!(span.instrumentation_lib.name, "test_tracer");
+    }
+
+    #[test]
+    fn tracing_tracer_start() {
```
shaun-cox, line 74: Just a nit on the apparent use of `tracing_` prefix in the function names... I'd prefer to rely on the parent module name which is already `trace::tests` and not redundantly add more `tracing_`.

shaun-cox (approved):
I like the direction.

opentelemetry-sdk/src/trace/mod.rs
cijothomas: oh yes! Will make this part of the next PR where I intend to add more tests, along with extracting common code to helpers.

PR #1307, opened 2023-10-18
Remove extra generic from with_http_client

ekzhang:
This generic type parameter is not being used anywhere, so when you try to use `with_http_client`, it requires you to specify a dummy type that implements `HttpClient + 'static`.

So this does not work:

```rust
new_pipeline().with_http_client(Arc::new(client))
```

But these work, even though they don't necessarily match the actual type of `client`:

```rust
new_pipeline::<Arc<dyn HttpClient>>().with_http_client(Arc::new(client))

new_pipeline::<reqwest::Client>().with_http_client(Arc::new(client))
```

## Changes

I changed the type signature of `with_http_client` to address this.

### DIFF
--- a/opentelemetry-datadog/CHANGELOG.md
+++ b/opentelemetry-datadog/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Do not set an empty span as the active span when the propagator does not find a remote span.
+- Change type signature of `with_http_client()` to use the provided generic as argument.
 
 ## V0.8.0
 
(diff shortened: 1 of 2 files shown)


cijothomas (approved):
Would be good to add changelog entry as well, given this is a public facing change

ekzhang:
Thanks, will do!

ekzhang:
@cijothomas — I added a changelog entry. I also updated the signature to match those of the other `with_http_client()` methods for the Zipkin and Jaeger exporters, since I realized that the Datadog exporter is the only inconsistent one.

So the type signature is now:

```rust
pub fn with_http_client<T: HttpClient + 'static>(mut self, client: T) -> Self;
```

Let me know if that makes sense!

cijothomas:
@open-telemetry/rust-approvers Anyone knows who is the best person to review datadog components?

shaun-cox (approved):
Thanks for doing this.

TommyCpp (approved):

PR #1313, opened 2023-10-20
Span links stored as Vector instead of EvictedQueue

cijothomas:
Continuing https://github.com/open-telemetry/opentelemetry-rust/pull/1293. This PR modifies Span links - they are now stored as Vec, and the limit enforcement is done without the previous "eviction" behavior. A simpler logic of keeping the 1st "N" links is performed now. Have not done more benchmarks given the perf gains are already proven in prior PRs.


opentelemetry-sdk/src/trace/links.rs
```diff
@@ -0,0 +1,11 @@
+//! # Span Links
+
+use opentelemetry::trace::Link;
+/// Stores span links along with dropped count.
+#[derive(Clone, Debug, Default, PartialEq)]
```
TommyCpp, line 5: Better to annotate the struct with `#[non_exhaustive]` so that we can add more fields without breaking existing code

TommyCpp (approved):

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -44,7 +44,7 @@ pub(crate) struct SpanData {
     /// Span events
     pub(crate) events: crate::trace::EvictedQueue<trace::Event>,
     /// Span Links
-    pub(crate) links: crate::trace::EvictedQueue<trace::Link>,
+    pub(crate) span_links: crate::trace::SpanLinks,
```
shaun-cox, line 47: fwiw, I liked `links` as the name without the `span_` prefix as everything in this struct relates to a span away.  e.g. `attributes` are not called `span_attributes` here.

opentelemetry-sdk/src/trace/links.rs
shaun-cox: If we don't compare them, I'd leave off the derive of `PartialEq`.

shaun-cox: Would it be appropriate to `impl Deref<Vec<Link>>` to avoid the double `span.span_links.links.` usage everywhere?

shaun-cox (approved):

opentelemetry-sdk/src/trace/links.rs
hdost: Since SpanLinks is public I would assume we should keep it `PartialEq` because if none uses it then there's no impact, but if we don't include it then people would have to write a wrapper or extension any time they wanted to

djc: If we're implementing `Deref` (but not `DerefMut`) might make sense to `Deref` to `Target = &[Link]` instead of `Vec<Link>`.

opentelemetry-sdk/src/trace/span.rs
hdost: Yea I'd second that, what's the value from a user perspective?

shaun-cox (approved):

opentelemetry-proto/src/transform/trace.rs
```diff
@@ -93,8 +93,13 @@ pub mod tonic {
                                 dropped_attributes_count: event.dropped_attributes_count,
                             })
                             .collect(),
-                        dropped_links_count: source_span.links.dropped_count(),
-                        links: source_span.links.into_iter().map(Into::into).collect(),
+                        dropped_links_count: source_span.links.dropped_count,
+                        links: source_span
+                            .links
+                            .links
```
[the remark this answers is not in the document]
cijothomas: yes! Looking at it now

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -210,18 +207,34 @@ impl opentelemetry::trace::Tracer for Tracer {
...
                 }
-                links.append_vec(link_options);
-            }
+                SpanLinks {
+                    links,
+                    dropped_count: dropped_count as u32,
+                }
+            } else {
+                SpanLinks::default()
+            };
```
lalitb, line 236: Nit - Above logic can be part of the `From` trait for `SpanLimits` Something like:
```rust
impl From<(Vec<Link>, &SpanLimits)> for SpanLinks {
    fn from((mut links, span_limits): (Vec<Link>, &SpanLimits)) -> Self {
        ...
        ...
       SpanLinks {
            links,
            dropped_count: dropped_count as u32,
        }
    }
 }
````
 But can be as enhancement separately.

lalitb (approved):

opentelemetry-sdk/src/trace/tracer.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/1344

cijothomas:
@jtescher could you re-review and merge if good?

PR #1314, opened 2023-10-23
Move RuntimeChannel type arg T to batch_message_channel and associated types

frigus02:
## Changes

`RuntimeChannel::batch_message_channel` needs to be generic over the message type. The type used to be declared on the `RuntimeChannel<T>` trait. This means a `RuntimeChannel` can only be used with one particular message type, which feels unfortunate.

```rust
fn install<R: RuntimeChannel<??::BatchMessage>>(runtime: R) {
    // Can't use the same runtime here. :-(
    TracerProvider::builder().with_batch_exporter(e, runtime);
    LoggerProvider::builder().with_batch_exporter(e, runtime);
}
```

This change moves the type argument to the `batch_message_channel<T>` function and the associated types `Receiver<T>` and `Sender<T>`. Channels are still specific to a message type, but a `RuntimeChannel` can be used with any number of message types.

```rust
fn install<R: RuntimeChannel>(runtime: R) {
    // It works. :-)
    TracerProvider::builder().with_batch_exporter(e, runtime);
    LoggerProvider::builder().with_batch_exporter(e, runtime);
}
```

This also means the `BatchMessage` types no longer need to be public.

  * not applicable, I think
  * Removes `logs::BatchMessage` and `trace::BatchMessage`
  * RuntimeChannel is public, so all changes to it here are breaking

### DIFF
--- a/opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
+++ b/opentelemetry-contrib/src/trace/exporter/jaeger_json.rs
@@ -8,7 +8,7 @@ use opentelemetry::trace::{SpanId, TraceError};
 use opentelemetry_sdk::{
     export::trace::{ExportResult, SpanData, SpanExporter},
     runtime::RuntimeChannel,
-    trace::{BatchMessage, Tracer, TracerProvider},
+    trace::{Tracer, TracerProvider},
 };
 use opentelemetry_semantic_conventions::SCHEMA_URL;
 use std::collections::HashMap;
@@ -213,7 +213,7 @@ fn opentelemetry_value_to_json(value: &opentelemetry::Value) -> (&str, serde_jso
 ///
 /// [`RuntimeChannel`]: opentelemetry_sdk::runtime::RuntimeChannel
 #[async_trait]
-pub trait JaegerJsonRuntime: RuntimeChannel<BatchMessage> + std::fmt::Debug {
+pub trait JaegerJsonRuntime: RuntimeChannel + std::fmt::Debug {
     /// Create a new directory if the given path does not exist yet
     async fn create_dir(&self, path: &Path) -> ExportResult;
     /// Write the provided content to a new file at the given path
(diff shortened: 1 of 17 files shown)


djc:
What are you trying to achieve with this change? What's your use case?

jtescher:
Looks like CI isn't passing, but this does appear to simplify the external interface which I'm supportive of 👍

frigus02:
Sorry, I should have really described my use case first:

I recently [added live metrics support to opentelemetry-application-insights](https://github.com/frigus02/opentelemetry-application-insights/pull/63). It's a separate span processor, which polls Application Insights and sends a handful of automatically collected metrics (e.g. requests per second) in a much more frequent interval to the server as long as a user is watching the metrics.

I couldn't quite use the batch span processor because the polling interval varies. It's shorter when someone is watching. Otherwise it has the same structure: a separate task/thread that calls the server comminuates with the main task via a channel. I think the nicest API here uses `RuntimeChannel`. Developers already have to provide that when configuring OpenTelemetry and it has everything I need. But I want to use my own message enum and I don't want developers to have to provide 2 different RuntimeChannels.

I want to provide a function like:

```rust
fn install_batch<R: RuntimeChannel>(runtime: R) {
    let live_metrics = LiveMetrics::new(runtime.clone());
    TracerProvider::builder().with_batch_exporter(e, runtime).with_span_processor(live_metrics)
}
```

frigus02:
Ha, but I see this isn't compatible with the MSRV:

> error[E0658]: generic associated types are unstable

I think they were stabilised in https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#generic-associated-types-gats

I might try if that's possible without generic associated types somehow.

djc (approved):
I think we can maybe bump the MSRV to 1.65?

hdost (approved):
I can push a patch on bumping msrv

frigus02:
Yeah, I don't know how to make this work without generic associated types. Bumping the MSRV sounds good to me.

frigus02:
Thanks for bumping the MSRV, hdost@. I rebased and also updated the changelog.

hdost (approved):
Thank you @frigus02 !

TommyCpp (approved):
Thanks!

PR #1315, opened 2023-10-24
[WIP] Suppress nested traces from instrumented dependencies

### LINKED ISSUE #1171: Need a way to suppress telemetry from SDKs own operation

lalitb:
Fixes #1171 

Need few changes before making it ready for review.
 - The SuppressionGuard struct should be runtime agnostic. As of now, it is using task-local from tokio runtime.
 - Need to add suppression logic across traces and logs, and OTLP-gRPC and OTLP-HTTP exporters.
Future change : As of now, Context uses thread-local storage. It can also optionally support task-local storage as supported by otel-js, otel-dotnet. And then Context can be leveraged for suppression.

## Changes

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/logs.rs
+++ b/opentelemetry-otlp/src/exporter/http/logs.rs
@@ -4,12 +4,14 @@ use async_trait::async_trait;
 use http::{header::CONTENT_TYPE, Method};
 use opentelemetry::logs::{LogError, LogResult};
 use opentelemetry_sdk::export::logs::{LogData, LogExporter};
+use opentelemetry_sdk::suppression::Supression;
 
 use super::OtlpHttpClient;
 
 #[async_trait]
 impl LogExporter for OtlpHttpClient {
     async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()> {
+        let _guard = Suppression::new();
         let client = self
             .client
             .lock()
--- a/opentelemetry-otlp/src/exporter/http/logs.rs
+++ b/opentelemetry-otlp/src/exporter/http/logs.rs
@@ -3,15 +3,18 @@ use std::sync::Arc;
 use async_trait::async_trait;
 use http::{header::CONTENT_TYPE, Method};
 use opentelemetry::logs::{LogError, LogResult};
-use opentelemetry_sdk::export::logs::{LogData, LogExporter};
-use opentelemetry_sdk::suppression::Supression;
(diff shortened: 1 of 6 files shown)


opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -3,13 +3,18 @@ use std::sync::Arc;
...
+    suppression::SuppressionGuard,
+};
+//use opentelemetry_sdk::suppression::Supression;
 
 use super::OtlpHttpClient;
 
 #[async_trait]
 impl LogExporter for OtlpHttpClient {
     async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()> {
+        let _guard = SuppressionGuard::new();
```
jtescher, line 17: hm holding this type of guard across await points does not generally work as expected. You would likely need to do a similar future wrapping strategy that we have in https://github.com/open-telemetry/opentelemetry-rust/blob/28d8cf9b6f6fd0f7037d7499e22d35d24f51fdcc/opentelemetry/src/trace/context.rs#L375-L384

lalitb: Thanks for the suggestion @jtescher. I tried the approach you suggested, basically holding the guard within Context, and try to propagate it across async functions. Seems I am still doing something different, as it is not working. Though I will debug if further, just in case you find some issue in the approach, or any other suggestions. I initially thought it would be straightforward  and tokio `task_local` value is propagated automatically through the chain of async method calls, but probably this is not :)

lalitb:
closing this PR as @shaun-cox and @TommyCpp had alternate suggestion to extend the Context structure to support suppressing nested logs from dependencies. Will be raising draft PR to further discuss that approach,

PR #1328, opened 2023-10-31
More distinct naming of metrics constructs in the SDK

shaun-cox:
## Fixes Some Naming Confusions

The SDK currently implements some API `trait`s using `struct`s with identical names as the `trait`. This can get confusing and leads sometimes to using aliasing when `use`ing types, which introduces even more names to have to sort out while reading the code.

## Changes

- rename SDK's `MeterProvider` to `SdkMeterProvider`.  This avoids the clash with the API's `MeterProvider` trait, and opens an easy door to implement other (non-default) `MeterProvider`s with distinguishing names.
- rename `Meter` to `SdkMeter`, for similar reasons.
- rename `InstrumentImpl` to `ResolvedMeasures` as that's all it holds and it doesn't implement any instrument-related trait.
- rename `InstProvider` to `InstrumentResolver`. There is an `InstrumentProvider` trait, which this struct does NOT implement, so avoiding that false association makes things a little clearer (IMO).

### DIFF
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -1,11 +1,13 @@
 use opentelemetry::metrics::Unit;
 use opentelemetry::Key;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
-use opentelemetry_sdk::metrics::{Aggregation, Instrument, MeterProvider, PeriodicReader, Stream};
+use opentelemetry_sdk::metrics::{
+    Aggregation, DefaultMeterProvider, Instrument, PeriodicReader, Stream,
+};
 use opentelemetry_sdk::{runtime, Resource};
 use std::error::Error;
 
-fn init_meter_provider() -> MeterProvider {
+fn init_meter_provider() -> DefaultMeterProvider {
     // for example 1
     let my_view_rename_and_unit = |i: &Instrument| {
         if i.name == "my_histogram" {
@@ -48,7 +50,7 @@ fn init_meter_provider() -> MeterProvider {
         //   Ok(serde_json::to_writer_pretty(writer, &data).unwrap()))
         .build();
     let reader = PeriodicReader::builder(exporter, runtime::Tokio).build();
-    MeterProvider::builder()
+    DefaultMeterProvider::builder()
         .with_reader(reader)
         .with_resource(Resource::new(vec![KeyValue::new(
             "service.name",
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -2,12 +2,12 @@ use opentelemetry::metrics::Unit;
 use opentelemetry::Key;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
 use opentelemetry_sdk::metrics::{
-    Aggregation, DefaultMeterProvider, Instrument, PeriodicReader, Stream,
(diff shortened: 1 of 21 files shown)


hdost:
Many of the other sdks opt for something like `SdkMeterProvider` and the like, but in general I'm in favor of re-naming. I do like the idea of it being `DefaultMeterProvider` especially since it's more clear that it's the default implementation 👍🏼.

cijothomas:
> Many of the other sdks opt for something like `SdkMeterProvider` and the like, but in general I'm in favor of re-naming. I do like the idea of it being `DefaultMeterProvider` especially since it's more clear that it's the default implementation 👍🏼.

I vote in below order of preference:
SdkMeterProvider
MeterProviderSdk
DefaultMeterProvider - at the lowest, as I think this can potentially be confused with default being "no-op". Probably just my confusion!

jtescher:
I think there being two defaults (API default and SDK default) make the default naming a little more tricky. Have other language impls found clever solutions? the current sdk meter provider matches the [MeterProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#meterprovider) naming in the spec which has some advantages too, although it is quite unfortunate that the traits then have to be imported as `_` or similar workarounds.

hdost:
Yea perhaps it is then best to stick with `SdkMeterProvider` and the like.

lalitb:
I don't have strong opinion on naming, but whatever we choose, good to be consistent with the other signals. As of now. logs and traces are using same name in both API and SDK (`Tracer/TracerProvider` and `Logger/LoggerProvider`).

While it may not be the best example, in `otel-cpp` we follow the existing practice of using the same names for these constructs in the API and SDK, under different namespace, and resort to aliasing when naming conflicts arise.

shaun-cox:
> I don't have strong opinion on naming, but whatever we choose, good to be consistent with the other signals. As of now, logs and traces are using same name in both API and SDK (`Tracer/TracerProvider` and `Logger/LoggerProvider`).
> 
> While it may not be the best example, in `otel-cpp` we follow the existing practice of using the same names for these constructs in the API and SDK, under different namespace, and resort to aliasing or using fully qualified names in case of conflict.

Agree.  If everyone likes the new terms `SdkMeterProvider`, `SdkMeter`, etc. I'll either extend this PR to cover changing `TracerProvider`, `LoggerProvider`, or I'll open a new one, as others prefer.

cijothomas:
> > I don't have strong opinion on naming, but whatever we choose, good to be consistent with the other signals. As of now, logs and traces are using same name in both API and SDK (`Tracer/TracerProvider` and `Logger/LoggerProvider`).
> > While it may not be the best example, in `otel-cpp` we follow the existing practice of using the same names for these constructs in the API and SDK, under different namespace, and resort to aliasing or using fully qualified names in case of conflict.
> 
> Agree. If everyone likes the new terms `SdkMeterProvider`, `SdkMeter`, etc. I'll either extend this PR to cover changing `TracerProvider`, `LoggerProvider`, or I'll open a new one, as others prefer.

My personal preference is to fix Metrics in this PR, and follow up with trace/logs in separate PR!

jtescher:
Wonder if we could have the best of both worlds by manually implementing the methods on the sdk's `MeterProvider` so the trait didn't need to be imported at all in the common case? could simplify the examples and general usage

hdost (approved):

jtescher (approved):
Can do further improvements in follow up PRs as well, can merge as is for the sake of moving forward

Issue #1189, opened 2023-08-07
Logs API/SDK - Move LogRecord implementation details from API to SDK


lalitb:
As of now, the Logs API contains the implementation details for creation of [LogRecord](https://github.com/open-telemetry/opentelemetry-rust/blob/77db498ead77b359ee93f443b2ae68b3cf22a6cb/opentelemetry-api/src/logs/record.rs#L11) object. This should be ideally moved to SDK. The API should have minimal implementation details and storage allocation, and hence minimal performance overhead. Anyone wanting to create custom logs SDK will otherwise have to bear the baggage of these allocations happening in API.

This is also how trace signal is implemented, API has a `Span` trait, and it's Noop implementation, while the implementation details are in the `Span` structure in SDK.

Issue #761, opened 2022-03-14
Infinite trace generator


hamiltop:
When I use the `opentelemetry_datadog` library with `reqwest` as the client, each batch of spans will generate an additional 2 spans. These spans look like this:

```
[[{'service': 'test', 'name': 'opentelemetry-datadog', 'resource': 'encode_headers', 'trace_id': 9313706354471130378, 'span_id': 2181458161362850175, 'parent_id': 0, 'start': 1647298645595811251, 'duration': 156890, 'error': 0, 'meta': {'busy_ns': '66163', 'code.namespace': 'hyper::proto::h1::role', 'code.filepath': '/home/hamiltop/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.17/src/proto/h1/role.rs', 'code.lineno': '1061', 'idle_ns': '90678'}}], [{'service': 'test', 'name': 'opentelemetry-datadog', 'resource': 'parse_headers', 'trace_id': 11785853278508097401, 'span_id': 9996945950193117364, 'parent_id': 0, 'start': 1647298645596617264, 'duration': 159966, 'error': 0, 'meta': {'code.namespace': 'hyper::proto::h1::role', 'code.lineno': '1183', 'busy_ns': '119852', 'code.filepath': '/home/hamiltop/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.17/src/proto/h1/role.rs', 'idle_ns': '36137'}}]]
```

I believe this is from the http request to datadog to save the original spans. Now, the request to send these 2 spans also generates 2 new spans, which generate 2 new spans, etc.

To reproduce:

1) Run this script as a mock datadog agent: https://gist.github.com/kongo2002/a4d560dca89e24bf27c8d59dc63f4b1e This will echo all incoming traces to stdout for easy inspecting.
2) Run this code:
```
use opentelemetry_datadog::{new_pipeline, ApiVersion};
use tracing::trace_span;
use tracing_subscriber::layer::SubscriberExt;

#[tokio::main]
async fn main() {
    let _tracer = new_pipeline()
        .with_agent_endpoint("http://localhost:8126")
        .with_service_name("test")
        .with_version(ApiVersion::Version03)
        .install_batch(opentelemetry::runtime::TokioCurrentThread)
        .unwrap();

    let telemetry = tracing_opentelemetry::layer().with_tracer(_tracer);

    let subscriber = tracing_subscriber::Registry::default().with(telemetry);

[...]

PR #1330, opened 2023-11-01
[WIP] Suppression of nested logs from dependencies

### LINKED ISSUE #1171: Need a way to suppress telemetry from SDKs own operation

lalitb:
Raising this PR as alternate approach to fix #1171, by extending the Context structure to store the suppression flag. The PR is to discuss this approach. The earlier approach #1315 was based on tokio runtime, and generalizing it would have been difficult.

**Change summary**

1. Move `FutureExt` and `WithContext` constructs from `trace/context.rs` to `context.rs`.
2. Add suppress_logs flag in `Context` struct
```rust
pub struct Context {
    #[cfg(feature = "trace")]
    pub(super) span: Option<Arc<SynchronizedSpan>>,
    pub suppress_logs: bool,
    entries: HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>,
}
```
3. Add  `with_suppression` method in Context to enable suppression:
```rust
impl Context {
  ...
    pub(super) fn with_suppression(&self) -> Self {
        Context {
            suppress_logs: true,
            entries: self.entries.clone(),
            ..self.clone()
        }
    }
}
```

4. Add `with_current_context_supp` method in FutureExt:
```rust
pub trait FutureExt: Sized {
    ...
    fn with_current_context_supp(self) -> WithContext<Self> {
        let otel_cx = Context::current().with_suppression();
        self.with_context(otel_cx)
    }
}
```
5. Suppress in Exporter::export method:
```rust
    async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()> {
        async {
            // export logic
       }
        .with_current_context_supp()
        .await
    }
```

6. Check if suppression is enabled in LogProcessor:
```rust
impl<R: RuntimeChannel<BatchMessage>> LogProcessor for BatchLogProcessor<R> {
    ...
    fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
        !Context::current().suppress_logs
    }
}
```

7. How to test:

run opentelemetry-otlp/examples/basic-otlp-http/ example, while collector is running (using docker-compose up).

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -105,7 +105,7 @@ impl LogProcessor for SimpleLogProcessor {
 
     #[cfg(feature = "logs_level_enabled")]
     fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
-        !Context::current().suppress_logs
+        !Context::current().suppression
     }
 }
 
@@ -134,7 +134,7 @@ impl<R: RuntimeChannel<BatchMessage>> LogProcessor for BatchLogProcessor<R> {
 
     #[cfg(feature = "logs_level_enabled")]
     fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
-        !Context::current().suppress_logs
+        !Context::current().suppression
     }
 
     fn force_flush(&self) -> LogResult<()> {
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -19,6 +19,8 @@ use std::{
     time::Duration,
 };
 
+use opentelemetry::Context;
+
 /// The interface for plugging into a [`Logger`].
 ///
 /// [`Logger`]: crate::logs::Logger
@@ -103,7 +105,7 @@ impl LogProcessor for SimpleLogProcessor {
 
     #[cfg(feature = "logs_level_enabled")]
     fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
-        true
+        !Context::current().suppress_logs
     }
 }
 
@@ -132,7 +134,7 @@ impl<R: RuntimeChannel<BatchMessage>> LogProcessor for BatchLogProcessor<R> {
 
     #[cfg(feature = "logs_level_enabled")]
     fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
-        true
+        !Context::current().suppress_logs
     }
 
     fn force_flush(&self) -> LogResult<()> {
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -12,7 +12,6 @@ use std::marker::PhantomData;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context as TaskContext, Poll};
-use std::thread;
 
 thread_local! {
     static CURRENT_CONTEXT: RefCell<Context> = RefCell::new(Context::default());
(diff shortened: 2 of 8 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -103,7 +105,7 @@ impl LogProcessor for SimpleLogProcessor {
 
     #[cfg(feature = "logs_level_enabled")]
     fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
-        true
+        !Context::current().suppress_logs
```
cijothomas, line 108: we should not tie this to the `logs_level_enabled` feature.

opentelemetry/src/context.rs
```diff
@@ -78,6 +84,7 @@ thread_local! {
 pub struct Context {
     #[cfg(feature = "trace")]
     pub(super) span: Option<Arc<SynchronizedSpan>>,
+    pub suppress_logs: bool,
```
cijothomas, line 87: You are experimenting with logs, but the final expectation is that this will be applicable to traces and logs? If so, the name can be suppress_instrumentation

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Thanks for the comment. I believe placing it within `logs_level_enabled` is preferable for early suppression detection, which would avoid the need to generate a `LogRecord`. I also plan to add this validation in `Logger::emit()` for instances when the `logs_level_enabled` feature is not enabled, will be doing it before making it ready for review.

opentelemetry/src/context.rs
lalitb: Good comment. Got similar suggestion from @shaun-cox. I changed it to `suppression`, but probably `suppress_instrumentation` is better name.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: Agree doing it at the earliest possible place is perf friendly!

Can we do `fn event_enabled` always (i.e even without logs_level_enabled feature) and check Suppression inside that? And if logs_level_enabled feature is enabled, then do the additional check?

lalitb: Yes we can do that, adding this validation within appender for logs and tracing, or better in LogRecord->Emit() or similar method once have implemented #1189. As of now, I am still bit struggling with propagating the future state across threads used by async task, so will revisit that once it is covered.

lalitb:
As this PR came into discussion in today's community meeting, summarising the challenge with the current solution - 
 
The PR adds the suppress_logs flag in the Context, and the Context object is propagated inside the async task. However, wrapping the future with `FutureExt::WithExt` will not cascade the wrapping behavior to all internal futures. So if these internal futures (which could be from external (say) Hyper crate) get invoked in a separate thread, they won't see the suppression flag from context. And we can't explicitly apply the wrapper to each async function of third-party crates.

lalitb:
closing this PR. The changes would be done based on the options discussed here - https://github.com/open-telemetry/opentelemetry-rust/issues/761#issuecomment-2121548678.

Issue #1331, opened 2023-11-01
Non-portable `sed` usage in semconv code generation


kriswuollett:
When running on macOS:

    sed: 1: "src/lib.rs": unterminated substitute in regular expression

Caused by [non-portable `sed` specification of in-file substitution (backup)](https://github.com/open-telemetry/opentelemetry-rust/blob/ed97a1bd9efbbde2f080bdad7331b91f277b9f6b/opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh#L46). The fix for this already exists in [open-telemetry/opentelemetry-cpp](https://github.com/open-telemetry/opentelemetry-cpp/blob/ca08c5a34ad4af1b6b392c8bf5fc26a5210a2f2c/tools/format.sh#L16-L19).

PR #1332, opened 2023-11-01
macOS dev fix for `sed`

kriswuollett:
- Allows `sed` to run correctly for code generation on macOS

Fixes: #1331


lalitb (approved):
Thanks for the fix. The bug got introduced  in #1312, didn't test it on MacOS :(

shaun-cox (approved):

Issue #1043, opened 2023-04-25
SDk/Sampler concepts leaked in API


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/trace/tracer.rs#L277
`SamplingResult` is the type returned by Samplers, and it is a pure SDK concept. 
If this was unintentional, we should remove it before 1.0 stable release.

Similar: https://github.com/open-telemetry/opentelemetry-rust/issues/1042

cijothomas:
I think the only "Sampling" related concepts that should be part of Tracing API are:
1. spanContext TraceFlags for SAMPLED.
2. span.IsRecording

Based on the comment (https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/tracer.rs#L174-L176), I think this is done to accommodate tracing-opentelemetry crate. Related discussion around that crate : https://github.com/open-telemetry/opentelemetry-rust/discussions/1032

cijothomas:
@hdost Moving back to Tracing API stable, as the fix for this (if we chose to), would be in the API itself!

cijothomas:
@open-telemetry/rust-approvers 
The natural fix for this would be to simply remove this from the API. But that'd probably (or surely?) break tracing-opentelemetry, unless tracing-opentelemetry can depend on the SDK (it does seem to [depend on SDK today](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/Cargo.toml#L29)...)... But even then, there is no such support in OTel spec to support externally sampled spans..

Given the popularity of `tracing` crate for both logging and distributed tracing in Rust ecosystem, I think we should keep this, but make this a well-documented decision by OTel Rust maintainers/approvers. At minimum, we need to[ document this as a conscious decision](https://github.com/open-telemetry/opentelemetry-rust/issues/1297) (and not an accidental leak/bug), and document we commit to keep support tracing-opentelemetry by specially treating the crate. 

Please share your thoughts on how best to proceed.

My proposal:
1. Document this in the same place https://github.com/open-telemetry/opentelemetry-rust/issues/1297 prefers (that is undecided as of now)
2. Call this out in this repo's main readme file itself. I can work on some wordings once I get an okay for this direction.

lalitb:
+ 1 on keeping it. We can also keep document it as doc comments for SamplingResult and SamplingDecision.

djc:
[...]

PR #1334, opened 2023-11-01
Add generated modules that output const &str for `tracing` compatibility

### LINKED ISSUE #1320: Need `&'static str`s from `semantic-conventions` for `tracing` field recording
I'd like to do something like the following with `tracing`:

```rust
event!(
    Level::INFO,
    { semconv::trace::CODE_FUNCTION } = "update_user",
    { semconv::trace::CODE_NAMESPACE } = "com.my-org",
    { my_org::semconv::USER_ID } = request.user_id,
    "Updating user"
);
```

The `tracing` crate requires values that look like [Rust identifiers, string literals, or constant expressions](https://github.com/tokio-rs/tracing/blob/f93cfa087e6ebdcbd8ecdcccca47d73c3a89ab94/tracing/src/lib.rs#L322-L333). But the string for values like code function are wrapped up in a `Key` which although `const` does not satisfy the `format_args` that is used by `tracing`: 

```
error: format argument must be a string literal
  --> sample/src/tracing_test/mod.rs:28:9
   |
28 |         { semconv::trace::CODE_FUNCTION } = "update_user",
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
help: you might be missing a string literal to format with
   |
28 |         "{} {} {} {}", { semconv::trace::CODE_FUNCTION } = "update_user",
   |         ++++++++++++++
```

I'd think that the easiest solution would be to create another submodule that is code generated with the just primary source of the key strings from which the module with the `const Key`s are derived, e.g.,:

```rust
const CODE_FUNCTION: &str = "code.function";
```

That way something like the above example would work for `tracing` as well as any other context in which `const` / `&'static str` is required.

Although the code generation can be on my end, automatically or manually, this seems basic enough and useful to others that I'd think it should be implemented here? Or is there some Rust technique or library I haven't learned of yet to make use of the existing constants instead of copy and pasting raw `str` values like `"code.function"` everywhere?

kriswuollett:
- Adds generated modules for `const` `&str` access to semantic convention keys, fixes #1320, via `semconv` submodule


shaun-cox (approved):
Thanks for the contribution!

opentelemetry-semantic-conventions/src/semconv/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod resource;
```
lalitb, line 1: Do we need semconv directory along with the generated files here, now that we are directly modifying the existing generated files ?

kriswuollett: No we don't need the dir. Removed it my recent force push. Will take off the WIP soon once I double check the tests and other code.

opentelemetry-semantic-conventions/src/semconv/resource.rs
```diff
@@ -0,0 +1,882 @@
+// DO NOT EDIT, this is an auto-generated file
```
jtescher, line 1: Was this file added by mistake? seems to be a dup of src/resource.rs

opentelemetry-semantic-conventions/src/semconv/trace.rs
```diff
@@ -0,0 +1,1599 @@
+// DO NOT EDIT, this is an auto-generated file
```
jtescher, line 1: Same with this one

opentelemetry-semantic-conventions/scripts/templates/header_resource.rs
```diff
@@ -7,14 +7,31 @@
 //!
 //! ## Usage
 //!
+//! `tracing`:
+//! 
 //! ```
+//! use opentelemetry_semantic_conventions as semconv;
+//! use tracing::span;
```
jtescher, line 14: Does tracing now need to be a dev dep?

jtescher (changes requested):
Overall supportive of this change, some nits left

opentelemetry-semantic-conventions/src/semconv/resource.rs
kriswuollett: Left in by mistake, removing.

opentelemetry-semantic-conventions/scripts/templates/header_resource.rs
kriswuollett: Seems like it should be, adding.

opentelemetry-semantic-conventions/scripts/templates/header_resource.rs
```diff
@@ -7,14 +7,31 @@
...
+//! [`tracing`]:
+//! 
 //! ```
+//! use opentelemetry_semantic_conventions as semconv;
+//! use tracing::span;
+//!
+//! let span = span!(
+//!     LEVEL::INFO,
+//!     "handle_request",
+//!     { semconv::resource::SERVICE_NAME = "my-service" },
```
cijothomas, line 19: I am trying to understand this line here.. Is this adding service name and ns to every span as attributes? If yes, that is not something we should show/encourage, as Resource is expected to be attached to the Provider level *not* to each span/log/metric. (In somebackends, it maybe ultimately stored for each span/log, but thats at exporter level//backend level)

opentelemetry-semantic-conventions/scripts/templates/header_resource.rs
```diff
@@ -7,14 +7,31 @@
 //!
 //! ## Usage
 //!
+//! [`tracing`]:
```
cijothomas, line 10: I think examples in this repo should show OTel Tracing API, not `tracing` crate macros!

[I agree with the decision](https://github.com/open-telemetry/opentelemetry-rust/issues/1043#issuecomment-1776529519) to specially treat `tracing` in the API/SDK given the importance of tracing crate to Rust ecosystem, but the examples in this repo should be showing components shipped from this repo.

cijothomas (changes requested):
Changes looks good! 

I am leaving my review as "request-changes", specifically for addressing the below points:
https://github.com/open-telemetry/opentelemetry-rust/pull/1334/files#r1387150519
https://github.com/open-telemetry/opentelemetry-rust/pull/1334/files#r1387146406

opentelemetry-semantic-conventions/scripts/templates/header_resource.rs
kriswuollett: OK, just removed.

opentelemetry-semantic-conventions/scripts/templates/header_resource.rs
kriswuollett: OK, lacking a specific suggested example to do, if any, for `tracing`, I've just removed the additional example block. I'll leave it to any follow ups that anyone would like to do for docs.

jtescher (approved):

PR #1343, opened 2023-11-06
Prepare for v0.21.0 release

jtescher:
See individual crate changelogs for details.


hdost (approved):
I think a number of people will be happy about this.  ☺️

lalitb (approved):

shaun-cox (approved):

cijothomas:
4 months ago, I'd have asked to wait for some more breaking changes, but I think we should let the release proceed, and **make sure we continue to do regular releases**.

PR #1346, opened 2023-11-07
[Tracing Appender] Propagate event name to exporters

lalitb:
As per the recent changes in Tokio Tracing library - https://github.com/tokio-rs/tracing/pull/2699, it now allows setting event name in tracing macros. E.g., in

```rust
error!(name: "my-event-name", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
```

As lots of backends support event/log name (ElasticSearch, Fluentd, Sentry etc), it should be possible for users to specify event name using tracing macros, and export them to these backend through custom otel-exporters.

The PR modifies tracing subscriber to propagates this event-name to exporters by adding it to attributes. And if the attributes already have the key with the same name, that would override the "name" metadata field. Also, modified user_events log exporter to utilize this field. As of now, user_events exporter expect user to provide this field as "event_name" attribute through tracing macros.

TBD - It is also good to have parity across the metadata field in tracing event, and the log-record, at-least log-record structure should optionally have the additive field from the event metadata (and `name` is one such field). I plan to subsequently propose adding this field in LogRecord implementation separately.

### DIFF
--- a/opentelemetry-appender-tracing/examples/basic.rs
+++ b/opentelemetry-appender-tracing/examples/basic.rs
@@ -23,6 +23,6 @@ fn main() {
     let layer = layer::OpenTelemetryTracingBridge::new(&provider);
     tracing_subscriber::registry().with(layer).init();
 
-    error!(target: "my-system", event_id = 20, event_name = "my-event_name", user_name = "otel", user_email = "otel@opentelemetry.io");
+    error!(target: "my-system", name: "my-event-name", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
     drop(provider);
 }
--- a/opentelemetry-appender-tracing/examples/basic.rs
+++ b/opentelemetry-appender-tracing/examples/basic.rs
@@ -23,6 +23,6 @@ fn main() {
     let layer = layer::OpenTelemetryTracingBridge::new(&provider);
     tracing_subscriber::registry().with(layer).init();
 
-    error!(target: "my-system", name: "my-event-name", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
+    error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
     drop(provider);
 }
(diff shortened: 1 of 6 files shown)


opentelemetry-appender-tracing/examples/basic.rs
```diff
@@ -23,6 +23,6 @@ fn main() {
     let layer = layer::OpenTelemetryTracingBridge::new(&provider);
     tracing_subscriber::registry().with(layer).init();
 
-    error!(target: "my-system", event_id = 20, event_name = "my-event_name", user_name = "otel", user_email = "otel@opentelemetry.io");
+    error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
```
cijothomas, line 26: ❤️ this!

opentelemetry-user-events-logs/src/logs/exporter.rs
```diff
@@ -52,7 +52,8 @@ pub(crate) struct UserEventsExporter {
 }
 
 const EVENT_ID: &str = "event_id";
-const EVENT_NAME: &str = "event_name";
+const EVENT_NAME_PRIMARY: &str = "event_name";
+const EVENT_NAME_SECONDARY: &str = "name";
```
cijothomas, line 56: Maybe reverse primary and secondary?

Lets add a comment indicating why we need to look for attributes with either of these names.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -103,6 +103,11 @@ where
         log_record.severity_number = Some(map_severity_to_otel_severity(meta.level().as_str()));
         log_record.severity_text = Some(meta.level().to_string().into());
 
+        // add the `name` metadata to attributes
+        // TBD - Propose this to be part of log_record metadata.
+        let vec = vec![("name", meta.name())];
```
cijothomas, line 108: just to confirm: OTLPExporters will export this as "name" ?

opentelemetry-user-events-logs/src/logs/exporter.rs
lalitb: The `name` metadata field defaults to `file:line` if it is not set using event macros, or with older version of tracing. So, making it primary will not allow to use `event_name`.

opentelemetry-appender-tracing/src/layer.rs
lalitb: yes, this will go within attributes across all the log exporters.

cijothomas (approved):
Please add changelogs too!

PR #1348, opened 2023-11-07
prometheus: Ignore unknown instrument units

jtescher:
## Motivation

The [metric unit semantic conventions] suggest that integer counts should use annotations (e.g. `{packet}`), which breaks the current unit appending logic as they are not properly escaped.

[metric unit semantic conventions]: https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/general/metrics.md#instrument-units

## Solution

Ignore unknown units (including annotations) as other language implementations currently do. This change also removes the `$` mapping as it is not UCUM.

### DIFF
--- a/opentelemetry-prometheus/CHANGELOG.md
+++ b/opentelemetry-prometheus/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## vNext
 
+### Fixed
+
+- Fix UCUM annotation escaping by ignoring unknown instrument units and annotations (#1348)
+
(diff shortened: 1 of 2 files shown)


TommyCpp (commented):
hmm IIUC by doing so users can only use UCUM as instrument?

jtescher:
@TommyCpp yeah this seems to be the common choice currently (e.g. go, java, etc)

TommyCpp (approved):

PR #1351, opened 2023-11-07
Return consistent `Meter` for a given `MeterProvider`

jtescher:
## Changes

Adds a meter cache to return consistent `Meter` instances for the same scope. Meters have their own instrument caches so this avoids double instrument registration.


shaun-cox:
@jtescher, I would kindly request we accept or reject #1328 before merging this, as otherwise I'll have merge conflicts to fix in #1328 I suspect... (and I've been waiting a day or so for final approval.) Thanks.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -123,15 +123,23 @@ impl MeterProvider for SdkMeterProvider {
...
-                Arc::new(NoopMeterCore::new())
-            };
+        if self.is_shutdown.load(Ordering::Relaxed) {
+            return Meter::new(Arc::new(NoopMeterCore::new()));
+        }
+
+        let scope = Scope::new(name, version, schema_url, attributes);
 
-        Meter::new(inst_provider)
+        if let Ok(mut meters) = self.meters.lock() {
```
cijothomas, line 132: Given the fact that obtaining `meter` is not free, we should mention in the documentation that, meter (and instruments) should be obtained once and cached by user to avoid perf hit. Not blocking this PR, but wanted to call it out here.

jtescher: I think we can do some benchmarking in a follow up and have better guidelines on metric perf in general. The use case for accessing a meter is different than a tracer or logger as instruments are more likely to be a fixed set. And even in those cases repeatedly creating/accessing instruments could be the place that needs perf considerations.

shaun-cox (approved):

lalitb (approved):

PR #1354, opened 2023-11-08
Improve contributing guidelines

cijothomas:
Hope they makes sense!


CONTRIBUTING.md
```diff
@@ -64,20 +70,29 @@ the repo to catch any issues locally.
...
 
 A PR is considered to be **ready to merge** when:
 
-* It has received approval from Collaborators/Maintainers.
-* Major feedback is resolved.
+* It has received approval from
+  [Approvers](https://github.com/open-telemetry/community/blob/main/community-membership.md#approver).
+  /
+  [Maintainers](https://github.com/open-telemetry/community/blob/main/community-membership.md#maintainer).
+* Major feedbacks are resolved.
```
lalitb, line 91: we can decide on the number of approvals good enough to merge, depending on the size and other factors. Can be done separate of this PR.

lalitb (approved):

CONTRIBUTING.md
TommyCpp: Yeah I'd like enforce 2 approval before merging given we have more approvers

cijothomas: For now, maybe maintainers can decide to wait for 2 approvals before merging on a case-by-case basis, instead of requiring/enforcing it for every PRs (just to keep the momentum going!). Would this work?
(That is how many others SIGs are operating. But happy to revisit the idea for this SIG)

TommyCpp: Yeah we should call it out in documentation though. So that people will review the PR with 1 approval

cijothomas: Do we want to? I guess maintainers/approvers are already doing so.... Writing something like this in official doc, without being too restrictive is hard 😄

shaun-cox (approved):

CONTRIBUTING.md
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/pull/1354/commits/6a53393c0fa233e3928bef877dbe7a89d6e0babd Added some wording! @TommyCpp see if that makes sense!

TommyCpp: Thanks for addressing that

TommyCpp (approved):

PR #1356, opened 2023-11-08
Modify existing http client/server example to demonstrate context propagation across async delegates

lalitb:
## Changes
Modify the existing [traceresponse ](https://github.com/open-telemetry/opentelemetry-rust/tree/v0.21.0/examples/traceresponse) example to make it more complete as an HTTP client/server example with below changes:

 **Server** - Enhance the  server's request handling, by adding support for `/echo` and `/health` endpoints. Upon receiving the request, the server now creates a child span, linked to the originating remote span, and the request is forwarded to its respective delegate async task. Furthermore, within each async task, a subsequent child span is spawned, parented by the initial child span. This nested span creation exemplifies the effective propagation of tracing context through the multiple layers of async execution.

**Client** - The client sends requests for `/echo` and `/health` within context of the client root span.

Also renamed the example from `traceresponse` to `tracing-http`.


examples/tracing-http/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+    let parent_cx = extract_context_from_request(&req);
+    let response = {
+        // Create a span parenting the remote client span.
+        let tracer = global::tracer("example/server");
+        let mut span = tracer
+            .span_builder("router")
+            .with_kind(SpanKind::Server)
+            .start_with_context(&tracer, &parent_cx);
+
+        span.add_event("dispatching request", vec![]);
```
cijothomas, line 65: seeing this makes me realize we can potentially change the api to do:
```
add_event(name)
add_event_with_attributes(name,attrib)
```

Separate topic, just shared as I saw it here!

examples/tracing-http/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let cx = Context::default().with_span(span);
+        match (req.method(), req.uri().path()) {
+            (&hyper::Method::GET, "/health") => handle_health_check(req).with_context(cx).await,
+            (&hyper::Method::GET, "/echo") => handle_echo(req).with_context(cx).await,
+            _ => {
+                let mut not_found = Response::default();
+                *not_found.status_mut() = StatusCode::NOT_FOUND;
+                Ok(not_found)
```
cijothomas, line 74: could we put this status code to the "router" span as well?

examples/tracing-http/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+        let mut span = tracer
+            .span_builder("router")
+            .with_kind(SpanKind::Server)
+            .start_with_context(&tracer, &parent_cx);
+
+        span.add_event("dispatching request", vec![]);
+
+        let cx = Context::default().with_span(span);
+        match (req.method(), req.uri().path()) {
+            (&hyper::Method::GET, "/health") => handle_health_check(req).with_context(cx).await,
```
cijothomas, line 69: what would have happened if we did not explicitly pass `cx` "with_context" ?

lalitb: The `handle_health_check` async function can run in separate thread, so `with_context` ensure the context (passed as argument) gets attached to the current_context during [polling ](https://github.com/open-telemetry/opentelemetry-rust/blob/47881b20a2b8e94d8e1cdbd4877852dd74cc07de/opentelemetry/src/trace/context.rs#L380)this future.

examples/tracing-http/src/server.rs
lalitb: yes, I too felt so.

examples/tracing-http/src/server.rs
cijothomas: Thanks for explaining. I see this is already documented as well : https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/trace/tracer.rs#L73

examples/tracing-http/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+        propagator.extract(&HeaderExtractor(req.headers()))
+    })
+}
+
+// Separate async function for the handle endpoint
+async fn handle_health_check(_req: Request<Body>) -> Result<Response<Body>, Infallible> {
+    let tracer = global::tracer("example/server");
+    let mut span = tracer
+        .span_builder("health_check")
+        .with_kind(SpanKind::Server)
```
cijothomas, line 29: This should be the internal span kind, right?

examples/tracing-http/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+        .start(&tracer);
+    span.add_event("Health check accessed", vec![]);
+    let cx = Context::default().with_span(span);
+    let mut res = Response::new(Body::from("Server is up and running!"));
+    let response_propagator: &dyn TextMapPropagator = &TraceContextResponsePropagator::new();
```
cijothomas, line 34: hint that returning context is optional, and based on the draft spec, linking to https://w3c.github.io/trace-context/#traceresponse-header

examples/tracing-http/src/server.rs
```diff
@@ -0,0 +1,107 @@
...
+    response_propagator.inject_context(&cx, &mut HeaderInjector(res.headers_mut()));
+    Ok(res)
+}
+
+// Separate async function for the echo endpoint
+async fn handle_echo(req: Request<Body>) -> Result<Response<Body>, Infallible> {
+    let tracer = global::tracer("example/server");
+    let mut span = tracer
+        .span_builder("echo")
+        .with_kind(SpanKind::Server)
```
cijothomas, line 44: internal span?

cijothomas (approved):

cijothomas:
@lalitb lets hold off this during the "move to contrib" phase.. The trace-response is moved to contrib, so lets add this as a separate working example in main repo.

lalitb:
closing this as discussed above. Will create a separate PR with the example without tracereponse header.

PR #1361, opened 2023-11-09
Remove span_builder from Tracer

cijothomas:
## Changes

Removed `span_builder` from `Tracer`. The returned `SpanBuilder` was anyway not connected to the `Tracer`, so users still had to provide a `Tracer` to the `start` method! Removed the method in favor of just using `SpanBuilder`'s associated method to get one!

I am thinking of this as a necessary cleanup of unnecessary public APIs. If there is a desire in the future to bring back this method, we can do so, but this time, it should be linked to Tracer automatically, so there won't be any need to pass it to the "start" !

### DIFF
--- a/examples/traceresponse/src/client.rs
+++ b/examples/traceresponse/src/client.rs
@@ -1,9 +1,10 @@
 use hyper::http::HeaderValue;
 use hyper::{body::Body, Client};
+use opentelemetry::trace::SpanBuilder;
 use opentelemetry::{
     global,
     propagation::TextMapPropagator,
-    trace::{SpanKind, TraceContextExt, Tracer},
+    trace::{SpanKind, TraceContextExt},
     Context, KeyValue,
 };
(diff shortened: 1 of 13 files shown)


jtescher:
Yeah the API here is a little unfortunate as rust doesn't have variadic params. The intention was to have a "simple" api via `tracer.span(...)` and a "full" api via `tracer.span_builder(...).build()` but it was always awkward as the builder needed a reference back to the tracer to actually create the span. Possibly worth considering if there are better solutions than the current builder approach?

shaun-cox (approved):

hdost (changes requested):
I was say let's have a solution for this first since we should probably have a solution to having span builder. Ideally we could have the span builder maintain a reference to the Tracer? Not sure why we couldn't do that.

jtescher:
@hdost yeah giving the builder a lifteime and a reference should be fine

cijothomas:
Closing old, inactive PRs. 
This could be resurrected in the future.

PR #1369, opened 2023-11-14
make severity of level computations const

shaun-cox:
## Changes

This is a drive-by change to improve something that I noticed could be a `const fn` instead of `fn` and to avoid conversion to and comparison of string slices.

### DIFF
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -18,10 +18,9 @@ where
 {
     fn enabled(&self, _metadata: &Metadata) -> bool {
         #[cfg(feature = "logs_level_enabled")]
-        return self.logger.event_enabled(
-            map_severity_to_otel_severity(_metadata.level()),
-            _metadata.target(),
-        );
+        return self
+            .logger
+            .event_enabled(severity_of_level(_metadata.level()), _metadata.target());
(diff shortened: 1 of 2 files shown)


opentelemetry-appender-log/src/lib.rs
```diff
@@ -61,7 +60,7 @@ where
     }
 }
 
-fn map_severity_to_otel_severity(level: Level) -> Severity {
+const fn severity_of_level(level: Level) -> Severity {
```
lalitb, line 63: nit - Can this take reference to `Level` ?

lalitb (approved):
Thanks.

opentelemetry-appender-log/src/lib.rs
shaun-cox: It can, it's just that its a `Copy` item and the `log` crate returns them by value, so ergonomically it was easier to not pass by reference.

Yes, the other file passes it by reference, but that is because the `tracing` API to get the level returns a reference to Level.

cijothomas:
> CI / test (stable) (pull_request) Successful in 15m

❤️ After the move to contrib! This used to be around 20-30mins before!

cijothomas (approved):

hdost:
@shaun-cox performance is his business 🤘

PR #1371, opened 2023-11-14
Remove API for Creating Histograms with signed integers.

lalitb:
Fixes #
As per the otel [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#histogram), the value to be recorded with histogram instrument should be non-negative. So removing the existing method to record signed integer values.

``
The value is expected to be non-negative. This API SHOULD be documented in a way to communicate to users that this value is expected to be non-negative. This API SHOULD NOT validate this value, that is left to implementations of the API.
``

## Changes


jtescher (approved):
Always nice to support a smaller API surface, can we get a changelog entry?

TommyCpp (approved):

shaun-cox (approved):

PR #1375, opened 2023-11-15
Fix metric collections during PeriodicReader shutdown

lalitb:
A nit fix in metric collection during `PeriodicReader` shutdown.

The `PeriodicReader` performs a one last collection before shutting down. However, currently, the shutdown flag is set before the  message is sent to initiate the collection process asynchronously. This result in situation where collection error's out as the reader is already shutdown:
```
     Finished dev [unoptimized + debuginfo] target(s) in 6.60s
     Running `/<path-to>/opentelemetry-rust/target/debug/metrics-basic`
     Error: Other("[Other(\"reader is shut down\")]")
```
The proposed PR resolves this by ensuring that the shutdown flag is only set after the completion of the collection."

## Changes


jtescher (approved):
Good call

Issue #1297, opened 2023-10-13
Document spec deviations with rationale


hdost:
While I'm not sure if for [this specific case](https://github.com/open-telemetry/opentelemetry-rust/pull/1295) we want to, but for the purposes of posterity and avoid needing to do archeological digging every time a decision is made it might be worth having a doc for this.

hdost:
The parentheses in this comment sounds like another deviation. https://github.com/open-telemetry/opentelemetry-rust/pull/1293#issuecomment-1760017241

tp06147:
(123) 456-7890

cijothomas:
From SIG call:
We already has some known deviations, so lets start with documenting that as an md. We can start with just linking to this issue, which can be used to make the md file

PR #1379, opened 2023-11-16
Improve cost of creating AttributeSets

KallDrexx:
## Changes

When attempting to call `Counter<T>::add()` or `UpDownCounter<T>::add()` on a frequent basis (e.g. per network packet), a significant amount of CPU time was being used by `AttributeSet::from()` calls.  

Replacing the `AttributeSet`'s inner type from a `BTreeSet` to `Vec` (+ `sort()` and deduplication)  improves metric throughput (presumably due to more efficient handling of smaller slices).  Likewise, pre-caching the hash for the `AttributeSet` made lookups much faster.

This PR also changes the metrics test to have attributes out of order, to ensure that throughput considers sorting cost.

#### main branch
Number of threads: 8
Throughput: 4,236,400 iterations/sec
Throughput: 4,253,800 iterations/sec
Throughput: 4,261,400 iterations/sec
Throughput: 4,308,400 iterations/sec
Throughput: 4,389,600 iterations/sec
Throughput: 4,440,800 iterations/sec
Throughput: 4,474,400 iterations/sec
Throughput: 4,460,200 iterations/sec

#### This PR 
Number of threads: 8
Throughput: 5,490,600 iterations/sec
Throughput: 5,496,800 iterations/sec
Throughput: 5,891,800 iterations/sec
Throughput: 5,909,600 iterations/sec
Throughput: 5,830,800 iterations/sec

This ends up being **31%** throughput improvement on my machine.


KallDrexx:
@cijothomas When you get a chance, what is the best way to fix the test failure this PR causes? 

Specifically, the [prometheus sanitation test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/tests/integration_test.rs#L137) uses duplicate attributes (`A.B`) and uses the previous de-duplication logic which would only result in `Q`.  However, my PR removes de-duplication logic and thus the final result has the two values concatenated.

Based on previous conversations, the C++ and .net libraries do not perform de-duplication, and as far as I"m aware the [the OpenTelemetry spec does not say they should be duplicated](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/tests/integration_test.rs#L137), which means this behavior might surprise users.

Should I modify the test to expect them to be concatenated?

cijothomas:
> @cijothomas When you get a chance, what is the best way to fix the test failure this PR causes?
> 
> Specifically, the [prometheus sanitation test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/tests/integration_test.rs#L137) uses duplicate attributes (`A.B`) and uses the previous de-duplication logic which would only result in `Q`. However, my PR removes de-duplication logic and thus the final result has the two values concatenated.
> 
> Based on previous conversations, the C++ and .net libraries do not perform de-duplication, and as far as I"m aware the [the OpenTelemetry spec does not say they should be duplicated](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/tests/integration_test.rs#L137), which means this behavior might surprise users.
> 
> Should I modify the test to expect them to be concatenated?

prometheus won't be able to scrap successfully if there is duplication, so maybe this is something to be handled at the promethes exporter level. @jtescher Do you have suggestions to unblock? Could we just modify the prometheus test to prevent the issue and document that SDK/Prometheus exporter does not de-dup, and is user responsibility?

jtescher:
Seems like there are a few places to investigate here, if the instrument `add` calls are with a fixed set of attributes we could look at changing the methods to accept attribute sets rather than slices that have to be converted which may sidestep this issue for some people. If this is more about supporting a large range of different attribute sets for instruments we could look at fixing this in exporters, or adding a `new_unchecked` style API where users have to ensure there are no dupes themselves.

KallDrexx:
> prometheus won't be able to scrap successfully if there is duplication, so maybe this is something to be handled at the promethes exporter level.

Are we sure about this?  The prometheus test seems pretty explicit that it *will* concatenate values if the keys are duplicated due to sanitation.  That is tested and still works after my change.  So it seems like prometheus already supports some level of duplication, this test is only making the explicit decision that duplicate keys due to sanitation are allowed but duplicate exact keys are not allowed.

KallDrexx:
> Seems like there are a few places to investigate here, if the instrument add calls are with a fixed set of attributes we could look at changing the methods to accept attribute sets rather than slices that have to be converted which may sidestep this issue for some people.

That only addresses performance concerns for users that have a stable set of attributes they use a lot.  This PR significantly improves performance for users who do not have a predictable set of attribute key/value combinations.

stress/src/metrics.rs
```diff
@@ -33,14 +33,16 @@ fn test_counter() {
...
 
     // each attribute has 10 possible values, so there are 1000 possible combinations (time-series)
     COUNTER.add(
         1,
         &[
-            KeyValue::new("attribute1", ATTRIBUTE_VALUES[index_first_attribute]),
+            KeyValue::new("attribute4", ATTRIBUTE_VALUES[index_fourth_attribute]),
             KeyValue::new("attribute2", ATTRIBUTE_VALUES[index_second_attribute]),
             KeyValue::new("attribute3", ATTRIBUTE_VALUES[index_third_attribute]),
+            KeyValue::new("attribute1", ATTRIBUTE_VALUES[index_first_attribute]),
```
shaun-cox, line 45: Maybe a comment about why we purposely unsort them here?

shaun-cox (approved):
Excellent improvement!  Thanks for the contribution.  Assuming other approvers/maintainers are okay with the behavior change, I wholeheartedly support from a performance perspective.

cijothomas:
> > prometheus won't be able to scrap successfully if there is duplication, so maybe this is something to be handled at the promethes exporter level.
> 
> Are we sure about this? The prometheus test seems pretty explicit that it _will_ concatenate values if the keys are duplicated due to sanitation. That is tested and still works after my change. So it seems like prometheus already supports some level of duplication, this test is only making the explicit decision that duplicate keys due to sanitation are allowed but duplicate exact keys are not allowed.

I tested OTel .NET with actual Prometheus to see how would Prometheus react if de-dedup is not done. Below are my findings.

Following is an output Prometheus is happily scrapping successfully:
```
# TYPE myCounter_total counter
# HELP myCounter_total A counter for demonstration purpose.
myCounter_total{color="red",name="apple"} 2999.700000000016 1700256532355
myCounter_total{color="yellow",name="lemon"} 30269.70000000016 1700256532355

# EOF
```
This was generated using below
```csharp
Counter.Add(9.9, new("name", "apple"), new("color", "red"));
Counter.Add(99.9, new("name", "lemon"), new("color", "yellow"));
```

Now the following is something prometheus is failing to scrape
```
# TYPE myCounter_total counter
# HELP myCounter_total A counter for demonstration purpose.
myCounter_total{color="red",name="orange",name="apple"} 11375.099999999758 1700256656251
myCounter_total{color="yellow",name="lemon"} 114785.09999999781 1700256656251

# EOF
```

This was generated using
```csharp
Counter.Add(9.9, new("name", "apple"), new("name", "orange"), new("color", "red")); // no dedup done in sdk..
Counter.Add(99.9, new("name", "lemon"), new("color", "yellow"));
```

In case of .NET, neither SDK, not prometheus exporter is deduping, so prometheus scrape is failing.
For OTel Rust, we need to decide if we are okay with similar behavior as OTel .NET OR we should do de-dedup in PrometheusExporter itself. 

My personal vote is to not do dedup in prometheus now, and get back to this based on feedback. If there are users affected by this, we should fix prom exporter.

Main point being - lets remove dedup from SDK and work towards the [goals](https://opentelemetry.io/community/mission/#we-value-_performance_)!

cijothomas:
```
# TYPE myCounter_total counter
# HELP myCounter_total A counter for demonstration purpose.
myCounter_total{color="red",name="orange",name="apple"} 11375.099999999758 1700256656251
myCounter_total{color="yellow",name="lemon"} 114785.09999999781 1700256656251
```
can prometheus scrape this correctly?

_Duplicate Label Names: In the first metric, myCounter_total{color="red",name="orange",name="apple"}, you have used the label name name twice. Prometheus expects each label name within a metric to be unique. Having duplicate label names is not valid in Prometheus and would likely cause an error during scraping._

Should have just asked AI 🤖

hdost:
> In case of .NET, neither SDK, not prometheus exporter is deduping, so prometheus scrape is failing.
> For OTel Rust, we need to decide if we are okay with similar behavior as OTel .NET OR we should do de-dedup in PrometheusExporter itself.

 .NET has a bad Prometheus exporter implementation ? I wouldn't think we should replicate that 🤕. I agree we want to get good performance, but we should also ensure a proper exporter output.

KallDrexx:
It's not clear to me why deduping should be opentelemetry's concern rather than a concern of the calling application.

What happens when deduping occurs?  Should the last value be kept?  The first?  What is last and first in different scenarios since all attributes are passed in at the same time?  Only the consuming application knows how deduping should accurately be done imo

hdost:
While we don't have to follow the limits there are clear guidelines for how to deal with attributes https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-limits
If we want to offload the work of deduplication to the exporter, then we will need to preserve the order in which they were added.

hdost:
Also I do want to say @KallDrexx thank you for looking into this it's very much appreciated. I apologize if I came if a bit strong to start.

hdost:
In a curiosity vein, for your specific use case do you have skeleton you could provide. Would it be possible to have a cached set of attributes or is it entirely dynamic for your network counter?

KallDrexx:
Those guidelines are far from clear in this scenario as far as I can tell.  Not only do they state that those guidelines aren't really applicable to metrics, Those dictate how truncation should happen if there is a finite number of allowed attributes which isn't the case here.

> For metrics the identity is very important to how values are aggregated.

Which is 100% why I think the library doing arbitrary deduplication is bad.

If the consumer calls `counter.add(5, {color: red, shape: circle})` what are they semantically trying to communicate?   Based on my experience in past observability projects (not opentelemetry specific) they ware looking at incrementing the counter for the intersection of all attributes specified.  

So as a concrete example calling 

```
counter.add(5, {color: red, shape: square});
counter.add(5, {color: red, shape: circle});
```

In my metrics UI I should see 10 for `color: red`, 5 for `shape: square`, 5 for `shape: circle`, 5 for `shape: circle, color: red` and 0 for `shape: circle, color: blue`.  The latter one is especially important because I can clearly tell that all the circles that were incremented were red and none were blue.

So what happens if the user calls `counter.add(5, {color: red, color: blue})`.  Semantically this means that they only want 5 added to the metrics where you are search for both red and blue, which is a totally different bucket than `red` and `blue` by themselves.  

So truncating now means that `counter.add(5, {color: red, color: blue})` is equivalent to `counter.add(5, {color: red})` which has a very different semantic meaning imo. 

This is not that crazy to me , because that's exactly how the prometheus exporter test works today.  If you look [at the test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/tests/integration_test.rs#L137) you see that it's assuming that if `C.D` and `C/D` are provided, they should get concatenated into a single `C_D` with the final attribute value being `Y;Z`.  While I don't have back story into that test (I didn't write it and am not a prometheus expert), my assumption is that this will get aggregated by prometheus as a single `Y;Z` value that shows up in filters, which is the only meaningful way to filter on values that are an intersection of two types of the same dimension.  

This leads me to believe that the current prometheus code is correct.  If you call `counter.add(5, {color: red, color: blue})` then the current prometheus exporter will export `color:red;blue`, which while will bloat the UI is the only way to make the metrics correctly identifiable.  

 That's why I believe that the caller of the SDK should do any de-duplication that involves complete removal of attributes, because they are the only one who know the valid way to do so.

So my next question is how do other opentelemetry SDKs handle this?

* [Python's add call](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/instrument.py#L150) requires a dictionary to be put in, explicitly forcing the consumer of the SDK to do any de-duplication themselves.  
* [Java's library](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/common/Attributes.java#L76) does de-dupe attributes but does a last one wins, which is different than rust's code today and the guidelines posted above.
* [C++](https://github.com/open-telemetry/opentelemetry-cpp/blob/cb603ad97f33e52340e627e1cb43ba73bb1d7ef0/api/include/opentelemetry/common/key_value_iterable.h#L17) does not seem to make any attempt to filter duplicate keys in their iterables (the interface doesn't specify it and the code I found doesn't do it
* I'm not totally sure what's going on in [Go](https://github.com/open-telemetry/opentelemetry-go/blob/main/metric/instrument.go#L218) but it doesn't appear that it's doing any filtering from the browsing I did
* Javascripts' library is expecting a javascript object to be passed in by attributes, which means it's expecting the caller to have have no-duplicates already when passed in (or handled duplicate keys themselves)

So every language library seems to have different considerations and no one follows that guideline you posted specifically.  Maybe that's a bad thing and the rust's current implementation is doing things right and everyone else is wrong, but is it ok to be the only one doing things "right" if it causes a performance hit?

KallDrexx:
> In a curiosity vein, for your specific use case do you have skeleton you could provide. Would it be possible to have a cached set of attributes or is it entirely dynamic for your network counter?

Our application is working with metrics for network I/O on a per packet basis, which is why we encountered the `AttributeSet::from()` costs.  I do plan on looking at adding a bounded instrument API to not have to bear the cost of creating attribute sets as often, but we will have to create new attribute sets regularly at some points, so I wanted to ensure that that code path was still fast.

cijothomas:
> > In case of .NET, neither SDK, not prometheus exporter is deduping, so prometheus scrape is failing.
> > For OTel Rust, we need to decide if we are okay with similar behavior as OTel .NET OR we should do de-dedup in PrometheusExporter itself.
> 
> .NET has a bad Prometheus exporter implementation ? I wouldn't think we should replicate that 🤕. I agree we want to get good performance, but we should also ensure a proper exporter output. This could be offloaded to the scrape time.

You are right! It could be treated as a bug in OTel .NET's Prometheus Exporter. OTel .NET's Prometheus is an **experimental** component, and there is a chance it'll never be stable released at all (in favor of [native OTLP support](https://prometheus.io/docs/prometheus/latest/feature_flags/#otlp-receiver) in Prometheus)

And I tried how would Prometheus's native OTLP ingestion deals with duplication! Surprised with the result! I send it `MyFruitCounter.Add(1, new("name", "apple"), new("name", "orange"), new("color", "red"));`, via OTLP in OTel .NET where we know neither the SDK nor OTLP Exporter is deduping, and Prometheus stored it as **_MyFruitCounter_total{color="red", job="unknown_service:getting-started-prometheus-grafana", name="orange;apple"}_**

@hdost Any ideas how should OTel Rust prometheus handles duplicate attributes? Should we send "name=orange", with a policy of "last-one-wins", OR should we do "name=orange;apple", just like Prometheus itself is doing?

hdost:
> > For metrics the identity is very important to how values are aggregated.
> 
> Which is 100% why I think the library doing arbitrary deduplication is bad.

Well I would agree that arbitrary dedup is bad and I would say what I said before in that we could have the exporter (the client from my perspective) have settings which could control how dedup is handled. It also wouldn't put the performance hit on the writer but on the subscriber. 

That's what I was trying to say here:
> If we want to offload the work of deduplication to the exporter, then we will need to preserve the order in which they were added.

Whatever we do for resolving Prometheus specifically requires uniqueness of labels, so whether to do it is not the question, but how. And there are Processors and Views which can be utilized to make a determination as to how this data is processed. It's the SDK which enables it. And to provide flexibility to users we could certainly expand that. 


>  That's why I believe that the caller of the SDK should do any de-duplication that involves complete removal of attributes, because they are the only one who know the valid way to do so.


> So every language library seems to have different considerations and no one follows that guideline you posted specifically.  Maybe that's a bad thing and the rust's current implementation is doing things right and everyone else is wrong, but is it ok to be the only one doing things "right" if it causes a performance hit?

I'm not going to die on the hill of right vs performance, but I think this comes from me not explaining my position well enough. As that's not what I was arguing. 

I would also say whatever "right" is should be consistent between implementations, otherwise what's the point of having a standard, but that's going to be decided here. If we decide to go against spec then that's why we have https://github.com/open-telemetry/opentelemetry-rust/issues/1297

KallDrexx:
I will also mention that if we *really* want duplication logic as part of `AttributeSet::from()`  then I can look at adding that in.  My suspicion is the majority of the performance gain is from switching from `BTreeSet` to `Vec`.  I'd rather not add the cost in if it's not needed.  

I'm not necessarily opposed to doing deduplication of keys on the exporting side (which I think may be what you are aiming for).  However to me it seems like:

* OpenTelemetry spec does not specify this behavior as far as I've been able to find
* Every implementation that I've been able to find seems to differ. 
  * Some APIs filter duplicate attribute keys on creation of the attribute set (rust right now is first wins, java is last wins)
  * Some APIs have API boundaries that make it impossible to submit duplicate keys in the first place (python, javascript)
  * Some APIs seem to do no filtering at attribute set time (c++, go I think?)

I haven't looked at the exporter code in these projects yet but it seems like there's no consistency between implementations currently.  So I'm not sure it's actually a deviation from the spec.

KallDrexx:
> If we want to offload the work of deduplication to the exporter, then we will need to preserve the order in which they were added.

I guess "preserving the order in which they are added" is an interesting concept.  For example

```
counter.add(1, {color: red, color: blue});
counter.add(1, {color: blue, color: red});
```

THis is actually technically an edge case in my PR as well.  But should these be in the same time series or different time series?  If the same, should order be preserved at export time?

jtescher:
Can see go's set handling [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/attribute/set.go) and which semantics it uses to dedup.

hdost:
> I will also mention that if we *really* want duplication logic as part of `AttributeSet::from()`  then I can look at adding that in.  My suspicion is the majority of the performance gain is from switching from `BTreeSet` to `Vec`.  I'd rather not add the cost in if it's not needed.  


I agree, so the only thing I would say for the difference on your implementation would be let's remove the sorting. 

> I'm not necessarily opposed to doing deduplication of keys on the exporting side (which I think may be what you are aiming for).  

Yes that's my thought here, I'd like Prometheus to export correctly (as possible). Though don't worry in this PR. 


> * Every implementation that I've been able to find seems to differ. 
> 
> I haven't looked at the exporter code in these projects yet but it seems like there's no consistency between implementations currently.  So I'm not sure it's actually a deviation from the spec.

Yea even where there's ambiguity I'd rather document what are we doing than have people not know until they read the source.

KallDrexx:
> I agree, so the only thing I would say for the difference on your implementation would be let's remove the sorting.

I can do that, but that does mean that 

```
counter.add(5, {color: red, shape: apple});
counter.add(5, {shape: apple, color: red});
```

will count as 2 different attribute sets and thus two different time series.  The only con I can see with that is it means that cardinality rules (the 2000 limit that's currently in place) is more likely to be hit, which means it's more likely that metrics will overflow (especially since there's no cleanup logic currently for time series/attribute sets AIUI).

cijothomas:
> > I agree, so the only thing I would say for the difference on your implementation would be let's remove the sorting.
> 
> I can do that, but that does mean that
> 
> ```
> counter.add(5, {color: red, shape: apple});
> counter.add(5, {shape: apple, color: red});
> ```
> 
> will count as 2 different attribute sets and thus two different time series. The only con I can see with that is it means that cardinality rules (the 2000 limit that's currently in place) is more likely to be hit, which means it's more likely that metrics will overflow (especially since there's no cleanup logic currently for time series/attribute sets AIUI).

we cannot do that. it must be aggregated to same time series!

TommyCpp:
In prometheus client(both in Rust and Go) generally you cannot produce duplicate labels. In Golang you need to pass a dictionary. In Rust you can pass a slice but it will panic if the label has duplication. So I think prometheus users does expected the deduplication or panic happen inside client.

Also per spec the attributes collections need to be deduplicated in the application. We don't necessary need to perform it in SDK but we need to make sure it's deduplicated before sending it out

> [Resources](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md), Metrics [data points](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#metric-points), [Spans](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes), Span [Events](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#add-events), Span [Links](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#link) and [Log Records](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) may contain a collection of attributes. The keys in each such collection are unique, i.e. there MUST NOT exist more than one key-value pair with the same key. The enforcement of uniqueness may be performed in a variety of ways as it best fits the limitations of the particular implementation.

Before this PR merging we should also check 

1. Other signals are not rely on this deduplication
3. Since we no longer duplicate the content, let's rename it to `AttributeCollection` or something else. `Set` implies deduplication in my mind

KallDrexx:
Thanks for that @TommyCpp , I had previously looked and never saw any mention of that, but now that you pointed to the exact paragraph wording I can see it [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md).  

So yes, this means my PR violates the opentelemetry specification.  The next paragraph also says 

> Normally for the telemetry generated using OpenTelemetry SDKs the attribute key-value pairs are set via an API that either accepts a single key-value pair or a collection of key-value pairs. Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value. See for example Span's [SetAttribute](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-attributes) API.

Which also means that [the current rust implementation isn't correct](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/attributes/set.rs#L118) since it's doing a first seen wins rather than last seen wins.  

So the first thing I will do is adjust the code to dedup with last seen wins keys to follow the opentelemetry step on `AttributeSet` creation, and we'll see the performance implications  of that.

That being said, this also tells me that we should utilize the type system better, and instead of `counter.add()` taking a slice of `KeyValue` pairs, it should instead take a `HashMap<Key, Value>` to enforce this in the type system.  This not only means that callers of the API eats the cost of deduplication, but this also means that since they are passing us a `HashMap` already formed callers of the API are better able to pre-form attribute sets and re-use them at no additional cost.

hdost:
Maybe we should discuss this at the next SIG

cijothomas:
Lot of interesting discussions. Thanks everyone for reviewing. 
1. I think this SIG already agrees on the importance of performance. We already removed de-dup from [Logs](https://github.com/open-telemetry/opentelemetry-rust/pull/1142), [SpanAttributes ](https://github.com/open-telemetry/opentelemetry-rust/pull/1293) achieving significant perf gains.
2. For metrics, we should do the same to achieve perf goals. This PR is doing that.
3. For metrics, we know of a well known exporter (prometheus) which could fail to work correctly without deduplication. This means we need to do de-duplication at the exporter(prometheus) side. A similar concern exists for other signals too - there may be users wanting to de-duplicate before exporting.
4. Either in this PR or a followup PR - lets to de-dup at Prometheus side. The Prometheus exporter is anyway pre-release (Otel -> Prometheus spec is also experimental), but I agree we need to make sure the exporter produces valid output for Prometheus to scrap. 
5. The lack of tests in general is concerning, but this is separately addressable. (One attempt in that direction is https://github.com/open-telemetry/opentelemetry-rust/pull/1382)

Given the long discussions, I think it is also good to discuss this in coming SIG meeting before merging.


@KallDrexx 
_" my PR violates the opentelemetry specification."._

Regd. this - the spec allows flexibility here.  The last sentence of [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-collections) allows **_"enforcement of uniqueness will likely be the responsibility of the recipient of this data."_**. We are leveraging that flexibility for Logs, Spans today, and this PR is adding it for Metrics.

KallDrexx:
So I plan to attend tomorrow's SIG, but for now I wanted to put together some data points from the stress test under various attribute set mechanisms

main: 954,240 (100%  baseline)
sorted vec (this PR): 1,583,000 (166%)
sorted + deduped vector: 1,151,440 (121%)
Hashmap: 1,877,720 (197%)


That last one is really interesting.  That last run involves [this change set](https://github.com/open-telemetry/opentelemetry-rust/compare/main...KallDrexx:opentelemetry-rust:attribute_set_hashmap?expand=1) which makes the attribute set hashmap backed instead of `Vec` or `BTreeSet` backed.  I need to double check everything to make sure this doesn't subtly break anything, but the performance gain over even `Vec` is really interesting, and does make it seem like we can maybe have our performance cake while still adhering to the OpenTelemetry spec.  I suspect part of the performance gains is maybe the caching of the hash that I'm doing, so it doesn't need to be computed multiple times.

lalitb:
> I suspect part of the performance gains is maybe the caching of the hash that I'm doing, so it doesn't need to be computed multiple times.

This is possible, we do the same in otel-cpp (and I think otel-js does the same), to pre-compute and store the hash along with attributes.

KallDrexx:
> This is possible, we do the same in otel-cpp (and I think otel-js does the same), to pre-compute and store the hash along with attributes.

Yep.  This definitely seems to be the key.  I updated my current PR to not only dedup the `Vec<HashKeyValue>` but also precompute the hash.

This gave the throughput numbers of 

```
 4,943,600.00 
 4,951,000.00 
 4,945,800.00 
 4,560,000.00 
 4,926,800.00 
```

which ended up being a 423% improvements over the stress test against main for me.

hdost:
> > This is possible, we do the same in otel-cpp (and I think otel-js does the same), to pre-compute and store the hash along with attributes.
> 
> 
> 
> Yep.  This definitely seems to be the key.  I updated my current PR to not only dedup the `Vec<HashKeyValue>` but also precompute the hash.
> 
> 
> 
> This gave the throughput numbers of 
> 
> 
> 
> ```
> 
>  4,943,600.00 
> 
>  4,951,000.00 
> 
>  4,945,800.00 
> 
>  4,560,000.00 
> 
>  4,926,800.00 
> 
> ```
> 
> 
> 
> which ended up being a 423% improvements over the stress test against main for me.
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 
> 

This sounds very promising, do you want to make this one as a specific separate PR for comparison ?

KallDrexx:
> This sounds very promising, do you want to make this one as a specific separate PR for comparison ?

I went ahead and updated this PR, because while precomputing the hash is new, I still did it in the vein of storing the attributes in a `Vec<HashedKeyValue>` (though I probably don't need `HashKeyValue` instead of just `KeyValue` since it's precomputed).

The `Vec<T>` is still going to be faster than a `BTreeSet` as the `BTreeSet` does more work to sort on every insert instead of once at the end (among other `Vec` optimizations for small lists).

So the new changes that include deduplication last-wins logic now passes all tests as can be seen on CI.  I still want to do some manual verification since we don't have good test coverage over attributes.  

The current stress test values are:

```
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust/stress$ cargo run --release --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 6.13s
     Running `/home/mshapiro/code/opentelemetry-rust/target/release/metrics`
Number of threads: 6
Throughput: 6,290,200 iterations/sec
Throughput: 6,562,400 iterations/sec
Throughput: 6,672,400 iterations/sec
Throughput: 6,636,600 iterations/sec
Throughput: 6,373,800 iterations/sec
Throughput: 6,460,800 iterations/sec
^C
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust/stress$ cargo run --release --bin metrics
    Finished release [optimized] target(s) in 0.08s
     Running `/home/mshapiro/code/opentelemetry-rust/target/release/metrics`
Number of threads: 6
Throughput: 6,304,000 iterations/sec
Throughput: 5,937,600 iterations/sec
Throughput: 5,824,600 iterations/sec
Throughput: 6,308,600 iterations/sec
Throughput: 6,110,200 iterations/sec
^C
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust/stress$ cargo run --release --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 9.07s
     Running `/home/mshapiro/code/opentelemetry-rust/target/release/metrics`
Number of threads: 6
Throughput: 1,248,800 iterations/sec
Throughput: 1,296,000 iterations/sec
Throughput: 1,266,400 iterations/sec
Throughput: 1,291,800 iterations/sec
Throughput: 1,306,200 iterations/sec
Throughput: 1,267,200 iterations/sec
Throughput: 1,246,800 iterations/sec
Throughput: 1,228,200 iterations/sec
^C
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust/stress$ cargo run --release --bin metrics
    Finished release [optimized] target(s) in 0.07s
     Running `/home/mshapiro/code/opentelemetry-rust/target/release/metrics`
Number of threads: 6
Throughput: 1,092,000 iterations/sec
Throughput: 1,102,000 iterations/sec
Throughput: 1,075,800 iterations/sec
Throughput: 1,102,000 iterations/sec
^C
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust/stress$ Throughput: 971,400 iterations/sec
cargo run --release --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 6.03s
     Running `/home/mshapiro/code/opentelemetry-rust/target/release/metrics`
Number of threads: 6
Throughput: 6,162,400 iterations/sec
Throughput: 5,684,200 iterations/sec
Throughput: 5,232,000 iterations/sec
Throughput: 5,629,600 iterations/sec
Throughput: 5,640,200 iterations/sec
^C
```

The first two runs and the final run are from the commit in this PR as of now.  The middle two runs are from the main branch (with the stress test modified to match the out of order sorting my PR changes).

Averaging these iteration/sec values gives main branch `1,210,266` iterations per second while my PR provided `6,168,066` iterations per second, a 510% improvement.  

To be sure where the performance difference was coming from, precomputing the hashes with a `BTreeSet` gave a speedup but only got up to about `2,959,560` iterations per second, which is far below what I'm getting for the Vec implementation.

I also tested this change against our application which encountered metrics being a bottleneck for us.  One function is ending up particularly heavy for us which is incrementing two counters for every network packet we get.  This brought the total CPU usage for that function from 11.93% down to 8.24%.  While I need to look a bit more at it, a significant amount of the performance gain ended up coming from improved call times to `Measure::call()`.

hdost:
Thank you again very much for working on this!

KallDrexx:
Ok I've double checked everything I wanted to, and things look good.  I had to modify the stress test a tiny bit (lower the number of attribute values)  since 4 attributes was causing us to go over the 2k cardinality limit.  

I've updated the PR description with the final benchmarks from my machine.  Would be nice if someone is able to verify similar numbers.

Interestingly enough, I did an experiment reducing the deduplication logic and ended up with consistently lower throughput (only by ~100k-200k iterations per second).  That doesn't quite make sense to me, especially since it's not actually detecting duplicates in the stress test.  

So from my perspective this PR is ready for review.  

Again, sorry for all the back and forth that went on while I did exploration :)

cijothomas:
> Would be nice if someone is able to verify similar numbers.

My laptop is giving a diff result:
Throughput: 4,471,200 iterations/sec  -- main branch with the stress test from this PR
Throughput: 5,492,000 iterations/sec -- this PR branch

lalitb:
| Would be nice if someone is able to verify similar numbers.

This is what I see (8 cores, wsl2 32GB)

Current PR:  ~6,286,000 iterations/sec 
Main branch (with stress test from current PR): ~ 2,592,000 iterations/sec

cijothomas:
> > Would be nice if someone is able to verify similar numbers.
> 
> My laptop is giving a diff result: Throughput: 4,471,200 iterations/sec -- main branch with the stress test from this PR Throughput: 5,492,000 iterations/sec -- this PR branch

another machine:

Throughput: 2,257,200 iterations/sec
Throughput: 2,236,200 iterations/sec
Throughput: 2,476,200 iterations/sec

vs


Throughput: 5,180,000 iterations/sec
Throughput: 5,242,400 iterations/sec
Throughput: 5,289,600 iterations/sec

opentelemetry-sdk/src/attributes/set.rs
```diff
@@ -104,37 +105,46 @@ impl Eq for HashKeyValue {}
...
+            .rev()
+            .filter_map(|k| {
+                if seen_keys.insert(k.key.clone()) {
+                    Some(HashKeyValue(k.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+        vec.sort();
```
jtescher, line 125: nit: Think we can have a small optimization here as we don't need stable sorting for items without duplicate keys?

```suggestion
        vec.sort_unstable();
```

opentelemetry-sdk/src/attributes/set.rs
```diff
@@ -104,37 +105,46 @@ impl Eq for HashKeyValue {}
...
-        AttributeSet(
-            values
-                .iter()
-                .map(|(key, value)| HashKeyValue(KeyValue::new(key.clone(), value.clone())))
-                .collect(),
-        )
+        let vec = values
+            .iter()
+            .map(|(key, value)| HashKeyValue(KeyValue::new(key.clone(), value.clone())))
+            .collect::<Vec<_>>();
```
jtescher, line 141: Need to sort this vec as well now to ensure order is maintained

jtescher (changes requested):
one last issue and a nit, other than that looks good!

opentelemetry-sdk/src/attributes/set.rs
KallDrexx: Good call. changed!

opentelemetry-sdk/src/attributes/set.rs
KallDrexx: Good catch, added!

opentelemetry-sdk/src/attributes/set.rs
```diff
@@ -104,37 +105,47 @@ impl Eq for HashKeyValue {}
...
-                        Some(HashKeyValue(kv.clone()))
-                    }
-                })
-                .collect(),
-        )
+        let mut seen_keys = HashSet::with_capacity(values.len());
+        let mut vec = values
+            .iter()
+            .rev()
+            .filter_map(|k| {
```
cijothomas, line 117:
```suggestion
            .filter_map(|kv| {
```

nit for readability

jtescher (approved):

cijothomas:
> Interestingly enough, I did an experiment reducing the deduplication logic and ended up with consistently lower throughput (only by ~100k-200k iterations per second). That doesn't quite make sense to me, especially since it's not actually detecting duplicates in the stress test.

I see otherwise! Removing the de-dup caused another 10% boost for me... (I cannot think of any reason why it showed lower throughput for you! Could you try re-running?

opentelemetry-sdk/src/attributes/set.rs
KallDrexx: done!

opentelemetry-sdk/src/attributes/set.rs
```diff
@@ -104,37 +105,47 @@ impl Eq for HashKeyValue {}
 ///
 /// This must implement [Hash], [PartialEq], and [Eq] so it may be used as
 /// HashMap keys and other de-duplication methods.
-#[derive(Clone, Default, Debug, Hash, PartialEq, Eq)]
-pub struct AttributeSet(BTreeSet<HashKeyValue>);
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub struct AttributeSet(Vec<HashKeyValue>, u64);
```
cijothomas, line 109: Trying to see how caching the hash is improve speed.. I believe this is because it can now be re-used when we lookup timeseries in [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/internal/sum.rs#L34)?

It'd help if we were needing the hash calculation more than once. But we don't need it for more than once.. So in this PR, the hashcalculation is simply moved from [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/internal/sum.rs#L34) to the AttributeSet creation, so I am not able to see how that would have helped with the perf improvement! 

Or the perf gains are purely from the Btee->Vec switch?

cijothomas: In other words, we should not make this change, as taking up more memory to store/cache the hash only benefits *if* there lookup/insert are needed more than once, and from my understanding, we don't have the lookup-more-than-once requirement..

KallDrexx: Putting a breakpoint on `Hash` shows there are several hidden hash calls within the hashmap.  For example:

image

So when the hash map grows, it has to re-hash all the time series in it.  So this maybe why we see a 500% increase in stress test performance but a bit muted (but still impactful) performance improvement in real life applications.  Precaching hashes improves this test case because we have 1k+ time series that sometimes need to be rehashed.  

Although eventually that should stabilize, and after a few thousand iterations I no longer see reserve calls requesting hashes, but I don't see a significant change in throughput.  

I did the following two runs:

```
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust$ cargo run --release --package stress --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 5.88s
     Running `target/release/metrics`
Number of threads: 6
Throughput: 5,525,800 iterations/sec
Throughput: 5,796,800 iterations/sec
Throughput: 5,553,600 iterations/sec
Throughput: 5,823,200 iterations/sec
Throughput: 5,942,600 iterations/sec
Throughput: 5,609,600 iterations/sec
^C
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust$ cargo run --release --package stress --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 5.86s
     Running `target/release/metrics`
Number of threads: 6
Throughput: 1,304,400 iterations/sec
Throughput: 1,314,800 iterations/sec
Throughput: 1,314,600 iterations/sec
Throughput: 1,158,400 iterations/sec
Throughput: 1,300,200 iterations/sec
Throughput: 1,332,600 iterations/sec
Throughput: 1,333,200 iterations/sec
```

The first run is with this PR as is, the 2nd run is commenting out the `impl Hash` block and adding `Hash` to the `derive()` attribute.

So pre-hashing is definitely providing an improvement on its own as far as  Ican tell.  This is backed up by profiling our actual application, as we see `Measure::call()` percentages go from 4.66% of our CPU time down to 2.30%.  I'm also pretty sure we saw a reduction in `hash_one` percentages but I didn't write them down in my notes.  

So I'm not totally sure where the extra hashing is coming from, but it does seem to add some benefits in real use cases.

lalitb: Good point. Looking into the code now I don't even see caching is going to help here. Even for the slight improvement, the implication of storing the cache could be considerable. We have a slightly different design in otel-cpp which helps as the lookup is not purely O(1) in the hot path.

cijothomas: > So when the hash map grows, it has to re-hash all the time series in it. So this maybe why we see a 500% increase in stress test performance but a bit muted (but still impactful) performance improvement in real life applications. Precaching hashes improves this test case because we have 1k+ time series that sometimes need to be rehashed.

We need to prevent this nonetheless. We can ensure HashMap has enough size by initializing it correctly, so rehashed would be avoided. We know the max size (default 2000), so this should be relatively easy to achieve! Let us address this separately, as that is another nice improvement by avoiding reallocations!

My stress test show no change when I removed Hash custom implementation!

KallDrexx: Not sure what's going on then I guess.

For comparison I tested the Main branch stress test, then added precaching of the hashes (same mechanism) to the `BtreeSet` implementation and got 

``` 
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust$ cargo run --release --package stress --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 6.36s
     Running `target/release/metrics`
Number of threads: 6
Throughput: 1,109,400 iterations/sec
Throughput: 1,150,800 iterations/sec
Throughput: 1,153,000 iterations/sec
Throughput: 1,137,800 iterations/sec
Throughput: 1,135,000 iterations/sec
^C
mshapiro@matt-ms-hpz:~/code/opentelemetry-rust$ Throughput: 1,098,600 iterations/sec
cargo run --release --package stress --bin metrics
   Compiling opentelemetry_sdk v0.21.1 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-sdk)
   Compiling opentelemetry-stdout v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-stdout)
   Compiling opentelemetry-appender-tracing v0.2.0 (/home/mshapiro/code/opentelemetry-rust/opentelemetry-appender-tracing)
   Compiling stress v0.1.0 (/home/mshapiro/code/opentelemetry-rust/stress)
    Finished release [optimized] target(s) in 5.67s
     Running `target/release/metrics`
Number of threads: 6
Throughput: 3,475,400 iterations/sec
Throughput: 3,722,400 iterations/sec
Throughput: 4,138,400 iterations/sec
Throughput: 4,164,400 iterations/sec
Throughput: 4,053,600 iterations/sec
```

I've tried the same tests in windows (instead of just WSL) and got similar results (though a bit less of a gap).

KallDrexx: As another data point, I changed `ValueMap::new()` to do `HashMap::with_capacity()` to reduce resizing and rehashing.  That made the first iteration report to be slightly better than vanilla `main`, but not by much.  Every throughput report afterwards was within the same variance I get with main.

cijothomas: rehashing impact cannot be tested using the current stress test, as there is no collect() triggered so the hashmap is never emptied, so after the initial fill-up, it wont need re-hash! 

Here's what I suggest:
1. We know from the tests that btree -> vec is definitely perf booster, so lets scope *this* PR to do just that.
2. For rehashing prevention, cache-the-hash, and avoid de-duplication logic, avoid sorting and more - lets do it separately on after other, so each PR is very focused and demonstrates clean perf gains.
(This PR already kick started a ton of ideas to improve perf!)

Thoughts?

lalitb (approved):
LGTM. Suggestion to remove the cache for hash, and take it as separate enhancement. There needs to be tradeoff between perf improvement, and storage allocation, and probably needs more investigation.

opentelemetry-sdk/src/attributes/set.rs
KallDrexx: I'm not opposed to splitting it out.  I can do that tomorrow.

stress/src/metrics.rs
```diff
@@ -33,14 +31,17 @@ fn test_counter() {
     let index_first_attribute = rng.gen_range(0..len);
     let index_second_attribute = rng.gen_range(0..len);
     let index_third_attribute = rng.gen_range(0..len);
+    let index_fourth_attribute = rng.gen_range(0..len);
```
cijothomas, line 34: lets not make change to the stress test itself in this PR. We can drive changes to stress test as a separate PR if we believe it should do unsorted order. (I do not believe it should, but want to avoid adding noise to this PR).

cijothomas (changes requested):
Agree with the replacement of Btree with Vec.! This is moving us in the right direction. Many thanks for doing that, and all the other productive discussions in this PR. I hope the next set of PRs will take us even closer to the perf goals!

Requesting changes only to make sure the following is done before merge:
1. Move the hash caching to a separate PR, after we demonstrate it is needed.
2. Revert changes to stress test.

I opened https://github.com/open-telemetry/opentelemetry-rust/pull/1389 to get more focused benchmark. Could you run that and update the PR description with the results? I also suggest re-run the stress tests as we don't see the 5x improvement in other machines!

opentelemetry-sdk/src/attributes/set.rs
KallDrexx: Ok it's 12:30am, I went to go to sleep and it finally dawns on me why precalculating the hashing improves my throughput on multiple machines more than others.  So I had to get out my laptop to verify, and yes, I know what's going on now.

You are correct that we are only computing the hash once, but *where* we compute the hash is important!  Precomputing the hash the way I am in this PR (that I'm going to put in its own PR) incurs the cost at `AttributeSet::From()` time.  Not precomputing the hash incurs the cost at `Measure::call()` time.  

This is important because incurring the cost during the hash lookup means we are doing more work inside the mutex lock, and thus delaying all other `Measure::call()` operations.  Precomputing the hash allows each counter to compute the hash in parallel, and thus less work inside the hash.

This is why we were seeing such variance of if it was helping or not.  It all came down to thread counts for the throughput stress test.  I verified this by testing with and without precomputed hashes with 6 threads on my PC, then with and without with 2 threads (fyi, 1 thread is broken).  

On my personal laptop (and on battery):
* 6 threads precomputed: 3.5M
* 6 threads non-precomputed: 2.3M
* 2 threads prcomputed: 2.5M
* 2 threads non-precomputed: 2.5M

So precomputing the hash is a concurrency optimization.

cijothomas: Awesome! It makes sense. Lets tackle it in the follow up PR.

If I learned anything over the last few years working in OTel - it is that there is no magic! Everything should be explainable. Thanks a lot!

KallDrexx:
Updated the PR to remove the changes to the stress test, and remove caching of the hash.  I will open up new PRs for these separately.

Will update the PR description with updated benchmarks and throughput values

cijothomas (approved):
LGTM! Thank you!

PR #1380, opened 2023-11-16
Normalized metadata in opentelemetry-appender-tracing.

Rockin2:
## Changes

* Normalized metadata for events from the tracing crate and the log crate.
* Remove redundant metadata.
* Follows https://opentelemetry.io/docs/specs/semconv/general/logs/ for metadata names.


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
+
+fn get_filename(filepath: &str) -> &str {
+    if let Some((_, filename)) = filepath.rsplit_once('/') {
+        return filename;
+    }
+    if let Some((_, filename)) = filepath.rsplit_once('\\') {
+        return filename;
+    }
+    filepath
+}
```
lalitb, line 50: Can we use `std::path::Path` ?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
+
+impl EventVisitor {
+    fn visit_metadata(&mut self, meta: &Metadata) {
+        self.log_record_attributes
+            .push(("log.name".into(), meta.name().into()));
```
lalitb, line 55: Should we remove the redundant attribute here - https://github.com/open-telemetry/opentelemetry-rust/blob/33b5e9b4e2b1ce7b1ed209ad61e2fbcad73e627e/opentelemetry-appender-tracing/src/layer.rs#L109C12-L109C12 ? This will break user_events log exporter, but I can raise a subsequent PR to handle it.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
-    log_record: &'a mut LogRecord,
+#[derive(Default)]
+struct EventVisitor {
+    log_record_attributes: Vec<(Key, AnyValue)>,
+    log_record_body: Option<AnyValue>,
 }
 
-impl<'a> tracing::field::Visit for EventVisitor<'a> {
+fn is_metadata(field: &str) -> bool {
+    if let Some(log_field) = field.strip_prefix("log.") {
```
lalitb, line 32: Why to strip `log.` prefix - the metadata from log/tracing crate won't be having it anyway. Or am I missing something :)

opentelemetry-appender-tracing/src/layer.rs
Rockin2: I though about it but I was not sure because std::path::Path behaviour is dependant on the host OS.

So two cases where I think it can cause problem:

1. Cross-compilation : If we compile on Linux the Windows binary, do std::path::Path will try splitting the path with '\' instead of '/'? Because logically the original path of source files will be with '/'.
2. Bare Metal (No OS) : Do std::path will be available?

But I'm not against using std::path::Path if people judge it is not necessary to handle these cases. :)

opentelemetry-appender-tracing/src/layer.rs
Rockin2: Do you talk about this line `let vec = vec![("name", meta.name())];`?

I think yes because when I read this doc https://opentelemetry.io/docs/specs/semconv/general/logs/, it gives me the impression that the best name for this attribute is "log.name". This doc reminds me the Elastic Common Schema. ECS gives you guidelines for custom field names to make you able to give the impression that your custom fields are like a part of the ECS even if it is not.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: A little quirk from me :p. All the metadata we have for the log lines begin with 'log.' so I though that checking if the name begins with 'log.' is quicker than always looping the LOG_METADATA_ATTRIBUTES array. But honestly I can remove it if it is confusing. :)

Rockin2:
@lalitb Thanks for your quick review :).

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
+        for log_metadata_field in LOG_METADATA_ATTRIBUTES {
```
shaun-cox, line 33: Rather than a `for` loop to search on string compares of every entry, I think a `match` on the set of literals in your `LOG_METADATA_ATTRIBUTES` might generate better code, as I could see the compiler sorting the branch check order to take advantage of bailing out early and avoiding having to check all entries.  e.g.

``` rust
fn is_metadata(field: &'static str) -> bool {
    field
        .strip_prefix("log.")
        .map(|remainder| {
            matches!(
                remainder,
                    "file.line"
                    | "file.name"
                    | "file.path"
                    | "file"
                    | "line"
                    | "module_path"
                    | "module.path"
                    | "name"
                    | "target"
            )
        })
        .unwrap_or(false)
}
```
(I've manually sorted the list of matches, but I'm not sure if its required for better code or not.)
This is also a little more concise than the current version, fwiw.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
         }
     }
 
     fn record_bool(&mut self, field: &tracing_core::Field, value: bool) {
-        if let Some(ref mut vec) = self.log_record.attributes {
-            vec.push((field.name().into(), value.into()));
-        } else {
-            let vec = vec![(field.name().into(), value.into())];
-            self.log_record.attributes = Some(vec);
+        if !is_metadata(field.name()) {
```
shaun-cox, line 103: isn't this a visitor for attributes of type `bool`?  Most (all?) of the metadata attributes are not of `type` bool, so isn't searching the list of metadata attributes to see if any of them match the name of this attribute a bit wasteful?  Or, I guess my question is, why do we need to check if this bool attribute is in the set of any metadata attributes that are not bool?  (And the same question for all the other visit methods other than `record_str`...)

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
+    }
+    filepath
+}
+
+impl EventVisitor {
+    fn visit_metadata(&mut self, meta: &Metadata) {
+        self.log_record_attributes
+            .push(("log.name".into(), meta.name().into()));
+        self.log_record_attributes
+            .push(("log.target".into(), meta.target().to_owned().into()));
```
cijothomas, line 57: Given this is based on [experimental ](https://opentelemetry.io/docs/specs/semconv/general/logs/)conventions, this may block us from releasing stable version of the appender. In other words, we'll be dependent on this semantic convention becoming stable, to release this crate as stable. I am not sure if that is a risk we should take, given the historical delays in stabilizing sem. conventions!

Open to ideas how to navigate this space....

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,122 @@
...
+
+        if let Some(module_path) = meta.module_path() {
+            self.log_record_attributes
+                .push(("log.module.path".into(), module_path.to_owned().into()));
+        }
+        if let Some(filepath) = meta.file() {
+            self.log_record_attributes
+                .push(("log.file.path".into(), filepath.to_owned().into()));
```
cijothomas, line 65: what does the "log.file.path" indicate here? Is that the line of code where the log was emitted? The [sem. convention doc ](https://opentelemetry.io/docs/specs/semconv/general/logs/#log-file) talks about the file to which log was emitted.

opentelemetry-appender-tracing/src/layer.rs
lalitb: Good point. We can also have these under the experimental feature flag, till the semantic conventions are stable.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: Thanks! I will test it and push it.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: I think you're right but if I want to be really really pedantic I could say someone could add an attribute of type bool in a log line with one of the metadata. But it seems really really unlikely. :p 

The logline is an Int so we could check only for this metadata in record_i64.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: I'm good with the feature flag. I can add it.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: Yes it is the source file to which the log was emitted. It is really useful to find a log line in the source code.

Rockin2: The line in the source code is in the log.file.line attribute.

lalitb: @Rockin2 From the [documentation ](https://opentelemetry.io/docs/specs/semconv/general/logs/#log-file)it seems attribute `log.file.path` doesn't refer to source code file path, but the destination file path. So if the source file (say) `auth.rs` has code to emit log to `/var/log/auth.log`:
```
log.file.path: /var/log/auth.log
log.file.name: auth.log
```

Rockin2: Ouch. You're right. Sorry :(. I will fix it.

Rockin2: I think I will rename those log.source.file.path and log.source.file.name.

Rockin2: I think the reason I was confused is the log crate to tracing crate bridge add an attribute called "file" that is the path to the source file that emitted the log line.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: @cijothomas Do you agree with the feature flag solution?

cijothomas: Yes, feature-flag is a good way to manage the risk.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,60 +1,115 @@
...
-struct EventVisitor<'a> {
-    log_record: &'a mut LogRecord,
+#[derive(Default)]
+struct EventVisitor {
+    log_record_attributes: Vec<(Key, AnyValue)>,
+    log_record_body: Option<AnyValue>,
 }
 
-impl<'a> tracing::field::Visit for EventVisitor<'a> {
+fn is_metadata(field: &'static str) -> bool {
```
cijothomas, line 19: this is called for each and every field in every log right? How does this affect the performance? We don't have much benchmarks, but there is stress test which may be leveraged to measure the impact.

Also - do you want to help with writing some logging specific benchmarks?

lalitb: +1 for the benchmark for logs. We need them eventually. I can add them sometime next week unless someone volunteers. For now, good to have metadata validation under the experimental flag, as most of these attributes are not part of the semantic convention either.

Rockin2:
I pushed new commits to cater all your comments.

Rockin2:
@cijothomas @lalitb @shaun-cox Hi everyone! I'm sorry to bother you but I want to merge this PR this week if possible. May you give feedback on my last commits please? Thank you! :)

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -99,23 +167,21 @@ where
...
-        log_record.attributes = Some(vec.into_iter().map(|(k, v)| (k.into(), v.into())).collect());
-
         // Not populating ObservedTimestamp, instead relying on OpenTelemetry
         // API to populate it with current time.
 
-        let mut visitor = EventVisitor {
-            log_record: &mut log_record,
-        };
+        let mut visitor = EventVisitor::default();
+        visitor.visit_metadata(meta);
```
lalitb, line 180: I am not very convinced with visiting metadata, as it seems all the metadata fields it is processing are not part of the semantic convention (including `log.name`). If that is the case, should we move this `visit_metadata` call under the experimental flag?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -99,23 +167,21 @@ where
         event: &tracing::Event<'_>,
         _ctx: tracing_subscriber::layer::Context<'_, S>,
     ) {
-        let meta = event.metadata();
+        let normalized_meta = event.normalized_metadata();
+        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
```
lalitb, line 171: Can we run the stress test to see if this is adding to any cost - probably compare with and without changes?

opentelemetry-appender-tracing/src/layer.rs
Rockin2: I renamed log.name just name like before. All the other fields in visit_metadata are under the experimental flag. Is it ok?

opentelemetry-appender-tracing/src/layer.rs
Rockin2: Seems to be this one https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/README.md :)

lalitb: yes, this is for logs - https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs

Rockin2: Interesting.

So on my machine, here is the result :

- Main : ~ 6.2M iterations/sec
- On my branch : ~ 6.3M iterations/sec (a little bit faster)
- With experimental attributes : ~ 4.7M iterations/sec (24% slower)

However, by modifying the code, I found the is_metadata function is really quick and has a very minimal impact on performance. (~6.3M vs ~6.2M).

It is adding the metadata attributes that makes the logger slower. Probably allocating memory that slow down the whole thing.

Rockin2: On Main, if I remove all attributes in the event in the test_log function, the performance goes up to 8M iterations/sec.

shaun-cox (commented):
I'm satisfied that the performance impacting changes are behind a feature flag.
I'll defer to Lalit on the overall approval of this PR.

opentelemetry-appender-tracing/src/layer.rs
Rockin2: @cijothomas Where are the basic criterion benchmarks for metrics and traces?

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/trace.rs
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/metric_counter.rs

Rockin2: Thanks! Do you think it is better to do the new benches in a new PR? I think it would be easier to compare the performance before and after this PR.

cijothomas: > Do you think it is better to do the new benches in a new PR?

Yes!

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,61 +1,129 @@
...
-    log_record: &'a mut LogRecord,
+#[derive(Default)]
+struct EventVisitor {
+    log_record_attributes: Vec<(Key, AnyValue)>,
+    log_record_body: Option<AnyValue>,
 }
 
-impl<'a> tracing::field::Visit for EventVisitor<'a> {
+#[cfg(feature = "experimental_metadata_attributes")]
+fn is_metadata(field: &'static str) -> bool {
```
cijothomas, line 20: I am not able to understand the need for `is_metadata` check. Can you help explain the need for this?

Rockin2: Unfortunately, when I have logs from the log crate, some metadata are in double.

So, if I modify the is_metadata function to always returns false, I have this :

````
otel-collector-1  | Body: Str(testlog)
otel-collector-1  | Attributes:
otel-collector-1  |      -> name: Str(log event)
otel-collector-1  |      -> log.target: Str(testotel)
otel-collector-1  |      -> log.module.path: Str(testotel)
otel-collector-1  |      -> log.source.file.path: Str(src/main.rs)
otel-collector-1  |      -> log.source.file.name: Str(main.rs)
otel-collector-1  |      -> log.source.file.line: Int(25)
otel-collector-1  |      -> log.target: Str(testotel)
otel-collector-1  |      -> log.module_path: Str(testotel)
otel-collector-1  |      -> log.file: Str(src/main.rs)
otel-collector-1  |      -> log.line: Str(25)
````

Comment (Rockin2) [reply opentelemetry-appender-tracing/src/layer.rs:20]: Here is why it is like that. https://docs.rs/tracing-log/latest/tracing_log/trait.NormalizeEvent.html

**Comment by lalitb** [comment]
@Rockin2 - Firstly, thanks for your contribution, and the patience during the review of this PR. Given the critical nature of this crate's performance, we are trying to ensure that any modifications are carefully considered and thoroughly reviewed :)

Couple of comments I have:
1. Do you think we can move the below code behind the feature flag. 
```rust
        let normalized_meta = event.normalized_metadata();
        let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
```
As per the stress test I ran locally, this normalize code to convert `log` fields to `tracing` metadata seems to affecting the performance. Once I move them behind the feature flag, the performance seems to be somewhat comparable to the main branch. 

2. The `tracing-log` dependency should be optional, and available only when the feature flag is enabled.

3. Also, not directly related to this PR, did you consider using the `opentelemetry-appender-logs` and `opentelemetry-appender-tracing` crates together, if the requirement is to export events from both `log` and `tracing` .

Rockin2:
> @Rockin2 - Firstly, thanks for your contribution, and the patience during the review of this PR. Given the critical nature of this crate's performance, we are trying to ensure that any modifications are carefully considered and thoroughly reviewed :)
> 
> Couple of comments I have:
> 
>     1. Do you think we can move the below code behind the feature flag.
> 
> 
> ```rust
>         let normalized_meta = event.normalized_metadata();
>         let meta = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
> ```
> 
> As per the stress test I ran locally, this normalize code to convert `log` fields to `tracing` metadata seems to affecting the performance, even when feature-flag was disabled. Though I didn't investigate it further, but once I moved them behind the feature flag, the performance seemed to be somewhat comparable to the main branch.
> 
> ```
>  PR  (feature-flag disabled): ~11,500,000 iterations/sec
>  PR and normalize behind feature-flag (feature-flag disabled): ~12,500,000 iterations/sec
>  main branch: ~12,500,000 iterations/sec
> ```
> 
>     2. The `tracing-log` dependency should be optional, and available only when the feature flag is enabled.
> 
>     3. Also, not directly related to this PR, did you consider using the `opentelemetry-appender-logs` and `opentelemetry-appender-tracing` crates together, if the requirement is to export events from both `log` and `tracing` .

1 and 2 : Yes I will do that. :)
3 : Unfortunately, only the tracing crate supports having multiple outputs. I have to output the logs to stdout and to an otel-collector with the otlp protocol. Also, I have an independent stdout envfilter. Tracing is very flexible.

lalitb (approved):
Looks good, as most of the non-standard attributes are behind feature flag. Thanks.

Rockin2:
> Looks good, as most of the non-standard attributes are behind feature flag. Thanks.

Thank you for your help! However, I cannot merge my PR even if you approved it.
Thank you also @cijothomas and @shaun-cox!

cijothomas:
> > > Looks good, as most of the non-standard attributes are behind feature flag. Thanks.
> > 
> > 
> > Thank you for your help! However, I cannot merge my PR even if you approved it. Thank you also @cijothomas and @shaun-cox!
> 
> It would be merged by one of the maintainers.

Will do shortly!

opentelemetry-appender-tracing/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 ### Added
 
 - Add log appender versions to loggers (#1182)
+- Normalized metadata fields with events from the log crate [#1380](https://github.com/open-telemetry/opentelemetry-rust/pull/1380)
+- Remove rendudant metadata [#1380](https://github.com/open-telemetry/opentelemetry-rust/pull/1380)
```
cijothomas, line 15:
```suggestion
- Remove redundant metadata [#1380](https://github.com/open-telemetry/opentelemetry-rust/pull/1380)
```

opentelemetry-appender-tracing/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
...
+- Follows [](https://opentelemetry.io/docs/specs/semconv/general/logs/) for metadata names [#1380](https://github.com/open-telemetry/opentelemetry-rust/pull/1380)
```
cijothomas, line 16: do we follow anything from the linked sem convention anymore?

Rockin2: No really. I can remove it.

opentelemetry-appender-tracing/CHANGELOG.md
cijothomas: Please callout that this removal of redundant data is only applicable for users doing logging via log crate.

Or put the entire changelog under a heading "Fixes the following for events emitted using log crate", to make it easy to follow changelog.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -196,11 +269,25 @@ mod tests {
...
         assert!(attributes.contains(&(Key::new("user_email"), "otel@opentelemetry.io".into())));
+        #[cfg(feature = "experimental_metadata_attributes")]
+        {
+            assert!(attributes.contains(&(Key::new("log.source.file.name"), "layer.rs".into())));
+            assert!(attributes.contains(&(
+                Key::new("log.module.path"),
+                "opentelemetry_appender_tracing::layer::tests".into()
+            )));
+            // The other 3 experimental_metadata_attributes are too unstable to check statically.
+            // Ex.: The path will be different on a Windows and Linux machine.
```
cijothomas, line 288: nit: we can still check if an attribute with that key is added, without validating the actual value? We are generally very low in code coverage, so trying to making sure new code is always supported by tests!

opentelemetry-appender-tracing/CHANGELOG.md
cijothomas: also, I'd suggest to mention that the changes are under feature flag, so folks reading the changelog can quickly discover that fact without having to open up the PR itself

Rockin2: I updated the CHANGELOG.MD. What do you think?

opentelemetry-appender-tracing/src/layer.rs
Rockin2: Will do it.

Rockin2: Done.

opentelemetry-appender-tracing/CHANGELOG.md
Rockin2: Done.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1,61 +1,130 @@
...
+            self.log_record_attributes.push((
+                "log.source.file.name".into(),
+                get_filename(filepath).to_owned().into(),
+            ));
+        }
+        #[cfg(feature = "experimental_metadata_attributes")]
+        if let Some(line) = meta.line() {
+            self.log_record_attributes
+                .push(("log.source.file.line".into(), line.into()));
+        }
```
TommyCpp, line 79: Can we wrap this into a function and annotate it with `#[cfg(feature = "experimental_metadata_attributes")]` once?

Rockin2: Done

Rockin2:
@cijothomas @TommyCpp I implemented the changes you asked. :)

PR #1381, opened 2023-11-17
Remove evicted q and hmap

### LINKED ISSUE #1283: Need for EvictedHashMap, EvictedQueue

cijothomas:
Fixes #1283

## Changes
Cleans up (now) unused files.


opentelemetry-sdk/Cargo.toml
```diff
@@ -57,10 +57,6 @@ rt-async-std = ["async-std"]
 name = "context"
 harness = false
 
-[[bench]]
-name = "key_value_map"
```
cijothomas, line 61: Goodbye to our trusted benchmark code. Thanks for the speedy (nanoseconds!) memories. Rest in peace!

shaun-cox (approved):

PR #1382, opened 2023-11-17
Attempt to write metric tests with inmemoryexp

cijothomas:
Attempting to write Metrics test to ensure SDK is doing the aggregation as expected. ~This is full of "hacks" to unblock.! Need to spend some time to understand the inner workings to fix this - If anyone can help, really appreciate it!~ Thanks @TommyCpp  who [suggested the fix](https://github.com/open-telemetry/opentelemetry-rust/pull/1382#discussion_r1398257009) for flush being stuck!

Such tests are very critical to make sure the aggregation is correct, especially considering we are considering big changes like the ones in this PR - https://github.com/open-telemetry/opentelemetry-rust/pull/1379

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -122,7 +122,7 @@ mod tests {
         assert_eq!(sum.data_points.len(), 2);
 
         // TODO: Don't think the order is guaranteed.
-        // Need to make it easy to write unit tests.s
+        // Need to make it easy to write unit tests.
         let data_point1 = &sum.data_points[0];
         data_point1
             .attributes
@@ -131,12 +131,12 @@ mod tests {
             .expect("kvp expected");
         assert_eq!(data_point1.value, 5);
 
-        let data_point1 = &sum.data_points[1];
-        data_point1
+        let data_point2 = &sum.data_points[1];
+        data_point2
             .attributes
             .iter()
             .find(|(k, v)| k.as_str() == "key1" && v.as_str() == "value2")
             .expect("kvp expected");
-        assert_eq!(data_point1.value, 3);
+        assert_eq!(data_point2.value, 3);
     }
 }
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -57,3 +57,86 @@ pub use meter_provider::*;
 pub use periodic_reader::*;
 pub use pipeline::Pipeline;
 pub use view::*;
+
+#[cfg(all(test, feature = "testing"))]
+mod tests {
+    use super::*;
+    use crate::{runtime, testing::metrics::InMemoryMetricsExporter};
+    use opentelemetry::{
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -57,3 +57,86 @@ pub use meter_provider::*;
...
+
+        // TODO: Unable to get this to work, as
+        // flush is taking forever.
+        // println!("flush started");
+        // meter_provider.force_flush().unwrap();
+        // println!("flush finished");
+        // The sleep is temporary workaround to
+        // unblock progress on writing tests like
+        // this.
+        sleep(Duration::from_millis(100)).await;
```
cijothomas, line 105: its risky to check-in this as-is, as this won't be reliable. So need to understand why flush/shutdown is not working inside unittests, and fix it..
@jtescher @TommyCpp Either of you can help given you have much more knowledge of tokio runtime/metric sdk internals!

TommyCpp: Tokio test by default uses single thread runtime, which can be tricky. See [here](https://docs.rs/tokio/latest/tokio/attr.test.html)

Maybe try annotate the test to use multi thread runtime

cijothomas: That did fix the Flush-not-progressing issue! Thank you!
I'll try to tweak the tests to make it reliable now.

hdost (approved):
Looks like a winner to me 🤘 one step closer to stability.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -57,3 +57,103 @@ pub use meter_provider::*;
...
+            .init();
+        counter.add(1, &[KeyValue::new("key1", "value1")]);
+        counter.add(1, &[KeyValue::new("key1", "value1")]);
+        counter.add(1, &[KeyValue::new("key1", "value1")]);
+        counter.add(1, &[KeyValue::new("key1", "value1")]);
+        counter.add(1, &[KeyValue::new("key1", "value1")]);
+
+        counter.add(1, &[KeyValue::new("key1", "value2")]);
+        counter.add(1, &[KeyValue::new("key1", "value2")]);
+        counter.add(1, &[KeyValue::new("key1", "value2")]);
```
lalitb, line 96: nit - attributes can be defined once, and reused - something like (not tested):
```cpp
let key_value1 = KeyValue::new("key1", "value1");
let key_value2 = KeyValue::new("key1", "value2");

for _ in 0..5 {
    counter.add(1, &[key_value1]); 

for _ in 0..3 {
    counter.add(1, &[key_value2]);
}
```

lalitb (approved):
LGTM. Nicely done :)

cijothomas:
This is far from ideal, but merging as is, give this works, and can be used to quickly validate that aggregations/merges are done correctly, especially when working on PRs like https://github.com/open-telemetry/opentelemetry-rust/pull/1379

Expecting to improve such tests in follow ups. Thanks everyone for helping review!

Issue #1378, opened 2023-11-16
[Feature]: `opentelemetry-appender-tracing` should respect information inside `tracing::Span`


0xForerunner:
It's great to be able to correlate logs with traces as is possible with the Grafana Tempo+Loki combo. This could be some sort of configuration option likely.

cijothomas:
The logs should be [correlated automatically](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_emitter.rs#L214-L217).. Are you not seeing that behavior?

lalitb:
Yes  -  this should be correlated automatically, as long as the application is instrumented using opentelemetry tracing (either through [tracing-opentelemetry](https://github.com/tokio-rs/tracing-opentelemetry) or directly using opentelemetry tracing api). Not if it is using any other (non-otel) tracing subscriber.

ewoolsey:
<img width="1608" alt="Screenshot 2023-11-16 at 9 24 20 AM" src="https://github.com/open-telemetry/opentelemetry-rust/assets/8366997/bda1003d-0784-4ad3-abb4-fcde1dc7bc03">
This is the information I'm getting in loki, with no trace_id information.

Logs are being generated with this code
```rust
        let logs = opentelemetry_otlp::new_pipeline()
            .logging()
            .with_exporter(opentelemetry_otlp::new_exporter().http())
            .with_log_config(
                opentelemetry_sdk::logs::config().with_resource(Resource::new(vec![
                    KeyValue::new("service.name", "liquidation-bot-rs"),
                ])),
            )
            .install_batch(opentelemetry_sdk::runtime::Tokio)?;

        let logging = OpenTelemetryTracingBridge::new(&logs.provider().unwrap())
            .with_filter(EnvFilter::from_default_env());
```

ewoolsey:
> Yes - this should be correlated automatically, as long as the application is instrumented using opentelemetry tracing (either through [tracing-opentelemetry](https://github.com/tokio-rs/tracing-opentelemetry) or directly using opentelemetry tracing api). Not if it is using any other (non-otel) tracing subscriber.

I can confirm things are properly instrumented with `tracing-opentelemetry`. The log in the screenshot above is inside a function with `#[instrument]` and I am correctly seeing traces and events.

cijothomas:
Can you use stdout exporter and share the output? We just want to confirm if this issue is with OTLP or Loki itself.

[...]

PR #1394, opened 2023-11-22
Propagate span_id & trace_id in logs records when converting trace events

maximebedard:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1378.

## Changes

Depends on https://github.com/tokio-rs/tracing-opentelemetry/pull/78.

This PR updates the dependency of tracing-opentelemetry to `0.22` and add a support for propagating the trace context when bridging trace events to logs. This is useful when sending otlp log records to tools such as datadog, which would then use the trace_id and span_id to correlate logs and traces.

To do so, I'm pulling the `OtelData` from the current span and converting it to a TraceContext that can be set on the log record.

### DIFF
--- a/opentelemetry-appender-tracing/Cargo.toml
+++ b/opentelemetry-appender-tracing/Cargo.toml
@@ -11,18 +11,19 @@ license = "Apache-2.0"
 rust-version = "1.65"
 
 [dependencies]
-opentelemetry = { path = "../opentelemetry", features = ["logs"] }
-opentelemetry_sdk = { path = "../opentelemetry-sdk", features = ["logs"] }
+opentelemetry = { version = "0.21", features = ["logs"] }
+opentelemetry_sdk = { version = "0.21", features = ["logs"] }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
-tracing-opentelemetry = "0.22"
+tracing-opentelemetry = { version = "0.22", optional = true }
 tracing-core = "0.1.31"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["registry", "std"] }
 once_cell = "1.13.0"
 
 [dev-dependencies]
-opentelemetry-stdout = { path = "../opentelemetry-stdout", features = ["logs"] }
+opentelemetry-stdout = { version = "0.2", features = ["logs"] }
 
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled", "opentelemetry_sdk/logs_level_enabled"]
+tracing = ["tracing-opentelemetry"]
 default = ["logs_level_enabled"]
 testing = ["opentelemetry_sdk/testing"]
--- a/opentelemetry-appender-tracing/Cargo.toml
+++ b/opentelemetry-appender-tracing/Cargo.toml
@@ -11,8 +11,8 @@ license = "Apache-2.0"
 rust-version = "1.65"
 
 [dependencies]
-opentelemetry = { version = "0.21", features = ["logs"] }
-opentelemetry_sdk = { version = "0.21", features = ["logs"] }
+opentelemetry = { version = "0.21", path = "../opentelemetry", features = ["logs"] }
+opentelemetry_sdk = { version = "0.21", path = "../opentelemetry-sdk", features = ["logs"] }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
 tracing-opentelemetry = { version = "0.22", optional = true }
 tracing-core = "0.1.31"
@@ -20,7 +20,7 @@ tracing-subscriber = { version = "0.3.0", default-features = false, features = [
 once_cell = "1.13.0"
 
 [dev-dependencies]
-opentelemetry-stdout = { version = "0.2", features = ["logs"] }
+opentelemetry-stdout = { version = "0.2", path = "../opentelemetry-stdout", features = ["logs"] }
 
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled", "opentelemetry_sdk/logs_level_enabled"]
--- a/opentelemetry-appender-tracing/Cargo.toml
+++ b/opentelemetry-appender-tracing/Cargo.toml
@@ -11,9 +11,10 @@ license = "Apache-2.0"
 rust-version = "1.65"
 
 [dependencies]
-opentelemetry = { version = "0.21", path = "../opentelemetry", features = ["logs"] }
-opentelemetry_sdk = { version = "0.21", path = "../opentelemetry-sdk", features = ["logs"] }
-tracing = {version = "0.1", default-features = false, features = ["std"]}
+opentelemetry = { path = "../opentelemetry", features = ["logs"] }
+opentelemetry_sdk = { path = "../opentelemetry-sdk", features = ["logs"] }
+tracing = { version = "0.1", default-features = false, features = ["std"] }
+tracing-opentelemetry = "0.22"
 tracing-core = "0.1.31"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["registry", "std"] }
 once_cell = "1.13.0"
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -1,7 +1,11 @@
-use opentelemetry::logs::{LogRecord, Logger, LoggerProvider, Severity};
+use opentelemetry::{
+    logs::{LogRecord, Logger, LoggerProvider, Severity, TraceContext},
+    trace::{SpanContext, TraceFlags, TraceState},
+};
 use std::borrow::Cow;
-use tracing_core::Level;
-use tracing_subscriber::Layer;
+use tracing_core::{Level, Subscriber};
+use tracing_opentelemetry::OtelData;
+use tracing_subscriber::{registry::LookupSpan, Layer};
 
 const INSTRUMENTATION_LIBRARY_NAME: &str = "opentelemetry-appender-tracing";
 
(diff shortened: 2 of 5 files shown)


cijothomas:
@maximebedard Could you sign the easycla part, which is a pre-req before accepting any contributions!

opentelemetry-appender-tracing/Cargo.toml
```diff
@@ -11,9 +11,10 @@ license = "Apache-2.0"
 rust-version = "1.65"
 
 [dependencies]
-opentelemetry = { version = "0.21", path = "../opentelemetry", features = ["logs"] }
-opentelemetry_sdk = { version = "0.21", path = "../opentelemetry-sdk", features = ["logs"] }
-tracing = {version = "0.1", default-features = false, features = ["std"]}
+opentelemetry = { path = "../opentelemetry", features = ["logs"] }
+opentelemetry_sdk = { path = "../opentelemetry-sdk", features = ["logs"] }
+tracing = { version = "0.1", default-features = false, features = ["std"] }
+tracing-opentelemetry = "0.22"
```
cijothomas, line 17: this looks odd to depend on tracing-opentelemetry!

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -90,20 +94,31 @@ where
...
-        event: &tracing::Event<'_>,
-        _ctx: tracing_subscriber::layer::Context<'_, S>,
-    ) {
+    fn on_event(&self, event: &tracing::Event<'_>, ctx: tracing_subscriber::layer::Context<'_, S>) {
         let meta = event.metadata();
         let mut log_record: LogRecord = LogRecord::default();
         log_record.severity_number = Some(severity_of_level(meta.level()));
         log_record.severity_text = Some(meta.level().to_string().into());
 
+        // Extract the trace_id & span_id from the opentelemetry extension.
```
cijothomas, line 107: https://github.com/open-telemetry/opentelemetry-rust/issues/1378 - Looks like this PR is attempting to solving the issue?

maximebedard: That's correct! I didn't notice the original issue, I'll update the PR description to include it.

opentelemetry-appender-tracing/Cargo.toml
maximebedard: Definitely. I believe it's necessary since it's where the OtelData is injected into the current span extension and where the `TraceId` and `SpanId` is generated. If there's a better way to retrieve the data, I would be happy to update the PR and remove this dependency.

maximebedard:
I'm getting the CLA situation sorted out with my employer. I'll provide updates ASAP.

opentelemetry-appender-tracing/Cargo.toml
lalitb: Can we just copy the `OtelData` definition from tracing-opentelemetry, doesn't seems to be a deep-rooted definition. And this will help get rid of dependency.

```rust
#[derive(Debug, Clone)]
pub struct OtelData {
    /// The parent otel `Context` for the current tracing span.
    pub parent_cx: opentelemetry::Context,

    /// The otel span data recorded during the current tracing span.
    pub builder: opentelemetry::trace::SpanBuilder,
}
```

In the future, we do want to have a single subscriber/layer for both traces and logs and can consolidate at that time.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -90,20 +94,31 @@ where
...
+        if let Some((trace_id, span_id)) = ctx.event_span(event).and_then(|span| {
```
lalitb, line 108: just curious, can't we get the currently active span using `ctx.lookup_current()`. How is `ctx.event_span(event)` different ?

ewoolsey:
Thanks for your hard work on this! Seems to be working well for me.

opentelemetry-appender-tracing/Cargo.toml
maximebedard: I'm not sure this works as the extension type internally uses a `HashMap<TypeId, _>`, and the TypeId of the copied definition would happen to be different. I tried an simpler example [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=09af3bc9eea68f47775794e44deea0bf) and the types are indeed not equal.

opentelemetry-appender-tracing/src/layer.rs
maximebedard: Yes I believe we can, updating!

opentelemetry-appender-tracing/Cargo.toml
lalitb: Thanks for the explanation. You are right, copying `OtelData` definition won't work here. Do you think we can make this dependency optional? And then check for the tracing context only if this dependency is met. This will ensure that the users who are using this crate solely for logging purpose are not paying cost of context lookup when no span is actually created. And this is valid also for users who are directly using the otel tracing API ( though we don't recommend it). 

As a side-note, as of now, when using both`tracing-opentelemetry` and `opentelemetry-appender-tracing`, the logs emitted with tracing macros (trace!, info!, event! etc) are exported both through SpanExporter as span-events, and the LogExporter as logs. This mayn't be desirable. It would be eventually good to integrate both `tracing-opentelemetry` and `opentelemetry-appender-tracing` crates.

maximebedard: Yes definitely, I'll see what I can do. I'll push an update ASAP 👍

lalitb: @maximebedard Just wondering if you are able to proceed further, or any help is required on this?

maximebedard: Appologies for the delay @lalitb, I was able to get the CLA situation sorted out and pushed a few commits to fix the dependency issues. I've gated the feature with the flag `tracing` for the lack of a better name, if you have any suggestions I'm all ears!

lalitb: Thanks @maximebedard 
Regarding standardizing flags naming, it is still under discussion, but close to finalise as
 - use `otel_unstable` for specs-compliant features that are not yet stable. These are temporary flags and cleaned up regularly.
 - use `<signal>-<feature-name>` for all other features. 

I think this feature falls under the second category. The signal is `logs` here, so we can use `logs_<feature-name>`. Eg `logs_tracing_context` or `logs_propogate_tracing_context`. You can decide on the name acccordingly.

ewoolsey: What else do we need to complete to get this merged?

lalitb: There are a few merge conflicts, and @maximebedard needs to fix the flag name as above.

ewoolsey: Seems like this PR has stagnated a bit, anything I can do to help move things forward? @maximebedard I could potentially take over if you don't have the time?

maximebedard: Appologies for not getting back to this one, but I don't think it's possible to fix it within the opentelemetry workspace as importing `tracing-opentelemetry` causes a circular dependency. Ultimately, the `OtelData` structure should probably be moved into the opentelemetry-sdk and consumed by the tracing-opentelemetry crate, which would avoid importing tracing-opentelemetry.

I'll close this MR since I can't make anymore progress on it until the `OtelData` structure is moved (see https://github.com/tokio-rs/tracing-opentelemetry/issues/33).

ewoolsey: So correct me if I'm wrong but this is what we need:
- Create PR here to create OtelData struct
- Create PR in `tracing-opentelemetry` to switch to using that struct
- Recreate this PR without the circular dep

I could begin that process if it would help

lalitb: Thanks @ewoolsey if you could give it a try.

lpj145:
What a amazing situation the one's top protocol to monitor services like open telemetry does't have a good integration with tracing, guys, we need to fix this thing... seriously, if you want to enable and add opentelemtry in another languages eg: tech stacks you just need to install it, configure under 3 lines, but in rust, you need todo a entire conversion just to trace works, but not for logs, by the way in the way that open telemtry follow we need to setup 3 connections of grpc just to have the data..... again... amazing...

opentelemetry-appender-tracing/Cargo.toml
ewoolsey: > Thanks @ewoolsey if you could give it a try.

@lalitb which crate do you think `OtelData` belongs in?

ewoolsey: Also, I just pulled this PR and I'm not getting any circular dependencies... Everything is compiling just fine for me.

lalitb: @ewoolsey Can you validate if the span_id and trace_id are populated correctly?

ewoolsey: Yes it's working correctly. I've got a project depending on this PR and it's functioning as expected.

lalitb: Ok, interesting. I think there won't be any error thrown by the Rust compiler for the circular dependency. Rust strong type system just allows that, but still this is a bad design. One option could be to reopen the PR for review to unblock, and follow up the circular dependency problem with a separate GitHub issue. @maximebedard - if you have any comment?

ewoolsey: I would love that personally. This is a pretty critical feature for my project and finding a way around it has been really tough.

maximebedard: The dependency issue is specific to the workspace, it would not be an issue if you use a released version of the opentelemetry crates and tracing-opentelemetry. We can see the issue in the CI job that failing here https://github.com/open-telemetry/opentelemetry-rust/actions/runs/7105523258/job/19342898991. It's not really possible to push a fix upstream unless we move the `tracing_opentelemetry::OtelData` into the current workspace.

I have a forked version that uses this patch, but I noticed that it doesn't fully solve the problem. The trace_id is only set for the root span (see https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L871-L874), any child span created would be missing the trace_id and not be correlated correctly.

hdost:
@maximebedard 👋🏼 He I am curious what made you close the PR?

lalitb:
> @maximebedard 👋🏼 Hey I am curious what made you close the PR?

It's summarised here - https://github.com/open-telemetry/opentelemetry-rust/pull/1394#discussion_r1468029776.

PR #1396, opened 2023-11-22
Metric stress test with unsorted attributes.

KallDrexx:
## Changes

To better account for sorting cost, don't pass in attributes pre-sorted, as that's unlikely to be done in real applications.  Likewise, add a fourth attribute for better sorting accountability.

### DIFF
--- a/stress/src/metrics.rs
+++ b/stress/src/metrics.rs
@@ -13,9 +13,8 @@ lazy_static! {
     static ref PROVIDER: SdkMeterProvider = SdkMeterProvider::builder()
         .with_reader(ManualReader::builder().build())
         .build();
-    static ref ATTRIBUTE_VALUES: [&'static str; 10] = [
-        "value1", "value2", "value3", "value4", "value5", "value6", "value7", "value8", "value9",
-        "value10"
+    static ref ATTRIBUTE_VALUES: [&'static str; 6] = [
+        "value1", "value2", "value3", "value4", "value5", "value6"
     ];
     static ref COUNTER: Counter<u64> = PROVIDER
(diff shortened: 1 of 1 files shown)


cijothomas:
> as that's unlikely to be done in real applications.

mm..Maybe we should ask users to do that to get maximum performance? It won't be enforced, so we can accept in unsorted order and do the sorting inside.. 
Sorting of attributes is still a perf-concern, and we need to solve it. OTel-cpp and OTel-dotnet has solved it (in diff. ways), so we should steal/copy-ideas-inspirations and do whatever it takes to fix the perf issue arising due to sorting!

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/compare/main...cijothomas:perf-sort?expand=1 This benchmarks shows 18-20% perf boost, if we avoid sorting. Ofcourse it cannot be simply done, as we need to ensure correctness, but just demonstrating that sorting is a perf eater! 

https://github.com/open-telemetry/opentelemetry-dotnet/pull/2805 - PR showing how much OTel .NET boosted perf by avoiding sorting!

KallDrexx:
I think asking them to pre-sort attributes, and pre-dedup for best performance is fine.  As far as I can tell from the `Counter_Add_Unsorted` and `Counter_Add_Sorted`, there's not a big difference for it to matter since we have to ensure they are sorted and deduplicated if they are not.

So for this PR I was mostly just trying to make sure the stress test assumes a worst case (and realistic) scenario, though I can see arguments for the existing code remaining for a "best case".

cijothomas:
Closing old PR. We dont plan to add stress tests for this particular case, but can be ad-hoc leveraged by contributors.

Issue #1300, opened 2023-10-13
[Feature Request] Deduplication of attributes


cijothomas:
As of today, Logs API/SDK [does not](https://github.com/open-telemetry/opentelemetry-rust/pull/1142/files) do de-deuplication.
Tracing API/SDK does do de-duplication for SpanAttributes, but it is being [removed](https://github.com/open-telemetry/opentelemetry-rust/pull/1293). There is no de-duplication done for Event Attributes and Link Attributes.
Metrics API/SDK does do de-duplication, but it could be [removed](https://github.com/open-telemetry/opentelemetry-rust/issues/1098).

Please comment on this issue, if you are an end-user who want a feature from this repo to do the de-duplication. This is typically the case, if you, as instrumentation author, cannot enforce it, and your backend does not support it either. When you comment here, please describe the scenario in detail, so that we can design a solution to cover that. In general, if we chose to offer a feature for this, it'll be opt-in basis, as there is significant overhead to do de-duplication.

hdost:
I'd like to bring up https://github.com/open-telemetry/opentelemetry-rust/issues/1297

cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/issues/3931#issuecomment-1991915593 From spec, only applicable for logs, still in discussion only.

PR #1398, opened 2023-11-22
Metric SDK - remove de-duplicating attributes for perf gain

### LINKED ISSUE #1098: Remove de-duping of attribute keys
In the current metrics SDK implementation, attribute sets with entries that have the same keys are de-duped. The "first" unique key is taken and the rest is dropped. This behavior is not fully defined in the spec so it is up to languages to decide the specific implementation. De-duping does not make sense for a couple of reasons:

1. Only the first key seen is kept, the rest are dropped. Is this "correct" behavior? Who is to say that the first key is the one that the user wants?
2. Performance improvement only exists for the case in which there are multiple of the same key, which is uncommon. The majority of the cases, keys within an AttributeSet will have different values, in which case, performance actually degrades because of the constant checks of key existence in the `seen` hashmap.

I ran 2 benchmark tests demonstrating this, the first `Counter/AddUniqueAttr` makes a call to add with 1million unique keys and the seconds makes a call to add with 1 million of the same keys.

With de-duping:
```
Counter/AddUniqueAttr   time:   [338.88 ms **342.29 ms** 345.65 ms]
                        change: [+26.244% +28.995% +31.699%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
Counter/AddDupeAttr     time:   [22.946 ms **23.200 ms** 23.475 ms]
                        change: [-90.635% -90.503% -90.361%] (p = 0.00 < 0.05)
                        Performance has improved.
```
Without de-duping:
```
Counter/AddUniqueAttr   time:   [236.92 ms **240.31 ms** 244.08 ms]
                        change: [-0.0665% +1.5192% +3.3808%] (p = 0.08 > 0.05)

Counter/AddDupeAttr     time:   [186.02 ms **187.53 ms** 189.18 ms]
                        change: [+1.8287% +2.9432% +4.0648%] (p = 0.00 < 0.05)
                        Performance has regressed.
```
Of course de-duping will have tremendous time improvements for duplicate attributes but this is probably not worth the overhead of additional time added with non-duplicate attributes.

Thoughts?

cijothomas:
Fixes #1098. This adds to the gains achieved from https://github.com/open-telemetry/opentelemetry-rust/pull/1379!

Similar to [Tracing ](https://github.com/open-telemetry/opentelemetry-rust/pull/1293)and Logs, this PR removes the attribute de-duplication from the SDK. Changelog is advising users to https://github.com/open-telemetry/opentelemetry-rust/issues/1300, feedback from which can be used to evaluate if we need to bring this back as a opt-in feature in OTel SDK.

## Changes
Remove de-deduplication of metric attributes from SDK.
Add test to validate the behavior and assert that it is a conscious choice.
Prometheus Exporter already handles this (when duplicates arise when sanitizing), but order is not guaranteed. Tests are modified to account for this.

## Performance gains
20% gain for Counter.Add() hot path, when 4 attributes are used. More attributes will show more gains.
Numbers from my machine:

main branch
```
Counter_Add_Sorted      time:   [650.97 ns 659.10 ns 668.39 ns]
  2 (2.00%) high mild
  4 (4.00%) high severe

Counter_Add_Unsorted    time:   [645.21 ns 650.88 ns 656.56 ns]
  2 (2.00%) high severe
```

This PR
```
Counter_Add_Sorted      time:   [527.00 ns 531.73 ns 537.88 ns]
                        change: [-18.633% -15.611% -11.545%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  10 (10.00%) high severe

Counter_Add_Unsorted    time:   [532.62 ns 537.23 ns 542.44 ns]
                        change: [-19.127% -17.803% -16.416%] (p = 0.00 < 0.05)
                        Performance has improved.
```

main branch
50% boost in the time needed for making AttributeSet from slice!
```
AttributeSet_without_duplicates
                        time:   [217.74 ns 220.07 ns 222.85 ns]
  3 (3.00%) high mild
  3 (3.00%) high severe

AttributeSet_with_duplicates
                        time:   [219.37 ns 229.60 ns 241.22 ns]
Found 15 outliers among 100 measurements (15.00%)
  2 (2.00%) high mild
  13 (13.00%) high severe
```

This PR:
```
AttributeSet_without_duplicates
                        time:   [99.515 ns 100.37 ns 101.33 ns]
                        change: [-54.598% -53.572% -52.608%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  7 (7.00%) high severe

AttributeSet_with_duplicates
                        time:   [105.90 ns 107.11 ns 108.55 ns]
                        change: [-54.672% -52.847% -51.089%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Stress Tests:
main branch
```
Number of threads: 4
Throughput: 7,660,800 iterations/sec
Throughput: 7,904,400 iterations/sec
Throughput: 7,685,400 iterations/sec
Throughput: 7,048,800 iterations/sec
Throughput: 7,251,000 iterations/sec
```

this PR
```
Number of threads: 4
Throughput: 8,148,800 iterations/sec
Throughput: 8,265,800 iterations/sec
Throughput: 8,422,600 iterations/sec
Throughput: 8,420,600 iterations/sec
Throughput: 8,473,200 iterations/sec
```

Stress tests use only 3 attributes, so the improvement seems smaller compared
to benchmarks which uses 4 attributes!

### DIFF
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -373,7 +373,7 @@ impl prometheus::core::Collector for Collector {
 /// Maps attributes into Prometheus-style label pairs.
 ///
 /// It sanitizes invalid characters and handles duplicate keys (due to
-/// sanitization or when user provides duplicate keys) by sorting and 
+/// sanitization or when user provides duplicate keys) by sorting and
 /// concatenating the values following the spec.
 fn get_attrs(kvs: &mut dyn Iterator<Item = (&Key, &Value)>, extra: &[LabelPair]) -> Vec<LabelPair> {
     let mut keys_map = BTreeMap::<String, Vec<String>>::new();
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -373,7 +373,8 @@ impl prometheus::core::Collector for Collector {
 /// Maps attributes into Prometheus-style label pairs.
 ///
 /// It sanitizes invalid characters and handles duplicate keys (due to
-/// sanitization) by sorting and concatenating the values following the spec.
+/// sanitization or when user provides duplicate keys) by sorting and 
+/// concatenating the values following the spec.
 fn get_attrs(kvs: &mut dyn Iterator<Item = (&Key, &Value)>, extra: &[LabelPair]) -> Vec<LabelPair> {
     let mut keys_map = BTreeMap::<String, Vec<String>>::new();
     for (key, value) in kvs {
--- a/opentelemetry-prometheus/tests/data/sanitized_labels.txt
+++ b/opentelemetry-prometheus/tests/data/sanitized_labels.txt
@@ -1,6 +1,6 @@
 # HELP foo_total a sanitary counter
 # TYPE foo_total counter
-foo_total{A_B="Q",C_D="Y;Z",otel_scope_name="testmeter",otel_scope_version="v0.1.0"} 24.3
+foo_total{A_B="Q;X",C_D="Y;Z",otel_scope_name="testmeter",otel_scope_version="v0.1.0"} 24.3
 # HELP otel_scope_info Instrumentation Scope metadata
 # TYPE otel_scope_info gauge
 otel_scope_info{otel_scope_name="testmeter",otel_scope_version="v0.1.0"} 1
--- a/opentelemetry-prometheus/tests/integration_test.rs
+++ b/opentelemetry-prometheus/tests/integration_test.rs
@@ -135,7 +135,13 @@ fn prometheus_exporter_integration() {
             builder: ExporterBuilder::default().without_units(),
             record_metrics: Box::new(|meter| {
                 let attrs = vec![
-                    // exact match, value should be overwritten
+                    // exact match, SDK will not do de-duplication,
+                    // values should be concatenated.
+                    // The order X;Q vs Q:X is not guaranteed.
+                    // We expect end-users to not produce duplicates in the
+                    // first place, but if that cannot be done,
+                    // we can offer a opt-in feature in SDK to do it.
+                    // https://github.com/open-telemetry/opentelemetry-rust/issues/1300
                     Key::new("A.B").string("X"),
                     Key::new("A.B").string("Q"),
                     // unintended match due to sanitization, values should be concatenated
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -24,6 +24,12 @@
 - **Breaking** Remove `TextMapCompositePropagator` [#1373](https://github.com/open-telemetry/opentelemetry-rust/pull/1373). Use `TextMapCompositePropagator` in opentelemetry API.
 
 - [#1375](https://github.com/open-telemetry/opentelemetry-rust/pull/1375/) Fix metric collections during PeriodicReader shutdown
+- **Breaking**
+  [#1397](https://github.com/open-telemetry/opentelemetry-rust/issues/1397)
+  Removes de-duplication of Metric attribute keys to achieve performance gains.
+  Please share [feedback
+  here](https://github.com/open-telemetry/opentelemetry-rust/issues/1300), if
+  you are affected.
 
 ## v0.21.1
 
--- a/opentelemetry-sdk/src/attributes/set.rs
+++ b/opentelemetry-sdk/src/attributes/set.rs
@@ -111,7 +111,7 @@ impl From<&[KeyValue]> for AttributeSet {
         let mut vec = values
             .iter()
             .rev()
-            .filter_map(|kv| Some(HashKeyValue(kv.clone())))
+            .map(|kv| HashKeyValue(kv.clone()))
             .collect::<Vec<_>>();
         vec.sort_unstable();
 
--- a/opentelemetry-sdk/src/attributes/set.rs
+++ b/opentelemetry-sdk/src/attributes/set.rs
@@ -111,9 +111,7 @@ impl From<&[KeyValue]> for AttributeSet {
         let mut vec = values
             .iter()
             .rev()
-            .filter_map(|kv| {
-                Some(HashKeyValue(kv.clone()))
-            })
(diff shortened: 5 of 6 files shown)


hdost:
So going back to our conversation from before the gain here is not the 7x we saw. 
I guess my interpretation was that we'd look at bounded instruments and/or allowing users to pass in prebuilt AttributeSets. 
If we're going to just break it this seems like we should put it behind a feature flag which was also something we discussed.

cijothomas:
> So going back to our conversation from before the gain here is not the 7x we saw. This is closer to a 6-8% improvement based on the stress test.

There is no 7x gain with this PR. I am trying to understanding why did you expect 7x here? Apologies if I missed something obvious..
The perf gains from this change are shared in the PR description. If we have more attributes, there will be more gains, of course, but just picked one example (3 attributes for stress, 4 attributes for criterion) to show the gains!

> we'd look at bounded instruments and/or allowing users to pass in prebuilt AttributeSets

Yes absolutely! That is a separate improvement, helping users who has key/values known ahead. This PR is not attempting that improvement.

> If we're going to just break it this seems like we should put it behind a feature flag

If users have duplicate keys in their attributes, that is already an indication of something wrong with their instrumentation. I don't think its fair to ask every user to pay for the perf penality to dedup! I am totally open to bringing de-dup as an opt-in feature, which is why I has opened [this issue to gather feedback](https://github.com/open-telemetry/opentelemetry-rust/issues/1300).

hdost:
TL;DR I think we should add the feature flag before merging this. 

> > So going back to our conversation from before the gain here is not the 7x we saw. This is closer to a 6-8% improvement based on the stress test.
> 
> 
> 
> There is no 7x gain with this PR. I am trying to understanding why did you expect 7x here? Apologies if I missed something obvious..
> 

My point was maybe not well made.
https://github.com/open-telemetry/opentelemetry-rust/pull/1379 we saw significant improvement on the order of 6-7x improvement (from what I understand and doesn't break any previous functionality. 
This PR breaks backwards compatibility and is not as significant improvement. 


> > we'd look at bounded instruments and/or allowing users to pass in prebuilt AttributeSets
> 
> Yes absolutely! That is a separate improvement, helping users who has key/values known ahead. This PR is not attempting that improvement.
> 

I think we'd be better off focusing on providing those to provide performance. And if we're going to break functionality I really think we need to be more thoughtful about it. Yes we want performance, but if we are constantly breaking things people have little incentive to move forward and with our latest release I don't think we've done a good job of providing a path forward. 

This is a little wider point but it's relevant. I don't want to be the barrier to progress, but I also don't want no one to use this because they experience frustration through upgrades. 

> 
> > If we're going to just break it this seems like we should put it behind a feature flag
> 
> 
> 
> If users have duplicate keys in their attributes, that is already an indication of something wrong with their instrumentation. I don't think its fair to ask every user to pay for the perf penality to dedup! I am totally open to bringing de-dup as an opt-in feature, which is why I has opened [this issue to gather feedback](https://github.com/open-telemetry/opentelemetry-rust/issues/1300). 
> 

On the point above I think that we should have a feature flag on this to allow people to work with the old behavior. Feedback is great but we don't give people the option to try both ways as they would have to use an older version which might miss other internal improvements.

cijothomas:
@hdost I'm not entirely clear on who might be affected by this change. The API itself hasn't been changed; it's more a shift in behavior from not deduplicating in the SDK to doing so in the Prometheus exporter. Is it common for anyone to intentionally send duplicate attributes, expecting the SDK to handle the deduplication? Considering there are no extensibility points like SpanProcessor/LogRecordProcessor, it seems unlikely that additional attributes could be added even inadvertently, leading to duplication. Thus, the individual emitting the measurement should have complete control and can take steps to avoid passing duplicate keys. If there are other scenarios I'm not considering, I'm open to understanding more about them. From what I've seen in OTel .NET, which has never performed deduplication either in the SDK or Prometheus Exporter, this issue of duplication hasn't been a point of concern.

I do understand the importance of being mindful with any changes that could potentially break existing workflows. However, in this instance, I'm not convinced that this change is of that nature. This is similar to the changes we've made previously with Logs and Traces. We didn't implement feature flags for traces or logs, and I'm not sure it's necessary for metrics either. I strongly believe that the **default behavior should prioritize efficiency**, and I fully support the idea of a feature flag or opt-in approach for the less efficient route involving deduplication, rather than defaulting to it.

Would you reconsider your objection to this PR?

hdost:
> TL;DR I think we should add the feature flag before merging this.

> I strongly believe that the **default behavior should prioritize efficiency**, and I fully support the idea of a feature flag or opt-in approach for the less efficient route involving deduplication, rather than defaulting to it.

I guess we're on the same page that there should be a feature flag.

As I suggested it could be something like: `metrics-enable-deduplication`

KallDrexx:
Maybe it's prudent to have concrete idea of what the ramifications of not-deduplicating is?   From my limited knowledge, there are two ramifications

1. You may end up with duplicated time series (e.g. `{color: red, shape: circle, color:red}` vs `{color:red, shape: cicle}` would be different time series). 
2. Some exporters may fail (e.g. prometheus non-otlp).

There's always a balance between performance and creating the pit of success for users, and it's hard for me to give much of an opinion without how accidentally violating the duplicate key rules in this case actually works.

Fwiw, I don't have a problem enabling this to be opt-in as a "I am aware of how I build my attributes and explicitly know that providing duplicate attributes may cause bad data".  I'm a big proponent of pit of success, and a feature flag to enable better performance without code changes doesn't seem too erroneous.

KallDrexx:
I guess my primary concerns is with `{color: red, shape: circle, color:red}` vs `{color:red, shape: cicle}`.

If we don't dedup, then these are stored in the `ValueMap` as two different attribute sets, which means a counter against one won't be applied to the other.  This means you could have two different values in both attribute sets.

I assume the backends will automatically aggregate them into a single time series in that case?

cijothomas:
> I assume the backends will automatically aggregate them into a single time series in that case?

Again, this is backend specific.

cijothomas:
Closing for now, as we are exploring other options, and will come back to this after https://github.com/open-telemetry/opentelemetry-rust/pull/1421

Issue #1405, opened 2023-11-27
[Feature]: Stress Test is not leveraging max threads 


cijothomas:
### Describe the solution you'd like:

https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/throughput.rs#L73 Shows that it only uses 1 less thread than it should be using, presumably to reserve one thread for the main loop, calculating the stats/displaying on console etc.
Given the main thread would be sleeping almost all the time, wondering if we should use all allowed threads for doing the actual work... 

In short, we are not fully pushing the CPU to its max, as I only see CPU utilization around 60-70% when running this. On comparison, similar stress test from https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/OpenTelemetry.Tests.Stress.Metrics shows 100% CPU utilization!

ThomsonTan:
How are cores in total in your test environment?
> I only see CPU utilization around 60-70% when running this

Issue #1387, opened 2023-11-21
[Feature]: Allow precreation of AttributeSets for metrics


KallDrexx:
### Related Problems?

In our application we manage counters that get incremented on a per-network packet basis.  Opentelemetry metrics has a high performance cost for this, and even with PR #1379 `AttributeSet::from()` ends up consuming 6.8% of our application's CPU time in one of our load tests.  In most of these cases, the attributes' keys and values are stable and thus cacheable.  

### Describe the solution you'd like:

Allowing the application calling opentelemetry's APIs to cache `AttributeSet`s and incur the costs only once per set of attributes can go a long way to improving performance.  

At a quick glance, we can either add a second `Counter.add()` method that takes an `AttributeSet` or make a backward incompatible change where `Counter::add()` no longer takes a `&[KeyValue]` and instead takes an `AttributeSet`.  

My opinion is the latter is a better long term approach, as it encourages consumers to think about the cost of creating `AttributeSets`.  Even if they are not in a scenario where they can cache the attributes (they aren't stable) this still gives the benefit of having less cloning required (today cloning is required since a slice is being passed in by reference, thus we can't consume the `KeyValue`s to form the `AttributeSet`.

Note that this does require pulling `AttributeSet` out of the `opentelemetry_sdk` crate and moving it to a higher level crate.

Note that I'm willing to do this work.

cijothomas:
Thanks for opening this!

Still trying to fully understand the value this brings... I think what this allows us is to avoid the cost of _turning a slice of attributes into AttributeSet_ by reusing the `AttributeSet`.. If we make that cost really small, then do we still need this? Once #1379 is completed, I expect that cost would be much smaller, and not worth breaking current API or introducing new api!

Also, the [bound instrument idea](https://github.com/open-telemetry/opentelemetry-rust/issues/1374), could be used if one knows keys+values ahead of time, so this would most likely be not required at all!

> Note that I'm willing to do this work
My suggestion: Let us wait for #1379 to finish, and re-evaluate this, before you spending more time on this!

KallDrexx:
So there are two parts to your response that I'll answer separately.

**The first is, is this needed after #1379 is merged?**  

I think so.  While #1379 does improve `AttributeSet::from()` times, it's still far from negligible.  As a reference, in our application tested against #1379, `AttributeSet::from()` still takes up about 6% of total CPU time in our application due to creating these over and over.  

[...]

PR #1407, opened 2023-11-27
Precalculated attribute set hashes

KallDrexx:
The hash of `AttributeSet`s are expensive to compute, as they have to be computed for each key and value in the attribute set.  This hash is used by the `ValueMap` to look up if we are already aggregating a time series for this set of attributes or not.  Since this hashmap lookup occurs inside a mutex lock, no other counters can execute their `add()` calls while this hash is being calculated, and therefore contention in high throughput scenarios exists.

This PR calculates and caches the hashmap at creation time. This improves throughput because the hashmap is calculated by the thread creating the `AttributeSet` and is performed outside of any mutex locks, meaning hashes can be computed in parallel and the time spent within a mutex lock is reduced.  As larger sets of attributes are used for time series, the benefits of reduction of lock times should be greater.

The stress test results of this change for different thread counts are:

| Thread Count | Main        |  PR           |
| -------------- | ---------- | --------- |
| 2                       | 3,376,040 | 3,310,920 |
| 3                       | 5,908,640 | 5,807,240 |
| 4                       | 3,382,040 | 8,094,960 |
| 5                       | 1,212,640 | 9,086,520 |
| 6                       | 1,225,280 | 6,595,600 |

The non-precomputed hashes starts feeling contention with 4 threads, and drops substantially after that while precomputed hashes doesn't start seeing contention until 6 threads, and even then we still have 5-6x more throughput after contention due to reduced locking times.

While these benchmarks may not be "realistic" (since most applications will be doing more work in between counter updates) it does show a benefit of better parallelism and the opportunity to reduce lock contention at the cost of only 8 bytes per time series (so a total of 16KB additional memory at maximum cardinality).

Bechmark results:

main: 

```
Counter_Add_Sorted      time:   [704.41 ns 710.47 ns 717.13 ns]
Found 14 outliers among 100 measurements (14.00%)
  10 (10.00%) high mild
  4 (4.00%) high severe

Counter_Add_Unsorted    time:   [723.78 ns 749.82 ns 781.35 ns]
  9 (9.00%) high mild
  2 (2.00%) high severe
```

PR: 
```
Counter_Add_Sorted      time:   [713.96 ns 717.19 ns 721.08 ns]
                        change: [+1.6730% +3.0990% +4.6842%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) high mild
  4 (4.00%) high severe

Counter_Add_Unsorted    time:   [713.50 ns 716.58 ns 720.07 ns]
                        change: [-10.737% -7.1220% -3.8036%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  1 (1.00%) high severe
```

Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/src/attributes/set.rs
+++ b/opentelemetry-sdk/src/attributes/set.rs
@@ -1,3 +1,4 @@
+use std::collections::hash_map::DefaultHasher;
 use std::collections::HashSet;
 use std::{
     cmp::Ordering,
@@ -104,8 +105,8 @@ impl Eq for HashKeyValue {}
 ///
 /// This must implement [Hash], [PartialEq], and [Eq] so it may be used as
 /// HashMap keys and other de-duplication methods.
-#[derive(Clone, Default, Debug, Hash, PartialEq, Eq)]
-pub struct AttributeSet(Vec<HashKeyValue>);
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub struct AttributeSet(Vec<HashKeyValue>, u64);
 
 impl From<&[KeyValue]> for AttributeSet {
     fn from(values: &[KeyValue]) -> Self {
@@ -123,7 +124,12 @@ impl From<&[KeyValue]> for AttributeSet {
             .collect::<Vec<_>>();
         vec.sort_unstable();
 
-        AttributeSet(vec)
+        let mut hasher = DefaultHasher::new();
+        for item in &vec {
+            item.hash(&mut hasher);
+        }
+
+        AttributeSet(vec, hasher.finish())
     }
 }
 
@@ -135,7 +141,12 @@ impl From<&Resource> for AttributeSet {
             .collect::<Vec<_>>();
         vec.sort_unstable();
 
-        AttributeSet(vec)
+        let mut hasher = DefaultHasher::new();
+        for item in &vec {
+            item.hash(&mut hasher);
+        }
+
+        AttributeSet(vec, hasher.finish())
     }
 }
 
@@ -161,5 +172,11 @@ impl AttributeSet {
     /// Iterate over key value pairs in the set
     pub fn iter(&self) -> impl Iterator<Item = (&Key, &Value)> {
         self.0.iter().map(|kv| (&kv.0.key, &kv.0.value))
+    }
+}
+
+impl Hash for AttributeSet {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write_u64(self.1)
     }
 }
--- a/opentelemetry-sdk/src/attributes/set.rs
+++ b/opentelemetry-sdk/src/attributes/set.rs
@@ -111,7 +111,7 @@ pub struct AttributeSet(Vec<HashKeyValue>, u64);
 impl From<&[KeyValue]> for AttributeSet {
     fn from(values: &[KeyValue]) -> Self {
         let mut seen_keys = HashSet::with_capacity(values.len());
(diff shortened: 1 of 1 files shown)


KallDrexx:
Also note that precomputing hashes has the opportunity for further benefits if bounded instruments is implemented, or if #1387 is done.

lalitb:
Thanks for the PR. Can you also add the benchmark result from main branch (without these changes) for comparison?

KallDrexx:
> Thanks for the PR. Can you also add the benchmark result from main branch (without these changes) for comparison?

Can you be more specific of which benchmarks you are looking for?  I posted stress test results between `main` and this pr, as well as the `counter_add` benchmarks between main and this pr.

KallDrexx:
I added the benchmark run I had done previously from main before switching back to the PR for a run.

lalitb:
> I added the benchmark run I had done previously from main before switching back to the PR for a run.

Yeah, I meant the benchmark from main. Probably I missed it if it was already there :)

opentelemetry-sdk/src/attributes/set.rs
```diff
@@ -135,7 +141,12 @@ impl From<&Resource> for AttributeSet {
             .collect::<Vec<_>>();
         vec.sort_unstable();
 
-        AttributeSet(vec)
+        let mut hasher = DefaultHasher::new();
+        for item in &vec {
+            item.hash(&mut hasher);
+        }
+
+        AttributeSet(vec, hasher.finish())
```
shaun-cox, line 149: This code is duplicated above.  Suggest a private `new` method accepting a `Vec` and put the hash computation in there, once.  (Also suggest seeing if the four-line for-loop could make use of `std::iter::fold` instead... e.g. `vec.iter().fold(&mut hasher, |&mut hasher, item| item.hash(hasher))` or something.

shaun-cox (changes requested):

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/issues/1405 might be skewing some results.. When we think we use 2 threads due to 2 cpus, we are doing 1 less actually!

cijothomas:
> > I added the benchmark run I had done previously from main before switching back to the PR for a run.
> 
> Yeah, I meant the benchmark from main. Probably I missed it if it was already there. Thanks :)

benchmarks are probably not the right tool to show gains from this change! (or anything related to improving contentions!) Stress test is proving to be helpful, despite some [limitations](https://github.com/open-telemetry/opentelemetry-rust/issues/1405)!

cijothomas:
> > benchmarks are probably not the right tool to show gains from this change! (or anything related to improving contentions!) Stress test is proving to be helpful, despite some #1405!
> 
> Not to see the gain, but to ensure that there is no adverse effect on benchmark results. Not related to this PR, but optimizing for concurrency can sometimes introduce additional overhead in single-threaded or low-load scenarios. So good to compare both :)

Good point! thanks

opentelemetry-sdk/src/attributes/set.rs
KallDrexx: Done

KallDrexx:
> #1405 might be skewing some results.. When we think we use 2 threads due to 2 cpus, we are doing 1 less actually!

I had noticed this the other day when I tried setting thread count to 1 and got 0 throughput.  

That being said even if you take 1 away from the `thread count` values in my table, you still get the same results.

cijothomas (approved):

jtescher (approved):

PR #1408, opened 2023-11-28
Invalid View to proceed as if view did not exist

cijothomas:
Trying to see if my interpretation of [this part of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#measurement-processing) is correct. In case of a view config that has invalid stream configuration, should we proceed as if the view did not exist completely? Or should we ignore the specific part of view that was invalid, and proceed with rest of the view config?

For this specific case, I simulated invalid stream config by providing invalid bounds. Should we ignore the bounds and fallback to default bounds? OR should we ignore the whole view config itself?

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -221,20 +221,20 @@ mod tests {
         let exporter = InMemoryMetricsExporter::default();
         let reader = PeriodicReader::builder(exporter.clone(), runtime::Tokio).build();
         let criteria = Instrument::new().name("test_histogram");
-        let stream_invalid_aggregation = Stream::new()           
-           .aggregation(Aggregation::ExplicitBucketHistogram 
-            {
+        let stream_invalid_aggregation = Stream::new()
+            .aggregation(Aggregation::ExplicitBucketHistogram {
                 boundaries: vec![0.9, 1.9, 1.2, 1.3, 1.4, 1.5], // invalid boundaries
                 record_min_max: false,
             })
-           .name("test_histogram_renamed")
-           .unit(Unit::new("test_unit_renamed")
-        );
+            .name("test_histogram_renamed")
+            .unit(Unit::new("test_unit_renamed"));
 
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -209,4 +209,57 @@ mod tests {
...
+        let stream_invalid_aggregation = Stream::new()
+            .aggregation(Aggregation::ExplicitBucketHistogram {
+                boundaries: vec![0.9, 1.9, 1.2, 1.3, 1.4, 1.5], // invalid boundaries
+                record_min_max: false,
+            })
+            .name("test_histogram_renamed")
+            .unit(Unit::new("test_unit_renamed"));
+
+        let view =
+            new_view(criteria, stream_invalid_aggregation).expect("Expected to create a new view");
```
cijothomas, line 233: Wondering how/where would we validate the view config, when one is using https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/view.rs#L56-L63 instead? Might require us to move the validation elsewhere....

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/1414 Opened this issue to track doing a proper fix.

opentelemetry-sdk/src/metrics/view.rs
```diff
@@ -139,9 +139,10 @@ pub fn new_view(criteria: Instrument, mask: Stream) -> Result<Box<dyn View>> {
             Ok(_) => agg = Some(ma.clone()),
             Err(err) => {
                 global::handle_error(MetricsError::Other(format!(
-                    "{}, not using aggregation with view. criteria: {:?}, mask: {:?}",
+                    "{}, Proceeding as if View did not exist. criteria: {:?}, mask: {:?}",
```
TommyCpp, line 142: nit: shouldn't this start with lower case?

TommyCpp (approved):

PR #1410, opened 2023-11-29
Add experimental synchronous gauge 

### LINKED ISSUE #1242: Support New instrument - Gauge
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#gauge New spec. 
Need API changes, plus corresponding SDK changes.

lalitb:
Fixes #1242 

## Changes

Adds synchronous gauge, as described in specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#gauge.

The feature is experimental in specs, so the API to create the Gauge instrument is behind the feature flag - `otel_unstable`. The feature flag can be removed once the specs are stable.

### DIFF
--- a/examples/metrics-basic/Cargo.toml
+++ b/examples/metrics-basic/Cargo.toml
@@ -11,3 +11,7 @@ opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["metrics", "
 opentelemetry-stdout = { path = "../../opentelemetry-stdout", features = ["metrics"]}
 tokio = { version = "1.0", features = ["full"] }
 serde_json = {version = "1.0"}
+
+[features]
+default = ["sync-gauge-experimental"]
+sync-gauge-experimental = ["opentelemetry/sync-gauge-experimental"]
--- a/examples/metrics-basic/Cargo.toml
+++ b/examples/metrics-basic/Cargo.toml
@@ -6,12 +6,8 @@ license = "Apache-2.0"
 publish = false
 
 [dependencies]
-opentelemetry = { path = "../../opentelemetry", features = ["metrics", "sync-gauge-experimental"] }
+opentelemetry = { path = "../../opentelemetry", features = ["metrics"] }
 opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["metrics", "rt-tokio"] }
 opentelemetry-stdout = { path = "../../opentelemetry-stdout", features = ["metrics"]}
 tokio = { version = "1.0", features = ["full"] }
 serde_json = {version = "1.0"}
-
-[features]
-default = ["sync-gauge-experimental"]
-sync-gauge-experimental = ["opentelemetry/sync-gauge-experimental"]
--- a/examples/metrics-basic/Cargo.toml
+++ b/examples/metrics-basic/Cargo.toml
@@ -6,7 +6,7 @@ license = "Apache-2.0"
 publish = false
 
 [dependencies]
-opentelemetry = { path = "../../opentelemetry", features = ["metrics"] }
+opentelemetry = { path = "../../opentelemetry", features = ["metrics", "sync-gauge-experimental"] }
 opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["metrics", "rt-tokio"] }
 opentelemetry-stdout = { path = "../../opentelemetry-stdout", features = ["metrics"]}
 tokio = { version = "1.0", features = ["full"] }
--- a/examples/metrics-basic/Cargo.toml
+++ b/examples/metrics-basic/Cargo.toml
@@ -6,8 +6,12 @@ license = "Apache-2.0"
 publish = false
 
 [dependencies]
-opentelemetry = { path = "../../opentelemetry", features = ["metrics"] }
+opentelemetry = { path = "../../opentelemetry", features = ["metrics", "otel_unstable"] }
 opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["metrics", "rt-tokio"] }
 opentelemetry-stdout = { path = "../../opentelemetry-stdout", features = ["metrics"]}
 tokio = { version = "1.0", features = ["full"] }
 serde_json = {version = "1.0"}
+
+[features]
+default = ["otel_unstable"]
+otel_unstable = ["opentelemetry/otel_unstable"]
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 ### Added
 
-
 - [#1410](https://github.com/open-telemetry/opentelemetry-rust/pull/1410) Add experimental synchronous gauge
 
 ### Changed
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## vNext
 
+### Added
+
+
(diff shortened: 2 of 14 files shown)


examples/metrics-basic/Cargo.toml
```diff
@@ -6,8 +6,12 @@ license = "Apache-2.0"
 publish = false
 
 [dependencies]
-opentelemetry = { path = "../../opentelemetry", features = ["metrics"] }
+opentelemetry = { path = "../../opentelemetry", features = ["metrics", "sync-gauge-experimental"] }
```
cijothomas, line 9: maybe we should develop some convention saying, all experimental features must begin with "experimental-feature"?

opentelemetry/src/metrics/mod.rs
```diff
@@ -179,6 +179,36 @@ pub trait InstrumentProvider {
         )))
     }
 
+    /// creates an instrument for recording independent values.
+    fn u64_gauge(
```
cijothomas, line 183: this needs to be under the feature flag?

opentelemetry-sdk/src/metrics/reader.rs
```diff
@@ -144,6 +145,7 @@ impl AggregationSelector for DefaultAggregationSelector {
             | InstrumentKind::UpDownCounter
             | InstrumentKind::ObservableCounter
             | InstrumentKind::ObservableUpDownCounter => Aggregation::Sum,
+            InstrumentKind::Gauge => Aggregation::LastValue,
```
cijothomas, line 148: Exporters need no change, as the aggregation is anyway LastValue, right?

opentelemetry/src/metrics/mod.rs
lalitb: This is not user-facing (not `pub`), so can't be left as it is. I have only kept the Meter API to create a Gauge instrument under the feature flag.

opentelemetry-sdk/src/metrics/reader.rs
lalitb: Yes, the default is LastValue, and since both Gauge and Observable Gauge are absolute values, the aggregation and exporters needn't change.

examples/metrics-basic/Cargo.toml
lalitb: Yeah, good point. Prefixing `experimental` gives more clarity. I can make this `experimental-sync-gauge` if we agree.

cijothomas: `experimental-sync-gauge` This looks good for me.

cijothomas (approved):
LGTM.
Lets also add changelog entry for api,sdks.

shaun-cox (changes requested):
Would like others to weigh in on the comment I left about feature flags design and where experimental features should reside, in general...

examples/metrics-basic/Cargo.toml
lalitb: > Should we consider an orthogonal unstable feature flag in general, which can be combined with other features which enable signals such as metrics, tracing, logs, rather than a unique feature flag name for all the combinations?

That would make things simpler, and easy to maintain. Just wondering if we need to take care of a scenario where the user wants to enable only a specific feature flag under (say) `metrics`. With generic flag like `unstable`, this mayn't be possible.

> Finally, how does this overlap with our previous decision about having opentelemetry crate be only for the stable API and the other opentelemetry-contrib crate being for the experimental or proposed, but not yet stable, additions?

Probably I missed the discussion. My proposal would be to have `opentelemetry-contrib` crate (under `otel-rust-contrib` repo) only for features that are not part of the specs. Convenience/wrapper API would be one such candidate. The main repo should have under feature flag
  - unstable implementation for stable/experimental specs feature
  - stable implementation of experimental specs feature.

From an end-user perspective, it would be easy to toggle the feature flag without changing dependencies. While we can debate whether this would increase the complexity of the code base without clear boundaries between stable and unstable implementation, but moving forward this seems manageable. Considering that all three signals – logs, traces, and metrics – are stable, significant changes in the specification aren't anticipated, making this strategy feasible.

cijothomas: > The main repo should have under the feature flag:
> 
> unstable implementation for stable/experimental specs feature
> stable implementation of experimental specs feature.

Agree with this.

TommyCpp (approved):

examples/metrics-basic/Cargo.toml
TommyCpp: I think we considered to use features for stable and non experimental feature to keep the compiling size minimal. Do we still see we need this going forward?

lalitb: Ok, as suggested above by @shaun-cox  and @jtescher, I have updated to use the conditional compilation of unstable/experimental features using the `otel_unstable` cfg flag - similar to how the `tokio` library using `tokio_unstable` for all unstable features.
So, to build the metrics with unstable features, we should be using:

```bash
$ RUSTFLAGS="--cfg=otel_unstable" cargo build --features "metrics"
```

Also updated the Contributing.md accordingly.

cijothomas: could we add to the changelog too?

lalitb: > could we add to the changelog too?

done

hdost: So would a good summary be:
For unstable/experimental features regardless of whether it's "experimental" in the spec or just "experimental" in our create we require the `otel_unstable` config flag.
For optional (non experimental) things we'll place them behind normal flags.

I would assume that certain things would maybe require both an `otel_unstable` config and a feature flag.

lalitb: > So would a good summary be:

Yes, rightly summarised -  `Configuration flag` for unstable features, and `feature flag` for optional features.

> Is there a reason we're choosing to use a configuration flag and not just a feature flag?

I felt using configuration flag for the unstable feature will keep our `Cargo.toml` clean, and avoid adding/enabling this flag in `Cargo.toml`  across multiple crates. Also, ease of testing, and writing examples - we can compile with or without feature with minimal changes. Also reduces the risk of inadvertently enabling this feature in Cargo.toml.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## vNext
 
+### Added
+
+- [#1410](https://github.com/open-telemetry/opentelemetry-rust/pull/1410) Add experimental synchronous gauge
```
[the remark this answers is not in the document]
lalitb: Have updated the opentelemetry crate CHANGELOG.md as below:

``
[#1410](https://github.com/open-telemetry/opentelemetry-rust/pull/1410) Add experimental synchronous gauge. This is behind the conditional flag, and can be enabled by setting the environment variable RUSTFLAGS="--cfg=otel_unstable".
``

Let me know if this is not self-explanatory, can try to elaborate further.

examples/metrics-basic/Cargo.toml
lalitb: Make sense, have changed it now to the feature flag `otel_unstable`, and updated the CONTRIBUTING.md accordingly.

cijothomas (approved):

examples/metrics-basic/Cargo.toml
cijothomas: All okay with resolving this? 
https://github.com/open-telemetry/opentelemetry-rust/issues/1411 can be used for more feedback on this.

shaun-cox (approved):

PR #1412, opened 2023-11-29
Remove old global meter provider shutdown method

jtescher:
## Changes

This change removes the old `global::shutdown_meter_provider` method which is not part of the metrics API spec, and properly documents the `SdkMeterProvider::shutdown` method which is spec compliant.

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -58,30 +58,29 @@ impl SdkMeterProvider {
     /// use opentelemetry_sdk::metrics::SdkMeterProvider;
     ///
     /// fn init_metrics() -> SdkMeterProvider {
+    ///     // Setup metric pipelines with readers + views, default has no
+    ///     // readers so nothing is exported.
     ///     let provider = SdkMeterProvider::default();
     ///
     ///     // Set provider to be used as global meter provider
     ///     let _ = global::set_meter_provider(provider.clone());
     ///
-    ///     // Setup metric pipelines with readers + views
-    ///
     ///     provider
     /// }
     ///
-    /// fn main() {
+    /// fn main() -> Result<(), Box<dyn std::error::Error>> {
     ///     let provider = init_metrics();
     ///
     ///     // create instruments + record measurements
     ///
     ///     // force all instruments to flush
-    ///     provider.force_flush().unwrap();
+    ///     provider.force_flush()?;
     ///
     ///     // record more measurements..
     ///
-    ///     // dropping provider and shutting down global provider ensure all
-    ///     // remaining metrics data are exported
-    ///     drop(provider);
-    ///     global::shutdown_meter_provider();
+    ///     // shutting down meter provider ensure all remaining metrics data
+    ///     // are exported
+    ///     provider.shutdown()?;
     /// }
     /// ```
     pub fn force_flush(&self) -> Result<()> {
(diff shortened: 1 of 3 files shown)


opentelemetry/src/global/metrics.rs
```diff
@@ -98,14 +98,6 @@ where
     *global_provider = GlobalMeterProvider::new(new_provider);
 }
 
-/// Shut down the current meter global meter provider.
```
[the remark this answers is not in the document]
jtescher: Likely yes, but they are more work as those are actually used currently so they can be done in dedicated follow up PRs

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -58,30 +58,31 @@ impl SdkMeterProvider {
...
-    ///     provider.force_flush().unwrap();
+    ///     provider.force_flush()?;
     ///
     ///     // record more measurements..
     ///
-    ///     // dropping provider and shutting down global provider ensure all
-    ///     // remaining metrics data are exported
-    ///     drop(provider);
-    ///     global::shutdown_meter_provider();
+    ///     // shutting down meter provider ensure all remaining metrics data
```
cijothomas, line 81: more accurately, we could say "shutdown ensures any cleanups required by provided is done, and also invokes shutdown on the readers.", as depending on the reader, shutdown may not cause any exports (like in manual reader case..).

cijothomas (approved):

lalitb (approved):

opentelemetry/CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ of entries, so moving from `IndexMap` to `HashMap` offers slight performance
 gains, and avoids `IndexMap` dependency. This affects `body` and `attributes` of
 `LogRecord`.
 [#1353](https://github.com/open-telemetry/opentelemetry-rust/pull/1353)
-
 - Add `TextMapCompositePropagator` [#1373](https://github.com/open-telemetry/opentelemetry-rust/pull/1373)
+- Remove `global::shutdown_meter_provider`, use `SdkMeterProvider::shutdown`
```
hdost, line 14: Nit: should this go under Removed section?

hdost (approved):

opentelemetry/CHANGELOG.md
jtescher: Yeah true

Oliboy50:
ℹ️ I opened [a discussion here](https://github.com/open-telemetry/opentelemetry-rust/discussions/1605) to find a valid replacement for when the MeterProvider is not easily accessible, since the removal of `global::shutdown_meter_provider`.

lalitb:
@Oliboy50 This issue is specific to the MetricsLayer provided by tracing-opentelemetry. And not while using the Otel Metrics API from this repo. Would you like to open a discussion within that repo. As I replied in our discussion:

> I think your workaround should work (though good to validate it), but the solution should be for the MetricsLayer layer to take either instance of Meter, or the non-mutable reference to MeterProvider as an argument, as done for [traces](https://github.com/tokio-rs/tracing-opentelemetry/blob/5e3354ec24debcfbf856bfd1eb7022459dca1e6a/src/layer.rs#L535) and [logs](https://github.com/open-telemetry/opentelemetry-rust/blob/4aa48271917c40921265ce4750ebed191d363e0d/opentelemetry-appender-tracing/src/layer.rs#L139). This will allow the MeterProvider instance to be used by the application for shutdown and flush calls.

Oliboy50:
Following your advices, I opened [a discussion there](https://github.com/tokio-rs/tracing-opentelemetry/discussions/114)

PR #1416, opened 2023-11-30
Add tests to validate spatial aggregation

cijothomas:
Added tests to validate SDK is doing "spatial aggregation" correctly. That term is no longer used explicitly in the spec, but the expected behavior is not currently implemented. Based on my experience, many implementations had/have this bug.

eg: .NET had this bug from day1 and not fixed as of today. https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/OpenTelemetry.Tests/Metrics/MetricApiTestsBase.cs#L883-L886

Related discussions on this topic:
https://github.com/open-telemetry/opentelemetry-specification/issues/2905
https://github.com/open-telemetry/opentelemetry-specification/issues/1874

The tests are ignored now (they won't pass today!), but I think its worth while to have them, while we work on Metrics SDK, to help us keep in mind some edge scenarios.

I'll dig more into fixing these. The shared discussions above only affected Observable instruments, but we are failing for Non-observable as well here, so there are likely more changes required.

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -347,11 +347,7 @@ mod tests {
 
         // find and validate the single datapoint
         let data_point = &sum.data_points[0];
-        assert_eq!(
-            data_point
-                .value,
-            300
-        );
+        assert_eq!(data_point.value, 300);
     }
 
     // "multi_thread" tokio flavor must be used else flush won't
@@ -430,10 +426,6 @@ mod tests {
         assert_eq!(sum.data_points.len(), 1);
         // find and validate the single datapoint
         let data_point = &sum.data_points[0];
-        assert_eq!(
-            data_point
-                .value,
-            30
-        );
+        assert_eq!(data_point.value, 30);
     }
 }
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -262,4 +262,163 @@ mod tests {
...
+        let sum = metric
+            .data
+            .as_any()
+            .downcast_ref::<data::Sum<u64>>()
+            .expect("Sum aggregation expected for ObservableCounter instruments by default");
+
+        // Expecting 1 time-series only, as the view drops all attributes resulting
+        // in a single time-series.
+        // This is failing today, due to lack of support for spatial aggregation.
+        assert_eq!(sum.data_points.len(), 1);
```
lalitb, line 346: Also, adding a  test to validate the value of aggregation?

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -262,4 +262,163 @@ mod tests {
...
+        let sum = metric
+            .data
+            .as_any()
+            .downcast_ref::<data::Sum<u64>>()
+            .expect("Sum aggregation expected for Counter instruments by default");
+
+        // Expecting 1 time-series only, as the view drops all attributes resulting
+        // in a single time-series.
+        // This is failing today, due to lack of support for spatial aggregation.
+        assert_eq!(sum.data_points.len(), 1);
```
lalitb, line 422: same as above - Checking for the aggregated value?

lalitb (approved):
LGTM. Thanks. We have the same problem in otel-cpp for the Observable counter. I think synchronous works fine.

cijothomas:
Merging. I'll look at fixing this for sync instruments.
 
Fixing for Observable is likely harder - haven't been able to solve in .NET, C++ yet, unfortunately :-( Hopefully Rust can fix that.)

PR #1421, opened 2023-12-01
Allow precreation of AttributeSets for metrics

### LINKED ISSUE #1387: [Feature]: Allow precreation of AttributeSets for metrics

KallDrexx:
Fixes #1387

## Changes

This PR creates a backwards incompatible change to the public metric interfaces to allow pre-created and cached attribute sets to be passed into `Counter`, `Histogram`, and `Observer` APIs instead of references to `KeyValue` slices.  

The first benefit of this change is that there is no more hidden string cloning/allocations occurring.  The `AttributeSet` that is passed in now contains an `Arc<T>` to it's internal data.  This allows each measurement used by the instrument to share its internal `KeyValue` pairs instead of cloning them for each instrument. Likewise, the caller of OpenTelemetry-rust can also pas the same attribute set into multiple instrument measurement calls without performing any cloning.

A second benefit is that if consumers have a stable set of `KeyValue` pairs for a set of counters, they can create an `AttributeSet` once and only pay the sorting + deduplication + hash calculation costs a single time instead of every measurement call.  

This should increase performance when metrics are used with common attributes, while having minimal impact when using dynamic attributes.  The one case where performance regression might be encountered are instruments that heavily utilize the `AttributeSet::retain()` function, as we now have to clone the underlying `KeyValue` pairs and create a new `AttributeSet` for the retained version now that `AttributeSet`s are internally immutable.

This is a breaking API change, however it now requires a previous call of `cntr.add(1, &[KeyValue::new("K", i)]);` to be changed to `cntr.add(1, [KeyValue::new("K", i)].into());`.  Specifically, a reference is no longer passed in and it has to be converted into an `AttributeSet`, which can be done with the `.into()` clause.

The `AttributeSet` type had to be raised to the root `opentelemetry` trait to be visible to all types.

## Metrics Stress Tests

```
main:
Number of threads: 6
Throughput: 6,459,800 iterations/sec
Throughput: 6,487,000 iterations/sec
Throughput: 6,085,600 iterations/sec
Throughput: 6,018,600 iterations/sec
```

```
pr:
Number of threads: 6
Throughput: 7,571,600 iterations/sec
Throughput: 7,786,000 iterations/sec
Throughput: 7,757,400 iterations/sec
Throughput: 7,617,400 iterations/sec
```

Likewise, a new stress test was added to show throughput with premade attribute sets
```
Number of threads: 6
Throughput: 10,310,000 iterations/sec
Throughput: 10,317,000 iterations/sec
Throughput: 10,384,000 iterations/sec
Throughput: 10,378,800 iterations/sec
Throughput: 10,299,600 iterations/sec
```

## Metric_Counter Benchmark

```
main:
Counter_Add_Sorted      time:   [727.79 ns 736.87 ns 747.75 ns]
  4 (4.00%) high mild
  7 (7.00%) high severe

Counter_Add_Unsorted    time:   [720.25 ns 727.28 ns 734.66 ns]
  3 (3.00%) high mild
  4 (4.00%) high severe
  ```

```
pr:
Counter_Add_Sorted      time:   [764.60 ns 779.08 ns 796.56 ns]
                        change: [-9.8051% -5.0974% -0.7162%] (p = 0.04 < 0.05)
  3 (3.00%) high mild
  4 (4.00%) high severe

Counter_Add_Unsorted    time:   [742.61 ns 752.64 ns 765.10 ns]
                        change: [+0.6259% +2.5097% +4.2566%] (p = 0.01 < 0.05)
  5 (5.00%) high severe

Counter_Add_Cached_Attributes
                        time:   [81.280 ns 81.979 ns 82.869 ns]
  2 (2.00%) high mild
  5 (5.00%) high severe
```

(note that the last benchmark is new)

## Metrics Benchmark

main: [metric.main.txt](https://github.com/open-telemetry/opentelemetry-rust/files/13531895/metric.main.txt)

pr: [metric.pr.txt](https://github.com/open-telemetry/opentelemetry-rust/files/13531896/metric.pr.txt)

### CODE REVIEW

File: opentelemetry/src/attributes/set.rs [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+
+impl PartialOrd for HashKeyValue {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for HashKeyValue {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.0.key.cmp(&other.0.key) {
```
cijothomas, line 42: I understand this is not introduced in this PR, but I am not able to follow the logic behind this comparison.. Wont a simple "self.0.key.cmp(&other.0.key)" be sufficient here? The question of Key's being equal should not occur as we de-dedup before sorting...

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -253,25 +253,25 @@ pub(crate) struct ResolvedMeasures<T> {
 }
 
 impl<T: Copy + 'static> SyncCounter<T> for ResolvedMeasures<T> {
-    fn add(&self, val: T, attrs: &[KeyValue]) {
+    fn add(&self, val: T, attrs: AttributeSet) {
```
TommyCpp, line 256: +1 on making adding a generic function here to replace `AttributeSet`

hdost:
> @KallDrexx can you do:
> 
> ```rust
>  pub fn add(&self, value: T, attributes: impl Into<AttributeSet>) {
>         self.0.add(value, attributes.into())
>  }
> ```
> 
> to avoid the api break?

I really think we should do this, and we can `impl From<>` if we haven't already, and I believe `into()` on self should be free as it will be detected by the compiler.

KallDrexx:
Sorry been a bit busy the last few days, will give this a try tomorrow.

KallDrexx:
The `Into<T>` suggestion works to get rid of `.into()` clauses, but making it a backwards compatible change is much more nuanced and depends what we mean.

Specifically, do we wish to retain being able to pass in a `&[KeyValue]`?    

The problem is the current `From` trait I created to support non-reference attribute set creation was 
```rust
impl<I: IntoIterator<Item = KeyValue>> From<I> for AttributeSet
where
    <I as IntoIterator>::IntoIter: DoubleEndedIterator + ExactSizeIterator,
{
    fn from(values: I) -> Self {
        AttributeSet(Arc::new(InternalAttributeSet::from(values)))
    }
}
```

This works well for `[KeyValue: N]`  type of values, as well as a `Vec<KeyValue>` or any other type of type that implements `IntoIterator<KeyValue>`.  **However**, `&[KeyValue]` does not implement `IntoIterator<KeyValue>` but instead implements `IntoIterator<&KeyValue>`.  

But you can't have `impl<I: IntoIterator<Item = KeyValue>> From<I> for AttributeSet` while also having `impl<'a, I: IntoIterator<Item = &'a KeyValue>> From<I> for AttributeSet`, because the compiler sees them as conflicting implementations.

I can fix this by removing these implementations and go for more concrete implementations, such as

```rust
impl<const N: usize> From<&[KeyValue; N]> for AttributeSet {
    fn from(value: &[KeyValue; N]) -> Self {
        AttributeSet(Arc::new(InternalAttributeSet::from(value.iter().cloned())))
    }
}

impl<const N: usize> From<[KeyValue; N]> for AttributeSet {
    fn from(value: [KeyValue; N]) -> Self {
        AttributeSet(Arc::new(InternalAttributeSet::from(value.into_iter())))
    }
}
```

However, this means you can only pass in references or owned fixed size arrays to create an attribute set.  You can't pass in a `hashset<KeyValue>`, a `HashMap<Key, Value>`, a `Vec<KeyValue>` or any other type that can be converted to an attribute set without us explicitely adding more `From<T>` implementations.

So the question becomes, do we want to support shared references and keep the API backward compatible, but enforce a limited set of iterator types that can be utilized, or make the API more broad?

KallDrexx:
Take away from SIG call:

1. Keeping hidden cloning out of the library's creation of `AttributeSet`s is important so consumers of the API can have better visibility into the costs of their calls.  Since `Key` and `Value` types can contained owned strings, it's required that we clone them if we take a shared reference, and therefore we seem to want the API to move away from taking in a shared reference to a slice.  
2. Having the `From` trait utilize generics and allow for an expansive set of collection/iterator types is important for maintaining the semantic meaning of how you create an attribute set.

I'm still open for discussion on it, but with that I'm going to go with the first code block I posted (`IntoIterator` `From<I>` implementation) for the moment.

hdost:
> Take away from SIG call:
> 
>     1. Keeping hidden cloning out of the library's creation of `AttributeSet`s is important so consumers of the API can have better visibility into the costs of their calls.  Since `Key` and `Value` types can contained owned strings, it's required that we clone them if we take a shared reference, and therefore we seem to want the API to move away from taking in a shared reference to a slice.
> 
>     2. Having the `From` trait utilize generics and allow for an expansive set of collection/iterator types is important for maintaining the semantic meaning of how you create an attribute set.
> 
> 
> I'm still open for discussion on it, but with that I'm going to go with the first code block I posted (`IntoIterator` `From<I>` implementation) for the moment.

Well unfortunately I missed the call because I was assuming we were going to meet at the later time of 9am 😓 I'll try to read over this.

opentelemetry-sdk/src/metrics/instrument.rs
KallDrexx: Made the outer `Counter` accept the `Into<AttributeSet>` but left the inner `Sync*` types alone.

hdost:
@KallDrexx you mention:
> But you can't have impl<I: IntoIterator<Item = KeyValue>> From<I> for AttributeSet while also having impl<'a, I: IntoIterator<Item = &'a KeyValue>> From<I> for AttributeSet, because the compiler sees them as conflicting implementations.

Since it sees the impls as conflicting in this case would it work for both situations, or does it give you a compiler error.

> Keeping hidden cloning out of the library's creation of AttributeSets is important ...

That makes sense in general I will have to listen to the recording to see the context I suppose.

KallDrexx:
> Since it sees the impls as conflicting in this case would it work for both situations, or does it give you a compiler error.

Unfortunately it does not.  The way the compiler detects conflicting implementations is complex but in this case the difference between `KeyValue` and `&KeyValue` inside the associated generic is not enough for it the compiler to know they don't conflict (or isn't enough for it to compile into two separate implementations).  It just sees two `From` implementations for `IntoIterator` and thinks they conflict.

Since `impl<I: IntoIterator<Item = KeyValue>> From for AttributeSet` explicitly states that it only accepts types that can be turned into iterators of owned `KeyValue`s, it does not accept iterators that return shared references of `KeyValue`s, and vice versa.  

> That makes sense in general I will have to listen to the recording to see the context I suppose.

Some of the context I mentioned in https://github.com/open-telemetry/opentelemetry-rust/issues/1387#issuecomment-1821558764.  Essentially, accepting a shared reference forces us to clone to be able to hold onto the string keys and values, but those clones aren't always necessary (e.g. it's the last `counter.add()` call for that set of attributes).  Those clones add extra allocations and overhead that's hard to debug even with a profiler.

So by removing the internal cloning and forcing the consumer of the library to clone, they become aware of when allocations happen, and may even be encouraged to better optimize their code (e.g. by storing an attribute set instead of re-using `KeyValue` iterators).

opentelemetry/src/attributes/set.rs
```diff
@@ -0,0 +1,363 @@
...
+    <I as IntoIterator>::IntoIter: DoubleEndedIterator + ExactSizeIterator,
+{
+    fn from(value: I) -> Self {
+        let iter = value.into_iter();
+        let mut seen_keys = HashSet::with_capacity(iter.len());
+        let vec = iter
+            .into_iter()
+            .rev()
+            .filter_map(|kv| {
+                if seen_keys.insert(kv.key.clone()) {
```
hdost, line 140: Not sure that we need to clone here since we own the reference.

KallDrexx:
After the last SIG meeting, we decided to keep backward compatibility by taking in a shared reference.  The rationale is that there's interest in moving the `Key` and `Value` types to take `Cow<String>` instead of an owned `String`.  If that's done then the cloning cost goes away.  So I'll make that change next

opentelemetry/src/attributes/set.rs
KallDrexx: fixed

KallDrexx:
Updated PR so iterators returning key value references can now be used, and therefore is backward compatible with existing calls in all cases *except* observables (see comment https://github.com/open-telemetry/opentelemetry-rust/pull/1421#issuecomment-1841377730 for why).  

Ready for re-review

opentelemetry/src/attributes/set.rs
```diff
@@ -0,0 +1,366 @@
...
+///
+/// Cloning of an attribute set is cheap, as all clones share a reference to the underlying
+/// attribute data.
+///
+/// This must implement [Hash], [PartialEq], and [Eq] so it may be used as
+/// HashMap keys and other de-duplication methods.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct AttributeSet(Arc<InternalAttributeSet>);
+
+impl<'a, I: IntoIterator<Item = &'a KeyValue>> From<I> for AttributeSet
```
jtescher, line 169: It's a little unfortunate to require this to always be references from a caller perspective. One alternative could be to make a new trait so this can accept both iterators over references and owned values? e.g. 

```rust
trait ToKv {
    fn to_kv(self) -> KeyValue;
}

impl ToKv for KeyValue {
    fn to_kv(self) -> KeyValue {
        self
    }
}

impl ToKv for &'_ KeyValue {
    fn to_kv(self) -> KeyValue {
        self.clone()
    }
}

impl<'a, KV, I> From<I> for AttributeSet
where
    KV: ToKv,
    I: IntoIterator<Item = KV>,
    <I as IntoIterator>::IntoIter: DoubleEndedIterator + ExactSizeIterator,
{
    fn from(values: I) -> Self {
        AttributeSet(Arc::new(InternalAttributeSet::from(
            values.into_iter().map(ToKv::to_kv),
        )))
    }
}
```

That would allow for a greater variety of arguments 

```rust
let _ = AttributeSet::from(vec![KeyValue::new("owned", true)]);
let _ = AttributeSet::from(&[KeyValue::new("owned", false)]);
```

examples/metrics-basic/src/main.rs
```diff
@@ -1,3 +1,4 @@
+use opentelemetry::attributes::AttributeSet;
```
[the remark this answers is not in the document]
KallDrexx: Moved `AttributeSet` reference to `opentelemetry` root

jtescher (approved):
This looks good to me, could update the examples to remove the extra reference when constructing the attribute sets to clean them up, but not strictly necessary.

stress/src/metrics_cached_attrs.rs
```diff
@@ -0,0 +1,53 @@
...
+fn main() {
+    throughput::test_throughput(test_counter);
+}
+
+fn test_counter() {
+    let mut rng = SmallRng::from_entropy();
+    let len = ATTRIBUTE_SETS.len();
+    let index = rng.gen_range(0..len);
+
+    COUNTER.add(1, ATTRIBUTE_SETS[index].clone());
```
lalitb, line 52: Can you please confirm if the change in this PR is backward incompatible (breaking) change for existing instrumentation (as mentioned in the PR description)? As there are no changes done in the stress/metrics.rs, and it seems to be building correctly - https://github.com/open-telemetry/opentelemetry-rust/blob/27d338dd954ef7681242ead47a8a61c964dcc428/stress/src/metrics.rs#L38-L45

KallDrexx: With Julian's `ToKv` trait suggestion this works in a backwards compatible way, since a `&[KeyValue]` implements the `ToKv` trait, it automatically gets coerced into an interator of `ToKv` values.

lalitb: Thanks for clarifying. So, the below code will still work as before, but there won't be any perf improvement in this case as a new instance of `(Internal)AttributeSet` gets created in each call to counter::add(). 

```rust
    // Create a vector of KeyValue pairs.
    let key_values = vec![
    ];

    // Record measurements using the Counter instrument.
    counter.add(10, key_values.as_ref());
    counter.add(10, key_values.as_ref());
```

While supported, this is not the recommended approach and is good to be documented (probably as separate PR).

KallDrexx: Actually, that specific example is backward compatible, but observables **are not** backward compatible.  Those must take an `AttributeSet` in because they rely on boxed traits instead of generics, and thus cannot be easily coerced without a bunch of overhead.

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -131,7 +132,7 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 
             span.add_event("Sub span event", vec![]);
 
-            histogram.record(1.3, &[]);
+            histogram.record(1.3, AttributeSet::default());
```
cijothomas, line 135: not in this PR, but just realized that this is somewhat awkward having to pass empty/default when no dimensions!

opentelemetry-prometheus/src/lib.rs
```diff
@@ -31,8 +32,10 @@
 //!     .with_description("Records values")
 //!     .init();
 //!
-//! counter.add(100, &[KeyValue::new("key", "value")]);
-//! histogram.record(100, &[KeyValue::new("key", "value")]);
+//! let attributes = AttributeSet::from(&[KeyValue::new("key", "value")]);
+//!
+//! counter.add(100, attributes.clone());
```
cijothomas, line 37: i am not sure how common is it to have 2 instruments used with same attributes.. maybe better to stick with existing examples only.

(Also counter's functionality is already provided by histogram, so this example of showing both could be confusing! But that is unrelated to this PR)

opentelemetry-sdk/src/testing/metrics/in_memory_exporter.rs
```diff
@@ -46,7 +47,8 @@ use std::sync::{Arc, Mutex};
 ///  // Create and record metrics using the MeterProvider
 ///  let meter = meter_provider.meter(std::borrow::Cow::Borrowed("example"));
 ///  let counter = meter.u64_counter("my_counter").init();
-///  counter.add(1, &[KeyValue::new("key", "value")]);
+///  let attributes = AttributeSet::from(&[KeyValue::new("key", "value")]);
```
cijothomas, line 50: i am not sure if we should change all examples to use AttributeSet.. maybe show it only where it gives some perf benefits.

opentelemetry/src/attributes/set.rs
cijothomas: @KallDrexx Gentle reminder for this! Not a blocker for this PR.

opentelemetry-prometheus/tests/integration_test.rs
```diff
@@ -44,27 +45,29 @@ fn prometheus_exporter_integration() {
...
                     .with_description("a simple counter")
                     .with_unit(Unit::new("ms"))
                     .init();
-                counter.add(5.0, &attrs);
-                counter.add(10.3, &attrs);
-                counter.add(9.0, &attrs);
-                let attrs2 = vec![
+                counter.add(5.0, attrs.clone());
+                counter.add(10.3, attrs.clone());
+                counter.add(9.0, attrs.clone());
```
lalitb, line 62: nit - do we need to clone() here ?

opentelemetry-prometheus/tests/integration_test.rs
```diff
@@ -73,59 +76,62 @@ fn prometheus_exporter_integration() {
...
                     .with_description("a simple counter without a total suffix")
                     .with_unit(Unit::new("ms"))
                     .init();
-                counter.add(5.0, &attrs);
-                counter.add(10.3, &attrs);
-                counter.add(9.0, &attrs);
-                let attrs2 = vec![
+                counter.add(5.0, attrs.clone());
+                counter.add(10.3, attrs.clone());
+                counter.add(9.0, attrs.clone());
```
lalitb, line 93: nit - same as earlier, we can remove clone() here?

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -96,7 +91,7 @@ impl<T: Number<T>> AggregateBuilder<T> {
         let filter = self.filter.as_ref().map(Arc::clone);
         move |n, mut attrs: AttributeSet| {
             if let Some(filter) = &filter {
-                attrs.retain(filter.as_ref());
+                attrs = attrs.clone_with(filter.as_ref());
```
lalitb, line 94: This will clone every HashKeyValue that passes the filter. Unfortunately, I don't think this would be easy to avoid. Not blocking this PR, maybe something we can look in the future.

lalitb (approved):
Thanks for the PR. Nicely done. Please update the ChangeLog to reflect breaking changes for Observables.

opentelemetry-sdk/src/metrics/internal/aggregate.rs
KallDrexx: Yeah I wasn't happy about this part of the change, but the only way precreated attribute sets really worked is by making them immutable.

opentelemetry-otlp/examples/basic-otlp/src/main.rs
KallDrexx: This is kind of an artifact of the original state of the PR being backward incompatible.  In the current state of the code `&[]` now works again.

opentelemetry-prometheus/src/lib.rs
KallDrexx: Since this example uses a `histogram`, and histograms couldn't be updated to take a `Into<AttributeSet>` either, the precreated attribute set is required here for at least that one.

opentelemetry-sdk/src/testing/metrics/in_memory_exporter.rs
KallDrexx: These examples were changed prior to a backward compatible solution being found, thus I had to update all the doctests to use precreated attributes for CI to pass.

opentelemetry/CHANGELOG.md
```diff
@@ -17,6 +17,16 @@ gains, and avoids `IndexMap` dependency. This affects `body` and `attributes` of
 `LogRecord`.
 [#1353](https://github.com/open-telemetry/opentelemetry-rust/pull/1353)
 - Add `TextMapCompositePropagator` [#1373](https://github.com/open-telemetry/opentelemetry-rust/pull/1373)
+- `Counters`, `UpDownCounters`, `Observables`, and `Histograms` now take an
+  `Into<AttributeSet>` as the parameter type for recording metric values. This
+  allows passing in a precreated `AttributeSet` for better performance when the
+  same set of attributes are used across instruments. This is backward
+  compatible with previous calls passing in `&[KeyValue]`.
+  [#1421](https://github.com/open-telemetry/opentelemetry-rust/pull/1421)
+- `Histograms` and `Gauges` no longer accept `&[KeyValue]` parameters for
```
cijothomas, line 26: Histograms?

opentelemetry/CHANGELOG.md
```diff
@@ -17,6 +17,16 @@ gains, and avoids `IndexMap` dependency. This affects `body` and `attributes` of
...
 [#1353](https://github.com/open-telemetry/opentelemetry-rust/pull/1353)
 - Add `TextMapCompositePropagator` [#1373](https://github.com/open-telemetry/opentelemetry-rust/pull/1373)
+- `Counters`, `UpDownCounters`, `Gauges`, and `Histograms` now take an
+  `Into<AttributeSet>` as the parameter type for recording metric values. This
+  allows passing in a precreated `AttributeSet` for better performance when the
+  same set of attributes are used across instruments. This is backward
+  compatible with previous calls passing in `&[KeyValue]`.
+  [#1421](https://github.com/open-telemetry/opentelemetry-rust/pull/1421)
+- Observable instruments no longer accept `&[KeyValue]` parameters for
+  `observe()` or `record()` calls, and only accept a precreated `AttributeSet`
```
cijothomas, line 27:
```suggestion
  `observe()` calls, and only accept a precreated `AttributeSet`
```

cijothomas: All observables all "observeABC()" only, there is no 'record() calls".

cijothomas:
This caused some unintended side-effects, opened a separate issue to discussion options : https://github.com/open-telemetry/opentelemetry-rust/issues/1508

PR #1424, opened 2023-12-04
Fix metric_counter bench

Rockin2:
Fixes #

metric_counter bench doesn't work because of metrics feature was not enables.

## Changes

### DIFF
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -66,6 +66,7 @@ harness = false
 [[bench]]
 name = "metric_counter"
 harness = false
+required-features = ["metrics"]
 
 [[bench]]
 name = "attribute_set"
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/Cargo.toml
```diff
@@ -66,6 +66,7 @@ harness = false
 [[bench]]
 name = "metric_counter"
 harness = false
+required-features = ["metrics"]
```
cijothomas, line 69: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/metric_counter.rs#L10 Could you fix this comment too?

cijothomas (approved):

opentelemetry-sdk/Cargo.toml
Rockin2: Done. Be aware you must activated the feature with --feature event if we defined it in the Cargo.toml. However, instead of a compilation error cargo gives a nice reminder to activate the feature. It fixes also the use case when we run all bench with cargo bench. :)

Issue #1413, opened 2023-11-29
Should this repo maintain examples showing tracing-opentelemetry?


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/tracing-http-propagator shows using Tracing API from this Repo to demonstrate how to do tracing + propagation.

https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/tracing-grpc This is using `tracing-opentelemetry` crate, achieving similar end-results. Should we remove those examples from this repo, and instead ask them to be maintained in https://github.com/tokio-rs/tracing-opentelemetry/tree/v0.1.x/examples? Since the split to contrib repo, https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/tracing-grpc is the only place where tracing-opentelemetry is used. We do link to the crate in many docs, and that is necessary, but maintaining examples using tracing-opentelemetry may not be desirable.

Related : [Discussion about OTel Tracing API vs Tokio Tracing ](https://github.com/open-telemetry/opentelemetry-rust/discussions/1032).

lalitb:
I feel it is good to have both `tracing-http` and `tracing-grpc` examples, as the latter also shows how to use the propagator with tonic metadata. How about removing `tracing-opentelemetry` dependency from this `tracing-grpc` example?

cijothomas:
Yes, it makes sense to maintain tracing-grpc. I was thinking of removing the tracing-opentelemetry part only and instead follow the tracing-http style of showing OTel Tracing API. If a user ends up with tracing-opentelemetry separately, it'll likely result in duplicate?, so a warning can be mentioned here and in tracing-opentelemetry about that.

TommyCpp:
Agree we should move those examples to `tracing-opentelemetry`. Different from most of example which uses the master branch. Examples using `tracing-opentelemtry` uses last released version. So probably better to move it to `tracing-opentelemetry` repo

PR #1426, opened 2023-12-05
Remove `tracing-opentelemetry` dependency from `tracing-grpc` example

lalitb:
Fixes (partially) #1413 

## Changes

A minimalistic version of `tracing-grpc` example without `opentelemetry-jaeger` and `tracing-*` dependencies

### DIFF
--- a/examples/tracing-grpc/src/client.rs
+++ b/examples/tracing-grpc/src/client.rs
@@ -1,7 +1,9 @@
 use hello_world::greeter_client::GreeterClient;
 use hello_world::HelloRequest;
 use opentelemetry::{global, propagation::Injector};
-use opentelemetry_sdk::{propagation::TraceContextPropagator, trace::TracerProvider};
+use opentelemetry_sdk::{
+    propagation::TraceContextPropagator, runtime::Tokio, trace::TracerProvider,
+};
 use opentelemetry_stdout::SpanExporter;
 
 use opentelemetry::{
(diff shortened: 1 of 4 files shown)


examples/tracing-grpc/src/client.rs
```diff
@@ -1,10 +1,23 @@
...
+use opentelemetry::{
+    trace::{SpanKind, TraceContextExt, Tracer},
+    Context, KeyValue,
+};
+
+fn init_tracer() {
+    global::set_text_map_propagator(TraceContextPropagator::new());
+    // Install stdout exporter pipeline to be able to retrieve the collected spans.
+    let provider = TracerProvider::builder()
+        .with_simple_exporter(SpanExporter::default())
```
TommyCpp, line 16: Since we have tokio here can we use batch span processor here?

lalitb: Good point. Updated to use the batch span processor with Tokio runtime.

shaun-cox (approved):

examples/tracing-grpc/src/client.rs
```diff
@@ -24,45 +39,46 @@ pub mod hello_world {
...
+        Ok(_res) => "OK".to_string(),
+        Err(status) => {
+            // Access the status code
+            let status_code = status.code();
+            status_code.to_string()
+        }
+    };
+    cx.span().add_event(
+        "Got response!".to_string(),
+        vec![KeyValue::new("status", status)],
```
cijothomas, line 72: I think we should be showing the recommended approach of storing this to Span status itself, if failed.

cijothomas (approved):
Looks good!

left couple of nits, non-blocking!

PR #1431, opened 2023-12-08
Bench for logs

Rockin2:
## Changes

Add benches for logs.

### DIFF
--- a/opentelemetry-sdk/benches/log.rs
+++ b/opentelemetry-sdk/benches/log.rs
@@ -1,4 +1,50 @@
+use async_trait::async_trait;
 use criterion::{criterion_group, criterion_main, Criterion};
+
+use opentelemetry::logs::{LogResult, Logger};
+use opentelemetry_sdk::export::logs::{LogData, LogExporter};
+use opentelemetry_sdk::logs::LoggerProvider;
+
+#[derive(Debug)]
+struct VoidExporter;
+
+#[async_trait]
+impl LogExporter for VoidExporter {
+    async fn export(&mut self, _batch: Vec<LogData>) -> LogResult<()> {
+        LogResult::Ok(())
+    }
+}
+
+fn log_benchmark_group<F: Fn(&dyn Logger)>(c: &mut Criterion, name: &str, f: F) {
+    let mut group = c.benchmark_group(name);
+
+    group.bench_function("no-context", |b| {
+        let provider = LoggerProvider::builder()
+            .with_simple_exporter(VoidExporter)
+            .build();
+    });
+
+    //group.bench_function("always-sample", |b| {
+    //let provider = sdktrace::TracerProvider::builder()
+    //.with_config(sdktrace::config().with_sampler(sdktrace::Sampler::AlwaysOn))
+    //.with_simple_exporter(VoidExporter)
+    //.build();
+    //let always_sample = provider.tracer("always-sample");
+
+    //b.iter(|| f(&always_sample));
+    //});
+
+    //group.bench_function("never-sample", |b| {
+    //let provider = sdktrace::TracerProvider::builder()
+    //.with_config(sdktrace::config().with_sampler(sdktrace::Sampler::AlwaysOff))
+    //.with_simple_exporter(VoidExporter)
+    //.build();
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/benches/log.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    group.finish();
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    log_benchmark_group(c, "simple-log", |logger| {
+        logger.emit(LogRecord::builder().with_body("simple log".into()).build())
+    });
+
+    log_benchmark_group(c, "long-log", |logger| {
```
cijothomas, line 63: not sure if we need to distinguish between body being short vs long. SDK is not doing any processing with Body content anyway, so this is less valuable.

cijothomas (approved):
Looks good to get started with benchmarking logging signal! Left couple nits, totally okay to consider for a future improvement as well.

opentelemetry-sdk/benches/log.rs
```diff
@@ -0,0 +1,119 @@
...
+                .with_attribute("event.id", 20)
+                .with_attribute("user.name", "otel")
+                .with_attribute("user.email", "otel@opentelemetry.io")
+                .build(),
+        )
+    });
+
+    log_benchmark_group(c, "full-log-with-9-attributes", |logger| {
+        logger.emit(
+            LogRecord::builder()
```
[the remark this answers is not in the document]
Rockin2: I could do that in another PR. :)

opentelemetry-sdk/benches/log.rs
Rockin2: Do you want me to delete this bench?

lalitb: I believe no harm in keeping. SDK does internally copy the argument passed in `with_body`, so gives a good idea of the cost for copy operation.

opentelemetry-sdk/benches/log.rs
```diff
@@ -0,0 +1,119 @@
...
+                .with_severity_text(Severity::Warn.name())
+                .with_attribute("name", "my-event-name")
+                .with_attribute("event.id", 20)
+                .with_attribute("user.name", "otel")
+                .with_attribute("user.email", "otel@opentelemetry.io")
+                .with_attribute("log.source.file.name", "log.rs")
+                .with_attribute("log.source.file.path", "opentelemetry_sdk/benches/log.rs")
+                .with_attribute("log.source.file.line", 96)
+                .with_attribute("log.module.path", "opentelemetry_sdk::benches::log")
+                .with_attribute("log.target", "opentelemetry_sdk::benches::log")
```
lalitb, line 112: Good to have the test for all the attribute types (or at-least - bool, int, float, string, byte) - https://github.com/open-telemetry/opentelemetry-rust/blob/443f68b1b8cd9016aec7f21ce913884f473c848c/opentelemetry/src/logs/record.rs#L79-L95)

opentelemetry-sdk/benches/log.rs
```diff
@@ -0,0 +1,119 @@
...
+            LogRecord::builder()
+                .with_body("full log".into())
+                .with_timestamp(now)
+                .with_observed_timestamp(now)
+                .with_severity_number(Severity::Warn)
+                .with_severity_text(Severity::Warn.name())
+                .with_attribute("name", "my-event-name")
+                .with_attribute("event.id", 20)
+                .with_attribute("user.name", "otel")
+                .with_attribute("user.email", "otel@opentelemetry.io")
```
lalitb, line 91: Can we also add test for `with_attributes` ?

opentelemetry-sdk/benches/log.rs
Rockin2: Ok. I can have a test for each attributes type

opentelemetry-sdk/benches/log.rs
Rockin2: The thing is with this test I don't know if I will create the Voc outside of the bench or inside? What is your opinion?

lalitb: Vec should be outside of the bench.

PR #1435, opened 2023-12-10
feat: make trace state can be propagated  for dropped sampler

### LINKED ISSUE #1399: [Bug]: trace state for Dropped sampler decision is not propagated
### What happened?

it seems that if a sampler makes the decision not to sample, even if in the decision trace_state is passed, it wont be propagated further
this is useful for us since we want to propagate some data even if spans are not sampled
Im not sure its a bug, but i know it does work in other language sdks (node, python and java are the ones we tried)

this is where the decision happens:
https://github.com/open-telemetry/opentelemetry-rust/blob/075d26cf23ad7c7541aca8272c863d65220248bd/opentelemetry-sdk/src/trace/tracer.rs#L94C1-L94C1

### API Version

0.21.0

### SDK Version

0.21.1

_No response_

### Relevant log output

_No response_

zengxilong:
Fixes #1399 

## Changes

Added a return value (SamplingDecision) for func `make_sampling_decision`, and keep the `TraceState` value when sampling decision is `Drop`.

### DIFF
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -265,6 +265,7 @@ mod tests {
                 KeyValue::new("extra_attr_key", "extra_attr_value"),
                 KeyValue::new("record_only_key", "record_only_value")
             ]
-        )
+        );
+        assert_eq!(span.span_context().trace_state().get("foo"), Some("bar"));
     }
 }
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -44,7 +44,9 @@ mod tests {
         trace::span_limit::{DEFAULT_MAX_EVENT_PER_SPAN, DEFAULT_MAX_LINKS_PER_SPAN},
     };
     use opentelemetry::testing::trace::TestSpan;
-    use opentelemetry::trace::{TraceContextExt, TraceState};
+    use opentelemetry::trace::{
+        SamplingDecision, SamplingResult, SpanKind, TraceContextExt, TraceState,
+    };
     use opentelemetry::{
         trace::{
             Event, Link, Span, SpanBuilder, SpanContext, SpanId, TraceFlags, TraceId, Tracer,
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -176,4 +180,92 @@ mod tests {
...
+            SpanBuilder::from_name("span")
+                .with_attributes(vec![KeyValue::new("extra_attr_key", "extra_attr_value")]),
+            &parent_context,
+        );
+        assert!(!span.span_context().trace_flags().is_sampled());
+        assert_eq!(
+            span.exported_data().unwrap().attributes,
+            vec![
+                KeyValue::new("extra_attr_key", "extra_attr_value"),
+                KeyValue::new("record_only_key", "record_only_value")
```
cijothomas, line 266: this is not really testing anything related to trace-state, right?

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -176,4 +180,92 @@ mod tests {
...
+        ) -> SamplingResult {
+            let trace_state = parent_context
+                .unwrap()
+                .span()
+                .span_context()
+                .trace_state()
+                .clone();
+            SamplingResult {
+                decision: SamplingDecision::RecordOnly,
+                attributes: vec![KeyValue::new("record_only_key", "record_only_value")],
```
cijothomas, line 231: dont think we need this attribute here, unless this is specifically testing if attributes from SamplingResult are added.

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -95,27 +95,43 @@ impl Tracer {
...
         match sampling_result {
             SamplingResult {
                 decision: SamplingDecision::Drop,
+                trace_state,
                 ..
-            } => None,
+            } => (
+                SamplingDecision::Drop,
+                TraceFlags::default(),
+                Vec::new(),
```
lalitb, line 107: The SDK should do minimum processing in case the spans are dropped. Do you think we can replace it with an empty slice, which can be statically allocated?

cijothomas (approved):
Looks okay, and is better spec compliant as well. 
Couple of comments are open, please address them before merging.
Many thanks for the test coverage added!

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -172,19 +188,20 @@ impl opentelemetry::trace::Tracer for Tracer {
...
+                self.make_sampling_decision(
+                    parent_cx,
+                    trace_id,
+                    &builder.name,
+                    &span_kind,
+                    builder.attributes.as_ref().unwrap_or(&Vec::new()),
+                    builder.links.as_deref().unwrap_or(&[]),
+                    provider.config(),
+                )
+            };
```
TommyCpp, line 204: I don't see we use `trace_state` in this part. Maybe we can just extract it before we process sampling result or make a sampling decision?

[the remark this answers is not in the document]
zengxilong: I agree, other language implementations do the same.

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -196,97 +213,95 @@ impl opentelemetry::trace::Tracer for Tracer {
...
+                        end_time,
+                        attributes: attribute_options,
+                        dropped_attributes_count,
+                        events: span_events,
+                        links: span_links,
+                        status,
+                    }),
+                    self.clone(),
+                    span_limits,
+                )
```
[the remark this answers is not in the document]
zengxilong: 👌

opentelemetry-sdk/src/trace/tracer.rs
zengxilong: According to @TommyCpp , `process_sampling_result ` can be deprecated

TommyCpp (approved):
Thanks for the change

lalitb (approved):
Looks Good. Thanks for the fix.

PR #1440, opened 2023-12-12
Add await to async trace documentation

jbr:
Fixes #

## Changes

Add an `await` to the trace module doc section that describes WithContext, as the returned future is #[must_use] and does nothing without being awaited. It might not be apparent to readers that [WithContext&lt;impl Future&gt; is Future](https://docs.rs/opentelemetry/latest/opentelemetry/trace/struct.WithContext.html#impl-Future-for-WithContext%3CT%3E)


lalitb (approved):
Good catch.

Issue #1445, opened 2023-12-16
[Feature]: opentelemetry-appender-tracing should use semantic conventions


bouk:
### Describe the solution you'd like:

There's an `experimental_metadata_attributes` feature on opentelemetry-appender-tracing that adds some attributes related to the code location: https://github.com/open-telemetry/opentelemetry-rust/blob/4e80e3a2ebf96a166917b169f7d27e44e83cd3da/opentelemetry-appender-tracing/src/layer.rs#L53-L90

These don't use the [semantic convention names](https://opentelemetry.io/docs/specs/semconv/general/attributes/#source-code-attributes), which they should so the field naming is consistent across tooling.

@Rockin2

Rockin2:
Great that we have a convention for those field like ECS! :) Can you assign me this issue please? :) I will do that tomorrow.

cijothomas:
Need to keep in mind that the conventions are still experimental, so we need to do this under feature flags as well.

PR #1448, opened 2023-12-18
Use good OpenTelemetry conventions for attributes in appender-tracing

Rockin2:
Also, only removed duplicates attributes from the log crate. I think it is a good compromise for a little more performance.

Fixes #
https://github.com/open-telemetry/opentelemetry-rust/issues/1445

## Changes

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -17,24 +17,12 @@ struct EventVisitor {
     log_record_body: Option<AnyValue>,
 }
 
+/// Logs from the log crate have duplicated attributes that we removed here.
 #[cfg(feature = "experimental_metadata_attributes")]
-fn is_metadata(field: &'static str) -> bool {
+fn is_duplicated_metadata(field: &'static str) -> bool {
     field
         .strip_prefix("log.")
-        .map(|remainder| {
-            matches!(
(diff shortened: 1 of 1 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -65,21 +53,21 @@ impl EventVisitor {
...
+                .push(("code.namespace".into(), module_path.to_owned().into()));
         }
 
         if let Some(filepath) = meta.file() {
             self.log_record_attributes
-                .push(("log.source.file.path".into(), filepath.to_owned().into()));
+                .push(("code.filepath".into(), filepath.to_owned().into()));
             self.log_record_attributes.push((
-                "log.source.file.name".into(),
+                "code.filename".into(),
```
lalitb, line 63: Do we need `code.filename` as it is not part of semantic convention, and `code.filepath` contains the fully qualified path with filename ?

Rockin2: It is easier to do an exact match with only the name of the file. Remember also the path can change between compilations. Also, search engines like ElasticSearch has a column type called keyword for exact match that is more efficient than text search.

lalitb (approved):
LGTM. Thanks for the changes.

hdost (approved):

PR #1452, opened 2023-12-22
Fix max_scale validation of exponential histogram configuration

inahga:
`metrics::Aggregation::validate()` has a bug that limits `max_scale` of a `Base2ExponentialHistogram` to the interval `[10, 20]` instead of the expected `[-10, 20]`.

## Changes

Fix the double negation so that the function validates correctly. I've also added some tests that exercise the entire validation function.


jtescher (approved):
Good catch, thanks @inahga

Issue #1453, opened 2023-12-24
[Bug]: Wrong StatusCode ordinal in stdout exporter


j-white:
### What happened?

An Error is recorded as status=1, and Ok as status=2.:
https://github.com/open-telemetry/opentelemetry-rust/blob/v0.21.1/opentelemetry-stdout/src/trace/transform.rs#L214

These should be flipped to match the standard:
https://github.com/open-telemetry/opentelemetry-proto/blob/v1.0.0/opentelemetry/proto/trace/v1/trace.proto#L269

### API Version

1.0

### SDK Version

0.21.1

### What Exporters are you seeing the problem on?

N/A

PR #1454, opened 2023-12-24
Fix StatusCode in stdout exporter (#1453)

j-white:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1453

## Changes

Swap the ordinals to match the spec.

### DIFF
--- a/opentelemetry-stdout/src/trace/transform.rs
+++ b/opentelemetry-stdout/src/trace/transform.rs
@@ -217,11 +217,11 @@ impl From<opentelemetry::trace::Status> for Status {
             },
             opentelemetry::trace::Status::Error { description } => Status {
                 message: Some(description),
-                code: 1,
+                code: 2,
             },
             opentelemetry::trace::Status::Ok => Status {
                 message: None,
-                code: 2,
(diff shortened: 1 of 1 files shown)


opentelemetry-stdout/src/trace/transform.rs
```diff
@@ -217,11 +217,11 @@ impl From<opentelemetry::trace::Status> for Status {
             },
             opentelemetry::trace::Status::Error { description } => Status {
                 message: Some(description),
-                code: 1,
+                code: 2,
```
lalitb, line 220: should we use the constants for status codes to enhance readability and maintainability. Something like:
```rust
const STATUS_CODE_UNSET: u32 = 0;
const STATUS_CODE_OK: u32 = 1;
const STATUS_CODE_ERROR: u32 = 2;
```

These constants can be used both in `is_zero` and `From`.

lalitb (approved):
LGTM, with nit suggestion.

jtescher (approved):
Thanks @j-white, think it's ok to merge as is and improve in follow up as well

Issue #1209, opened 2023-08-17
Perf bottleneck when Loggers upgrade weak ref to LoggerProvider


cijothomas:
Noticed while stress testing Logs, but should be applicable to Tracing as well, as they follow the same pattern, where Logger (Tracer), holds a `Weak` ref to LoggerProvider (TracerProvider), and in hoth path (logger.EmitLog, span.Record), the weak reference to provider is upgraded to Arc, to obtain things like processor list, resource etc. from the provider. This weak -> Arc upgrade seems to be the bottleneck.

Here's how I tested:
1. Ran the [stress test for logs](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs), in my box, and it shows ~7M/sec throughput.
2. Modified the NoOpLogProcessor in the above test to [ return `false`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs#L61), which should make the throughput skyrocket as we don't have to do anything like creating LogRecord or invoke processors etc. This did increase the throughput to ~35M/sec, but I was expecting 200+M/sec, as I get similar throughout when using tracing+no-op-tracing-subscriber, without any OpenTelemetry component.
3. After some manual explorations, found that the upgrading of [weak to Arc](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_emitter.rs#L198) in the hot path of checking if [event is enabled](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_emitter.rs#L234) is the bottleneck.
4. I refactored the code so that Logger is holding an Arc instead of Weak ([draft code](https://github.com/open-telemetry/opentelemetry-rust/compare/main...cijothomas:cijothomas/logger_arc_to_provider?expand=1) ), so there is no longer the need for upgrading to Arc in hotpath. With this change, the throughput shot to ~250M/sec (in line with what was seen with using tracing+no-op subscriber alone).

The weak ref to provider was introduced for Tracing (and then followed for Logging), in this [PR](https://github.com/open-telemetry/opentelemetry-rust/pull/229). Using Arc does mean that the Provider won't be dropped and shutdown won't be signaled until the last Logger is dropped, but that seems okay to me. In the case of Logging, Logger's are held only by the appenders (tokio-tracing subscribers etc), which, when dropped, will drop their Logger as well, allowing provider to be dropped....
[...]

PR #1446, opened 2023-12-18
Use Arc to store reference to LoggerProvider from Logger

bIgBV:
Fixes #
Design discussion issue (if applicable) #1209 

## Changes
Switch from using a `Weak` reference to `LoggerProvider` from within the `Logger` to using an `Arc`.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1455, opened 2023-12-26
Change to owned LoggerProvider

hdost:
The interface to an item shouldn't take an inner value since it's considered inner, this also allows for further optimizations in the future as it hides the complexity from the user.

Relates #1209

## Changes

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -13,10 +13,7 @@ use opentelemetry::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 
-use std::{
-    borrow::Cow,
-    sync::{Arc, Weak},
-};
+use std::{borrow::Cow, sync::Arc};
 
 #[derive(Debug, Clone)]
 /// Creator for `Logger` instances.
@@ -55,16 +52,11 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
     }
 
     fn library_logger(&self, library: Arc<InstrumentationLibrary>) -> Self::Logger {
-        Logger::new(library, Arc::downgrade(&self.inner))
+        Logger::new(library, self.clone())
     }
 }
 
 impl LoggerProvider {
-    /// Build a new logger provider.
-    pub(crate) fn new(inner: Arc<LoggerProviderInner>) -> Self {
-        LoggerProvider { inner }
-    }
-
     /// Create a new `LoggerProvider` builder.
     pub fn builder() -> Builder {
         Builder::default()
@@ -91,7 +83,7 @@ impl LoggerProvider {
     /// Shuts down this `LoggerProvider`, panicking on failure.
     pub fn shutdown(&mut self) -> Vec<LogResult<()>> {
         self.try_shutdown()
-            .expect("canont shutdown LoggerProvider when child Loggers are still active")
+            .expect("cannot shutdown LoggerProvider when child Loggers are still active")
     }
 
     /// Attempts to shutdown this `LoggerProvider`, succeeding only when
@@ -108,7 +100,7 @@ impl LoggerProvider {
 }
 
 #[derive(Debug)]
-pub(crate) struct LoggerProviderInner {
+struct LoggerProviderInner {
     processors: Vec<Box<dyn LogProcessor>>,
     config: Config,
 }
@@ -179,13 +171,13 @@ impl Builder {
 /// [`LogRecord`]: opentelemetry::logs::LogRecord
 pub struct Logger {
     instrumentation_lib: Arc<InstrumentationLibrary>,
-    provider: Weak<LoggerProviderInner>,
+    provider: LoggerProvider,
 }
 
 impl Logger {
     pub(crate) fn new(
         instrumentation_lib: Arc<InstrumentationLibrary>,
-        provider: Weak<LoggerProviderInner>,
+        provider: LoggerProvider,
     ) -> Self {
         Logger {
             instrumentation_lib,
@@ -194,8 +186,8 @@ impl Logger {
     }
 
     /// LoggerProvider associated with this logger.
-    pub fn provider(&self) -> Option<LoggerProvider> {
-        self.provider.upgrade().map(LoggerProvider::new)
+    pub fn provider(&self) -> &LoggerProvider {
+        &self.provider
     }
 
     /// Instrumentation library information of this logger.
@@ -207,16 +199,14 @@ impl Logger {
 impl opentelemetry::logs::Logger for Logger {
     /// Emit a `LogRecord`.
     fn emit(&self, record: LogRecord) {
-        let provider = match self.provider() {
-            Some(provider) => provider,
-            None => return,
-        };
+        let provider = self.provider();
+        let config = provider.config();
+        let processors = provider.log_processors();
         let trace_context = Context::map_current(|cx| {
             cx.has_active_span()
                 .then(|| TraceContext::from(cx.span().span_context()))
         });
-        let config = provider.config();
-        for processor in provider.log_processors() {
+        for p in processors {
             let mut record = record.clone();
             if let Some(ref trace_context) = trace_context {
                 record.trace_context = Some(trace_context.clone())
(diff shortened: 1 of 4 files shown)


lalitb:
Trying to understand more here - so with this PR, each logger holds its own complete instance of LoggerProvider, which would be different from the other loggers, and also different from the global LoggerProvider. And will calling shutdown on the global loggerProvider ensure that the logs through these loggers are not exported ?

Also, would there be increased memory usage if many Logger instances are created, as each carries a full copy of the LoggerProvider and its internal state.

hdost:
> Trying to understand more here - so with this PR, each logger holds its own complete instance of LoggerProvider, which would be different from the other loggers, and also different from the global LoggerProvider. And will calling shutdown on the global loggerProvider ensure that the logs through these loggers are not exported ?

The inner is still the same, so it should, I'll double check that there's a test. 

> Also, would there be increased memory usage if many Logger instances are created, as each carries a full copy of the LoggerProvider and its internal state.

It won't be a full copy, the inner is an Arc, so the outer would be an additional pointer memory wise. 

> 
> 
> Also, please change the PR title to reflect LoggerProvider instead of TracerProvider :)

Done 😅

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -55,16 +52,11 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
     }
 
     fn library_logger(&self, library: Arc<InstrumentationLibrary>) -> Self::Logger {
-        Logger::new(library, Arc::downgrade(&self.inner))
+        Logger::new(library, self.to_owned())
```
TommyCpp, line 55: nit: `self.clone()` probably cleaner IMO

TommyCpp (approved):

lalitb:
> It won't be a full copy, the inner is an Arc, so the outer would be an additional pointer memory wise.

Ah good point. I didn't realize that inner is Arc type, and since there is no `owned` trait implementation, the `to_owned()` will actually do `clone`, an so all outer loggerprovider instances would share the single `inner` instance.

lalitb:
> And will calling shutdown on the global loggerProvider ensure that the logs through these loggers are not exported ?

It would be good to validate this use-case. It seems [shutdown_logger_provider](https://github.com/open-telemetry/opentelemetry-rust/blob/f9b08c10809e3b39497e21cccaa0d0c49b0216dc/opentelemetry/src/global/logs.rs#L134) will invoke shutdown on the LogProcessors when they are are [dropped](https://github.com/open-telemetry/opentelemetry-rust/blob/f9b08c10809e3b39497e21cccaa0d0c49b0216dc/opentelemetry-sdk/src/logs/log_emitter.rs#L116). And now if any of the loggers have it's reference in separate thread, the shutdown will not be invoked.  One option can be to add a new `LoggerProvider::Shutdown()` method, and explicitly invoke it from within `shutdown_logger_provider` . This new method will invoke shutdown on all processors, and so ensure that all the existing logs are flushed. And no new logs can be exported with processors in shutdown state.

TommyCpp:
Hmm I think this PR is the same as #1446 ?

hdost:
> Hmm I think this PR is the same as #1446 ?

Yes it's meant to be an alternative because right now the initial interface requires passing the under which is inherently leaky as an interface. This aims to remove the necessity.

lalitb:
Surprisingly, this PR has better improvement over #1446, as visible from the stress test:
```bash
/home/labhas/opentelemetry-rust/stress$ cargo run --bin logs --release
```

#1455 (current PR):
Number of threads: 8
Throughput: 20,641,400 iterations/sec
Throughput: 20,336,600 iterations/sec

#1446:
Number of threads: 8
Throughput: 14,935,400 iterations/sec
Throughput: 15,633,600 iterations/sec

#main branch:
Number of threads: 8
Throughput: 11,473,200 iterations/sec
Throughput: 11,909,200 iterations/sec

hdost:
That's a nice benefit as well. I believe it's because we're really doing less things in this one. 


Side note: I really would love to run whatever machines y'all are running because my same stress benchmarks never get as good of results. Maybe after the next refresh for my laptop 😅

lalitb:
> > And will calling shutdown on the global loggerProvider ensure that the logs through these loggers are not exported ?
> 
> It would be good to validate this use-case. It seems [shutdown_logger_provider](https://github.com/open-telemetry/opentelemetry-rust/blob/f9b08c10809e3b39497e21cccaa0d0c49b0216dc/opentelemetry/src/global/logs.rs#L134) will invoke shutdown on the LogProcessors when the inner instance is [dropped](https://github.com/open-telemetry/opentelemetry-rust/blob/f9b08c10809e3b39497e21cccaa0d0c49b0216dc/opentelemetry-sdk/src/logs/log_emitter.rs#L116). And now if any of the loggers have it's reference in separate thread, the shutdown will not be invoked. One option can be to add a new `LoggerProvider::Shutdown()` method, and explicitly invoke it from within `shutdown_logger_provider` . This new method will invoke shutdown on all processors, and so ensure that all the existing logs are flushed. And no new logs can be exported with processors in shutdown state.

Thinking about this scenario again,  it seems that the potential issue of leaking providers might not be as significant if we can document it clearly to explain the risk. Specifically, providers could be leaked if any of the tasks indefinitely retain their reference, which can result in a shutdown not triggering.

lalitb (approved):

cijothomas:
> > Hmm I think this PR is the same as #1446 ?
> 
> Yes it's meant to be an alternative because right now the initial interface requires passing the under which is inherently leaky as an interface. This aims to remove the necessity.

Just catching up on PRs! Could you describe any pros/cons of this vs the alternate in the PR desc?  Also, if feasible, can you add the stress test numbers you are observing. I am not seeing the same gains as Lalit has posted above, so wondering what is the actual gain! (i'll try to run more tests and post back as well soon)

lalitb:
My earlier results were on the Azure dev box, but I can see perf improvement on my Windows desktop with WSL2:

#main branch:
     Running `/home/labhas/obs/ot/opentelemetry-rust/target/release/logs`
Number of threads: 10
Throughput: 9,659,000 iterations/sec
Throughput: 9,166,800 iterations/sec
Throughput: 9,068,400 iterations/sec
Throughput: 9,013,600 iterations/sec

#this PR:
     Running `/home/labhas/obs/ot/opentelemetry-rust/target/release/logs`
Number of threads: 10
Throughput: 28,168,200 iterations/sec
Throughput: 26,013,800 iterations/sec
Throughput: 23,196,200 iterations/sec
Throughput: 24,413,400 iterations/sec

hdost:
Are we still thinking that we need to have the testing before this gets merged?

I'm trying to thing what would be a valid testing scenario that we'd want to run.

cijothomas:
> Are we still thinking that we need to have the testing before this gets merged?
> 
> I'm trying to thing what would be a valid testing scenario that we'd want to run.

From SIG Call:
We don't need to block the PR for tests - a targetted test can be added as a followup.
Scenario: https://github.com/open-telemetry/opentelemetry-rust/pull/1455#issuecomment-1877690239 
There are other issues reporting about shutdown scenarios, maybe we need a comprehensive testing targetting shutdowns, separate from this PR.

opentelemetry/CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ gains, and avoids `IndexMap` dependency. This affects `body` and `attributes` of
   `observe()` calls, and only accept a precreated `AttributeSet`
   value.
   [#1421](https://github.com/open-telemetry/opentelemetry-rust/pull/1421)
+- Turned off events for NoopLogger to save on operations
+  [1455](https://github.com/open-telemetry/opentelemetry-rust/issues/1455)
```
cijothomas, line 31:
```suggestion
  [1455](https://github.com/open-telemetry/opentelemetry-rust/pull/1455)
```

cijothomas (approved):
LGTM. 
Maybe good to add changelog entry to the sdk.
We can definitely follow up to do targeted tests for shutdown part.

cijothomas:
I ran stress test for logs after modifying the no-op processor to [return false ](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs#L61) for event_enabled.
main branch:

Number of threads: 8
Throughput: 75,537,000 iterations/sec
Throughput: 77,545,600 iterations/sec

This pr:

Number of threads: 8
Throughput: 579,008,400 iterations/sec
Throughput: 586,900,800 iterations/sec

On the other hand, if using a [no-op loglayer](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs#L18) with event_enabled returning false, the same test gives below:

Number of threads: 8
Throughput: 1,106,156,200 iterations/sec
Throughput: 1,069,867,600 iterations/sec

This test shows that OTel still has a lot more room for further optimizations. Those can be explored in the future. This PR already boosts the perf significantly!

PR #1456, opened 2023-12-26
chore: Pin nightly version for check-external-types

hdost:
This is is because the versions are extremely tied to each other.

Fixes #

## Changes


TommyCpp (approved):
why do we have to use nightly toolchain here though?

hdost:
> why do we have to use nightly toolchain here though?

It's currently using features only available in nightly.

PR #1252, opened 2023-09-06
Update tonic to 0.10.0

rodoufu:
Fixes #

Fixes and issue when using tonic 0.10.0

## Changes

Update tonic version to 0.10.0

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (if applicable)

PR #1457, opened 2023-12-27
Workspace dependencies

rodoufu:
Adds shared dependencies to the workspace.
It improves the experience while managing the dependencies.

## Changes

Picked the shared dependencies of the member projects and used added then to the workspace dependencies, https://crates.io/crates/work_dep can help with that. 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,3 +32,44 @@ resolver = "2"
 # See function names in profiling reports.
 # 2/true is too much, 0 is not enough, 1 is just right for back traces
 debug = 1
+
+[workspace.dependencies]
+bytes = "1"
+log = "0.4"
+
+tonic = "0.9"
+grpcio = "0.12"
+prost = "0.11"
+
+serde = "1.0"
+serde_json = "1.0"
+
+http = "0.2"
+reqwest = "0.11"
+isahc = "1.4"
+hyper = "0.14"
+url = "2.2"
+
+async-trait = "0.1"
+async-std = "1.10"
+rand = "0.8"
+
+tokio = "1"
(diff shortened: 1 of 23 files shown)


rodoufu:
Adding the the workspace dependencies will make it easier to then fix this other one https://github.com/open-telemetry/opentelemetry-rust/pull/1252 when tonic is updated to 0.10.0

lalitb:
I like this approach - seems this will ensure the version consistency for dependencies across the crates, and (possibly) decrease the build time.

Cargo.toml
```diff
@@ -32,3 +32,46 @@ resolver = "2"
 # See function names in profiling reports.
 # 2/true is too much, 0 is not enough, 1 is just right for back traces
 debug = 1
+
+[workspace.dependencies]
```
djc, line 36: Please alphabetize the dependencies. I would probably also just remove the grouping.

djc (approved):
I generally prefer this approach, and have used it successfully in other projects.

Cargo.toml
rodoufu: Done.

rodoufu: The `CI / docs` failed because of:
> Error: API rate limit exceeded for 40.76.119.131. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

I don't think I have permission to retry that.

lalitb (approved):
LGTM based on my Rust expertise. Good to have reviewed by @hdost , @TommyCpp too :)

TommyCpp (approved):
Seems to be a nice improvement

Issue #1327, opened 2023-10-31
Add integration test for OTLP exporters


TommyCpp:
In order to increase our confidence when releasing a new version. We should have integration tests for important exporters(OTLP, jager, etc). 

Currently we have limited integration test for jaeger, and CI setup to run integration tests on PRs. But we want to expand/move it to OTLP exporters

hdost:
Maybe consider using https://docs.rs/testcontainers/latest/testcontainers/ for starting the container. https://github.com/open-telemetry/opentelemetry-java/blob/main/integration-tests/otlp/src/main/java/io/opentelemetry/integrationtest/OtlpExporterIntegrationTest.java This is the example from the Java repo.

cijothomas:
I also want to share my own experience that, leveraging In-MemoryExporter for intergration tests is another great way of improving confidence in the overall SDK! Yes, it is not true e2e, but does help a lot. Integration test for OTLP (as proposed in this issue), combined with in-memory exporter tests should be a good boost to coverage, and improve our velocity.

https://github.com/open-telemetry/opentelemetry-rust/pull/1305

hdost:
Are there any other Tests we want to add?
Maybe we can also add a daily/weekly build of the integration test?

TommyCpp:
> Maybe we can also add a daily/weekly build of the integration test?

yeah we could set something up but I want to extend the integration test coverage first

hdost:
> > Maybe we can also add a daily/weekly build of the integration test?
> 
> 
> 
> yeah we could set something up but I want to extend the integration test coverage first

Which cases do you want to cover maybe we can split it up a bit.

hdost:
Btw i put this for SDK 
Stable because it could reveal issues in the underlying SDK implementation.

cijothomas:
Closing as we now have the infra for integration tests and basic test for traces.

PR #1462, opened 2024-01-02
fix(proto): use camelCase for json (de)serialization.

TommyCpp:
Part of the effort of #1327 as we need  json formats for assertation in integration tests. 

## Changes

- add configuration to serde to deserialize the json in camelCase field name
- add custom (de)serialization for traceId, spanId as they are case-insensitive hex encoded string(see [here](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding))
- add custom (de)serialization for `KeyValue`
- add tests for above, and a test using example json files

* [ x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [] Changes in public API reviewed (if applicable)

### DIFF
--- a/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.common.v1.rs
+++ b/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.common.v1.rs
@@ -3,6 +3,7 @@
 /// object containing arrays, key-value lists and primitives.
 #[cfg_attr(feature = "with-schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "with-serde", serde(rename_all = "camelCase"))]
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct AnyValue {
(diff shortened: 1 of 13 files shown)


jtescher (approved):
Nice addition! Little unfortunate about the `hex` dependency but it's probably fine for this crate as it does need to have arbitrary bytes <-> hex conversions

lalitb (approved):
LGTM.

TommyCpp:
> Nice addition! Little unfortunate about the `hex` dependency but it's probably fine for this crate as it does need to have arbitrary bytes <-> hex conversions

If we feel strong about it we can implement our own methods to do the bytes <-> hex conversion?

sl0thentr0py:
hey we need this, can you make a release for this crate?

opentelemetry-proto/src/proto/tonic/opentelemetry.proto.common.v1.rs
```diff
@@ -68,18 +72,28 @@ pub struct KeyValueList {
...
 #[cfg_attr(feature = "with-schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "with-serde", serde(rename_all = "camelCase"))]
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct KeyValue {
     #[prost(string, tag = "1")]
     pub key: ::prost::alloc::string::String,
     #[prost(message, optional, tag = "2")]
+    #[cfg_attr(
```
hwrdtm, line 82: May I ask why this block was added? It is inconsistent with what the serde semantics are as implemented on the (Golang-based) OTEL Collector - [see this example](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/ptrace/json_test.go#L91) 

In my local fork I've removed this and only then am I able to successfully unmarshal from within the Collector.

PR #1477, opened 2024-01-16
feat(otlp): add integration test setup for traces

### LINKED ISSUE #1327: Add integration test for OTLP exporters

TommyCpp:
#1327 

## Changes
- remove jaeger integration test
- add integration test setup.


cijothomas:
Thanks @TommyCpp ! Will you be removing the existing jaeger based integration test in favor of this?

opentelemetry-otlp/tests/integration_test/src/asserter.rs
```diff
@@ -0,0 +1,50 @@
+use opentelemetry_proto::tonic::trace::v1::ResourceSpans;
+use std::fs::File;
+
+// assert two resource spans is the same except for the timestamps in it
```
cijothomas, line 4: we can do similar exception for spanid/traceid as well, to keep this simpler.

TommyCpp: I think for span ID and trace ID I'd rather use a `IncrementIdGenerator` so that we get predictable IDs. Because we want to make sure the parent-child relationship is correct

opentelemetry-otlp/tests/integration_test/expected/traces.json
```diff
@@ -0,0 +1,102 @@
...
+              "endTimeUnixNano": "1703985537070572718",
+              "attributes": [
+                {
+                  "key": "lemons",
+                  "value": {
+                    "stringValue": "five"
+                  }
+                }
+              ],
+              "events": [
```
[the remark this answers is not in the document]
TommyCpp: Good point I will add it

TommyCpp:
> Will you be removing the existing jaeger based integration test in favor of this?

I think it's fine to keep the integration test in jaeger in place? That only tests jaeger so it's not mutally exclusive. 

I plan to add a jaeger container here to integrate test otlp exporter w/ jaeger backend too in follow up PRs

opentelemetry-otlp/tests/integration_test/src/asserter.rs
cijothomas: In my view, that is partially defeating the purpose of integration tests.
Integration tests should be executing same code as a user would be doing, and custom IdGenerator deviates from that. 
Testing parent-child relation is not the goal of integration tests, right? 

Here's what I am thinking:
1. Execute the same hello world style example code + OTLP Exporter
2. Setup OTel Collector and have it save telemetry to file.
3. Validate that the file has expected telemetry. Things like attributes of diff. types are received, span events are received, metrics are correct etc. - You can of course validate that span's parent-child relation is maintained without pre-knowing the actual traceid/spanid.

As a first step, I'd suggest to start with simply validating that a file is created by collector, and we can add validation of its content as a next PR, to keep things moving easily.

TommyCpp: Yeah the current setup complete the 1 and 2. 

> Integration tests should be executing same code as a user would be doing, and custom IdGenerator deviates from that.

I think the `IdGenerator` is a small enough component that self-contained and can be well tested so the chance we missed anything bu not using the `RandomIdGenerator` is minimal

> You can of course validate that span's parent-child relation is maintained without pre-knowing the actual traceid/spanid.

Yeah it's boils down to a compare two forests. But I didn't find a easy way to write the expected span tress before asserting. I think a prediable ID is a good start point for now until we start testing context progatation through multiple containers. 

In other words, if we are only testing export from single container to collector. `IncrementIdGenerator` can satisify the requirment to be unique in this simple system. Randomness is not required by spec(though it's preferred)

cijothomas: > Yeah the current setup complete the 1 and 2.


Not quite as this is executing diff. code than user would be (due to idgenerator). IMHO, Intergration tests should treat the library as blackbox only, and dont change anything at all, including idgenerator.

TommyCpp: Not sure I understand. 

> the current setup complete the 1 and 2.

The current setup includes the following steps:
- Execute the same hello world style example code + OTLP Exporter. This is done by spin up a collector container using testcontainer create and mount a local file to collector container.
- Setup OTel Collector and have it save telemetry to file. The result will be written to the mounted file by collector container

I think the difference is on how do we assert the result. The current setup opt in a quick appoaches where we parse the json output from collector and examine the `ResourceSpan` is expected, which need the predictable trace IDs and span IDs. 

I agree on the long term we should take a step further and rebuild the span relationship from `ResourceSpan`s but not sure if it should be a blocker for this PR

cijothomas: > examine the ResourceSpan is expected, which need the predictable trace IDs and span IDs.

Why need predictable IDs? We don't use predictable timestamps, right? I was asking to do same for traceid,spanid as well. Anything that is auto-generated (and hence unpredictable), can be avoided from the assertion. Anything user-provided, and hence predictable can be validated.

To reiterate my opinion:
We should not use custom id generator, and instead only execute test app as a normal user would execute. I understand IdGenerator is a small change only, but that is still partially defeating the purpose of integration tests with otlp!

TommyCpp: >  I was asking to do same for traceid,spanid as well

I think the integration tests don't add much values if we cannot validate parent-children relationship between spans thus I am hesitiate to just ignore the trace IDs and span IDs. For example, say we have a bug that caused every span to be a root span. Without assertion on span IDs and traces IDs the integration test won't catch this. 

> Why need predictable IDs? We don't use predictable timestamps, right?

One point is we do allow users to provide their own `IdGenerator` while we don't allow users to bring their own clock

Either case if you feel strong about the predictable IdGenerator I can explore other options to assert the span relationships 🙂

cijothomas: > I think the integration tests don't add much values if we cannot validate parent-children relationship

Agree. But we dont need predictable ids for that. We just need to validate that for the root span, `parentSpanId` will be empty, and the `parentSpanId` of the child will be the `spanId` of the parent. The values themselves can be anything! Its totally okay to add this validation in the future.

TommyCpp: > We just need to validate that for the root span, parentSpanId will be empty, and the parentSpanId of the child will be the spanId of the parent

This only validate the number of root spans and non root spans. To fully validate the span relationship you need to build two span forests and build a map between each spans in expected result and actual results, then validate they are the same. We will also need to handle links, which make the question even harder that we need build and compare two directed cyclic graph.

Prediable ID avoid all those operation with small drawback that we are not testing the random generation part of the operation, which IMO is a good trade off.

TommyCpp:
Looking at the output of the file exporter. I find the timestamp fields are all string while they should be uint64(the example json from the otlp protocol has them as ints). I will open an issue on file exporter side to see if that's expected. Until then probably won't be able to merge this

cijothomas:
> Looking at the output of the file exporter. I find the timestamp fields are all string while they should be uint64(the example json from the otlp protocol has them as ints). I will open an issue on file exporter side to see if that's expected. Until then probably won't be able to merge this

I think we can unblock ourselves by merging this PR just after checking a file is generated. Validating its contents can be done as a next PR, so we can keep things moving.

opentelemetry-otlp/tests/integration_test/src/asserter.rs
cijothomas: _but we did seen a lot of issues with context propagation_

That should be tested via InMemoryExporter, right? Why its part of integration tests ?

opentelemetry-otlp/tests/integration_test/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+once_cell = "1.17"
+opentelemetry = { path = "../../../opentelemetry", features = ["metrics", "logs"] }
+opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "logs", "testing"] }
+opentelemetry-otlp = { path = "../../../opentelemetry-otlp", features = ["tonic", "metrics", "logs"] }
+opentelemetry-semantic-conventions = { path = "../../../opentelemetry-semantic-conventions" }
+opentelemetry-proto = { path = "../../../opentelemetry-proto", features = ["gen-tonic-messages", "trace", "with-serde"] }
+tokio = { version = "1.0", features = ["full"] }
+log = { version = "0.4.17" }
```
cijothomas, line 16: do we use this anywhere?

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    let docker = Cli::default();
+    let mut image =
+        RunnableImage::from(collector_image).with_container_name(COLLECTOR_CONTAINER_NAME);
+
+    for port in [
+        4317,  // gRPC port
+        4318,  // HTTP port
+        55681, // OpenTelemetry protocol port
```
cijothomas, line 71: do we need this port?

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
```diff
@@ -0,0 +1,95 @@
...
+        8888,  // Prometheus metrics exposed by the collector
```
cijothomas, line 72: do we need this? I think only 4317, 4318 should be needed for testing the gRPC and Http based exporting and nothing more.

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let collector_container = docker.run(image);
+
+    tokio::time::sleep(Duration::from_secs(5)).await;
+    traces::traces().await.unwrap();
+
+    // wait for file to flush to disks
+    // ideally we should use volume mount but otel collector file exporter doesn't handle permission too well
+    // bind mount mitigate the issue by set up the permission correctly on host system
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
cijothomas, line 88: mm.. These generally cause random failures in the CI checks...One way to mitigate this is by checking for the file's presence for upto N seconds, sleeping in between checks. (A good candidate for this future improvement.)

opentelemetry-otlp/tests/integration_test/tests/traces.rs
```diff
@@ -0,0 +1,93 @@
...
+                .tonic()
+                .with_endpoint("http://localhost:4317"),
+        )
+        .with_trace_config(
+            sdktrace::config()
+                .with_resource(Resource::new(vec![KeyValue::new(
+                    opentelemetry_semantic_conventions::resource::SERVICE_NAME,
+                    "basic-otlp-tracing-example",
+                )]))
+                // use increment id generator so we get predictable test results
```
cijothomas, line 29: we don't validate this so, I suggest to remove this customization.

opentelemetry-otlp/tests/integration_test/src/images.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn volumes(&self) -> Box<dyn Iterator<Item = (&String, &String)> + '_> {
+        Box::new(self.volumes.iter())
+    }
+
+    fn expose_ports(&self) -> Vec<u16> {
+        vec![
+            // 4317,  // gRPC port, defined in Dockerfile
+            // 4318,  // HTTP port, defined in Dockerfile
+            // 55681, // OpenTelemetry protocol port, defined in Dockerfile
+            8888, // Prometheus metrics exposed by the collector
```
lalitb, line 33: nit - do we need all these ports exposed?

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
TommyCpp: Removed both

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
TommyCpp: We can add retry with backoff once we have more complicate integration tests

opentelemetry-otlp/tests/integration_test/src/asserter.rs
```diff
@@ -0,0 +1,221 @@
+use opentelemetry_proto::tonic::trace::v1::{ResourceSpans, Span, TracesData};
```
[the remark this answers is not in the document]
TommyCpp: I don't have a strong preference but feel like we can refactor as we adding more signals to it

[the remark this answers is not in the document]
lalitb: Agree.

opentelemetry-otlp/tests/integration_test/tests/traces.rs
```diff
@@ -0,0 +1,90 @@
...
+    let right = read_spans_from_json(File::open("./expected/failed_traces.json").unwrap());
+
+    TraceAsserter::new(right, left).assert();
+}
+
+#[test]
+pub fn test_assert_span_eq() {
+    let spans = read_spans_from_json(File::open("./expected/traces.json").unwrap());
+
+    TraceAsserter::new(spans.clone(), spans).assert();
```
lalitb, line 89: Is this test meant to validate the TraceAsserter logic, as it is always testing two identical spans?

opentelemetry-otlp/tests/integration_test/tests/traces.rs
```diff
@@ -0,0 +1,90 @@
...
+pub fn assert_traces_results(result: &str, _: &str) {
+    // let left = read_spans_from_json(File::open(expected).unwrap());
+    // let right = read_spans_from_json(File::open(result).unwrap());
+
+    // TraceAsserter::new(left, right).assert();
+    //
+    // we cannot read result json file because the timestamp was represents as string instead of u64.
+    // need to fix it on json file exporter
+
+    assert!(File::open(result).unwrap().metadata().unwrap().size() > 0)
```
lalitb, line 73: Just to confirm, as of now we are not testing the success scenario, only testing the creation of result file?

TommyCpp: Yes

opentelemetry-otlp/tests/integration_test/tests/traces.rs
TommyCpp: Yes. it's like meta testing for testing tool

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn create_temporary_result_file(&self) -> File {
+        let file = File::create(self.result_file_path()).unwrap();
+        file.set_permissions(std::fs::Permissions::from_mode(0o666))
+            .unwrap();
+        file
+    }
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[ignore] // skip when running unit test
```
cijothomas, line 51: do we only expect to run this manually? Or we expect this to be run in CI as well?

lalitb (approved):
LGTM, thanks. The validation of exported data can be in subsequent PR.

opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
cijothomas: nvm. found that we run even ignored in intergration tests.

cijothomas (approved):

PR #1481, opened 2024-01-18
Fix error message caused by race condition when using PeriodicReader

### LINKED ISSUE #1244: Race when using PeriodicReader to export metrics
When setting up a `PeriodicReader`, the thread to start exporting beings immediately.

This does not give enough time to guarantee that `MeterProviderBuilder#build` has been called, resulting in the `sdk_producer` within the `PeriodicReader` not being set.

This results in a call to the telemetry global error handler which probably isn't fatal, but does produce an alarming message.

izquierdo:
Fixes #1244

## Changes

This defers starting the exporting thread until `MeterProviderBuilder#build` has been called.


lalitb:
@izquierdo The changes look to be in the right direction. Wondering if you plan to make this ready for review. Thanks.

izquierdo:
> @izquierdo The changes look to be in the right direction. Wondering if you plan to make this ready for review. Thanks.

Yes, I haven't done so yet because of the new lines that are missing test coverage. Will mark it for review after updating the tests. Thank you.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -392,3 +404,25 @@ impl MetricReader for PeriodicReader {
...
+        let reader = PeriodicReader::builder(exporter.clone(), runtime::Tokio).build();
+
+        let mut rm = ResourceMetrics {
+            resource: Resource::empty(),
+            scope_metrics: Vec::new(),
+        };
+        let result = reader.collect(&mut rm);
+        result.expect_err("error expected when reader is not registered");
+    }
+}
```
lalitb, line 428: Do you think it would be easy to add the test for success scenario (something like):
 - create PeriodicReader 
 - Create SdkMeterProvider instance / or Pipelines with the above created reader.
 - This should invoke worker, and collect cycle.
 
If not straightforward, we can park it for later. (Ideally, this test should have existed at first place before this PR.)

lalitb (approved):
Nicely done. Thanks for the PR :)

opentelemetry-sdk/src/metrics/periodic_reader.rs
izquierdo: I'll give it a shot :+1:

cijothomas:
Oops! Merge confict... Could you resolve that?

cijothomas:
Thanks @izquierdo for your contribution! Hope you can continue with more! We could really use more hands here.

PR #1403, opened 2023-11-26
fix(sdk, zipkin, otlp): use `temp_env` for testing

TommyCpp:
Fixes #1383
Design discussion issue (if applicable) #

## Changes

Replace all env locks with `temp-env` create

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (if applicable)

PR #1496, opened 2024-01-26
Use temp_env when constructing BatchConfig from environment variables in log_processor.rs test

### LINKED ISSUE #1490: Flaky test
---- logs::log_processor::tests::test_default_batch_config_adheres_to_specification stdout ----
thread 'logs::log_processor::tests::test_default_batch_config_adheres_to_specification' panicked at opentelemetry-sdk/src/logs/log_processor.rs:504:9:
assertion `left == right` failed
  left: 2.046s
 right: 30s

failures:
    logs::log_processor::tests::test_default_batch_config_adheres_to_specification

mattbodd:
Fixes #1490 
 
## Description
When constructing a `BatchConfig` instance using the `default` method, the `BatchConfigBuilder::init_from_env_vars` method is implicitly called. In a multi-threaded environment, this can lead to unsynchronized reads and writes of common environment variables which is what happens when running the tests with multiple threads in `opentelemetry-sdk/src/logs/log_processor.rs`.

This change updates one flaky test, `test_default_batch_config_adheres_to_specification`, that constructed a `BatchConfig` using the `default` method to use `temp_env::with_vars_unset` which will synchronize other tests' accesses to environment variables (assuming they use `temp_env` methods as well), and will ensure that the environment variables specified are unset which is what the test expects.

## Changes

* Use `temp_env::with_vars_unset` in `test_default_batch_config_adheres_to_specification` to ensure default values are used rather than environment variable overrides

## Validation
* Reproduce bug described in #1490 locally by repeatedly running all `log_processor` tests
  * As is the nature of concurrent bugs, failures were non-deterministic but occurred after only a handful of attempts
* Attempt to reproduce the issue after leveraging `temp_env::with_vars_unset` and could not after 50+ attempts of running all `log_processor` tests


cijothomas (approved):
Thanks.
Would be good to do an audit in the repo to see if such serial-enforcement is required elsewhere too.

TommyCpp (commented):
We have used `temp_env` in the past to avoid conflicting env vars https://github.com/open-telemetry/opentelemetry-rust/pull/1403. Can we look into doing the same here?

mattbodd:
@TommyCpp `temp_env` uses a singleton mutex to synchronize accesses but the problem is that while our tests synchronize reads and writes with `temp_env`, `opentelemetry_sdk::logs::log_processor::BatchConfigBuilder::init_from_env_vars()` uses unsynchronized reads of environment variables. We could conditionally (`#[cfg(test)]`) use `temp_env` while reading environment variables outside of tests to make use of `temp_env`'s implicit synchronization or we can synchronize accesses across our tests with something like `serial_test`.

TommyCpp:
> temp_env uses a singleton mutex to synchronize accesses but the problem is that while our tests synchronize reads and writes with temp_env

temp_env only provides method to sync on writing env vars and users are expected to use standard `env::var` function to read from it. I notice that we already using temp vars in those tests it's worth figure out if we are setting it correctly as we are running some issues with it.

mattbodd:
Thanks for the feedback @TommyCpp. I agree with your suggestion and think it's clearer to use `temp_env` as you described than introduce an additional way of synchronizing environment variable accesses. I removed usage of `serial_test` in favor of using `temp_env::with_vars_unset` in `test_default_batch_config_adheres_to_specification`.

lalitb (approved):

cijothomas (approved):

PR #1503, opened 2024-01-31
Enable doc generation for testing module for InMemory*Exporters.

### LINKED ISSUE #1466: [Feature]: Add an Arc to ManualReader or make SharedReader public
### Related Problems?

See #1329 for previous discussion.

Currently, `ManualReader` is not useful on its own. I tried using it in a unit test, but `MeterProviderBuilder::with_reader()` takes ownership of the `MetricReader` implementation that is passed in. Thus, I can't call `ManualReader::collect()` later. I am aware of three uses of `ManualReader`, and all three create their own newtype around `Arc<ManualReader>`, and implement `TemporalitySelector`, `AggregationSelector`, and `MetricReader` on it, dispatching to the inner reader in the natural way. This is done in `opentelemetry-sdk/benches/metric.rs` and `opentelemetry-prometheus/src/lib.rs` in this repository, and in `tracing-opentelemetry`, in `tests/metrics_publishing.rs`.

I think it would make sense to change `ManualReader` so that it is an `Arc` around an inner struct, and thus can be cloned.

### Considered Alternatives

Alternately, the `SharedReader` struct currently in the benchmark could be moved into the library proper, and made public. This would allow users to opt-in to the `Arc`, and reduce the boilerplate they have to write.

### Additional Context

_No response_

lalitb:
Fixes #1466

## Changes

Enable doc generation for InMemory*Exporter when built with the `testing` feature. As it should be possible for users to use these exporters for testing purpose.

The screenshot of the doc from this PR:
image

image


hdost (commented):
Looks good for the most part will have to run docs locally to double check.

lalitb:
> Looks good for the most part will have to run docs locally to double check.

Yes agree, have validated locally and shared the screenshot of the generated docs in the PR description.

cijothomas:
> Looks good for the most part will have to run docs locally to double check.

Okay to merge?

cijothomas (approved):

opentelemetry-sdk/src/testing/trace/span_exporters.rs
```diff
@@ -0,0 +1,168 @@
...
+        events: SpanEvents::default(),
+        links: SpanLinks::default(),
+        status: Status::Unset,
+        resource: config.resource,
+        instrumentation_lib: InstrumentationLibrary::default(),
+    }
+}
+
+#[derive(Debug)]
+pub struct TestSpanExporter {
```
TommyCpp, line 44: Can we add some documents for `TestSpanExporter`? If we are making it public

TommyCpp (commented):
`testing` used to be a collection of helpers that used by other crates in this repo for unit tests. I understand it maybe helpful to open it up so users can test their custom exporters. But I think we should make sure every helper has basic documentation before merging this

opentelemetry-sdk/src/testing/trace/span_exporters.rs
lalitb: This is not made public. One InMemory*Exporters are public.

TommyCpp: Ah I see

lalitb:
> `testing` used to be a collection of helpers that used by other crates in this repo for unit tests. I understand it maybe helpful to open it up so users can test their custom exporters. But I think we should make sure every helper has basic documentation before merging this

Currently, only the memory exporters are made open for external users, as required by the specs. Apart from that, the `testing` module has TestSpanExporter and TestMetricReader. I believe it should be possible to replace them in unit-tests with the in-memory exporters as separate PR. So kept their docs hidden. Let me know if there are other thoughts.

TommyCpp (approved):
Sounds good! Thanks for fixing it

PR #1515, opened 2024-02-07
Add add_link API 

lalitb:
## Changes

The API is already supported as experimental in specification:
https://github.com/open-telemetry/opentelemetry-specification/blob/v1.29.0/specification/trace/api.md#add-link

And planned to make stable soon - https://github.com/open-telemetry/opentelemetry-specification/issues/3865. So not added under experimental flag.


opentelemetry-sdk/src/trace/span.rs
```diff
@@ -170,6 +170,28 @@ impl opentelemetry::trace::Span for Span {
         });
     }
 
+    /// Add link to this `Span`
```
hdost, line 173:
```suggestion
    /// Add `Link` to this `Span`
```

hdost (commented):
Looks good for the most part. I would have to look at the commit log but if this Link api was released in a previous version of the `opentelemetry` crate that we should have a CHANGELOG entry for the signature change and addition of the `with_context`

TommyCpp (commented):
We need to wrap this API with `otel_unstable` as it's experimental

cijothomas:
> We need to wrap this API with `otel_unstable` as it's experimental

Do we need to? The spec is about to get stable soon, so its not worth the effort to hide it under feature-flag now and later undo that.

opentelemetry/src/trace/mod.rs
```diff
@@ -302,11 +302,24 @@ pub struct Link {
 }
 
 impl Link {
-    /// Create a new link.
-    pub fn new(span_context: SpanContext, attributes: Vec<KeyValue>) -> Self {
+    /// Create new `Link`
+    pub fn new(
+        span_context: SpanContext,
+        attributes: Vec<KeyValue>,
+        dropped_attributes_count: u32,
```
cijothomas, line 309: not really introduced in this PR, but the API exposing things like `dropped_attributes_count` is not good :( It is up-to the SDK to enforce the limit and store the count of attributes, not that of API.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-limits

TommyCpp:
> Do we need to? The spec is about to get stable soon, so its not worth the effort to hide it under feature-flag now and later undo that.

Emm shall we wait for the spec PR to merge before we merge this one then? I am mostly worried we merge this and then spec made some changes before stablizing. Not sure how long it usually takes for spec to stablize

cijothomas:
> > Do we need to? The spec is about to get stable soon, so its not worth the effort to hide it under feature-flag now and later undo that.
> 
> Emm shall we wait for the spec PR to merge before we merge this one then? I am mostly worried we merge this and then spec made some changes before stablizing. Not sure how long it usually takes for spec to stablize

We are still pre 1.0, so its still okay? Or we can hold off from merging for few days as well and see if the spec actually gets stable.

lalitb:
> We are still pre 1.0, so its still okay? Or we can hold off from merging for few days as well and see if the spec actually gets stable.

Yeah, will add the `otel_unstable` flag if it is not stable after (say) a couple of weeks.

opentelemetry/src/trace/mod.rs
cijothomas: @lalitb You don't plan to address this in this PR right?

lalitb: This needs to be addressed for both Span Events and Span Links. Let me address that separately. Will create an issue for that.

opentelemetry/src/trace/span.rs
```diff
@@ -154,6 +154,21 @@ pub trait Span {
...
+    /// to connect spans from different traces or within the same trace. Attributes can be attached to the link to
+    /// provide additional context or metadata.
+    ///
+    /// # Arguments
+    ///
+    /// * `span_context` - The `SpanContext` of the span to link to. This represents the target span's unique identifiers
+    ///   and trace information.
+    /// * `attributes` - A vector of `KeyValue` pairs that describe additional attributes of the link. These attributes
+    ///   can include any contextual information relevant to the link between the spans.
+    ///
```
TommyCpp, line 169:
```suggestion
    /// Adds [`Link`] to another [`SpanContext`].
    ///
    /// This method allows linking the current span to another span, identified by its `SpanContext`. Links can be used
    /// to connect spans from different traces or within the same trace. Attributes can be attached to the link to
    /// provide additional context or metadata.
    ///
    /// # Arguments
    ///
    /// * `span_context` - The `SpanContext` of the span to link to. This represents the target span's unique identifiers
    ///   and trace information.
    /// * `attributes` - A vector of `KeyValue` pairs that describe additional attributes of the link. These attributes
    ///   can include any contextual information relevant to the link between the spans.
    ///
    /// [`Link`]: crate::trace::Link
```
nit: we can link the documentation using [identifier] format in rust doc. I find it to be useful for users so we should try to do more of those

TommyCpp (approved):
one nit otherwise LGTM

opentelemetry/src/trace/mod.rs
hdost: @lalitb

lalitb: Done now - #1629

PR #1516, opened 2024-02-08
Add tests to confirm observable counter aggregation bug

cijothomas:
Observable counter has an aggregation bug. Most likely same issue exists for ObservableUpDownCounter as well.
This PR just adds a test that will fail, and hence ignored for now.
Will look at the fix separately.


opentelemetry-sdk/benches/metric_counter.rs
```diff
@@ -69,6 +70,17 @@ fn counter_add(c: &mut Criterion) {
     });
 }
 
-criterion_group!(benches, criterion_benchmark);
+#[cfg(not(target_os = "windows"))]
```
cijothomas, line 73: unrelated change, sorry! Hope its minimal (copy from other benches) and hence okay!

lalitb (approved):
LGTM once conflict fixed.

lalitb:
We can also consider the use of parameterized tests for scenarios where only the inputs and expected outputs differ, to reduce redundant test code.  Rust doesn't support it directly, but there are external crates providing this option e.g., - https://crates.io/crates/parameterized_test , or we can create a custom macro as suggested here - https://stackoverflow.com/questions/34662713/how-can-i-create-parameterized-tests-in-rust/

Not related to this PR, for future :)

PR #1518, opened 2024-02-08
Use consistent style of passing attributes to metrics

cijothomas:
Metric attributes were passed inconsistently throughout the repo. Some used &[] and some used [].as_ref(). This modifies everywhere to use the former, as I believe that is more idiomatic than using as_ref() and more concise too.

### DIFF
--- a/opentelemetry/src/metrics/meter.rs
+++ b/opentelemetry/src/metrics/meter.rs
@@ -83,8 +83,7 @@ pub trait MeterProvider {
 /// // Record measurements using the counter instrument add()
 /// u64_counter.add(
 ///     10,
-///     &
-///     [
+///     &[
 ///         KeyValue::new("mykey1", "myvalue1"),
 ///         KeyValue::new("mykey2", "myvalue2"),
 ///     ]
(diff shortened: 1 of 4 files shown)


opentelemetry/src/metrics/meter.rs
```diff
@@ -83,10 +83,11 @@ pub trait MeterProvider {
 /// // Record measurements using the counter instrument add()
 /// u64_counter.add(
 ///     10,
+///     &
```
hdost, line 86: Nit: Remove the new line

opentelemetry/src/metrics/meter.rs
```diff
@@ -95,10 +96,11 @@ pub trait MeterProvider {
 /// // Record measurements using the counter instrument add()
 /// f64_counter.add(
 ///     3.15,
+///     &
```
hdost, line 99: Nit

hdost (approved):
Minor notes. Looks 👍 good. There are times when & and .as_ref() aren't always the same like with Option, but in this case it's fine.

lalitb (approved):

PR #1519, opened 2024-02-08
Reduce cost of no attribute counters

KallDrexx:
Fixes #

## Changes

This PR utilizes atomics to make counters fast when an empty set of attributes are passed in.  This atomic bypasses the hashmap lookups and mutex locking.

### main
```
Counter/AddNoAttrs      time:   [59.516 ns 59.947 ns 60.457 ns]
  2 (2.00%) high mild
  4 (4.00%) high severe

Counter/AddNoAttrsDelta time:   [62.342 ns 63.361 ns 64.458 ns]
  1 (1.00%) high mild
  1 (1.00%) high severe

Counter/AddOneAttr      time:   [144.01 ns 146.76 ns 150.11 ns]
  5 (5.00%) high mild
  8 (8.00%) high severe

Counter/AddOneAttrDelta time:   [144.58 ns 145.39 ns 146.28 ns]
  1 (1.00%) high mild
  5 (5.00%) high severe
```

### PR
```
Counter/AddNoAttrs      time:   [30.140 ns 30.390 ns 30.652 ns]
                        change: [-49.181% -48.265% -47.331%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  8 (8.00%) high severe
Counter/AddNoAttrsDelta time:   [29.836 ns 30.040 ns 30.327 ns]
                        change: [-53.397% -52.564% -51.790%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  3 (3.00%) high severe
Counter/AddOneAttr      time:   [145.48 ns 145.84 ns 146.24 ns]
                        change: [-0.3877% +0.8455% +2.1208%] (p = 0.20 > 0.05)
  1 (1.00%) high mild
  7 (7.00%) high severe
Counter/AddOneAttrDelta time:   [147.37 ns 148.07 ns 148.85 ns]
                        change: [+0.5418% +2.2081% +3.9805%] (p = 0.01 < 0.05)
  4 (4.00%) high mild
  7 (7.00%) high severe
```

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -172,3 +172,75 @@ impl AtomicallyUpdate<f64> for f64 {
         F64AtomicTracker::new()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn can_add_and_get_u64_atomic_value() {
+        let atomic = u64::new_atomic_tracker();
+        atomic.add(15);
+        atomic.add(10);
+
+        let value = atomic.get_value();
+        assert_eq!(value, 25);
+    }
+
+    #[test]
+    fn can_reset_u64_atomic_value() {
+        let atomic = u64::new_atomic_tracker();
+        atomic.add(15);
+
+        let value = atomic.get_and_reset_value();
+        let value2 = atomic.get_value();
+
+        assert_eq!(value, 15, "Incorrect first value");
+        assert_eq!(value2, 0, "Incorrect second value");
+    }
+
+    #[test]
+    fn can_add_and_get_i64_atomic_value() {
+        let atomic = i64::new_atomic_tracker();
+        atomic.add(15);
+        atomic.add(-10);
+
+        let value = atomic.get_value();
+        assert_eq!(value, 5);
+    }
+
+    #[test]
+    fn can_reset_i64_atomic_value() {
+        let atomic = i64::new_atomic_tracker();
+        atomic.add(15);
+
+        let value = atomic.get_and_reset_value();
+        let value2 = atomic.get_value();
+
+        assert_eq!(value, 15, "Incorrect first value");
+        assert_eq!(value2, 0, "Incorrect second value");
+    }
+
+    #[test]
+    fn can_add_and_get_f64_atomic_value() {
+        let atomic = f64::new_atomic_tracker();
+        atomic.add(15.3);
+        atomic.add(10.4);
+
+        let value = atomic.get_value();
+
+        assert!(f64::abs(25.7 - value) < 0.0001);
+    }
+
+    #[test]
+    fn can_reset_f64_atomic_value() {
+        let atomic = f64::new_atomic_tracker();
+        atomic.add(15.5);
+
+        let value = atomic.get_and_reset_value();
+        let value2 = atomic.get_value();
+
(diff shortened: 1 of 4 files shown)


cijothomas:
Consider a changelog entry, as users upgrading will learn about the perf improvement of zero-attributes, and can leverage it to achieve bound-instrument (not technically, but a reasonable workaround).

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -71,3 +88,87 @@ impl Number<f64> for f64 {
...
+
+impl AtomicallyUpdate<i64> for i64 {
+    type AtomicTracker = AtomicI64;
+
+    fn new_atomic_tracker() -> Self::AtomicTracker {
+        AtomicI64::new(0)
+    }
+}
+
+pub(crate) struct F64AtomicTracker {
```
cijothomas, line 136: it maybe too much to write the inmemoryexporter test for each type, but I think we'll need to cover it somehow, given we have different implementations for different type. Maybe some unit tests in this file can help, not a blocker for this PR

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -10,26 +11,39 @@ use opentelemetry::{global, metrics::MetricsError};
...
-    Number,
+    AtomicTracker, Number,
 };
 
 /// The storage for sums.
-#[derive(Default)]
 struct ValueMap<T: Number<T>> {
     values: Mutex<HashMap<AttributeSet, T>>,
+    has_no_value_attribute_value: AtomicBool,
+    no_attribute_value: T::AtomicTracker,
```
cijothomas, line 21: do we expect Histograms to gain from a similar optimization? or the gains won't be that much as we anyway need some locking for histograms anyway?

cijothomas (commented):
Left some nits, but overall looks very good.

opentelemetry-sdk/src/metrics/internal/sum.rs
KallDrexx: I haven't looked into histograms at all yet.  I just started with counters as they were the part I was most familiarly with.

KallDrexx:
Unit tests and changelog updated

opentelemetry-sdk/src/metrics/internal/sum.rs
cijothomas: Np. I'll check histograms later.

cijothomas (approved):

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb, line 21: Sorry, missed out on reviewing this. It seems  ValueMap instance will either store a HashMap or a single value (`no_attribute_value` along with `has_no_value_attribute_value`), but not both. Should we instead use enum to segregate them, for efficient memory utilization. For no-attributes scenario, even for empty `values` object, the ovehead from metdata for Mutex and HashMap would be more than that of enum. Ignore if this doesn't makes sense, still going through the implementation :)

KallDrexx: This is not true.  You can call Add with no attributes and then add again with an attribute set.  So one counter can have both

cijothomas: @KallDrexx use case relies on user discipline to ensure no attributes are passed, not something SDK can assume and do optimizations!

PR #1521, opened 2024-02-08
Test successful collection by PeriodicReader

izquierdo:
## Changes

Add a success scenario test for `PeriodicReader` as suggested in https://github.com/open-telemetry/opentelemetry-rust/pull/1481#discussion_r1479307855

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -438,7 +438,7 @@ mod tests {
         // Assert
         receiver
             .recv_timeout(interval * 2)
-            .expect("message should be available in channel");
+            .expect("message should be available in channel, indicating a collection occurred");
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -409,20 +409,52 @@ impl MetricReader for PeriodicReader {
 mod tests {
     use super::PeriodicReader;
     use crate::{
-        metrics::data::ResourceMetrics, metrics::reader::MetricReader, runtime,
-        testing::metrics::InMemoryMetricsExporter, Resource,
+        metrics::data::ResourceMetrics, metrics::reader::MetricReader, metrics::SdkMeterProvider,
+        runtime, testing::metrics::InMemoryMetricsExporter, Resource,
     };
+    use opentelemetry::metrics::MeterProvider;
+    use std::sync::mpsc;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn registration_triggers_collection() {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -409,20 +409,52 @@ impl MetricReader for PeriodicReader {
...
-        metrics::data::ResourceMetrics, metrics::reader::MetricReader, runtime,
-        testing::metrics::InMemoryMetricsExporter, Resource,
+        metrics::data::ResourceMetrics, metrics::reader::MetricReader, metrics::SdkMeterProvider,
+        runtime, testing::metrics::InMemoryMetricsExporter, Resource,
     };
+    use opentelemetry::metrics::MeterProvider;
+    use std::sync::mpsc;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn registration_triggers_collection() {
```
cijothomas, line 419: not sure I follow the name of the test, as the trigger for collection is the time interval only...
The test is validating that, during collection, triggered by time interval, the observable's callback(s) are invoked.

izquierdo: The aim is to test that creating the `SdkMeterProvider` instance triggers the start of the worker that will periodically collect (and export) metrics. This is the line being tested:

```rust
        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
```

Perhaps a better name would be `registration_triggers_worker`?

Using the observable callback is a roundabout way to confirm that the worker thread is running. Maybe better to use `meter_provider.force_flush()` and `exporter.get_finished_metrics()` instead, as in [other tests](https://github.com/open-telemetry/opentelemetry-rust/blob/3f327a1ecbe9db3ad67323565c3347884ea5a2d3/opentelemetry-sdk/src/metrics/mod.rs#L407-L412)? I also want to avoid test flakiness here.

cijothomas: calling force_flush() would anyway flush the metric, even without the worker starting?... I need check in detail, but this testing logic sounds good to me.

cijothomas: builder_triggers_periodicreader_worker - as alternative suggestion. Or you can just add a code comment with your comment, as that was good explanation.

cijothomas (approved):
Thanks. Left a suggestion about rename and/or leaving a code comment to make it clear what are we testing with this particular test, given it may not be very obvious.

opentelemetry-sdk/src/metrics/periodic_reader.rs
lalitb: > Using the observable callback is a roundabout way to confirm that the worker thread is running.

I believe this test looks good to validate the `periodic exporter metric collection cycle` through the `mpsc` channel. We don't need to validate the measurements and their aggregation, as there are other tests for the same.

lalitb (approved):

PR #1523, opened 2024-02-11
fix: don't return empty string when trace header is not present

### LINKED ISSUE #1475: [Feature]: Avoid parsing empty strings into `TraceId`
Missing trace state value is unwrapped to into an empty string to then be parsed rather than directly returning `TraceState::default()` which internally just stores a None.

https://github.com/open-telemetry/opentelemetry-rust/blob/27d338dd954ef7681242ead47a8a61c964dcc428/opentelemetry-sdk/src/propagation/trace_context.rs#L93-L95

Maybe this doesn't matter given that `FromStr` is pretty lightweight. I don't think its use of empty Vecs would require any allocation anyways 🤷

https://github.com/open-telemetry/opentelemetry-rust/blob/27d338dd954ef7681242ead47a8a61c964dcc428/opentelemetry/src/trace/span_context.rs#L401-L406

TommyCpp:
Fixes #1475

## Changes

- if trace header is not present. Just return default 


hdost (approved):
Looks good! are there tests we want to add?

lalitb (approved):

Issue #1124, opened 2023-06-21
Performance: send resource data to span processors once, not for every span when it ends


shaun-cox:
[`sdk::trace::Span::ensure_ended_and_exported`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/span.rs#LL192C20-L192C56) clones the provider's config resource for every span processor and every span before sending it to the span processor.

But since the resource is unique to the provider which owns all the span processors, this data could be sent ahead of time, just once, to each span processor.

Not only would this reduce CPU consumption by not cloning, but it will also reduce the size of each message sent through a span processor channel by about 64 bytes (estimated) which will save CPU all by itself.

(I can work on a PR for this but wanted to hear of any objections here before starting that work.)

cijothomas:
+1
The same idea could be used in other signals as well (logs, metrics).

lalitb:
I may be missing something basic here, but can't we have [build_export_data](https://github.com/open-telemetry/opentelemetry-rust/blob/320957751e688ca8a27f60a0c51f46f1c3fb08b8/opentelemetry-sdk/src/trace/span.rs#L218) take reference to the Resource object, just like it does for Tracer  ? Probably I am thinking C++ way, and rust lifetime scoping makes it difficult to pass reference here :)

djc:
By "sent", do you mean it would just be passed at initialization? Or do we need some other (more clever) path to pass it?

Anyway, this sounds like it could make a lot of sense.

shaun-cox:
> By "sent", do you mean it would just be passed at initialization?

At provider initialization, yes, but not technically at span processor initialization, as that happens while [`sdk::trace::provider::Builder`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/provider.rs#L156-L218) is accumulating span processors.  Once `build` is invoked to create the `TraceProvider`, we'd have to invoke a new method on `SpanProcessor` to pass the configuration's `Resource`. The `SpanProcessor` implementations will then "send" it to their exporter counter-part on the receive end of the channel separating processor from exporter.

(Hope that makes more sense. In hindsight, the title of this issue should have said "send resource data to exporter once", or "send resource data through span processor to exporter".)

shaun-cox:
First, I'm going to use the following to disambiguate *exported* `SpanData` from *core* `SpanData`:
[...]

PR #1526, opened 2024-02-12
Optimize Resource Sharing Across Exporters with Arc Implementation

lalitb:
## Changes

This PR introduces a refactor of the `Resource` structure to improve the efficiency of resource sharing across exporters for owned resource attributes. By encapsulating the Resource data within a `ResourceInner` struct and wrapping it with an Arc (Atomic Reference Counting), we eliminate the need for deep copies of resource data, thereby reducing memory overhead and improving performance.

### Existing
```rust
#[derive(Clone, Debug, PartialEq)]
pub struct Resource {
    attrs: HashMap<Key, Value>,
    schema_url: Option<Cow<'static, str>>,
}
```
### PR
```rust
/// Inner structure of `Resource` holding the actual data.
/// This structure is designed to be shared among `Resource` instances via `Arc`.
struct ResourceInner {
    attrs: HashMap<Key, Value>,
    schema_url: Option<Cow<'static, str>>,
}

/// An immutable representation of the entity producing telemetry as attributes.
/// Utilizes `Arc` for efficient sharing and cloning.
#[derive(Clone, Debug, PartialEq)]
pub struct Resource {
    inner: Arc<ResourceInner>,
}
```
The stress test results (default has 4 resource attributes) for the main and PR branch. The results are more pronounced if the resource is added with `LoggerProviderBuilder::with_resource` and `TracerProviderBuilder::with_resource` methods as indicated [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1526#issuecomment-1939725867), as the resource created this way are owned, and can't be optimized by the compiler.

##### main
Number of threads: 10
Throughput: 8,584,000 iterations/sec
Throughput: 8,560,000 iterations/sec
Throughput: 8,793,800 iterations/sec
Throughput: 8,465,200 iterations/sec

#### PR
Number of threads: 10
Throughput: 9,059,800 iterations/sec
Throughput: 8,719,200 iterations/sec
Throughput: 8,831,400 iterations/sec
Throughput: 8,767,800 iterations/sec

#### main
Number of threads: 10
Throughput: 23,298,600 iterations/sec
Throughput: 21,358,600 iterations/sec
Throughput: 21,797,000 iterations/sec

#### PR

Number of threads: 10
Throughput: 25,054,600 iterations/sec
Throughput: 24,265,000 iterations/sec
Throughput: 24,576,600 iterations/sec
Throughput: 23,008,000 iterations/sec
```

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -49,6 +49,10 @@ Optimize Resource Sharing Across Exporters with Arc Implementation
   - `LoggerProviderInner` is no longer `pub (crate)`
   - `Logger.provider()` now returns `&LoggerProvider` instead of an `Option<LoggerProvider>`
 
+- [1519](https://github.com/open-telemetry/opentelemetry-rust/pull/1519) Performance improvements 
+    when calling `Counter::add()` and `UpDownCounter::add()` with an empty set of attributes
+    (e.g. `counter.Add(5, &[])`)
+    
 ### Fixed
 
 - [#1481](https://github.com/open-telemetry/opentelemetry-rust/pull/1481) Fix error message caused by race condition when using PeriodicReader
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -8,6 +8,9 @@
 - [#1471](https://github.com/open-telemetry/opentelemetry-rust/pull/1471) Configure batch log record processor via [`OTEL_BLRP_*`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#batch-logrecord-processor) environment variables and via `OtlpLogPipeline::with_batch_config`
 - [#1503](https://github.com/open-telemetry/opentelemetry-rust/pull/1503) Make the documentation for In-Memory exporters visible.
 
+- [#1526](https://github.com/open-telemetry/opentelemetry-rust/pull/1526) 
+Optimize Resource Sharing Across Exporters with Arc Implementation
+
 ### Changed
 
 - **Breaking**
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -46,6 +46,10 @@
   - `LoggerProviderInner` is no longer `pub (crate)`
   - `Logger.provider()` now returns `&LoggerProvider` instead of an `Option<LoggerProvider>`
 
+### Fixed
+
+- [#1481](https://github.com/open-telemetry/opentelemetry-rust/pull/1481) Fix error message caused by race condition when using PeriodicReader
+
 ## v0.21.2
 
 ### Fixed
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -9,7 +9,7 @@
 - [#1503](https://github.com/open-telemetry/opentelemetry-rust/pull/1503) Make the documentation for In-Memory exporters visible.
 
 - [#1526](https://github.com/open-telemetry/opentelemetry-rust/pull/1526) 
-Optimize Resource Sharing Across Exporters with Arc Implementation
+Performance Improvement : Creating Spans and LogRecords are now faster, by avoiding expensive cloning of `Resource` for every Span/LogRecord.
 
 ### Changed
 
--- a/opentelemetry-sdk/benches/log.rs
+++ b/opentelemetry-sdk/benches/log.rs
@@ -39,12 +39,6 @@ fn log_benchmark_group<F: Fn(&dyn Logger)>(c: &mut Criterion, name: &str, f: F) 
     group.bench_function("with-context", |b| {
         let provider = LoggerProvider::builder()
             .with_simple_exporter(VoidExporter)
-            .with_config(Config::default().with_resource(Resource::new(vec![
-                KeyValue::new("service.name", "my-service"),
-                KeyValue::new("service.version", "1.0.0"),
-                KeyValue::new("service.environment", "production"),
-                KeyValue::new("service.instance.id", "1234"),
-            ])))
             .build();
 
         let logger = provider.logger("with-context");
--- a/opentelemetry-sdk/benches/log.rs
+++ b/opentelemetry-sdk/benches/log.rs
@@ -7,10 +7,11 @@ use criterion::{criterion_group, criterion_main, Criterion};
 use opentelemetry::logs::{AnyValue, LogRecord, LogResult, Logger, LoggerProvider as _, Severity};
 use opentelemetry::trace::Tracer;
 use opentelemetry::trace::TracerProvider as _;
-use opentelemetry::Key;
+use opentelemetry::{Key, KeyValue};
 use opentelemetry_sdk::export::logs::{LogData, LogExporter};
-use opentelemetry_sdk::logs::LoggerProvider;
+use opentelemetry_sdk::logs::{Config, LoggerProvider};
 use opentelemetry_sdk::trace::{config, Sampler, TracerProvider};
+use opentelemetry_sdk::Resource;
 
 #[derive(Debug)]
 struct VoidExporter;
@@ -27,6 +28,12 @@ fn log_benchmark_group<F: Fn(&dyn Logger)>(c: &mut Criterion, name: &str, f: F) 
 
     group.bench_function("no-context", |b| {
         let provider = LoggerProvider::builder()
+            .with_config(Config::default().with_resource(Resource::new(vec![
+                KeyValue::new("service.name", "my-service"),
+                KeyValue::new("service.version", "1.0.0"),
+                KeyValue::new("service.environment", "production"),
+                KeyValue::new("service.instance.id", "1234"),
+            ])))
             .with_simple_exporter(VoidExporter)
             .build();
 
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

**Comment by cijothomas** [comment]
Can you also add benchmarks to logs, so we can how much this would improve the common use case (like 5-10 Resource attributes)? 
(Would be good to share the stress test for the 5-10 attributes, to see how much this impacts normal usages.)

**Comment by lalitb** [comment]
Just to add, the stress test uses NoOpLogProcessor, so the performance is measured for creating the LogData structure for each event, and not processing by exporter.

> (Would be good to share the stress test for the 5-10 attributes, to see how much this impacts normal usages.)

With PR, the throughput doesn't change as we increase the number of attributes, as we are just doing shared resource increment and no real data copy. However, with existing code, the throughput reduces drastically as we increase the resource attributes.

**5 resource attributes:**

*main:*
Number of threads: 8
Throughput: 12,364,800 iterations/sec
Throughput: 12,177,400 iterations/sec
Throughput: 12,345,800 iterations/sec

*PR:*
Number of threads: 8
Throughput: 22,066,200 iterations/sec
Throughput: 22,142,600 iterations/sec
Throughput: 21,524,400 iterations/sec

**10 resource attributes:**
*main:*
Number of threads: 8
Throughput: 8,696,200 iterations/sec
Throughput: 8,750,200 iterations/sec
Throughput: 8,767,400 iterations/sec
Throughput: 8,888,200 iterations/sec

*PR:*
Number of threads: 8
Throughput: 22,376,000 iterations/sec
Throughput: 21,866,200 iterations/sec
Throughput: 22,334,200 iterations/sec


**100 resource attributes:**
*main:*
Number of threads: 8
Throughput: 1,357,600 iterations/sec
Throughput: 1,324,000 iterations/sec
Throughput: 1,353,000 iterations/sec

*PR:*
Number of threads: 8
Throughput: 22,715,400 iterations/sec
Throughput: 21,450,800 iterations/sec
Throughput: 22,922,000 iterations/sec


> Can you also add benchmarks to logs,

Modified the benchmark to include attributes in common scenario:
```rust
                KeyValue::new("service.name", "my-service"),
                KeyValue::new("service.version", "1.0.0"),
                KeyValue::new("service.environment", "production"),
                KeyValue::new("service.instance.id", "1234"),
```

The results:
*main:*

simple-log/no-context   time:   [254.13 ns 255.45 ns 257.16 ns]
simple-log/with-context time:   [251.62 ns 252.16 ns 252.76 ns]
simple-log-with-int/no-context   time:   [299.27 ns 300.59 ns 302.25 ns]
simple-log-with-int/with-context   time:   [302.25 ns 302.83 ns 303.50 ns]

*PR:*

simple-log/no-context   time:   [167.73 ns 168.10 ns 168.54 ns]
simple-log/with-context time:   [162.33 ns 162.68 ns 163.13 ns]
simple-log-with-int/no-context time:    time:   [224.12 ns 225.28 ns 226.71 ns]
simple-log-with-int/with-context:  time:   [218.04 ns 218.28 ns 218.55 ns]

opentelemetry-sdk/benches/log.rs
```diff
@@ -27,6 +28,12 @@ fn log_benchmark_group<F: Fn(&dyn Logger)>(c: &mut Criterion, name: &str, f: F)
 
     group.bench_function("no-context", |b| {
         let provider = LoggerProvider::builder()
+            .with_config(Config::default().with_resource(Resource::new(vec![
```
cijothomas, line 31: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/config.rs#L19-L20

Wondering why config stores `Resource` inside a `Cow` ?

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -34,13 +34,22 @@ use opentelemetry::{Key, KeyValue, Value};
 use std::borrow::Cow;
 use std::collections::{hash_map, HashMap};
 use std::ops::Deref;
+use std::sync::Arc;
 use std::time::Duration;
 
+/// Inner structure of `Resource` holding the actual data.
```
cijothomas, line 40: while this does improve the perf very nicely, I think we should further explore other ideas from https://github.com/open-telemetry/opentelemetry-rust/issues/1124

Even with this PR, we are still passing the Resource to each span, for every invocation. (for cases like simple/re-entrant processor, this is still unacceptable). We should explore is Processors/Exporters, who cares about Resource, can ask its provider for the Resource, if it wants, instead of passing it every time. 

This approach is used in OTel .NET : https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.Console/ConsoleActivityExporter.cs#L115

Not a blocker for this PR, but this is definitely something worth continuing to fine tune further.

opentelemetry-sdk/benches/log.rs
TommyCpp: It's to avoid performance overhead. `Cow` provides cheap copy until the data changes or an ownership is required

opentelemetry-sdk/src/resource/mod.rs
lalitb: > Even with this PR, we are still passing the Resource to each span, for every invocation

Resource is only passed to processor after cloning, as per the code here - https://github.com/open-telemetry/opentelemetry-rust/blob/3f327a1ecbe9db3ad67323565c3347884ea5a2d3/opentelemetry-sdk/src/trace/span.rs#L206

hdost: > Even with this PR, we are still **passing the Resource to each span**, for every invocation. (for cases like simple/re-entrant processor, this is still unacceptable). We should explore if Processors/Exporters, who cares about Resource, can ask its provider for the Resource, if it wants, instead of passing it every time. 
> 

Yea this seems like it would be a thing to remove. Especially since it's something that is configured per provider anyhow.

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -105,8 +134,11 @@ impl Resource {
         for detector in detectors {
             let detected_res = detector.detect(timeout);
             for (key, value) in detected_res.into_iter() {
-                // using insert instead of merge to avoid clone.
-                resource.attrs.insert(key, value);
+                // This call ensures that if the Arc is not uniquely owned,
+                // the data is cloned before modification, preserving safety.
+                // If the Arc is uniquely owned, it simply returns a mutable reference to the data.
```
hdost, line 139: This is a constructor, at what point would the inner not be uniquely owned?
That being said is there a reason we shouldn't put the `make_mut` before the loop?

hdost (commented):
Excellent gains ☺️ thanks again.

opentelemetry-sdk/src/resource/mod.rs
cijothomas: > essor after cloning, as per the code here -

Yes. What I was saying was - there is no need to even pass the cloned one (however cheap cloning is) to the processors each time! The size of the struct passed to the processors can be reduced by chosing not to include Resource in it!.

Agree this is a follow up candidate.

opentelemetry-sdk/src/resource/mod.rs
lalitb: I believe it would be safe to move `make_mut` before loop at this point. The `inner` is ref-incremented only during LogData/SpanData creation.

hdost (approved):
Thank you for the work again!

TommyCpp (approved):
Great performance improvements!

opentelemetry-sdk/benches/log.rs
cijothomas, line 31: @lalitb I think we can the benchmarks as is because there is default Resource added anyway with 4 attributes. And show the gains for the default case. Also include a changelog entry, so folks upgrading knows that the upcoming version has noticeable perf gains.

opentelemetry-sdk/src/resource/mod.rs
cijothomas: @TommyCpp also shared possibly only passing the Tracer, from which instrumentation-scope can be obtained, and from which Resource can also be obtained by first obtaining the TracerProvider. Will table this and come back later.

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -92,8 +108,21 @@ impl Resource {
         KV: IntoIterator<Item = KeyValue>,
         S: Into<Cow<'static, str>>,
     {
+        let schema_url_str = schema_url.into();
```
cijothomas, line 111: let schema_url_str = schema_url.into();
 let normalized_schema_url = if schema_url_str.is_empty() {
            None
        } else {
            Some(schema_url_str)
        };
        let mut attrs = HashMap::new();
        for kv in kvs {
            attrs.insert(kv.key, kv.value);
        }

        Resource {
            inner: Arc::new(ResourceInner {
                attrs,
                schema_url: normalized_schema_url ,
            }),
        }

^ Isn't this easier, as opposed to created a Resource with just attributes, and then cloning it to form a new one with schema_url added ? 

(I might be missing something obvious....)

cijothomas (approved):
LGTM.
Before merge, I'd suggest to add changelog entry to let users know of the perf gains.
Also, please update PR description with before/after numbers.

opentelemetry-sdk/src/resource/mod.rs
lalitb: Good comment. Updated accordingly.

opentelemetry-sdk/benches/log.rs
lalitb: Done.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 - [#1471](https://github.com/open-telemetry/opentelemetry-rust/pull/1471) Configure batch log record processor via [`OTEL_BLRP_*`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#batch-logrecord-processor) environment variables and via `OtlpLogPipeline::with_batch_config`
 - [#1503](https://github.com/open-telemetry/opentelemetry-rust/pull/1503) Make the documentation for In-Memory exporters visible.
 
+- [#1526](https://github.com/open-telemetry/opentelemetry-rust/pull/1526) 
+Optimize Resource Sharing Across Exporters with Arc Implementation
```
cijothomas, line 12:
```suggestion
Performance Improvement : Creating Spans and LogRecords are now faster, by avoiding expensive cloning of `Resource` for every Span/LogRecord.
```

Issue #1427, opened 2023-12-05
[Feature]: Port opentelemetry-http to hyper 1.x


Cogitri:
### Related Problems?

https://github.com/hyperium/tonic/issues/1307

### Describe the solution you'd like:

Hyper 1.0 has been released and as such it'd be nice to upgrade opentelemetry-http's `HttpClient` to the new version

TommyCpp:
With the upgrade I think we should also upgrade the version of `http` to `1.0`. However it's currently blocked because some of the http client implementation is still on pre 1.0 and having both `http 1.0` create and `http 0.2` create caused some issues.

We can either

- Delete all http client that hasn't support http 1.0 yet and upgrade the `opentelemtry-http` create to use the `http 1.0` types. This will require some changes on downsteam create uses those http client defined in `opentelemetry-http`.
- Wait until http clients to upgrade to http 1.0. However, some clients hasn't upgrade in a while so I am not sure if they will upgrade to 1.0 soon

lalitb:
As I understand in the community meeting this week, it was agreed to go with the first option above i.e., delete all the HTTP clients not supporting HTTP 1.0, and bring them back once the support is there.

TommyCpp:
Other than the http cteates. Tonic also depends on the `http 0.2`. We are currently blocked on it https://github.com/hyperium/tonic/issues/1579

hdost:
cc: @KallDrexx

KallDrexx:
After going through this upgrade for our project, I would definitely advise waiting on this until when/if tonic updates.  I had to abort our upgrade for now because of the tonic dependency, and couldn't get them to work side by side easily, at least in the few hours I tried.

KallDrexx:
I take that back.  It took some figuring out but in our product I was able to get Tower middle ware with hyper 1.0 and http 1.0 to work with tonic 0.2.   So it looks like it should be possible to support both and not have to wait for tonic's upgrade (which doesn't seem to be progressing very fast).

hdost:
Would it need to be in the form of a temporary feature or something like that?

TommyCpp:
I can look into the hyper middle ware or if you want feel free to take this issue @KallDrexx

KallDrexx:
[...]

PR #1528, opened 2024-02-13
Add basic doc comments for opentelemetry-http

cijothomas:
Basic doc comments, as it was never clear what was the purpose of opentelemetry-http crate.
Related issue https://github.com/open-telemetry/opentelemetry-rust/issues/1427 (This PR is in no way fixing that issue!)

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -9,8 +9,8 @@ use opentelemetry::propagation::{Extractor, Injector};
 
 /// Helper for injecting headers into HTTP Requests. This is used for OpenTelemetry context
 /// propagation over HTTP.
-/// See https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/tracing-http-propagator/README.md
-/// for an example of how to use this.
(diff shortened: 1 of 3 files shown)


opentelemetry-http/src/lib.rs
```diff
@@ -7,6 +7,10 @@ pub use bytes::Bytes;
 pub use http::{Request, Response};
 use opentelemetry::propagation::{Extractor, Injector};
 
+/// Helper for injecting headers into HTTP Requests. This is used for OpenTelemetry context
```
cijothomas, line 10: From a quick glance, this can be moved out to its own crate, as this does not require any particular httpclient implementation, so can move quickly. 

Would it help anyone if we move this to own crate? I guess there are users who need this part (for context propagation over http), but does not need an exporting mechanism over http (as they can use alternatives like gRPC)?

lalitb (approved):

opentelemetry-http/src/lib.rs
cijothomas: @TommyCpp suggested its better to keep crate count low, but we can achieve this via feature flags.. Will have to follow up separately for that.

PR #1534, opened 2024-02-15
Remove the option of using grpcio for gRPC

### LINKED ISSUE #1524: [Feature]: remove the grpcio gRPC backend
### Related Problems?

_No response_

The grpcio crate seems to be mostly used for internal TiKV purposes, and as a result doesn't get much maintenance. While it finally gained an option to use prost as a codec, apparently its prost update has gotten [stuck](https://github.com/tikv/grpc-rs/pull/636) for six months, making an upgrade to prost 0.12 and recent tonic harder. Given the amount of maintenance this project's dependencies are getting and given the limited popularity of the grpcio implementation (50k downloads in past 90 days vs 7.9M for tonic in the same time frame), I propose we remove the grpcio implementation from the opentelemetry core crates.

I would be willing to do the legwork if necessary.

### Considered Alternatives

Alternatively, maybe there is some way to cut the grpcio dependencies off from the other crates in a way that it doesn't block tonic work? But if none of the maintainers is willing to champion doing the work to keep grpcio going, I think removing support for it is the better choice.

### Additional Context

_No response_

djc:
Fixes #1524


shaun-cox (approved):
Thanks for doing this!

cijothomas (approved):
LGTM.
Can you add an entry in changelog.md for the opentelemetry-otlp

djc:
Added changelog entries in -proto and -otlp.

PR #1540, opened 2024-02-16
Add markdown link CI check

lalitb:
## Changes

Shamelessly copied from https://github.com/open-telemetry/opentelemetry-specification/blob/5d9cef817415a15aaa4883e184a38adec79096ed/.github/workflows/checks.yaml#L36, and fixed broken links.

This will recursively check all the markdown files (ignoring CHANGELOG.md), and fail if any dead link is found.

If a CI check isn't necessary, I can modify the PR to exclude it, focusing solely on the updates to the broken links.


scripts/markdown-link-check-config.json
```diff
@@ -0,0 +1,10 @@
+{
+"httpHeaders": [
+    {
+      "urls": ["https://crates.io"],
+      "headers": {
+        "Accept": "text/html"
```
lalitb, line 6: HTTP request on crates.io fails with a `404` error, if the `Accept: text/html` header is not specified.

opentelemetry-sdk/README.md
```diff
@@ -34,3 +34,8 @@ From the root directory, run the following command:
 ```sh
 cargo bench
 ```
+
+## Supported Rust Versions
```
cijothomas, line 38: we need to put it in main readme, and everywhere else can just point to it

.github/workflows/markdown-link-check.yml
```diff
@@ -0,0 +1,25 @@
+name: Markdown link check
+
+on:
+  pull_request:
```
cijothomas, line 4: maybe once a week is fine?

opentelemetry-sdk/README.md
lalitb: Yeah I thought so, but then kept it consistent with other crates from this repo. 

https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-rust+%22Supported+Rust+Versions%22&type=code

.github/workflows/markdown-link-check.yml
lalitb: As of now it only runs for PRs having markdown changes and takes ~40 secs to complete. But no strong opinion, once a week works fine too if we all agree :)

TommyCpp: if it's ~40 sec maybe we can just run it for every PR that touches markdown?

cijothomas (approved):

PR #1553, opened 2024-02-17
Safer endpoint Uri construction

### LINKED ISSUE #997: opentelemetry-otlp: '/v1/traces' is not appended to endpoint for http exporter

jaymell:
Make endpoint Uri construction less fragile by stripping out duplicate slashes.

Fixes #997 

## Changes

The functionality to append the signal path to a user-supplied tracing endpoint is already in place. However, it does so in a way that is likely to break for any user who's passing the string value of a `Uri` or `Url`, which will have a trailing slash appended to them. This attempts to fix that issue.

There were other possible changes discussed but ever implemented/merged in #1056. This PR attempts to keep it simple by just changing existing behavior to not break in a common use case.


hdost (approved):
Thank you for the contribution!

lalitb (approved):

TommyCpp (approved):

Issue #1326, opened 2023-10-31
Release documentation and process


TommyCpp:
We need to document the release process to help us release more frequently. We should have a step-by-step guide to release a create

TommyCpp:
#1327 should help

jtescher:
Working on getting github group access for @open-telemetry/rust-maintainers on crates.io, at that point we can document how any maintainer can run the release process.

hdost:
Update from the SIG meeting @jtescher is working with Community Managers to allow access so that the github group can be used to add users to crates.io

hdost:
I saw just a couple days ago that the org should now be readable by crates.io

hdost:
The release process could be something like:

- Check pending PRs and validate if they should be a blocker for a next release (merge as necessary)
- Create a release PR 
  - For each crate 
    - Bump appropriate version 
    - Update change logs to reflect 
    - update API/SDK version as necessary 
- Merge the PR
  - Get reviews from other Maintainers 
  - Ensure that there haven't been any interfering PRs 
- Tag the release commit (TBD convention)
- Create Github Release 
- Publish to crates.io using the version as of the release commit 
  - TBD include steps or links `cargo release`
- Post to #otel-rust on CNCF Slack.

cijothomas:
We have the process documented and have been following it for last few releases without trouble. Closing this issue.

PR #1555, opened 2024-02-18
Add RELEASING.md documentation.

hdost:
While we try to de-centralize the release process we should establish consistent practice.

Relates #1326


RELEASING.md
```diff
@@ -0,0 +1,51 @@
+# Releasing OpenTelemetry Rust
+
+The primary audience for this is the SIG Maintaners. It provides the list of steps for how to release the crates and the
```
cijothomas, line 3:
```suggestion
The primary audience for this is the SIG Maintainers. It provides the list of steps for how to release the crates and the
```

RELEASING.md
```diff
@@ -0,0 +1,51 @@
...
+  - Ensure that there haven't been any interfering PRs
+3. Tag the release commit based on the [tagging convention](#tagging-convention)
+4. Create Github Release
+5. [Publish](#publishing-crates) to crates.io using the version as of the release commit
+6. Post to [#otel-rust](https://cloud-native.slack.com/archives/C03GDP0H023) on CNCF Slack.
+
+
+## Tagging Convention
+
+For each crate: it should be `<crate-name>-<version>` `<version>` being the simple `X.Y.Z`.
```
cijothomas, line 43: should we consider aligning the same version for all crates? We moved all non-core crates to contrib, so should be easier if there is a single version for all core crates.?

RELEASING.md
```diff
@@ -0,0 +1,51 @@
...
+* Bring it up at a SIG meeting, this can usually get some of these questions answered sooner than later. It will also
+  help establish a person to perform the release. Ideally this can be someone different each time to ensure that the
+  process is documented.
+
+## Steps to Release
+
+1. Create a release PR
+  - For each crate
+    - Bump appropriate version
+    - Update change logs to reflects
```
cijothomas, line 29:
```suggestion
    - Update change logs to reflect release date
```

Was that the intention?

RELEASING.md
```diff
@@ -0,0 +1,51 @@
...
+    - Update API/SDK version as necessary
+  - If there's a large enough set of changes consider writing a migration guide.
+2. Merge the PR
+  - Get reviews from other Maintainers
+  - Ensure that there haven't been any interfering PRs
+3. Tag the release commit based on the [tagging convention](#tagging-convention)
```
TommyCpp, line 35:
```suggestion
3. Tag the release commit based on the [tagging convention](#tagging-convention). It should usually be a bump on minor version before 1.0
```

RELEASING.md
```diff
@@ -0,0 +1,51 @@
+# Releasing OpenTelemetry Rust
+
+The primary audience for this is the SIG Maintaners. It provides the list of steps for how to release the crates and the
+considerations to make before releasing the create. It may provide use to consumers of the crate if/when we develop a
```
lalitb, line 4:
```suggestion
considerations to make before releasing the crate. It may provide use to consumers of the crate if/when we develop a
```

RELEASING.md
```diff
@@ -0,0 +1,51 @@
...
+release cadence.
+
+## Release cadence
+
+As of Feb 2024, there is no established cadence for the OpenTelemetry crates. The balance is required between too many
+breaking changes in a single release, and since we have instability flipping between implementations across 0.x
+releases. Perhaps once we've established a 1.0 we can consider a consistent release cadence.
```
lalitb, line 11:
```suggestion
releases. Perhaps once we've established a 1.0 we can consider adopting a consistent release cadence.
```

RELEASING.md
```diff
@@ -0,0 +1,51 @@
...
+  process is documented.
+
+## Steps to Release
+
+1. Create a release PR
+  - For each crate
+    - Bump appropriate version
+    - Update change logs to reflects
+    - Update API/SDK version as necessary
+  - If there's a large enough set of changes consider writing a migration guide.
```
lalitb, line 31:
```suggestion
  - If there's a large enough set of changes, consider writing a migration guide.
```

RELEASING.md
hdost: Meant to say release version.

RELEASING.md
hdost: Pro: makes releasing simpler 
Con: Potentially creating more noise on the releases post 1.0

Not opposed in the short term. Perhaps we start by re-aligning them and we can revisit after 1.0 if it becomes more painful. I suppose after 1.x if nothing changes in a version we can just skip a version for that particular crate.

hdost: Serial Version Pros
- Follows general convention 
- Avoids crate consumer confusion
- Removes dependence from api/sdk on exporter releases. A case where an exporter needs a breaking change without breaking API for example. 

Unified Version Pros
- Simplifies releasing for maintainers, maybe? 


Common Considerations 
- You still need to think about which crates to release. (This assumes we don't do worthless releases, which I'm not in favor of)

I could be swayed either way. Looking at it again I don't really see the advantage of the unified versioning. 

Am I missing any points? I can update my post.

hdost: Cc: @djc

djc: Yeah, I wouldn't do unified versioning (instead documing a common reference point). Also, chances are it won't be a big deal after 1.0? Most crates will be at 1.0 together and will only evolve slowly from there.

RELEASING.md
djc: Note that for a bunch of these crates, semver stability might be limited by dependencies that appear in the public API.

RELEASING.md
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/tags We never used to push tags for each crate separately, so in some way, we were following the unified versioning (in terms of git tag only!)

hdost: I realize that currently we only tag API SDK , but that's the part I was looking to change.

hdost: From my perspective right now I'd like to document our current process, and we can approach the versioning question later. After the SIG we discussed this I believe since we didn't have a specific consensus we should continue with operating as we already do.

TommyCpp (approved):
I think it's a good start point. For next step we should have a script that:

1. automaticaly validates the versions(maybe by querying crates.io?)
2. help publish each create instead of having releaser does it one by one

lalitb (approved):

cijothomas (approved):
lets merge the current process, and then iterate and improve it over next few releases.

Issue #1509, opened 2024-02-03
[Bug]: constructing a tonic builder defaults to connect to 4318?


colemickens:
### What happened?

I thought I saw the issue in the code last night, but now I'm not convinced I quite grok the issue, but here's what I'm seeing:

As part of my telemetry pipeline creation, I do this:
```
                let mut tonic = opentelemetry_otlp::new_exporter().tonic();
                let exporter = SpanExporterBuilder::Tonic(tonic);
```

However, this tries to connect to 4318 by default, rather than 4317 as I'd expect.

```
// Endpoints per protocol https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md
const OTEL_EXPORTER_OTLP_GRPC_ENDPOINT_DEFAULT: &str = "http://localhost:4317";
const OTEL_EXPORTER_OTLP_HTTP_ENDPOINT_DEFAULT: &str = "http://localhost:4318";
```

Right now I have to do this:
```
                let mut tonic = opentelemetry_otlp::new_exporter().tonic();
                tonic.export_config().endpoint = "http://localhost:4317".to_string();
                let exporter = SpanExporterBuilder::Tonic(tonic);
```

but it doesn't feel right that I need to?

### API Version

```
opentelemetry_sdk = { version = "0.21.2", features = ["rt-tokio"] }
opentelemetry-otlp = { version = "0.14", features = [
  "http-proto",
  "reqwest-client",
  "trace" # TODO: ensure we need this
] }
opentelemetry = "0.21"
tracing-opentelemetry = "0.22"
```

### SDK Version

(see above)

### What Exporters are you seeing the problem on?

OTLP

### Relevant log output

```shell
DEBUG connection error: not connected
DEBUG resolving host="localhost"
DEBUG connecting to [::1]:4318
DEBUG connected to [::1]:4318
DEBUG binding client connection
DEBUG client connection bound
DEBUG send frame=Settings { flags: (0x0), enable_push: 0, initial_window_size: 2097152, max_frame_size: 16384 }
DEBUG service.ready=true processing request
DEBUG Connection: Connection::poll; connection error error=GoAway(b"", FRAME_SIZE_ERROR, Library) peer=Client
DEBUG Connection: send frame=GoAway { error_code: FRAME_SIZE_ERROR, last_stream_id: StreamId(0) } peer=Client
[...]

PR #1556, opened 2024-02-18
opentelemetry-otlp: Default the port correctly.

hdost:
Current behavior uses http-proto port even when there's no desire to use it.

Relates #1509


TommyCpp (approved):
Should we add a unit test to prevent future regression?

lalitb (approved):
LGTM. CI test is failing for missing import.

hdost:
Tests added feel free to merge.

PR #1539, opened 2024-02-16
Bump versions to prepare for release v0.22

djc:
Hope this helps. Is the release process written down anywhere?

PR #1564, opened 2024-02-21
Adding two level hashing in metrics hashmap 

lalitb:
## Changes

This PR is opened just for the discussion, of one of the possible changes to improve the performance of concurrent recording/updation of measurements in the hot-path. In today's (Feb 20th, 2024) community meeting, we discussed multiple approaches 
- Using [`dashmap`](https://crates.io/crates/dashmap) - sharded hashmap. 
- Using [`flurry`](https://crates.io/crates/flurry) - uses fine-grained atomic locks in hashmap
- Using [`sharde-slab`](https://crates.io/crates/sharded-slab) - building hashmap using concurrent data-structures provided by this crate.
- Using `thread-local` hashmaps in the hot-path to record measurements and then merge them during collection. The collection would be a performance-heavy process. And the SDK doesn't have control over the life cycle of threads created by the application, so exiting a thread by the application will also remove all it's aggregated data.
- Cijo spoke about the lock-free approach for hot-path as used in otel-dotnet, where the first level of hashmap only stores the index to the aggregated data. 

The approach in this PR is modifying the `ValueMap` store values in two-level hashing to minimize lock contention. It's basically a more simpler form of sharding as provided by `dashmap`.

Existing:

```rust
    values: Mutex<HashMap<AttributeSet, T>>, 
}
```

PR:
```
    buckets: Arc<HashMap<u8, Mutex<HashMap<AttributeSet, T>>>>, // ? or RWLock
    total_count: AtomicUsize,
}
```

First level of hashing is to distribute the values across the fixed set of 256 buckets.  Each bucket is guarded by its own mutex and contains a second-level hash map for storing AttributeSet to aggregation mapping. 

Please note
 - This approach won't see any improvement in single-threaded application, as we still lock. Infact, the complexity added by sharding may slightly impact the performance. I didn't see any perf improvement/degradation in benchmark tests as they are single threaded.
 - If all the attributes are hashed to the same bucket ( rare scenario), there won't be any perf improvement as all the threads will contest for the same lock.
 - This approach may impact [`locality of reference`](https://en.wikipedia.org/wiki/Locality_of_reference) during collection cycle, as the OS can allocate these buckets in different segments of memory, and so mayn't be cached together.
 
All above are the limitation of any sharded hashmap (including `dashmap`).

This is the result of our metrics stress test in my machine. However, I don't think results are really consistent across machines, so good if someone can test it too:

| Threads  | Main Throughput (millions/sec)   | PR  Throughput (millions/sec)  | PR Throughput (millions/sec) with hashbrown |
| ---|---|---|---|
| 3  | 6.6  | 6.6   | 6.8
| 5  | 10.1  |12.5   | 13.2
| 8 | 5.6  |  18.9 | 20.0
| 10 |  4.5 | 22.0  |  24.0

Based on the above results- with PR, the perf seems to be increasing significantly with the number of threads. In the main branch, the performance increases with threads till a threshold and then starts degrading.

Benchmark result. This PR won't improve performance there, as these are single-threaded tests. There is slight retrogress for two-level indirection, which gets compensated (in fact improved) with hashbrown/ahash:

| Test | time (Main branch) | time (PR branch) | time (PR with hashbrown) |
| ---|---|---|---|
| AddNoAttrs | 30.994 ns  | 31.286 ns  |31.544 ns
|AddNoAttrsDelta  | 31.009 ns  |31.009 ns   | 31.455 ns
| AddOneAttr | 169.76 ns |  184.00 ns| 154.39 ns
| AddOneAttrDelta |172.54 ns| 185.37 ns  |  157.08 ns
| AddThreeAttr |396.93 ns |393.36 ns | 368.69 ns| 
| AddThreeAttrDelta |396.41 ns | 393.68 ns|368.60 ns |
| AddFiveAttr |706.21 ns |718.80 ns |672.07 ns | 

Hashbrown/ahash seems widely adopted crates in terms of number of downloads, and also secure enough for external ddos attacks - https://github.com/tkaitchuck/aHash/wiki/How-aHash-is-resists-DOS-attacks


cijothomas:
Thanks! This is amazing!
As discussed in the SIG call, a key bottleneck today is the contention among every Update() calls (irrespective of if they are updating the same metric point (timeseries) or not). This eases that a lot, leading to way better scaling, as seen in the throughput increases for higher number of cores!

Lets sit on this for 1- 2 days to see if any concerns, before proceeding to do similar for other instrument types. Also, lets do this after the [pending release](https://github.com/open-telemetry/opentelemetry-rust/pull/1539).

hdost:
@lalitb love the performance gains! Were you able to test this impl against some of the single level hashmap alternative implementations?

> Lets sit on this for **1- 2 days to see if any concerns**, before proceeding to do similar for other instrument types. Also, lets do this after the [pending release](https://github.com/open-telemetry/opentelemetry-rust/pull/1539).

@cijothomas If it's truly for discussion a week seems like a more reasonable timeframe.

lalitb:
> Were you able to test this impl against some of the single level hashmap alternative implementations?

@hdost Do you have any alternative implemetations in mind, can test that.

I tested `dashmap` (which is also a sharded hashmap) but the performance was worse than our existing implementation - https://github.com/open-telemetry/opentelemetry-rust/compare/main...lalitb:metrics-dashmap?expand=1


`main branch (using hashmap)`:
Number of threads: 8
Throughput: 6,640,600 iterations/sec
Throughput: 6,203,800 iterations/sec

`using dashmap`:
Number of threads: 8
Throughput: 1,096,800 iterations/sec
Throughput: 1,184,600 iterations/sec

In general, as tested separately, the dashmap was much faster for concurrent read-only operations, but our scenario is concurrent updates. Also,  we were using dashmap in the old metrics implementation, but was removed in the new PR  (didn't find the reason in history). In general, reluctant to take dependency on any external crate unless it is widely used :)

hdost:
From what it seems there's no public interface effect with this so let's take the win for now and see if we can't make further improvements later 👍

hdost (approved):

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,22 +58,28 @@ impl<T: Number<T>> ValueMap<T> {
...
+                let size = values.len();
+                match values.entry(attrs) {
+                    Entry::Occupied(mut occupied_entry) => {
+                        let sum = occupied_entry.get_mut();
+                        *sum += measurement;
+                    }
+                    Entry::Vacant(vacant_entry) => {
+                        if is_under_cardinality_limit(size) {
+                            vacant_entry.insert(measurement);
+                            self.total_count.fetch_add(1, Ordering::SeqCst);
```
TommyCpp, line 74: Do we need `SeqCst` here? Given we only +1 here the order of which happens first doesn't really matter that much?

lalitb: Yes, good point. We can make it to `Ordering::Relaxed`. I'm also exploring possibilities to eliminate this `total_count` variable. Despite being atomic, it still encounters thread contention. The goal of this count is to reserve the size of the output vector in the collection phase.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -16,9 +18,10 @@ use super::{
 
 /// The storage for sums.
 struct ValueMap<T: Number<T>> {
-    values: Mutex<HashMap<AttributeSet, T>>,
+    buckets: Arc<[Mutex<Option<HashMap<AttributeSet, T>>>; 256]>,
```
lalitb, line 21: using hashing function from [ahash](https://crates.io/crates/ahash) crates help get more juice. I think we can have optional dependency on this crate - so use DefaultHasher from std collection by default,  and ahash::AHasher if application includes this crates.

opentelemetry-sdk/Cargo.toml
```diff
@@ -54,6 +56,7 @@ testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-async-std",
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
+use_hashbrown = ["hashbrown", "ahash"]
```
lalitb, line 59: `ahash` and `hashbrown` have been included as an _optional_ dependencies. Should the user opt to incorporate these dependencies, the HashMap and hashing functions supplied by these crates will be utilized. By default, the standard HashMap is used. Both crates seem to have wide customer adoption (based on the number of downloads). But if there are any concerns, these can be removed. The performance o

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: `total_count` (used to maintain the data-points count) has been removed, and the count is calculated by iteration over the buckets during the collect cycle.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,22 +85,33 @@ impl<T: Number<T>> ValueMap<T> {
...
+                        let sum = occupied_entry.get_mut();
+                        *sum += measurement;
+                    }
+                    Entry::Vacant(vacant_entry) => {
+                        if is_under_cardinality_limit(size) {
+                            vacant_entry.insert(measurement);
+                        } else {
+                            // TBD - Update total_count ??
+                            values
+                                .entry(STREAM_OVERFLOW_ATTRIBUTE_SET.clone())
```
cijothomas, line 110: do we need to store the overflow attribute in each bucket? or once only?

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -1,22 +1,35 @@
...
+    HashMap,
+};
 
 use super::{
     aggregate::{is_under_cardinality_limit, STREAM_OVERFLOW_ATTRIBUTE_SET},
     AtomicTracker, Number,
 };
 
+type BucketValue<T> = Mutex<Option<HashMap<AttributeSet, T>>>;
+type Buckets<T> = Arc<[BucketValue<T>; 256]>;
```
cijothomas, line 29: why 256? Is that something that should be done based on processor count?

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: overflow attribute is handled separately, and is stored only once.

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: These are not designed as buckets for each thread/core. The selection of a fixed size of 256 buckets was based on confining the hash value to a `u8` size, and given a default cardinality limit of 2000, with a uniform distribution, each bucket would contain approximately 10 metric points. This can be made configurable if we want the application to define the optimal size of buckets based on their attribute distribution.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,22 +85,33 @@ impl<T: Number<T>> ValueMap<T> {
...
+            let bucket_index = Self::hash_to_bucket(&attrs) as usize; // Ensure index is usize for array indexing
+            let bucket_mutex = &self.buckets[bucket_index];
+            let mut bucket_guard = bucket_mutex.lock().unwrap();
+
+            if bucket_guard.is_none() {
+                *bucket_guard = Some(HashMap::new()); // Initialize the bucket if it's None
+            }
+
+            if let Some(ref mut values) = *bucket_guard {
+                let size = values.len();
```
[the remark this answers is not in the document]
lalitb: Implementing this marginal optimization would need modification in current methodology due to Rust's borrowing rules.  We borrow the `values` hashmap earlier in the code, and with change there would be additional borrow required to find the length. I feel doing the changes will affect the code readibility, not sure if it is worth given that the obtaining the size of the hashmap is an O(1) operation :)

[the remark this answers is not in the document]
cijothomas: got it. its a minor thing, not a blocker.

opentelemetry-sdk/Cargo.toml
cijothomas: given it is possible that the input (keys/values) can come from external sources, the default may be more safe against collision attacks, so lets add a warning/guide to advice endu sers on not relying on use_hashbrown flag, unless they control the keys/values.

lalitb:
> > I see you've got some further improvements, how's the performance improvement for you ?
> 
> The additional improvements are with (optionally) using the fast hashing with hashbrown/ahash. Will update the improvements later from the same machine I used as base earlier. As of now, I am replicating the same changes for other aggregation.

Updated the latest throughputs in the PR description. The perf-boost using hashbrown/ahash is marginal.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -29,12 +43,41 @@ impl<T: Number<T>> Default for ValueMap<T> {
 
 impl<T: Number<T>> ValueMap<T> {
     fn new() -> Self {
+        let buckets = std::iter::repeat_with(|| Mutex::new(None))
+            .take(BUCKET_COUNT)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap_or_else(|_| panic!("Incorrect length"));
```
cijothomas, line 50: you'll remove the panic calls right?

lalitb: this will never reach, as the bucket count is fixed. Just kept for readability.

lalitb:
Also added benchmark results, they are single-threaded, so the changes won't improve those., though we see the improvement with the use of hashbrown/ahash.

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: ok this is removed now.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -29,12 +43,41 @@ impl<T: Number<T>> Default for ValueMap<T> {
 
 impl<T: Number<T>> ValueMap<T> {
     fn new() -> Self {
+        let buckets = std::iter::repeat_with(|| Mutex::new(None))
+            .take(BUCKET_COUNT)
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap();
```
cijothomas, line 50: if we are never expecting to reach unwrap(), add a code comment. I am not sure if there is existing precedence in this repo for that.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -29,12 +43,41 @@ impl<T: Number<T>> Default for ValueMap<T> {
...
+    // Hash function to determine the bucket
+    fn hash_to_bucket(key: &AttributeSet) -> u8 {
+        #[cfg(not(feature = "use_hashbrown"))]
+        let mut hasher = DefaultHasher::new();
+        #[cfg(feature = "use_hashbrown")]
+        let mut hasher = AHasher::default();
+
+        key.hash(&mut hasher);
+        // Use the 8 least significant bits directly, avoiding the modulus operation.
+        hasher.finish() as u8
```
cijothomas, line 68: trying to understand more here.. AttributeSet, which is the key here, is calculating hash already and storing it. That uses the default hasher only, so how does it know that this is using a different hashing algorithm?

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: Oh, [panic](https://github.com/open-telemetry/opentelemetry-rust/pull/1564#discussion_r1506756305) has been removed, unwrap() will be always called to convert the vector into an array reference.

opentelemetry-sdk/src/testing/metrics/in_memory_exporter.rs
```diff
@@ -285,6 +285,9 @@ impl TemporalitySelector for InMemoryMetricsExporter {
 #[async_trait]
 impl PushMetricsExporter for InMemoryMetricsExporter {
     async fn export(&self, metrics: &mut ResourceMetrics) -> Result<()> {
+        if metrics.scope_metrics.is_empty() || metrics.scope_metrics[0].metrics.is_empty() {
```
cijothomas, line 288: this is unexpected, right? if nothing to be exported, the sdk should not call exporters? Or this is to signal to exporters that export is trigged, but have no metric to export.
(unrelated to this PR I guess...)

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: The `hasher` is conditionally initialized either as `AHasher`, or `DefaultHasher`, based on whether `use_hashbrown` feature is enabled or not. So 

```rust
key.hash(&mut hasher);
```

will use the appropriate hashing algorithm based on the hasher object. passed.

opentelemetry-sdk/src/testing/metrics/in_memory_exporter.rs
lalitb: Yes, I did notice the export triggered with empty metric data. Let me check it further if it is related to the PR. Don't merge till I resolve this comment :)

opentelemetry-sdk/src/metrics/internal/sum.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/5e67cae02030a98e255d1e9ac1bd25d9bef70963/opentelemetry-sdk/src/attributes/set.rs#L179

it's just returning the hash already calculated using default hash right?

lalitb: > it's just returning the hash already calculated using default hash right?

No.
```rust
impl Hash for AttributeSet {
    fn hash<H: Hasher>(&self, state: &mut H) {
        state.write_u64(self.1)
    }
}
```

Hasher `H` is a generic/template in the above method, so based on the hasher instance used, the hash would be calculated.

opentelemetry-sdk/src/testing/metrics/in_memory_exporter.rs
lalitb: Ok, have removed this check. Seems it was caused by some thread conflict caused by running multiple tests concurrently in my local setup. The CI runs them sequentially so it shouldn't be an issue.

opentelemetry-sdk/Cargo.toml
lalitb: Have added this comment in Cargo.toml, purposefully didn't add the WARNING word :)

```
# Enable use_hashbrown for critical performance when measurement attributes are secure and not prone to external manipulation.
```

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: Modified the code to handle error from unwrap().

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,22 +90,39 @@ impl<T: Number<T>> ValueMap<T> {
...
-                    *sum += measurement;
+        } else {
+            let bucket_index = Self::hash_to_bucket(&attrs) as usize;
+            let bucket_mutex = &self.buckets[bucket_index];
+
+            let mut bucket_guard = match bucket_mutex.lock() {
+                Ok(guard) => guard,
+                Err(e) => {
+                    eprintln!("Failed to acquire lock due to: {:?}", e);
+                    return; // TBD - retry ?
```
bIgBV, line 101: In this failure path, does the measurement just get dropped? The only reason why a mutex lock would fail is if the mutex was poisoned. Reading the [documentation on mutex poisoning](https://doc.rust-lang.org/std/sync/struct.Mutex.html#poisoning)

> Once a mutex is poisoned, all other threads are unable to access the data by default as it is likely tainted (some invariant is not being upheld).

We should probably account for this error state and try to recover from the failure. A mutex is usually poisoned because another thread panicked while holding the lock. Since the underlying data this mutex is protecting is a map of measurements, I don't think there are external invariants that would be affected if we re-acquired the guard from within the mutex. This can be done using the [`into_inner`](https://doc.rust-lang.org/std/sync/struct.PoisonError.html#method.into_inner) method on the `PoisonError` type. This would get us access to the underlying data to update the measurement if we want to, but we would still need to re-instantiate the bucket with a new mutex.

Since `into_inner` gives us access to the inner hashmap, we can probably `mem::replace` the specific bucket with a fresh mutex with the existing data in the error path. 

This would definitely be a performance hit, but I think it's worth it because of the data integrity it affords.

lalitb: Thanks @bIgBV. This is the right approach. Seems we are not handling lock even in the earlier single-level hashmap. I will create a separate issue to handle this, to be fixed in a separate PR.

lalitb: created #1594 to handle mutex poisoning.

opentelemetry-sdk/Cargo.toml
```diff
@@ -54,6 +56,8 @@ testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-async-std",
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
+# Enable use_hashbrown for critical performance when measurement attributes are secure and not prone to external manipulation.
```
cijothomas, line 59: Enable use_hashbrown to improve hashing performance in Metrics aggregation.
This feature should be used carefully, especially when the key/value pairs in measurement attributes are derived from external or untrusted sources.

lalitb: Done.

opentelemetry-sdk/src/metrics/internal/sum.rs
cijothomas: i am not sure how is this handled.. it looks to me like we'll store overflow for each of the hasmaps.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,22 +90,39 @@ impl<T: Number<T>> ValueMap<T> {
...
             self.has_no_value_attribute_value
                 .store(true, Ordering::Release);
-        } else if let Ok(mut values) = self.values.lock() {
-            let size = values.len();
-            match values.entry(attrs) {
-                Entry::Occupied(mut occupied_entry) => {
-                    let sum = occupied_entry.get_mut();
-                    *sum += measurement;
+        } else {
+            let bucket_index = Self::hash_to_bucket(&attrs) as usize;
```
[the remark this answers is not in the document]
lalitb: Yeah returning `hasher.finish() as u8 as usize` from `hash_to_bucket`

TommyCpp (approved):
Nice work. Thanks

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: Good point, this was indeed overflow every hashmap. Have fixed it now. Will also add tests to validate this.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -29,12 +42,46 @@ impl<T: Number<T>> Default for ValueMap<T> {
...
+        #[cfg(feature = "use_hashbrown")]
+        let mut hasher = AHasher::default();
+
+        key.hash(&mut hasher);
+        // Use the 8 least significant bits directly, avoiding the modulus operation.
+        hasher.finish() as u8 as usize
+    }
+
+    // Calculate the total length of data points across all buckets.
+    fn total_data_points_count(&self) -> usize {
```
cijothomas, line 72: is that different from total_unique_entries? 🤔

lalitb: No, this is just a left over, have removed it now.

lalitb: This was a leftover, have removed it now.

lalitb:
As discussed during the community meeting, have added the tests for overflow and concurrent measurement recordings.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -161,6 +161,189 @@ mod tests {
...
+            counter.add(1, &[KeyValue::new("key1", attribute_value)]);
+        }
+        meter_provider.force_flush().unwrap();
+
+        // Assert
+        let resource_metrics = exporter
+            .get_finished_metrics()
+            .expect("metrics are expected to be exported.");
+        // Every collect cycle produces a new ResourceMetrics (even if no data is collected).
+        // TBD = This needs to be fixed, and then below assert should validate for one entry
```
lalitb, line 203: Will create an issue for this.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -161,6 +161,189 @@ mod tests {
...
+
+        // Act
+        let meter = meter_provider.meter("test");
+        let counter = meter
+            .u64_counter("my_counter")
+            .with_unit(Unit::new("my_unit"))
+            .init();
+
+        // sleep for random ~5 milis to avoid recording during first collect cycle
+        // (TBD: need to fix PeriodicReader to NOT collect data immediately after start)
```
lalitb, line 187: Will create an issue for this.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,26 +75,45 @@ impl<T: Number<T>> ValueMap<T> {
...
+            return;
         }
+
+        let bucket_index = Self::hash_to_bucket(&attrs);
+        let (is_new_entry, should_use_overflow) = {
+            let bucket_mutex = &self.buckets[bucket_index];
+            let bucket_guard = bucket_mutex.lock().unwrap();
+
+            let is_new_entry = if let Some(bucket) = &*bucket_guard {
+                !bucket.contains_key(&attrs)
```
cijothomas, line 87: in the common path where attributes are already existing, now we have to acquire lock once, do lookup, release lock, and the re-acquire the lock, and do the lookup+update. 

Apart from the perf hit, this **loses the atomicity of the update**. It is possible that, between the time we release the lock and re-acquire, other entries might have occurred and the limit was hit, so this attribute should be going into over-flow.

We need to ensure atomicity and avoid the two-step lock-release-re-lock.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust/pull/1564/files#r1516624961

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: Good point,  made the required changes. All operations should be (theoretically) atomic now, and perf should be fast for common scenarios of attributes already existing.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,26 +95,41 @@ impl<T: Number<T>> ValueMap<T> {
...
+
+        // if the attribute is not present in bucket, lets unlock the bucket
+        drop(bucket_guard);
+        // Attempt to first increment the total unique entries if under limit.
+        let under_limit = self.try_increment();
+        // Determine the bucket index for the attributes
+        let (bucket_index, attrs) = if under_limit {
+            (bucket_index, attrs) // the index remains same
+        } else {
+            // TBD - Should we log, as this can flood the logs ?
```
cijothomas, line 121: shouldn't log. the existing code was incorrectly doing it.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -43,26 +95,41 @@ impl<T: Number<T>> ValueMap<T> {
...
+            (OVERFLOW_BUCKET_INDEX, STREAM_OVERFLOW_ATTRIBUTE_SET.clone())
+        };
+
+        // Lock and update the relevant bucket
+        let mut final_bucket_guard = self.buckets[bucket_index].lock().unwrap();
```
cijothomas, line 126: this wont work, because by the time we reach and acquire this lock, collect() may have triggered and wiped the hashmap clean, and another updates could have occurred as well making our assumption that this attribute is not-an-overflow invalid.

let under_limit = self.try_increment();
It is entirely possible that this thread loses its CPU right after the above statement is executed. When it gets a chance to execute again, under_limit could have changed already. We need to do this whole thing atomically.

lalitb: There are two changes in the design
 - We increment before the actual insert.
 - The increment i.e, try_increment() is atomic.

There is a narrow window where a measurement with new set of attributes go to overflow index even when there is empty index (this can happen during collect), which should be fine. However, there won't be a scenario where the number of data-points exceed the cardinality limit.

lalitb: And thanks for the thorough review of this PR, thinking of all possible scenarios and edge conditions :)

cijothomas: take this example:
thread1 is executing update with attributes (foo=bar)., the current total_unique is 100, with limit as 2000
say bucket_index is 4
take lock
(foo=bar) is not present. 
drop lock
try_increment is called which increments total_unique to 101,  which is within the limit., so this thread will go and update bucket_index 4.
====thread1 lost its CPU===

// in other threads
collect() occurred and it reset total_unique to 0
other updates() occurred and take total_unique to 2000 (limit hit). None of these updates() were having (foo=bar) as attribute. 
At this stage (foo=bar) should be going into overflow.

====thread1 got back its CPU===
It takes lock to update bucket_index 4 with (foo=bar). But this is not correct. (foo=bar) should be going to overflow only.

lalitb: > collect() occurred and it reset total_unique to 0

It won't cleanup to 0, but will only adjust(decrement) according to the number of entries read.

cijothomas: collect() read and drained the entire hashmaps. Won't it hit 0 then?

lalitb: In the above scenario, when the number of entries in hashmap is 100, while the unique_count is 101 (for foo=bar), if collect happens before foo=bar is inserted, the count will get reset to 1 (not 0), and then the insert happens after/during the collect.

cijothomas: take the example where 5 threads try to update with (foo=bar) which is not existing currently. The current unique_count is 1999, and limit is 2000.
All 5 threads realize entry not found, so they all attempt to increment the unique_count. 
first thread succeeds, and attempts to store into correct bucket but remaining 4 threads will attempt to put foo=bar into overflow. I believe this is the part which you said might be okay?

However, consider the same scenario, but the current unique_count was just 10.
All 5 threads realize entry not found, so they all attempt to increment the unique_count.
All threads would succeed, so unique_count reaches 15. (though it should actually be 11)
One threads would then **insert** the hashmap and **rest** would update the hashmap.

Collect() runs, and it drains the map, but when it drains (foo=bar), it'll reduce unique_count by 1 (from 15 to 14). So we now have 4 wasted entries. No matter how many collect() runs afterwards, we'll never reclaim that wasted spot... If we run stress test long enough (may weeks), we might see all entries going to overflow, even though there is plenty of space.

Please confirm if the above is correct.

lalitb:
Just an update, I am still working on it, particularly to mitigate the challenges around concurrency and atomicity - The main issue is making sure the operations of checking if we're under the cardinality limit and then acting on it (like inserting a new entry) happen seamlessly without any race conditions or inconsistencies due to concurrent updates. This actually turned out to be more complex than initially anticipated, especially when trying to do so without compromising on performance. Still working on it, hoping to have a solution during this week :)

hdost (approved):

cijothomas:
> Just an update, I am still working on it, particularly to mitigate the challenges around concurrency and atomicity - The main issue is making sure the operations of checking if we're under the cardinality limit and then acting on it (like inserting a new entry) happen seamlessly without any race conditions or inconsistencies due to concurrent updates. This actually turned out to be more complex than initially anticipated, especially when trying to do so without compromising on performance. Still working on it, hoping to have a solution during this week :)

Let us know if this is ready for another review. If not, we can mark as draft again.

@hdost I am waiting for confirmation that this is ready for review, as we have open conversations that are not marked resolved.

lalitb:
> Let us know if this is ready for another review. If not, we can mark as draft again.

Sorry for the delay. Will revisit it now. If it takes more time, will move it to draft.

lalitb:
There have been substantial perf improvements in metrics instrumentation with #1833.  Will revisit this PR after logs beta release. Closing this for now to keep the PR list minimal.

PR #1567, opened 2024-02-22
Use a builder pattern for loggers and tracers

### LINKED ISSUE #1527: Instrumentation scope - move to builder pattern?
https://github.com/open-telemetry/opentelemetry-rust/pull/1021 added supported for Attributes to instrumentation scope, which already had version, schema_url.
This started with just name and evolved to add more and more (because spec added them). https://github.com/open-telemetry/opentelemetry-rust/pull/1021#discussion_r1161364638 called out the need for potentially making this to a builder pattern, so as to protect us from breaking change in the future, if there are more stuff added to scope.

```
let tracer = TracerProvider.tracer("tracer_name"); // common case. this is still a good idea to keep to keep things simple for majority.
let tracer = TracerProvider::tracer_builder("tracer_name") // advanced scenario.
    .with_version("1.0.0")
    .with_attributes([kv])
    .build();
```

Opening an issue to explore if moving to builder pattern makes sense here (all 3 signals), and if yes, implement the actual change.
Also, there could be other places where such changes are required to make things cleaner/more idiomatic and also protect from future breaking changes.

izquierdo:
Fixes #1527
Design discussion issue (if applicable) #1527

## Changes

This is currently just an exploratory PR to aid in the design discussion of the attached issue. May upgrade this PR and submit for review depending on the discussion result.


opentelemetry/src/trace/tracer_provider.rs
```diff
@@ -45,40 +45,11 @@ pub trait TracerProvider {
         )
     }
 
-    /// Returns a new versioned tracer with a given name.
```
hdost, line 48: I wouldn't necessarily remove the old way immediately. We could instead mark it deprecated and encourage use of the new one.

cijothomas:
@izquierdo Do you want us to review now, or should we wait till PR is marked non-draft?

izquierdo:
> @izquierdo Do you want us to review now, or should we wait till PR is marked non-draft?

Please wait until the PR is marked as non-draft.

However, I made a comment here: https://github.com/open-telemetry/opentelemetry-rust/issues/1527#issuecomment-1969954371

izquierdo:
Let me open this for review as is, with loggers and tracers. Will leave metrics out of this PR.

opentelemetry/src/common.rs
```diff
@@ -501,4 +502,92 @@ impl InstrumentationLibrary {
             attributes: attributes.unwrap_or_default(),
         }
     }
+
+    /// Create a new builder to create an [InstrumentationLibrary]
+    pub fn builder(name: impl Into<Cow<'static, str>>) -> InstrumentationLibraryBuilder {
```
TommyCpp, line 507:
```suggestion
    pub fn builder<T: Into<Cow<'static, str>>>(name: T) -> InstrumentationLibraryBuilder {
```

opentelemetry/src/trace/tracer_provider.rs
```diff
@@ -94,13 +129,42 @@ pub trait TracerProvider {
...
+        self.library_builder = self.library_builder.with_version(version);
+        self
+    }
+
+    pub fn with_schema_url(mut self, schema_url: impl Into<Cow<'static, str>>) -> Self {
+        self.library_builder = self.library_builder.with_schema_url(schema_url);
+        self
+    }
+
+    pub fn with_attributes(mut self, attributes: impl Into<Vec<KeyValue>>) -> Self {
```
TommyCpp, line 161: To make the API even more flixable we can use something similar to https://github.com/open-telemetry/opentelemetry-rust/blob/e98c61ce73865b56917695dc0ab89ba3eca6c454/opentelemetry/src/trace/tracer.rs#L333-L335

opentelemetry/src/common.rs
TommyCpp: I think the type is complicated enough to move out of the paramters

TommyCpp (commented):
Seems aligned with the discussion in #1527 overall looks good

opentelemetry/src/trace/tracer_provider.rs
cijothomas: I have a different opinion on this. I am okay with removing this right away (_i.e in this PR itself_). When we do the next release (0.23 from 0.22), it's allowed to have breaking changes. And in my opinion, its better to clean up **now** than delaying actual removal for another release/accumulating debt for future release.
I agree its more disruptive for end users, but I prefer to do things sooner than delaying. (users anyway have to change code at some point, why not make them do it earlier.)

@hdost @TommyCpp @lalitb do you strongly feel we should keep old api for another release? If everyone prefer to keep this for now, I am okay to accept that. Maybe I am missing something.

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -55,4 +55,45 @@ mod tests {
...
+    fn logger_attributes() {
+        let provider = LoggerProvider::builder().build();
+        let logger = provider
+            .logger_builder("test_logger")
+            .with_schema_url("https://opentelemetry.io/schema/1.0.0")
+            .with_attributes(vec![(KeyValue::new("test_k", "test_v"))])
+            .build();
+        let instrumentation_library = logger.instrumentation_library();
+        let attributes = &instrumentation_library.attributes;
+        assert_eq!(
```
cijothomas, line 69: assert for name as well?

opentelemetry/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 - Remove `urlencoding` crate dependency. [#1613](https://github.com/open-telemetry/opentelemetry-rust/pull/1613)
 
+### Changed
+
+- Deprecate `versioned_logger()` and `versioned_tracer()` in favor of `logger_builder()` and `tracer_builder()` respectively [1567](https://github.com/open-telemetry/opentelemetry-rust/pull/1567).
```
cijothomas, line 15: could we make the changelog more self-contained? for example, we can include actual changes expected by endusers right in this changelog itself.
eg:
provider.versioned_logger(name,version,schema,attr) should now be replaced with provider.builder().name().version()....

^ something like this? In general, I am trying to see if we can make chanelogs more self-contained wherever feasible, so end users can just read changelog and understand the necessary changes. For bigger ones, of course, link to the PR number, and we can give details in the PR description.

opentelemetry/src/trace/tracer_provider.rs
lalitb: I vote for being less disruptive for API-related changes, and follow deprecation followed by removal.

TommyCpp: I generally are OK if we move fast for API. But in this case I think it's a widely used API so it's probably worth a deprecation + remove in next release to control the size of changes needed for a version bump

cijothomas: Thanks for sharing your thoughts. I agree. Lets proceed with "mark-deprecated-first, then actual removal in next release" approach for anything very common, like in this case.

lalitb: If we plan to selectively deprecate API changes, and somtimes also directly have breaking changes in consecutive releases, we can add this explicitly in README, to set a clear expectation. Even with the status (alpha, beta) indication, users can expect the breaking changes to be deprecated first.

Have seen some concern raised by the user in twitter, so probably thinking more in that aspect :)

image

lalitb: Yeah sure. Will add this in the readme for the crates.

opentelemetry/CHANGELOG.md
hdost: Something like ?
Before:
```rust
let tracer = provider.versioned_tracer(
    "my-tracer-name",
    Some("X.Y.Z"),
    Some("https://opentelemetry.io/schema/1.0.0""),
    Some(vec![KeyValue::new("key", "value")]),
);
```
After:
```rust     
let tracer = provider.tracer_builder("my-tracername")
    .with_version(env!("CARGO_PKG_VERSION"))
    .with_schema_url(semcov::SCHEMA_URL)
    .with_attributes(vec![KeyValue::new("key", "value")])
    .build();
```

TommyCpp (approved):
LGTM. We probably want to revise the change log as suggested above

opentelemetry/CHANGELOG.md
izquierdo: Updated the changelog -- please check it out

opentelemetry/CHANGELOG.md
```diff
@@ -11,6 +11,56 @@
...
+    Some("https://opentelemetry.io/schema/1.0.0"),
+    Some(vec![KeyValue::new("key", "value")]),
+);
+```
+
+After:
+
+```
+let logger = provider
+    .logger_builder("my-tracername")
```
cijothomas, line 33:
```suggestion
    .logger_builder("my-logger-name")
```

opentelemetry/CHANGELOG.md
```diff
@@ -11,6 +11,56 @@
...
+### Changed
+
+- Deprecate `versioned_logger()` in favor of `logger_builder()` [1567](https://github.com/open-telemetry/opentelemetry-rust/pull/1567).
+
+Before:
+
+```
+let logger = provider.versioned_logger(
+    "my-logger-name",
+    Some("X.Y.Z"),
```
cijothomas, line 23:
```suggestion
    Some(env!("CARGO_PKG_VERSION")),
```

opentelemetry/CHANGELOG.md
```diff
@@ -11,6 +11,56 @@
...
+);
+```
+
+After:
+
+```
+let logger = provider
+    .logger_builder("my-tracername")
+    .with_version(env!("CARGO_PKG_VERSION"))
+    .with_schema_url(semcov::SCHEMA_URL)
```
cijothomas, line 35:
```suggestion
    .with_schema_url("https://opentelemetry.io/schema/1.0.0")
```

opentelemetry/CHANGELOG.md
```diff
@@ -11,6 +11,56 @@
...
+    Some("https://opentelemetry.io/schema/1.0.0"),
+    Some(vec![KeyValue::new("key", "value")]),
+);
+```
+
+After:
+
+```
+let tracer = provider
+    .tracer_builder("my-tracername")
```
cijothomas, line 57:
```suggestion
    .tracer_builder("my-tracer-name")
```

cijothomas (approved):
Thanks!
I left couple of nit comments to improve changelog, but Looks good overall.
Will you help with replicating this for Meter as well?

izquierdo:
> Will you help with replicating this for Meter as well?

Yes! But in a separate PR

PR #1057, opened 2023-05-09
Add OTLPMetricPipeline builder methods for delta and lowmemory

cijothomas:
## Changes

Add new builder methods to help build OTLP Metrics Pipeline with desired Temporality. This PR specifically adds
builders to support the "Delta" and "LowMemory" temporality preference from the spec:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md#additional-configuration

Note that the env variable support for the same (OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE), is *not* part of this PR, and will be offered as a separate PR.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1568, opened 2024-02-22
otlp: Add DeltaTemporalitySelector

svix-jplatte:
cc #1057, #1511

## Changes

Adds a temporality selector that prefers delta temporality, since some services like DataDog work better with it.

* [ ] ~~Unit tests added/updated (if applicable)~~


opentelemetry-otlp/src/metric.rs
```diff
@@ -247,6 +247,36 @@ impl<RT, EB: Debug> Debug for OtlpMetricPipeline<RT, EB> {
             .finish()
     }
 }
+/// A temporality selector that returns [`Delta`][Temporality::Delta] for all
+/// instruments except `UpDownCounter` and `ObservableUpDownCounter`.
+///
+/// This should be used over the default if you export your data to a service
+/// like DataDog that works better with delta aggregation¹.
```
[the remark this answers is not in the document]
svix-jplatte: Actually I find that naming DataDog is a major reason to have this in the crate in the first place: This way, it's more likely that people realize they should be using it (when they're exporting to DataDog).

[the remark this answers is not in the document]
cijothomas: I don't think so. Datadog (or any other vendor) can document it in their docs to use delta temporality. There is no need to talk about any vendors in the OTel Exporters.

NewRelic is another example of vendor expecting delta. My employer (Microsoft) also has backends expecting delta. We don't want to list the vendors here.

To make it easy for end users, some vendors do [distributions](https://opentelemetry.io/docs/concepts/distributions/) as well.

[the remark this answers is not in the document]
svix-jplatte: I'd much rather list all of these than none, but your choice :shrug:

cijothomas (approved):
Thanks!

Will have to hold off from merging for few hours, as we are in the middle of a release.

opentelemetry-otlp/src/metric.rs
```diff
@@ -248,6 +258,36 @@ impl<RT, EB: Debug> Debug for OtlpMetricPipeline<RT, EB> {
...
+/// A temporality selector that returns [`Delta`][Temporality::Delta] for all
+/// instruments except `UpDownCounter` and `ObservableUpDownCounter`.
+///
+/// This temporality selector is equivalent to OTLP Metrics Exporter's
+/// `Delta` temporality preference (see [its documentation][exporter-docs]).
+///
+/// [exporter-docs]: https://github.com/open-telemetry/opentelemetry-specification/blob/a1c13d59bb7d0fb086df2b3e1eaec9df9efef6cc/specification/metrics/sdk_exporters/otlp.md#additional-configuration
+#[non_exhaustive]
+#[derive(Debug)]
+pub struct DeltaTemporalitySelector;
```
cijothomas, line 270: on second thought, do we need this to be pub? I am thinking maybe this is better moved to the sdk itself as a pub struct.

The `pub fn with_delta_temporality` alone should be sufficient for users to leverage this, right?

svix-jplatte: Ah well, if it's public it certainly shouldn't be `#[non_exhaustive]` (or it needs a c'tor). I'm not sure what the best solution is, I kind of dislike making it private since that makes it impossible to reuse its `TemporalitySelector` impl as part of another `TemporalitySelector` impl, but I have no idea how useful that would actually be.

Also seems like it might pollute the root namespace a little bit, but since the `metric` module is private, re-exporting at the crate root seemed like the only sensible choice.

cijothomas: Changing temporality is a very common ask, so I am anticipating that OTel SDK itself will provide the building blocks for that, so that every exporter can leverage it!
There is a lot of duplication today for this!
eg: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/mod.rs#L603

let me think more and come back. Hopefully there are others who can chime in with a better solution.

lalitb: Within core metrics Exporters, the OTLP exporter is the only valid scenario (per specs) for providing Delta Temporality configuration, as Prometheus is cumulative. If so, we can let it be here. Other vendor-specific exporters can independently decide on implementing their temporality selector, including the option for delta, if necessary.

cijothomas: we can still achieve the goal without making the struct public, just the builder method public - @lalitb is this possible?

lalitb (approved):

opentelemetry-otlp/src/metric.rs
lalitb: Yes, that should be the right way.

cijothomas:
@svix-jplatte We just completed the release, but it caused conflict! Could you resolve the conflicts, and we are ready to merge?

PR #1192, opened 2023-08-08
Metric stdout exporter to print timestamp in human readable format

cijothomas:
Addresses https://github.com/open-telemetry/opentelemetry-rust/issues/1191 for metrics.

## Changes

~Update stdout metric exporter to write timestamp in human readable format.~
Writes timestamp in human-friendly format as well. The existing unix_ns format is kept as-is.

Example outout with this PR:
"startTime":"2023-08-08 00:02:41.404","time":"2023-08-08 00:02:41.404"
```json
[...]

PR #1562, opened 2024-02-19
fix: disable workspace default feature

TommyCpp:
Follow up on https://github.com/open-telemetry/opentelemetry-rust/pull/1486#issuecomment-1933152551

## Changes

- Disable default feature for some workspace dependencies to allow each create to decide which feature to enable.
- Remove integration tests from jaeger-propagator

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1569, opened 2024-02-23
Switch from the chrono `clock` feature to `now`

edmorley:
In #1192, `chrono` was added as a dependency of the `opentelemetry-stdout` crate in order to support outputting timestamps in human readable format.

In that PR, all Chrono features were disabled apart from the `clock` feature.

However, since that change landed, `chrono` [v0.4.32](https://github.com/chronotope/chrono/releases/tag/v0.4.32) has added support for an even finer-grained feature named `now`, which is a subset of the `clock` feature - that excludes timezone support, and so avoids pulling in many timezone related crates.

`opentelemetry-stdout` only uses chrono's UTC features, so we can safely switch from using the `clock` feature to using `now` instead.

After this change, the following transitive dependencies are no longer pulled in:

- `android-tzdata`
- `android_system_properties`
- `cc`
- `core-foundation-sys`
- `iana-time-zone`
- `iana-time-zone-haiku`
- `windows-core`
- `windows-targets`
- `windows_aarch64_gnullvm`
- `windows_aarch64_msvc`
- `windows_i686_gnu`
- `windows_i686_msvc`
- `windows_x86_64_gnu`
- `windows_x86_64_gnullvm`
- `windows_x86_64_msvc`

See:
https://github.com/chronotope/chrono/pull/1343
https://github.com/chronotope/chrono/blob/main/README.md#crate-features

### DIFF
--- a/opentelemetry-stdout/CHANGELOG.md
+++ b/opentelemetry-stdout/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## vNext
 
+### Changed
+
+- Adjusted `chrono` features to reduce number of transitive dependencies. [#1569](https://github.com/open-telemetry/opentelemetry-rust/pull/1569)
+
 ## v0.2.0
 
 ### Changed
--- a/opentelemetry-stdout/Cargo.toml
+++ b/opentelemetry-stdout/Cargo.toml
@@ -22,7 +22,7 @@ logs = ["opentelemetry/logs", "opentelemetry_sdk/logs", "async-trait", "thiserro
 
 [dependencies]
 async-trait = { workspace = true, optional = true }
-chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
+chrono = { version = "0.4.34", default-features = false, features = ["now"] }
 thiserror = { workspace = true, optional = true }
 futures-util = { workspace = true, optional = true }
 opentelemetry = { version = "0.21", path = "../opentelemetry", default_features = false }


shaun-cox (approved):
Thanks for the contribution!

hdost (approved):

edmorley:
The `coverage` CI job is failing for unrelated errors in another crate - any ideas?

```
 ---- exporter::config::collector::tests::test_set_collector_endpoint stdout ----
thread 'exporter::config::collector::tests::test_set_collector_endpoint' panicked at opentelemetry-jaeger/src/exporter/config/collector/mod.rs:575:9:
assertion `left == right` failed
  left: "ConfigError { pipeline_name: \"collector\", config_name: \"collector_endpoint\", reason: \"invalid uri from environment variable, invalid uri character\" }"
 right: "ConfigError { pipeline_name: \"collector\", config_name: \"collector_endpoint\", reason: \"invalid uri from the builder, invalid format\" }"
```

cijothomas:
@hdost We should not merge PRs while release is pending. We need to resolve conflicts again in https://github.com/open-telemetry/opentelemetry-rust/pull/1539

hdost:
> @hdost We should not merge PRs while release is pending. We need to resolve conflicts again in https://github.com/open-telemetry/opentelemetry-rust/pull/1539

There were no conflicts from this merge. We were still waiting on the no defaults PR at the time, so even if there were conflicts it would have been 1 small line diff and as this is a leaf had no downstream affects on other crates. 

In general, I agree we need to establish some sort of process, but there was only a net good considering how often we get asks for releases.

cijothomas:
> > @hdost We should not merge PRs while release is pending. We need to resolve conflicts again in #1539
> 
> There were no conflicts from this merge. We were still waiting on the no defaults PR #1562 at the time, so even if there were conflicts it would have been 1 small line diff and as this is a leaf had no downstream affects on other crates.
> 
> In general, I agree we need to establish some sort of process, but there was only a net good considering how often we get asks for releases.

Yes. I only meant to re-highlight the need to establish a process. Fully agree with your reasoning for merging this.

PR #1570, opened 2024-02-23
Disable default features on reqwest

rbtcollins:
reqwest's default features cause openssl to be linked in, and defeats the point
of supporting rustls feature flags. Disable default features to avoid this.

This was a regression from the migration to workspace dep versions.


rbtcollins:
@TommyCpp it is! But its narrower, and it passes CI :) 

Happy to close it if you prefer.

TommyCpp (approved):

Issue #1489, opened 2024-01-24
Remove aws id generator from core sdk


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/id_generator/aws.rs must be moved to contrib crate and should not be part of core sdk.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#id-generators

sreeo:
how do we proceed with the removal from the core repo? Do we add a deprecated message/warning and release a version?

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/1560 You can copy some of the ideas from this PR. (That PR removes an entire crate, but this is just a functionality, so adjust wordings accordingly.)

sreeo:
Would this be the right time to raise a PR where the XrayIdGenerator is removed from the core repo? Or do we wait for a few releases?

cijothomas:
> Would this be the right time to raise a PR where the XrayIdGenerator is removed from the core repo? Or do we wait for a few releases?

Hey sorry I missed this message until now! It should be removed now : https://github.com/open-telemetry/opentelemetry-rust/pull/1820

PR #1573, opened 2024-02-23
Chore/1489 deprecate aws id generator

sreeo:
Fixes #

## Changes
Deprecates the XrayIdGenerator in the opentelemetry core sdk. 
Relates [#1489](https://github.com/open-telemetry/opentelemetry-rust/issues/1489)

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -1,12 +1,12 @@
 # Changelog
 
-## Deprecation Notice
-
-XrayIdGenerator has been migrated to the opentelemetry-aws crate starting from 0.10.0. [#33](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/33)
-
 
 ## vNext
 
+
+### Deprecated
+
+- XrayIdGenerator in the opentelemetry-sdk has been deprecated and moved to version 0.10.0 of the opentelemetry-aws crate.
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
 # Changelog
 
+## Deprecation Notice
```
lalitb, line 3: should this instead be a "Migration Notice" or "Functionality moved"  or something similar? As this component is moved to another crate without an immediate plan to remove it from the ecosystem entirely?

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -1,7 +1,13 @@
...
+
+XrayIdGenerator has been migrated to the opentelemetry-aws crate starting from 0.10.0. [#33](https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/33)
```
cijothomas, line 5: the message is not very clear. Could you rephrase so that it is clear that we are referrung to 0.10.0 of the aws create.

Also, this can be under a "##Removed" section.

lalitb: I guess has been deprecated in this crate, not removed yet.

lalitb (approved):

opentelemetry-sdk/CHANGELOG.md
hdost: It can however be moved inside of the version. I think this may have been inspired by my jaeger deprecation, but in this case we're only deprecating a component. Where as for jaeger we're deprecating the crate. 

Could be:

```
# Changelog

## vNext

### Deprecated
...

```

sreeo: how about this?
`
### Deprecated

- XrayIdGenerator in the opentelemetry-sdk has been deprecated and moved to version 0.10.0 of the opentelemetry-aws crate.
`

Issue #1306, opened 2023-10-17
Consistent readme file for all components


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-http has no link to crates.io, docs.
https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-user-events-metrics has similar issue, has even less info than opentelemetry-http.
https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-stdout has all the details, but the description could be improved.

Opening an issue to review the main readme.md file for each component, and make it consistent throughout.

PR #1574, opened 2024-02-24
Fix invalid import paths under "Getting Started"

marcusirgens:
Updates the "Getting Started" code sample so that it will compile. Also adds a list of required dependencies, so users won't have to figure out the repository structure and how to find the dependencies in the sample.

I would also like to suggest that the maintainers find a way to perform regression tests against this code snippet, it's probably one of the most important pieces of code in the project.

  Unsure about this - this is not trivial, and also user-facing, but it might not be appropriate for the changelog.
  
  I guess this reviews a previous change in the public API?

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ Project versioning information and stability guarantees can be found
 ```rust
 use opentelemetry::{
     global,
-    sdk::trace::TracerProvider,
     trace::{Tracer, TracerProvider as _},
 };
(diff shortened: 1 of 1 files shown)


hdost (approved):

cijothomas:
Thanks. This is also related to https://github.com/open-telemetry/opentelemetry-rust/issues/1306

PR #1578, opened 2024-02-26
Support URL-encoded values for `OTEL_EXPORTER_OTLP_HEADERS`

### LINKED ISSUE #1577: URL encoded values for `OTEL_EXPORTER_OTLP_HEADERS` are not supported
### What happened?

The Rust SDK should support URL encoded values in `OTEL_EXPORTER_OTLP_HEADERS`. For backward compatibility, also non-URL encoded values should be supported.

Both this settings should work:
```
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic%20AAA"
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic AAA"
```

Currently only non-URL encoded values are supported.

**Additional context**

The OTel specification requires values for `OTEL_EXPORTER_OTLP_HEADERS` to be formatted conforming with the W3C Baggage specification.

The W3C Baggage spec [requires values to be URL encoded](https://www.w3.org/TR/baggage/#value):

> A value contains a string whose character encoding MUST be UTF-8 [[Encoding](https://www.w3.org/TR/baggage/#bib-encoding)]. Any characters outside of the baggage-octet range of characters MUST be percent-encoded. Characters which are not required to be percent-encoded MAY be percent-encoded.

This has been discussed in open-telemetry/opentelemetry-specification#3832, the result being that the behavior described in the OTel and W3C specification is desired and SDKs should support URL encoded values.

### API Version

`main` branch, is likely an issue for all released versions

### SDK Version

`main` branch, is likely an issue for all released versions

OTLP

### Relevant log output

```shell
$ export OTEL_EXPORTER_OTLP_HEADERS=Authorization=Basic%20NzAxN[...]
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `/home/johannes/Source/opentelemetry-rust/target/debug/basic-otlp`
OpenTelemetry trace error occurred. HTTP status client error (401 Unauthorized) for url (https://otlp-gateway-prod-us-east-0.grafana.net/otlp/v1/traces)
$ export OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic NzA[...]"
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `/home/johannes/Source/opentelemetry-rust/target/debug/basic-otlp`
$
```

pyohannes:
Fixes #1577

## Changes

This PR adds support for URL-encoded values in `OTEL_EXPORTER_OTLP_HEADERS` and related headers, which is required by the specification.

For backwards compatibility, both URL-encoded and non-URL-encoded values are supported, for example:

```
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic%20AAA"
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Basic AAA"
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,4 @@ tracing = { version = "0.1", default-features = false }
 tracing-core = { version = "0.1", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 url = { version = "2.2", default-features = false }
+urlencoding = { version = "2.1.3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,4 +66,3 @@ tracing = { version = "0.1", default-features = false }
 tracing-core = { version = "0.1", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 url = { version = "2.2", default-features = false }
-urlencoding = { version = "2.1.3" }
--- a/README.md
+++ b/README.md
@@ -100,10 +100,14 @@ The following crates are maintained in this repo:
 * [`opentelemetry-sdk`] This is the OpenTelemetry SDK crate, and contains the
   official OpenTelemetry SDK implementation. It contains Logging SDK, Metrics
   SDK, and Tracing SDK. It also contains propagator implementations.
-* [`opentelemetry-otlp`] exporter for sending logs, metrics and traces in the
-  OTLP format to an endpoint accepting OTLP, like the [OpenTelemetry
-  collector](https://opentelemetry.io/docs/collector/), or [vendor specific
-  endpoints](https://opentelemetry.io/ecosystem/vendors/).
+* [`opentelemetry-otlp`] - exporter to send telemetry (logs, metrics and traces)
+  in the [OTLP
+  format](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/protocol)
+  to an endpoint accepting OTLP. This could be the [OTel
+  Collector](https://github.com/open-telemetry/opentelemetry-collector),
+  telemetry backends like [Jaeger](https://www.jaegertracing.io/),
+  [Prometheus](https://prometheus.io/docs/prometheus/latest/feature_flags/#otlp-receiver)
+  or [vendor specific endpoints](https://opentelemetry.io/ecosystem/vendors/).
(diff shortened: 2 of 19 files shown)


hdost (approved):

lalitb:
This may not be required/specified in the specs, but for consistency should we also support the URL encoded header values specified as config:

https://github.com/open-telemetry/opentelemetry-rust/blob/ab9415a8cd8b288d985e94a2482e261e9941ed9f/opentelemetry-otlp/src/exporter/http/mod.rs#L120

Cargo.toml
```diff
@@ -66,3 +66,4 @@ tracing = { version = "0.1", default-features = false }
 tracing-core = { version = "0.1", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
 url = { version = "2.2", default-features = false }
+urlencoding = { version = "2.1.3" }
```
TommyCpp, line 69: It seems this repo is unmaintained(last commit indicates the status is `looking for maintainer`). Would be the logic simple enough to implement on our own

pyohannes: Makes sense, it isn't too hard to implement. `urlencoding` seems to be quite widely used according to [statistics here](https://crates.io/crates/urlencoding), I didn't see it's unmaintained.

Let me know your preference, I can add an encoding-implementation if that's desired.

pyohannes:
> This may not be required/specified in the specs, but for consistency should we also support the URL encoded header values specified as config:

I think that makes sense, I added support for decoding URL-encoded values also to `with_headers`.

Cargo.toml
cijothomas: > I think if it's easy we should just implement it ourselves. I want to avoid depending on creates that doesn't have active maintainers

+1

pyohannes: Sounds good, I removed the dependency and added an inline implementation.

lalitb (approved):
Thanks for the fix.

Cargo.toml
stormshield-fabs: `urlencoding`  is also used in the `opentelemetry` crate for the Baggage

PR #1585, opened 2024-02-27
Support http/json protocol

ramgdev:
Fixes #

## Changes

This change will export data in JSON format when `http/json` protocol is specified for a signal.


lalitb:
@ramgdev Thanks for volunteering to add the support. This looks to be in the right direction, and good to review once ready.

opentelemetry-otlp/Cargo.toml
```diff
@@ -73,6 +74,8 @@ http-proto = ["prost", "opentelemetry-http", "opentelemetry-proto/gen-tonic-mess
 reqwest-blocking-client = ["reqwest/blocking", "opentelemetry-http/reqwest"]
 reqwest-client = ["reqwest", "opentelemetry-http/reqwest"]
 reqwest-rustls = ["reqwest", "reqwest/rustls-tls-native-roots"]
+# http json
+http-json = ["serde_json", "prost", "opentelemetry-http", "opentelemetry-proto/gen-tonic-messages", "http", "trace", "metrics"]
```
TommyCpp, line 78: We can enable the `with-serde` feature of `opentelemetry-proto` here?

opentelemetry-otlp/src/exporter/http/trace.rs
```diff
@@ -68,23 +68,31 @@ impl SpanExporter for OtlpHttpClient {
...
-
-    Ok((buf, "application/x-protobuf"))
+    let buf;
+    let ctype;
+    #[cfg(all(feature = "http-proto", not(feature = "http-json")))]{
+        use prost::Message;
+        buf = req.encode_to_vec();
+        ctype = "application/x-protobuf";
+    }
+    #[cfg(all(feature = "http-json", not(feature = "http=proto")))]{
```
TommyCpp, line 85:
```suggestion
    #[cfg(all(feature = "http-json", not(feature = "http-proto")))]{
```

TommyCpp (commented):
Thanks for getting this start! It overall looks good. One thing to consider is we should make features additive. This means we should support enable both `http-json` and `http-proto` and let users config it in code.

opentelemetry-otlp/Cargo.toml
ramgdev: That's a good idea. I'll test it out.

ramgdev:
@TommyCpp I agree users should be able to enable both `http-proto` and `http-json` features on the crate. I'll test it and perhaps use the protocol to decide on certain things in the code. Good catch.

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -28,18 +28,32 @@ pub const OTEL_EXPORTER_OTLP_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_PROTOCOL";
...
+    feature = "grpc-tonic",
+    not(all(feature = "http-proto", feature = "http-json"))
+))]
+/// Default protocol, using grpc
 pub const OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT: &str = OTEL_EXPORTER_OTLP_PROTOCOL_GRPC;
-#[cfg(not(any(any(feature = "grpc-tonic", feature = "http-proto"))))]
+
+#[cfg(not(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json")))]
 /// Default protocol if no features are enabled.
 pub const OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT: &str = "";
```
ramgdev, line 52: @TommyCpp is this a valid default? Should this match the default feature enabled in Cargo.toml instead?

TommyCpp: No it's not. The `default` feature in Cargo.toml contains `grpc-tonic` so we shouldn't encounter unless users explicitly disable the default feature and didn't enable anything else.

Regardls

ramgdev:
@TommyCpp are there existing tests covering the function `build_body` function in the http exporters?

TommyCpp:
> `build_body`

Not exactly for the http exporter but in `opentelemetry-proto` we have this test https://github.com/open-telemetry/opentelemetry-rust/blob/e98c61ce73865b56917695dc0ab89ba3eca6c454/opentelemetry-proto/tests/json_deserialize.rs

ramgdev:
The latest failures do not show up on pre-commit tests locally.  
```
running 11 tests
test exporter::config::agent::tests::set_socket_address ... ok
test exporter::config::collector::http_client::collector_client_tests::test_bring_your_own_client ... ignored
test exporter::config::collector::tests::test_collector_exporter ... ignored
test exporter::config::collector::tests::test_resolve_endpoint ... ok
test exporter::config::collector::tests::test_resolve_timeout ... ok
test exporter::config::collector::tests::test_set_collector_endpoint ... ok
test exporter::config::tests::test_read_from_env ... ok
test exporter::config::tests::test_set_service_name ... ok
test exporter::tests::error_message_should_contain_details ... ok
test exporter::tests::ignores_user_set_values ... ok
test exporter::tests::test_set_status ... ok

test result: ok. 9 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

ramgdev:
Pulling back from review to fix test failures when running `cargo test --all`.

ramgdev:
@TommyCpp please take a look when you get a chance.

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -52,21 +52,38 @@ impl LogExporter for OtlpHttpClient {
...
     let req = ExportLogsServiceRequest {
         resource_logs: logs.into_iter().map(Into::into).collect(),
     };
-    let mut buf = vec![];
-    req.encode(&mut buf).map_err(crate::Error::from)?;
-
-    Ok((buf, "application/x-protobuf"))
+    let buf;
+    let ctype;
+    match default_protocol() {
```
lalitb, line 68: can we make it more idiomatic (something like):
```rust
let (buf, ctype) = match default_protocol() {
    Protocol::HttpJson => {
             ...
             (json.into(), "application/json")
   }
   _ => {
             ..
             (req.encode_to_vec(), "application/x-protobuf")
   }
}
```

opentelemetry-otlp/src/exporter/http/metrics.rs
```diff
@@ -48,20 +48,28 @@ impl MetricsClient for OtlpHttpClient {
     }
 }
 
-#[cfg(feature = "http-proto")]
+#[cfg(any(feature = "http-proto", feature = "http-json"))]
 fn build_body(metrics: &mut ResourceMetrics) -> Result<(Vec<u8>, &'static str)> {
```
TommyCpp, line 52: For this part. I wonder if we can make the feature additive. 

One way I see we can do it is:

- Add `protocol` field in `OtelHttpClient`
- Move `build_body` as part of the `OtelHttpClient`.

Then we just ask users to pass protocol in building pipline(it's already part of the `export_config`. And based on the feature the corresponding protocol may be available

TommyCpp (commented):
Overall looks good. My concern is mostly around `http-json` overriding with `http-proto` when both are enabled

ramgdev:
> Overall looks good. My concern is mostly around `http-json` overriding with `http-proto` when both are enabled

That's where I had most trouble changing the code. It seems that having a single default protocol is the source of the issue. One cannot have both `http-json` and `http-photo` unless they attach it to a specific signal, right? Ultimately the export is going to happen on only format. That means we have to build this default using a key that combines both format and signal rather than just the format.

opentelemetry-otlp/src/exporter/http/metrics.rs
ramgdev: Are we still talking about exporting in one format per signal per pipeline? If so, it should be possible with the above code, no? For e.g I can enable both features `http-json` and `http-photo` and build 2 trace pipelines that export in each of those formats. Let me know if I'm missing something.

TommyCpp: Yes by tuning the features it's possible to export in either `proto` or `json`. My preference is to ask users to explicitly tell the pipeline which protocol to rather then to config it using features. i.e it should be possible to enable both `http-json` and `http-proto` but choose to use `http-proto`. I think it current setup doesn't support this.

TommyCpp:
> It seems that having a single default protocol is the source of the confusion.

Yeah I think it's reasonable to rely on features if the user didn't provide any configuration on which protocol to use. But we should allow(or even promote) configure the protocol explicitly

opentelemetry-otlp/src/exporter/http/metrics.rs
ramgdev: Understood. Would you want `HttpExporterBuilder` to receive protocol via `with_protocol`?

ramgdev:
>I think it's reasonable to rely on features

This is still a bit tricky when both `http-json` and `http-proto` features are enabled. Are we ok to pick `http-json` in that case?  Or should we `require` the protocol to be specified when building the pipeline for `http`?

TommyCpp:
> Are we ok to pick `http-json` in that case? Or should we `require` the protocol to be specified when building the pipeline for `http`?

Yeah I think it's OK for us to pick `http-json` if user didn't provide one. i.e have a feature override in `default_protocol`. But in builder we should give users option to override it. Let me know if it makes sense

ramgdev:
@TommyCpp sorry, just got to this. Please take a look when you get a chance. Thanks.

hdost (approved):

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -28,18 +28,29 @@ pub const OTEL_EXPORTER_OTLP_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_PROTOCOL";
 /// Compression algorithm to use, defaults to none.
 pub const OTEL_EXPORTER_OTLP_COMPRESSION: &str = "OTEL_EXPORTER_OTLP_COMPRESSION";
 
-#[cfg(feature = "http-proto")]
+#[cfg(feature = "http-json")]
+/// Default protocol, using http-json.
+pub const OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT: &str = OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON;
```
lalitb, line 33: Will `http/json`  be the default if all or any two (including `http-json`) feature is enabled? If I remember correctly, http/protobuf was the default protocol in that case as per the specs.

lalitb (approved):
Thanks for the support. Nicely done.

opentelemetry-otlp/src/exporter/mod.rs
ramgdev: The spec says `SDK-dependent, but will typically be either http/protobuf or grpc`.  The feature `http-json` is not enabled in default feature list for `opentelemetry_otlp` crate. If someone is enabling it explicitly, it is reasonable to assume they want to go with `http/json` as the protocol imho. I'm open to change if that is not the case.

cijothomas:
Thanks @ramgdev for your first PR! Hoping to see many more, if your time/passion permits!

opentelemetry-otlp/src/exporter/mod.rs
lalitb: Thanks. This is what I see in the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/e6eccba97ebaffbbfad6d4358408a2cead0ec2df/specification/protocol/exporter.md#specify-protocol):

``
If no configuration is provided the default transport SHOULD be http/protobuf unless SDKs have good reasons to choose grpc as the default (e.g. for backward compatibility reasons when grpc was already the default in a stable SDK release).
``

This means the default enabled in `Cargo.toml` should actually be `http-proto`  and not [grpc-tonic](https://github.com/open-telemetry/opentelemetry-rust/blob/cc80ec6aa46cb42b61e935d5146ae58e115d3580/opentelemetry-otlp/Cargo.toml#L64).

Not related to this PR, but separate issue :)

ramgdev: Ah. I must have referred to the doc instead https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol

Agree, the spec requires it to be `http-proto` as default. Will open an issue and try to address it.

cijothomas: Thanks @ramgdev

PR #1586, opened 2024-02-27
Rewrite readme page for opentelemetry crate

cijothomas:
https://crates.io/crates/opentelemetry The description shown in the crates page is very generic, so this PR is trying to make it crate specific. Not adding examples/getting-started, as it is probably better to stay in docs.rs.

If the direction is good, will fix every other crate (over time...)

### DIFF
--- a/opentelemetry/README.md
+++ b/opentelemetry/README.md
@@ -1,10 +1,6 @@
-![OpenTelemetry — An observability framework for cloud-native software.][splash]
-
-[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
-
 # OpenTelemetry Rust
 
-The Rust [OpenTelemetry](https://opentelemetry.io/) implementation.
+This crate contains the [OpenTelemetry](https://opentelemetry.io/) API for Rust.
 
 [![Crates.io: opentelemetry](https://img.shields.io/crates/v/opentelemetry.svg)](https://crates.io/crates/opentelemetry)
 [![Documentation](https://docs.rs/opentelemetry/badge.svg)](https://docs.rs/opentelemetry)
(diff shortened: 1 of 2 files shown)


opentelemetry/README.md
```diff
@@ -15,124 +11,84 @@ The Rust [OpenTelemetry](https://opentelemetry.io/) implementation.
 
 ## Overview
 
-OpenTelemetry is a collection of tools, APIs, and SDKs used to instrument,
-generate, collect, and export telemetry data (metrics, logs, and traces) for
-analysis in order to understand your software's performance and behavior. You
-can export and analyze them using [Prometheus], [Jaeger], and other
-observability tools.
+OpenTelemetry is an Observability framework and toolkit designed to create and
```
cijothomas, line 14: this can be copied for every crate. (I copied from opentelemetry.io)

opentelemetry/README.md
```diff
@@ -15,124 +11,84 @@ The Rust [OpenTelemetry](https://opentelemetry.io/) implementation.
...
+various exporter crates such as
+[opentelemetry-otlp](https://crates.io/crates/opentelemetry-otlp). This
+architecture ensures that the final application can light up the instrumentation
+by integrating an SDK implementation.
+
+Library authors are recommended to depend on this crate *only*. This approach is
+also aligned with the design philosophy of existing telemetry solutions in the
+Rust ecosystem, like `tracing` or `log`, where these crates only offer a facade
+and the actual functionality is enabled through additional crates.
+
```
hdost, line 88: It would be good to add a reference to the registry https://opentelemetry.io/ecosystem/registry/?language=rust for Exporters, Propagator, Instrumentation, and more.
I took the liberty to add the missing elements this morning https://github.com/open-telemetry/opentelemetry.io/pull/4062

hdost (commented):
Looks good overall! Small request.

lalitb (approved):
LGTM. nicely done.

opentelemetry/README.md
cijothomas: Thanks! 
I am going to skip that suggestion in this PR, and add it next. I am still thinking what is the best way to ensure success - simply linking to registry won't necessarily help much unless the user is already familiar with OTel space.
Will come back to this.

PR #1597, opened 2024-03-04
Add test to confirm bug in metrics sdk

cijothomas:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -590,6 +590,113 @@ mod tests {
         );
     }
 
+    // "multi_thread" tokio flavor must be used else flush won't
+    // be able to make progress!
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[ignore = "Bug bug."]
+    async fn delta_memory_efficiency_test() {
+        // Run this test with stdout enabled to see output.
+        // cargo test delta_memory_efficiency_test --features=metrics,testing -- --nocapture
+
+        // Arrange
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -590,6 +590,113 @@ mod tests {
         );
     }
 
+    // "multi_thread" tokio flavor must be used else flush won't
+    // be able to make progress!
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[ignore = "Bug bug."]
```
cijothomas, line 596: don't have a tracking issue. There are open PRs touching aggregations, and more to come, and will review how best to fix this considering upcoming/pending changes.

lalitb: Can we add a tracking issue if not already there? And add more details why is this being ignored. I assume it is related to discussion here - https://cloud-native.slack.com/archives/C03GDP0H023/p1709591127115979

lalitb: There should be a way to ignore the specific assertion, rather than whole test, probably we can figure that separately.

lalitb: I think it's good to have tracking issues for things not working, even if we plan to revisit the design.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/1598

lalitb (approved):
Thanks for validating the issue.

PR #1600, opened 2024-03-05
Add more tests to Metric SDK aggregation

cijothomas:
Adding more tests to test meter and instrument identify part, and whether they get merged or not.

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -870,12 +870,7 @@ mod tests {
         metrics: &'a Vec<ScopeMetrics>,
         name: &'a str,
     ) -> Option<&'a ScopeMetrics> {
-        for scope_metric in metrics {
-            if scope_metric.scope.name == name {
-                return Some(scope_metric);
-            }
-        }
-        None
+        metrics.iter().find(|&scope_metric| scope_metric.scope.name == name)
     }
 
     struct DeltaTemporalitySelector();
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -283,7 +283,7 @@ mod tests {
             let datapoint1 = &sum1.data_points[0];
             assert_eq!(datapoint1.value, 10);
         } else {
-            assert!(false, "No MetricScope found for 'test.meter1'");
+            panic!("No MetricScope found for 'test.meter1'");
         }
 
         if let Some(scope2) = scope2 {
@@ -303,7 +303,7 @@ mod tests {
             let datapoint2 = &sum2.data_points[0];
             assert_eq!(datapoint2.value, 5);
         } else {
-            assert!(false, "No MetricScope found for 'test.meter2'");
+            panic!("No MetricScope found for 'test.meter2'");
         }
     }
 
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -867,7 +867,7 @@ mod tests {
     }
 
     fn find_scope_metric<'a>(
-        metrics: &'a Vec<ScopeMetrics>,
+        metrics: &'a [ScopeMetrics],
         name: &'a str,
     ) -> Option<&'a ScopeMetrics> {
         metrics.iter().find(|&scope_metric| scope_metric.scope.name == name)
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -60,6 +60,8 @@ pub use view::*;
 
 #[cfg(all(test, feature = "testing"))]
 mod tests {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
...
+        // Arrange
+        let exporter = InMemoryMetricsExporter::default();
+        let reader = PeriodicReader::builder(exporter.clone(), runtime::Tokio).build();
+        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
+
+        // Act
+        // Meters are identical except for scope attributes, but scope attributes are not an identifying property.
+        // Hence there should be a single metric stream output for this test.
+        let meter1 = meter_provider.versioned_meter(
+            "test.meter1",
```
lalitb, line 323:
```suggestion
            "test.meter",
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
...
+            Some("v0.1.0"),
+            Some("schema_url"),
+            Some(vec![KeyValue::new("key", "value1")]),
+        );
+        let meter2 = meter_provider.versioned_meter(
+            "test.meter1",
```
lalitb, line 329:
```suggestion
            "test.meter",
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
...
+            resource_metrics[0].scope_metrics.len() == 1,
+            "There should be a single scope as the meters are identical"
+        );
+        assert!(
+            resource_metrics[0].scope_metrics[0].metrics.len() == 1,
+            "There should be single metric for the scope as instruments are identical"
+        );
+
+        let scope = &resource_metrics[0].scope_metrics[0].scope;
+        assert_eq!(scope.name, "test.meter1");
```
lalitb, line 367:
```suggestion
        assert_eq!(scope.name, "test.meter");
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
...
+        assert_eq!(scope.version, Some(Cow::Borrowed("v0.1.0")));
+        assert_eq!(scope.schema_url, Some(Cow::Borrowed("schema_url")));
+
+        // Should we validate this, as this is a user error
+        assert_eq!(scope.attributes, vec![KeyValue::new("key", "value1")]);
```
lalitb, line 372: Can let it be here. I think specs doesn't say anything whether the attributes should get overwritten or not/.

lalitb (approved):
Thanks for the tests.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
         assert_eq!(datapoint.value, 15);
     }
 
+    // "multi_thread" tokio flavor must be used else flush won't
+    // be able to make progress!
```
TommyCpp, line 220: nit: don't think we need this for every test but not a strong opinion

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
...
+
+        let counter2 = meter2
+            .u64_counter("my_counter")
+            .with_unit(Unit::new("my_unit"))
+            .with_description("my_description")
+            .init();
+
+        let attribute = vec![KeyValue::new("key1", "value1")];
+        counter1.add(10, &attribute);
+        counter2.add(5, &attribute);
```
TommyCpp, line 245: We can just loop over `test.meter1` and `test.meter2` here?

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -214,6 +216,178 @@ mod tests {
...
+                .expect("Sum aggregation expected for Counter instruments by default");
+
+            // Expecting 1 time-series.
+            assert_eq!(sum2.data_points.len(), 1);
+
+            let datapoint2 = &sum2.data_points[0];
+            assert_eq!(datapoint2.value, 5);
+        } else {
+            panic!("No MetricScope found for 'test.meter2'");
+        }
```
TommyCpp, line 307: We can probably simplify it a little bit if we use for loop here.

TommyCpp (commented):
the test looks good. I'd add some more details on what we are testing here. 

I personally like table driven unit tests so if we can use for loop here may save us some duplication

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: agree. i am intentionally writing these tests without moving things to helper methods as I don't have a full grasp yet on how much we should leverage these tests vs something else. Once I get better idea, will move to helper methods wherever applicable, and remove all the code duplication!

PR #1603, opened 2024-03-06
Add External Type Checking

hdost:
The purpose of this is to ensure we understand the API surface area. Anything exposed by a crate like a type can affect the possible compatibility. This means that if we bump an external type it could cause breaking changes. This tracks them to limit this possibility.

Affected Crates:
- opentelemetry-otlp
- opentelemetry-zipkin

Fixes #

## Changes

### DIFF
--- a/opentelemetry-otlp/allowed-external-types.toml
+++ b/opentelemetry-otlp/allowed-external-types.toml
@@ -1,0 +1,22 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+# This is used with cargo-check-external-types to reduce the surface area of downstream crates from
+# the public API. Ideally this can have a few exceptions as possible.
+allowed_external_types = [
+    "opentelemetry::*",
+    "opentelemetry_http::*",
+    "opentelemetry_sdk::*",
+    # http is a pre 1.0 crate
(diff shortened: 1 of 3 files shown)


opentelemetry-otlp/allowed-external-types.toml
```diff
@@ -0,0 +1,22 @@
+# Copyright The OpenTelemetry Authors
```
lalitb, line 1: (sorry to nit pick) - Not related to this PR, we don't include copyright/license headers in individual files, we have the `LICENSE` file in each crate, and also license information in `crates.toml`. I noticed that the header is present in other `allowed-external-types.toml`'s too, which seems like something we should address in a separate cleanup :)

lalitb (approved):
LGTM. Probably we should be adding this static external type check for other crates not having it separately.

opentelemetry-otlp/allowed-external-types.toml
TommyCpp: +1

TommyCpp (approved):

hdost:
> LGTM. Probably we should be adding this static external type check for other crates not having it separately.

To my knowledge the tool only supports a per crate configuration ot workspace level.

PR #1607, opened 2024-03-08
Adding noop metrics benchmark

lalitb:
Fixes #

## Changes
Adding benchmark for Counter and Vector of KeyValue creation. Rest of the instruments should have benchmarks same as the 
counter. 

Results:
**NoopCounter_NoAttributes:**  1.5685 ns 
**NoopCounter_AddWithInlineStaticAttributes:** 15.452 ns
**NoopCounter_AddWithStaticArray:** 15.354 ns
**NoopCounter_AddWithDynamicAttributes:** 25.692 ns
**CreateVector_KeyValue:**  24.515 ns
**CreateDynamicVector_StringPair:** 117.25 ps


opentelemetry-sdk/benches/noop_metrics.rs
```diff
@@ -0,0 +1,85 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use opentelemetry::{
```
cijothomas, line 2: since we don't need sdk for these tests, could we keep them as part of opentelemetry project itself?

opentelemetry-sdk/benches/noop_metrics.rs
```diff
@@ -0,0 +1,85 @@
...
+    create_keyvalue_vector(c);
+}
+
+fn noop_counter_add(c: &mut Criterion) {
+    let attribute_values = [
+        "value1", "value2", "value3", "value4", "value5", "value6", "value7", "value8", "value9",
+        "value10",
+    ];
+
+    let mut rng = SmallRng::from_entropy();
```
cijothomas, line 27: since we are measuring noops only, there is no need of random attributes.

lalitb: Yes, kept the code as it is not costed as part of the benchmark. Will make it static array.

opentelemetry-sdk/benches/noop_metrics.rs
lalitb: yes good point. Will move it there.

TommyCpp (approved):
LGTM

hdost (approved):

cijothomas:
> CreateVector_KeyValue: 24.515 ns
> CreateDynamicVector_StringPair: 117.25 ps

This is....interesting. Wondering if it is due to copy or clone or something else inside Key,Value,KeyValue in `opentelemetry`! Should steal/learn/copy from how `tracing` handles similar needs.

lalitb:
> > CreateDynamicVector_StringPair: 117.25 ps
> 
> This is....interesting. Wondering if it is due to copy or clone or something else inside Key,Value,KeyValue in `opentelemetry`! Should steal/learn/copy from how `tracing` handles similar needs.

Yes, `117.25 ps`  is interesting,  it appears to be shorter than the instruction cycle. This might not reflect the actual cost, - the compiler could be doing some optimization in terms of creating the tuples for the vector at compile time. Or worst, the vector though created is not being used, so compiler might eliminate this code altogether. Let me check if it is possible to turn off compiler optimization in the benchmark tests and get true cost.

cijothomas:
> > > CreateDynamicVector_StringPair: 117.25 ps
> > 
> > 
> > This is....interesting. Wondering if it is due to copy or clone or something else inside Key,Value,KeyValue in `opentelemetry`! Should steal/learn/copy from how `tracing` handles similar needs.
> 
> Yes, `117.25 ps` is interesting, it appears to be shorter than the instruction cycle. This might not reflect the actual cost, - the compiler could be doing some optimization in terms of creating the tuples for the vector at compile time. Or worst, the vector though created is not being used, so compiler might eliminate this code altogether. Let me check if it is possible to turn off compiler optimization in the benchmark tests and get true cost.

Yes please! We need to make sure compiler treat this as a semi-black-box.
But I was more curious about the 24ns it took when only "no-op" is in play!

lalitb:
> But I was more curious about the 24ns it took when only "no-op" is in play!

There should be the any internal copy happening for Key and Value in this specific test. As we are passing string literals for Key and Value, they should just get type-converted to [OtelString::Static](https://github.com/open-telemetry/opentelemetry-rust/blob/e816cb926df990c64ff6c958fa0a38912acab94a/opentelemetry/src/common.rs#L141). The run time memory overhead should be for the vector memory allocation, which stores the KeyValue objects, which internally refer to the static string liternals.

cijothomas:
> > But I was more curious about the 24ns it took when only "no-op" is in play!
> 
> There should not be the any internal copy happening for Key and Value in this specific test. As we are passing string literals for Key and Value, they should just get type-converted to [OtelString::Static](https://github.com/open-telemetry/opentelemetry-rust/blob/e816cb926df990c64ff6c958fa0a38912acab94a/opentelemetry/src/common.rs#L141). The run time memory overhead should be for the vector allocation, which stores the KeyValue objects, which internally refer to the static string liternals.

Yes I think so too. What is(was) misleading is that, the `CreateDynamicVector_StringPair` shows that allocating vec is cheap. But as you said, its most likely due to compiler optimizations, and not doing what we think it is doing!!

PR #1609, opened 2024-03-08
Avoid enabling the transport feature of tonic if not needed

sandhose:
## Changes

This disables the default features of `tonic` in `opentelemetry-proto`, so that the `transport` feature doesn't get enabled by default.
This shrinks the dependency tree when using `opentelemetry-otlp` with only the HTTP transport not the GRPC one.

* N/A Unit tests added/updated (if applicable)
* N/A Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* N/A Changes in public API reviewed (if applicable)

### DIFF
--- a/opentelemetry-proto/Cargo.toml
+++ b/opentelemetry-proto/Cargo.toml
@@ -48,7 +48,7 @@ with-schemars = ["schemars"]
 with-serde = ["serde", "hex"]
 
 [dependencies]
-tonic = { workspace = true, optional = true, features = ["codegen", "prost"] }
+tonic = { workspace = true, optional = true, default-features = false, features = ["codegen", "prost"] }
 prost = { workspace = true, optional = true }
 opentelemetry = { version = "0.22", default-features = false, path = "../opentelemetry" }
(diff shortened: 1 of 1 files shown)


opentelemetry-proto/Cargo.toml
```diff
@@ -48,7 +48,7 @@ with-schemars = ["schemars"]
 with-serde = ["serde", "hex"]
 
 [dependencies]
-tonic = { workspace = true, optional = true, features = ["codegen", "prost"] }
```
TommyCpp, line 51: Since we are using workspace dependencies can we disable it in workspace?

cijothomas: @sandhose Could you make this suggestion?

TommyCpp:
Actually we are already have the default feature disabled in workspace. If you still see it somewhere, could you list the feature set you are using?

sandhose:
My bad, for some reason I thought the feature got enabled in my tree, but that was just me not looking long enough at my dependency tree… 😅 
And of course, I opened this PR without actually checking and then was out of office for a week… sorry about that!

PR #1612, opened 2024-03-11
Use regular lock for simplespanprocessor

### LINKED ISSUE #1409: [Feature]: SimpleSpanProcessor to be consistent with SimpleLogRecordProcessor
### Related Problems?

_No response_

https://github.com/open-telemetry/opentelemetry-rust/pull/1308 modified SimpleLogRecordProcessor to simply take a Mutex protected call to Exporter. 
Proposing to make the similar change to SimpleSpanProcessor, effectively reversing https://github.com/open-telemetry/opentelemetry-rust/pull/502.

We'll get to eliminate the crossbeam-channel by doing this, and also ensures there is more consistency between signals.

### Considered Alternatives

_No response_

### Additional Context

_No response_

cijothomas:
Fixes #1409

## Changes

Match SimpleSpanProcessor to align with SimpleLogRecordProcessor.
Still haven't addressed [this](https://github.com/open-telemetry/opentelemetry-rust/pull/1308/files#r1367122070) yet.

### DIFF
--- a/opentelemetry-sdk/src/testing/trace/span_exporters.rs
+++ b/opentelemetry-sdk/src/testing/trace/span_exporters.rs
@@ -7,13 +7,15 @@ use crate::{
     InstrumentationLibrary,
 };
 use async_trait::async_trait;
-use crossbeam_channel::{unbounded, Receiver, SendError, Sender};
 use futures_util::future::BoxFuture;
 pub use opentelemetry::testing::trace::TestSpan;
 use opentelemetry::trace::{
     SpanContext, SpanId, SpanKind, Status, TraceFlags, TraceId, TraceState,
 };
-use std::fmt::{Display, Formatter};
+use std::{
+    fmt::{Display, Formatter},
+    sync::{Arc, Mutex},
+};
 
 pub fn new_test_export_span_data() -> SpanData {
     let config = Config::default();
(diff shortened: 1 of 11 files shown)


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -729,17 +683,17 @@ mod tests {
 
     #[test]
     fn simple_span_processor_on_end_calls_export() {
-        let (exporter, rx_export, _rx_shutdown) = new_test_exporter();
-        let mut processor = SimpleSpanProcessor::new(Box::new(exporter));
+        let exporter = TestSpanExporter::new();
```
cijothomas, line 686: I dont know if this approach for testing is any better than the channels based approach earlier. If desired I can use the channels based approach back, but without using cross-bean, but using channels from std itself.

TommyCpp: Let's build add a buffer in TestSpanExporter to save all spans exported?

opentelemetry-sdk/src/testing/trace/span_exporters.rs
```diff
@@ -40,42 +42,47 @@ pub fn new_test_export_span_data() -> SpanData {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct TestSpanExporter {
-    tx_export: Sender<SpanData>,
-    tx_shutdown: Sender<()>,
+    pub export_called: Arc<Mutex<bool>>,
+    pub shutdown_called: Arc<Mutex<bool>>,
```
TommyCpp, line 48: We can use `AtomicBool` for simpliciy

opentelemetry-sdk/src/trace/span_processor.rs
cijothomas: that'd be using inmemoryexporter then right? This test is merely testing if simpleprocessors calls exporter.export or not.

TommyCpp: Feels like we can deprecate `TestSpanExporter` in favor of the in memory span exporter?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -707,6 +660,7 @@ where
 
 #[cfg(all(test, feature = "testing", feature = "trace"))]
 mod tests {
+    // cargo test trace::span_processor::tests:: --features=trace,testing
```
TommyCpp, line 663: nit: do we need this? the feature flag above seems to be clear to me what features needed to run this test

opentelemetry-sdk/src/trace/span_processor.rs
cijothomas: yes that'd be good. let me tackle it as a follow up. There could be other places where we create testexporters, and that can be replaced with InMemoryExporter

TommyCpp: 👍

opentelemetry-sdk/src/trace/span_processor.rs
cijothomas: I put it there just to make it easy for other devs to run these tests by copying and running the command, instead of having to type it manually every time! I use VSCode, and it does not offer any guided UI experience to run tests from a particular file.

TommyCpp:
Given the perf concern in #1409 I think it's worth call out explicitly in the comments of `SimpleSpanProcessor` that throughput may not be ideal and if one is looking for performance they should consider `BatchSpanProcessor`

cijothomas:
> Given the perf concern in #1409 I think it's worth call out explicitly in the comments of `SimpleSpanProcessor` that throughput may not be ideal and if one is looking for performance they should consider `BatchSpanProcessor`

added for span and log processor.

opentelemetry-sdk/src/testing/trace/span_exporters.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/pull/1612#discussion_r1520864465

TommyCpp (approved):
LGTM!

lalitb (approved):

Issue #1590, opened 2024-02-29
urlencoding dependency removal


cijothomas:
See https://github.com/open-telemetry/opentelemetry-rust/pull/1578#discussion_r1507777504

TommyCpp:
close via #1615

PR #1615, opened 2024-03-12
remove dependency of urlencoding crate

sreeo:
Fixes #
#1590 

## Changes
Remove urlencoding dependency.

### DIFF
--- a/opentelemetry/src/baggage.rs
+++ b/opentelemetry/src/baggage.rs
@@ -16,10 +16,10 @@
 //! [W3C Baggage]: https://w3c.github.io/baggage
 use crate::{Context, Key, KeyValue, Value};
 use once_cell::sync::Lazy;
+use std::borrow::Cow;
 use std::collections::{hash_map, HashMap};
 use std::fmt;
 use std::iter::FromIterator;
-use urlencoding::encode;
 
 static DEFAULT_BAGGAGE: Lazy<Baggage> = Lazy::new(Baggage::default);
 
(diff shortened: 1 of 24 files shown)


cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1590

opentelemetry/src/baggage.rs
```diff
@@ -282,10 +282,29 @@ impl FromIterator<KeyValueMetadata> for Baggage {
     }
 }
 
+fn encode(s: Cow<'_, str>) -> String {
```
lalitb, line 285: nit - can this function take `&str` as it is not modifying the input string?

opentelemetry/src/baggage.rs
```diff
@@ -282,10 +282,29 @@ impl FromIterator<KeyValueMetadata> for Baggage {
...
+    let special_characters = ['.', '-', '_', '~'];
+    let mut encoded_string = String::new();
```
lalitb, line 287: nit - we can preallocate the encoded string with the required length `s.len()`?

opentelemetry/src/baggage.rs
```diff
@@ -282,10 +282,29 @@ impl FromIterator<KeyValueMetadata> for Baggage {
...
+
+    for char in s.as_ref().chars() {
+        if char.is_ascii_alphanumeric() || special_characters.contains(&char) {
+            encoded_string.push(char);
+        } else if char == ' ' {
+            encoded_string.push_str("%20");
+        } else {
+            // Convert to a two-digit hexadecimal representation
+            for byte in char.to_string().as_bytes() {
```
lalitb, line 296: That seems inefficient, converting char to string, and then to bytes. Instead, can we directly iterate over the original string as bytes?

opentelemetry/src/baggage.rs
sreeo: yep

opentelemetry/src/baggage.rs
sreeo: yep

opentelemetry/src/baggage.rs
sreeo: yeah converting to char makes it easy to encode non ascii chars which could be multi byte.

lalitb: I meant something like this (not tested):
```rust
  for byte in s.as_bytes() {
        match *byte {
                    //alphanumeric...
                   b'a'.. b'z' | b 'A'.. b'Z' | ... => encoded_string.push(*byte as char);
                   b' ' =>  encoded_string.push_str("%20"),
                    //percentile encode all other type
                   _  =>  encoded_string.push_str(&format!("%{:02X}", byte)),
        }
  }
```

hdost (approved):

opentelemetry/src/baggage.rs
sreeo: hmm, let me try this out.

opentelemetry/src/baggage.rs
```diff
@@ -282,10 +281,53 @@ impl FromIterator<KeyValueMetadata> for Baggage {
...
+                    i = end - 1;
+                }
+            }
+        }
+
+        i += 1;
+    }
+
+    encoded_string
+}
```
lalitb, line 325: @sreeo Not sure if I am missing something, but we don't need separate handling for ASCII (non-alphanumeric) and UTF-8. Both need to be byte-wise percentile-encoded. I tested your unit-test examples on something I suggested earlier, and seems to work - https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f433e6f80889b0680bab004f926ecaa.  Though it is difficult to estimate the encoded size, it should be fine to have initially set to the original string size, and not over-estimate. Sorry for being bit critical from perf prospective here :)

sreeo: No worries, your suggested code is simpler than the current one and passes all tests. Replaced it.

lalitb (approved):
Thanks for the fix.

Issue #1514, opened 2024-02-07
Add Testing for Shutdown Scenarios.


hdost:
> > And will calling shutdown on the global loggerProvider ensure that the logs through these loggers are not exported ?
> 
> It would be good to validate this use-case. It seems [shutdown_logger_provider](https://github.com/open-telemetry/opentelemetry-rust/blob/f9b08c10809e3b39497e21cccaa0d0c49b0216dc/opentelemetry/src/global/logs.rs#L134) will invoke shutdown on the LogProcessors when the inner instance is [dropped](https://github.com/open-telemetry/opentelemetry-rust/blob/f9b08c10809e3b39497e21cccaa0d0c49b0216dc/opentelemetry-sdk/src/logs/log_emitter.rs#L116). And now if any of the loggers have it's reference in separate thread, the shutdown will not be invoked. One option can be to add a new `LoggerProvider::Shutdown()` method, and explicitly invoke it from within `shutdown_logger_provider` . This new method will invoke shutdown on all processors, and so ensure that all the existing logs are flushed. And no new logs can be exported with processors in shutdown state.

Thinking about this scenario again,  it seems that the potential issue of leaking providers might not be as significant if we can document it clearly to explain the risk. Specifically, providers could be leaked if any of the tasks indefinitely retain their reference, which can result in a shutdown not triggering.

_Originally posted by @lalitb in https://github.com/open-telemetry/opentelemetry-rust/issues/1455#issuecomment-1877690239_

bIgBV:
Looking into this, I'd love some clarification on how the scenario would work. My current approach has been to set up an integration test based on https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/logs-basic/src/main.rs and extend it such that we have multiple background tasks with set up with references to the `LoggerProvider` instance through an `Arc`.

The shutdown I wanted to test was to create a forced shutdown through an interrupt. I initially tried this with setting up multiple exporters with a single `LoggerProvider` and got a panic from this line https://github.com/open-telemetry/opentelemetry-rust/blob/36c2a8e0a3c7077b824af78119bfa72ce31a6cd6/opentelemetry-sdk/src/logs/log_emitter.rs#L86 which I think is expected behavior. 

Would taking the same approach with multiple `LoggerProvider` references in separate tasks be a similar scenario?

hdost:
[...]

PR #1620, opened 2024-03-13
Add test to validate log shutdown

cijothomas:
Related to https://github.com/open-telemetry/opentelemetry-rust/issues/1514, not quite the finish, but just a initial test.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -13,7 +13,7 @@ use opentelemetry::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 
-use std::{borrow::Cow, sync::Arc, thread};
+use std::{borrow::Cow, sync::Arc};
 
 #[derive(Debug, Clone)]
 /// Creator for `Logger` instances.
@@ -319,7 +319,7 @@ mod tests {
         // logger. Since that is the last logger, the provider (inner provider)
         // is finally dropped, triggering shutdown
         *signal_to_end.lock().unwrap() = true;
-        let _h = handle.join().unwrap();
+        handle.join().unwrap();
         assert!(*shutdown_called.lock().unwrap());
 
         // flush is never called by the sdk.
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -13,7 +13,7 @@ use opentelemetry::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 
-use std::{borrow::Cow, sync::Arc};
+use std::{borrow::Cow, sync::Arc, thread};
 
 #[derive(Debug, Clone)]
 /// Creator for `Logger` instances.
@@ -243,6 +243,7 @@ mod tests {
     use opentelemetry::global::{logger, set_logger_provider, shutdown_logger_provider};
     use opentelemetry::logs::Logger;
     use std::sync::{Arc, Mutex};
+    use std::thread;
 
     #[test]
     fn shutdown_test() {
@@ -251,6 +252,8 @@ mod tests {
         // Arrange
         let shutdown_called = Arc::new(Mutex::new(false));
         let flush_called = Arc::new(Mutex::new(false));
+        let signal_to_end = Arc::new(Mutex::new(false));
+        let signal_to_thread_started = Arc::new(Mutex::new(false));
         let logger_provider = LoggerProvider::builder()
             .with_log_processor(LazyLogProcessor::new(
                 shutdown_called.clone(),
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -57,4 +58,51 @@ mod tests {
...
+    fn logging_sdk_shutdown_test() {
+        // Arrange
+        let exporter: InMemoryLogsExporter = InMemoryLogsExporter::default();
+        let logger_provider = LoggerProvider::builder()
+            .with_log_processor(SimpleLogProcessor::new(Box::new(exporter.clone())))
+            .build();
+        set_logger_provider(logger_provider);
+
+        // Act
+        let logger = logger("test-logger");
```
cijothomas, line 72: this *MUST* occur after we set global::set_logger_provider, else it'll get permanently tied to noop provider. The current implementation does not have a mechanism to update already obtained logger with updated meter_provider. (same for traces,metrics too)

Just an observation. I'll send some docs updates to make this clear to end users. (open to discuss more on if this is the desired behavior or not!)

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -57,4 +58,51 @@ mod tests {
...
+            .first()
+            .expect("Atleast one log is expected to be present.");
+        assert_eq!(log.instrumentation.name, "test-logger");
+        assert_eq!(log.record.severity_number, Some(Severity::Error));
+        let attributes: Vec<(Key, AnyValue)> = log
+            .record
+            .attributes
+            .clone()
+            .expect("Attributes are expected");
+        assert_eq!(attributes.len(), 2);
```
lalitb, line 106: Should we validate for exported logs before doing shutdown, as SimpleLogProcessor will export them synchronously. And to ensure that shutdown has really happened, try calling emit() on existing logger instance. There shouldn't be any exporter happening this time.

cijothomas: thats a good point. I should use batchprocessor here instead. Or a custom one, which never publishes logs to exporter, except when shutdown/flush called - i think this is the best way to test

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -236,3 +236,99 @@ impl opentelemetry::logs::Logger for Logger {
...
+        // Drop one of the logger. Still not enough!
+        drop(logger1);
+        assert!(!*shutdown_called.lock().unwrap());
+
+        // drop logger2, which is the only remaining logger, and this will
+        // finally drop the inner provider, triggering shutdown.
+        drop(logger2);
+        assert!(*shutdown_called.lock().unwrap());
+
+        // flush is never called by the sdk.
```
hdost, line 292: So in this case this is false because we are intentionally testing SDK. The actual log processors will run the flush if they are shutdown.

hdost (approved):

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: sort of...

The built-in processors we ship (simple, batch), won't have to call flush(), but do the _equivalent_ of flush. In a diff. test, I intend to test that it is indeed causing the effect of flush(), but not necessarily by calling flush() itself.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#shutdown-1
_Shutdown MUST include the effects of ForceFlush._

lalitb (approved):
Thanks. Nice way to validate the scenario described in #1514 :)

lalitb:
> In a diff. test, I intend to test that it is indeed causing the effect of flush(), but not necessarily by calling flush() itself.

Looking into the batch processor code, it seems to be doing the flush before shutdown. But good to validate this.

https://github.com/open-telemetry/opentelemetry-rust/blob/5ef3fd31dfbaffe5d31b97be9663a010bf5c6f8c/opentelemetry-sdk/src/logs/log_processor.rs#L212-L220

cijothomas:
@hdost @lalitb Could you re-review? I added a separate-thread-logger scenario too.

cijothomas:
> > In a diff. test, I intend to test that it is indeed causing the effect of flush(), but not necessarily by calling flush() itself.
> 
> Looking into the batch processor code, it seems to be doing the flush before shutdown. But good to validate this in separate test.
> 
> https://github.com/open-telemetry/opentelemetry-rust/blob/5ef3fd31dfbaffe5d31b97be9663a010bf5c6f8c/opentelemetry-sdk/src/logs/log_processor.rs#L212-L220

yes. that'd be a targetted test for processors to ensure shutdown() gives same effect as force_flush().

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -236,3 +236,132 @@ impl opentelemetry::logs::Logger for Logger {
...
+                if *signal_to_end_clone.lock().unwrap() {
+                    break;
+                }
+            }
+        });
+
+        // wait for the spawned thread to start before calling shutdown This is
+        // very important - if shutdown is called before the spawned thread
+        // obtains its logger, then the logger will be no-op one, and the test
+        // will pass, but it will not be testing the intended scenario.
```
lalitb, line 289: ok, this is to avoid `NoopLoggerProvider` getting used?

lalitb (approved):
Nice :)

cijothomas:
@hdost @TommyCpp I'll go ahead and merge this to continue working on this. Please let me know if you have any feedback on this approach.

PR #1622, opened 2024-03-14
Remove unnecessary flush from tests

cijothomas:
SimpleProcessor is used in tests, and flush is not required.

### DIFF
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -110,7 +110,10 @@ mod tests {
         span.set_attribute(KeyValue::new("attribute1", "value1"));
         span.add_event("test-event".to_string(), vec![]);
         span.set_status(Status::error("cancelled"));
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -112,8 +110,10 @@ mod tests {
         span.set_attribute(KeyValue::new("attribute1", "value1"));
         span.add_event("test-event".to_string(), vec![]);
         span.set_status(Status::error("cancelled"));
-        drop(span);
-        provider.force_flush();
+        span.end();
```
TommyCpp, line 113: Let's also test implicitly drop span will end it

cijothomas: that is covered in others tests in this file itself. So we have coverage for explicit end() and implicit via drop() both.

TommyCpp (approved):

Issue #1592, opened 2024-03-01
OTLP Exporter builder issues


cijothomas:
Methods like `install_batch` created unintended side-effects https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/span.rs#L143-L150
1. It sets global providers, which may not be what user has intended.
2. It returns a tracer with instrumentation scope set to "opentelemetry-otlp", which is incorrect as well.

The builder methods used in OTLP to setup the pipeline must be revisited to ensure that there are no accidental side effects like above. Also, returning tracer with hardcoded name is defeating the purpose of instrumentation scope, as all tracers will have same scope (and that too incorrect one). This is particularly hard when used with `tracer-opentelemetry`, as it'll result in traces which all have same scope!

cijothomas:
I used traces as an example - same/similar issue for all signals exist and need a unified fix.

mattbodd:
I'm interested in taking this on

NickLarsenNZ:
Adding a permalink (for the link in the description):
https://github.com/open-telemetry/opentelemetry-rust/blob/d5392dc1df13b51553577a51d736eac9b29e7125/opentelemetry-otlp/src/span.rs#L143-L150

cijothomas:
Both logs, and metrics have completed this cleanup. Neither returns a Logger/Meter, nor set a Global Provider.
Tracing is still needing to be fixed to make it consistent with the rest.

TommyCpp:
I will do it for the tracing pipelines

NickLarsenNZ:
> Both logs, and metrics have completed this cleanup. Neither returns a Logger/Meter, nor set a Global Provider.
> Tracing is still needing to be fixed to make it consistent with the rest.

Is the logs one linked anywhere in this issue? I didn't see it, but am on mobile and maybe missing the obvious.

lalitb:
> Is the logs one linked anywhere in this issue?

Not linked, but it was removed as part of global cleanups in #1691 .

cijothomas:
> > Is the logs one linked anywhere in this issue?
> 
> Not linked, but it was removed as part of global cleanups in #1691 .

@NickLarsenNZ The logging should not affect end users.. It is only going to affect authors of bridges/appenders.. Curious if you are writing one?

utpilla:
I can work on this if this still needs help.

cijothomas:
@ThomsonTan Will see if this is something he can take.

PR #1623, opened 2024-03-15
Add global::meter_provider_shutdown

### LINKED ISSUE #1619: [OTEL 0.22] shutdown SdkMeterProvider when built from opentelemetry-otlp?
### Discussed in https://github.com/open-telemetry/opentelemetry-rust/discussions/1605

<div type='discussions-op-text'>

<sup>Originally posted by **Oliboy50** March  6, 2024</sup>
## Context
We built a simple lib in my company which abstract away most of the OpenTelemetry complexity (using `tracing-opentelemetry` internally) so my colleagues and I just have to use our simple lib and `tracing` in all our company projects.

## Question

My question is about how to properly shutdown (if even needed?), the `SdkMeterProvider` used internally in this lib.

Because the `SdkMeterProvider` is built using the `opentelemetry-otlp` crate, and it disappears immediately in our lib internal API, such as:

```rust
pub fn build_metrics_layer() -> tracing_opentelemetry::MetricsLayer<S>
    where
        S: tracing::Subscriber + for<'span> tracing_subscriber::registry::LookupSpan<'span>,
{
    let meter_provider = opentelemetry_otlp::new_pipeline()
        .metrics(opentelemetry_sdk::runtime::Tokio)
        .with_exporter(some_exporter())
        .build()
        .expect("Failed to build OTLP meter provider");

    MetricsLayer::new(meter_provider)

    // meter_provider is gone here, 
    // so we cannot call `meter_provider.shutdown()` when closing the application :'(
}
```

Then in our company projects, we use `tracing` such as:

```rust
let registry = tracing_subscriber::registry()
    .with(tracing_subscriber::filter::EnvFilter::from_default_env())
    .with(build_metrics_layer());
tracing::subscriber::set_global_default(registry).expect("Failed to set global tracing subscribers");

// do our stuff, such as run an HTTP API server
// ...

// here we want to shutdown the HTTP API server
// but we can't since there is no more `opentelemetry::global::shutdown_meter_provider()` :'(
```

So how are we supposed to shutdown the SdkMeterProvider used internally, without using `opentelemetry::global::shutdown_meter_provider()`? 🙏 

I tried to see if `opentelemetry-otlp` or `tracing-opentelemetry` was doing the shutdown stuff for us, but I wasn't able to find something there :'(</div>

lalitb:
Fixes #1619 

## Changes

- Add `global::shutdown_meter_provider()` which internally sets the global provider to `NoopMeterProvider`.
- Add `Drop` implementation for `SdkMeterProvider`, which ensures that the pipeline shutdown is invoked once `SdkMeterProvider` goes out of scope.
- Update examples to use `global::shutdown_meter_provider()`
- 

### DIFF
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -1,3 +1,4 @@
+use opentelemetry::global::{meter_provider, set_meter_provider, shutdown_meter_provider};
 use opentelemetry::metrics::Unit;
 use opentelemetry::Key;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
@@ -7,7 +8,7 @@ use opentelemetry_sdk::metrics::{
 use opentelemetry_sdk::{runtime, Resource};
 use std::error::Error;
 
-fn init_meter_provider() -> SdkMeterProvider {
+fn init_meter_provider() {
     // for example 1
     let my_view_rename_and_unit = |i: &Instrument| {
         if i.name == "my_histogram" {
@@ -50,7 +51,7 @@ fn init_meter_provider() -> SdkMeterProvider {
         //   Ok(serde_json::to_writer_pretty(writer, &data).unwrap()))
         .build();
     let reader = PeriodicReader::builder(exporter, runtime::Tokio).build();
-    SdkMeterProvider::builder()
+    let provider = SdkMeterProvider::builder()
         .with_reader(reader)
         .with_resource(Resource::new(vec![KeyValue::new(
             "service.name",
@@ -59,13 +60,14 @@ fn init_meter_provider() -> SdkMeterProvider {
         .with_view(my_view_rename_and_unit)
         .with_view(my_view_drop_attributes)
         .with_view(my_view_change_aggregation)
-        .build()
+        .build();
+    set_meter_provider(provider);
 }
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
-    let meter_provider = init_meter_provider();
-    let meter = meter_provider.meter("mylibraryname");
+    init_meter_provider();
+    let meter = meter_provider().meter("mylibraryname");
 
     // Example 1 - Rename metric using View.
     // This instrument will be renamed to "my_histogram_renamed",
@@ -151,6 +153,6 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // Metrics are exported by default every 30 seconds when using stdout exporter,
     // however shutting down the MeterProvider here instantly flushes
     // the metrics, instead of waiting for the 30 sec interval.
-    meter_provider.shutdown()?;
+    shutdown_meter_provider();
     Ok(())
 }
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -1,4 +1,4 @@
-use opentelemetry::global::{meter_provider, set_meter_provider, shutdown_meter_provider};
+use opentelemetry::global;
 use opentelemetry::metrics::Unit;
 use opentelemetry::Key;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
@@ -61,13 +61,13 @@ fn init_meter_provider() {
         .with_view(my_view_drop_attributes)
         .with_view(my_view_change_aggregation)
         .build();
-    set_meter_provider(provider);
+    global::set_meter_provider(provider);
 }
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     init_meter_provider();
-    let meter = meter_provider().meter("mylibraryname");
+    let meter = global::meter_provider().meter("mylibraryname");
 
     // Example 1 - Rename metric using View.
     // This instrument will be renamed to "my_histogram_renamed",
@@ -153,6 +153,6 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // Metrics are exported by default every 30 seconds when using stdout exporter,
     // however shutting down the MeterProvider here instantly flushes
     // the metrics, instead of waiting for the 30 sec interval.
-    shutdown_meter_provider();
+    global::shutdown_meter_provider();
     Ok(())
 }
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -67,7 +67,7 @@ fn init_meter_provider() {
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     init_meter_provider();
-    let meter = global::meter_provider().meter("mylibraryname");
+    let meter = global::meter("mylibraryname");
 
     // Example 1 - Rename metric using View.
     // This instrument will be renamed to "my_histogram_renamed",
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -19,7 +19,7 @@ fn init_meter_provider() {
             "metrics-basic-example",
         )]))
         .build();
-    set_meter_provider(provider);
+    opentelemetry::global::set_meter_provider(provider);
 }
 
 #[tokio::main]
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -1,4 +1,4 @@
-use opentelemetry::global::{meter_provider, set_meter_provider, shutdown_meter_provider};
+use opentelemetry::global::{self, meter_provider};
 use opentelemetry::metrics::Unit;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
 use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -19,7 +19,7 @@ fn init_meter_provider() {
             "metrics-basic-example",
         )]))
         .build();
-    opentelemetry::global::set_meter_provider(provider);
+    global::set_meter_provider(provider);
 }
 
 #[tokio::main]
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -28,7 +28,7 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     init_meter_provider();
 
     // Create a meter from the above MeterProvider.
-    let meter = global::meter_provider().meter("mylibraryname");
+    let meter = global::meter("mylibraryname");
 
     // Create a Counter Instrument.
     let counter = meter.u64_counter("my_counter").init();
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -1,4 +1,4 @@
-use opentelemetry::global::{self, meter_provider};
+use opentelemetry::global;
 use opentelemetry::metrics::Unit;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
 use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
@@ -28,7 +28,7 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     init_meter_provider();
 
     // Create a meter from the above MeterProvider.
-    let meter = meter_provider().meter("mylibraryname");
+    let meter = global::meter_provider().meter("mylibraryname");
 
     // Create a Counter Instrument.
     let counter = meter.u64_counter("my_counter").init();
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -1,32 +1,34 @@
+use opentelemetry::global::{meter_provider, set_meter_provider, shutdown_meter_provider};
 use opentelemetry::metrics::Unit;
 use opentelemetry::{metrics::MeterProvider as _, KeyValue};
 use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::{runtime, Resource};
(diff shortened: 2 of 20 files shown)


TommyCpp (approved):

opentelemetry-stdout/examples/basic.rs
```diff
@@ -43,11 +44,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     );
     span.end();
 
-    let meter = meter_provider.meter("stdout-test");
+    let meter = global::meter("stdout-test");
```
cijothomas, line 47: 💯 love this! A lot of people asked about how to pass around the provider everywhere, but this neatly solves that problem!

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -113,6 +114,16 @@ impl SdkMeterProvider {
     }
 }
 
+impl Drop for SdkMeterProvider {
```
cijothomas, line 117: please add this to changelog too.

examples/metrics-basic/src/main.rs
```diff
@@ -1,32 +1,34 @@
...
-    SdkMeterProvider::builder()
+    let provider = SdkMeterProvider::builder()
         .with_reader(reader)
         .with_resource(Resource::new(vec![KeyValue::new(
             "service.name",
             "metrics-basic-example",
         )]))
-        .build()
+        .build();
+    set_meter_provider(provider);
```
cijothomas, line 22:
```suggestion
    opentelemetry::global::set_meter_provider(provider);
```
or 
```
global::set_meter_provider(provider);
```

I think these versions makes it easier to convey the intent?

examples/metrics-basic/src/main.rs
```diff
@@ -146,6 +148,6 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // Metrics are exported by default every 30 seconds when using stdout exporter,
     // however shutting down the MeterProvider here instantly flushes
     // the metrics, instead of waiting for the 30 sec interval.
-    meter_provider.shutdown()?;
+    shutdown_meter_provider();
```
cijothomas, line 151:
```suggestion
    global::shutdown_meter_provider();
```

cijothomas:
can you add a targeted test for this? Similar to https://github.com/open-telemetry/opentelemetry-rust/pull/1620

lalitb:
> can you add a targeted test for this? Similar to #1620

Good point. Have added the tests now.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
   dependency on crossbeam-channel.
   [1612](https://github.com/open-telemetry/opentelemetry-rust/pull/1612/files)
 - [#1422](https://github.com/open-telemetry/opentelemetry-rust/pull/1422)
-  Fix metrics aggregation bug when using Views to drop attributes.  
+  Fix metrics aggregation bug when using Views to drop attributes.
+- [#1623](https://github.com/open-telemetry/opentelemetry-rust/pull/1623) Add Drop implementation for SdkTracerProvider
```
cijothomas, line 10:
```suggestion
- [#1623](https://github.com/open-telemetry/opentelemetry-rust/pull/1623) Add Drop implementation for SdkTracerProvider, which shuts down
metricreaders, thereby allowing metrics still in memory to be flushed out.
```

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
   dependency on crossbeam-channel.
   [1612](https://github.com/open-telemetry/opentelemetry-rust/pull/1612/files)
 - [#1422](https://github.com/open-telemetry/opentelemetry-rust/pull/1422)
-  Fix metrics aggregation bug when using Views to drop attributes.  
+  Fix metrics aggregation bug when using Views to drop attributes.
+- [#1623](https://github.com/open-telemetry/opentelemetry-rust/pull/1623) Add Drop implementation for SdkTracerProvider, which shuts down
```
cijothomas, line 10:
```suggestion
- [#1623](https://github.com/open-telemetry/opentelemetry-rust/pull/1623) Add Drop implementation for SdkMeterProvider, which shuts down
```

examples/metrics-advanced/src/main.rs
```diff
@@ -59,13 +60,14 @@ fn init_meter_provider() -> SdkMeterProvider {
...
+        .build();
+    global::set_meter_provider(provider);
 }
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
-    let meter_provider = init_meter_provider();
-    let meter = meter_provider.meter("mylibraryname");
+    init_meter_provider();
+    let meter = global::meter_provider().meter("mylibraryname");
```
cijothomas, line 70:
```suggestion
    let meter = global::meter("mylibraryname");
```

examples/metrics-basic/src/main.rs
```diff
@@ -1,32 +1,34 @@
...
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // Initialize the MeterProvider with the stdout Exporter.
-    let meter_provider = init_meter_provider();
+    init_meter_provider();
 
     // Create a meter from the above MeterProvider.
-    let meter = meter_provider.meter("mylibraryname");
+    let meter = global::meter_provider().meter("mylibraryname");
```
cijothomas, line 31:
```suggestion
    let meter = global::meter("mylibraryname");
```

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -113,6 +114,16 @@ impl SdkMeterProvider {
...
+    fn drop(&mut self) {
+        if self.is_shutdown.load(Ordering::Relaxed) {
+            return;
+        }
+        if let Err(err) = self.shutdown() {
```
cijothomas, line 122: we need to set is_shutdown to true, right?

cijothomas: nvm

cijothomas: i think its best to move the check of is already shutdown to inside the shutdown() method itself, so drop need not worry about that.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -44,19 +44,23 @@ fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
...
             opentelemetry_otlp::new_exporter()
                 .http()
                 .with_export_config(export_config),
         )
-        .build()
+        .build();
+    match provider {
+        Ok(provider) => Ok(()),
+        Err(err) => Err(err),
+    }
```
TommyCpp, line 63:
```suggestion
      provider.map(|_| ())
```
Should do the trick

Does OTLP pipeline already set the provider as global meter provider?

opentelemetry-sdk/src/metrics/meter_provider.rs
TommyCpp: +1 on that. This doesn't seem to fit Drop's job of `clean up resource`

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
cijothomas: OTLP Pipeline does that, but I consider that as a bug, and should be fixed : https://github.com/open-telemetry/opentelemetry-rust/issues/1592

opentelemetry-sdk/src/metrics/meter_provider.rs
lalitb: Good point. Remove the chek from `drop`. shutdown() is already validating it.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -113,6 +114,13 @@ impl SdkMeterProvider {
     }
 }
 
+impl Drop for SdkMeterProvider {
+    fn drop(&mut self) {
+        if let Err(err) = self.shutdown() {
+            global::handle_error(err);
```
cijothomas, line 120: this might cause false alarms.. it is normal for users to call shutdown of their own, and then drop will call it again, causing the global error handler to print a message. It is unactionable for users, and they haven't anything wrong either.... Might need to revisit.

cijothomas (approved):
thanks! 
we need follow ups for the below:
1. Multi shutdown errors can cause unwanted confusion. (user calling shutdown() themselves, + drop)
2. after shutdown and drop, we need to reclaim full memory used for instruments. This might not be possible with current design.

hdost:
Isn't this a restore of sorts?

Should we consider a 0.22.2 patch?

cijothomas:
> Isn't this a restore of sorts?
> 
> Should we consider a 0.22.2 patch?

Good point. Lets discuss in the next community meeting and decide.
This is still not ideal, so we may want some follow ups before doing patch release.

cijothomas:
On rethinking, I think we can live without shutdown on the `opentelemetry` crate. Or we can keep it, but rename it to something like `reset_global_provider` or something!

The drop impl to invoke shutdown is still required in sdk.

The only drawback is when app owners don't have meter_provider - but that should only occur when using tracing-opentelemetry or others, and can be fixed in those places. The OTLP Exporter in the repo also exposes pipeline builders, which also would require tweaks to account for this.

TommyCpp:
> I think we can live without shutdown on the opentelemetry crate. Or we can keep it, but rename it to something like reset_global_provider or something!

I actually prefer `shutdown` to keep the concept in `global` module the same as the `MeterProvider` itself. But I agree app owner should keep control of the meter provider. 

> The only drawback is when app owners don't have meter_provider - but that should only occur when using tracing-opentelemetry or others, and can be fixed in those places

Hmm the app owner should always has control of providers even in `tracing-opentelemetry`. It looks like we are taking a `MeterProvider` in tracing-opentelemetry for some reasons. https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/metrics.rs#L332

lalitb:
1. The `global::meter_provider_shutdown()` method was introduced to align with the existing shutdown methods for other signals - global::tracer_provider_shutdown() and global::logger_provider_shutdown. However, `MeterProvider::Drop()` was indeed required to ensure the cleanup logic is always executed upon application termination.

2. Ideally, the OpenTelemetry crate should not include any `global::<signal>_provider_shutdown()/reset()` methods. There should be no scenario where the instrumentation library is required to call these methods.

3. The application should always have the handle of the `opentelemetry_sdk::<signal>Provider`. This will ensure that they can always invoke shutdown and forceflush on the handle. For this to work:
     -   There may be differing opinions on this, but it's worth discussing the possibility of eliminating the exporter-specific pipeline conveniences like `OtlpTracePipeline`, `OtlpMetricPipeline`, `OtlpLogPipeline`, and `ZipkinPipeline`. These pipelines internally create a provider and set it as global, and then we don't have the handle to it.

    - The tokio-tracing layer or subscriber should either maintain a reference to the <signal>Provider, similar to the approach used for [logs](https://github.com/open-telemetry/opentelemetry-rust/blob/4aa48271917c40921265ce4750ebed191d363e0d/opentelemetry-appender-tracing/src/layer.rs#L139), or take ownership of the Tracer, Meter, or Logger, as implemented for [traces](https://github.com/tokio-rs/tracing-opentelemetry/blob/5e3354ec24debcfbf856bfd1eb7022459dca1e6a/src/layer.rs#L535). The adaptation of this approach for metrics within the `tracing-opentelemetry` is missing and needs to be fixed. This issue has been previously brought up in discussions [here](https://github.com/open-telemetry/opentelemetry-rust/discussions/1605#discussioncomment-8699931) and [here](https://github.com/tokio-rs/tracing-opentelemetry/discussions/114). It appears necessary for someone, possibly us, to formally create an issue in the tracing-opentelemetry repository to address this.

TommyCpp:
Just to clearify: are we proposing removing support of the global `Tracer/Meter/LoggerProvider` or we are only thinking about removing `global::<single>_shutdown` method

cijothomas:
> There may be differing opinions on this, but it's worth discussing the possibility of eliminating the exporter-specific pipeline conveniences like OtlpTracePipeline, OtlpMetricPipeline, OtlpLogPipeline, and ZipkinPipeline. These pipelines internally create a provider and set it as global, and then we don't have the handle to it.

yes agree. its covered in https://github.com/open-telemetry/opentelemetry-rust/issues/1592

cijothomas:
> Ideally, the OpenTelemetry crate should not include any global::<signal>_provider_shutdown()/reset() methods. There should be no scenario where the instrumented-library is required to call these methods.
> 
> The application should always have the handle of the opentelemetry_sdk::<signal>Provider

I am in agreement. We need to revert adding global_shutdown part of this PR, and follow through and cleanup other signals too.

lalitb:
> Just to clearify: are we proposing removing support of the global Tracer/Meter/LoggerProvider or we are only thinking about removing global::<single>_shutdown method.


Suggestion is to remove the `global::<signal>_shutdown` method without introducing any replacements such as `global::<signal>_reset`. Since these methods are intended exclusively for use by the application owner, they do not have a place in the API. While instrumented libraries can achieve a similar shutdown or reset effect by executing `set_meter_provider(NoopMeterProvider::new());`, the presence of a dedicated method might misleadingly imply that it is designed for use by libraries.

TommyCpp:
I am OK if we want to remove `global::<signal>_shutdown` method from API. 

However, for end users, if they set the global signal prvoider using `set_<signal>_provider` method, they will lose the handle of the provider. For them to shutdown the provider, they need to call `let _ = set_<singal>_provider(NoopProvider::new())` method, which seem to be unintuitive

lalitb:
> However, for end users, if they set the global signal prvoider using set_<signal>_provider method, they will lose the handle of the provider. For them to shutdown the provider, they need to call let _ = set_<singal>_provider(NoopProvider::new()) method to regain the handle of the provider, which seem to be unintuitive

Yes, that's a good point. This is an issue as of now too as there is no way to call force_flush after the call to set_\<signal\>_provider.

PR #1624, opened 2024-03-15
Remove os and process resource detectors

rogercoll:
## Changes

- Os and Process resource detectors have been moved to the [opentelemetry-contrib-rust crate](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-resource-detectors) due to their experimental state in the semantic convention.

- Use public crate constants for stable resources: telemetry and service name


rogercoll:
This PR should close https://github.com/open-telemetry/opentelemetry-rust/pull/1610

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -16,18 +16,17 @@
 //! SDK.
 //!
 //! - [`EnvResourceDetector`] - detect resource from environmental variables.
-//! - [`OsResourceDetector`] - detect OS from runtime.
-//! - [`ProcessResourceDetector`] - detect process information.
 //! - [`TelemetryResourceDetector`] - detect telemetry SDK's information.
+//!
+//! The OS and Process resource detectors are now packaged separately in the `opentelemetry-resource-detector` [contrib crate](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-resource-detectors), due to their experimental nature in semantic conventions.
```
cijothomas, line 21:
```suggestion
//! The OS and Process resource detectors are now packaged separately in the `opentelemetry-resource-detector` [crate](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-resource-detectors).
```

opentelemetry-sdk/src/resource/env.rs
```diff
@@ -76,15 +76,15 @@ pub struct SdkProvidedResourceDetector;
...
+            super::SERVICE_NAME,
             env::var(OTEL_SERVICE_NAME)
                 .ok()
                 .filter(|s| !s.is_empty())
                 .map(Value::from)
                 .or_else(|| {
                     EnvResourceDetector::new()
                         .detect(Duration::from_secs(0))
-                        .get(Key::new("service.name"))
+                        .get(Key::new(super::SERVICE_NAME))
```
cijothomas, line 87: this is the name of the env variable, right? Not semantic convention?

cijothomas:
We also need a changelog, and mention the alternate option. (That also means we need to do a release of the new crate)

hdost (changes requested):
Thank you for the patch! please add an entry in the Changelog, with a Breaking and provide a link to the new crate.

opentelemetry-sdk/src/resource/attributes.rs
```diff
@@ -0,0 +1,32 @@
...
+pub(crate) const SERVICE_NAME: &str = "service.name";
+
+/// The language of the telemetry SDK.
+pub(crate) const TELEMETRY_SDK_LANGUAGE: &str = "telemetry.sdk.language";
+
+/// The name of the telemetry SDK as defined above.
+///
+/// The OpenTelemetry SDK MUST set the `telemetry.sdk.name` attribute to `opentelemetry`.
+/// If another SDK, like a fork or a vendor-provided implementation, is used, this SDK MUST set the
+/// `telemetry.sdk.name` attribute to the fully-qualified class or module name of this SDK&#39;s main entry point
```
TommyCpp, line 17:
```suggestion
/// `telemetry.sdk.name` attribute to the fully-qualified class or module name of this SDK's main entry point
```

opentelemetry-sdk/src/resource/os.rs
```diff
@@ -1,46 +0,0 @@
...
-
-/// Detect runtime operating system information.
-///
-/// This detector uses Rust's [`OS constant`] to detect the operating system type and
-/// maps the result to the supported value defined in [`OpenTelemetry spec`].
-///
-/// [`OS constant`]: https://doc.rust-lang.org/std/env/consts/constant.OS.html
-/// [`OpenTelemetry spec`]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/os.md
-#[derive(Debug)]
-pub struct OsResourceDetector;
```
TommyCpp, line 18: Actually would it make sense to mark this as deprecate for the next version and removed later?

opentelemetry-sdk/src/resource/env.rs
rogercoll: Not the env variable, but the service name key value within the `OTEL_RESOURCE_ATTRIBUTES` key value.

This is the fallback value retrieved from the EnvResourceDetector, in which the service name can be defined directy on the key values list: `OTEL_RESOURCE_ATTRIBUTES="service.name=service1,another=test"`

In this case we try to get the "service.name" (semantic convention) from the parsed key values.

rogercoll:
Change log entry added in https://github.com/open-telemetry/opentelemetry-rust/pull/1624/commits/28c90b3e0fefa789b4a2119f41a0b94f64db008f

opentelemetry-sdk/src/resource/os.rs
rogercoll: Sounds good to me, should I open a new PR and keep this one until next release?

cijothomas: I think it is okay to move faster here - i.e remove right now. We only need to ensure that the alternate (the new crate) is published before the next release of sdk crate.

opentelemetry-sdk/src/resource/env.rs
```diff
@@ -76,15 +76,15 @@ pub struct SdkProvidedResourceDetector;
 impl ResourceDetector for SdkProvidedResourceDetector {
```
cijothomas, line 76: not something added in this PR, this `SdkProvidedResourceDetector` does not sound very clear. Maybe we should remove that pub struct, and provide a method like `add_default_resource` instead.
We also need to have a test for each providers(log/trace/meter) to ensure that
1. If users specifies *no* explicit resource (neither in code, nor in env vars), then the default is added
2. If users specifies *no* explicit resource in code, but env vars, then we get the env vars one.
3. Users can have resource without the default.
4. Users can have defaults + their own.
5. Validate precedence in all scenarios.
+ anything else missing.

This is a good candidate for follow up PR @rogercoll if you are interested in continuing this work!

cijothomas (approved):

opentelemetry-sdk/src/resource/env.rs
TommyCpp: We already have it coverage for `TracerProvider` [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/provider.rs#L277)

TommyCpp: Hmm I don't think we need explicit need a method to add default resource. It sounds like `Default` trait is a better fit and we are using it.

cijothomas: > We already have unit tests mentioned above for `TracerProvider` [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/provider.rs#L277)

Thanks for pointing. We need it for all signals. Also https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/provider.rs#L291 is only validating if service name is populated, it should do telemetry resource as well, right?

TommyCpp: Ah good point. Yeah we probably need to expand the coverage to other signal and other resources

rogercoll: Make sense to me, I will follow up on this topic for the other signals.

hdost (approved):

cijothomas (approved):

PR #1628, opened 2024-03-19
Treat log's key-values as attributes in log bridge

KodrAus:
## Changes

This PR supports `log`'s newly stabilized structured logging API in the `opentelemetry-appender-log` crate.

As of the `0.4.21` release, `log` has stable support for structured logging. Its model is to decouple the serialization framework choice from producers and consumers, so I've opted to use `serde` here to convert `log`'s structured values into `AnyValue`s. The infrastructure isn't specific to `log`, so if it either already exists somewhere in the SDK that I can use, or would be worth moving into the SDK under some `serde` feature I'd be happy to do that too.

One open question  I've got is whether I need to deal with deduplication when constructing the attributes list, or whether that's dealt with elsewhere.


cijothomas:
> One open question I've got is whether I need to deal with deduplication when constructing the attributes list, or whether that's dealt with elsewhere.

As of today, its not done anywhere in the API, SDK, Exporter. There is a good chance of a feature flag/option to make it happen. (unsure where,, could be just in otlp exporter only)

cijothomas:
> I've opted to use serde here to convert log's structured values into AnyValues.

mm.. This may require some discussion.. The log-bridge only depends on `opentelemetry` (which is otel's API crate) and it does not (and should not) deal with concerns such as serialization etc. That is left for the exporters (which are based off of the `opentelemetry-sdk` crate).

KodrAus:
The serialization via `serde` is really just an internal conversion from `log`'s value type into `AnyValue`s. `log` doesn't define its own complete data-model, instead deferring to a serialization framework like `serde` to provide one for it. From there, an exporter will receive `AnyValue`s which it can choose to serialize however it wants.

If we didn't want to force a dependency on `serde` here (which is perfectly reasonable) we can still convert primitive types coming from `log` into `AnyValue`s via its own simplified visitor API, but won't be able to convert complex types like arrays or objects without needing to format them as strings. I could make the `serde` dependency optional for anyone that wants to opt-in to complete support.

lalitb:
> If we didn't want to force a dependency on serde here (which is perfectly reasonable) we can still convert primitive types coming from log into AnyValues via its own simplified visitor API, but won't be able to convert complex types like arrays or objects without needing to format them as strings. I could make the serde dependency optional for anyone that wants to opt-in to complete support.

+ 1  on that. As of now we only support primitive types with [tokio-tracing appender](https://github.com/open-telemetry/opentelemetry-rust/blob/a7a47a745dfe9ded79ffb64722fc53847638d47a/opentelemetry-appender-tracing/src/layer.rs#L80C6-L80C27), so should be good to have similar support with log-appender, and then using serde optionally for complete support.

KodrAus:
Ok, I can make that change :+1: Any preference on what that feature should be called? `log_attributes_full`? The `tracing` appender may end up with an equivalent feature once they finish integrating `valuable`, unless they make that dependency required in `tracing`.

TommyCpp:
In the other repo we have the feature `with_serde` to annotate proto generated files with serde attributes. Maybe we can borrow it here

KodrAus:
Ok, I've made the `serde` dependency optional, gated by a `with_serde` feature. [This test case](https://github.com/open-telemetry/opentelemetry-rust/blob/8c91cb3f27cb466a28c767aedee58fbe9e0fa230/opentelemetry-appender-log/src/lib.rs#L790-L824) gives you a reasonable idea of how values are converted with and without the feature enabled.

KodrAus:
I can improve the test coverage here once we’re happy with the overall shape of things. Is there anywhere I’d need to plug this new feature into CI?

opentelemetry-appender-log/examples/logs-basic-in-memory.rs
```diff
@@ -25,7 +25,10 @@ async fn main() {
     log::set_max_level(Level::Info.to_level_filter());
 
     // Emit logs using macros from the log crate.
-    error!("hello from {}. My price is {}", "apple", 2.99);
+    let user = "apple";
```
lalitb, line 28: nit s/user/fruit

opentelemetry-appender-log/examples/logs-basic-in-memory.rs
```diff
@@ -25,7 +25,10 @@ async fn main() {
...
+    let price = 2.99;
+
+    error!(user, price; "hello from {user}. My price is {price}");
```
lalitb, line 31: Do you think it is possible to add a serde specific example under feature flag ?

opentelemetry-appender-log/src/lib.rs
```diff
@@ -70,13 +74,569 @@ const fn severity_of_level(level: Level) -> Severity {
...
+fn log_attributes(kvs: impl log::kv::Source) -> Vec<(Key, AnyValue)> {
+    struct AttributeVisitor(Vec<(Key, AnyValue)>);
+
+    impl<'kvs> log::kv::VisitSource<'kvs> for AttributeVisitor {
+        fn visit_pair(
+            &mut self,
+            key: log::kv::Key<'kvs>,
+            value: log::kv::Value<'kvs>,
+        ) -> Result<(), log::kv::Error> {
+            let key = Key::from(String::from(key.as_str()));
```
lalitb, line 86: Does log's Key API provides way to check if the key is a borrowed reference, or an owned string. As otel::key internally stores string reference, so we can avoid the allocation if possible ?

opentelemetry-appender-log/src/lib.rs
```diff
@@ -70,13 +74,569 @@ const fn severity_of_level(level: Level) -> Severity {
...
+            }
+
+            fn visit_bool(&mut self, value: bool) -> Result<(), log::kv::Error> {
+                self.0 = Some(AnyValue::Boolean(value));
+
+                Ok(())
+            }
+
+            fn visit_str(&mut self, value: &str) -> Result<(), log::kv::Error> {
+                self.0 = Some(AnyValue::String(StringValue::from(value.to_owned())));
```
lalitb, line 124: Same as key, if log's value provides the way to check if the string is borrowed or owned ?

lalitb (commented):
changes look good. Some minor comments

opentelemetry-appender-log/Cargo.toml
```diff
@@ -12,12 +12,16 @@ edition = "2021"
 
 [dependencies]
 opentelemetry = { version = "0.22", path = "../opentelemetry", features = ["logs"]}
-log = { workspace = true, features = ["kv_unstable", "std"]}
+log = { workspace = true, features = ["kv", "std"]}
+serde = { workspace = true, optional = true, features = ["std"] }
 
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled"]
+with_serde = ["log/kv_serde", "serde"]
```
lalitb, line 20: At other places, we use `with-serde` as feature name. Is `with_serde` naming intentional?

hdost:
Some relevant comments here https://github.com/open-telemetry/opentelemetry-specification/pull/3938

opentelemetry-appender-log/src/lib.rs
KodrAus: It doesn’t currently, but we’ve left the door open for it.

opentelemetry-appender-log/src/lib.rs
KodrAus: This is another case where it doesn’t, but we’ve left room for it.

opentelemetry-appender-log/Cargo.toml
KodrAus: I just tried to follow the convention I saw in naming of other features. I can change this to kebab-case 👍

KodrAus:
Thanks for the review @lalitb. I’ll make those changes, fill out the test coverage, and leave some docs on how to enable key-value support, and how various kinds of values are represented.

opentelemetry-appender-log/src/lib.rs
lalitb: Looks like it does [support](https://github.com/rust-lang/log/blob/7bdb8aac88c848df2668149d67a6acc971e1a8e0/src/kv/value.rs#L992), but should be fine to support it as separate PR.

KodrAus: Ah yes, that method lets us attempt to pull a string for a specific lifetime `'v` from a `Value<'v>`, but for `StringValue` we'd need that lifetime to be `'static`. It's _almost_ what we'd need here.

KodrAus:
This should be ready for another review now. I've added some crate-level docs to `opentelemetry-appender-log`. I tried to follow any conventions I saw in other docs, but please point out anywhere I should be doing things differently.

lalitb (approved):

TommyCpp (approved):
LGTM. Thanks for the change!

KodrAus:
The build failure here looks unrelated; is kicking off a fresh build likely to resolve it?

opentelemetry-appender-log/Cargo.toml
```diff
@@ -12,12 +12,17 @@ edition = "2021"
...
+serde = { workspace = true, optional = true, features = ["std"] }
 
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled"]
+with-serde = ["log/kv_serde", "serde"]
 default = ["logs_level_enabled"]
 
 [dev-dependencies]
 opentelemetry_sdk = { path = "../opentelemetry-sdk", features = [ "testing", "logs_level_enabled" ] }
+opentelemetry-stdout = { path = "../opentelemetry-stdout", features = ["logs"]}
```
cijothomas, line 25: were you planning to replace the current in-memory exporter in the example with the stdout? I think that'd be a nice addition!

KodrAus: Ah yes, I thought I had done this but must have reverted the change to the actual file somewhere along the line. I'll add it back in.

opentelemetry-appender-log/src/lib.rs
```diff
@@ -1,5 +1,104 @@
...
+//! | Tuple           | [`AnyValue::ListAny`] | Requires the `with-serde` feature, otherwise it will be stringified using [`std::fmt::Debug`]                           |
+//! | Tuple struct    | [`AnyValue::ListAny`] | Requires the `with-serde` feature, otherwise it will be stringified using [`std::fmt::Debug`]                           |
+//! | Tuple variant   | [`AnyValue::Map`]     | An internally-tagged map. Requires the `with-serde` feature, otherwise it will be stringified using [`std::fmt::Debug`] |
+//! | Map             | [`AnyValue::Map`]     | Requires the `with-serde` feature, otherwise it will be stringified using [`std::fmt::Debug`]                           |
+//! | Struct          | [`AnyValue::Map`]     | Requires the `with-serde` feature, otherwise it will be stringified using [`std::fmt::Debug`]                           |
+//! | Struct variant  | [`AnyValue::Map`]     | An internally-tagged map. Requires the `with-serde` feature, otherwise it will be stringified using [`std::fmt::Debug`] |
+//!
+//! # Feature Flags
+//!
+//! This library provides the following Cargo features:
```
cijothomas, line 90: 💯 love this doc!

cijothomas (approved):
Thanks! Could you add a changelog.md file too?

PR #1640, opened 2024-03-25
fix: send propagator error to global error handler

TommyCpp:
One issue with our propagator setup is when there is an error we usually just return default context and throw the error away. 

## Changes

- Add a new error type for `propagator` 
- Export errors from extracting and injecting to global error handler rather then ignore them
- No change made to the result of extract and injec


TommyCpp:
Need to add change log

cijothomas (approved):
LGTM. Will you be adding similar to the trace_context propagator as well?

Issue #1625, opened 2024-03-15
[Bug]: TracerProvider can be dropped unknowingly


SZenglein:
### What happened?

Consider this example code:   
(I am using tracing_opentelemetry, but it should be reproducible without it)  
```rust
let telemetry =  {
    // Create a new OpenTelemetry trace pipeline that prints to stdout
    let provider = TracerProvider::builder()
        .with_simple_exporter(
            opentelemetry_otlp::new_exporter()
                .tonic()
                .with_endpoint("http:localhost:4317")
                .with_timeout(Duration::from_secs(3))
                .build_span_exporter()?,
        )
        .with_config(opentelemetry_sdk::trace::Config::default().with_resource(
            Resource::new(vec![KeyValue::new(
                opentelemetry_semantic_conventions::resource::SERVICE_NAME,
                "emloc-mgr-api",
            )]),
        ))
        .build();
    
    let tracer = provider.tracer("emloc-mgr-api");
    
    // Create and return a tracing layer with the configured tracer
    tracing_opentelemetry::layer().with_tracer(tracer)
};
```

Since the provider is not used afterwards, it is very easy to accidentally drop it when executing inside a function or conditional block.  
There is no compiler warning or any documentation note regarding this behaviour or that it must be kept alive manually. This results in very hard to debug problem, because code that compiles fine without a warning can stop working just by moving a value into a scope.

I suspect this is because `TracerProvider` creates the `Tracer` by downgrading the `Arc` to a `Weak`. 

IMHO this should at the very least be documented on the `TracerProvider` struct, maybe a warning can be added to the Drop Implementation, too.

### API Version

0.15

### SDK Version

0.22.1

TommyCpp:
Yeah it's a known behavior. Users need to manually hold reference to provider or pass the provider to global tracer provider. 

Agree on we should update documentation and possible looking into compiler warnings

lalitb:
Will it make sense for TraceProvider to have similar behavior as LoggerProvider (#1455) - Let `Tracer`'s own `TracerProvider` (with shared referenece to `TracerProviderInner`), which would ensure that it won't be dropped till all instances of Tracer's are dropped.  Though the main motive for this change was performance, but I guess this is a good side-effect to have.

TommyCpp:
Ah you are right. Let me give it a try

TommyCpp:
[...]

PR #1643, opened 2024-03-26
feat(logs): make logger shutdown &self

TommyCpp:
This PR aim to address the issue brought up in https://github.com/open-telemetry/opentelemetry-rust/issues/1625#issuecomment-2002680753

In summary we need to:
- Ensure logger provider shut down will block creating new (functional) logger.
- Ensure no log can be emitted from logger once the logger provider shutdown.

We discussed the solution of keeping `Weak` reference from `LoggerProvider` to `Logger`s but as I implemented this solution it seems to complicated. I revisited why we need mutable reference during `LoggerProvider` and found that `LogProcessor` shutdown doesn't have to take a mutable reference. 

## Changes

- make `LogProcessor` shutdown taking `&self` instead of `&mut self`. 
   - This decouple the `shutdown` from `drop`. If one `LogProcessor` is shared across multiple thread, any thread can call `shutdown` to stop the `LogProcessor` from emitting more logs. But this doesn't mean the `LogProcessor` will drop. 
   - However, `drop` will call `shutdown`
   - This implementation stops the log emitting after `shutdown` in `BatchLogProcessor`. Emitting new logs after shutdown will result in a `LogError` saying the receiver on the worker task has already closed. 
   - for `SimpleLogProcessor` we need a field to mark if the processor has been shutdown we also need to check it everytime before emitting the logs. 

- Add a field in `LoggerProvider` to mark if the logger provider has shutdown. If it has, return a noop logger

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -13,6 +13,7 @@
   `ProcessResourceDetector` resource detectors, use the
   [`opentelemetry-resource-detector`](https://crates.io/crates/opentelemetry-resource-detectors) instead.
 - Baggage propagation error will be reported to global error handler [#1640](https://github.com/open-telemetry/opentelemetry-rust/pull/1640)
+- Make `LoggerProvider`'s `shutdown` method taking immutable reference [#1643](https://github.com/open-telemetry/opentelemetry-rust/pull/1643)
 
 ## v0.22.1
 
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -13,7 +13,7 @@
   `ProcessResourceDetector` resource detectors, use the
   [`opentelemetry-resource-detector`](https://crates.io/crates/opentelemetry-resource-detectors) instead.
 - Baggage propagation error will be reported to global error handler [#1640](https://github.com/open-telemetry/opentelemetry-rust/pull/1640)
-- Make `LoggerProvider`'s `shutdown` method taking immutable reference [#1643](https://github.com/open-telemetry/opentelemetry-rust/pull/1643)
+- Make `shutdown` method in `LoggerProvider` and `LogProcessor` taking immutable reference [#1643](https://github.com/open-telemetry/opentelemetry-rust/pull/1643)
 
 ## v0.22.1
 
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -13,7 +13,11 @@
   `ProcessResourceDetector` resource detectors, use the
   [`opentelemetry-resource-detector`](https://crates.io/crates/opentelemetry-resource-detectors) instead.
 - Baggage propagation error will be reported to global error handler [#1640](https://github.com/open-telemetry/opentelemetry-rust/pull/1640)
-- Make `shutdown` method in `LoggerProvider` and `LogProcessor` taking immutable reference [#1643](https://github.com/open-telemetry/opentelemetry-rust/pull/1643)
+- Improves `shutdown` behavior of `LoggerProvider` and `LogProcessor` [#1643](https://github.com/open-telemetry/opentelemetry-rust/pull/1643).
+  - `shutdown` can be called by any clone of the `LoggerProvider` without the need of waiting on all `Logger` drops. Thus, `try_shutdown` has been removed. 
+  - `shutdown` methods in `LoggerProvider` and `LogProcessor` now takes a immutable reference
+  - After `shutdown`, `LoggerProvider` will return noop `Logger`
+  - After `shutdown`, `LogProcessor` will not process any new logs
 
 ## v0.22.1
 
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -269,14 +269,63 @@ mod tests {
 
     use super::*;
     use opentelemetry::global::{logger, set_logger_provider, shutdown_logger_provider};
-    use opentelemetry::logs::Logger;
-    use opentelemetry::{Key, KeyValue, Value};
     use opentelemetry::logs::{Logger, LoggerProvider as _};
+    use opentelemetry::{Key, KeyValue, Value};
     use std::fmt::{Debug, Formatter};
     use std::sync::atomic::AtomicU64;
     use std::sync::{Arc, Mutex};
     use std::thread;
 
+    struct ShutdownTestLogProcessor {
+        is_shutdown: Arc<Mutex<bool>>,
+        counter: Arc<AtomicU64>,
+    }
+
+    impl Debug for ShutdownTestLogProcessor {
+        fn fmt(&self, _f: &mut Formatter<'_>) -> std::fmt::Result {
+            todo!()
+        }
+    }
+
+    impl ShutdownTestLogProcessor {
+        pub(crate) fn new(counter: Arc<AtomicU64>) -> Self {
+            ShutdownTestLogProcessor {
+                is_shutdown: Arc::new(Mutex::new(false)),
+                counter,
+            }
+        }
+    }
+
+    impl LogProcessor for ShutdownTestLogProcessor {
+        fn emit(&self, _data: LogData) {
+            self.is_shutdown
+                .lock()
+                .map(|is_shutdown| {
+                    if !*is_shutdown {
+                        self.counter
+                            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+                    }
+                })
+                .expect("lock poisoned");
+        }
+
+        fn force_flush(&self) -> LogResult<()> {
+            Ok(())
+        }
+
+        fn shutdown(&self) -> LogResult<()> {
+            self.is_shutdown
+                .lock()
+                .map(|mut is_shutdown| *is_shutdown = true)
+                .expect("lock poisoned");
+            Ok(())
+        }
+
+        #[cfg(feature = "logs_level_enabled")]
+        fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
+            true
+        }
+    }
     #[test]
     fn test_logger_provider_default_resource() {
         let assert_resource = |provider: &super::LoggerProvider,
@@ -403,57 +452,6 @@ mod tests {
         assert_eq!(no_service_name.config().resource.len(), 0);
     }
 
-    struct ShutdownTestLogProcessor {
-        is_shutdown: Arc<Mutex<bool>>,
-        counter: Arc<AtomicU64>,
-    }
-
-    impl Debug for ShutdownTestLogProcessor {
-        fn fmt(&self, _f: &mut Formatter<'_>) -> std::fmt::Result {
-            todo!()
-        }
-    }
-
-    impl ShutdownTestLogProcessor {
-        pub(crate) fn new(counter: Arc<AtomicU64>) -> Self {
-            ShutdownTestLogProcessor {
-                is_shutdown: Arc::new(Mutex::new(false)),
-                counter,
-            }
-        }
-    }
-
-    impl LogProcessor for ShutdownTestLogProcessor {
-        fn emit(&self, _data: LogData) {
-            self.is_shutdown
-                .lock()
-                .map(|is_shutdown| {
-                    if !*is_shutdown {
-                        self.counter
-                            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
-                    }
-                })
-                .expect("lock poisoned");
-        }
-
-        fn force_flush(&self) -> LogResult<()> {
-            Ok(())
-        }
-
-        fn shutdown(&self) -> LogResult<()> {
-            self.is_shutdown
-                .lock()
-                .map(|mut is_shutdown| *is_shutdown = true)
-                .expect("lock poisoned");
-            Ok(())
-        }
-
-        #[cfg(feature = "logs_level_enabled")]
-        fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
-            true
-        }
-    }
-
     #[test]
     fn shutdown_test() {
         let counter = Arc::new(AtomicU64::new(0));
@@ -524,36 +522,19 @@ mod tests {
             thread::sleep(std::time::Duration::from_millis(10));
         }
 
-        // Intentionally *not* calling shutdown/flush on the provider, but
-        // instead relying on shutdown_logger_provider which causes the global
-        // provider to be dropped, leading to the sdk logger provider's drop to
-        // be called, which is expected to call shutdown on processors.
+        // shutdown logger provider explicitly call shutdown on the logger
         shutdown_logger_provider();
 
         // Assert
 
-        // shutdown_logger_provider is necessary but not sufficient, as loggers
-        // hold on to the the provider (via inner provider clones).
-        assert!(!*shutdown_called.lock().unwrap());
+        // shutdown_logger_provider should be shutdown regardless if the logger is dropped.
+        assert!(*shutdown_called.lock().unwrap());
 
         // flush is never called by the sdk.
         assert!(!*flush_called.lock().unwrap());
 
-        // Drop one of the logger. Not enough!
-        drop(logger1);
-        assert!(!*shutdown_called.lock().unwrap());
-
-        // drop logger2, which is the only remaining logger in this thread.
-        // Still not enough!
-        drop(logger2);
-        assert!(!*shutdown_called.lock().unwrap());
-
-        // now signal the spawned thread to end, which causes it to drop its
-        // logger. Since that is the last logger, the provider (inner provider)
-        // is finally dropped, triggering shutdown
         *signal_to_end.lock().unwrap() = true;
         handle.join().unwrap();
-        assert!(*shutdown_called.lock().unwrap());
 
         // flush is never called by the sdk.
         assert!(!*flush_called.lock().unwrap());
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -80,6 +80,20 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
         }
         Logger::new(library, self.clone())
     }
+
+    /// Shuts down this `LoggerProvider`
+    fn shutdown(&self) -> Vec<LogResult<()>> {
+        // mark itself as already shutdown
+        self.is_shutdown
+            .store(true, std::sync::atomic::Ordering::Relaxed);
+        // propagate the shutdown signal to processors
+        // it's up to the processor to properly block new logs after shutdown
+        self.inner
+            .processors
+            .iter()
+            .map(|processor| processor.shutdown())
+            .collect()
+    }
 }
 
 impl LoggerProvider {
@@ -104,27 +118,6 @@ impl LoggerProvider {
             .iter()
             .map(|processor| processor.force_flush())
             .collect()
-    }
-
-    /// Shuts down this `LoggerProvider`, panicking on failure.
-    pub fn shutdown(&mut self) -> Vec<LogResult<()>> {
-        // mark itself as already shutdown
-        self.is_shutdown
-            .store(true, std::sync::atomic::Ordering::Relaxed);
-        // propagate the shutdown signal to processors
-        // it's up to the processor to properly block new logs after shutdown
-        self.inner
-            .processors
-            .iter()
-            .map(|processor| processor.shutdown())
-            .collect()
-    }
-
-    /// Attempts to shutdown this `LoggerProvider`, succeeding only when
-    /// all cloned `LoggerProvider` values have been dropped.
-    // todo: remove this
-    pub fn try_shutdown(&mut self) -> Option<Vec<LogResult<()>>> {
-        Some(self.shutdown())
     }
 }
 
@@ -464,7 +457,7 @@ mod tests {
     #[test]
     fn shutdown_test() {
         let counter = Arc::new(AtomicU64::new(0));
-        let mut logger_provider = LoggerProvider::builder()
+        let logger_provider = LoggerProvider::builder()
             .with_log_processor(ShutdownTestLogProcessor::new(counter.clone()))
             .build();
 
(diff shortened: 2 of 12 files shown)


lalitb:
This seems like a much more streamlined approach. In particular, the LoggerProvider (and so the LogProcessors) can be shut down without needing to wait for all loggers to be dropped. 

> [TODO] for SimpleLogProcessor we need a field to mark if the processor has been shutdown we also need to check it everytime before emitting the logs.

I believe it should be acceptable if the check is performed atomically. This isn't related to the current PR, but just to reiterate so I don't forget, this check would also need to be added in ReentrantLogProcessor for [ETW ](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/eaa0f5f49cd9ccc19933157341dd01da0e0aef2b/opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs#L19) and [user_events ](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/eaa0f5f49cd9ccc19933157341dd01da0e0aef2b/opentelemetry-user-events-logs/src/logs/reentrant_logprocessor.rs#L21) exporter.

lalitb:
@TommyCpp Do you plan to make it ready for review? The changes look good to me.

TommyCpp:
Sorry busy week. Will get it done this weekend

TommyCpp:
> this check would also need to be added in ReentrantLogProcessor for [ETW ](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/eaa0f5f49cd9ccc19933157341dd01da0e0aef2b/opentelemetry-etw-logs/src/logs/reentrant_logprocessor.rs#L19) and [user_events ](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/eaa0f5f49cd9ccc19933157341dd01da0e0aef2b/opentelemetry-user-events-logs/src/logs/reentrant_logprocessor.rs#L21) exporter.

I think we should also make it clear that `LogProcessor` needs to make sure no new logs get processed after shutdown

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -82,20 +101,23 @@ impl LoggerProvider {
...
+        self.inner
+            .processors
+            .iter()
+            .map(|processor| processor.shutdown())
+            .collect()
     }
 
     /// Attempts to shutdown this `LoggerProvider`, succeeding only when
     /// all cloned `LoggerProvider` values have been dropped.
+    // todo: remove this
```
lalitb, line 118: nit - If we are keeping `try_shutdown()` for backward compatibility, good to update the existing comments, as now it doesn't just attempt, but really shutdown the LoggerProvider.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -59,18 +62,25 @@ pub trait LogProcessor: Send + Sync + Debug {
...
         }
     }
 }
 
 impl LogProcessor for SimpleLogProcessor {
     fn emit(&self, data: LogData) {
+        // noop after shutdown
+        if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
+            return;
+        }
```
lalitb, line 82: Do we need this check, or else rely on the exporter to check this before actual export. As  I don't think batchprocessor is doing this check. Just want to ensure both the processors are consistent here :)
Also, good to compare the [benchmark](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/log.rs) if this validation has any effect on the performance.

TommyCpp: > As I don't think batchprocessor is doing this check. 

After batch processor shuts down the channel used to send logs to background task will error out because the recevier has been dropped. So we get the noop(well, in this case it will emit an error) free

TommyCpp:
Simple log processor took some hit as expected but it's acceptable IMHO
```
simple-log/no-context   time:   [109.01 ns 109.09 ns 109.20 ns]
                        change: [-2.4832% -2.3350% -2.1759%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  1 (1.00%) high severe
simple-log/with-context time:   [109.56 ns 109.62 ns 109.69 ns]
                        change: [+2.0309% +2.1353% +2.2391%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild

simple-log-with-int/no-context
                        time:   [146.61 ns 146.71 ns 146.81 ns]
                        change: [-3.8572% -3.6060% -3.4182%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
simple-log-with-int/with-context
                        time:   [146.41 ns 146.55 ns 146.70 ns]
                        change: [+2.7780% +2.9471% +3.1086%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild

simple-log-with-double/no-context
                        time:   [144.09 ns 144.30 ns 144.49 ns]
                        change: [-2.8585% -2.7094% -2.5562%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
simple-log-with-double/with-context
                        time:   [143.65 ns 143.75 ns 143.85 ns]
                        change: [-0.4290% -0.2448% -0.0662%] (p = 0.01 < 0.05)
  2 (2.00%) low mild
  1 (1.00%) high mild

simple-log-with-string/no-context
                        time:   [138.76 ns 138.87 ns 139.00 ns]
                        change: [-7.9410% -7.7945% -7.6528%] (p = 0.00 < 0.05)
                        Performance has improved.
  8 (8.00%) high mild
simple-log-with-string/with-context
                        time:   [147.51 ns 147.61 ns 147.74 ns]
                        change: [+2.4718% +2.6915% +2.8987%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high severe

simple-log-with-bool/no-context
                        time:   [144.49 ns 144.57 ns 144.64 ns]
                        change: [-1.3301% -1.2371% -1.1458%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low severe
  2 (2.00%) low mild
  4 (4.00%) high mild
simple-log-with-bool/with-context
                        time:   [147.11 ns 147.26 ns 147.41 ns]
                        change: [+2.9351% +3.1008% +3.2648%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) high mild

simple-log-with-bytes/no-context
                        time:   [167.25 ns 167.39 ns 167.53 ns]
                        change: [+3.4868% +3.5855% +3.6920%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
  1 (1.00%) high severe
simple-log-with-bytes/with-context
                        time:   [164.25 ns 164.52 ns 164.77 ns]
                        change: [-3.4532% -3.2357% -3.0189%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild

simple-log-with-a-lot-of-bytes/no-context
                        time:   [158.78 ns 158.94 ns 159.09 ns]
                        change: [-6.1546% -5.9827% -5.7952%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) high mild
  2 (2.00%) high severe
simple-log-with-a-lot-of-bytes/with-context
                        time:   [170.54 ns 171.53 ns 173.46 ns]
                        change: [+3.6391% +4.1090% +5.1313%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high severe

simple-log-with-vec-any-value/no-context
                        time:   [203.13 ns 203.24 ns 203.36 ns]
                        change: [-6.0749% -5.9717% -5.8644%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  3 (3.00%) high mild
simple-log-with-vec-any-value/with-context
                        time:   [211.36 ns 211.71 ns 212.08 ns]
                        change: [-1.1547% -0.9804% -0.8161%] (p = 0.00 < 0.05)

simple-log-with-inner-vec-any-value/no-context
                        time:   [276.62 ns 276.81 ns 277.00 ns]
                        change: [+0.8487% +0.9492% +1.0494%] (p = 0.00 < 0.05)
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
simple-log-with-inner-vec-any-value/with-context
                        time:   [274.47 ns 274.97 ns 275.65 ns]
                        change: [-1.1943% -0.8690% -0.5281%] (p = 0.00 < 0.05)
  7 (7.00%) high severe

simple-log-with-map-any-value/no-context
                        time:   [233.37 ns 233.47 ns 233.59 ns]
                        change: [-2.3170% -2.1937% -2.0710%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
simple-log-with-map-any-value/with-context
                        time:   [233.39 ns 233.55 ns 233.71 ns]
                        change: [+0.4716% +0.7618% +1.0012%] (p = 0.00 < 0.05)
  3 (3.00%) high mild
  3 (3.00%) high severe

simple-log-with-inner-map-any-value/no-context
                        time:   [345.49 ns 345.65 ns 345.80 ns]
                        change: [+4.0620% +4.1413% +4.2130%] (p = 0.00 < 0.05)
                        Performance has regressed.
  3 (3.00%) low mild
  2 (2.00%) high mild
simple-log-with-inner-map-any-value/with-context
                        time:   [348.78 ns 349.29 ns 350.15 ns]
                        change: [+5.3044% +5.5604% +5.8275%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe

long-log/no-context     time:   [107.13 ns 107.18 ns 107.24 ns]
                        change: [+0.4446% +0.5142% +0.5924%] (p = 0.00 < 0.05)
  1 (1.00%) high mild
long-log/with-context   time:   [109.07 ns 109.29 ns 109.69 ns]
                        change: [+1.3800% +1.5142% +1.6832%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) high mild
  1 (1.00%) high severe

full-log/no-context     time:   [109.70 ns 109.79 ns 109.89 ns]
                        change: [+3.4539% +3.5922% +3.7297%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) high mild
full-log/with-context   time:   [111.41 ns 111.47 ns 111.55 ns]
                        change: [+0.1869% +0.2796% +0.3707%] (p = 0.00 < 0.05)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe

full-log-with-4-attributes/no-context
                        time:   [246.83 ns 246.95 ns 247.08 ns]
                        change: [-0.3164% -0.0559% +0.1822%] (p = 0.68 > 0.05)
  1 (1.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe
full-log-with-4-attributes/with-context
                        time:   [249.19 ns 249.38 ns 249.59 ns]
                        change: [-4.3881% -4.1521% -3.9539%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe

full-log-with-9-attributes/no-context
                        time:   [442.82 ns 443.27 ns 443.83 ns]
                        change: [-3.3624% -3.0116% -2.6846%] (p = 0.00 < 0.05)
                        Performance has improved.
full-log-with-9-attributes/with-context
                        time:   [454.05 ns 454.27 ns 454.51 ns]
                        change: [-1.8147% -1.7175% -1.6242%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  2 (2.00%) high severe

full-log-with-attributes/no-context
                        time:   [282.22 ns 282.36 ns 282.52 ns]
                        change: [-5.0554% -4.1746% -3.4991%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
full-log-with-attributes/with-context
                        time:   [292.55 ns 292.78 ns 293.09 ns]
                        change: [-2.2768% -1.9404% -1.6585%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
```

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -13,12 +13,27 @@ use opentelemetry::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 
+use std::sync::atomic::AtomicBool;
 use std::{borrow::Cow, sync::Arc};
 
+use lazy_static::lazy_static;
```
shaun-cox, line 19: I think `once_cell` is used more predominantly in the codebase vs. `lazy_static`.
However, there is now support in the standard library via `std::sync::OnceLock` for this too, which would eliminate both dependencies, so please consider if that would be better.
e.g. https://stackoverflow.com/a/27826181/4954738

TommyCpp: Good point. Let me revise

TommyCpp: Ah I don't think we can use `std::sync::OnceLock` as our MSRV is still in 1.65 while the `OnceLock` is introcuded in `1.70`

opentelemetry/src/global/logs.rs
```diff
@@ -25,6 +26,9 @@ pub trait ObjectSafeLoggerProvider {
         &self,
         library: Arc<InstrumentationLibrary>,
     ) -> Box<dyn Logger + Send + Sync + 'static>;
+
+    /// shutdown the logger provider, logs emitted after this will not be processed.
+    fn shutdown(&self) -> Vec<LogResult<()>>;
```
lalitb, line 31: why is this added? I don't think we should have explicit shutdown method at the API.

opentelemetry/src/global/logs.rs
```diff
@@ -132,5 +140,6 @@ where
 
 /// Shut down the current global [`LoggerProvider`].
 pub fn shutdown_logger_provider() {
-    let _ = set_logger_provider(NoopLoggerProvider::new());
+    let logger_provider = set_logger_provider(NoopLoggerProvider::new());
+    let _ = ObjectSafeLoggerProvider::shutdown(&logger_provider);
```
lalitb, line 148: I think we need to revisit `shutdown_*_provider` methods as they are at the API level. The instrumented libraries shouldn't have the method to shutdown the LoggerProvider enabled by the application.

opentelemetry/src/global/logs.rs
TommyCpp: This was added to support the global module shutdown behavior

opentelemetry/src/global/logs.rs
TommyCpp: Discussed in SIG meeting and we will address this separate issue

TommyCpp:
@open-telemetry/rust-approvers can I get another review?

opentelemetry/src/logs/record.rs
```diff
@@ -373,10 +373,13 @@ impl LogRecordBuilder {
     }
 
     /// Sets the `event_name` of a record.
-    pub fn with_name(self, name: Cow<'static, str>) -> Self {
+    pub fn with_name<T>(self, name: T) -> Self
```
cijothomas, line 376: this is unrelated?

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -13,12 +13,25 @@ use opentelemetry::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 
+use std::sync::atomic::AtomicBool;
 use std::{borrow::Cow, sync::Arc};
 
+use once_cell::sync::Lazy;
+
+// a no nop logger provider used as placeholder when the provider is shutdown
```
cijothomas, line 21:
```suggestion
// a no-op logger provider used as placeholder when the provider is shutdown
```

lalitb (approved):

opentelemetry/src/logs/record.rs
TommyCpp: Oh I made the change while I wrote unit testing IIRC. This make it easier for users to input `&'static str`

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -46,7 +47,9 @@ pub trait LogProcessor: Send + Sync + Debug {
     /// Force the logs lying in the cache to be exported.
     fn force_flush(&self) -> LogResult<()>;
     /// Shuts down the processor.
-    fn shutdown(&mut self) -> LogResult<()>;
+    /// After shutdown returns the log processor should stop processing any logs.
+    /// It's up to the implementation on when to drop the LogProcessor.
+    fn shutdown(&self) -> LogResult<()>;
```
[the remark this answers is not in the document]
TommyCpp: They could but I don't think it's necessary. Conceptually the exporter should only be owned by one processor(User can of course build shared ownership inside exporter use `Inner`s)?

Let me know if you disagree and we can bring parity between processor and exporter

opentelemetry-sdk/src/testing/logs/in_memory_exporter.rs
```diff
@@ -83,14 +86,24 @@ impl InMemoryLogsExporterBuilder {
     /// Creates a new instance of `InMemoryLogsExporter`.
     ///
     pub fn new() -> Self {
-        Self {}
+        Self {
+            reset_on_shutdown: true,
+        }
+    }
+
+    /// If set, the records will not be [`InMemoryLogsExporter::reset`] on shutdown.
```
cijothomas, line 94: i am not sure if there is sufficient usecases to expose this new public api. I'd suggest to remove this, and only add, if there is enough demand.

opentelemetry/src/global/logs.rs
```diff
@@ -25,6 +26,9 @@ pub trait ObjectSafeLoggerProvider {
         &self,
         library: Arc<InstrumentationLibrary>,
     ) -> Box<dyn Logger + Send + Sync + 'static>;
+
+    /// shutdown the logger provider, logs emitted after this will not be processed.
```
cijothomas, line 30: Was thinking about this a bit more, specifically for logging.
It makes sense to have these Global things for metrics, traces, so that application can retrive the global provider and then get tracers,meters from them.

But for logs - these is no such use case. LoggerProviders are *not* meant for end users to get loggers from. It is only required for the log bridges. 

The expected usage pattern shown here : https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/examples/basic.rs

It works totally fine as-is. The user creates provider and provides a ref to it, to the bridges, but retains ownership. And they can continue to keep ownership and call shutdown at the end of the application itself. What scenario requires setting the global provider for logging?

@TommyCpp @lalitb 
Sorry for this question after agreeing with the direction of this PR overall :(
I think we should treat logs differently from other signals. The idea shown in this PR can be used for other signals, but not for logs...

Please share your thoughts..

cijothomas (changes requested):
This is probably not required for logging signal...
See https://github.com/open-telemetry/opentelemetry-rust/pull/1643/files#r1579840322

opentelemetry-sdk/src/testing/logs/in_memory_exporter.rs
TommyCpp: I need it for unit test. I can make it pub(crate)

opentelemetry/src/global/logs.rs
TommyCpp: It works totally fine as-is.

Then we should remove shutdown_logger_provider from global module. Otherwise users will just try to call it without actually shutting down anything

cijothomas: Yes and yes.
We need to keep shutdown only for the sdk!

lalitb: >  What scenario requires setting the global provider for logging?

While this repo provides/maintains the tokio-tracing logging bridge, which doesn't need setting global provider, the bridge-api should be extensible enough for users to create their appenders for some other logging API. We can't assume the requirements of that appender, and in that case good to provide the support recommended by [API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#loggerprovider):
```
Thus, the API SHOULD provide a way to set/register and access a global default LoggerProvider.
```

lalitb: Regarding shutdown_signal_provider, I believe it should be removed from global for all signals, as it gives an impression that these can be used by instrumentation libraries. We have an issue for that.

cijothomas: > > What scenario requires setting the global provider for logging?
> 
> While this repo provides/maintains the tokio-tracing logging bridge, which doesn't need setting global provider, the bridge-api should be extensible enough for users to create their appenders for any logging framework. We can't assume the requirements of that appender, and in that case good to provide the support recommended by [API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#loggerprovider):
> 
> ```
> Thus, the API SHOULD provide a way to set/register and access a global default LoggerProvider.
> ```

Both the appenders we have be hosting in this repo did not require global::logger, and it works by accepting LoggerProvider in its new() method. This can be used by other appenders as well.

(If there is good demand, we can expose it, but I'd suggest to keep the public API _as minimal as possible_ for initial stable release.)

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   `ProcessResourceDetector` resource detectors, use the
   [`opentelemetry-resource-detector`](https://crates.io/crates/opentelemetry-resource-detectors) instead.
 - Baggage propagation error will be reported to global error handler [#1640](https://github.com/open-telemetry/opentelemetry-rust/pull/1640)
+- Make `shutdown` method in `LoggerProvider` and `LogProcessor` taking immutable reference [#1643](https://github.com/open-telemetry/opentelemetry-rust/pull/1643)
```
cijothomas, line 16: Lets describe all the changes this PR is making:
1. Loggers obtained after shutdown is now no-ops.
2. Shutdown requires immutable ref for provider/processor.
3. Simple and batch processors will ignore new logrecord after shutdown.

cijothomas (approved):
Thanks! I have a suggestion to make changelog better to reflect the actuals.
Also please update the PR description to reflect the current state, for easy paper-trail in the future.

PR #1649, opened 2024-03-29
Add benchmark for opentelemetry-appender-tracing

lalitb:
Fixes #
## Changes

Added basic benchmarks as of now for the below scenarios with results:
  - `log_no_subscriber` - No subscriber registered - `312.83 ps`
  - `log_subscriber_log_level_disabled`  - Subscriber with event_enabled false - `15.016 ns`    
       // tokio::event_enabled() -> processor::event_enabled() -> exporter::event_enabled()
  - `log_subscriber_log_level_enabled`  - Subscriber with event_enabled true -  `380.09 ns`   
        // tokio::event_enabled() -> processor::event_enabled() -> exporter::event_enabled() 
       //+ tracing::info!() -> processor::emit() -> exporter::export()

image

### DIFF
--- a/opentelemetry-appender-tracing/benches/logs.rs
+++ b/opentelemetry-appender-tracing/benches/logs.rs
@@ -2,119 +2,100 @@ use async_trait::async_trait;
 use criterion::{criterion_group, criterion_main, Criterion};
 use opentelemetry::logs::LogResult;
 use opentelemetry::KeyValue;
-use opentelemetry_appender_tracing::layer;
-use opentelemetry_sdk::export::logs::LogData;
-use opentelemetry_sdk::{
-    logs::{Config, LoggerProvider},
-    Resource,
-};
+use opentelemetry_appender_tracing::layer as tracing_layer;
+use opentelemetry_sdk::export::logs::{LogData, LogExporter};
+use opentelemetry_sdk::logs::{Config, LogProcessor, LoggerProvider};
+use opentelemetry_sdk::Resource;
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::Registry;
 
 #[derive(Debug, Clone)]
-struct VoidExporter {
-    is_enabled: bool,
+struct NoopExporter {
+    enabled: bool,
 }
 
 #[async_trait]
-impl opentelemetry_sdk::export::logs::LogExporter for VoidExporter {
-    async fn export(&mut self, _batch: Vec<LogData>) -> LogResult<()> {
+impl LogExporter for NoopExporter {
+    async fn export(&mut self, _: Vec<LogData>) -> LogResult<()> {
         LogResult::Ok(())
     }
 
-    fn event_enabled(
-        &self,
-        _level: opentelemetry::logs::Severity,
-        _target: &str,
-        _name: &str,
-    ) -> bool {
-        self.is_enabled
+    fn event_enabled(&self, _: opentelemetry::logs::Severity, _: &str, _: &str) -> bool {
+        self.enabled
     }
 }
 
-// disabled log processor
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


cijothomas:
I ❤️ it that we start talking in terms of nanosecs and now picosecs!

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -0,0 +1,102 @@
...
+        name: &str,
+    ) -> bool {
+        self.exporter.event_enabled(level, target, name)
+    }
+}
+
+fn benchmark_no_subscriber(c: &mut Criterion) {
+    c.bench_function("log_no_subscriber", |b| {
+        b.iter(|| {
+            info!("my-event-name");
```
cijothomas, line 66: to more closely match practical uses, lets add name/target + 3-5 fields ?

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -0,0 +1,102 @@
...
+                info!("my-event-name");
+            });
+        });
+    });
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    benchmark_no_subscriber(c);
+    benchmark_with_subscriber(c, false, "log_subscriber_disabled");
+    benchmark_with_subscriber(c, true, "log_subscriber_enabled");
```
cijothomas, line 98: I think it'd be good to add a tracing-subscriber itself, that is noop, and returns false for is_enabled(). So we can truly compare - what is the overhead introduced by OTel when not_enabled! It should be negligible as we copy level/target and instrumentation name, and pass refs to it to processor..
(https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs#L71-L74)

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_emitter.rs#L210 This clone needs some rethinking. Currently, every processor, even the ones that can stream event directly will have to suffer from this clone cost. I think we should just pass ref to logrecord to processors, and if a processor wants to batch it, then it can do the clone and pay the cost. This should give a significant cost reduction.

opentelemetry-appender-tracing/benches/logs.rs
lalitb: yes done.

opentelemetry-appender-tracing/benches/logs.rs
lalitb: Good point, added now.

lalitb:
> https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_emitter.rs#L210 This clone needs some rethinking. Currently, every processor, even the ones that can stream event directly will have to suffer from this clone cost. I think we should just pass ref to logrecord to processors, and if a processor wants to batch it, then it can do the clone and pay the cost. This should give a significant cost reduction.

#1650

lalitb:
The noop layer's benchmark test shows it takes about 12.5 nanoseconds when is_enabled off. This seems to be the smallest delay we can't avoid for our otel subscriber, even after making it as fast as possible. One option is to register the subscriber only when the opentelemetry-sdk is available, through feature-flag, as @TommyCpp suggested, to make things faster for applications that don't use otel-sdk for logging.

cijothomas (approved):

lalitb:
Also checked-in the benchmark results for record. It's beneficial to keep these for future comparisons. Please let me know if there are any issues.

cijothomas:
> Also checked-in the benchmark results for record. It's beneficial to keep these for future comparisons. Please let me know if there are any issues.

yes, I personally love keeping some numbers for quick reference!

PR #1667, opened 2024-04-17
Add benchmarks to show cost of random generation

cijothomas:
Metric benchmarks and stress test includes the cost of random generation in it. This adds a dedicated test to tell how much of the overall cost is arising from the random generation itself, so we get a more realistic idea of the real cost of actual otel code.

Example run:

Counter_Add_Sorted       time:   [796.38 ns 812.20 ns 830.03 ns]
Counter_Add_Unsorted   time:   [772.11 ns 789.88 ns 810.13 ns]
Random_Generator_5      time:   [301.08 ns 307.36 ns 314.33 ns]

### DIFF
--- a/opentelemetry-sdk/benches/metric_counter.rs
+++ b/opentelemetry-sdk/benches/metric_counter.rs
@@ -67,6 +67,17 @@ fn counter_add(c: &mut Criterion) {
             );
         });
     });
+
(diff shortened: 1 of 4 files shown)


lalitb:
Cool, we actually are 300ns better :)

lalitb (approved):

PR #1671, opened 2024-04-18
fix: fail open when upload code coverage fails

TommyCpp:
## Changes

Saw some issues when uploading test reports. This shouldn't block us from getting coverage for the PR and I don't think we should fail on it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,4 +141,4 @@ jobs:
       - name: Upload to codecov.io
         uses: codecov/codecov-action@v3
         with:
-          fail_ci_if_error: true
+          fail_ci_if_error: false


cijothomas (approved):

TommyCpp:
I tried to look it last week but didn't get to the root cause. It looked like a rate limiting issue. So maybe the next step is to make code coverage opt in via labels. We don't need coverage test for every PRs

lalitb (approved):
Thanks. Regarding  upload failure, looks like tokenless uploading is unsupported in codecov server - in that case we need to pick the token value from - https://app.codecov.io/gh/open-telemetry/opentelemetry-rust/settings - and add it as `CODECOV_TOKEN` repository secret to https://github.com/open-telemetry/opentelemetry-rust/settings/secrets/actions and https://github.com/open-telemetry/opentelemetry-rust/settings/secrets/dependabot

TommyCpp:
> Thanks. Regarding upload failure, looks like tokenless uploading is unsupported in codecov server - in that case we need to pick the token value from - https://app.codecov.io/gh/open-telemetry/opentelemetry-rust/settings - and add it as `CODECOV_TOKEN` repository secret to https://github.com/open-telemetry/opentelemetry-rust/settings/secrets/actions and https://github.com/open-telemetry/opentelemetry-rust/settings/secrets/dependabot. I don't have admin access to check the settings, so can't validate it :(

I cannot access the URLs 😞

PR #1675, opened 2024-04-22
Run a stress thread for each logical core

### LINKED ISSUE #1405: [Feature]: Stress Test is not leveraging max threads

ThomsonTan:
Fixes #1405

Currently the stress test run a thread for each physical core with physical core reserved for the main loop. For CPU with hyper-thread enabled, this actually will take less than 50% CPU time for the whole stress run. Fix this by spawning the number of threads as the number of logical cores, but still reserve one logical core for main loop.

Task Manager shows the CPU usage as 100% and stable.

## Changes

### DIFF
--- a/stress/src/throughput.rs
+++ b/stress/src/throughput.rs
@@ -26,7 +26,7 @@ where
         STOP.store(true, Ordering::SeqCst);
     })
     .expect("Error setting Ctrl-C handler");
-    let num_threads = num_cpus::get_physical();
+    let num_threads = num_cpus::get();
     println!("Number of threads: {}", num_threads);
     let mut handles = Vec::with_capacity(num_threads);
     let func_arc = Arc::new(func);


cijothomas:
@ThomsonTan Thanks for looking into it. Can you run the stress tests (pick any!), and share the before/after numbers as well to see what impact switching to logical cores has on overall throughout!

ThomsonTan:
> @ThomsonTan Thanks for looking into it. Can you run the stress tests (pick any!), and share the before/after numbers as well to see what impact switching to logical cores has on overall throughout!

Sure, here are the data collected before and after this PR on my 16 cores devbox.

## Before the PR
* CPU Usage on Cores:
image

* Overall CPU usage:
image

* Throughput
> Number of threads: 8
> Throughput: 4,351,400 iterations/sec

## With this PR
* CPU Usage on Cores
image

* Overall CPU Usage:
image

* Throughput
> Number of threads: 16
> Throughput: 5,961,800 iterations/sec

lalitb (approved):

lalitb:
> Number of threads: 8
Throughput: 4,351,400 iterations/sec

Are these results with release build, if not can you share that number too?

ThomsonTan:
> > Number of threads: 8
> > Throughput: 4,351,400 iterations/sec
> 
> Are these results with release build, if not can you share that number too?

Yes, the previous throughput number is based on debug build. Below is the data from release build. (CPU usage data looks similar and ignore here).

## Throughput before the PR
> Number of threads: 8
> Throughput: 12,116,200 iterations/sec

## Throughput with the PR
> Number of threads: 16
> Throughput: 18,980,000 iterations/sec

cijothomas:
@ThomsonTan I see similar results for logs. But for metrics, it looks like the throughtput drops actually! Might have to do with metrics sdk implementation itself!

TommyCpp:
> Might have to do with metrics sdk implementation itself

We probably do more context switching in metrics than logs because of the aggregation of data points?

lalitb:
> But for metrics, it looks like the throughtput drops actually! 

This behavior has been documented in the PR description : https://github.com/open-telemetry/opentelemetry-rust/pull/1564

``
Based on the above results- with PR, the perf seems to be increasing significantly with the number of threads. In the main branch, the performance increases with threads till a threshold and then starts degrading.

``

cijothomas:
Thanks for your first PR @ThomsonTan !

We understand the throughput is dropping for metrics, but that is [separately](https://github.com/open-telemetry/opentelemetry-rust/pull/1675#issuecomment-2070588100)being addressed.!

stress/src/throughput.rs
```diff
@@ -26,7 +26,7 @@ where
         STOP.store(true, Ordering::SeqCst);
     })
     .expect("Error setting Ctrl-C handler");
-    let num_threads = num_cpus::get_physical();
+    let num_threads = num_cpus::get();
```
cijothomas, line 29: feel like the same advice I constantly give to my little kids! No physical fights 🤣

Issue #1571, opened 2024-02-23
OpenTelemetry Tracing API vs Tokio-Tracing API for Distributed Tracing


cijothomas:
## Background

The Rust ecosystem has two prominent tracing APIs: the OpenTelemetry Tracing
API (Otel for short), delivered through the `opentelemetry` crate, and the
[Tokio tracing API](https://github.com/tokio-rs/tracing), provided by the
`tracing` crate. The OTel Tracing API adheres to the [OpenTelemetry
specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md),
ensuring alignment with OpenTelemetry Tracing implementations in other
languages like C++, Java etc. Conversely, the Tokio tracing ecosystem, which predates
OpenTelemetry, boasts widespread adoption, with many popular libraries already
instrumented. The [tracing-opentelemetry](https://github.com/tokio-rs/tracing-opentelemetry) crate, maintained outside of
OpenTelemetry repositories, enables applications instrumented with tracing to
work with OpenTelemetry.

## The issue

The coexistence of the OTel Tracing API and Tokio-Tracing poses a dilemma,
forcing end users to choose between two competing APIs. This situation
complicates the decision-making process due to the absence of comprehensive
documentation comparing the two options. A significant concern is the lack of
tested interoperability between the APIs, which can result in issues, especially
in applications where different layers use different tracing APIs, potentially
leading to incomplete traces. This also impacts the [log correlation](https://github.com/open-telemetry/opentelemetry-rust/issues/1378) scenarios as well.

### A Comparison with OTel .NET

The OpenTelemetry .NET community encountered a similar challenge when the OTel
Tracing API was introduced, as the .NET [runtime ](https://github.com/dotnet/runtime) library (shipped as the
[DiagnosticSource ](https://www.nuget.org/packages/system.diagnostics.diagnosticsource) package) already had a similar API in place. This issue was
resolved through collaboration between OTel .NET maintainers and the .NET
runtime team, leading to the alignment of the .NET runtime's tracing API with
the OTel specifications. This approach was later applied to the Metrics API as
well. While the decision by OTel .NET to prioritize the .NET Runtime library's
API over its own for tracing/metrics has generally been successful, it has not
been without its challenges. Despite declaring stability years ago, OTel .NET
has yet to implement certain aspects of the OTel specification fully.

Although the outcomes in the .NET ecosystem might not directly forecast the
[...]

PR #1694, opened 2024-04-30
Reflect status of each component

cijothomas:
As discussed in previous community meetings, the main readme is now updated to show the status of individual components as opposed to the overall signal. This gives better transparency, and additionally, we expect API to always stabilize before SDK, so it is easy to reflect them with a breakdown.
Only showing the API/SDK/OTLP Exporter (Plus one appender for logs, without which logs are not really useable) for each signal as we are prioritizing them for initial GA release. (The list could change, and can be updated accordingly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,18 @@ observability tools.
 
 ## Project Status
 
-| Signal  | Status     |
-| ------- | ---------- |
-| Logs    | Alpha*     |
-| Metrics | Alpha      |
-| Traces  | Beta       |
+| Signal/Component      | Overall Status     |
+| --------------------  | ------------------ |
+| Logs-API              | Alpha*             |
+| Logs-SDK              | Alpha              |
+| Logs-OTLP Exporter    | Alpha              |
+| Logs-Appender-Tracing | Alpha              |
+| Metrics-API           | Alpha              |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,11 +25,18 @@ observability tools.
...
+| Signal/Component      | Overall Status     |
+| --------------------  | ------------------ |
+| Logs-API              | Alpha*             |
+| Logs-SDK              | Alpha              |
+| Logs-OTLP Exporter    | Alpha              |
+| Logs-Appender-Tracing | Alpha              |
+| Metrics-API           | Alpha              |
+| Metrics-SDK           | Alpha              |
+| Metrics-OTLP Exporter | Alpha              |
+| Traces-API            | Beta               |
```
lalitb, line 37: should we add a note referring to #1571, as a indication to the users that there may be changes in future in regards to the traces API.

cijothomas: That was also suggested in the SIG call too, but I think we decided to not mention it, until a decision is made on that front. (as of today, there are proposals, but no firm decision has been made.)

lalitb (approved):

cijothomas:
Merging quickly as this is not changing any status, but just adding breakdown.

PR #1695, opened 2024-05-01
Minor cleanups to Metrics API docs

cijothomas:
Nothing major, mostly fixing copy-paste errors and other small fixes.

### DIFF
--- a/opentelemetry/src/global/mod.rs
+++ b/opentelemetry/src/global/mod.rs
@@ -93,7 +93,7 @@
 //! use opentelemetry::{global, KeyValue};
 //!
 //! fn init_meter() {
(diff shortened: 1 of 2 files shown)


opentelemetry/src/global/mod.rs
```diff
@@ -93,6 +93,8 @@
 //! use opentelemetry::{global, KeyValue};
 //!
 //! fn init_meter() {
+//!     // Swap this no-op provider with an actual metrics provider,
```
lalitb, line 96:
```suggestion
//!     // Swap this no-op provider with an actual meter provider,
```

lalitb (approved):
Thanks.

PR #1702, opened 2024-05-02
Moving LogRecord implementation to SDK

### LINKED ISSUE #1189: Logs API/SDK - Move LogRecord implementation details from API to SDK

lalitb:
Fixes #1189 

## Changes

Opening as draft to get the initial feedback. The changes are:
 - Created `LogRecord`  traits in the API.
 - Added corresponding struct `LogRecord` with implementation in the SDK.

**Earlier:**
```rust
        logger.emit(
            LogRecord::builder()
                .with_body("full log")
                .with_timestamp(now)
                .with_observed_timestamp(now)
                .with_severity_number(Severity::Warn)
                .with_severity_text(Severity::Warn.name())
                .with_attribute("name", "my-event-name")
                .with_attribute("event.id", 20)
                .with_attribute("user.name", "otel")
                .with_attribute("user.email", "otel@opentelemetry.io")
                .with_attribute("code.filename", "log.rs")
                .with_attribute("code.filepath", "opentelemetry_sdk/benches/log.rs")
                .with_attribute("code.lineno", 96)
                .with_attribute("code.namespace", "opentelemetry_sdk::benches::log")
                .with_attribute("log.target", "opentelemetry_sdk::benches::log")
                .build(),
        )

```

**With PR:**
```rust
        let mut log_record = logger.create_log_record();
        log_record.set_body("full log".into());
        log_record.set_timestamp(now);
        log_record.set_observed_timestamp(now);
        log_record.set_severity_number(Severity::Warn);
        log_record.set_severity_text(Severity::Warn.name().into());
        log_record.add_attribute("name", "my-event-name");
        log_record.add_attribute("event.id", 20);
        log_record.add_attribute("user.name", "otel");
        log_record.add_attribute("user.email", "otel@opentelemetry.io");
        log_record.add_attribute("code.filename", "log.rs");
        log_record.add_attribute("code.filepath", "opentelemetry_sdk/benches/log.rs");
        log_record.add_attribute("code.lineno", 96);
        log_record.add_attribute("code.namespace", "opentelemetry_sdk::benches::log");
        log_record.add_attribute("log.target", "opentelemetry_sdk::benches::log");
        logger.emit(log_record);
```

Note - with the new design, the log record creation through chaining required cloning eventually while emitting, so haven't implemented the chaining. The tracing appender also doesn't require chaining as all these log data are populated separately.

This allows us to change the internal SDK implementation for LogRecord without breaking the Bridge API. While there are no current perf benefits with the approach, it enables us to add more optimizations in the SDK to optionally allow the serialization of the log records in the export format without any intermediate allocations within the SDK. I plan to add those optimizations in subsequent PRs once this is finalized. 

**Benchmark**

| Test Name                                        | Main                            | PR                             |
|--------------------------------------------------|---------------------------------|--------------------------------|
| simple-log/no-context                           | [105.28 ns]                     | [89.896 ns]                    |
| simple-log/with-context time                    | [115.49 ns]                     | [90.293 ns]                    |
| simple-log-with-int/no-context                  | [141.40 ns]                     | [123.36 ns]                    |
| simple-log-with-int/with-context                | [147.15 ns]                     | [123.61 ns]                    |
| simple-log-with-double/no-context               | [145.43 ns]                     | [123.32 ns]                    |
| simple-log-with-double/with-context             | [153.33 ns]                     | [122.28 ns]                    |
| simple-log-with-string/no-context               | [144.01 ns]                     | [122.61 ns]                    |
| simple-log-with-string/with-context             | [149.85 ns]                     | [123.61 ns]                    |
| simple-log-with-bool/no-context                 | [146.57 ns]                     | [121.77 ns]                    |
| simple-log-with-bool/with-context               | [152.56 ns]                     | [123.12 ns]                    |
| simple-log-with-bytes/no-context                | [161.11 ns]                     | [139.43 ns]                    |
| simple-log-with-bytes/with-context              | [171.69 ns]                     | [140.73 ns]                    |
| simple-log-with-a-lot-of-bytes/no-context       | [164.77 ns]                     | [140.72 ns]                    |
| simple-log-with-a-lot-of-bytes/with-context     | [172.50 ns]                     | [142.45 ns]                    |
| simple-log-with-vec-any-value/no-context        | [205.88 ns]                     | [194.93 ns]                    |
| simple-log-with-vec-any-value/with-context      | [214.82 ns]                     | [194.06 ns]                    |
| simple-log-with-inner-vec-any-value/no-context  | [275.55 ns]                     | [256.61 ns]                    |
| simple-log-with-inner-vec-any-value/with-context| [280.79 ns]                     | [255.82 ns]                    |
| simple-log-with-map-any-value/no-context        | [234.09 ns]                     | [207.90 ns]                    |
| simple-log-with-map-any-value/with-context      | [244.29 ns]                     | [207.57 ns]                    |
| simple-log-with-inner-map-any-value/no-context  | [335.55 ns]                     | [303.23 ns]                    |
| simple-log-with-inner-map-any-value/with-context| [344.18 ns]                     | [304.65 ns]                    |
| long-log/no-context                             | [106.17 ns]                     | [89.779 ns]                    |
| long-log/with-context                           | [118.93 ns]                     | [91.673 ns]                    |
| full-log/no-context                             | [143.31 ns]                     | [95.529 ns]                    |
| full-log/with-context                           | [145.20 ns]                     | [96.720 ns]                    |
| full-log-with-4-attributes/no-context           | [315.31 ns]                     | [227.49 ns]                    |
| full-log-with-4-attributes/with-context         | [312.81 ns]                     | [235.11 ns]                    |
| full-log-with-9-attributes/no-context           | [548.88 ns]                     | [406.99 ns]                    |
| full-log-with-9-attributes/with-context         | [501.88 ns]                     | [418.26 ns]                    |
| full-log-with-attributes/no-context             | [322.71 ns]                     | [277.63 ns]                    |
| full-log-with-attributes/with-context           | [328.19 ns]                     | [279.08 ns]                    |

**Stress Test** (20 threads)
Main: ~35 million iterations/sec
PR:    ~39 million iterations/sec

### DIFF
--- a/opentelemetry-sdk/src/logs/record.rs
+++ b/opentelemetry-sdk/src/logs/record.rs
@@ -1,5 +1,5 @@
 use opentelemetry::{
-    logs::{AnyValue, LogRecord, Severity},
+    logs::{AnyValue, Severity},
     trace::{SpanContext, SpanId, TraceFlags, TraceId},
     Key,
 };
@@ -9,7 +9,7 @@ use std::{borrow::Cow, time::SystemTime};
 #[non_exhaustive]
 /// LogRecord represents all data carried by a log record, and
 /// is provided to `LogExporter`s as input.
-pub struct SdkLogRecord {
+pub struct LogRecord {
     /// Record timestamp
     pub timestamp: Option<SystemTime>,
 
@@ -31,7 +31,7 @@ pub struct SdkLogRecord {
     pub attributes: Option<Vec<(Key, AnyValue)>>,
 }
 
-impl LogRecord for SdkLogRecord {
+impl opentelemetry::logs::LogRecord for LogRecord {
     fn set_timestamp(&mut self, timestamp: SystemTime) {
         self.timestamp = Some(timestamp);
     }
@@ -103,7 +103,7 @@ impl From<&SpanContext> for TraceContext {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use opentelemetry::logs::{AnyValue, LogRecord, Severity};
+    use opentelemetry::logs::{AnyValue, LogRecord as _, Severity};
     use opentelemetry::trace::{SpanContext, SpanId, TraceFlags, TraceId};
     use std::borrow::Cow;
     use std::time::SystemTime;
@@ -120,7 +120,7 @@ mod tests {
 
     #[test]
     fn test_set_timestamp() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let now = SystemTime::now();
         log_record.set_timestamp(now);
         assert_eq!(log_record.timestamp, Some(now));
@@ -128,7 +128,7 @@ mod tests {
 
     #[test]
     fn test_set_observed_timestamp() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let now = SystemTime::now();
         log_record.set_observed_timestamp(now);
         assert_eq!(log_record.observed_timestamp, Some(now));
@@ -136,7 +136,7 @@ mod tests {
 
     #[test]
     fn test_set_span_context() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let span_context = SpanContext::new(
             trace_id_from_u128(123),
             span_id_from_u64(456),
@@ -161,7 +161,7 @@ mod tests {
 
     #[test]
     fn test_set_severity_text() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let severity_text: Cow<'static, str> = "ERROR".into(); // Explicitly typed
         log_record.set_severity_text(severity_text);
         assert_eq!(log_record.severity_text, Some(Cow::Borrowed("ERROR")));
@@ -169,7 +169,7 @@ mod tests {
 
     #[test]
     fn test_set_severity_number() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let severity_number = Severity::Error;
         log_record.set_severity_number(severity_number);
         assert_eq!(log_record.severity_number, Some(Severity::Error));
@@ -177,7 +177,7 @@ mod tests {
 
     #[test]
     fn test_set_body() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let body = AnyValue::String("Test body".into());
         log_record.set_body(body.clone());
         assert_eq!(log_record.body, Some(body));
@@ -185,7 +185,7 @@ mod tests {
 
     #[test]
     fn test_set_attributes() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         let attributes = vec![(Key::new("key"), AnyValue::String("value".into()))];
         log_record.set_attributes(attributes.clone());
         assert_eq!(log_record.attributes, Some(attributes));
@@ -193,7 +193,7 @@ mod tests {
 
     #[test]
     fn test_set_attribute() {
-        let mut log_record = SdkLogRecord::default();
+        let mut log_record = LogRecord::default();
         log_record.set_attribute("key", "value");
         assert_eq!(
             log_record.attributes,
--- a/opentelemetry-sdk/src/logs/record.rs
+++ b/opentelemetry-sdk/src/logs/record.rs
@@ -1,92 +1,156 @@
 use opentelemetry::{
-    trace::{SpanContext, SpanId, TraceFlags, TraceId},
+    logs::{AnyValue, LogRecord, LogRecordBuilder, Severity},
+    trace::{SpanContext, SpanId, TraceContextExt, TraceFlags, TraceId},
     Key,
-    logs::{LogRecord, AnyValue, Severity},
 };
 use std::{borrow::Cow, time::SystemTime};
-    
-    #[derive(Debug, Clone, Default)]
-    #[non_exhaustive]
-    /// LogRecord represents all data carried by a log record, and
-    /// is provided to `LogExporter`s as input.
-    pub struct SdkLogRecord {
-        /// Record timestamp
-        pub timestamp: Option<SystemTime>,
-    
-        /// Timestamp for when the record was observed by OpenTelemetry
-        pub observed_timestamp: Option<SystemTime>,
-    
-        /// Trace context for logs associated with spans
-        pub trace_context: Option<TraceContext>,
-    
-        /// The original severity string from the source
-        pub severity_text: Option<Cow<'static, str>>,
-        /// The corresponding severity value, normalized
-        pub severity_number: Option<Severity>,
-    
-        /// Record body
-        pub body: Option<AnyValue>,
-    
-        /// Additional attributes associated with this record
-        pub attributes: Option<Vec<(Key, AnyValue)>>,
-    }
-    
-    impl LogRecord for SdkLogRecord {     
-        fn set_timestamp(&mut self, timestamp: SystemTime) -> &mut Self{
-            self.timestamp = Some(timestamp);
-            self
-        }
-    
-        fn set_observed_timestamp(&mut self, timestamp: SystemTime) -> &mut Self{
-            self.observed_timestamp = Some(timestamp);
-            self
-        }
-    
-        fn set_span_context(&mut self, span_context: &SpanContext) -> &mut Self{
-            self.trace_context =  Some(TraceContext {
-                span_id: span_context.span_id(),
-                trace_id: span_context.trace_id(),
-                trace_flags: Some(span_context.trace_flags()),
-            });
-            self
-        }
-    
-        fn set_severity_text(&mut self, severity_text: Option<Cow<'static, str>>) -> &mut Self{
-            self.severity_text = severity_text.into();
-            self
-        }
-    
-        fn set_severity_number(&mut self, severity_number: Severity) -> &mut Self{
-            self.severity_number = Some(severity_number);
-            self
-        }
-    
-        fn set_body(&mut self, body: Option<AnyValue>) -> &mut Self{
-            self.body = body;
-            self
-        }
-    
-        fn set_attributes(&mut self, attributes: Vec<(Key, AnyValue)>) -> &mut Self{
-            self.attributes = Some(attributes);
-            self
-        }
-    
-        fn set_attribute<K,V>(&mut self, key: K, value: V) -> &mut Self
-        where
-            K: Into<Key>,
-            V: Into<AnyValue>,
-        {
-            if self.attributes.is_none() {
-                self.attributes = Some(Vec::new());
-            }
-            if let Some(ref mut attributes) = self.attributes {
-                attributes.push((key.into(), value.into()));
-            }
-            self
-        }
+
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+/// LogRecord represents all data carried by a log record, and
+/// is provided to `LogExporter`s as input.
+pub struct SdkLogRecord {
+    /// Record timestamp
+    pub timestamp: Option<SystemTime>,
+
+    /// Timestamp for when the record was observed by OpenTelemetry
+    pub observed_timestamp: Option<SystemTime>,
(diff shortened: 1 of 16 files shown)


opentelemetry-sdk/src/logs/record.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+        if let Some(ref mut attributes) = self.attributes {
+            attributes.push((key.into(), value.into()));
+        }
+    }
+}
+
+/// Implementation for LogRecordBuilder
+#[derive(Debug, Default)]
+pub struct SdkLogRecordBuilder {
```
TommyCpp, line 83: Do we really need a builder for `LogRecord`? Looking at the trait all it provides are method to change fields. So `LogRecord` by itself looks like a builder to me.

cijothomas:
> While there are no current perf benefits with the approach,

Can you post benchmarks/stress tests with the changes, to confirm that we are not regressing either. (Just curiosity! I agree that this PR is the right direction)

lalitb:
> > While there are no current perf benefits with the approach,
> 
> Can you post benchmarks/stress tests with the changes, to confirm that we are not regressing either. (Just curiosity! I agree that this PR is the right direction)

Have attached the results in the description. Surprisingly I see the improvement, but don't want to read much out of that as of now :)

opentelemetry-sdk/src/logs/record.rs
lalitb: Yes, I think I over-engineered a bit. Have removed the builder trait now :)

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -0,0 +1,203 @@
...
+    trace::{SpanContext, SpanId, TraceFlags, TraceId},
+    Key,
+};
+use std::{borrow::Cow, time::SystemTime};
+
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+/// LogRecord represents all data carried by a log record, and
+/// is provided to `LogExporter`s as input.
+pub struct SdkLogRecord {
```
TommyCpp, line 12: We have be overriding names like `TracerProvider` between API and SDK for a while. I think it we should be consistent here. Maybe just `LogRecord`? Users can rename as needed when importing

lalitb: Yes, renamed it to `LogRecord` now.

opentelemetry/src/logs/record.rs
```diff
@@ -1,79 +1,46 @@
...
 use std::{borrow::Cow, collections::HashMap, time::SystemTime};
 
-#[derive(Debug, Clone)]
-#[non_exhaustive]
-/// LogRecord represents all data carried by a log record, and
-/// is provided to `LogExporter`s as input.
-pub struct LogRecord {
-    /// Event name. Optional as not all the logging API support it.
-    pub event_name: Option<Cow<'static, str>>,
+/// Abstract interface for managing log records. The implementation is provided by the SDK
```
cijothomas, line 7: Is the term "Abstract interface" common in Rust? 

Suggestion:
"SDK implemented trait for managing log records"

opentelemetry/src/logs/record.rs
```diff
@@ -1,79 +1,46 @@
...
+/// Abstract interface for log records in an observability framework.
```
cijothomas, line 8: consider removing this line.. its redundant since above line is good enough.

opentelemetry/src/logs/record.rs
```diff
@@ -1,79 +1,46 @@
...
+pub trait LogRecord {
+    /// Sets the creation timestamp.
```
cijothomas, line 10:
```suggestion
    /// Sets the time when the event occurred measured by the origin clock, i.e. the time at the source.
```

Liberally "steal" from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp ;)

opentelemetry/src/logs/record.rs
```diff
@@ -1,79 +1,46 @@
...
-            trace_context: None,
-            severity_text: None,
-            severity_number: None,
-            body: None,
-            attributes: None,
-        }
-    }
-}
+    /// Adds multiple attributes.
+    fn set_attributes(&mut self, attributes: Vec<(Key, AnyValue)>);
```
cijothomas, line 29: "set" inherently gives the impression that there will be overriding if key is duplicated, making an SDK implementation which does not de-dup bit awkward.. 
Maybe add_attributes, add_attribute are better suited?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// The original severity string from the source
+    pub severity_text: Option<Cow<'static, str>>,
+    /// The corresponding severity value, normalized
+    pub severity_number: Option<Severity>,
+
+    /// Record body
+    pub body: Option<AnyValue>,
+
+    /// Additional attributes associated with this record
+    pub attributes: Option<Vec<(Key, AnyValue)>>,
```
cijothomas, line 31: nit: is there any recommended way of using Option<None> vs empty vec! here?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -0,0 +1,203 @@
...
+/// span.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct TraceContext {
+    /// Trace id
+    pub trace_id: TraceId,
+    /// Span Id
+    pub span_id: SpanId,
+    /// Trace flags
+    pub trace_flags: Option<TraceFlags>,
```
cijothomas, line 90: wondering why is this optional?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        if let Some(ref mut attrs) = self.attributes {
+            attrs.push((key.into(), value.into()));
+        } else {
+            self.attributes = Some(vec![(key.into(), value.into())]);
+        }
+    }
+}
+
+/// TraceContext stores the trace data for logs that have an associated
```
cijothomas, line 80:
```suggestion
/// TraceContext stores the trace context for logs that have an associated
```

opentelemetry/src/logs/record.rs
```diff
@@ -1,79 +1,46 @@
...
 
-    /// Record timestamp
-    pub timestamp: Option<SystemTime>,
+    /// Sets the observed event timestamp.
+    fn set_observed_timestamp(&mut self, timestamp: SystemTime);
 
-    /// Timestamp for when the record was observed by OpenTelemetry
-    pub observed_timestamp: SystemTime,
+    /// Associates a span context.
+    fn set_span_context(&mut self, context: &SpanContext);
```
cijothomas, line 17: I am totally okay to remove this method. As of today, SDK sets this automatically. For now, we can rely on that, and later add this API, if needed.

cijothomas: I think we should remove this, as SDK currently overwrites this. Better to not offer this API now, and wait for the tokio-tracing discussions to finish before exposing this.

cijothomas (commented):
LGTM.
Waiting to add changelog, and remove the API to set span/trace context before approving.

opentelemetry-sdk/src/logs/record.rs
lalitb: I felt Option would be better, specifically for the no attributes scenario. The empty vector still has some minimal heap presence, along with memory required for storing length and capacity.

opentelemetry-sdk/src/logs/record.rs
lalitb: Oh, this was existing code in API. I can have a look into this separately.

opentelemetry/src/logs/record.rs
lalitb: Makes sense, it's optional as per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#logger-operations) too:

``
The [Context](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/README.md) associated with the LogRecord. The API MAY implicitly use the current Context as a default behavior.
``

lalitb: Just to confirm, you mean both `set_span_context` and `set_context` ?

opentelemetry/src/logs/record.rs
lalitb: Done.

opentelemetry/src/logs/record.rs
lalitb: Done.

opentelemetry/src/logs/record.rs
lalitb: Agree, and changed accordingly.

opentelemetry/src/logs/record.rs
lalitb: Ok removed both for now :)

lalitb:
> LGTM. Waiting to add changelog, and remove the API to set span/trace context before approving.

changelog is added now. Thanks for the review.

cijothomas (approved):
Thanks! Great to see API getting leaner and delegating things to the SDK!

opentelemetry-sdk/src/logs/record.rs
utpilla: From the [docs](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.new), it looks like empty vector does not allocate:

> The vector will not allocate until elements are pushed onto it.

lalitb: Yes, I read somewhere that some minimal bookkeeping allocation done on heap initially, but this may be wrong. Irrespective of that, I think enum allocation for None would be minimal as compared to the empty vector :)

utpilla: The allocation for the `Option` should be same regardless of whether it was `None` or `Some`, right? It should just depend on the type `T` that's provided to the `Option`.

```rust
let option_none: Option<Vec<(&str, u32)>> = None;

let mut vec = Vec::<(&str, u32)>::new();
vec.push(("a", 1));
vec.push(("b", 2));
vec.push(("c", 3));

let option_some = Some(vec);

println!("{}", mem::size_of::<Option<Vec<(&str, u32)>>>()); //24
println!("{}", mem::size_of::<Vec::<(&str, u32)>>()); //24
println!("{}", mem::size_of_val(&option_none)); //24
println!("{}", mem::size_of_val(&option_some)); //24
```

utpilla: > But happy to change if you feel otherwise :)

I'm not sure if using `Option` is the more idiomatic thing but I think `Option` adds another layer of indirection to check if there were any attributes present. Even if the `Option` is a `Some` variant, it could have an empty vector wrapped in `Some`. With using `Vec` as the type for the `attributes` field, we don't have this indirection. Other than that, I don't know of any other benefit to using `Vec` here 😄

lalitb: Good points. I guess one of the benefits of moving LogRecord to SDK is that we can change this internal representations afterwards too :)

cijothomas:
@TommyCpp could you take another look?

@lalitb the CI is red, I think it is related to doc failure in this pr. can you fix that?

cijothomas:
@lalitb also - can you update PR desc. to reflect the final change in the PR? (Set vs Add and anything else) Thank you!

lalitb:
> @lalitb the CI is red, I think it is related to doc failure in this pr. can you fix that?
done

> @lalitb also - can you update PR desc. to reflect the final change in the PR? (Set vs Add and anything else) Thank you!
done.

TommyCpp (approved):
Looks good. Thanks for the change

PR #1704, opened 2024-05-02
Code changes for the new Clippy version and running test on Windows

utpilla:
## Changes
- There's a newer version of the rust toolchain which brings in newer clippy lints:
   - https://rust-lang.github.io/rust-clippy/master/index.html#/implied_bounds_in_impls
   - https://rust-lang.github.io/rust-clippy/master/index.html#/assigning_clones
   - https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_clippy_cfg
   - https://rust-lang.github.io/rust-clippy/master/index.html#/let_unit_value

- Fix some tests to resolve the complaints from `cargo test` on Windows

After these changes, the commands specified in the [test.sh](https://github.com/open-telemetry/opentelemetry-rust/blob/main/scripts/test.sh) file run successfully on Windows:
```sh

cargo test --workspace --all-features "$@" -- --test-threads=1

# See https://github.com/rust-lang/cargo/issues/5364
cargo test --manifest-path=opentelemetry/Cargo.toml --no-default-features

# Run global tracer provider test in single thread
cargo test --manifest-path=opentelemetry/Cargo.toml --all-features -- --ignored --test-threads=1

cargo test --manifest-path=opentelemetry/Cargo.toml --all-features
cargo test --manifest-path=opentelemetry-jaeger/Cargo.toml --all-features -- --test-threads=1
cargo test --manifest-path=opentelemetry-zipkin/Cargo.toml --all-features
```

### DIFF
--- a/opentelemetry-jaeger/src/exporter/config/collector/http_client.rs
+++ b/opentelemetry-jaeger/src/exporter/config/collector/http_client.rs
@@ -129,6 +129,7 @@ impl CollectorHttpClient {
 }
 
 #[cfg(test)]
+#[allow(dead_code)]
 pub(crate) mod test_http_client {
     use async_trait::async_trait;
     use bytes::Bytes;
--- a/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
@@ -74,8 +74,20 @@ static COMMON_ATTRIBUTES: Lazy<[KeyValue; 4]> = Lazy::new(|| {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
-    init_tracer()?;
-    init_metrics()?;
+    let result = init_tracer();
+    assert!(
+        result.is_ok(),
+        "Init tracer failed with error: {:?}",
+        result.err()
+    );
+
+    let result = init_metrics();
+    assert!(
+        result.is_ok(),
+        "Init metrics failed with error: {:?}",
+        result.err()
+    );
+    
     // Opentelemetry will not provide a global API to manage the logger provider. Application users must manage the lifecycle of the logger provider on their own. Dropping logger providers will disable log emitting.
     let logger_provider = init_logs().unwrap();
 
--- a/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
@@ -74,8 +74,8 @@ static COMMON_ATTRIBUTES: Lazy<[KeyValue; 4]> = Lazy::new(|| {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
-    let _ = init_tracer()?;
-    let _ = init_metrics()?;
+    init_tracer()?;
+    init_metrics()?;
     // Opentelemetry will not provide a global API to manage the logger provider. Application users must manage the lifecycle of the logger provider on their own. Dropping logger providers will disable log emitting.
     let logger_provider = init_logs().unwrap();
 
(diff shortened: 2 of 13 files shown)


opentelemetry-otlp/tests/integration_test/tests/integration_tests.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(unix)]
```
utpilla, line 1: These tests use `use std::os::unix::fs::PermissionsExt` which cannot be used on Windows.

opentelemetry-prometheus/tests/integration_test.rs
```diff
@@ -405,7 +405,12 @@ fn gather_and_compare(registry: prometheus::Registry, expected: String, name: &'
     let encoder = TextEncoder::new();
     let metric_families = registry.gather();
     encoder.encode(&metric_families, &mut output).unwrap();
-    let output_string = String::from_utf8(output).unwrap();
+
+    let output_string = if cfg!(windows) {
+        String::from_utf8(output).unwrap().replace('\n', "\r\n")
```
TommyCpp, line 410: We can build a util function for testing that takes a string and 
1. if os == windows do `replace('\n', "\r\n")`
2. if not return

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -435,9 +435,11 @@ mod tests {
             })
             .expect("callback registration should succeed");
 
+        _ = meter_provider.force_flush();
```
TommyCpp, line 438: why `force_flush` here?

utpilla: Just to trigger the [callback code](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/periodic_reader.rs#L434) deterministically. This lets us assert things on the receiver end with certainty instead of waiting for `interval * 2` milliseconds and hoping that the sender sends the data in that time.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -74,8 +74,8 @@ static COMMON_ATTRIBUTES: Lazy<[KeyValue; 4]> = Lazy::new(|| {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
-    let _ = init_tracer()?;
-    let _ = init_metrics()?;
+    init_tracer()?;
```
lalitb, line 77: If the error was reported for unused return, can we instead handle the return value. Something like (not tested):
```rust
        let result = init_tracer();
        assert!(result.is_ok(), "Init tracer failed with error: {:?}", result.err());
```

opentelemetry-jaeger/src/exporter/config/collector/http_client.rs
```diff
@@ -129,6 +129,7 @@ impl CollectorHttpClient {
 }
 
 #[cfg(test)]
+#[allow(dead_code)]
```
utpilla, line 132: This is to suppress the clippy warning for the struct not being constructed. Check this for more details: https://github.com/open-telemetry/opentelemetry-rust/actions/runs/8943870592/job/24569519425?pr=1704

opentelemetry-prometheus/tests/integration_test.rs
utpilla: Updated.

opentelemetry-jaeger/src/exporter/config/collector/http_client.rs
lalitb: I think the issue is that this module `test_http_client` is always defined, while used only when the feature `rt-tokio` is enabled. Should be define this module only when the `rt-tokio` is enabled. 

```rust
#[cfg(test)]
#[cfg(feature = "rt-tokio")]
pub(crate) mod test_http_client {
    ..
}
```

lalitb (approved):
LGTM apart from the nit comment. Thanks for fixing the rust toolchain upgrade issues also :)

TommyCpp (approved):
👍 Thanks for the fix

opentelemetry-jaeger/src/exporter/config/collector/http_client.rs
utpilla: That's a good idea. Covered this in #1711

PR #1706, opened 2024-05-06
fix: URLs in OTLP exporters

### LINKED ISSUE #1618: [Bug]: opentelemetry-otlp inconsistency between http and grpc export endpoint for traces
### What happened?

When using the http and grpc exporter for traces you have to fill the endpoint value but when using the http one you have to specify the `/v1/traces` path and not with the grpc one. You can see in your examples [here](https://github.com/open-telemetry/opentelemetry-rust/blob/a8b308ae929e078a1fbb591f208ddc54bb68a055/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs#L42) and [here](https://github.com/open-telemetry/opentelemetry-rust/blob/a8b308ae929e078a1fbb591f208ddc54bb68a055/opentelemetry-otlp/examples/basic-otlp/src/main.rs#L24) it's both different. But under the hood it's configuring the struct `ExportConfig` and in the documentation for [endpoint](https://docs.rs/opentelemetry-otlp/0.15.0/opentelemetry_otlp/struct.ExportConfig.html#structfield.endpoint) it's explicit that it should be the `base address`. So I think there is a bug in the configuration here.

To keep backward compatibility maybe it could be interesting to directly check if the address already contains `/v1/traces` we don't modify the URL.

### API Version

0.13.0

### SDK Version

0.13.0

OTLP

### Relevant log output

_No response_

TommyCpp:
Fixes #1618

## Changes

- Clearify the documentation that the `endpoint` in `ExporterConfig` should be complete. i.e exporter will not append signal path onto it.

- Removed the signal path for default grpc endpoints. As in grpc protocol, [path is decided by the package name](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md). Users can still add path if they manually config it
> Path is case-sensitive. Some gRPC implementations may allow the Path format shown above to be overridden, but this functionality is strongly discouraged. gRPC does not go out of its way to break users that are using this kind of override, but we do not actively support it, and some functionality (e.g., service config support) will not work when the path is not of the form shown above.

- Don't append anything if the users add a path in `ExportConfig`. URLs from env vars don't append anything so we should keep them consistent

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixed
 
 - URL encoded values in `OTEL_EXPORTER_OTLP_HEADERS` are now correctly decoded. [#1578](https://github.com/open-telemetry/opentelemetry-rust/pull/1578)
+- OTLP exporter will not change the URL added through `ExportConfig` [#1706](https://github.com/open-telemetry/opentelemetry-rust/pull/1706)
+- Default grpc endpoint will not have path based on signal(e.g `/v1/traces`) [#1706](https://github.com/open-telemetry/opentelemetry-rust/pull/1706)
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


TommyCpp:
block on #1704 for the new version fix

cijothomas (approved):

lalitb (approved):

PR #1707, opened 2024-05-06
[Metrics API] InstrumentBuilder modified to take instrumentprovider instead of meter

cijothomas:
`InstrumentBuilder` modified to store `InstrumentProvider` instead of `Meter`. There is no need to keep anything from `Meter` other than the `InstrumentProvider` anyway, so this just reduced on indirection. Tried only for sync instruments - if this is good, I'll update the async-instruments too with the same pattern.

### DIFF
--- a/opentelemetry/src/metrics/instruments/mod.rs
+++ b/opentelemetry/src/metrics/instruments/mod.rs
@@ -25,22 +25,22 @@ pub trait AsyncInstrument<T>: Send + Sync {
 }
 
 /// Configuration for building a sync instrument.
-pub struct InstrumentBuilder<T> {
-    instrument_provider: Arc<dyn InstrumentProvider + Send + Sync>,
+pub struct InstrumentBuilder<'a, T> {
+    instrument_provider: &'a dyn InstrumentProvider,
     name: Cow<'static, str>,
     description: Option<Cow<'static, str>>,
     unit: Option<Unit>,
     _marker: marker::PhantomData<T>,
 }
 
-impl<T> InstrumentBuilder<T>
+impl<'a, T> InstrumentBuilder<'a, T>
 where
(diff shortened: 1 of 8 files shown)


opentelemetry/src/metrics/instruments/mod.rs
```diff
@@ -23,22 +25,22 @@ pub trait AsyncInstrument<T>: Send + Sync {
 }
 
 /// Configuration for building a sync instrument.
-pub struct InstrumentBuilder<'a, T> {
-    meter: &'a Meter,
+pub struct InstrumentBuilder<T> {
+    instrument_provider: Arc<dyn InstrumentProvider + Send + Sync>,
```
lalitb, line 29: Need to check the flow, but would this `dyn` has the cost when the hot path is used as below:
```rust
meter.f64_up_down_counter().add()
```

cijothomas: No. This is only in the instrument creation time. No change in the hotpath in this PR.

cijothomas: oh wait.. if you are asking about "what if someone keeps creating instruments repeatedily?" . Then, yes, the perf is going to be very bad with or without this change. That is why docs should strongly warn against it!

TommyCpp: Or we can do something like `&'a I where I: InstrumentProvider + Send + Sync`?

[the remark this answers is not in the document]
cijothomas: yes! made the changes to keep ref to InstrumentBuilder. The Arc is still used by the Meter itself, so don't know if this makes any difference actually!

TommyCpp (approved):

Issue #1712, opened 2024-05-06
Public noops?


cijothomas:
There are a lot of NoOp things that are declared [public](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/metrics/noop.rs#L17). I am not sure if there is any need to have them public at all.

Issue is to track removal of `public` from NoOps, unless required.

cijothomas:
Closing as per this https://github.com/open-telemetry/opentelemetry-rust/pull/1713#issuecomment-2096720535
Not a lot of exposure risk here anyway.

PR #1713, opened 2024-05-06
Reduce public exposure of noops

cijothomas:
Part of https://github.com/open-telemetry/opentelemetry-rust/issues/1712
A lot of others noops are used by opentelemetry-sdk. I think sdk can find different ways to handle this. Will explore that later.

### DIFF
--- a/opentelemetry/src/metrics/noop.rs
+++ b/opentelemetry/src/metrics/noop.rs
@@ -14,13 +14,13 @@ use std::{any::Any, borrow::Cow, sync::Arc};
 
 /// A no-op instance of a `MetricProvider`
 #[derive(Debug, Default)]
-pub struct NoopMeterProvider {
+pub(crate) struct NoopMeterProvider {
     _private: (),
 }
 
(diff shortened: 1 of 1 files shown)


lalitb:
Not sure if I am interpreting correct, but seems the specs says they should be available(exposed?) in the API surface - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/noop.md. Though in general agree to remove if they won't be ever used in libraries.

cijothomas:
> Not sure if I am interpreting correct, but seems the specs says they should be available(exposed?) in the API surface - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/noop.md. Though in general agree to remove if they won't be ever used in libraries.

It could be added back to public api of `opentelemetry`, if necessary. As long as all the APIs are no-op, even without exposing the no-op as-is, we should be good.

note: I don't think it is extra risky to expose the noops anyway, as the traits are anyway public!

PR #1714, opened 2024-05-06
Add CI coverage for Windows

utpilla:
## Changes
- Update CI workflow to run jobs on Windows as well

Note: Certain jobs from the CI workflow can only be run on Linx for now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,10 @@ jobs:
   test:
     strategy:
       matrix:
-        # test all Rust versions on ubuntu-latest
+        # test both stable and beta versions of Rust on ubuntu-latest
         os: [ubuntu-latest]
         rust: [stable, beta]
-        # test stable Rust on Windows
+        # test only stable version of Rust on Windows
         include:
           - rust: stable
             os: windows-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,8 @@ jobs:
   external-types:
     strategy:
       matrix:
-        os: [ windows-latest, ubuntu-latest ]
         example: [opentelemetry, opentelemetry-sdk, opentelemetry-otlp, opentelemetry-zipkin]
-    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,9 +34,12 @@ jobs:
       run: rustup set profile minimal
     - uses: arduino/setup-protoc@v3
     - name: Test
-      run: ./scripts/test.sh
+      run: bash ./scripts/test.sh
   lint:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ windows-latest, ubuntu-latest ]
```
cijothomas, line 41: does it matter if lint checks are done on windows/linux/either?

utpilla: Good point! Updated the workflow to run `docs` and `lint` jobs only on Linux.

lalitb (approved):
Thanks.

cijothomas:
Consider doing windows for only the test/stable for now, given it is much slower. We can add it for all, later.
Also see what other repos like tracing-rs is doing.

.github/workflows/ci.yml
```diff
@@ -32,7 +34,7 @@ jobs:
       run: rustup set profile minimal
     - uses: arduino/setup-protoc@v3
     - name: Test
-      run: ./scripts/test.sh
+      run: bash ./scripts/test.sh
```
ThomsonTan, line 37: Is there some scenario where the `bash` prefix is required?

utpilla: `bash` prefix lets us run `.sh` files on Windows.

TommyCpp (approved):
Looks good. Thanks

utpilla:
> Consider doing windows for only the test/stable for now, given it is much slower. We can add it for all, later. Also see what other repos like tracing-rs is doing.

@cijothomas I checked the tracing repo and for non-Linux platforms, they only test using the stable version of Rust. Check [this](https://github.com/tokio-rs/tracing/blob/master/.github/workflows/CI.yml#L194-L210). I have updated the workflow to do the same. I have also updated code coverage workflow to only run on Linux.

cijothomas (approved):

PR #1715, opened 2024-05-06
Simpler usage of Observable instruments

cijothomas:
There are 2 ways today to provide callbacks for Observable Instruments today:
1. As part of instrument creation itself (via `with_callback`) 
2. After instrument creation, via `meter.register_callback`

After careful consideration, I'm leaning towards removing option 2. While it offers flexibility by allowing one callback to be associated with multiple instruments, and the ability to "un-register", it significantly increases complexity and requires more extensive testing to ensure correctness. It also has a lot of public API exposure, so removing this from 1st stable release also allow us to maintain a manageable scope.

In the upcoming PR(s), I'll make changes to modify all examples/tests to utilize option 1 (this PR just adds a single test) If there are no scenarios where option 2 is essential, I'll proceed with its removal in subsequent iterations.

@open-telemetry/rust-approvers Please share your thoughts on removing option2. (To be clear, this is proposed to be removed to **keep scope in control**, as this is not only complex to use, but equally challenging to ensure sufficient test coverage/correctness)

Edit: Updated all examples and tests to use option1. (There is dire shortage of tests overall :( )

### DIFF
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -43,22 +43,20 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     );
 
     // Create a ObservableCounter instrument and register a callback that reports the measurement.
-    let observable_counter = meter
+    let _observable_counter = meter
         .u64_observable_counter("my_observable_counter")
         .with_description("My observable counter example description")
         .with_unit(Unit::new("myunit"))
+        .with_callback(|observer| {
+            observer.observe(
+                100,
+                &[
+                    KeyValue::new("mykey1", "myvalue1"),
+                    KeyValue::new("mykey2", "myvalue2"),
+                ],
+            )
+        })
         .init();
-
-    meter.register_callback(&[observable_counter.as_any()], move |observer| {
-        observer.observe_u64(
-            &observable_counter,
-            100,
-            &[
-                KeyValue::new("mykey1", "myvalue1"),
-                KeyValue::new("mykey2", "myvalue2"),
-            ],
-        )
-    })?;
 
     // Create a UpCounter Instrument.
     let updown_counter = meter.i64_up_down_counter("my_updown_counter").init();
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -71,22 +71,20 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     );
 
     // Create a Observable UpDownCounter instrument and register a callback that reports the measurement.
-    let observable_up_down_counter = meter
+    let _observable_up_down_counter = meter
         .i64_observable_up_down_counter("my_observable_updown_counter")
         .with_description("My observable updown counter example description")
         .with_unit(Unit::new("myunit"))
+        .with_callback(|observer| {
+            observer.observe(
+                100,
+                &[
+                    KeyValue::new("mykey1", "myvalue1"),
+                    KeyValue::new("mykey2", "myvalue2"),
+                ],
+            )
+        })
         .init();
-
-    meter.register_callback(&[observable_up_down_counter.as_any()], move |observer| {
-        observer.observe_i64(
-            &observable_up_down_counter,
-            100,
-            &[
-                KeyValue::new("mykey1", "myvalue1"),
-                KeyValue::new("mykey2", "myvalue2"),
-            ],
-        )
-    })?;
 
     // Create a Histogram Instrument.
     let histogram = meter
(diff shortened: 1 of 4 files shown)


lalitb:
> It also has a lot of public API exposure, so removing this from 1st stable release also allow us to maintain a manageable scope.

Should we remove it from the code, or move it under the experimental flag?

cijothomas:
> > It also has a lot of public API exposure, so removing this from 1st stable release also allow us to maintain a manageable scope.
> 
> Should we remove it from the code, or move it under the experimental flag?

I'd prefer to remove it from code completely. Putting under experimental flag without investing in tests is not really helpful. Also, with feature flags, we still need to make sure things build successfully, somewhat defeating the goal of reducing scope!

Not all languages have implemented this capability, and it is not a MUST in the spec either. Let us start with offering just the simple API.

TommyCpp (commented):
I think we should be cautious on removing APIs we advertised in examples. Even to control the scope. If we remove too many of those APIs people may not want to upgrade to a stable version. Or they will take the oppounity to switch to something more stable

TommyCpp (commented):
I think we should be cautious on removing APIs we advertised in examples. Even to control the scope. If we remove too many of those APIs people may not want to upgrade to a stable version. Or they will take the oppounity to switch to something more stable

cijothomas:
> I think we should be cautious on removing APIs we advertised in examples. Even to control the scope. If we remove too many of those APIs people may not want to upgrade to a stable version. Or they will take the oppounity to switch to something more stable

Yes, I completely understand your concern! It is a delicate balance. Delaying the stable release indefinitely also poses its own drawbacks. I am hesitant to keep **any** API without thoroughly considering its use cases and adequate test coverage, which sometimes conflicts with preserving backward compatibility and keeping initial scope in check. Given the [planned beta](https://github.com/open-telemetry/opentelemetry-rust/milestone/13) release is closer, now is the apt moment to accept breaking API changes while we're still in the alpha phase.

lalitb:
Can we instead document "option 2" as unstable, and clearly state that it is not part of the beta release.  Additionally, also mark it as `deprecated`. This way, it won't block us for beta release, and would be heads up to users that this MAY be removed before the stable release.  Just concerned about removing it abruptly even in current alpha state :)

cijothomas:
> Just concerned about removing it abruptly even in current alpha state :)

I'll be happy to discuss more in the community call, but if we cannot break public API in an *alpha* release, what is the whole point of the notion of "alpha"?

TommyCpp:
> but if we cannot break public API in an alpha release, what is the whole point of the notion of "alpha"?

We can and probably should break public API in alpha release. And I am OK with removing advance API like `view`s. But this API seems like a fundational one and one we have been advertising(using example) a lot. I personally think we should have it by the time we GA the metrics signal. Thus, marking it as unstable/experiemental during beta phase seems most ideal in this case. 

Either case given we won't have it stable we should still change the examples

cijothomas:
> But this API seems like a fundational one and one we have been advertising(using example) a lot

We are not losing functionality, just a diff. way to do the same. The examples are modified to show the alternate option (which, btw, already existed, but examples didn't show it well)

cijothomas:
> I personally think we should have it by the time we GA the metrics signal.

Any particular reason why we need that? We have other (simpler) ways to register callbacks already!

examples/metrics-basic/src/main.rs
```diff
@@ -43,23 +43,21 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
 
     // Create a ObservableCounter instrument and register a callback that reports the measurement.
-    let observable_counter = meter
+    let _observable_counter = meter
         .u64_observable_counter("my_observable_counter")
         .with_description("My observable counter example description")
         .with_unit(Unit::new("myunit"))
+        .with_callback(|observer| {
+            observer.observe(
+                100,
```
cijothomas, line 52: this is still type-safe. i.e passing 100.0 will be blocked by compiler. But the API itself is more consistent with sync versions, where we don't use record_i64 or record_u64 etc. instead they are just record,add.

examples/metrics-basic/src/main.rs
```diff
@@ -43,23 +43,21 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
+                &[
+                    KeyValue::new("mykey1", "myvalue1"),
+                    KeyValue::new("mykey2", "myvalue2"),
+                ],
+            )
+        })
         .init();
 
-    meter.register_callback(&[observable_counter.as_any()], move |observer| {
```
cijothomas, line 52: From Community Meeting 5/7:

Lalit, Zhongyang:
1. Mark this under unstable flag.
2. Unregister callback is only available with this.
3. This is part of API, so higher bar to change than SDK/View/Exporter
4. Removing this and adding in future might be difficult to achieve without breaking change.

Zhongyang: Opposed to removing it, because it was part of documented API, not because it is *required*.
Suggestion: Mark as unstable, and make it just compile - so let user take the risk. (docs are showing the right way).
Okay with: If this is too complex to achieve, Zhongyang won't block from proceeding with removal.

Lalit: +1 to above.

lalitb (approved):
LGTM for these examples/tests changes.

TommyCpp (approved):

Issue #1642, opened 2024-03-25
Cost of Key, Value abstractions


cijothomas:
Benchmarks from [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/benches/attributes.rs) shows that the cost of creating abstractions like Key,Value in `opentelemetry` crate is non-trivial. 
Take the example below:
```
c.bench_function("CreateOTelKeyValue", |b| {
        b.iter(|| {
            let _v1 = black_box(KeyValue::new("attribute1", "value1"));
        });
    });

    c.bench_function("CreateTupleKeyValue", |b| {
        b.iter(|| {
            let _v1 = black_box(("attribute1", "value1"));
        });
    });
```
Its 7.7199 **ns** vs 786.74 **ps** (10x!)

It may be very-hard or even impossible to avoid the overhead. Opening an issue to facilitate further discussions. Some options include
1. Do nothing, and accept this as the cost ("OTel tax"). This is simply listed only, not really an option we can afford to take!
2. Make Key,Value etc. as Traits with no-op implementation in `opentelemetry` crate, with actual implementations moved to the sdk crate. This should avoid/reduce perf hit when just `opentelemetry` is used.
3. See if we can steal/borrow idea from `tracing` crate, which uses [macros](https://github.com/tokio-rs/tracing/blob/master/tracing/src/macros.rs) that avoid evaluating the attributes completely, unless enabled. The prior art here is definitely worth exploring further.

I also want to emphasize why is this issue is relevant/important! For an application owner who uses OTel, this may not be a significant concern. But this would be a serious (probably blocker) for a library author to bet on OpenTelemetry as the instrumentation API. This is because, by simply enabling OTel API calls, the library's own performance will degrade, even when *nobody* is interested in the telemetry (i.e the final app owner does not enable OTel itself). Whether the few nanosecs overhead is significant or not depends on the library's own latency without OTel! If it is in the order of milliseconds, then probably not a big issue, but Rust is used in perf sensitive scenarios where it won't be surprising to find libraries that perform its core operation in nanosecs order.

cijothomas:
@lalitb has agreed to dig more into where exactly are the overheads arising from, and see if we can bring it to more acceptable levels, before attempting suggestions 2,3 from above or something-else .

cijothomas:
[...]

PR #1716, opened 2024-05-07
Use static dispatch to validate value types for public APIs

utpilla:
Related to #1642 

## Changes
- Use static dispatch to validate the value types

Here are the benchmark results:
| Benchmark Method | Execution Time|
|-----------------------|-----------------|
|CreateOTelKey_Static|1.8588 ns|
|CreateOTelKey_Owned|45.364 ns|
|CreateOTelKey_Arc|46.610 ns|
|CreateOTelKeyValue|4.3361 ns|
|CreateTupleKeyValue|1.2396 ns|
|**ProvideKeyValueUsingGenericMethod** |**621.28 ps**|
|ProvideMultipleKeyValuesUsingGenericMethod |3.4066 ns|
|CreateOtelKeyValueVector|64.357 ns|
|CreateTupleKeyValueVector|46.892 ns|

### DIFF
--- a/opentelemetry/benches/attributes.rs
+++ b/opentelemetry/benches/attributes.rs
@@ -40,6 +40,12 @@ fn attributes_creation(c: &mut Criterion) {
         });
     });
 
+    c.bench_function("CreateTupleKeyValueUsingGenerics", |b| {
+        b.iter(|| {
+            let _v1 = black_box(no_op("attribute1", "value1"));
+        });
+    });
+
     #[allow(clippy::useless_vec)]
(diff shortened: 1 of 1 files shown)


opentelemetry/benches/attributes.rs
```diff
@@ -65,6 +71,17 @@ fn attributes_creation(c: &mut Criterion) {
     });
 }
 
+
+trait OTelValueType {}
+
+impl OTelValueType for u32 {}
```
utpilla, line 77: We can implement this trait for all the OpenTelemetry supported types.

PR #1718, opened 2024-05-07
Metric attributes need sort with keys only

cijothomas:
`HashKeyValue` is a simple wrapper around `KeyValue` to provide hash, eq, ord implementations.
The `Ord` implementation is required to ensure that attributes are treated the same irrespective of the order in which they are passed. This only requires a sort with the `Key` alone.

This PR simplifies the `Ord` implementation to be based of the `Key` part alone from the `KeyValue`. Also added tests to validate that all measurements aggregate to the same point, irrespective of the order in which KeyValue pairs are passed.

(This is part of a somewhat big refactor of Metrics SDK to significantly improve performance - offering short PRs to make reviews easy. Also intentionally keeping the tests modelled like integration tests, instead of unit tests - this is to reduce churn when refactoring, as lot of internals are expected to change, but the end user facing API need to be relatively stable, as we move towards beta release.)

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -607,6 +607,104 @@ mod tests {
         assert_eq!(data_point.value, 30);
     }
 
+    // "multi_thread" tokio flavor must be used else flush won't
+    // be able to make progress!
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_aggregation_attribute_order() {
+        // Run this test with stdout enabled to see output.
+        // cargo test counter_aggregation_attribute_order --features=metrics,testing -- --nocapture
+
+        // Arrange
+        let exporter = InMemoryMetricsExporter::default();
+        let reader = PeriodicReader::builder(exporter.clone(), runtime::Tokio).build();
+        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
+
+        // Act
+        let meter = meter_provider.meter("test");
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -607,6 +607,104 @@ mod tests {
         assert_eq!(data_point.value, 30);
     }
 
+    // "multi_thread" tokio flavor must be used else flush won't
+    // be able to make progress!
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_aggregation_attribute_order() {
```
utpilla, line 613: Add a test to ensure that the dedupe + sort is case-sensitive as well.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -607,6 +607,104 @@ mod tests {
...
+            .data
+            .as_any()
+            .downcast_ref::<data::Sum<u64>>()
+            .expect("Sum aggregation expected for Counter instruments by default");
+
+        // Expecting 1 time-series.
+        assert_eq!(
+            sum.data_points.len(),
+            1,
+            "Expected only one data point as attributes are same, but just reordered."
```
utpilla, line 695: We could probably add an assertion to verify that the attributes of the datapoint match what we expect.

lalitb (approved):

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: coming next PR

PR #1724, opened 2024-05-08
Minor: Add tests to confirm Meter creation behavior

cijothomas:
Very basic test. Need to eventually get to more solid ones, but this is a start to ensure things continue to work while I am working through refactoring some Metric sdk internals.

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -253,6 +253,7 @@ mod tests {
     use crate::testing::metrics::metric_reader::TestMetricReader;
     use crate::Resource;
     use opentelemetry::global;
+    use opentelemetry::metrics::MeterProvider;
     use opentelemetry::{Key, KeyValue, Value};
     use std::env;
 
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -433,4 +434,28 @@ mod tests {
         // Now the shutdown should be invoked
         assert!(reader.is_shutdown());
     }
+
+    #[test]
+    fn same_meter_reused_same_scope() {
+        let provider = super::SdkMeterProvider::builder().build();
+        let _meter1 = provider.meter("test");
```
TommyCpp, line 441:
```suggestion
        // provider should reuse the meter if the name is the same
        let _meter1 = provider.meter("test");
```

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -433,4 +434,28 @@ mod tests {
...
+        let _meter2 = provider.meter("test");
+        assert_eq!(provider.inner.meters.lock().unwrap().len(), 1);
+        let _meter3 =
```
TommyCpp, line 444:
```suggestion
        // don't reuse the meter if the version and schema URL is different
        let _meter3 = 
```

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -433,4 +434,28 @@ mod tests {
...
+            provider.versioned_meter("test", Some("1.0.0"), Some("http://example.com"), None);
+        let _meter4 =
+            provider.versioned_meter("test", Some("1.0.0"), Some("http://example.com"), None);
+        let _meter5 =
+            provider.versioned_meter("test", Some("1.0.0"), Some("http://example.com"), None);
```
TommyCpp, line 449: Maybe also add one for schema URL? Not a blocker

lalitb (approved):

opentelemetry-sdk/src/metrics/meter_provider.rs
cijothomas: Yes, a ton of new tests will be added soon! 
There are more considerations for Meter identities.. not quite sure how to tackle them all. Adding basic tests now, just to be alerted myself while refactoring!

PR #1725, opened 2024-05-08
[Logs] fixing couple of issues in LogRecord move

lalitb:
Fixes #

## Changes 
- event_name got removed from original LogRecord structure
- observed_timestamp is required parameter, default to ts when received by otel-sdk, so made it non-optional.

### DIFF
--- a/opentelemetry-sdk/src/logs/record.rs
+++ b/opentelemetry-sdk/src/logs/record.rs
@@ -5,16 +5,19 @@ use opentelemetry::{
 };
 use std::{borrow::Cow, time::SystemTime};
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
 /// LogRecord represents all data carried by a log record, and
 /// is provided to `LogExporter`s as input.
 pub struct LogRecord {
+    /// Event name. Optional as not all the logging API support it.
+    pub event_name: Option<Cow<'static, str>>,
+
     /// Record timestamp
     pub timestamp: Option<SystemTime>,
 
     /// Timestamp for when the record was observed by OpenTelemetry
-    pub observed_timestamp: Option<SystemTime>,
+    pub observed_timestamp: SystemTime,
 
     /// Trace context for logs associated with spans
     pub trace_context: Option<TraceContext>,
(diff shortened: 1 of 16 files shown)


opentelemetry-sdk/src/logs/record.rs
```diff
@@ -31,13 +34,35 @@ pub struct LogRecord {
...
+        self.event_name = Some(name.into());
+    }
+
     fn set_timestamp(&mut self, timestamp: SystemTime) {
         self.timestamp = Some(timestamp);
     }
 
     fn set_observed_timestamp(&mut self, timestamp: SystemTime) {
-        self.observed_timestamp = Some(timestamp);
+        self.observed_timestamp = timestamp;
```
utpilla, line 65: I think we shouldn't allow users to change observed timestamp as that only depends on when the SDK collected the event.

lalitb: Agree. Let's wait for more comment if all agree :)

TommyCpp: We need to make the `observed_timestamp` field private then. I think that makes sense

lalitb: Going through the [specs ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#emit-a-logrecord) it say's:

```
Emit a LogRecord
The effect of calling this API is to emit a LogRecord to the processing pipeline.

The API MUST accept the following parameters:
...
[Observed Timestamp] If unspecified the implementation SHOULD set it equal to the current time.
```

My interpretation is that the API should provide interface to set the `observed timestamp`, and use current time if not set.

lalitb: Ok, I have reverted `observed_timestamp` type back to Optional, and kept it's setter function for the appenders. If the timestamp is not set, the SDK will set it to it's current time.

TommyCpp: Ah if the spec says so then we should probably leave it as it is

opentelemetry/src/logs/record.rs
```diff
@@ -3,6 +3,11 @@ use std::{borrow::Cow, collections::HashMap, time::SystemTime};
 
 /// SDK implemented trait for managing log records
 pub trait LogRecord {
+    /// Sets the `event_name` of a record
+    fn set_event_name<T>(&mut self, name: T)
```
cijothomas, line 7: can we provide a {} empty implementation, so SDKs need not implement it, unless they care about it.

cijothomas (approved):

opentelemetry/src/logs/record.rs
lalitb: Good point. done.

PR #1636, opened 2024-03-21
[Logs SDK] Send resource once to processor and exporter, and not for every event. 

lalitb:
Fixes  Part of #1124 (for logs)
Design discussion issue (if applicable) #

## Changes
1. Send resource once to processor at the startup:
    a. Add `LogExporter::set_resource(&resource)` method. 
    b. Add `LogProcessor::set_resource(&resource)` method, which invokes `set_resource` on all the exporters.
    c. The LogExporter would store the resource in original or serialized format, and merge it with each event during export.
    d. The LoggerProvider invokes `set_resource` on all the processors during its initialization.
    e. Remove `resource` from the `LogData` structure, which is exported to the exporters through `LogExporter::emit(logdata)`. This ensures that resource is not sent to the exporter with each event.
~~2. Avoid cloning LogRocord during emit if single processor is configured.~~ Will create separate PR for this

call sequence -
`LoggerProvider::build()` -> `LogProcessor::set_resource()` -> `LogExporter::set_resource()`

Currently, these changes apply exclusively to Logs. Once there is consensus and these changes are merged, will extend them to traces and metrics. Additionally, changes made in #1526 can be undone after implementing this across all signals.

**Benchmark**:

***Main branch***:
```
Number of threads: 8
Throughput: 16,356,200 iterations/sec
Throughput: 15,894,400 iterations/sec
Throughput: 16,018,200 iterations/sec
Throughput: 16,344,800 iterations/sec
```

***PR branch***: (with Resource propagation at startup )
```
Number of threads: 8
Throughput: 19,301,400 iterations/sec
Throughput: 19,789,000 iterations/sec
Throughput: 20,133,600 iterations/sec
Throughput: 19,763,000 iterations/sec
```

~~***PR branch***: (with Resource propagation at startup + avoid LogRecord copy for single processor)~~ will create separate PR for this
```
Number of threads: 8
Throughput: 28,997,800 iterations/sec
Throughput: 29,057,800 iterations/sec
Throughput: 28,755,800 iterations/sec
Throughput: 29,712,000 iterations/sec
```

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (if applicable)

PR #1726, opened 2024-05-09
Support LogProcessors chaining through mutable reference

### LINKED ISSUE #1650: Investigate Potential to bypass LogRecord cloning when emitting LogRecord to LogProcessor(s)
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_emitter.rs#L210 This clone needs some rethinking. Currently, every processor, even the ones that can stream event directly will have to suffer from this clone cost. I think we should just pass ref to logrecord to processors, and if a processor wants to batch it, then it can do the clone and pay the cost. This should give a significant cost reduction.

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry-rust/issues/1649#issuecomment-2026570453_

lalitb:
Fixes #1650 
As discussed [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1636#discussion_r1576580577), this PR modifies LogProcessor to take mutable reference to LogData, and further pass it as Owned/Borrowed type to the Exporter. 

**Existing**:
```rust
pub trait LogProcessor: Send + Sync + Debug {
    // ...
    fn emit(&self, data: LogData);
   // ...
}

#[async_trait]
    // ... 
    async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()>;
   // ..
```

**This PR**:
```rust
pub trait LogProcessor: Send + Sync + Debug {
    // ...
    fn emit(&self, data: &mut LogData);
   // ..
}

#[async_trait]
   // ...
   async fn export<'a>(&mut self, batch: Vec<Cow<'a, LogData>>) -> LogResult<()>
  // ...
}
```

- BatchLogProcessor clones the received mutable references of `LogData` before enqueueing them and passes(i.e moves) this cloned data to the exporter
- SimpleLogProcessor passes received mutable references of `LogData` directly to the exporter as borrowed type, without cloning.
- Exporters accept a slice of  `LogData` references and process them accordingly.
- These changes:
      - Enables passing modified LogData between processors in a chain.
      - Allows LogData to be passed from the SDK to the exporter without cloning, such as with `SimpleLogProcessor` and for exporters like` opentelemetry-user-events-logs` and `opentelemetry-etw-logs` that use a `reentrant log processor`.

This also adds certain constraints for implementing custom exporters and log-processors, else their compilation will fail:
**Exporters:**
The lifetime parameter `'a` in `Vec<Cow<'a, LogData>>` ensures that the exporter do not retain references to the logdata beyond its intended lifespan. If there is a need to modify or store LogData beyond the duration of the export call, the data must be cloned. 

**Asynchronous Log-Processors:**
When a LogProcessor processes `LogData` asynchronously, such as in the case of `BatchLogProcessor`, there is a risk that the data may outlive the scope from which it was passed. To prevent any issues, LogData should always be cloned before any asynchronous processing begins.

**Synchronous Log Processors:**
For synchronous processing within a LogProcessor, as in `SimpleLogProcessor`, it is important to ensure that references to LogData are contained within the bounds of the emit function.  The `LogData` should be passed as a reference if all processing is completed within the emit function. 

**Benchmark for otel-sdk/logs**:
**with PR**: full-log-with-attributes/with-context:  444.79 ns
**main repo**:full-log-with-attributes/with-context: 644.23 ns

**Which is 32% improvement**

**Benchmark for otel-tracing-appender/logs**:
**with PR**: 446 ns
**main repo**: 588 ns

**which is 25% improvement.**

**Stress test**
**With PR**: 
Number of threads: 16
Throughput: 39,695,400 iterations/sec
Throughput: 39,273,800 iterations/sec

**main repo**:
Number of threads: 16
Throughput: 25,536,600 iterations/sec
Throughput: 27,350,800 iterations/sec

**which is 45 % improvement in the stress results.**

## Changes


lalitb:
Also the result of stress test. There is perf improvement as we no longer do clone in log_emitter.rs anymore, and we use NoOpLogProcessor  there.

**Existing**:
``` 
Number of threads: 16
Throughput: 36,961,600 iterations/sec
Throughput: 37,929,800 iterations/sec
```

**PR**:
```
Number of threads: 16
Throughput: 64,182,600 iterations/sec
Throughput: 64,921,000 iterations/sec
Throughput: 63,776,800 iterations/sec
```

will compare and add bench too.

lalitb:
**Benchmark**

| Test Name                                        | Main                            | PR                             |
|--------------------------------------------------|---------------------------------|--------------------------------|
| simple-log/no-context                           | [128.97 ns]                     | [89.746 ns]                    |
| simple-log/with-context time                    | [130.56 ns]                     | [89.683 ns]                    |
| simple-log-with-int/no-context                  | [171.34 ns]                     | [106.89 ns]                    |
| simple-log-with-int/with-context                | [ 173.99 ns]                     | [106.38 ns]                    |
| simple-log-with-double/no-context               | [172.04 ns]                     | [106.54 ns]                    |
| simple-log-with-double/with-context             | [173.93 ns]                     | [106.53 ns]                    |
| simple-log-with-string/no-context               | [172.15 ns]                     | [107.70 ns]                    |
| simple-log-with-string/with-context             | [174.03 ns]                     | [106.98 ns]                    |
| simple-log-with-bool/no-context                 | [174.00 ns]                     | [111.34 ns]                    |
| simple-log-with-bool/with-context               | [178.27 ns]                     | [109.40 ns]                    |
| simple-log-with-bytes/no-context                | [209.17 ns]                     | [131.57 ns]                    |
| simple-log-with-bytes/with-context              | [211.31 ns]                     | [131.33 ns]                    |
| simple-log-with-a-lot-of-bytes/no-context       | [216.39 ns]                     | [126.85 ns]                    |
| simple-log-with-a-lot-of-bytes/with-context     | [218.21 ns]                     | [124.83 ns]                    |
| simple-log-with-vec-any-value/no-context        | [285.70 ns]                     | [165.38 ns]                    |
| simple-log-with-vec-any-value/with-context      | [282.09 ns]                     | [169.18 ns]                    |
| simple-log-with-inner-vec-any-value/no-context  | [393.67 ns]                     | [223.43 ns]                    |
| simple-log-with-inner-vec-any-value/with-context| [391.62 ns]                     | [224.96 ns]                    |
| simple-log-with-map-any-value/no-context        | [328.11 ns]                     | [177.65 ns]                    |
| simple-log-with-map-any-value/with-context      | [329.72 ns]                     | [176.94 ns]                    |
| simple-log-with-inner-map-any-value/no-context  | [482.51 ns]                     | [257.19 ns]                    |
| simple-log-with-inner-map-any-value/with-context| [484.81 ns]                     | [252.65 ns]                    |
| long-log/no-context                             | [128.64 ns]                     | [90.043 ns]                    |
| long-log/with-context                           | [129.22 ns]                     | [89.758 ns]                    |
| full-log/no-context                             | [140.29 ns]                     | [98.945 ns]                    |
| full-log/with-context                           | [139.50 ns]                     | [99.310 ns]                    |
| full-log-with-4-attributes/no-context           | [311.59 ns]                     | [196.60 ns]                    |
| full-log-with-4-attributes/with-context         | [299.92 ns]                     | [190.10 ns]                    |
| full-log-with-9-attributes/no-context           | [517.14 ns]                     | [365.58 ns]                    |
| full-log-with-9-attributes/with-context         | [527.35 ns]                     | [367.39 ns]                    |
| full-log-with-attributes/no-context             | [421.30 ns]                     | [248.99 ns]                    |
| full-log-with-attributes/with-context           | [426.44 ns]                     | [239.60 ns]                    |

cijothomas:
Thanks for looking into this! The perf win is amazing! 

Would the below statements be accurate?
1. LogProcessors now get a mutable ref to LogData. If they want to make changes to LogData, they are allowed to do so, but they need to be aware that it'll be reflected in other processors registered after them.
2. If a LogProcessor is an Exporting processor, and it has to batch LogData, then it *must* clone the data. Due to cloning, any changes made to the original one *after* the processor is invoked won't be visible to this LogProcessor.
3. Currently the clone cost is paid by the SDK itself. But this PR makes it possible to write pipeline without any cloning (simple/re-entrant processors who export sync.). BatchingProcessors need to clone, but that is no worse than current.
4. The most common use case of 1 or more enrichment/redactor processor, followed by BatchProcessor would be much faster with this change. Infact this was not even feasible before as enrichment/redactions got applied to the clone only
5. The high-perf use case of 1 or more enrichment/redactor processor, following by Re-entrantProcessors would be much much faster now as we completely eliminate cloning.

Robert's suggestion of offering the ability to run independent pipelines (one pipeline with redaction processor exporting redacted data, and another pipeline exporting without redaction is possible with this change also, but it requires cloning to ensure the changes are only affecting that pipeline only)

lalitb:
> Would the below statements be accurate?

Nicely summarised the changes :) Yes to all.

lalitb:
Have added unit-tests. This is ready for review now.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -251,20 +251,21 @@ impl opentelemetry::logs::Logger for Logger {
...
+        if let Some(ref trace_context) = trace_context {
+            log_record.trace_context = Some(trace_context.clone());
+        }
+        if log_record.observed_timestamp.is_none() {
+            log_record.observed_timestamp = Some(SystemTime::now());
+        }
+
+        let mut data = LogData {
+            record: log_record,
+            instrumentation: self.instrumentation_library().clone(),
```
lalitb, line 264: @utpilla pointed out that instrumentation data shouldn't be allowed to be modified within processors. Will create a separate issue to track that.

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -19,7 +19,13 @@ impl LogExporter for OtlpHttpClient {
                 _ => Err(LogError::Other("exporter is already shut down".into())),
             })?;
 
-        let (body, content_type) = { self.build_logs_export_body(batch, &self.resource)? };
+        //TODO :avoid cloning when logdata is borrowed?
+        let owned_batch = batch
+            .iter()
+            .map(|&log_data| log_data.clone()) // Converts Cow to owned LogData
```
cijothomas, line 25: There is no 🐮 now 🤣

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -13,8 +13,9 @@ use std::fmt::Debug;
 /// `LogExporter` defines the interface that log exporters should implement.
 #[async_trait]
 pub trait LogExporter: Send + Sync + Debug {
-    /// Exports a batch of [`LogData`].
-    async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()>;
+    /// Exports a batch of reference to `LogData`.
+    async fn export<'a>(&mut self, batch: &'a [&'a LogData]) -> LogResult<()>;
```
cijothomas, line 17: is the exported type a slice of ref to LogData? Previously, it was a concrete type `Vec`. Please call it out in changelog. 
(I agree with this change, just making sure the change is aptly reflected in changelog)

opentelemetry-otlp/src/exporter/http/logs.rs
lalitb: unfortunately, had to bring `her` back in latest commit. Too much milking needed in  batchprocessor without that.

opentelemetry-sdk/src/export/logs/mod.rs
lalitb: Had to revert the slice changes as they complicated the handling in the batch processor. Previously, the batch processor directly moved the logs from the queue to the export method. Now, we pass a reference to the logs, track which logs have been passed, and then delete them from the queue. Since the queue can receive new logs while the export is happening, it is necessary to keep track of the logs that were exported. Also, didn't see any improvement in stress and batch tests with the changes.

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -65,9 +66,11 @@ impl LogExporter for TonicLogsClient {
             None => return Err(LogError::Other("exporter is already shut down".into())),
         };
 
+        // TODO: Avoid cloning when logdata is borrowed?
```
cijothomas, line 69: trying to understand the TODO here - if the data is borrowed or not, does it matter for the exporter? Isn't it serializing the data inside export() call itself, and no longer need to keep the original data around....

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -777,11 +797,114 @@ mod tests {
...
+
+    #[derive(Debug)]
+    struct FirstProcessor {
+        pub(crate) logs: Arc<Mutex<Vec<LogData>>>,
+    }
+
+    impl LogProcessor for FirstProcessor {
+        fn emit(&self, data: &mut LogData) {
+            data.record.attributes.get_or_insert(vec![]).push((
+                Key::from_static_str("processed_by"),
```
cijothomas, line 813: nit: the body can also be modified, so that can be leveraged easily for testing too.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -6,6 +6,19 @@
   "trace", "metrics" and "logs".
 - Removed dependency on `ordered-float`.
 
+- **Breaking** [1726](https://github.com/open-telemetry/opentelemetry-rust/pull/1726)
+  Update `LogProcessor::emit() method to take mutable reference to LogData. This is breaking
+  change for LogProcessor developers. If the processor needs to invoke the exporter 
+  asynchronously, it should clone the data to ensure it can be safely processed without 
+  lifetime issues. Any changes made to the log data in this method will be reflected in
```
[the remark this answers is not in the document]
lalitb: Good point. Have updated accordingly.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Yes, added test for modifying body too.

opentelemetry-otlp/src/exporter/tonic/logs.rs
lalitb: As of now, we are always cloning which can be avoided. Have updated the TODO to reflect that. Will address this separately.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -8,13 +8,14 @@ use opentelemetry::{
...
 };
+use std::borrow::Cow;
 use std::fmt::Debug;
 
 /// `LogExporter` defines the interface that log exporters should implement.
 #[async_trait]
 pub trait LogExporter: Send + Sync + Debug {
     /// Exports a batch of [`LogData`].
-    async fn export(&mut self, batch: Vec<LogData>) -> LogResult<()>;
+    async fn export<'a>(&mut self, batch: Vec<Cow<'a, LogData>>) -> LogResult<()>;
```
lalitb, line 18: TODO: To remove the dependency on constructs reflecting the implementation details e.g. `Vec` here.

cijothomas (approved):
Please update the  [stress test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs#L6) numbers with this change, so it is easy to track.

Also https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/benches/logs.rs#L13

lalitb:
> Please update the [stress test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs#L6) numbers with this change, so it is easy to track.
> 
> Also https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/benches/logs.rs#L13

Done, also updated the results in the description.

PR #1729, opened 2024-05-10
Use shared slices instead of references to owned vectors as return types

utpilla:
## Changes
- Use shared slices (a more abstract type) instead of references to owned vectors as the return types of `TracerProvider.span_processors()` and `LoggerProvider.log_processors()`
- Shared slices offer enough information for the callers of these methods and hides implementation details

### DIFF
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -62,7 +62,7 @@ impl TracerProvider {
     }
 
     /// Span processors associated with this provider
-    pub fn span_processors(&self) -> &Vec<Box<dyn SpanProcessor>> {
+    pub fn span_processors(&self) -> &[Box<dyn SpanProcessor>] {
         &self.inner.processors
     }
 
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -62,7 +62,7 @@ impl TracerProvider {
     }
 
     /// Span processors associated with this provider
-    pub fn span_processors(&self) -> &Vec<Box<dyn SpanProcessor>> {
+    pub fn span_processors(&self) -> &[Box<dyn SpanProcessor>] {
```
cijothomas, line 65:
```suggestion
    pub(crate) fn span_processors(&self) -> &[Box<dyn SpanProcessor>] {
```

Don't think this method should be public? Unrelated to this PR, but sharing here, as this is considered breaking given this method is public. We should audit all public API and trim everything unnecessary to avoid issues as we move towards stable.

cijothomas (approved):
Agree with the change. Please do update changelog with a breaking change warning (affecting plugin authors only), and as a follow up, lets trim all unnecessary public APIs to better handle this in future.

(Assuming this is indeed a breaking change due to `pub fn` in providers)

lalitb (approved):

opentelemetry-sdk/src/trace/provider.rs
TommyCpp: +1 let's make it `pub(crate)`

cijothomas: lets do the full audit and cleanup as a follow up to keep scope in control for this PR.

PR #1737, opened 2024-05-11
Refactor Metrics tests and add more

cijothomas:
Trying to leverage helper methods as much as possible. Not fully done, but moving in that direction.
Added UpDownCounter tests.

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -986,16 +986,10 @@ mod tests {
         key: &str,
         value: &str,
     ) -> Option<&'a DataPoint<T>> {
-        for datapoint in data_points {
-            if datapoint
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -848,13 +753,13 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn no_attr_delta_up_down_counter() {
```
TommyCpp, line 755: This test shares lots of duplication with the method below. Can we merge them into one as use table based unit testing here?

cijothomas: yes. not fixing too many in this PR, but will do shorter fixes.

TommyCpp (approved):

PR #1739, opened 2024-05-11
Modify metrics example to shutdown without global shutdown method

cijothomas:
To see if we need to keep global::shutdown() for metrics.


lalitb (approved):

TommyCpp (approved):
IIRC we never introduce the behavior that requires all loggers(meters for metrics) to be dropped before one can allow shutdown so it should work

Issue #1679, opened 2024-04-23
Revisit shutdown_signal_provider methods at API level


lalitb:
I think we need to revisit `shutdown_*_provider` methods as they are at the API level. The instrumented libraries shouldn't have the method to shutdown the LoggerProvider enabled by the application.

_Originally posted by @lalitb in https://github.com/open-telemetry/opentelemetry-rust/pull/1643#discussion_r1576473716_

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/1643#discussion_r1579840322 We may be able to this entirely at API level, for the logging signal. (still need it/similar capability for metrics,traces)

TommyCpp:
> We may be able to this entirely at API level

To clearify you mean remove shutdown, get logger, get logger provider, set logger provider entirely for logging signal?

cijothomas:
For v0.23 release:
This is not an issue for logs, as logs removed the global shutdown completely.
Not an issue for traces, as global shutdown existed even before this release.
For metrics, this is newly introduced in v0.23. An example on how to shutdown even without the global_shutdown : https://github.com/open-telemetry/opentelemetry-rust/pull/1739

PR #1743, opened 2024-05-11
Remove global shutdown from metrics and doc fixes

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1679


lalitb:
Just to summarise my understanding of the current behavior with this PR:
- The global shutdown is now removed for `metrics` and `logs`. And shutdown will happen when
     - explicitly invoked from any of the Sdk*Provider's, Or
     - All the references to the providers are dropped. 
- From `traces`, global shutdown will drop the global provider, and invoke shutdown (through drop) if there are no other provider references.

lalitb (approved):

PR #1738, opened 2024-05-11
Prepare v0.23 release

cijothomas:
Preparing the next release.
No change in overall [status](https://github.com/open-telemetry/opentelemetry-rust/tree/main?tab=readme-ov-file#project-status) and breaking changes are still expected in next release.

The next release, which is planned for June 30 2024, is expected to update status of [Logs](https://github.com/open-telemetry/opentelemetry-rust/issues/1733), [Metrics](https://github.com/open-telemetry/opentelemetry-rust/issues/1719) to Beta.

Opening as draft to discuss the below in next Community meeting on Tuesday May 14th 9 AM PT.
Blockers to discuss:
1. https://github.com/open-telemetry/opentelemetry-rust/issues/1679
2. In-flight or upcoming PRs that must be part of this release.

Issue #1146, opened 2023-07-10
Proposal to improve internal error logging


TommyCpp:
Presently, an internal logging mechanism is in place that leverages eprintln! to alert users of errors within the Opentelemetry pipeline. Users have the option to customize this logging by introducing their own error handler.

However, this existing approach exhibits a couple of issues:

- Absence of granularity in internal logs, making it challenging for users to isolate valuable information.
- Insufficient metadata, which makes error messages less informative for users.

To improve user-friendliness of our internal error handling, I am proposing the development of an enhanced internal logging tool with the following attributes:

- Minimal reliance on `opentelemetry`. Given that `opentelemetry` is typically used for diagnosing problems in applications and other libraries, it's important that we avoid using its log pillar for error reporting.
- Different levels for the internal logs. As telemetry should remain stable and never panic, the maximum error level should be capped at ERROR. Additionally, levels such as WARNING and DEBUG can be introduced.
- Inclusion of metadata to aid users in issue identification and resolution. Key metadata, in my view, includes:
   - Pillars (traces vs metrics vs logs)
   - Components, which could differ across various pillars. For instance, trace should contain `SpanProvider`, `SpanProcessor`, `SpanExporter`.
- To help user filter different level of messages. We should provide some kind of configuration to set the maximum log level. [log creates' compile time filters](https://docs.rs/log/latest/log/#compile-time-filters) is a good example and we can provide something similar

To implement these enhancements, I think we should convert the current `Error` enum into an `Error` structure that will include all the aforementioned metadata and level information. So something like 
```rust
pub struct Error {
    level: LogLevel,
    pillar: Pillar,
    component: Component,
    error: String,
}
```

shaun-cox:
It might be worth seeing if we can leverage the [tokio tracing](https://docs.rs/tracing/latest/tracing/index.html#using-the-macros) crate for this implementation.  It already defines concepts of log levels and components (which it calls targets) and has a rich facility for filtering which logs go where. I realize there might appear to be a circular dependency between their crate and `opentelemetry`, but I suspect it might be possible to avoid if we (and they) have api vs. sdk crates and our sdk only depends on their api and vice versa.

lalitb:
[...]

Issue #1042, opened 2023-04-25
SDK/Export concepts leaked into API.


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/common.rs#L426
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/logs/mod.rs#L27
The above are part of Public API of the OpenTelemetry-API create. Its about "Export" which is an SDK only concept. 
Wondering if this is intentional or accidental? If it is not intentional, we should move them into SDK befoe 1.0 stable.

(There are few other similar issues, will open separate issues to track them)

cijothomas:
See additional comment to be fixed while resolving this issue:https://github.com/open-telemetry/opentelemetry-rust/pull/1750#discussion_r1598799927

cijothomas:
@lalitb Can you address for logs while fixing https://github.com/open-telemetry/opentelemetry-rust/issues/2175 ? error handler is the reason for this issue, so would be good to close them together.

lalitb:
> @lalitb Can you address for logs while fixing https://github.com/open-telemetry/opentelemetry-rust/issues/2175 ? error handler is the reason for this issue, so would be good to close them together.

Yes.

PR #1750, opened 2024-05-13
Modify loggerprovider shutdown, flush to return single Result and handle repeat shutdown calls

cijothomas:
This is consistent with Metrics shutdown.
The overall result contains formatted version of combined errors, if any.
Added test to validate that repeated shutdown calls are handled.

### DIFF
--- a/opentelemetry-appender-tracing/examples/basic.rs
+++ b/opentelemetry-appender-tracing/examples/basic.rs
@@ -24,5 +24,5 @@ fn main() {
     tracing_subscriber::registry().with(layer).init();
 
     error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
-    drop(provider);
+    let _ = provider.shutdown();
 }
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -4,8 +4,8 @@ use crate::{
     runtime::RuntimeChannel,
 };
 use opentelemetry::{
-    global::{self},
-    logs::LogResult,
+    global,
+    logs::{LogError, LogResult},
     trace::TraceContextExt,
     Context, InstrumentationLibrary,
 };
@@ -13,7 +13,10 @@ use opentelemetry::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 
-use std::{borrow::Cow, sync::Arc};
+use std::{
+    borrow::Cow,
+    sync::{atomic::Ordering, Arc},
+};
 use std::{sync::atomic::AtomicBool, time::SystemTime};
 
 use once_cell::sync::Lazy;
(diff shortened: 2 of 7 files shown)


opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -97,25 +100,46 @@ impl LoggerProvider {
     }
 
     /// Force flush all remaining logs in log processors and return results.
-    pub fn force_flush(&self) -> Vec<LogResult<()>> {
-        self.log_processors()
-            .iter()
-            .map(|processor| processor.force_flush())
-            .collect()
+    pub fn force_flush(&self) -> LogResult<()> {
```
TommyCpp, line 103: TracerProvider returns a list of results from processors. If there a reason why we prefer single result over multiple results here? 

IMO it only helps simplify the use case where users want to fail when any processor fails. But it's relatively easy for users to do it on their own (Just `.iter().filter().is_empty()` should be enough)

However, it will make it harder for users that care about one processor over another to decide whether the force_flush is worth retrying. For example, let's say I have a simple processor that print to stdout and a batch processor to print to backend, with simple processor mostly used for debugging purpose. Then with this change I need to prase the error results to determine if the error comes from batch processor. 

I also don't think performance will improves just by chaning the return types but let me know if you think it will

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -97,25 +100,46 @@ impl LoggerProvider {
...
+                }
+            }
+
+            if errs.is_empty() {
+                Ok(())
+            } else {
+                Err(LogError::Other(format!("{errs:?}").into()))
+            }
+        } else {
+            Err(LogError::Other("logger provider already shut down".into()))
```
TommyCpp, line 141: We should make a dedicate LogError variant for repeat shutdown.

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: This is not done for any performance reasons, but simply to ensure consistency.
`MeterProvider` always returns single Result, and not a vec of Results from each of its readers. This PR makes `LoggerProvider` consistent with metrics.
(Neither is consistent with Traces, as Traces do have global shutdown, but others dont have that. However traces don't have shutdown on the provider itself.)

> However, it will make it harder for users that care about one processor over another to decide whether the force_flush is worth retrying. For example, let's say I have a simple processor that print to stdout and a batch processor to print to backend, with simple processor mostly used for debugging purpose. Then with this change I need to prase the error results to determine if the error comes from batch processor.

Agree this not very ideal.

I can revert changes to `force_flush` from this PR, as it probably need more discussion and alignment.

TommyCpp: Sounds good. Let's open an issue then?

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: Agree to fix it. Not in this PR, as the current structure of placing these errors in the API is not correct, and needs a refactoring. Will cover it as part of https://github.com/open-telemetry/opentelemetry-rust/issues/1042

TommyCpp: I'd mark this as releasing blocking and add a todo in code too. It will be very anti-intutive if the users has to parse the error string to understand if the errors is caused by repeated shutdown

opentelemetry-appender-tracing/examples/basic.rs
```diff
@@ -24,5 +24,5 @@ fn main() {
     tracing_subscriber::registry().with(layer).init();
 
     error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
-    drop(provider);
+    let _ = provider.shutdown();
```
cijothomas, line 27: I think we should be promoting shutdown() instead of drop, as shutdown() has a feedback mechanism, so users know at least something went wrong with shutdown (and possibly do something else to indicate lack of clean shutdown)

@lalitb @TommyCpp any thoughts on this?

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: > I'd mark this as releasing blocking

yes for stable or even beta release. But not for https://github.com/open-telemetry/opentelemetry-rust/pull/1738 which is still alpha for logs.

opentelemetry-appender-tracing/examples/basic.rs
lalitb: Agree on explicit shutdown.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Sorry if this is already answered. What is wrong with returning vec of LogResult for shutdown? Is it only for consistency with metrics?

cijothomas: yes. And also, if any individual sub-component faced issue while shutdown, they should be emitting their own logging. Shutdown() just need to capture the overall state. But I am not very sure we are doing it correctly. The proper solution should be coming once we address https://github.com/open-telemetry/opentelemetry-rust/issues/1146 (that also is very much related to https://github.com/open-telemetry/opentelemetry-rust/issues/761 as it is hard to solve internal logging without fixing the infinite telemetry/circular dep problem)

lalitb: Ok thanks, that makes sense.

opentelemetry-appender-tracing/examples/basic.rs
TommyCpp: Agree on this

lalitb (approved):

TommyCpp (approved):

PR #1752, opened 2024-05-13
OTLP Examples fixed to be consistent

cijothomas:
Fix the example to use same Resource for all signals.
The http one is now matching the grpc one, though I plan to send a follow up later to maintain single example with feature flag to control transport, as opposed to copying a lot.

Also, the http one is not working for logs, metrics. (this was same before this PR also, so not a blocker for this PR.)


cijothomas:
> Also, the http one is not working for logs, metrics. (this was same before this PR also, so not a blocker for this PR.)

https://github.com/open-telemetry/opentelemetry-rust/pull/1706 might have broken the endpoint.

TommyCpp:
> > Also, the http one is not working for logs, metrics. (this was same before this PR also, so not a blocker for this PR.)
> 
> #1706 might have broken the endpoint.

The metrics endpoint should work as after #1706 we won't change anything from user input in builders. For logs, we need to manually append the `/v1/logs` path to the base URL

cijothomas:
> > > Also, the http one is not working for logs, metrics. (this was same before this PR also, so not a blocker for this PR.)
> > 
> > 
> > #1706 might have broken the endpoint.
> 
> The metrics endpoint should work as after #1706 we won't change anything from user input in builders. For logs, we need to manually append the `/v1/logs` path to the base URL

Its still not working. :( Could you take a look ?

_"Error: Other("[Other(reqwest::Error { kind: Status(400), url: Url { scheme: \"http\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"localhost\")), port: Some(4318), path: \"/v1/logs\", query: None, fragment: None } })]")"_

lalitb (approved):
LGTM. Assuming the issue of OTLP HTTP Log Exporter would be handled separately.

TommyCpp:
> > > > Also, the http one is not working for logs, metrics. (this was same before this PR also, so not a blocker for this PR.)
> > > 
> > > 
> > > #1706 might have broken the endpoint.
> > 
> > 
> > The metrics endpoint should work as after #1706 we won't change anything from user input in builders. For logs, we need to manually append the `/v1/logs` path to the base URL
> 
> Its still not working. :( Could you take a look ?
> 
> _"Error: Other("[Other(reqwest::Error { kind: Status(400), url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(4318), path: "/v1/logs", query: None, fragment: None } })]")"_

Yeah taking a look now

lalitb:
> "Error: Other("[Other(reqwest::Error { kind: Status(400), url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(4318), path: "/v1/logs", query: None, fragment: None } })]")"

Bad Request, Could also be related to PR for sending Resource separately, and then combining before export. Let me check too :(

TommyCpp:
Alright I think there are some issues with OTLP http/json exporter. After set the http binary protocol everything works fine 
https://github.com/cijothomas/opentelemetry-rust/pull/1

PR #1755, opened 2024-05-14
feat: clean up trace SDK APIs

TommyCpp:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -27,6 +27,10 @@
  reference or owned`LogData`. If the exporter needs to process the log data
  asynchronously, it should clone the log data to ensure it can be safely processed without
  lifetime issues.
+- Clean up public methods in SDK.
+    - [`TracerProvider::span_processors`] and [`TracerProvider::config`] was removed as it's not part of the spec.
+    - Added `non_exhaustive` annotation to [`trace::Config`]. Marked [`config`] as deprecated since it's only a wrapper for `Config::default`
+    - Removed [`Tracer::tracer_provder`] and [`Tracer::instrument_libraries`] as it's not part of the spec.
 
 - **Breaking** [1836](https://github.com/open-telemetry/opentelemetry-rust/pull/1836) `SpanProcessor::shutdown` now takes an immutable reference to self. Any reference can call shutdown on the processor. After the first call to `shutdown` the processor will not process any new spans. 
 
@@ -61,13 +65,7 @@
   - After `shutdown`, `LogProcessor` will not process any new logs
 - Moving LogRecord implementation to the SDK. [1702](https://github.com/open-telemetry/opentelemetry-rust/pull/1702).
     - Relocated `LogRecord` struct to SDK, as an implementation for the trait in the API.
-- **Breaking** [#1729](https://github.com/open-telemetry/opentelemetry-rust/pull/1729)
-  - Update the return type of `TracerProvider.span_processors()` from `&Vec<Box<dyn SpanProcessor>>` to `&[Box<dyn SpanProcessor>]`.
-  - Update the return type of `LoggerProvider.log_processors()` from `&Vec<Box<dyn LogProcessor>>` to `&[Box<dyn LogProcessor>]`.
-- Update `opentelemetry` dependency version to 0.23
-- Update `opentelemetry-http` dependency version to 0.12
-- **Breaking** [#1750](https://github.com/open-telemetry/opentelemetry-rust/pull/1729)
-  - Update the return type of `LoggerProvider.shutdown()` from `Vec<LogResult<()>>` to `LogResult<()>`.
+
 
 ## v0.22.1
 
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -65,7 +65,13 @@
   - After `shutdown`, `LogProcessor` will not process any new logs
 - Moving LogRecord implementation to the SDK. [1702](https://github.com/open-telemetry/opentelemetry-rust/pull/1702).
     - Relocated `LogRecord` struct to SDK, as an implementation for the trait in the API.
-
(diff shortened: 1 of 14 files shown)


opentelemetry-sdk/CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
   - After `shutdown`, `LogProcessor` will not process any new logs
 - Moving LogRecord implementation to the SDK. [1702](https://github.com/open-telemetry/opentelemetry-rust/pull/1702).
     - Relocated `LogRecord` struct to SDK, as an implementation for the trait in the API.
+- Clean up public methods in SDK.
```
cijothomas, line 31: lets split this PR into smaller, focused PRs.

cijothomas: Specifically, the jaeger cleanup can be a separate one.

TommyCpp: > jaeger cleanup

No longer needed

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -10,12 +10,14 @@ use std::env;
 use std::str::FromStr;
 
 /// Default trace configuration
+#[deprecated(since = "0.23.0", note = "Use Config::default() instead")]
```
cijothomas, line 13: As shared in https://github.com/open-telemetry/opentelemetry-rust/issues/1810, and https://github.com/open-telemetry/opentelemetry-rust/pull/1788#issuecomment-2120868947 I am not seeing strong value in this `Config` struct abstraction, so this might change further depending on how that issue is going to be resolved.

Okay to proceed with this PR.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
  reference or owned`LogData`. If the exporter needs to process the log data
  asynchronously, it should clone the log data to ensure it can be safely processed without
  lifetime issues.
+- Clean up public methods in SDK.
+    - [`TracerProvider::span_processors`] and [`TracerProvider::config`] was removed as it's not part of the spec.
```
cijothomas, line 29: `TracerProvider::span_processors` -- is still public? Was that also intended to be pub(crate) ?

lalitb:
I think [this](https://github.com/open-telemetry/opentelemetry-rust/pull/1755/files#diff-ea02c6cde804746d1e52866ae343ef68302abf2e053092ddfabae1846b04bd6e) update to opentelemetry-proto submodule is not intended, as we are already using the latest version :)

opentelemetry-sdk/CHANGELOG.md
TommyCpp: Errr probably messed up when rebasing

opentelemetry-sdk/src/trace/config.rs
TommyCpp: Don't think it's conflicting even if we want to remove the Config. I think we can merge as it is and iterate on it

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -61,13 +65,7 @@
   - After `shutdown`, `LogProcessor` will not process any new logs
 - Moving LogRecord implementation to the SDK. [1702](https://github.com/open-telemetry/opentelemetry-rust/pull/1702).
     - Relocated `LogRecord` struct to SDK, as an implementation for the trait in the API.
-- **Breaking** [#1729](https://github.com/open-telemetry/opentelemetry-rust/pull/1729)
```
cijothomas, line 64: we cannot remove this as it is already part of shipped code!

lalitb (approved):

djc:
It seems a little unfortunate that this PR doesn't provide any kind of context or rationale in the description. This removes the `opentelemetry_sdk::trace::Tracer::provider()` method, which tracing-opentelemetry relies upon in its `PreSampledTracer`. I understand that this is a niche use case but as I understand it it's pretty important to tracing-opentelemetry. From what I can see there is no obvious alternative to this API -- how might we make `PreSampledTracer` work in the next release?

See the code here: https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/tracer.rs#L67. This ends up accessing `provider.config().id_generator.new_span_id()` and `provider.config().id_generator.new_trace_id()`.

TommyCpp:
Sorry for missing content in PR descriptions. 

As for way forward, we can either:

1. Re-open `provider()` method on tracer.
2. Or it seems the `tracing-opentelemetry` cares more about the configuration associated with the tracer provider. Maybe it's better if `Tracer` return a reference to current configuration. One benifit we get is we can bound the lifetime of the configuration so if the configuration won't change while the users are reading it(TracerProvider may update the config if needed, we don't support it today but may support it in the future?)

cijothomas:
The rationale is mentioned in the changelog: https://github.com/open-telemetry/opentelemetry-rust/pull/1755/files#diff-14dfb7552119b2eca23adb12f587f7afba4f354c8dce5bcd2ae0e8a6a258e932R30-R31
`PreSampledTracer` is not a concept in OpenTelemetry, so I can't give a workaround for that use case, unfortunately.

It was unintentional to break `tracing-opentelemetry`, sorry about that! At the same time, it is also undesirable to expose public APIs not warranted by the spec.

@TommyCpp I am okay to revert the provider() part to prevent breaking `tracing-opentelemetry` this time. I'll hold off tagging and publishing to crates.io to resolve this. Can you make a PR reverting provider() being marked crate only?

The only way I see such issues from occurring again is by resolving https://github.com/open-telemetry/opentelemetry-rust/issues/1571

NOBLES5E:
By the way, `TracerProvider::config()` also needs to be marked `public` to prevent breaking `tracing-opentelemetry`.

djc:
Alternatively if we want to provide the minimal capability via the Tracer API we could only export access to the id generator or even specifically new_trace_id() and new_span_id() methods.

djc:
Submitted https://github.com/open-telemetry/opentelemetry-rust/pull/1934.

Issue #1659, opened 2024-04-09
[Bug]: unclear how to setup http(hyper) without agent pipeline


ibigbug:
### What happened?

```
opentelemetry = "0.22"
opentelemetry_sdk = { version = "0.22", features = ["rt-tokio"] }
tracing-opentelemetry = "0.23"
opentelemetry-jaeger-propagator = "0.1.0"
opentelemetry-jaeger = { version = "0.21", features = ["collector_client", "hyper_collector_client", "rt-tokio"] }
opentelemetry-otlp = { version = "0.15.0", features = ["http-proto"] }
```

these are the enabled features

```
global::set_text_map_propagator(opentelemetry_jaeger_propagator::Propagator::new());

        let otlp_exporter = opentelemetry_otlp::new_exporter()
            .http()
            .with_endpoint(jager_endpoint);
        let tracer =
            opentelemetry_otlp::new_pipeline()
                .tracing()
                .with_exporter(otlp_exporter)
                .with_trace_config(trace::config().with_resource(Resource::new(vec![
                    KeyValue::new("service.name", "clash-rs"),
                ])))
                .install_batch(opentelemetry_sdk::runtime::Tokio)?;

```

these are the setup

problem is

```
failed to setup logging: Exporter otlp encountered the following error(s): no http client, you must select one from features or provide your own implementation
```

how do i active the default hyper client?

ty

### API Version

x

### SDK Version

x

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
Can you check this example, and see if it helps?

cijothomas:
This is not resolved, as shown here : https://github.com/open-telemetry/opentelemetry-rust/pull/1756#issuecomment-2109176774

ibigbug:
```
125 |         Some(tracing_opentelemetry::layer().with_tracer(tracer))
    |                                             ----------- ^^^^^^ the trait `PreSampledTracer` is not implemented for `opentelemetry_sdk::trace::Tracer`
    |                                             |
    |                                             required by a bound introduced by this call
    |
``` 

how to have `PreSampledTracer` implemented for a tracer below?

```
   let tracer = opentelemetry_otlp::new_pipeline()
            .tracing()
            .with_exporter(otlp_exporter)
            .with_trace_config(
                trace::config()
                    .with_sampler(Sampler::AlwaysOn)
[...]

PR #1756, opened 2024-05-14
Show working example for OTLP Http

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1659
Based on https://github.com/open-telemetry/opentelemetry-rust/pull/1752#issuecomment-2109116504, the example for otlp-http is modified to use HTTP/protobuf instead of json.
Also modified the example to use log appenger instead of tracing appender, as `tracing` causes https://github.com/open-telemetry/opentelemetry-rust/issues/761, and cannot be used without fixing.

### DIFF
--- a/opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
+++ b/opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
@@ -9,11 +9,10 @@ publish = false
 once_cell = { workspace = true }
 opentelemetry = { path = "../../../opentelemetry" }
 opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "metrics", "logs"] }
-opentelemetry-otlp = { path = "../..", features = ["http-proto", "http-json", "reqwest-client", "logs"] }
+opentelemetry-otlp = { path = "../..", features = ["http-proto", "reqwest-client", "logs"] }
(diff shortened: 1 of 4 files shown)


lalitb (approved):

TommyCpp:
I am not sure if the provided example will help users understand how to enable hyper clients. As it uses reqwest clients

cijothomas:
> I am not sure if the provided example will help users understand how to enable hyper clients. As it uses reqwest clients

Agree. Should we fix that in the doc itself, instead of maintaining full blown examples for each transport/clients?

ibigbug:
yeah i think a doc would be super helpfule. i'm struggling updating to the latest versions with a combination of 

```
# opentelemetry
opentelemetry = "0.23"
opentelemetry_sdk = { version = "0.23", features = ["rt-tokio"] }
tracing-opentelemetry = "0.23"
opentelemetry-jaeger-propagator = "0.2.0"
opentelemetry-jaeger = { version = "0.22", features = ["collector_client", "hyper_collector_client", "rt-tokio"] }
opentelemetry-otlp = { version = "0.16.0", features = ["http-proto"] }
```

also to be frank i feel the rust community library versions are a bit fragile, it sometimes require a fair bit of work to upgrade all eco system crates when 1 crate gets a breaking change

Issue #1559, opened 2024-02-18
Flaky test


cijothomas:
```
failures:

---- metrics::tests::second_delta_export_does_not_give_no_attr_value_if_add_not_called stdout ----
thread 'metrics::tests::second_delta_export_does_not_give_no_attr_value_if_add_not_called' panicked at opentelemetry-sdk/src/metrics/mod.rs:674:13:
No scope metrics in latest export
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    metrics::tests::second_delta_export_does_not_give_no_attr_value_if_add_not_called
```

cijothomas:
failures:

---- metrics::tests::second_delta_export_does_not_give_no_attr_value_if_add_not_called stdout ----
thread 'metrics::tests::second_delta_export_does_not_give_no_attr_value_if_add_not_called' panicked at opentelemetry-sdk/src/metrics/mod.rs:674:13:
No scope metrics in latest export
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    metrics::tests::second_delta_export_does_not_give_no_attr_value_if_add_not_called

test result: FAILED. 116 passed; 1 failed; 7 ignored; 0 measured; 0 filtered out; finished in 2.01s

error: test failed, to rerun pass `-p opentelemetry_sdk --lib`

PR #1766, opened 2024-05-15
Fix periodic reader to trigger first exporter at the interval

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1559

`PeriodicReader` makes the first export right away, instead of waiting for the interval. This is mostly unobservable/no-impact in real systems, but this is causing instability in the tests, as tests rely on force_flush() and expect results from a single export, not the result from multiple exports. With this change, the tests, when all [were run together](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/mod.rs#L82-L83) locally, passes 100% of the time. Before this PR, almost every time, there is at least 1-3 failures. Naturally, the CI flakiness should be gone as well.

Separately, we should be able to leverage `ManualReader` for the tests to get predictable behavior and avoid needing `tokio::test`. Will fix it later, as that requires larger refactoring.

(This issue is visible, if running the opentelemetry-otlp examples too - instead of seeing one metric, resulting from the drop triggered shutdown, one would have seen two metrics without this fix. Also added windows specific instructions to test this in OTLP+ Collector)

### DIFF
--- a/opentelemetry-otlp/examples/basic-otlp-http/README.md
+++ b/opentelemetry-otlp/examples/basic-otlp-http/README.md
@@ -36,14 +36,14 @@ On Unix based systems use:
 
 ```shell
 # From the current directory, run `opentelemetry-collector`
-docker run --rm -it -p 4317:4317 -v $(pwd):/cfg otel/opentelemetry-collector:latest --config=/cfg/otel-collector-config.yaml
+docker run --rm -it -p 4318:4318 -v $(pwd):/cfg otel/opentelemetry-collector:latest --config=/cfg/otel-collector-config.yaml
 ```
 
 On Windows use:
 
 ```shell
 # From the current directory, run `opentelemetry-collector`
-docker run --rm -it -p 4317:4317 -v "%cd%":/cfg otel/opentelemetry-collector:latest --config=/cfg/otel-collector-config.yaml
+docker run --rm -it -p 4318:4318 -v "%cd%":/cfg otel/opentelemetry-collector:latest --config=/cfg/otel-collector-config.yaml
 ```
 
 Run the app which exports logs, metrics and traces via OTLP to the collector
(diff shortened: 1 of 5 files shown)


TommyCpp:
> instead of seeing one metric, resulting from the drop triggered shutdown, one would have seen two metrics without this fix

If we skip one export before this change. Shouldn't we see one export vs two exports?

cijothomas:
> > instead of seeing one metric, resulting from the drop triggered shutdown, one would have seen two metrics without this fix
> 
> If we skip one export before this change. Shouldn't we see one export vs two exports?

Correct. We now only see one export. Without this, there would be two exports! (You can use Collector to verify, or the unit tests)

cijothomas:
To make it more clear:
Before the PR:
Exports were triggered at T, T+60, T+120 and so on, where T is the time meterprovider was built.

With this change:
Exports are triggered at T+60, T+120 and so on, where T is the time meterprovider was built.

TommyCpp (approved):
Nvm I get the change diff upside down. Good find!

lalitb (approved):

cijothomas:
If anyone is curious, this is how I have been testing:

```rust
use futures_util::{
    stream::{self, FusedStream},
    StreamExt,
};
use opentelemetry_sdk::runtime;
use opentelemetry_sdk::runtime::Runtime;
use std::error::Error;
use std::time::{Duration, SystemTime};

#[derive(Debug)]
enum MyMessage {
    Frequent,
    Rare,
}
struct Poller;

impl Poller {
    async fn run(mut messages: impl Unpin + FusedStream<Item = MyMessage>) {
        while let Some(message) = messages.next().await {
            println!(
                "Received message: {:?} time: {:?}",
                message,
                SystemTime::now()
            );
        }
    }
}

#[tokio::main]
    let rt = runtime::Tokio;
    let stream1 = rt
        .interval(tokio::time::Duration::from_secs(15))
        // .skip(1)
        .map(|_| MyMessage::Frequent);
    let stream2 = rt
        .interval(tokio::time::Duration::from_secs(1500))
        // .skip(1)
        .map(|_| MyMessage::Rare);

    println!(
        "Finished building stream1 and stream2 {:?}",
        SystemTime::now()
    );
    let messages = Box::pin(stream::select(stream1, stream2));
    rt.spawn(Box::pin(Poller::run(messages)));

    loop {
        tokio::time::sleep(Duration::from_secs(10)).await;
    }
}

```

PR #1767, opened 2024-05-15
Try to use authenticated request to install protoc in GH

cijothomas:
Error: Error: API rate limit exceeded for 13.105.117.234. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

Constantly hitting this limit, since we added windows and macOS to test matrix.. Attempting to fix...


lalitb (approved):

cijothomas:
Looks like this is working!

PR #1774, opened 2024-05-15
Add script to tag and publish to crates.io

cijothomas:
Storing what I used (partially) for the release.
This can be used if we are going to do manual release like today.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+packages=(
+    "opentelemetry"
+    "opentelemetry-http"
+    "opentelemetry-semantic-conventions"
+    "opentelemetry-jaeger-propagator"
+    "opentelemetry-sdk"    
+    "opentelemetry-proto"
+    "opentelemetry-otlp"
+    "opentelemetry-stdout"
+    "opentelemetry-jaeger"
+    "opentelemetry-zipkin"
+    "opentelemetry-prometheus"
+    "opentelemetry-appender-log"
+    "opentelemetry-appender-tracing"
+
+    # Add more packages as needed, in the right order
+)
+
+# Set the current directory to one level above the scripts directory
+current_dir=$(pwd)/..
+cd "$current_dir"  # Change to the current directory
+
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -0,0 +1,63 @@
...
+    "opentelemetry-proto"
+    "opentelemetry-otlp"
+    "opentelemetry-stdout"
+    "opentelemetry-jaeger"
+    "opentelemetry-zipkin"
+    "opentelemetry-prometheus"
+    "opentelemetry-appender-log"
+    "opentelemetry-appender-tracing"
+
+    # Add more packages as needed, in the right order
```
TommyCpp, line 18:
```suggestion
    # Add more packages as needed, in the right order. A package should only be published after all it's dependencies have been published
```

lalitb:
Should we update the [RELEASING.md](https://github.com/open-telemetry/opentelemetry-rust/blob/main/RELEASING.md) about (optionally) using this script - with and without dry run?

scripts/publish.sh
```diff
@@ -0,0 +1,63 @@
...
+
+        if [[ -n "$name" && -n "$version" ]]; then
+            echo "Found package $name with version $version" in cargo.toml
+            
+            # Tag the version in git
+            tag="${name}-${version}"
+            tag_message="${name} ${version} release."
+            # uncomment the following lines after verifying all looks good.
+            # git tag -a "$tag" -m "\"$tag_message\""
+            # git push origin "$tag"
```
lalitb, line 44: Script can take a mandatory parameter for dry-run and publish

```bash
$ publish --dryrun # will do dry-run (or make it default)
$ publish --publish # will do publish
```

This way, no need to uncomment lines for publishing :)

cijothomas: dryrun will fail anyway as dependencies are not going to be there. The commented out is just to prevent accidentally running and causing issues!

Just wanted to have something simple, as long term, we could use something like https://github.com/MarcoIeni/release-plz

lalitb (approved):

PR #1776, opened 2024-05-15
Try rename opentelemetry_sdk to opentelemetry-sdk

cijothomas:
Fixes #

## Changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ The example above requires the following packages:
 # Cargo.toml
 [dependencies]
 opentelemetry = "0.22"
-opentelemetry_sdk = "0.22"
+opentelemetry-sdk = "0.22"
 opentelemetry-stdout = { version = "0.3", features = ["trace"] }
 ```
 
--- a/examples/logs-basic/Cargo.toml
+++ b/examples/logs-basic/Cargo.toml
@@ -7,7 +7,7 @@ publish = false
 
 [dependencies]
 opentelemetry = { path = "../../opentelemetry", features = ["logs"] }
-opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["logs"] }
+opentelemetry-sdk = { path = "../../opentelemetry-sdk", features = ["logs"] }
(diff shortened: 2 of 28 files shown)


shaun-cox:
This might have ramification on how or if this can be pushed to crates.io, I suspect.  Should we consult with someone with more knowledge of the cargo publishing process to understand the consequences?

cijothomas:
> This might have ramification on how or if this can be pushed to crates.io, I suspect. Should we consult with someone with more knowledge of the cargo publishing process to understand the consequences?

yes, I intend to do some research before attempting! Thanks.

hdost:
We could consider posting a notice to https://rustsec.org/contributing.html about `opentelemetry_sdk` -> `opentelemetry-sdk`
Perhaps also we could try to write a wrapper for at least one version which just has `opentelemetry_sdk` wrapping `opentelemetry-sdk` directly.

demurgos:
Is there some related issue with the discussion about the rename? The naming consistency should be fixed indeed, but I'd like to read about why you picked to normalize using `-` (compared to normalizing the other crates to `_`). The reason why I'm asking is that both styles are present on `crates.io`. `-` feels a bit more popular, but many larger crates also use `_` since it means that the crate name and import name match.

cijothomas:
> Is there some related issue with the discussion about the rename? The naming consistency should be fixed indeed, but I'd like to read about why you picked to normalize using `-` (compared to normalizing the other crates to `_`). The reason why I'm asking is that both styles are present on `crates.io`. `-` feels a bit more popular, but many larger crates also use `_` since it means that the crate name and import name match.

There is no other discussion in this repo about this topic. Since opentelemetry_sdk was the outlier, I tried to make it change to use "-", instead of trying every other crate switch to "_".


> but many larger crates also use `_` since it means that the crate name and import name match.

Yes I noticed that too!

cijothomas:
Closing now, keeping a new issue to track this: https://github.com/open-telemetry/opentelemetry-rust/issues/2826

PR #1777, opened 2024-05-15
Exclude stdout exporter from codecoverage

cijothomas:


lalitb:
Why are we excluding it ? It's part of the specs, and should be well tested, irrespective of whether it's used for debugging for production.

cijothomas:
> Why are we excluding it ? It's part of the specs, and should be well tested, irrespective of whether it's used for debugging for production.

It should be well test, but manually? Are you suggesting we should setup tests by reading from stdout?
(Don't think its worth it)

lalitb:
> Are you suggesting we should setup tests by reading from stdout?

No, I thought we should have unit-tests for specific components e.g., [transform](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-stdout/src/logs/transform.rs), and probably some mock test for stdout. This can't be a priority yet, as there are other things to stabilize, but good to visit sometime. Should we fine to remove for now, if we all agree.

cijothomas:
> > Are you suggesting we should setup tests by reading from stdout?
> 
> No, I thought we should have unit-tests for specific components e.g., [transform](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-stdout/src/logs/transform.rs), and probably some mock test for stdout. This can't be a priority yet, as there are other things to stabilize, but good to visit sometime. Should we fine to remove for now, if we all agree.

Agree. Just reminded that we did have bugs like https://github.com/open-telemetry/opentelemetry-rust/pull/1747 before!

PR #1778, opened 2024-05-15
Main readme to have runnable example

cijothomas:
I am not sure if we should maintain a lengthy code in the [main readme](https://github.com/open-telemetry/opentelemetry-rust?tab=readme-ov-file#getting-started).. If we chose to, we need to make sure it is showing all signals, and is a runnable example.

(Personally, I am okay to remove this from main readme file, but open to suggestions)


README.md
```diff
@@ -89,9 +160,14 @@ The example above requires the following packages:
 ```toml
 # Cargo.toml
 [dependencies]
-opentelemetry = "0.22"
-opentelemetry_sdk = "0.22"
-opentelemetry-stdout = { version = "0.3", features = ["trace"] }
+opentelemetry = { version="0.23", features = ["trace", "logs", "metrics"]}
+opentelemetry_sdk = { version="0.23", features = ["rt-tokio", "metrics"]}
```
TommyCpp, line 164:
```suggestion
opentelemetry = { version = "0.23", features = ["trace", "logs", "metrics"]}
opentelemetry_sdk = { version = "0.23", features = ["rt-tokio", "metrics"]}
```
For consistency

lalitb:
The example could be bit overwhelming for newcomers due to its length. Will it make sense to split it in 3 relatively smaller examples for logs, metrics and traces ?

TommyCpp:
> The example could be bit overwhelming for newcomers due to its length. Will it make sense to split it in 3 relatively smaller examples for logs, metrics and traces ?

Probably can link to smaller examples? I feel like if we split it up we will just repeat a lot of setups

cijothomas:
> The example could be bit overwhelming for newcomers due to its length. Will it make sense to split it in 3 relatively smaller examples for logs, metrics and traces ?

I am not sure what would be a good way... Keeping 3 examples will make the main readme even longer..
The opentelemetry-stdout has single example, but split by signal feature... 

Maybe we should steal something like https://github.com/tokio-rs/axum/tree/main/examples/readme ? We can keep an example in the examples directory, named readme, which can be referred from the readme.md file. It can ensure main readme.md file is still small, and example is always kept up-to-date...

cijothomas:
Marking to draft, as I need to decide if we can do any better than this.. Please keep sharing the feedback.

TommyCpp:
> and example is always kept up-to-date

I don't think that's a small task given we are improving APIs on a daily basis. I'd rather we get a stable place first and then add more examples.

lalitb:
> I feel like if we split it up we will just repeat a lot of setups

We can a common `toml` for all 3 examples. Apart from that setup would be different in all three.

TommyCpp:
> > > and example is always kept up-to-date
> > 
> > 
> > I don't think that's a small task given we are improving APIs on a daily basis. I'd rather we get a stable place first and then add more examples. We cut down lots of examples earlier to keep the maintaince burdan low
> 
> We are already keeping example in stdout/otlp up-to-date! (it is also a good way for PR reviewers to see the impact of breaking changes, as we ourselves show the changes required in the example. Given most changelogs link to PR#, it also helps end users to know what changes they must do in their own project).
> 
> Not very sustainable path, but once we announce stable release, this should be resolved. Until then a lot of users will be 😞.

Just to be clear I am not against the idea of breaking changes on APIs now or adding more examples. My worries are given we haven't stable yet, adding too many examples will add unnecessary burdan on us as maintainers, as well as contributors.

cijothomas:
Closing old, inactive PRs. Feel free to reopen when ready. Thanks.

Issue #1761, opened 2024-05-14
Remove ordered_float dependency


cijothomas:
The SDK currently has a dependency on [ordered_float](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/Cargo.toml#L21). It is used to provide Hash, PartialEq implementations, for `F64` values. While the crate is very popular, it is still a 3rd party dependency, [maintained](https://github.com/reem/rust-ordered-float) under a personal account.
Proposing to remove the dependency, and make `Hash`/`PartialEq` implementation in the `opentelemetry` crate itself, where `Value` is defined. 

The "correctness" part is not very important - i.e it is undefined if the following would get aggregated into a single metric point or not.
```rust
counter.Add(10, &[KeyValue::new("key", F64.Nan)]
counter.Add(10, &[KeyValue::new("key", F64.Nan)]
```

As a bonus, we can avoid the [`HashKeyValue`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/attributes/set.rs#L12) wrapper around `KeyValue` used in Metrics, which solely exist today to provide hash/partialeq implementations.

cijothomas:
Closed via https://github.com/open-telemetry/opentelemetry-rust/pull/1806

PR #1780, opened 2024-05-16
Implement Hasheable Float for use in metrics

cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-rust/issues/1761
This adds Hash implementation for KeyValue, by providing a wrapper around F64. This is only planned to be used in Metrics aggregation.

HashKeyValue is replaces with KeyValue in sdk.

Alternatively, we could just state that float values in attributes for Metrics are not supported. It was never tested before either!

No perf regression, but slight gains, probably due to one less indirection?.

```txt
     Running benches/attribute_set.rs (/home/cijothomas/repos/opentelemetry-rust/target/release/deps/attribute_set-c5774977e4c1461a)
Gnuplot not found, using plotters backend
AttributeSet_without_duplicates
                        time:   [399.63 ns 400.22 ns 401.05 ns]
                        change: [-0.7145% -0.4216% -0.1156%] (p = 0.01 < 0.05)
  1 (1.00%) low mild
  2 (2.00%) high mild
  6 (6.00%) high severe

AttributeSet_with_duplicates
                        time:   [326.62 ns 326.88 ns 327.18 ns]
                        change: [-1.9789% -1.4644% -1.0568%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe
```


opentelemetry/src/common.rs
```diff
@@ -422,6 +424,55 @@ impl KeyValue {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+struct F64Hasheable(f64);
```
utpilla, line 428:
```suggestion
struct F64Hashable(f64);
```

opentelemetry/src/common.rs
```diff
@@ -594,3 +645,74 @@ impl InstrumentationLibraryBuilder {
...
+
+        let kv1 = KeyValue::new("key", 1.0);
+        let kv2 = KeyValue::new("key", 1.01);
+        assert_ne!(kv1, kv2);
+
+        let kv1 = KeyValue::new("key", std::f64::NAN);
+        let kv2 = KeyValue::new("key", std::f64::NAN);
+        assert_ne!(kv1, kv2, "NAN is not equal to itself");
+
+        let kv1 = KeyValue::new("key", std::f64::INFINITY);
```
utpilla, line 671: Add a test for negative infinity as well.

opentelemetry/src/common.rs
```diff
@@ -594,3 +645,74 @@ impl InstrumentationLibraryBuilder {
...
+        }
+    }
+
+    #[test]
+    fn hash_kv_float() {
+        let kv1 = KeyValue::new("key", 1.0);
+        let kv2 = KeyValue::new("key", 1.0);
+        assert_eq!(hash_helper(&kv1), hash_helper(&kv2));
+
+        let kv1 = KeyValue::new("key", 1.001);
```
utpilla, line 691: Add a test case to check the hash values for two different values is also different.

utpilla (approved):

opentelemetry/src/common.rs
```diff
@@ -594,3 +645,106 @@ impl InstrumentationLibraryBuilder {
...
+            let random_value = rng.gen::<f64>();
+            let kv1 = KeyValue::new("key", random_value);
+            let kv2 = KeyValue::new("key", random_value);
+            assert_eq!(hash_helper(&kv1), hash_helper(&kv2));
+        }
+    }
+
+    #[test]
+    fn hash_kv_order() {
+        let kv1 = KeyValue::new("a", 1.0);
```
utpilla, line 723: You could remove this as it's already being tested in the `for` loop.

opentelemetry/src/common.rs
```diff
@@ -422,6 +424,55 @@ impl KeyValue {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for KeyValue {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.key.cmp(&other.key)
+    }
+}
```
lalitb, line 472: I have some concerns about adding this behavior to the public API. The `KeyValue` struct is a generic data structure in our API, and its behavior should be consistent with how Rust handles (say) tuples. Specifically:

- Two `KeyValue` instances should be considered equal if both their key and value fields are equal.
- Both key and value should be considered for ordering comparisons.

The current proposal adds ordering functionality based solely on the `key` field. While this might suit our specific use case, such as using KeyValue in HashMap or HashSet where only keys are compared, it seems different from the intuitive behavior.
 
Maybe these are only my thoughts :)

lalitb: Can we add a (say) `KeyOnly` wrapper on top of `KeyValue` to enable using `KeyValue` in scenarios that need key-only ordering and comparison? I guess that can remain in the public API.

cijothomas: thanks for raising this! It is a valid concern.
I think the doc for KeyValue can explain its behavior. It is OpenTelemetry Key,Value, not something else, so I think we can afford to do OpenTelemetry specific things.

`Ord` is currently only used for Metrics, but they could be used for other signals too, to achieve de-duplication.(deduping a sorted list can be done without hashset allocation). In those cases also, `Ord` would be based on Key only.

In other words, KeyValue is not required to match regular (key,value) tuples, and we need to document the intend in the doc. Otherwise, we could have just used (key,value) tuple itself, instead of implemented our own wrapper `KeyValue`.

lalitb: We also provide `KeyValue` attributes as argument in our Sampling API, and user may want to use some filtering logic in custom log-processors based on the attributes, and these sampling/filtering decision could require the default  `Ord` logic. 
But I don't have strong opinion here, if we can clearly document the behavior as part of this PR, we should be good.

cijothomas: > could require the default Ord logic

That part, I am not sure! The fact that we are using a custom struct, and not (key,value) tuple, is to convey that it is *not* a normal tuple, and it's behavior is diff. from normal tuples. If a user intend to leverage a custom struct's trait implementations, I expect they read the doc for the same.

(I am yet to add docs comments in this PR, will add and update).

cijothomas:
> Sorry haven't get a chance to look at it. Will do it today

Moved to draft for now, as other changes are likely conflicting with this PR. and I'd prefer to finish those first and comeback to this. Also, I modified this PR so that the hash/eq implementations are restricted to "metrics" feature flag only in api.

cijothomas:
@utpilla and I discussed another alternative option - for Metrics, if the Attribute's value is `f64`, convert to string (`to_string()`) and use string's built-in hash/eq capabilities. And add a doc warning saying, Attributes with values of type `f64` will be stringified for metric aggregation.

@lalitb @TommyCpp any objections for doing this?

utpilla:
> Creating a `String` requires heap allocation and can be expensive.

That should be a one-time allocation (only when storing the key-value pair on the Hashmap), right?

> Also using string representation also means `(key, 1.00)`'s hash will be differnt from `(key, 1.0)`'s hash, which is counterintuitive IMO

Is it even possible to have a float value stringified to "1.0" or "1.00"? Shouldn't it be just converted to "1"?

```rust
let num1 :f64 = 1.0;
let num2 :f64 = 1.00;
    
assert_eq!("1", num1.to_string());
assert_eq!("1", num2.to_string());
```

lalitb:
> @lalitb @TommyCpp any objections for doing this?

why not to directly do bitwise hashing. Apart from performance, it would be good to test that there is no precision loss in conversion from the float to string. `to_string()` uses the default precision of `Display` trait, and can't be changed.

TommyCpp:
> Is it even possible to have a float value stringified to "1.0" or "1.00"? Shouldn't it be just converted to "1"?
> 
> ```rust
> let num1: f64 = 1.0;
> let num2: f64 = 1.00;
>     
> assert_eq!("1", num1.to_string());
> assert_eq!("1", num2.to_string());
> ```

Hmm yeah you are right

TommyCpp:
> > @lalitb @TommyCpp any objections for doing this?
> 
> why not to directly do bitwise hashing. Apart from performance, it would be good to test that there is no precision loss in conversion from the float to string. `to_string()` uses the default precision of `Display` implementation, and can't be changed.

+1 It's not clear to me what benifit the stringify brings compare with bitwise hashing?

cijothomas:
> > > @lalitb @TommyCpp any objections for doing this?
> > 
> > 
> > why not to directly do bitwise hashing. Apart from performance, it would be good to test that there is no precision loss in conversion from the float to string. `to_string()` uses the default precision of `Display` implementation, and can't be changed.
> 
> +1 It's not clear to me what benifit the stringify brings compare with bitwise hashing?

🤣 Now I am not sure. I was trying to avoid having to own the hashing logic ourselves, and steer users away from using float for dimension values in the first place! (No metric backend that I am aware of support float dimension values, it'll be anyway stringified at exporter/ingestion level)

I'll stick the bitwise hashing for now, and also include a note/warning for users to avoid f64 values for Metric KeyValues.

opentelemetry/src/common.rs
lalitb: I noticed this hashing and ordering is moved (back) to `metrics` directory, which means it would be available only if the metrics feature is enabled (though we enable it by default now) ? Should be fine, just confirming.

cijothomas: true. I'll prepare a new PR to avoid dealing with conflict.

cijothomas: >  moved (back) to metrics directory

But now in opentelemetry, not in sdk.

utpilla:
> Good point you are right. Display implementation will do some conversion before stringify it

@TommyCpp I know we are not considering this anymore but just for my own learning, I think it probably doesn't have to do stringification. I believe the in-memory representation of the float values `1.0` and `1.00` is the same:

```rust

let num1: f64 = 1.0;
let num2: f64 = 1.00;
    
println!("{}", num1.to_bits()); // 4607182418800017408
println!("{}", num2.to_bits()); // 4607182418800017408
    
assert_eq!(num1.to_bits(), num2.to_bits());
```

cijothomas:
Closing in favor of https://github.com/open-telemetry/opentelemetry-rust/pull/1806

TommyCpp:
> > Good point you are right. Display implementation will do some conversion before stringify it
> 
> @TommyCpp I know we are not considering this anymore but just for my own learning, I think it probably doesn't have to do stringification. I believe the in-memory representation of the float values `1.0` and `1.00` is the same:
> 
> ```rust
> let num1: f64 = 1.0;
> let num2: f64 = 1.00;
>     
> println!("{}", num1.to_bits()); // 4607182418800017408
> println!("{}", num2.to_bits()); // 4607182418800017408
>     
> assert_eq!(num1.to_bits(), num2.to_bits());
> ```

Yep they should have the same underlying representation right after compiling from literal value to f64

Issue #1709, opened 2024-05-06
[Metrics API] - Add metrics to default features


cijothomas:
[Default ](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/Cargo.toml#L34) is only Trace today. This issue is to track adding Metrics to the default.

(Whether to add Logs or not is separately tracked)

PR #1783, opened 2024-05-19
Add logs and metrics to default features in api, sdk and otlp crate

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1709
All 3 signals are planned to reach beta in the next release. This PR is changing the default for api,sdk,otlp crates to include all three signals by default. The separate feature flags are still maintained, so if anyone wants to only pick only, it is still possible.

### DIFF
--- a/examples/logs-basic/README.md
+++ b/examples/logs-basic/README.md
@@ -1,16 +1,17 @@
-# Log Appender for API -  Example
+# OpenTelemetry Log Appender for log -  Example
 
 This example shows how to use the opentelemetry-appender-log crate, which is a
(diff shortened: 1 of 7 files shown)


examples/logs-basic/README.md
```diff
@@ -1,16 +1,17 @@
-# Log Appender for API -  Example
+# OpenTelemetry Log Appender for log -  Example
```
cijothomas, line 1: sorry, unrelated changes was sitting in my local! Hope this is okay.

lalitb (approved):
Thanks. Good to check if we need to add this for opentelemetry-proto too.

cijothomas:
> Thanks. Good to check if we need to add this for opentelemetry-proto too.

I left that as it is not clear why would end users need that directly...

PR #1785, opened 2024-05-19
Minor cleanups of readme files

cijothomas:
Towards #1306 
1. Consistent readme for more crates.
2. Moved releasing.md file to under scrips directory. This is only intended for maintainers, and this will keep the top level leaner.
3. Fix some jaeger cleanup leftovers.
4. Nit doc fixes.
(not fully sure of the current content of readme.md files. Will revisit once more before next release to ensure consistency)


opentelemetry-http/README.md
```diff
@@ -1,37 +1,30 @@
...
 ![OpenTelemetry — An observability framework for cloud-native software.][splash]
 
 [splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
 
-# OpenTelemetry HTTP
-
-The Rust [OpenTelemetry](https://opentelemetry.io/) implementation.
+This crate contains helper implementations for sending HTTP requests. Uses
+include propagating and extracting context over http, exporting telemetry,
+requesting sampling strategies.
```
lalitb, line 9: Something like - 

This crate facilitates HTTP header propagation and provides a trait-based HTTP client abstraction for telemetry export using OpenTelemetry. It includes utilities for injecting and extracting headers and supports various HTTP clients such as `reqwest`, `isahc`, and `hyper`.

TommyCpp: > provides a trait-based HTTP client abstraction for telemetry export using OpenTelemetry

Honestly we probably should note this won't be needed after http 1.0 upgrade

scripts/RELEASING.md
```diff
@@ -42,7 +42,7 @@ A draft PR can be created, but before releasing consider the following:
 5. [Publish](#publishing-crates) to crates.io using the version as of the release commit
 6. Post to [#otel-rust](https://cloud-native.slack.com/archives/C03GDP0H023) on CNCF Slack.
 
-[Publish.sh](./scripts/publish.sh) may be used to automate steps 3 and 5.
+[Publish.sh](./publish.sh) may be used to automate steps 3 and 5.
```
lalitb, line 45: why is RELEASING.md moved under scripts? It is not really a script, and root level actually seemed fine.

opentelemetry-http/README.md
lalitb: yes good point.

scripts/RELEASING.md
cijothomas: just looking for a way to hide it from top level directory, to keep the top level directory leaner. (this is not meant for anyone but maintainers to see)
I can move it anywhere else, if there are good alternatives!

lalitb: I don't see any other place to hide it. But I see most of the other languages otel-go, `otel-cpp, otel-java, otel-python have it a top level. Though not that a big issue either, should be fine :)

lalitb (approved):

scripts/RELEASING.md
cijothomas: yes. it is immaterial where it is kept. I just prefer to keep top-level leaner and reserve it for things that must absolutely in the root like the toml file etc!

PR #1789, opened 2024-05-20
Use warning in rustdocs for stdout

cijothomas:
Trying to leverage ["Warning"](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block) in our docs, since https://github.com/rust-lang/rust/pull/106561/files is merged!

Rendered output in local:

image


TommyCpp (approved):
Nice! We should use it more

Issue #1719, opened 2024-05-07
Metrics beta/rc/stable release timelines


cijothomas:
Extracting from [parent issue](https://github.com/open-telemetry/opentelemetry-rust/issues/1678) to own issue to gather explicit feedbacks, as details are hard to notice, when embedded inside the parent issue, especially the breaking changes.

### Background: 
Re-emphasizing the core issues from https://github.com/open-telemetry/opentelemetry-rust/issues/1572
1. Performance is not acceptable. (heavy contentions)
2. Lack of test coverage to prove correctness, stability
3. Shutdown/Cleanup issues
4. No ability to run Metrics SDK without an async runtime
5. Hard to troubleshoot

## Release plan for Metrics
Components in scope are:
Metrics API (`opentelemetry` crate)
Metrics SDK (`opentelemetry-sdk` crate)
Metrics InMemory Exporter (`opentelemetry-sdk` crate)
Metrics OTLP Exporter (`opentelemetry-otlp` crate)
Metrics stdout Exporter (`opentelemetry-stdout` crate)

Following sub-sections provides details on items we plan for each release, along with rough timelines. 

### Beta release - ETA June 30 2024
Breaking changes are still possible in beta, but has a higher bar than alpha releases.

The following functionality is explicitly being removed:
1. Metrics API's ability to add/remove Callbacks *after* Observable Instruments has been created. There will still be ability to provide callback(s) at the time of Instrument Creation. The current API to add/remove callbacks after instrument creation needs to be revisited to ensure functionality/usefulness/correctness. Removing it from initial release to keep the scope manageable.

Link to discussion around this, as there are concerns: https://github.com/open-telemetry/opentelemetry-rust/pull/1715/files#r1592783145

2. Metrics SDK's View feature. Would need to gather feedback about the common use cases where Views are needed. My own experience is that, it is used most frequently to provide custom Histogram bounds. Before removing View feature, the ability to provide custom bounds will be provided at the API level. The main reason why it is removed from initial release is to keep scope manageable. Views bring a whole lot of new complexities and additional test requirements.
4. Prometheus/Pull Exporter. 

### RC release - ETA Nov 30 2024
While breaking changes are still possible, it has extremely high bar here.

### Stable release - ETA Jan 30 2024
No breaking changes allowed after. 

Post the 1st stable release, the following features will be added (listed *not* in priority order)
1. Add support for multiple readers
2. Add support for Views
3. Add support for Exemplars
[...]

PR #1792, opened 2024-05-20
[Metric SDK] - Avoid exposing AttributeSet to exporters - Part1

cijothomas:
`AttributeSet` is how attributes are stored internally for Metric Aggregation, and is also exposed to exporters. This PR exposes only a `Vec<KeyValue>` to exporters (done only for SUM, HISTOGRAM to keep PR's small)

In the next set of PRs, this will be extended to all Aggregations, and then `AttributeSet` can be completely kept pub(crate) instead of exposing it. Also, attributes can be further hidden by just exposing an Iterator only, instead of even a Vec!.

This affects some export authors. Not adding changelog in this PR, as I plan to get it covered in one shot when above are achieved.

Also removed `opentelemetry-prometheus` from CI for now, as it is not planned for the upcoming release : https://github.com/open-telemetry/opentelemetry-rust/issues/1719

### DIFF
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -97,7 +97,7 @@ impl<T: fmt::Debug + Send + Sync + 'static> Aggregation for Sum<T> {
 pub struct DataPoint<T> {
     /// Attributes is the set of key value pairs that uniquely identify the
     /// time series.
-    pub attributes: AttributeSet,
+    pub attributes: Vec<KeyValue>,
     /// The time when the time series was started.
     pub start_time: Option<SystemTime>,
     /// The time when the time series was recorded.
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -143,7 +143,7 @@ impl<T: fmt::Debug + Send + Sync + 'static> Aggregation for Histogram<T> {
 #[derive(Debug)]
 pub struct HistogramDataPoint<T> {
     /// The set of key value pairs that uniquely identify the time series.
-    pub attributes: AttributeSet,
+    pub attributes: Vec<KeyValue>,
     /// The time when the time series was started.
     pub start_time: SystemTime,
     /// The time when the time series was recorded.
--- a/opentelemetry-sdk/src/metrics/internal/sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/sum.rs
@@ -1,4 +1,5 @@
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::vec;
 use std::{
     collections::{hash_map::Entry, HashMap},
     sync::Mutex,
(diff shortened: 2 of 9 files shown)


opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -141,7 +143,10 @@ impl<T: Number<T>> Sum<T> {
 
         for (attrs, value) in values.drain() {
             s_data.data_points.push(DataPoint {
-                attributes: attrs,
+                attributes: attrs
+                    .iter()
+                    .map(|(k, v)| KeyValue::new(k.clone(), v.clone()))
+                    .collect(),
```
cijothomas, line 149: not worried about the extra cost here, as this is in collect() path only, to focus on fixing the hot path first. The collect() path can be refactored afterwards.

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -97,7 +97,7 @@ impl<T: fmt::Debug + Send + Sync + 'static> Aggregation for Sum<T> {
 pub struct DataPoint<T> {
     /// Attributes is the set of key value pairs that uniquely identify the
     /// time series.
-    pub attributes: AttributeSet,
+    pub attributes: Vec<KeyValue>,
```
utpilla, line 100: Could we use a boxed slice instead of a Vec? It would save 8 bytes (`usize` length) per instance since unlike we wouldn't need an additional pointer to track the capacity. That could save a considerable amount of memory if we are going to have a lot of these `DataPoints`.

```suggestion
    pub attributes: Box<[KeyValue]>,
```

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: We can implement `From` trait, and use it throughout?

lalitb (approved):
Nit comment. Also, Collect cycle would need to be revisited if we need to clone to convert from AttributeSet. Better to add a TODO there?

TommyCpp (commented):
What's the reason of disabling the Prometheus exporter? I'd rather not have component that cannot compile in the repo. Happy to help fix any compiling issues

cijothomas:
> What's the reason of disabling the Prometheus exporter? I'd rather not have component that cannot compile in the repo. Happy to help fix any compiling issues

What other option we have ?

From desc:

> Also removed opentelemetry-prometheus from CI for now, as it is not planned for the upcoming release : https://github.com/open-telemetry/opentelemetry-rust/issues/1719

opentelemetry-sdk/src/metrics/data/mod.rs
cijothomas: As mentioned in the PR desc, will revisit the public APIs. This PR is to unblock the unwanted exposure of `AttributeSet` outside of core sdk.

cijothomas:
We absolutely need it in main repo, as it is a spec-ed out component. (though experimental).
I can remove the code from repo completely and bring it back post 1.0, or keep it as-is, and remove from CI, re-enable CI post 1.0.

Happy to discuss other ideas! Maybe good topic for the community meeting tommorow.

PR #1794, opened 2024-05-21
[Metric SDK] - Avoid exposing AttributeSet to exporters - Part2

cijothomas:
Continuing https://github.com/open-telemetry/opentelemetry-rust/pull/1792
This applies changes to ExponentialHistograms
With this change `AttributeSet` is only required within the SDK only. (It is still marked public, expecting to clear that along with the benchmarks that use it next)

With this, I think I can continue with https://github.com/open-telemetry/opentelemetry-rust/pull/1780#issuecomment-2121379009


cijothomas:
image
Semver check works nicely!

TommyCpp:
> image Semver check works nicely!

Are we expecting with this change the semcheck will fail until a new version release? If so probably we can disable it until the next release?

cijothomas:
> > image Semver check works nicely!
> 
> Are we expecting with this change the semcheck will fail until a new version release? If so probably we can disable it until the next release?

That is a good point. Keeping it enabled means all PRs will be red until new release. But I believe the CI check show exactly what are the breaking changes (not for a given PR, but cumulative...)... 
Let me think more about it.

TommyCpp:
> > > image Semver check works nicely!
> 
> > 
> 
> > Are we expecting with this change the semcheck will fail until a new version release? If so probably we can disable it until the next release?
> 
> 
> 
> That is a good point. Keeping it enabled means all PRs will be red until new release. But I believe the CI check show exactly what are the breaking changes (not for a given PR, but cumulative...)... 
> 
> Let me think more about it.

Can we make the check opt-in like integration test? And we can manually run it if we feel it's needed? Can make it mandatory for all crates once one of them reaches 1.0

PR #1795, opened 2024-05-21
chore: add prometheus back with API and SDK pinned at 0.23

TommyCpp:
## Changes

Keep prometheus exporter with lint checks but depended on fixed version to avoid any breaking changes in SDK and API impacing it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,8 +7,6 @@ members = [
     "examples/*",
     "stress",
 ]
-# Crates temporarily excluded from the workspace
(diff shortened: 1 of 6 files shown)


lalitb (approved):
Thanks, makes sense to have it pinned to last released version.

cijothomas (approved):
Could we include a note in opentelemetry-prometheus readme and main docs.rs content to indicate it only works with api/sdk 0.23, and will be brought back with 1.0 of api/sdk only?

PR #1802, opened 2024-05-22
Add name to info logs in OTLP Example

cijothomas:
Without this, name simply defaults to the source file including line number.


lalitb (approved):
Thanks.

TommyCpp (approved):

PR #1815, opened 2024-05-24
Fix overflow for counter metric

cijothomas:
Overflow handling for Counter fixed to:
1. Allow 2000 unique entries, *excluding* the overflow and 0 attributes case. i.e we will store 2002 entries, specially treating the 0 attribute and overflow to optimize performance (later)
2. Fix the error handling to be triggered once per collection interval at max.
3. Avoid unnecessary cloning of overflow atttribute.
4. Added tests and perf tests to have targeted testing for overflow scenario.

Note: Did not touch Histogram,Gauge, as there are no tests today for them, and their overflow logic looks incorrectly implemented. Will comeback to them after Counter is fully resolved.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/sum.rs
@@ -56,11 +56,13 @@ impl<T: Number<T>> ValueMap<T> {
                     if is_under_cardinality_limit(size) {
                         vacant_entry.insert(measurement);
                     } else {
-                        values
-                            .entry(STREAM_OVERFLOW_ATTRIBUTE_SET.clone())
-                            .and_modify(|val| *val += measurement)
-                            .or_insert(measurement);
-                        global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow.".into()));
+                        if let Some(val) = values.get_mut(&STREAM_OVERFLOW_ATTRIBUTE_SET) {
+                            *val += measurement;
+                            return;
+                        } else {
+                            values.insert(STREAM_OVERFLOW_ATTRIBUTE_SET.clone(), measurement);
+                            global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Further overflows to same metric until next collect will not be logged.".into()));
+                        }
                     }
                 }
             }
--- a/opentelemetry-sdk/src/metrics/internal/sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/sum.rs
@@ -55,14 +55,12 @@ impl<T: Number<T>> ValueMap<T> {
                 Entry::Vacant(vacant_entry) => {
                     if is_under_cardinality_limit(size) {
                         vacant_entry.insert(measurement);
+                    } else if let Some(val) = values.get_mut(&STREAM_OVERFLOW_ATTRIBUTE_SET) {
+                        *val += measurement;
+                        return;
                     } else {
-                        if let Some(val) = values.get_mut(&STREAM_OVERFLOW_ATTRIBUTE_SET) {
-                            *val += measurement;
-                            return;
-                        } else {
-                            values.insert(STREAM_OVERFLOW_ATTRIBUTE_SET.clone(), measurement);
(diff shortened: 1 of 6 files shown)


lalitb (approved):
LGTM. Thanks.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -55,12 +55,12 @@ impl<T: Number<T>> ValueMap<T> {
...
+                        *val += measurement;
+                        return;
                     } else {
-                        values
-                            .entry(STREAM_OVERFLOW_ATTRIBUTE_SET.clone())
-                            .and_modify(|val| *val += measurement)
-                            .or_insert(measurement);
-                        global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow.".into()));
+                        values.insert(STREAM_OVERFLOW_ATTRIBUTE_SET.clone(), measurement);
+                        global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Further overflows to same metric until next collect will not be logged.".into()));
```
utpilla, line 63: > Further overflows to same metric until next collect will not be logged."

Is it because we clear the `HashMap` during every collect?

utpilla, line 63:
```suggestion
                        global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -161,6 +161,35 @@ mod tests {
...
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_overflow_delta() {
+        // Arrange
+        let mut test_context = TestContext::new(Temporality::Delta);
+        let counter = test_context.u64_counter("test", "my_counter", None);
+
+        // Act
+        // Record measurements with A:0, A:1,.......A:1999, which just fits in the 2000 limit
+        for v in 0..2000 {
```
utpilla, line 172: It would be good to also test the overflow scenario when a measurement with zero tags has also been recorded.

stress/src/metrics_overflow.rs
```diff
@@ -0,0 +1,43 @@
...
+
+fn main() {
+    for v in 0..2001 {
+        COUNTER.add(100, &[KeyValue::new("A", v.to_string())]);
+    }
+    throughput::test_throughput(test_counter);
+}
+
+fn test_counter() {
+    COUNTER.add(1, &[KeyValue::new("A", "2001")]);
```
utpilla, line 42: I think we should use random `ATTRIBUTE_VALUES` here to have a fair comparison with the regular stress test for metrics. Emitting the same dimension here would always give better results than this: https://github.com/open-telemetry/opentelemetry-rust/blob/aca4b52e20ee6728c91eb74744adfffa68e061d6/stress/src/metrics.rs#L2-L6

opentelemetry-sdk/src/metrics/internal/sum.rs
cijothomas: yes.

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: yes, good point. (we'd be expecting 2000+1+1 = 2002 points). Will add next.

stress/src/metrics_overflow.rs
cijothomas: The goal for this test was to make sure that in the event of some rogue input, the process won't crash due to metrics i.e are we keeping our memory under cap? are we doing excessive logging (like we were doing prior to this PR) ? etc.

I think this test can do the below to simulate that?
`COUNTER.add(1, &[KeyValue::new("A", i++)]);`

Will keep iterating on this.

utpilla: We could remove `ATTRIBUTE_VALUES` if we are not using them.

PR #1830, opened 2024-05-24
[Trace SDK] Send resource once to processor and exporter, and not with every span 

### LINKED ISSUE #1124: Performance: send resource data to span processors once, not for every span when it ends

lalitb:
Fixes #1124 

## Changes

The changes were done earlier for Logs SDK in #1636 . Creating similar PR for Trace SDK.

**Stress test**: (~ 12% improvement)
before:
```
Number of threads: 16
Throughput: 9,432,600 iterations/sec
Throughput: 9,403,800 iterations/sec
```
after:
```
Number of threads: 16
Throughput: 11,028,400 iterations/sec
Throughput: 11,029,800 iterations/sec
```
**Benchmarks**: (~10% improvement):
start-end-span/always-sample: 
before: 297.20, after: 269.55


opentelemetry-zipkin/src/exporter/mod.rs
```diff
@@ -231,6 +231,8 @@ impl trace::SpanExporter for Exporter {
             self.local_endpoint.clone(),
         ))
     }
+
+    fn set_resource(&mut self, _resource: &Resource) {}
```
cijothomas, line 235: surprised that zipkin wasn't using Resources!

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -92,6 +93,8 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
     /// Shuts down the processor. Called when SDK is shut down. This is an
     /// opportunity for processors to do any cleanup required.
     fn shutdown(&mut self) -> TraceResult<()>;
+    /// Set the resource for the log processor.
+    fn set_resource(&mut self, _resource: &Resource);
```
cijothomas, line 97: provide an empty {}, so we can go easy on those test processors who don't need this

opentelemetry-sdk/src/testing/trace/span_exporters.rs
```diff
@@ -61,6 +60,8 @@ impl SpanExporter for TokioSpanExporter {
     fn shutdown(&mut self) {
         self.tx_shutdown.send(()).unwrap();
     }
+
+    fn set_resource(&mut self, _resource: &crate::Resource) {}
```
cijothomas, line 64: please call this out in the sdk changelog.

opentelemetry-zipkin/src/exporter/mod.rs
lalitb: Yes, it initializes the resource detectors during startup, and then it never changes. Will create a issue for this.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Yes, done.

opentelemetry-sdk/src/testing/trace/span_exporters.rs
lalitb: Added changelog, and the stress test. Seems we don't have benchmark tests of spans with resource to measure it.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/trace.rs#L72-L75 Won't this be enough?

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -77,11 +76,11 @@ impl Span {
     /// overhead.
     pub fn exported_data(&self) -> Option<crate::export::trace::SpanData> {
         let (span_context, tracer) = (self.span_context.clone(), &self.tracer);
-        let resource = self.tracer.provider()?.config().resource.clone();
+        //let resource = self.tracer.provider()?.config().resource.clone();
```
cijothomas, line 79: let us remove the commented out code too.

opentelemetry-stdout/src/trace/exporter.rs
```diff
@@ -107,6 +117,7 @@ impl SpanExporterBuilder {
     pub fn build(self) -> SpanExporter {
         SpanExporter {
             writer: Some(self.writer.unwrap_or_else(|| Box::new(stdout()))),
+            resource: Resource::default(),
```
cijothomas, line 120: could we do empty here instead, as default is not really empty, and will unnecessarily waste cycles to populate the default, only to throw it away and override with SetResource.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -307,12 +307,16 @@ impl OtlpHttpClient {
     fn build_trace_export_body(
         &self,
         spans: Vec<SpanData>,
```
TommyCpp, line 309: Can we remove the `Resource` from `SpanData` then?

opentelemetry-sdk/src/testing/trace/in_memory_exporter.rs
```diff
@@ -142,4 +145,9 @@ impl SpanExporter for InMemorySpanExporter {
     fn shutdown(&mut self) {
         self.reset()
     }
+
+    fn set_resource(&mut self, resource: &Resource) {
+        let mut res_guard = self.resource.lock().expect("Resource lock poisoned");
+        *res_guard = resource.clone();
```
TommyCpp, line 151:
```suggestion
        self.resource.lock().map(|mut res_guard| {
            *res_guard = resource.clone();
        })
```

opentelemetry-stdout/src/trace/transform.rs
```diff
@@ -9,17 +9,27 @@ pub struct SpanData {
...
+        Vec<opentelemetry_sdk::export::trace::SpanData>,
+        &opentelemetry_sdk::Resource,
+    )> for SpanData
+{
+    fn from(
+        (sdk_spans, sdk_resource): (
+            Vec<opentelemetry_sdk::export::trace::SpanData>,
+            &opentelemetry_sdk::Resource,
+        ),
+    ) -> Self {
```
TommyCpp, line 23: might as well create a new constructor instead of a tuple from here

opentelemetry-otlp/src/exporter/tonic/trace.rs
```diff
@@ -66,14 +70,21 @@ impl SpanExporter for TonicTracesClient {
             }
         };
 
+        // TODO: Avoid cloning here.
+        let resource_spans = {
+            batch
+                .into_iter()
+                .map(|log_data| (log_data, &self.resource))
```
TommyCpp, line 77: same^

opentelemetry-zipkin/src/exporter/mod.rs
lalitb: Created an issue #1843

opentelemetry-stdout/src/trace/transform.rs
lalitb: yes `From` with tuple is somewhat weird. Created the new instead. Thanks.

opentelemetry-otlp/src/exporter/tonic/trace.rs
lalitb: done.

opentelemetry-sdk/src/testing/trace/span_exporters.rs
lalitb: yes thanks. Have updated the benchmarks result in the description.

cijothomas (approved):

TommyCpp (approved):
Looks good. Thanks

PR #1831, opened 2024-05-25
Fix tests to make precommit.sh run on Windows

utpilla:
## Changes
- Running `precommit.sh` on Windows locally fails due to some tests
- Fixing those tests in this PR. These were missed out in #1704

### DIFF
--- a/opentelemetry-prometheus/tests/integration_test.rs
+++ b/opentelemetry-prometheus/tests/integration_test.rs
@@ -414,11 +414,10 @@ fn gather_and_compare(registry: prometheus::Registry, expected: String, name: &'
 
 ///  Returns a String which uses the platform specific new line feed character.
 fn get_platform_specific_string(input: String) -> String {
-    if cfg!(windows) {
-        input.replace('\n', "\r\n")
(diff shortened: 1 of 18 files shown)


opentelemetry-prometheus/tests/integration_test.rs
```diff
@@ -406,6 +406,7 @@ fn gather_and_compare(registry: prometheus::Registry, expected: String, name: &'
     let metric_families = registry.gather();
     encoder.encode(&metric_families, &mut output).unwrap();
 
+    let expected = get_platform_specific_string(expected);
```
lalitb, line 409: nit - should we move this normalization to immediately after the file reading? Ideally, gather_and_compare should be agnostic of platform-specific conversion.

utpilla: Since the method is gathering data from `prometheus::Registry` in its definition, we would still need a conversion for the gathered data.

lalitb (approved):

cijothomas (approved):

Issue #1740, opened 2024-05-11
Metrics SDK improvements


cijothomas:
Opening a parent issue to track Metrics SDK improvements for [Stable release](https://github.com/open-telemetry/opentelemetry-rust/issues/1719).

**Background**
The primary function of the Metrics SDK is to accept a number `<T>` along with a slice of KeyValue pairs `<T>, &[KeyValue]`, aggregating these measurements in memory and exporting the aggregated values to Readers/Exporters as needed. Our main goals are ensuring **correctness, thread-safety, memory-efficiency and high performance**, particularly on the "hot path" where measurement reporting occurs, as this demands the utmost efficiency. Correctness/thread-safety/memory-efficiency requires extensive testing via unit tests *and* stress testing.

**Performance issues**
1. Cloning and allocation on hot path -  A significant portion of the overhead involves cloning the incoming slice to prepare `AttributeSet`. We can avoid this in many cases by using a thread-local Vec, which would reduce memory allocations, but still requires copy. Copy as well can be avoided by carefully designing temp data structures to hold references only.
2. Sorting of the Keys - This is a "identity" [requirement](https://github.com/open-telemetry/opentelemetry-rust/pull/1718), so cannot be avoided entirely. However, it is possible to avoid this in the common case, by storing both sorted and original orders for quick lookups.
3. De-deduplication of Attributes with same Keys - This is another "identify" requirement, but similar idea as above can be used to avoid this in the common case.
4. Contention - The use of a Mutex around the HashMap for aggregations leads to heavy contention. Replacing it with a `RwLock` and applying interior mutability could lessen this issue, though sharding may be necessary for further scalability improvements as demonstrated [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1564).
5. Memory efficiency, mostly affecting delta - https://github.com/open-telemetry/opentelemetry-rust/issues/1598 . This is relatively easy to fix.
6. Memory issue - https://github.com/open-telemetry/opentelemetry-rust/issues/1566. This is also relatively easy to fix.

Some issues like calculating hash outside of lock, special casing 0-attributes etc. were addressed already. Also, a lot of ideas were discussed in the past (Community meetings, PRs, issues). I have attempted prototyping several of them here: https://github.com/cijothomas/metrics-mini/tree/main/metrics/src. A lot of the issues from 1,2,3, part of 4 has been addressed in the prototype, giving huge performance improvements. I plan to incorporate them to this repo soon.

[...]

PR #1833, opened 2024-05-25
Metrics Aggregation - Improve throughput by 10x

cijothomas:
Part 1 of https://github.com/open-telemetry/opentelemetry-rust/issues/1740

Modifies Sum aggregation (used by Counter/UpDownCounters), to have less contention by using `RwLock` instead of `Mutex` to access `HashMap` of values. Updates (the hot path) now only need `read()` lock as it leverages interior mutability to update the underlying value. This effectively makes the `HashMap` read-heavy, and only need `read()` locks in hot path, **significantly reducing contention**, and thereby boosting throughput.

Perf numbers from metrics stress test confirms the above: we jump from **3 M/sec to 35 M/sec, i.e 10X jump!** 
Criterion based benchmarks uses single thread and hence, won't show contention. They are not expected to change at all with the changes in this PR - and results show no change.

This PR focused on throughput only, and the next set of PRs (which require more refactoring), will boost the benchmarks as well significantly.


lalitb:
Nice work! updating the value atomically under read lock seems to have boosted the throughout a lot :) Changes look good on quick scan, will review it thoroughly. and though we can get some more juice with #1564 , we are hopefully good for now with this piece of code.

cijothomas:
> Nice work! updating the value atomically under read lock seems to have boosted the throughout a lot :) Changes look good on quick scan, will review it thoroughly. and though we can get some more juice with #1564 , we are hopefully good for now with these changes.

Once contention is avoided (this PR), we are unlikely to gain _hot path_ perf by just sharding alone. But that'll surely help ease the spikes when collect() thread runs, as we can do locks on smaller section instead of whole. We can revisit the sharding logics from 1564.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -1,8 +1,9 @@
+// use std::f32::consts::E;
```
TommyCpp, line 1: Is it needed

hdost (approved):

lalitb (approved):

utpilla (approved):

opentelemetry-sdk/src/metrics/internal/sum.rs
cijothomas: No! Removed.

cijothomas:
> > Nice work! updating the value atomically under read lock seems to have boosted the throughout a lot :) Changes look good on quick scan, will review it thoroughly. and though we can get some more juice with #1564 , we are hopefully good for now with these changes.
> 
> Once contention is avoided (this PR), we are unlikely to gain _hot path_ perf by just sharding alone. But that'll surely help ease the spikes when collect() thread runs, as we can do locks on smaller section instead of whole. We can revisit the sharding logics from 1564.

Correction. Sharding still helps updates() compete less with other updates() that need to insert a new KVP combination.

Issue #1817, opened 2024-05-24
Add metrics aggregation tests to ensure correctness


cijothomas:
The [current tests](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/mod.rs#L164-L176) are only covering Counters.

We need to replicate the same for all instruments. 
Multi-threaded tests are nice to have, but not a must before beta.

utpilla:
I would like to take this up.

cijothomas:
Thanks! Assigned. Lack of test is giving me less confidence to do big refactoring to get perf boost. For now I'll focus on Counters, which do have some coverage.

cijothomas:
Closing as we have this mostly covered.

PR #1834, opened 2024-05-26
Add tests for Histogram metrics

cijothomas:
A small part to help kick start https://github.com/open-telemetry/opentelemetry-rust/issues/1817 
Also, stress test for Histograms. (Histograms have to update multiple memory points, so it must leverage different synchronization mechanisms, and hence must be tested separately)


stress/Cargo.toml
```diff
@@ -4,11 +4,16 @@ version = "0.1.0"
 edition = "2021"
 publish = false
 
-[[bin]] # Bin to run the metrics stress tests
+[[bin]] # Bin to run the metrics stress tests for Counter
 name = "metrics"
```
TommyCpp, line 8: Can we just rename this with _counter suffix?

cijothomas: I renamed the files, and temporarily retaining "metrics" as bin name for counters, just to make it easy to run it, given it is run most frequently. once we are done with refactoring, will use full names

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -199,6 +200,20 @@ mod tests {
         counter_aggregation_helper(Temporality::Delta);
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn histogram_aggregation_cumulative() {
+        // Run this test with stdout enabled to see output.
+        // cargo test histogram_aggregation_cumulative --features=metrics,testing -- --nocapture
```
lalitb, line 206: nit - this comment is valid for all the tests in general, do we need to add specifically for these tests ?

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1007,6 +1022,65 @@ mod tests {
...
+        assert_eq!(data_point1.min.unwrap(), *values_kv1.iter().min().unwrap());
+        assert_eq!(data_point1.max.unwrap(), *values_kv1.iter().max().unwrap());
+
+        let data_point2 =
+            find_histogram_datapoint_with_key_value(&histogram.data_points, "key1", "value2")
+                .expect("datapoint with key1=value2 expected");
+        assert_eq!(data_point2.count, values_kv2.len() as u64);
+        assert_eq!(data_point2.sum, values_kv2.iter().sum::<u64>());
+        assert_eq!(data_point2.min.unwrap(), *values_kv2.iter().min().unwrap());
+        assert_eq!(data_point2.max.unwrap(), *values_kv2.iter().max().unwrap());
```
lalitb, line 1081: I assume there would be separate tests to validate the delta/cumulative aggregation of these time-series data ?

lalitb (approved):
LGTM. Thanks.

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: No strong opinion. I think it makes it easy to copy paste locally, to run one test alone. (I personally use it heavily)

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: absolutely! These basic tests check if the datapoint says it is Cumulative or Delta. Actual validation of the values - not yet done.

PR #1835, opened 2024-05-27
Fix http json example

### LINKED ISSUE #1763: OTLP Exporter with http/json is not working
See : 
https://github.com/open-telemetry/opentelemetry-rust/pull/1752#issuecomment-2109116504

ramgdev:
Fixes #1763 

## Changes

Use string value for event attribute

### DIFF
--- a/opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
+++ b/opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
@@ -9,7 +9,7 @@ publish = false
 once_cell = { workspace = true }
 opentelemetry = { path = "../../../opentelemetry" }
 opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "metrics", "logs"] }
-opentelemetry-otlp = { path = "../..", features = ["http-proto", "reqwest-client", "logs"] }
+opentelemetry-otlp = { path = "../..", features = ["http-proto", "http-json", "reqwest-client", "logs"] }
 opentelemetry-appender-tracing = { path = "../../../opentelemetry-appender-tracing", default-features = false}
 opentelemetry-semantic-conventions = { path = "../../../opentelemetry-semantic-conventions" }
 
--- a/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
@@ -32,7 +32,8 @@ fn init_logs() -> Result<sdklogs::LoggerProvider, opentelemetry::logs::LogError>
         .with_exporter(
             opentelemetry_otlp::new_exporter()
                 .http()
-                .with_endpoint("http://localhost:4318/v1/logs"),
+                .with_protocol(opentelemetry_otlp::Protocol::HttpJson)
+                .with_endpoint("http://dp-ramji-work.fyre.ibm.com:4318/v1/logs"),
         )
         .install_batch(opentelemetry_sdk::runtime::Tokio)
 }
@@ -43,7 +44,8 @@ fn init_tracer_provider() -> Result<sdktrace::TracerProvider, TraceError> {
         .with_exporter(
             opentelemetry_otlp::new_exporter()
                 .http()
-                .with_endpoint("http://localhost:4318/v1/traces"),
+                .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
+                .with_endpoint("http://dp-ramji-work.fyre.ibm.com:4318/v1/traces"),
         )
         .with_trace_config(Config::default().with_resource(RESOURCE.clone()))
         .install_batch(opentelemetry_sdk::runtime::Tokio)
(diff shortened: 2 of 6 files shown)


ramgdev:
@lalitb @TommyCpp please take a look at the changes when you get a chance. I had to disable the call to serialize `int` values to `string` so it works correctly when deserializing from JSON. I think JSON deserializer expects `StringValue` for all string attribute values. 

My editor picked up some formatting changes in addition.

opentelemetry-proto/src/proto.rs
```diff
@@ -46,14 +46,14 @@ pub(crate) mod serializers {
...
-            S: Serializer,
+    where
+        S: Serializer,
     {
         // Serialize any_value::Value using its own implementation
         // If value is None, it will be serialized as such
         match value {
             Some(any_value) => match &any_value.value {
-                Some(Value::IntValue(i)) => serialize_i64_to_string(i, serializer),
+                //Some(Value::IntValue(i)) => serialize_i64_to_string(i, serializer),
```
lalitb, line 56: @ramgdev does commenting it out resolve the export issue? From my understanding, [proto i64/u64 should be converted to JSON strings](https://protobuf.dev/programming-guides/proto3/#json)."

ramgdev: @lalitb yes, this change fixes the exporter issue in the example. I see this in the JSON mapping doc ```JSON value will be a decimal string. Either numbers or strings are accepted.```  Since using numbers in serialization help with deserializing back into either standalone value or value wrapped in `opentelemetry::Value` enum, this should work imho.

ramgdev: Not sure if I am looking at the wrong place. Looks like the attribute spec is asking `int64` to be encoded as number, not string. https://opentelemetry.io/docs/specs/otel/common/#attribute

ramgdev: Doesn't that introduce ambiguity for deserializer without the explicit use of `Value::String` during serialization?

lalitb:
@ramgdev Please go through the fix mentioned in the comment [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1856#issuecomment-2149082284). Do you think it resolves this issue also ?

ramgdev:
> @ramgdev Please go through the fix mentioned in the comment [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1856#issuecomment-2149082284). Do you think it resolves this issue also ?

@lalitb thanks. It looks like it will take care of this issue as well. I will wait to test it after that PR merges and remove the edits from  `proto.rs` in this PR.

lalitb:
> > @ramgdev Please go through the fix mentioned in the comment [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1856#issuecomment-2149082284). Do you think it resolves this issue also ?
> 
> @lalitb thanks. It looks like it will take care of this issue as well. I will wait to test it after that PR merges and remove the edits from `proto.rs` in this PR.

@ramgdev fyi - #1882 is merged.

opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 once_cell = { workspace = true }
 opentelemetry = { path = "../../../opentelemetry" }
 opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "metrics", "logs"] }
-opentelemetry-otlp = { path = "../..", features = ["http-proto", "reqwest-client", "logs"] }
+opentelemetry-otlp = { path = "../..", features = ["http-proto", "http-json", "reqwest-client", "logs"] }
```
lalitb, line 12: @ramgdev Which format would be used by default? Do we need to update the README.md to specify how to switch the format?

ramgdev: @lalitb please see the latest commit. The example shows how to specify the protocol when creating the exporters. We wanted to encourage clients to be explicit about their protocol choice rather than relying on the default. Hope this serves that purpose.

lalitb: Thanks @ramgdev, this provides clarity. I was also considering another option: introducing a feature flag in the Cargo.toml for this example. This would allow users to select the protocol when running the example, similar to how #1861 introduces a feature flag to select the HTTP client (either hyper or reqwest).

ramgdev: The example is using a debug exporter. So, even if we offer a feature flag to change the protocol, they won't see any difference when they run it. Plus, it is easy enough to modify the code to change the endpoint and/or protocol if they really want. So, I'm thinking this is in a good state to illustrate the sdk usage.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -32,7 +32,8 @@ fn init_logs() -> Result<sdklogs::LoggerProvider, opentelemetry::logs::LogError>
         .with_exporter(
             opentelemetry_otlp::new_exporter()
                 .http()
-                .with_endpoint("http://localhost:4318/v1/logs"),
+                .with_protocol(opentelemetry_otlp::Protocol::HttpJson)
```
lalitb, line 35:
```suggestion
                .with_protocol(opentelemetry_otlp::Protocol::HttpBinary)
```

Let's keep all the three exporters as binary for consistency, as we already have the README.md updated for HTTP/JSON support

lalitb (approved):
LGTM with nit change. Thanks for the PR.

opentelemetry-proto/src/proto.rs
ramgdev: resolved by https://github.com/open-telemetry/opentelemetry-rust/pull/1882

ramgdev:
closing to open a separate PR

PR #1845, opened 2024-05-29
Add multi-threaded test for Counter

utpilla:
Towards #1817 

## Changes
- Add multi-threaded test for Counter

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -150,6 +150,7 @@ mod tests {
     use rand::{rngs, Rng, SeedableRng};
     use std::borrow::Cow;
     use std::sync::{Arc, Mutex};
+    use std::thread;
 
     // Run all tests in this mod
     // cargo test metrics::tests --features=metrics,testing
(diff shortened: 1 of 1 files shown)


cijothomas (approved):

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1032,6 +1033,61 @@ mod tests {
         assert!(resource_metrics.is_empty(), "No metrics should be exported as no new measurements were recorded since last collect.");
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_multithreaded() {
```
lalitb, line 1037: nit - Should we instead convert the existing single threaded test to multi-threaded. Or if we need both the tests, how about having a helper which takes number of threads, and temporality as argument?

cijothomas: Maybe keep threading one separate for now, as we need to add more complexities to the threading one (introduce multiple collections etc.). Once we are confident that coverage is good, we can optimize code re-use.

PR #1846, opened 2024-05-29
Add self-diagnostics example

lalitb:
## Changes

Example to demonstrate using `tracing` as a global error handler for errors generated by the OpenTelemetry Metrics SDK. In this example, measurements are recorded to exceed the cardinality limit, which triggers the error to be logged. This error is then emitted to stdout using `opentelemetry-appender-tracing` subscriber.


examples/self-diagnostics/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+opentelemetry-appender-tracing = { path = "../../opentelemetry-appender-tracing"}
+tokio = { workspace = true, features = ["full"] }
+serde_json = { workspace = true }
+tracing = { workspace = true, features = ["std"]}
+tracing-core = { workspace = true }
+tracing-subscriber = { version = "0.3.18", features = ["env-filter","registry", "std"]}
+
+[features]
+default = ["otel_unstable"]
+otel_unstable = ["opentelemetry/otel_unstable"]
```
cijothomas, line 21: what is requiring the unstable feature flag?

examples/self-diagnostics/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    if let Err(err) = set_error_handler(custom_error_handler) {
+        eprintln!("Failed to set custom error handler: {}", err);
+    }
+    //init tracing subscriber to send error logs generated by custom error handler
+    // to OTEL stdout exporter.
+    let exporter = opentelemetry_stdout::LogExporter::default();
+    let provider: LoggerProvider = LoggerProvider::builder()
+        .with_simple_exporter(exporter)
+        .build();
+    let filter = EnvFilter::new("error"); // only push errors
```
cijothomas, line 35: can you add a readme.md and explain how this works and how infinite logging is prevented here?
(I am not sure if we prevent that actually though... For example, if OTLP Endpoint is not reachable, and we use this approach - would we be in infinite loop ?)

lalitb: > For example, if OTLP Endpoint is not reachable, and we use this approach - would we be in an infinite loop

It shouldn't be an infinite loop with OTLP Metrics Exporter + stdout LogExporter. Should we update the example to use the OTLP exporters for both logs and metrics, as that could need some filters?

examples/self-diagnostics/Cargo.toml
lalitb: yeah copy-paste, will remove it.

examples/self-diagnostics/src/main.rs
cijothomas: Yes. That'd be the most common scenario (using OTLP), so showing how exactly to filter would be helpful for majority users.

examples/self-diagnostics/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+fn custom_error_handler(err: OtelError) {
+    match err {
+        OtelError::Metric(err) => error!("OpenTelemetry metrics error occurred: {}", err),
+        OtelError::Trace(err) => error!("OpenTelemetry trace error occurred: {}", err),
+        OtelError::Log(err) => error!("OpenTelemetry log error occurred: {}", err),
+        OtelError::Propagation(err) => error!("OpenTelemetry propagation error occurred: {}", err),
+        OtelError::Other(err_msg) => error!("OpenTelemetry error occurred: {}", err_msg),
+        _ => error!("An unknown OpenTelemetry error occurred"), //won't reach here
+    }
+}
```
TommyCpp, line 22: This is pretty much the same as the default handler but using `traing::error!` macros?

lalitb: yes true. This is to demonstrate using OpenTelemetry logging pipeline as custom handler for self-diagnostics.

ramgdev:
Is there an easy way to catch from the error handler when spans are being dropped?

lalitb:
Moving to draft to address some concerns.

cijothomas:
> Is there an easy way to catch from the error handler when spans are being dropped? Ideally, I'd like to use this technique to build a status provider for the export pipeline.

Can you elaborate more? Or maybe create a new issue ? This PR is just showing how to override error handler to route logs via `tracing`.

examples/self-diagnostics/src/main.rs
lalitb: There is a potential risk of an infinite loop when using the OTLP Metrics Exporter and the OTLP Log Exporter as a custom handler. If the OTLP Log Exporter generates an error, it can trigger this loop. We can manage this in our code; currently, it is being handled by the custom handler.

TommyCpp (approved):

examples/self-diagnostics/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+        .expect("Error setting Ctrl-C handler");
+
+    println!("Press Ctrl-C to continue...");
+    rx.recv().expect("Could not receive from channel.");
+    println!("Got Ctrl-C, Doing shutdown and existing.");
+
+    // Metrics are exported by default every 30 seconds when using stdout exporter,
+    // however shutting down the MeterProvider here instantly flushes
+    // the metrics, instead of waiting for the 30 sec interval.
+    meter_provider.shutdown()?;
```
utpilla, line 142: Use `force_flush` instead of `shutdown`?

examples/self-diagnostics/docker-compose.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+  # Collector
+  otel-collector:
+    image: otel/opentelemetry-collector:latest
+    command: ["--config=/etc/otel-collector-config.yaml", "${OTELCOL_ARGS}"]
+    volumes:
+      - ./otel-collector-config.yaml:/etc/otel-collector-config.yaml
+    ports:
+      - "4318:4318"   # OTLP HTTP receiver
+
```
utpilla, line 12: nit: Remove extra lines.

examples/self-diagnostics/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+    let (tx, rx) = channel();
+
+    ctrlc::set_handler(move || tx.send(()).expect("Could not send signal on channel."))
+        .expect("Error setting Ctrl-C handler");
+
+    println!("Press Ctrl-C to continue...");
+    rx.recv().expect("Could not receive from channel.");
+    println!("Got Ctrl-C, Doing shutdown and existing.");
+
+    // Metrics are exported by default every 30 seconds when using stdout exporter,
```
utpilla, line 139: Copy-paste error? This example is using OTLP exporter with 1 sec export interval.

examples/self-diagnostics/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    // Initialize the MeterProvider with the stdout Exporter.
+    let meter_provider = init_meter_provider();
+
+    // Create a meter from the above MeterProvider.
+    let meter = global::meter("example");
+    // Create a Counter Instrument.
+    let counter = meter.u64_counter("my_counter").init();
+
+    // Record measurements with unique key-value pairs to exceed the cardinality limit
```
utpilla, line 118: It would be good to mention that by default, cardinality limit is 2000.

examples/self-diagnostics/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+license = "Apache-2.0"
+publish = false
+
+[dependencies]
+opentelemetry = { path = "../../opentelemetry" }
+opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["rt-tokio"]}
+opentelemetry-stdout = { path = "../../opentelemetry-stdout"}
+opentelemetry-appender-tracing = { path = "../../opentelemetry-appender-tracing"}
+tokio = { workspace = true, features = ["full"] }
+serde_json = { workspace = true }
```
utpilla, line 14: Is this dependency needed?

examples/self-diagnostics/src/main.rs
lalitb: This was delibrate to call shutdown twice, and trigger "already shutdown" error.

examples/self-diagnostics/docker-compose.yaml
lalitb: thanks done.

examples/self-diagnostics/src/main.rs
lalitb: Good catch. Done

examples/self-diagnostics/Cargo.toml
lalitb: Not needed, have removed it. I think we should add a CI check to catch these errors.

examples/self-diagnostics/src/main.rs
utpilla: The collector logs added in the README file do not show anything for "already shutdown" error though. There is only any entry for exceeding cardinality limit.

examples/self-diagnostics/otel-collector-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+exporters:
+  logging:
+    loglevel: debug
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [logging]
```
utpilla, line 21: Use `debug` exporter.

examples/self-diagnostics/src/main.rs
```diff
@@ -0,0 +1,150 @@
...
+    let (tx, rx) = channel();
+
+    ctrlc::set_handler(move || tx.send(()).expect("Could not send signal on channel."))
+        .expect("Error setting Ctrl-C handler");
+
+    println!("Press Ctrl-C to continue...");
+    rx.recv().expect("Could not receive from channel.");
+    println!("Got Ctrl-C, Doing shutdown and existing.");
+
+    // Metrics are exported by default every 1 seconds when using stdout exporter,
```
utpilla, line 139:
```suggestion
    // MeterProvider is configured with an OTLP Exporter to export metrics every 1 second,
```

examples/self-diagnostics/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+opentelemetry-stdout = { path = "../../opentelemetry-stdout"}
+opentelemetry-appender-tracing = { path = "../../opentelemetry-appender-tracing"}
+tokio = { workspace = true, features = ["full"] }
+tracing = { workspace = true, features = ["std"]}
+tracing-core = { workspace = true }
+tracing-subscriber = { version = "0.3.18", features = ["env-filter","registry", "std"]}
+opentelemetry-otlp = { path = "../../opentelemetry-otlp", features = ["http-proto", "reqwest-client", "logs"] }
+once_cell ={ version = "1.19.0"}
+ctrlc = "3.4"
+
```
utpilla, line 20: nit: Remove the additional line in the end similar to other `.toml` files.

examples/self-diagnostics/src/main.rs
lalitb: Yes, I meant the extra shutdown was to demonstrate that the error is triggered but not logged, as the pipeline is already closed with the first shutdown. But it seems it was causing more confusion, so have removed the second shutdown :)

utpilla (approved):

hdost (approved):

PR #1850, opened 2024-05-31
Small cleanup in Logs SDK.

lalitb:
## Changes

- Set `LoggerProvider::log_processors()` visibility to crate scope.
- Set `LoggerProvider::resource()` visibility to crate scope.
- Modify Builder::build() to first create the `LoggerProvider`, and then propagate the resource to the processor, to address below warning after above scope changes:

```rust
warning: method `resource` is never used
   --> opentelemetry-sdk/src/logs/log_emitter.rs:94:19
    |
87  | impl LoggerProvider {
    | ------------------- method in this implementation
...
94  |     pub(crate) fn resource(&self) -> &Resource {
    |                   ^^^^^^^^
    |
note: the lint level is defined here
   --> opentelemetry-sdk/src/lib.rs:110:5
    |
110 |     unused
    |     ^^^^^^
    = note: `#[warn(dead_code)]` implied by `#[warn(unused)]`
```
- 

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -41,7 +41,9 @@
     - `SpanData` doesn't have the resource attributes. The `SpanExporter::export()` method needs to merge it 
       with the earlier preserved resource before export.
 
-- **Breaking** [1836](https://github.com/open-telemetry/opentelemetry-rust/pull/1836) `SpanProcessor::shutdown` now takes an immutable reference to self. Any reference can call shutdown on the processor. After the first call to `shutdown` the processor will not process any new spans. 
+- **Breaking** [1836](https://github.com/open-telemetry/opentelemetry-rust/pull/1836) `SpanProcessor::shutdown` now takes an immutable reference to self. Any reference can call shutdown on the processor. After the first call to `shutdown` the processor will not process any new spans.
+
+- **Breaking** [1850] (https://github.com/open-telemetry/opentelemetry-rust/pull/1850) `LoggerProvider::log_processors()` and `LoggerProvider::resource()` are not public methods anymore. They are only used within the `opentelemetry-sdk` crate.
 
 ## v0.23.0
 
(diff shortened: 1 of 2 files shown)


utpilla:
> Modify Builder::build() to first create the LoggerProvider, and then propagate the resource to the processor, to address below warning after above scope changes:

Could we just remove those methods instead of force using them just to avoid the "unused" warnings?

lalitb:
> > Modify Builder::build() to first create the LoggerProvider, and then propagate the resource to the processor, to address below warning after above scope changes:
> 
> Could we just remove those methods instead of force using them just to avoid the "unused" warnings?

Yes, thought of doing that, but just kept it to be consistent with recent traces changes #1755. And open to remove from both if we don't see them required.

utpilla:
> Yes, thought of doing that, but just kept it to be consistent with recent traces changes #1755. And open to remove from both signals if we see them bloating the code.

I would be in favor of removing unused methods instead of keeping them around.

cijothomas (approved):

tisonkun:
I have some code rely on this API:

```rust
        for processor in provider.log_processors() {
            let record = record.clone();
            let mut data = opentelemetry_sdk::export::logs::LogData {
                record,
                instrumentation: self.instrumentation_library().clone(),
            };
            processor.emit(&mut data);
        }
```

It seems no migration is provided?

tisonkun:
Rewrite with `Logger`:

```rust
        let provider = self.provider.clone();
        let logger = provider.library_logger(self.library.clone());

        let mut record = opentelemetry_sdk::logs::LogRecord::default();
        record.observed_timestamp = Some(SystemTime::now());
        record.severity_number = Some(log_level_to_otel_severity(log_record.level()));
        record.severity_text = Some(log_record.level().as_str().into());
        record.body = Some(AnyValue::from(log_record.args().to_string()));

        logger.emit(record);
        Ok(())
```

cijothomas:
@tisonkun OTel Logging Bridge is *NOT* meant for end users to use. It is only meant for authoring log-bridges(appenders) only. Are you using OTel Logging Bridge to report logs in your application?

PR #1854, opened 2024-06-02
test: fix MockLogExporter

TommyCpp:
# Changes

Don't think the current unit test is actually working. It's passing because the default resource also has 4 key value paris. 

Fixed it to use mutex and change the test cases so it's has more key values than default resource

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -224,7 +224,7 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                                 &timeout_runtime,
                                 logs.split_off(0),
                             )
-                            .await;
+                                .await;
 
                             if let Err(err) = result {
                                 global::handle_error(err);
@@ -239,7 +239,7 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                             &timeout_runtime,
                             logs.split_off(0),
                         )
-                        .await;
+                            .await;
 
                         if let Some(channel) = res_channel {
                             if let Err(result) = channel.send(result) {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -745,9 +750,11 @@ mod tests {
                 KeyValue::new("k2", "v3"),
                 KeyValue::new("k3", "v3"),
                 KeyValue::new("k4", "v4"),
+                KeyValue::new("k5", "v5"),
             ]))
             .build();
-        assert_eq!(exporter.get_resource().unwrap().into_iter().count(), 4);
+        tokio::time::sleep(Duration::from_secs(2)).await; // set resource in batch span processor is not blocking. Should we make it blocking?
```
TommyCpp, line 756: Is it intented that `set_resource` is non blocking in batch processor? cc @lalitb

lalitb: This was kept non-blocking to be consistent with the other methods. Also, I didn't feel it's theoretically possible for the log event to reach the exporter before the resource is fully set? But apart from that, the latency added by making it non-blocking should be acceptable as it is one-time configuration.

lalitb: Thinking more, the force-flush and shutdown are blocking, so indeed set_resource is not consistent. We can make it blocking - this will also help us capturing failure.

lalitb (approved):

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: Yes please. (i.e make it blocking)
We need to avoid sleeps as much as possible in tests (they generally makes things unstable too)

Issue #1801, opened 2024-05-22
Tracer does weak->Arc upgrade twice in hot path leading to bottleneck


cijothomas:
`Tracer` holds a weakref to its `TracerProvider` and attempts to [upgrade that to an `Arc`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/tracer.rs#L59) *twice* in the hot path - one during span start and then again during span end. This is affecting throughput due to contention.

This issue is already [fixed for Logs](https://github.com/open-telemetry/opentelemetry-rust/issues/1209), opening a new issue to track fixing the same for traces.

TommyCpp:
1. Make the span processor shutdown from any reference #1836 
2. Add `shutdown` to sdk tracer provider
3. Remove `global::shutdown_global_tracer_provider`
4. Now tracer provider can take a `TracerProviderInner` Arc pointer.

cijothomas:
This is done. Closing the issue.

PR #1855, opened 2024-06-02
feat: add `shutdown` in `TracerProvider`

TommyCpp:
Towards #1801

## Changes

- Add `shutdown` in `TracerProvider`.
- `Tracer` now takes a `TracerProvider` instead of `Weak<TracerProviderInner>`, removing the upgrading cost
- A span will NOT started when `TracerProvider` has already shutdown
- A span will NOT be exportered when `TracerProvider` has already shutdown
- Minor fix for LoggerProvider

### DIFF
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -158,7 +158,7 @@ impl TracerProvider {
             .is_ok()
         {
             // propagate the shutdown signal to processors
-            // it's up to the processor to properly block new logs after shutdown
+            // it's up to the processor to properly block new spans after shutdown
             let mut errs = vec![];
             for processor in &self.inner.processors {
                 if let Err(err) = processor.shutdown() {
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -25,7 +25,7 @@ use std::sync::Arc;
 const DEFAULT_COMPONENT_NAME: &str = "rust.opentelemetry.io/sdk/tracer";
 static PROVIDER_RESOURCE: OnceCell<Resource> = OnceCell::new();
 
-// a no nop logger provider used as placeholder when the provider is shutdown
+// a no nop tracer provider used as placeholder when the provider is shutdown
 static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider {
     inner: Arc::new(TracerProviderInner {
         processors: Vec::new(),
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -9,17 +9,36 @@
 //! not duplicate this data to avoid that different [`Tracer`] instances
 //! of the [`TracerProvider`] have different versions of these data.
 use crate::runtime::RuntimeChannel;
-use crate::trace::{BatchSpanProcessor, SimpleSpanProcessor, Tracer};
+use crate::trace::{
+    BatchSpanProcessor, Config, RandomIdGenerator, Sampler, SimpleSpanProcessor, SpanLimits, Tracer,
+};
 use crate::{export::trace::SpanExporter, trace::SpanProcessor};
 use crate::{InstrumentationLibrary, Resource};
-use once_cell::sync::OnceCell;
+use once_cell::sync::{Lazy, OnceCell};
+use opentelemetry::trace::TraceError;
 use opentelemetry::{global, trace::TraceResult};
 use std::borrow::Cow;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 /// Default tracer name if empty string is provided.
 const DEFAULT_COMPONENT_NAME: &str = "rust.opentelemetry.io/sdk/tracer";
 static PROVIDER_RESOURCE: OnceCell<Resource> = OnceCell::new();
+
+// a no nop logger provider used as placeholder when the provider is shutdown
+static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider {
+    inner: Arc::new(TracerProviderInner {
+        processors: Vec::new(),
+        config: Config {
+            // cannot use default here as the default resource is not empty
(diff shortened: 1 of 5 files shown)


opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -9,18 +9,37 @@
...
 use opentelemetry::{global, trace::TraceResult};
 use std::borrow::Cow;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 /// Default tracer name if empty string is provided.
 const DEFAULT_COMPONENT_NAME: &str = "rust.opentelemetry.io/sdk/tracer";
 static PROVIDER_RESOURCE: OnceCell<Resource> = OnceCell::new();
 
+// a no nop logger provider used as placeholder when the provider is shutdown
```
lalitb, line 28:
```suggestion
// a no nop tracer provider used as placeholder when the provider is shutdown
```

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -114,11 +147,41 @@ impl TracerProvider {
...
+    ///
+    /// Note that shut down doesn't means the TracerProvider has dropped
+    pub fn shutdown(&self) -> TraceResult<()> {
+        if self
+            .is_shutdown
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .is_ok()
+        {
+            // propagate the shutdown signal to processors
+            // it's up to the processor to properly block new logs after shutdown
```
lalitb, line 161:
```suggestion
            // it's up to the processor to properly block new spans after shutdown
```

lalitb (approved):
LGTM once the tests are added.

cijothomas (approved):
LGTM. A changelog entry is recommended, as this is removing a significant perf bottleneck. (i tried stress testing - 8-9M/sec to 20M/sec with this PR which is huge improvement)

Do you plan to follow up with removal of global shutdowns similar to what is done for metrics? https://github.com/open-telemetry/opentelemetry-rust/pull/1743

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -52,8 +76,11 @@ impl Default for TracerProvider {
 
 impl TracerProvider {
     /// Build a new tracer provider
-    pub(crate) fn new(inner: Arc<TracerProviderInner>) -> Self {
-        TracerProvider { inner }
+    pub(crate) fn new(inner: TracerProviderInner) -> Self {
+        TracerProvider {
+            inner: Arc::new(inner),
+            is_shutdown: Arc::new(AtomicBool::new(false)),
```
utpilla, line 82: Is there any benefit to creating another `Arc` pointer just for shutdown? Could we reuse the existing `Arc` pointer `inner` to also hold `is_shutdown` inside it?

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -39,9 +58,14 @@ impl Drop for TracerProviderInner {
...
 
 /// Creator and registry of named [`Tracer`] instances.
+///
+/// `TracerProvider` is lightweight container holding pointers to `SpanProcessor` and other components.
+/// Cloning and dropping them will not stop the span processing. To stop span processing, users
+/// must either call `shutdown` method explicitly, or drop every clone of `TracerProvider`.
 #[derive(Clone, Debug)]
 pub struct TracerProvider {
     inner: Arc<TracerProviderInner>,
+    is_shutdown: Arc<AtomicBool>,
```
utpilla, line 68: Is there any benefit to creating another Arc pointer just for shutdown? Could we reuse the existing Arc pointer inner to also hold is_shutdown inside it?

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -114,11 +147,41 @@ impl TracerProvider {
...
+            }
+
+            if errs.is_empty() {
+                Ok(())
+            } else {
+                Err(TraceError::Other(format!("{errs:?}").into()))
+            }
+        } else {
+            Err(TraceError::Other(
+                "tracer provider already shut down".into(),
```
utpilla, line 176: We should probably use `Mutex` or `RwLock` instead of atomic variable to track shutdown. With the current setup, if there are two threads that are racing to shutdown the tracerprovider, the thread which fails the compare and swap and immediately come to the else condition. It wouldn't return an error message saying "tracer provider already shut down" without even waiting or verifying that the other thread has indeed completed the shutdown.

lalitb: Will rephrase the error to say "tracer provider is already shutting down or has been shut down" be useful here, instead of using the Mutex/RwLock ?

utpilla: It depends on what kind of experience we are targeting. I feel that shutting down a tracer provider does not have to be a perf optimized operation so it's okay to use locks. Shutdown would anyway not be a frequent scenario. I also like the clear status that locking offers about whether the provider is shutdown or not.

cijothomas: The mutex/rwlock/atomic - is checked in hot path, so it needs to be performant!

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_processor.rs#L93-L97 Logs. 
Looks like we are inconsistent here. so that need to be taken care as well!

utpilla: > The mutex/rwlock/atomic - is checked in hot path, so it needs to be performant!

Got it! In that case, we should make use of `Arc<Cell<bool>>` to track `is_shutdown` instead of atomics as that should offer better performance.

lalitb: why `Arc<Cell<bool>>` - this can cause race condition and data corruption right, as it's not thread safe.

utpilla: My bad! I thought we could use `Mutex` in conjunction with `Arc<Cell<T>>` to save the atomic read operation on hot-path. Rust wouldn't allow that.

TommyCpp: `Cell<T>` is not `Send` and `Sync` so compiler won't allow this

PR #1857, opened 2024-06-03
Small change to return error from periodic exporter

lalitb:
## Changes

The `collect_and_export()` drops the error in case of failure during export, and always return success. Fixing that. Testing using end-to-end test for OTLP metrics exporter, with no collector running. Not adding the unit-test for now - can be done by creating a failing MockExporter (if required).

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -44,7 +44,6 @@
 - **Breaking** [1836](https://github.com/open-telemetry/opentelemetry-rust/pull/1836) `SpanProcessor::shutdown` now takes an immutable reference to self. Any reference can call shutdown on the processor. After the first call to `shutdown` the processor will not process any new spans.
 
 - **Breaking** [1850] (https://github.com/open-telemetry/opentelemetry-rust/pull/1850) `LoggerProvider::log_processors()` and `LoggerProvider::resource()` are not public methods anymore. They are only used within the `opentelemetry-sdk` crate.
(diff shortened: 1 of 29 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -260,7 +260,7 @@ impl<RT: Runtime> PeriodicReaderWorker<RT> {
         pin_mut!(timeout);
 
         match future::select(export, timeout).await {
-            Either::Left(_) => Ok(()),
+            Either::Left((res, _)) => res, // return the result.
```
cijothomas, line 263: good catch! might be worth adding as a bug fix chanelog entry.

cijothomas (approved):

TommyCpp (approved):
Good catch

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -45,6 +45,8 @@
 
 - **Breaking** [1850] (https://github.com/open-telemetry/opentelemetry-rust/pull/1850) `LoggerProvider::log_processors()` and `LoggerProvider::resource()` are not public methods anymore. They are only used within the `opentelemetry-sdk` crate.
 
+- [1857](https://github.com/open-telemetry/opentelemetry-rust/pull/1857) Fixed an issue in the `collect_and_export()` method of the periodic exporter where errors during export were dropped, ensuring proper error handling and reporting.
```
cijothomas, line 48: `collect_and_export()` is not a public method. I'd suggest to phrase this in a way applicable to end users.

something like:
```Fixed an issue in Metrics SDK which prevented export errors from being send to global error handler. With the fix, errors occurring during export like OTLP Endpoint unresponsive shows up in stderr by default.```

lalitb: Yes true. Updated.

PR #1858, opened 2024-06-04
Improve multi-threaded test for Counter

utpilla:
Follow-up to #1845 

## Changes
- Update Counter multi-threaded test to call Collect during updates
- This offers better coverage for multi-threaded scenarios

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -1081,16 +1081,6 @@ mod tests {
                 assert!(sum.is_monotonic, "Counter should produce monotonic.");
                 assert_eq!(sum.temporality, temporality);
 
-                if let Temporality::Cumulative = temporality {
-                    assert_eq!(
-                        sum.temporality,
-                        Temporality::Cumulative,
-                        "Should produce cumulative"
-                    );
-                } else {
-                    assert_eq!(sum.temporality, Temporality::Delta, "Should produce delta");
-                }
-
                 // find and validate key1=value1 datapoint
                 let data_point = find_datapoint_with_key_value(&sum.data_points, "key1", "value1")
                     .expect("datapoint with key1=value1 expected");
@@ -1590,7 +1580,12 @@ mod tests {
                 "no metrics were exported"
             );
 
-            assert_eq!(self.resource_metrics.len(), invocation_count, "Expected collect to be called {} times", invocation_count);
+            assert_eq!(
+                self.resource_metrics.len(),
+                invocation_count,
+                "Expected collect to be called {} times",
+                invocation_count
+            );
 
             let result = self
                 .resource_metrics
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1553,5 +1573,56 @@ mod tests {
...
+                .expect("metrics expected to be exported");
+
+            assert!(
+                !self.resource_metrics.is_empty(),
+                "no metrics were exported"
+            );
+
+            assert!(
+                self.resource_metrics.len() == invocation_count,
+                "Expected single resource metrics."
```
TommyCpp, line 1595: nit: it's probably not the right message? 
```suggestion
                "Expected {} resource metrics and expect {}.",
    self.resource_metrics.len(),
    invocation_count
```

utpilla: Good catch! It was a copy-paste error. I have changed this to use `assert_eq!` instead of `assert`. Let me know what you think.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1047,45 +1048,64 @@ mod tests {
...
+        // We invoke `test_context.flush_metrics()` six times.
+        let sums =
+            test_context.get_from_multiple_aggregations::<data::Sum<u64>>("my_counter", None, 6);
+
+        let values = sums
+            .iter()
+            .map(|sum| {
+                assert_eq!(sum.data_points.len(), 1); // Expecting 1 time-series.
+                assert!(sum.is_monotonic, "Counter should produce monotonic.");
+                assert_eq!(sum.temporality, temporality);
```
cijothomas, line 1082: this check is duplicate with the one occuring just below

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1047,45 +1048,64 @@ mod tests {
...
-                counter.add(1, &[KeyValue::new("key1", "value1")]);
-            }));
-        }
-
-        for thread in update_threads {
-            thread.join().unwrap();
+        for i in 0..10 {
+            thread::scope(|s| {
+                s.spawn(|| {
+                    counter.add(1, &[KeyValue::new("key1", "value1")]);
```
cijothomas, line 1054: add 1-3 diff. timeseries
add 1 with 0 attributes.

cijothomas: and a separate test for overflow part.

opentelemetry-sdk/src/metrics/mod.rs
utpilla: Fixed

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1047,45 +1048,54 @@ mod tests {
...
-                counter.add(1, &[KeyValue::new("key1", "value1")]);
-                counter.add(1, &[KeyValue::new("key1", "value1")]);
-                counter.add(1, &[KeyValue::new("key1", "value1")]);
-                counter.add(1, &[KeyValue::new("key1", "value1")]);
-            }));
-        }
-
-        for thread in update_threads {
-            thread.join().unwrap();
+        for i in 0..10 {
```
cijothomas, line 1051: nit: maybe use 2x number of cores in the machine?

cijothomas (approved):

PR #1869, opened 2024-06-06
Adding target as LogRecord field, and using it as scope.name in OTLP exporter

### LINKED ISSUE #1683: Appender-tracing and target
OpenTelemetry-Appender-Tracing is currently ignoring the `target` from the Metadata. It is stored as an [attribute "log.target"](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L52) when using experimental feature flag.

`target` from `tracing` is what OTel calls [instrumentation scope's name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#get-a-logger). Hence, the `target` should be stored as instrumentation_scope.name. 

Currently, the instrumentation_scope is hardcoded to be ["opentelemetry-appender-tracing"](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L11), which makes it less useful, as *every* `LogRecord` will have the same information, defeating the purpose of scope/target in OpenTelemetry.

Suggested fixes:
1. Modify the appender to use a different Logger for each `target` it encounters. This would likely kill perf/throughput due to the need of a read or read+insert to a lock-protected-hashmap in the hotpath. This could be mitigated via thread-local hashmap, so the contention can be solved at the expense of more memory. (IIUC, OTel Java uses this approach, but not sure if they use thread-local optimization or not)
2. Store `target` as a special attribute `_tracing.target` in the LogRecord. In the OTLP Exporter thread, specially treat "_tracing.target" and use that to construct instrumentation_scope_name. (and then do not export this special attribute). OTel .NET does something like this to avoid performance bottleneck: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4941

lalitb:
Fixes #1683 

## Changes

- Add an API method `LogRecord::set_target(&mut self, target: Cow<'static, str>)`.
- Introduce `target` as a field in `opentelemetry_sdk::logs::LogRecord`:
```rust
pub struct LogRecord {
    /// Target of the log record
    pub target: Option<Cow<'static, str>>,
}
```
- The opentelemetry-appender-tracing module will call set_target() using the [tracing::metadata::target](https://github.com/tokio-rs/tracing/blob/382ee01dc1cf428c071cad1e3f2e6c6427e70f87/tracing-core/src/metadata.rs#L63) value.
- The opentelemetry-appender-logs module will call set_target() using the [logs::metadata::target](https://github.com/rust-lang/log/blob/c879b011a8ac662545adf9484d9a668ebcf9b814/src/lib.rs#L1081) value.
- The OTLP logs exporter will:
   - Copy the `target` to `scope.name` before exporting the record.

### DIFF
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## vNext
 
 - [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) Utilize the `LogRecord::set_target()` method to pass the tracing target to the SDK.
+  Exporters might use the target to override the instrumentation scope, which previously contained "opentelemetry-appender-tracing".
 
 ## v0.4.0
 
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) Utilize the `LogRecord::set_target()` method to pass the tracing target to the SDK.
+
 ## v0.4.0
 
 - Removed unwanted dependency on opentelemetry-sdk.
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -172,6 +172,7 @@ where
         let mut log_record = self.logger.create_log_record();
         log_record.set_severity_number(severity_of_level(meta.level()));
         log_record.set_severity_text(meta.level().to_string().into());
+        log_record.set_target(meta.target());
 
         let mut visitor = EventVisitor::new(&mut log_record);
         visitor.visit_metadata(meta);
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -172,7 +172,7 @@ where
         let mut log_record = self.logger.create_log_record();
         log_record.set_severity_number(severity_of_level(meta.level()));
         log_record.set_severity_text(meta.level().to_string().into());
-        log_record.set_target(meta.target());
+        log_record.set_target(meta.target().to_string());
 
         let mut visitor = EventVisitor::new(&mut log_record);
         visitor.visit_metadata(meta);
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -172,7 +172,7 @@ where
         let mut log_record = self.logger.create_log_record();
         log_record.set_severity_number(severity_of_level(meta.level()));
         log_record.set_severity_text(meta.level().to_string().into());
-        log_record.set_target(meta.target().to_string());
+        log_record.set_target(meta.target());
 
         let mut visitor = EventVisitor::new(&mut log_record);
         visitor.visit_metadata(meta);
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -18,7 +18,7 @@ now use `.with_resource(RESOURCE::default())` to configure Resource when using
 - Bump MSRV to 1.70 [#1840](https://github.com/open-telemetry/opentelemetry-rust/pull/1840)
 - Fixing the OTLP HTTP/JSON exporter. [#1882](https://github.com/open-telemetry/opentelemetry-rust/pull/1882) - The exporter was broken in the
   previous release.
-- **Breaking** [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) The OTLP logs exporter now overrides the [InstrumentationScope::name](https://github.com/open-telemetry/opentelemetry-proto/blob/b3060d2104df364136d75a35779e6bd48bac449a/opentelemetry/proto/common/v1/common.proto#L73) field with the `target` from `LogRecord`, instead of the appender's name as done previously.
+- **Breaking** [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) The OTLP logs exporter now overrides the [InstrumentationScope::name](https://github.com/open-telemetry/opentelemetry-proto/blob/b3060d2104df364136d75a35779e6bd48bac449a/opentelemetry/proto/common/v1/common.proto#L73) field with the `target` from `LogRecord`, if target is populated.
 
 ## v0.16.0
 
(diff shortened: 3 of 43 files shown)


lalitb:
Moving to draft to add tests, but should be good to review.

opentelemetry-proto/src/transform/common.rs
```diff
@@ -42,21 +42,47 @@ pub mod tonic {
...
+            data: (
+                opentelemetry_sdk::InstrumentationLibrary,
+                Option<Cow<'static, str>>,
+            ),
+        ) -> Self {
+            let (library, target) = data;
             InstrumentationScope {
-                name: library.name.into_owned(),
+                name: target
+                    .map(|t| t.replace("::", "."))
```
cijothomas, line 60: please do this in appender only.

lalitb: Not doing it on purpose. Not all exporters would like to use the target as scope, and mayn't want this conversion.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -172,6 +172,7 @@ where
         let mut log_record = self.logger.create_log_record();
         log_record.set_severity_number(severity_of_level(meta.level()));
         log_record.set_severity_text(meta.level().to_string().into());
+        log_record.set_target(meta.target().to_string());
```
lalitb, line 175: cijo - Can we use the static string instead of copy.

opentelemetry-proto/src/transform/common.rs
lalitb: ok, the conversion has been removed from the exporter now.

opentelemetry-appender-tracing/src/layer.rs
lalitb: We can't avoid the copy as the `metadata.target` is not the static str. It is str reference.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ now use `.with_resource(RESOURCE::default())` to configure Resource when using
 - Bump MSRV to 1.70 [#1840](https://github.com/open-telemetry/opentelemetry-rust/pull/1840)
 - Fixing the OTLP HTTP/JSON exporter. [#1882](https://github.com/open-telemetry/opentelemetry-rust/pull/1882) - The exporter was broken in the
   previous release.
+- **Breaking** [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) The OTLP logs exporter now populates the [InstrumentationScope::name](https://github.com/open-telemetry/opentelemetry-proto/blob/b3060d2104df364136d75a35779e6bd48bac449a/opentelemetry/proto/common/v1/common.proto#L73) field with the name of the target/component emitting the logs, instead of the appender's name as done previously.
```
cijothomas, line 21: OTLP Exporter now overrides InstrumentationScope's name with `target` from `LogRecord`, if it is present.

opentelemetry-appender-tracing/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) Utilize the `LogRecord::set_target()` method to pass the tracing target to the SDK.
```
cijothomas, line 5: add a note saying that, exporters might use target to override the instrumentation scope, which previously contained "opentelemetry-appender-tracing".

opentelemetry-appender-tracing/src/layer.rs
cijothomas, line 175: the logger created by appender  - please modify it to not populate version, as it'll be misleading now.

opentelemetry/src/logs/record.rs
```diff
@@ -10,6 +10,11 @@ pub trait LogRecord {
     {
     }
 
+    /// Sets the `target` of a record
```
cijothomas, line 13: some note about the need for this as opposed to instrumentation scope would be helpful.
also mention that exporters would use target to override instrumentation scope when present.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: also - there is some experimental feature where target was added as an attribute. lets remove that now?

cijothomas (commented):
Looks good, overall. Left some comments to be addressed before marking as approved.

opentelemetry-appender-tracing/src/layer.rs
lalitb: > the logger created by appender - please modify it to not populate version, as it'll be misleading now.

Other thought  - we can keep it, and allow users to customize the logger-name, for those interested to have the single logger, and use that created by appender?

> also - there is some experimental feature where target was added as an attribute. lets remove that now?

Yeah good point. This is basically the "log.target" coming from log-to-tracing conversion. Will handle it as separate PR ?

cijothomas: the version will be misleading. it'll be the version of appender-crate, but the name will be that of the tracing:target. I'd suggest to remove version completely to avoid this confusion..

lalitb: Agree, as of now I have removed `version` and `attributes` population from OTLP exporter. We should make this configurable in OTLP exporter - default to override the scope.name with target, unless configured otherwise. Will create a separate PR for that (and a tracing task). I think the current way of configuration is bit messy with no separate way to provide log/trace/metrics specific configuration.

opentelemetry-otlp/CHANGELOG.md
lalitb: Done.

opentelemetry-appender-tracing/src/layer.rs
lalitb: #1890 to track this.

opentelemetry/src/logs/record.rs
lalitb: Done.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: > This is basically the "log.target" coming from log-to-tracing conversion. Will handle it as separate PR ?
Lets add an issue to track this too.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ now use `.with_resource(RESOURCE::default())` to configure Resource when using
 - Bump MSRV to 1.70 [#1840](https://github.com/open-telemetry/opentelemetry-rust/pull/1840)
 - Fixing the OTLP HTTP/JSON exporter. [#1882](https://github.com/open-telemetry/opentelemetry-rust/pull/1882) - The exporter was broken in the
   previous release.
+- **Breaking** [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) The OTLP logs exporter now overrides the [InstrumentationScope::name](https://github.com/open-telemetry/opentelemetry-proto/blob/b3060d2104df364136d75a35779e6bd48bac449a/opentelemetry/proto/common/v1/common.proto#L73) field with the `target` from `LogRecord`, instead of the appender's name as done previously.
```
cijothomas, line 21:
```suggestion
- **Breaking** [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) The OTLP logs exporter now overrides the [InstrumentationScope::name](https://github.com/open-telemetry/opentelemetry-proto/blob/b3060d2104df364136d75a35779e6bd48bac449a/opentelemetry/proto/common/v1/common.proto#L73) field with the `target` from `LogRecord`, if target is populated.
```

cijothomas (approved):
Left a suggestion to improve changelog wording, rest looks good.

opentelemetry-appender-tracing/src/layer.rs
lalitb: #1894 for tracking.

PR #1871, opened 2024-06-06
Make the stress threads number configurable via command line argument

### LINKED ISSUE #1055: Implement configurable number of cpus for stress testing
for a follow up PR - make the count configurable instead of defaulting to num_cpus. We can use it to measure how good we scale by trying different thread count from 1, 2, ...till num_cpus.

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry-rust/pull/1045#discussion_r1181795387_

ThomsonTan:
Fixes #1055 

## Changes

### DIFF
--- a/stress/src/throughput.rs
+++ b/stress/src/throughput.rs
@@ -45,7 +45,10 @@ where
 
         if arg_num > 0 {
             if arg_num > num_cpus::get() {
-                println!("Specified {} threads which is larger than the number of logical cores ({})!", arg_num, num_threads);
+                println!(
+                    "Specified {} threads which is larger than the number of logical cores ({})!",
+                    arg_num, num_threads
+                );
             }
             num_threads = arg_num as usize;
         } else {
--- a/stress/src/throughput.rs
+++ b/stress/src/throughput.rs
@@ -34,7 +34,7 @@ where
 
     let mut num_threads = num_cpus::get();
     if args.len() >= 2 {
-        let arg = args[1].parse::<i32>();
+        let arg = args[1].parse::<usize>();
 
         if !arg.is_ok() {
             eprintln!("Invalid command line argument '{}' as number of threads. Make sure the value is a positive integer.", args[1]);
--- a/stress/src/throughput.rs
+++ b/stress/src/throughput.rs
@@ -30,23 +30,23 @@ where
     })
     .expect("Error setting Ctrl-C handler");
 
-    let args: Vec<String> = env::args().collect();
-
     let mut num_threads = num_cpus::get();
-    if args.len() >= 2 {
-        let arg = args[1].parse::<usize>();
+    let mut args_iter = env::args();
+
+    if let Some(arg_str) = args_iter.nth(1) {
+        let arg = arg_str.parse::<usize>();
 
         if !arg.is_ok() {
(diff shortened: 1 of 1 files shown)


stress/src/throughput.rs
```diff
@@ -28,7 +29,21 @@ where
         STOP.store(true, Ordering::SeqCst);
     })
     .expect("Error setting Ctrl-C handler");
-    let num_threads = num_cpus::get();
+
+    let args: Vec<String> = env::args().collect();
+
+    let mut num_threads = num_cpus::get();
+    if args.len() == 2 {
+        let arg = args[1].parse::<i32>().unwrap();
```
lalitb, line 37: nit - this can panic if the argument is not numeric. Good to handle that.

TommyCpp (approved):

stress/src/throughput.rs
```diff
@@ -28,7 +29,28 @@ where
         STOP.store(true, Ordering::SeqCst);
     })
     .expect("Error setting Ctrl-C handler");
-    let num_threads = num_cpus::get();
+
+    let args: Vec<String> = env::args().collect();
```
utpilla, line 33: nit: You could the `nth` method to get the first command line argument instead of collecting the arguments into a new vector.

```rust
if let Some(num_threads_arg) = env::args().nth(1) {
    // validate the command line argument
}
```

utpilla: We should probably look at using some crate like [clap](https://docs.rs/clap/latest/clap/) to support named command line arguments.

stress/src/throughput.rs
```diff
@@ -28,7 +29,28 @@ where
...
+
+    let mut num_threads = num_cpus::get();
+    let mut args_iter = env::args();
+
+    if let Some(arg_str) = args_iter.nth(1) {
+        let arg = arg_str.parse::<usize>();
+
+        if !arg.is_ok() {
+            eprintln!("Invalid command line argument '{}' as number of threads. Make sure the value is a positive integer.", arg_str);
+            std::process::exit(1);
```
utpilla, line 41: I think we should use `Result.expect` instead of calling `process.exit`.

Something like this:

```rust
if let Some(arg_str) = args_iter.nth(1) {
    let arg = arg_str.parse::<usize>().expect("Provide a positive value for number of threads");
}

File: stress/src/throughput.rs [hunk 4]
```diff
@@ -28,7 +29,28 @@ where
...
+        }
+
+        let arg_num = arg.unwrap();
+
+        if arg_num > 0 && arg_num <= num_threads {
```
utpilla, line 46: I think when the user has provided a value greater than the number of logical CPUs, we should run the stress test with the maximum allowed value instead of simply exiting. We could log a message saying something like "the value provided was higher than the number of logical cores on the machine. Stress test would be run with n threads instead."

ThomsonTan: Ok. Updated, stay with the passed in threads even if it is larger than the available logical cores.

stress/src/throughput.rs
hdost: I think in this case the slightly more verbose method is ok, as it provides more user-friendly error message.

PR #1877, opened 2024-06-10
fix msrv

### LINKED ISSUE #1874: [Bug]: CI failure
### What happened?

https://github.com/open-telemetry/opentelemetry-rust/actions/runs/9455476592/job/26045353359?pr=1871

error: package `icu_properties v1.5.0` cannot be built because it requires rustc 1.67 or newer, while the currently active rustc version is 1.65.0
Either upgrade to rustc 1.67 or newer, or use
cargo update -p icu_properties@1.5.0 --precise ver
where `ver` is the latest version of `icu_properties` supporting rustc 1.65.0

### API Version

0.22

### SDK Version

0.22

_No response_

### Relevant log output

_No response_

lalitb:
Fixes #1874 

## Changes

The `opentelemetry-http` has an indirect dependency on the `url` crate via `reqwest`. Pinning the `url` crate to a version that supports Rust 1.65.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,4 +45,4 @@ tokio-stream = "0.1.1"
 tracing = { version = "0.1", default-features = false }
 tracing-core = { version = "0.1", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
-url = { version = "2.2", default-features = false }
+url = { version = "=2.5.0", default-features = false } #pinning the version supporting rustc 1.65


cijothomas (approved):
changelog ?

lalitb:
> changelog ?

do we need it, as it won't affect customers. They can still use their own version of `url` crate for their usage.

Cargo.toml
```diff
@@ -45,4 +45,4 @@ tokio-stream = "0.1.1"
 tracing = { version = "0.1", default-features = false }
 tracing-core = { version = "0.1", default-features = false }
 tracing-subscriber = { version = "0.3", default-features = false }
-url = { version = "2.2", default-features = false }
+url = { version = "=2.5.0", default-features = false } #pinning the version supporting rustc 1.65
```
ThomsonTan, line 48: Why need ping this version to support the given rustc?

lalitb: If not pinned, the higher version ` 2.5.1` will be used, which requires rust `1.67`. The previous configured version `2.2` did not have any impact since `url` is an indirect dependency, and thus, the rules defined by the direct dependency that uses it are applied.

cijothomas:
> > changelog ?
> 
> do we need it, as it won't affect customers. They can still continue using their own version of `url` crate for their application usage.

there is bump from 2.2 to 2.5.1 right?

lalitb:
> there is bump from 2.2 to 2.5.1 right?

Not actually. With URL "=2.5.0", we are ensuring that the `reqwest `crate in otel-http doesn't use `url` crate version beyond "2.5.0". Customers needn't be interested to know as long as it doesn't affect their usage of otel crates with the currently supported rust version. And this doesn't anyway affect their direct usage of these crates (reqwest, url) in customer's application.

Issue #1744, opened 2024-05-11
[Metrics SDK] Shutdown false alarm


cijothomas:
Originally discussed here: https://github.com/open-telemetry/opentelemetry-rust/pull/1623/files#r1527013980

```
#[test]
    fn test_shutdown_false_alarm() {
        let reader = TestMetricReader::new();
        let provider = super::SdkMeterProvider::builder()
            .with_reader(reader.clone())
            .build();

        // Initially, shutdown should not be called
        assert!(!reader.is_shutdown());

        // Shutdown the provider manually.
        provider.shutdown().unwrap();
        // Now the shutdown should be invoked
        assert!(reader.is_shutdown());

        // end of this block drops the provider,
        // which try to shutdown again,
        // triggering an error message.
        // that is unactionable for users....
    }
```

cijothomas:
Not very high priority to fix, as the typical pattern would be for user to make a clone of the provider and hand it over to the set_global method, so even when their original provider is dropped, the meter_provider's inner is not dropped, as there is one clone still let.

PR #1878, opened 2024-06-11
[opentelemetry-otlp] basic-otlp-http example - trivial cleanup

markdingram:
avoids metrics log message on shutdown -

discussed in https://github.com/open-telemetry/opentelemetry-rust/issues/1744


lalitb (approved):
Thanks for fixing the example.

cijothomas (approved):

PR #1889, opened 2024-06-18
Implement cardinality limits for metrics streams

### LINKED ISSUE #1065: Implement cardinality limits for metrics streams

bnjjj:
Fixes #1065

## Changes

Adding a new function called `set_stream_cardinality_limit` in `opentelemetry_sdk::metrics` module to set the stream cardinality limit for metric streams.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## vNext
 
+- Implement function to set cardinality limits for metrics streams using `set_stream_cardinality_limit` function.
 - Add "metrics", "logs" to default features. With this, default feature list is
   "trace", "metrics" and "logs".
 - Add `with_resource` on Builder for LoggerProvider, replacing the `with_config`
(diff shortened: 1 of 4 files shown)


cijothomas:
> I think this needs to be implemented as per: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#configuration-1
> 
> In particular cardinality can be set on views rather than globally. Then metric readers. Then the global limit of 2000.

We are removing Views from initial stable release, so this can only be on a MetricReader or most likely on MeterProvider level only for now.

cijothomas (changes requested):
it is unclear to this works. See https://github.com/open-telemetry/opentelemetry-rust/pull/1889#issuecomment-2177247224

The ideal way is via views, but we should hold off from offering that, given Views are being removed for 1st stable release.

cijothomas:
Closing inactive PRs. This is definitely something we need for metrics, and hoping to bring it in next 1-2 releases.

PR #1898, opened 2024-06-27
Make SetResource return status, and call it synchronously in Batch[Log|Span]Processor

lalitb:
## Changes
As discussed [here](https://github.com/open-telemetry/opentelemetry-rust/pull/1854#discussion_r1623663493)

Make `SetResource` return status, and call it blockng in Batch[Log|Span]Processor. This makes the setting of resources during initialization deterministic.

Earlier

```rust
LogExporter::set_resource(&mut self, resource: &opentelemetry_sdk::Resource);
LogProcessor::set_resource(&self, _resource: &Resource);

SpanExporter::set_resource(&mut self, resource: &opentelemetry_sdk::Resource);
SpanProcessor::set_resource(&self, _resource: &Resource);
```

Now:

```rust
LogExporter::set_resource(&mut self, resource: &opentelemetry_sdk::Resource) -> LogResult<()>;
LogProcessor::set_resource(&self, _resource: &Resource) -> LogResult<()>;

SpanExporter::set_resource(&mut self, resource: &opentelemetry_sdk::Resource) -> -> ExportResult
SpanProcessor::set_resource(&self, _resource: &Resource) -> -> ExportResult
```


opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -14,17 +14,43 @@ use std::fmt::Debug;
...
     /// Chek if logs are enabled.
     fn event_enabled(&self, _level: Severity, _target: &str, _name: &str) -> bool {
         true
     }
     /// Set the resource for the exporter.
-    fn set_resource(&mut self, _resource: &Resource) {}
+    /// This function SHOULD only be called once during the initialization of the exporter.
+    /// This function SHOULD complete or abort within some timeout. This function SHOULD be
+    /// implemented as a blocking API
+    fn set_resource(&mut self, _resource: &Resource) -> LogResult<()> {
```
utpilla, line 51: Why does `set_resource` need to return a `Result` or impose timeout limit on it? Shouldn't it always be a matter of assigning/copying a vector?

lalitb: Don't want to make assumptions on what the exporter does within `set_resource()`. The exporter may choose to directly do some custom serialization for resource (say JSON) within this method which can fail. So better to have the status.

cijothomas:
> This makes the setting of resources during initialization deterministic.

Can you elaborate more? I am not sure I follow the intention. What is broken and how is that being fixed ?

lalitb:
> Can you elaborate more? I am not sure I follow the intention. What is broken and how is that being fixed ?

This was discussed and agreed here: https://github.com/open-telemetry/opentelemetry-rust/pull/1854#discussion_r1623663493
To summarise:
- Nothing is broken. 
- The intention is to:
     - Catch the status from the set_resource(), and ensure that error (if any) get's logged.
     - Avoid unnecessary sleeps in tests, as discussed in above link.

This is neither a priority nor a required change. It was agreed upon earlier, and I just wanted to make any such interface changes before the beta release. We can even kill this PR if we now agree with earlier interface :)

cijothomas:
> > Can you elaborate more? I am not sure I follow the intention. What is broken and how is that being fixed ?
> 
> This was discussed and agreed here: [#1854 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/1854#discussion_r1623663493) To summarise:
> 
> * Nothing is broken.
> * The intention is to:
>   
>   * Catch the status from the set_resource(), and ensure that error (if any) get's logged.
>   * Making Exporter::set_resource() call blocking in batch-processor, and so avoid unnecessary sleeps in tests, as discussed in above link.
> 
> This is neither a priority nor a required change. It was agreed upon earlier, and I just wanted to make this interface changes before the beta release. We can even kill this PR if we now agree with earlier interface :)

*Nothing is broken* - Thanks for the context! I agree we need a solution to avoid random sleep, but not sure if the problem needs changes **beyond** BatchProcessors. BatchProcessors should be able to store the Resource reliably, irrespective of the fact that it uses channels for its internal comms. Is it possible to fix BatchProcessor alone?

lalitb:
> Is it possible to fix BatchProcessor alone?

Yes, it is possible. However, `Exporter::set_resource()` function can do more than just copy/store resource attributes; it could allow the exporter to pre-serialize them into the required format, and use that in every export. This serialization can fail. Previously, this step would be performed in `Exporter::emit()`, where any errors could have been returned.

cijothomas:
> > Is it possible to fix BatchProcessor alone?
> 
> Yes, it is possible. However, `Exporter::set_resource()` function can do more than just copy/store resource attributes; it could allow the exporter to pre-serialize them into the required format, and use that in every export. This serialization can fail. Previously, this step would be performed in `Exporter::emit()`, where any errors could have been returned.

You are right that exporter can do much more than just store the attributes. In fact, I think that is the right thing for exporters to do - pre-serialize once at startup those things which won't change. 

What is the SDK supposed to do, if the resource serialization fails in processor/exporter? Should it keep sending normal telemetry or abort that processor? Or should we just let Exporters/Processors deal with it (like logging, or fabricating default resource etc.)

lalitb:
> What is the SDK supposed to do, if the resource serialization fails in processor/exporter? 

As of now, the error is logged by SDK, and continues emitting telemetry to exporter. However if we decide, the status gives a flexibility for  `LoggerProvider::builder().with_resource("...").build()` to return NoopLoggerProvider if one or more exporter(s) fail to set resource.

cijothomas:
Closing old, inactive PRs. Please reopen or make a new one when continuing.

PR #1899, opened 2024-06-27
Add flamegraph for some benchmarks

cijothomas:
To get flamegraphs. Only done for couple of benchmarks which are in the area actively being worked on.


opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -173,5 +174,16 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+    name = benches;
+    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    targets = criterion_benchmark
+}
+#[cfg(target_os = "windows")]
+criterion_group! {
+    name = benches;
+    config = Criterion::default();
+    targets = criterion_benchmark
+}
```
TommyCpp, line 188: May be we can use https://crates.io/crates/cfg-if to make things easier. Especially in dev tools

lalitb (approved):

PR #1901, opened 2024-06-28
Benchmarks for AnyValue

cijothomas:
For discussion on the cost of AnyValue abstraction, as opposed to Value abstraction (which itself is expensive)


opentelemetry/benches/anyvalue.rs
```diff
@@ -0,0 +1,44 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use opentelemetry::{logs::AnyValue, Value};
+
+// Run this benchmark with:
+// cargo bench --bench anyvalue
+// Results:
+// CreateOTelValueString 1-2 ns
+// CreateOTelAnyValueString 15 ns
```
cijothomas, line 8: @lalitb The difference is huge, so maybe worth modifying LogRecord to stick with `Value`, and not `AnyValue`. None of the appenders are capable of leveraging `AnyValue` anyway.

lalitb: Good find. Should we just move the complex variants within `AnyValue` under the feature flag. If someone wants for their appender, it comes with a cost.

cijothomas: I am okay to completely remove `AnyValue` and replace it with `Value` as well for now. Let us discuss more before making any changes. (This will not affect tokio-tracing, as they don't have any way of producing AnyValue

lalitb: With this change (non-primitives to be enclosed as `Box` types):

```rust
/// Value types for representing arbitrary values in a log record.
pub enum AnyValue {
    /// An integer value
    Int(i64),
    /// A double value
    Double(f64),
    /// A string value
    String(StringValue),
    /// A boolean value
    Boolean(bool),
    /// A byte array
    Bytes(Box<Vec<u8>>),
    /// An array of `Any` values
    ListAny(Box<Vec<AnyValue>>),
    /// A map of string keys to `Any` values, arbitrarily nested.
    Map(Box<HashMap<Key, AnyValue>>),
}
```

The benchmark seems comparable:

```
CreateOTelValueString    time:   [3.3011 ns 3.3352 ns 3.3736 ns]
CreateOTelAnyValueString time:   [2.9156 ns 2.9417 ns 2.9722 ns]
```

---

lalitb (approved):

opentelemetry/benches/anyvalue.rs
cijothomas: @lalitb will explore making changes to AnyValue separately. We can merge this PR without waiting.

cijothomas: @TommyCpp to look at this too.

PR #1907, opened 2024-07-02
Update metric benchmarks

utpilla:
## Changes
- Update metrics benchmarks for API
- Use `iter_batched` method to generate the dynamic attributes before each iteration

### DIFF
--- a/opentelemetry/benches/metrics.rs
+++ b/opentelemetry/benches/metrics.rs
@@ -1,4 +1,4 @@
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
 use opentelemetry::{global, metrics::Counter, KeyValue};
 
 // Run this benchmark with:
@@ -28,10 +28,10 @@ fn counter_add(c: &mut Criterion) {
             counter.add(
                 1,
                 &[
-                    KeyValue::new("attribute1", "value1"),
-                    KeyValue::new("attribute2", "value2"),
-                    KeyValue::new("attribute3", "value3"),
-                    KeyValue::new("attribute4", "value4"),
+                    black_box(KeyValue::new("attribute1", "value1")),
+                    black_box(KeyValue::new("attribute2", "value2")),
+                    black_box(KeyValue::new("attribute3", "value3")),
+                    black_box(KeyValue::new("attribute4", "value4")),
                 ],
             );
         });
@@ -51,16 +51,26 @@ fn counter_add(c: &mut Criterion) {
     });
 
     c.bench_function("Counter_AddWithDynamicAttributes", |b| {
-        b.iter(|| {
-            let kv = vec![
-                KeyValue::new("attribute1", "value1"),
-                KeyValue::new("attribute2", "value2"),
-                KeyValue::new("attribute3", "value3"),
-                KeyValue::new("attribute4", "value4"),
-            ];
+        b.iter_batched(
+            || {
+                let value1 = black_box("a".repeat(6)); // Repeat character six times to match the length of value strings used in other benchmarks
+                let value2 = black_box("b".repeat(6));
+                let value3 = black_box("c".repeat(6));
+                let value4 = black_box("d".repeat(6));
 
-            counter.add(1, &kv);
-        });
+                (value1, value2, value3, value4)
+            },
+            |values| {
+                let kv = &[
+                    KeyValue::new("attribute1", values.0),
+                    KeyValue::new("attribute2", values.1),
+                    KeyValue::new("attribute3", values.2),
+                    KeyValue::new("attribute4", values.3),
+                ];
+                
+                counter.add(1, kv);
+            },
+            BatchSize::SmallInput);
     });
 }
 
--- a/opentelemetry/benches/metrics.rs
+++ b/opentelemetry/benches/metrics.rs
@@ -67,10 +67,11 @@ fn counter_add(c: &mut Criterion) {
                     KeyValue::new("attribute3", values.2),
                     KeyValue::new("attribute4", values.3),
                 ];
-                
+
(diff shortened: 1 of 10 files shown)


opentelemetry/benches/metrics.rs
```diff
@@ -51,16 +51,27 @@ fn counter_add(c: &mut Criterion) {
...
-        b.iter(|| {
-            let kv = vec![
-                KeyValue::new("attribute1", "value1"),
-                KeyValue::new("attribute2", "value2"),
-                KeyValue::new("attribute3", "value3"),
-                KeyValue::new("attribute4", "value4"),
-            ];
+        b.iter_batched(
+            || {
+                let value1 = black_box("a".repeat(6)); // Repeat character six times to match the length of value strings used in other benchmarks
```
lalitb, line 56: nit - do we need black_box for creating these strings?

lalitb (approved):
Thanks for fixing these. Should we also add the benchmarks results as done here -  https://github.com/open-telemetry/opentelemetry-rust/blob/ad990d6d33aa73b35e0e3b792b1fe3ea4251b92b/opentelemetry/benches/anyvalue.rs#L6-L10

also add the machine configuration used for testing.

opentelemetry/benches/metrics.rs
```diff
@@ -28,10 +28,10 @@ fn counter_add(c: &mut Criterion) {
             counter.add(
                 1,
                 &[
-                    KeyValue::new("attribute1", "value1"),
-                    KeyValue::new("attribute2", "value2"),
-                    KeyValue::new("attribute3", "value3"),
-                    KeyValue::new("attribute4", "value4"),
+                    black_box(KeyValue::new("attribute1", "value1")),
```
cijothomas, line 31: curious : what is the actual need for this here? are we worried that compiler will do some optimizations here?

lalitb: Good point. Should it be good to get the benchmark result with optimization (if any) as that would be the real scenario?

utpilla: > are we worried that compiler will do some optimizations here?

Yes, that was my motivation to add `black_box`.

cijothomas: i am not convinced we need to wrap in black_box.. If we were simply creating a KeyValue, then it makes sense, as compiler might think we are not using it anywhere so try to optimize in a way that does not match real scenarios.
In this case, Counter API is used, so we probably don't need black_box.

utpilla: > compiler might think we are not using it anywhere so try to optimize in a way

The compiler could make other optimizations as well.

I benchmarked this method with and without `black_box` and there is a considerable perf difference around 32%

Without black_box: 13.523 ns
With black_box: 19.802

opentelemetry/benches/metrics.rs
```diff
@@ -1,4 +1,4 @@
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};
```
cijothomas, line 1: could you add the numbers here, similar to https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/metric_counter.rs#L1-L13 for  easy reference.

opentelemetry/benches/metrics.rs
cijothomas: we want to measure the real user scenario right? Won't compiler be doing same optimization (the one wihtthout blackbox) in their code?

utpilla: Yeah, it makes sense to mimic the real user scenario.

opentelemetry/benches/metrics.rs
utpilla: No, we can remove these as well.

opentelemetry/benches/metrics.rs
utpilla: Updated.

opentelemetry/benches/metrics.rs
utpilla: Updated.

opentelemetry/benches/metrics.rs
```diff
@@ -44,23 +58,34 @@ fn counter_add(c: &mut Criterion) {
...
-            let kv = vec![
-                KeyValue::new("attribute1", "value1"),
-                KeyValue::new("attribute2", "value2"),
-                KeyValue::new("attribute3", "value3"),
-                KeyValue::new("attribute4", "value4"),
-            ];
+    c.bench_function("AddWithDynamicAttributes", |b| {
+        b.iter_batched(
+            || {
+                let value1 = "value1".to_string(); // Repeat character six times to match the length of value strings used in other benchmarks
```
cijothomas, line 70:
```suggestion
                let value1 = "value1".to_string();
```

opentelemetry/benches/metrics.rs
```diff
@@ -1,4 +1,19 @@
...
+    criterion = "0.5.1"
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
+    RAM: 64.0 GB
+    | Test                                                | Average time|
+    |-----------------------------------------------------|-------------|
+    | NoAttributes                                        | 1.1616 ns   |
+    | AddWithInlineStaticAttributes                       | 13.296 ns   |
+    | AddWithStaticArray                                  | 1.1612 ns   |
+    | AddWithDynamicAttributes                            | 110.40 ns   |
```
utpilla, line 12: This benchmark method runs much faster when run in isolation. It takes around 80 ns instead of 110 ns.

PR #1488, opened 2024-01-24
Add event_name field to `LogRecord`

bIgBV:
Fixes #
Design discussion issue (if applicable) #1349 

## Changes

Add `event_name` field to the `LogRecord` struct.

The `event_name` field is manually passed around through attributes to exporters which then have to parse it out. This change would being it in line with existing logging and tracing frameworks in the rust ecosystem, and provide a performance boost to downstream consumers who depend on the `event_name` field.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1908, opened 2024-07-03
Predefine Vector Capacity for LogRecord Attributes

### LINKED ISSUE #1896: LogRecord Attributes - preallocate attributes vec?
image
Since we don't pre-allocate, the vec starts with size 1, and often requires re-allocating. Would we start with a default size of, say 5? 
(don't ask me why 5, just a magic number!)

Opening an issue to track this improvement. The change can improve current throughput, benchmark by 10-20% (as they all need ~5 attributes.)

lalitb:
Fixes #1896 

**Stress test**

**_Before_**:
Throughput: 57,703,000 iterations/sec
Memory usage: 3.70 MB
CPU usage: 99.87512%
Virtual memory usage: 2249.05 MB

**_With PR:_**
Throughput: 70,031,600 iterations/sec
Memory usage: 3.98 MB
CPU usage: 100%
Virtual memory usage: 2249.05 MB


opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -10,7 +10,7 @@
     | noop_layer_disabled         | 12 ns       |
     | noop_layer_enabled          | 25 ns       |
     | ot_layer_disabled           | 19 ns       |
-    | ot_layer_enabled            | 371 ns      |
+    | ot_layer_enabled            | 270 ns      |
```
cijothomas, line 13: ❤️ this! Could you update stress tests too?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -37,6 +39,22 @@ pub struct LogRecord {
...
+        LogRecord {
+            event_name: None,
+            target: None,
+            timestamp: None,
+            observed_timestamp: None,
+            trace_context: None,
+            severity_text: None,
+            severity_number: None,
+            body: None,
+            attributes: Some(Vec::with_capacity(PREALLOCATED_ATTRIBUTE_CAPACITY)), // Pre-allocate for perf optimization. This may change in future.
```
cijothomas, line 53: question: Should we micro-optimize this to only allocate attributes, when first attribute is added, so that if someone is simply using body only, and no attributes, then we can save some space...

cijothomas (approved):

opentelemetry-sdk/src/logs/record.rs
lalitb: That's a good point. Also to add, as of now we are always adding `meta.name` to [attributes](https://github.com/open-telemetry/opentelemetry-rust/blob/ad990d6d33aa73b35e0e3b792b1fe3ea4251b92b/opentelemetry-appender-tracing/src/layer.rs#L43), so we won't get benefit of above optimization. The event-name should be instead populated as  the struct member `LogRecord::name` (after #1488). Can incorporate both these changes as separate PR.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -37,6 +39,22 @@ pub struct LogRecord {
     pub attributes: Option<Vec<(Key, AnyValue)>>,
 }
 
+impl Default for LogRecord {
```
utpilla, line 42: Since we are only overriding one field, it might be better to keep the auto implementation of `Default` trait and instead update the call site like this:

```rust
fn create_log_record(&self) -> Self::LogRecord {
        LogRecord { attributes : Some(Vec::with_capacity(PREALLOCATED_ATTRIBUTE_CAPACITY)), .. Default::default() }
    }
```

lalitb: yes this indeed is the better approach. Have modified accordingly.  Thanks.

utpilla (approved):

opentelemetry-appender-tracing/benches/logs.rs
lalitb: updated stress. The benchmark shows improvement from 391 ns -> 305 ns. Updated that too.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -246,7 +247,11 @@ impl opentelemetry::logs::Logger for Logger {
     type LogRecord = LogRecord;
 
     fn create_log_record(&self) -> Self::LogRecord {
-        LogRecord::default()
+        // Reserve attributes memory for perf optimization. This may change in future.
+        LogRecord {
+            attributes: Some(Vec::with_capacity(PREALLOCATED_ATTRIBUTE_CAPACITY)),
+            ..Default::default()
```
ThomsonTan, line 253: Just curious, what is the `..Default` for?

stress/src/logs.rs
```diff
@@ -3,7 +3,7 @@
     OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
     Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
     RAM: 64.0 GB
-    53 M/sec
+    69 M/sec
```
ThomsonTan, line 6: Does this mean that more memory is consumed in benchmark?

cijothomas: @lalitb maybe use https://github.com/open-telemetry/opentelemetry-rust/tree/main/stress#feature-flags and compare the before and after results, so we know what is the impact (for the stress test!). (We don't need to update this file with that number, but can leave it as a comment in this PR for easy reference in the future)

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: `..Default::default()` means that all the remaining fields of the struct that are not explicitly set will be initialized to their default values. From `C++` prospective, this can be thought of as using constructor overloading, where one constructor initializes specific fields, and another constructor (the default constructor) initializes the rest. `Rust` doesn't have concept of function overloading, so this is the pattern used.

stress/src/logs.rs
lalitb: > Does this mean that more memory is consumed in benchmark?

This is a measure of throughput, 69 Million events/sec

lalitb: Good point, update the PR description with memory stats.

PR #1856, opened 2024-06-03
Fix serde

hwrdtm:
Make compatible with Collector serde shape

## Changes

This PR:
- Flattens some `data` and `value` fields in order to be consistent with the JSON schema implemented on the OTEL Collector
  - See [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/data/protogen/metrics/v1/metrics.pb.go#L469) for the struct defs and JSON tags
  - Also see [here](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.3.1/examples/metrics.json) for an example resource metric in JSON
- Make the `common::v1::KeyValue::value` field serde into an `Option<AnyValue>` rather than the direct serialized value in order to be consistent with the JSON schema implemented on the OTEL Collector
  - See [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/data/protogen/common/v1/common.pb.go#L288) for the struct defs and JSON tags
  - Also see [here](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.3.1/examples/trace.json#L40-L42) for an example resource span in JSON
  - Similar comment made on an earlier PR: https://github.com/open-telemetry/opentelemetry-rust/pull/1462/files#r1623330058

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1911, opened 2024-07-04
Flatten field + fix u64 to string serde

hwrdtm:
## Changes

This PR:
- Is a continuation of what was originally surfaced https://github.com/open-telemetry/opentelemetry-rust/pull/1856.
- Flattens some `data` and `value` fields in order to be consistent with the JSON schema implemented on the OTEL Collector
  - See [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/internal/data/protogen/metrics/v1/metrics.pb.go#L469) for the struct defs and JSON tags
  - Also see [here](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.3.1/examples/metrics.json) for an example resource metric in JSON
- Adds more fields to be correctly doing serde between `u64` and `String`
- Without both of these changes the `./expected/metrics.json` file **wouldn't even deserialize correctly**.

## Testing

- Run `cargo test --workspace --all-features --test grpc_build -- --nocapture` to generate grpc files. Running the first time is expected to fail, as it will create the files.
- Run `cargo test --workspace --all-features --test metrics 'test_serde' -- --nocapture` to run the new test case introduced in this PR.

cc @lalitb


hwrdtm:
Any thoughts here?

lalitb:
Thanks for the PR @hwrdtm . Will review it tomorrow.

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
```diff
@@ -0,0 +1,23 @@
...
+fn test_serde() {
+    let metrics = read_metrics_from_json(File::open("./expected/metrics.json").unwrap());
+
+    let json = serde_json::to_string_pretty(&MetricsData {
+        resource_metrics: metrics,
+    })
+    .expect("Failed to serialize metrics");
+
+    // Write to file.
+    let mut file = File::create("./expected/serialized_metrics.json").unwrap();
```
TommyCpp, line 16: We can use [tempfile](https://crates.io/crates/tempfile) here

hwrdtm: I am not sure we should, leaving the expected shape here committed to source allows us to almost use a snapshot-testing technique. 

The problem is, if the serde logic is wrong, then the serialization from `./expected/metrics.json` would be wrong, and the deserialization back into a file would also be wrong, **but because both the serde is wrong in the same manner, the `MetricsAsserter` would assert their shapes to be the same**. By leaving the expected file committed to source, we'll be able to notice that the actual errors in serde.
- For example, try commenting out all the changes in `grpc_build.rs` in this PR, generating the proto files once, and then running the metrics integration test introduced in this PR. I also logged the `left` and `right` and I can see that they are both incorrectly serde **in the same manner** (`data: None`):

```
left: [
    ResourceMetrics {
        resource: Some(
            Resource {
                attributes: [
                    KeyValue {
                        key: "service.name",
                        value: Some(
                            AnyValue {
                                value: Some(
                                    StringValue(
                                        "my.service",
                                    ),
                                ),
                            },
                        ),
                    },
                ],
                dropped_attributes_count: 0,
            },
        ),
        scope_metrics: [
            ScopeMetrics {
                scope: Some(
                    InstrumentationScope {
                        name: "my.library",
                        version: "1.0.0",
                        attributes: [
                            KeyValue {
                                key: "my.scope.attribute",
                                value: Some(
                                    AnyValue {
                                        value: Some(
                                            StringValue(
                                                "some scope attribute",
                                            ),
                                        ),
                                    },
                                ),
                            },
                        ],
                        dropped_attributes_count: 0,
                    },
                ),
                metrics: [
                    Metric {
                        name: "my.counter",
                        description: "I am a Counter",
                        unit: "1",
                        metadata: [],
                        data: None,
                    },
                    Metric {
                        name: "my.gauge",
                        description: "I am a Gauge",
                        unit: "1",
                        metadata: [],
                        data: None,
                    },
                    Metric {
                        name: "my.histogram",
                        description: "I am a Histogram",
                        unit: "1",
                        metadata: [],
                        data: None,
                    },
                ],
                schema_url: "whatever",
            },
        ],
        schema_url: "whatever",
    },
]
right: [
    ResourceMetrics {
        resource: Some(
            Resource {
                attributes: [
                    KeyValue {
                        key: "service.name",
                        value: Some(
                            AnyValue {
                                value: Some(
                                    StringValue(
                                        "my.service",
                                    ),
                                ),
                            },
                        ),
                    },
                ],
                dropped_attributes_count: 0,
            },
        ),
        scope_metrics: [
            ScopeMetrics {
                scope: Some(
                    InstrumentationScope {
                        name: "my.library",
                        version: "1.0.0",
                        attributes: [
                            KeyValue {
                                key: "my.scope.attribute",
                                value: Some(
                                    AnyValue {
                                        value: Some(
                                            StringValue(
                                                "some scope attribute",
                                            ),
                                        ),
                                    },
                                ),
                            },
                        ],
                        dropped_attributes_count: 0,
                    },
                ),
                metrics: [
                    Metric {
                        name: "my.counter",
                        description: "I am a Counter",
                        unit: "1",
                        metadata: [],
                        data: None,
                    },
                    Metric {
                        name: "my.gauge",
                        description: "I am a Gauge",
                        unit: "1",
                        metadata: [],
                        data: None,
                    },
                    Metric {
                        name: "my.histogram",
                        description: "I am a Histogram",
                        unit: "1",
                        metadata: [],
                        data: None,
                    },
                ],
                schema_url: "whatever",
            },
        ],
        schema_url: "whatever",
    },
]
```

lalitb: @hwrdtm I understand that if the serde logic is incorrect, serialization and deserialization would be incorrect in same way, and the actual error would get unnoticed. However, how is the committing the expected output in git helping here, as we are not using it for comparison anywhere? Probably I am missing something :)

hwrdtm: It produces a git diff that the author would hopefully notice and dig into. It's not foolproof, but is consistent with snapshot testing paradigms, because sometimes we may actually wish to change the snapshot, and so the git diff is not always necessarily an error, and human judgement is helpful here.

lalitb: Thanks for clarification.

lalitb (approved):

PR #1918, opened 2024-07-09
Initial attempt at adding metric semconvs.

tbrockman:
Fixes # `TBD`
Slack context # https://cloud-native.slack.com/archives/C03GDP0H023/p1720453950905069

## Changes

* Creates a new `semantic_metrics.rs.j2` Jinja template for producing `metric.rs`, which contains Metric semantic conventions.
* Stores all attributes `const &str` in `attributes.rs` file
* Other files (`resource.rs`, `trace.rs`) re-export attributes from `attribute.rs` to preserve backwards compatibility (though it might be desirable to break compatibility here as it seems `resource.rs` and `trace.rs` both contain duplicated attributes from `attribute_group`)

Generated docstring VSCode examples:
image
image

As mentioned in Slack, it seems like the code generation will be migrated to `weaver`, but I'd already wrote this and perhaps this could be useful in the interim.


cijothomas:
> to preserve backwards compatibility

The crate is pre-release (with no planned stable release date as of today), so don't worry too much about back compatibility. Feel free to do the correct thing, without feeling burdened by the need of back-compatbility for this crate.

tbrockman:
> > to preserve backwards compatibility
> 
> The crate is pre-release (with no planned stable release date as of today), so don't worry too much about back compatibility. Feel free to do the correct thing, without feeling burdened by the need of back-compatbility for this crate.

Oh awesome, I wasn't really aware of the current release status, thanks for the heads up!

Changes made to remove those attributes from `trace.rs` and `resource.rs`, making them available through `attribute.rs`.

tbrockman:
Current CI failures are unrelated:
```bash
error: package `cc v1.1.0` cannot be built because it requires rustc 1.67 or newer, while the currently active rustc version is 1.65.0
Either upgrade to rustc 1.67 or newer, or use
cargo update -p cc@1.1.0 --precise ver
where `ver` is the latest version of `cc` supporting rustc 1.65.0
Error: Process completed with exit code 1.
```
Looks like https://github.com/open-telemetry/opentelemetry-rust/pull/1924 might fix this, will update when that's merged.

aumetra:
> Looks like https://github.com/open-telemetry/opentelemetry-rust/pull/1924 might fix this, will update when that's merged.

Yep, my PR adds an entry to the `patch_dependencies.sh` script to downgrade the crate to the last version that fits the semver requirements (namely `cc v1.0.105`)

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
```diff
@@ -51,16 +70,15 @@ fi
...
 
-# Remove the messaging.client_id definition along with its comments from the generated files
-#   - semconv "messaging.client_id" is deprecated
-#   - semconv "messaging.client.id" is to be used instead
-#   - Now because we use:
-#      	pub const {{attribute.fqn | to_const_name}}: &str = "{{attribute.fqn}}";
-#     to generate the consts, where to_const_name replaces '.' with '_', we need to remove the old definition
-#	  to avoid conflicts with the new one. Refer - https://github.com/open-telemetry/semantic-conventions/issues/1031
-"${SED[@]}" '/\/\/\/ Deprecated, use `messaging.client.id` instead\./{N;N;N;N;d;}' src/trace.rs src/resource.rs
-"${SED[@]}" '/pub const MESSAGING_CLIENT_ID: &str = "messaging.client_id";/{N;d;}' src/trace.rs src/resource.rs
```
lalitb, line 64: This special handling for `messaging.client_id` was to resolve the conflict as same variable was getting created for new and deprecated value. Is this no more an issue?

tbrockman: Hopefully! I moved the logic to the Jinja template itself: https://github.com/open-telemetry/opentelemetry-rust/pull/1918/files#diff-1a84552ce3e0d372a5f57ad55a985d0dd2a8af846561449c804ef3d835dfae66R14-R16

If there are multiple which resolve to the same name variable name, it takes the first non-deprecated. If there aren't any, it returns the first of whatever deprecated variable is available.

lalitb: Thanks, this looks cleaner approach. Though not sure if I am missing something, as of now I only see the variable name with deprecated value in this PR i.e., 
```rust
pub const MESSAGING_CLIENT_ID: &str = "messaging.client_id";
```
while it should be:
```rust
pub const MESSAGING_CLIENT_ID: &str = "messaging.client.id";

```

lalitb:
@tbrockman  Can you also add changelog for the support of metrics semconvs, along with the breaking change for consolidation of attributes.

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
tbrockman: Ah poop, good catch. I'd guess `selectattr('stability_string', '!=', 'StabilityLevel.DEPRECATED')` (or something similar) isn't functioning as I expect. I'll look into fixing it when I get a chance.

tbrockman:
> @tbrockman Can you also update changelog for the support of metrics semconvs, along with the breaking change for consolidation of attributes.

Can do.

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
tbrockman: @lalitb Ah I misunderstood the potential values for StabilityLevel (looks like the original `StabilityLevel.DEPRECATED` level referenced in the script doesn't exist? https://github.com/open-telemetry/build-tools/blob/ffcffc73b3575942932de31cdae0ffe35c181186/semantic-conventions/src/opentelemetry/semconv/model/semantic_attribute.py#L40). Explains why there are no `#[deprecated]` macros anywhere. Correct key to reference seems to be `attribute.deprecated`.

tbrockman: Believe this should be fixed now!

tbrockman:
@lalitb believe all related `Cargo.toml`'s and `CHANGELOG.md`'s should be up-to-date, let me know if I've missed anything.

opentelemetry-semantic-conventions/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "opentelemetry-semantic-conventions"
-version = "0.15.0"
+version = "0.16.0"
```
lalitb, line 3: version bump is done once before crate publishing, we can revert this change for now.

opentelemetry-zipkin/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "opentelemetry-zipkin"
-version = "0.21.0"
+version = "0.22.0"
```
lalitb, line 3: version bump is done once before crate publishing, we can revert this change for now.

opentelemetry-zipkin/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## vNext
 
+## v0.22.0
+
+### Changed
+
+- Update `opentelemetry-semantic-conventions` dependency version to 0.16
```
lalitb, line 9: This changelog can be removed for now (see my comments below)

opentelemetry-zipkin/Cargo.toml
```diff
@@ -31,7 +31,7 @@ once_cell = { workspace = true }
 opentelemetry = { version = "0.23", path = "../opentelemetry" }
 opentelemetry_sdk = { version = "0.23", path = "../opentelemetry-sdk", features = ["trace"] }
 opentelemetry-http = { version = "0.12", path = "../opentelemetry-http" }
-opentelemetry-semantic-conventions = { version = "0.15", path = "../opentelemetry-semantic-conventions" }
+opentelemetry-semantic-conventions = { version = "0.16", path = "../opentelemetry-semantic-conventions" }
```
lalitb, line 34: Keep it "0.15" as of now. This version here would be bumped before publish.

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
lalitb: Thanks @tbrockman yes it should have been `attribute.deprecated`. The variable seems to be created correctly now.

lalitb:
> @lalitb believe all related `Cargo.toml`'s and `CHANGELOG.md`'s should be up-to-date, let me know if I've missed anything.

@tbrockman Just few nit comments, should be good to go otherwise. Thanks for the contribution.

opentelemetry-zipkin/Cargo.toml
lalitb: Unrelated to this PR - Zipkin exporter (just like other crates in this repo) shouldn't be using variables from semconv crate till it is stable. Will raise a separate PR for the cleanup.

tbrockman:
@lalitb makes sense, I wasn't entirely sure what the release process entailed. Happy to contribute, thanks for making the process quick and painless 😬

opentelemetry-semantic-conventions/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## vNext
 
+## v0.16.0
```
lalitb, line 5: nit- remove version from here too, and keep them under `vNext`.

lalitb (approved):
Thanks @tbrockman . One nit comment.

PR #1674, opened 2024-04-22
Upgrade to hyper/http v1.0

aumetra:
Fixes #1427

Blocked by:
- https://github.com/hyperium/tonic/pull/1670
- https://github.com/tokio-rs/axum/pull/2693

## Changes

- Update all hyper and http dependencies to v1.x
- Update reqwest to v0.12
- Remove isahc since it's still stuck on http v0.2

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1924, opened 2024-07-10
Remove `isahc` support

aumetra:
Removes `isahc` as part of the http v1 upgrade (see <https://github.com/open-telemetry/opentelemetry-rust/pull/1674#discussion_r1668348354>)

## Changes

**Breaking** change: Remove the `isahc` client from `opentelemetry-http`. Progress on that client has been pretty much non-existent lately and is therefore stuck on http v0.2.

In preparation for the upgrade, we need to remove that client first since supporting it, while possible, would add a maintainance burden in form of the compatibility layer.

Also the author has mentioned that they are attempting to pick up development again, meaning as soon as it reaches support for v1, it can be added again (it's not a lot of code tbh)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ futures-executor = "0.3"
 futures-util = { version = "0.3", default-features = false }
 hyper = { version = "0.14", default-features = false }
 http = { version = "0.2", default-features = false }
-isahc = { version = "1.4", default-features = false }
 log = "0.4.21"
 once_cell = "1.13"
 ordered-float = "4.0"
--- a/.cspell.json
+++ b/.cspell.json
@@ -37,7 +37,6 @@
         "Dirkjan",
         "EPYC",
         "hasher",
-        "isahc",
         "Isobel",
         "jaegertracing",
         "Kühle",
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -19,13 +19,13 @@ debug = 1
 async-std = "1.10"
 async-trait = "0.1"
 bytes = "1"
+cc = "=1.0.105" # pinning version supporting rustc 1.65
```
mladedav, line 22: Why is this needed?

I think it's acceptable that msrv does not work with latest dependencies' versions if that's the reason. But preventing higher versions here could lead to resolution problems if any other library in someone's dependency tree uses newer `cc`.

aumetra: Fair but the CI won't finish otherwise. We either need to adjust the CI or keep this

aumetra: Well, good enough. I also felt free to remove the `url` pinning (for the same reason)

djc (approved):
LGTM!

lalitb (approved):
LGTM. Thanks.

TommyCpp (approved):

PR #1928, opened 2024-07-11
Insert tracing event name into LogRecord::event_name instead of attributes

lalitb:
## Changes
`event_name` field was added in `LogRecord` as part of #1488, however it was never populated with the `name` metadata in tracing appender. The `name` was instead added to the attributes. Ensuring this is now populated properly.  
- This may need changes in the custom exporters to serialize this field directly from `LogRecord::event_name` instead of iterating over the attributes.
- The downstreams depending on the OTLP transformation will no longer receive this as attribute. We can separately decide if OTLP transformation code should add it as attribute (if not already existing). 


cijothomas (approved):
Please add a changelog entry, as this breaking behavior.
Are you planning to modify OTLPExporter to sent this as attribute  (with same name as before), so this change will be transparent to the OTLPExporter users?

lalitb:
> Please add a changelog entry, as this breaking behavior.
Are you planning to modify OTLPExporter to sent this as attribute (with same name as before), so this change will be transparent to the OTLPExporter users?

Added changelog. 
Regarding OTLP Exporter, we need to iterate over attributes (we don't do right now) to add if not already existing.  Should we really do that, or let the users invoke a custom processor to handle that before calling batch processor if they need this field?

cijothomas:
> Added changelog.
> Regarding OTLP Exporter, we need to iterate over attributes (we don't do right now) to add if not already existing

OTLP Should always add it to attributes. Why the condition "add if not already existing"?

lalitb:
> OTLP Should always add it to attributes. Why the condition "add if not already existing"?

info!("message", name = "my_name")
info!(name: "event_name1", "message", name = "some_other_name")

Should we let the OTLP attribute vector have both of these fields? Then we don't need to iterate, just add.

lalitb:
> In above scenario, there would be entry for LogRecord::event_name and also with "name" key in attributes. Should we let the OTLP attribute vector have both of these fields? Then we don't need to iterate, just add.

Thinking again, we can keep the existing behavior - to have two entries with "name" key in the attributes vector in these specific scenarios - which means we can just add without iterating to check.

cijothomas:
> > OTLP Should always add it to attributes. Why the condition "add if not already existing"?
> 
> ```rust
> info!("message", name = "my_name")
> info!(name: "event_name1", "message", name = "some_other_name")
> ```
> 
> In above scenario, there would be entry for LogRecord::event_name and also with "name" key in attributes. Should we let the OTLP attribute vector have both of these fields? Then we don't need to iterate, just add.

yes please. (If the user does `error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io", name = "1", name = "2", name = 3);` we'll have name repeated 4 times.)

lalitb:
> yes please. (If the user does error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io", name = "1", name = "2", name = 3); we'll have name repeated 4 times.)

Make sense. Done, along with updating the changelog. Also triggered integration test to validate.

opentelemetry-appender-tracing/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - [1869](https://github.com/open-telemetry/opentelemetry-rust/pull/1869) Utilize the `LogRecord::set_target()` method to pass the tracing target to the SDK.
   Exporters might use the target to override the instrumentation scope, which previously contained "opentelemetry-appender-tracing".
 
+- **Breaking** [1928](https://github.com/open-telemetry/opentelemetry-rust/pull/1928) Insert tracing event name into LogRecord::event_name instead of attributes.
+   - Custom exporters must now serialize this field directly from LogRecord::event_name instead of iterating over the attributes.
```
cijothomas, line 9:
```suggestion
   - If using a custom exporter, then they must serialize this field directly from LogRecord::event_name instead of iterating over the attributes. OTLP Exporter is modified to handle this.
```

opentelemetry-proto/src/transform/logs.rs
```diff
@@ -92,11 +92,22 @@ pub mod tonic {
                 severity_number: severity_number.into(),
                 severity_text: log_record.severity_text.map(Into::into).unwrap_or_default(),
                 body: log_record.body.map(Into::into),
-                attributes: log_record
-                    .attributes
-                    .map(Attributes::from_iter)
-                    .unwrap_or_default()
-                    .0,
+                attributes: {
```
cijothomas, line 95: can you fix stdout exporter too?

lalitb: done

PR #1935, opened 2024-07-15
Fix Stress Test

utpilla:
## Changes
- Fixed the Stress to not double count throughput

Here's a comparison of the stress test results:

| Test    | main branch              | Current PR               |
|---------|--------------------------|--------------------------|
|[Logs](https://github.com/open-telemetry/opentelemetry-rust/blob/5922205d07e10db56cbb4731eebbcbea14500b51/stress/src/logs.rs#L52-L54)     |62,050,400 iterations/sec |31,245,400 iterations/sec |
|[Metrics](https://github.com/open-telemetry/opentelemetry-rust/blob/5922205d07e10db56cbb4731eebbcbea14500b51/stress/src/metrics_counter.rs#L43-L66)|19,652,600 iterations/sec |9,550,400 iterations/sec  |
|[Traces](https://github.com/open-telemetry/opentelemetry-rust/blob/5922205d07e10db56cbb4731eebbcbea14500b51/stress/src/traces.rs#L54-L64)     |13,256,000 iterations/sec |6,496,600 iterations/sec  |

### DIFF
--- a/stress/src/logs.rs
+++ b/stress/src/logs.rs
@@ -1,9 +1,9 @@
 /*
     Stress test results:
-    OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
-    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
-    69 M/sec
+    ~31 M/sec
 */
 
 use opentelemetry_appender_tracing::layer;
--- a/stress/src/metrics_counter.rs
+++ b/stress/src/metrics_counter.rs
@@ -1,9 +1,9 @@
 /*
     Stress test results:
-    OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
-    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
-    35 M /sec
+    ~9.5 M /sec
 */
 
 use lazy_static::lazy_static;
--- a/stress/src/metrics_histogram.rs
+++ b/stress/src/metrics_histogram.rs
@@ -1,9 +1,9 @@
 /*
     Stress test results:
-    OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
-    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
-    4.6M /sec
+    ~1.8 M/sec
 */
 
 use lazy_static::lazy_static;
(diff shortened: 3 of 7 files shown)


stress/src/throughput.rs
```diff
@@ -69,8 +69,8 @@ where
     let worker_stats_shared_monitor = Arc::clone(&worker_stats_shared);
 
     let handle_main_thread = thread::spawn(move || {
-        let mut start_time = Instant::now();
-        let mut end_time = start_time;
+        let mut last_collect_time = Instant::now();
+        let mut current_time: Instant;
```
lalitb, line 73: nit - is this required?

lalitb, line 73: nit - do we need to declare it here, or can be done inside loop?

utpilla: I have moved the `current_time` variable inside the loop.

> I believe cost of creating is negligible if we are trying to avoid that.

Yeah, since we would always initialize it to `Instant::now()`, most likely the compiler should optimize it.

stress/src/throughput.rs
```diff
@@ -79,12 +78,14 @@ where
         let mut system = System::new_all();
 
         loop {
-            let elapsed = end_time.duration_since(start_time).as_secs();
+            let current_time = Instant::now();
+            let elapsed = current_time.duration_since(last_collect_time).as_secs();
             if elapsed >= SLIDING_WINDOW_SIZE {
```
ThomsonTan, line 83: Is this check still needed? This seems always true because the 5 secnods sleep is included.

lalitb: yes, it's unclear to me too why we sleep more than SLIDING_WINDOW_SIZE. Can possibly be other way.

utpilla: There are quite a few simplifications that can be made to the Stress Test. I intend to cover them in a separate PR and just fix the calculation error in this PR for easier review. Would you want me to include other changes in this PR itself?

lalitb: Thanks, I think this PR is important to get the correct throughput for now. Rest can follow.

lalitb (approved):

stress/src/logs.rs
```diff
@@ -1,9 +1,9 @@
 /*
     Stress test results:
-    OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
-    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
-    69 M/sec
+    ~31 M/sec
```
cijothomas, line 6: I am seeing different numbers with the AMD hardware! (its exactly half of the current in all cases).

Maybe, stick with same hardware which @lalitb and myself has , so the numbers are comparable easily.

stress/src/metrics_counter.rs
```diff
@@ -1,9 +1,9 @@
 /*
     Stress test results:
-    OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
-    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
-    35 M /sec
+    ~9.5 M /sec
```
cijothomas, line 6: ~17 in AMD

stress/src/metrics_histogram.rs
```diff
@@ -1,9 +1,9 @@
 /*
     Stress test results:
-    OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
-    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+    OS: Ubuntu 22.04.4 LTS (5.15.153.1-microsoft-standard-WSL2)
+    Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
-    4.6M /sec
+    ~1.8 M/sec
```
cijothomas, line 6: ~3 in AMD

PR #1955, opened 2024-07-23
Fix PartialEq for Value in opentelemetry-stdout

### LINKED ISSUE #1952: [Bug]: Stack overflow in opentelemetry-stdout
### What happened?

When running with opentelemetry-stdout exporter, there are reliable stack overflows. This is running with fastrace 0.6, but it looks like a fairly clear bug/code loop in the OT exporter.

The code that is highlighted in lldb is in https://github.com/open-telemetry/opentelemetry-rust/blob/a1f02faf7c3a4f25f0ccebbc7e0cb8ccf1e80d82/opentelemetry-stdout/src/common.rs#L87:
```rust
impl PartialEq for Value {
    fn eq(&self, other: &Self) -> bool {
        match (&self, &other) {
            (Value::Double(f), Value::Double(of)) => OrderedFloat(*f).eq(&OrderedFloat(*of)),
            (non_double, other_non_double) => non_double.eq(other_non_double),
        }
    }
}
```

This looks like an infinite loop?

### API Version

opentelemetry 0.23
opentelemetry-stdout 0.4.0

### SDK Version

0.23

N/A

### Relevant log output

```shell
When running the program, I get this error:

thread 'fastrace-global-collector' has overflowed its stack
fatal runtime error: stack overflow
```

Running with lldb:
```
* thread #12, name = 'fastrace-global-collector', stop reason = EXC_BAD_ACCESS (code=2, address=0x17127bfa8)
    frame #0: 0x0000000100160938 service`_$LT$opentelemetry_stdout..common..Value$u20$as$u20$core..cmp..PartialEq$GT$::eq::h655d1b993f78f9a1(self=<unavailable>, other=<unavailable>) at common.rs:99
   96  	}
   97  	
   98  	impl PartialEq for Value {
-> 99  	   fn eq(&self, other: &Self) -> bool {
   100 	       match (&self, &other) {
   101 	           (Value::Double(f), Value::Double(of)) => OrderedFloat(*f).eq(&OrderedFloat(*of)),
   102 	           (non_double, other_non_double) => non_double.eq(other_non_double),
```
```

### PR DESCRIPTION
**Pull request by fiadliel**
Fixes #1952 

## Changes

PartialEq's previous implementation could call itself again with the same arguments. Here the underlying values are extracted and compared. If the enum types are different, false is returned.

### DIFF
--- a/opentelemetry-stdout/src/common.rs
+++ b/opentelemetry-stdout/src/common.rs
@@ -86,9 +86,21 @@ pub(crate) enum Value {
 
 impl PartialEq for Value {
     fn eq(&self, other: &Self) -> bool {
-        match (&self, &other) {
+        match (self, other) {
+            (Value::Bool(b), Value::Bool(ob)) => b.eq(ob),
+            (Value::Int(i), Value::Int(oi)) => i.eq(oi),
             (Value::Double(f), Value::Double(of)) => OrderedFloat(*f).eq(&OrderedFloat(*of)),
-            (non_double, other_non_double) => non_double.eq(other_non_double),
+            (Value::String(s), Value::String(os)) => s.eq(os),
+            (Value::Array(a), Value::Array(oa)) => a.eq(oa),
+            (Value::KeyValues(kv), Value::KeyValues(okv)) => kv.eq(okv),
+            (Value::BytesValue(b), Value::BytesValue(ob)) => b.eq(ob),
+            (Value::Bool(_), _) => false,
+            (Value::Int(_), _) => false,
+            (Value::Double(_), _) => false,
+            (Value::String(_), _) => false,
+            (Value::Array(_), _) => false,
+            (Value::KeyValues(_), _) => false,
+            (Value::BytesValue(_), _) => false,
         }
     }
 }

### CODE REVIEW

File: opentelemetry-stdout/src/common.rs [hunk 1]
```diff
@@ -86,9 +86,21 @@ pub(crate) enum Value {
...
+        match (self, other) {
+            (Value::Bool(b), Value::Bool(ob)) => b.eq(ob),
+            (Value::Int(i), Value::Int(oi)) => i.eq(oi),
             (Value::Double(f), Value::Double(of)) => OrderedFloat(*f).eq(&OrderedFloat(*of)),
-            (non_double, other_non_double) => non_double.eq(other_non_double),
+            (Value::String(s), Value::String(os)) => s.eq(os),
+            (Value::Array(a), Value::Array(oa)) => a.eq(oa),
+            (Value::KeyValues(kv), Value::KeyValues(okv)) => kv.eq(okv),
+            (Value::BytesValue(b), Value::BytesValue(ob)) => b.eq(ob),
+            (Value::Bool(_), _) => false,
```
utpilla, line 97: Use a simpler wild card pattern instead?

```rust
_ => false,
```

fiadliel: My concern was that if a new variant were added to `Value`, `PartialEq` would compile but have an incorrect result. With this version, there would be a compile error. But if you favor conciseness, I can change.

utpilla: Good point!

lalitb (approved):

Issue #1920, opened 2024-07-09
LogRecord Attributes - avoid heap allocations


cijothomas:
We recently decided to [pre-allocate up to 8](https://github.com/open-telemetry/opentelemetry-rust/pull/1908) attributes, to avoid reallocation/copy costs, but attributes are still always on the heap. 
Opening an issue to track improving this, so that, up to 8 attributes can be stack allocated, and then fallback to heap for more attributes. This, along with https://github.com/open-telemetry/opentelemetry-rust/issues/1919, allow one to write logs without any allocation when using synchronous exporters like etw/user_events that require no batching.

We could add a dependency on [SmallVec](https://docs.rs/smallvec/) or make a home-grown solution. For an example implementation, [.NET's `TagList`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/TagList.netcore.cs) can be referred, which optimizes for up to 8 attributes.

We may need a feature flag to control this, as some scenarios (eg: Kernel mode) has less stack available.

lalitb:
fixed in #1965 . The feature flag as discussed in #2055 would be added as required. Will create a separate issue for supporting kernel mode applications.

PR #1913, opened 2024-07-08
[WIP] Box complex types in AnyValue enum

lalitb:
Fixes #1903 
Design discussion issue (if applicable) #

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #1965, opened 2024-07-25
Preallocate storage for log attributes on stack

lalitb:
towards #1920

## Changes

- The GrowableArray is introduced to replace the preallocated vector for storing log attributes, aiming to use stack allocation first and only allocate heap memory when necessary. TBD - Make it optional under feature flag.

The benchmark and stress results as below. A negligible improvement in perf for stress, while benchmark has further improved from 5%-10%.

| Metric                                      | Main Branch                | PR Branch                  | Improvement (%) |
|---------------------------------------------|----------------------------|----------------------------|-----------------|
| Throughput (iterations/sec)                 | 42,239,400 to 44,855,600   | 43,380,200 to 44,410,600   | -0.13%          |
| ot_layer_enabled (ns)                       | 218.92 to 219.50           | 206.11 to 207.56           | 5.57%           |
| Logging_Comparable_To_Appender (ns)         | 96.679 to 96.822           | 85.466 to 85.646           | 11.51%          |
| log_noop_processor (ns)                     | 101.78 to 102.16           | 86.313 to 86.503           | 15.23%          |
| log_cloning_processor (ns)                  | 184.26 to 184.57           | 177.25 to 177.56           | 3.80%           |
| log_clone_and_send_to_channel_processor (ns)| 648.33 to 657.47           | 668.22 to 670.25           | -2.36%          |
| LogExporterWithFuture (ns)                  | 130.25 to 130.58           | 133.11 to 133.64           | -2.24%          |
| LogExporterWithoutFuture (ns)               | 98.344 to 99.553           | 90.973 to 91.901           | 7.45%           |


opentelemetry-sdk/src/logs/record.rs
```diff
@@ -34,7 +44,7 @@ pub struct LogRecord {
     pub body: Option<AnyValue>,
 
     /// Additional attributes associated with this record
-    pub attributes: Option<Vec<(Key, AnyValue)>>,
+    pub attributes: AttributesGrowableArray,
```
cijothomas, line 47: Does this force `GrowableArray` to be public? Can we avoid that?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -1,10 +1,20 @@
+use super::growable_array::GrowableArray;
 use opentelemetry::{
     logs::{AnyValue, Severity},
     trace::{SpanContext, SpanId, TraceFlags, TraceId},
     Key,
 };
 use std::{borrow::Cow, time::SystemTime};
 
+// According to a Go-specific study mentioned on https://go.dev/blog/slog,
+// up to 5 attributes is the most common case. We have chosen 8 as the default
```
cijothomas, line 10: i am inclined to stick with 5 itself.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -1,10 +1,20 @@
...
+// capacity for attributes to avoid reallocation in common scenarios.
+const PREALLOCATED_ATTRIBUTE_CAPACITY: usize = 8;
+
+/// A vector of `Option<(Key, AnyValue)>` with default capacity.
+pub type AttributesGrowableArray =
+    GrowableArray<Option<(Key, AnyValue)>, PREALLOCATED_ATTRIBUTE_CAPACITY>;
```
cijothomas, line 16: `Option<(Key, AnyValue)>` - Why `Option` is needed?

opentelemetry-sdk/src/logs/record.rs
lalitb: `GrowableArray` is public for now. We can make it private, and only keep it's iterators public, which can be used by the exporter.

opentelemetry-sdk/src/logs/record.rs
lalitb: While preallocating an array, we need to store some default value (as in Rust we can't have uninitialised variables). And  `AnyValue` being heterogenous doesn't have a clear default state, so it doesn't implement the `Default` trait. Which makes it difficult for us to create pre-allocated array of `(Key, AnyValue)` tuple. `Option` is internally enum, with `None` as default state, along with allocating the enough room for maximum storage needed for  `(Key, AnyValue)`.

opentelemetry-sdk/src/logs/record.rs
lalitb: I should be fine with either of them. Will make it 5.

opentelemetry-sdk/src/logs/record.rs
cijothomas: We'll need to lookat boxing the complex types to keep enum size under control, now that we are preallocated/stack-allocating them.

opentelemetry-sdk/src/logs/record.rs
cijothomas: Yes please. We certainly don't want to leak any implementation information public.

opentelemetry-sdk/src/logs/record.rs
lalitb: yes exactly my thoughts. Need to revisit #1913 from size prospective, if not perf.

lalitb:
Have also included the Boxing of AnyValue enum as part of this PR, as that overall affects the size of the preallocated buffer, and is a small change. There is an improvement with the bench, but stress throughput has gone dropped. Need to look into that.

**_full-log-with-4-attributes/no-context:_**
main:  ~170 ns
PR: 132 ns


**_stress test_**:
main: 35,526,000 iterations/sec
PR: 21,113,200 iterations/sec

Need to explore that further.

lalitb:
ok with latest updates:

**opentelemetry_sdk/benches**:
main:
full-log-with-4-attributes/no-context
                        time:   [204.50 ns 207.92 ns 211.58 ns]

PR:
full-log-with-4-attributes/no-context
                        time:   [129.55 ns 135.27 ns 142.12 ns]


**opentelemetry-appender-tracing benches:**
main:
ot_layer_enabled        time:   [476.21 ns 496.47 ns 518.38 ns]

PR:
ot_layer_enabled        time:   [454.39 ns 467.50 ns 482.72 ns]


**stress**: (no of threads: 8)
main:
Throughput: 5,707,600 iterations/sec
Throughput: 5,745,000 iterations/sec
Throughput: 4,936,400 iterations/sec
Throughput: 5,634,600 iterations/sec
Throughput: 6,127,000 iterations/sec

PR:
Throughput: 6,331,000 iterations/sec
Throughput: 6,913,800 iterations/sec
Throughput: 5,927,800 iterations/sec
Throughput: 7,158,400 iterations/sec
Throughput: 7,184,000 iterations/sec
Throughput: 6,085,400 iterations/sec

cijothomas:
> Have also included the Boxing of AnyValue enum as part of this PR, as that overall affects the size of the preallocated buffer, and is a small change. There is an improvement with the bench, but stress throughput has dropped drastically. Need to look into that, along with memory stats.
> 
> **_full-log-with-4-attributes/no-context:_** main: ~170 ns PR: 132 ns
> 
> **_stress test_**: main: 35,526,000 iterations/sec PR: 21,113,200 iterations/sec

@utpilla and myself has observed that some stress tests perform poorer, when making some improvements to Metrics SDK!! (benchmarks show good improvement). The changes are logically improvements, but the stress test somehow ends up showing lower...

lalitb:
> @utpilla and myself has observed that some stress tests perform poorer, when making some improvements to Metrics SDK!! (benchmarks show good improvement). The changes are logically improvements, but the stress test somehow ends up showing lower...

Ok, in my case, I had [experimental_metadata_attributes](https://github.com/open-telemetry/opentelemetry-rust/blob/7a9076032d270ae283c0549c7ef2f03833bbd2e5/opentelemetry-appender-tracing/Cargo.toml#L33) left enabled in my PR, which resulted in more attributes getting added to the log_record, and eventually slowing the stress test. But agree, the scale of improvements observed in benchmark sometimes doesn't reflect back in the stress.

opentelemetry-sdk/Cargo.toml
```diff
@@ -51,6 +51,9 @@ testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-async-std",
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
+memory-profiling = []
```
utpilla, line 54: Do we want this to be a crate feature?

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
+use std::array::IntoIter;
+
+/// The default max capacity for the stack portation of `GrowableArray`.
```
utpilla, line 3:
```suggestion
/// The default max capacity for the stack portion of `GrowableArray`.
```

opentelemetry-sdk/Cargo.toml
lalitb: Good catch, leftover of mem stats I though of adding earlier. Removed now.

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
...
+/// A hybrid vector that starts with a fixed-size array and grows dynamically with a vector.
+///
+/// `GrowableArray` uses an internal fixed-size array (`initial`) for storing elements until it reaches
+/// `MAX_STACK_CAPACITY`. When this capacity is exceeded, additional elements are stored in a heap-allocated
+/// vector (`additional`). This structure allows for efficient use of stack memory for small numbers of elements,
+/// while still supporting dynamic growth.
+///
+pub(crate) struct GrowableArray<
+    T: Default + Clone + PartialEq,
+    const MAX_STACK_CAPACITY: usize = DEFAULT_MAX_STACK_CAPACITY,
```
utpilla, line 18: I would be in favor of removing these `const` generics for simplicity.

opentelemetry-sdk/Cargo.toml
```diff
@@ -91,3 +91,8 @@ required-features = ["metrics"]
 name = "log"
 harness = false
 required-features = ["logs"]
+
+#[[bench]]
```
utpilla, line 95: Remove this as well?

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
...
+    /// Pushes a value into the `GrowableArray`.
+    ///
+    /// If the internal array (`initial`) has reached its capacity (`MAX_STACK_CAPACITY`), the value is pushed
+    /// into the heap-allocated vector (`additional`). Otherwise, it is stored in the array.    #[inline]
+    pub(crate) fn push(&mut self, value: T) {
+        if self.count < MAX_STACK_CAPACITY {
+            self.initial[self.count] = value;
+            self.count += 1;
+        } else {
+            if self.additional.is_none() {
```
utpilla, line 62: It might be more efficient to simply `match` on `self.additional.as_mut()`:

```rust
match self.additional.as_mut() {
    Some(v) => v.push(value),
    None => v = Some(Vec::with_capacity(INITIAL_VEC_CAPACITY)),
}
```

Using `self.additional.is_none()` and `self.additional.as_mut().unwrap()` would lead to more match statements anyway, I guess.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: I tried the rust ideomatic way of specifing the size as compile time constants, taking _inspiration_ from rust-lang array, and `smallvec`. But happy to change if this looks complex. What do you suggest?

utpilla: Since this we are not marking this data structure public and our usage of it doesn't involve dynamic configuration, I think we should keep the code simple and less for easier maintenance.

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
...
+    ///
+    /// The iterator yields elements from the internal array (`initial`) first, followed by elements
+    /// from the vector (`additional`) if present. This allows for efficient iteration over both
+    /// stack-allocated and heap-allocated portions.
+    ///
+    #[inline]
+    pub(crate) fn iter(&self) -> GrowableArrayIter<'_, T, MAX_STACK_CAPACITY> {
+        if self.additional.is_none() || self.additional.as_ref().unwrap().is_empty() {
+            GrowableArrayIter::StackOnly {
+                iter: self.initial.iter().take(self.count),
```
utpilla, line 99: Do we need to use `take`? It should have the same behavior as plain `iter()`, right?

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: Are you suggesting replacing the `const` generics with the `const` variable?

```rust
const MAX_STACK_CAPACITY = 5;
const INITIAL_VEC_CAPACITY =10;
pub(crate) struct GrowableArray
{
    initial: [T; MAX_STACK_CAPACITY ],
    additional: Option<Vec<T>>,
    count: usize,
}
```

My initial thoughts were that this would allow flexibility if we want to introduce the feature-flag, for heap-only allocation requirements:
```rust
#[cfg(feature = "heap-only")]
pub(crate) type AttributesGrowableArray =
    GrowableArray<Option<(Key, AnyValue)>, 0, 10>;
#[cfg(not(feature = "heap-only"))]
pub(crate) type AttributesGrowableArray =
    GrowableArray<Option<(Key, AnyValue)>, 10>;
```

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: Thanks, this looks better. Will update.

opentelemetry-sdk/Cargo.toml
lalitb: Good catch. Removed now :)

opentelemetry-sdk/src/logs/growable_array.rs
utpilla: > Are you suggesting replacing the const generics with the const variable?

Yes.

> My initial thoughts were that this would allow flexibility if we want to introduce the feature-flag, for (say) heap-only allocation requirements:

Is there any demand for heap-only allocation? Even if it's there, that should still be possible without using `const` generics, I think. We could simply update the `const` value. Would something like this work?
```rust
const MAX_STACK_CAPACITY: usize = match () {
    #[cfg(feature = "heap-only")]
    () => 0,
    _ => 10
  };

const INITIAL_VEC_CAPACITY : usize = match () {
    #[cfg(feature = "heap-only")]
    () => 10,
    _ => 0,
  };
```

lalitb: > Would something like this work?

This will work, but then we need to add more such constants if we want to later use this array in trace/event attributes with different size/feature requirements. The existing approach provides a cleaner way to specify these constants at the place of creation. I do agree `const` variable is a much simpler approach if we don't plan to use this array at other places. We can revisit it later and change it according?

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
...
+            GrowableArrayIntoIter::Mixed {
+                stack_iter,
+                heap_iter,
+            } => stack_iter.next().or_else(|| heap_iter.next()),
+        }
+    }
+}
+
+// Implement `IntoIterator` for a reference to `GrowableArray`
+impl<'a, T: Default + Clone + PartialEq + 'a, const INITIAL_CAPACITY: usize> IntoIterator
```
utpilla, line 171: Could we remove this? Anyone trying to iterate over references of the attributes could simply use `.iter()`, right?

```diff
    #[test]
    fn test_ref_iter() {
        let mut collection = GrowableArray::<i32>::new();
        for i in 0..15 {
            collection.push(i);
        }
-        let iter = &collection;
+        let iter = collection.iter();
        let mut count = 0;
        for value in iter {
            assert_eq!(*value, count);
            count += 1;
        }
        assert_eq!(count, 15);
    }
```

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
...
+    fn test_len() {
+        let mut collection = GrowableArray::<i32>::new();
+        for i in 0..15 {
+            collection.push(i);
+        }
+        assert_eq!(collection.len(), 15);
+    }
+
+    #[test]
+    fn test_into_iter() {
```
utpilla, line 268: It would be nice to test this with different lengths each representing a unique state of the `GrowableArray`:

1. Empty attributes
2. When the number of attributes is less than the max stack capacity
3. When the number of attributes exactly matches the max stack capacity
4. When the number of attributes exceeds stack capacity but not the initial vec capacity
5. When the number of attributes exceeds stack capacity and also the initial vec capacity

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,323 @@
...
+    },
+}
+
+impl<T: Default + Clone + PartialEq, const INITIAL_CAPACITY: usize> Iterator
+    for GrowableArrayIntoIter<T, INITIAL_CAPACITY>
+{
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self {
```
utpilla, line 160: It's possible to avoid creating an `enum` and matching on its variants on every `next()` call. We could have the required iterator transform ready in `into_iterator` so we don't have to keep checking the presence of elements in vector each time. I tried something like this:
```rust
// Implement `iter` method for `GrowableArray`
#[inline]
pub(crate) fn iter(&self) -> impl Iterator<Item = &T> {
    if self.additional.is_none() || self.additional.as_ref().unwrap().is_empty() {
        self.initial.iter().take(self.count).chain([].iter()) // Chaining with an empty array
         // so that both `if` and `else` branch return the same type
    } else {
        self.initial.iter().take(self.count).chain(self.additional.as_ref().unwrap().iter())
    }
}
```


```rust
// Implement `IntoIterator` for `GrowableArray`
impl<T: Default + Clone + PartialEq, const INITIAL_CAPACITY: usize> IntoIterator
    for GrowableArray<T, INITIAL_CAPACITY>
{
    type Item = T;
    type IntoIter = GrowableArrayIntoIter<T, INITIAL_CAPACITY>;

    fn into_iter(self) -> Self::IntoIter {
        GrowableArrayIntoIter::<T, INITIAL_CAPACITY>::new(self)
    }
}

#[derive(Debug)]
pub(crate) struct GrowableArrayIntoIter<T: Default + Clone + PartialEq, const INITIAL_CAPACITY: usize>
{
    iter: std::iter::Chain<std::iter::Take<std::array::IntoIter<T, INITIAL_CAPACITY>>, std::vec::IntoIter<T>>,
}

impl<T: Default + Clone + PartialEq, const INITIAL_CAPACITY: usize> GrowableArrayIntoIter<T, INITIAL_CAPACITY> {
    fn new(source: GrowableArray<T, INITIAL_CAPACITY>) -> Self{
        Self {
            iter: Self::get_iterator(source)
        }
    }

    fn get_iterator(source: GrowableArray<T, INITIAL_CAPACITY>) -> std::iter::Chain<std::iter::Take<std::array::IntoIter<T, INITIAL_CAPACITY>>, std::vec::IntoIter<T>> {
        if source.additional.is_none() || source.additional.as_ref().unwrap().is_empty() {
            source.initial.into_iter().take(source.count).chain(Vec::<T>::new().into_iter())
        } else {
            source.initial.into_iter().take(source.count).chain(source.additional.unwrap().into_iter())
        }
    }
}

impl<T: Default + Clone + PartialEq, const INITIAL_CAPACITY: usize> Iterator for GrowableArrayIntoIter<T, INITIAL_CAPACITY> {
    type Item = T;

    fn next(&mut self) -> Option<Self::Item> {
        self.iter.next()
    }
}
```

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: Thanks, added tests.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: Very nice. Thanks for trying this out. This improves the iteration cost. Have incorporated these changes.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: Yes makes sense. We can always bring this back later if needed.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: ok looking again, I think this has to be actually changed as:
```
        match self.additional.as_mut() {
            Some(v) => v.push(value),
            None => {
                let mut vec = Vec::with_capacity(INITIAL_VEC_CAPACITY);
                vec.push(value);
                self.additional = Some(vec);
            }
        }
```

while `match` is more rust idiomatic, in this case, if-else looks easier to follow. We can revisit it if any performance issues are observed.

lalitb:
@utpilla Thanks for reviewing this in depth, very good comments. Have tried addressing them all. Let me know if you still see any issues.

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,369 @@
...
+        for value in iter {
+            assert_eq!(*value, count);
+            count += 1;
+        }
+        assert_eq!(count, 15);
+    }
+
+    #[test]
+    fn test_key_value_pair_storage_growable_array() {
+        let mut collection = GrowableArray::<KeyValuePair>::new();
```
utpilla, line 241: It might be better to test `GrowableArray` with the type that actually gets used by `LogRecord` which is `Option<(Key, AnyValue)>` instead of `struct KeyValuePair(Key, AnyValue)`.

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,369 @@
...
+        let mut collection = GrowableArray::<i32>::new();
+        for i in 0..DEFAULT_MAX_STACK_CAPACITY - 1 {
+            collection.push(i as i32);
+        }
+        assert_eq!(collection.len(), DEFAULT_MAX_STACK_CAPACITY - 1);
+
+        for i in 0..DEFAULT_MAX_STACK_CAPACITY - 1 {
+            assert_eq!(collection.get(i), Some(&(i as i32)));
+        }
+        assert_eq!(collection.get(DEFAULT_MAX_STACK_CAPACITY), None);
```
utpilla, line 289: The next index after inserting elements in the collection here should be `DEFAULT_MAX_STACK_CAPACITY - 1`?

```suggestion
        assert_eq!(collection.get(DEFAULT_MAX_STACK_CAPACITY - 1), None);
```

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,369 @@
...
+        assert_eq!(
+            collection.len(),
+            DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY - 1
+        );
+
+        for i in 0..(DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY - 1) {
+            assert_eq!(collection.get(i), Some(&(i as i32)));
+        }
+        assert_eq!(
+            collection.get(DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY),
```
utpilla, line 333: Same as this [comment](https://github.com/open-telemetry/opentelemetry-rust/pull/1965/files#r1698884942).

utpilla (approved):
Looks good to me! Left a few minor comments about tests.

opentelemetry-sdk/src/logs/growable_array.rs
utpilla: For this test, `DEFAULT_MAX_STACK_CAPACITY - 1` is the first out of bounds index. Shouldn't we test that since we are adding `DEFAULT_MAX_STACK_CAPACITY - 2` elements to the collection?

lalitb: ok I got what you mean. Have also added for `DEFAULT_MAX_STACK_CAPACITY - 1`. `DEFAULT_MAX_STACK_CAPACITY` was to ensure to test past the max size.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: yes, have updated the tests for this. Thanks once again :)

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,370 @@
...
+        assert_eq!(
+            collection.len(),
+            DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY + 5
+        );
+
+        for i in 0..(DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY + 5) {
+            assert_eq!(collection.get(i), Some(&(i as i32)));
+        }
+        assert_eq!(
+            collection.get(DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY + 6),
```
utpilla, line 360: This one as well.

```suggestion
            collection.get(DEFAULT_MAX_STACK_CAPACITY + DEFAULT_INITIAL_VEC_CAPACITY + 5),
```

opentelemetry/src/logs/record.rs
```diff
@@ -59,11 +59,11 @@ pub enum AnyValue {
     /// A boolean value
     Boolean(bool),
     /// A byte array
-    Bytes(Vec<u8>),
+    Bytes(Box<Vec<u8>>),
```
cijothomas, line 62: if feasible, could we make this in a separate, small PR.
or update the changelog to indicate this change.

opentelemetry-appender-log/src/lib.rs
```diff
@@ -429,7 +429,7 @@ mod any_value {
         }
 
         fn serialize_bytes(self, v: &[u8]) -> Result<Self::Ok, Self::Error> {
-            Ok(Some(AnyValue::Bytes(v.to_owned())))
+            Ok(Some(AnyValue::Bytes(Box::new(v.to_owned()))))
```
cijothomas, line 432: The side effect of boxing is that, this forces a heap allocation which maybe undesirable. I think its best to make the Boxing change separate, so that can be independently tested and reviewed.

opentelemetry-appender-tracing/Cargo.toml
```diff
@@ -33,7 +33,6 @@ pprof = { version = "0.13", features = ["flamegraph", "criterion"] }
 experimental_metadata_attributes = ["dep:tracing-log"]
 logs_level_enabled = ["opentelemetry/logs_level_enabled"]
 
-
```
cijothomas, line 36: nit: this was unintended?

opentelemetry-sdk/Cargo.toml
```diff
@@ -91,3 +91,4 @@ required-features = ["metrics"]
 name = "log"
 harness = false
 required-features = ["logs"]
+
```
cijothomas, line 94: nit: unintentional ?

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,370 @@
+/// The default max capacity for the stack portion of `GrowableArray`.
+const DEFAULT_MAX_STACK_CAPACITY: usize = 10;
```
cijothomas, line 2:
```suggestion
const DEFAULT_MAX_STACK_CAPACITY: usize = 5;
```

Add a link to the study in Golang : https://go.dev/blog/slog so we know where is 5 coming from.

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,370 @@
...
+/// `MAX_STACK_CAPACITY`. When this capacity is exceeded, additional elements are stored in a heap-allocated
+/// vector (`additional`). This structure allows for efficient use of stack memory for small numbers of elements,
+/// while still supporting dynamic growth.
+///
+pub(crate) struct GrowableArray<
+    T: Default + Clone + PartialEq,
+    const MAX_STACK_CAPACITY: usize = DEFAULT_MAX_STACK_CAPACITY,
+    const INITIAL_VEC_CAPACITY: usize = DEFAULT_INITIAL_VEC_CAPACITY,
+> {
+    initial: [T; MAX_STACK_CAPACITY],
```
cijothomas, line 19: nit: initial -> inline, additional -> overflow could better convey the design.

opentelemetry-appender-log/src/lib.rs
lalitb: The boxing was done for types which are already allocated on heap. But we can discuss it independently. Will remove it

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: This is handled separately for LogRecord in record.rs. Didn't want to impose the recommended size of LogRecord attributes here:
```rust
// According to a Go-specific study mentioned on https://go.dev/blog/slog,
// up to 5 attributes is the most common case.
const PREALLOCATED_ATTRIBUTE_CAPACITY: usize = 5;

/// A vector of `Option<(Key, AnyValue)>` with default capacity.
pub(crate) type AttributesGrowableArray =
    GrowableArray<Option<(Key, AnyValue)>, PREALLOCATED_ATTRIBUTE_CAPACITY>;
```

opentelemetry-appender-log/src/lib.rs
lalitb: This is removed. It does affect the bench/memory, have updated them. Would be good to bring it back.

opentelemetry-sdk/Cargo.toml
lalitb: yes it was, removed the newline now.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: yeah true. updated accordingly.

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,370 @@
+/// The default max capacity for the stack portion of `GrowableArray`.
```
TommyCpp, line 1: This is pretty generic and we may use it in other signals right, would it make sense for us to move it out of `logs` and put it in it's own mod?

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// Creates a new `GrowableArray` with the default initial capacity.
+    #[allow(dead_code)]
+    pub(crate) fn new() -> Self {
+        Self::default()
+    }
+
+    /// Pushes a value into the `GrowableArray`.
+    ///
+    /// If the internal array (`inline`) has reached its capacity (`MAX_INLINE_CAPACITY`), the value is pushed
+    /// into the heap-allocated vector (`overflow`). Otherwise, it is stored in the array.    #[inline]
```
TommyCpp, line 54: Why not `#[inline]`?

opentelemetry-sdk/src/logs/growable_array.rs
```diff
@@ -0,0 +1,370 @@
...
+    pub(crate) fn push(&mut self, value: T) {
+        if self.count < MAX_INLINE_CAPACITY {
+            self.inline[self.count] = value;
+            self.count += 1;
+        } else {
+            if self.overflow.is_none() {
+                // Initialize the vector with a specified capacity
+                self.overflow = Some(Vec::with_capacity(INITIAL_OVERFLOW_CAPACITY));
+            }
+            self.overflow.as_mut().unwrap().push(value);
```
TommyCpp, line 64: Can we use [get_or_insert_with](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with) here

lalitb: Yes, that's more idiomatic. Done.

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: Oh, it got hidden in the comments by mistake. fixed now. Thanks

opentelemetry-sdk/src/logs/growable_array.rs
lalitb: I initially thought we would move this code to the global space only if needed, but I think it's more visible and usable there. Done.

opentelemetry-sdk/src/growable_array.rs
```diff
@@ -0,0 +1,374 @@
...
+            Some(&self.inline[index])
+        } else if let Some(ref overflow) = self.overflow {
+            overflow.get(index - MAX_INLINE_CAPACITY)
+        } else {
+            None
+        }
+    }
+
+    /// Returns the number of elements in the `GrowableArray`.
+    #[allow(dead_code)]
```
lalitb, line 84: Seems clippy is not able to recognize the use of these methods(`len`, `push`, and `iter`) when GrowableArray is used through the type alias within the `logs` module. So adding `dead_code` attribute to keep clippy happy for now.

TommyCpp (approved):
Nice work. Thanks for working on this

opentelemetry-appender-log/src/lib.rs
```diff
@@ -621,7 +621,7 @@ mod any_value {
 
         fn end(self) -> Result<Self::Ok, Self::Error> {
             Ok(Some(AnyValue::Map({
-                let mut variant = HashMap::new();
+                let mut variant = HashMap::default();
```
cijothomas, line 624: are these related to this PR?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -186,17 +214,25 @@ mod tests {
         let mut log_record = LogRecord::default();
         let attributes = vec![(Key::new("key"), AnyValue::String("value".into()))];
         log_record.add_attributes(attributes.clone());
-        assert_eq!(log_record.attributes, Some(attributes));
+
+        let mut expected_attributes = AttributesGrowableArray::new();
```
cijothomas, line 218: is it intentional that these tests rely on implementation detail, by relying on GrowableArray?

lalitb: Oh this is not really required, have removed now.

opentelemetry-appender-log/src/lib.rs
lalitb: these were the related to lint error with box'ing change. I didn't revert this while revering box'ing, as both are similar, and `default` is more explicit.

cijothomas (changes requested):
Hold off merging until https://github.com/open-telemetry/opentelemetry-rust/pull/1985 is in, which is a subset of this PR.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -89,33 +98,31 @@ impl opentelemetry::logs::LogRecord for LogRecord {
...
+        self.attributes.len()
     }
 
     #[allow(dead_code)]
-    /// Returns true if the `LogRecord` contains the specified attribute.
+    /// Checks if the `LogRecord` contains the specified attribute.
     pub(crate) fn attributes_contains(&self, key: &Key, value: &AnyValue) -> bool {
-        self.attributes.as_ref().map_or(false, |attrs| {
-            attrs.iter().any(|(k, v)| k == key && v == value)
+        self.attributes.iter().any(|opt| {
```
utpilla, line 120: Would using `Option` methods to achieve this be more idiomatic here?

```rust
self.attributes.iter().any(|opt| {
    opt.as_ref().map_or(false, |(k, v)| k == key && v == value)
})
```

utpilla (approved):

opentelemetry-sdk/src/growable_array.rs
```diff
@@ -0,0 +1,374 @@
+/// The default max capacity for the stack portion of `GrowableArray`.
+const DEFAULT_MAX_INLINE_CAPACITY: usize = 10;
+/// The default initial capacity for the vector portion of `GrowableArray`.
+const DEFAULT_INITIAL_OVERFLOW_CAPACITY: usize = 5;
+
+#[derive(Debug, Clone, PartialEq)]
+/// A hybrid vector that starts with a fixed-size array and grows dynamically with a vector.
+///
+/// `GrowableArray` uses an internal fixed-size array (`initial`) for storing elements until it reaches
```
cijothomas, line 9:
```suggestion
/// `GrowableArray` uses an internal fixed-size array (`inline`) for storing elements until it reaches
```

opentelemetry-sdk/src/growable_array.rs
```diff
@@ -0,0 +1,374 @@
...
+/// `MAX_INLINE_CAPACITY`. When this capacity is exceeded, additional elements are stored in a heap-allocated
+/// vector (`additional`). This structure allows for efficient use of stack memory for small numbers of elements,
```
cijothomas, line 11:
```suggestion
/// vector (`overflow`). This structure allows for efficient use of stack memory for small numbers of elements,
```

opentelemetry-sdk/src/growable_array.rs
TommyCpp: not a strong opinion, but `inline` in Rust mostly refers to the compiler optimization so maybe we can use something else to avoid confusion?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -819,15 +820,21 @@ mod tests {
 
     impl LogProcessor for FirstProcessor {
         fn emit(&self, data: &mut LogData) {
-            // add attribute
-            data.record.attributes.get_or_insert(vec![]).push((
+            // Ensure attributes is initialized if it is empty
```
cijothomas, line 823: can we modify the test processor to do what a user would do, instead of directly relying on implementation internals like GrowableArray?

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -83,14 +82,19 @@ mod tests {
         let attributes: Vec<(Key, AnyValue)> = log
             .record
             .attributes
-            .clone()
-            .expect("Attributes are expected");
+            .iter()
```
cijothomas, line 85: isn't this change fully internal only? Why we are forced to change processors to account for this?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -254,7 +258,11 @@ mod tests {
             severity_text: Some(Cow::Borrowed("ERROR")),
             severity_number: Some(Severity::Error),
             body: Some(AnyValue::String("Test body".into())),
-            attributes: Some(vec![(Key::new("key"), AnyValue::String("value".into()))]),
+            attributes: {
```
cijothomas, line 261: Can we modify the test to do what end users will do to populate attributes, instead of relying on internal impln?

cijothomas (commented):
The change itself is very good. Awesome work and thanks for taking us one more step closer to achieving zero heap logging!

I left suggestion about not modifying test code to rely on the new internal implementation details. I want to fix https://github.com/open-telemetry/opentelemetry-rust/issues/1995, and re-do all benchmarks, to gauge the actual gains from this change (and less variations, hopefully), before merging.

opentelemetry-sdk/src/growable_array.rs
cijothomas: I don't see the point of confusion.. Inline can refer to code being in-lined, or memory being inline (when using structs etc.)... But this is purely internal, so not a concern.

(I think I suggested inline, overflow from .NET runtimes's similar naming. .NET also has the compiler inlining 
optimization!)

opentelemetry-sdk/src/logs/record.rs
lalitb: This is supposed to be unit-test to validate internal constructs. The user can never create the log record by direct assignment as tested here. And the test won't need any changes in future if we change the attributes implementation details.

lalitb:
> The change itself is very good. Awesome work and thanks for taking us one more step closer to achieving zero heap logging!
> 
> I left suggestion about not modifying test code to rely on the new internal implementation details. I want to fix #1995, and re-do all benchmarks, to gauge the actual gains from this change (and less variations, hopefully), before merging.

Thanks @cijothomas for review. Have resolved the tests to not rely on the internal implementation details of the attribute storage. Let me know if you still see any issues.

cijothomas:
@lalitb and I observed that in some machines the perf is getting de-graded (benchmarks and stress tests) with the change. We are trying to understand what might be causing that, even though we greatly reduce heap allocation. Holding off merge until we understand the source.

lalitb:
Adding the relevant part of flamegraph before (main branch) and after (PR branch):
Main branch:
main

PR branch:
staticallocationPR

The cost of execution of ` __memmove_avx_unaligned_erms` function has increased from `1.5%` to `4.0%`.  This is the low-level C function, which typically represents that some data is being physically copied or moved in memory. In our case, the data movement is happening here:

https://github.com/open-telemetry/opentelemetry-rust/blob/d583695d30681ee1bd910156de27d91be3711822/opentelemetry-sdk/src/logs/log_emitter.rs#L276-L279

i.e, `LogRecord` object is moved inside the `LogData` instance. It seems that switching the `LogRecord::attributes` type from a pre-allocated vector to a fixed-size array is changing the behavior of our code in terms of memory operations.

- In the case of an array, the entire contiguous array is moved from one location to another. 
- In the case of vector, only it's metadata fields (pointer, capacity, and length) are moved.

The cost of moving the data (i.e., the array) outweighs the benefit of initial creation and population. This leads to increased latency. The option could be using references or smart pointers (Rc/Arc) over attribute array to mitigate this issue, but they come with their own trade-offs - increased complexity, lifetime management challenge, potential overhead of indirections. 

Also attaching the full flamegraph svg files if someone is interested in doing more analysis :)

Main branch:

flamegraph_ex_main


PR branch:

flamegraph_ex_pr

cijothomas:
Thanks @lalitb for getting to the root cause of the observed behavior! This is similar to what we faced in OTel .NET where using stack allocation to avoid heap (and gc pressure) actually increased hot path cost, due to increase in amount of stack to stack copying! https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/Benchmarks/Metrics/MetricsBenchmarks.cs#L42-L52

utpilla:
> The cost of execution of ` __memmove_avx_unaligned_erms` (shown with arrow above) function has increased from `1.5%` to `4.0%`. This is the low-level C function, which typically represents that some data is being physically copied or moved in memory. In our case, the data movement is happening here:

Great analysis @lalitb!

lalitb:
I have not converted the `LogData` export structure to contain the CoW of the `LogRecord` and `InstrumentationLibrary`:

Before:
```rust
#[derive(Clone, Debug)]
pub struct LogData {
    /// Log record
    pub record: LogRecord,
    /// Instrumentation details for the emitter who produced this `LogEvent`.
    pub instrumentation: InstrumentationLibrary,
}
```

After:
```rust
#[derive(Clone, Debug)]
pub struct LogData<'a> {
    /// Log record, which can be borrowed or owned.
    pub record: Cow<'a, LogRecord>,
    /// Instrumentation details for the emitter who produced this `LogEvent`.
    pub instrumentation: Cow<'a, InstrumentationLibrary>,
}
``` 

This allows us to avoid copying of LogRecord structure when used with a synchronous processor.  
- Simple processor (and also other concurrent processors) gets the reference to the LogData structure, borrows it, and sends it to the exporter. This involves NO copy/move. The exporter can decide whether to own it based on its logic.
- The batch processor gets the reference to the LogData structure, converts it into the owned version, and pushes it to the queue. This involves clone, which effectively copies the attributes too. This can't be avoided.
- 
I can see ~20% improvement with this:

```
Appender:
PR: ot_layer_enabled        time:   [210.69 ns 211.52 ns 212.46 ns]
Main: ot_layer_enabled        time:   [253.75 ns 254.01 ns 254.33 ns]

SDK:
PR:  [104.99 ns 105.09 ns 105.21 ns]
Main: [136.19 ns 136.30 ns 136.41 ns]

Stress:
PR: 
Throughput: 43,832,200 iterations/sec
Throughput: 44,060,200 iterations/sec
Throughput: 44,089,600 iterations/sec
Throughput: 43,246,400 iterations/sec
Throughput: 42,510,000 iterations/sec

Main:
Throughput: 39,540,600 iterations/sec
Throughput: 39,462,800 iterations/sec
Throughput: 39,574,800 iterations/sec
Throughput: 39,052,600 iterations/sec
```

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -274,8 +266,8 @@ impl opentelemetry::logs::Logger for Logger {
         }
 
         let mut data = LogData {
-            record: log_record,
-            instrumentation: self.instrumentation_library().clone(),
+            record: Cow::Borrowed(&log_record),
```
cijothomas, line 269: thinking out loud -- what is the need of a separate `LogData` struct itself? Can we use `LogRecord` itself?

LogRecord struct can add ref (not Arc else we'd need to touch heap again!) to InstrumentationLibrary. In create_log_record() method, the Logger can set the InstrumentationLibrary into the `LogRecord` right at creation time. With this, sdk does not need to create another structure (LogData), and instead can pass `LogRecord` as-is.

(I haven't thought through 'a issues, if any)

lalitb: Need to do more thinking around this. First thoughts are - If LogRecord directly holds a reference to InstrumentationLibrary, we need to carefully manage the lifetimes of these references. This could complicate the API and usage patterns, particularly if InstrumentationLibrary has a shorter lifetime than the LogRecord (eg logger gets destroyed while the LogData is in queue).  Doable, but unless I am overthinking, it seems complex :)

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/log_processor.rs#L9-L10 Can you check the effect on this benchmark with/without latest commit? (Trying to see if our benchmarks are actually helping us make more informed decisions!)

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: If any component (Batchprocessor) needs to queue it/store beyond the 'a, it should clone and store. Just like it does today.

lalitb:
Before:
log_noop_processor      time:   [133.99 ns 134.10 ns 134.23 ns]
log_cloning_processor   time:   [229.78 ns 229.93 ns 230.09 ns]
log_clone_and_send_to_channel_processor  time:   [617.49 ns 618.84 ns 620.54 ns]


After:
log_noop_processor      time:   [133.99 ns 134.10 ns 134.23 ns]
log_cloning_processor   time:   [301.43 ns 301.76 ns 302.09 ns]   
log_clone_and_send_to_channel_processor time:   [697.19 ns 697.86 ns 698.57 ns]

Had to change the log_clone_and_send_to_channel_processor  to clone the logrecord, and instrumentation library, Also, I believe we need to revisit the log_cloning_processor   - We would only clone in the batch, never in simple/concurrent processor.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Yes, but cloning/storing would be difficult with LogRecording containing `a reference. As of now, things are much simpler as LogRecord either has owning data, or Cow.

cijothomas:
I am also seeing same perf results!(42 M/sec in stress test) Thanks for continuously striving to get us better and better!

I just compared with OTel .NET's similar [setup in same machine](https://github.com/open-telemetry/opentelemetry-dotnet/pull/5790#issue-2471252217), and OTel .NET gets around 46 M/sec, so we are getting pretty close to match .NET (which is heavily optimized already and GA for years). With the next set of planned improvements, we should easily exceed that as well.

I requested to extract out the changes to LogData/lifetime into own PR, so this one just has only the change to stack-allocate attributes and nothing more.

lalitb:
The PR now only contains the growable array related changes. Should be ready to merge if there are no further comments. The feature-flag for heap allocated attributes would be added subsequently.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -1,10 +1,19 @@
...
     trace::{SpanContext, SpanId, TraceFlags, TraceId},
     Key,
 };
 use std::{borrow::Cow, time::SystemTime};
 
+// According to a Go-specific study mentioned on https://go.dev/blog/slog,
+// up to 5 attributes is the most common case.
+const PREALLOCATED_ATTRIBUTE_CAPACITY: usize = 5;
+
+/// A vector of `Option<(Key, AnyValue)>` with default capacity.
```
cijothomas, line 13: nit: No longer a vector, so comment is incorrect.

cijothomas (approved):
LGTM. Left a nit about benchmark results not in sync with the PR desc.
A combined changelog highlighting all perf improvements could be added in a follow up?

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -10,7 +10,7 @@
     | noop_layer_disabled         | 12 ns       |
     | noop_layer_enabled          | 25 ns       |
     | ot_layer_disabled           | 19 ns       |
-    | ot_layer_enabled            | 250 ns      |
+    | ot_layer_enabled            | 219 ns      |
```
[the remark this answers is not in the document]
lalitb: Good catch. Yes it should be ~203. Will fix this now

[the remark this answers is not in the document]
cijothomas: Also, main branch is not updated with right numbers. That should be done first, else the gains will be incorrectly assumed to be coming from stack allocation of attributes, where a good chunk is already from the previous optimizations.

lalitb:
> LGTM. Left a nit about benchmark results not in sync with the PR desc. A combined changelog highlighting all perf improvements could be added in a follow up?

Have fixed the benchmark result as per the desc, along with the comment. Yes planning to add the combined changelog as follow up.  With this, we've hit the 100th comment on this PR :)

cijothomas:
> > LGTM. Left a nit about benchmark results not in sync with the PR desc. A combined changelog highlighting all perf improvements could be added in a follow up?
> 
> Have fixed the benchmark result as per the desc, along with the comment. Yes planning to add the combined changelog as follow up. With this, we've hit the 100th comment on this PR :)

merged the other PR, so this need to handle the conflict. Sorry I wasn't paying attention to the merge order, Hopefully this is an easy conflict to resolve and we are good merge.

This is comment 102 🤣 ! (We still haven't broken the records of some PRs in OTel specs)

lalitb:
> merged the other PR, so this need to handle the conflict. Sorry I wasn't paying attention to the merge order, Hopefully this is an easy conflict to resolve and we are good merge.

Yes, fixed the merge conflict now :)

cijothomas:
The idea of using stack for attributes is not super common, and may be undesired in some scenarios. We do have precedence in Go, .NET and even in Rust for similar changes. But this change is done as a **purely internal implementation detail**, so it is possible to change this in the future, without any breaking change. 
With that, I'll hit the merge button now! 

Thanks Lalit for continuously pursuing better performance and thanks everyone who helped review.

Issue #1879, opened 2024-06-11
[Feature]: metrics macro API like log/trace?


kelko:
### Describe the solution you'd like:

opentelemetry-rust uses a logging-bridge to connect to log/trace macros and also discusses currently dropping the own tracing SDK in favor of Tokios trace macros for tracing (#1689). Similar to e.g. .NET OTel leveraging System.Diagnostics & Microsoft.Extension.ILogger instead additional libs.

The log & trace macros have the really nice effect, that one doesn't need to pass around optional objects around a lot, the macros are "just there", even if no logging/tracing is turned on. And they are a nice abstraction from the actual logging library. My libs don't need to care about specifics.

With this background:

Would it be possible to make such an abstraction macro for metrics as well? Creating the concrete Meter instance via `meter!(name)` and later derive instruments from that meter. And at startup I either connect an OTel metrics provider to the meter-backend. Or I don't.
But while writing the library I don't need to care about the meter potentially being None. It will work anyway code-wise, but just don't do much (like log!/trace!).

Then (over time) rust developers only need 3 abstraction libs for observability, all similar in how they are used: log, trace & meter.
I would be willing to help out, but I don't (yet) have much experience with rust macro development.

### Considered Alternatives

Passing along Option<Meter>, which I currently do in my code. And it's cumbersome, compared to log!/trace!

cijothomas:
option 3 from https://github.com/open-telemetry/opentelemetry-rust/issues/1642 calls for macros approach. This is not yet pursued actively, but I can totally imagine that coming in the future. Leveraging macros for performance *and* easier-to-use APIs is a good addition.

PR #1972, opened 2024-07-28
[WIP] Add Counted Macro

manuelgdlvh:
Fixes #

## Changes
I am developing a macros crate designed to streamline the generation of observability-related code across various types of metrics and logging functionalities. The initial focus of this crate is on implementing a basic counter, but our goal is to support a wide range of observability features

The crate leverages Rust's powerful attribute macros to simplify the configuration and integration process. By using these attribute macros, developers can avoid repetitive boilerplate code and configure observability tools with maximum flexibility. This approach ensures that the observability code remains concise, readable, and maintainable, allowing developers to focus more on their core application logic.

Key benefits of this macros crate include:

    Reduction of Boilerplate Code: By abstracting common patterns and setups into macros, we significantly reduce the amount of boilerplate code developers need to write and maintain.

    Consistency: The macros enforce consistent implementation of observability features across different parts of the codebase, reducing the likelihood of errors and inconsistencies.

    Flexibility: The attribute macros are designed to be highly configurable, allowing developers to customize the generated code to fit their specific needs without sacrificing simplicity.

First steps looks like (LazyLock would be changed to ensure version compatibility):

Screenshot from 2024-07-29 00-07-26


cijothomas (changes requested):
There is definitely an interest if offering macros for not just metrics, but also other things. https://github.com/open-telemetry/opentelemetry-rust/issues/1879

I am also particularly interested in using it to solve some perf issues like https://github.com/open-telemetry/opentelemetry-rust/issues/1642 by leveraging macros.

We are not quite ready with a plan on how to make this happen in this repo. Until then, this is best hosted in opentelemetry-rust-contrib repo. Once it grows/stabilizes, it can be re-evaluated for bringing to core api/sdk itself.

Would you like to first describe the design (rough one) in an issue in the contrib repo, so we can discuss more?

manuelgdlvh:
Hi @cijothomas ,

It seems correct to me to include these first steps in the contrib crate.

I am currently working on the metrics part, more specifically on the counters part, and when it's finished, I plan to create a macro for Timer metrics that generates count and sum of times (I am open to hearing other priorities or needs to start with those).

Regarding what you mentioned about KeyValue, the first implementation is intended for const labels whose performance could be improved with static values, avoiding the repetitive creation of these structures.

For dynamic labels, we would need to change the way these values are passed, but I currently don't see how using a procedural macro could improve performance without changing the internal implementation of this part (which is not initially intended to be covered by these feature/s).

Within this first week, I would like to have the Counted macro ready, with the described design and implementation, a roadmap of possible interesting macros for the future, and to look into the performance issues you mentioned in depth and how to handle them efficiently, both dynamic and static.

Thanks for your time!

cijothomas:
> don't see how using a procedural macro could improve performance 

It won't improve perf of the actual SDK. I was referring to avoiding the un-necessary overhead when OTel SDK is not even enabled in this issue : https://github.com/open-telemetry/opentelemetry-rust/issues/1642

cijothomas:
Closing the PR in this repo, as it was agreed this will be on contrib repo.

Issue #1954, opened 2024-07-23
OTel Metrics with no heap allocation


cijothomas:
Opening a parent issue to track making it possible to do OTel Metrics with zero heap allocation. Since Metrics are always pre-aggregated and exported once every few seconds in background, the main goal is to achieve zero allocation in the hot path of reporting measurements.

This requires support from

1. OTel Metrics API - the API accepts slice of KV, so users can pass stack allocated KV pair array.
2. OTel Metrics SDK - The Sdk is currently allocating and copying to Heap in hot path. There is a way to avoid this completely for existing timeseries (the common case is updating existing points, not creating new ones). For non-hot path (i.e Exporting), it is also possible to avoid allocation by re-using the memory for each export.
3. Metric Exporter - It maybe possible to reuse the buffer required to hold serialization data

The overall flow in hot path would be:
user created stack allocated array of KV pairs -> OTel Metric API -> OTel Metric SDK -> Use the slice to do lookup and find the metric point to update -> Update

After the initial "hydration" phase, then measurements with same KV pairs require no allocation, as SDK simply needs the slice to perform lookup. If a previously unseen (or not seen for at least one cycle) measurement is provided, it'd need allocation.

Also, need some tooling to verify the above claims. Zero heap hot path is achieved on prototyping separately (the prototyped code will be used as reference to refactor large parts of Metrics SDK implementation shortly).

cijothomas:
We have now achieved the goal of no heap allocation in the hot path. This will be part of the upcoming release (~Aug 30th).

The export/collect path is still being worked on, so this issue is not closed, but this is in a future release.

utpilla:
Related issue #1740.

PR #1977, opened 2024-07-31
MetricRefactor - Part1 - Move AttributeSet inside aggregations

cijothomas:
Part 1 of metric sdk refactor to achieve perf gains and [no-heap-allocation](https://github.com/open-telemetry/opentelemetry-rust/issues/1954). The heap allocation and perf cost is due to the usage of `AttributeSet` which clones and stores the incoming slice of KV pairs, after sorting + de-duplicating, which can be eliminated.

This PR modifies traits to only need `&[KeyValue]` (which is what API accepts from user) instead of `AttributeSet`. There is no perf gain in this PR, as the `AttributeSet` is still constructed inside each aggregations. This is done to keep the scope of PR very small, and easy to review.
In the next PR, Sum aggregation will be targeted to achieve the perf gains and no-heap alloc, later extending to other instruments.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -3,7 +3,10 @@ use std::{marker, sync::Arc};
 use once_cell::sync::Lazy;
 use opentelemetry::KeyValue;
 
-use crate::metrics::data::{Aggregation, Gauge, Temporality};
+use crate::metrics::{
+    data::{Aggregation, Gauge, Temporality},
+    AttributeSet,
+};
 
 use super::{
     exponential_histogram::ExpoHistogram,
@@ -14,8 +17,10 @@ use super::{
 };
 
 const STREAM_CARDINALITY_LIMIT: u32 = 2000;
-pub(crate) static STREAM_OVERFLOW_ATTRIBUTE_SET: Lazy<Vec<KeyValue>> =
-    Lazy::new(|| vec![KeyValue::new("otel.metric.overflow", "true")]);
+pub(crate) static STREAM_OVERFLOW_ATTRIBUTE_SET: Lazy<AttributeSet> = Lazy::new(|| {
+    let key_values: [KeyValue; 1] = [KeyValue::new("otel.metric.overflow", "true")];
+    AttributeSet::from(&key_values[..])
+});
 
 /// Checks whether aggregator has hit cardinality limit for metric streams
 pub(crate) fn is_under_cardinality_limit(size: usize) -> bool {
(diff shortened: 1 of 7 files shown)


opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -94,11 +94,14 @@ impl<T: Number<T>> AggregateBuilder<T> {
     /// Wraps the passed in measure with an attribute filtering function.
     fn filter(&self, f: impl Measure<T>) -> impl Measure<T> {
         let filter = self.filter.clone();
-        move |n, mut attrs: AttributeSet| {
-            if let Some(filter) = &filter {
-                attrs.retain(filter.as_ref());
-            }
-            f.call(n, attrs)
+        move |n, attrs: &[KeyValue]| {
+            let filtered_attrs: Vec<KeyValue> = if let Some(filter) = &filter {
```
utpilla, line 98: Could we update it to this to avoid the allocation when there is no filter?

```rust
move |n, attrs: &[KeyValue]| {
    if let Some(filter) = &filter {
        let filtered_attrs: Vec<KeyValue> = attrs.iter().filter(|kv| filter(kv)).cloned().collect();
        f.call(n, &filtered_attrs);
    } else {
        f.call(n, attrs);
    };
}
```

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -94,11 +94,14 @@ impl<T: Number<T>> AggregateBuilder<T> {
...
+                attrs.iter().filter(|kv| filter(kv)).cloned().collect()
+            } else {
+                // TODO: This is unwanted allocation. We should avoid it.
```
utpilla, line 101: Just noticed that you added a TODO comment 😀.

I think [this](https://github.com/open-telemetry/opentelemetry-rust/pull/1977/files#r1699088220) should fix it. Up to you if you want to do it in this PR or not.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -129,6 +129,16 @@ impl AttributeSet {
     pub fn iter(&self) -> impl Iterator<Item = (&Key, &Value)> {
         self.0.iter().map(|kv| (&kv.key, &kv.value))
     }
+
+    /// Returns a slice of the key value pairs in the set
+    pub fn as_slice(&self) -> &[KeyValue] {
```
utpilla, line 134: These two methods are not being used in this PR. Do we need them for later?

utpilla (approved):
Looks good! Left some suggestions.

lalitb (approved):
LGTM. These seems to be all internal changes, so no changelog required.

Issue #1966, opened 2024-07-25
[Docs]: Compatibility matrix


xd009642:
### Describe the solution you'd like:

This is very much just a this-crate issue. But the usage of traits from different otel crates is widespread enough that whenever I attempt to upgrade I end up in some compatibility nightmare trying to get all the versions aligned to the correct one. As far as I see if it, the solution would be either to move versions up lockstep and disregarding semver (or at least broadening it's definition to cover all the various crates), or alternatively document for each opentelemetry version what the latest compatible version for the other crates are.

I'd rather not spend another hour of my life trying to solve issues like `the trait `JaegerTraceRuntime` is not implemented for `Tokio`` whenever I upgrade this crate. Unfortunately, the opentelemetry upgrades are the most painful upgrades I've found in the Rust ecosystem :cry:

lalitb:
> solution would be either to move versions up lockstep and disregarding semver (or at least broadening it's definition to cover all the various crates), or alternatively document for each opentelemetry version what the latest compatible version for the other crates are.

@xd009642 In the current publishing process, we coordinate the versions of all OpenTelemetry-related crates to be updated simultaneously, ensuring compatibility across the board. The `tracing-opentelemetry` crate is an exception as it is part of `tokio` ecosystem, and so is upgraded separately afterwards. This does bring the `is not implemented for` errors in interim. While discussions around #1571 are still ongoing, I agree that there needs to be better synchronization to address these issues.

ibigbug:
vote up for this.

I'm still facing trait issue when every individual package is up to date

```
# opentelemetry
opentelemetry = "0.24"
opentelemetry_sdk = { version = "0.24.1", features = ["rt-tokio"] }
tracing-opentelemetry = "0.25"
opentelemetry-jaeger-propagator = "0.3"
opentelemetry-otlp = { version = "0.17", features = ["grpc-tonic"] }
```

```
error[E0277]: the trait bound `opentelemetry_sdk::trace::TracerProvider: opentelemetry::trace::Tracer` is not satisfied
   --> clash_lib/src/app/logging.rs:119:57
    |
119 |         Some(tracing_opentelemetry::layer().with_tracer(tracer))
    |                                             ----------- ^^^^^^ the trait `opentelemetry::trace::Tracer` is not implemented for `opentelemetry_sdk::trace::TracerProvider`
    |                                             |
    |                                             required by a bound introduced by this call
    |
[...]

PR #1979, opened 2024-07-31
Update README with guidance on version updates

lalitb:
## Changes

Adding a disclaimer that it is possible to have breaking changes with minor version updates, till we are stable. Based on discussions here - https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/98#issuecomment-2261199603 and https://github.com/open-telemetry/opentelemetry-rust/issues/1966#issuecomment-2256452286

While Versioning.md could have been the better place to add this, README.md will get better attention :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,12 @@ See the [examples](./examples) directory for different integration patterns.
 
 ## Overview of crates
 
-The following crates are maintained in this repo:
+OpenTelemetry Rust is composed of multiple interdependent crates, each of which 
+may have different version numbers. To maintain compatibility and ensure proper 
+functionality, it is recommended to update all related crates simultaneously 
+when new versions are released. While it would be ideal to have fewer breaking 
+changes with minor version updates, this is currently not feasible until all 
+signals within the ecosystem reach stability.
 
 * [`opentelemetry`] This is the OpenTelemetry API crate, and is the crate
   required to instrument libraries and applications. It contains Context API,
--- a/README.md
+++ b/README.md
@@ -98,12 +98,12 @@ See the [examples](./examples) directory for different integration patterns.
 
 ## Overview of crates
 
-OpenTelemetry Rust is composed of multiple interdependent crates, each of which 
-may have different version numbers. To maintain compatibility and ensure proper 
-functionality, it is recommended to update all related crates simultaneously 
-when new versions are released. While it would be ideal to have fewer breaking 
-changes with minor version updates, this is currently not feasible until all 
-signals within the ecosystem reach stability.
+OpenTelemetry Rust is composed of multiple interdependent crates as given below, 
+each of which may have different version numbers. To maintain compatibility 
+and ensure proper functionality, it is recommended to update all related 
+crates simultaneously when new versions are released. While it would be ideal
+to have fewer breaking changes with minor version updates, this is currently 
+not feasible until all signals within the ecosystem reach stability.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,7 +98,12 @@ See the [examples](./examples) directory for different integration patterns.
 
 ## Overview of crates
 
-The following crates are maintained in this repo:
+OpenTelemetry Rust is composed of multiple interdependent crates as given below, 
+each of which may have different version numbers. To maintain compatibility 
+and ensure proper functionality, it is recommended to update all related 
+crates simultaneously when new versions are released.  Although it would be 
+ideal to avoid breaking changes with minor version updates, this is currently
```
utpilla, line 105: This makes it sound like are we are required to avoid such breaking changes, but we would not be able to do it. It's completely valid under Semver compatibility rules to make breaking changes with minor version updates for `0.x.x` versions.

If the intention is to warn users who might have misconception about what kind of breaking changes are allowed, maybe we could just reiterate the relevant Semver compatibility rule. Something similar to the text in this [doc](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility):

> Versions are considered compatible if their left-most non-zero major/minor/patch component is the same. For example, 0.1.0 and 0.1.2 are compatible, but 0.1.0 and 0.2.0 are not.

lalitb: > It's completely valid under Semver compatibility rules to make breaking changes with minor version updates for 0.x.x versions.

This is what I read from [semver](https://semver.org/):
```
MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
PATCH version when you make backward compatible bug fixes
```

lalitb: Oh you mean with `0` as major version.

lalitb: The issue I observe is that in the Rust ecosystem, most of crates are in 0.x.x versions but are considered stable, with minimal breaking changes occurring with minor version updates. I just wanted to emphasize that for OpenTelemetry, this would be comparatively more frequent.

utpilla: Got it. Thanks for providing the context! Maybe we can put this text under its own section in the doc to highlight this point even more. Something like this: https://github.com/rust-random/rand?tab=readme-ov-file#versions

cijothomas: I am unsure what is the issue we are trying to address with this wording? Can you elaborate.
Is it to convey that we are not stable?
Is it to convey that breaking changes are possible within minor version bump, until a crate hits 1.0?
Or more?

lalitb: > I am unsure what is the issue we are trying to address with this wording? Can you elaborate.
> Is it to convey that we are not stable?
> Is it to convey that breaking changes are possible within minor version bump, until a crate hits 1.0?
> Or more?

Maybe I haven't put the wordings properly. The main emphasis was this

> To maintain compatibility  and ensure proper functionality, it is recommended to update all related 
crates simultaneously when new versions are released. 

 i.e., to ask users to update to the latest versions of all OpenTelemetry crates together when new releases are made. Additionally, if there are any indirect dependencies on OpenTelemetry crates (e.g. such as those coming through `tracing-opentelemetry`), they should ensure that the versions are same (preferably latest). If not, reach out to the repo providing this dependency asking them to update.

But looks like the PR is causing more confusion, and less value.  Would be good to close this :)

lalitb:
Closing this for now. We should be adding the guidance for the version updates because of the interlinked crates published from this repo, will come up with better wordings after discussing in community meeting.

PR #1985, opened 2024-08-05
Hide LogRecord attributes Implementation Details from exporters

lalitb:
## Changes

Hide the implementation details of LogRecord attributes by:

- Making the LogRecord attributes private to the opentelemetry-sdk crate.
- Creating wrapper methods attributes_iter(), attributes_contains(), and attributes_len() to expose the functionality of these attributes to the exporter.

These changes will enable modifications to the internal implementation of these attributes without requiring any changes in the exporter code.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -13,9 +13,9 @@
   [#1973](https://github.com/open-telemetry/opentelemetry-rust/pull/1973)
   - **Breaking** [#1985](https://github.com/open-telemetry/opentelemetry-rust/pull/1985)
   Hide LogRecord attributes Implementation Details from processors and exporters.
-  The custom exporter's can't directly access the `LogData::LogRecord::attributes`, as
+  The custom exporters and processors can't directly access the `LogData::LogRecord::attributes`, as
   these are private to opentelemetry-sdk. Instead, they would now use LogRecord::attributes_iter()
-  and LogRecord::attributes_len() methods to access them.
+  method to access them.
 
 
 ## v0.24.1
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -11,6 +11,11 @@
   first interval instead of doing it right away.
   [#1970](https://github.com/open-telemetry/opentelemetry-rust/pull/1970)
   [#1973](https://github.com/open-telemetry/opentelemetry-rust/pull/1973)
+  - **Breaking** [#1985](https://github.com/open-telemetry/opentelemetry-rust/pull/1985)
+  Hide LogRecord attributes Implementation Details from exporters.
+  The custom exporter's can't directly access the `LogData::LogRecord::attributes`, as
+  these are private to opentelemetry-sdk. Instead, they would now use LogRecord::attributes_iter()
+  and LogRecord::attributes_len() methods to access them.
 
 
 ## v0.24.1
--- a/opentelemetry-sdk/src/logs/record.rs
+++ b/opentelemetry-sdk/src/logs/record.rs
@@ -110,8 +110,9 @@ impl LogRecord {
         self.attributes.as_ref().map_or(0, |attrs| attrs.len())
     }
 
+    #[allow(dead_code)]
     /// Returns true if the `LogRecord` contains the specified attribute.
-    pub fn attributes_contains(&self, key: &Key, value: &AnyValue) -> bool {
+    pub(crate) fn attributes_contains(&self, key: &Key, value: &AnyValue) -> bool {
         self.attributes.as_ref().map_or(false, |attrs| {
             attrs.iter().any(|(k, v)| k == key && v == value)
         })
--- a/opentelemetry-sdk/src/logs/record.rs
+++ b/opentelemetry-sdk/src/logs/record.rs
@@ -34,7 +34,7 @@ pub struct LogRecord {
     pub body: Option<AnyValue>,
 
     /// Additional attributes associated with this record
-    pub attributes: Option<Vec<(Key, AnyValue)>>,
+    pub(crate) attributes: Option<Vec<(Key, AnyValue)>>,
 }
 
 impl opentelemetry::logs::LogRecord for LogRecord {
@@ -94,6 +94,27 @@ impl opentelemetry::logs::LogRecord for LogRecord {
         } else {
             self.attributes = Some(vec![(key.into(), value.into())]);
         }
+    }
+}
+
+impl LogRecord {
+    /// Provides an iterator over the attributes in the `LogRecord`.
+    pub fn attributes_iter(&self) -> impl Iterator<Item = &(Key, AnyValue)> {
+        self.attributes
+            .as_ref()
+            .map_or_else(|| [].iter(), |attrs| attrs.iter())
+    }
+
+    /// Returns the number of attributes in the `LogRecord`.
+    pub fn attributes_len(&self) -> usize {
+        self.attributes.as_ref().map_or(0, |attrs| attrs.len())
+    }
+
+    /// Returns true if the `LogRecord` contains the specified attribute.
+    pub fn attributes_contains(&self, key: &Key, value: &AnyValue) -> bool {
+        self.attributes.as_ref().map_or(false, |attrs| {
+            attrs.iter().any(|(k, v)| k == key && v == value)
+        })
     }
 }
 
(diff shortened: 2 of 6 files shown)


opentelemetry-sdk/src/logs/record.rs
```diff
@@ -97,6 +97,27 @@ impl opentelemetry::logs::LogRecord for LogRecord {
...
+            .as_ref()
+            .map_or_else(|| [].iter(), |attrs| attrs.iter())
+    }
+
+    /// Returns the number of attributes in the `LogRecord`.
+    pub fn attributes_len(&self) -> usize {
+        self.attributes.as_ref().map_or(0, |attrs| attrs.len())
+    }
+
+    /// Returns true if the `LogRecord` contains the specified attribute.
```
cijothomas, line 113: I think this is not required to be public. If an exporter needs it, then can write a helper function to get this by using the above 2 methods.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -34,7 +34,7 @@ pub struct LogRecord {
     pub body: Option<AnyValue>,
 
     /// Additional attributes associated with this record
-    pub attributes: Option<Vec<(Key, AnyValue)>>,
+    pub(crate) attributes: Option<Vec<(Key, AnyValue)>>,
```
cijothomas, line 37: please add a changelog entry to the sdk crate.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@
   first interval instead of doing it right away.
   [#1970](https://github.com/open-telemetry/opentelemetry-rust/pull/1970)
   [#1973](https://github.com/open-telemetry/opentelemetry-rust/pull/1973)
+  - **Breaking** [#1985](https://github.com/open-telemetry/opentelemetry-rust/pull/1985)
+  Hide LogRecord attributes Implementation Details from exporters.
```
cijothomas, line 15:
```suggestion
  Hide LogRecord attributes Implementation Details from processors and exporters.
```

Same for processors also, right?

opentelemetry-sdk/src/logs/record.rs
cijothomas: I think we need to offer API to modify an existing entry, remove an existing entry too. These are critical to ensure we allow redaction, enrichment capabilities. Lets track it as separate issue for adding in the future.

opentelemetry-sdk/CHANGELOG.md
lalitb: yes, custom processors not part of otel-sdk crate.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -97,6 +97,28 @@ impl opentelemetry::logs::LogRecord for LogRecord {
...
+impl LogRecord {
+    /// Provides an iterator over the attributes in the `LogRecord`.
+    pub fn attributes_iter(&self) -> impl Iterator<Item = &(Key, AnyValue)> {
+        self.attributes
+            .as_ref()
+            .map_or_else(|| [].iter(), |attrs| attrs.iter())
+    }
+
+    /// Returns the number of attributes in the `LogRecord`.
+    pub fn attributes_len(&self) -> usize {
```
utpilla, line 109: This method also need not be made public?

lalitb: Without this, the only way to get the length would be to iterate over all the entries. I can keep it private if we are sure this won't be used by exporters/processors.

utpilla: I would be in favor of keeping the public API surface smaller unless we already have users asking for this feature. This could always be added later if there's a significant demand for it.

utpilla: Also, if we do decide to give users the information about length of the iterator, we should probably do it by overriding the `Iterator` traits methods instead of coming up with our own. That's probably the idiomatic way of doing it.

- [Iterator::size_hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint)
- [Iterator::count](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count)

lalitb: [Iterator::count](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count) will iterate over the entries to find the size - not sure if we really want this, as that would have o(n) complexity.

cijothomas: Given there is no other performant way to obtain length, I think we'll need to expose attributes length in some way, but lets start without it and add later.

lalitb: I can remove it if we all agree. As of now, it's used in otel-appender-tracing/logs unit-test to validate the attributes, will create the wrapper over the attributes_iter() there.

cijothomas (approved):
LGTM.
Its probably better not to expose len() method initially, not a blocker if we expose it as well.

opentelemetry-sdk/src/logs/record.rs
cijothomas: if the only use is in test, lets remove it actually and make a helper method in tests.

utpilla: > [Iterator::count](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count) will iterate over the entries to find the size - not sure if we really want this, as that would have o(n) complexity.

Yes, we don't want that. That's the default behavior of the count method. My suggestion was to override these methods and not use them as-is. You should be able to override them to have a more performant implementation.

cijothomas: > > [Iterator::count](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.count) will iterate over the entries to find the size - not sure if we really want this, as that would have o(n) complexity.
> 
> Yes, we don't want that. That's the default behavior of the count method. My suggestion was to override these methods and not use them as-is. You should be able to override them to have a more performant implementation.

That is good! I didn't realize we could provide own implementation to return the known size without iterating.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@
   first interval instead of doing it right away.
   [#1970](https://github.com/open-telemetry/opentelemetry-rust/pull/1970)
   [#1973](https://github.com/open-telemetry/opentelemetry-rust/pull/1973)
+  - **Breaking** [#1985](https://github.com/open-telemetry/opentelemetry-rust/pull/1985)
+  Hide LogRecord attributes Implementation Details from processors and exporters.
+  The custom exporter's can't directly access the `LogData::LogRecord::attributes`, as
+  these are private to opentelemetry-sdk. Instead, they would now use LogRecord::attributes_iter()
+  and LogRecord::attributes_len() methods to access them.
```
utpilla, line 18: Remove the mention of `attributes_len()`.

opentelemetry-sdk/src/logs/record.rs
lalitb: > That is good! I didn't realize we could provide own implementation to return the known size without iterating.

We shouldn't be doing that. Any method over Iterator trait should be consuming the iterator, iterating over it and giving the required result.  If we need to provide length, we should be providing a separate method, just like provided by other containers.

opentelemetry-sdk/CHANGELOG.md
lalitb: Thanks, removed it :)

utpilla (approved):

opentelemetry-sdk/src/logs/record.rs
utpilla: > Any method over Iterator trait should be consuming the iterator, iterating over it and giving the required result.

Is this a requirement? The docs for [implementing iterator](https://doc.rust-lang.org/std/iter/index.html#implementing-iterator) says that we could write a custom implementation for the iterator methods (without repeatedly calling `next()`):

_"Also note that Iterator provides a default implementation of methods such as nth and fold which call next internally. **However, it is also possible to write a custom implementation of methods like nth and fold if an iterator can compute them more efficiently without calling next.**"_

lalitb: The issue is, any method on the iterator should start processing from the current position of the iterator. So Iter::count() shouldn't be giving the actual size of underlying storage, but the length from the current position.
```rust
fn main() {
    let vec = vec![1, 2, 3, 4, 5];
    let mut iter = vec.iter();
    let mut count = iter.clone().count();
    println!("count: {}", count); // 5

    iter.next();
    count = iter.count();

    println!("new count: {}", count); // 4
}
```

PR #1989, opened 2024-08-06
Metric refactor - 2x perf and allocation free

### LINKED ISSUE #1954: OTel Metrics with no heap allocation

cijothomas:
Part 2 of https://github.com/open-telemetry/opentelemetry-rust/issues/1740, addressing 1,2,3. Also closes #1954 for SUM.

Refactoring SUM aggregation logic to operate entirely on user provided slice of KeyValues. In hot-path, we avoid **allocating, copying, sorting, de-duplicating.**, leading to **2X or more gains** (with 4 attributes. More attributes = much more gains), as long as user do not keep changing the order of attributes.

This is mostly based on the prototypes did outside this repo in https://github.com/cijothomas/metrics-mini/tree/main/metrics
Once the approach is settled, we can replicate to all aggregations. (This PR is only fixing SUM)

Benchmarks:
```
Counter_Add_Sorted      time:   [192.64 ns 193.03 ns 193.48 ns]
                        change: [-66.328% -66.034% -65.769%] (p = 0.00 < 0.05)
                        Performance has improved.
Counter_Add_Unsorted    time:   [209.09 ns 209.41 ns 209.74 ns]
                        change: [-63.198% -62.905% -62.670%] (p = 0.00 < 0.05)
                        Performance has improved.
Counter_Overflow        time:   [895.61 ns 898.58 ns 902.11 ns]
                        change: [+54.000% +58.218% +61.362%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

Stress Test also shows small gains (17M/sec to 20M/sec). https://github.com/open-telemetry/opentelemetry-rust/pull/1833 fixed the major contention, so this PR has limited impact on stress test, but the raw speed is significantly improved.

**Stress Test with memory stats:**

Main:
Throughput: 17,842,800 iterations/sec
Memory usage: 5.06 MB
CPU usage: 98.61875%
Virtual memory usage: 2249.73 MB

PR:
Throughput: 18,869,600 iterations/sec
Memory usage: 4.23 MB
CPU usage: 99.2016%
Virtual memory usage: 2249.55 MB

Visible memory reduction as well from ~5 MB to ~4.2 MB

*note*: Overflow perf regressed, and that is acceptable. The main goal with overflow feature is to protect SDK from consuming unlimited memory, so it is acceptable that this path is slower. This is rare case, and indicates user needs to adjust cardinality.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/sum.rs
@@ -74,7 +74,7 @@ impl<T: Number<T>> ValueMap<T> {
                         } else if let Some(value_to_update) = values.get(sorted_attrs.as_slice()) {
                             value_to_update.add(measurement);
                             return;
-                        } else if is_under_cardinality_limit(self.count.load(Ordering::Acquire)) {
+                        } else if is_under_cardinality_limit(self.count.load(Ordering::SeqCst)) {
                             let new_value = T::new_atomic_tracker();
                             new_value.add(measurement);
                             let new_value = Arc::new(new_value);
@@ -85,7 +85,7 @@ impl<T: Number<T>> ValueMap<T> {
                             // Insert sorted order
                             values.insert(sorted_attrs, new_value);
 
-                            self.count.fetch_add(1, Ordering::Release);
+                            self.count.fetch_add(1, Ordering::SeqCst);
                         } else if let Some(overflow_value) =
                             values.get_mut(STREAM_OVERFLOW_ATTRIBUTES.as_slice())
                         {
--- a/opentelemetry-sdk/src/metrics/internal/sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/sum.rs
@@ -54,13 +54,11 @@ impl<T: Number<T>> ValueMap<T> {
             // Try incoming order first
             if let Some(value_to_update) = values.get(attrs) {
                 value_to_update.add(measurement);
-                return;
             } else {
                 // Then try sorted order.
                 let sorted_attrs = AttributeSet::from(attrs).into_vec();
                 if let Some(value_to_update) = values.get(sorted_attrs.as_slice()) {
                     value_to_update.add(measurement);
-                    return;
                 } else {
                     // Give up the lock, before acquiring write lock.
                     drop(values);
@@ -70,10 +68,8 @@ impl<T: Number<T>> ValueMap<T> {
                         // value.
                         if let Some(value_to_update) = values.get(attrs) {
                             value_to_update.add(measurement);
-                            return;
                         } else if let Some(value_to_update) = values.get(sorted_attrs.as_slice()) {
                             value_to_update.add(measurement);
-                            return;
                         } else if is_under_cardinality_limit(self.count.load(Ordering::SeqCst)) {
                             let new_value = T::new_atomic_tracker();
                             new_value.add(measurement);
@@ -90,7 +86,6 @@ impl<T: Number<T>> ValueMap<T> {
                             values.get_mut(STREAM_OVERFLOW_ATTRIBUTES.as_slice())
                         {
                             overflow_value.add(measurement);
-                            return;
                         } else {
                             let new_value = T::new_atomic_tracker();
                             new_value.add(measurement);
@@ -152,7 +147,7 @@ impl<T: Number<T>> Sum<T> {
 
         // Max number of data points need to account for the special casing
         // of the no attribute value + overflow attribute.
-        let n = self.value_map.count.load(Ordering::Relaxed) + 2;
+        let n = self.value_map.count.load(Ordering::SeqCst) + 2;
         if n > s_data.data_points.capacity() {
             s_data
                 .data_points
@@ -223,14 +218,11 @@ impl<T: Number<T>> Sum<T> {
         let s_data = s_data.unwrap_or_else(|| new_agg.as_mut().expect("present if s_data is none"));
         s_data.temporality = Temporality::Cumulative;
         s_data.is_monotonic = self.monotonic;
-        s_data.data_points.clear();
+        s_data.data_points.clear();        
 
-        let values = match self.value_map.values.write() {
-            Ok(v) => v,
-            Err(_) => return (0, None),
-        };
-
-        let n = values.len() + 1;
+        // Max number of data points need to account for the special casing
+        // of the no attribute value + overflow attribute.
+        let n = self.value_map.count.load(Ordering::SeqCst) + 2;
         if n > s_data.data_points.capacity() {
             s_data
                 .data_points
@@ -252,6 +244,11 @@ impl<T: Number<T>> Sum<T> {
                 exemplars: vec![],
             });
         }
+
+        let values = match self.value_map.values.write() {
+            Ok(v) => v,
+            Err(_) => return (0, None),
+        };
 
         // TODO: This will use an unbounded amount of memory if there
         // are unbounded number of attribute sets being aggregated. Attribute
@@ -316,14 +313,11 @@ impl<T: Number<T>> PrecomputedSum<T> {
         let s_data = s_data.unwrap_or_else(|| new_agg.as_mut().expect("present if s_data is none"));
         s_data.data_points.clear();
         s_data.temporality = Temporality::Delta;
-        s_data.is_monotonic = self.monotonic;
+        s_data.is_monotonic = self.monotonic;        
 
-        let mut values = match self.value_map.values.write() {
-            Ok(v) => v,
-            Err(_) => return (0, None),
-        };
-
-        let n = values.len() + 1;
+        // Max number of data points need to account for the special casing
+        // of the no attribute value + overflow attribute.
+        let n = self.value_map.count.load(Ordering::SeqCst) + 2;
         if n > s_data.data_points.capacity() {
             s_data
                 .data_points
@@ -348,6 +342,11 @@ impl<T: Number<T>> PrecomputedSum<T> {
                 exemplars: vec![],
             });
         }
+
+        let mut values = match self.value_map.values.write() {
+            Ok(v) => v,
+            Err(_) => return (0, None),
+        };
 
         let default = T::default();
         for (attrs, value) in values.drain() {
@@ -399,14 +398,11 @@ impl<T: Number<T>> PrecomputedSum<T> {
         let s_data = s_data.unwrap_or_else(|| new_agg.as_mut().expect("present if s_data is none"));
         s_data.data_points.clear();
         s_data.temporality = Temporality::Cumulative;
-        s_data.is_monotonic = self.monotonic;
+        s_data.is_monotonic = self.monotonic;        
 
-        let values = match self.value_map.values.write() {
-            Ok(v) => v,
-            Err(_) => return (0, None),
-        };
-
-        let n = values.len() + 1;
+        // Max number of data points need to account for the special casing
+        // of the no attribute value + overflow attribute.
+        let n = self.value_map.count.load(Ordering::SeqCst) + 2;
         if n > s_data.data_points.capacity() {
             s_data
                 .data_points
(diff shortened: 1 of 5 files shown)


opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -35,42 +39,65 @@ impl<T: Number<T>> ValueMap<T> {
...
             has_no_value_attribute_value: AtomicBool::new(false),
             no_attribute_value: T::new_atomic_tracker(),
+            count: AtomicUsize::new(0),
         }
     }
 }
 
 impl<T: Number<T>> ValueMap<T> {
-    fn measure(&self, measurement: T, attrs: AttributeSet) {
+    fn measure(&self, measurement: T, attrs: &[KeyValue]) {
```
cijothomas, line 48: this is the key - we operate on incoming ref/slice itself **without allocating, copying, sorting, de-duplicating**. If the look up fails, the "slow" path will do everything as before. The net effect is, for users who use metrics API normally (i.e without duplicates, and without changing the order of attributes), we'll have highest performance and zero heap allocations.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -8,19 +10,21 @@ use std::{
...
+pub(crate) static STREAM_OVERFLOW_ATTRIBUTES: Lazy<Vec<KeyValue>> =
+    Lazy::new(|| vec![KeyValue::new("otel.metric.overflow", "true")]);
 
 /// The storage for sums.
 struct ValueMap<T: Number<T>> {
-    values: RwLock<HashMap<AttributeSet, T::AtomicTracker>>,
+    values: RwLock<HashMap<Vec<KeyValue>, Arc<T::AtomicTracker>>>,
     has_no_value_attribute_value: AtomicBool,
     no_attribute_value: T::AtomicTracker,
+    count: AtomicUsize,
```
cijothomas, line 27: count is tracked separately, as number of entries in HashMap is no longer accurate, as we insert original+sorted in hashmap.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -8,19 +10,21 @@ use std::{
...
-};
+use super::{aggregate::is_under_cardinality_limit, AtomicTracker, Number};
+
+pub(crate) static STREAM_OVERFLOW_ATTRIBUTES: Lazy<Vec<KeyValue>> =
+    Lazy::new(|| vec![KeyValue::new("otel.metric.overflow", "true")]);
 
 /// The storage for sums.
 struct ValueMap<T: Number<T>> {
-    values: RwLock<HashMap<AttributeSet, T::AtomicTracker>>,
+    values: RwLock<HashMap<Vec<KeyValue>, Arc<T::AtomicTracker>>>,
```
cijothomas, line 24: The metric value is `Arc` shared to ensure any order results in contributing to same timeseries. Arc equality is also leveraged in export time to de-dup entries.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -152,23 +175,29 @@ impl<T: Number<T>> Sum<T> {
...
-                attributes: attrs
-                    .iter()
-                    .map(|(k, v)| KeyValue::new(k.clone(), v.clone()))
-                    .collect(),
-                start_time: Some(prev_start),
-                time: Some(t),
-                value: value.get_value(),
-                exemplars: vec![],
-            });
+            if seen.insert(Arc::as_ptr(&value)) {
```
cijothomas, line 185: Dedup sorted+original attributes!
@utpilla I leveraged Arc pointer to de-dup. Storing MP in diff. array was tried as well, but this looks good to me.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -925,6 +925,62 @@ mod tests {
         assert_eq!(data_point1.value, 6);
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_aggregation_attribute_ordering() {
```
cijothomas, line 929: added this to investigate some issue, that was false alarm. Keeping this test for now. It is not that useful as-is, but can be augmented in subsequent PRs to make sure we cover all scenarios via testing.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -35,42 +39,64 @@ impl<T: Number<T>> ValueMap<T> {
...
+                            new_value.add(measurement);
+                            let new_value = Arc::new(new_value);
+
+                            // Insert original order
+                            values.insert(attrs.to_vec(), new_value.clone());
+
+                            // Insert sorted order
+                            values.insert(sorted_attrs, new_value);
+
+                            self.count.fetch_add(1, Ordering::SeqCst);
```
utpilla, line 88: nit: Let's either remove the explicitly added `return` statements from each of these `if`, `else` blocks or have them added for each `if`, `else` block.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -125,12 +150,9 @@ impl<T: Number<T>> Sum<T> {
...
 
-        let mut values = match self.value_map.values.write() {
-            Ok(v) => v,
-            Err(_) => return (0, None),
-        };
-
-        let n = values.len() + 1;
+        // Max number of data points need to account for the special casing
+        // of the no attribute value + overflow attribute.
+        let n = self.value_map.count.load(Ordering::Relaxed) + 2;
```
utpilla, line 155: Let's start with `SeqCst` and then investigate if we could relax the ordering without affecting correctness?

```suggestion
        let n = self.value_map.count.load(Ordering::SeqCst) + 2;
```

utpilla, line 155: We should update this calculation for cumulative as well?

cijothomas: yes, but it is not a mandatory step, as the vec will just grow if pre-allocated size is not big enough. In cumulative, once grown, it remains in that size forever, so not much of an issue.
Will make the change to make it consistent, but this is an area (exporting), that needs revisit/refactor.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -925,6 +925,62 @@ mod tests {
...
+        // to be treated as same attributes)
+        // start with unsorted order
+
+        let attribute_values = [
+            "value1", "value2", "value3", "value4", "value5", "value6", "value7", "value8",
+            "value9", "value10",
+        ];
+        let mut rng = rngs::SmallRng::from_entropy();
+
+        for _ in 0..1000000 {
```
utpilla, line 948: Let's lower the number of iterations to have a faster test execution time.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -925,6 +925,62 @@ mod tests {
...
+                    KeyValue::new("attribute4", attribute_values[index_fourth_attribute]),
+                ],
+            );
+        }
+
+        test_context.flush_metrics();
+
+        let sum = test_context.get_aggregation::<data::Sum<u64>>("my_counter", None);
+
+        // Expecting 1 time-series.
```
utpilla, line 974:
```suggestion
        // Expecting 1600 time-series.
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -925,6 +925,62 @@ mod tests {
...
+        assert_eq!(sum.data_points.len(), 1600);
+
+        // find and validate key1=value2 datapoint
```
utpilla, line 977:
```suggestion
        // find and validate metric overflow datapoint
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -925,6 +925,62 @@ mod tests {
...
+        let data_point1 =
```
utpilla, line 978: nit: Rename the variable to `overflow_datapoint`.

stress/src/metrics_counter.rs
```diff
@@ -4,6 +4,9 @@
     Hardware: Intel(R) Xeon(R) Platinum 8370C CPU @ 2.80GHz, 16vCPUs,
     RAM: 64.0 GB
     ~9.5 M /sec
+
+    Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
```
utpilla, line 8: You probably intended to replace the previous entry instead of appending a new one.

cijothomas: keeping both :D as we have AMD and Intel machines, so no harm keeping both :)

utpilla: Okay 😀 We should update the stress test results for Intel then. It should have gone up from 9.5 M/sec.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -35,42 +39,59 @@ impl<T: Number<T>> ValueMap<T> {
...
+
+                            // Insert original order
+                            values.insert(attrs.to_vec(), new_value.clone());
+
+                            // Insert sorted order
+                            values.insert(sorted_attrs, new_value);
+
+                            self.count.fetch_add(1, Ordering::SeqCst);
+                        } else if let Some(overflow_value) =
+                            values.get_mut(STREAM_OVERFLOW_ATTRIBUTES.as_slice())
```
utpilla, line 86: I think we could just use the `get` method here.

```suggestion
                            values.get(STREAM_OVERFLOW_ATTRIBUTES.as_slice())
```

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -152,23 +169,29 @@ impl<T: Number<T>> Sum<T> {
...
+                    exemplars: vec![],
+                });
+            }
         }
 
         // The delta collection cycle resets.
         if let Ok(mut start) = self.start.lock() {
             *start = t;
         }
+        self.value_map.count.store(0, Ordering::Release);
```
utpilla, line 194: Use `SeqCst` ordering for this and also the `Precomputed` one.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -125,12 +145,9 @@ impl<T: Number<T>> Sum<T> {
...
 
-        let mut values = match self.value_map.values.write() {
-            Ok(v) => v,
-            Err(_) => return (0, None),
-        };
-
-        let n = values.len() + 1;
+        // Max number of data points need to account for the special casing
+        // of the no attribute value + overflow attribute.
+        let n = self.value_map.count.load(Ordering::SeqCst) + 2;
```
utpilla, line 150: I'm not sure if this calculation is important for correctness of the export. If it is, then we should do this under the `self.value_map.values.write()` lock. Otherwise, we read a certain value here and before we reserve capacity for data points, some update thread(s) could enter the write lock and increment the count further.

cijothomas: It does not look important for correctness. Even if it is off by few, it just results in vec resizing in the middle of collect.
We can revisit this.

utpilla (approved):
Great work!

PR #1991, opened 2024-08-06
Metric benchmark - more isolation

cijothomas:


opentelemetry-sdk/benches/metric_counter.rs
```diff
@@ -79,9 +83,16 @@ fn counter_add(c: &mut Criterion) {
             );
         });
     });
+}
 
-    c.bench_function("Counter_Add_Unsorted", |b| {
-        let counter = create_counter("Counter_Add_Unsorted");
+fn counter_add_unsorted(c: &mut Criterion) {
+    let attribute_values = [
```
utpilla, line 89: Could we make this a static array? It has the same value for every benchmark.

utpilla (approved):

opentelemetry-sdk/benches/metric_counter.rs
cijothomas: done

PR #1644, opened 2024-03-26
Fix observable counter cumulative aggregation

stormshield-fabs:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1517 (incorrect cumulative aggregation for observable counters).

## Changes

The observable counters rely on `PrecomputedSum` that was using the same `ValueMap` implementation as `Sum` (its counterpart for synchronous counters), effectively adding all observations (also drastically reducing the counter range).

This PR specializes `ValueMap` with an `O` (operation) parameter so that observations can be performed with the correct semantics (`AddAssign` for synchronous counters, `Assign` for observable ones). It also updates the `PrecomputedSum::cumulative` aggregation to return the current value.

The tests (`observable_counter_delta` and `observable_counter_cumulative`) are from https://github.com/open-telemetry/opentelemetry-rust/pull/1516 with tiny fixes.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (none)

PR #1992, opened 2024-08-07
Fix few Observable Counter and UpDownCounter bugs

### LINKED ISSUE #1517: ObservableCounter aggregation bug for cumulative
Observable Counter/UpDownCounter has aggregation bug while using cumulative temporality.
Shown in https://github.com/open-telemetry/opentelemetry-rust/pull/1516

This is the tracking issue for the fix.

cijothomas:
Fixes #1517 and other bugs. Added tests.
Copying ideas from @stormshield-fabs 's original PR: https://github.com/open-telemetry/opentelemetry-rust/pull/1644

Additionally, the previous implementation of the Observable relied on the assumption that the ValueMap is drained after ever collect. This is an area that could be tweaked in the future. This PR removed that dependency anyway, so this is better future proof.

Focused on fixing the bug only. The code is poorly structured now, and could benefit from a refactoring. The Aggregation could be made generic similar to what was done in #1644 or other ways to make it manageable. Currently, there are tons of code duplication. Will attempt this after ensuring all known bugs are fixed and there is good test coverage to confidently do major refactorings.


opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -25,40 +25,54 @@ struct ValueMap<T: Number<T>> {
...
+            assign_only,
         }
     }
 }
 
 impl<T: Number<T>> ValueMap<T> {
     fn measure(&self, measurement: T, attrs: &[KeyValue]) {
         if attrs.is_empty() {
-            self.no_attribute_value.add(measurement);
+            if self.assign_only {
```
cijothomas, line 52: @stormshield-fabs Could you help refactor this to use generics, so we dont have this duplication everywhere and get better perf?

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -629,8 +667,9 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[ignore = "Spatial aggregation is not yet implemented."]
```
cijothomas, line 670: this got accidentally fixed in some other PR, but this PR breaks it. spatial aggregation is complex and needs a separate discussion altogether!

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -16,7 +16,10 @@
   The custom exporters and processors can't directly access the `LogData::LogRecord::attributes`, as
   these are private to opentelemetry-sdk. Instead, they would now use LogRecord::attributes_iter()
   method to access them.
-
+- Fixed various Metric aggregation bug related to
+  ObservableCounter,UpDownCounter including
+  [1517](https://github.com/open-telemetry/opentelemetry-rust/issues/1517).
+  [#1990](https://github.com/open-telemetry/opentelemetry-rust/pull/1990)
```
utpilla, line 22:
```suggestion
  [#1992](https://github.com/open-telemetry/opentelemetry-rust/pull/1992)
```

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -25,40 +25,54 @@ struct ValueMap<T: Number<T>> {
     has_no_value_attribute_value: AtomicBool,
     no_attribute_value: T::AtomicTracker,
     count: AtomicUsize,
+    assign_only: bool, // if true, only assign incoming value, instead of adding to existing value
```
utpilla, line 28: Could we rename this to something more intuitive like `is_sum_precomputed`? I think that would better explain the reasoning behind the logic when reading the code:

```rust
if self.is_sum_precomputed {
    value_to_update.store(measurement);
} else {
    value_to_update.add(measurement);
}
```

opentelemetry-sdk/src/metrics/internal/sum.rs
utpilla: Let's extract this repeated snippet to a method for now? It would also make it easier to review the future PR that would refactor it using generics.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -334,11 +368,16 @@ impl<T: Number<T>> PrecomputedSum<T> {
             .has_no_value_attribute_value
             .swap(false, Ordering::AcqRel)
         {
+            let default = T::default();
+            let delta = self.value_map.no_attribute_value.get_value()
+                - *reported.get(&vec![]).unwrap_or(&default);
+            new_reported.insert(vec![], self.value_map.no_attribute_value.get_value());
```
utpilla, line 374: We should be storing the computed `delta` value here and not do another atomic read. Otherwise, we risk storing an incorrect value from another update thread.

opentelemetry-sdk/src/metrics/mod.rs
lalitb: If there is no issue to track this, probably good to have one.

opentelemetry-sdk/src/metrics/internal/sum.rs
lalitb: Good catch. The `self.value_map.no_attribute_value` fetched once in line 372  should be used as `new_reported` value instead of fetching again.

cijothomas:
Closing in favor of https://github.com/open-telemetry/opentelemetry-rust/pull/2004

PR #1993, opened 2024-08-07
Box complex types in AnyValue enum (Improve perf by 10%, and size reduction by ~60%)

### LINKED ISSUE #1903: Performance Degradation with AnyValue Enum
As discussed in linked issue, There is a noticeable performance degradation when using the [AnyValue ](https://github.com/open-telemetry/opentelemetry-rust/blob/d5c86d9adc6f0e3f2612dcfd92b8b64e88ab20d7/opentelemetry/src/logs/record.rs#L51) enum compared to the [Value](https://github.com/open-telemetry/opentelemetry-rust/blob/d5c86d9adc6f0e3f2612dcfd92b8b64e88ab20d7/opentelemetry/src/common.rs#L243) enum. This needs to be investigated further.

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry-rust/pull/1901#discussion_r1658025252_

lalitb:
Fixes #1903

## Changes

The AnyValue enum has been modified to box complex types, reducing the effective size of the enum for commonly used scenarios like integers and strings.

Existing:
```rust
pub enum AnyValue {
    /// An integer value
    Int(i64),
    /// A double value
    Double(f64),
    /// A string value
    String(StringValue),
    /// A boolean value
    Boolean(bool),
    /// A byte array
    Bytes(Box<Vec<u8>>),
    /// An array of `Any` values
    ListAny(Box<Vec<AnyValue>>),
    /// A map of string keys to `Any` values, arbitrarily nested.
    Map(Box<HashMap<Key, AnyValue>>),
}
```

Modified:

```rust
pub enum AnyValue {
    /// An integer value
    Int(i64),
    /// A double value
    Double(f64),
    /// A string value
    String(StringValue),
    /// A boolean value
    Boolean(bool),
    /// A byte array
    Bytes(Vec<u8>),
    /// An array of `Any` values
    ListAny(Vec<AnyValue>),
    /// A map of string keys to `Any` values, arbitrarily nested.
    Map(HashMap<Key, AnyValue>),
}
```

Since the size of an enum in Rust is determined by the size of its largest variant, boxing the complex types (which are rarely used) can reduce the effective size of the AnyValue enum for the more frequently used cases like integers and strings.

**Size Comparison**
Refer - https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9d2c7b8cc2a603afe23284816faae242

```
Without Box: 56 bytes
With Box: 24 bytes
```

By boxing, the overall size of the enum is reduced, leading to potential performance improvements. Complex types already allocate `data` on the heap; boxing simply means that both the `data` and `metadata` will now be heap-allocated.

**Performance Benchmark**:

**opentelemetry-sdk/benches -  full-log-with-4-attributes**
PR:  116.32 ns 
Main: 127.22 ns

**opentelemetry-appender-tracing - ot-layer-enabled**
PR:  305.00 ns 
Main: 325.00 ns

### DIFF
--- a/opentelemetry-appender-log/Cargo.toml
+++ b/opentelemetry-appender-log/Cargo.toml
@@ -18,6 +18,7 @@ serde = { workspace = true, optional = true, features = ["std"] }
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled"]
 with-serde = ["log/kv_serde", "serde"]
+default = ["logs_level_enabled", "with-serde"]
 
 [dev-dependencies]
 opentelemetry_sdk = { path = "../opentelemetry-sdk", features = [ "testing", "logs_level_enabled" ] }
--- a/opentelemetry-appender-log/Cargo.toml
+++ b/opentelemetry-appender-log/Cargo.toml
@@ -18,7 +18,6 @@ serde = { workspace = true, optional = true, features = ["std"] }
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled"]
 with-serde = ["log/kv_serde", "serde"]
-default = ["logs_level_enabled", "with-serde"]
 
 [dev-dependencies]
 opentelemetry_sdk = { path = "../opentelemetry-sdk", features = [ "testing", "logs_level_enabled" ] }
--- a/opentelemetry/CHANGELOG.md
+++ b/opentelemetry/CHANGELOG.md
@@ -45,7 +45,7 @@ pub enum AnyValue {
 }
 ```
 So the custom log appenders should box these types while adding them in message body, or
-attribute values. Similarlty, the custom exporters should derefernce these complex type values
+attribute values. Similarly, the custom exporters should dereference these complex type values
 before serializing.
 
 ## v0.24.0
--- a/opentelemetry/CHANGELOG.md
+++ b/opentelemetry/CHANGELOG.md
@@ -2,6 +2,52 @@
 
 ## vNext
 
+- **BREAKING** [#1993](https://github.com/open-telemetry/opentelemetry-rust/pull/1993) Box complex types in AnyValue enum
+Before:
+```rust
+#[derive(Debug, Clone, PartialEq)]
+pub enum AnyValue {
+    /// An integer value
+    Int(i64),
+    /// A double value
+    Double(f64),
+    /// A string value
+    String(StringValue),
+    /// A boolean value
+    Boolean(bool),
+    /// A byte array
+    Bytes(Box<Vec<u8>>),
+    /// An array of `Any` values
+    ListAny(Box<Vec<AnyValue>>),
+    /// A map of string keys to `Any` values, arbitrarily nested.
+    Map(Box<HashMap<Key, AnyValue>>),
+}
+```
+
+After:
+```rust
+#[derive(Debug, Clone, PartialEq)]
+pub enum AnyValue {
+    /// An integer value
+    Int(i64),
+    /// A double value
+    Double(f64),
+    /// A string value
+    String(StringValue),
+    /// A boolean value
+    Boolean(bool),
+    /// A byte array
+    Bytes(Vec<u8>),
+    /// An array of `Any` values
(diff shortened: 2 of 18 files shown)


opentelemetry/src/logs/record.rs
```diff
@@ -59,11 +59,11 @@ pub enum AnyValue {
     /// A boolean value
     Boolean(bool),
     /// A byte array
-    Bytes(Vec<u8>),
+    Bytes(Box<Vec<u8>>),
```
cijothomas, line 62: please add changelog entry as this is breaking change.

opentelemetry-appender-log/Cargo.toml
```diff
@@ -18,6 +18,7 @@ serde = { workspace = true, optional = true, features = ["std"] }
 [features]
 logs_level_enabled = ["opentelemetry/logs_level_enabled"]
 with-serde = ["log/kv_serde", "serde"]
+default = ["logs_level_enabled", "with-serde"]
```
cijothomas, line 21: is this needed?

cijothomas (approved):
Thanks for continuing to make perf get better! 
LGTM with a changelog entry.

Could you also share stress tests with memory stats to see if it can show the improvement in memory as well?

opentelemetry-appender-log/Cargo.toml
lalitb: Oh, this was only for testing. Good catch. Let me remove it.

lalitb:
> Thanks for continuing to make perf get better! LGTM with a changelog entry.
> 
> Could you also share stress tests with memory stats to see if it can show the improvement in memory as well?

Here are the memory stats. We can see ~10% reduction.

PR:
```
Throughput: 37,162,800 iterations/sec
Memory usage: 3.62 MB
CPU usage: 99.490875%
Virtual memory usage: 2249.08 MB


Throughput: 37,184,400 iterations/sec
Memory usage: 3.62 MB
CPU usage: 100%
Virtual memory usage: 2249.08 MB
```

Main:
```
Throughput: 35,348,600 iterations/sec
Memory usage: 4.12 MB
CPU usage: 99.83837%
Virtual memory usage: 2249.08 MB


Throughput: 36,159,000 iterations/sec
Memory usage: 4.12 MB
CPU usage: 99.863014%
Virtual memory usage: 2249.08 MB


Throughput: 36,223,800 iterations/sec
Memory usage: 4.25 MB
CPU usage: 99.78797%
Virtual memory usage: 2249.08 MB
```

, I think we can get better/detailed insights by integrating custom allocator (jemalloc) with the stress tes

opentelemetry/src/logs/record.rs
lalitb: Done

opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,52 @@
 
 ## vNext
 
+- **BREAKING** [#1993](https://github.com/open-telemetry/opentelemetry-rust/pull/1993) Box complex types in AnyValue enum
+Before:
```
utpilla, line 6: It looks like you have mistakenly switched Before and After.

opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,52 @@
...
+    /// A byte array
+    Bytes(Vec<u8>),
+    /// An array of `Any` values
+    ListAny(Vec<AnyValue>),
+    /// A map of string keys to `Any` values, arbitrarily nested.
+    Map(HashMap<Key, AnyValue>),
+}
+```
+So the custom log appenders should box these types while adding them in message body, or
+attribute values. Similarlty, the custom exporters should derefernce these complex type values
```
utpilla, line 48:
```suggestion
attribute values. Similarly, the custom exporters should dereference these complex type values
```

opentelemetry/CHANGELOG.md
lalitb: Oh good catch. Fixed now :)

utpilla (approved):

PR #1994, opened 2024-08-07
Export event-name in OTLP Exporter optionally, if the feature-flag is enabled

### LINKED ISSUE #1941: [Log] EventName should be exporter under feature flag only
We currently export EventName (populated by Log, Tracing appenders to LogRecord), as an attribute named "name" in OTLP.  Given no stable spec/convention covering this attribute, this should be under feature flag only, so users know it is not stable and can potentially change in the future.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-proto/src/transform/logs.rs#L103

lalitb:
Fixes #1941

## Changes

Added feature flag `populate-logs-event-name` in `opentelemetry-proto` and `opentelemetry-otlp` crates. The feature flag in the `opentelemetry-proto` crate will be enabled when it is explicitly enabled for the `opentelemetry-otlp` crate. And this will result in `LogRecord::event_name` to be populated in the OTLP log attributes.

TBD - Update Changelog.


opentelemetry-proto/Cargo.toml
```diff
@@ -29,7 +29,7 @@ path = "tests/json_deserialize.rs"
 
 
 [features]
-default = ["full"]
+populate-logs-event-name = []
```
cijothomas, line 32: why change default?

cijothomas, line 32: is there an easy way to document the feature-flags in general? For this one, we'd like to document that this is experimental, and the attribute name is "name", and is subject to change.

opentelemetry-proto/src/transform/logs.rs
```diff
@@ -98,6 +98,7 @@ pub mod tonic {
                         .map(Attributes::from_iter)
                         .unwrap_or_default()
                         .0;
+                    #[cfg(feature = "populate-logs-event-name")]
```
cijothomas, line 101: should we do similar to consoleexporter?

opentelemetry-proto/Cargo.toml
lalitb: There is no feature in this crate. So `default` had no effect. And with feature `populate-logs-event-name`. we want this to be switched off by default.

opentelemetry-proto/src/transform/logs.rs
lalitb: Yes done.

opentelemetry-proto/Cargo.toml
lalitb: We can document in README.md for this crate, or else in Cargo.toml itself, as done by serde.rs - https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml

cijothomas: https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/#feature-flags Lets add it here itself along with rest of other flags.

opentelemetry-otlp/Cargo.toml
```diff
@@ -56,6 +56,7 @@ temp-env = { workspace = true }
 trace = ["opentelemetry/trace", "opentelemetry_sdk/trace", "opentelemetry-proto/trace"]
 metrics = ["opentelemetry/metrics", "opentelemetry_sdk/metrics", "opentelemetry-proto/metrics"]
 logs = ["opentelemetry/logs", "opentelemetry_sdk/logs", "opentelemetry-proto/logs"]
+populate-logs-event-name = ["opentelemetry-proto/populate-logs-event-name"]
```
cijothomas, line 59: please add changelog too, as this is now opt-in, so affecting end users.

opentelemetry-proto/Cargo.toml
lalitb: done.

opentelemetry-otlp/Cargo.toml
lalitb: done

opentelemetry-proto/src/transform/logs.rs
```diff
@@ -98,14 +98,22 @@ pub mod tonic {
...
                         })
                         .collect();
-                    if let Some(event_name) = log_record.event_name.as_ref() {
-                        attributes.push(KeyValue {
-                            key: "name".into(),
-                            value: Some(AnyValue {
-                                value: Some(Value::StringValue(event_name.to_string())),
-                            }),
-                        });
+                    #[cfg(feature = "populate-logs-event-name")]
```
utpilla, line 101: I think it would be simpler to use `cfg!` macro instead of the attribute here:

```rust
let mut attributes: Vec<KeyValue> = log_record
    .attributes_iter()
    .map(|kv| KeyValue {
        key: kv.0.to_string(),
        value: Some(AnyValue {
            value: Some(kv.1.clone().into()),
        }),
    })
    .collect();

if cfg!(feature = "populate-logs-event-name") {
    if let Some(event_name) = &log_record.event_name {
        attributes.push(KeyValue {
            key: "name".into(),
            value: Some(AnyValue {
                value: Some(Value::StringValue(event_name.to_string())),
            }),
        });
    }
}
attributes
```

lalitb: I think `cfg!` macro has runtime overhead, so the code (both if/else branch) would be part of the compiled binary.

opentelemetry-stdout/Cargo.toml
```diff
@@ -20,6 +20,7 @@ default = ["trace", "metrics", "logs"]
 trace = ["opentelemetry/trace", "opentelemetry_sdk/trace", "futures-util"]
 metrics = ["async-trait", "opentelemetry/metrics", "opentelemetry_sdk/metrics"]
 logs = ["opentelemetry/logs", "opentelemetry_sdk/logs", "async-trait", "thiserror", "opentelemetry_sdk/logs_level_enabled"]
+populate-logs-event-name = []
```
cijothomas, line 23: please add changelog and possibly doc for this too.

cijothomas (approved):
LGTM, please add changelog for stdout too.

opentelemetry-proto/src/transform/logs.rs
```diff
@@ -98,14 +98,22 @@ pub mod tonic {
...
+                    {
```
utpilla, line 102: Remove the extra set of braces wrapping the `if/else` block? You don't have that for the stdout exporter.

utpilla (approved):

opentelemetry-proto/src/transform/logs.rs
lalitb: yeah it's weird but I got the lint error without the braces - https://github.com/open-telemetry/opentelemetry-rust/actions/runs/10292895062/job/28488187666?pr=1994

lalitb:
> LGTM, please add changelog for stdout too.

done.

Issue #1995, opened 2024-08-07
Log Benchmarks are repeatedily creating LogProviders


cijothomas:
As https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/log.rs#L52 line is executed repeatedly for each iteration, the cost that the benchmark measures include the cost of creating Provider itself. It could be minor, but we should only be testing the true of Log emit only, as creating LoggerProvider, Logger is one time, and emission is the repeating code for end users.

Similar issue was noted in metrics and fixed : https://github.com/open-telemetry/opentelemetry-rust/pull/1991

We need to review every benchmarks to fix this issue, but priority is to get the log ones fixed, so it is easily to measure the impact of all the active work on improving Log performance @lalitb is doing.  Hopefully, this can also reduce the noise/variations we are observing making it difficult to compare gains.

PR #1996, opened 2024-08-07
Add benchmarks for measuring cost of Logger, Provider creation

cijothomas:
Related to #1995 
This adds a dedicated benchmark to show the cost of creating provider, and logger. These are expected to be one-time operation and end users are expected to re-use them.

### DIFF
--- a/opentelemetry-sdk/benches/log.rs
+++ b/opentelemetry-sdk/benches/log.rs
@@ -2,6 +2,17 @@
 //! So to run test named "full-log-with-attributes/with-context" you would run `$ cargo bench --bench log -- --exact full-log-with-attributes/with-context`
 //! To run all tests for logs you would run `$ cargo bench --bench log`
 //!
+/*
+The benchmark results:
+criterion = "0.5.1"
+OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
+Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+RAM: 64.0 GB
+| Test                           | Average time|
+|--------------------------------|-------------|
+| Logger_Creation                |  30 ns      |
+| LoggerProvider_Creation        | 909 ns      |
+*/
 
 use std::collections::HashMap;
 use std::time::SystemTime;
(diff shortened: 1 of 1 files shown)


lalitb (approved):

opentelemetry-sdk/benches/log.rs
```diff
@@ -2,6 +2,17 @@
...
+/*
+The benchmark results:
+criterion = "0.5.1"
+OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
+Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
+RAM: 64.0 GB
+| Test                           | Average time|
+|--------------------------------|-------------|
+| Logger_Creation                |  30 ns      |
+| LoggerProvider_Creation        | 909 ns      |
```
cijothomas, line 14: Something doesn't add up. This shows the cost of provider creation is 900ns. https://github.com/open-telemetry/opentelemetry-rust/issues/1995 says provider is repeatedly created for each log, but yet, the benchmarks for that is hardly 200ns. 
@lalitb @utpilla we need to spend some time understanding this!

utpilla: I did some investigation into this. Here's what I observed:
1. The number of times the code inside `b.iter` is called is way higher than the number of times the code outside (which is under `c.bench_func`).
2. The code under `b.iter` is what gets benchmarked meaning the time measurement reported is for the time spent in running the code inside `b.iter` and NOT `c.bench_func`.


Here is a sample benchmark to demonstrate this behavior:

```rust
fn criterion_benchmark(c: &mut Criterion) {
    benchmark_test(c);
}

fn benchmark_test(c: &mut Criterion) {
    let mut func_count: u64 = 0;
    let mut iter_count: u64 = 0;
    c.bench_function("Test", |b| {

        wait_micro_seconds(10); // spin for 10 micro seconds
        func_count += 1; // increment count to check how many times this is executed

        b.iter(|| {
            wait_micro_seconds(5); // spin for 5 micro seconds
            iter_count += 1; // increment count to check how many times this is executed
        });
    });

    println!("func_count: {:?}", func_count);
    println!("iter_count: {:?}", iter_count);
}

fn wait_micro_seconds(duration: u128) {
    let start = Instant::now();
    let mut elapsed = 0;
    while elapsed < duration {
        elapsed = start.elapsed().as_micros();
    }
}
```

The time reported by the benchmark is ~5.06 us which is roughly the time spent in running the code inside `b.iter` in this example.

Also, I'm printing the counts after the benchmark has run. Here is the count that was printed on my machine:
```
func_count: 120
iter_count: 2038375
```

Since, the difference between the counts is huge, my guess would be that the code which is inside just `bench_func` is only run during warm-up (to compute the number of iterations for the actual benchmark which is inside `b.iter`).

cijothomas: That makes sense. Lets make all benchmarks properly isolated to avoid any confusions/false alarms.

lalitb: So, just to be clear, any code running between the two closures won't be reported in the benchmark results, even though it is repeatedly run in lesser count?

Issue #1953, opened 2024-07-23
OTel Logs with no heap allocation


cijothomas:
Opening a parent issue to track making it possible to do OTel Logging with zero heap allocation. This requires support from
1. Logging library (`tracing` is used as example, which already has zero allocation and delegates most work to layer/subscriber)
2. OTel Bridge API - this delegates most work to OTel Logging SDK and has no heap allocation now.
3. OTel Log SDK - Currently, this is where all heap allocation occurs!
4. Exporter - Exporting to etw, user_events can be synchronously, and hence can be fully done without heap allocation.

Making OTel Log SDK heap allocation free requires addressing the below at the minimum:
1. https://github.com/open-telemetry/opentelemetry-rust/issues/1919 - Avoid vec! for passing `LogRecord` to Exporter.
2. https://github.com/open-telemetry/opentelemetry-rust/issues/1920 - Avoid allocating vec! for storing upto "N" attributes
3. https://github.com/open-telemetry/opentelemetry-rust/issues/1921 - Avoid String allocation. For `tracing` mostly
4. https://github.com/open-telemetry/opentelemetry-rust/pull/2795 - Fixed unnecessarily allocating string
5. https://github.com/open-telemetry/opentelemetry-rust/pull/2799 target no longer allocates

**For sync exporters**

The overall flow would be (using `tracing` as example)
tracing -> otel-tracing-layer -> create `LogRecord` on stack, populate it using callback/visitor from `tracing-subscriber`. -> pass mut ref of `LogRecord` to exporting processor -> exporter ->serialize and write to transport (etw/user-events/lttng) 

The above can be achieved fully on stack. This issue is targeting the above only.

**For async exporters**
For exporters requiring Batching for efficiency (eg: OTLP over gRPC, HTTP), some Ring buffer or Channels would be needed. There would be clone/copy costs for those scenarios, but it maybe possible to avoid heap allocation entirely. This will be tracked separately.

Also need some tooling to ensure the assumptions are correct. i.e measure the actual heap allocation and confirm it is indeed 0! Above is based on manual code inspection only.

cijothomas:
Closing as we now have fixed all allocation by default in the SDK except https://github.com/open-telemetry/opentelemetry-rust/issues/1921 which requires changes in `tracing` itself.

[...]

PR #1997, opened 2024-08-07
LogAppender - 5% perf gain and reduce heap alloc for SeverityText

cijothomas:
Part of https://github.com/open-telemetry/opentelemetry-rust/issues/1953
Gains are between 5-10% in my machine.

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -168,7 +168,7 @@ where
         log_record.set_target(meta.target().to_string());
         log_record.set_event_name(meta.name());
         log_record.set_severity_number(severity_of_level(meta.level()));
-        log_record.set_severity_text(meta.level().to_string().into());
+        log_record.set_severity_text(severity_text_of_level(meta.level()).into());
         let mut visitor = EventVisitor::new(&mut log_record);
         #[cfg(feature = "experimental_metadata_attributes")]
         visitor.visit_experimental_metadata(meta);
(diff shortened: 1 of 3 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -201,6 +201,16 @@ const fn severity_of_level(level: &Level) -> Severity {
     }
 }
 
+const fn severity_text_of_level(level: &Level) -> &'static str {
+    match *level {
+        Level::TRACE => "TRACE",
```
cijothomas, line 206: `Severity::Warn.name()` could have been used as well, but I think this would be the fastest. I believe compiler is smart enough to avoid storing the string duplicates...

lalitb:
Nicely done.

lalitb (approved):

opentelemetry-appender-tracing/src/layer.rs
utpilla: We can get rid of this method. `meta.level().as_str()` does exactly this.

https://github.com/tokio-rs/tracing/blob/master/tracing-core/src/metadata.rs#L524-L535

cijothomas: 🤣 I am considering removing populating severitytext at all, and let exporter figure it out from the severitynumber.

utpilla: Haha okay. Let's do that right away or remove this method until make the change to stop populating severity text?

PR #1999, opened 2024-08-07
[logs-appender-tracing] Use method from tracing-core instead of creating our own

utpilla:
Follow-up to #1997 

## Changes
- Use `Level.as_str()` method from the tracing-core create instead of creating our own mapping method


lalitb (approved):
LGTM.

cijothomas (approved):

PR #2004, opened 2024-08-08
Use generics to dispatch updates in `ValueMap`

stormshield-fabs:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/1517 (incorrect cumulative aggregation for observable counters).

This is a follow-up to https://github.com/open-telemetry/opentelemetry-rust/pull/1992 as I cannot push to the original branch.

It re-introduces the ideas from https://github.com/open-telemetry/opentelemetry-rust/pull/1644 : the `ValueMap` with an `O` (operation) parameter so that observations can be performed with the correct semantics (`Increment` for synchronous counters, `Assign` for observable ones). 

I've accounted for the pending review comments on the initial PR and refactored the `measure` method a bit to limit right-ward drift.

Edit to include the additional gains with this change:

Additionally, the previous implementation of the Observable relied on the assumption that the ValueMap is drained after ever collect. This is an area that could be tweaked in the future. This PR removed that dependency anyway, so this is better future proof.


opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -19,89 +20,122 @@ use super::{aggregate::is_under_cardinality_limit, AtomicTracker, Number};
 pub(crate) static STREAM_OVERFLOW_ATTRIBUTES: Lazy<Vec<KeyValue>> =
     Lazy::new(|| vec![KeyValue::new("otel.metric.overflow", "true")]);
 
+/// Abstracts the update operation for a measurement.
+trait Operation {
+    fn update_tracker<T: 'static>(tracker: &dyn AtomicTracker<T>, value: T);
```
cijothomas, line 25: Can we avoid the dyn dispatch here? This is metrics hot path, and we need to try and make it as fast as possible.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -331,29 +365,30 @@ impl<T: Number<T>> PrecomputedSum<T> {
...
         };
 
-        let default = T::default();
         for (attrs, value) in values.drain() {
-            let delta = value.get_value() - *reported.get(&attrs).unwrap_or(&default);
-            if delta != default {
-                new_reported.insert(attrs.clone(), value.get_value());
-            }
+            let delta = value.get_value() - *reported.get(&attrs).unwrap_or(&T::default());
+            new_reported.insert(attrs.clone(), value.get_value());
```
utpilla, line 391: Similar to the no-attributes case, we shouldn't do a new atomic read on `value` here.

lalitb: @utpilla I believe this looks fine, as the hashmap is locked (in line 384) here during collect cycle, so there won't be any inserts happening.

utpilla: Ah yes.

utpilla: We could still avoid paying the cost of another atomic read. Since, we take a lock, we are certain that the value of the variable wouldn't change from what we get on the first atomic read (on line 390). We could simply use that value.

Not a major perf benefit though as this is not the hot path. 😄

opentelemetry-sdk/src/metrics/internal/sum.rs
stormshield-fabs: Done

opentelemetry-sdk/src/metrics/internal/sum.rs
stormshield-fabs: Ah didn't see this one, fixed.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -16,7 +16,10 @@
   The custom exporters and processors can't directly access the `LogData::LogRecord::attributes`, as
   these are private to opentelemetry-sdk. Instead, they would now use LogRecord::attributes_iter()
   method to access them.
-
+- Fixed various Metric aggregation bug related to
+  ObservableCounter,UpDownCounter including
+  [#1517](https://github.com/open-telemetry/opentelemetry-rust/issues/1517).
+  [#1992](https://github.com/open-telemetry/opentelemetry-rust/pull/1992)
```
utpilla, line 22:
```suggestion
  [#2004](https://github.com/open-telemetry/opentelemetry-rust/pull/2004)
```

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -19,89 +20,122 @@ use super::{aggregate::is_under_cardinality_limit, AtomicTracker, Number};
 pub(crate) static STREAM_OVERFLOW_ATTRIBUTES: Lazy<Vec<KeyValue>> =
     Lazy::new(|| vec![KeyValue::new("otel.metric.overflow", "true")]);
 
+/// Abstracts the update operation for a measurement.
+trait Operation {
```
utpilla, line 24: Another way to implement a customizable behavior for update operation would be to use a function pointer. We could add a new field to the `ValueMap` struct and set the desired behavior for update during the struct creation and use the function pointer to execute the update logic.

```rust
   ...
   update: fn(&T::AtomicTracker, T),
}

    fn new(is_sum_precomputed: bool) -> Self {
        ValueMap {
            values: RwLock::new(HashMap::new()),
            has_no_value_attribute_value: AtomicBool::new(false),
            no_attribute_value: T::new_atomic_tracker(),
            count: AtomicUsize::new(0),
            update: if is_sum_precomputed { Self::store } else { Self::add },
        }
    }

    fn store(tracker: &T::AtomicTracker, measurement: T) {
        tracker.store(measurement);
    }

    fn add(tracker: &T::AtomicTracker, measurement: T) {
        tracker.add(measurement);
    }
}

// Use (self.update)(&value_to_update, measurement); to update the atomic tracker value in the measure method.
```

@stormshield-fabs What are your thoughts on using generics vs function pointers? Any interesting pros or cons?

stormshield-fabs: I think this works too and is much more readable! 
It might introduce an additional call indirection (since the compiler can inline the generic version) but it's probably not super impactful.

I can update this PR to use function pointers if the other reviewers agree

cijothomas: Generic's advantage of allowing inline is worth keeping the generic approach.. (Its hard to measure the impact with our benchmarks i think.. this is most likely within noise level of that!)

Let us go with the generic now, and revisit as we work through rest of Metrics SDK issues.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -19,89 +20,122 @@ use super::{aggregate::is_under_cardinality_limit, AtomicTracker, Number};
...
-            }
+impl<T: Number<T>, O: Operation> ValueMap<T, O> {
+    fn measure(&self, measurement: T, attributes: &[KeyValue]) {
+        if attributes.is_empty() {
+            O::update_tracker(&self.no_attribute_tracker, measurement);
+            self.has_no_attribute_value.store(true, Ordering::Release);
+            return;
+        }
+
+        let Ok(trackers) = self.trackers.read() else {
```
cijothomas, line 86: thanks for reorganizing this! One of the next steps (in another PR) is to add internal metrics whenever we are dropping a measurement.

cijothomas (approved):
LGTM, Thanks for the extra cleanups.

utpilla (approved):

Issue #1975, opened 2024-07-30
[Bug]: Metrics - Gauge stops reporting after being recorded


cpaika:
### What happened?

In our application we periodically record a Gauge<u64> every 60 seconds.  We use the `opentelemetry-prometheus` exporter and host a metrics endpoint where we scrape our prometheus metrics from OpenTelemetry.

We have other metrics (Histogram, Counter) that always show up on our prometheus metrics endpoint.  However for gauges, if more than a second goes by between calls to `record` for the gauge - the gauge stops being exposed on our prometheus endpoint, the metric value, description, and type all dissapear.  It then returns when we record the Gauge. 

We tried switching to the `opentelemetry_stdout` library instead of prometheus and saw the same behavior - the gauge would disappear from the metrics exporter within a second or two after it was last recorded, and then reappear when we recorded a new value.

For context, our prometheus instance scrapes our application periodically every 5s.  So with the Gauge intermittently deregistering itself, we have a race condition where we may not be recording the Gauge value.

Is this known, intended behavior? I wasn't able to find a reference to this in the [opentelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#gauge) on gauges.  If it is intended - is there any way to customize the Gauge expiration so we can have it persist in the metrics provider context for longer?

### API Version

0.24

### SDK Version

0.24

### What Exporter(s) are you seeing the problem on?

Prometheus

cijothomas:
Can you confirm if you are using ObservableGauge or Gauge?

cpaika:
> Can you confirm if you are using ObservableGauge or Gauge?

Yes we're using Gauge, not the `ObservableGauge`

cijothomas:
I tried the below and it seems working fine for me.. Could you compare what is the difference between the repo app and this? 

For Gauge, as long as new values are reported, it'll be exported.. Maybe that is the wrong behavior?


```
use opentelemetry::global;
use opentelemetry::KeyValue;
use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
use opentelemetry_sdk::{runtime, Resource};
use std::error::Error;
use std::time::Duration;

fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default()
        // uncomment the below lines to pretty print output.
        //  .with_encoder(|writer, data|
[...]

PR #2010, opened 2024-08-12
Add a special test for synchronous instruments (cumulative aggregation)

utpilla:
## Changes
- Add a test to ensure that synchronous instruments (for cumulative aggregation) continue to report the previously reported values when no new measurements are made between the previous and current collect

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -1251,7 +1251,7 @@ mod tests {
         counter_multithreaded_aggregation_helper(Temporality::Cumulative);
     }
 
-    #[tokio::test(flavor= "multi_thread", worker_threads = 1)]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1251,6 +1251,118 @@ mod tests {
...
+
+        synchronous_instruments_cumulative_with_gap_in_measurements_helper("counter");
+        synchronous_instruments_cumulative_with_gap_in_measurements_helper("updown_counter");
+        synchronous_instruments_cumulative_with_gap_in_measurements_helper("histogram");
+
+        /* Synchronous Gauge has an aggregation bug. Uncomment the code below to run the test for gauge
+        once this issue is fixed: https://github.com/open-telemetry/opentelemetry-rust/issues/1975
+        */
+
+        // synchronous_instruments_cumulative_with_gap_in_measurements_helper("gauge");
```
utpilla, line 1267: This test fails for synchronous Gauge. We need to fix #1975 before we can run the test for Gauge.

cijothomas (approved):
Thanks!

PR #2015, opened 2024-08-13
Remove ability to specify observable callback on Meter

cijothomas:
Continuation of https://github.com/open-telemetry/opentelemetry-rust/pull/1715
This actually removes the ability to register callbacks on Meter directly. Initially, I was considering to keep this under experimental flag as suggested by @TommyCpp, but given we have to make it compile always, the effort is not worth the effort.

Even after removal of this, the currently documented Observable instruments is exposing too much and is *not* working the way it is expected to. The following example shows a user is able to call `observe` directly on the returned instrument. This was *not* the intend behind observable instruments. I'll do further cleanups for Observables in general, but this is just 1st step in keeping public API under control, and more importantly, keeping scopes in control for stable release.

```
    // Create a ObservableCounter instrument and register a callback that reports the measurement.
    let observable_counter = meter
        .with_description("My observable counter example description")
        .with_unit("myunit")
        .with_callback(|observer| {
            observer.observe(
                100,
                &[
                ],
            )
        })
        .init();

   // the following code should be throwing errors, but it is working just fine, making
  // observable operate like sync instruments only!
    observable_counter.observe(150, &[
        KeyValue::new("foo", "f"),
        KeyValue::new("bar", "b"),
    ]); 
```

There is no ability to "unregister" callback. That is not required by the spec either. If there is need for one, even with the current implementation, we should be able to show how to achieve that.
(To be clear, the ability to unregister was not really there earlier too from what I can tell...)

### DIFF
--- a/opentelemetry/CHANGELOG.md
+++ b/opentelemetry/CHANGELOG.md
@@ -48,10 +48,12 @@ So the custom log appenders should box these types while adding them in message 
 attribute values. Similarly, the custom exporters should dereference these complex type values
 before serializing.
 
-*Breaking* : Removed the ability to register callbacks for Observable
-instruments on Meter directly. If you were using `meter.register_callback` to
-provide the callback, provide them using `with_callback` method, while creating
-the Observable instrument itself.
+*Breaking* :
+[#2015](https://github.com/open-telemetry/opentelemetry-rust/pull/2015) Removed
+the ability to register callbacks for Observable instruments on Meter directly.
+If you were using `meter.register_callback` to provide the callback, provide
+them using `with_callback` method, while creating the Observable instrument
+itself.
 [1715](https://github.com/open-telemetry/opentelemetry-rust/pull/1715/files)
 shows the exact changes needed to make this migration. If you are starting new,
 refer to the
--- a/opentelemetry/CHANGELOG.md
+++ b/opentelemetry/CHANGELOG.md
@@ -49,7 +49,14 @@ attribute values. Similarly, the custom exporters should dereference these compl
 before serializing.
 
 *Breaking* : Removed the ability to register callbacks for Observable
-instruments on Meter directly.
+instruments on Meter directly. If you were using `meter.register_callback` to
+provide the callback, provide them using `with_callback` method, while creating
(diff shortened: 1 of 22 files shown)


opentelemetry/src/metrics/meter.rs
```diff
@@ -66,31 +63,43 @@ pub trait MeterProvider {
     ) -> Meter;
 }
 
-/// Provides access to instrument instances for recording measurements.
-///
-/// ```
+/// Provides the ability to create instruments for recording measurements or
```
cijothomas, line 66: Will send a follow up to make doc comments better for entire Metric API soon.

opentelemetry/CHANGELOG.md
```diff
@@ -48,6 +48,16 @@ So the custom log appenders should box these types while adding them in message
...
 
+*Breaking* : Removed the ability to register callbacks for Observable
+instruments on Meter directly. If you were using `meter.register_callback` to
+provide the callback, provide them using `with_callback` method, while creating
+the Observable instrument itself.
+[1715](https://github.com/open-telemetry/opentelemetry-rust/pull/1715/files)
+shows the exact changes needed to make this migration. If you are starting new,
+refer to the
+[examples](https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/metrics-basic/src/main.rs)
+to learn how to provide Observable callbacks.
```
utpilla, line 59: Add a link to this PR as well?

utpilla (approved):

lalitb (approved):

cijothomas:
@TommyCpp I am proceeding with merging this to continue progress. I understand you/Lalit earlier raised concerns in the linked discussion but this seems the right way to make the API easier to use and reduce the amount of code/scope.

neoeinstein:
Is there a recommendation for how to manage "epoch" based metrics? For example, I had been using this to advance the `jemalloc` epoch for cached statistics _once_ per observable collection. Is the expectation that I should now be advancing the epoch for each of the instruments in each of the callbacks? This seems like it will cause a reasonably significant regression in performance for that collection, but I can give it a try.

PR #2018, opened 2024-08-13
Fix metrics aggregation bug for Sum and Precomputed Sum to avoid duplicate export

utpilla:
We missed adding a few `HashSet` checks to avoid duplicate exports in #1989. The affected aggregations are:
- Sum (Cumulative)
- PrecomputedSum (Delta)
- PrecomputedSum (Cumulative) 

## Changes
- Fix the affected aggregations
- Update the existing related tests to test for both Delta and Cumulative temporality

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -835,72 +835,77 @@ mod tests {
         // Run this test with stdout enabled to see output.
         // cargo test counter_aggregation_attribute_order_sorted_first --features=testing -- --nocapture
 
-        // Arrange
-        let mut test_context = TestContext::new(Temporality::Delta);
-        let counter = test_context.u64_counter("test", "my_counter", None);
+        counter_aggregation_attribute_order_sorted_first_helper(Temporality::Delta);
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -835,145 +835,155 @@ mod tests {
...
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn counter_aggregation_attribute_order_unsorted_first() {
         // Run this test with stdout enabled to see output.
         // cargo test counter_aggregation_attribute_order_unsorted_first --features=testing -- --nocapture
 
-        // Arrange
-        let mut test_context = TestContext::new(Temporality::Delta);
-        let counter = test_context.u64_counter("test", "my_counter", None);
+        counter_aggregation_attribute_order_unsorted_first_helper(Temporality::Delta);
+        counter_aggregation_attribute_order_unsorted_first_helper(Temporality::Cumulative);
```
utpilla, line 917: The test for cumulative aggregation would fail without this fix.

cijothomas (approved):
more gaps in testing identified.
As mentioned in community call, I'll do a full check on the overall coverage, and reorganize the tests so it is easy to tell what is covered and what is not!

Thanks!

PR #2020, opened 2024-08-13
Benchmark tests for batch log exporter:

lalitb:
## Changes

We are missing the benchmark tests on the BatchLogExporter, which means any change in the log pipeline doesn't measure it's effect on the batch pipeline. Adding tests to measure that.
Two tests are added:
- Measure time to create, emit and flush a log record
- Measure time to create and emit enough logs to trigger export and wait for export completion

Will enhance/extend them further when required.

### DIFF
--- a/opentelemetry-sdk/benches/batch_log_processor.rs
+++ b/opentelemetry-sdk/benches/batch_log_processor.rs
@@ -1,0 +1,138 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use opentelemetry::logs::LoggerProvider as _;
+use opentelemetry::logs::Severity;
+use opentelemetry::logs::{LogRecord, Logger};
+use opentelemetry_sdk::export::logs::LogExporter;
+use opentelemetry_sdk::logs::LoggerProvider;
+use opentelemetry_sdk::logs::{BatchConfigBuilder, BatchLogProcessor};
+use opentelemetry_sdk::resource::Resource;
+use opentelemetry_sdk::runtime::Tokio;
+use std::sync::{Arc, Condvar, Mutex};
+use std::time::SystemTime;
+use tokio::runtime::Runtime;
+
+#[derive(Debug)]
+struct NoopLogExporter;
+
+#[async_trait::async_trait]
+impl LogExporter for NoopLogExporter {
+    async fn export<'a>(
+        &mut self,
+        _batch: Vec<std::borrow::Cow<'a, opentelemetry_sdk::export::logs::LogData>>,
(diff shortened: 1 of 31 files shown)


opentelemetry-sdk/benches/batch_log_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+        Ok(())
+    }
+
+    fn shutdown(&mut self) {}
+
+    fn set_resource(&mut self, _resource: &Resource) {}
+}
+
+fn benchmark_batch_log_processor(c: &mut Criterion) {
+    // Benchmark 1: Measure time to emit and flush a log record
```
cijothomas, line 55: can you make a benchmark to know the hot path cost alone? Benchmark 1 is doing flush and benchmark 2 is doing many items till export gets triggered, so neither is matching the cost typically expected by the user which is the cost of creating logrecord, calling emit, passing to processor, cloning by batch processor, queueing to the channel.

opentelemetry-sdk/benches/batch_log_processor.rs
```diff
@@ -0,0 +1,138 @@
+use criterion::{criterion_group, criterion_main, Criterion};
```
cijothomas, line 1: please share the bench results for easy reference!

lalitb: Have added it. This mayn't be highly accurate, I believe slightly over-reported. But for now, we have baseline to compare. And can be improved over.

opentelemetry-sdk/benches/batch_log_processor.rs
lalitb: Yes, @utpilla also suggested same :). Have done now. Thanks.

opentelemetry-sdk/benches/batch_log_processor.rs
```diff
@@ -0,0 +1,95 @@
...
+                .build();
+            let logger = provider.logger("test-logger");
+
+            b.iter_custom(|iters| {
+                let mut count = 0;
+                let mut total_duration = std::time::Duration::new(0, 0);
+                for _ in 0..iters {
+                    if count == max_queue_size - 1 {
+                        // flush the queue outside the timing loop
+                        provider.force_flush();
```
cijothomas, line 70: what is the need for this flush?

lalitb: The queue gets full rapidly without this, even with a 1sec upload interval.

lalitb:
closing it for #2026

PR #2021, opened 2024-08-13
Fix aggregation for Gauge

### LINKED ISSUE #1975: [Bug]: Metrics - Gauge stops reporting after being recorded

utpilla:
Fixes #1975 

## Changes
- Added dedicated collect handling logic for Delta and Cumulative temporality for Gauge
- Added a test to ensure that asynchronous instruments with Cumulative temporality keep reporting the last recorded measurement if there was no new measurement made from the last collect

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -106,11 +106,7 @@ impl<T: Number<T>> AggregateBuilder<T> {
     }
 
     /// Builds a last-value aggregate function input and output.
-    ///
-    /// [Builder::temporality] is ignored and delta is always used.
     pub(crate) fn last_value(&self) -> (impl Measure<T>, impl ComputeAggregation) {
-        // Delta temporality is the only temporality that makes semantic sense for
-        // a last-value aggregate.
         let lv_filter = Arc::new(LastValue::new());
         let lv_agg = Arc::clone(&lv_filter);
         let t = self.temporality;
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -113,6 +113,7 @@ impl<T: Number<T>> AggregateBuilder<T> {
         // a last-value aggregate.
         let lv_filter = Arc::new(LastValue::new());
         let lv_agg = Arc::clone(&lv_filter);
+        let t = self.temporality;
 
         (
             self.filter(move |n, a: &[KeyValue]| lv_filter.measure(n, a)),
@@ -127,7 +128,12 @@ impl<T: Number<T>> AggregateBuilder<T> {
                 };
                 let g = g.unwrap_or_else(|| new_agg.as_mut().expect("present if g is none"));
 
-                lv_agg.compute_aggregation(&mut g.data_points);
+                match t {
+                    Some(Temporality::Delta) => {
+                        lv_agg.compute_aggregation_delta(&mut g.data_points)
+                    }
+                    _ => lv_agg.compute_aggregation_cumulative(&mut g.data_points),
+                }
 
                 (g.data_points.len(), new_agg.map(|a| Box::new(a) as Box<_>))
             },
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1404,6 +1400,133 @@ mod tests {
...
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn asynchronous_instruments_cumulative_with_gap_in_measurements() {
+        // Run this test with stdout enabled to see output.
+        // cargo test asynchronous_instruments_cumulative_with_gap_in_measurements --features=testing -- --nocapture
+
+        asynchronous_instruments_cumulative_with_gap_in_measurements_helper("counter");
+        asynchronous_instruments_cumulative_with_gap_in_measurements_helper("updown_counter");
+        asynchronous_instruments_cumulative_with_gap_in_measurements_helper("gauge");
```
utpilla, line 1410: The test for `Gauge` would fail without this fix.

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -113,6 +113,7 @@ impl<T: Number<T>> AggregateBuilder<T> {
         // a last-value aggregate.
         let lv_filter = Arc::new(LastValue::new());
         let lv_agg = Arc::clone(&lv_filter);
+        let t = self.temporality;
```
lalitb, line 116: nit - should we update the function comment, as the temporality is no longer ignored now?

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -127,7 +128,12 @@ impl<T: Number<T>> AggregateBuilder<T> {
                 };
                 let g = g.unwrap_or_else(|| new_agg.as_mut().expect("present if g is none"));
 
-                lv_agg.compute_aggregation(&mut g.data_points);
+                match t {
+                    Some(Temporality::Delta) => {
+                        lv_agg.compute_aggregation_delta(&mut g.data_points)
```
lalitb, line 133: nit - how about naming these methods delta() and cumulative(), to be consistent with naming in other aggregates?

lalitb (approved):

cijothomas (approved):
LGTM, please add a changelog entry for the bugfix

opentelemetry-sdk/src/metrics/internal/aggregate.rs
utpilla: The code setup for `last_value` is not consistent with other aggregates. `delta` and `cumulative` of other aggregates return a tuple `(usize, Option<Box<dyn Aggregation>>)` whereas the `compute_aggregation` methods of `last_value` do not return anything.

I would like to make the code setup for `Gauge` consistent with other aggregates in a different PR. Is it okay to fix it then?

opentelemetry-sdk/src/metrics/internal/aggregate.rs
utpilla: Thanks! Updated.

PR #2024, opened 2024-08-14
Refactor Metric test - part1

cijothomas:
Added a md file to track the key scenarios. It is not filled correctly, that'll be done over next several PRs.
Rearrange, rename Counter tests, to match the order in md file.

If the overall direction is reasonable, I can extend this to all tests.

(Code re-use is still minimal. That can be handled separately)

### DIFF
--- a/opentelemetry-sdk/src/metrics/test.md
+++ b/opentelemetry-sdk/src/metrics/test.md
@@ -1,0 +1,24 @@
+# Test Coverage for Metric Instruments
+
+Tests are located in [mod.rs](mod.rs)
+
+// TODO: Fill this correctly.
+
+## Sync Instruments
+
+| Test Type                  | Counter (Delta) | Counter (Cumulative) | UpDownCounter (Delta) | UpDownCounter (Cumulative) | Gauge (Delta) | Gauge (Cumulative) | Histogram (Delta) | Histogram (Cumulative) |
+|----------------------------|-----------------|----------------------|----------------------|----------------------------|---------------|--------------------|-------------------|------------------------|
+| Regular aggregation test   | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| No-attribute test          | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| Overflow test              | [Yes]         | [No]               | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| Attr Order Sorted First    | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/test.md
```diff
@@ -0,0 +1,24 @@
...
+|----------------------------|-----------------|----------------------|----------------------|----------------------------|---------------|--------------------|-------------------|------------------------|
+| Regular aggregation test   | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| No-attribute test          | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| Overflow test              | [Yes]         | [No]               | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| Attr Order Sorted First    | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| Attr Order Unsorted First  | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
```
utpilla, line 15: Need to add entries for multi-threaded test and other data types such as `f64` and `i64`.

opentelemetry-sdk/src/metrics/test.md
```diff
@@ -0,0 +1,24 @@
...
+Tests are located in [mod.rs](mod.rs)
+
+// TODO: Fill this correctly.
+
+## Sync Instruments
+
+| Test Type                  | Counter (Delta) | Counter (Cumulative) | UpDownCounter (Delta) | UpDownCounter (Cumulative) | Gauge (Delta) | Gauge (Cumulative) | Histogram (Delta) | Histogram (Cumulative) |
+|----------------------------|-----------------|----------------------|----------------------|----------------------------|---------------|--------------------|-------------------|------------------------|
+| Regular aggregation test   | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
+| No-attribute test          | [Yes]         | [Yes]              | [No]              | [No]                    | [No]      | [No]            | [No]           | [No]                |
```
utpilla, line 12: For implementation of these tests, I think we could combine regular aggregation and no-attribute into a single test to keep the code size for tests less. Check this [test](https://github.com/open-telemetry/opentelemetry-rust/blob/b6ab7287ff787eab93b9150ba1fd5d8a1334f15d/opentelemetry-sdk/src/metrics/mod.rs#L1807-L1809) for example which conditionally emits a zero-attribute measurement along with the regular measurements.

opentelemetry-sdk/src/metrics/test.md
utpilla: Also, need to add entries for special scenarios such as testing the export results when no new measurement has been emitted since the last collect.

utpilla (approved):
Building a test matrix is a good step towards ensuring good test coverage. Left some comments about adding more entries to the test matrix. We could keep updating it as required.

PR #2027, opened 2024-08-14
Add log export benchmark to measure the cost paid for async abstraction

cijothomas:
The current export method is marked async. Even when scenarios where export is done synchronously, the async is causing some unnecessary overheads (likely related to creating/updating the async state machines etc...)
This PR just shows the overhead is ~10% (250 ns ->275 ns). Not sure if this is mimicking the actual exporter correctly, would love to get some feedback on this.

### DIFF
--- a/opentelemetry-sdk/benches/log_exporter.rs
+++ b/opentelemetry-sdk/benches/log_exporter.rs
@@ -10,20 +10,17 @@
     | LogExporterWithoutFuture       | 255 ns      |
 */
 
-
 use std::sync::Mutex;
 use std::time::SystemTime;
 
 use async_trait::async_trait;
 use criterion::{criterion_group, criterion_main, Criterion};
 
-use opentelemetry::logs::{
-   LogRecord as _, LogResult, Logger as _, LoggerProvider as _, Severity,
-};
+use opentelemetry::logs::{LogRecord as _, LogResult, Logger as _, LoggerProvider as _, Severity};
 
 use opentelemetry_sdk::export::logs::LogData;
 use opentelemetry_sdk::logs::LogProcessor;
-use opentelemetry_sdk::logs::{LoggerProvider};
+use opentelemetry_sdk::logs::LoggerProvider;
 use pprof::criterion::{Output, PProfProfiler};
 use std::fmt::Debug;
 
@@ -34,8 +31,8 @@ pub trait LogExporterWithFuture: Send + Sync + Debug {
     async fn export(&mut self, batch: Vec<LogData>);
 }
 
-pub trait LogExporterWithoutFuture : Send + Sync + Debug {
-    fn export(&mut self, batch: Vec<LogData>);    
+pub trait LogExporterWithoutFuture: Send + Sync + Debug {
+    fn export(&mut self, batch: Vec<LogData>);
 }
 
 #[derive(Debug)]
@@ -43,15 +40,13 @@ struct NoOpExporterWithFuture {}
 
 #[async_trait]
 impl LogExporterWithFuture for NoOpExporterWithFuture {
-    async fn export(&mut self, _batch: Vec<LogData>){
-    }
+    async fn export(&mut self, _batch: Vec<LogData>) {}
 }
 
 #[derive(Debug)]
 struct NoOpExporterWithoutFuture {}
 impl LogExporterWithoutFuture for NoOpExporterWithoutFuture {
-    fn export(&mut self, _batch: Vec<LogData>) {
-    }
+    fn export(&mut self, _batch: Vec<LogData>) {}
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/benches/log_exporter.rs
```diff
@@ -0,0 +1,178 @@
...
+        ExportingProcessorWithFuture {
+            exporter: Mutex::new(exporter),
+        }
+    }
+}
+
+impl LogProcessor for ExportingProcessorWithFuture {
+    fn emit(&self, data: &mut LogData) {
+        let mut exporter = self.exporter.lock().expect("lock error");
+        futures_executor::block_on(exporter.export(vec![data.clone()]));
```
TommyCpp, line 68: Do we block on runtime our LogProcrssor? It's generally not really performat

TommyCpp: We could try https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on and see if there is difference. Maybe spin up the runtime as part of the Processor so we can reuse them

cijothomas:
> So this is measuring the case when the user doesn't have an async runtime?

Its immaterial if user has async runtime or not. The exporter (eg: exporters to systems like Windows ETW, Linux user_events) does not need async runtime, as it is executing synchronously.

TommyCpp:
> So

OK so it's testing the case where the exporter doesn't need async runtime. Does it covers the case where the exporter requires async runtime. e.g HTTP or gRPC exporter based on async HTTP and gRPC crate?

TommyCpp (commented):
> likely related to creating/updating the async state machines etc.

State machine should create during compiling time. Given the async function doesn't do anything the state machine probably only have one state. I think the overhead is mostly coming from the runtime setup here rather than the state machine itself.

cijothomas:
> > Its immaterial if user has async runtime or not. The exporter (eg: exporters to systems like Windows ETW, Linux user_events) does not need async runtime, as it is executing synchronously.
> 
> OK so it's testing the case where the exporter doesn't need async runtime. Does it covers the case where the exporter requires async runtime. e.g HTTP or gRPC exporter based on async HTTP and gRPC crate?

It doesn't.

opentelemetry-sdk/benches/log_exporter.rs
lalitb: +1. This is not testing the async runtime. Better to create the async runtime outside of the timing loop, and then block using tokio runtime.

cijothomas: Modified to use Tokio runtime's `block_on`, and  perf drastically dropped. (280ns to 350 ns)
```
#[derive(Debug)]
struct ExportingProcessorWithFuture {
    exporter: Mutex<NoOpExporterWithFuture>,
    rt: Runtime,
}

impl ExportingProcessorWithFuture {
    fn new(exporter: NoOpExporterWithFuture) -> Self {
        let rt = Runtime::new().unwrap();
        ExportingProcessorWithFuture {
            exporter: Mutex::new(exporter),
            rt,      
        }
    }
}

impl LogProcessor for ExportingProcessorWithFuture {
    fn emit(&self, data: &mut LogData) {
        let mut exporter = self.exporter.lock().expect("lock error");
        // futures_executor::block_on(exporter.export(vec![data.clone()]));
        self.rt.block_on(exporter.export(vec![data.clone()]));
    }

    fn force_flush(&self) -> LogResult<()> {
        Ok(())
    }

    fn shutdown(&self) -> LogResult<()> {
        Ok(())
    }
}
```

TommyCpp:
> There is no runtime here

`futures_executor::block_on` is a runtime(well, maybe `async executor` will be a better name) as async tasks doesn't run without a execturor

> [All asynchronous computation occurs within an executor](https://docs.rs/futures-executor/latest/futures_executor/)

It doesn't change tha fact async tasks has overhead because of the need of exector

opentelemetry-sdk/benches/log_exporter.rs
```diff
@@ -0,0 +1,178 @@
...
+#[derive(Debug)]
+struct NoOpExporterWithoutFuture {}
+impl LogExporterWithoutFuture for NoOpExporterWithoutFuture {
+    fn export(&mut self, _batch: Vec<LogData>) {}
+}
+
+#[derive(Debug)]
+struct ExportingProcessorWithFuture {
+    exporter: Mutex<NoOpExporterWithFuture>,
+}
```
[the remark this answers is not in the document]
cijothomas: want planning to merge this as is. I started with a simpler example then slowly moved to match log exporter scenario and ended up with this. 😂
The whole point was to validate if async has overhead or not based on the last sig discussion.

TommyCpp:
> The discussion about runtime etc. is not very relevant (apologies my knowledge of async runtime is very limited).

Yep. Sorry I kind of hijacked the discussion 😞 


> I have some workarounds

Would it make sense to offer two versions(one for sync, one for async). It's not uncommon in Rust eco system. One example would be `reqwest`, which offers `async` and [blokcing](https://docs.rs/reqwest/latest/reqwest/blocking/index.html) implementation

TommyCpp (approved):
Benchmark looks good. I have a few thoughts on how to setup runtime/try differnt runtime to minimize the cost. I can explore it offline once we merge it

lalitb (approved):

cijothomas:
Opened an issue to track the change : https://github.com/open-telemetry/opentelemetry-rust/issues/2031
Wasn't planning to merge this benchmark originally, but no harm in keeping it either.

PR #2017, opened 2024-08-13
Use ValueMap for Gauge - Throughput increased around 8x

utpilla:
Follow-up to #2012 
Similar to #1833 and #1989 from @cijothomas 

## Changes
- Update `Gauge` to use `ValueMap`

<details>
<summary>Machine information</summary>
OS: Ubuntu 22.04.3 LTS (5.15.146.1-microsoft-standard-WSL2)
Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
RAM: 64.0 GB
</details>

### Benchmarks

There is a **~63%** increase in benchmark performance. This is coming from `ValueMap`'s optimized lookup on the hot path. 

| Gauge_Add_With_Three_Random_Attributes  | Average time|
|----------------------------------------------------|---------------|
| main branch                                                       | 483.78 ns     |
| **PR**                                                          | **178.37 ns**     |

### Stress Test

That's an improvement of nearly 8x for throughput. This is coming from `ValueMap`'s reduced thread contention and optimized lookup on the hot path.

| 16 threads updating random trackers               | Throughput|
|----------------------------------------------------|---------------|
| main branch                                                       |   1.5 M/sec |
| **PR**                                                                 |  11.5 M/sec   |

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #2033, opened 2024-08-16
Use ValueMap for Histogram- Throughput increased by 5x

utpilla:
Follow-up to #2012 
Similar to #1833. #1989, #2017 

## Changes
- Update `Histogram` to use `ValueMap`

<details>
<summary>Machine information</summary>
RAM: 64.0 GB
</details>

### Benchmarks

There is a **~60%** increase in benchmark performance. This is coming from `ValueMap`'s optimized lookup on the hot path. 

| Histogram_Record_With_Three_Random_Attributes  | Average time |
|----------------------------------------------------------|--------------- |
| main branch                                                               | 509.21 n         |
| **PR**                                                                         | **193.04 ns**|

### Stress Test

Throughput has improved by 5 times. This is coming from `ValueMap`'s reduced thread contention and optimized lookup on the hot path.

| 16 threads updating random trackers | Throughput |
|-------------------------------------|------------|
| main branch                         | 1.8 M/sec  |
| **PR**                              | **9 M/sec**|

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/histogram.rs
@@ -1,14 +1,99 @@
-use std::{collections::HashMap, sync::Mutex, time::SystemTime};
+use std::any::Any;
+use std::collections::HashSet;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use std::{sync::Mutex, time::SystemTime};
 
+use crate::metrics::data::HistogramDataPoint;
 use crate::metrics::data::{self, Aggregation, Temporality};
-use crate::{metrics::data::HistogramDataPoint, metrics::AttributeSet};
 use opentelemetry::KeyValue;
-use opentelemetry::{global, metrics::MetricsError};
 
-use super::{
-    aggregate::{is_under_cardinality_limit, STREAM_OVERFLOW_ATTRIBUTE_SET},
-    Number,
-};
+use super::Number;
+use super::{AtomicTracker, AtomicallyUpdate, Operation, ValueMap};
+
+struct HistogramUpdate;
+
+impl Operation for HistogramUpdate {
+    fn update_tracker<T: 'static, AT: AtomicTracker<T>>(tracker: &AT, value: T, index: usize) {
+        println!("index:{:?}", index);
+        tracker.update_histogram(index, value);
+    }
+}
+
+struct HistogramTracker<T> {
+    buckets: Mutex<Buckets<T>>,
+}
+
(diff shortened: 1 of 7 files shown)


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -48,40 +48,54 @@ impl Operation for Assign {
 ///
 /// This structure is parametrized by an `Operation` that indicates how
 /// updates to the underlying value trackers should be performed.
-pub(crate) struct ValueMap<T: Number<T>, O> {
+pub(crate) struct ValueMap<AU: AtomicallyUpdate<T>, T: Number<T>, O> {
```
utpilla, line 51: `ValueMap` struct takes three generic parameters now. There is scope to further refactor this. We could update `AtomicTracker` in a way that would allow us to get rid of `Operation` trait. I didn't want to move too many things around in this PR.

opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -1,14 +1,49 @@
...
+
+impl<T: Number<T>> AtomicTracker<T> for HistogramTracker<T> {
+    fn update_histogram(&self, index: usize, value: T) {
+        let mut buckets = match self.buckets.lock() {
+            Ok(guard) => guard,
+            Err(_) => return,
+        };
+
+        buckets.bin(index, value);
+        buckets.sum(value);
```
utpilla, line 33: One difference now is that we would always update the sum. Previously, this was happening only when `self.record_sum` is `true`. I think we are not required to provide `record_sum` as a configuration option in the first place.

cijothomas: spec:

> 
Arithmetic sum of Measurement values in population. This SHOULD NOT be collected when used with instruments that record negative measurements (e.g. UpDownCounter or ObservableGauge).

This scenario occurs only if a View is used to aggregate UpDownCounter/Gauge as Histogram. Lets address this (if we chose to), in a future PR.

cijothomas (approved):
Nice gains!
The generic/awkwardness is not a blocker as they are purely internal implementations, and can be improved without breaking.

opentelemetry-sdk/src/metrics/internal/histogram.rs
cijothomas: Additionally, I am unsure if SUM should be not reported, when -ve measurements are provided, which are currently not blockeable at API level for f64...

Proto wording is not super clear to me...Lets follow up later.
https://github.com/open-telemetry/opentelemetry-proto/blob/v0.9.0/opentelemetry/proto/metrics/v1/metrics.proto#L538-L543

PR #2035, opened 2024-08-19
Add lifetime specifier with LogData struct members - 10% perf improvement log pipeline

lalitb:
## Changes

1. The Exporter::export interface is modified to accept the vector of  pair/tuple of reference to LogRecord and InstrumentationLib, instead of Cow of LogData.

Before:
```rust
async fn export<'a>(&mut self, batch: Vec<Cow<'a, LogData>>) -> LogResult<()>;
```

After:
```rust
async fn export( &mut self, batch: Vec<(&LogRecord, &InstrumentationLibrary)>) -> LogResult<()>;
```
This simplifies the processing for exporters, as they don't need to check whether the object is borrowed, or owned.  Exporter needs to ensure to create the copy of LogRecord and/or InstrumentationLib whenever needed.

2. The `LogData` structure is not part of the export interface. So it has been moved from `opentelemetry_sdk::export::logs` to `opentelemetry_sdk::logs::` namespace. It may be in the future removed altogether.
3. The LogData structure has been changed as below:

Before:
```rust
#[derive(Clone, Debug)]
pub struct LogData {
    /// Log record
    pub record: LogRecord,
    /// Instrumentation details for the emitter who produced this `LogEvent`.
    pub instrumentation: InstrumentationLibrary,
}
```
 
After:
```rust
#[derive(Clone, Debug)]
pub struct LogData<'a> {
    /// Log record, which can be borrowed or owned.
    pub record: Cow<'a, LogRecord>,
    /// Instrumentation details for the emitter who produced this `LogEvent`.
    pub instrumentation: Cow<'a, InstrumentationLibrary>,
}
```
This allows us to avoid copying the LogRecord structure when used with a synchronous processor.
    - Simple processor (and also other concurrent processors) gets the reference to the LogData structure, borrows it, and sends it to the exporter. This involves NO copy/move. The exporter can decide whether to own it based on its logic.
    - The batch processor gets the reference to the LogData structure, converts it into the owned version, and pushes it to the queue. This involves clone, which effectively copies the attributes too. This can't be avoided.

The perf matrix:
| Benchmark                                    | Metric                     | Main Result                              | PR Result                                | Change (%)               | Performance |
|----------------------------------------------|----------------------------|------------------------------------------|------------------------------------------|--------------------------|-------------|
| **Stress Test (16 Threads)**                 | Throughput (iterations/sec) | 39,377,400                                | 42,628,400                                | +8.26%                   | Improved    |
| **ot_layer_enabled**                         | Time (ns)                  | [256.95 ns, 257.48 ns]                    | [224.21 ns, 225.78 ns]                    | -12.59%                  | Improved    |
| **Logging_Comparable_To_Appender**           | Time (ns)                  | [132.90 ns, 133.09 ns]                    | [106.11 ns, 106.31 ns]                    | -20.16%                  | Improved    |
| **log_noop_processor**                       | Time (ns)                  | [134.45 ns, 134.88 ns]                    | [109.77 ns, 109.97 ns]                    | -18.29%                  | Improved    |
| **log_cloning_processor**                    | Time (ns)                  | [231.31 ns, 232.42 ns]                    | [126.61 ns, 126.83 ns]                    | -45.29%                  | Improved    |
| **log_clone_and_send_to_channel_processor**  | Time (ns)                  | [613.20 ns, 615.45 ns]                    | [600.58 ns, 603.57 ns]                    | -2.22%                   | Improved    |


cijothomas:
Is the change to Exporter trait the reason for perf improvement? Or is the changes to LogProcessor traits? Its not clear which one is the real reason, and also wondering if they are related?

lalitb:
Lots of conflict with #2041. Will close this and better raise a new PR.

PR #2040, opened 2024-08-20
Simplified stdout exporter for logs and traces

cijothomas:
The current implementation of stdout exporter is not very easy to read by default or even with optional json formatting. Modified the stdout exporter implementation to print logs,traces in an easier to read format. (Of course, this is subjective!). I tried to mimic [OTel Collector Debug exporters detailed output format.](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/debugexporter#detailed-verbosity) closely. ~If everyone feels this is better, I'll continue and extend this for Metrics too, and fix up any gaps. (i.e ensure every field is exported etc.) This can also eliminate the transformation code we maintain for stdout exporter.~  Update after initial PR review: All signals are covered now.

Note that the output format is not fixed and users are already advised not to rely on any particular format. Stdout exporter is *ONLY* designed/meant for learning purposes and not any other scenarios.

**Example Logs output**
**Before:**
```txt
{"resourceLogs":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"basic-stdout-example"}},{"key":"telemetry.sdk.name","value":{"stringValue":"opentelemetry"}},{"key":"telemetry.sdk.language","value":{"stringValue":"rust"}},{"key":"telemetry.sdk.version","value":{"stringValue":"0.24.1"}}]},"scopeLogs":[{"scope":{"name":"opentelemetry-appender-tracing","version":"0.5.0"},"logRecords":[{"timeUnixNano":null,"time":null,"observedTimeUnixNano":1724181491423171133,"observedTime":"2024-08-20 19:18:11.423","severityNumber":17,"severityText":"ERROR","attributes":[{"key":"event_id","value":{"intValue":20}},{"key":"user_name","value":{"stringValue":"otel"}},{"key":"user_email","value":{"stringValue":"otel@opentelemetry.io"}}],"droppedAttributesCount":0}]}]}]}
```
**With this PR**
```txt
Resource
         telemetry.sdk.version=String(Static("0.24.1"))
         telemetry.sdk.name=String(Static("opentelemetry"))
         telemetry.sdk.language=String(Static("rust"))
         service.name=String(Static("basic-stdout-example"))
Log #0
         EventName: "my-event-name"
         Target (Scope): "my-system"
         Observed Timestamp: 2024-08-20 19:10:07.275165
         SeverityText: "ERROR"
         SeverityNumber: Error
         Attributes:
                 event_id: Int(20)
                 user_name: String(Owned("otel"))
                 user_email: String(Owned("otel@opentelemetry.io"))
```

Example Span output
**Before:**
```txt
{"resourceSpans":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"basic-stdout-example"}},{"key":"telemetry.sdk.name","value":{"stringValue":"opentelemetry"}},{"key":"telemetry.sdk.language","value":{"stringValue":"rust"}},{"key":"telemetry.sdk.version","value":{"stringValue":"0.24.1"}}]},"scopeSpans":[{"scope":{"name":"stdout-test"},"spans":[{"traceId":"ef2090ff07115b5e7cc26949402d5032","spanId":"0783277ce34d8f0c","parentSpanId":"0000000000000000","name":"test_span","kind":1,"startTimeUnixNano":1724181491423412359,"startTime":"2024-08-20 19:18:11.423","endTimeUnixNano":1724181491423418960,"endTime":"2024-08-20 19:18:11.423","attributes":[{"key":"test_key","value":{"stringValue":"test_value"}}],"droppedAttributesCount":0,"events":[{"name":"test_event","attributes":[{"key":"test_event_key","value":{"stringValue":"test_event_value"}}],"droppedAttributesCount":0,"timeUnixNano":1724181491423417485,"time":"2024-08-20 19:18:11.423"}],"droppedEventsCount":0,"flags":1,"droppedLinksCount":0,"status":{}}]}]}]}
```
**With this PR**
```txt
Resource
         telemetry.sdk.version=String(Static("0.24.1"))
         telemetry.sdk.name=String(Static("opentelemetry"))
         telemetry.sdk.language=String(Static("rust"))
         service.name=String(Static("basic-stdout-example"))
Span #0
         Instrumentation Scope
                 Name: "stdout-example"
                 Version: "v1."
                 SchemaUrl: "schema_url"
                 Scope Attributes:
                         scope_key: String(Static("scope_value"))

         Name: "test_span"
         TraceId: f8d3787cde7f1f976bf0ec2e31cd4ebc
         SpanId: 3688ee24904752ad
         ParentSpanId: 0000000000000000
         Kind: Internal
         Start time: 2024-08-20 19:10:07.275416
         End time: 2024-08-20 19:10:07.275421
         Status: Unset
         Attributes:
         Dropped attributes count: 0
                 test_attribute_key: String(Static("test_attribute_value"))
         Events:
                 Name: "test_event"
                 Timestamp: 2024-08-20 19:10:07.275420
                 Attributes:
                         test_event_key: String(Static("test_event_value"))
```

Metrics output with this PR
```txt
Metrics
Resource
         ->  service.name=String(Static("basic-stdout-example"))
         ->  telemetry.sdk.name=String(Static("opentelemetry"))
         ->  telemetry.sdk.version=String(Static("0.24.1"))
         ->  telemetry.sdk.language=String(Static("rust"))
        Instrumentation Scope #0
                Name         : stdout-example
Metric #0
                Name         : example_counter
                Description  : 
                Unit         : 
                Type         : Sum
                Sum DataPoints
                Monotonic    : true
                DataPoint #0
                        Value        : 2
                        Attributes   :
                                 ->  name: apple
                                 ->  color: green
                DataPoint #1
                        Value        : 2
                        Attributes   :
                                 ->  name: apple
                                 ->  color: red
                DataPoint #2
                        Value        : 12
                        Attributes   :
                                 ->  name: banana
                                 ->  color: yellow
```


opentelemetry-stdout/src/logs/exporter.rs
```diff
@@ -1,38 +1,24 @@
...
-    Box<dyn Fn(&mut dyn Write, crate::logs::transform::LogData) -> LogResult<()> + Send + Sync>;
-
-/// A [`LogExporter`] that writes to [`Stdout`] or other configured [`Write`].
-///
-/// [`LogExporter`]: opentelemetry_sdk::export::logs::LogExporter
-/// [`Write`]: std::io::Write
-/// [`Stdout`]: std::io::Stdout
+/// An OpenTelemetry exporter that writes Logs to stdout on export.
 pub struct LogExporter {
-    writer: Option<Box<dyn Write + Send + Sync>>,
```
utpilla, line 21: I think the good thing about this was it would have allowed users to easily write the output to streams (File, stderr, etc.) other than `stdout`. I would be in favor of keeping the `opentelemetry-stdout` project simple, so I support this change. If there is an actual demand for such an exporter that lets users write to any stream that implements the `Write` trait, we could consider offering such an exporter in the contrib repo.

utpilla: We would need a CHANGELOG entry though as this is a breaking change.

opentelemetry-stdout/src/logs/exporter.rs
```diff
@@ -1,38 +1,24 @@
...
     resource: Resource,
-}
-
-impl LogExporter {
-    /// Create a builder to configure this exporter.
-    pub fn builder() -> LogExporterBuilder {
-        Default::default()
-    }
+    is_shutdown: atomic::AtomicBool,
+    resource_emitted: bool,
```
utpilla, line 12: Similar to `is_shutdown`, `resource_emitted` needs thread synchronization as well?

utpilla, line 12: Also, what's the reason for printing resource information only once? I think it should be okay to keep printing the resource with every export for easy reference. Unlike the collector, `stdout` of the user application could be used for printing a lot of other things, right?

opentelemetry-stdout/src/trace/exporter.rs
```diff
@@ -22,106 +20,136 @@ impl fmt::Debug for SpanExporter {
...
 
-        Box::pin(std::future::ready(res))
+                self.resource.iter().for_each(|(k, v)| {
+                    println!("\t ->  {}={:?}", k, v);
+                });
+
+                print_spans(batch);
+            }
+
+            Box::pin(futures_util::future::ready(Ok(())))
```
utpilla, line 57: Is the change from `std::future::ready()` to `futures_util::future::ready()` intentional? Does this work?
```suggestion
            Box::pin(std::future::ready(Ok(())))
```

opentelemetry-stdout/src/logs/exporter.rs
cijothomas: Do we? Processor ensures export() calls are never done concurrently.

cijothomas: Resource is fixed and won't change.. So I wasn't sure if its worth showing the same info repeatedly! (It kinda pollutes the output, so maybe we can make an opt-in flag to keep printing Resource)

utpilla: Ah yes. I was wondering about the SimpleProcessor case, but we take a Mutex there anyway.

opentelemetry-stdout/src/logs/exporter.rs
cijothomas: There is demand to make an exporter to stdout that can be relied on production. Its not yet spec-ed out, but that is something we should add, once the spec is good. (I think of it as an OTLP Exporter using stdout/file as the transport instead of http/gRPC)

(There is some experimental specs for them I believe)

opentelemetry-stdout/src/logs/exporter.rs
```diff
@@ -44,108 +30,77 @@ impl fmt::Debug for LogExporter {
...
+        if self.is_shutdown.load(atomic::Ordering::SeqCst) {
+            return Err("exporter is shut down".into());
         } else {
-            Err("exporter is shut down".into())
+            if self.resource_emitted {
+                print_logs(batch);
+            } else {
+                self.resource_emitted = true;
+                println!("Resource");
+                if self.resource.schema_url().is_some() {
```
utpilla, line 43: You could simplify this a bit and avoid calling `resource.schema_url()` twice.

```rust
if let Some(schema_url) = self.resource.schema_url() {
    println!("\t Resource SchemaUrl: {:?}", schema_url);
}
``

File: opentelemetry-stdout/src/logs/exporter.rs [hunk 5]
```diff
@@ -44,108 +30,77 @@ impl fmt::Debug for LogExporter {
...
-            + Sync
-            + 'static,
-    {
-        self.encoder = Some(Box::new(encoder));
-        self
-    }
+fn print_logs(batch: Vec<Cow<'_, LogData>>) {
+    for (i, log) in batch.into_iter().enumerate() {
+        println!("Log #{}", i);
+        if let Some(event_name) = &log.record.event_name {
```
utpilla, line 70: Do we not have to print information about `LogData.InstrumentationLibrary`?

cijothomas:
I'll leave this open for a day or two, to make sure others have a chance to share feedback. 
(There might be some merge conflicts with other in-flight Logs PR, but this PR is easy to resolve them, so this can go last anyway)

opentelemetry-stdout/src/logs/exporter.rs
cijothomas: yes and no! Its hard (perf wise) to populate InstrumentationLibrary in hot path, so Log appenders are populating it in target.... At somepoint, the target has to be mapped back to Instrumentation Library. Need to follow up on that later.

opentelemetry-stdout/src/trace/exporter.rs
```diff
@@ -22,106 +20,136 @@ impl fmt::Debug for SpanExporter {
...
-        self.encoder = Some(Box::new(writer));
-        self
-    }
+        print_header = true;
+        for event in span.events.iter() {
+            if print_header {
+                println!("\t Events:");
+                print_header = false;
+            }
+            println!("\t\t Name: {:?}", event.name);
```
TommyCpp, line 120: Why do we want to print debug info for `event.name` here?

TommyCpp: Also we can just use it variable inline `println!("format {local_variable} arguments");`

opentelemetry-stdout/src/trace/exporter.rs
```diff
@@ -22,106 +20,136 @@ impl fmt::Debug for SpanExporter {
...
-            }),
+        print_header = true;
+        for link in span.links.iter() {
+            if print_header {
+                println!("\t Links:");
+                print_header = false;
+            }
+            println!("\t\t TraceId: {:?}", link.span_context.trace_id());
+            println!("\t\t SpanId: {:?}", link.span_context.span_id());
+            println!("\t\t Attributes:");
```
TommyCpp, line 144: Can we do 
```
span.links.iter()
    .enumerate()
    .for_each(|(index, link)| {
        if index == 0 {
            println!("\t Links:");
        }

        println!("\t\t TraceId: {:?}", link.span_context.trace_id());
        println!("\t\t SpanId: {:?}", link.span_context.span_id());
        ...
   })
```
no need for the `print_header` variable, if our intention here is to print once

cijothomas: Thanks! Yea that looks much better!
I believe next version of clippy will be able to detect/suggest these improvements automatically! (it already suggested fixes when I maintained "i" for counting iterations!)

opentelemetry-stdout/src/logs/exporter.rs
cijothomas: Also - for OTel collector, resource can be anything as there could be  any number of apps sending to Collector. (It still tried to print Resource once per batch)

utpilla: Once per batch might offer a better user experience than ONLY once throughout the application lifetime. It all depends on the usage pattern. If a user is emitting a lot of other stuff to stdout and also testing multiple batches of export, then they have to always scroll up on the console to where the resource info was first printed.

lalitb:
Thanks for the effort on simplifying the output. I agree the existing json format (even with serde pretty-print) is not easy read.
I have another suggestion. If we can keep the current design (transform, encoder, writer) intact, and instead provide a a built-in `encoder` that formats logs in a human-readable way (as in this PR) and make it the default option in the builder. This way, we can maintain the existing flexibility provided by the current design by abstracting the writing and encoding process. I understand this is not for production, and most of the users won't care of the existing flexibility, but preserving ensures the advance use-cases, without much of maintenance effort.

cijothomas:
> Thanks for the effort on simplifying the output. I agree the existing json format (even with serde pretty-print) is not easy read. I have another suggestion. If we can keep the current design (transform, encoder, writer) intact, and instead provide a a built-in `encoder` that formats logs/spans in a human-readable way (as in this PR) and make it the default option in the builder. This way, we can maintain the existing flexibility provided by the current design by abstracting the writing and encoding process. I understand this is not for production, and most of the users won't care of the existing flexibility, but preserving ensures the advance use-cases, without much of maintenance effort.

I was initially thinking of keeping existing and providing an option to do the current output format... Then I felt that it'd possibly convey wrong intent that we intend to keep such flexibility! Removing all such flexibility, makes our intent pretty clear - this is only for learning/testing purpose and nothing more.

lalitb:
> > Thanks for the effort on simplifying the output. I agree the existing json format (even with serde pretty-print) is not easy read. I have another suggestion. If we can keep the current design (transform, encoder, writer) intact, and instead provide a a built-in `encoder` that formats logs/spans in a human-readable way (as in this PR) and make it the default option in the builder. This way, we can maintain the existing flexibility provided by the current design by abstracting the writing and encoding process. I understand this is not for production, and most of the users won't care of the existing flexibility, but preserving ensures the advance use-cases, without much of maintenance effort.
> 
> I was initially thinking of keeping existing and providing an option to do the current output format... Then I felt that it'd possibly convey wrong intent that we intend to keep such flexibility! Removing all such flexibility, makes our intent pretty clear - this is only for learning/testing purpose and nothing more.

I see your point in potential confusion coming with the flexibility provided. We can also have a balance by documenting that the existing flexibility doesn’t make the current exporter suitable for any production use-case. No strong opinion though, i eventually go with what all agree :)

cijothomas:
Given there is general acceptance to move towards this PR, I'll continue polishing this (and add metrics too). Marking draft while I work through them and address review feedbacks.

opentelemetry-stdout/src/logs/exporter.rs
cijothomas: Issue with **Once per batch** is that, the default pairing of stdout exporter is simple processor, so the batch has only one item, making it effectively emit Resource with each item!

opentelemetry-stdout/examples/basic.rs
```diff
@@ -64,21 +64,92 @@ fn init_logs() -> opentelemetry_sdk::logs::LoggerProvider {
...
-    span.set_attribute(KeyValue::new("test_key", "test_value"));
+    use opentelemetry::trace::{
+        SpanContext, SpanId, TraceFlags, TraceId, TraceState, TracerProvider,
+    };
+
+    let tracer = global::tracer_provider()
+        .tracer_builder("stdout-example")
+        .with_version("v1")
+        .with_schema_url("schema_url")
+        .with_attributes(vec![KeyValue::new("scope_key", "scope_value")])
```
utpilla, line 75: Let's showcase using arrays instead of vec in examples.

```suggestion
        .with_attributes([KeyValue::new("scope_key", "scope_value")])
```

opentelemetry-stdout/src/logs/exporter.rs
utpilla: Did you miss this one?

opentelemetry-stdout/src/trace/exporter.rs
```diff
@@ -22,106 +20,129 @@ impl fmt::Debug for SpanExporter {
...
-        let res = if let Some(writer) = &mut self.writer {
-            (self.encoder)(writer, crate::trace::SpanData::new(batch, &self.resource)).and_then(
-                |_| {
-                    writer
-                        .write_all(b"\n")
-                        .map_err(|err| TraceError::Other(Box::new(err)))
-                },
-            )
+        if self.is_shutdown.load(atomic::Ordering::SeqCst) {
+            Box::pin(futures_util::future::ready(Err(TraceError::from(
```
utpilla, line 37: Use `std::future` here as well?

```suggestion
            Box::pin(std::future::ready(Err(TraceError::from(
```

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -76,57 +76,139 @@ impl PushMetricsExporter for MetricsExporter {
...
+            });
+
+        metric.metrics.iter().enumerate().for_each(|(i, metric)| {
+            println!("Metric #{}", i);
+            println!("\t\tName         : {}", &metric.name);
+            println!("\t\tDescription  : {}", &metric.description);
+            println!("\t\tUnit         : {}", &metric.unit);
+
+            let data = metric.data.as_any();
+            if let Some(hist) = data.downcast_ref::<data::Histogram<i64>>() {
```
utpilla, line 113: We only allow `i64` for UpDownCounter and Gauge, right? [Histograms](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/meter.rs#L376-L399) should only be tested for `u64` and `f64`? 
```suggestion
            if let Some(hist) = data.downcast_ref::<data::Histogram<u64>>() {
```

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -76,57 +76,139 @@ impl PushMetricsExporter for MetricsExporter {
...
+    println!("\t\tMonotonic    : {}", sum.is_monotonic);
+    print_data_points(&sum.data_points);
+}
+
+fn print_gauge<T: std::fmt::Debug + DataPointValue>(gauge: &data::Gauge<T>) {
+    println!("\t\tGauge DataPoints");
+    print_data_points(&gauge.data_points);
+}
+
+fn print_histogram<T: std::fmt::Debug + DataPointValue>(histogram: &data::Histogram<T>) {
```
utpilla, line 161: nit: It would be less verbose to simply import `use std::fmt::Debug;` than to always write the fully qualified name.

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -76,57 +76,139 @@ impl PushMetricsExporter for MetricsExporter {
...
+        }
+
+        println!("\t\t\tAttributes   :");
+        for kv in data_point.attributes.iter() {
+            println!("\t\t\t\t ->  {}: {}", kv.key, kv.value.as_str());
+        }
+    }
+}
+
+trait DataPointValue {}
```
utpilla, line 199: Why is this trait needed?

cijothomas: was a way to ensure <T> (metric value) can be printed for f64,i64,u64

opentelemetry-stdout/src/logs/exporter.rs
cijothomas: oops. some refactoring while conflict resolution might have messed up. will fix

opentelemetry-stdout/src/metrics/exporter.rs
utpilla: It should be able to print that without using this trait.

cijothomas: T has no bounds specified! The <T> in the data structure exposed to exporter is totally unrelated to the <T> in hot path..I think we can fix this when we refactor exporter/collect part of metrics

utpilla: I don't follow. Does this trait have anything to do with collect?

> T has no bounds specified! 

Why do we need this additional bound? If you simply remove this trait `DataPointValue` and update the methods that you have added in the PR to only have the `Default` bound, everything should work just fine.

```diff
- fn print_sum<T: std::fmt::Debug + DataPointValue>(sum: &data::Sum<T>)
+ fn print_sum<T: std::fmt::Debug>(sum: &data::Sum<T>)

// Same thing for `print_gauge`, `print_histogram`, `print_data_points`, `print_hist_data_points`
```

If it works without this additional trait bound, what value is this new trait adding?

cijothomas: Removed that unnecessary trait. (I am not sure why I even added that earlier 🤣 )

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -76,57 +77,132 @@ impl PushMetricsExporter for MetricsExporter {
...
+            println!("\t\tUnit         : {}", &metric.unit);
+
+            let data = metric.data.as_any();
+            if let Some(hist) = data.downcast_ref::<data::Histogram<u64>>() {
+                println!("\t\tType         : Histogram");
+                print_histogram(hist);
+            } else if let Some(hist) = data.downcast_ref::<data::Histogram<f64>>() {
+                println!("\t\tType         : Histogram");
+                print_histogram(hist);
+            } else if let Some(_hist) = data.downcast_ref::<data::ExponentialHistogram<i64>>() {
```
utpilla, line 120: Similar to https://github.com/open-telemetry/opentelemetry-rust/pull/2040#discussion_r1727888812, I think we only support `u64` and `f64` for ExponentialHistograms.

utpilla (approved):

cijothomas:
Merging now, given there is general agreement for this direction. Happy to address more feedback in follow ups.

aochagavia:
Hello there! Is there any suggested crate for people like me who were relying on the older behavior against all warnings? I was using `opentelemetry-stdout` to store a program run's traces as a JSON file, then loading it into Jaeger for inspection. I understand that you can use an exporter that talks to Jaeger directly, but I'm using this for debugging / profiling and it's useful to have the results in a JSON file.

cijothomas:
> Hello there! Is there any suggested crate for people like me who were relying on the older behavior against all warnings? I was using `opentelemetry-stdout` to store a program run's traces as a JSON file, then loading it into Jaeger for inspection. I understand that you can use an exporter that talks to Jaeger directly, but I'm using this for debugging / profiling and it's useful to have the results in a JSON file.

Based on https://github.com/open-telemetry/opentelemetry-specification/pull/4183 we can add a new exporter that'll output OTLP as json, and this time without any "don't use for prod" warning!
No timelines yet, but we are happy to get contributions for the same.

PR #2041, opened 2024-08-21
Simplify LogExporter::Export interface

lalitb:
## Changes

1. The Exporter::export interface is modified to accept the vector of  pair/tuple of reference to LogRecord and InstrumentationLib, instead of Cow of LogData.

Before:
```rust
async fn export<'a>(&mut self, batch: Vec<Cow<'a, LogData>>) -> LogResult<()>;
```

After:
```rust
async fn export( &mut self, batch: Vec<(&LogRecord, &InstrumentationLibrary)>) -> LogResult<()>;
```
This simplifies the processing for exporters, as they don't need to check whether the object is borrowed, or owned.  Exporter needs to ensure to create the copy of LogRecord and/or InstrumentationLib whenever needed.

2. The `LogData` structure is not part of the export interface. So it has been moved from `opentelemetry_sdk::export::logs` to `opentelemetry_sdk::logs::` namespace. 


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -309,8 +312,13 @@ where
     if batch.is_empty() {
         return Ok(());
     }
+    // Convert the Vec<&LogData> to Vec<(&LogRecord, &InstrumentationLibrary)>
+    let export_batch = batch
+        .iter()
+        .map(|log_data| (&log_data.record, &log_data.instrumentation))
+        .collect();
```
cijothomas, line 319: BatchProcessor is now doing an extra cloning - 1st one occurs when it receives the `&mut LogData`, so it can save it to channel. And its drained to a Vec! later, which was passed as-is to Exporter previously. Now another Vec! needs to be created to pass on to exporter.
Something we can revisit and address in the future.

cijothomas (approved):

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -3,13 +3,15 @@ use std::sync::Arc;
...
+use opentelemetry::InstrumentationLibrary;
+use opentelemetry_sdk::export::logs::LogExporter;
+use opentelemetry_sdk::logs::LogRecord;
 
 use super::OtlpHttpClient;
 
 #[async_trait]
 impl LogExporter for OtlpHttpClient {
-    async fn export<'a>(&mut self, batch: Vec<std::borrow::Cow<'a, LogData>>) -> LogResult<()> {
+    async fn export(&mut self, batch: Vec<(&LogRecord, &InstrumentationLibrary)>) -> LogResult<()> {
```
TommyCpp, line 14: Hmm can we wrap it with a struct, otherwise it will hard to add new field/data in the future

lalitb: I think we still have flexibility to add new fields in `LogRecord` structure, as that represents the actual Log datamodel.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: This operation is only creating references and collecting them, no cloning of data is happening.

cijothomas: Isn't the collect making a new Vec! ?

lalitb: Oh yes, I was talking specifically about cloning. Need to revisit this later.

Issue #2047, opened 2024-08-23
[Bug]: Using `futures_executor::block_on` blocks forever in WASM


mendess:
### What happened?

Using this library in WASM causes the program to block forever and use 100% CPU, due to the usage of `futures_executor::block_on`

### API Version

0.24.0

### SDK Version

0.24.1

### What Exporter(s) are you seeing the problem on?

N/A

Using `futures_executor::block_on` is harmful to performance and correctness.

In single-threaded contexts, blocking in place leads to blocking the entire runtime forever, and `block_on` causes that thread to busy loop until the future resolves, using 100% of CPU.

In a multithreaded environment it is also harmful as it blocks an entire thread, and if that thread happens to have other tasks scheduled on it, this will cause those tasks to be starved of resources, worse still if the future the thread is blocking on is waiting for a task that's scheduled on that same thread via a mechanism such as the tokio LIFO slot, leading to a deadlock.

See this [draft PR](https://github.com/open-telemetry/opentelemetry-rust/pull/2048) as a first attempt to address this problem.

In order to make this work for WASM I introduced a `spawn_future` function, which does the right thing in WASM but continues using the bad `futures_executor::block_on` function in other environments. With this patch I stopped having problems in the WASM project I'm working on, thus proving that this was the root of the issue.

Next I made some more methods `async`. The trait methods I made async are:
  - `LogProcessor::emit`
  - `LogProcessor::force_flush`
  - `LogProcessor::shutdown`
  - `SpanProcessor::on_end`
  - `SpanProcessor::force_flush`
  - `SpanProcessor::shutdown`

This avoids having to call `spawn_future` inside these functions. I couldn't however avoid calling this inside the `Drop` implementations that needed to call `SpanProcessor::shutdown`, as well as, inside `periodic_reader.rs` because I was unsure as to how best to proceed. So there are still some open questions.

cijothomas:
Thanks for opening the issue with detailed analysis and the PR! I left a comment on the PR https://github.com/open-telemetry/opentelemetry-rust/pull/2048/files#r1729165305 . 

We need to see what is the right way to execute all the backgrounds tasks (batch exporting, metric reader). @ThomsonTan was also looking into this. Tagging @TommyCpp and @lalitb also to share ideas/concerns.

cijothomas:
[...]

Issue #2031, opened 2024-08-16
Should Exporter force async?


cijothomas:
Exporter traits are [now written as async](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/export/logs/mod.rs#L15-L18). This works great if the exporter actually needs to use async. But when exporter is not needing async (like when exporting to kernel tracing systems like Windows ETW, Linux user_events), the async is causing unnecessary overhead, as shown in the [benchmarks](https://github.com/open-telemetry/opentelemetry-rust/pull/2027/files).

This issue is to track addressing this. One way could be offer [multiple methods](https://github.com/open-telemetry/opentelemetry-rust/pull/2027#issuecomment-2290117486). Another would be model etw/user-events exporter as a processor itself,

demurgos:
I'm coming from #2071. My opinion is that `block_on` should be avoided at all cost because of bad interactions with single-threaded runtimes and of the hangs that it causes.

If performance is a concern then multiple methods are a good approach I think.

demurgos:
> the async is causing unnecessary overhead, as shown in the [benchmarks](https://github.com/open-telemetry/opentelemetry-rust/pull/2027/files).

I ran the benchmarks myself to compare, and I agree that there's an overhead when using the `async_trait` crate. Here are my results:
```
LogExporterWithFuture
                        time:   [59.786 ns 59.883 ns 59.987 ns]
LogExporterWithoutFuture
                        time:   [50.156 ns 50.266 ns 50.392 ns]
```

However I also added an impl using native async trait support now that Rust supports it for a while:

```
LogExporterWithNativeFuture
                         time:   [51.791 ns 51.936 ns 52.102 ns
```

This suggests that there is no real overhead when using native async traits. This means that there may be no need to require multiple methods if async is good enough.

cijothomas:
@demurgos can you send a PR to modify the benchmark to use the native async trait part?

demurgos:
@cijothomas Here you go: https://github.com/open-telemetry/opentelemetry-rust/pull/2143

cijothomas:
Closing as this is taken care for logs by moving to native async, that has minimal overhead. However, etw/user-event exporter may still be modeled as processor to avoid even this minimal overhead.

PR #2048, opened 2024-08-23
[WIP] Remove usage of `futures_executor::block_on` in WASM targets

mendess:
See issue #2047 for discussion about this PR

## Changes

This PR contains 5 changes:
 - first it makes the `LogExporter::export` function return a future that doesn't capture self's lifetime, this is similar to what `SpanExporter` does. This will be important for a follow up commit
 - It adds a runtime agnostic way of spawning futures such that in WASM it uses the JS scheduler to poll futures instead of blocking the main thread
 - Makes the `LogProcessor` async, this is where the first commit is relevant. As it allows me to get around std MutexGuard not being Sync
 - Makes the same change to SpanProcessor, in order to further remove the calls to `block_on`
 - Finally replaces the final two places in the sdk where `block_on` is used explicitly, this however should instead be done by making these two changed methods async, but I was unsure how to proceed with the fallout of that change on the codebase, so settled for this.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -45,6 +45,7 @@ const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE_DEFAULT: usize = 512;
 /// The interface for plugging into a [`Logger`].
 ///
 /// [`Logger`]: crate::logs::Logger
+#[async_trait::async_trait]
 pub trait LogProcessor: Send + Sync + Debug {
     /// Called when a log record is ready to processed and exported.
     ///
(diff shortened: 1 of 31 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -57,13 +58,13 @@ pub trait LogProcessor: Send + Sync + Debug {
     /// # Parameters
     /// - `record`: A mutable reference to `LogData` representing the log record.
     /// - `instrumentation`: The instrumentation library associated with the log record.
-    fn emit(&self, data: &mut LogRecord, instrumentation: &InstrumentationLibrary);
+    async fn emit(&self, data: &mut LogRecord, instrumentation: &InstrumentationLibrary);
```
cijothomas, line 61: There is already a concern about forcing `async`  everywhere, even in places where it won't do any good, but actually lowers performance. https://github.com/open-telemetry/opentelemetry-rust/issues/2031

Agree that the blocking is an issue that needs to be solved, but I am not convinced that this change is an acceptable solution :(

Noah-Kennedy: Blocking the runtime in this method is very dicey and will at best have significant negative performance implications and at worst lock things up entirely, and should only be done if no other solutions are present.

Given that this has been observed causing production issues for folks, it's probably best to just take the approach that is correct and optimize it (if benchmarks even suggest that there is a problem) rather than do this trick.

Noah-Kennedy: Besides, if AsyncDrop does land, it's very easy to just undo this change and make the drop async

cijothomas: I am not sure I follow. What is the negative performance implication being referred to here? To be clear, `BatchLogRecordProcessor` is the one that requires a runtime and benefit from async, but that is *not* the only way to export logs. We have exporters to export to Windows ETW, Linux user-events, where exporting is done synchronously, inline within the logging call itself, and has no need for async and runtimes.

lalitb: Agree with @cijothomas. Async runtime was an optional dependency till now - only required for batch*processor (in logs and traces) and periodic-reader (in metrics). And there was plan to have sync and async version of LogExporter::export() method to make it more complete. However, with this change, the runtime would be always required, even for custom synchronous processors.

cijothomas:
Closing inactive PR. The linked issue is still open, and we can use that for further discussions, and PR can be resurrected when ready to work on it.

PR #2050, opened 2024-08-23
Use event.name as attribute key for exporting Log event name

cijothomas:
This is still being discussed in Event SIG and semantic conventions, but I propose to name it "event.name" before the upcoming release. There is high chance that "event.name" will be the final name, so we can avoid yet another breaking change after the coming release. 
This is still under feature-flag.

(Similar change for stdout exporter is needed, but that'll be done the current refactoring of stdout exporter to avoid conflicts)

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -3,10 +3,11 @@
 ## vNext
 
 - **Breaking**
-[1994](https://github.com/open-telemetry/opentelemetry-rust/pull/1994) The
-logrecord event-name is added as an attribute only if the feature flag
+The logrecord event-name is added as an attribute only if the feature flag
 `populate-logs-event-name` is enabled. The name of the attribute is changed from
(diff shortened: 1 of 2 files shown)


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 
 ## vNext
 
-- **Breaking** [1994](https://github.com/open-telemetry/opentelemetry-rust/pull/1994) The logrecord event-name is added as attribute with
-key `name` only if the feature flag `populate-logs-event-name` is enabled.
+- **Breaking**
+[1994](https://github.com/open-telemetry/opentelemetry-rust/pull/1994) The
```
utpilla, line 6: It would be nice to add the link to this PR as well as this is the PR which uses `event.name` as the attribute name.

utpilla (approved):

lalitb (approved):

PR #2052, opened 2024-08-25
fix broken docs.rs build

tsfotis:
docs.rs build fails [here](https://docs.rs/crate/opentelemetry-stdout/0.5.0/builds/1292782) with error `error[E0658]: #[doc(cfg)] is experimental...`.
Documentation build also fails on crates that depend on opentelemetry-stdout and build docs with `--cfg docsrs` and without `--no-deps`.

A simple `#![cfg_attr(docsrs, feature(doc_cfg))]` would have been enough but from a quick search I found [this](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-rust+doc_cfg&type=code) convention to be used across this repo.

## Changes

- enable doc_cfg feature on docsrs
- enable docsrs and all features build on docs.rs


lalitb:
Thanks for the fix. Wondering if we just remove the `cfg` attribute, as all three features - `logs`, `traces` and `metrics` are now enabled by default?

tsfotis:
Then it becomes a question of how we want the docs to look on docs.rs.
On one hand we remove some maybe unwanted _clutter_, but on the other we might improve usability for users who opt-in to `default-features = false` or even show users that they can opt-out on something they do not use.
I am 50/50 on this, but is something that should be the same across all creates in this repo to help with readability.

I made this change in the last commit, but I kept the `#![deny(rustdoc::broken_intra_doc_links)]` which is a good default :).

lalitb:
@tsfotis Sorry, I was not clear enough. I meant removing the error'ing code - 

https://github.com/open-telemetry/opentelemetry-rust/blob/a2c8c7dea196c80183d3be6ce6fa80b3fb939be9/opentelemetry-stdout/src/lib.rs#L21C1-L22C1.

tsfotis:
Aaa ok.

> Thanks for the fix. Wondering if we just remove the `cfg` attribute, as all three features - `logs`, `traces` and `metrics` are now enabled by default?

If we just remove this the build will still fail, maybe you mean also remove this?

lalitb:
@tsfotis Had a look into the failure logs again, removing `cfg` attribute is not the right approach, infact your first approach was indeed correct: 

https://github.com/open-telemetry/opentelemetry-rust/pull/2052/commits/57573e5203abc6cb8566c655aa8dda4da967dff4

If you can revert to first commit, we can get it through quickly. Really sorry about all the confusion :(

tsfotis:
@TommyCpp I updated the proto generated files and pushed the changes.

Though there is also the `opentelemetry-otlp/tests/integration_test/expected/serialized_traces.json` which gets updated, should I include this as well?

lalitb (approved):

lalitb:
> Though there is also the opentelemetry-otlp/tests/integration_test/expected/serialized_traces.json which gets updated, should I include this as well?

I believe shouldn't be required, as the integration tests are successful in this PR. Thanks for fixing the proto files :)

PR #2054, opened 2024-08-26
fix: AtomicI64, AtomicU64 for build mipsel processor

TGLuong:
## Changes

Use `atomic_shim` for mipsel processor. Because when build for mipsel architecture, it don't have `AtomicI64`, `AtomicU64` in std.

### DIFF
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -1,5 +1,4 @@
-use opentelemetry::global;
-use opentelemetry::KeyValue;
+use opentelemetry::{global, KeyValue};
 use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::{runtime, Resource};
 use std::error::Error;


lalitb:
Thanks for the PR. Couple of comments for now:
1. any reason for using [atomic_shim](https://crates.io/crates/atomic-shim) instead of [portable-atomic](https://crates.io/crates/portable-atomic) ?  Haven't used either of them, but from the stats (number of downloads and dependents), the later seems to be more widely used.
2. Can we include this crate only when compiling for MIPS (and possibly PowerPC), as done here - https://github.com/esp-rs/rust/blob/1b259e0b8b2129dd3d62819a32f6d9b287e6b86c/compiler/rustc_data_structures/Cargo.toml#L50

TGLuong:
> Thanks for the PR. Couple of comments for now:
> 
> 1. any reason for using [atomic_shim](https://crates.io/crates/atomic-shim) instead of [portable-atomic](https://crates.io/crates/portable-atomic) ?  Haven't used either of them, but from the stats (number of downloads and dependents), the later seems to be more widely used.
> 2. Can we include this crate only when compiling for MIPS (and possibly PowerPC), as done here - https://github.com/esp-rs/rust/blob/1b259e0b8b2129dd3d62819a32f6d9b287e6b86c/compiler/rustc_data_structures/Cargo.toml#L50

Hi @lalitb, thanks for your comments:
1. I search for Atomic library and only seen [atomic_shim](https://crates.io/crates/atomic-shim). I just have check [portable-atomic](https://crates.io/crates/portable-atomic) and it seem to be better.
2. Let me change to use this crate only for MIPS and PowerPC

PR #2055, opened 2024-08-26
Add Feature Flag to Disable Array Preallocation for Log Record Attributes in Limited Stack Environments

### LINKED ISSUE #1920: LogRecord Attributes - avoid heap allocations

lalitb:
Fixes #1920 

## Changes
This PR introduces a feature flag, `no-preallocate-attributes-array`, that disables the default behavior of preallocating a fixed-size array for storing log record attributes. Instead of using a preallocated array, this feature flag enables the use of the vector part of `GrowableArray`, which allocates memory dynamically when the first entry is added.

Rationale: Kernel mode and other environments with limited stack space require careful management of stack allocations. By using a dynamically growing vector instead of preallocating a fixed-size array, we can reduce the stack footprint.

Note - We could have directly used the `Vector` instead of `GrowableArray` for this scenario, however didn't find any visible difference in stress tests with either approach, so kept `GrowableArray` to be consistent.

### DIFF
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -51,6 +51,7 @@ testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-async-std", 
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
+no-preallocate-attributes-array = []
 
 [[bench]]
 name = "context"
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -51,7 +51,7 @@ testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-async-std", 
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
-no-preallocate-attributes-array = []
+disable_stack_log_attributes = []
 
 [[bench]]
 name = "context"
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -38,6 +38,12 @@
     ```
     This change simplifies the processing required by exporters. Exporters no longer need to determine if the LogData is borrowed or owned, as they now work directly with references. As a result, exporters must explicitly create a copy of LogRecord and/or InstrumentationLibrary when needed, as the new interface only provides references to these structures.
 
+- [#2055](https://github.com/open-telemetry/opentelemetry-rust/pull/2055)
+  New Feature Flag: `disable_stack_log_attributes`
+  * Default Behavior: By default, the `LogRecord::attributes` field uses a combination of stack and heap allocation for storing attributes. This approach is efficient for handling a limited number of attributes with minimal heap allocation.
+
+  * Behavior with `disable_stack_log_attributes` Enabled: When the `disable_stack_log_attributes` feature flag is enabled, the LogRecord::attributes field only uses vector-based heap allocation, bypassing stack allocation entirely. This setting is beneficial for applications that run in constrained environment with limited stack, and thus avoiding potential stack overflow.
+
 ## v0.24.1
 
 - Add hidden method to support tracing-opentelemetry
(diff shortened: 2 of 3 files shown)


opentelemetry-sdk/Cargo.toml
```diff
@@ -51,6 +51,7 @@ testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-async-std",
 rt-tokio = ["tokio", "tokio-stream"]
 rt-tokio-current-thread = ["tokio", "tokio-stream"]
 rt-async-std = ["async-std"]
+no-preallocate-attributes-array = []
```
cijothomas, line 54: Alternate suggestion:
`disable_stack_log_attributes`

"Preallocate" is potentially misleading, we can pre-allocate in heap as well.

cijothomas: please also add a changelog entry for this. There is a general lacking of documentaion for each of the feature flags, but that is something we need to work on independent of this PR.

utpilla:
> Rationale: Kernel mode and other environments with limited stack space require careful management of stack allocations. By using a dynamically growing vector instead of preallocating a fixed-size array, we can reduce the stack footprint.

Do we have users who have tried to run OpenTelemetry Rust in Kernel mode? I'm asking mainly to understand if the lack of this feature is an actual blocker for anyone.

lalitb:
> Do we have users who have tried to run OpenTelemetry Rust in Kernel mode? 

We don't yet support kernel mode, because of usage of `std` constructs. That needs to be addressed separately. Irrespective of that, kernel stack size is limited (8kb in x32 and 16 kb in x64) for Linux kernel, so this feature-flag would be relevant once the kernel support is there.

cijothomas:
> > Do we have users who have tried to run OpenTelemetry Rust in Kernel mode?
> 
> We don't yet support kernel mode, because of usage of `std` constructs. That needs to be addressed separately. Irrespective of that, kernel stack size is limited (8kb in x32 and 16 kb in x64) for Linux kernel, so this feature-flag would become relevant once the kernel support is there.

Is it better to delay adding this feature flag until we reach that state? Just to keep overall feature-flags under control.
We can have a tracking issue to gauge the overall interest in the feature (there could be other scenarios beyond kernel mode code)

utpilla:
> Is it better to delay adding this feature flag until we reach that state?

I would be in favor of that.

> Irrespective of that, kernel stack size is limited (8kb in x32 and 16 kb in x64) for Linux kernel, so this feature-flag would become relevant once the kernel support is there.

It'd be good to try it out first and check if we run into issues. (once we start supporting this scenario).

We allocate 10 instances of `Option<(Key, AnyValue)>` on stack. The size of `Option<(Key, AnyValue)>` should be around 48 bytes. 480 bytes of stack allocation for a stack size of 8kb/16kb is probably okay?

lalitb:
> Is it better to delay adding this feature flag until we reach that state?

sure, I actually prefer that. The feature flag requirement was added in #1920, so create the PR :)

> We allocate 10 instances of Option<(Key, AnyValue)> on stack. The size of Option<(Key, AnyValue)> should be around 48 bytes. 480 bytes of stack allocation for a stack size of 8kb/16kb is probably okay?

I think tracing needs to be conservative for any stack allocation in kernel mode, specifically these pre-allocations in stack. Thinking about the deep call chains, and recursion in actual kernel mode application, the stack size can grow fast. But, there would be other places we need to look too, and good to revisit once we think about kernel-mode :)

lalitb:
closing as discussed above.

PR #2057, opened 2024-08-27
Introduce LogBatch Struct to Replace Vec in LogExporter Interface

### LINKED ISSUE #1919: LogExporter need vec of LogRecord?
`LogProcessor` is forced to construct a Vec to pass the `LogRecord` to the Exporter (because Exporter's exporter method expects Vec), thereby requiring heap allocation.
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/log_processor.rs#L104

For sync exporters like etw, the heap allocation is not required, but SDK currently forces it.

Opening an issue to track addressing this. Options include having the exporter take a slice or some iterator so internal implementation details is not required to be exposed, but also can avoid the heap allocation.

lalitb:
Fixes #1919 

### Existing interface:

```rust
    /// Exports a batch of [LogRecord, InstrumentationLibrary].
    async fn export(&mut self, batch: Vec<(&LogRecord, &InstrumentationLibrary)>) -> LogResult<()>;
}
```
The existing interface uses a Vec to pass batches of log records and instrumentation libraries. This approach can lead to unnecessary heap allocations and does not provide the flexibility to change the backend data source.

Three approaches were considered to replace the use of Vec with more flexible abstractions:

### Approach 1: Using Iterator as below:
```rust
async fn export<'a, I>(&mut self, batch: I) -> LogResult<()>
where
    I: IntoIterator<Item = (&'a LogRecord, &'a InstrumentationLibrary)> + Send + 'a;
```

While this approach allows for flexibility in the types of data structures that can be passed, adding a generic parameter `I` makes the LogExporter trait non-object-safe. In Rust, non-object-safe traits cannot be turned into trait objects, which complicates regular usage patterns of the LogExporter trait, such as storing different exporters in a collection of Box<dyn LogExporter>.

### Approach 2: Using slice:
```rust
   async fn export(&mut self, batch: &[(&LogRecord, &InstrumentationLibrary)]) -> LogResult<()>;
```

This approach maintains object safety by avoiding generics in the trait method signature. Also provides  efficient way to handle batches of log records without unnecessary heap allocations.

### Chosen Solution: Using `LogBatch`
The third approach introduces a custom struct, `LogBatch`, to encapsulate the slice of log records and instrumentation libraries. Here's how it works:
```rust
pub struct LogBatch<'a> {
    data: &'a [(&'a LogRecord, &'a InstrumentationLibrary)],
}

impl<'a> LogBatch<'a> {
    pub(crate) fn new(
        data: &'a [(&'a LogRecord, &'a InstrumentationLibrary)],
    ) -> LogBatch<'a> {
        LogBatch { data }
    }
}

impl LogBatch<'_> {
    pub fn iter(&self) -> impl Iterator<Item = (&LogRecord, &InstrumentationLibrary)> {
        self.data.iter().map(|(record, library)| (*record, *library))
    }
}
```

Rational behind selection
 - Maintains Object Safety for `LogExporter` trait  by avoiding generics.
 - This solution avoids unnecessary heap allocations. LogBatch directly wraps a slice, ensuring that log records and instrumentation libraries are handled efficiently.
 - Comparing with option 2- Allows us to change the internal container for storage of LogRecord to something non-contiguous which can't be directly converted to slice.
 - 


cijothomas:
@lalitb Any measurable perf difference? (Irrespective of that, we still want to get rid of Vec!)

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -14,7 +14,7 @@ use std::fmt::Debug;
 #[async_trait]
 pub trait LogExporter: Send + Sync + Debug {
     /// Exports a batch of [`LogRecord`, `InstrumentationLibrary`].
-    async fn export(&mut self, batch: Vec<(&LogRecord, &InstrumentationLibrary)>) -> LogResult<()>;
```
cijothomas, line 17: please make a changelog entry

TommyCpp:
> generated file has changed but it's a CI environment, please rerun this test locally and commit the changes

Means we need to regenerate the proto files and commit. Maybe we changed the `opentelemetry-proto` version?

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -34,7 +34,7 @@ struct NoopExporter {
 
 #[async_trait]
 impl LogExporter for NoopExporter {
-    async fn export(&mut self, _: Vec<(&LogRecord, &InstrumentationLibrary)>) -> LogResult<()> {
+    async fn export(&mut self, _: &[(&LogRecord, &InstrumentationLibrary)]) -> LogResult<()> {
```
TommyCpp, line 37: Is it possible spec adding more fields than `LogRecord` and `InstrumentationLibrary`? Having a tuple here means we cannot add another field without a breaking change

lalitb:
> Means we need to regenerate the proto files and commit. Maybe we changed the opentelemetry-proto version?

tonic version has got bumped up. Added more details in #2059. We may have to regenerate proto files if there is no fix from tonic.

opentelemetry-appender-tracing/benches/logs.rs
lalitb: The specs for logs (including data model) is already stable. And any additive change (if done) would be in Log Data model, which is covered by LogRecord struct.

TommyCpp:
Yeah let's just regenerate

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -10,7 +10,7 @@
     | noop_layer_disabled         | 12 ns       |
     | noop_layer_enabled          | 25 ns       |
     | ot_layer_disabled           | 19 ns       |
-    | ot_layer_enabled            | 203 ns      |
+    | ot_layer_enabled            | 196 ns      |
```
lalitb, line 13: not related to this PR (as the processor is noop). This was consistent, so updated here.

lalitb:
Moving to draft. There seems to be some issue in batch processor handling. Will make it ready after fixing that.

opentelemetry-appender-tracing/benches/logs.rs
lalitb: reverted, so no longer relevant. Pls ignore.

lalitb:
> @lalitb Any measurable perf difference? (Irrespective of that, we still want to get rid of Vec!)

there should be definitely specifically for concurrent processor/exporter flow, but we don't have benchmark for that. The stress test uses noop processor, so it won't measure the improvement.

opentelemetry-sdk/benches/log_exporter.rs
```diff
@@ -6,8 +6,8 @@
     RAM: 64.0 GB
     | Test                           | Average time|
     |--------------------------------|-------------|
-    | LogExporterWithFuture          | 280 ns      |
-    | LogExporterWithoutFuture       | 255 ns      |
+    | LogExporterWithFuture          | 108 ns      |
+    | LogExporterWithoutFuture       | 88 ns      |
```
lalitb, line 10: This improvement is on top of #2061, so #2061 needs to be merged first.

lalitb:
> Moving to draft. There seems to be some issue in batch processor handling. Will make it ready after fixing that.

This is fixed. Making it ready for review.

opentelemetry-sdk/benches/log_exporter.rs
```diff
@@ -6,8 +6,8 @@
     RAM: 64.0 GB
     | Test                           | Average time|
     |--------------------------------|-------------|
-    | LogExporterWithFuture          | 122 ns      |
-    | LogExporterWithoutFuture       | 89 ns      |
+    | LogExporterWithFuture          | 108 ns      |
+    | LogExporterWithoutFuture       | 88 ns      |
```
cijothomas, line 10: interesting! The one with future saw a decent gain by avoiding vec!, but the other has negligible gains!

cijothomas (approved):

opentelemetry-sdk/benches/log_exporter.rs
utpilla: I think the compiler might be optimizing away the creation of `vec` in the non-futures case, since it knows the `NoopExporter` doesn't do anything with it anyway. It probably cannot do that optimization for the one with Future.

@lalitb I think we would see the perf gain for the without future case as well, if you `black_box` the argument to export method.


```diff
fn emit(&self, record: &mut LogRecord, library: &InstrumentationLibrary) {
        let mut exporter = self.exporter.lock().expect("lock error");
-       futures_executor::block_on(exporter.export(vec![(record, library)]));
+       futures_executor::block_on(exporter.export(black_box(vec![(record, library)])));
    }
    
fn emit(&self, record: &mut LogRecord, library: &InstrumentationLibrary) {
        self.exporter
            .lock()
            .expect("lock error")
-           .export(vec![(record, library)]));
+           .export(black_box(vec![(record, library)]));
    }

Comment (cijothomas) [reply opentelemetry-sdk/benches/log_exporter.rs:10]: @lalitb would you like to update benchmarks? If not, we can merge now, and tweak benchmark in future.

Comment (lalitb) [reply opentelemetry-sdk/benches/log_exporter.rs:10]: Yes will update it later, lets merge it now.

**Comment by utpilla** [comment]
> ### Approach 1: Using Iterator as below:
> ```rust
> async fn export<'a, I>(&mut self, batch: I) -> LogResult<()>
> where
>     I: IntoIterator<Item = (&'a LogRecord, &'a InstrumentationLibrary)> + Send + 'a;
> ```
> 
> While this approach allows for flexibility in the types of data structures that can be passed, adding a generic parameter `I` makes the LogExporter trait non-object-safe. In Rust, non-object-safe traits cannot be turned into trait objects, which complicates regular usage patterns of the LogExporter trait, such as storing different exporters in a collection of Box.

@lalitb Could we try this with a concrete implementation instead of generics so that `LogExporter` is still object-safe? Something like this:

```rust
async fn export(&mut self, batch: LogBatchIterator<'_>) -> LogResult<()>;

pub(crate) struct LogBatchIterator<'a> {
    data: &'a [(&'a LogRecord, &'a InstrumentationLibrary)],
}

impl<'a> LogBatchIterator<'a> {
    pub(crate) fn new(
        data: &'a [(&'a LogRecord, &'a InstrumentationLibrary)],
    ) -> LogBatchIterator<'a> {
        LogBatchIterator {
            data,
        }
    }
}

impl LogBatchIterator<'_> {
    pub fn iter(&self) -> impl Iterator<Item = (&LogRecord, &InstrumentationLibrary)> {
        self.data.iter().map(|(record, library)| (*record, *library))
    }
}
```

The LogProcessors would create `LogBatchIterator` struct and pass that to the exporters. The exporters would then call `LogBatchIterator.iter()` to iterate over `(&LogRecord, `&InstrumentationLibrary)`.

lalitb:
> The LogProcessors would create LogBatchIterator struct and pass that to the exporters. The exporters would then call LogBatchIterator.iter() to iterate over (&LogRecord, &InstrumentationLibrary)`.

thanks @utpilla . I thought about the custom iterator, but felt that the cost of creating and maintaining the iterator would add to the overhead, so didn’t proceed further. The slice is zero cost abstraction, and also no cost associated with converting vector to slice.

utpilla:
> thanks @utpilla . I thought about the custom iterator, but felt that the cost of creating and maintaining the iterator would add to the overhead, so didn’t proceed further. Specifically for concurrent scenario where we only export one event per batch. The slice is zero cost abstraction, and also no cost associated with converting vector to slice.

The advantage of using a custom iterator is that it allows us to hide the implementation details from the public API and thus offers the flexibility to change the underlying implementation later on if needed. A simple custom iterator shouldn't add any significant overhead (would need to measure it though)

Using a slice instead of vec is definitely way better but we would still be tying the type of the argument to a slice. Maybe for the `export` method, we would never need to change this from slice to something else. But in general, abstracting away the type from slice to a custom iterator should allow us to easily change the underlying implementation from `vec` to slice to a `GrowableArray` or anything else.

opentelemetry-sdk/benches/log_exporter.rs
lalitb: I don't see any difference after adding the black_box. Probably, the compiler is not optimizing it out.

lalitb:
> A simple custom iterator shouldn't add any significant overhead (would need to measure it to confirm)

Yes, the overhead would be minimal, approximately 2-3 nanoseconds per event export for the concurrent/simple exporter, as tested with log_exporter benchmark from changes [here](https://github.com/open-telemetry/opentelemetry-rust/compare/main...lalitb:opentelemetry-rust:exporter-vector-custom-iterator). It introduces an additional level of indirection to access log-record, which we tried to eliminate by removing `LogData`. While the slice can abstract over arrays and vectors. If we decide on GrowableArray - we can add a `inline_slice()` and `oveflow_slice()` methods and invoke export on each. 
I am still inclined with using slice, but can move to custom iterator if we all decide :)

opentelemetry-sdk/benches/log_exporter.rs
utpilla: Did you add the `black_box` to the main branch or this PR? It won't help with this PR as you have already removed the `vec!` creation. If you add `black_box` to the main branch, then you should see a higher baseline for the non-future case and then we would see the perf improvement from this PR.

TommyCpp:
I like the custom iterator idea, mostly because we can make changes to `LogBatch` later. Though as pointed out earlier it's unlikely to change?

lalitb:
> I like the custom iterator idea, mostly because we can make changes to LogBatch later. Though as pointed out earlier it's unlikely to change?

Yes, the log datamodel is stable, so there won't be changes in the event-data passed to the exporter. And the additive changes can be still added to `LogRecord`. The only scenario would be if we change the internal storage of LogRecords (currently Vector for batch-processor) to something highly non-contiguous which can't be converted to slice. Though that is unlikely, but we can go forward with the `LogBatch` abstraction to cover that. I believe it should be fine taking the cost of 2-3 ns for that :)

opentelemetry-sdk/benches/log_exporter.rs
lalitb: Thanks @utpilla , yes adding black_box to `main` gives result:
LogExporterWithFuture: 121
LogExporterWithoutFuture: 100

And with PR (LogBatch changes):
LogExporterWithFuture: 112
LogExporterWithoutFuture: 92

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -10,11 +10,81 @@ use opentelemetry::{
...
+///
+/// The `LogBatch` struct holds a collection of log records along with their associated
+/// instrumentation libraries. This structure is used to group log records together for efficient
+/// export operations.
+///
+/// # Type Parameters
+/// - `'a`: The lifetime of the references to the log records and instrumentation libraries.
+///
+#[derive(Debug)]
+pub struct LogBatch<'a> {
```
utpilla, line 23: Could we rename this to `LogsBatch` (plural logs) as it may contain multiple entries?

utpilla: I didn't find a consistent usage of singular vs plural (log vs logs) even in the spec. Exporters use singular form of the word such as LogExporter but when talking about the SDK it's singular again: Logs SDK, Metrics SDK etc. 😄

lalitb: Actually in this case, `LogBatch` makes sense, as `batch` already indicate multiple logs. So this name implies "single batch that contains multiple log entries."

utpilla: Hmm. I'm confused. 😄

The general usage of the word batch still is accompanied with a plural noun. E.g. "batch of students" and not "batch of student". So, I thought it might make sense to say `LogsBatch`.

Anyway, I think what you're saying also makes sense, so I'll go with `LogBatch`.

utpilla (approved):
Looks good! The PR title/description needs to be updated to reflect the new changes.

Issue #1849, opened 2024-05-31
[Bug]: example code is misleading about how `with_view` works


CobaltCause:
### What happened?

The way [these variable names](https://github.com/open-telemetry/opentelemetry-rust/blob/82fa485359745e5bda9eadc9c080bb84f98a4649/examples/metrics-advanced/src/main.rs#L12-L34) are worded make it sound like that's the only effect each view has and neglects to mention that the implementation of these closures also drops all pieces of information about each instrument that aren't specified again via `Stream` combinators. To demonstrate:

```console
$ cargo run -qp metrics-advanced | jq '.resourceMetrics.scopeMetrics[].metrics[].name'
"my_histogram_renamed"
""
""
```

It's quite surprising that the names are dropped despite no indication that this would happen.

Here's a small program that demonstrates how this is problematic and also demonstrates how to achieve the desired (and implied) behavior:

```rust
// [dependencies]
// opentelemetry = { version = "0.23.0", features = ["metrics"] }
// opentelemetry-prometheus = "0.16.0"
// opentelemetry_sdk = "0.23.0"
// prometheus = "0.13.4"

use std::error::Error;

use opentelemetry::metrics::{MeterProvider, Unit};
use opentelemetry_prometheus::exporter;
use opentelemetry_sdk::metrics::{
    new_view, Aggregation, Instrument, SdkMeterProvider, Stream,
};
use prometheus::{Registry, TextEncoder};

fn main() {
    if let Err(e) = minimal_repro() {
        eprintln!("error: {e}");
    }
    if this_works().is_err() {
        unreachable!()
    }
}

// This does what the example implies you should do to only change the
// boundaries but leave everything else the same.
fn minimal_repro() -> Result<(), Box<dyn Error>> {
    let registry = Registry::new();
    let exporter = exporter().with_registry(registry.clone()).build()?;
    let provider = SdkMeterProvider::builder()
        .with_reader(exporter)
        .with_view(|_: &Instrument| {
            Some(Stream::new().aggregation(
                Aggregation::ExplicitBucketHistogram {
                    boundaries: vec![1., 2., 3.],
                    record_min_max: true,
                },
            ))
        })
        .build();
    let meter = provider.meter("example");
    let histogram =
        meter.u64_histogram("histogram").with_unit(Unit::new("seconds")).init();
    histogram.record(1, &[]);
    let encoded = TextEncoder::new().encode_to_string(&registry.gather())?;
    println!("{encoded}");
[...]

Issue #1810, opened 2024-05-23
OTLPExporter Pipeline issues


cijothomas:
OTLP Exporters are configured via the helper methods offered by OTLP crate, using the idea of "pipeline", (eg: `opentelemetry_otlp::new_pipeline().logging()...`). This hides a lot of LoggerProvider details from the user, and has somewhat reasonable consistency across signals. - opentelemetry_otlp::new_pipeline().logging() vs opentelemetry_otlp::new_pipeline().tracing() vs opentelemetry_otlp::new_pipeline().metrics(), but I see some issues/challenges with that approach.

1. It is inconsistent with other exporters. For example, the examples for `stdout` show a different way of configuring the LoggerProvider, via `LoggerProvider::builder().with_simple_exporter(exporter).build()`. This does not hide `LoggerProvider` concepts from user. Most users get started with OpenTelemetry using stdout exporters, but then switching to otlp is a big shift in the way things are configured.
2. Composability is harder.. The OTLP Pipeline returns a built `LoggerProvider`, so user lose the ability to customize the `LoggerProvider` themselves. For eg: if user want to add a `RedactionProcessor`, it is not possible.
3. Exporters need to be aware of more things than it need to - One possible way to solve the above problem is to expose methods in OTLPPipeleine to allow adding processors. But then the Exporter pipeline need to always mimic the functionality additionally added in the provider. Even today, OTLPExporters does have knowledge about the runtime (tokio etc.), which it should not have.
4. The `install_simple/batch` is odd naming (of course, this is subjective!). 

The below is the way using LoggerProviderBuilder, shown in all examples except OTLP!
```
fn init_logs() -> Result<opentelemetry_sdk::logs::LoggerProvider, LogError> {
    let exporter = opentelemetry_otlp::new_exporter()
        .tonic()
        .with_endpoint("http://localhost:4317")
        .build_log_exporter()?;
    let provider: LoggerProvider = LoggerProvider::builder()
        .with_batch_exporter(exporter, runtime::Tokio)
        .with_resource(RESOURCE.clone())
        .build();

    Ok(provider)
}
```

The below uses OTLP::pipeline/helps, and shown in almost all OTLP examples.
```
fn init_logs() -> Result<opentelemetry_sdk::logs::LoggerProvider, LogError> {
    opentelemetry_otlp::new_pipeline()
        .logging()
        .with_resource(RESOURCE.clone())
        .with_exporter(
[...]

Issue #1241, opened 2023-08-31
Implement instrument advisory (hint) for Metric API


cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-advisory-parameters New spec. The advisory needs to be exposed on Metric API, and then the SDK must leverage them for aggregation.

alfiejfs:
Hi, is this still planned for the beta release on August 2nd? I see it mentioned in #1719 but I don't see it in the milestone. And is this just a temporary API patch whilst views are removed, or is it intended to stay? Cheers :)

cijothomas:
> Hi, is this still planned for the beta release on August 2nd? I see it mentioned in #1719 but I don't see it in the milestone. And is this just a temporary API patch whilst views are removed, or is it intended to stay? Cheers :)

Yes. (Note we got delay by few weeks).
Will fix milestone, thanks for catching it. It'll be a permanent feature, as it is part of OTel's Stable specification.

cijothomas:
This didn't land for the current release, will be covered in the next release, ~3 weeks from now. @utpilla is working on this.

cijothomas:
Closing as this is covered by 0.26 release for histogram hints.

PR #2085, opened 2024-09-08
AggregationSelector is not needed anymore

fraillt:
Hello,

I'm not sure if this is a right way to approach this, but basically I think that `AggregationSelector` trait, and many `with_aggregation_selector` configuration methods on various public SDK types is not needed anymore, because newly introduced [Views](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view) feature fully replaces this trait and removing `AggregationSelector` only makes things simpler/more understandable.

I created this PR as a proof to show that we can simply remove it without loosing any current functionality.

The only issue is this sentence in SDK:
> If the MeterProvider has no View registered, take the Instrument and apply the default Aggregation on the basis of instrument kind according to the [MetricReader](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#metricreader) instance’s aggregation property.

If I'm not missing anything, I think it could be changed to:
> If the MeterProvider has no View registered, take the Instrument and apply the [default aggregation](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#default-aggregation).

What do you think, do you agree, or I'm missing something important here?

## Changes

* Remove `AggregationSelector` trait
* Remove `with_aggregation_selector` configuration function from many public types.

### DIFF
--- a/opentelemetry-sdk/src/metrics/reader.rs
+++ b/opentelemetry-sdk/src/metrics/reader.rs
@@ -4,7 +4,6 @@ use std::{fmt, sync::Weak};
 use opentelemetry::metrics::Result;
 
 use super::{
-    aggregation::Aggregation,
     data::{ResourceMetrics, ScopeMetrics, Temporality},
     instrument::InstrumentKind,
     pipeline::Pipeline,
@@ -24,9 +23,7 @@ use super::{
 ///
 /// Pull-based exporters will typically implement `MetricReader` themselves,
 /// since they read on demand.
-pub trait MetricReader:
-    AggregationSelector + TemporalitySelector + fmt::Debug + Send + Sync + 'static
-{
+pub trait MetricReader: TemporalitySelector + fmt::Debug + Send + Sync + 'static {
     /// Registers a [MetricReader] with a [Pipeline].
     ///
     /// The pipeline argument allows the `MetricReader` to signal the sdk to collect
@@ -93,67 +90,5 @@ impl DefaultTemporalitySelector {
 impl TemporalitySelector for DefaultTemporalitySelector {
     fn temporality(&self, _kind: InstrumentKind) -> Temporality {
         Temporality::Cumulative
-    }
-}
-
-/// An interface for selecting the aggregation and the parameters for an
-/// [InstrumentKind].
-pub trait AggregationSelector: Send + Sync {
-    /// Selects the aggregation and the parameters to use for that aggregation based on
-    /// the [InstrumentKind].
-    fn aggregation(&self, kind: InstrumentKind) -> Aggregation;
-}
-
-impl<T> AggregationSelector for T
-where
-    T: Fn(InstrumentKind) -> Aggregation + Send + Sync,
-{
-    fn aggregation(&self, kind: InstrumentKind) -> Aggregation {
-        self(kind)
-    }
-}
-
-/// The default aggregation and parameters for an instrument of [InstrumentKind].
-///
-/// This [AggregationSelector] uses the following selection mapping per [the spec]:
-///
-/// * Counter ⇨ Sum
-/// * Observable Counter ⇨ Sum
-/// * UpDownCounter ⇨ Sum
-/// * Observable UpDownCounter ⇨ Sum
-/// * Gauge ⇨ LastValue
-/// * Observable Gauge ⇨ LastValue
-/// * Histogram ⇨ ExplicitBucketHistogram
-///
-/// [the spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.19.0/specification/metrics/sdk.md#default-aggregation
-#[derive(Clone, Default, Debug)]
-pub struct DefaultAggregationSelector {
-    pub(crate) _private: (),
-}
-
-impl DefaultAggregationSelector {
-    /// Create a new default aggregation selector.
-    pub fn new() -> Self {
-        Self::default()
-    }
-}
-
-impl AggregationSelector for DefaultAggregationSelector {
-    fn aggregation(&self, kind: InstrumentKind) -> Aggregation {
-        match kind {
-            InstrumentKind::Counter
-            | InstrumentKind::UpDownCounter
-            | InstrumentKind::ObservableCounter
-            | InstrumentKind::ObservableUpDownCounter => Aggregation::Sum,
-            InstrumentKind::Gauge => Aggregation::LastValue,
-            InstrumentKind::ObservableGauge => Aggregation::LastValue,
-            InstrumentKind::Histogram => Aggregation::ExplicitBucketHistogram {
-                boundaries: vec![
-                    0.0, 5.0, 10.0, 25.0, 50.0, 75.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0,
-                    5000.0, 7500.0, 10000.0,
-                ],
-                record_min_max: true,
-            },
-        }
     }
 }
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -78,7 +78,7 @@ impl Default for HttpConfig {
 /// ```
 /// # #[cfg(feature="metrics")]
 /// use opentelemetry_sdk::metrics::reader::{
-///     DefaultAggregationSelector, DefaultTemporalitySelector,
+///     DefaultTemporalitySelector,
 /// };
 ///
 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -91,7 +91,6 @@ impl Default for HttpConfig {
 /// let metrics_exporter = opentelemetry_otlp::new_exporter()
 ///     .http()
 ///     .build_metrics_exporter(
-///         Box::new(DefaultAggregationSelector::new()),
 ///         Box::new(DefaultTemporalitySelector::new()),
 ///     )?;
 ///
@@ -252,7 +251,6 @@ impl HttpExporterBuilder {
     #[cfg(feature = "metrics")]
     pub fn build_metrics_exporter(
         mut self,
-        aggregation_selector: Box<dyn opentelemetry_sdk::metrics::reader::AggregationSelector>,
         temporality_selector: Box<dyn opentelemetry_sdk::metrics::reader::TemporalitySelector>,
     ) -> opentelemetry::metrics::Result<crate::MetricsExporter> {
         use crate::{
@@ -267,11 +265,7 @@ impl HttpExporterBuilder {
             OTEL_EXPORTER_OTLP_METRICS_HEADERS,
         )?;
 
-        Ok(crate::MetricsExporter::new(
-            client,
-            temporality_selector,
-            aggregation_selector,
-        ))
+        Ok(crate::MetricsExporter::new(client, temporality_selector))
     }
 }
 
--- a/opentelemetry-otlp/src/exporter/tonic/mod.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/mod.rs
@@ -97,7 +97,7 @@ fn resolve_compression(
 /// ```no_run
 /// # #[cfg(feature="metrics")]
 /// use opentelemetry_sdk::metrics::reader::{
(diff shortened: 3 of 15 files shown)


cijothomas:
This looks okay to me, thanks. We need have CLA agreement signed (one-time) before accepting any changes to the repo, as this is the mandatory requirement from OpenTelemetry/Linux-Foundation organization which hosts the project.

Though Views allow one to change default aggregation, I haven't observed demand for that capability yet!. That capability is *not* exposed in OTel .NET Metrics, and no one has ever asked for it so far (its been GA since late 2021). We could even that capability from OTel Rust too for first stable release to keep public API surface minimal.

fraillt:
Thanks for guidance and your input!

For my use case it was enough when `AggregatorSelector` had signature of `fn aggregator_for(&self, descriptor: &Descriptor) -> Option<Arc<dyn Aggregator + Send + Sync>>`. It wasn't super convenient to use, but having instrument `name` and `kind` was enough to set proper histogram bucket sizes.
`Views` are more powerful and I like to have a power, but in my opinion they are too powerful, especially considering that they are not requested that much :). Regarding .NET i'm a bit surprised, maybe they have non-compatible (with SDK) ways to achieve same things? (worst case you implement your own `MetricReader` :))

In my opinion, supporting 0-N number of views was a mistake, because it's not obvious how things behave without reading very carefully SDK (or even better source code), and what is worse, you'll get warnings if you use `*` to change default.
If it would be exactly 1 view, things could be much simpler. I would be happy to have a simple interface with a `fn change_view(&self, stream: StreamView) -> StreamView`.

* Default implementation would be `stream`
* Drop everything would be `stream.aggregation(Drop)`
* For more complicated scenarios `StreamView` could have some helper functions like `matches_name` that could use `*` and `?`, but for the most part you would write simple code.
```rust
if stream.matches_name("histogram*") {
  return stream.aggregation(Sum)
} 
if stream.scope.name =="mylib" {
  return stream.allowed_attributes([Key::new("size")])
} 
if stream.scope.name == "otherlib" && if stream.name == "my_metric" {
  return stream.description("even nicer description for other lib")
    .unit("X")
}
stream // leave unchanged default
// or .stream.aggregation(Drop) to drop the rest
```

Anyway, thanks again, and I'll try to get CLA signed.

cijothomas:
> Regarding .NET i'm a bit surprised, maybe they have non-compatible (with SDK) ways to achieve same things? (worst case you implement your own MetricReader :))

No there is no other way. OTel .NET simply don't allow one to change aggregation. Ofcourse it does allow one to change HistogramBuckets, and also flip between Explicit vs Exponential Histogram, but thats about it. One cannot take a Counter and ask it to be aggregated as Histogram and things like this.

cijothomas:
@fraillt Your concerns are valid. Views does have valid use cases, but it is also true that its somewhat easy to mis-use and cause issues. My original plan was to [remove Views completely from 1st stable release.](https://github.com/open-telemetry/opentelemetry-rust/issues/1719). It is entirely possible we do that, or possible that we trim it down.

The most common use case for Views is to change Histogram Buckets - @utpilla is looking at fixing https://github.com/open-telemetry/opentelemetry-rust/issues/1241 so one can change Histogram Buckets *without* having to learn about Views.

Is your scenario also just about customizing histogram buckets?

Irrespective of this, the direction of this PR is good.

fraillt:
Yes, currently I need to customize histogram buckets, but I actually I might want to use rename or aggregation drop functionality as well... :)
Don't get me wrong, I actually think that views are very good addition, and given what they provide I think they have good design. It's just it might be too powerful and that power adds a costs in terms of user experience.

Anyway, what should I do to finish my work? Is updating appropriate CHANGELOG.md files would be enough?

cijothomas:
_Anyway, what should I do to finish my work? Is updating appropriate CHANGELOG.md files would be enough?_

Please pick latest main, and see if the CI turns green. I'll get back to reviewing this after the https://github.com/open-telemetry/opentelemetry-rust/pull/2088 release is done! Thanks for waiting!

fraillt:
This test is flaky.
```
 ---- prometheus_exporter_integration stdout ----
thread 'prometheus_exporter_integration' panicked at opentelemetry-prometheus/tests/integration_test.rs:412:5:
assertion `left == right` failed: sanitized attributes to labels
```
Do you know if anyone is working on it? What should I do?

opentelemetry-prometheus/Cargo.toml
```diff
@@ -21,13 +21,13 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 once_cell = { workspace = true }
-opentelemetry = { version = "0.24", default-features = false, features = ["metrics"] }
-opentelemetry_sdk = { version = "0.24", default-features = false, features = ["metrics"] }
+opentelemetry = { default-features = false, features = ["metrics"], path = "../opentelemetry" }
```
cijothomas, line 24: prometheus was intentionally excluded to use latest sdk to keep 1.0 scope minimized. It is okay to do another release with prometheus picking up newest api/sdk, but please do that into a separate PR, to keep PRs focused on one thing at time.

cijothomas (changes requested):
The changes look good. Thank you for your contirbution!

Requesting changes just to remove the Prometheus exporter changes from this PR. Please submit that as a separate PR, and we can get it in, and do a release of Prometheus as well.

opentelemetry-sdk/src/metrics/reader.rs
```diff
@@ -136,10 +117,9 @@ impl DefaultAggregationSelector {
     pub fn new() -> Self {
         Self::default()
     }
-}
 
-impl AggregationSelector for DefaultAggregationSelector {
-    fn aggregation(&self, kind: InstrumentKind) -> Aggregation {
+    /// Default aggregation for instrument kind
+    pub fn aggregation(&self, kind: InstrumentKind) -> Aggregation {
```
utpilla, line 122: We don't want this method to be public. In fact, since this is only used in `pipeline`, we could simply remove the struct `DefaultAggregationSelector` along with its methods and move this to `pipeline.rs` as a module level function similar to `aggregate_fn` and `is_aggregator_compatible`.

cijothomas (approved):
Thanks a lot for your contributions! Would love to see more contributions.

utpilla (approved):

domenicquirl:
Hi folks, I just had a miserable time upgrading our `opentelemetry` versions because we make use of custom histogram buckets and after bumping `opentelemetry-otlp` the `AggregationSelector` was just gone. I'm still severely confused as to how I am now and will be in the future supposed to do this, so I wanted to point out a few things I ran into in hopes it helps improving them:

This change made in into the changelog for `opentelemetry-sdk` ([v0.26.0](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#v0260)), but there is nothing in the corresponding [entry](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/CHANGELOG.md#v0260) for `opentelemetry-otlp`, even though this PR removed `with_aggregation_selector` from the `OtlpMetricPipeline`. This should probably be listed as a breaking change.

The changelog entry that does exist for `opentelemetry-sdk` just says that the same functionality "should" be achievable by using views. It does not say whether this is actually possible, nor how one would do so.

Views seem to be under-documented in general. While https://github.com/open-telemetry/opentelemetry-rust/issues/1060 exists and examples were added for usage from `opentelemetry-sdk` in https://github.com/open-telemetry/opentelemetry-rust/pull/1170, I was not able to find any official documentation on how to use views at all with `opentelemetry-otlp`.  As far as I was able to find out, there is now way to actually do so directly through the `OtlpMetricsPipeline` returned from `opentelemetry_otlp::new_pipeline().metrics()`, which is how meter providers are constructed in the crate documentation. I did find https://github.com/open-telemetry/opentelemetry-rust/issues/1241, but was again left wondering how I was supposed to make use of the advisory in this context.

In the end, I think I made it work without using an `OtlpMetricsPipeline` by constructing exporter, reader and provider individually following https://github.com/open-telemetry/opentelemetry-rust/issues/1174#issuecomment-1660646295, which is also out of date already but pointed me in the right direction. 

When reading about views in https://github.com/open-telemetry/opentelemetry-rust/issues/1849, I stumbled across https://github.com/open-telemetry/opentelemetry-rust/issues/1849#issuecomment-2142966646, which seems to indicate that this method of customization will _stop working_ in upcoming releases, maybe being re-added in the future. You again suggest using https://github.com/open-telemetry/opentelemetry-rust/issues/1241 instead, but at this point I am just lost as to what plugs in where.

Apologies for highjacking this PR for my comments, but it does feel like the OTLP crate got the short end of the stick with these changes. Putting aside my confusion, imo at minimum a changelog entry should be added for the breaking change in `opentelemetry-otlp` and it would be nice if that showed or linked to how to replace an aggregation selector with a view or however else is supposed to be the "correct" way of customizing things here going forward.

fraillt:
Hi,
I'm not deeply involved in all the details, but I think that Views are the only option at the moment... 
The reason why AggregationSelector was removed by this PR, is because at some point it became basically useless. It changed signature from `Descriptor -> Aggregator` to `InstrumentKind -> Aggregation`.
IMHO, views will not be removed, (because there's no replacement for them yet), and there are at least two people (you and me) who needs this functionality. :) at the same time I agree, its not trivial to use them, (I actually found it easier to just go source code and figure out how they work...) so I think there will still be some changes.

utpilla:
@domenicquirl Apologies for the confusion! Since this was more of an API/SDK feature than OTLP exporter, we didn't update the Changelog for OTLP Exporter. However, as you pointed out, we could have done a better job at updating the SDK changelog.

The best way to do customize histogram buckets is through the `with_boundaries` Hint API. This API was added in #2135 and it is independent of the exporter being used. You can find the usage in the metrics-basic example: https://github.com/open-telemetry/opentelemetry-rust/blob/16c0e1034384f3897ccecc76b2aa1fb55f1f231a/examples/metrics-basic/src/main.rs#L86-L93

Just FYI Histogram Hint API only covers customization of buckets. Other aggregation customizations would still be covered by Views. We would be focusing on improving and fixing Views after the stable release.

domenicquirl:
Thanks for following up! Very much appreciate your acknowledgement regarding the changelog.

> The best way to do customize histogram buckets is through the `with_boundaries` Hint API. This API was added in #2135 and it is independent of the exporter being used. You can find the usage in the metrics-basic example:

@utpilla I've seen both the hint / advisory API and the example by now, but I'm actually using `tracing_opentelemetry` with `opentelemetry_otlp`, so there's two more layers between my code and anything from `opentelemetry_sdk`. I can set up a `MeterProvider` which I pass to the tracing layer, but it is the [`MetricsLayer`](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/struct.MetricsLayer.html) which is responsible for creating the histograms and which would thus need to pass boundaries for them. At the moment, it doesn't, and there isn't really a way to hook into its event processing that I know of that would allow me to do this.

It might be that this is fine and `tracing_opentelemetry` just needs time to catch up. There's an open issue on the repo (https://github.com/tokio-rs/tracing-opentelemetry/issues/108) about supporting custom histogram buckets using the advisory API, but it hasn't seen much action. 

That said, I think even for users of just `opentelemetry_otlp` without tracing it would be nice if there was a way to achieve this via the `OtlpMetricsPipeline`. It's reasonably more complicated having to drop down to the individual components and it's not just required for histograms: the pipeline currently doesn't offer any functionality for views (perhaps because of the temporary removal that was considered?).

cijothomas:
@domenicquirl Thanks for taking time reporting the issue (very well written!).

Couple of follow ups:
1. The OtlpMetricsPipeline has a questionable future, depending on https://github.com/open-telemetry/opentelemetry-rust/issues/1810. Configuring SDK (i.e MeterProvider) should be done in an exporter-neutral way, but OtlpMetricsPipeline breaks that by putting too much responsibility to an exporter.
2. Can you share your use-cases of needing customizing AggregationSelector. We are quite likely to remove View from 1.0 scope, but want to understand common use-cases for needing Views. (The most common use case I am seeing is the histogram buckets, then cardinality capping/dropping-attributes.)

PR #2089, opened 2024-09-10
Update KeyValue to avoid copying non-static string slices

utpilla:
Related to #1642

## Changes
- Update `Value` struct to avoid copying non-static string slices

### Benchmarks
- There is around a **32% improvement in perf** for the scenario where the attribute values are not static string slices
- No regression with other existing scenarios 

### DIFF
--- a/opentelemetry/benches/attributes.rs
+++ b/opentelemetry/benches/attributes.rs
@@ -10,7 +10,7 @@ RAM: 64.0 GB
 | CreateTupleKeyValue                              |      671 ps |
 | CreateOtelKeyValueArray                          |     18.4 ns |
 | CreateOtelKeyValueArrayWithMixedValueTypes       |     18.1 ns |
-| CreateOtelKeyValueArrayWithNonStaticValues       |     90.1 ns |
+| CreateOtelKeyValueArrayWithNonStaticValues       |     61.2 ns |
 | CreateTupleKeyValueArray                         |     2.73 ns |
 */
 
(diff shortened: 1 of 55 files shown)


opentelemetry/benches/attributes.rs
```diff
@@ -10,7 +10,7 @@ RAM: 64.0 GB
 | CreateTupleKeyValue                              |      671 ps |
 | CreateOtelKeyValueArray                          |     18.4 ns |
 | CreateOtelKeyValueArrayWithMixedValueTypes       |     18.1 ns |
-| CreateOtelKeyValueArrayWithNonStaticValues       |     90.1 ns |
+| CreateOtelKeyValueArrayWithNonStaticValues       |     61.2 ns |
```
utpilla, line 13: @cijothomas The changes in this PR are helpful for this scenario.

cijothomas: Can you add metric e2e benchmarks too? the counter.add and friends

cijothomas:
1. Would love to add benchmarks for Metrics to see the overall gains in metric hot path.
2. Given this is doable without disrupting users, I'd prefer to park this and get back post RC stage, so we can focus on fixing critical bugs/feature gaps now.

As usual, great work! Thanks for constantly pursuing to get the highest performance. From my knowledge of other OTel clients, none of them allow alloc-free recordings when the key/values are a slice/part of an existing string, so this is a great improvement!

cijothomas:
@utpilla can you resurrect this for 0.28 consideration?

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/issues/2449 Closing inactive PR, we have issue to track this and linked to this PR.

opentelemetry/benches/attributes.rs
utpilla: I have added the benchmarks for Counter and Histogram in #2502

PR #2095, opened 2024-09-10
Remove range requirement, patch in shell script

### LINKED ISSUE #2094: [Bug]: Support tokio >1.38
### What happened?

The opentelemetry project currently requires tokio "~1.38.0", as changed in this commit https://github.com/open-telemetry/opentelemetry-rust/commit/57bcd59989ecea1eca7b6b8c73dd87fddeb4e07c

The comment indicates that this is to not require a bump of OTel's msrv from 1.65 to 1.70.
However, this requirement prevents projects that want to use OTel's crates from using newer tokio versions, even if they have a rust version that is way newer than 1.70 (as is my case).

Perhaps the requirement could be reverted to just "1" to allow consumers to choose the tokio version, with some other workaround for CI? I am not sure what the best way forward is.

### API Version

0.24

### SDK Version

0.24

_No response_

### Relevant log output

_No response_

aumetra:
Fixes #2094

## Changes

Remove the range requirement from the tokio dependency, enabling usage of more recent versions of tokio.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ temp-env = "0.3.6"
 thiserror = { version = "1", default-features = false }
 tonic = { version = "0.12", default-features = false }
 tonic-build = "0.12"
-tokio = { version = "~1.38.0", default-features = false } #1.39 needs msrv bump to rustc 1.70
+tokio = { version = "1", default-features = false }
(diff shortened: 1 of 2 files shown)


lalitb (approved):
thanks.

PR #2096, opened 2024-09-10
[Logs] Initial draft to handle batch export into background thread

### LINKED ISSUE #2066: [Feature]: remove async runtime dependency in batch processors
### Related Problems?

As mentioned in item 7 in #1968, the current bath processes requires an async runtime (see [here](https://github.com/open-telemetry/opentelemetry-rust/blob/29fd682203cd6c677d0e9ca2e14f2634c1aeafd6/opentelemetry-sdk/src/trace/span_processor.rs#L449), there are pros and cons for this dependency than creating a dedicated background thread to export for each batch processor, while the latter approach is implemented in both [OpenTelemetry C++]() and [OpenTelemetry .NET]() SDKs.

* Pros
  * No dedicated thread requirement.
  * Reuse the existing runtime if the application already depends on it.
* Cons
  * Handle telemetry export in the application's async runtime and interfere with the business logic processing.
  * Force adding the dependence on async runtime which may not be available for some environment like IoT or others.
  * The dedicated background exporting thread number could be unbound.

Based on the current implementation in other SDKs and the pros and cons above, I suggest we switch to export in background thread instead of sending the export task to user configured async runtime.

### Considered Alternatives

_No response_

### Additional Context

_No response_

ThomsonTan:
Fixes #2066

## Changes

This is an early draft which implements the proposal mentioned in #2066, and looking for feedback.


cijothomas:
See https://github.com/open-telemetry/opentelemetry-rust/compare/main...lalitb:thread-runtime?expand=1 also to get some ideas about solving same/similar problems.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -351,20 +353,15 @@ where
...
-    }
+    let _export = exporter.export(LogBatch::new(log_vec.as_slice()));
+    // let timeout = runtime.delay(time_out);
+    pin_mut!(_export);
+    // pin_mut!(timeout);
+    // match future::select(export, export).await {
+    //     Either::Left((export_res, _)) => export_res,
+    //     Either::Right((_, _)) => ExportResult::Err(LogError::ExportTimedOut(time_out)),
+    // }
+    ExportResult::Ok(())
```
lalitb, line 364: So, we are expecting export method to complete in a definite time-period, and we are not handling timeout outside of export method ? any reason why it is so ?

lalitb:
@ThomsonTan - can we move this as separate processor (say) `BatchLogProcessorWithOwnThread` and keeping existing `BatchLogProcessor`, and then once we validate all the scenarios, we can rename it back - as done two-step process for metrics in #2292 and #2403. That way, we can merge it more quickly and keep doing incremental changes on top before being stable. The next release can bring it as separate processor, and then subsequent stable release for logs will have it as default?

lalitb:
Also, can you please fix the merge conflicts - would be good to review on that :),

lalitb:
Also @ThomsonTan - Can you test it with all OTLP scenarios - OTLP HTTP (hyper, reqwest, reqwest-blocing) and OTLP gRPC, and share the result. Thanks.

lalitb:
Also, adding this to v0.28 milestone.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: yes. Just chatted with @ThomsonTan about this. If a rogue exporter takes forever, then BatchProcessor's will get stuck behind that. For now, lets us document this limitation and move on.
(Metrics have the same problem. Its slightly worse in metrics, as observable callbacks can also go rogue and entire Metrics SDK will halt to function)

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -538,11 +585,13 @@ where
 enum BatchMessage {
     /// Export logs, usually called when the log is emitted.
     ExportLog((LogRecord, InstrumentationScope)),
+    /// ForceFlush flush the current buffer to the backend
+    ForceFlush(mpsc::SyncSender<ExportResult>),
```
cijothomas, line 589: there is another Flush in the enum below, is that meant to be removed?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -39,6 +41,14 @@ const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE: &str = "OTEL_BLRP_MAX_EXPORT_BATCH_SIZE";
 /// Default maximum batch size.
 const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE_DEFAULT: usize = 512;
 
+/// Default timeout for forceflush and shutdown.
+const OTEL_LOGS_DEFAULT_FORCEFLUSH_TIMEOUT: Duration = Duration::from_secs(1);
+const OTEL_LOGS_DEFAULT_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(1);
+
+/// environment variable name for forceflush and shutdown timeout.
+const OTEL_LOGS_FORCEFLUSH_TIMEOUT_NAME: &str = "OTEL_METRIC_EXPORT_INTERVAL";
```
cijothomas, line 49: I don't think these are meant for force_flush....

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -150,33 +160,49 @@ impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
...
+            );
+            return;
+        }
+
+        let result = self.sender.send(BatchMessage::ExportLog((
             record.clone(),
             instrumentation.clone(),
         )));
 
+        if let Err(err) = result {
```
cijothomas, line 199: please remove this logging as we are tracking drop via counts

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
             .try_send(BatchMessage::Shutdown(res_sender))
             .map_err(|err| LogError::Other(err.into()))?;
 
-        futures_executor::block_on(res_receiver)
-            .map_err(|err| LogError::Other(err.into()))
-            .and_then(std::convert::identity)
+        let (sender, receiver) = mpsc::sync_channel(1);
```
cijothomas, line 260: there is a one shot::channel, and another try_send just above this. I guess they were leftovers and should be removed.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+        receiver.recv_timeout(self.shutdown_timeout).map_err(|err| {
+            if err == RecvTimeoutError::Timeout {
+                LogError::ExportTimedOut(self.shutdown_timeout)
+            } else {
+                LogError::Other(err.into())
+            }
+        })??;
+
+        if let Some(handle) = self.handle.lock().unwrap().take() {
+            handle.join().unwrap();
```
cijothomas, line 273: emit internal log when thread is exiting?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                    Err(RecvTimeoutError::Timeout) => {
+                        export_with_timeout_sync(config.max_export_timeout, exporter.as_mut(), logs.split_off(0), &mut last_export_time);
+                    }
+                    Err(err) => global::handle_error(LogError::Other(err.into())),
                 }
             }
-        }));
+        });
+
+        let forceflush_timeout = env::var(OTEL_LOGS_FORCEFLUSH_TIMEOUT_NAME)
```
cijothomas, line 342: I don't think we need timeout for flush.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+            .ok()
+            .and_then(|v| v.parse().map(Duration::from_millis).ok())
+            .unwrap_or(OTEL_LOGS_DEFAULT_FORCEFLUSH_TIMEOUT);
+        let shutdown_timeout = env::var(OTEL_LOGS_SHUTDOWN_TIMEOUT_NAME)
```
cijothomas, line 346: agree with using timeout for shutdown, but this env variable is not required. Lets accept a timeout in the shutdown() method itself, and if none provided, use 5 sec as default. (5 sec is not from any spec, just my guess of a good default)

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
             .map_err(|err| LogError::Other(err.into()))?;
 
-        futures_executor::block_on(res_receiver)
-            .map_err(|err| LogError::Other(err.into()))
-            .and_then(std::convert::identity)
+        let (sender, receiver) = mpsc::sync_channel(1);
+        self.sender.try_send(BatchMessage::Shutdown(sender))
+                    .map_err(|err| LogError::Other(err.into()))?;
+
+        receiver.recv_timeout(self.shutdown_timeout).map_err(|err| {
```
cijothomas, line 264: nice way of enforcing shutdown timeout! Love it

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
-                            )
-                            .await;
-
-                            if let Err(err) = result {
-                                otel_error!(
-                                    name: "BatchLogProcessor.Export.Error",
-                                    error = format!("{}", err)
-                                );
-                            }
+            logs.reserve(config.max_export_batch_size);
```
cijothomas, line 289: I think Vec: has a new method that takes capacity, so you can use that instead of reserve.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+
+            loop {
+                let remaining_time_option = config.scheduled_delay.checked_sub(last_export_time.elapsed());
+                let remaining_time = match remaining_time_option {
+                    Some(remaining_time) => remaining_time,
+                    None => config.scheduled_delay,
+                };
+
+                match receiver.recv_timeout(remaining_time) {
+                    Ok(BatchMessage::ExportLog(data)) => {
```
cijothomas, line 299: trying to understand the amount of allocation/copying overall...
when emit() is called, we close the record and store it in heap, and a pointer to the heap is passed to the channel. In our thread, we receive the pointer as a message and add that to the vec, so its just the cost of copying a pointer, and not the entire LogRecord again.. is this right understanding?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -216,136 +257,143 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                        logs.push(data);
```
cijothomas, line 300: I think I mentioned this in the GitHub issue too, but something we need to further discuss. The channel has capacity of 500 (example). The vec also can hold 500. So it is possible for around 1000 item to be in memory (channel + vector) at a time?

cijothomas:
Closing in favor of #2436

PR #2097, opened 2024-09-10
Fix metrics sorted/deduped bug AND extract common functionality for better code usage/consistency.

fraillt:
Fixes #
https://github.com/open-telemetry/opentelemetry-rust/pull/2085#issuecomment-2339729168

Basically the issue was, that introduced in #1989.
It made success path super fast, by avoiding any copying/sorting/deduplicating of attributes, which is very cool, 
BUT...
in order to achieve that in attribute set tracker it was inserted two records: `original_list` AND `sorted_deduped_list`.
Later, during aggregation, there was introduced deduplication mechanism, so that only one attribute set would be selected for specific tracker.
However, since rust uses randomness in hashing, the order in which these two attribute sets (original and deduped) is returned is random.
So the end result is that everytime application is restarted, `MetricReader` will get `original_list`  OR `sorted_dedup_list` at random.

### Bug fix

* by introducing separate list for `sorted_deduped` attributes. 
This should also improve performance for collection stage as it only needs to go through this list of already `sorted_deduped` attributes.
* added `attributes_should_be_sorted_and_deduplicated` test, to make sure this wouldn't happen again.

### Other improvements

* extracted common functionality used across all (except one) aggregators into few functions and classes/struts. 
  * this reduces code duplication by 4x, and makes implementations more consistent.
  * all logic related to updating aggregates for specific attribute sets is more isolated an in one place.
* several improvements related to `AttributeSet`
  * use faster hashing function `rustc-hash`.
  * improved sorting/deduping implementation.
  * `sorted_deduped` list uses AttributeSet as key, for faster hashing/lookup.
  * moved `AttributeSet` to it's own file, in order to better protect internal implementation to guarantee sorting&deduping.

### DIFF
--- a/opentelemetry-sdk/src/metrics/attribute_set.rs
+++ b/opentelemetry-sdk/src/metrics/attribute_set.rs
@@ -1,0 +1,63 @@
+use std::hash::{Hash, Hasher};
+
+use opentelemetry::{Key, KeyValue, Value};
+use rustc_hash::FxHasher;
+
+/// A unique set of attributes that can be used as instrument identifiers.
+///
+/// This must implement [Hash], [PartialEq], and [Eq] so it may be used as
+/// HashMap keys and other de-duplication methods.
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub(crate) struct AttributeSet(Vec<KeyValue>, u64);
+
+impl From<&[KeyValue]> for AttributeSet {
+    fn from(values: &[KeyValue]) -> Self {
+        let mut vec = Vec::from_iter(values.into_iter().cloned());
+        vec.sort_by(|a, b| a.key.cmp(&b.key));
+
+        // we cannot use vec.dedup_by because it will remove last duplicate not first
+        if vec.len() > 1 {
+            let mut i = vec.len() - 1;
+            while i != 0 {
+                let is_same = unsafe { vec.get_unchecked(i - 1).key == vec.get_unchecked(i).key };
+                if is_same {
+                    vec.remove(i - 1);
+                }
+                i -= 1;
+            }
+        }
+
+        let hash = calculate_hash(&vec);
+        AttributeSet(vec, hash)
+    }
+}
+
+fn calculate_hash(values: &[KeyValue]) -> u64 {
+    let mut hasher = FxHasher::default();
+    values.iter().fold(&mut hasher, |mut hasher, item| {
+        item.hash(&mut hasher);
+        hasher
+    });
+    hasher.finish()
+}
+
+impl AttributeSet {
+    /// Iterate over key value pairs in the set
+    pub(crate) fn iter(&self) -> impl Iterator<Item = (&Key, &Value)> {
+        self.0.iter().map(|kv| (&kv.key, &kv.value))
+    }
+
+    pub(crate) fn into_inner(self) -> Vec<KeyValue> {
+        self.0
+    }
+
+    pub(crate) fn as_ref(&self) -> &Vec<KeyValue> {
+        &self.0
+    }
+}
+
+impl Hash for AttributeSet {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        state.write_u64(self.1)
+    }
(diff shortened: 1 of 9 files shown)


cijothomas:
@fraillt Good work here! Just wondering if you hit the issue in a real application OR just that the prom tests started failing due to this, that led you to dig deep into this?

fraillt:
Indeed solves same issue as #2093,
I just added few extra improvements, but probably it makes sense to borrow idea of sorting/deduping only when assigning to DataPoints.

I don't use this version in production yet, but in production we have custom MetricReader that expects sorted attributes, so I would be caching it in prod as well :)

cijothomas:
> I don't use this version in production yet, but in production we have custom MetricReader that expects sorted attributes, so I would be caching it in prod as well :)

Can you clarify if you expect **sorted** attributes or **sorted+de-duped** attributes? OTel spec ensures uniqueness (i.e no de-duplications), but not ordering.

My main question was - how did the instrumentation end up with duplicate attributes in the first place? That feels like an instrumentation bug to me to be fixed.

fraillt:
At the moment, our production setup is a bit custom...
We have custom exporter (MetricReader) that exports metrics in more concise way (compared to prometheus exporter).
So for this reason we need to output sorted+deduped attributes.
Our current implementation relies on fields to be sorted+deduped from Opentelemetry... so it's either works this way in opentelemetry, or we have some issues that we don't notice :))

cijothomas:
> At the moment, our production setup is a bit custom... We have custom exporter (MetricReader) that exports metrics in more concise way (compared to prometheus exporter). So for this reason we need to output sorted+deduped attributes. Our current implementation relies on fields to be sorted+deduped from Opentelemetry... so it's either works this way in opentelemetry, or we have some issues that we don't notice :))

Got it. You need to do the sorting yourself, as that is not something guaranteed by OTel spec. It *may* work, but that is not a guarantee, so its best if your custom MetricReader does not rely on undocumented behavior.

fraillt:
Updated revision, with few notable changes:


* grouped some functionality related to attribute set tracking in new classes: `AttributeSetTracker`, and `LockFreeNoAttribsTracker`.
* introduced `collect_data_points_reset` and `collect_data_points_readonly` reusable functions that is responsible for collecting all data points. This reduced code duplication 5x and now all optimizations will be defined in one file.
* borrowed ideas from #2093 and introduced separate `sorted_deduped` list for fast iteration at collection time.
* restored back performance, by keeping unmodified attribs set in separate list.

Overall performance has increased, at least in few places:

* in collecting phase, (as we don't need to check if attribute set was seen, and attributes are already sorted_deduped)
* also performance in crease in 
```
Counter_Overflow        time:   [165.23 ns 165.31 ns 165.38 ns]
                        change: [-74.110% -74.031% -73.967%] (p = 0.00 < 0.05)
                        Performance has improved.
```

opentelemetry-sdk/src/metrics/attribute_set.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use opentelemetry::{Key, KeyValue, Value};
+use rustc_hash::FxHasher;
+
+/// A unique set of attributes that can be used as instrument identifiers.
+///
+/// This must implement [Hash], [PartialEq], and [Eq] so it may be used as
+/// HashMap keys and other de-duplication methods.
+#[derive(Clone, Default, Debug, PartialEq, Eq)]
+pub(crate) struct AttributeSet(pub(crate) Vec<KeyValue>, u64);
```
cijothomas, line 11: its not clear what is the reason why this is being changed? can you update the PR desc. with the reasoning. 
if the goal is to get predictable attribute set in readers, then https://github.com/open-telemetry/opentelemetry-rust/pull/2093 looks sufficient to achieve that, without having to touch the hot path code.

fraillt:
I'll close this PR for few reasons:
* #2093 solves the issue with less disruptive way (less code refactorings)
* #2114 focuses on improve interface where it is most needed first.

Issue #374, opened 2020-12-01
feat(opentelemetry-jaeger): Add feature to enable statically linking isahc


kjvalencik:
# Summary

Default features in `isahc` are disabled which includes statically linking cURL. Add a feature flag to `opentelemetry-jaeger` to enable static linking.

## Use Case

Some environments do not have a system libcurl or may have an incompatible version. For example, iOS.

## API Sketch

```toml
opentelemetry-jaeger = { version = "0.9", features = ["collector_client", "static_curl"] }
```

```toml
[features]
collector_client = ["isahc", "http"]
static_curl = ["isahc/static-curl"]
```

jtescher:
I think this can be accomplished once https://github.com/open-telemetry/opentelemetry-rust/issues/337 is introduced. At that point you can bring whichever http client you prefer.

kjvalencik:
@jtescher That's awesome! That would be perfect. For my use case, `isahc` isn't a great choice either because it requires opening a UDP socket.

I'll close this issue since that feature is a superset of this one.

Issue #318, opened 2020-10-30
Add code coverage reporting


jtescher:
Follow up from https://github.com/open-telemetry/opentelemetry-rust/issues/143

TommyCpp:
I think https://codecov.io/ is a good choice for displaying result as many OSS projects are using it.

PR #2098, opened 2024-09-10
Use weaver for semantic convention codegen

lquerel:
## Changes

This PR marks the first step in the migration to Weaver for semconv code generation. The following changes have been made:
- Utilized Weaver’s new capabilities to simplify code generation as much as possible.
- Moved to SemConv v1.27.0.
- Added a blank line between each declaration for improved readability.
- Attribute notes have been included in comments.
- String examples are now represented using double quotes.
- Boolean examples are represented with true or false, instead of 'True' or 'False'.
- Number examples are represented without single quotes.
- Experimental attributes are now gated behind the `semconv_experimental` feature.
- Deprecated attributes include a note explaining the reason for deprecation.
- Sorted metric attributes by name.

Note: This PR represents the first step of the plan described [here](https://github.com/open-telemetry/opentelemetry-rust/issues/2100).


opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
```diff
@@ -20,54 +20,24 @@ git fetch origin "v$SPEC_VERSION"
 git reset --hard FETCH_HEAD
 cd "$CRATE_DIR"
 
-docker run --rm \
```
lquerel, line 23: We only need to call Weaver once, as the `weaver.yaml` file specifies what needs to be generated.

opentelemetry-semantic-conventions/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "opentelemetry-semantic-conventions"
-version = "0.25.0"
+version = "0.26.0"
```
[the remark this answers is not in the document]
lquerel: Fixed in [2b300bb](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/2b300bb06ff398e1ddd1d2ee09d55ad230ab09e1)

opentelemetry-semantic-conventions/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## vNext
 
+## v0.26.0
+### Changed
+
+- Starting with this version, this crate will use Weaver for the generation of
```
TommyCpp, line 8: Could you also include a note on `semconv_experimental`?

TommyCpp (commented):
Per my understanding this PR only changes the generation tool we use. The end results are still a list of static string right?

opentelemetry-semantic-conventions/scripts/templates/registry/rust/attribute.rs.j2
```diff
@@ -0,0 +1,24 @@
+{%- import 'macros.j2' as attr_macros -%}
+// DO NOT EDIT, this is an auto-generated file
+//
+// If you want to update the file:
+// - Edit the template at scripts/templates/registry/rust/attributes.rs.j2
+// - Run the script at scripts/generate-consts-from-spec.sh
```
TommyCpp, line 6: From CI prespective I think we need a CI check to make sure the commited version of `attribute.rs` is consistent with the one generate from script? I am thinking about the case where I accidentially typed something in the generated files and commit it.

lquerel:
> Per my understanding this PR only changes the generation tool we use. The end results are still a list of static string right?

@TommyCpp @cijothomas 

Yes, the initial intent was to minimize differences with the existing code generation first and then propose a more advanced integration that removes static strings entirely. However, if the community prefers to move directly to a step where static strings are replaced, such as for attribute declarations like:

```
pub const HTTP_REQUEST_METHOD: crate::attributes::AttributeKey<HttpRequestMethod> = crate::attributes::AttributeKey::new("http.request.method");
```

I am happy to update this PR accordingly. This would be straightforward since I already have a proof of concept (POC) for the step 2 described in #2100. Even if we don’t move directly to step 2, which is a fully type-safe API for Rust semconv, we could create an intermediate step (let’s call it step 1.5) where static strings are no longer visible to users or the compiler. I’d appreciate more feedback on this approach.

opentelemetry-semantic-conventions/CHANGELOG.md
lquerel: Added a short description in the changelog. See [c6af804](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/c6af8048553a39eee58f93a737437d80f9537a0e)

opentelemetry-semantic-conventions/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+  the semantic conventions.
+- Introduced a new feature `semconv_experimental` to enable experimental semantic conventions.
```
cijothomas, line 10: do you anticipate further features flags per signal/area so users can just opt-in to those that they need? 
Are there concerns usually about too big assembly with all the unused constants?

lquerel: We could indeed introduce a feature for each namespace (e.g., `db`, `container`, `url`, etc.) and for each signal type (e.g., `metric`, `span`, `event`, etc.). This would allow us to use configurations like `#[cfg(any(feature = "metrics", feature = "db"))]`. 

This should be simple to implement, so please let me know if you’d like to proceed in this direction.

cijothomas: not for this PR, I wanted to ask if it is feasible/easy, if demand arises for such a need!

opentelemetry-semantic-conventions/scripts/templates/registry/rust/attribute.rs.j2
lquerel: I agree. One option would be to update Weaver to generate a checksum (e.g., SHA256) for the generated code and store it in a file alongside the generated code. This would give us everything needed to create a CI check to ensure that no manual modifications have been made to the code in the repository, assuming we’ve also prevented rust fmt from altering the generated code. I can add this option to Weaver in a future release if the Rust Client SDK community is interested.

lquerel: Just created this feature request in Weaver. https://github.com/open-telemetry/weaver/issues/362

cijothomas: Thanks. This can be a follow up and not a blocker for the PR. @TommyCpp Do you agree to resolve this ?

TommyCpp: Yeah, a follow up sounds good

opentelemetry-semantic-conventions/src/attribute.rs
```diff
@@ -1,60 +1,79 @@
 // DO NOT EDIT, this is an auto-generated file
 //
 // If you want to update the file:
-// - Edit the template at scripts/templates/semantic_attributes.rs.j2
+// - Edit the template at scripts/templates/registry/rust/attributes.rs.j2
```
lalitb, line 4: The attributes.rs file currently defines semconv constants, which are then re-exported in various other modules. If there aren't any (or too many) shared definitions used across multiple modules (i.e trace.rs, metrics.rs, resource.rs), would it make sense to move these constants to their respective modules where they actually belong? The attributes.rs can grow over time as we add more definitions, and add the semconv for logs later, could be difficult to manage even though auto-generated.

I know this has been the existing behavior, but then right time to decide any such change.

opentelemetry-semantic-conventions/CHANGELOG.md
lalitb, line 10: This is breaking change ? Just prefix it with  `**Breaking**` if so ?

lalitb:
> pub const HTTP_REQUEST_METHOD: crate::attributes::AttributeKey<HttpRequestMethod> = crate::attributes::AttributeKey::new("http.request.method");

@lquerel  This looks awesome - in this case it will ensure that HTTP_REQUEST_METHOD will always have valid HTTP verbs. Just thinking in terms of the performance - we will possibly encapsulate these verbs in enum, and then encapsulate the `&static str for `http.request.method` in a structure `AttributeKey`. Do you see any performance overhead for using them in hot path (as comparison to the existing approach) ?

lquerel:
> > pub const HTTP_REQUEST_METHOD: crate::attributes::AttributeKey<HttpRequestMethod> = crate::attributes::AttributeKey::new("http.request.method");
> 
> @lquerel This looks awesome - in this case it will ensure that HTTP_REQUEST_METHOD will always have valid HTTP verbs. Just thinking in terms of the performance - we will possibly encapsulate these verbs in enum, and then encapsulate the `'&static str` for `http.request.method` in a structure `AttributeKey`. Do you see any performance overhead for (specifically) using them in hot path (as comparison to the existing approach) ?

@lalitb Yes, the verbs are indeed represented as an enum in the POC, as shown in the full declaration below. Regarding performance, `HTTP_REQUEST_METHOD` is a constant, so the overhead is minimal and primarily localized in the `AttributeKey::value` method, which can potentially be inlined. I also believe the current POC can be slightly optimized to avoid cloning the `&str`. While I don’t anticipate a noticeable overhead, we should confirm this with a proper benchmark. This approach is intermediary; the ultimate goal is to provide a fully type-safe API, where all hashmaps are replaced by structs for both mandatory and optional attributes. Additionally, metrics, spans, etc., will be predeclared with all their metadata such as instrument, description, unit, etc. (more efficient, and easier for end-users), and the SDK will be optimized to leverage this code generation. Once we reach that stage, I expect a significant reduction in overhead (but it's not this PR to be clear).

```rust
/// Attributes for the `client` namespace.
/// Attributes for the `error` namespace.
/// Attributes for the `exception` namespace.
/// Attributes for the `http` namespace.
/// Attributes for the `network` namespace.
/// Attributes for the `server` namespace.
/// Attributes for the `system` namespace.
/// Attributes for the `url` namespace.

/// A typed attribute key.
pub struct AttributeKey<T> {
    key: Key,
    phantom: std::marker::PhantomData<T>
}

impl <T> AttributeKey<T> {
    /// Returns a new [`AttributeKey`] with the given key.
    #[must_use]
    pub(crate) const fn new(key: &'static str) -> AttributeKey<T> {
        Self {
            key: Key::from_static_str(key),
            phantom: std::marker::PhantomData
        }
    }

    /// Returns the key of the attribute.
    #[must_use]
    pub fn key(&self) -> &Key {
        &self.key
    }
}

impl AttributeKey<StringValue> {
    /// Returns a [`KeyValue`] pair for the given value.
    #[must_use]
    pub fn value(&self, v: StringValue) -> KeyValue {
        KeyValue::new(self.key.clone(), v)
    }
}

impl AttributeKey<i64> {
    /// Returns a [`KeyValue`] pair for the given value.
    #[must_use]
    pub fn value(&self, v: i64) -> KeyValue {
        KeyValue::new(self.key.clone(), v)
    }
}

impl AttributeKey<f64> {
    /// Returns a [`KeyValue`] pair for the given value.
    #[must_use]
    pub fn value(&self, v: f64) -> KeyValue {
        KeyValue::new(self.key.clone(), v)
    }
}

impl AttributeKey<bool> {
    /// Returns a [`KeyValue`] pair for the given value.
    #[must_use]
    pub fn value(&self, v: bool) -> KeyValue {
        KeyValue::new(self.key.clone(), v)
    }
}

/// HTTP request method.
/// 
/// HTTP request method value SHOULD be "known" to the instrumentation.
/// By default, this convention defines "known" methods as the ones listed in [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-methods)
/// and the PATCH method defined in [RFC5789](https://www.rfc-editor.org/rfc/rfc5789.html).
/// 
/// If the HTTP request method is not known to instrumentation, it MUST set the `http.request.method` attribute to `_OTHER`.
/// 
/// If the HTTP instrumentation could end up converting valid HTTP request methods to `_OTHER`, then it MUST provide a way to override
/// the list of known HTTP methods. If this override is done via environment variable, then the environment variable MUST be named
/// OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS and support a comma-separated list of case-sensitive known HTTP methods
/// (this list MUST be a full override of the default known method, it is not a list of known methods in addition to the defaults).
/// 
/// HTTP method names are case-sensitive and `http.request.method` attribute value MUST match a known HTTP method name exactly.
/// Instrumentations for specific web frameworks that consider HTTP methods to be case insensitive, SHOULD populate a canonical equivalent.
/// Tracing instrumentations that do so, MUST also set `http.request.method_original` to the original value
///
/// ## Examples:
/// - GET
/// - POST
/// - HEAD
pub const HTTP_REQUEST_METHOD: crate::attributes::AttributeKey<HttpRequestMethod> = crate::attributes::AttributeKey::new("http.request.method");

/// HTTP request method
#[derive(Debug, Clone)]
#[non_exhaustive]
pub enum HttpRequestMethod {
    /// CONNECT method.
/// 
/// none
    Connect,
    /// DELETE method.
    Delete,
    /// GET method.
    Get,
    /// HEAD method.
    Head,
    /// OPTIONS method.
    Options,
    /// PATCH method.
    Patch,
    /// POST method.
    Post,
    /// PUT method.
    Put,
    /// TRACE method.
    Trace,
    /// Any HTTP method that the instrumentation has no prior knowledge of.
    Other,
    /// This variant allows defining a custom entry in the enum.
    _Custom(String),
}

impl HttpRequestMethod {
    /// Returns the string representation of the [`HttpRequestMethod`].
    #[must_use]
    pub fn as_str(&self) -> &str {
        match self {
            HttpRequestMethod::Connect => "CONNECT",
            HttpRequestMethod::Delete => "DELETE",
            HttpRequestMethod::Get => "GET",
            HttpRequestMethod::Head => "HEAD",
            HttpRequestMethod::Options => "OPTIONS",
            HttpRequestMethod::Patch => "PATCH",
            HttpRequestMethod::Post => "POST",
            HttpRequestMethod::Put => "PUT",
            HttpRequestMethod::Trace => "TRACE",
            HttpRequestMethod::Other => "_OTHER",
            HttpRequestMethod::_Custom(v) => v.as_str(),
            // Without this default case, the match expression would not
            // contain any variants if all variants are annotated with the
            // 'semconv_experimental' feature and the feature is not enabled.
            #[allow(unreachable_patterns)]
            _ => unreachable!(),
        }
    }
}

impl core::fmt::Display for HttpRequestMethod {
    /// Formats the value using the given formatter.
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        write!(f, "{}", self.as_str())
    }
}

impl crate::attributes::AttributeKey<HttpRequestMethod> {
    /// Returns a [`KeyValue`] pair for the given value.
    #[must_use]
    pub fn value(&self, v: &HttpRequestMethod) -> opentelemetry::KeyValue {
        opentelemetry::KeyValue::new(self.key.clone(), v.to_string())
    }
}
```

opentelemetry-semantic-conventions/src/attribute.rs
lalitb: Just to add - this is not a blocker for the PR, something to discuss and plan for subsequent if all agree.

lalitb:
> This approach is intermediary; the ultimate goal is to provide a fully type-safe API, where all hashmaps are replaced by structs for both mandatory and optional attributes

Thanks @lquerel for the details. We have seen the perf issues with the value part of `KeyValue` in the past, so the concern. I would vote for keeping the existing PR close to the current approach of exposing the `&static str`, and then go for the type-safe API, unless there are different thoughts by other @open-telemetry/rust-maintainers / @open-telemetry/rust-approvers .

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
```diff
@@ -6,7 +6,7 @@ CRATE_DIR="${SCRIPT_DIR}/../"
 
 # freeze the spec version and generator version to make generation reproducible
 SPEC_VERSION=1.27.0
-SEMCOVGEN_VERSION=0.25.0
+WEAVER_VERSION=v0.9.2
```
lalitb, line 9: To clarify, the process for generating the code remains unchanged, correct? Specifically, we still increment the spec/weaver version here and run this script as usual?

lquerel: Yes correct.

lalitb:
@lquerel Can you make it ready for review if all are aligned with the direction - specifically the CI is failing for now.

opentelemetry-semantic-conventions/scripts/templates/registry/rust/resource.rs.j2
```diff
@@ -1,21 +1,39 @@
...
-//!         KeyValue::new(semconv::resource::SERVICE_NAME, "my-service"),
-//!         KeyValue::new(semconv::resource::SERVICE_NAMESPACE, "my-namespace"),
-//!     ])))
-//!     .build();
-//! ```
+{%- import 'macros.j2' as attr_macros -%}
+// DO NOT EDIT, this is an auto-generated file
+//
+// If you want to update the file:
+// - Edit the template at scripts/templates/registry/rust/attributes.rs.j2
```
lalitb, line 5: there is no file with this name?

opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
```diff
@@ -0,0 +1,79 @@
+{%- import 'macros.j2' as metric_macros -%}
+// DO NOT EDIT, this is an auto-generated file
+//
+// If you want to update the file:
+// - Edit the template at scripts{{template}}
```
lalitb, line 5: some issue with auto-generation, as file name is missing?

opentelemetry-semantic-conventions/scripts/templates/registry/rust/attribute.rs.j2
```diff
@@ -0,0 +1,24 @@
...
+
+//! # Semantic Attributes
+//!
+//! The entire set of semantic attributes (or [conventions](https://opentelemetry.io/docs/concepts/semantic-conventions/)) defined by the project. The resource, metric, and trace modules reference these attributes.
+
+{% for root_ns in ctx %}
+    {% for attr in root_ns.attributes | rejectattr("name", "in", params.excluded_attributes) %}
+{{ [attr.brief, concat_if("\n\n# Notes\n\n", attr.note), attr_macros.examples(attr)] | comment }}
```
tbrockman, line 14: Nit: Title/Heading1/`#` results in a kinda large tooltip, any thoughts about reducing the size of the headers here and elsewhere?

opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
```diff
@@ -0,0 +1,79 @@
...
+//! let histogram = meter
+//!     .u64_histogram(semconv::metric::HTTP_SERVER_REQUEST_DURATION)
+//!     .with_unit("By")
+//!     .with_description("Duration of HTTP server requests.")
+//!     .init();
+//! ```
+
+{% for root_ns in ctx %}
+    {% for metric in root_ns.metrics %}
+{{ ["## Description\n\n", metric.brief, concat_if("\n\n# Notes\n\n", metric.note), metric_macros.examples(metric)] | comment }}
```
tbrockman, line 34: Nit
```suggestion
{{ ["## Description\n\n", metric.brief, concat_if("\n\n## Notes\n\n", metric.note), metric_macros.examples(metric)] | comment }}
```
Or increase the size of all headers to match notes here.

opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
```diff
@@ -0,0 +1,79 @@
...
+                {% endif %}
+            {% else %}
+                {% set req_level = "Unspecified" %}
+                {% set req_message = '' %}
+            {% endif %}
+/// | [`crate::attribute::{{ attribute.name | screaming_snake_case }}`] | `{{ req_level }}`{{ (': ' + req_message if req_message else '') }}
+        {% endfor %}
+        {% if metric.examples %}
+///
+/// # Examples
```
tbrockman, line 64: Same as above:
```suggestion
/// ## Examples
```

opentelemetry-semantic-conventions/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## vNext
 
+## v0.26.0
```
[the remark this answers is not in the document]
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

opentelemetry-semantic-conventions/CHANGELOG.md
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

opentelemetry-semantic-conventions/scripts/templates/registry/rust/resource.rs.j2
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

opentelemetry-semantic-conventions/scripts/templates/registry/rust/attribute.rs.j2
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
lquerel: Fixed in [4970ee9](https://github.com/open-telemetry/opentelemetry-rust/pull/2098/commits/4970ee9314ca5cb72376b21bb807ec14da0d4cc8)

lquerel:
@lalitb I’ve fixed most of the issues, but there are still a few problems related to unescaped markdown links in the semconv registry. `cargo doc` rejects markdown text like `blabla [0,n] blabla`. Instead of finding a workaround in the Rust template, I plan to add a new configuration for the comment filter to automatically escape invalid links. See: [Issue #374](https://github.com/open-telemetry/weaver/issues/374). This should be part of the next release.

lalitb:
> @lalitb I’ve fixed most of the issues, but there are still a few problems related to unescaped markdown links in the semconv registry. `cargo doc` rejects markdown text like `blabla [0,n] blabla`. Instead of finding a workaround in the Rust template, I plan to add a new configuration for the comment filter to automatically escape invalid links. See: [Issue #374](https://github.com/open-telemetry/weaver/issues/374). This should be part of the next release.

Thanks @lquerel. It seems if the CI is breaking, we need to wait for the next release. When is it planned?

lquerel:
@lalitb We are targeting a new release at the beginning of next week.

lquerel:
@lalitb @cijothomas Weaver release v0.10.0 fixed the issue with non-escaped invalid links present in the semconv registry that were causing problems with cargo doc. This PR is now ready for final review.

hdost (approved):

TommyCpp (approved):
LGTM. Thanks for getting start with this

lalitb (approved):
LGTM to start with. Thanks

opentelemetry-semantic-conventions/scripts/generate-consts-from-spec.sh
```diff
@@ -20,54 +20,24 @@ git fetch origin "v$SPEC_VERSION"
...
 if [[ "$(uname)" = "Darwin" ]]; then
   SED=(sed -i "")
 fi
 
 # Keep `SCHEMA_URL` key in sync with spec version
-"${SED[@]}" "s/\(opentelemetry.io\/schemas\/\)[^\"]*\"/\1$SPEC_VERSION\"/" src/lib.rs
+"${SED[@]}" "s/\(opentelemetry.io\/schemas\/\)[^\"]*\"/\1$SPEC_VERSION\"/" scripts/templates/registry/rust/weaver.yaml
+
+docker run --rm \
+  --mount type=bind,source=$CRATE_DIR/semantic-conventions/model,target=/home/weaver/source,readonly \
```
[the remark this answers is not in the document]
lquerel: Yes, there are advantages. For context, I’ve copied the relevant change log entry below.

(#322, #312,
#319, #318,
#312, #304
by @jsuereth ) Many improvements have been made to the creation of the Weaver Docker image,
which is now scoring an A on the Scout Docker image score.
- Add Weaver docker image to dependabot tracking,
- Add build attestations,
- Stop using root user the docker image,
- Use official docker action to build docker image,
Update docker to use release build.

lquerel:
@lalitb @TommyCpp Do you know why the merge into the main is blocked?

PR #2103, opened 2024-09-11
feat: API to modify/remove an existing entry from LogRecord attributes

### LINKED ISSUE #1986: API to modify/remove an existing entry from LogRecord attributes
I think we need to offer API to modify an existing entry, remove an existing entry too. These are critical to ensure we allow redaction, enrichment capabilities. Lets track it as separate issue for adding in the future.

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry-rust/pull/1985#discussion_r1704710222_

lalitb:
Fixes #1986

## Changes

This PR adds two new methods, `update_attribute` and `delete_attribute`, to the public API of the `LogRecord` struct. 
```rust
pub fn update_attribute(&Key, &AnyValue) -> Option<AnyValue>
```
- Updates the value of the first occurrence of an attribute with the specified key.
- If the attribute already exists, its value is replaced, and the old value is returned.
- If the attribute is not found, it is added as a new key-value pair, and None is returned.

```rust
pub fn remove_attribute(&mut self, key: &Key) -> usize
```
- Removes all the occurrence of  attributes with the specified key.
- The count of deleted attributes is returned .


opentelemetry-sdk/src/growable_array.rs
```diff
@@ -87,6 +87,72 @@ impl<
...
+    /// - If the value is found in the overflow vector, it is removed, and the remaining elements in the vector are shifted left to maintain order.
+    ///
+    /// # Arguments
+    ///
+    /// - `value`: A reference to the value to be deleted.
+    ///
+    /// # Returns
+    ///
+    /// - `Some(T)`: The deleted value, if found.
+    /// - `None`: If the value was not found in either the array or the vector.
```
cijothomas, line 110:
```suggestion
    /// - `None`: If the value was found in neither the array nor the vector
```

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -124,6 +125,76 @@ impl LogRecord {
...
+    /// Updates the first occurrence of an attribute with the specified key.
+    ///
+    /// This method searches for the first occurrence of the attribute with the given key
+    /// in the `attributes` collection. If the key is found, its value is updated with the
+    /// provided value. If the key is not found, the attribute is added.
+    ///
+    /// # Arguments
+    ///
+    /// - `key`: A reference to the key of the attribute to update.
+    /// - `value`: A reference to the new value for the attribute.
```
cijothomas, line 138: why ref, and not owned value?

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -124,6 +125,76 @@ impl LogRecord {
...
+            *attr = Some((key.clone(), value.clone()));
+            return old_value;
+        }
+
+        // If not found, add a new attribute
+        self.add_attribute(key.clone(), value.clone());
+        None
+    }
+
+    /// Deletes the first occurrence of an attribute with the specified key.
```
cijothomas, line 163: should it delete all occurrences instead of first? I understand that we don't de-dup, so if a user believes they accidentally put a PII/something into LogRecord, and if they ended up with duplicates too, the expectation when one calls delete is that that key/value is completely eliminated.

opentelemetry-sdk/src/logs/record.rs
lalitb: Good point. Have made it owned now.

opentelemetry-sdk/src/logs/record.rs
lalitb: I though delete first provides flexibility, and can be called repeatedly to delete all. But I understand, difficult to get use-case where user would like to delete specific/first entry. Have update it to delete all occurrences. Also renamed it to `remove_attribute`, to be consistent with `remove` method of Vector and HashMap.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -124,6 +125,73 @@ impl LogRecord {
...
+        {
+            // Take the old value and update the attribute
+            let old_value = attr.take().map(|(_, v)| v);
+            *attr = Some((key.clone(), value));
+            return old_value;
+        }
+
+        // If not found, add a new attribute
+        self.add_attribute(key.clone(), value.clone());
+        None
```
TommyCpp, line 160: Small nit on code style

maybe easier to read if we do 


```rust
match self
            .attributes
            .iter_mut()
            .find(|opt| opt.as_ref().map(|(k, _)| k == key).unwrap_or(false)) {
                     Some(attr) => {
                               // Take the old value and update the attribute
                               let old_value = attr.take().map(|(_, v)| v);
                               *attr = Some((key.clone(), value));
                               old_value
                        }
                     None => self.add_attribute(key.clone(), value.clone());
}
```

TommyCpp: I was little confused why we need to `add_attribute` regardless until I see the early return

TommyCpp: Or you have already tried that and compiler complains because the we are still holding mutable reference to self when iterator found nothing?

lalitb: > Or you have already tried that and compiler complains because the we are still holding mutable reference to self when iterator found nothing?

@TommyCpp yes, i did try that, and got the error for still holding the mutable reference to self. Looks like the borrow checker is bit conservative and doesn't automatically drop iterators/temporaries until the end of the scope where they are used.

cijothomas:
This is important, but trying to wrap up critical metric stuff/internal logging for the upcoming release, and will get back to this right after that.

cijothomas:
@lalitb Can you resurrect the PR?

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -64,6 +64,16 @@
  `LogExporter` trait no longer requires a mutable ref to `self`. If the exporter
  needs to mutate state, it should rely on interior mutability.
  [2764](https://github.com/open-telemetry/opentelemetry-rust/pull/2764)
+- Added Two new methods to the LogRecord struct's public API:
```
cijothomas, line 67:
```suggestion
- Added Two new methods to the `SdkLogRecord` struct's public API:
```

Can you share why we are not adding these new APIs to the `LogRecord` trait itself?

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -64,6 +64,16 @@
...
+```rust
+  update_attribute(&Key, &AnyValue) -> Option<AnyValue>
+```
+  - Updates the value of the first occurrence of an attribute with the specified key. 
+  - If the key exists, the old value is returned. If not, the new key-value pair is added, and None is returned.
```
cijothomas, line 72: the behavior of update-or-insert might call for a different name than `update_attribute`. is `set_attribute` better sounding?

opentelemetry-sdk/src/growable_array.rs
```diff
@@ -87,6 +87,71 @@ impl<
...
+    ///
+    /// # Returns
+    ///
+    /// - `Some(T)`: The deleted element, if found.
+    /// - `None`: If the index is out of bounds for both the internal array and the overflow vector.
+    ///
+    #[allow(dead_code)]
+    pub(crate) fn remove_at(&mut self, index: usize) -> Option<T> {
+        if index < self.count {
+            let removed_value = self.inline[index].clone();
```
cijothomas, line 114: is clone necessary? Can we "take" ownership?

opentelemetry-sdk/src/growable_array.rs
```diff
@@ -87,6 +87,71 @@ impl<
         self.count + self.overflow.as_ref().map_or(0, Vec::len)
     }
 
+    /// Removes the element at a specific position (index) while preserving the order.
```
cijothomas, line 90: Thanks for detailed writeup!

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -201,6 +202,73 @@ impl SdkLogRecord {
             .flatten()
             .any(|(k, v)| k == key && v == value)
     }
+
+    /// Updates the first occurrence of an attribute with the specified key.
+    ///
+    /// This method searches for the first occurrence of the attribute with the given key
+    /// in the `attributes` collection. If the key is found, its value is updated with the
+    /// provided value. If the key is not found, the attribute is added.
```
utpilla, line 210: > If the key is not found, the attribute is added.

We already have an `add_attribute` method which is meant for adding a new attribute. I think we should do one of the following:
1. Either make `update_attribute` strictly an update operation.
2. Or rename the method to `add_or_update_attribute` to reflect its true behavior.

opentelemetry-sdk/src/logs/record.rs
```diff
@@ -201,6 +202,73 @@ impl SdkLogRecord {
             .flatten()
             .any(|(k, v)| k == key && v == value)
     }
+
+    /// Updates the first occurrence of an attribute with the specified key.
```
utpilla, line 206: Similar to `remove_attribute`, we should update all the occurrences of the given attribute key. For example, I think it would make it easier for redaction processor authors to redact a "secret" attribute with certainty.

opentelemetry-sdk/CHANGELOG.md
utpilla: > Can you share why we are not adding these new APIs to the LogRecord trait itself?

+1

opentelemetry-sdk/src/growable_array.rs
```diff
@@ -87,6 +87,71 @@ impl<
...
+    ///
+    /// This function performs the following operations:
+    ///
+    /// - If the index points to an element in the internal array (`inline`):
+    ///   - Removes the element at the specified index.
+    ///   - Shifts the remaining elements in the internal array to the left to fill the gap, preserving the order.
+    ///   - If an overflow vector exists:
+    ///     - Moves the first element from the overflow vector into the last position of the internal array, if available.
```
utpilla, line 98: Can we avoid doing this? It seems like an unnecessary expensive operation. We maintain the `count` of the items in the array. It should be okay to have the `count` less than the array length while also having the `vec`.

cijothomas:
@lalitb Could you address review comments and resurrect this when you get a chance?

PR #2108, opened 2024-09-12
Mark `ObjectSafeProvider` and `GlobalMeterProvider::new()` as private

utpilla:
## Changes
- Mark `ObjectSafeProvider` and `GlobalMeterProvider::new()` as private
- These shouldn't be required by the users either. None of code samples for metrics use these.


lalitb (approved):
Make sense. The user would only use set_meter_provider() and meter_provider() methods which are public. Thanks.

PR #2109, opened 2024-09-12
update trait `Number<T>` to `Number`

fraillt:
## Changes

Simply updated `trait Number<T>` by removing unnecessary generic type from it.

There is exactly these changes (you can repeat it in 1 min)
* search for regex `Number(<[a-zA-Z0-9]*>)` replace with `Number`
* in `trait Number` replace `T` with `Self`.

What is the value? It's easier to understand to newcomers such as myself :)

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -71,7 +71,7 @@ pub(crate) struct AggregateBuilder<T> {
 
 type Filter = Arc<dyn Fn(&KeyValue) -> bool + Send + Sync>;
 
-impl<T: Number<T>> AggregateBuilder<T> {
+impl<T: Number> AggregateBuilder<T> {
     pub(crate) fn new(temporality: Option<Temporality>, filter: Option<Filter>) -> Self {
         AggregateBuilder {
             temporality,
--- a/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
@@ -32,7 +32,7 @@ struct ExpoHistogramDataPoint<T> {
     zero_count: u64,
 }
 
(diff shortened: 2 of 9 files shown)


cijothomas (approved):
Thanks for another cleanup/simplification. The existing thing was probably done to meet some needs in the Histogram aggregation, not quite sure why. 
This is still good to merge, and we can do another deep review/cleanup of Histograms.

(Histograms SUM should always be double, but looks like we have it as <T>. Good to check it further after this PR.)

cijothomas:
Looks like CLA is failing, similar to https://github.com/communitybridge/easycla/issues/4329

utpilla (approved):

fraillt:
> Looks like CLA is failing, similar to [communitybridge/easycla#4329](https://github.com/communitybridge/easycla/issues/4329)

Sorry... I keep forgetting ;(

fraillt:
> This is still good to merge, and we can do another deep review/cleanup of Histograms.

I'm actually looking at this right now :) it seams that both `ExpoHistogram` and `Histogram` could share some logic.
At the moment I see at least few issues:
* `ExpoHistogram` - uses it's own implementation for attribute tracking, instead of utilizing `ValueMap`, and for this reason it doesn't have `is_under_cardinality_limit` check, and other improvements.
* `Histogram` - uses `ValueMap` which is optimized for AtomicOperations, but histogram measurements cannot be performed in atomic maneir, fo this reason every measurement get's atleast to locks: one for attribute map, another for actually updating histogram aggregate.

So If I'll create a new revision soon to address this ;)

cijothomas:
ExponentialHistogram was left out from initial set of changes to leverage ValueMap, just to keep scope under control!
Yes if we can somehow avoid the dual lock (1st one is really a read lock) for histogram, that'd be awesome.

cijothomas:
Merging, as CLA issue is resolved. Other proposals looks good, happy to accept them via follow up PRs!

PR #2112, opened 2024-09-12
[Breaking] Enforce meter name, version schema_url to be static string slices 

utpilla:
## Context
- `MeterProvider` trait is not object safe because the methods of `MeterProvider` trait have generic parameters
-  Since it's not object safe, we cannot use it as it is to switch from a no-op provider to an SDK provider
-  We have a public struct `GlobalMeterProvider` and a trait `ObjectSafeMeterProvider` to work around this
- `Meter` name, version and schema_url should be static string slices. Doing so would also make `MeterProvider` object safe.

## Changes
- Make `MeterProvider` trait object safe by updating its method signatures to accept `&'static str` types instead of the generic `impl Into<Cow<'static, str>>`
- Remove the public struct `GlobalMeterProvider` and `ObjectSafeMeterProvider` trait as they are no longer necessary

### DIFF
--- a/opentelemetry/src/metrics/meter.rs
+++ b/opentelemetry/src/metrics/meter.rs
@@ -29,23 +29,18 @@ pub trait MeterProvider {
     /// let provider = global::meter_provider();
     ///
     /// // meter used in applications
-    /// let meter = provider.meter("my_app");
+    /// let meter = provider.meter("my_app".to_string());
     ///
     /// // meter used in libraries/crates that optionally includes version and schema url
     /// let meter = provider.versioned_meter(
-    ///     "my_library",
-    ///     Some(env!("CARGO_PKG_VERSION")),
-    ///     Some("https://opentelemetry.io/schema/1.0.0"),
+    ///     "my_library".to_string(),
+    ///     Some(env!("CARGO_PKG_VERSION").to_string()),
+    ///     Some("https://opentelemetry.io/schema/1.0.0".to_string()),
     ///     Some(vec![KeyValue::new("key", "value")]),
     /// );
     /// ```
-    fn meter(&self, name: impl Into<Cow<'static, str>>) -> Meter {
-        self.versioned_meter(
-            name,
-            None::<Cow<'static, str>>,
-            None::<Cow<'static, str>>,
-            None,
-        )
+    fn meter(&self, name: String) -> Meter {
+        self.versioned_meter(name, None, None, None)
     }
 
     /// Returns a new versioned meter with a given name.
@@ -56,9 +51,9 @@ pub trait MeterProvider {
     /// default name will be used instead.
     fn versioned_meter(
         &self,
-        name: impl Into<Cow<'static, str>>,
-        version: Option<impl Into<Cow<'static, str>>>,
-        schema_url: Option<impl Into<Cow<'static, str>>>,
+        name: String,
+        version: Option<String>,
+        schema_url: Option<String>,
         attributes: Option<Vec<KeyValue>>,
     ) -> Meter;
 }
--- a/opentelemetry/src/metrics/meter.rs
+++ b/opentelemetry/src/metrics/meter.rs
@@ -29,17 +29,17 @@ pub trait MeterProvider {
     /// let provider = global::meter_provider();
     ///
     /// // meter used in applications
-    /// let meter = provider.meter("my_app".to_string());
+    /// let meter = provider.meter("my_app");
     ///
     /// // meter used in libraries/crates that optionally includes version and schema url
     /// let meter = provider.versioned_meter(
-    ///     "my_library".to_string(),
-    ///     Some(env!("CARGO_PKG_VERSION").to_string()),
-    ///     Some("https://opentelemetry.io/schema/1.0.0".to_string()),
+    ///     "my_library",
+    ///     Some(env!("CARGO_PKG_VERSION")),
+    ///     Some("https://opentelemetry.io/schema/1.0.0"),
     ///     Some(vec![KeyValue::new("key", "value")]),
     /// );
     /// ```
-    fn meter(&self, name: String) -> Meter {
+    fn meter(&self, name: &'static str) -> Meter {
         self.versioned_meter(name, None, None, None)
     }
 
@@ -51,9 +51,9 @@ pub trait MeterProvider {
     /// default name will be used instead.
     fn versioned_meter(
         &self,
-        name: String,
-        version: Option<String>,
-        schema_url: Option<String>,
+        name: &'static str,
+        version: Option<&'static str>,
+        schema_url: Option<&'static str>,
         attributes: Option<Vec<KeyValue>>,
     ) -> Meter;
 }
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -83,6 +83,7 @@ mod reqwest {
         }
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
     #[async_trait]
     impl HttpClient for reqwest::blocking::Client {
         async fn send(&self, request: Request<Vec<u8>>) -> Result<Response<Bytes>, HttpError> {
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Update `async-std` dependency version to 1.13
 - *Breaking* - Remove support for `MetricProducer` which allowed metrics from
   external sources to be sent through OpenTelemetry.
-  [#2105](https://github.com/open-telemetry/opentelemetry-rust/pull/2105)  
+  [#2105](https://github.com/open-telemetry/opentelemetry-rust/pull/2105)
+- Feature: `SimpleSpanProcessor::new` is now public [#2119](https://github.com/open-telemetry/opentelemetry-rust/pull/2119)
 
 ## v0.25.0
 
--- a/opentelemetry-sdk/benches/metric.rs
+++ b/opentelemetry-sdk/benches/metric.rs
@@ -149,7 +149,10 @@ fn bench_counter(view: Option<Box<dyn View>>, temporality: &str) -> (SharedReade
         builder = builder.with_view(view);
     }
     let provider = builder.build();
-    let cntr = provider.meter("test").u64_counter("hello").init();
+    let cntr = provider
+        .meter("test".to_string())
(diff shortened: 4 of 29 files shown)


TommyCpp (commented):
If the concern here is object safety. I wonder if we can ask the implementation of `MeterProvider` to be `Sized`. Something like this https://rust-lang.github.io/api-guidelines/flexibility.html#traits-are-object-safe-if-they-may-be-useful-as-a-trait-object-c-object

TommyCpp:
> One less public API to maintain

I don't think it's a huge overhead. We basically only repeating the `MeterProvider` trait implementation here, which has been pretty stable. In generally I want to use generic parameters as much as possible. I think in this case the maintance overhead is not worth the regression on readbitliy or ergonomics

utpilla:
> If the concern here is object safety. I wonder if we can ask the implementation of `MeterProvider` to be `Sized`. Something like this https://rust-lang.github.io/api-guidelines/flexibility.html#traits-are-object-safe-if-they-may-be-useful-as-a-trait-object-c-object

@TommyCpp The concern here is the extra bit of code (`GlobalMeterProvider` + `ObjectSafeMeterProvider`) which exists ONLY to get around `MeterProvider` not being object safe. I think we should get rid of this code as it seems unnecessary.

The `Self:Sized` approach won't help as it would exclude the methods from the trait object which means we wouldn't be able to call `meter()` or `versioned_meter()` on the `MeterProvider` trait object.

utpilla:
> > One less public API to maintain
> 
> I don't think it's a huge overhead. We basically only repeating the `MeterProvider` trait implementation here, which has been pretty stable. In generally I want to use generic parameters as much as possible. I think in this case the maintance overhead is not worth the regression on readbitliy or ergonomics

Apart from reducing the maintenance overhead, removing the public API also allows us to change things in the global metrics setup more freely if required in future. We won't be tying the setup process to any type other than `MeterProvider`.

There is no regression on readability/ergonomics, if the user is using `global::meter()` to create the meter. I believe our documentation also showcases this as the approach to create `Meter`. The breaking change will affect users only if they create meters using `SdkMeterProvider.meter()`. If we feel that there would be a lot of users who might be calling `meter()` on `SdkMeterProvider` instead of `global` we could consider not making this change.

TommyCpp:
> removing the public API also allows us to change things in the global metrics setup more freely if required in future

Generally I'd agree. But in this case `GlobalMeterProvider` is just implementing the `MeterProvider`. Thus I see little risk `GlobalMeterProvider` will prevent us from making any breaking changes.

TommyCpp:
> There is no regression on readability/ergonomics, if the user is using `global::meter()` to create the meter

True. I am thinking about the regression on readability/ergonomics for users that don't use `global` module. I feel even if they are probably not the majority, the reward we get from this PR doesn't warrant making their life harder. But it's pretty subjective

opentelemetry/src/metrics/meter.rs
```diff
@@ -29,23 +29,18 @@ pub trait MeterProvider {
...
     /// );
     /// ```
-    fn meter(&self, name: impl Into<Cow<'static, str>>) -> Meter {
-        self.versioned_meter(
-            name,
-            None::<Cow<'static, str>>,
-            None::<Cow<'static, str>>,
-            None,
-        )
+    fn meter(&self, name: String) -> Meter {
```
cijothomas, line 42: `fn meter(&self, name: &'static str)`

@TommyCpp @utpilla Can we restrict meter name to be static strings only? Are there scenarios which we prohibit with this? This should avoid the concern Zhongyang raised about reduced ergonomics, I think.

(I am generally motivated by `tracing` crates strategy of restricting things to 'static in many places. In this case, this is just meter name, so I think we can steal that strategy.)

TommyCpp: That should work! But insetad of `'static str` can we do `'a str`. IIUC we need to copy the string anyway so we just need to scope the lifetime of the string as long as the `meter` function. I expect most users to use static string so we will only impact users that doesn't using global module + doesn't use `'static str`

I was trying to think of any case someone won't have it as `'static str` . And the only viable cases (I can think of) is probably [fetching meter name from environment variables](https://doc.rust-lang.org/std/env/fn.var.html).

TommyCpp: Per my understanding having a lifetime parameter on inputs only doesn't impact object safety but let me know if the compiler starts yelling 🤣

utpilla: @TommyCpp Are you suggesting changing the API to this?

```diff
- fn meter(&self, name: impl Into<Cow<'static, str>>) -> Meter 
+ fn meter(&self, name: &str) -> Meter
```
Doing this, would still break users who could have been passing an owned `String` as the value for `name`. They would now have to update the call sites to use `String.as_str()`. Essentially, no matter how we remove generics (Into<T>) from argument type, we would end up breaking some or the other use case. Is that still okay with you?

Having said that, I'm okay with using `&str` as the argument type for meter `name`.

cijothomas: & 'a str also suffer from same concern as this PR. I think all of us are agreeing for `&'static str`, so let us proceeed with that?
(The rare scenario where meter name comes from somewhere else, Box::leak can rescue. This should be acceptable as the scenario itself is rare and also, there is usually 1-2 meter in any app!).

cijothomas: @TommyCpp Can you confirm if we can proceed with `&'static str`?
(A similar optimization can be ported to traces (and eventually logs) too)

lalitb: > and eventually logs

logs doesn't have these global components to remove to reduce the API surface.

cijothomas: Logs - not today, but possible we need to expose it, as it maybe [promoted as a user facing API](https://github.com/open-telemetry/oteps/pull/265)

TommyCpp: > we would end up breaking some or the other use case

I am still leaning towards `'a str` as it's a good middleground IMO. I will skip on why `'a &str` or `'static &str` is better than `String` as it's clear to everyone. 

Comparing `'static &str` and `'a &str`. let's say you are a user that somehow are using `String` to fetch a meter. `s.to_str()` is a better change than [`s.leak()`](https://doc.rust-lang.org/nightly/alloc/string/struct.String.html#method.leak) as the `leak` function(as the name sugguest?) can cause memory leak 

Plus I think semantically there is no reason why the meter name needs to be `'static str` if it's used only to fetch the meter object

cijothomas: > Comparing 'static &str and 'a &str. let's say you are a user that somehow are using String to fetch a meter. s.to_str() is a better change than [s.leak()](https://doc.rust-lang.org/nightly/alloc/string/struct.String.html#method.leak) as the leak function(as the name sugguest?) can cause memory leak

I don't fully follow the argument that it cause memory leak. Can you elaborate? Are there scenarios where someone keeps creating Meters repeatedly?

Here is my thinking:
1. Meter name is meant to indicate the scope of the instrumentation like package name, module name etc. They can always fit the `& 'static str`.
2. It is possible for someone to make meter name using dynamic strings, but I am not aware of any real reason why is that needed. There can be reasons like, "My meter name comes from env variable or an incoming http header etc.", but I don't think they are worth optimizing for. (TBH, I have never seen such needs ever).
3. Even if someone really needs 2, then they can use leak api, to keep the meter name alive for the rest of the application. The heap memory used for meter name will remain forever, even after dropping otel metrics completely. With static, the same amount of memory would have still been used, but just allocated in a different location. With &'a str, same amount of memory would be used (as we copy it), but this can be cleaned up with drop, but such create-cleanup-re-create-repeat is not a pattern we are promoting. 
4. If someone keeps creating new Meters for, say, each incoming request, then yes, they'll leak memory. But I don't think we need to worry about that. MeterProvider, Meter (and instruments) are meant to be created once, and reused. If users are repeatedly trying to create Meters, then that itself is an issue.

> Plus I think semantically there is no reason why the meter name needs to be 'static str if it's used only to fetch the meter object

I actually think 'static str does promote the idea that meter name is for scope, that is known ahead. Which is why I am proposing to make that case, irrespective of the use case shown in this PR. In other words, with or without this PR, I suggest to keep meter scope names 'static str only.

(I think we need to further exploring the idea of 'static str in other places like Attribute Keys, Instrument Names etc. but I don't have a strong argument to push for that, though `tracing` does that today.)

TommyCpp: Ah good point on the Meter name should be a compling value. Let's move forward with `'static str`. We can include pointer to the spec so users understand the constrain here

> I don't fully follow the argument that it cause memory leak. Can you elaborate

If we only consider the SDK we provided then there is no concern on memory leak. But there is no guarantee which SDK end users are using, depends on the underlying implementation it may retain the string more than needed. 

The tricky part is API can be used by library users. So we are basically forcing the library users to leak memory to an unknown implementation(as SDK is chosen by users). But if it's supposed to be a compiling value then there is no reason why library users shouldn't have a `'static str` already.

cijothomas: Thanks. Lets proceed by making &'static for MeterName.

cijothomas (changes requested):
This is good cleanup, but lets do it by making 'static str for MeterNames, as discussed here : https://github.com/open-telemetry/opentelemetry-rust/pull/2112#discussion_r1761521317

opentelemetry/src/metrics/meter.rs
```diff
@@ -56,9 +51,9 @@ pub trait MeterProvider {
     /// default name will be used instead.
     fn versioned_meter(
         &self,
-        name: impl Into<Cow<'static, str>>,
```
cijothomas, line 59: @stormshield-fabs FYI. https://github.com/open-telemetry/opentelemetry-rust/issues/1527 is still to be added for Metrics, which will remove versioned_meter in favor of builder pattern.

cijothomas (approved):
LGTM.
Please follow up with a breaking change entry to changelog for this (and previous 2 PRs that changed public api)

lalitb (approved):
LGTM. Based on the following assumptions:
- Meter names are generally recommended to be static strings defined at compile time.
- There will be no need to make object unsafe addition to `MeterProvider` trait in future.

cijothomas:
@TommyCpp merging now to unblock progress on api. We can follow up if there are additional feedback to be addressed.

PR #2113, opened 2024-09-12
Mark SdkMeter as pub(crate)

utpilla:
## Changes
- `SdkMeter` should not be `public`

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter.rs
+++ b/opentelemetry-sdk/src/metrics/meter.rs
@@ -42,7 +42,7 @@ const INSTRUMENT_UNIT_INVALID_CHAR: &str = "characters in instrument unit must b
 /// See the [Meter API] docs for usage.
 ///
 /// [Meter API]: opentelemetry::metrics::Meter
-pub struct SdkMeter {
+pub(crate) struct SdkMeter {
     scope: Scope,
     pipes: Arc<Pipelines>,
     u64_resolver: Resolver<u64>,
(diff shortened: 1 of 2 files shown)


cijothomas (approved):

TommyCpp (commented):
This is a breaking change so maybe a change note?

cijothomas:
> This is a breaking change so maybe a change note?

+1. We also need to add one for https://github.com/open-telemetry/opentelemetry-rust/pull/2085 too

cijothomas:
Merging. Missing changelogs can be added in a follow up for both together.

hdost:
Can i get a rational on this? Perhaps I am missing some context. Is it because users should use a builder and not the SdkMeter directly?

cijothomas:
> Can i get a rational on this? Perhaps I am missing some context. Is it because users should use a builder and not the SdkMeter directly?

From the examples:

    let meter = global::meter("mylibraryname");
// OR
// let meter = meter_provider.meter("mylibraryname");

    // Create a Counter Instrument from above meter
    let counter = meter.u64_counter("my_counter").init();

SdkMeter not required to do this, and OTel's expectation is also rely on API only, not SDK constructs like SdkMeter, as that would tie their instrumentation to a specific SDK implementation (negating one of the original goals of OTel).

I also look at this from another angle - some pub thing was removed, and still all examples continue to compile, indicating that this should not have been public in the first place. Should work _most_ of the time to help clean up public APIs to the absolute minimum required.

PR #2114, opened 2024-09-13
Unify Histogram and ExpHistogram aggregation

fraillt:
Fixes #
Introducing common abstraction solved several issues with `ExpoHistogram`:
* added `is_under_cardinality_limit`
* improved performance GREATLY +100%.

Also fixed broken benchmark for histogram collection.

## Changes

Provided abstractions to unify `Histogram` and `ExpoHistogram` behaviour.
There's two new types:
* `trait Aggregator` - provides aggregation specific implementation
* `struct AttributeSetAggregation` - has two tasks:
  * select and update specific aggregator very efficiently. (very similar to current `ValueMap` implementation.)
  * provide functions to collect (and reset) all measurements for DataPoints.
  
 Generally performance for Histogram is not changes much, only collection phase was improved 10-30% percent (the more data you have the bigger the improvement)

### DIFF
--- a/opentelemetry-sdk/benches/metric.rs
+++ b/opentelemetry-sdk/benches/metric.rs
@@ -3,6 +3,7 @@ use std::sync::{Arc, Weak};
 
 use criterion::{criterion_group, criterion_main, Bencher, Criterion};
 use opentelemetry::{
+    global,
     metrics::{Counter, Histogram, MeterProvider as _, Result},
     Key, KeyValue,
 };
@@ -344,7 +345,7 @@ fn counters(c: &mut Criterion) {
 const MAX_BOUND: usize = 100000;
 
 fn bench_histogram(bound_count: usize) -> (SharedReader, Histogram<u64>) {
-    let mut bounds = vec![0; bound_count];
+    let mut bounds: Vec<usize> = vec![0; bound_count];
     #[allow(clippy::needless_range_loop)]
     for i in 0..bounds.len() {
         bounds[i] = i * MAX_BOUND / bound_count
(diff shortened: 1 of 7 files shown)


fraillt:
> @fraillt I don't completely follow the need for unifying abstractions for the two histograms?
> 
> The regular histogram is already using the efficient lookup offered by `ValueMap`. If all we want is `ExponentialHistogram` to have an efficient lookup and respect cardinality limit, why can't we just update `ExponentialHistogram` to use `ValueMap` (similar to the regular Histogram)?

Good question.
I tried to look at how to improve ExponentialHistogram (it was lacking performances and features), but it wasn't possible to easily do that with existing `ValueMap` implementation.
* I would have to add yet another method in `AtomicTracker` it already has `update_histogram(bucket_size)` (removed in this PR), so I would had to add yet another method `update_exp_histogram(....)`.
* also `ValueMap` has `measure(measurement, attrs, index)` method, the last argument is only for histograms, but for exp histogram I think i needed few more arguments...
* I would be forced to store more data in HistogramDataPoint, because interface doesn't allow to pass in configuration
I think that's the reason why ExpHistogram didnt use `ValueMap`, because it doesn't fit well...

So I had a choice:
* either modify existing transaction (and all other aggregations with it) to somehow make it work with `ExpHistogram`, with the cost that it will not be efficient anyway.
* or avoid refactoring existing code and come up with efficient abstraction for histograms.

I choose the other option :) What is nice about it, is that it should work perfectly fine with counter types as well!!!
Eventually, I would like to have one efficient/consistent abstraction for all metrics and I think it is achievable.
As a comparison you can compare Histogram implementation.
- main branch - 352 lines of code
- my branch - 201 lines of code
Not saying that lines of code is good or bad, but in my opinion this new interface abstracted a lot of details and now you can focus on core functionality instead.

cijothomas:
@fraillt Thanks for continuing your great work! I'd suggest to keep the scope of each PR very small, so it is easy to review and share feedbacks. (Metrics hot path code reviews need lot of focused attention, so smaller PRs are really appreciated here.)

Also, we'd like to keep ExponentialHistogram and ExplicitHistogram separate, and let ExponentialHistogram pickup `ValueMap` for its own improved speed. ExponentialHistogram is a relatively new entrant, and we need to do extensive testing before marking it stable. I highly doubt if we can do that by 1.0, so separating it is preferred.

if you can join our community meetings (Tuesdays 9 AM Pacific), that'd be a good time to discuss some of the design ideas/other improvements! Most of the meeting time is spent discussing non-trivial problems, and you can definitely help there too, if you have the bandwidth to join. (If the time is not friendly to your time zone, we can do ad-hoc meetings to allow more participations)

fraillt:
@cijothomas Would be interesting to join community meeting :) Timing is alright as well (19 PM EEST my time).

fraillt:
I'll close this, because as you pointed out, this might be too big to review.
I have created #2117 that uses same new `Aggregator` interface, but should be more easy to review, as I tried to make it as small as possible.
This revision might serve as an example what I want to do next, in smaller steps. Providing unified `collect` function is really important, as I have encountered several subtle bugs or redundant `clone`s that probably happened due to copy/paste, so having unified `collect` function(s) could solve these inconsistencies, and provide performance improvements across all metrics.

cijothomas:
> @cijothomas Would be interesting to join community meeting :) Timing is alright as well (19 PM EEST my time).

Thanks! See ya at the meeting!

PR #2119, opened 2024-09-16
feature: expose `SimpleSpanProcessor::new`

### LINKED ISSUE #2060: [Regression]: opentelemetry_sdk::TracerProvider::span_processors no longer public
### What happened?

PR https://github.com/open-telemetry/opentelemetry-rust/pull/1755 did some API clean-ups, but it was too aggressive and turned some useful APIs private. In particular `opentelemetry_sdk::TracerProvider::span_processors` was `pub` but is now only `pub(crate)`.

I have an HTTP server that acts as an OTLP proxy. A tracer provider is configured with exporters. This tracers is then used either to directly write traces for the main server, or to forward traces received through `POST /v1/traces/`.
To forward already existing spans received through OTLP, you need access to the `span_processors`. This was previously exposed by `opentelemetry_sdk::TracerProvider::span_processors` in version `0.23.0` but since version `0.24.0` it is no longer exposed.

Please expose this method so it is possible to wrap a tracer provider into a higher level tracer provider adapter.

### API Version

0.24.0

### SDK Version

0.24.1

OTLP

### Relevant log output

```shell
error[E0624]: method `span_processors` is private
  --> crates/core/src/opentelemetry/mod.rs:28:43
   |
28 |       Self::Regular(provider) => provider.span_processors(),
   |                                           ^^^^^^^^^^^^^^^ private method
   |
  ::: /home/demurgos/.cargo/registry/src/index.crates.io-6f17d22bba15001f/opentelemetry_sdk-0.24.1/src/trace/provider.rs:92:5
   |
92 |     pub(crate) fn span_processors(&self) -> &[Box<dyn SpanProcessor>] {
   |     ----------------------------------------------------------------- private method defined here
```

demurgos:
Fixes #2060

## Changes

This commit update the visibility of the method `SimpleSpanProcessor::new` from `pub(crate)` to `pub`. This enables consumers to create `SimpleSpanProcessor` values for use with `trace::provider::Builder::with_span_processor`.

In particular, this fixes a consistency issue: the `BatchSpanProcessor` struct may already be built thanks to its `pub` builder. With this change, both processors in this repo may be built manually.

* [x] Unit tests added/updated (if applicable) (not applicable)

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Update `async-std` dependency version to 1.13
 - *Breaking* - Remove support for `MetricProducer` which allowed metrics from
   external sources to be sent through OpenTelemetry.
-  [#2105](https://github.com/open-telemetry/opentelemetry-rust/pull/2105)  
+  [#2105](https://github.com/open-telemetry/opentelemetry-rust/pull/2105)
+- Feature: `SimpleSpanProcessor::new` is now public [#2119](https://github.com/open-telemetry/opentelemetry-rust/pull/2119)
 
 ## v0.25.0
 
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -109,7 +109,8 @@ pub struct SimpleSpanProcessor {
 }
 
 impl SimpleSpanProcessor {
-    pub(crate) fn new(exporter: Box<dyn SpanExporter>) -> Self {
+    /// Create a new [SimpleSpanProcessor] using the provided exporter.
+    pub fn new(exporter: Box<dyn SpanExporter>) -> Self {
         Self {
             exporter: Mutex::new(exporter),
         }


lalitb (approved):
looks good, specifically that it allows us to remove `with_simple_processor` and `with_batch_processor` methods in future as reasoned [here](https://github.com/open-telemetry/opentelemetry-rust/issues/2060#issuecomment-2352701790).

TommyCpp (approved):

cijothomas:
While I agree with this change (Thanks for fixing the inconsistency), I just want to call out that `SimpleSpanProcessor` is *not* going to scale well if you use it on any PROD scenarios. It makes blocking calls, so a Span.End might be blocked waiting for an http response, so Simpleprocessor is best left for debug/learning purposes. As long as you are leveraging it for test purposes no issues!

demurgos:
@cijothomas Thanks, I confirm that I only use it in development. In production I use the bathed span processor.

The simple span processor has the benefit of being immediate, which helps when debugging locally with traces sent to stdout.

PR #2125, opened 2024-09-17
Update CHANGELOG

utpilla:
## Changes
- Update CHANGELOG with the recent public API changes

### DIFF
--- a/opentelemetry/CHANGELOG.md
+++ b/opentelemetry/CHANGELOG.md
@@ -3,9 +3,11 @@
 ## vNext
 
 - **BREAKING** Public API changes:
-  - **Removed**: `Key.bool()`, `Key.i64()`, `Key.f64()`, `Key.string()`, `Key.array()` [#2090](https://github.com/open-telemetry/opentelemetry-rust/issues/2090)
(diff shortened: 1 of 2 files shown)


opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- **BREAKING** Public API changes:
+  - **Removed**: `Key.bool()`, `Key.i64()`, `Key.f64()`, `Key.string()`, `Key.array()` [#2090](https://github.com/open-telemetry/opentelemetry-rust/issues/2090)
```
cijothomas, line 6: for places where it is feasible to provide alternatives, lets do that right in the changelog.md itself, so its easy to perform this migration.

for things are unlikely to affect users (like removing ObjectSafe..), can we mention something like "Removed abc, as this was unnecessarily public"/similar wording?

utpilla: I have updated the CHANGELOG to add the relevant information. Let me know if it needs further improvement.

cijothomas (approved):

PR #2128, opened 2024-09-19
Improve internal opentelemetry logging

### LINKED ISSUE #1146: Proposal to improve internal error logging

lalitb:
Fixes #1146 

## Changes

Created the draft PR to discuss the approach as mentioned [here](https://github.com/open-telemetry/opentelemetry-rust/issues/1146#issuecomment-2357990216)

Have added comment on some of the open issues.

Sample out of the appender-tracing example:

Scenario 1: only output internal `error`:

image

Scenario 2: output `info` and above:
image

Scenario 3: output 'debug' and above:
image


opentelemetry-appender-tracing/Cargo.toml
```diff
@@ -12,16 +12,16 @@ rust-version = "1.65"
 
 [dependencies]
 log = { workspace = true, optional = true }
-opentelemetry = { version = "0.25", path = "../opentelemetry", features = ["logs"] }
+opentelemetry = { version = "0.25", path = "../opentelemetry", features = ["logs","experimental-internal-debugging"] }
```
lalitb, line 15: name of the experimental flag - `experimental-internal-debugging` ?

opentelemetry-appender-tracing/examples/basic.rs
```diff
@@ -15,9 +18,19 @@ fn main() {
...
-    error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");
+    let env_filter = EnvFilter::from_default_env();
+    let otel_layer = layer::OpenTelemetryTracingBridge::new(&provider);
+    let internal_log_layer = fmt::Layer::default()
+        .with_writer(std::io::stdout) // Writes to stdout
+        .pretty()
+        .with_filter(filter_fn(|meta| meta.target().starts_with("opentelemetry"))) // Custom filter function
+        .with_filter(env_filter);
+    tracing_subscriber::registry()
+        .with(internal_log_layer)
```
lalitb, line 29: as the `internal_log_layer` can be added only after the `LoggerProvider` is created, any internal logs till that time won't be logged. This would be the case for logs and traces, and not metrics.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -157,6 +157,9 @@ where
         event: &tracing::Event<'_>,
         _ctx: tracing_subscriber::layer::Context<'_, S>,
     ) {
+        if event.metadata().target().starts_with("opentelemetry") {
+            return;
+        }
```
lalitb, line 162: To prevent infinite logging, events with the target prefix `opentelemetry` are suppressed at the `on_event` level.  Suppressing at the `enabled` level would block these events from reaching other layers (both higher and lower in the order). While suppressing at the `on_event` level introduces a minimal overhead, this approach is safer as it ensures that if the application owner forgets to handle suppression, the application won't crash with a stack overflow. We need to discuss the best approach based on these considerations.

opentelemetry/src/global/internal_logging.rs
```diff
@@ -0,0 +1,130 @@
+#![allow(unused_macros)]
+/// Macro for logging messages at the general log level in OpenTelemetry.
+///
+/// # Fields:
+/// - `target`: The component or module generating the log. This identifies which crate or system part is logging the message (e.g., "opentelemetry-sdk", "opentelemetry-otlp").
+/// - `name`: The operation or action being logged. This provides context on what the system is doing at the time the log is emitted (e.g., "sdk_initialization", "exporter_start").
+/// - `signal`: The type of telemetry data related to the log. This could be "trace", "metric", or "log", representing the OpenTelemetry signal.
+///
```
lalitb, line 8: The `target`, `name`, and `signal` fields are mandatory. While recommendations are provided for what these fields should contain, their values are not strictly enforced. Further discussion is needed to determine whether a more specific representation or stricter validation should be implemented for these fields.

opentelemetry/src/global/internal_logging.rs
```diff
@@ -0,0 +1,130 @@
...
+///     target: "opentelemetry-sdk",
+///     name: "sdk_initialization",
+///     signal: "trace",
+///     "Initializing the OpenTelemetry SDK for traces"
+/// );
+/// ```
+#[macro_export]
+macro_rules! otel_log {
+    (target: $target:expr, name: $name:expr, signal: $signal:expr, $($arg:tt)*) => {
+        #[cfg(all(feature = "experimental-internal-debugging"))]
```
lalitb, line 21: no associated cost if the feature flag is disabled, or if the level if disabled, or no subscriber is attached.

opentelemetry-appender-tracing/examples/basic.rs
```diff
@@ -15,9 +18,19 @@ fn main() {
         )]))
         .with_simple_exporter(exporter)
         .build();
-    let layer = layer::OpenTelemetryTracingBridge::new(&provider);
-    tracing_subscriber::registry().with(layer).init();
-
-    error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");
+    let env_filter = EnvFilter::from_default_env();
```
cijothomas, line 21: this example is to show the usage of appender-tracing, and we can leave it untouched. We need to fix https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/self-diagnostics to showcase this, if needed.
I am unsure if we need to show this with a full blown example, but no harm doing it in https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/self-diagnostics

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -49,6 +50,9 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
         attributes: Option<Vec<opentelemetry::KeyValue>>,
     ) -> Logger {
         let name = name.into();
+        otel_info!(target: "opentelemetry-sdk", name: "logger_versioned_creation", signal: "log",
```
cijothomas, line 53: target
I think the usage of component name for target looks good. Though we need to discuss further about making it even more specific, like opentelemetry-sdk.log.logemitter?

name
Looks good. We need to develop conventions. (like using "." vs "_" separator).

signal
I don't think this is required. Target+Name should be sufficient.

message
I think we should only promote key/values pairs, and leave the message out, unless there is a strong need.
Lets also use a named field as that is proved much faster : https://github.com/open-telemetry/opentelemetry-rust/pull/2001

opentelemetry/src/global/internal_logging.rs
```diff
@@ -0,0 +1,130 @@
+#![allow(unused_macros)]
+/// Macro for logging messages at the general log level in OpenTelemetry.
```
cijothomas, line 2: a warning at the top to indicate that this meant for user by OpenTelemetry components itself.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: So, 
target := <component/crate>.<signal>.<module/struct>
name := <operation-name> eg logger_creation
And any unstructured data (if required) as "message" attribute.

So for batch::emit method:

target := opentelemetry_sdk.logs.log_processor
name := batch_emit
event_count := 10 (say)

cijothomas: that looks very reasonable for first release.!

opentelemetry-appender-tracing/src/layer.rs
cijothomas: this part I am unsure. I don't think the layer should make this decision. and leave it upto users to apply filtering/layering correctly (we should show example of doing it correctly).
There are genuine cases for users to send Otel internal logs via OTel itself. (eg: Metrics cadinality limit hit, and droppign metrics. this is an valid log. and its valid to send it via OTel itself.)

opentelemetry/src/global/internal_logging.rs
```diff
@@ -0,0 +1,200 @@
+#![allow(unused_macros)]
```
shaun-cox, line 1: Why expose this from the `opentelemetry` (API) package if it is seemingly only going to be used by SDK or implementation specific crates in the project? Won't changes to this stuff necessitate a possible semver change in `opentelemetry` API crate itself, unnecessarily?

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -49,6 +50,14 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
         attributes: Option<Vec<opentelemetry::KeyValue>>,
     ) -> Logger {
         let name = name.into();
+        otel_info!(
+            target: otel_target!("logs"),
```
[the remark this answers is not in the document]
cijothomas: if we skip target, the default target might be good enough maybe?

opentelemetry-appender-tracing/src/layer.rs
lalitb: > There are genuine cases for users to send Otel internal logs via OTel itself.

Good point. Although, this would be tricky - as anything like below can cause the infinite loop if not intercepted:

If info/debug level is enabled:

otel_info("log", "log_processor_emit") -> otel-appender-tracing(OAT->  otel_info("log", "log_processor_emit") ->  OAT -> otel_info("log", "log_processor_emit") -> OAT -> otel_info("log", "log_processor_emit") -> .. and continues eventually crashing

If warn/error is enabled (as default):

`otel_warn("metrics", "cardinality_limit_hit") -> otel-appender-tracing(OAT) -> otel_warn("logs", "batch-already-shutdown") -> OAT -> otel_warn("logs", "batch-already-shutdown") -> .... and continues eventually crashing.

Probably, we should send it once, and then start intercepting as done [here](https://github.com/open-telemetry/opentelemetry-rust/blob/d652dc82fcb828909877c3492f85d3a8206efd6e/examples/self-diagnostics/src/main.rs#L21), or maintain some state in thread-local storage.

opentelemetry-sdk/src/logs/log_emitter.rs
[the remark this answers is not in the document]
lalitb: The otel_target macro expands to `crate.module.<signal>` where, signal is passed as argument. In this case it is `log`. @cijothomas do you men we don't need any of these, and just operation_name is enough?

opentelemetry/src/global/internal_logging.rs
lalitb: I kept it at the level of custom error handler which is within API. I thought there could be the cases where we may want to used within API e.g., debugging `Context` implementation.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: That should be [solved irrespective](https://github.com/open-telemetry/opentelemetry-rust/issues/761) of logs, using Context, right?

opentelemetry/src/global/internal_logging.rs
cijothomas: i am not sure if there are better ways to "hide" it. It should be public API, so it follows the same sem ver rules, and it can be used by OTel itself + any other components extending otel. For example, if someone authors an exporter, they can also use this. 
We can doc it so user end up misusing it for any other logging!

> Won't changes to this stuff necessitate a possible semver change in opentelemetry API crate itself, unnecessarily?
We should treat this API as just any other API, so it cannot make breaking changes. As long as we are doing a minimal API (super thin wrapper around tracing), we should be able to stabilize it by 1.0, and keep it non-breaking forever (or until 2.0 ships)

opentelemetry-appender-tracing/Cargo.toml
cijothomas, line 15:
```suggestion
opentelemetry = { version = "0.25", path = "../opentelemetry", features = ["logs","experimental-internal-logs"] }
```

opentelemetry-sdk/src/logs/log_emitter.rs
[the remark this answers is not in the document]
TommyCpp: Or can we can try procedural macros if we want more control over the function & types

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -83,19 +97,37 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
...
+            name: "logger_provider_resource",
+            resource_count= self.inner.resource.len()
+        );
         &self.inner.resource
     }
 
     /// Force flush all remaining logs in log processors and return results.
     pub fn force_flush(&self) -> Vec<LogResult<()>> {
+        otel_info!(
```
TommyCpp, line 127: Hmm how do we decide when to use `info` vs `debug` here?

opentelemetry/src/global/internal_logging.rs
```diff
@@ -0,0 +1,200 @@
...
+///     target: "opentelemetry-otlp",
+///     name: "export_warning",
+///     error_code = 404,
+///     version = "1.0.0"
+/// );
+/// ```
+#[macro_export]
+macro_rules! otel_warn {
+    // Without extra attributes
+    (target: $target:expr, name: $name:expr) => {
```
TommyCpp, line 102: Can we default target to current package if not provided?

cijothomas:
As per discussion in the Community meeting today:
1. It is better to expose an API from `opentelemetry` crate, as that is the common crate everyone depends on
2. This API closely mimics `tracing` macros for 5 log levels - error,warning,info,debug,trace
3. We can remove feature flag, and do this by default.
4. Remove the existing global error handler and replace with `tracing`
5. In the future, we can offer alternates - eg: use a feature flag to use `log` or `another-logging-crate`, or even user-provided callback.

@lalitb please add if my missed anything. Let's use this PR itself to sort this out.

examples/self-diagnostics/src/main.rs
```diff
@@ -33,6 +33,7 @@ impl ErrorState {
 }
 
 static GLOBAL_ERROR_STATE: Lazy<Arc<ErrorState>> = Lazy::new(|| Arc::new(ErrorState::new()));
+static SEEN_EVENT_NAMES: Lazy<Mutex<HashSet<String>>> = Lazy::new(|| Mutex::new(HashSet::new()));
```
cijothomas, line 36: I'll cleanup this example in a separate PR.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: This is my take -  A good practice would be to use `otel_debug` for more granular logging, especially in hot-path scenarios. Meanwhile, `otel_info` can be reserved for higher-level events like SDK initialization, shutdown, and actions such as `force-flush`. For emitting logs, traces, and metrics, `otel_debug` is appropriate.

opentelemetry/src/global/internal_logging.rs
lalitb: @shaun-cox The current macros are lightweight and closely resemble the tracing macros. Since not all opentelemetry-* crates depend on opentelemetry-sdk, exposing these macros from the API package appears to be the most suitable approach. 

Let me know if you see any concerns here.

lalitb:
@cijothomas 

- It is better to expose an API from opentelemetry crate, as that is the common crate everyone depends on: Done
- This API closely mimics tracing macros for 5 log levels - error,warning,info,debug,trace: Do we need trace, or debug and info will suffice ? As of now we have - error, warning, info, debug.
- We can remove feature flag, and do this by default.  - Done
- Remove the existing global error handler and replace with tracing - I prefer to do this as separate PR, to keep this PR scope limited.

opentelemetry-sdk/Cargo.toml
```diff
@@ -28,7 +28,7 @@ url = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["rt", "time"], optional = true }
 tokio-stream = { workspace = true, optional = true }
 http = { workspace = true, optional = true }
-tracing = {workspace = true, optional = true}
+tracing = {workspace = true}
```
cijothomas, line 31: @lalitb Looking at this again, I think we should be do a feature-flag, and make it part of default. If someone really wants to avoid `tracing` dependency, they can do so by removing feature flags.
(Tracing causes assembly size increase due to statics, so the more callsites we have, the more the size.. some people may avoid to avoid that...)
Thoughts?
ff can be "internal-logs" (without the experimental)

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -234,7 +235,11 @@ mod tests {
             .with_simple_exporter(exporter.clone())
             .build();
 
-        let layer = layer::OpenTelemetryTracingBridge::new(&logger_provider);
+        let non_opentelemetry_filter = tracing_subscriber::filter::filter_fn(|metadata| {
+            !metadata.target().starts_with("opentelemetry")
+        }); // filter internal logs
+        let layer = layer::OpenTelemetryTracingBridge::new(&logger_provider)
```
cijothomas, line 241: nit: lets move this to a common helper, to avoid the repetition

opentelemetry-sdk/Cargo.toml
lalitb: This is done.

lalitb: Just realized that this will also make the errors and warnings optional depending on the flag. Will update the macros to dump them using `eprintln` if the flag is disabled.

cijothomas (commented):
The changes look good! Thank you!.

I'd suggest to removing adding logs throughout into its own separate PR. We should be doing it very carefully (irrespective of the level), so its best to start with adding error level ones for the failure scenarios (that can be used to remove the current global handler), and anything else on a case-by-case basis, few at a time. (I am trying to be conservative here, as using `tracing` is still a very new thing, so lets start small)

lalitb:
> I'd suggest to removing adding logs throughout into its own separate PR. 

Yes, this PR doesn't add any new logs. Just updates the logs which were added by the previous PR directly using tracing.

opentelemetry-sdk/Cargo.toml
lalitb: Have added eprintln for errors and warning if the flag is disabled. Please have a look, if there is any concern.

cijothomas: I think its best to do that along with removal of current global error handler.

lalitb: Make sense. done.

opentelemetry-appender-tracing/src/layer.rs
lalitb: This is done now.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -83,19 +93,33 @@ impl opentelemetry::logs::LoggerProvider for LoggerProvider {
...
     /// Create a new `LoggerProvider` builder.
     pub fn builder() -> Builder {
+        otel_info!(
+            name: "logger_provider_builder"
+        );
         Builder::default()
     }
 
     pub(crate) fn log_processors(&self) -> &[Box<dyn LogProcessor>] {
+        otel_debug!(
```
utpilla, line 103: It should be good enough to get the processor count information just once during LoggerProvider build. Is there any value in repeatedly logging it?

cijothomas: to unblock and merge this PR quickly, I think its best to remove all internal logging that are not error/warning, and we can selective bring them one by one in future PRs.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -104,6 +128,9 @@ impl LoggerProvider {
 
     /// Shuts down this `LoggerProvider`
     pub fn shutdown(&self) -> LogResult<()> {
+        otel_info!(
```
utpilla, line 131: This should be logged only in case of a successful shutdown? Only in this case:
```rust
if errs.is_empty() {
    Ok(())
}
```

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -221,6 +260,9 @@ impl Logger {
         instrumentation_lib: Arc<InstrumentationLibrary>,
         provider: LoggerProvider,
     ) -> Self {
+        otel_info!(
+            name: "logger_new"
```
utpilla, line 264: We should probably add some identification information for the logger such as name, version etc.

lalitb:
All the debug/info has been removed now.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -98,16 +99,12 @@ impl LogProcessor for SimpleLogProcessor {
     fn emit(&self, record: &mut LogRecord, instrumentation: &InstrumentationLibrary) {
         // noop after shutdown
         if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
+            otel_warn!(
```
utpilla, line 102: This could continue to be the hot path even after shutdown so it might not be a great idea to keep logging warnings for this. Let's a TODO to revisit, if we should remove the logging here or make it less severe.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -147,6 +147,9 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
             global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
+            otel_warn!( name: "ValueMap.measure",
+                message = "Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."
```
utpilla, line 151: nit: We can remove the word "Warning" as the the log category is a warning anyway.

```suggestion
                message = "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."
```

utpilla (approved):

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -234,8 +246,7 @@ mod tests {
             .with_simple_exporter(exporter.clone())
             .build();
 
-        let layer = layer::OpenTelemetryTracingBridge::new(&logger_provider);
-        let subscriber = tracing_subscriber::registry().with(layer);
+        let subscriber = create_tracing_subscriber(exporter.clone(), &logger_provider);
```
cijothomas, line 249: why is exporter required to be passed?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -513,7 +521,7 @@ mod tests {
             let span_id = cx.span().span_context().span_id();
 
             // logging is done inside span context.
-            log::error!("log from log crate");
+            log::error!(target: "my-system", "log from log crate");
```
cijothomas, line 524: do we need these changes?

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -147,6 +147,9 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
             global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
+            otel_warn!( name: "ValueMap.measure",
```
cijothomas, line 150: These are user actionable logs, so it should use a end user friendly name like "MetricCarinalityLimitHit"/similar,. 
Good to follow up separately on all logs.

cijothomas (approved):

opentelemetry-appender-tracing/src/layer.rs
lalitb: Yes we can remove the target, will follow up in subsequent cleanups.

opentelemetry-appender-tracing/src/layer.rs
lalitb: Forgot to remove as part of updating the method. Will cleanup in subsequent PR.

cschramm:
I observed obscure build issues with opentelemetry_sdk 0.27.0 and traced them down to this code, which requires tracing 0.1.39 while the workspace specifies 0.1.0 as the minimum version. While that is unfortunate, at least the dependency should reflect it.

lalitb:
> I observed obscure build issues with opentelemetry_sdk 0.27.0 and traced them down to this code, which requires tracing 0.1.39 while the workspace specifies 0.1.0 as the minimum version. While that is unfortunate, at least the dependency should reflect it.

Can you provide more details, where do you see 0.1.39 dependency added?

cschramm:
Sure. The code is using the name syntax of the tracing macros which was added in https://github.com/tokio-rs/tracing/pull/2699 and released in [0.1.39](https://github.com/tokio-rs/tracing/releases/tag/tracing-0.1.39). Thus, when you `cargo update -p tracing --precise 0.1.38` (or to 0.1.37 if you like, as 0.1.38 got yanked), `cargo check -p opentelemetry_sdk` fails.

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/issues/2373 Opening a new issue to track and address this.

PR #2130, opened 2024-09-19
Update Histogram Builder

utpilla:
## Changes
- Introduce a new builder for Histogram. Histograms have unique compared to other instruments so having its own builder type would help adding Histogram-specific configuration such as custom bounds
- This builder struct has been marked `non_exhaustive` to allow for adding more configuration in the future.
- Update `InstrumentProvider` trait methods for creating histograms to only take the builder as the input. This reduces the number of public API changes when adding newer configuration. If this approach seems better then we can implement it for other instruments as well.

### DIFF
--- a/opentelemetry/src/metrics/instruments/mod.rs
+++ b/opentelemetry/src/metrics/instruments/mod.rs
@@ -6,7 +6,7 @@ use std::borrow::Cow;
 use std::marker;
 use std::sync::Arc;
 
-use super::InstrumentProvider;
+use super::{Histogram, InstrumentProvider};
 
 pub(super) mod counter;
 pub(super) mod gauge;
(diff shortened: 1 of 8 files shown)


opentelemetry/src/metrics/instruments/mod.rs
```diff
@@ -24,6 +24,94 @@ pub trait AsyncInstrument<T>: Send + Sync {
     fn as_any(&self) -> Arc<dyn Any>;
 }
 
+/// Configuration for building a Histogram.
+#[non_exhaustive]
```
cijothomas, line 28: nit: add a comment on why this is relevant.

opentelemetry/src/metrics/mod.rs
```diff
@@ -273,22 +273,12 @@ pub trait InstrumentProvider {
     }
 
     /// creates an instrument for recording a distribution of values.
-    fn f64_histogram(
-        &self,
-        _name: Cow<'static, str>,
-        _description: Option<Cow<'static, str>>,
-        _unit: Option<Cow<'static, str>>,
-    ) -> Result<Histogram<f64>> {
+    fn f64_histogram(&self, _builder: HistogramBuilder<'_, f64>) -> Result<Histogram<f64>> {
```
cijothomas, line 276: this is crucial, and allow the future extensibility we are after.
Agree with this, lets follow up next, by leveraging this to add histogram custom bound support.

cijothomas (approved):
LGTM. We can revisit this in the near future, after adding similar for Counters, and then trying this to support Hints/InstrumentAdvisory.

PR #2133, opened 2024-09-20
PeriodicReader::build make interval registration in same thread/task

fraillt:
## Changes

Previous behavior was that in `PeriodicReader::build` interval was registered before spawning a task. Which means that we effectively need already running runtime in order to register an `interval`.

This revision moves `interval` registration within `spawn` function, which allow to register `PeriodicReader` independently of any runtime.

You can see in test `collection_triggered_by_interval`, that this works perfectly fine without any runtime when using `TokioCurrentThread`.


cijothomas (approved):
LGTM!

PR #2138, opened 2024-09-21
[Public API cleanup] Remove unnecessary conversion methods for builder and instruments

utpilla:
## Changes
- Remove `TryFrom` implementations for instruments which are there only to call the required method on `InstrumentProvider`.
- This logic can be kept internal by providing dedicated implementations of `try_init` and `init` for each instrument type
- I have used macros to do that

On the main branch:
```rust
let counter = meter.u64_counter("my_counter").init();

// Be keeping the `TryFrom` implementations, we also allow a user to do this.
let counter: Counter<u64> = meter.u64_counter("my_counter").try_into().unwrap();
```

This PR would ensure that an instrument can be built only by using the recommended `try_init` or `init` methods.


opentelemetry/src/metrics/instruments/mod.rs
```diff
@@ -162,25 +164,39 @@ where
...
-    ///
-    /// Panics if the instrument cannot be created. Use
-    /// [`try_init`](InstrumentBuilder::try_init) if you want to handle errors.
-    pub fn init(self) -> T {
-        T::try_from(self).unwrap()
-    }
+macro_rules! build_instrument {
+    ($name:ident, $inst:ty) => {
+        impl<'a> InstrumentBuilder<'a, $inst> {
+            /// Validate the instrument configuration and creates a new instrument.
```
TommyCpp, line 172:
```suggestion
             #[doc = concat!("Validate the instrument configuration and creates a new `",  stringify!($inst), "`.")]
```

This will include the instrument name in documentation. `/// ...` is just syntax sugar for `#[doc]`

utpilla: @TommyCpp I tried this but it didn't work. I expected it to show up on the tool tip on VSCode but it didn't show anything at all.

TommyCpp: Screenshot from 2024-09-25 23-38-33

I tried it on my localhost. The `cargo doc` generates the expected value. Not sure if it's because VS code didn't expand the macros correctly?

I also can see the doc on my RustRover. 
Screenshot from 2024-09-25 23-41-56

utpilla: @TommyCpp I have made the changes.

cijothomas (approved):

cijothomas:
@TommyCpp Please let us know if there are other feedbacks, we can address them in follow ups. 
Merging now to unblock progress.

Issue #2056, opened 2024-08-26
[Bug]: `PeriodicReader::shutdown()` deadlocks on current thread runtime.


inahga:
### What happened?

When using the `SdkMeterProvider` with a `PeriodicReader` on a tokio `current_thread` runtime, the program hangs when the `SdkMeterProvider` leaves scope.

Example reproducer: Take the `registration_triggers_collection()` test function in `opentelemetry-sdk/src/metrics/periodic_reader.rs`, and run it on a `current_thread` runtime:

<details>

```rust
    #[tokio::test]
    async fn registration_triggers_collection() {
        // Arrange
        let interval = std::time::Duration::from_millis(1);
        let exporter = InMemoryMetricsExporter::default();
        let reader = PeriodicReader::builder(exporter.clone(), runtime::Tokio)
            .with_interval(interval)
            .build();
        let (sender, receiver) = mpsc::channel();

        // Act
        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
        let meter = meter_provider.meter("test");
        let _counter = meter
            .u64_observable_counter("testcounter")
            .with_callback(move |_| {
                sender.send(()).expect("channel should still be open");
            })
            .init();

        _ = meter_provider.force_flush();

        // Assert
        receiver
            .try_recv()
            .expect("message should be available in channel, indicating a collection occurred");
    }
```

</details>

The problematic code looks to be in `PeriodicReader::shutdown()`, where we send a shutdown signal and wait for receipt on a oneshot channel in a `block_on` statement. When executed in the context of a `Drop`, there is no runtime to advance the other side of the channel, so we hang indefinitely.

It's not a big deal, it can be worked around by using a multi-thread runtime, or by calling `shutdown()` ahead of time. It's just unexpected and easy to trip up on, since tokio tests default to a current thread runtime.

More than happy to fix it myself, but wanted to get maintainer input first. (It may be the right answer is to not use a current thread runtime?)

### API Version

0.24.0

### SDK Version

0.24.1

### What Exporter(s) are you seeing the problem on?

N/A

cijothomas:
This could be some design bug in the metric aggregation. [Related to BatchProcessing as well](https://github.com/open-telemetry/opentelemetry-rust/issues/1968), as that is another place where OTel does background processing.

I don't have a good answer at the moment, but you can expect a solution soon. We just finished redesigning the metrics hot path (the aggregation path), and the next major item left is to revisit the periodicreader part.

[...]

Issue #1949, opened 2024-07-20
[Feature]: Name spawned threads


howardjohn:
### Describe the solution you'd like:

The `TokioCurrentThread` spawns threads without names. This makes debugging hard. For example:
```
thread '<unnamed>' panicked at /home/john/.rhack/tokio-1.37.0/src/runtime/time/entry.rs:557:9:
A Tokio 1.x context was found, but it is being shutdown.
```

is quite hard to trace back to otel.

I would like each thread spawned by this library to be [named](https://doc.rust-lang.org/std/thread/#configuring-threads)

### Considered Alternatives

Use the multithreaded tokio runtime which does not spawn. This is currently causing deadlocks (https://github.com/open-telemetry/opentelemetry-rust/issues/1395), so its not acceptable.

howardjohn:
BTW, I don't get `A Tokio 1.x context was found, but it is being shutdown.` after upgrading -- was just an example of a type of issue that would have been easier to debug if the thread was named.

lalitb:
@howardjohn If you'd like to contribute the fix, please feel free to proceed. We eventually plan to eliminate the dependency on the async runtime for batch processors. I'm currently working on this for logs (#1943), and a similar approach would be replicated to traces as well.

howardjohn:
BTW turns out the error was actually still happening, opened https://github.com/open-telemetry/opentelemetry-rust/issues/1963. I missed it due to not setting the global trace provider after upgrading :facepalm:

cijothomas:
Closing as thread created by PeriodicReader, BatchProcessor uses named thread.

Issue #1507, opened 2024-02-01
[Bug]: PeriodicReader uses sync::Mutex.


bIgBV:
### What happened?

The `PeriodicReader` internally uses a [`sync::Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html) to synchronize on the `PeriodicReaderInner` type:

https://github.com/open-telemetry/opentelemetry-rust/blob/d19187db46ec445143ecdc0271794f71dce3055d/opentelemetry-sdk/src/metrics/periodic_reader.rs#L201C1-L205C2

This can possibly cause deadlocks if the mutex is held across an `.await` point(https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use). And it looks like there is at least one such case of a user reporting this on the [#otel-rust](https://cloud-native.slack.com/archives/C03GDP0H023/p1706210769680649) slack channel. 

### API Version

0.21

### SDK Version

0.21.2

### What Exporters are you seeing the problem on?

N/A

### Relevant log output

```shell
$ RUST_LOG=debug cargo run -- --poll-interval 30s
[2024-01-25T19:25:41Z INFO  nomad_otel_metrics_scraper] Polling http://localhost:4646/ every 30s
{
  "resourceMetrics": {
    "resource": {
      "attributes": [
        {
          "key": "telemetry.sdk.version",
          "value": {
            "stringValue": "0.21.2"
          }
        },
        {
          "key": "telemetry.sdk.language",
          "value": {
            "stringValue": "rust"
          }
        },
        {
          "key": "telemetry.sdk.name",
          "value": {
            "stringValue": "opentelemetry"
          }
        },
        {
          "key": "service.name",
          "value": {
            "stringValue": "unknown_service"
          }
        }
      ]
    },
    "scopeMetrics": []
  }
}
^C[2024-01-25T19:25:43Z INFO  nomad_otel_metrics_scraper] Provider, as we know it. MeterProvider {
        pipes: Pipelines(
            [
                Pipeline,
            ],
        ),
        meters: Mutex {
            data: {
                InstrumentationLibrary {
                    name: "nomad_metrics",
                    version: None,
                    schema_url: None,
                    attributes: [],
                }: Meter {
                    scope: InstrumentationLibrary {
                        name: "nomad_metrics",
                        version: None,
                        schema_url: None,
                        attributes: [],
                    },
                },
            },
            poisoned: false,
            ..
[...]

PR #2142, opened 2024-09-24
Use dedicated thread for Metrics PeriodicReader

cijothomas:
1. This eliminates the need of async runtime, and avoids all the bugs associated with (in)correct usage of async runtimes and other bugs like https://github.com/open-telemetry/opentelemetry-rust/issues/1507 and https://github.com/open-telemetry/opentelemetry-rust/issues/2056
Given the usage of dedicated thread, PeriodicReader is no longer at the mercy of user threads not blocking threads and causing issues like [this](https://github.com/open-telemetry/opentelemetry-rust/discussions/2058#discussioncomment-10629598).
2. Callbacks for Observable instruments can now make async calls with help of futures executor/others without deadlocking, as the callbacks are now executed from a normal thread, and avoids issues like https://github.com/open-telemetry/opentelemetry-rust/discussions/2063
3. This also uses named thread, addressing https://github.com/open-telemetry/opentelemetry-rust/issues/1949
4. User thread is never blocked in this. Blocking calls, if any, are only from OTel's own background thread.
5. Flush - is designed to be blocking the calling thread - this should not be a concern as Flush() is rarely required.
6. Shutdown - This PR keeps in sync, but will offer an async version of it, so users can call shutdown and await without blocking thread. This can be done without dependency on async-runtimes in the OTel SDK.
7. This is a working prototype, and I expect to iterate over based on feedback. Have used internal logs liberally for now. This will be revisited to keep it to essentials only before stabilizing.
8. A notable limitation as-is is the lack of time out enforcement at PeriodicReader, as it is now delegated to exporters. A future PR will modify the exporters to actually enforce this, to keep the scope in control for this PR. Or explore how to enforce timeouts from PeriodicReader itself.

(The ability for users to bring-own-async-runtime - this can be offered as an opt-in feature. How exactly - TBD. Most likely we need to do it post 1.0)

Note: The key change is in periodicreader.rs only. Rest are cascading effects on removing runtime requirement, passing timeout to exporters. The non-relevant changes to examples are to be removed before PR merge - it is added to help anyone run locally and observe the logs themselves.


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -187,17 +147,116 @@ where
 #[derive(Clone)]
 pub struct PeriodicReader {
     exporter: Arc<dyn PushMetricsExporter>,
-    inner: Arc<Mutex<PeriodicReaderInner>>,
```
cijothomas, line 190: Mutex requirement was removed and this eliminates a lot of potential deadlocks.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -187,17 +147,116 @@ where
...
+            exporter: Arc::new(exporter),
+            inner: Arc::new(PeriodicReaderInner {
+                message_sender,
+                is_shutdown: AtomicBool::new(false),
+                producer: Mutex::new(None),
+            }),
+        };
+        let cloned_reader = reader.clone();
+
+        thread::spawn(move || {
```
cijothomas, line 178: this thread is the meat of the change.

demurgos:
I'm worried about support for situation where threads may be problematic, either because their are not supported or because or due to the use of functions unsafe to call in the presence of threads. For example, what about `WebAssembly`?

Would it be possible to have thread not required and instead be another runtime to chose from? Or exposing a lower level interface based on channels and let the use chose its concurrency solution?

cijothomas:
> I'm worried about support for situations where threads may be problematic, either because their are not supported or because or due to the use of functions unsafe to call in the presence of threads (non-reentrant function calls). For example, what about `WebAssembly`?
> 
> Would it be possible to have thread not required and instead be another runtime to chose from? Or exposing a lower level interface based on channels and let the use chose its concurrency solution?

Yes that is very valid point, which is already mentioned in the desc:
_The ability for users to bring-own-async-runtime - this can be offered as an opt-in feature. How exactly - TBD. Most likely we need to do it post 1.0)_

Yes we expect there'd be scenarios where OTel spawning background threads is not feasible, and OTel will offer a way to "bring your own runtime".

cijothomas:
@sandersaares Could you take a look?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -187,17 +147,116 @@ where
...
+            let mut interval_start = Instant::now();
+            let mut remaining_interval = interval;
+            println!("PeriodicReader Thread started.");
+            loop {
+                match message_receiver.recv_timeout(remaining_interval) {
+                    Ok(Message::Flush(response_sender)) => {
+                        println!("Performing ad-hoc export due to Flush.");
+                        if let Err(_e) = cloned_reader.collect_and_export() {
+                            response_sender.send(false).unwrap();
```
sandersaares, line 187: I am guessing the background worker here probably should not panic if whoever sent it the flush message is no longer there to receive the response - without awareness of any big picture design, my first instinct is to ignore "nobody listening for me anymore" style errors. Perhaps in current code paths this should never happen (did not check all places) but I can easily imagine "don't bother waiting for the flush to complete" a reasonable thing to do at times.

sandersaares (commented):
Basic premise seems workable. I suspect my lack of OTel familiarity prevents me from making more useful comments here.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -187,17 +147,116 @@ where
...
+
+        thread::spawn(move || {
+            let mut interval_start = Instant::now();
+            let mut remaining_interval = interval;
+            println!("PeriodicReader Thread started.");
+            loop {
+                match message_receiver.recv_timeout(remaining_interval) {
+                    Ok(Message::Flush(response_sender)) => {
+                        println!("Performing ad-hoc export due to Flush.");
+                        if let Err(_e) = cloned_reader.collect_and_export() {
```
TommyCpp, line 186: This can just be `is_err` function

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -187,17 +147,116 @@ where
...
+    }
+
+    fn collect_and_export(&self) -> Result<()> {
+        let mut rm = ResourceMetrics {
+            resource: Resource::empty(),
+            scope_metrics: Vec::new(),
+        };
+        self.collect(&mut rm)?;
+
+        // TODO: Enforce timeout
```
TommyCpp, line 257: +1, since the exporter may not be SDK exporter we need to guard against the possiblity it's deadlock

opentelemetry-sdk/src/metrics/periodic_reader.rs
TommyCpp: Just to double check this is not related to async -> background change right? Just an addon optimzation?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -187,17 +147,116 @@ where
...
+                        }
+                    }
+                    Ok(Message::Shutdown(response_sender)) => {
+                        // Perform final export and break out of loop and exit the thread
+                        println!("Performing final export and shutting down.");
+                        if let Err(_e) = cloned_reader.collect_and_export() {
+                            response_sender.send(false).unwrap();
+                        } else {
+                            response_sender.send(true).unwrap();
+                        }
```
TommyCpp, line 214:
```suggestion   
response_sender.send(cloned_reader.collect_and_export().is_err())
```

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: I am not sure if there is any way to enforce a timeout with just futures_executor...

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: it is a required fix this to work, else flush/shutdown cannot operate without deadlocking. But I can try to make that change to own PR.

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: i don't plan to keep any on the unwraps. Everything would be handled and logged.

opentelemetry-sdk/src/metrics/periodic_reader.rs
TommyCpp: This is a basically a blocking call
```
futures_executor::block_on(self.exporter.export(&mut rm))?;
```

So in order to interrupte it to have a timeout we probably need a separate thread to act as a ticker and do a `select` between those two. Because the background thread will be blocked on exporting

TommyCpp: Not saying we need to handle timeout here. Just we need to guard against a high latency or dead lock exporter

cijothomas: yes those require another thread. Have to explore further to find an acceptable solution.

lalitb: If we agree to take dependency on [futures_timer](https://crates.io/crates/futures-timer), this seems to be working as I tested (courtesy chatgpt):
```rust
    let export_future = Box::pin(exporter.export()).fuse();
    let timeout_future = Box::pin(Delay::new(Duration::from_secs(5)).map(|_| Err::<&'static str, &'static str>("Export timed out"))).fuse();
    let combined_future = future::select(export_future, timeout_future);
    let result = block_on(combined_future);
    match result {
        future::Either::Left((Ok(message), _)) => Ok(message),           // Export succeeded
        future::Either::Left((Err(e), _)) => Err(e),                    // Export failed
        future::Either::Right((Err(e), _)) => Err(e),                   // Timeout occurred
        _ => Err("Unknown error"),                                      // Fallback for other outcomes
    }
```

cijothomas: if we can depend on external crate, we can solve it via async-std or tokio!

lalitb: Yes, but timer utilities in async-std or tokio would need the async runtime to be already running. `futures_timer` doesn't need that.

lalitb: Oh, but I think `futures_timer` internally may be spawning a background thread, which will defeat the purpose.

cijothomas: Updated based on some offline conversations:
1. @ThomsonTan explored how to achieve this by doing manual polling of the future. This does work, but likely requires us to mimic functionality typically done inside executors! The basic idea of looping with a small sleep, polling for the future can work, but we'll likely need to leverage a waker and park the thread etc..
2. Alternative option is to make this a problem for the exporter to solve. The periodic reader simply passes the timeout (whatever is remaining after doing collect etc.), to the exporter. The exporter which deals with async io/network call, can use the timeout when it makes the calls. This idea works only for well behaved exporters. 

If 1 turns out to be too complex to replicate in a telemetry library, and we still want to avoid extra dependencies, then 2 seems the viable alternate. (Based on my quick check, OTel .NET does the same - it relies on exporter to respect the timeout)

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: It is infeasible to make this fix separate.

.github/workflows/ci.yml
```diff
@@ -106,6 +106,7 @@ jobs:
         os: [windows-latest, ubuntu-latest]
         rust: [1.65.0, 1.71.1]
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
```
ThomsonTan, line 109: Is this from some separated issue?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -95,8 +87,9 @@ where
         self
     }
 
-    /// Configures the time a [PeriodicReader] waits for an export to complete
-    /// before canceling it.
+    /// Configures the timeout for an export to complete. PeriodicReader itself
+    /// does not enforce timeouts. Instead timeout is passed on to the exporter
```
ThomsonTan, line 91: nit: `enforce timeout`, or there could be multiple timeout values?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -112,93 +105,194 @@ where
...
 }
 
 /// A [MetricReader] that continuously collects and exports metric data at a set
 /// interval.
 ///
-/// By default it will collect and export data every 60 seconds, and will cancel
-/// export attempts that exceed 30 seconds. The export time is not counted
-/// towards the interval between attempts.
+/// By default, PeriodicReader will collect and export data every
+/// 60 seconds. The export time is not counted towards the interval between
```
ThomsonTan, line 116: "The export time is not counted towards the interval between attempts", does this mean if the export is too slow like exeeding 60 seconds, multiple export then can to be outstanding(pending), and new export will still issued after another 60 seconds?

cijothomas: I have left a code comment to explain the behavior when export itself takes more than the interval. This PR's behavior is to trigger the next one immediately and then continue to honor the internal for the next one onwards.

cijothomas: @ThomsonTan any additional comments on this?

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: fixed.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -61,7 +61,7 @@ fn init_tracer_provider() -> Result<sdktrace::TracerProvider, TraceError> {
 
 fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, MetricsError> {
     opentelemetry_otlp::new_pipeline()
-        .metrics(opentelemetry_sdk::runtime::Tokio)
+        .metrics()
```
utpilla, line 64: Have you verified that this still works? I believe `reqwest` relies on Tokio. Now that the outgoing Http call through `reqwest` is being made on a background thread without Tokio (or any runtime), it might be problematic. Looking to ensure that we don't hit this issue in particular:
```
there is no reactor running, must be called from the context of Tokio runtime
```

Related to #248, #2137

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -209,271 +303,502 @@ impl fmt::Debug for PeriodicReader {
...
-
-    async fn run(mut self, mut messages: impl FusedStream<Item = Message> + Unpin) {
-        while let Some(message) = messages.next().await {
-            if !self.process_message(message).await {
-                break;
+        if let Ok(response) = response_rx.recv() {
+            self.is_shutdown
+                .store(true, std::sync::atomic::Ordering::Relaxed);
+            if response {
+                Ok(())
```
utpilla, line 440: Join the background thread before considering it a successful shutdown?

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
cijothomas: Excellent point!
Based on offline discussion, yes this appear to cause issues when using both `reqwest` and `hyper`. These libraries seem to have a strong requirement that it cannot work unless they are inside tokio runtime.
`tonic` seem to work fine without issues.

Will check if there are ways to work around the http library limitations.

lalitb: > Will check if there are ways to work around the http library limitations.

I did a quick test with SimpleLogExporter + OTLP with request::blocking::Client - this works without need of tokio runtime. This should also work with background thread for batch in that case. Enforcing this for background thread could be one option.

cijothomas: > > Will check if there are ways to work around the http library limitations.
> 
> I did a quick test with Simple Log Processor + OTLP exporter with request::blocking::Client - this works without need of tokio runtime. This should also work with background thread for batch in that case. Enforcing this for background thread could be one option.

Yes. But there is no blocking version for libraries like tonic, effectively limiting exporting to only support http with reqwest::blockingClient.
Will need to redesign after doing a comparison of all feasible options.

cijothomas: > tonic seem to work fine without issues.

This maybe incorrect, as it may have worked during shutdown only. Apologies for the confusion!

lalitb: Yes, `tonic` is tricky, there is no alternative either.

cijothomas: @lalitb I have another correction to make about tonic, based on more testing:
The changes in this PR works fine with tonic, as long as the main function of the app is a tokio one. Yes the tonic::export call is made from our background thread, still this works.
If the main function is not a tokio main, then the app panics at meterprovider build itself. It looks like we attempt to create a grpc channel at build(), and that fails due to lack of tokio runtime.

opentelemetry-sdk/src/metrics/periodic_reader.rs
ThomsonTan: Join the background thread could block the user thread if the background thread is some how blocking (like blocking in the export call), so the timeout parameter could be ignored.

Could we try to join it if it can be done within the specified timeout, and give up the joining operation once timeout is reached?

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
lalitb: > The changes in this PR works fine with tonic, as long as the main function of the app is a tokio one. Yes the tonic::export call is made from our background thread, still this works.

Interesting, I was thinking the spawned thread (std::thread::spawn) doesn't run in the tokio runtime and so will not have access to tokio runtime. thanks for confirming. 

Trying to summarize the scenarios:

 These work:

```
 tokio::main -> background-thread -> gRPC (tonic), HTTP (hyper, reqwest, reqwest-blocking) 
 tokio::main -> simple-exporter -> gRPC (tonic), HTTP (hyper, reqwest, reqwest-blocking)  (assuming we do filtering to avoid infinite loop).
 tokio::main -> simple-exporter -> HTTP (reqwest-blocking) 
 main -> simple-exporter -> HTTP (reqwest-blocking)
```

And this doesn't work:
```
 main -> background-thread -> gRPC (tonic), HTTP (hyper, reqwest, reqwest-blocking) 
 tokio::main(current_thread) -> simple-exporter -> gRPC (tonic), HTTP (hyper, reqwest, reqwest-blocking) # hangs

```

how about this ?

```
tokio::main(current_thread) -> background-thread -> gRPC (tonic), HTTP (hyper, reqwest, reqwest-blocking) 
```

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: @ThomsonTan Such a situation will not arise in current implementation. If `response_rx` got the message back, the background thread will be exiting itself, and not doing anything else.

cijothomas:
This is now merged via alternate PR, Closing. Thanks everyone for thorough reviews!

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
maxwellflitton: How is this now implemented? I'm trying to get a batch exporter working with `reqwest` JSON and I'm running in Tokio but I get the error for not finding a reactor. I guess I just have to move it to a blocking `reqwest` client?

PR #2143, opened 2024-09-24
feature: refactor and extend `log_exporter` benchmark

demurgos:
Design discussion issue: #2031

## Changes

This commit refactors and extends the `log_exporter` benchmark.
1. Rename to be more precise: `FooWithFuture` becomes `FooAsyncTrait`, `FooWithoutFuture` becomes `FooSync`
2. Add variant using native Rust support for async traits
3. Share workload code between implementations
4. Use blackbox hint for implementations under test
5. Expand module documentation
6. Move each impl into its own submodule

* [x] Unit tests added/updated (if applicable): N/A
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes: N/A
* [x] Changes in public API reviewed (if applicable): N/A


cijothomas:
@demurgos Thanks in advance for your patience! I am clearing my backlog and will get to this soon!

opentelemetry-sdk/benches/log_exporter.rs
```diff
@@ -25,147 +31,232 @@ use opentelemetry_sdk::logs::LogRecord;
...
-impl LogProcessor for ExportingProcessorWithFuture {
-    fn emit(&self, record: &mut LogRecord, library: &InstrumentationLibrary) {
-        let mut exporter = self.exporter.lock().expect("lock error");
-        let logs = [(record as &LogRecord, library)];
-        futures_executor::block_on(exporter.export(LogBatch::new(&logs)));
+mod style_async_native {
+    use super::*;
+
+    /// Async style using Rustc support for async traits
+    pub trait LogExporterAsyncNative: Send + Sync + Debug {
```
cijothomas, line 94: From the benchmarks, this looks to be close enough with sync counterpart. Wondering if there is a reason we can't try this in the Exporter traits? The <T> for output might make things complex...?
(I only added this benchmarks to see if there is actually some price to be paid for introducing async.)

@demurgos Do you have some bandwidth to modify existing exporter trait to use async native style? The existing benchmarks should show improvement with this change, based on what i observe in this PR.

(Still pretty new to Rust async, so I might be missing something obvious!)

demurgos: Hey :wave: 
I was also interested in trying to impl a version using native async. I'm pretty familiar with it, but the lib is pretty large. I'll try to send something, but I can't promise any timeline.

opentelemetry-sdk/benches/log_exporter.rs
```diff
@@ -25,147 +31,232 @@ use opentelemetry_sdk::logs::LogRecord;
...
+        fn export(&mut self, batch: LogBatch<'_>) -> impl std::future::Future<Output = ()> + Send;
     }
 
-    fn force_flush(&self) -> LogResult<()> {
-        Ok(())
+    #[derive(Debug)]
+    pub struct NoOpExporterAsyncNative;
+
+    impl LogExporterAsyncNative for NoOpExporterAsyncNative {
+        #[inline]
```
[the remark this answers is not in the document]
demurgos: Oh, I missed the `#[inline]`, I thought I applied it everywhere. I'll update my PR.

cijothomas:
@lalitb Can you help continue this?

lalitb:
> @lalitb Can you help continue this?

Had a quick look. The `impl Future` approach is indeed much faster (around ~8 times faster in my local tests), which makes sense as it uses static dispatch with future type is known at compile time and so avoids overhead of heap allocations and dynamic dispatch that comes with using `Pin<Box<dyn Future>>` or `async-trait`. However, it requires Rust 1.75 or newer since `impl Trait` in traits was only stabilized in that version. Since our current `msrv` is 1.71, we can't use this approach without updating our Rust version.

Before going forward, it's good to discuss if we need to bump-up our msrv to 1.75 for otel-sdk, and all exporters.

cijothomas:
Closing this, as we have a PR already open to make this happen : https://github.com/open-telemetry/opentelemetry-rust/pull/2374

PR #2147, opened 2024-09-25
Add tests for periodic reader from various RT combinations

cijothomas:
Adding tests, ignoring the one that would fail.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -383,7 +383,7 @@ mod tests {
     #[test]
     fn collection_triggered_by_interval_tokio_current() {
         collection_triggered_by_interval_helper(runtime::TokioCurrentThread);
-    }    
+    }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn collection_triggered_by_interval_from_tokio_multi_one_thread_on_runtime_tokio() {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -424,4 +436,31 @@ mod tests {
...
         );
     }
+
+    fn collection_triggered_by_interval_helper<RT>(runtime: RT)
+    where
+        RT: crate::runtime::Runtime,
+    {
+        let interval = std::time::Duration::from_millis(1);
+        let exporter = InMemoryMetricsExporter::default();
+        let reader = PeriodicReader::builder(exporter.clone(), runtime)
```
utpilla, line 446: Do we expect to merge #2142 soon? If that's the case, then these tests would have to be removed/skipped. Maybe add these tests post #2142 once we have a mechanism to accept custom runtimes?

cijothomas: I am using these tests to primarily validate that when 2142 approach is merged, we'll not have hangs/deadlocks, not matter which async runtime user is on.

(Ofcourse, will modify these tests to match 2142 later, as there will be no need of specifying runtime, and reducing test matrix too)

utpilla: Not as inconvenient but I guess the simpler thing (for review) would be to have the tests added in #2142 itself. Otherwise, we just increase the amount of diff for review.

utpilla (approved):

hdost (approved):

PR #2152, opened 2024-09-28
Improve internal opentelemetry logging - directly using tracing mcros

lalitb:
## Changes

#2128 needs more discussion before it can be finalized. 
For now, this PR adds a more simplistic approach to directly use the tracing macro in the code for internal logging.

### DIFF
--- a/examples/self-diagnostics/src/main.rs
+++ b/examples/self-diagnostics/src/main.rs
@@ -2,8 +2,9 @@ use opentelemetry::global::{self, set_error_handler, Error as OtelError};
 use opentelemetry::KeyValue;
 use opentelemetry_appender_tracing::layer;
 use opentelemetry_otlp::WithExportConfig;
+use tracing_subscriber::filter::{EnvFilter, LevelFilter};
+use tracing_subscriber::fmt;
 use tracing_subscriber::prelude::*;
-use tracing_subscriber::EnvFilter;
 
 use std::error::Error;
 use tracing::error;
@@ -59,6 +60,7 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
         )
         .install_batch(opentelemetry_sdk::runtime::Tokio)
         .unwrap();
+    let cloned_provider = provider.clone();
 
     // Add a tracing filter to filter events from crates used by opentelemetry-otlp.
     // The filter levels are set as follows:
@@ -72,11 +74,22 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
         .add_directive("hyper=error".parse().unwrap())
         .add_directive("tonic=error".parse().unwrap())
         .add_directive("reqwest=error".parse().unwrap());
-    let cloned_provider = provider.clone();
-    let layer = layer::OpenTelemetryTracingBridge::new(&cloned_provider);
+    let opentelemetry_filter =
+        tracing_subscriber::filter::filter_fn(|metadata| metadata.target() == "opentelemetry");
+
+    let non_opentelemetry_filter =
+        tracing_subscriber::filter::filter_fn(|metadata| metadata.target() != "opentelemetry");
+
+    let fmt_opentelemetry_layer = fmt::layer()
+        .with_filter(LevelFilter::DEBUG)
+        .with_filter(opentelemetry_filter);
+    let otel_layer = layer::OpenTelemetryTracingBridge::new(&cloned_provider)
+        .with_filter(non_opentelemetry_filter.clone());
+
     tracing_subscriber::registry()
-        .with(filter)
-        .with(layer)
+        .with(fmt_opentelemetry_layer)
+        .with(fmt::layer().with_filter(filter))
+        .with(otel_layer)
         .init();
     provider
 }
--- a/examples/self-diagnostics/src/main.rs
+++ b/examples/self-diagnostics/src/main.rs
@@ -74,15 +74,27 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
         .add_directive("hyper=error".parse().unwrap())
         .add_directive("tonic=error".parse().unwrap())
         .add_directive("reqwest=error".parse().unwrap());
-    let opentelemetry_filter =
-        tracing_subscriber::filter::filter_fn(|metadata| metadata.target() == "opentelemetry");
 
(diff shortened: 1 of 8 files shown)


examples/self-diagnostics/src/main.rs
```diff
@@ -72,11 +74,22 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
         .add_directive("hyper=error".parse().unwrap())
         .add_directive("tonic=error".parse().unwrap())
         .add_directive("reqwest=error".parse().unwrap());
-    let cloned_provider = provider.clone();
-    let layer = layer::OpenTelemetryTracingBridge::new(&cloned_provider);
+    let opentelemetry_filter =
+        tracing_subscriber::filter::filter_fn(|metadata| metadata.target() == "opentelemetry");
```
cijothomas, line 78: == "opentelemetry"  -> starts_with("opentelemetry")

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -101,6 +101,13 @@ impl LogProcessor for SimpleLogProcessor {
             return;
         }
 
+        #[cfg(all(feature = "experimental-internal-logs"))]
+        tracing::debug!(
+            name: "simple_log_processor_emit",
+            target: "opentelemetry",
```
cijothomas, line 107:
```suggestion
            target: "opentelemetry_sdk",
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -217,6 +224,12 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                     // Log has finished, add to buffer of pending logs.
                     BatchMessage::ExportLog(log) => {
                         logs.push(log);
+                        #[cfg(feature = "experimental-internal-logs")]
+                        tracing::debug!(
+                            name: "batch_log_processor_record_count",
+                            target: "opentelemetry",
```
cijothomas, line 230:
```suggestion
                            target: "opentelemetry_sdk",
```

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -234,6 +234,8 @@ struct PeriodicReaderWorker<RT: Runtime> {
 
 impl<RT: Runtime> PeriodicReaderWorker<RT> {
     async fn collect_and_export(&mut self) -> Result<()> {
+        #[cfg(feature = "experimental-internal-logs")]
+        tracing::debug!(name: "collect_and_export", target: "opentelemetry", message ="Periodic export triggered.");
```
cijothomas, line 238: this could be triggered due to flush as well.

examples/self-diagnostics/src/main.rs
```diff
@@ -72,11 +74,22 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
...
+
+    let non_opentelemetry_filter =
```
cijothomas, line 80: I believe we already have decided in a previous SIG meeting to refactor this example. Can you clean it up.
IMHO, all it needs to show is
1. How to setup gathering internal logs via a layer
2. How to avoid sending the internal logs back to otel (show the filter).
3. Show the common case of hyper,tonic etc filtering.

cijothomas (changes requested):
This looks ideal for the coming release. thank you!
Requesting following changes before merge:
1. use opentelemetry-sdk, opentelemetry-otlp etc. as target.
2. if showing example, it should filter target by "starts-with" opentelemetry.

The self-diagnostics example require other refactorings, but that is okay to be done as follow up.

examples/self-diagnostics/src/main.rs
lalitb: Can you clarify what to cleanup, as I believe the current example is just doing above three.

lalitb:
> This looks ideal for the coming release. thank you! Requesting following changes before merge:
> 
> 1. use opentelemetry-sdk, opentelemetry-otlp etc. as target.
> 2. if showing example, it should filter target by "starts-with" opentelemetry.
> 
> The self-diagnostics example require other refactorings, but that is okay to be done as follow up.

Thanks, these are taken care now. Also, added the debug logs in opentelemetry-otlp.

lalitb:
The CI is failing for the issue separate to this PR, will check that and raise a separate PR for that.

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -19,6 +19,16 @@ pub struct ResourceMetrics {
     pub scope_metrics: Vec<ScopeMetrics>,
 }
 
+impl ResourceMetrics {
+    /// Returns the total number of metrics across all `ScopeMetrics`.
+    pub fn metrics_count(&self) -> usize {
```
cijothomas, line 24: lets not expose any new public API. if a component needs it, they can do it on their own for now. While refactoring the collect/export part of metrics, we can discuss what APIs it'd need to expose publically.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -234,6 +234,8 @@ struct PeriodicReaderWorker<RT: Runtime> {
 
 impl<RT: Runtime> PeriodicReaderWorker<RT> {
     async fn collect_and_export(&mut self) -> Result<()> {
+        #[cfg(feature = "experimental-internal-logs")]
+        tracing::debug!(name: "collect_and_export", target: "opentelemetry-sdk", status = "started");
```
cijothomas, line 238:
```suggestion
        tracing::debug!(name: "metrics_collect_and_export", target: "opentelemetry-sdk", status = "started");
```

cijothomas (approved):
Looks good enough for us to explore leveraging tracing for internal logs. Conventions around name,targets, the exact list of logs and attribute in them can be discussed after the initial release.

examples/self-diagnostics/src/main.rs
cijothomas: The part where the examples show storing 1st message and ignoring rest.
image

opentelemetry-sdk/src/metrics/data/mod.rs
lalitb: Fixed this by using the functionality directly where required.

PR #2180, opened 2024-10-07
Nits: Collect config fix in examples

cijothomas:
Config fix
CI check to keep running all msrv targets, even if one fails.


cijothomas:
@lalitb Integration Check CI was green even without fixing the config file. Looks like integration tests were somehow passing, lets follow up separately.

PR #2181, opened 2024-10-08
[WIP] Cleanup Global Error handler

### LINKED ISSUE #2175: Remove global error handler in favor of internal logs
- Internal logs support is now available (#2128), and it would make sense to use it for internal errors.
- This allows the removal of the global error handler.
- Current downside: it won’t be possible to register a custom handler for errors at this time. However, plan is to add support for registering custom handlers for internal logs.
- If "internal-logs" is enabled, tracing `error` macro would be used for handling error, else the `eprintln` would be used.

lalitb:
Fixes #2175

## Changes

In Progress. TODO:
 - Move TraceError and LogError from API to SDK. API should be only returning no-op constructs, and not the error objects. MetricsError is still being used in its API, so need to be handled separately.
 - Better handling for `<signal>Error`, `CoW` and few other types inside otel macros. As of now, the called needs to use `format!` to invoke their Debug impl, which is cumbersome. 

For future improvement:
- Allow applications/users to plugin their log handler.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -1,12 +1,11 @@
 use super::{BatchLogProcessor, LogProcessor, LogRecord, SimpleLogProcessor, TraceContext};
 use crate::{export::logs::LogExporter, runtime::RuntimeChannel, Resource};
-use opentelemetry::otel_warn;
 use opentelemetry::{
-    global,
     logs::{LogError, LogResult},
     trace::TraceContextExt,
     Context, InstrumentationLibrary,
 };
+use opentelemetry::{otel_error, otel_warn};
 
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
@@ -146,7 +145,7 @@ impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+                otel_error!(name: "LoggerProviderInner.Drop",  otel_name = "LoggerProviderInner.Drop", error = format!("{:?}", err));
             }
         }
     }
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -12,7 +12,6 @@ use futures_util::{
 #[cfg(feature = "logs_level_enabled")]
 use opentelemetry::logs::Severity;
 use opentelemetry::{
-    global,
     logs::{LogError, LogResult},
     otel_error, otel_warn, InstrumentationLibrary,
 };
@@ -118,7 +117,7 @@ impl LogProcessor for SimpleLogProcessor {
                 name: "simple_log_processor_emit_error",
                 error = format!("{:?}", err)
             );
-            global::handle_error(err);
+            otel_error!(name: "SimpleLogProcessor.emit", otel_error = "SimpleLogProcessor.emit", error = format!("{:?}", err));
         }
     }
 
@@ -175,7 +174,7 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
                 name: "batch_log_processor_emit_error",
                 error = format!("{:?}", err)
             );
-            global::handle_error(LogError::Other(err.into()));
+            otel_error!(name: "BatchLogProcessor.emit", otel_error = "BatchLogProcessor.emit", error = format!("{:?}", LogError::Other(err.into())));
         }
     }
 
@@ -246,7 +245,7 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                                     name: "batch_log_processor_export_error",
                                     error = format!("{:?}", err)
                                 );
-                                global::handle_error(err);
+                                otel_error!(name: "BatchLogProcessor.export", otel_error = "BatchLogProcessor.export", error = format!("{:?}", err));
                             }
                         }
                     }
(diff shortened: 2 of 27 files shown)


opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -146,7 +145,7 @@ impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+                otel_error!(name: "LoggerProviderInner.Drop",  otel_name = "LoggerProviderInner.Drop", error = format!("{:?}", err));
```
cijothomas, line 148: any warning/error level message must be user understandable and should not refer to internal constructs like 'Inner'.
name:LoggerProvideShutdownOnDrop
msg = "LoggerProvider shutdown() failed when invoked from Drop"
error = "error"

Its okay to handle this in a follow up - so this PR can focus on replacing global error handler with internal logs.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -175,7 +174,7 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
                 name: "batch_log_processor_emit_error",
                 error = format!("{:?}", err)
             );
-            global::handle_error(LogError::Other(err.into()));
+            otel_error!(name: "BatchLogProcessor.emit", otel_error = "BatchLogProcessor.emit", error = format!("{:?}", LogError::Other(err.into())));
```
cijothomas, line 177: Need to re-visit this, as this can flood if channel is full.

opentelemetry/src/global/error_handler.rs
```diff
@@ -47,40 +43,3 @@ impl<T> From<PoisonError<T>> for Error {
         Error::Other(err.to_string())
     }
 }
-
-struct ErrorHandler(Box<dyn Fn(Error) + Send + Sync>);
```
cijothomas, line 51: I'd suggest to split the PR into smaller ones - 
1. Review and replace global handler with error! internal logs, one crate or even small scope in one PR
2. When we are done with all crates, removing the global handler can be a simple PR.
3. Revisit each internal log - if it is not user actionable, and only good for sdk owners, then it should be debug level.

opentelemetry/src/global/internal_logging.rs
```diff
@@ -50,12 +50,24 @@ macro_rules! otel_warn {
         {
             tracing::warn!(name: $name, target: env!("CARGO_PKG_NAME"), "");
         }
+        #[cfg(not(feature = "internal-logs"))]
```
cijothomas, line 53: i think this is best left for Error level only.

opentelemetry/src/trace/context.rs
```diff
@@ -55,7 +55,9 @@ impl SpanRef<'_> {
         if let Some(ref inner) = self.0.inner {
             match inner.lock() {
                 Ok(mut locked) => f(&mut locked),
-                Err(err) => global::handle_error(err),
+                Err(_err) => {
+                    otel_error!(name: "SpanRef.with_inner_mut", otel_name = "SpanRef.with_inner_mut", error = format!("{:?}", _err));
```
cijothomas, line 59: instead of repeating otel_name, lets do it in the wrapper itself for now, until `tracing:fmt` is fixed to display event name.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: We can have name to follow pattern like:
name  = `<Module>.<Funtion>.<Action>`
So in this case it would be `LoggerProvider.Drop.Error` or `LoggerProvider.Drop.ShutdownError`. The actual error message would be contained in the error `attribute`. And if required, more attributes can be there for specific field values.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: This needs to be discussed whether it's necessary to handle these errors internally by implementing logic to prevent log flooding. This could involve using an "already-sent" flag and tracking the last sent timestamp to reduce the frequency of repeated error messages. 
Or else, let the user handle this flooding by adding a filter logic.

opentelemetry/src/trace/context.rs
lalitb: yes, I thought so, but then this will force this behavior for anyone using these macros in their custom exporters or processors.  Thought not a big reason for not adding it in macros :)

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: > Or else, let the user handle this flooding by adding a filter logic.

OTel SDK, by default, should not burden the user with such requirements!
Few thoughts:
1. If the item is being dropped due to buffer being full, emit a log only once in the app lifetime. If item is dropped for other unknown reasons, it maybe fine to emit log for each occurrence. I prefer that this should be a simple bool flag only and not a sophisticated throttling mechanism.
2. We should use an internal metric for items dropped. This may require more refactoring/designs.
3. Keep the count of items dropped (using atomic uint etc.). During shutdown/drop, if that count is greater than zero, emit a warning log "N items were dropped due to buffer full".

All of these are good for future additions, so no need to tackle in this PR. 
My only minimum bar for RC release - do not log for each time item is dropped due to buffer is full. (We had similar issue for Metrics overflow that was fixed recently)

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: module.func.action pattern is good pattern to follow, for debug logging that owners of this crate use to debug. I don't think we can have such a pattern for user-actionable ones, as they don't need to know which module/component is triggering this error.

lalitb:
closing, the PR has been split to multiple small PRs/

PR #2184, opened 2024-10-08
Global Log handler cleanup - Logs SDK

lalitb:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -1,9 +1,8 @@
 use super::{BatchLogProcessor, LogProcessor, LogRecord, SimpleLogProcessor, TraceContext};
 use crate::{export::logs::LogExporter, runtime::RuntimeChannel, Resource};
-use opentelemetry::otel_warn;
 use opentelemetry::{
-    global,
     logs::{LogError, LogResult},
+    otel_error, otel_warn,
     trace::TraceContextExt,
     Context, InstrumentationLibrary,
 };
@@ -114,9 +113,9 @@ impl LoggerProvider {
             let mut errs = vec![];
             for processor in &self.inner.processors {
                 if let Err(err) = processor.shutdown() {
-                    otel_warn!(
-                        name: "logger_provider_shutdown_error",
-                        error = format!("{:?}", err)
+                    otel_error!(
+                        name: "LoggerProvider.Shutdown.Error",
+                        error = err
                     );
                     errs.push(err);
                 }
@@ -128,10 +127,13 @@ impl LoggerProvider {
                 Err(LogError::Other(format!("{errs:?}").into()))
             }
         } else {
+            let error = LogError::Other("logger provider already shut down".into());
+
             otel_warn!(
-                name: "logger_provider_already_shutdown"
+                name: "LoggerProvider.Shutdown.AlreadyShutdown",
+                error = error
             );
-            Err(LogError::Other("logger provider already shut down".into()))
+            Err(error)
         }
     }
 }
@@ -146,7 +148,10 @@ impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+                otel_warn!(
+                    name: "LoggerProvider.Drop.AlreadyShutdown",
+                    error = err
+                );
             }
         }
     }
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -2,7 +2,7 @@ use super::{BatchLogProcessor, LogProcessor, LogRecord, SimpleLogProcessor, Trac
 use crate::{export::logs::LogExporter, runtime::RuntimeChannel, Resource};
 use opentelemetry::{
     logs::{LogError, LogResult},
-    otel_debug, otel_error, otel_warn,
+    otel_debug,
     trace::TraceContextExt,
     Context, InstrumentationLibrary,
 };
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -113,22 +113,19 @@ impl LoggerProvider {
             let mut errs = vec![];
             for processor in &self.inner.processors {
                 if let Err(err) = processor.shutdown() {
-                    if let Err(err) = processor.shutdown() {
-                        match err {
-                            // Specific handling for mutex poisoning
-                            LogError::MutexPoisoned(_) => {
-                                otel_debug!(
-                                    name: "LoggerProvider.Shutdown.MutexPoisoned",
-                                );
-                            }
-                            _ => {
-                                otel_error!(
-                                    name: "LoggerProvider.Shutdown.Error",
-                                    error = format!("{err}")
-                                );
-                            }
+                    match err {
+                        // Specific handling for mutex poisoning
+                        LogError::MutexPoisoned(_) => {
+                            otel_debug!(
+                                name: "LoggerProvider.Shutdown.MutexPoisoned",
+                            );
                         }
-                        errs.push(err);
+                        _ => {
+                            otel_debug!(
+                                name: "LoggerProvider.Shutdown.Error",
+                                error = format!("{err}")
+                            );
+                        }
                     }
                     errs.push(err);
                 }
@@ -142,7 +139,7 @@ impl LoggerProvider {
             }
         } else {
             let error = LogError::AlreadyShutdown("LoggerProvider".to_string());
-            otel_warn!(
+            otel_debug!(
                 name: "LoggerProvider.Shutdown.AlreadyShutdown",
             );
             Err(error)
@@ -160,7 +157,7 @@ impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
-                otel_warn!(
+                otel_debug!(
                     name: "LoggerProvider.Drop.ShutdownError",
                     error = format!("{}", err)
                 );
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -115,7 +115,7 @@ impl LoggerProvider {
                 if let Err(err) = processor.shutdown() {
                     otel_error!(
                         name: "LoggerProvider.Shutdown.Error",
-                        error = err
+                        error = format!("{err}")
                     );
                     errs.push(err);
                 }
@@ -128,10 +128,8 @@ impl LoggerProvider {
             }
         } else {
             let error = LogError::Other("logger provider already shut down".into());
-
             otel_warn!(
                 name: "LoggerProvider.Shutdown.AlreadyShutdown",
-                error = error
             );
             Err(error)
         }
@@ -149,7 +147,7 @@ impl Drop for LoggerProviderInner {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
                 otel_warn!(
-                    name: "LoggerProvider.Drop.AlreadyShutdown",
+                    name: "LoggerProvider.Drop.ShutdownError",
                     error = err
                 );
             }
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -28,7 +28,19 @@ static NOOP_LOGGER_PROVIDER: Lazy<LoggerProvider> = Lazy::new(|| LoggerProvider 
 });
 
 #[derive(Debug, Clone)]
-/// Creator for `Logger` instances.
+/// Handles the creation and coordination of [`Logger`]s.
+///
+/// All `Logger`s created by a `LoggerProvider` will share the same
+/// [`Resource`] and have their created log records processed by the
+/// configured log processors. This is a clonable handle to the `LoggerProvider`
+/// itself, and cloning it will create a new reference, not a new instance of a
+/// `LoggerProvider`. Dropping the last reference will trigger the shutdown of
+/// the provider, ensuring that all remaining logs are flushed and no further
+/// logs are processed. Shutdown can also be triggered manually by calling
+/// the [`shutdown`](LoggerProvider::shutdown) method.
+///
+/// [`Logger`]: opentelemetry::logs::Logger
+/// [`Resource`]: crate::Resource
 pub struct LoggerProvider {
     inner: Arc<LoggerProviderInner>,
 }
@@ -109,33 +121,15 @@ impl LoggerProvider {
             .is_ok()
         {
             // propagate the shutdown signal to processors
-            // it's up to the processor to properly block new logs after shutdown
-            let mut errs = vec![];
-            for processor in &self.inner.processors {
-                if let Err(err) = processor.shutdown() {
-                    match err {
-                        // Specific handling for mutex poisoning
-                        LogError::MutexPoisoned(_) => {
-                            otel_debug!(
-                                name: "LoggerProvider.Shutdown.MutexPoisoned",
-                            );
-                        }
-                        _ => {
-                            otel_debug!(
-                                name: "LoggerProvider.Shutdown.Error",
-                                error = format!("{err}")
-                            );
-                        }
-                    }
-                    errs.push(err);
-                }
-            }
-
+            let errs = self.inner.shutdown();
             if errs.is_empty() {
                 Ok(())
             } else {
-                // consolidate errors from all the processors - not all may be user errors
-                Err(LogError::Other(format!("{errs:?}").into()))
+                otel_warn!(
+                    name: "logger_provider_shutdown_error",
+                    error = format!("{:?}", errs)
+                );
+                Err(LogError::Other(format!("{:?}", errs).into()))
             }
         } else {
             let error = LogError::AlreadyShutdown("LoggerProvider".to_string());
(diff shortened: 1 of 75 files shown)


opentelemetry/src/global/internal_logging.rs
```diff
@@ -108,7 +114,28 @@ macro_rules! otel_error {
...
+                            }),+,
+                            ""
+                    )
         }
     };
 }
+
+/// Helper macro to format a value using Debug if available, falling back to Display
+#[macro_export]
+macro_rules! format_value {
```
lalitb, line 130: This avoid calling format() explicitly by the caller.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -128,10 +127,13 @@ impl LoggerProvider {
                 Err(LogError::Other(format!("{errs:?}").into()))
             }
         } else {
+            let error = LogError::Other("logger provider already shut down".into());
+
             otel_warn!(
-                name: "logger_provider_already_shutdown"
+                name: "LoggerProvider.Shutdown.AlreadyShutdown",
+                error = error
```
cijothomas, line 134: what is the need for passing error here? It does not give any additional information to user, right?

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -146,7 +148,10 @@ impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+                otel_warn!(
+                    name: "LoggerProvider.Drop.AlreadyShutdown",
```
cijothomas, line 152: not sure how we assume that the error is due to "AlreadyShutdown".

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: It doesn't add any extra value. I tried for consistency by including both the error and the name everywhere in case of error/warning. But no strong opinion, we can remove if we agree.

lalitb: Have removed this now.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Yes good catch I have changed it to `LoggerProvider.Drop.ShutdownError`. Have kept it as warning, as the error is during shutdown of LoggerProvider, and should be less destructive ( I think)

opentelemetry/src/global/internal_logging.rs
```diff
@@ -104,11 +123,46 @@ macro_rules! otel_error {
...
+                            $($key = {
+                                    $crate::format_value!($value)
+                            }),+,
+                            ""
+                    )
+        }
+        #[cfg(not(feature = "internal-logs"))]
+        {
+            {
+                let _ = ($name, $($value),+);
```
lalitb, line 148: this is to suppress the warning for unused variables - for the case where the "internal-logs" is disabled. The compiler should optimize out these assignments anyway.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -261,24 +255,12 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
...
-                                global::handle_error(LogError::from(format!(
-                                    "failed to send flush result: {:?}",
-                                    result
-                                )));
+                            if let Err(send_error) = channel.send(result) {
                                 otel_error!(
-                                    name: "batch_log_processor_flush_error",
-                                    error = format!("{:?}", result),
-                                    message = "Failed to send flush result"
+                                    name: "BatchLogProcessor.Flush.SendResultError",
```
cijothomas, line 260: `BatchLogProcessor.Flush.SendResultError` -- This does not look very user friendly. They need to know that Flush failed, but this is too much into implementation detail.

lalitb: BatchLogProcessor.Flush.SendResultError indicates that the Flush `may` have been successful, but there was failure to send the flush result to the channel. This is definitely useful for the developers of Otel Rust, if not much useful for the users.

cijothomas: if it is not useful for end users, then make it debug level.

> was failure to send the flush result to the channel.

These are implementation detail.

opentelemetry/src/global/internal_logging.rs
```diff
@@ -104,11 +123,46 @@ macro_rules! otel_error {
...
     };
 }
+
+/// Helper macro to format a value using Debug if available, falling back to Display
+#[macro_export]
+macro_rules! format_value {
+    ($value:expr) => {{
+        // Try Debug first
+        let debug_result = std::fmt::format(format_args!("{:?}", $value));
+        if debug_result.starts_with('<') || debug_result.contains("::") {
```
cijothomas, line 161: Can you add some details on why is this needed? Not sure I am able to understand this part...

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: But this is an error , how can it be at debug level?

cijothomas: if an error occur with flush, then user should know about it already as Flush should return `Result`. What additional value this gives to user? 
We shouldn't send error level logs (error/warn are typically enabled by the user) to user, unless it is actionable/giving-some-information which they would not have known otherwise. Flush failures can be easily communicated back to user via returned Result.

An example on the opposite side - say metric carinality limit hit and we are dropping metrics - there is no way to indicate that to user without internal log, so it is adding additional value to user.

lalitb: Have made it warn now - so we agree to have all errors not actionable by user as warning ?

opentelemetry/src/global/internal_logging.rs
cijothomas: unsure of this.. does format! make every attribute value a stringified version?

opentelemetry/src/global/internal_logging.rs
lalitb: This was to try formatting with `Debug`, and fallback to `Display` if there is any error. But now in latest commit I have simplified now to always use `Debug`.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: ok. Changed it to otel_debug.

opentelemetry/src/global/internal_logging.rs
cijothomas: Please do these changes separately. I am not sure if this is needed, as this is making the "thin" wrapper not so thin.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -128,10 +127,11 @@ impl LoggerProvider {
                 Err(LogError::Other(format!("{errs:?}").into()))
             }
         } else {
+            let error = LogError::Other("logger provider already shut down".into());
             otel_warn!(
-                name: "logger_provider_already_shutdown"
+                name: "LoggerProvider.Shutdown.AlreadyShutdown",
```
cijothomas, line 132: this looks redundant, as returned Result already contains this information. If we need to keep this, then maybe keep it at debug/info level.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -114,9 +113,9 @@ impl LoggerProvider {
             let mut errs = vec![];
             for processor in &self.inner.processors {
                 if let Err(err) = processor.shutdown() {
-                    otel_warn!(
-                        name: "logger_provider_shutdown_error",
-                        error = format!("{:?}", err)
+                    otel_error!(
```
cijothomas, line 116: Same as https://github.com/open-telemetry/opentelemetry-rust/pull/2184/files#r1793832380, this feels redundant, given Result will carry this info already.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -146,7 +146,10 @@ impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         for processor in &mut self.processors {
             if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+                otel_warn!(
```
cijothomas, line 149: i think we should do more changes to this part - this is the automatically invoked shutdown upon drop right? I believe we should first check if user has already called shutdown, and if yes, return without any logs (maybe a debug level log?)

if user has not called shutdown, then we need to call shutdown ourselves. Even here - should we log for each processor? or just log once saying shutdownfailed. The detailed failure reason should be part of the individual processor's own log.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -100,7 +99,8 @@ impl LogProcessor for SimpleLogProcessor {
         // noop after shutdown
         if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
             otel_warn!(
-                name: "simple_log_processor_emit_after_shutdown"
+                name: "SimpleLogProcessor.Export.AfterShutdown",
+                error = LogError::Other("Attempted to export a log after processor shutdown".into())
```
cijothomas, line 103: is there any advantage of wrapping this into LogError?

opentelemetry/src/global/internal_logging.rs
lalitb: Without this, we need to stringy each attribute at the caller site.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -115,10 +115,9 @@ impl LogProcessor for SimpleLogProcessor {
             });
         if let Err(err) = result {
             otel_error!(
-                name: "simple_log_processor_emit_error",
-                error = format!("{:?}", err)
+                name: "SimpleLogProcessor.Export.Error",
```
cijothomas, line 118: its not easy to tell if this is useful or not.. The failure could be due to
1. Mutex poisoning
2. Exporter facing issues like network etc.
3. Futures-Executor can fail itself too.

I don't have a good solution in mind, so probably okay to just replace current global error handler with otel_error! for now until we can find a better alternative.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: So if the error is returned to the user, we decide not to log it even if this could be actionable ?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -172,10 +168,9 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
 
         if let Err(err) = result {
             otel_error!(
-                name: "batch_log_processor_emit_error",
-                error = format!("{:?}", err)
+                name: "BatchLogProcessor.Export.Error",
```
cijothomas, line 171: will this be triggered when channel is full? If yes, we need to rethink this, as this can spam the log output.

opentelemetry/src/global/internal_logging.rs
cijothomas: does users of `tracing` have to it themselves? If yes, we (i.e all otel components doing this logging) can do the same.

Also, I don't think that every attribute should be stringified. There can be integer attributes like item_count, floating attributes like sampling_ratio etc. By stringifying in the helper, we lose the type information for everyone, even when they did not intend to do so.

lalitb: > does users of tracing have to it themselves? If yes, we (i.e all otel components doing this logging) can do the same.

yes, tracing crate has to do it themself. 

> By stringifying in the helper, we lose the type information for everyone, even when they did not intend to do so.

I didn't get this ?

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: https://github.com/open-telemetry/opentelemetry-rust/pull/2184#discussion_r1793852851

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: not to log or log at a very low severity.

Do you see any reason why this should be logged, given Result already carries this information? "Internal logging" is more for things that are otherwise hard to figure out, right?

opentelemetry/src/global/internal_logging.rs
cijothomas: By stringifying all attributes, won't it always result getting serialized as string, even when user provide an integer/float/bool value?
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L86

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: I prefer logging at a lower severity (debug) for internal debugging. When we request logs from the customer at the debug level, they should include everything necessary for troubleshooting, even if the customer disregards the shutdown result.

cijothomas: That is a good point! Agree with debug! here. It won't cause noise in normal operations, but when we ask users to report bug/issues, we can ask to collect all logs, including debug, so we'll have enough to investigate.

opentelemetry/src/global/internal_logging.rs
lalitb: Make sense, have removed the formatting .

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Changed this to debug now.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Changed this to debug now.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: > i think we should do more changes to this part - this is the automatically invoked shutdown upon drop right? I believe we should first check if user has already called shutdown, and if yes, return without any logs (maybe a debug level log?)

Modified it accordingly, and add debug log for already shutdown.

> if user has not called shutdown, then we need to call shutdown ourselves. Even here - should we log for each processor? or just log once saying shutdownfailed. The detailed failure reason should be part of the individual processor's own log.

The Processor::shutdown() returns the error, and we are now logging each to debug (instead of warn).

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: I find it useful - it adds convenience on top of  [std::error::Error](https://doc.rust-lang.org/std/error/trait.Error.html), by declaring various error types as enum and mapping it to actual error string. We don't need to add the error string while returning the error.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Have modified it to differentiate between MutexPoisoning and Exporter failure - I think we can further extend it in future.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Yes, this error only triggers when channel is full or closed. We need to add some throttling or logic to prevent flooding - have added the TODO for now, as we need common strategy for such flooding.

opentelemetry/src/logs/mod.rs
```diff
@@ -30,6 +30,14 @@ pub enum LogError {
     #[error("Exporter timed out after {} seconds", .0.as_secs())]
     ExportTimedOut(Duration),
 
+    /// Processor is already shutdown
+    #[error("{0} already shutdown")]
```
cijothomas, line 34: we are trying to get rid of this as part of https://github.com/open-telemetry/opentelemetry-rust/issues/1042 , so not sure why we need to add more to this?

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: Agree we need a common strategy, but lets remove the error log from here. It'll flood as-is when buffer is full.

opentelemetry/src/logs/mod.rs
lalitb: This will be moved to SDK eventually once global error handler is removed.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -140,14 +151,38 @@ impl LoggerProvider {
...
     resource: Resource,
+    is_shutdown: AtomicBool,
 }
 
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
-        for processor in &mut self.processors {
-            if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+        if self
```
cijothomas, line 159: I agree with this change, but please make it own separate PR, so we can keep this PR strictly for replacing global error handler with internal logger.

opentelemetry/src/global/internal_logging.rs
```diff
@@ -50,11 +50,30 @@ macro_rules! otel_warn {
         {
             tracing::warn!(name: $name, target: env!("CARGO_PKG_NAME"), "");
         }
+        #[cfg(not(feature = "internal-logs"))]
```
cijothomas, line 53: Agree with this change, but for future - please make them into own separate PR to make it easier to review.

cijothomas (approved):
LGTM.
Please keep this change separate with own changelog entry:
https://github.com/open-telemetry/opentelemetry-rust/pull/2184/files#r1796235052

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: should we remove altogether, or make it otel_debug for now - with comment to change it to otel_error once throttling is ready.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: This was done in response to the comment - https://github.com/open-telemetry/opentelemetry-rust/pull/2184#discussion_r1793842371. Should we keep it in this PR with separate changelog entry?

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: either of them are fine with me, though I slightly prefer removing altogether, as I don't know if we can ship a throttling solution for next release.

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: I always prefer short, very focused PRs! 
One can easily review them and continuously make progress by merging short PRs quickly. Slow and steady!

lalitb:
internal-log and shutdown related changes are now merged separately in #2195 and #2196. Just incase if there needs another look, else will go with merge.

lalitb:
> internal-log and shutdown related changes are now merged separately in #2195 and #2196. Just incase if there needs another look, else will go with merge.

Merging it to make the cleanup faster, please comment if any changes required will incorporate in separate PR.

PR #2185, opened 2024-10-08
Global error handler cleanup - Metrics SDK

lalitb:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
@@ -101,7 +101,7 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
                 // With a scale of -10 there is only two buckets for the whole range of f64 values.
                 // This can only happen if there is a max size of 1.
                 otel_error!(
-                    name: "ExpoHistogramDataPoint.Scale.Underflow",
+                    name: "ExponentialHistogramDataPoint.Scale.Underflow",
                     current_scale = self.scale,
                     scale_delta = scale_delta,
                     max_size = self.max_size,
--- a/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
@@ -100,6 +100,7 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
             if (self.scale - scale_delta as i8) < EXPO_MIN_SCALE {
                 // With a scale of -10 there is only two buckets for the whole range of f64 values.
                 // This can only happen if there is a max size of 1.
+                //TBD - check for throttling requirements if can happen too frequent.
                 otel_error!(
                     name: "ExponentialHistogramDataPoint.Scale.Underflow",
                     current_scale = self.scale,
@@ -107,7 +108,8 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
                     max_size = self.max_size,
                     min_scale = EXPO_MIN_SCALE,
                     record_min_max = self.record_min_max,
-                    record_sum = self.record_sum
+                    record_sum = self.record_sum,
+                    error = format!("The measurement will be dropped due to scale underflow. Check the histogram configuration")
                 );
                 return;
             }
--- a/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashMap, f64::consts::LOG2_E, sync::Mutex, time::SystemTime};
 
 use once_cell::sync::Lazy;
-use opentelemetry::{metrics::MetricsError, KeyValue};
+use opentelemetry::{otel_error, KeyValue};
 
 use crate::{
     metrics::data::{self, Aggregation, Temporality},
@@ -100,9 +100,15 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
             if (self.scale - scale_delta as i8) < EXPO_MIN_SCALE {
                 // With a scale of -10 there is only two buckets for the whole range of f64 values.
                 // This can only happen if there is a max size of 1.
-                opentelemetry::global::handle_error(MetricsError::Other(
-                    "exponential histogram scale underflow".into(),
-                ));
+                otel_error!(
+                    name: "ExpoHistogramDataPoint.Scale.Underflow",
+                    current_scale = self.scale,
+                    scale_delta = scale_delta,
+                    max_size = self.max_size,
+                    min_scale = EXPO_MIN_SCALE,
+                    record_min_max = self.record_min_max,
+                    record_sum = self.record_sum
+                );
                 return;
             }
             // Downscale
--- a/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
@@ -100,7 +100,11 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
             if (self.scale - scale_delta as i8) < EXPO_MIN_SCALE {
                 // With a scale of -10 there is only two buckets for the whole range of f64 values.
                 // This can only happen if there is a max size of 1.
-                //TBD - check for throttling requirements if can happen too frequent.
+
+                // This error is logged when a measurement is dropped because its value is either too high
+                // or too low, falling outside the allowable range defined by the scale and max_size.
+                // If these values are expected, the user should adjust the histogram configuration
+                // to accommodate the range.
                 otel_error!(
                     name: "ExponentialHistogramDataPoint.Scale.Underflow",
                     current_scale = self.scale,
@@ -109,6 +113,7 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
                     min_scale = EXPO_MIN_SCALE,
                     record_min_max = self.record_min_max,
                     record_sum = self.record_sum,
+                    value = format!("{:?}", v),
                     error = format!("The measurement will be dropped due to scale underflow. Check the histogram configuration")
                 );
                 return;
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -148,8 +148,8 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
             //TODO -  include name of meter, instrument
             otel_warn!( name: "MetricCardinalityLimitReached",
-                error = MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()),
-                cardinality_limit = cardinality_limit()
+                error = format!("{}", MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into())),
+                cardinality_limit = cardinality_limit() as u64,
             );
         }
     }
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -12,12 +12,12 @@ use std::ops::{Add, AddAssign, Sub};
 use std::sync::atomic::{AtomicBool, AtomicI64, AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, RwLock};
 
-use aggregate::is_under_cardinality_limit;
+use aggregate::{cardinality_limit, is_under_cardinality_limit};
 pub(crate) use aggregate::{AggregateBuilder, ComputeAggregation, Measure};
 pub(crate) use exponential_histogram::{EXPO_MAX_SCALE, EXPO_MIN_SCALE};
 use once_cell::sync::Lazy;
 use opentelemetry::metrics::MetricsError;
-use opentelemetry::{otel_error, KeyValue};
+use opentelemetry::{otel_warn, KeyValue};
 
 use crate::metrics::AttributeSet;
 
@@ -146,8 +146,10 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             let new_tracker = AU::new_atomic_tracker(self.buckets_count);
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
-            otel_error!( name: "MetricCardinalityLimitReached",
-                error = MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into())
+            //TODO -  include name of meter, instrument
+            otel_warn!( name: "MetricCardinalityLimitReached",
+                error = MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()),
+                cardinality_limit = cardinality_limit()
             );
         }
     }
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -16,7 +16,6 @@ use aggregate::{cardinality_limit, is_under_cardinality_limit};
 pub(crate) use aggregate::{AggregateBuilder, ComputeAggregation, Measure};
 pub(crate) use exponential_histogram::{EXPO_MAX_SCALE, EXPO_MIN_SCALE};
 use once_cell::sync::Lazy;
-use opentelemetry::metrics::MetricsError;
 use opentelemetry::{otel_warn, KeyValue};
 
 use crate::metrics::AttributeSet;
@@ -148,7 +147,7 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
             //TODO -  include name of meter, instrument
             otel_warn!( name: "MetricCardinalityLimitReached",
-                error = format!("{}", MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into())),
+                message = format!("{}", "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."),
                 cardinality_limit = cardinality_limit() as u64,
             );
         }
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -147,7 +147,7 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
             //TODO -  include name of meter, instrument
             otel_warn!( name: "MetricCardinalityLimitReached",
-                message = format!("{}", "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."),
+                message = format!("{}", "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric will not be logged until next collect."),
                 cardinality_limit = cardinality_limit() as u64,
             );
         }
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -146,7 +146,7 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             let new_tracker = AU::new_atomic_tracker(self.buckets_count);
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
-            otel_error!( name: "ValueMap.Measure.Overflow",
+            otel_error!( name: "MetricCardinalityLimitReached",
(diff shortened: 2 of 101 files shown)


opentelemetry/src/global/internal_logging.rs
```diff
@@ -50,11 +50,29 @@ macro_rules! otel_warn {
         {
             tracing::warn!(name: $name, target: env!("CARGO_PKG_NAME"), "");
         }
+        #[cfg(not(feature = "internal-logs"))]
```
lalitb, line 53: The changes in `internal-logs` are same as #2184. Needn't be reviewed twice.

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -414,15 +413,13 @@ where
...
-
-                global::handle_error(MetricsError::Other(format!(
-                    "duplicate metric stream definitions, names: ({} and {}), descriptions: ({} and {}), kinds: ({:?} and {:?}), units: ({:?} and {:?}), and numbers: ({} and {})",
-                    existing.name, id.name,
-                    existing.description, id.description,
-                    existing.kind, id.kind,
-                    existing.unit, id.unit,
-                    existing.number, id.number,
-               )))
+                otel_warn!(name: "InstrumentSync.AddSync.DuplicateMetricStreamDefinitions",
```
cijothomas, line 416: These names must be user friendly, and not internal implementation name based.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -264,20 +263,26 @@ impl<RT: Runtime> PeriodicReaderWorker<RT> {
         match message {
             Message::Export => {
                 if let Err(err) = self.collect_and_export().await {
-                    global::handle_error(err)
+                    otel_error!(name: "PeriodicReaderWorker.Export.Error",error = err);
```
cijothomas, line 266: you can ignore this file initially, to avoid conflict for my PR on the same file.

opentelemetry-sdk/src/metrics/pipeline.rs
lalitb: Have updated the names, and also reverted changes from periodic reader.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -146,9 +146,8 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             let new_tracker = AU::new_atomic_tracker(self.buckets_count);
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
-            global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
-            otel_warn!( name: "ValueMap.measure",
-                message = "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."
+            otel_error!( name: "ValueMap.Measure.Overflow",
```
cijothomas, line 149:
```suggestion
            otel_error!( name: "MetricCardinalityLimitReached",
```

This should be user actionable. We should include name of meter, instrument along with the configured limit, but those requires major refactoring.

cijothomas: I think this can be a WARN. We can reserve ERROR for the case where entire telemetry is lost/dropped. In this case, we are not dropping entire measurement but folding into overflow.

lalitb: Have changed ERROR to WARN, and added the cardinality limit. With  TODO to add instrument name and metric name.

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
```diff
@@ -100,9 +100,15 @@ impl<T: Number> ExpoHistogramDataPoint<T> {
             if (self.scale - scale_delta as i8) < EXPO_MIN_SCALE {
                 // With a scale of -10 there is only two buckets for the whole range of f64 values.
                 // This can only happen if there is a max size of 1.
-                opentelemetry::global::handle_error(MetricsError::Other(
-                    "exponential histogram scale underflow".into(),
-                ));
+                otel_error!(
```
cijothomas, line 103: I don't know the inner workings enough to give a strong opinion - but unless this is a auto recoverable error, this can flood the error log.

opentelemetry-sdk/src/metrics/manual_reader.rs
```diff
@@ -84,9 +84,7 @@ impl MetricReader for ManualReader {
             if inner.sdk_producer.is_none() {
                 inner.sdk_producer = Some(pipeline);
             } else {
-                global::handle_error(MetricsError::Config(
-                    "duplicate reader registration, did not register manual reader".into(),
-                ))
+                otel_error!(name: "ManualReader.RegisterPipeline.DuplicateRegistration");
```
cijothomas, line 87: info/debug only. Even if a user gets this message, they won't know what to do. Its helpful to us only.

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -74,7 +74,7 @@ impl SdkMeter {
     {
         let validation_result = validate_instrument_config(builder.name.as_ref(), &builder.unit);
         if let Err(err) = validation_result {
-            global::handle_error(err);
+            otel_error!(name: "SdkMeter.CreateCounter.ValidationError", error = format!("{}", err));
```
cijothomas, line 77: Agree Error is the right severity for this, but this is still not so user-friendly.
Since this is user facing, the internal log should be clearly actionable, along with the impact of this error.
In this case, the impact is - measurements reported using this instrument will be ignore.
The "why" is also not so obvious. 
As an end-user, I'd prefer to see something like

Name: InstrumentCreationFailed
InstrumentName: A_non_asccii_Metric
MeterName: MeterName
Reason: Instrument Name contains non-ascii charactors. Link to spec, optionally.
Message: Measurements from this instruments will be ignored.

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -90,7 +90,7 @@ impl SdkMeter {
         {
             Ok(counter) => Ok(counter),
             Err(err) => {
-                global::handle_error(err);
+                otel_error!(name: "SdkMeter.CreateCounter.Error", error = format!("{}", err));
```
cijothomas, line 93: Same comment as https://github.com/open-telemetry/opentelemetry-rust/pull/2185/files#r1796240076
I dont think a user would know the difference between ValidationError vs Error.

cijothomas:
@lalitb Let us know if this is ready for another review. Unfortunately, it got some conflicts too to be resolved, hopefully simple ones.

lalitb:
> @lalitb Let us know if this is ready for another review. Unfortunately, it got some conflicts too to be resolved, hopefully simple ones.

Resolved the conflicts. Will go through the review comments now.

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
lalitb: As per what I can understand this part of code, this error occurs with restrictive `max_size` configuration, while the application is recording measurements with values that are far apart than what allowed by `max_size`.  And error would be logged whenever the faulty measurement is recorded. If these faulty measurements are not frequent, the error log won't be flooded, else it can. Again, either some kind of throttling or simply flag to log only once need to be added. Let me know what you suggest, else I can keep TODO to revisit.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -146,9 +146,10 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             let new_tracker = AU::new_atomic_tracker(self.buckets_count);
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
-            global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
-            otel_warn!( name: "ValueMap.measure",
-                message = "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."
+            //TODO -  include name of meter, instrument
+            otel_warn!( name: "MetricCardinalityLimitReached",
+                error = format!("{}", MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into())),
```
utpilla, line 151: Since this is a warning should we use `message` or `warning` as the key name instead of `error`? 
```suggestion
                message = format!("{}", MetricsError::Other("Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into())),
```

utpilla, line 151: Also, could we simply log the message string instead of wrapping it in `MetricsError::Other` enum variant?

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -146,9 +145,10 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             let new_tracker = AU::new_atomic_tracker(self.buckets_count);
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
-            global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
-            otel_warn!( name: "ValueMap.measure",
-                message = "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."
+            //TODO -  include name of meter, instrument
+            otel_warn!( name: "MetricCardinalityLimitReached",
+                message = format!("{}", "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."),
```
utpilla, line 150: Not from this PR but I think the below sentence sounds better.

```suggestion
                message = format!("{}", "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric will not be logged until next collect."),
```

opentelemetry-sdk/src/metrics/manual_reader.rs
```diff
@@ -76,10 +76,11 @@ impl MetricReader for ManualReader {
...
             if inner.sdk_producer.is_none() {
                 inner.sdk_producer = Some(pipeline);
-            } else {
-                global::handle_error(MetricsError::Config(
-                    "duplicate reader registration, did not register manual reader".into(),
-                ))
+            }  else {
+                otel_debug!(
+                    name: "ManualReader.RegisterPipeline.DuplicateRegistration",
+                    error = "The pipeline is already registered to the Reader. Registering pipeline multiple times is not allowed."
```
utpilla, line 82: Should we use `message` here as well instead of `error`?

opentelemetry-sdk/src/metrics/manual_reader.rs
```diff
@@ -76,10 +76,11 @@ impl MetricReader for ManualReader {
             // Only register once. If producer is already set, do nothing.
             if inner.sdk_producer.is_none() {
                 inner.sdk_producer = Some(pipeline);
-            } else {
-                global::handle_error(MetricsError::Config(
-                    "duplicate reader registration, did not register manual reader".into(),
-                ))
+            }  else {
+                otel_debug!(
+                    name: "ManualReader.RegisterPipeline.DuplicateRegistration",
```
utpilla, line 81: I think it's better to leave out the implementation details from the event names. We might or might not change this method name or we might rename pipeline to something else in the future.

```suggestion
                    name: "ManualReader.DuplicateRegistration",
```

utpilla: We should come up with some convention all these events would follow. For example, something like this:
`{crate}.{component}.{optional subcomponent}.{eventName}`

`Sdk.MetricReader.ManualReader.DuplicateRegistration`

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -107,34 +116,41 @@ impl SdkMeter {
...
                     self.pipes
                         .register_callback(move || callback(cb_inst.as_ref()));
                 }
-
                 ObservableCounter::new()
             }
-            Err(err) => {
-                global::handle_error(err);
+            Err(_err) => {
+                // TBD - Add error handling
```
utpilla, line 153: Let's be consistent with the usage of `TODO`/`TBD` in the repo so that we can search for these actionable items easily. I did a [quick search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-rust%20TODO&type=code) on the repo and mostly we have been using `TODO`. There is only one instance where we have used `TBD`.

```suggestion
                // TODO - Add error handling
```

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -75,14 +75,18 @@ impl SdkMeter {
     {
         let validation_result = validate_instrument_config(builder.name.as_ref(), &builder.unit);
         if let Err(err) = validation_result {
-            global::handle_error(err);
+            otel_error!(
+                name: "SdkMeter.CreateCounter.InstrumentCreationFailed", 
+                meter_name = self.scope.name.as_ref(),
+                instrument_name = builder.name.as_ref(),
+                error = format!("Measurements from the instrument will be ignored. Reason: {}", err));
```
utpilla, line 82: If we are using dedicated events for each instrument, then we might as well be more specific in the message.

```suggestion
                error = format!("Measurements from the counter will be ignored. Reason: {}", err));
```

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -245,36 +245,56 @@ impl<RT: Runtime> PeriodicReaderWorker<RT> {
...
         match message {
             Message::Export => {
                 if let Err(err) = self.collect_and_export().await {
-                    global::handle_error(err)
+                    match err {
+                        MetricsError::ExportTimeout(_, _) => {
+                            otel_error!( name: "PeriodicReader.ExportFailed", error  = format!("{:?}", err));
+                        }
+                        MetricsError::ReaderShutdown => {
+                            otel_debug!( name: "PeriodicReader.ReaderShutdown", error = format!("{:?}", err));
```
utpilla, line 265: I think we could use `otel_error` for each of these variants as they all seem important. Metrics export is anyway not a hot-path operation so we don't have to worry about flooding of error messages.

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -414,15 +413,18 @@ where
...
-                    existing.unit, id.unit,
-                    existing.number, id.number,
-               )))
+                otel_warn!(name: "Instrument.DuplicateMetricStreamDefinitions",
+                    name = format!("{}",id.name),
+                    description = format!("{}", id.description),
+                    kind = format!("{:?}", id.kind),
+                    unit = format!("{}",id.unit),
+                    number = format!("{}", id.number),
+                    existing_name = format!("{}", existing.name),
```
utpilla, line 422: I think we can skip logging this set of attributes twice. The event name says that we have a duplicate stream so it should already be understood that we have two identical sets of attributes.

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -710,8 +712,7 @@ where
 
         if errs.is_empty() {
             if measures.is_empty() {
-                // TODO: Emit internal log that measurements from the instrument
-                // are being dropped due to view configuration
+                // Error is logged elsewhere.
```
utpilla, line 715: Remove the `if` check?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -245,36 +245,56 @@ impl<RT: Runtime> PeriodicReaderWorker<RT> {
...
-            Either::Right(_) => {
-                otel_error!(
-                    name: "collect_and_export",
-                    status = "timed_out"
-                );
-                Err(MetricsError::Other("export timed out".into()))
-            }
+            Either::Right(_) => Err(MetricsError::ExportTimeout(
+                "PeriodicReader".into(),
+                self.timeout.as_nanos(),
```
utpilla, line 250: Why are we converting this into nanoseconds?

utpilla: Also, we don't seem to be using the string or time duration in logging at all. Do we need them?

opentelemetry-sdk/src/metrics/periodic_reader.rs
lalitb: I thought of doing same way, @cijothomas comment in earlier PRs was to only keep user-actionable one's in the error, and rest in debug - irrespective of not in hot-path.

opentelemetry-sdk/src/metrics/periodic_reader.rs
lalitb: we are doing it here

    #[error("Metrics reader {0} failed with timeout. Max configured timeout: {1} ns")]
    ExportTimeout(String, u128),

opentelemetry-sdk/src/metrics/pipeline.rs
lalitb: I am not very sure of the logic here. Can you check once. As we return at line 414 if they are same.

opentelemetry-sdk/src/metrics/meter.rs
lalitb: Yes, TODO is something we follow in repo, just that I am used to TBD :)

opentelemetry-sdk/src/metrics/manual_reader.rs
lalitb: We are already adding crate name in the macro. Trying to follow the naming as 
`ModuleName.SecondLevel.OptionalThirdLevel.EventName`

second lvel could be struct / method in this module.

utpilla: Internal modules could still be moved around or refactored so I think we should rely on:
`component name (spec concepts such as MeterProvider, MetricReader, etc.) -> sub component(s) -> event name`

opentelemetry-sdk/src/metrics/periodic_reader.rs
utpilla: Two questions here:
1. Why are we adding these new variants to the `MetricsError` enum? Would the end user be able to consume these variants? They only seem to be used in internal methods.
2. If we decide that these variants have to be public, why use nanoseconds for time out? Why not milliseconds?

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: Shutdown failures are propagated to users via Result anyway, so no need of another Error level log.

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
cijothomas: Unless we are 100% sure this cannot cause flooding of logs, lets remove the log from here, and leave a TODO to add logging once we understand more.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -146,9 +145,10 @@ impl<AU: AtomicallyUpdate<T>, T: Number, O: Operation> ValueMap<AU, T, O> {
             let new_tracker = AU::new_atomic_tracker(self.buckets_count);
             O::update_tracker(&new_tracker, measurement, index);
             trackers.insert(STREAM_OVERFLOW_ATTRIBUTES.clone(), Arc::new(new_tracker));
-            global::handle_error(MetricsError::Other("Warning: Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged.".into()));
-            otel_warn!( name: "ValueMap.measure",
-                message = "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric until next collect will not be logged."
+            //TODO -  include name of meter, instrument
+            otel_warn!( name: "MetricCardinalityLimitReached",
+                message = format!("{}", "Maximum data points for metric stream exceeded. Entry added to overflow. Subsequent overflows to same metric will not be logged until next collect."),
```
cijothomas, line 150: what is the need for `format`?

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -75,14 +75,18 @@ impl SdkMeter {
     {
         let validation_result = validate_instrument_config(builder.name.as_ref(), &builder.unit);
         if let Err(err) = validation_result {
-            global::handle_error(err);
+            otel_error!(
```
cijothomas, line 78: otel_error!(
                name: "InstrumentCreationFailed", 
                meter_name = self.scope.name.as_ref(),
                instrument_name = builder.name.as_ref(),
                message = "Measurements from this instrument will be ignored."
                reason = fmt(err))

^ I prefer this version. Please see if this is better for end users.

cijothomas: Error vs Warning - this is debatable. We need to find some agreement to follow throughout.
If entire telemetry flow is affected, then Error is apt.
for missing measurements from a single instrument - warn *might* be sufficient. This is good to be followed up separately anyway.

cijothomas:
@lalitb @utpilla This is getting too big with large number of comments, making it even harder to keep up. 
Can you reduce scope so we can focus on one file/small-section-within-a-file at a time? A lot of internal logging requires reviewers to also understand the overall flow and its a lot easier to do with very targeted PR.

lalitb:
> @lalitb @utpilla This is getting too big with large number of comments, making it even harder to keep up. Can you reduce scope so we can focus on one file/small-section-within-a-file at a time? A lot of internal logging requires reviewers to also understand the overall flow and its a lot easier to do with very targeted PR.

I can try that. Can move this to draft for now, and will create further smalls along with incorporating the existing comments. Let me know if that's fine ?

cijothomas:
> > @lalitb @utpilla This is getting too big with large number of comments, making it even harder to keep up. Can you reduce scope so we can focus on one file/small-section-within-a-file at a time? A lot of internal logging requires reviewers to also understand the overall flow and its a lot easier to do with very targeted PR.
> 
> I can try that. Can move this to draft for now, and will create further smalls along with incorporating the existing comments. Let me know if that's fine ?

Yes that works!

lalitb:
closing, the PRs has been split to multiple small PRs.

PR #2186, opened 2024-10-09
[Metrics API] Update HistogramBuilder 

utpilla:
## Changes
- Use `Histogram<T>` instead of `T` for `HistogramBuilder` APIs
- This is to be consistent with `InstrumentBuilder` and `AsynInstrumentBuilder` APIs


cijothomas (approved):
Okay to skip changelog? Is this like a "technically breaking change" , but all examples compile just fine, so maybe okay...

utpilla:
> Okay to skip changelog? Is this like a "technically breaking change" , but all examples compile just fine, so maybe okay...

Yes, I don't expect this change to break anyone unless they have some weird Histogram setup where **they explicitly add the type of the builder** and then call `init` on the builder.

```rust
let histogram_builder: HistogramBuilder<'_, opentelemetry::metrics::Histogram<f64>> = meter
        .f64_histogram("my_histogram");

let histogram = histogram_builder.init();
```

I don't expect people to be doing this so I think we can skip the CHANGELOG update.

PR #2195, opened 2024-10-11
Avoid redundant shutdown in LoggerProvider::drop when already shut down

lalitb:
## Changes

Prevents LoggerProvider::drop from calling shutdown if the provider has already been shut down. 

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -2,6 +2,7 @@ use super::{BatchLogProcessor, LogProcessor, LogRecord, SimpleLogProcessor, Trac
 use crate::{export::logs::LogExporter, runtime::RuntimeChannel, Resource};
 use opentelemetry::otel_warn;
 use opentelemetry::{
+    global,
     logs::{LogError, LogResult},
     trace::TraceContextExt,
     Context, InstrumentationLibrary,
@@ -114,7 +115,29 @@ impl LoggerProvider {
 
     /// Shuts down this `LoggerProvider`
     pub fn shutdown(&self) -> LogResult<()> {
-        self.inner.shutdown()
+        if self
+            .inner
+            .is_shutdown
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .is_ok()
+        {
+            // propagate the shutdown signal to processors
+            let errs = self.inner.shutdown();
+            if errs.is_empty() {
+                Ok(())
+            } else {
+                otel_warn!(
+                    name: "logger_provider_shutdown_error",
+                    error = format!("{:?}", errs)
+                );
+                Err(LogError::Other(format!("{:?}", errs).into()))
+            }
+        } else {
+            otel_warn!(
+                name: "logger_provider_already_shutdown"
+            );
+            Err(LogError::Other("logger provider already shut down".into()))
+        }
     }
 }
 
@@ -126,43 +149,26 @@ struct LoggerProviderInner {
 }
 
 impl LoggerProviderInner {
-    pub(crate) fn shutdown(&self) -> LogResult<()> {
-        if self
-            .is_shutdown
-            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
-            .is_ok()
-        {
-            // propagate the shutdown signal to processors
-            // it's up to the processor to properly block new logs after shutdown
-            let mut errs = vec![];
-            for processor in &self.processors {
-                if let Err(err) = processor.shutdown() {
-                    otel_warn!(
-                        name: "logger_provider_shutdown_error",
-                        error = format!("{:?}", err)
-                    );
-                    errs.push(err);
-                }
+    /// Shuts down the `LoggerProviderInner` and returns any errors.
+    pub(crate) fn shutdown(&self) -> Vec<LogError> {
+        let mut errs = vec![];
+        for processor in &self.processors {
+            if let Err(err) = processor.shutdown() {
+                errs.push(err);
             }
-
-            if errs.is_empty() {
-                Ok(())
-            } else {
-                Err(LogError::Other(format!("{errs:?}").into()))
-            }
-        } else {
-            otel_warn!(
-                name: "logger_provider_already_shutdown"
-            );
-            Err(LogError::Other("logger provider already shut down".into()))
         }
+        errs
     }
 }
 
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
-        let _ = self.shutdown();
-        //TODO - handle error during shutdown
+        if !self.is_shutdown.load(Ordering::Relaxed) {
+            let errs = self.shutdown();
+            if !errs.is_empty() {
+                global::handle_error(LogError::Other(format!("{:?}", errs).into()));
+            }
+        }
     }
 }
 
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -2,7 +2,6 @@ use super::{BatchLogProcessor, LogProcessor, LogRecord, SimpleLogProcessor, Trac
 use crate::{export::logs::LogExporter, runtime::RuntimeChannel, Resource};
 use opentelemetry::otel_warn;
 use opentelemetry::{
-    global,
     logs::{LogError, LogResult},
     trace::TraceContextExt,
     Context, InstrumentationLibrary,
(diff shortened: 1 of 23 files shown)


opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -136,22 +135,25 @@ impl LoggerProvider {
     }
 }
 
+impl Drop for LoggerProvider {
```
cijothomas, line 138: what is the reason for implementing Drop for the LoggerProvider, as opposed to the LoggerProviderInner?

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -136,22 +135,25 @@ impl LoggerProvider {
...
+    fn drop(&mut self) {
```
cijothomas, line 139: I believe we want to replicate the Metrics behavior here as well https://github.com/open-telemetry/opentelemetry-rust/pull/2156/files 
But this looks different. Can you explain why is this differently handled? Specifically, why is Drop implemented for LoggerProvider as opposed to doing it to the inner one, given LoggerProvider is cloned for each logger?

Also, this still causes the false alarm to the user, when shutdown is triggered by Drop even when user has shutdown it already?

cijothomas (changes requested):
I have some comment left here:
https://github.com/open-telemetry/opentelemetry-rust/pull/2195/files#r1797268610

Please check if it makes sense. I am not I understand the reasoning fully why LoggerProvider is doing things differently than MeterProvider..

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: yes good catch. I tried to optimize it incorrectly. it. Have added the validation in LoggerProviderInner::drop now.

lalitb: Also updated the unit-test to test with multiple LoggerProvider instances , sharing a clone of LoggerProviderInner.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -140,13 +140,20 @@ impl LoggerProvider {
...
     resource: Resource,
+    is_shutdown: AtomicBool,
 }
 
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
-        for processor in &mut self.processors {
-            if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+        if self
```
cijothomas, line 148: i don't think drop should modify the is_shutdown flag here. it should just call the shutdown() method if not already shutdown, and nothing more. the shutdown() method imlementation should know how to handle shutdown() and invoke processors etc - that logic is not required to be duplicated in drop.

lalitb: shutdown() is part of LoggerProvider, and can't be called from within LoggerProviderInner.  We can instead create a private `shutdown()` method inside the LoggerProviderInner, and let it be called from both LoggerProvider::shutdown() and LoggerProviderInner::drop().

cijothomas: i think we should move shutdown() logic to be inside inner one, with the loggerprovider delgating everything, including shutdown to the inner.

LoggerProvier/Meterprovider/Tracerprovider - they are just shared ref to the actual provider. So entire logic should be moved inside inner. When working on periodicreader, I think i managed to achieve the same for PeriodicReader.

image

If this makes sense, then lets do it everywhere, starting with LoggerProvider

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/meter_provider.rs#L25 The doc comment here is something we should replicate too.

lalitb: Have added a crate private method in LoggerProviderInner, and let it be called from both Drop and Shutdown. The error/warn/debug handling can be ignored for now (didn't change it), will handle it in the separate PR.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -136,19 +159,10 @@ impl LoggerProvider {
...
-}
-
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
-        for processor in &mut self.processors {
-            if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
-            }
-        }
+        let _ = self.shutdown();
```
cijothomas, line 164: this is still calling shutdown, even when is_shutdown is already true?

lalitb: shutdown take care of checking is_shutdown, so we dont need to do it here

cijothomas: it should still be done in drop to avoid the false alarm to users, as shutdown() will now emit an internal log about "already shutdown" if user has already invoked shutdown.
See https://github.com/open-telemetry/opentelemetry-rust/pull/2156/files for reference.

lalitb: This already have a TODO in next line to be taken care in #2184

cijothomas: This is a logical flaw that drop is unnecessarily calling shutdown, when it is not supposed to. 2184 just fixes the confusing internal log, but isn't the right behavior is for drop to *not* call shutdown itself?

lalitb: I am not sure why this is a flaw - The MeterProvider has a flaw for checking the is_shutdown flag twice - once is drop, and then in shutdown. I think we should fix that. The `Inner::shutdown` should check the is_shutdown, and return the Result status to the caller, caller then handles the Error accordingly.

lalitb: I have instead removed is_shutdown check from LoggerProviderInner::shutdown(), and added both in LoggerProviderInner::drop and LoggerProvider::Shutdown. Let me know if this works, will update TracerProvider PR too accordingly.

cijothomas: I am not seeing any flaw with MeterProvider. It is expected that shutdown() checks if already shutdown and return appropriate Result. It is also expected that Drop should not even call shutdown() if user has already done so. In my view, calling shutdown from drop is SDK being good citizen. If user has forgotten to call shutdown, then drop will do it for them. But if user has already done their job, drop should detect that and don't call shutdown again.

If you can send a PR with a proposal to address issue with MeterProvider, that could help me see what is the current issue vs proposed solution.

lalitb: > If you can send a PR with a proposal to address issue with MeterProvider, that could help me see what is the current issue vs proposed solution.

Please review the updated changes in this PR. If it looks good, I will apply similar changes for TracerProvider and MeterProvider. The issue with MeterProvider is that the is_shutdown check is performed twice during a drop, which is unnecessary:
```
MeterProviderInner::Drop -> checks is_shutdown -> calls MeterProviderInner::shutdown -> checks is_shutdown again -> invokes shutdown

MeterProvider::shutdown -> calls MeterProviderInner::shutdown -> checks is_shutdown -> invokes shutdown
```

The `MeterProviderInner::shutdown` method is private and won’t be called directly by the user. We have two possible approaches for handling the is_shutdown check:

- Option 1: Perform the is_shutdown check within `MeterProviderInner::shutdown` and skip the check in both `MeterProviderInner::Drop` and` MeterProvider::shutdown`.
- Option 2: Skip the is_shutdown check in `MeterProviderInner::shutdown` and instead perform the check in both `MeterProviderInner::Drop` and `MeterProvider::shutdown`.

I earlier implemented option 1 in this PR, but now changed it to option 2 based on your concern.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -24,15 +24,26 @@ static NOOP_LOGGER_PROVIDER: Lazy<LoggerProvider> = Lazy::new(|| LoggerProvider
...
+/// Handles the creation and coordination of [`Logger`]s.
+///
+/// All `Logger`s created by a `LoggerProvider` will share the same
+/// [`Resource`] and have their created log records processed by the
+/// configured log processors. This is a clonable handle to the `LoggerProvider`
+/// itself, and cloning it will create a new reference, not a new instance of a
+/// `LoggerProvider`. Dropping the last reference will trigger the shutdown of
+/// the provider, ensuring that all remaining logs are flushed and no further
+/// logs are processed. Shutdown can also be triggered manually by calling
+/// the [`shutdown`](LoggerProvider::shutdown) method.
```
cijothomas, line 41: 💯 excellent doc comment!

cijothomas (approved):

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas: This looks good!

PR #2197, opened 2024-10-11
Avoid redundant shutdown in TracerProvider::drop when already shut down

lalitb:
## Changes

changes similar to #2195 for TracerProvider

### DIFF
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -640,6 +640,9 @@ mod tests {
         // existing tracer becomes noops after shutdown
         let _ = test_tracer_1.start("test");
         assert!(assert_handle.started_span_count(2));
+
+        // also existing tracer's tracer provider are in shutdown state
+        assert!(test_tracer_1.provider().is_shutdown());
     }
 
     #[derive(Debug)]
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -645,15 +645,11 @@ mod tests {
     #[derive(Debug)]
     struct CountingShutdownProcessor {
         shutdown_count: Arc<AtomicU32>,
-        flush_called: Arc<AtomicBool>,
     }
 
     impl CountingShutdownProcessor {
-        fn new(shutdown_count: Arc<AtomicU32>, flush_called: Arc<AtomicBool>) -> Self {
-            CountingShutdownProcessor {
-                shutdown_count,
-                flush_called,
-            }
+        fn new(shutdown_count: Arc<AtomicU32>) -> Self {
+            CountingShutdownProcessor { shutdown_count }
         }
     }
 
@@ -667,7 +663,6 @@ mod tests {
         }
 
         fn force_flush(&self) -> TraceResult<()> {
-            self.flush_called.store(true, Ordering::SeqCst);
             Ok(())
         }
 
@@ -687,7 +682,6 @@ mod tests {
             let shared_inner = Arc::new(TracerProviderInner {
                 processors: vec![Box::new(CountingShutdownProcessor::new(
                     shutdown_count.clone(),
-                    flush_called.clone(),
                 ))],
                 config: Config::default(),
                 is_shutdown: AtomicBool::new(false),
@@ -723,13 +717,11 @@ mod tests {
     #[test]
     fn drop_after_shutdown_test_with_multiple_providers() {
         let shutdown_count = Arc::new(AtomicU32::new(0));
-        let flush_called = Arc::new(AtomicBool::new(false));
 
         // Create a shared TracerProviderInner and use it across multiple providers
         let shared_inner = Arc::new(TracerProviderInner {
             processors: vec![Box::new(CountingShutdownProcessor::new(
                 shutdown_count.clone(),
-                flush_called.clone(),
             ))],
             config: Config::default(),
             is_shutdown: AtomicBool::new(false),
@@ -754,6 +746,7 @@ mod tests {
             // TracerProvider2 should observe the shutdown state but not trigger another shutdown
             let shutdown_result2 = tracer_provider2.shutdown();
             assert!(shutdown_result2.is_err());
+            assert_eq!(shutdown_count.load(Ordering::SeqCst), 1);
 
             // Both tracer providers will be dropped at the end of this scope
         }
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -134,7 +134,7 @@ impl Drop for TracerProviderInner {
             let _ = self.shutdown(); // errors are handled within shutdown
         } else {
             otel_debug!(
-                name: "TracerProviderProvider.Drop.AlreadyShutdown"
+                name: "TracerProvider.Drop.AlreadyShutdown"
             );
         }
     }
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -13,8 +13,8 @@
 ///
 /// ## Cloning and Shutdown
 ///
-/// The `TracerProvider` is designed to be lightweight and clonable. Cloning a `TracerProvider`
-/// creates a new reference to the same provider, not a new instance. Dropping the last reference
+/// The `TracerProvider` is designed to be clonable. Cloning a `TracerProvider`  creates a
+/// new reference to the same provider, not a new instance. Dropping the last reference
 /// to the `TracerProvider` will automatically trigger its shutdown. During shutdown, the provider
 /// will flush all remaining spans, ensuring they are passed to the configured processors.
 /// Users can also manually trigger shutdown using the [`shutdown`](TracerProvider::shutdown)
@@ -136,7 +136,7 @@ impl Drop for TracerProviderInner {
 
 /// Creator and registry of named [`Tracer`] instances.
 ///
-/// `TracerProvider` is a lightweight container holding pointers to `SpanProcessor` and other components.
+/// `TracerProvider` is a container holding pointers to `SpanProcessor` and other components.
 /// Cloning a `TracerProvider` instance and dropping it will not stop span processing. To stop span processing, users
 /// must either call the `shutdown` method explicitly or allow the last reference to the `TracerProvider`
 /// to be dropped. When the last reference is dropped, the shutdown process will be automatically triggered
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -36,8 +36,8 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider 
             span_limits: SpanLimits::default(),
             resource: Cow::Owned(Resource::empty()),
         },
+        is_shutdown: AtomicBool::new(true),
     }),
-    is_shutdown: Arc::new(AtomicBool::new(true)),
 });
 
 /// TracerProvider inner type
@@ -45,13 +45,20 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider 
 pub(crate) struct TracerProviderInner {
     processors: Vec<Box<dyn SpanProcessor>>,
     config: crate::trace::Config,
+    is_shutdown: AtomicBool,
 }
 
 impl Drop for TracerProviderInner {
     fn drop(&mut self) {
-        for processor in &mut self.processors {
-            if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+        if self
+            .is_shutdown
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .is_ok()
+        {
+            for processor in &self.processors {
+                if let Err(err) = processor.shutdown() {
+                    global::handle_error(err);
+                }
             }
         }
     }
@@ -65,7 +72,6 @@ impl Drop for TracerProviderInner {
 #[derive(Clone, Debug)]
 pub struct TracerProvider {
     inner: Arc<TracerProviderInner>,
-    is_shutdown: Arc<AtomicBool>,
 }
 
 impl Default for TracerProvider {
@@ -79,7 +85,6 @@ impl TracerProvider {
     pub(crate) fn new(inner: TracerProviderInner) -> Self {
         TracerProvider {
             inner: Arc::new(inner),
-            is_shutdown: Arc::new(AtomicBool::new(false)),
         }
     }
 
@@ -101,7 +106,7 @@ impl TracerProvider {
     /// true if the provider has been shutdown
     /// Don't start span or export spans when provider is shutdown
     pub(crate) fn is_shutdown(&self) -> bool {
-        self.is_shutdown.load(Ordering::Relaxed)
+        self.inner.is_shutdown.load(Ordering::Relaxed)
     }
 
     /// Force flush all remaining spans in span processors and return results.
@@ -153,6 +158,7 @@ impl TracerProvider {
     /// Note that shut down doesn't means the TracerProvider has dropped
     pub fn shutdown(&self) -> TraceResult<()> {
         if self
+            .inner
             .is_shutdown
             .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
             .is_ok()
@@ -215,7 +221,7 @@ impl opentelemetry::trace::TracerProvider for TracerProvider {
     }
 
     fn library_tracer(&self, library: Arc<InstrumentationLibrary>) -> Self::Tracer {
-        if self.is_shutdown.load(Ordering::Relaxed) {
+        if self.inner.is_shutdown.load(Ordering::Relaxed) {
             return Tracer::new(library, NOOP_TRACER_PROVIDER.clone());
         }
         Tracer::new(library, self.clone())
@@ -292,7 +298,12 @@ impl Builder {
             p.set_resource(config.resource.as_ref());
         }
 
-        TracerProvider::new(TracerProviderInner { processors, config })
+        let is_shutdown = AtomicBool::new(false);
+        TracerProvider::new(TracerProviderInner {
+            processors,
+            config,
+            is_shutdown,
+        })
     }
 }
 
@@ -391,6 +402,7 @@ mod tests {
                 Box::from(TestSpanProcessor::new(false)),
             ],
             config: Default::default(),
+            is_shutdown: AtomicBool::new(false),
         });
 
         let results = tracer_provider.force_flush();
@@ -534,6 +546,7 @@ mod tests {
         let tracer_provider = super::TracerProvider::new(TracerProviderInner {
             processors: vec![Box::from(processor)],
             config: Default::default(),
+            is_shutdown: AtomicBool::new(false),
         });
 
         let test_tracer_1 = tracer_provider.tracer("test1");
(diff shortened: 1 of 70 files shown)


opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -36,22 +36,29 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider
...
 impl Drop for TracerProviderInner {
     fn drop(&mut self) {
-        for processor in &mut self.processors {
-            if let Err(err) = processor.shutdown() {
-                global::handle_error(err);
+        if self
+            .is_shutdown
+            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .is_ok()
+        {
```
cijothomas, line 57: the logic should be
drop() 
{
1. check if already shutdown. if yes, do nothing and return
2. if no, set the is_shutdown to false, and then call shutdown
}

This PR is not doing 1 I think...

cijothomas: Cancel that. This should be same as https://github.com/open-telemetry/opentelemetry-rust/pull/2195/files#r1797436168
i.e Drop should just call shutdown() if not already shutdown, and let shutdown() handle the details on how to do shutdown etc.

lalitb: > check if already shutdown. if yes, do nothing and return

compare_exchage does check if already shutdown.

 > This should be same as https://github.com/open-telemetry/opentelemetry-rust/pull/2195/files#r1797436168

same comment - https://github.com/open-telemetry/opentelemetry-rust/pull/2195/files#r1797443148 :)

lalitb: Added private TracerProviderInner::Shutdown() to be called from LoggerProvider::Shutdown and  LoggerProviderInner::Drop

cijothomas (changes requested):
Lets make it similar to https://github.com/open-telemetry/opentelemetry-rust/pull/2195

lalitb:
> Lets make it similar to #2195

Done.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -36,36 +96,60 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider
...
+    }
+}
+
+impl Drop for TracerProviderInner {
+    fn drop(&mut self) {
+        if !self.is_shutdown.load(Ordering::Relaxed) {
+            let _ = self.shutdown(); // errors are handled within shutdown
+        } else {
+            otel_debug!(
+                name: "TracerProviderProvider.Drop.AlreadyShutdown"
```
utpilla, line 137:
```suggestion
                name: "TracerProvider.Drop.AlreadyShutdown"
```

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +636,69 @@ mod tests {
...
+            // Shutdown explicitly first
+            assert!(provider.shutdown().is_ok());
+
+            // Drop all providers - should not trigger another shutdown in TracerProviderInner::drop
+            drop(provider);
+            drop(provider2);
+            drop(provider3);
+
+            // Verify shutdown was called exactly once
+            assert!(assert_handle.0.is_shutdown.load(Ordering::SeqCst));
```
utpilla, line 676: How does this verify that shutdown was called only once? It looks like it's only verifying that shutdown was called (could have been called once or multiple times)

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -36,36 +96,60 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider
...
     }
 }
 
 /// Creator and registry of named [`Tracer`] instances.
 ///
-/// `TracerProvider` is lightweight container holding pointers to `SpanProcessor` and other components.
-/// Cloning and dropping them will not stop the span processing. To stop span processing, users
-/// must either call `shutdown` method explicitly, or drop every clone of `TracerProvider`.
+/// `TracerProvider` is a lightweight container holding pointers to `SpanProcessor` and other components.
+/// Cloning a `TracerProvider` instance will not stop span processing. To stop span processing, users
```
utpilla, line 146: Did you mean cloning and dropping the `TracerProvider` instance?

```suggestion
/// Cloning a `TracerProvider` instance and dropping it will not stop span processing. To stop span processing, users
```

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +636,69 @@ mod tests {
...
+            });
+
+            // Create multiple providers sharing same inner
+            let provider2 = provider.clone();
+            let provider3 = provider.clone();
+
+            // Drop providers without explicit shutdown
+            drop(provider);
+            drop(provider2);
+
```
utpilla, line 696: We should verify that shutdown was not called by asserting that `assert_handle.0.is_shutdown` is still `false`.

opentelemetry/src/trace/mod.rs
```diff
@@ -200,6 +200,10 @@ pub enum TraceError {
     #[error("Exporting timed out after {} seconds", .0.as_secs())]
     ExportTimedOut(time::Duration),
 
+    /// already shutdown error
+    #[error("{0} already shutdown")]
+    AlreadyShutdown(String),
```
utpilla, line 205: Do we expect to use this variant for anything other than `TracerProvider`?

opentelemetry-sdk/src/trace/provider.rs
lalitb: Good point. I think I should be using [CountingShutdownProcessor](https://github.com/open-telemetry/opentelemetry-rust/blob/8d84a76b8f8aaa0d625ab67a16320680c276f6b5/opentelemetry-sdk/src/logs/log_emitter.rs#L764) which was added in #2195.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,73 @@
...
+/// are processed or exported after shutdown.
+///
+/// ## Span Processing and Force Flush
+///
+/// The `TracerProvider` manages the lifecycle of span processors, which are responsible for
+/// collecting, processing, and exporting spans. To ensure all spans are processed before shutdown,
+/// users can call the [`force_flush`](TracerProvider::force_flush) method at any time to trigger
+/// an immediate flush of all pending spans for **batch processors** to the exporters. Note that
+/// calling [`force_flush`](TracerProvider::force_flush) is optional before shutdown, as `shutdown`
+/// will automatically trigger a flush for batch processors, but not for simple processors.
```
cijothomas, line 36:
```suggestion
/// will automatically trigger a flush for batch processors, but not for simple processors.
```

Flush is triggered for all processors, irrespective of batch vs simple or something else. It is upto processor to decide if it is a no-op or not.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,73 @@
...
+/// operations on its associated `Tracer` instances will result in no-ops, ensuring that no spans
+/// are processed or exported after shutdown.
+///
+/// ## Span Processing and Force Flush
+///
+/// The `TracerProvider` manages the lifecycle of span processors, which are responsible for
+/// collecting, processing, and exporting spans. To ensure all spans are processed before shutdown,
+/// users can call the [`force_flush`](TracerProvider::force_flush) method at any time to trigger
+/// an immediate flush of all pending spans for **batch processors** to the exporters. Note that
+/// calling [`force_flush`](TracerProvider::force_flush) is optional before shutdown, as `shutdown`
```
cijothomas, line 35: not sure if we mentioning "optional" for force_flush is a good idea.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,73 @@
...
+///     provider
+/// }
+///
+/// fn main() {
+///     let provider = init_tracing();
+///
+///     // create spans...
+///
+///     // Flush all spans before shutdown (optional for batch processors)
+///     for result in provider.force_flush() {
```
cijothomas, line 59: why we need to show force_flush in here?

opentelemetry-sdk/src/trace/provider.rs
lalitb: Have made this use CountingShutdownProcessor now.

opentelemetry-sdk/src/trace/provider.rs
lalitb: please have a look again, the tests are more focused now.

opentelemetry/src/trace/mod.rs
lalitb: Yes, thought to use them for the processors and exporters too. But I believe we can customize it later if required. For now, made it static for TracerProvider.

opentelemetry-sdk/src/trace/provider.rs
lalitb: I have removed the mentioning of batch and simple processors here, kept is generic

opentelemetry-sdk/src/trace/provider.rs
lalitb: Yes removed now.

opentelemetry-sdk/src/trace/provider.rs
cijothomas: thanks! In future prs, we can do another sweep and make sure we don't ever require/recommend users to do force_flush. Explicity shutdown or implicit via drop is sufficient.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,67 @@
...
+/// ## Tracer Creation
+///
+/// New [`Tracer`] instances are always created through a `TracerProvider`. These `Tracer`s share
+/// a common configuration, which includes the [`Resource`], span processors, sampling strategies,
+/// and span limits. This avoids the need for each `Tracer` to maintain its own version of these
+/// configurations, ensuring uniform behavior across all instances.
+///
+/// ## Cloning and Shutdown
+///
+/// The `TracerProvider` is designed to be lightweight and clonable. Cloning a `TracerProvider`
```
cijothomas, line 16: I don't think TracerProvider is lightweight. It is pretty heavy, and we expect user to create it only once. It is correct to mention cloning is cheap as it is just creating a new ref.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,67 @@
...
+/// creates a new reference to the same provider, not a new instance. Dropping the last reference
+/// to the `TracerProvider` will automatically trigger its shutdown. During shutdown, the provider
+/// will flush all remaining spans, ensuring they are exported to the configured exporters.
```
cijothomas, line 19:
```suggestion
/// will flush all remaining spans, ensuring they are passed to the configured processors.
```

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,67 @@
...
+///
+/// Once shut down, the `TracerProvider` transitions into a disabled state. In this state, further
+/// operations on its associated `Tracer` instances will result in no-ops, ensuring that no spans
+/// are processed or exported after shutdown.
+///
+/// ## Span Processing and Force Flush
+///
+/// The `TracerProvider` manages the lifecycle of span processors, which are responsible for
+/// collecting, processing, and exporting spans. To ensure all spans are processed before shutdown,
+/// users can call the [`force_flush`](TracerProvider::force_flush) method at any time to trigger
```
cijothomas, line 31: lets remove force_flush mention here. I have seen many users doing force_flush in their code (and block their threads).. Not sure why, but lets make sure official docs don't recommend it.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -36,36 +90,60 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider
...
+        }
     }
 }
 
 /// Creator and registry of named [`Tracer`] instances.
 ///
-/// `TracerProvider` is lightweight container holding pointers to `SpanProcessor` and other components.
-/// Cloning and dropping them will not stop the span processing. To stop span processing, users
-/// must either call `shutdown` method explicitly, or drop every clone of `TracerProvider`.
+/// `TracerProvider` is a lightweight container holding pointers to `SpanProcessor` and other components.
```
cijothomas, line 139: not introduced in this PR, but advertising TracerProvider as lightweight is incorrect, and can lead to users repeatedly creating them, instead of doing it once.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +631,136 @@ mod tests {
...
+        assert!(noop_tracer.provider().is_shutdown());
 
         // existing tracer becomes noops after shutdown
         let _ = test_tracer_1.start("test");
         assert!(assert_handle.started_span_count(2));
     }
+
+    #[derive(Debug)]
+    struct CountingShutdownProcessor {
+        shutdown_count: Arc<Mutex<i32>>,
```
cijothomas, line 648: nit: Atomics maybe easier here

opentelemetry-sdk/src/trace/provider.rs
lalitb: Have reworded it so that it doesn't look as recommendation. I think it's better to at-least document since we provide it.
```
/// ## Span Processing and Force Flush
///
/// The `TracerProvider` manages the lifecycle of span processors, which are responsible for
/// collecting, processing, and exporting spans. The [`force_flush`](TracerProvider::force_flush) method
/// invoked at any time will trigger an immediate flush of all pending spans (if any) to the exporters.
/// This will block the user thread till all the spans are passed to exporters
```

opentelemetry-sdk/src/trace/provider.rs
lalitb: Done

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -1,13 +1,67 @@
...
+/// a common configuration, which includes the [`Resource`], span processors, sampling strategies,
+/// and span limits. This avoids the need for each `Tracer` to maintain its own version of these
+/// configurations, ensuring uniform behavior across all instances.
+///
+/// ## Cloning and Shutdown
+///
+/// The `TracerProvider` is designed to be lightweight and clonable. Cloning a `TracerProvider`
+/// creates a new reference to the same provider, not a new instance. Dropping the last reference
+/// to the `TracerProvider` will automatically trigger its shutdown. During shutdown, the provider
+/// will flush all remaining spans, ensuring they are passed to the configured exporters.
```
cijothomas, line 19:
```suggestion
/// will flush all remaining spans, ensuring they are passed to the configured processors.
```

TracerProvider, LoggerProvider does not do anything with Exporters, so the above wording is more accurate. 

we still need some mention of exporters, as users don't care about this distinction.

opentelemetry/src/trace/mod.rs
```diff
@@ -200,6 +200,10 @@ pub enum TraceError {
     #[error("Exporting timed out after {} seconds", .0.as_secs())]
     ExportTimedOut(time::Duration),
 
+    /// already shutdown error
+    #[error("TracerProvider already shutdown")]
+    AlreadyShutdown,
```
utpilla, line 205: Should we rename this to be more specific?

```suggestion
    TracerProviderAlreadyShutdown,
```

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +630,135 @@ mod tests {
...
+            let shutdown_result = tracer_provider1.shutdown();
+            assert!(shutdown_result.is_ok());
+
+            // Verify that shutdown was called exactly once
+            assert_eq!(shutdown_count.load(Ordering::SeqCst), 1);
+
+            // TracerProvider2 should observe the shutdown state but not trigger another shutdown
+            let shutdown_result2 = tracer_provider2.shutdown();
+            assert!(shutdown_result2.is_err());
+
```
utpilla, line 757: Add `assert_eq!(shutdown_count.load(Ordering::SeqCst), 1);` here as well to ensure that explicitly calling shutdown on an already shutdown TracerProvider doesn't call shutdown again.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +630,135 @@ mod tests {
...
 
         // existing tracer becomes noops after shutdown
         let _ = test_tracer_1.start("test");
         assert!(assert_handle.started_span_count(2));
     }
+
+    #[derive(Debug)]
+    struct CountingShutdownProcessor {
+        shutdown_count: Arc<AtomicU32>,
+        flush_called: Arc<AtomicBool>,
```
utpilla, line 648: Would you be adding tests to check `flush_called` later?

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +630,135 @@ mod tests {
...
+
+            // TracerProvider2 should observe the shutdown state but not trigger another shutdown
+            let shutdown_result2 = tracer_provider2.shutdown();
+            assert!(shutdown_result2.is_err());
+
+            // Both tracer providers will be dropped at the end of this scope
+        }
+
+        // Verify that shutdown was only called once, even after drop
+        assert_eq!(shutdown_count.load(Ordering::SeqCst), 1);
```
utpilla, line 762: Also verify that `force_flush` was not called similar to the previous test.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -36,36 +90,60 @@ static NOOP_TRACER_PROVIDER: Lazy<TracerProvider> = Lazy::new(|| TracerProvider
             span_limits: SpanLimits::default(),
             resource: Cow::Owned(Resource::empty()),
         },
+        is_shutdown: AtomicBool::new(true),
```
utpilla, line 93: Why have we initialized the no-op tracer provider as an already shutdown provider?

I know this wouldn't make much difference in functionality but semantically it would be weird if I call shutdown on the global provider and get an error saying it has already been shut down.

lalitb: global [shutdown_tracer_provider ](https://github.com/lalitb/opentelemetry-rust/blob/ea4b5e4d992610bb23a059f05253d0c47e262015/opentelemetry/src/global/trace.rs#L421) naming is bit confusing. It doesn't invoke shutdown, only decrements the TracerProviderInner reference. Once all the tracers are dropped, the shutdown will get invoked through `Drop`.

And the user will never get hold of NOOP_TRACER_PROVIDER instance. It's just internally used to create the tracer instance after shutdown.

opentelemetry-sdk/src/trace/provider.rs
lalitb: Good point. Have update the comment to remove the lightweight reference.

opentelemetry-sdk/src/trace/provider.rs
lalitb: Have updated the comment to remove the lightweight reference.

opentelemetry-sdk/src/trace/provider.rs
lalitb: Not in this PR, as this is only for redundant shutdown.

opentelemetry-sdk/src/trace/provider.rs
lalitb: Good point. Done.

opentelemetry-sdk/src/trace/provider.rs
lalitb: Have removed the flush_called from the processor as we are not using it. Thanks for pointing out.

opentelemetry/src/trace/mod.rs
lalitb: done.

utpilla: It doesn't show the change.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -554,14 +630,128 @@ mod tests {
...
+        fn shutdown(&self) -> TraceResult<()> {
+            self.shutdown_count.fetch_add(1, Ordering::SeqCst);
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn drop_test_with_multiple_providers() {
+        let shutdown_count = Arc::new(AtomicU32::new(0));
+        let flush_called = Arc::new(AtomicBool::new(false));
```
utpilla, line 678: Remove `flush_called` from here as well.

opentelemetry-sdk/src/trace/provider.rs
utpilla: Okay so then we could initialize no-op tracer provider with `is_shutdown` as `false`?

utpilla (approved):

opentelemetry/src/trace/mod.rs
lalitb: oh missed pushing one of the files, done now.

opentelemetry-sdk/src/trace/provider.rs
lalitb: oh yes, done. thanks.

opentelemetry-sdk/src/trace/provider.rs
lalitb: > Okay so then we could initialize no-op tracer provider with is_shutdown as false?

Making it `false` fails this test - https://github.com/open-telemetry/opentelemetry-rust/blob/ea4b5e4d992610bb23a059f05253d0c47e262015/opentelemetry-sdk/src/trace/provider.rs#L531
And this test make me realise user can access it through tracer.provider(), and can try to invoke shutdown on it. But feel free to raise an issue if there are any improvements here.

cijothomas (approved):

Issue #1745, opened 2024-05-12
[Bug]: `opentelemetry_appender_tracing` does not work with blocking otlp exporter


cortopy:
### What happened?

The basic [example](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/examples/basic.rs) for `opentelemetry_appender_tracing` is working fine for me.

The same example also works if I use an OTLP exporter with tonic and Tokio runtime within an async context.

However, if I modify the example for `opentelemetry_appender_tracing` to use an OTLP exporter with `reqwest::blocking`, the program enters into a deadlock.

This is the code I'm running:
```rust
use opentelemetry::KeyValue;
use opentelemetry_appender_tracing::layer;
use opentelemetry_sdk::{
    logs::{Config, LoggerProvider},
    Resource,
};
use tracing::error;
use tracing_subscriber::prelude::*;

fn main() {
    // This exporter is built with reqwest:blocking and will hang
    let otl_exporter = opentelemetry_otlp::new_exporter()
        .http()
        .build_log_exporter()
        .unwrap();
    let provider: LoggerProvider = LoggerProvider::builder()
        .with_config(
            Config::default().with_resource(Resource::new(vec![KeyValue::new(
                "service.name",
                "log-appender-tracing-example",
            )])),
        )
        .with_simple_exporter(otl_exporter)
        .build();
    let layer = layer::OpenTelemetryTracingBridge::new(&provider);
    tracing_subscriber::registry().with(layer).init();

    error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
    drop(provider);
}

```

The opentelemetry-otlp crate is used with both reqwest blocking and tonic within an internal config crate:

```
opentelemetry-otlp = { version = "0.15.0", features = ["logs", "grpc-tonic", "http-proto", "reqwest-blocking-client", "serialize", "metrics", "trace"] }
```

### API Version

I don't know

### SDK Version

0.22.1

### What Exporters are you seeing the problem on?

OTLP

cijothomas:
I'll take a look today.

cijothomas:
I was able to easily repro this, but unsure why is this deadlocked, despite using reqwest::blocking client. Will investigate more and get back. There are several errors related to async runtimes and we are actively investigating a proper fix to take care of all of them, including this issue!.

cijothomas:
[...]

Issue #2071, opened 2024-08-31
[Bug]: Assigning `opentelemetry_otlp` layer causes hang when using `tokio::spawn`.


drwhut:
### What happened?

Hey! I'm having an issue when trying to use an `opentelemetry_otlp` layer in a `tracing_subscriber`, when spawning another task with `tokio::spawn`.

I've managed to make a minimal reproduction example. Here's the `Cargo.toml` that I have currently:
```toml
[dependencies]
opentelemetry = "0.24"
opentelemetry-otlp = "0.17"
tokio = { version = "1", features = ["full", "test-util"] }
tracing = "0.1"
tracing-opentelemetry = "0.25"
tracing-subscriber = { version = "0.3", features = ["env-filter"] }
```

And this is the code that hangs:
```rust
use opentelemetry::trace::TracerProvider;
use tracing_subscriber::filter::{EnvFilter, LevelFilter};
use tracing_subscriber::{fmt, prelude::*};
use tracing::info;

#[tracing::instrument]
async fn some_function() {
    info!("some_function()");
    inner_function().await;
    info!("inner function done.");
}

#[tracing::instrument]
async fn inner_function() {
    info!("inner function!");
}

#[tokio::main]
async fn main() {
    let fmt_layer = fmt::layer().with_target(false);

    let filter_layer = EnvFilter::builder()
        .with_default_directive(LevelFilter::INFO.into())
        .from_env_lossy();

    let otlp_exporter = opentelemetry_otlp::new_exporter().tonic();

    let tracer = opentelemetry_otlp::new_pipeline()
        .tracing()
        .with_exporter(otlp_exporter)
        .install_simple()
        .unwrap()
        .tracer("test_server");

    let otel_layer = tracing_opentelemetry::layer().with_tracer(tracer);

    tracing_subscriber::registry()
        .with(filter_layer)
        .with(fmt_layer)
        .with(otel_layer)
        .init();

    info!("calling...");
    let handle = tokio::spawn(some_function());
    handle.await.unwrap();
    info!("done!");
}

```

If I run the code as is, this is the output that I get:
```
2024-08-29T15:13:10.063000Z  INFO calling...
2024-08-29T15:13:10.063108Z  INFO some_function: some_function()
2024-08-29T15:13:10.063138Z  INFO some_function:inner_function: inner function!
```

[...]

PR #2199, opened 2024-10-12
Add tests to confirm known deadlock issues

cijothomas:
Adding test to repro https://github.com/open-telemetry/opentelemetry-rust/issues/1745

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -208,17 +208,20 @@ const fn severity_of_level(level: &Level) -> Severity {
 #[cfg(test)]
 mod tests {
     use crate::layer;
-    use opentelemetry::logs::Severity;
+    use async_trait::async_trait;
+    use opentelemetry::logs::{LogResult, Severity};
     use opentelemetry::trace::TracerProvider as _;
     use opentelemetry::trace::{TraceContextExt, TraceFlags, Tracer};
     use opentelemetry::{logs::AnyValue, Key};
+    use opentelemetry_sdk::export::logs::{LogBatch, LogExporter};
     use opentelemetry_sdk::logs::{LogRecord, LoggerProvider};
     use opentelemetry_sdk::testing::logs::InMemoryLogsExporter;
     use opentelemetry_sdk::trace;
     use opentelemetry_sdk::trace::{Sampler, TracerProvider};
-    use tracing::error;
+    use tracing::{error, warn};
     use tracing_subscriber::prelude::__tracing_subscriber_SubscriberExt;
-    use tracing_subscriber::Layer;
+    use tracing_subscriber::util::SubscriberInitExt;
+    use tracing_subscriber::{EnvFilter, Layer};
 
     pub fn attributes_contains(log_record: &LogRecord, key: &Key, value: &AnyValue) -> bool {
         log_record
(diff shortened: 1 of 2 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -238,6 +241,70 @@ mod tests {
...
+        tracing_subscriber::registry()
+            .with(filter)
+            .with(layer)
+            .init();
+        warn!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[ignore = "While this test runs fine, this uses global subscriber and does not play well with other tests."]
+    async fn batch_processor_deadlock() {
```
lalitb, line 296: should this be `batch_processor_no_deadlock` ?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -238,6 +241,70 @@ mod tests {
...
     // cargo test --features=testing
+
+    #[derive(Clone, Debug, Default)]
+    struct ReentrantLogExporter;
+
+    #[async_trait]
+    impl LogExporter for ReentrantLogExporter {
+        async fn export(&mut self, _batch: LogBatch<'_>) -> LogResult<()> {
+            // This will cause a deadlock as the export itself creates a log
+            // while still within the lock of the SimpleLogProcessor.
```
lalitb, line 252: wondering if we should bring back the message queue back for simple log processor ?

lalitb (approved):
Thanks for the tests.  nit comments, with the lint fix - should be good to mege.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: that might mitigate the issue temporarily, but we need to solve it via context based suppression I think.

lalitb: In context-based suppression, there may still be cases where we want to loop back the internal events at least once. This could potentially result in a deadlock situation.

cijothomas: Not sure I follow that. Maybe add a unit test to cover the scenario, so its easy to track.

lalitb: I meant - event with the suppression logic in place, we may want the internal errors to loop-back to the pipeline once, hoping that the error is transient  and doesn't cause infinite loop. In case of non-transient errors, the subsequent error is suppressed.  But this can be something discussed when we look into suppression logic.

Not related to this, but moving to message queue will fix hang  issue - https://github.com/open-telemetry/opentelemetry-rust/issues/2071#issuecomment-2327358510

cijothomas: > we may want the internal errors to loop-back to the pipeline once, hoping that the error is transient and doesn't cause infinite loop

Can you share a scenario where this is needed? I am not sure how often it is feasible for a component to determine if an error is transient or not.

> Not related to this, but moving to message queue will also fix hang issue - https://github.com/open-telemetry/opentelemetry-rust/issues/2071#issuecomment-2327358510

Yes, likely. But this needs a full redesign. I don't think adding a queue in the middle helps all cases.

PR #2204, opened 2024-10-12
Chore/metrics advanced

pitoniak32:
Issue That prompted these changes: https://github.com/open-telemetry/opentelemetry-rust/issues/1060

References:
- mention of wanting alternate Temporality example: https://github.com/open-telemetry/opentelemetry-rust/issues/1060#issuecomment-1834421336
- mention of duplication in benches: https://github.com/open-telemetry/opentelemetry-rust/pull/1057#discussion_r1205798111

## Changes
`DeltaTemporalitySelector`:
- Moving the `DeltaTemporalitySelector` to same file as `DefaultTemporalitySelector`.
- make the visibility of the struct `pub` so its usable for pipeline setup.

Example:
- Update the advanced metrics example to use `DeltaTemporalitySelector`.

### DIFF
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -1,6 +1,7 @@
 use opentelemetry::global;
 use opentelemetry::Key;
 use opentelemetry::KeyValue;
+use opentelemetry_sdk::metrics::reader::DeltaTemporalitySelector;
 use opentelemetry_sdk::metrics::{
     Aggregation, Instrument, PeriodicReader, SdkMeterProvider, Stream,
 };
@@ -44,7 +45,11 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
         }
     };
 
-    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default().build();
+    // Build exporter using Delta Temporality.
+    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default()
+        .with_temporality_selector(DeltaTemporalitySelector::new())
+        .build();
+
     let reader = PeriodicReader::builder(exporter, runtime::Tokio).build();
     let provider = SdkMeterProvider::builder()
         .with_reader(reader)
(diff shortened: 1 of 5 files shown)


opentelemetry-otlp/src/metric.rs
```diff
@@ -169,7 +169,7 @@ where
     ///
     /// [exporter-docs]: https://github.com/open-telemetry/opentelemetry-specification/blob/a1c13d59bb7d0fb086df2b3e1eaec9df9efef6cc/specification/metrics/sdk_exporters/otlp.md#additional-configuration
     pub fn with_delta_temporality(self) -> Self {
-        self.with_temporality_selector(DeltaTemporalitySelector)
+        self.with_temporality_selector(DeltaTemporalitySelector::new())
```
cijothomas, line 172: the future of these methods in OTLP crate is unknown based on https://github.com/open-telemetry/opentelemetry-rust/pull/1811
However, this change itself is fine.

examples/metrics-advanced/src/main.rs
```diff
@@ -44,7 +45,11 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
         }
     };
 
-    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default().build();
+    // Build exporter using Delta Temporality.
```
cijothomas, line 48: While the concept is somewhat advanced, it is a common ask. Could you make this in the metrics-basic itself, as a code comment, so users can pick between delta vs cumulative by removing commented out code?

(It is possible that we'll have different ways to configure this in the future, but this change is still very good)

cijothomas (approved):
LGTM.
If you can make the change in metrics-basic, that'd be sweet. It is okay to do it in a follow up as well.

pitoniak32:
> LGTM.
> 
> If you can make the change in metrics-basic, that'd be sweet. It is okay to do it in a follow up as well.

Awesome! I will make the requested changes for this tomorrow for the example.

PR #2210, opened 2024-10-14
[Metrics-API] Update asynchronous instruments to not allow calling `observe` outside of a callback

utpilla:
All the asynchronous instruments implement the trait `AsyncInstrument` which allows users to call `observe` method outside of the callback
```rust
let observable_counter = meter
        .with_callback(|observer| {
            observer.observe(
                100,
                &[
                ],
            )
        })
        .init();

observable_counter.observe(500, &[]); // This should not be allowed
```

## Changes
- This PR removes the implementation of `AsyncInstrument` for each of the asynchronous instruments to not allow calling `observe` outside of a callback (as a side benefit, we also reduce the public API surface)
- That also means that we can completely remove `AsyncInstrument` from each of asynchronous instruments as it was only used for implementation of itself

### DIFF
--- a/opentelemetry/src/metrics/instruments/gauge.rs
+++ b/opentelemetry/src/metrics/instruments/gauge.rs
@@ -1,4 +1,4 @@
-use crate::{metrics::AsyncInstrument, KeyValue};
+use crate::KeyValue;
 use core::fmt;
 use std::sync::Arc;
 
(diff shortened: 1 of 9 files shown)


cijothomas (approved):
Lets mention in changelog as this is breaking (though unlikely to affect anyone in practice)

opentelemetry/src/metrics/instruments/gauge.rs
```diff
@@ -33,7 +33,9 @@ impl<T> Gauge<T> {
 /// An async instrument that records independent readings.
 #[derive(Clone)]
 #[non_exhaustive]
-pub struct ObservableGauge<T>(Arc<dyn AsyncInstrument<T>>);
+pub struct ObservableGauge<T> {
+    _marker: std::marker::PhantomData<T>,
```
ThomsonTan, line 37: nit: Declare the full type name of `PhantomData` in `use`?

lalitb (approved):

PR #2211, opened 2024-10-14
[Metrics-API] Update asynchronous instruments to only allow one callback

utpilla:
## Changes
- This PR updates the `AynshronousInstrumentBuilder` and related code to only allow for one callback
- If a user provides multiple callbacks for an instrument by calling `with_callback` multiple times, we select the last provided `with_callback` for the instrument


lalitb:
Is this for initial release, as the specs does allow specifying the multiple callbacks for instruments ?

cijothomas (changes requested):
Removing this capability is not yielding a significant simplification in overall design, so I think its best to leave this capability untouched and not break any existing users.

PR #2215, opened 2024-10-16
[Metrics API] Remove unnecessary auto trait implementations

utpilla:
## Changes
- Remove unnecessary `Ord` and `PartialOrd` implementation for `KeyValue`
- This was only used for sorting a vec of `KeyValue` pairs in the SDK which could be done using `sort_unstable_by` method anyway

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -104,7 +104,7 @@ fn calculate_hash(values: &[KeyValue]) -> u64 {
 
 impl AttributeSet {
     fn new(mut values: Vec<KeyValue>) -> Self {
-        values.sort_unstable();
+        values.sort_unstable_by(|a, b| a.key.cmp(&b.key));
         let hash = calculate_hash(&values);
         AttributeSet(values, hash)
     }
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -104,7 +104,7 @@ fn calculate_hash(values: &[KeyValue]) -> u64 {
 
 impl AttributeSet {
     fn new(mut values: Vec<KeyValue>) -> Self {
-        values.sort_unstable();
+        values.sort_unstable_by(|a, b| a.key.cmp(&b.key));
```
lalitb, line 107: will this be a breaking change if the user is sorting the KeyValue using the sort/sort_unsable before calling the metrics API?

utpilla: Yes, but I don't expect anyone to be doing that. We don't ask to sort the metric attributes before recording a measurement.

lalitb: Yes agree. Good to have changlog entry just in case someone is doing that :)

lalitb (approved):
LGTM with changelog

opentelemetry-sdk/src/metrics/mod.rs
utpilla: Updated.

Issue #2099, opened 2024-09-11
Cleanup Temporality Selector


cijothomas:
Similar to https://github.com/open-telemetry/opentelemetry-rust/pull/2085, we should remove this trait/implementations from public API. The exporters can offer https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/metric.rs#L172-L180 `with_delta` method to switch temporality to delta from the default cumulative. This reduces a lot of unnecessary code in metrics sdk without significantly losing any functionality. OTLP Exporter only allows 3 options : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md

(We can ofcourse `add with_lowmemory_temporality` if needed)

cijothomas:
Take care by https://github.com/open-telemetry/opentelemetry-rust/pull/2217 closing!

PR #2217, opened 2024-10-16
example: add temporality selector option to metrics-basic

pitoniak32:
Follow up PR from this comment: https://github.com/open-telemetry/opentelemetry-rust/pull/2204#pullrequestreview-2364955914

## Changes

Add Temporality selector option to metrics-basic example


examples/metrics-basic/src/main.rs
```diff
@@ -6,7 +6,16 @@ use std::error::Error;
 use std::vec;
 
 fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
-    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default().build();
+    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default()
+        // Build exporter using Default (Cumulative) Temporality Selector.
+        .with_temporality_selector(
```
cijothomas, line 11: @utpilla suggested this can be further simplified. An enum should be better here, delta/cumulative/lowmemory being the enum variants.

@pitoniak32 Do you have bandwidth to help make this change? Hoping to make it done by next release, so as to avoid breaking users after that release.

pitoniak32: Yeah for sure, I can take a look tomorrow morning. When would the next release be? Also, Is there an open issue for this currently?

cijothomas: Next release is ~Oct 22-25.
No issue with all the details, https://github.com/open-telemetry/opentelemetry-rust/issues/2099 is the closest. Hope that is good enough to get started.

pitoniak32: I took a first go at this before work this morning. I will take another look and try and clean things up later today.

examples/metrics-basic/src/main.rs
```diff
@@ -6,7 +6,12 @@ use std::error::Error;
 use std::vec;
 
 fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
-    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default().build();
+    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default()
+        // Build exporter using Default (Cumulative) Temporality Selector.
+        .with_default_temporality_selector()
```
utpilla, line 11: I think a simpler interface would be to just have one method `with_temporality` which takes in an enum value. No need to have separate methods for delta, default, cumulative. Something like this:
```rust
.with_temporality(Temporality::Delta)
// or 
.with_temporality(Temporality::Cumulative)
```

If the user doesn't explicitly provide the temporality information using `with_temporality` we just use the default.

cijothomas: +1

pitoniak32: I was thinking along those lines also, this would prevent the use of the `LowMemory` option, if that's a fair trade off I can go ahead with this

cijothomas: we don't need to prevent LowMemory. It can be added to the Enum. Either right away or a future addition

pitoniak32: I switched over to using the `Temporality` enum directly

examples/metrics-basic/src/main.rs
```diff
@@ -1,12 +1,17 @@
...
+use opentelemetry_sdk::metrics::{data::Temporality, PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::{runtime, Resource};
 use std::error::Error;
 use std::vec;
 
 fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
-    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default().build();
+    let exporter = opentelemetry_stdout::MetricsExporterBuilder::default()
+        // Build exporter using Default (Cumulative) Temporality.
+        .with_temporality(Temporality::default())
```
utpilla, line 11: For a better user experience, let's not put any examples that explicitly use `Temporality::default()`. It should be enough to have a comment mention that : by default, `Cumulative` temporality will be used.

We don't want to give the impression that one needs to explicitly add `.with_temporality(Temporality::default())` to use `Cumulative` aggregation.

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -37,6 +39,35 @@ pub enum InstrumentKind {
     ObservableGauge,
 }
 
+impl InstrumentKind {
+    /// Select the [Temporality] preference based on [InstrumentKind]
+    ///
+    /// [exporter-docs]: https://github.com/open-telemetry/opentelemetry-specification/blob/a1c13d59bb7d0fb086df2b3e1eaec9df9efef6cc/specification/metrics/sdk_exporters/otlp.md#additional-configuration
+    pub fn temporality_preference(&self, temporality: Temporality) -> Temporality {
```
utpilla, line 46: This method for internal use only.

```suggestion
    pub(crate) fn temporality_preference(&self, temporality: Temporality) -> Temporality {
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -945,7 +944,11 @@ mod tests {
 
         assert_eq!(sum.data_points.len(), 1, "Expected only one data point");
         assert!(!sum.is_monotonic, "Should not produce monotonic.");
-        assert_eq!(sum.temporality, Temporality::Delta, "Should produce Delta");
+        assert_eq!(
```
utpilla, line 947: Rename this test method to `no_attr_up_down_counter` now that its behavior has changed.

opentelemetry-otlp/src/lib.rs
```diff
@@ -183,7 +183,7 @@
 //!         .with_resource(Resource::new(vec![KeyValue::new("service.name", "example")]))
 //!         .with_period(Duration::from_secs(3))
 //!         .with_timeout(Duration::from_secs(10))
-//!         .with_temporality_selector(DefaultTemporalitySelector::new())
+//!         .with_temporality(Temporality::default())
```
utpilla, line 186: Following the same thing as https://github.com/open-telemetry/opentelemetry-rust/pull/2217#discussion_r1805266835, we could just remove this.

utpilla (approved):
Looks good overall!

The public APIs exposed on `MetricReader` and `PushMetricExporter` traits might be updated in the future. That's mainly because it could be beneficial to push the "temporality + instrument kind" configuration solely to the exporter. That discussion and decision can happen independent of this PR.

pitoniak32:
> Looks good overall!
> 
> 
> 
> The public APIs exposed on `MetricReader` and `PushMetricExporter` traits might be updated in the future. That's mainly because it could be beneficial to push the "temporality + instrument kind" configuration solely to the exporter. That discussion and decision can happen independent of this PR.

Awesome! I would be happy to take part in the discussion and decision around this topic when it comes up, I have been enjoying contributing to this code base and would like to take part more often :)

I will also update the CHANGELOG now that the larger concerns were addressed

opentelemetry-sdk/src/metrics/mod.rs
pitoniak32: I updated to `no_attr_up_down_counter_always_cumulative`, and updated the assertion reason to indicate why `Cumulative` would be selected when `Delta` is provided for clarity in the future.

cijothomas:
> > Looks good overall!
> > The public APIs exposed on `MetricReader` and `PushMetricExporter` traits might be updated in the future. That's mainly because it could be beneficial to push the "temporality + instrument kind" configuration solely to the exporter. That discussion and decision can happen independent of this PR.
> 
> Awesome! I would be happy to take part in the discussion and decision around this topic when it comes up, I have been enjoying contributing to this code base and would like to take part more often :)
> 
> I will also update the CHANGELOG now that the larger concerns were addressed

Thanks!
some rough ideas:
1. Make temporality selection purely a configuration on MetricReader, and keep exporters away from it.
2. Exporters may offer methods that take MeterProviderBuilder as input, and returns the MeterProviderBuilder as output to make it easy for end users to enable exporters of their choice without having to worry about temporality.
eg:
meter_provider::builder()
.configure_resource(..)
.addFooExporter();

AddFooExporter() is offered by vendors/exporter authors which will add the right metric_reader with correct temporality etc automatically.

Of course this is not a requirement, so users can add reader themselves.

Alternatively, these traits can be provided with default implementation, so exporters need to override only if they wish to.

Closely related to that is the below issue. I am not a fan of OTLPExporter exposing duplicate concepts as that of SDK, so this could use some cleanups.
https://github.com/open-telemetry/opentelemetry-rust/issues/1810

@pitoniak32 If you are up for it, please take a look at the above! Very happy to see more contributions from you.

pitoniak32:
> @pitoniak32 If you are up for it, please take a look at the above! Very happy to see more contributions from you.

For sure! I will take a look at this one after work today

PR #2218, opened 2024-10-17
Add Unit Tests for Sync and Async Log Exporters - with and without runtime - for SimpleLogProcessor

lalitb:
## Changes
There aren't many tests for SimpleLogExporter with various combinations. So added these:

**SyncLogExporter:** - A synchronous log exporter that stores log records in memory without requiring a runtime. 

- `test_simple_processor_sync_exporter_without_runtime`: Verifies log export without any runtime.
- `test_simple_processor_sync_exporter_with_runtime`:  Validates export processing in a Tokio multi-threaded runtime.
- `test_simple_processor_sync_exporter_with_current_thread_runtime`: Tests processing within a single-threaded runtime.

**AsyncLogExporter:** - An asynchronous log exporter that processes log records in a separate Tokio task, requiring a Tokio runtime.

- `test_simple_processor_async_exporter_without_runtime`: Ensures a panic occurs when attempting async export without a runtime.
- `test_simple_processor_async_exporter_with_runtime`: Validates async log export with a multi-threaded (single worker thread) Tokio runtime.
- `test_simple_processor_async_exporter_with_current_thread_runtime`: Ignored due to hanging behavior in the single-threaded runtime.
- `test_simple_processor_async_exporter_with_multi_thread_runtime_all_cores_blocked`: Ignored due to hanging behaviour with all threads blocked.
- `test_simple_processor_async_exporter_with_multi_thread_runtime` : Validates async log export with a multi-threaded (number of worker threads equal to the number of cores) Tokio runtime.

Will add similar tests for batch log exporter too.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -555,7 +555,6 @@ mod tests {
     use opentelemetry::{logs::LogResult, KeyValue};
     use std::sync::{Arc, Mutex};
     use std::time::Duration;
-    //use num_cpus;
 
     #[derive(Debug, Clone)]
     struct MockLogExporter {
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -1164,7 +1164,7 @@ mod tests {
             exporter.clone(),
         ))));
 
-        let concurrent_emit = 5; // number of worker threads + 1
+        let concurrent_emit = 4; // number of worker threads
 
         let mut handles = vec![];
         // try send `concurrent_emit` events concurrently
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -553,10 +553,8 @@ mod tests {
     use opentelemetry::InstrumentationLibrary;
     use opentelemetry::Key;
     use opentelemetry::{logs::LogResult, KeyValue};
-    use std::sync::{
-        atomic::{AtomicU32, Ordering},
-        Arc, Mutex,
-    };
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::{Arc, Mutex};
     use std::time::Duration;
 
     #[derive(Debug, Clone)]
@@ -1019,9 +1017,7 @@ mod tests {
     #[tokio::test(flavor = "multi_thread")]
     async fn test_simple_processor_sync_exporter_with_multi_thread_runtime() {
         let exporter = InMemoryLogsExporterBuilder::default().build();
-        let processor = Arc::new(Mutex::new(SimpleLogProcessor::new(Box::new(
-            exporter.clone(),
-        ))));
+        let processor = Arc::new(SimpleLogProcessor::new(Box::new(exporter.clone())));
 
         let mut handles = vec![];
         for _ in 0..10 {
@@ -1029,10 +1025,7 @@ mod tests {
             let handle = tokio::spawn(async move {
                 let mut record: LogRecord = Default::default();
                 let instrumentation: InstrumentationLibrary = Default::default();
-                processor_clone
-                    .lock()
-                    .unwrap()
-                    .emit(&mut record, &instrumentation);
+                processor_clone.emit(&mut record, &instrumentation);
             });
             handles.push(handle);
         }
@@ -1059,20 +1052,20 @@ mod tests {
 
     #[derive(Debug, Clone)]
     struct LogExporterThatRequiresTokio {
-        export_count: Arc<AtomicU32>,
+        export_count: Arc<AtomicUsize>,
     }
 
     impl LogExporterThatRequiresTokio {
         /// Creates a new instance of `LogExporterThatRequiresTokio`.
         fn new() -> Self {
             LogExporterThatRequiresTokio {
-                export_count: Arc::new(AtomicU32::new(0)),
+                export_count: Arc::new(AtomicUsize::new(0)),
             }
         }
 
         /// Returns the number of logs stored in the exporter.
         fn len(&self) -> usize {
-            self.export_count.load(Ordering::Acquire) as usize
+            self.export_count.load(Ordering::Acquire)
         }
     }
 
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -1074,14 +1074,14 @@ mod tests {
     }
 
     #[derive(Debug, Clone)]
-    struct AsyncLogExporter {
+    struct LogExporterThatRequiresTokioSpawn {
         logs: Arc<Mutex<Vec<(LogRecord, InstrumentationLibrary)>>>,
     }
 
-    impl AsyncLogExporter {
-        /// Creates a new instance of `AsyncLogExporter`.
+    impl LogExporterThatRequiresTokioSpawn {
+        /// Creates a new instance of `LogExporterThatRequiresTokioSpawn`.
         pub(crate) fn new() -> Self {
-            AsyncLogExporter {
+            LogExporterThatRequiresTokioSpawn {
                 logs: Arc::new(Mutex::new(Vec::new())),
             }
         }
@@ -1096,7 +1096,7 @@ mod tests {
     use tokio::time::sleep;
 
     #[async_trait::async_trait]
-    impl LogExporter for AsyncLogExporter {
+    impl LogExporter for LogExporterThatRequiresTokioSpawn {
         async fn export(&mut self, batch: LogBatch<'_>) -> LogResult<()> {
             println!("Exporting...");
 
@@ -1118,7 +1118,7 @@ mod tests {
     fn test_simple_processor_async_exporter_without_runtime() {
         // Use `catch_unwind` to catch the panic caused by missing Tokio runtime
         let result = std::panic::catch_unwind(|| {
-            let exporter = AsyncLogExporter::new();
+            let exporter = LogExporterThatRequiresTokioSpawn::new();
             let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
             let mut record: LogRecord = Default::default();
@@ -1153,7 +1153,7 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn test_simple_processor_async_exporter_with_runtime() {
-        let exporter = AsyncLogExporter::new();
+        let exporter = LogExporterThatRequiresTokioSpawn::new();
         let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
         let mut record: LogRecord = Default::default();
@@ -1166,7 +1166,7 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_simple_processor_async_exporter_with_multi_thread_runtime() {
-        let exporter = AsyncLogExporter::new();
+        let exporter = LogExporterThatRequiresTokioSpawn::new();
         let processor = Arc::new(Mutex::new(SimpleLogProcessor::new(Box::new(
             exporter.clone(),
         ))));
@@ -1195,7 +1195,7 @@ mod tests {
     #[tokio::test(flavor = "current_thread")]
     #[ignore] // This test hangs as of now.
     async fn test_simple_processor_async_exporter_with_current_thread_runtime() {
-        let exporter = AsyncLogExporter::new();
+        let exporter = LogExporterThatRequiresTokioSpawn::new();
 
         let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -1184,7 +1184,7 @@ mod tests {
         for handle in handles {
             handle.await.unwrap();
         }
-        assert_eq!(exporter.len(), 2);
+        assert_eq!(exporter.len(), 4);
     }
 
     #[tokio::test(flavor = "multi_thread")]
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -96,6 +96,7 @@ impl SimpleLogProcessor {
 
 impl LogProcessor for SimpleLogProcessor {
     fn emit(&self, record: &mut LogRecord, instrumentation: &InstrumentationLibrary) {
+        println!("SimpleLogProcessor::Emit");
         // noop after shutdown
         if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
             // this is a warning, as the user is trying to log after the processor has been shutdown
@@ -553,7 +554,10 @@ mod tests {
     use opentelemetry::InstrumentationLibrary;
     use opentelemetry::Key;
     use opentelemetry::{logs::LogResult, KeyValue};
-    use std::sync::{Arc, Mutex};
+    use std::sync::{
+        atomic::{AtomicU32, Ordering},
+        Arc, Mutex,
+    };
     use std::time::Duration;
 
     #[derive(Debug, Clone)]
@@ -1066,44 +1070,37 @@ mod tests {
         let instrumentation: InstrumentationLibrary = Default::default();
 
         processor.emit(&mut record, &instrumentation);
-        processor.force_flush().unwrap();
-        processor.shutdown().unwrap();
 
         assert_eq!(exporter.get_emitted_logs().unwrap().len(), 1);
     }
 
     #[derive(Debug, Clone)]
-    struct LogExporterThatRequiresTokioSpawn {
-        logs: Arc<Mutex<Vec<(LogRecord, InstrumentationLibrary)>>>,
+    struct LogExporterThatRequiresTokio {
+        event_count: Arc<AtomicU32>,
     }
 
-    impl LogExporterThatRequiresTokioSpawn {
-        /// Creates a new instance of `LogExporterThatRequiresTokioSpawn`.
-        pub(crate) fn new() -> Self {
-            LogExporterThatRequiresTokioSpawn {
-                logs: Arc::new(Mutex::new(Vec::new())),
+    impl LogExporterThatRequiresTokio {
+        /// Creates a new instance of `LogExporterThatRequiresTokio`.
+        fn new() -> Self {
+            LogExporterThatRequiresTokio {
+                event_count: Arc::new(AtomicU32::new(0)),
             }
         }
 
         /// Returns the number of logs stored in the exporter.
-        pub(crate) async fn len(&self) -> usize {
-            let logs = self.logs.lock().unwrap();
-            logs.len()
+        fn len(&self) -> usize {
+            self.event_count.load(Ordering::Acquire) as usize
         }
     }
 
-    use tokio::time::sleep;
-
     #[async_trait::async_trait]
-    impl LogExporter for LogExporterThatRequiresTokioSpawn {
+    impl LogExporter for LogExporterThatRequiresTokio {
         async fn export(&mut self, batch: LogBatch<'_>) -> LogResult<()> {
-            // Simulate minimal dependency on tokio by sleeping for a short duration
-            sleep(Duration::from_millis(50)).await;
+            // Simulate minimal dependency on tokio by sleeping asynchronously for a short duration
+            tokio::time::sleep(Duration::from_millis(50)).await;
 
-            let logs = Arc::clone(&self.logs);
-            let mut logs_lock = logs.lock().unwrap();
-            for (log_record, instrumentation) in batch.iter() {
-                logs_lock.push((log_record.clone(), instrumentation.clone()));
+            for _ in batch.iter() {
+                self.event_count.fetch_add(1, Ordering::Acquire);
             }
             Ok(())
         }
@@ -1113,13 +1110,13 @@ mod tests {
     fn test_simple_processor_async_exporter_without_runtime() {
         // Use `catch_unwind` to catch the panic caused by missing Tokio runtime
         let result = std::panic::catch_unwind(|| {
-            let exporter = LogExporterThatRequiresTokioSpawn::new();
+            let exporter = LogExporterThatRequiresTokio::new();
             let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
             let mut record: LogRecord = Default::default();
             let instrumentation: InstrumentationLibrary = Default::default();
 
-            // This will panic because `tokio::spawn` or an async operation is called without a runtime.
+            // This will panic because an async operation is called without a runtime.
             processor.emit(&mut record, &instrumentation);
         });
 
@@ -1128,27 +1125,24 @@ mod tests {
             result.is_err(),
             "The test should fail due to missing Tokio runtime, but it did not."
         );
+        let panic_payload = result.unwrap_err();
+        let panic_message = panic_payload
+            .downcast_ref::<String>()
+            .map(|s| s.as_str())
+            .or_else(|| panic_payload.downcast_ref::<&str>().copied())
+            .unwrap_or("No panic message");
 
-        if let Err(panic) = result {
-            let panic_message = panic
-                .downcast_ref::<String>()
-                .map(|s| s.as_str())
-                .or_else(|| panic.downcast_ref::<&str>().copied())
-                .unwrap_or("No panic message");
-
-            assert!(
-                panic_message.contains("no reactor running")
-                    || panic_message
-                        .contains("must be called from the context of a Tokio 1.x runtime"),
-                "Expected panic message about missing Tokio runtime, but got: {}",
-                panic_message
-            );
-        }
+        assert!(
+            panic_message.contains("no reactor running")
+                || panic_message.contains("must be called from the context of a Tokio 1.x runtime"),
+            "Expected panic message about missing Tokio runtime, but got: {}",
+            panic_message
+        );
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn test_simple_processor_async_exporter_with_runtime() {
-        let exporter = LogExporterThatRequiresTokioSpawn::new();
+        let exporter = LogExporterThatRequiresTokio::new();
         let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
         let mut record: LogRecord = Default::default();
@@ -1156,23 +1150,23 @@ mod tests {
 
         processor.emit(&mut record, &instrumentation);
 
-        assert_eq!(exporter.len().await, 1);
+        assert_eq!(exporter.len(), 1);
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
-    #[ignore]
-    // all threads are blocked running blocked::export(), and the exporter further needs tokio
-    // runtime to progress on this blocked thread, resulting in deadlock.
-    async fn test_simple_processor_async_exporter_with_multi_thread_runtime_all_cores_blocked() {
-        let exporter = LogExporterThatRequiresTokioSpawn::new();
+    //#[ignore]
+    // All worker threads except one are blocked, waiting for the export operation to complete.
+    // The exporter, which isn't blocked, requires the runtime to proceed, but no free worker threads are available, resulting in a deadlock.
+    async fn test_simple_processor_async_exporter_with_all_runtime_worker_threads_blocked() {
+        let exporter = LogExporterThatRequiresTokio::new();
         let processor = Arc::new(Mutex::new(SimpleLogProcessor::new(Box::new(
             exporter.clone(),
         ))));
 
-        let concurrent_emit = 5; // number of worker sthreads + 1
+        let concurrent_emit = 5; // number of worker threads + 1
 
         let mut handles = vec![];
-        // send 2 events concurrently
+        // try send `concurrent_emit` events concurrently
         for _ in 0..concurrent_emit {
             let processor_clone = Arc::clone(&processor);
             let handle = tokio::spawn(async move {
@@ -1189,12 +1183,12 @@ mod tests {
         for handle in handles {
             handle.await.unwrap();
         }
-        assert_eq!(exporter.len().await, 2);
+        assert_eq!(exporter.len(), 2);
     }
 
     #[tokio::test(flavor = "multi_thread")]
     async fn test_simple_processor_async_exporter_with_multi_thread_runtime() {
-        let exporter = LogExporterThatRequiresTokioSpawn::new();
+        let exporter = LogExporterThatRequiresTokio::new();
 
         let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
@@ -1203,7 +1197,7 @@ mod tests {
 
         processor.emit(&mut record, &instrumentation);
 
-        assert_eq!(exporter.len().await, 1);
+        assert_eq!(exporter.len(), 1);
     }
 
     #[tokio::test(flavor = "current_thread")]
@@ -1211,7 +1205,7 @@ mod tests {
               // complete the export, and the exporter further needs tokio runtime to progress
               // on this blocked thread, resulting in deadlock.
     async fn test_simple_processor_async_exporter_with_current_thread_runtime() {
-        let exporter = LogExporterThatRequiresTokioSpawn::new();
+        let exporter = LogExporterThatRequiresTokio::new();
 
         let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
 
@@ -1220,6 +1214,6 @@ mod tests {
 
         processor.emit(&mut record, &instrumentation);
 
-        assert_eq!(exporter.len().await, 1);
+        assert_eq!(exporter.len(), 1);
     }
 }
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -555,6 +555,7 @@ mod tests {
     use opentelemetry::{logs::LogResult, KeyValue};
     use std::sync::{Arc, Mutex};
     use std::time::Duration;
+    //use num_cpus;
 
     #[derive(Debug, Clone)]
     struct MockLogExporter {
@@ -1159,16 +1160,21 @@ mod tests {
         assert_eq!(exporter.len().await, 1);
     }
 
-    #[tokio::test(flavor = "multi_thread")]
-    async fn test_simple_processor_async_exporter_with_multi_thread_runtime() {
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    #[ignore]
+    // all threads are blocked running blocked::export(), and the exporter further needs tokio
+    // runtime to progress on this blocked thread, resulting in deadlock.
+    async fn test_simple_processor_async_exporter_with_multi_thread_runtime_all_cores_blocked() {
         let exporter = LogExporterThatRequiresTokioSpawn::new();
         let processor = Arc::new(Mutex::new(SimpleLogProcessor::new(Box::new(
             exporter.clone(),
         ))));
 
+        let concurrent_emit = 5; // number of worker sthreads + 1
+
         let mut handles = vec![];
-        // send 4 events concurrently
-        for _ in 0..4 {
+        // send 2 events concurrently
+        for _ in 0..concurrent_emit {
             let processor_clone = Arc::clone(&processor);
             let handle = tokio::spawn(async move {
                 let mut record: LogRecord = Default::default();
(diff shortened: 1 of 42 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,172 @@ mod tests {
                 == AnyValue::String("Updated by FirstProcessor".into())
         );
     }
+
+    #[derive(Debug, Clone)]
+    struct SyncLogExporter {
```
cijothomas, line 1000: Isn't this same as InMemoryLogExporter?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,172 @@ mod tests {
...
+    impl LogExporter for SyncLogExporter {
+        async fn export(&mut self, batch: LogBatch<'_>) -> LogResult<()> {
+            let mut logs = self.logs.lock().unwrap();
+            for (log_record, instrumentation) in batch.iter() {
+                logs.push((log_record.clone(), instrumentation.clone()));
+            }
+            Ok(())
+        }
+
+        fn shutdown(&mut self) {}
```
cijothomas, line 1014: are these empty methods required?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,172 @@ mod tests {
...
+        let exporter = SyncLogExporter {
+            logs: Arc::new(Mutex::new(Vec::new())),
+        };
+        let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+        let mut record: LogRecord = Default::default();
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
```
cijothomas, line 1030: why flush and shutdown? shutdown alone is good enough right?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,172 @@ mod tests {
...
+            let logs = Arc::clone(&self.logs);
+
+            // Collect batch into an owned Vec to ensure the data can be moved into the tokio::spawn task
+            let batch_owned: Vec<(LogRecord, InstrumentationLibrary)> = batch
+                .iter()
+                .map(|(log_record, instrumentation)| (log_record.clone(), instrumentation.clone()))
+                .collect();
+
+            // Spawn a Tokio task to process the log batch asynchronously
+            tokio::spawn(async move {
```
cijothomas, line 1087: wondering if this is really a scalable way of testing. Should we restrict to testing the features offered by otlp specificially?
otlp + tonic from tokio main, normal main
otlp + request ...
otlp + request blocking

lalitb: Yes but it would be difficult to test that in unit-test with collector. I think we can extend functional test for that. The current test is more generic to cover any exporter which internally uses tokio runtime.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: yes using InMemoryLogsExporter now.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: No, removed them.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Yes as of now, force_flush() is empty. Just as safeguard, if we change it in future, this will test it out.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
+        processor.shutdown().unwrap();
+
+        assert_eq!(exporter.get_emitted_logs().unwrap().len(), 1);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
ThomsonTan, line 1016: Also test for runtime other than tokio?

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: since this test is specifically for simple, we don't actually need either flush or shutdown, so they can be removed.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+                logs: Arc::new(Mutex::new(Vec::new())),
+            };
+            let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+            let mut record: LogRecord = Default::default();
+            let instrumentation: InstrumentationLibrary = Default::default();
+
+            // This will panic because `tokio::spawn` is called without a runtime
+            processor.emit(&mut record, &instrumentation);
+            processor.force_flush().unwrap();
```
cijothomas, line 1094: please remove flush and shutdown

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+            processor.shutdown().unwrap();
+
+            // Assert to make sure the failure occurs
+            assert_eq!(exporter.logs.lock().unwrap().len(), 1);
```
cijothomas, line 1098: will this be executed ever?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
+        processor.shutdown().unwrap();
+
+        assert_eq!(exporter.get_emitted_logs().unwrap().len(), 1);
+    }
+
+    #[derive(Debug, Clone)]
+    struct AsyncLogExporter {
```
cijothomas, line 1051: lets use a better name, better reflecting its functionality. It is specifically using Tokio spawn, so maybe `LogExporterThatRequiresTokioSpawn`

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+        let exporter = AsyncLogExporter {
+            logs: Arc::new(Mutex::new(Vec::new())),
+        };
+        let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+        let mut record: LogRecord = Default::default();
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
```
cijothomas, line 1119: remove flush and shutdown

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+        });
+
+        // Verify that the panic occurred, indicating the absence of a Tokio runtime
+        assert!(
+            result.is_err(),
+            "The test should fail due to missing Tokio runtime, but it did not."
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
cijothomas, line 1108: have a variation for worker_thread >1 too

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+            processor.force_flush().unwrap();
+            processor.shutdown().unwrap();
+
+            // Assert to make sure the failure occurs
+            assert_eq!(exporter.logs.lock().unwrap().len(), 1);
+        });
+
+        // Verify that the panic occurred, indicating the absence of a Tokio runtime
+        assert!(
+            result.is_err(),
```
ThomsonTan, line 1103: Is it possible to check the exact panic type, in case there are other panic raised other than missing runtime in scope?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +995,148 @@ mod tests {
...
+
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
+        processor.shutdown().unwrap();
+
+        assert_eq!(exporter.logs.lock().unwrap().len(), 1);
+    }
+
+    #[tokio::test(flavor = "current_thread")]
+    #[ignore] // This test hangs as of now.
```
cijothomas, line 1126: its unclear from the comment where/why this hangs.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: agree its best to cover it via the integration test. I did some of them manually and it is not scaleable.

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan, line 1126: Sounds like it hangs with the same reason as batch processor, probably link them together?

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan: And also test `worker_threads = 0`?

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Just to be clear, we still keep current generic  test here with runtime - to cover the hang scenario, and panic due to missing runtime.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Yes, this can be extended in future. As of now, we don't have the use-case for async-std.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: no, have removed this.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: thanks, this looks better. renamed.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: done

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Have added for `#[tokio::test(flavor = "multi_thread"]`, which configures number of threads as number of core.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Done. Thanks.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Do you have issue for that ?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+        pub(crate) async fn len(&self) -> usize {
+            let logs = self.logs.lock().unwrap();
+            logs.len()
+        }
+    }
+
+    use tokio::time::sleep;
+
+    #[async_trait::async_trait]
+    impl LogExporter for LogExporterThatRequiresTokioSpawn {
```
cijothomas, line 1098:
```suggestion
    impl LogExporter for LogExporterThatRequiresTokio {
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+        async fn export(&mut self, batch: LogBatch<'_>) -> LogResult<()> {
+            // Simulate minimal dependency on tokio by sleeping for a short duration
+            sleep(Duration::from_millis(50)).await;
```
cijothomas, line 1101:
```suggestion
            tokio::time::sleep(Duration::from_millis(50)).await;
```

I think this makes it easy to spot that this is tokio sleep

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
+        processor.shutdown().unwrap();
+
+        assert_eq!(exporter.get_emitted_logs().unwrap().len(), 1);
+    }
+
+    #[derive(Debug, Clone)]
+    struct LogExporterThatRequiresTokioSpawn {
+        logs: Arc<Mutex<Vec<(LogRecord, InstrumentationLibrary)>>>,
```
cijothomas, line 1077: lets just keep an atomic count of the number of times this was called. No need of storing anything else, as the goal of this test is not validating that.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+    async fn test_simple_processor_async_exporter_with_multi_thread_runtime_all_cores_blocked() {
+        let exporter = LogExporterThatRequiresTokioSpawn::new();
+        let processor = Arc::new(Mutex::new(SimpleLogProcessor::new(Box::new(
+            exporter.clone(),
+        ))));
+
+        let concurrent_emit = 5; // number of worker sthreads + 1
+
+        let mut handles = vec![];
+        // send 2 events concurrently
```
cijothomas, line 1175: not sure I follow this.. each task take a mutex lock to invoke emit on processor, so processors are called serially only right?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+        let exporter = InMemoryLogsExporterBuilder::default()
+            .keep_records_on_shutdown()
+            .build();
+        let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+        let mut record: LogRecord = Default::default();
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
+        processor.force_flush().unwrap();
```
utpilla, line 1069: Do we need to force flush and shutdown here? Other tests only need processor emit to make the assertion.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+    fn test_simple_processor_async_exporter_without_runtime() {
+        // Use `catch_unwind` to catch the panic caused by missing Tokio runtime
+        let result = std::panic::catch_unwind(|| {
+            let exporter = LogExporterThatRequiresTokioSpawn::new();
+            let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+            let mut record: LogRecord = Default::default();
+            let instrumentation: InstrumentationLibrary = Default::default();
+
+            // This will panic because `tokio::spawn` or an async operation is called without a runtime.
```
utpilla, line 1122: Did you mean `tokio::sleep`?

```suggestion
            // This will panic because `tokio::sleep` is called without Tokio runtime.
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+            for (log_record, instrumentation) in batch.iter() {
+                logs_lock.push((log_record.clone(), instrumentation.clone()));
+            }
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn test_simple_processor_async_exporter_without_runtime() {
+        // Use `catch_unwind` to catch the panic caused by missing Tokio runtime
```
utpilla, line 1114: Nice!

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+            processor.emit(&mut record, &instrumentation);
+        });
+
+        // Verify that the panic occurred and check the panic message for the absence of a Tokio runtime
+        assert!(
+            result.is_err(),
+            "The test should fail due to missing Tokio runtime, but it did not."
+        );
+
+        if let Err(panic) = result {
```
utpilla, line 1132: nit: Since we ensure that `result` is `Err`, we could use `unwrap_err` instead of `if let Err(panic) = result`

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+    }
+
+    #[derive(Debug, Clone)]
+    struct LogExporterThatRequiresTokioSpawn {
+        logs: Arc<Mutex<Vec<(LogRecord, InstrumentationLibrary)>>>,
+    }
+
+    impl LogExporterThatRequiresTokioSpawn {
+        /// Creates a new instance of `LogExporterThatRequiresTokioSpawn`.
+        pub(crate) fn new() -> Self {
```
utpilla, line 1082: Remove `pub(crate)` for these methods?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+            LogExporterThatRequiresTokioSpawn {
+                logs: Arc::new(Mutex::new(Vec::new())),
+            }
+        }
+
+        /// Returns the number of logs stored in the exporter.
+        pub(crate) async fn len(&self) -> usize {
```
utpilla, line 1089: This method need not be asynchronous, right?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +994,232 @@ mod tests {
...
+        processor.emit(&mut record, &instrumentation);
+
+        assert_eq!(exporter.len().await, 1);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    #[ignore]
+    // all threads are blocked running blocked::export(), and the exporter further needs tokio
+    // runtime to progress on this blocked thread, resulting in deadlock.
+    async fn test_simple_processor_async_exporter_with_multi_thread_runtime_all_cores_blocked() {
```
utpilla, line 1166: The issue is due to worker threads getting blocked not the cores.

```suggestion
    async fn test_simple_processor_async_exporter_with_all_runtime_worker_threads_blocked() {
```

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Yes the processor::emit is called serially. However this does spawn's  4 out of 5 requested async tasks (on 4 available worker threads) concurrently, one task is able to move forward, and invoke exporter::export() method. The rest of the tokio tasks are blocked waiting for the processor lock. So the running exporter task doesn't get any free worker thread to run the tokio's sleep method, resulting in deadlock.
This mimics the scenario where use invokes tracing event concurrently from max available worker threads, and all but one wait on the SimpleLogProcessor export lock to be released, resulting in the deadlock.

lalitb: This is somehow to demonstrate that even in case the tokio::main is invoked in multiple thread mode, there is possibility of deadlock just because the futures::block_on ensures any async calls inside it run on the current worker thread.

lalitb:
@cijothomas @ThomsonTan @utpilla thanks for the thorough review, and valuable comments. I have tried to address all of them. Please have a look once you get time.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: > And also test worker_threads = 0?

seems tokio doesn't allow configuring with `worker_threads = 0`

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,223 @@ mod tests {
...
+    fn test_simple_processor_async_exporter_without_runtime() {
+        // Use `catch_unwind` to catch the panic caused by missing Tokio runtime
+        let result = std::panic::catch_unwind(|| {
+            let exporter = LogExporterThatRequiresTokio::new();
+            let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+            let mut record: LogRecord = Default::default();
+            let instrumentation: InstrumentationLibrary = Default::default();
+
+            // This will panic because an async operation is called without a runtime.
```
utpilla, line 1119: The panic is specifically caused by `tokio::sleep`. The export method being an async operation does not cause this error. We didn't have this issue for InMemoryExporter.

lalitb: I meant tokio async operation :) Have clarified this now.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,223 @@ mod tests {
...
+    // All worker threads except one are blocked, waiting for the export operation to complete.
+    // The exporter, which isn't blocked, requires the runtime to proceed, but no free worker threads are available, resulting in a deadlock.
+    // This mimics the scenario where the
+    async fn test_simple_processor_async_exporter_with_all_runtime_worker_threads_blocked() {
+        let exporter = LogExporterThatRequiresTokio::new();
+        let processor = Arc::new(Mutex::new(SimpleLogProcessor::new(Box::new(
+            exporter.clone(),
+        ))));
+
+        let concurrent_emit = 5; // number of worker threads + 1
```
utpilla, line 1167: This would hang even with 4 threads.

```suggestion
        let concurrent_emit = 4; // number of worker threads
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,223 @@ mod tests {
...
+
+        processor.emit(&mut record, &instrumentation);
+
+        assert_eq!(exporter.len(), 1);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    #[ignore]
+    // All worker threads except one are blocked, waiting for the export operation to complete.
+    // The exporter, which isn't blocked, requires the runtime to proceed, but no free worker threads are available, resulting in a deadlock.
```
utpilla, line 1159: Let's a provide a more specific reasoning here? How about something like this?

We spawn enough number of tasks to have each worker thread execute export. While executing export, each of these worker threads park themselves when they hit tokio `sleep` (since `sleep` is called within `futures_executor::block_on`). We need a worker thread to react to the completion of sleep and unpark these threads but there is none available. As a result, the worker threads never get unparked and the spawned tokio tasks don't ever complete.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,223 @@ mod tests {
...
+                    .unwrap()
+                    .emit(&mut record, &instrumentation);
+            });
+            handles.push(handle);
+        }
+
+        for handle in handles {
+            handle.await.unwrap();
+        }
+        assert_eq!(exporter.len(), 2);
```
utpilla, line 1187: Even though we don't execute this line, let's make this match the number of worker threads to show the correct intent.

```suggestion
        assert_eq!(exporter.len(), 4);
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,223 @@ mod tests {
...
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_simple_processor_async_exporter_with_runtime() {
+        let exporter = LogExporterThatRequiresTokio::new();
+        let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));
+
+        let mut record: LogRecord = Default::default();
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
```
utpilla, line 1151: I think it would be nice to add a note here that this works only because the export method executes synchronously here. If this was under a tokio spawn, the export would have hung again as there is only one worker thread.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: oh yes thanks.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: @utpilla actually only one worker thread is blocked on the tokio `sleep`, rest are blocked on the `mutex` for processor. I have added the comment to make it clear as below:
```
    // This test demonstrates a potential deadlock scenario in a multi-threaded Tokio runtime.
    // It spawns Tokio tasks equal to the number of runtime worker threads (4) to emit log events.
    // Each task attempts to acquire a mutex on the SimpleLogProcessor. Only one task obtains the lock,
    // while the others are blocked, waiting for its release.
    //
    // The task holding the lock invokes the LogExporterThatRequiresTokio, which performs an
    // asynchronous operation (e.g., network I/O simulated by `tokio::sleep`). This operation
    // requires yielding control back to the Tokio runtime to make progress.
    //
    // However, all worker threads are occupied:
    // - One thread is executing the async exporter operation
    // - Three threads are blocked waiting for the mutex
    //
    // This leads to a deadlock as there are no available threads to drive the async operation
    // to completion, preventing the mutex from being released. Consequently, neither the blocked
    // tasks nor the exporter can proceed.
```

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: thanks @utpilla. The export method is asynchronous here (ie with tokio `sleep`) just like the current_thread (`test_simple_processor_async_exporter_with_current_thread_runtime`) scenario. The difference is that we still have worker_thread to execute this async task. I have updated the comments for both as below:

    `#[tokio::test(flavor = "multi_thread", worker_threads = 1)]`:

    // This test uses a multi-threaded runtime setup with a single worker thread. Note that even
    // though only one worker thread is created, it is distinct from the main thread. The processor
    // emits a log event, and the exporter performs an async operation that requires the runtime.
    // The single worker thread handles this operation without deadlocking, as long as no other
    // tasks occupy the runtime.

    `#[tokio::test(flavor = "multi_thread")]`

    // This test uses a multi-threaded runtime setup with the default number of worker threads.
    // The processor emits a log event, and the exporter, which requires the runtime for its async
    // operations, can access one of the available worker threads to complete its task. As there
    // are multiple threads, the exporter can proceed without blocking other tasks, ensuring the
    // test completes successfully.

    #[tokio::test(flavor = "current_thread")]
    #[ignore]
    // This test uses a current-thread runtime, where all operations run on the main thread.
    // The processor emits a log event while the runtime is blocked using `futures::block_on`
    // to complete the export operation. The exporter, which performs an async operation and
    // requires the runtime, cannot progress because the main thread is already blocked.
    // This results in a deadlock, as the runtime cannot move forward.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -96,6 +96,7 @@ impl SimpleLogProcessor {
 
 impl LogProcessor for SimpleLogProcessor {
     fn emit(&self, record: &mut LogRecord, instrumentation: &InstrumentationLibrary) {
+        println!("SimpleLogProcessor::Emit");
```
cijothomas, line 99: lets use internal logs for these.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,250 @@ mod tests {
                 == AnyValue::String("Updated by FirstProcessor".into())
         );
     }
+
+    #[test]
+    fn test_simple_processor_sync_exporter_without_runtime() {
+        let exporter = InMemoryLogsExporterBuilder::default()
+            .keep_records_on_shutdown()
```
cijothomas, line 1005: wondering if keep_record_on_shutdown is relevant here?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,250 @@ mod tests {
                 == AnyValue::String("Updated by FirstProcessor".into())
         );
     }
+
+    #[test]
+    fn test_simple_processor_sync_exporter_without_runtime() {
```
cijothomas, line 1003: nit: We can write a helper and call them for all the cominations:
1. Normal test
2. Tokio test - current thread
3. Tokio test - multi-thread
4. Tokio test - multi-thread with thread=1

we can eventually prepare a test matrix.md file for quick visualization.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,250 @@ mod tests {
...
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
+
+        assert_eq!(exporter.get_emitted_logs().unwrap().len(), 1);
+    }
+
+    #[derive(Debug, Clone)]
+    struct LogExporterThatRequiresTokio {
+        event_count: Arc<AtomicU32>,
```
cijothomas, line 1079:
```suggestion
        export_count: Arc<AtomicU32>,
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,250 @@ mod tests {
...
+            "Expected panic message about missing Tokio runtime, but got: {}",
+            panic_message
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    #[ignore]
+    // This test demonstrates a potential deadlock scenario in a multi-threaded Tokio runtime.
+    // It spawns Tokio tasks equal to the number of runtime worker threads (4) to emit log events.
+    // Each task attempts to acquire a mutex on the SimpleLogProcessor. Only one task obtains the lock,
```
cijothomas, line 1147: If we use `tracing` with tracing-appender, then we can more closely mimic user scenario. The test acquiring lock on SimpleProcessor, while simulates things, is not really what end users does.

maybe a follow up is to write this test using `tracing` macros to simulate user scenarios.

cijothomas (approved):
Thanks for adding coverage for all the combinations! 
There are some comments left, which I think can be addressed via followups, as needed.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +998,250 @@ mod tests {
...
+
+        let mut handles = vec![];
+        // try send `concurrent_emit` events concurrently
+        for _ in 0..concurrent_emit {
+            let processor_clone = Arc::clone(&processor);
+            let handle = tokio::spawn(async move {
+                let mut record: LogRecord = Default::default();
+                let instrumentation: InstrumentationLibrary = Default::default();
+                processor_clone
+                    .lock()
```
utpilla, line 1177: Remove this mutex around the SimpleLogProcessor? The `processor.emit` code already takes a lock before calling `exporter.export` so this does not seem to add any value on top of that.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: I see. Just a general note, I feel that the `Mutex` used by `SimpleLogProcessor` is hiding away the main issue here which is `futures_executor::block_on`. For this particular test, yes since we use a `SimpleLogProcessor` other worker threads are waiting on the `Mutex`. That might lead us into thinking that `Mutex` is the real issue here but it's not.

Even if we were to remove the `Mutex`, and use something like a Reentrant/no-lock processor we would still hit this issue as long as we have `futures_exectuor::block_on`. Try this code which takes no mutex and spawns a task on each worker thread to wait for some time. This would still hang due to the reasoning stated in this [comment](https://github.com/open-telemetry/opentelemetry-rust/pull/2218#discussion_r1807055338).

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
async fn test_simple_processor_async_exporter_with_all_runtime_worker_threads_blocked() {
        let concurrent_emit = 4; // number of worker threads

        let mut handles = vec![];
        for _ in 0..concurrent_emit {
            let handle = tokio::spawn(async move {
                futures_executor::block_on(async { tokio::time::sleep(Duration::from_millis(50)).await });
            });
            handles.push(handle);
        }

        for handle in handles {
            handle.await.unwrap();
        }
}

File: opentelemetry-sdk/src/logs/log_processor.rs (hunk 28 above)
Comment (utpilla) [reply opentelemetry-sdk/src/logs/log_processor.rs]: What I meant was this test works only because we execute `processor.emit` on the same original thread. If we were to do that as a spawned task this would have hung as we would have blocked the only worker thread we have by calling `futures_executor::block_on`.

This code with just a single worker thread will hang:

```rust
async fn test_simple_processor_async_exporter_with_runtime() {
        let exporter = LogExporterThatRequiresTokio::new();
        let processor = SimpleLogProcessor::new(Box::new(exporter.clone()));

        let mut record: LogRecord = Default::default();
        let instrumentation: InstrumentationLibrary = Default::default();

        let handle =  tokio::spawn(async move {
            processor.emit(&mut record, &instrumentation);
        });

        handle.await.unwrap();

        // processor.emit(&mut record, &instrumentation);

        assert_eq!(exporter.len(), 1);
    }

File: opentelemetry-sdk/src/logs/log_processor.rs (hunk 26 above)
Comment (cijothomas) [reply opentelemetry-sdk/src/logs/log_processor.rs]: +1 to above.
I left https://github.com/open-telemetry/opentelemetry-rust/pull/2218#discussion_r1808032002 to mean same thing, but above is a very good explanation!

Comment (lalitb) [reply opentelemetry-sdk/src/logs/log_processor.rs]: @utpilla In case of SimpleLogProcessor, we will never have concurrent execution of `block_on` because of the mutex here - https://github.com/open-telemetry/opentelemetry-rust/blob/84c151c0b6f755b71c466f98d54a6336ebe159aa/opentelemetry-sdk/src/logs/log_processor.rs#L110

So actually it is the combination of mutex and block_on in the processor code that is causing deadlock :)

However, I have removed the mutex from the unit-test, as there is already mutex as above, as also you indicated here - https://github.com/open-telemetry/opentelemetry-rust/pull/2218#discussion_r1808080261. Hopefully this will make it more clear now.

@cijothomas  - In this test now that we don't lock processor, we invoke SimpleLogProcessor::emit concurrently from multiple tokio worker threads, which would be the similar scenario when the user invokes multiple tracing macros concurrently.

File: opentelemetry-sdk/src/logs/log_processor.rs (hunk 28 above)
Comment (lalitb) [reply opentelemetry-sdk/src/logs/log_processor.rs]: yes I think I got your point now. I believe the scenario you indicated above is similar to our "multi_thread" with multiple workers test where we spawn tokio tasks. Do you want to still make it explicit here ?

File: opentelemetry-sdk/src/logs/log_processor.rs (hunk 34 above)
Comment (lalitb) [reply opentelemetry-sdk/src/logs/log_processor.rs:1177]: Very good point. Have removed this now.

File: opentelemetry-sdk/src/logs/log_processor.rs (hunk 31 above)
Comment (lalitb) [reply opentelemetry-sdk/src/logs/log_processor.rs:1003]: Yes, good to have a test matrix. Will do it after adding tests for batch.

File: opentelemetry-sdk/src/logs/log_processor.rs [hunk 35]
```diff
@@ -994,4 +989,237 @@ mod tests {
...
+        let instrumentation: InstrumentationLibrary = Default::default();
+
+        processor.emit(&mut record, &instrumentation);
+
+        assert_eq!(exporter.get_emitted_logs().unwrap().len(), 1);
+    }
+
+    #[derive(Debug, Clone)]
+    struct LogExporterThatRequiresTokio {
+        export_count: Arc<AtomicU32>,
```
cijothomas, line 1062: `Arc` needed here?

lalitb: yes, Atomic is not clonable, and need to keep the clone of exporter to check the count later.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +989,237 @@ mod tests {
...
+        ))));
+
+        let mut handles = vec![];
+        for _ in 0..10 {
+            let processor_clone = Arc::clone(&processor);
+            let handle = tokio::spawn(async move {
+                let mut record: LogRecord = Default::default();
+                let instrumentation: InstrumentationLibrary = Default::default();
+                processor_clone
+                    .lock()
```
utpilla, line 1033: Same as this https://github.com/open-telemetry/opentelemetry-rust/pull/2218#discussion_r1808080261

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -994,4 +989,237 @@ mod tests {
...
+        /// Creates a new instance of `LogExporterThatRequiresTokio`.
+        fn new() -> Self {
+            LogExporterThatRequiresTokio {
+                export_count: Arc::new(AtomicU32::new(0)),
+            }
+        }
+
+        /// Returns the number of logs stored in the exporter.
+        fn len(&self) -> usize {
+            self.export_count.load(Ordering::Acquire) as usize
```
utpilla, line 1075: nit: We could simply use an `AtomicUsize` to begin with to avoid this casting.

utpilla (approved):

PR #2220, opened 2024-10-18
Unify builders across signals

### LINKED ISSUE #1527: Instrumentation scope - move to builder pattern?

### LINKED ISSUE #2164: Metrics API - Move to builder pattern for creating Meters
Related to https://github.com/open-telemetry/opentelemetry-rust/issues/1527, but for Metrics
This is part of Metrics API, and hence must be done by Metrics API RC.

stormshield-fabs:
Fixes #1527 and fixes #2164

## Changes

The consensus reached during the SIG meeting is that we'd like to unify the builder APIs across signals, limit duplication and keep our traits object-safe. 
- `opentelemetry`
    - Removed deprecated method `InstrumentationLibrary::new`
    - Renamed `InstrumentationLibrary` to `InstrumentationScope`
    - Renamed `InstrumentationLibraryBuilder` to `InstrumentationScopeBuilder`
    - Removed deprecated methods `LoggerProvider::versioned_logger` and `TracerProvider::versioned_tracer`
    - Removed methods `LoggerProvider::logger_builder`, `TracerProvider::tracer_builder` and `MeterProvider::versioned_meter`
    - Replaced these methods with `LoggerProvider::logger_with_scope`, `TracerProvider::logger_with_scope`, `MeterProvider::meter_with_scope`
    - Replaced `global::meter_with_version` with `global::meter_with_scope`
    - Added `global::tracer_with_scope`
- `opentelemetry_sdk`:
    - Removed `InstrumentationLibrary` re-export and its `Scope` alias, use `opentelemetry::InstrumentationLibrary` instead.
    - Unified builders across signals
      - Removed deprecated `LoggerProvider::versioned_logger`, `TracerProvider::versioned_tracer`
      - Removed `MeterProvider::versioned_meter`
      - Replaced these methods with `LoggerProvider::logger_with_scope`, `TracerProvider::logger_with_scope`, `MeterProvider::meter_with_scope`

## Migration guide (example for `logger` but the changes are similar across all 3 signals)

```rust
let provider = LoggerProvider::builder().build();

// Before:
let library = std::sync::Arc::new(
    InstrumentationLibrary::builder("test_logger")
        .with_version("v1.2.3")
        .with_schema_url("https://opentelemetry.io/schema/1.0.0")
        .build(),
);
let logger = provider.library_logger(library); // no longer available, use provider.logger_with_scope


// or maybe (was deprecated):
let logger = provider.versioned_logger(
  "test_logger",
  Some("v1.2.3".into()),
  Some("https://opentelemetry.io/schema/1.0.0".into()),
);

// After:
let scope = InstrumentationScope::builder("test_logger")
  .with_version("v1.2.3")
  .with_schema_url("https://opentelemetry.io/schema/1.0.0")
  .build();

let logger = provider.logger_with_scope(scope);
```

### DIFF
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -97,11 +97,11 @@
 use log::{Level, Metadata, Record};
 use opentelemetry::{
     logs::{AnyValue, LogRecord, Logger, LoggerProvider, Severity},
-    Key,
+    InstrumentationLibrary, Key,
 };
 #[cfg(feature = "experimental_metadata_attributes")]
 use opentelemetry_semantic_conventions::attribute::{CODE_FILEPATH, CODE_LINENO, CODE_NAMESPACE};
-use std::borrow::Cow;
+use std::sync::Arc;
 
 pub struct OpenTelemetryLogBridge<P, L>
 where
@@ -170,11 +170,13 @@ where
     L: Logger + Send + Sync,
 {
     pub fn new(provider: &P) -> Self {
-        OpenTelemetryLogBridge {
-            logger: provider
-                .logger_builder("opentelemetry-log-appender")
-                .with_version(Cow::Borrowed(env!("CARGO_PKG_VERSION")))
+        let library = Arc::new(
+            InstrumentationLibrary::builder("opentelemetry-log-appender")
+                .with_version(env!("CARGO_PKG_VERSION"))
                 .build(),
+        );
+        OpenTelemetryLogBridge {
+            logger: provider.library_logger(library),
             _phantom: Default::default(),
         }
     }
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -97,11 +97,10 @@
 use log::{Level, Metadata, Record};
 use opentelemetry::{
     logs::{AnyValue, LogRecord, Logger, LoggerProvider, Severity},
-    InstrumentationLibrary, Key,
+    InstrumentationScope, Key,
 };
 #[cfg(feature = "experimental_metadata_attributes")]
 use opentelemetry_semantic_conventions::attribute::{CODE_FILEPATH, CODE_LINENO, CODE_NAMESPACE};
-use std::sync::Arc;
 
 pub struct OpenTelemetryLogBridge<P, L>
 where
@@ -170,13 +169,12 @@ where
     L: Logger + Send + Sync,
 {
     pub fn new(provider: &P) -> Self {
-        let library = Arc::new(
-            InstrumentationLibrary::builder("opentelemetry-log-appender")
-                .with_version(env!("CARGO_PKG_VERSION"))
-                .build(),
-        );
+        let scope = InstrumentationScope::builder("opentelemetry-log-appender")
+            .with_version(env!("CARGO_PKG_VERSION"))
+            .build();
+
         OpenTelemetryLogBridge {
-            logger: provider.library_logger(library),
+            logger: provider.logger_with_scope(scope),
             _phantom: Default::default(),
         }
     }
--- a/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
@@ -3,7 +3,7 @@ use opentelemetry::{
     global,
     metrics::MetricsError,
     trace::{TraceContextExt, TraceError, Tracer},
-    InstrumentationLibrary, KeyValue,
+    InstrumentationScope, KeyValue,
 };
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
 use opentelemetry_otlp::WithExportConfig;
@@ -18,7 +18,7 @@ use opentelemetry_sdk::{
     logs::{self as sdklogs},
     Resource,
 };
-use std::{error::Error, sync::Arc};
+use std::error::Error;
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
@@ -128,14 +128,13 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         .init();
 
     let common_scope_attributes = vec![KeyValue::new("scope-key", "scope-value")];
-    let library = Arc::new(
-        InstrumentationLibrary::builder("basic")
-            .with_version("1.0")
-            .with_attributes(common_scope_attributes)
-            .build(),
-    );
-    let tracer = global::library_tracer(library.clone());
-    let meter = global::library_meter(library);
+    let scope = InstrumentationScope::builder("basic")
+        .with_version("1.0")
+        .with_attributes(common_scope_attributes)
+        .build();
+
+    let tracer = global::tracer_with_scope(scope.clone());
+    let meter = global::meter_with_scope(scope);
 
     let counter = meter
         .u64_counter("test_counter")
--- a/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
@@ -2,8 +2,8 @@ use once_cell::sync::Lazy;
 use opentelemetry::{
     global,
     metrics::MetricsError,
-    trace::{TraceContextExt, TraceError, Tracer, TracerProvider as _},
-    KeyValue,
+    trace::{TraceContextExt, TraceError, Tracer},
+    InstrumentationLibrary, KeyValue,
 };
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
 use opentelemetry_otlp::WithExportConfig;
@@ -18,7 +18,7 @@ use opentelemetry_sdk::{
     logs::{self as sdklogs},
     Resource,
 };
-use std::error::Error;
+use std::{error::Error, sync::Arc};
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
@@ -128,16 +128,14 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         .init();
 
     let common_scope_attributes = vec![KeyValue::new("scope-key", "scope-value")];
-    let tracer = global::tracer_provider()
-        .tracer_builder("basic")
-        .with_attributes(common_scope_attributes.clone())
-        .build();
-    let meter = global::meter_with_version(
-        "basic",
-        Some("v1.0"),
-        Some("schema_url"),
-        Some(common_scope_attributes.clone()),
+    let library = Arc::new(
+        InstrumentationLibrary::builder("basic")
+            .with_version("1.0")
+            .with_attributes(common_scope_attributes)
+            .build(),
     );
+    let tracer = global::library_tracer(library.clone());
+    let meter = global::library_meter(library);
 
     let counter = meter
         .u64_counter("test_counter")
--- a/opentelemetry-otlp/examples/basic-otlp/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp/src/main.rs
@@ -1,9 +1,9 @@
 use once_cell::sync::Lazy;
 use opentelemetry::logs::LogError;
 use opentelemetry::metrics::MetricsError;
-use opentelemetry::trace::{TraceContextExt, TraceError, Tracer, TracerProvider};
+use opentelemetry::trace::{TraceContextExt, TraceError, Tracer};
 use opentelemetry::KeyValue;
-use opentelemetry::{global, InstrumentationLibrary};
+use opentelemetry::{global, InstrumentationScope};
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
 use opentelemetry_otlp::{LogExporter, MetricsExporter, SpanExporter, WithExportConfig};
 use opentelemetry_sdk::logs::LoggerProvider;
@@ -11,7 +11,6 @@ use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::trace::Config;
 use opentelemetry_sdk::{runtime, trace as sdktrace, Resource};
 use std::error::Error;
-use std::sync::Arc;
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
@@ -106,14 +105,13 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         .init();
 
     let common_scope_attributes = vec![KeyValue::new("scope-key", "scope-value")];
-    let library = Arc::new(
-        InstrumentationLibrary::builder("basic")
-            .with_version("1.0")
-            .with_attributes(common_scope_attributes)
-            .build(),
-    );
-    let tracer = global::tracer_provider().library_tracer(library.clone());
-    let meter = global::library_meter(library);
+    let scope = InstrumentationScope::builder("basic")
+        .with_version("1.0")
+        .with_attributes(common_scope_attributes)
+        .build();
+
+    let tracer = global::tracer_with_scope(scope.clone());
+    let meter = global::meter_with_scope(scope);
 
     let counter = meter
         .u64_counter("test_counter")
--- a/opentelemetry-otlp/examples/basic-otlp/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp/src/main.rs
@@ -1,12 +1,9 @@
 use once_cell::sync::Lazy;
-use opentelemetry::global;
 use opentelemetry::logs::LogError;
 use opentelemetry::metrics::MetricsError;
-use opentelemetry::trace::{TraceError, TracerProvider};
-use opentelemetry::{
-    trace::{TraceContextExt, Tracer},
-    KeyValue,
-};
+use opentelemetry::trace::{TraceContextExt, TraceError, Tracer, TracerProvider};
+use opentelemetry::KeyValue;
+use opentelemetry::{global, InstrumentationLibrary};
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
 use opentelemetry_otlp::{LogExporter, MetricsExporter, SpanExporter, WithExportConfig};
 use opentelemetry_sdk::logs::LoggerProvider;
@@ -14,6 +11,7 @@ use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::trace::Config;
 use opentelemetry_sdk::{runtime, trace as sdktrace, Resource};
 use std::error::Error;
+use std::sync::Arc;
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
@@ -108,16 +106,14 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
         .init();
 
     let common_scope_attributes = vec![KeyValue::new("scope-key", "scope-value")];
-    let tracer = global::tracer_provider()
-        .tracer_builder("basic")
-        .with_attributes(common_scope_attributes.clone())
-        .build();
-    let meter = global::meter_with_version(
-        "basic",
-        Some("v1.0"),
-        Some("schema_url"),
-        Some(common_scope_attributes.clone()),
+    let library = Arc::new(
+        InstrumentationLibrary::builder("basic")
+            .with_version("1.0")
+            .with_attributes(common_scope_attributes)
+            .build(),
     );
+    let tracer = global::tracer_provider().library_tracer(library.clone());
+    let meter = global::library_meter(library);
 
     let counter = meter
         .u64_counter("test_counter")
--- a/opentelemetry-sdk/src/export/logs/mod.rs
+++ b/opentelemetry-sdk/src/export/logs/mod.rs
@@ -11,11 +11,11 @@ use std::fmt::Debug;
 /// A batch of log records to be exported by a `LogExporter`.
 ///
 /// The `LogBatch` struct holds a collection of log records along with their associated
-/// instrumentation libraries. This structure is used to group log records together for efficient
+/// instrumentation scopes. This structure is used to group log records together for efficient
 /// export operations.
 ///
 /// # Type Parameters
-/// - `'a`: The lifetime of the references to the log records and instrumentation libraries.
+/// - `'a`: The lifetime of the references to the log records and instrumentation scopes.
 ///
 #[derive(Debug)]
 pub struct LogBatch<'a> {
@@ -31,11 +31,11 @@ impl<'a> LogBatch<'a> {
     ///
     /// * `data` - A slice of tuples, where each tuple consists of a reference to a `LogRecord`
     ///   and a reference to an `InstrumentationScope`. These tuples represent the log records
-    ///   and their associated instrumentation libraries to be exported.
+    ///   and their associated instrumentation scopes to be exported.
     ///
     /// # Returns
     ///
-    /// A `LogBatch` instance containing the provided log records and instrumentation libraries.
+    /// A `LogBatch` instance containing the provided log records and instrumentation scopes.
     ///
     /// Note - this is not a public function, and should not be used directly. This would be
     /// made private in the future.
@@ -46,7 +46,7 @@ impl<'a> LogBatch<'a> {
 }
 
 impl LogBatch<'_> {
-    /// Returns an iterator over the log records and instrumentation libraries in the batch.
+    /// Returns an iterator over the log records and instrumentation scopes in the batch.
     ///
     /// Each item yielded by the iterator is a tuple containing references to a `LogRecord`
     /// and an `InstrumentationScope`.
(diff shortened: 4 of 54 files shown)


opentelemetry-appender-log/src/lib.rs
```diff
@@ -170,11 +170,10 @@ where
     L: Logger + Send + Sync,
 {
     pub fn new(provider: &P) -> Self {
+        let library =
+            Arc::new(InstrumentationLibrary::builder("opentelemetry-log-appender").build());
```
cijothomas, line 174: lets keep the with_version as before.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -121,16 +121,14 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
-        .build();
-    let meter = global::meter_with_version(
-        "basic",
-        Some("v1.0"),
-        Some("schema_url"),
-        Some(common_scope_attributes.clone()),
+    let library = Arc::new(
+        InstrumentationLibrary::builder("basic")
+            .with_version("1.0")
+            .with_attributes(common_scope_attributes.clone())
```
cijothomas, line 127:
```suggestion
            .with_attributes(common_scope_attributes)
```

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -112,16 +110,14 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
-        .build();
-    let meter = global::meter_with_version(
-        "basic",
-        Some("v1.0"),
-        Some("schema_url"),
-        Some(common_scope_attributes.clone()),
+    let library = Arc::new(
+        InstrumentationLibrary::builder("basic")
+            .with_version("1.0")
+            .with_attributes(common_scope_attributes.clone())
```
cijothomas, line 116:
```suggestion
            .with_attributes(common_scope_attributes)
```

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -112,16 +110,14 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
+            .build(),
     );
+    let tracer = global::tracer_provider().library_tracer(library.clone());
```
cijothomas, line 119: wondering why tracing is done slightly differently than metric?

opentelemetry-stdout/examples/basic.rs
```diff
@@ -64,16 +64,20 @@ fn init_logs() -> opentelemetry_sdk::logs::LoggerProvider {
...
+        InstrumentationLibrary,
     };
 
-    let tracer = global::tracer_provider()
-        .tracer_builder("stdout-example")
-        .with_version("v1")
-        .with_schema_url("schema_url")
-        .with_attributes([KeyValue::new("scope_key", "scope_value")])
-        .build();
+    let library = Arc::new(
```
cijothomas, line 74: From a useability standpoint, I think its best to not require Arc here. Cloning InstrumentationLibrary should be cheap anyway.

opentelemetry/src/global/metrics.rs
```diff
@@ -44,26 +44,26 @@ pub fn meter(name: &'static str) -> Meter {
...
+/// use opentelemetry::InstrumentationLibrary;
 /// use opentelemetry::KeyValue;
 ///
-/// let meter = meter_with_version(
-///     "io.opentelemetry",
-///     Some("0.17"),
-///     Some("https://opentelemetry.io/schemas/1.2.0"),
-///     Some(vec![KeyValue::new("key", "value")]),
+/// let library = Arc::new(
+///     InstrumentationLibrary::builder("io.opentelemetry")
```
cijothomas, line 58: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter

I think we should rename InstrumentationLibrary to InstrumentationScope everywhere, as Scope is what user provides.

cijothomas (commented):
Left few comments but looks good overall.
Please check the comments and see if they make sense.

cijothomas:
> There were some concerns about having `Arc` in the `library_signal` APIs, but someone also mentioned we want to avoid unnecessary copies. We could hide the `Arc` inside the `InstrumentationLibrary` if needed.

yes I suggest to hide Arc inside.

opentelemetry/src/global/metrics.rs
cijothomas: @lalitb @utpilla Please share your thoughts on this.

utpilla: Yes. `InstrumentationScope` also matches the term used in [proto file](https://github.com/open-telemetry/opentelemetry-proto/blob/7312bdf63218acf27fe96430b7231de37fd091f2/opentelemetry/proto/common/v1/common.proto#L69-L81).

lalitb: +1 for `InstrumentationScope`

stormshield-fabs: done

opentelemetry-otlp/examples/basic-otlp/src/main.rs
stormshield-fabs: implemented `global::library_tracer`

opentelemetry-stdout/examples/basic.rs
stormshield-fabs: `InstrumentationLibrary` was already cheap to clone, except for the `attributes`. I've put them in a `Cow` so we no longer need to use `Arc`.

opentelemetry/src/common.rs
```diff
@@ -402,82 +402,55 @@ pub trait ExportError: std::error::Error + Send + Sync + 'static {
...
     pub schema_url: Option<Cow<'static, str>>,
 
     /// Specifies the instrumentation scope attributes to associate with emitted telemetry.
-    pub attributes: Vec<KeyValue>,
+    pub attributes: Cow<'static, Vec<KeyValue>>,
 }
 
-// Uniqueness for InstrumentationLibrary/InstrumentationScope does not depend on attributes
-impl Eq for InstrumentationLibrary {}
+// Uniqueness for InstrumentationScope/InstrumentationScope does not depend on attributes
```
cijothomas, line 431: there was a recent clarification that scope attributes also contribute to uniqueness test.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md#logs-1

Let fix in a follow up PR to keep this PR's scope.

opentelemetry-stdout/examples/basic.rs
utpilla: I noticed that you changed it to `pub attributes: Cow<'static, Vec<KeyValue>>`. How does wrapping the vec with `Cow` help?

Did you mean to change that to cow of static slice instead? `Cow<'static, [KeyValue]>`?

cijothomas: Its okay to not optimize for cloning cost. 
Cloning same instrumentation-scope is to be done at max for 2 times (one for each signal), so it is okay!
(Or on the future, we can do Scope backed by a Arc<InnerScope> so clone is super cheap (if we ever need that))

opentelemetry/src/metrics/meter.rs
```diff
@@ -6,13 +6,38 @@ use crate::metrics::{
...
+    ///
+    /// ```
+    /// use std::sync::Arc;
+    /// use opentelemetry::InstrumentationScope;
+    /// use opentelemetry::metrics::MeterProvider;
+    /// use opentelemetry_sdk::metrics::SdkMeterProvider;
+    ///
+    /// let provider = SdkMeterProvider::default();
+    ///
+    /// // logger used in applications/binaries
```
utpilla, line 28: Looks like a copy-paste error. Replace logger with `meter` in the comments and variable names.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -145,34 +145,14 @@ impl Drop for SdkMeterProviderInner {
...
-        if let Some(a) = attributes {
-            builder = builder.with_attributes(a);
-        }
-
-        let scope = builder.build();
-
         if let Ok(mut meters) = self.inner.meters.lock() {
             let meter = meters
-                .entry(scope)
+                .entry(InstrumentationScope::clone(&scope))
```
utpilla, line 155: No need to clone here.

```suggestion
                .entry(scope)
```

opentelemetry/src/common.rs
lalitb, line 431:
```suggestion
// Uniqueness for InstrumentationScope does not depend on attributes
```

opentelemetry/src/common.rs
```diff
@@ -402,82 +402,55 @@ pub trait ExportError: std::error::Error + Send + Sync + 'static {
...
+/// An instrumentation scope should be named to follow any naming conventions
 /// of the instrumented library (e.g. 'middleware' for a web framework).
 ///
 /// See the [instrumentation libraries] spec for more information.
 ///
 /// [instrumentation libraries]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/overview.md#instrumentation-libraries
 #[derive(Debug, Default, Clone)]
 #[non_exhaustive]
-pub struct InstrumentationLibrary {
+pub struct InstrumentationScope {
```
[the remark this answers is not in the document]
cijothomas: this should be a relative easy renaming for users, so I don't think its worth maintaining the alias for few releases. 
This is part of `opentelemetry` and we are aiming to declare it RC in a week and stable soon after that.

opentelemetry/src/common.rs
```diff
@@ -556,13 +529,13 @@ impl InstrumentationLibraryBuilder {
...
-    pub fn build(self) -> InstrumentationLibrary {
-        InstrumentationLibrary {
+    /// Create a new [InstrumentationScope] from this configuration
+    pub fn build(self) -> InstrumentationScope {
+        InstrumentationScope {
             name: self.name,
             version: self.version,
             schema_url: self.schema_url,
-            attributes: self.attributes.unwrap_or_default(),
+            attributes: Cow::Owned(self.attributes.unwrap_or_default()),
```
lalitb, line 538: Just to confirm, if users want to take advantage of borrowed attributes, they can bypass the builder pattern and directly initialize the `InstrumentationScope` struct.

opentelemetry/src/common.rs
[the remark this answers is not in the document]
lalitb: > this should be a relative easy renaming for users, so I don't think its worth maintaining the alias for few releases.
This is part of opentelemetry and we are aiming to declare it RC in a week and stable soon after that.

Yes, but in current form, this will break their instrumentation build, which is not good. We can still remove the deprecated code after reaching RC.

opentelemetry/src/logs/logger.rs
```diff
@@ -30,128 +30,37 @@ pub trait LoggerProvider {
...
-    /// let library = std::sync::Arc::new(
-    ///     InstrumentationLibrary::builder(env!("CARGO_PKG_NAME"))
-    ///         .with_version(env!("CARGO_PKG_VERSION"))
-    ///         .with_schema_url("https://opentelemetry.io/schema/1.0.0")
-    ///         .build(),
-    /// );
-    /// let logger = provider.library_logger(library);
-    /// ```
-    fn library_logger(&self, library: Arc<InstrumentationLibrary>) -> Self::Logger;
+    fn library_logger(&self, scope: InstrumentationScope) -> Self::Logger;
```
cijothomas, line 55: we need to rename `library_logger` to something more indicative of its purpose...
provider.logger("name")
provider.logger_with_scope(scope);

open to suggestions.

opentelemetry/src/common.rs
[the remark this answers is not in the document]
cijothomas: > We can still remove the deprecated code after reaching RC.

You are right! However, this PR will break users pretty bad as it is not just renaming `InstrumentationLibrary`. I don't think its feasible to make this in a non-breaking way, and I feel its best to do this and unblock RC release of all APIs, so users know there won't be any further breaking change for instrumentations in the future.

(It might be possible to find ways to smoothen breaking change, but that effort might delay RC again and may not yield much value ultimately)

opentelemetry/src/common.rs
cijothomas: Tracked via https://github.com/open-telemetry/opentelemetry-rust/issues/2225

cijothomas:
@stormshield-fabs Thanks a lot! I think the following are to be addressed before merge. We also need to write a migration guide in the PR description showing before/after, so users can easily (relatively) react to this.

Lets keep InstrumentationScope simple, and accept the perf cost now, given it is infrequent operation.
https://github.com/open-telemetry/opentelemetry-rust/pull/2220#discussion_r1809335125

Rename library-signal to something like signal_with_scope/better names.
https://github.com/open-telemetry/opentelemetry-rust/pull/2220#discussion_r1809701673

opentelemetry/src/logs/logger.rs
lalitb: `logger_with_scope` looks fine to me.

opentelemetry-stdout/examples/basic.rs
stormshield-fabs: I've reverted the `attributes` type, but this raises the question of using `Cow` for the other struct members.

opentelemetry/src/metrics/meter.rs
stormshield-fabs: fixed

opentelemetry-sdk/src/metrics/meter_provider.rs
stormshield-fabs: fixed

opentelemetry/src/common.rs
stormshield-fabs: fixed

opentelemetry/src/common.rs
stormshield-fabs: Reverted `attributes` to its initial type as per comments above.

cijothomas (approved):
LGTM.
Could you modify the PR description to show the before/after example for logs/metrics/traces, so users can easily do the migration?

cijothomas:
@stormshield-fabs oops we just cause some merge conflicts! Could you resolve them.
We can merge with one more approval.
(Also, please update pr description to show before/after code snippets to make it easy for users to migrate)

opentelemetry/src/common.rs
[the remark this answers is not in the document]
cijothomas: From SIG meeting discussion:
1. We acknowledge the amount of breaking changes
2. We are delaying the 0.27 release by a week, to ensure we can finally declare `opentelemetry` crates logs, metrics, common as RC, to assure end users that the breaking changes are ending finally!
3. It is still technically possible to make breaking changes after RC, but the bar is extremely high.
4. We also should be declaring the `opentelemetry` metrics,logs,common things stable in the next ~1 month or so.
5. Because of this, its best to accept the cost of this breaking change now.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -45,48 +45,15 @@ pub struct LoggerProvider {
...
-    fn versioned_logger(
-        &self,
-        name: impl Into<Cow<'static, str>>,
-        version: Option<Cow<'static, str>>,
-        schema_url: Option<Cow<'static, str>>,
-        attributes: Option<Vec<opentelemetry::KeyValue>>,
-    ) -> Logger {
-        let name = name.into();
-        let component_name = if name.is_empty() {
-            Cow::Borrowed(DEFAULT_COMPONENT_NAME)
```
utpilla, line 64: Are we changing the behavior to not override empty names with the default component name?

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -13,16 +13,14 @@ use crate::trace::{
...
+use crate::{Resource, Scope};
 use once_cell::sync::{Lazy, OnceCell};
 use opentelemetry::trace::TraceError;
 use opentelemetry::{global, trace::TraceResult};
 use std::borrow::Cow;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
-/// Default tracer name if empty string is provided.
-const DEFAULT_COMPONENT_NAME: &str = "rust.opentelemetry.io/sdk/tracer";
```
utpilla, line 25: Same as this https://github.com/open-telemetry/opentelemetry-rust/pull/2220/files#r1811937347

opentelemetry/src/global/trace.rs
```diff
@@ -395,6 +389,33 @@ pub fn tracer(name: impl Into<Cow<'static, str>>) -> BoxedTracer {
...
+/// use opentelemetry::InstrumentationScope;
+/// use opentelemetry::KeyValue;
+///
+/// let scope = InstrumentationScope::builder("io.opentelemetry")
+///     .with_version("0.17")
+///     .with_schema_url("https://opentelemetry.io/schema/1.2.0")
+///     .with_attributes(vec![(KeyValue::new("key", "value"))])
+///     .build();
+///
+/// let meter = tracer_with_scope(scope);
```
utpilla, line 411:
```suggestion
/// let tracer = tracer_with_scope(scope);
```

opentelemetry-sdk/src/lib.rs
```diff
@@ -146,6 +146,7 @@ pub mod trace;
 #[doc(hidden)]
 pub mod util;
 
-pub use instrumentation::{InstrumentationLibrary, Scope};
+#[doc(inline)]
+pub use opentelemetry::InstrumentationScope as Scope;
```
lalitb, line 150: I believe we can remove this aliasing. Or atleast not expose this to end-user, if we want to use it internally in sdk. I understand this is coming from instrumentation.rs, but is not necessary.

opentelemetry-sdk/src/logs/log_emitter.rs
stormshield-fabs: I suggest we move the `signal` methods to `named_signal` and let the trait implementers handle the logic around empty names in their implementations of `signal`:

```rust
trait LoggerProvider {
    fn named_logger(&self, name: impl Into<Cow<'static, str>>) -> Self::Logger {
        let scope = InstrumentationScope::builder(name).build();
        self.logger_with_scope(scope)
    }
    
    fn logger(&self, name: impl Into<Cow<'static, str>>) -> Self::Logger;

    fn logger_with_scope(&self, scope: InstrumentationScope) -> Self::Logger;
}
```

Or maybe we should simply get rid of `named_signal`?

stormshield-fabs: I went with the overrides in a separate commit, let me know what you think!

opentelemetry-sdk/src/lib.rs
stormshield-fabs: done

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -144,32 +144,25 @@ impl Drop for SdkMeterProviderInner {
         }
     }
 }
+
+/// Default logger name if empty string is provided.
```
utpilla, line 148:
```suggestion
/// Default meter name if empty string is provided.
```

opentelemetry/src/global/metrics.rs
```diff
@@ -38,32 +38,26 @@ pub fn meter(name: &'static str) -> Meter {
     meter_provider().meter(name)
 }
 
-/// Creates a [`Meter`] with the name, version and schema url.
+/// Creates a [`Meter`] with the given instrumentation library.
```
utpilla, line 41:
```suggestion
/// Creates a [`Meter`] with the given instrumentation scope.
```

opentelemetry/src/global/metrics.rs
```diff
@@ -38,32 +38,26 @@ pub fn meter(name: &'static str) -> Meter {
...
 ///
-/// - name SHOULD uniquely identify the instrumentation scope, such as the instrumentation library (e.g. io.opentelemetry.contrib.mongodb), package, module or class name.
-/// - version specifies the version of the instrumentation scope if the scope has a version
-/// - schema url specifies the Schema URL that should be recorded in the emitted telemetry.
-///
-/// This is a convenient way of `global::meter_provider().versioned_meter(...)`
+/// This is a shortcut `global::meter_provider().meter_with_scope(...)`
```
utpilla, line 43: nit:

```suggestion
/// This is a simpler alternative to `global::meter_provider().meter_with_scope(...)`
```

opentelemetry/src/global/mod.rs
```diff
@@ -50,16 +50,21 @@
...
-//!         with_version(env!("CARGO_PKG_VERSION")).
-//!         with_schema_url("https://opentelemetry.io/schemas/1.17.0").
-//!         build();
+//!
+//!     let scope = InstrumentationScope::builder("my_library-name")
+//!         .with_version(env!("CARGO_PKG_VERSION"))
+//!         .with_schema_url("https://opentelemetry.io/schemas/1.17.0")
+//!         .build();
+//!
+//!     let tracer = global::tracer_provider().tracer_with_scope(scope);
```
utpilla, line 67:
```suggestion
//!     let tracer = global::tracer_with_scope(scope);
```

opentelemetry/src/global/trace.rs
```diff
@@ -395,6 +389,33 @@ pub fn tracer(name: impl Into<Cow<'static, str>>) -> BoxedTracer {
     tracer_provider().tracer(name.into())
 }
 
+/// Creates a [`Tracer`] with the given instrumentation library
+/// via the configured [`GlobalTracerProvider`].
+///
+/// This is a shortcut `global::tracer_provider().tracer_with_scope(...)`
```
utpilla, line 395: nit:
```suggestion
/// This is a simpler alternative to `global::tracer_provider().tracer_with_scope(...)`
```

opentelemetry/src/logs/logger.rs
```diff
@@ -30,128 +30,37 @@ pub trait LoggerProvider {
...
-        }
-        builder.build()
-    }
-
-    /// Returns a new builder for creating a [`Logger`] instance
-    ///
-    /// The `name` should be the application name or the name of the library
-    /// providing instrumentation. If the name is empty, then an
-    /// implementation-defined default name may be used instead.
+    /// Returns a new logger with the given instrumentation library.
```
utpilla, line 33:
```suggestion
    /// Returns a new logger with the given instrumentation scope.
```

opentelemetry/src/global/trace.rs
```diff
@@ -395,6 +389,33 @@ pub fn tracer(name: impl Into<Cow<'static, str>>) -> BoxedTracer {
     tracer_provider().tracer(name.into())
 }
 
+/// Creates a [`Tracer`] with the given instrumentation library
```
utpilla, line 392:
```suggestion
/// Creates a [`Tracer`] with the given instrumentation scope
```

opentelemetry/src/metrics/meter.rs
```diff
@@ -6,13 +6,38 @@ use crate::metrics::{
...
+    ///
+    /// // meter used in libraries/crates that optionally includes version and schema url
+    /// let scope = InstrumentationScope::builder(env!("CARGO_PKG_NAME"))
+    ///     .with_version(env!("CARGO_PKG_VERSION"))
+    ///     .with_schema_url("https://opentelemetry.io/schema/1.0.0")
+    ///     .build();
+    ///
+    /// let meter = provider.meter_with_scope(scope);
+    /// ```
+    fn meter_with_scope(&self, scope: InstrumentationScope) -> Meter;
```
utpilla, line 39: nit:

It would be nice to be consistent with the ordering of these methods. Both `LoggerProvider` and `TracerProvider` traits have the simple function (`logger`, `tracer`) defined declared before the `signal_with_scope` method.

utpilla (approved):
Left some minor suggestions.

Great work @stormshield-fabs!!

opentelemetry-appender-log/src/lib.rs
```diff
@@ -170,11 +169,12 @@ where
     L: Logger + Send + Sync,
 {
     pub fn new(provider: &P) -> Self {
+        let scope = InstrumentationScope::builder("opentelemetry-log-appender")
```
lalitb, line 172: This scope name should be `opentelemetry-appender-log` , to be consistent with [tracing appender](https://github.com/open-telemetry/opentelemetry-rust/blob/8bd529a6d629aff7482b875cfc39275a8a71eaeb/opentelemetry-appender-tracing/src/layer.rs#L13), so as to follow the convention of `opentelemetry-appender-<name>`. 
But this is not related to PR, I will raise a separate PR for this.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -22,8 +20,8 @@ use std::fmt::Debug;
 #[derive(Debug)]
 pub struct LogBatch<'a> {
     /// The data field contains a slice of tuples, where each tuple consists of a reference to
-    /// a `LogRecord` and a reference to an `InstrumentationLibrary`.
-    data: &'a [(&'a LogRecord, &'a InstrumentationLibrary)],
+    /// a `LogRecord` and a reference to an `InstrumentationScope`.
```
lalitb, line 23: nit - change line 14, 18 (and few more) in this file to indicate `instrumentation scope` instead of `instrumentation libraries`.

lalitb (approved):
LGTM. Thanks for the changes.

cijothomas:
@stormshield-fabs Can you resolve conflicts and address the nits, and we can merge!

As suggested offline, if you can add a migration guide in the PR description, similar to https://github.com/open-telemetry/opentelemetry-rust/pull/2221, that'd be awesome!

(The next release is a flood of breaking changes, so we need to write an overall migration guide, which will refer to the sub-migration-guides within each PR description.)

lalitb:
@stormshield-fabs  I took the liberty of fixing minor merge conflicts in this PR, as it was introduced by changes from my own PR. Hope that's fine :)

PR #2221, opened 2024-10-19
feat: replace OtlpPipeline with exporter builders

### LINKED ISSUE #1810: OTLPExporter Pipeline issues

pitoniak32:
Fixes #1810

# Migration Guide

To move from `opentelemetry_otlp::new_exporter()`, and `opentelemetry_otlp::new_pipeline().{tracing,metrics,logging}()`, you will need to select the `Exporter`, and `Provider` for the signal you are using.

Below there is a guide with details about Exporters and Providers. There are also Examples for each type of signal.

## Exporter Guide
The following are the available exporters:
- SpanExporter
- MetricsExporter
- LogExporter

The exporter interface should have the same options as with the old method. For example, a tonic grpc exporter being configured old vs new:
```rust
// old
let old_exporter = opentelemetry_otlp::new_exporter()
  .tonic()
  .with_endpoint("http://localhost:4317");

// new
let new_exporter = SpanExporter::builder()
  .tonic()
  .with_endpoint("http://localhost:4317")
  .build()?; // note: you need to build the exporter after setting configuration.
```

## Provider Guide
The following are the available providers:
- TracingProvider
- SdkMeterProvider
- LoggerProvider

The provider interface should be similar to the old method. For example, a `TracerProvider` being configured old vs new:
```rust
// old
let tracer_provider: TracerProvider = opentelemetry_otlp::new_pipeline()
    .tracing()
    .with_exporter(old_exporter /* ^^ See exporter guide above ^^ */)
    .with_trace_config(Config::default().with_resource(RESOURCE.clone()))
    .install_batch(runtime::Tokio)?;

// new
let tracer_provider: TracerProvider = TracerProvider::builder()
    .with_batch_exporter(new_exporter /* ^^ See exporter guide above ^^ */, runtime::Tokio)
    .with_config(Config::default().with_resource(RESOURCE.clone()))
    .build(); // note: you need to build the provider after setting configuration.
```

# Signal Examples
### Trace Example
```rust
// old
fn init_tracer_provider() -> Result<sdktrace::TracerProvider, TraceError> {
        .tracing()
        .with_exporter(
                .tonic()
                .with_endpoint("http://localhost:4317"),
        )
        .with_trace_config(Config::default().with_resource(RESOURCE.clone()))
        .install_batch(runtime::Tokio)
}

// new
fn init_tracer_provider() -> Result<sdktrace::TracerProvider, TraceError> {
        .with_tonic()
        .with_endpoint("http://localhost:4317")
        .build()?;
    Ok(sdktrace::TracerProvider::builder()
        .with_config(Config::default().with_resource(RESOURCE.clone()))
        .with_batch_exporter(exporter, runtime::Tokio)
        .build())
}
```

### Metric Example
```rust
// old
fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, MetricsError> {
    let export_config = ExportConfig {
        endpoint: "http://localhost:4317".to_string(),
        ..ExportConfig::default()
    };
        .metrics(runtime::Tokio)
        .with_exporter(
                .tonic()
                .with_export_config(export_config),
        )
}

// new
fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, MetricsError> {
    let exporter = MetricsExporter::builder().with_tonic().build()?;

    Ok(SdkMeterProvider::builder()
        .with_reader(reader)
        .with_resource(RESOURCE.clone())
        .build())
}
```
### Log Example
```rust
// old
fn init_logs() -> Result<opentelemetry_sdk::logs::LoggerProvider, LogError> {
        .logging()
        .with_resource(RESOURCE.clone())
        .with_exporter(
                .tonic()
                .with_endpoint("http://localhost:4317"),
        )
        .install_batch(runtime::Tokio)
        .with_batch_exporter(exporter, runtime::Tokio)
        .build())
}

// new
fn init_logs() -> Result<opentelemetry_sdk::logs::LoggerProvider, LogError> {
    let exporter = LogExporter::builder()
        .with_tonic()
        .with_endpoint("http://localhost:4317")
        .build()?;
  
    Ok(LoggerProvider::builder()
        .with_resource(RESOURCE.clone())
        .with_exporter(
            exporter,
        )
        .with_batch_exporter(exporter, runtime::Tokio)
        .build())
}
```

## Changes

Started to replace the `OTLPPipeline` pattern in the `opentelemetry-otlp` crate.

### DIFF
--- a/examples/self-diagnostics/src/main.rs
+++ b/examples/self-diagnostics/src/main.rs
@@ -1,7 +1,8 @@
 use opentelemetry::global::{self, set_error_handler, Error as OtelError};
 use opentelemetry::KeyValue;
 use opentelemetry_appender_tracing::layer;
-use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_otlp::{LogExporter, MetricsExporter, WithExportConfig};
+use opentelemetry_sdk::metrics::PeriodicReader;
 use tracing_subscriber::filter::{EnvFilter, LevelFilter};
 use tracing_subscriber::fmt;
 use tracing_subscriber::prelude::*;
@@ -51,15 +52,16 @@ fn custom_error_handler(err: OtelError) {
 }
 
 fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
-    let provider = opentelemetry_otlp::new_pipeline()
-        .logging()
-        .with_exporter(
-            opentelemetry_otlp::new_exporter()
-                .http()
-                .with_endpoint("http://localhost:4318/v1/logs"),
-        )
-        .install_batch(opentelemetry_sdk::runtime::Tokio)
+    let exporter = LogExporter::builder()
+        .with_http()
+        .with_endpoint("http://localhost:4318/v1/logs")
+        .build()
         .unwrap();
+
+    let provider = opentelemetry_sdk::logs::LoggerProvider::builder()
+        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::Tokio)
+        .build();
+
     let cloned_provider = provider.clone();
 
     // Add a tracing filter to filter events from crates used by opentelemetry-otlp.
@@ -107,16 +109,20 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
 }
 
 fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
-    let provider = opentelemetry_otlp::new_pipeline()
-        .metrics(opentelemetry_sdk::runtime::Tokio)
-        .with_period(std::time::Duration::from_secs(1))
-        .with_exporter(
-            opentelemetry_otlp::new_exporter()
-                .http()
-                .with_endpoint("http://localhost:4318/v1/metrics"),
-        )
+    let exporter = MetricsExporter::builder()
+        .with_tonic()
+        .with_endpoint("http://localhost:4318/v1/metrics")
         .build()
         .unwrap();
+
+    let reader = PeriodicReader::builder(exporter, opentelemetry_sdk::runtime::Tokio)
+        .with_interval(std::time::Duration::from_secs(1))
+        .build();
+
+    let provider = opentelemetry_sdk::metrics::SdkMeterProvider::builder()
+        .with_reader(reader)
+        .build();
+
     let cloned_provider = provider.clone();
     global::set_meter_provider(cloned_provider);
     provider
--- a/examples/self-diagnostics/src/main.rs
+++ b/examples/self-diagnostics/src/main.rs
@@ -110,7 +110,7 @@ fn init_logger_provider() -> opentelemetry_sdk::logs::LoggerProvider {
 
 fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
     let exporter = MetricsExporter::builder()
-        .with_tonic()
+        .with_http()
         .with_endpoint("http://localhost:4318/v1/metrics")
         .build()
         .unwrap();
--- a/examples/tracing-jaeger/src/main.rs
+++ b/examples/tracing-jaeger/src/main.rs
@@ -4,27 +4,26 @@ use opentelemetry::{
     trace::{TraceContextExt, TraceError, Tracer},
     KeyValue,
 };
-use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_sdk::trace::TracerProvider;
 use opentelemetry_sdk::{runtime, trace as sdktrace, Resource};
 use opentelemetry_semantic_conventions::resource::SERVICE_NAME;
 
 use std::error::Error;
 
 fn init_tracer_provider() -> Result<opentelemetry_sdk::trace::TracerProvider, TraceError> {
-    opentelemetry_otlp::new_pipeline()
-        .tracing()
-        .with_exporter(
-            opentelemetry_otlp::new_exporter()
-                .tonic()
-                .with_endpoint("http://localhost:4317"),
-        )
-        .with_trace_config(
+    let exporter = opentelemetry_otlp::SpanExporter::builder()
+        .with_tonic()
+        .build()?;
+
+    Ok(TracerProvider::builder()
+        .with_batch_exporter(exporter, runtime::Tokio)
+        .with_config(
             sdktrace::Config::default().with_resource(Resource::new(vec![KeyValue::new(
                 SERVICE_NAME,
                 "tracing-jaeger",
             )])),
         )
-        .install_batch(runtime::Tokio)
+        .build())
 }
 
 #[tokio::main]
--- a/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
@@ -6,9 +6,14 @@ use opentelemetry::{
     KeyValue,
 };
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
-use opentelemetry_otlp::Protocol;
-use opentelemetry_otlp::{HttpExporterBuilder, WithExportConfig};
-use opentelemetry_sdk::trace::{self as sdktrace, Config};
+use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_otlp::{LogExporter, MetricsExporter, Protocol, SpanExporter};
+use opentelemetry_sdk::{
+    logs::LoggerProvider,
+    metrics::{PeriodicReader, SdkMeterProvider},
+    runtime,
+    trace::{self as sdktrace, Config, TracerProvider},
+};
 use opentelemetry_sdk::{
     logs::{self as sdklogs},
     Resource,
@@ -17,6 +22,9 @@ use std::error::Error;
 use tracing::info;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::EnvFilter;
+
+#[cfg(feature = "hyper")]
+use opentelemetry_otlp::WithHttpConfig;
 
 #[cfg(feature = "hyper")]
 mod hyper;
@@ -28,47 +36,46 @@ static RESOURCE: Lazy<Resource> = Lazy::new(|| {
     )])
 });
 
-fn http_exporter() -> HttpExporterBuilder {
-    let exporter = opentelemetry_otlp::new_exporter().http();
-    #[cfg(feature = "hyper")]
-    let exporter = exporter.with_http_client(hyper::HyperClient::default());
-    exporter
-}
-
 fn init_logs() -> Result<sdklogs::LoggerProvider, opentelemetry::logs::LogError> {
-    opentelemetry_otlp::new_pipeline()
-        .logging()
+    let exporter_builder = LogExporter::builder()
+        .with_http()
+        .with_endpoint("http://localhost:4318/v1/logs")
+        .with_protocol(Protocol::HttpBinary);
+
+    #[cfg(feature = "hyper")]
+    let exporter_builder = exporter_builder.with_http_client(hyper::HyperClient::default());
+
+    let exporter = exporter_builder.build()?;
+
+    Ok(LoggerProvider::builder()
+        .with_batch_exporter(exporter, runtime::Tokio)
         .with_resource(RESOURCE.clone())
-        .with_exporter(
-            http_exporter()
-                .with_protocol(Protocol::HttpBinary) //can be changed to `Protocol::HttpJson` to export in JSON format
-                .with_endpoint("http://localhost:4318/v1/logs"),
-        )
-        .install_batch(opentelemetry_sdk::runtime::Tokio)
+        .build())
 }
 
 fn init_tracer_provider() -> Result<sdktrace::TracerProvider, TraceError> {
-    opentelemetry_otlp::new_pipeline()
-        .tracing()
-        .with_exporter(
-            http_exporter()
-                .with_protocol(Protocol::HttpBinary) //can be changed to `Protocol::HttpJson` to export in JSON format
-                .with_endpoint("http://localhost:4318/v1/traces"),
-        )
-        .with_trace_config(Config::default().with_resource(RESOURCE.clone()))
-        .install_batch(opentelemetry_sdk::runtime::Tokio)
+    let exporter = SpanExporter::builder()
+        .with_http()
+        .with_protocol(Protocol::HttpBinary) //can be changed to `Protocol::HttpJson` to export in JSON format
+        .with_endpoint("http://localhost:4318/v1/traces")
+        .build()?;
+    Ok(TracerProvider::builder()
+        .with_batch_exporter(exporter, runtime::Tokio)
+        .with_config(Config::default().with_resource(RESOURCE.clone()))
+        .build())
 }
 
 fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, MetricsError> {
-    opentelemetry_otlp::new_pipeline()
-        .metrics(opentelemetry_sdk::runtime::Tokio)
-        .with_exporter(
-            http_exporter()
-                .with_protocol(Protocol::HttpBinary) //can be changed to `Protocol::HttpJson` to export in JSON format
-                .with_endpoint("http://localhost:4318/v1/metrics"),
-        )
+    let exporter = MetricsExporter::builder()
+        .with_http()
+        .with_protocol(Protocol::HttpBinary) //can be changed to `Protocol::HttpJson` to export in JSON format
+        .with_endpoint("http://localhost:4318/v1/metrics")
+        .build()?;
+
+    Ok(SdkMeterProvider::builder()
+        .with_reader(PeriodicReader::builder(exporter, runtime::Tokio).build())
         .with_resource(RESOURCE.clone())
-        .build()
+        .build())
 }
 
 #[tokio::main]
--- a/opentelemetry-otlp/examples/basic-otlp/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp/src/main.rs
@@ -8,7 +8,9 @@ use opentelemetry::{
     KeyValue,
 };
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
-use opentelemetry_otlp::{ExportConfig, WithExportConfig};
+use opentelemetry_otlp::{LogExporter, MetricsExporter, SpanExporter, WithExportConfig};
+use opentelemetry_sdk::logs::LoggerProvider;
+use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::trace::Config;
 use opentelemetry_sdk::{runtime, trace as sdktrace, Resource};
 use std::error::Error;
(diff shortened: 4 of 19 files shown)


pitoniak32:
I have started looking into replicating `opentelemetry-stdout` patterns. I have gotten to a place that with very basic and defaulted behavior its similar. I just wanted to confirm that it seems reasonable before I start getting into the details of allowing more customization to the exporter builders.

examples/tracing-jaeger/src/main.rs
```diff
@@ -4,27 +4,26 @@ use opentelemetry::{
...
+    Ok(TracerProvider::builder()
+        .with_batch_exporter(exporter, runtime::Tokio)
+        .with_config(
             sdktrace::Config::default().with_resource(Resource::new(vec![KeyValue::new(
                 SERVICE_NAME,
                 "tracing-jaeger",
             )])),
         )
-        .install_batch(runtime::Tokio)
+        .build())
```
pitoniak32, line 26: this would be an example of the refactored interface (minus the fact that currently its not capable of configuring the tonic endpoint or other things)

pitoniak32:
my current thoughts on configuring the `tonic / http` exporter options would be something like this
```rust
    .with_tonic(
        TonicExporter::builder().with_endpoint("0.0.0.0:4317").build()?
    )
    .build()?;
```

but it feels kind of clunky to have 3 different builders, although it is more modular...

pitoniak32:
This might be another issue / PR, but I'm noticing the `SdkMeterProvider` and `TracerProvider` have different interfaces for configuring options like `Resource` for example. I have a few ideas for possibly unifying this if its open for discussion in another location?

cijothomas:
> This might be another issue / PR, but I'm noticing the `SdkMeterProvider` and `TracerProvider` have different interfaces for configuring options like `Resource` for example. I have a few ideas for possibly unifying this if its open for discussion in another location?

Yes, ensuring consistency across signals wherever feasible is definitely something we need to tackle soon!

examples/self-diagnostics/src/main.rs
```diff
@@ -51,15 +52,16 @@ fn custom_error_handler(err: OtelError) {
...
-        )
-        .install_batch(opentelemetry_sdk::runtime::Tokio)
+    let exporter = LogExporter::builder()
+        .with_http()
+        .with_endpoint("http://localhost:4318/v1/logs")
+        .build()
         .unwrap();
+
+    let provider = opentelemetry_sdk::logs::LoggerProvider::builder()
+        .with_batch_exporter(exporter, opentelemetry_sdk::runtime::Tokio)
```
cijothomas, line 62: I like this direction.

opentelemetry-sdk/src/metrics/exporter.rs
```diff
@@ -11,7 +11,7 @@ use super::data::Temporality;
 ///
 /// This is the final component in the metric push pipeline.
 #[async_trait]
-pub trait PushMetricsExporter: Send + Sync + 'static {
+pub trait PushMetricExporter: Send + Sync + 'static {
```
cijothomas, line 14: lets do this in its own PR, to keep the scope for each PR very focused.

opentelemetry-stdout/examples/basic.rs
```diff
@@ -28,7 +28,7 @@ static RESOURCE: Lazy<Resource> = Lazy::new(|| {
 fn init_trace() {
     let exporter = opentelemetry_stdout::SpanExporter::default();
     let provider = TracerProvider::builder()
-        .with_simple_exporter(exporter)
+        .with_batch_exporter(exporter, runtime::Tokio)
```
cijothomas, line 31: lets keep it simple itself. stdout is only for learning purposes. and simple exporter is also only meant for that.

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) mod tonic;
 
 /// Configuration for the OTLP exporter.
 #[derive(Debug)]
-pub struct ExportConfig {
+pub struct ExporterConfig {
```
cijothomas, line 69: also suggest to keep the renaming to a separate, small pr, so it is easier to review.

cijothomas:
> I have started looking into replicating `opentelemetry-stdout` patterns. I have gotten to a place that with very basic and defaulted behavior its similar. I just wanted to confirm that it seems reasonable before I start getting into the details of allowing more customization to the exporter builders.
> 
> If you have a chance to review this @cijothomas I would really appreciate it!

Took a brief look and I like this direction. Left some comments, mostly about limiting the scope of PR. We really prefer short PRs each solving one particular aspect.

pitoniak32:
> Took a brief look and I like this direction. Left some comments, mostly about limiting the scope of PR. We really prefer short PRs each solving one particular aspect.

I'll go through and remove some of the changes I made that weren't specific to this issue 👍

pitoniak32:
As an output of this PR I would like to open a discussion / issue to talk about the following refactors:
- Renaming the types named `Metics*` to `Metric*` to follow the spec.
- Renaming `opentelemetry::metrics::Result` to `opentelemetry::metrics:MetricResult` (or doing the opposite and using the mod path for `LogResult`, and `TraceResult`).
- Modifying the `TracerProvider` to follow a similar configuration approach that is used for the other "Providers" instead of or in addition to the `with_config(...)` option for consistency across signals.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -28,47 +36,46 @@ static RESOURCE: Lazy<Resource> = Lazy::new(|| {
     )])
 });
 
-fn http_exporter() -> HttpExporterBuilder {
-    let exporter = opentelemetry_otlp::new_exporter().http();
+fn init_logs() -> Result<sdklogs::LoggerProvider, opentelemetry::logs::LogError> {
+    let exporter_builder = LogExporter::builder()
+        .with_http()
+        .with_endpoint("http://localhost:4318/v1/logs")
```
cijothomas, line 42: is it possible for endpoint, protocol to take defaults, so users does not have to specify this (unless they want nondefaults)

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -24,43 +26,37 @@ static RESOURCE: Lazy<Resource> = Lazy::new(|| {
...
-        .build()
+        .build())
 }
 
 fn init_logs() -> Result<opentelemetry_sdk::logs::LoggerProvider, LogError> {
-    opentelemetry_otlp::new_pipeline()
-        .logging()
+    let exporter = LogExporter::builder()
+        .with_tonic()
+        .with_endpoint("http://localhost:4317")
```
cijothomas, line 53: one inconsistency I notice between grpc and http is the end point - one has to append /v1/logs in http, but grpc does not require that. This is something we can address in a follow up.
I am thinking something like:
with_tonic() or with_http() is enough for builder to know which endpoint, protocol to use by default.
builder.with_tonic().build() -- this should just work with defaults
builder.with_http().build() -- this should also just work with defaults.

if user wish to change defaults, then they can use methods like with_endpoint, with_protocol etc.

Lets come back to this after merging this.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -614,11 +640,46 @@ mod tests {
...
+            HashMap::new(),
+            |mut acc, (k, v)| {
+                acc.insert(k.clone(), v.clone());
+                acc
+            },
+        );
+        let builder = HttpExporterBuilder {
+            http_config: HttpConfig {
+                client: None,
+                headers: Some(HashMap::from([("k1".to_string(), "v1".to_string())])),
```
cijothomas, line 659: nit: reuse initial_headers variable from above

opentelemetry-otlp/src/logs.rs
```diff
@@ -107,106 +132,3 @@ impl opentelemetry_sdk::export::logs::LogExporter for LogExporter {
         self.client.set_resource(resource);
     }
 }
-
-/// Recommended configuration for an OTLP exporter pipeline.
```
cijothomas, line 111: We'll need to make a detailed changelog entry as this is breaking change. I think its best to include before/after code snippets in changelog itself for some common scenario, and a link to the PR.
This PR's description can contain the full details and a mini migration guide.

cijothomas (approved):
LGTM. We'll need some follow ups, as noted in comments.
Most important thing is a detailed migration guide, which can be written in this PR's description for users to follow.
Also the changelog can mention the before/after for a common scenario for quick reference.

Given the significance of this change, I'll merge once we have 3 approvals:
@TommyCpp @lalitb @utpilla @hdost Please review.

cijothomas:
> As an output of this PR I would like to open a discussion / issue to talk about the following refactors:
> 
> * Renaming the types named `Metics*` to `Metric*` to follow the spec.
> * Renaming `opentelemetry::metrics::Result` to `opentelemetry::metrics:MetricResult` (or doing the opposite and using the mod path for `LogResult`, and `TraceResult`).
> * Modifying the `TracerProvider` to follow a similar configuration approach that is used for the other "Providers" instead of or in addition to the `with_config(...)` option for consistency across signals.

:100: Very good points. Lets do it as follow ups.

pitoniak32:
> LGTM. We'll need some follow ups, as noted in comments.
> 
> Most important thing is a detailed migration guide, which can be written in this PR's description for users to follow.
> 
> Also the changelog can mention the before/after for a common scenario for quick reference.

Awesome! I'll address comments and add the migration details / CHANGELOG entry sometime today 🙏

opentelemetry-otlp/examples/basic-otlp/src/main.rs
pitoniak32: > sounds good, after my most recent update, the `with_{tonic,http}()` is enough to determine the `http://localhost:431{7,8}` and the tests for this are updated also.
> 
> 
> 
> I will however need to look at the spec to see how the `/v1/*` is handled for the two, and I would be happy to create a new pr to get that sorted out.

EDIT: After looking back through the code, this should also already be the case, I made some changes to make it a little more clear

opentelemetry-otlp/src/lib.rs
```diff
@@ -36,12 +36,11 @@
 //!
 //! fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
 //!     // First, create a OTLP exporter builder. Configure it as you need.
-//!     let otlp_exporter = opentelemetry_otlp::new_exporter().tonic();
+//!     let otlp_exporter = opentelemetry_otlp::SpanExporter::builder().with_tonic().build()?;
```
lalitb, line 39: nit - remove the `new_pipeline` reference from doc comment above (line 27)

opentelemetry-otlp/src/lib.rs
```diff
@@ -70,10 +69,15 @@
...
-//! let tracer = opentelemetry_otlp::new_pipeline()
-//!     .tracing()
-//!     .with_exporter(opentelemetry_otlp::new_exporter().tonic())
-//!     .install_batch(opentelemetry_sdk::runtime::AsyncStd)?;
+//! let tracer = opentelemetry_sdk::trace::TracerProvider::builder()
+//!        .with_batch_exporter(
+//!            opentelemetry_otlp::SpanExporter::builder()
+//!                .with_tonic()
+//!                .build()?,
+//!            opentelemetry_sdk::runtime::AsyncStd,
```
lalitb, line 77: nit - not related to this PR, but I believe tonic has hard dependency on the tokio runtime. We should be changing this example to use `opentelemetry_sdk::runtime::Tokio`

opentelemetry-otlp/src/lib.rs
pitoniak32: updated the doc comment to be more accurate for these changes.

opentelemetry-otlp/src/lib.rs
pitoniak32: updated

lalitb (approved):
LGTM. Thanks for the cleanup, nicely done :)

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -405,11 +378,64 @@ fn add_header_from_string(input: &str, headers: &mut HashMap<HeaderName, HeaderV
     }));
 }
 
+/// Expose interface for modifying builder config.
+pub trait HasHttpConfig {
```
utpilla, line 382: Why are these new traits `HasHttpConfig` and `WithHttpConfig` needed? Could we have these methods as part of `HttpExporterBuilder like earlier`?

pitoniak32: Those are so the config options are available on the Span / Metric / Log Exporter builders when with_http is used

pitoniak32: If they are on the HttpExporterBuilder we would have to take that as a parameter for the with_http

utpilla: I don't follow. How does `with_http` influence `with_http_client`?
- `with_http` would bring in the default client (reqwest)
- `with_http_client` allows the user to bring their custom client

Both of these methods can reside in `impl HttpExporterBuilder`. What value are these new public traits adding?

pitoniak32: for example, using the new traits the user gains access to the http specific config options without needing to configure a separate `HttpExporterBuilder` / `TonicExporterBuilder`. So in the following case, after `.with_http()` the two http specific methods become available, but the tonic specific ones aren't.
```rust
    .with_http()
    .with_http_client(...) // non default client like hyper
    .with_headers(...) // non default headers to send
    .build()?;

    .with_tonic()
    .with_metadata(...) // non default metadata to send
    .build()?;
```

Without the traits we would have to do something like:
```rust
    .with_http(HttpExporter::builder().with_headers(...).build())
    .build()?;
```

pitoniak32: >How does `with_http` influence `with_http_client`?

`with_http()` alone will use the default values for http_client / headers, but after you say `with_http()` you can configure that http exporter with non default values like `.with_http_client(hyper)` to set a non default client before you call `build()`

cijothomas: To unblock this PR (its pretty big and will cause merge conflicts if delayed more), lets track this separately as a follow up ?

As per the SIG/Community call yesterday, we are delaying 0.27 release by a week, so we can do follow ups as necessary after merge. Also, I think changes to this area might be less intrusive to end users (not 100% sure).

@lalitb @utpilla Please share your thoughts on merging this as-is and follow up on these extra traits afterwards?

cijothomas:
@pitoniak32 Very good migration guide in the PR description! We can keep editing it, as we get feedback from users.

opentelemetry-otlp/src/exporter/http/mod.rs
lalitb: Yes, this is good discussion. Traits are not affecting the API surface, so can be revisited aftewards.

utpilla: This PR brings a lot of new public API additions:
- `pub struct HttpConfig`
- `pub trait HasHttpConfig`
- `pub trait WithHttpConfig`
- `pub trait HasTonicConfig`
- `pub trait WithTonicConfig`
- `pub struct NoExporterBuilderSet`
- `pub struct TonicExporterBuilderSet(TonicExporterBuilder)`
- pub struct HttpExporterBuilderSet(HttpExporterBuilder)`

I would like us to achieve this with a reduce public API surface. I haven't looked at this deeply to assess whether it's feasible or not.

I also agree that even though these are public, we don't expect users to be using them. However, a reduced public API surface generally eases the maintenance burden in the long run. I would prefer that we attempt to simplify this if we are merging it now. My question would be if that simplification process demands any breaking change in the usage (especially setup of the exporter), would that be acceptable?

cijothomas: > My question would be if that simplification process demands any breaking change in the usage (especially setup of the exporter), would that be acceptable?

In my opinion -  Yes, given SDK/OTLP Exporter crates are still going to be alpha/beta for at-least few more weeks.

@pitoniak32 We can merge this as-is now. Would you have some bandwidth to see if this can be further simplified as a follow up?

pitoniak32: Yeah I can look more into this, I won't be able to get started until tomorrow, my main desktop motherboard needs to be RMAd because of memory errors :( so I need to setup my dev env on another computer but I should still be able to look into it

pitoniak32: > @pitoniak32 We can merge this as-is now. Would you have some bandwidth to see if this can be further simplified as a follow up?

Should we have a discussion around what would be ideally included in the new solution? I joined the otel rust slack if you would like to chat there, or would it be better to open a discussion / issue? I can also just jump into trying alternatives and open a draft PR. But it could be good to align on what would be a nice interface to configure these exporters

utpilla: Thanks @pitoniak32! I'll reach out to you on Slack with my initial thoughts.

cijothomas: @utpilla okay to merge now?
(Trying to avoid delaying big PRs to avoid conflicts)

utpilla (approved):
Looks good overall from the user-facing changes.

opentelemetry-otlp/src/exporter/http/mod.rs
utpilla: Yes.

keltia:
Guys, I know this is still 0.x and stuff, but breaking the API on almost every release is, from a user point of view, quite painful.

frederikhors:
Hi dear people, there is no guid for `with_batch_config()`, how to use it now with 0.27?

frederikhors:
@pitoniak32, can you help me with `with_batch_config()`, how to use it now with 0.27?

pitoniak32:
> @pitoniak32, can you help me with `with_batch_config()`, how to use it now with 0.27?

Sure! Do you have a specific issue? Did you checkout the migration guide in the overview comment already?

frederikhors:
Yeah, but I cannot find `with_batch_config()` in there.

lalitb:
@frederikhors Can you try something like (not tested)
```rust
fn init_tracer_provider() -> Result<sdktrace::TracerProvider, TraceError> {
    // Create a span exporter
        .with_http()
        .with_protocol(Protocol::HttpBinary) 
        .with_endpoint("http://localhost:4318/v1/traces")
        .build()?;

    let batch_config = BatchConfigBuilder::default()
        .with_max_queue_size(2048)
        .build();

    let batch_processor = sdktrace::BatchSpanProcessor::builder(exporter, runtime::Tokio)
        .with_batch_config(batch_config)
        .build();

    Ok(TracerProvider::builder()
        .with_span_processor(batch_processor)
        .with_config(Config::default().with_resource(RESOURCE.clone()))
        .build())
}
```

frederikhors:
It works.

I'm having only this error now:

```
error[E0599]: no method named `tracer` found for struct `opentelemetry_sdk::trace::TracerProvider` in the current scope
   |
87 |             .with_tracer(otlp_tracer_provider.tracer("otlp-tracer"))
   |                                               ^^^^^^
   |
  ::: C:\Users\Fred\.cargo\registry\src\index.crates.io-6f17d22bba15001f\opentelemetry-0.27.0\src\trace\tracer_provider.rs:31:8
   |
31 |     fn tracer(&self, name: impl Into<Cow<'static, str>>) -> Self::Tracer {
   |        ------ the method is available for `opentelemetry_sdk::trace::TracerProvider` here
   |
   = help: items from traits can only be used if the trait is in scope
help: trait `TracerProvider` which provides `tracer` is implemented but not in scope; perhaps you want to import it
   |
1  + use opentelemetry::trace::TracerProvider;
   |
help: there is a method `boxed_tracer` with a similar name
   |
87 |             .with_tracer(otlp_tracer_provider.boxed_tracer("otlp-tracer"))
   |                                               ~~~~~~~~~~~~
```

for the code:

```rust
let layer = tracing_opentelemetry::layer()
    .with_tracer(otlp_tracer_provider.tracer("otlp-tracer"))
    .with_filter(EnvFilter::from("INFO"))
    .boxed();
```

lalitb:
Does importing `use opentelemetry::trace::TracerProvider as _;` helps?

frederikhors:
> Try importing `use opentelemetry::trace::TracerProvider as _;` if that works?

It works. But ??? Is this OK?

lalitb:
> It works. But ??? Is this OK?

Yes, this is typical of Rust — the trait must be in scope for its methods to be accessible.

cijothomas:
Thanks all. If there are any issues/concerns with this change that need help - please open a new issue.

Issue #2046, opened 2024-08-23
[Log SDK] Cost associated with populating LogRecord::observed_timestamp


lalitb:
In Log SDK, the LogRecord::observed_timestamp is populated as:
 
https://github.com/open-telemetry/opentelemetry-rust/blob/673c328522b99e61ec5c91fd905ac4ffe5a0da08/opentelemetry-sdk/src/logs/log_emitter.rs#L270

However, after all the optimizations, the cost associated with this call is visible considerably ~20%:

![image](https://github.com/user-attachments/assets/2d837b60-a90b-4066-a2c2-5e1a7a63bfc5)

Not sure about the solution for now, raising this issue to track the cost.

cijothomas:
You are right! Obtaining the timestamp is indeed quite expensive. Wondering if there are solutions like checking system Ticks since start and using that to populate timestamp.

lalitb:
Some old thread in rust community  - https://users.rust-lang.org/t/fast-get-current-time/10381/2. It talks about using [coarsetime](https://github.com/jedisct1/rust-coarsetime) or directly using [libc](https://crates.io/crates/libc).

> like checking system Ticks since start and using that to populate timestamp.

Yes, need to see if there is platform/arch-independent way to achieve it.

lalitb:
Some comparison between `SystemTime::now()` and directly using `libc::clock_gettime()`:

```rust
fn get_libc_clock(tv: &mut timespec) { 
    unsafe {
        libc::clock_gettime(libc::CLOCK_MONOTONIC_COARSE, tv);
    }
}
fn benchmark_time_methods(c: &mut Criterion) {
    c.bench_function("SystemTime::now", |b| {
        b.iter(|| black_box(std::time::SystemTime::now()));
    });
    let mut tv = libc::timespec {
        tv_sec: 0,
        tv_nsec: 0,
    };
    c.bench_function("libc::clock_gettime", |b| {
        b.iter(|| black_box(get_libc_clock(&mut tv)));
    });
}
```

o/p is:
```
SystemTime::now         time:   [24.313 ns 24.881 ns 25.466 ns]
libc::clock_gettime     time:   [5.5160 ns 5.6096 ns 5.7068 ns]
```
So, `libc` is 4 times faster than `SystemTime`, however it needs `unsafe` keyword. Another option could be directly using `coarsetime` crate, which internally uses `libc` and so similar perf gains.

[...]

PR #2231, opened 2024-10-23
adds timestamps

montekki:
Fixes #

## Changes

Unset timestamps in logs break services such as Quickwit, although if my understanding is correct these timestamps _must_ be set as a part of OTEL spec.

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -173,6 +173,8 @@ where
         log_record.set_event_name(meta.name());
         log_record.set_severity_number(severity_of_level(meta.level()));
         log_record.set_severity_text(meta.level().as_str());
+        log_record.set_timestamp(std::time::SystemTime::now());
+        log_record.set_observed_timestamp(std::time::SystemTime::now());
         let mut visitor = EventVisitor::new(&mut log_record);
         #[cfg(feature = "experimental_metadata_attributes")]
         visitor.visit_experimental_metadata(meta);


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -173,6 +173,8 @@ where
         log_record.set_event_name(meta.name());
         log_record.set_severity_number(severity_of_level(meta.level()));
         log_record.set_severity_text(meta.level().as_str());
+        log_record.set_timestamp(std::time::SystemTime::now());
```
lalitb, line 176: As per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp) - `This field is optional, it may be missing if the source timestamp is unknown.`

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -173,6 +173,8 @@ where
...
+        log_record.set_observed_timestamp(std::time::SystemTime::now());
```
lalitb, line 177: This is already done in [SDK](https://github.com/open-telemetry/opentelemetry-rust/blob/8bd529a6d629aff7482b875cfc39275a8a71eaeb/opentelemetry-sdk/src/logs/log_emitter.rs#L303), why do we need it here.

lalitb:
timestamp setting is costly (ref - #2046), let's not do it unless really required by specs.

opentelemetry-appender-tracing/src/layer.rs
montekki: Is this path in SDK hit when this `Bridge` is added as a layer into `tracing_subscriber` ? On the QuickWit side I see that this timestamp was not populated which needed a workaround: quickwit-oss/quickwit#5366

lalitb: > Is this path in SDK hit when this Bridge is added as a layer into tracing_subscriber

yes it does.

lalitb (changes requested):
As discussed in PR.

cijothomas:
@montekki I am closing this PR as it looks like there is no missing functionality. If you believe otherwise, we can re-open to discuss more.

PR #2233, opened 2024-10-24
Exclude benchmark setup duration using iter_batched

### LINKED ISSUE #1805: Avoid random generation cost from all perf tests
https://github.com/open-telemetry/opentelemetry-rust/pull/1800 has made the overhead very small, but we should be able to avoid it completely. See https://github.com/open-telemetry/opentelemetry-rust/pull/1800#discussion_r1609408461 for infomation.

zarvd:
Fixes #1805

## Changes

Use [iter_batched](https://docs.rs/criterion/latest/criterion/struct.Bencher.html#method.iter_batched) for benchmarking `Counter_Add_Sorted` and `Counter_Add_Unsorted` to exclude the setup duration from the benchmark results.

With `LargeInput`
```
Counter_Created
Counter_Add_Sorted      time:   [170.49 ns 170.63 ns 170.89 ns]
                        change: [-11.149% -10.899% -10.583%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  4 (4.00%) high severe

Counter_Created
Counter_Add_Unsorted    time:   [172.93 ns 172.97 ns 173.01 ns]
                        change: [-11.878% -11.673% -11.426%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  5 (5.00%) high severe
```

With `SmallInput`
```
Counter_Created
Counter_Add_Sorted      time:   [167.97 ns 168.04 ns 168.11 ns]
                        change: [-2.0473% -1.7160% -1.4582%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe

Counter_Created
Counter_Add_Unsorted    time:   [171.76 ns 171.94 ns 172.18 ns]
                        change: [-1.0523% -0.8236% -0.6261%] (p = 0.00 < 0.05)
  1 (1.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe

```

OS: Ubuntu 24.04 (6.8.0-1014-azure)
Hardware: AMD EPYC 7763 64-Core (16) @ 2.44 GHz
RAM: 64.0 GB

### DIFF
--- a/opentelemetry-sdk/benches/metrics_counter.rs
+++ b/opentelemetry-sdk/benches/metrics_counter.rs
@@ -12,7 +12,7 @@
     | ThreadLocal_Random_Generator_5 |  37 ns      |
 */
 
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use opentelemetry::{
     metrics::{Counter, MeterProvider as _},
     KeyValue,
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/benches/metrics_counter.rs
```diff
@@ -57,62 +57,72 @@ fn criterion_benchmark(c: &mut Criterion) {
...
-                    KeyValue::new("attribute3", ATTRIBUTE_VALUES[index_third_attribute]),
-                    KeyValue::new("attribute1", ATTRIBUTE_VALUES[index_first_attribute]),
-                    KeyValue::new("attribute4", ATTRIBUTE_VALUES[index_fourth_attribute]),
-                ],
-            );
-        });
+        b.iter_batched(
+            || {
+                // 4*4*10*10 = 1600 time series.
+                CURRENT_RNG.with(|rng| {
```
cijothomas, line 99: nit: most likely obvious, but maybe leave a comment here saying iter_batched() is used to exclude setup time (rand generations) from the meaasurements.

cijothomas (approved):
Thank you!
Would love to see more contributions!

opentelemetry-sdk/benches/metrics_counter.rs
```diff
@@ -57,62 +57,72 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+                    1,
+                    &[
+                        KeyValue::new("attribute1", ATTRIBUTE_VALUES[index_first_attribute]),
+                        KeyValue::new("attribute2", ATTRIBUTE_VALUES[index_second_attribute]),
+                        KeyValue::new("attribute3", ATTRIBUTE_VALUES[index_third_attribute]),
+                        KeyValue::new("attribute4", ATTRIBUTE_VALUES[index_fourth_attribute]),
+                    ],
+                );
+            },
+            BatchSize::LargeInput,
```
utpilla, line 88: Does it need to be `BatchSize::LargeInput`? Could we use `BatchSize::SmallInput` instead as it has a lower measurement overhead?

lalitb: The only reason of using large batch I can think here is if the setup overhead is significant, and the large batch size relatively reduces the impact. But otherwise, small batch size is good enough.

utpilla: From the [docs](https://docs.rs/criterion/latest/criterion/enum.BatchSize.html):

> Generally speaking, almost all benchmarks should use `SmallInput`. If the input or the result of the benchmark routine is large enough that `SmallInput` causes out-of-memory errors, `LargeInput` can be used to reduce memory usage at the cost of increasing the measurement overhead.

I would prefer using `SmallInput` whenever possible.

zarvd: Good catch! Fixed. PTAL thanks.

utpilla (approved):

opentelemetry-sdk/benches/metrics_counter.rs
utpilla: @zarvd It would be great if you could also confirm that running these benchmarks with `SmallInput` did not run into any issues.

zarvd: It’s working fine on my Mac and the Ubuntu VM. I’ve also updated the PR description with the results.

opentelemetry-sdk/benches/metrics_counter.rs
zarvd: IMHO, it’s obvious, so I’ll leave it as is. 😀

PR #2234, opened 2024-10-24
Global error handler cleanup - Counter and Observable counter creation

lalitb:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter.rs
+++ b/opentelemetry-sdk/src/metrics/meter.rs
@@ -124,7 +124,7 @@ impl SdkMeter {
                 name: "InstrumentCreationFailed", 
                 meter_name = self.scope.name.as_ref(),
                 instrument_name = builder.name.as_ref(),
-                message = "Callbacks for this observable counter will be invoked.",
+                message = "Callbacks for this observable counter will not be invoked.",
                 reason = format!("{}", err));
             return ObservableCounter::new();
         }
@@ -142,7 +142,7 @@ impl SdkMeter {
                         name: "InstrumentCreationFailed",
                         meter_name = self.scope.name.as_ref(),
                         instrument_name = builder.name.as_ref(),
-                        message = "Callbacks for this observable counter will be invoked. Check View Configuration."
+                        message = "Callbacks for this observable counter will not be invoked. Check View Configuration."
                     );
                     return ObservableCounter::new();
                 }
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -75,14 +76,20 @@ impl SdkMeter {
...
         let validation_result = validate_instrument_config(builder.name.as_ref(), &builder.unit);
         if let Err(err) = validation_result {
-            global::handle_error(err);
+            otel_error!(
+                name: "InstrumentCreationFailed",
+                meter_name = self.scope.name.as_ref(),
+                instrument_name = builder.name.as_ref(),
+                message = "Measurements from this counter will be ignored.",
+                reason = format!("{}", err)
+            );
```
lalitb, line 85: Cijo's comment:
otel_error!(
name: "InstrumentCreationFailed",
meter_name = self.scope.name.as_ref(),
instrument_name = builder.name.as_ref(),
message = "Measurements from this instrument will be ignored."
reason = fmt(err))

^ I prefer this version. Please see if this is better for end users.

lalitb: Utkarsh comment: 
change
`"Measurements from this instrument will be ignored."` to
`""Measurements from this counter will be ignored.",`

cijothomas, line 85: 2024-10-24T05:32:23.524520Z ERROR main opentelemetry_sdk meter_name="mylibraryname" instrument_name="-my_counter" Measurements from this counter will be ignored. reason="Invalid instrument configuration: instrument name must start with an alphabetic character"

^ This is what we get when I used fmt module. The reason is a slightly extra verbose:
Current: _Invalid instrument configuration: instrument name must start with an alphabetic character_
Proposal: _Instrument name must start with an alphabetic character_

This is a minor thing we can change in the future.

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -107,19 +120,30 @@ impl SdkMeter {
     {
         let validation_result = validate_instrument_config(builder.name.as_ref(), &builder.unit);
         if let Err(err) = validation_result {
-            global::handle_error(err);
+            otel_error!(
+                name: "InstrumentCreationFailed", 
+                meter_name = self.scope.name.as_ref(),
+                instrument_name = builder.name.as_ref(),
+                message = "Measurements from this observable counter will be ignored.",
```
cijothomas, line 127: nit: if we are doing different message for each kind of instrument, then for observable, we need to call out that the registered callbacks will not be invoked.

cijothomas (approved):
Left a nit suggestion, but LGTM for merging.

opentelemetry-sdk/src/metrics/meter.rs
lalitb: This is good point. Have updated it.

utpilla (approved):

TommyCpp (approved):

PR #2241, opened 2024-10-24
refactor: update metrics Result to be MetricResult

pitoniak32:
related to comment: https://github.com/open-telemetry/opentelemetry-rust/pull/2221#issuecomment-2424967497

## Changes

Make the `MetricResult` / `MetricError` consistent with `LogResult` / `LogError`, and `TraceResult` / `TraceError`

Another option could be to have all signals type be `Result` / `Error`, and determine which one is used from the module path `metrics::Result<()>` / `metrics::Error`?

### DIFF
--- a/opentelemetry-sdk/src/metrics/exporter.rs
+++ b/opentelemetry-sdk/src/metrics/exporter.rs
@@ -1,7 +1,7 @@
 //! Interfaces for exporting metrics
 use async_trait::async_trait;
 
-use opentelemetry::metrics::Result;
+use opentelemetry::metrics::MetricResult;
 
 use crate::metrics::data::ResourceMetrics;
 
@@ -18,16 +18,16 @@ pub trait PushMetricsExporter: Send + Sync + 'static {
     /// implement any retry logic. All errors returned by this function are
     /// considered unrecoverable and will be reported to a configured error
     /// Handler.
-    async fn export(&self, metrics: &mut ResourceMetrics) -> Result<()>;
+    async fn export(&self, metrics: &mut ResourceMetrics) -> MetricResult<()>;
 
     /// Flushes any metric data held by an exporter.
-    async fn force_flush(&self) -> Result<()>;
+    async fn force_flush(&self) -> MetricResult<()>;
 
     /// Releases any held computational resources.
     ///
     /// After Shutdown is called, calls to Export will perform no operation and
     /// instead will return an error indicating the shutdown state.
-    fn shutdown(&self) -> Result<()>;
+    fn shutdown(&self) -> MetricResult<()>;
 
     /// Access the [Temporality] of the MetricsExporter.
     fn temporality(&self) -> Temporality;
--- a/opentelemetry/src/metrics/mod.rs
+++ b/opentelemetry/src/metrics/mod.rs
@@ -22,12 +22,12 @@ pub use instruments::{
 pub use meter::{Meter, MeterProvider};
 
 /// A specialized `Result` type for metric operations.
-pub type MetricResult<T> = result::Result<T, MetricsError>;
+pub type MetricResult<T> = result::Result<T, MetricError>;
 
 /// Errors returned by the metrics API.
 #[derive(Error, Debug)]
 #[non_exhaustive]
-pub enum MetricsError {
+pub enum MetricError {
     /// Other errors not covered by specific cases.
     #[error("Metrics error: {0}")]
     Other(String),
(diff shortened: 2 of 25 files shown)


pitoniak32:
@cijothomas as we discussed in https://github.com/open-telemetry/opentelemetry-rust/pull/2221 I opened this to address one of the bullets brought up

opentelemetry/src/metrics/mod.rs
```diff
@@ -22,12 +22,12 @@ pub use instruments::{
 pub use meter::{Meter, MeterProvider};
 
 /// A specialized `Result` type for metric operations.
-pub type Result<T> = result::Result<T, MetricsError>;
+pub type MetricResult<T> = result::Result<T, MetricError>;
```
cijothomas, line 25: I am unsure if this Result/MetricError need to be kept public. For sure, we are removing it from the `opentelemetry` crate (see https://github.com/open-telemetry/opentelemetry-rust/issues/1042).

I believe these Results (public) are only useful when returning a Result from end-user APIs. We have very few such APIs, and most of the time we would be relying on internal logs.
eg of user facing api which benefit from a public Result is ForceFlush(), Shutdown().

It'd be better if we take a step back and see if there would be any actual need of MetricResult in the future, and if not, lets clean it up. And also, keep it in sdk crate only as opentelemetry's API almost never need it.

@pitoniak32 What do you think?

pitoniak32: I think that makes sense. The only argument I would have in favor of keeping it would be, in my personal project code where I'm doing the setup for these apis I would use these result/error types as the return type from the setup function (if I wasn't already using anyhow or translating to custom errors) just to make things cleaner and avoid doing a Box dyn error. But this is a very specific use case, just wanted to play devils advocate on this one.

pitoniak32: I would say definitely that the Result types can be removed from the public API but I could see the Errors being useful to keep for users to translate into their own errors, or return from setup functions

pitoniak32: Would you also do the rename of metrics items as part of this too?

opentelemetry-sdk/src/metrics/exporter.rs
```diff
@@ -18,16 +18,16 @@ pub trait PushMetricsExporter: Send + Sync + 'static {
...
     /// Flushes any metric data held by an exporter.
-    async fn force_flush(&self) -> Result<()>;
+    async fn force_flush(&self) -> MetricResult<()>;
 
     /// Releases any held computational resources.
     ///
     /// After Shutdown is called, calls to Export will perform no operation and
     /// instead will return an error indicating the shutdown state.
-    fn shutdown(&self) -> Result<()>;
+    fn shutdown(&self) -> MetricResult<()>;
```
lalitb, line 30: Not related to this PR, but the export interface of metrics, logs and traces are not consistent with each other:
 - `SpanExporter::shutdown()` and `LogExporter::shutdown()` does't return the status of the shutdown. `MetricsExporter::shutdown` return status.
 - `LogsExporter::force_flush()` is missing. `SpanExporter::force_flush()` is not async method, while `MetricsExporter::force_flush()` is an async method.

We need to discuss and make them consistent separate of this PR.

opentelemetry/src/metrics/mod.rs
```diff
@@ -22,12 +22,12 @@ pub use instruments::{
...
 
 /// Errors returned by the metrics API.
 #[derive(Error, Debug)]
 #[non_exhaustive]
-pub enum MetricsError {
+pub enum MetricError {
```
lalitb, line 30: With these changes, `MetricReader`, `MetricError`, and `MetricResult` are singular, while PushMetricsExporter is plural. Should we PushMetricsExporter as separate PR ?

opentelemetry/src/metrics/mod.rs
lalitb: No, the rename for `MetricError` and `MetricResult` can go through as part of this PR.

opentelemetry/src/metrics/mod.rs
pitoniak32: Yeah the end goal will be have that one as singular too, I was going to open another PR for those changes, keeping this one scoped to the result and error types

lalitb (approved):
LGTM.

pitoniak32:
> edit: We can have a single, consolidated changelog after #1042.
so I should omit adding a changelog update on this pr in favor of waiting for 1042?

cijothomas (approved):
Thanks. As discussed in the PR already, we need to re-evaluate custom Error types and only make the relevant ones as public.

Issue #2145, opened 2024-09-25
Unify metrics DataPoints collection


fraillt:
## Problem

At the moment there's zero code reuse in DataPoints collection phase. Due to too much copy/paste there are few issues:
* there are inefficiencies by doing unnecessary `.clone()`
* I noticed at least one bug in `PrecomputedSum::delta` where `no_attribute_tracker` uses `get_value` instead of `get_and_reset_value`.
* hard to understand the core logic of aggregator, versus implementation details of how DataPoints are collected.
* makes very hard to experiment with performance improvements in `measuring` and/or `collecting` phase. E.g. if I change something within ValueMap, I need to go and update at minimum 5 aggregators (if we add `ExpoHistogram`) * 2 (delta/comulative) = 10 places.
* makes very hard to review, because simple change will affect at least 10 places.

## Solution

Hide implementation details of `ValueMap`, and provide simple `collect_cumulative` and `collect_delta` functions instead (similar idea as in #2114).
In the future we'll be able to experiment and improve `ValueMap` implementation and get benefits across all aggregators.

cijothomas:
Agreed. In other implementations like OTel .NET, the entire logic of keeping track of metric points, look up, etc are done in a common place (i.e not repeated for each instument type).

Related:https://github.com/open-telemetry/opentelemetry-rust/issues/1566

PR #2117, opened 2024-09-14
ValueMap interface change

fraillt:
## Changes

Absolute minimum changes to `ValueMap` in order to provide interface that can be applied for all kinds of metrics.
I've tried to make this revision as small as possible, (leaving out optimization opportunities and bugs that noticed along the way), so it would be easier to review.
The benefit of this new interface is that it can be applied efficiently to all histograms. I have a proof in #2114 (which uses same interface) that it can be elegantly applied to `ExpoHistogram` as well.

Few more points/notes that might be helpful for reviewer:
* `Histogram` might be a bit harder to review, because it required most changes, but essentially there's configuration extracted into new `BucketsConfig` type, and `Aggregator` is implemented for `Mutex<Buckets<T>>`, the rest is basically compiler-driven-development.
* `LastValue`, `Sum`, `PrecomputedSum` should be trivial to review, I just had to implement `Aggregator` interface for `Increment` and `Assign` functionality, which is pretty trivial.
* Changes to `AtomicTracker` and `AtomicallyUpdate` had to be made as well, either to get rid of "unused code" warning, or required for `Increment` and `Assign`. Generally these interfaces got simplified as well, as they no longer contain histogram-only related stuff.

I really want to unify common/important code so all metrics could benefit from it.
And I really hope this revision is small enough to review efficiently :) Happy reviewing :)

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [x] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #2247, opened 2024-10-25
Metrics collect stress test

fraillt:
## Changes

Added two tests (delta and cumulative temporality) to measuring how much metrics collection phase impact measurements throughput.
Implementation consists of running collection phase in the loop, while simultaneously testing measurements throughput.
The main concern isn't the speed of collection phase itself, but ensuring that Opentelemetry Metrics doesn't significantly contribute to p99 latency.
I didn't customize it for different metrics type, idea is to eventually land #2117, so that I could unify collection phase for other metrics #2145.

Additionally, significantly improved code reuse for `stress` crate.

Here are some results on my machine:
```
cargo run --release --package stress --bin metrics_collect -- delta
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/metrics_collect delta`
     17.38 measurements/ms
     52.54 measurements/it
   3023.88 μs/it
   
cargo run --release --package stress --bin metrics_collect -- cumulative
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/metrics_collect cumulative`
      9.14 measurements/ms
     24.35 measurements/it
   2665.30 μs/it
```   

For curiosity I have also implemented #2117 and #2145 locally, and results are as follows:
```
cargo run --release --package stress --bin metrics_collect -- delta
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/metrics_collect delta`
     56.38 measurements/ms
    154.24 measurements/it
   2735.73 μs/it
   
cargo run --release --package stress --bin metrics_collect -- cumulative
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/metrics_collect cumulative`
    840.84 measurements/ms
   2077.64 measurements/it
   2470.90 μs/it   
```

### DIFF
--- a/stress/src/metrics_collect.rs
+++ b/stress/src/metrics_collect.rs
@@ -1,0 +1,239 @@
+use std::{
+    cell::RefCell,
+    ops::DerefMut,
+    sync::{
+        atomic::{AtomicBool, AtomicUsize, Ordering},
+        Arc, Barrier, Weak,
+    },
+    time::{Duration, Instant},
+};
+
+use lazy_static::lazy_static;
+use opentelemetry::{
+    metrics::{Histogram, MeterProvider, MetricResult},
+    KeyValue,
+};
+use opentelemetry_sdk::{
+    metrics::{
+        data::{ResourceMetrics, Temporality},
+        reader::MetricReader,
+        InstrumentKind, ManualReader, Pipeline, SdkMeterProvider,
+    },
+    Resource,
+};
+
+use rand::{
+    rngs::{self, SmallRng},
+    Rng, SeedableRng,
+};
+
+use clap::{Parser, ValueEnum};
+
+#[derive(Debug, Clone, Copy, ValueEnum)]
+enum CliTemporality {
+    Cumulative,
+    Delta,
+}
+
+/// Simple program to greet a person
+#[derive(Parser, Debug)]
+#[command(
+    version,
+    about = "Measure metrics performance while collecting",
+    long_about = "The purpose of this test is to see how collecing interferre with measurements.\n\
+    Most of the test measure how fast is collecting phase, but more important is\n\
+    that it doesn't \"stop-the-world\" while collection phase is running."
+)]
+struct Cli {
+    /// Select collection phase temporality
+    temporality: CliTemporality,
+}
+
+lazy_static! {
+    pub static ref ATTRIBUTE_VALUES: [&'static str; 10] = [
+        "value1", "value2", "value3", "value4", "value5", "value6", "value7", "value8", "value9",
+        "value10"
+    ];
+}
+
+thread_local! {
+
+    /// Store random number generator for each thread
+    pub static CURRENT_RNG: RefCell<rngs::SmallRng> = RefCell::new(rngs::SmallRng::from_entropy());
+}
+
+fn main() {
+    let cli = Cli::parse();
+    let temporality = match cli.temporality {
+        CliTemporality::Cumulative => Temporality::Cumulative,
+        CliTemporality::Delta => Temporality::Delta,
+    };
+    let reader = SharedReader::new(
+        ManualReader::builder()
+            .with_temporality(temporality)
+            .build(),
+    );
+    let provider = SdkMeterProvider::builder()
+        .with_reader(reader.clone())
+        .build();
+    // use histogram, as it is a bit more complicated during
+    let histogram = provider.meter("test").u64_histogram("hello").build();
+
+    calculate_measurements_during_collection(histogram, reader).print_results();
+}
+
+fn calculate_measurements_during_collection(
+    histogram: Histogram<u64>,
+    reader: SharedReader,
+) -> MeasurementResults {
+    // we don't need to use every single CPU, better leave other CPU for operating system work,
+    // so our running threads could be much more stable in performance.
+    // just for the record, this is has HUGE effect on my machine (laptop intel i7-1355u)
+    let num_threads = num_cpus::get() / 2;
+
+    let mut res = MeasurementResults {
+        total_measurements_count: 0,
+        total_time_collecting: 0,
+        num_iterations: 0,
+    };
+    let start = Instant::now();
+    while start.elapsed() < Duration::from_secs(3) {
+        res.num_iterations += 1;
+        let is_collecting = AtomicBool::new(false);
+        let measurements_while_collecting = AtomicUsize::new(0);
+        let time_while_collecting = AtomicUsize::new(0);
+        let barrier = Barrier::new(num_threads + 1);
+        std::thread::scope(|s| {
+            // first create bunch of measurements,
+            // so that collection phase wouldn't be "empty"
+            let mut handles = Vec::new();
+            for _t in 0..num_threads {
+                handles.push(s.spawn(|| {
+                    for _i in 0..1000 {
+                        CURRENT_RNG.with(|rng| {
+                            histogram.record(
+                                1,
+                                &random_attribute_set3(
+                                    rng.borrow_mut().deref_mut(),
+                                    ATTRIBUTE_VALUES.as_ref(),
+                                ),
+                            );
+                        });
+                    }
+                }));
+            }
+            for handle in handles {
+                handle.join().unwrap();
+            }
+
+            // simultaneously start collecting and creating more measurements
+            for _ in 0..num_threads - 1 {
+                s.spawn(|| {
+                    barrier.wait();
+                    let now = Instant::now();
+                    let mut count = 0;
+                    while is_collecting.load(Ordering::Acquire) {
+                        CURRENT_RNG.with(|rng| {
+                            histogram.record(
+                                1,
+                                &random_attribute_set3(
+                                    rng.borrow_mut().deref_mut(),
+                                    ATTRIBUTE_VALUES.as_ref(),
+                                ),
+                            );
+                        });
+                        count += 1;
+                    }
+                    measurements_while_collecting.fetch_add(count, Ordering::AcqRel);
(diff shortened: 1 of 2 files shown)


utpilla:
@fraillt Thanks for the continued interest in making metrics better! I have two suggestions around sending PRs that I would like you to consider. These are mainly directed towards making the PR review process easy and simple (which would in turn result in the getting PRs merged faster).

1. Send targeted PRs aimed at solving one thing at a time. For example, there's a lot of things going on this PR. There's a new test being added as the PR title suggests, there is moving around of files, refactoring around reuse of logic across projects. Each of these different additions/enhancements could be added incrementally instead of doing it in a single shot.

2. Avoid force pushing the commit. With force push, we have no way to know how many files changed since your last commit. We then have to review all the files again to ensure that the latest commit only made the changes that we were expecting. Retaining the individual commits makes it easy for reviewer.

fraillt:
> @fraillt Thanks for the continued interest in making metrics better! I have two suggestions around sending PRs that I would like you to consider. These are mainly directed towards making the PR review process easy and simple (which would in turn result in the getting PRs merged faster).
> 
>     1. Send targeted PRs aimed at solving one thing at a time. For example, there's a lot of things going on this PR. There's a new test being added as the PR title suggests, there is moving around of files, refactoring around reuse of logic across projects. Each of these different additions/enhancements could be added incrementally instead of doing it in a single shot.
> 
>     2. Avoid force pushing the commit. With force push, we have no way to know how many files changed since your last commit. We then have to review all the files again to ensure that the latest commit only made the changes that we were expecting. Retaining the individual commits makes it easy for reviewer.

I'll try to be more careful regarding force-push in the future.
Regarding targeted PR, I see your point, that targeted PRs are easier to review, but sometimes they doesn't make sense in isolation. For instance I moved `SharedReader` from `opentelemetry-sdk/benches/metric.rs` to `opentelemetry-sdk/src/testing/metrics/metric_reader.rs`, in isolation that looks like a pointless effort, but I do need this functionality for `metrics_collect.rs`.

Basically what I'm trying to say, is that I'm afraid that if I create PR just for refactoring, it might look worthless on its own...

All changes that was done in this PR does relate to one file that brings new functionality `stress/src/bin/metrics_collect.rs`.

utpilla:
> Basically what I'm trying to say, is that I'm afraid that if I create PR just for refactoring, it might look worthless on its own...
> 
> All changes that was done in this PR does relate to one file that brings new functionality `stress/src/bin/metrics_collect.rs`.

I think we could still have a simpler split up of PRs. Something like this:
1. Send a PR that adds the new stress test. (along with the boiler plate code: `SharedReader` etc.) You don't have to add the reuse related changes at this point.
2. Once that gets merged, you could send a refactoring PR to allow for reuse of `SharedReader`. This would showcase the motivation for reuse as well.

fraillt:
I force-pushed, because it basically delete everything I did, except for one file, so I guess it is ok :).
I'll not force-push, once someone reviews and I'll update my revision to address review comments.

stress/src/metrics_collect.rs
```diff
@@ -0,0 +1,239 @@
...
+                handle.join().unwrap();
+            }
+
+            // simultaneously start collecting and creating more measurements
+            for _ in 0..num_threads - 1 {
+                s.spawn(|| {
+                    barrier.wait();
+                    let now = Instant::now();
+                    let mut count = 0;
+                    while is_collecting.load(Ordering::Acquire) {
```
utpilla, line 135: I think we could have a much simpler and effective setup here. If we want to know whether running collect stops the world, it's better to have spawn a thread that keeps calling `reader.collect` on a `loop`. And have other threads, record measurements simultaneously.

stress/src/metrics_collect.rs
```diff
@@ -0,0 +1,239 @@
...
+    let start = Instant::now();
+    while start.elapsed() < Duration::from_secs(3) {
+        res.num_iterations += 1;
+        let is_collecting = AtomicBool::new(false);
+        let measurements_while_collecting = AtomicUsize::new(0);
+        let time_while_collecting = AtomicUsize::new(0);
+        let barrier = Barrier::new(num_threads + 1);
+        std::thread::scope(|s| {
+            // first create bunch of measurements,
+            // so that collection phase wouldn't be "empty"
```
utpilla, line 108: This can be done without spawning new threads. (before you spawn threads for running collect and recording measurements)

stress/src/metrics_collect.rs
fraillt: I did that initially. The problem was that collection phase was not realistic as it had 0 measurements and basically held lock in a loop. So I decided to make it more realistic, by generating some measurements so collection wouldn't be empty.

utpilla: > The problem was that collection phase was not realistic as it had 0 measurements and basically held lock in a loop

I don't follow. You could always record some measurements before you call collect. My concern is around the way this test is setup. The test is using some custom "iterations" which seems unnecessary. You could simply do this instead:
1. Emit some measurements.
2. Spawn a thread that runs collect in a loop.
3. Spawn additional threads that record measurements.
4. Calculate throughput for the measurements recorded in step 3.

If doing the above, leads to zero measurements being recorded, then we have our answer: `Collect is going to "stop the world" as long as it runs`.

As I mentioned in this [comment](https://github.com/open-telemetry/opentelemetry-rust/pull/2247#discussion_r1823416061), unless we plan to use a more efficient synchronization mechanism in `ValueMap`, this stress test would not be adding any value.

fraillt: First of all, this test is adding value, as it is able to measure difference with different "collect" implementation.
Second, running collect in the loop doesn't work with delta temporality, because after first run there will be zero measurements and all other iterations will basically hold write lock in the loop. While in reality no one runs collect in loop and usually there are many measurements and its important that write lock is held as short as possible (which is not the case with current implementation, hence I was able to improve it and this test actually measures it).
I could probably have different tests for delta an cumulative temporality, but I decided to emulate realistic environment.

stress/src/metrics_collect.rs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            let collect_handle = s.spawn(|| {
+                let mut rm = ResourceMetrics {
+                    resource: Resource::empty(),
+                    scope_metrics: Vec::new(),
+                };
+                is_collecting.store(true, Ordering::Release);
+                barrier.wait();
+                reader.collect(&mut rm).unwrap();
```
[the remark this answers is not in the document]
fraillt: For delta temporality I was able to reduce the amount of time a write lock is held while collecting.
For cumulative temporality I simply changed to read lock (currently there is write lock, although cumulative temporality doesn't modify anything, and IMO locking individual measurements is sufficient and much better, then "stop the world until all measurements are read" approach). Although read lock is better, but measurements with new attribute-sets will still be locked. I see two more alternatives: 
* clone entire attribute-set hashmap (acquire read lock, clone hashmap, release lock, iterate individual measurements)
* implement some sort of sharding (more throughtput at the cost of single measurement performance)

In any case, there are ways to improve it:) so I created this PR so we could actually measure different ideas.

stress/src/metrics_collect.rs
utpilla: > While in reality no one runs collect in loop

In reality, no one runs the collect the way it's run in this stress test either. Collect would usually be run periodically in 10/30/60 seconds intervals.

> First of all, this test is adding value, as it is able to measure difference with different "collect" implementation.

> usually there are many measurements and its important that write lock is held as short as possible (which is not the case with current implementation, hence I was able to improve it and this test actually measures it).

This test [claims](https://github.com/open-telemetry/opentelemetry-rust/pull/2247/files#diff-7cd495a1f764caabee463e9e132e20d67bcf0de3da1ea032c6baf39fcf8dbe9bR44-R45) to measure update throughput "while collect is running". The actual implementation of this test however relies on squeezing in some updates around collect (either before collect executes or after it). That's not the same as testing "while collect is running".

I would love to improve collect to take the write lock for the shortest possible time, but it's better tested using a benchmark.

stress/src/metrics_collect.rs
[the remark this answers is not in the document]
utpilla: These are good ideas!

> For cumulative temporality I simply changed to read lock (currently there is write lock, although cumulative temporality doesn't modify anything, 

Great! In that case, we should be able test update throughput while `collect` runs in a loop for `Cumulative`. 🙂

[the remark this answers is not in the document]
lalitb: > implement some sort of sharding (more throughtput at the cost of single measurement performance)

I did try sharding approach earlier in #1564 - to reduce the write-local duration by locking only the specific shard of measurements during collection. However, I encountered some concurrency issues. It’s still an approach worth revisiting at some point.

stress/src/metrics_collect.rs
fraillt: >  The actual implementation of this test however relies on squeezing in some updates before collect runs.
There's a lot of truth to this statement, but none the less it appears to measure something useful  :)

| temporality | type of change                               | before change (measurements/ms) | after change (measurements/ms) |
|-------------|----------------------------------------------|---------------------------------|--------------------------------|
| Cumulative  | changed write to read lock of hashmap        | 9                               | 840                            |
| Delta       | reduce the amount of time write lock is held | 17                              | 56                             |

Ideally I would like to "catch" changes to how attribute-set hashmap are locked with both: existing types of measurements (with existing attribute-set and new attribute-set combination).
Maybe we need to tests these things separately... I don't know, any ideas are welcome :)

fraillt:
#2323 solves same problem in much more precise way.

PR #2249, opened 2024-10-25
Global error handler cleanup - Metrics View 

lalitb:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-sdk/src/metrics/view.rs
+++ b/opentelemetry-sdk/src/metrics/view.rs
@@ -2,7 +2,6 @@ use super::instrument::{Instrument, Stream};
 use glob::Pattern;
 use opentelemetry::{
     metrics::{MetricsError, Result},
-    otel_warn,
 };
 
 fn empty_view(_inst: &Instrument) -> Option<Stream> {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/view.rs
```diff
@@ -102,19 +102,25 @@ impl View for Box<dyn View> {
 /// ```
 pub fn new_view(criteria: Instrument, mask: Stream) -> Result<Box<dyn View>> {
     if criteria.is_empty() {
-        global::handle_error(MetricsError::Config(format!(
-            "no criteria provided, dropping view. mask: {mask:?}"
-        )));
+        otel_warn!(
```
cijothomas, line 105: for a method that returns `Result`, I don't think there is any need of internal logging. If we really want to do some log, then it should be Debug level.

cijothomas (changes requested):
See: https://github.com/open-telemetry/opentelemetry-rust/pull/2249/files#r1817683313

opentelemetry-sdk/src/metrics/view.rs
lalitb: The actual error is getting lost in all these cases, as we just return success with empty view.

cijothomas: if we always return success, then that looks like a flaw. Since it is not introduced in this PR, lets remove the error logging from here, and put a todo (optional). We need to rework the View and can fix these then.

lalitb: Added TODO, and removed logging.

cijothomas (approved):

PR #2250, opened 2024-10-25
Global error handler cleanup - Jaeger Propagator

lalitb:
Fixes #

## Changes

### DIFF
--- a/opentelemetry-jaeger-propagator/src/propagator.rs
+++ b/opentelemetry-jaeger-propagator/src/propagator.rs
@@ -1,8 +1,7 @@
-use opentelemetry::propagation::PropagationError;
 use opentelemetry::{
-    global::{self, Error},
+    otel_error,
     propagation::{text_map_propagator::FieldIter, Extractor, Injector, TextMapPropagator},
-    trace::{SpanContext, SpanId, TraceContextExt, TraceError, TraceFlags, TraceId, TraceState},
+    trace::{SpanContext, SpanId, TraceContextExt, TraceFlags, TraceId, TraceState},
     Context,
 };
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


opentelemetry-jaeger-propagator/src/propagator.rs
```diff
@@ -82,10 +81,11 @@ impl Propagator {
 
         let parts = header_value.split_terminator(':').collect::<Vec<&str>>();
         if parts.len() != 4 {
-            global::handle_error(Error::Propagation(PropagationError::extract(
-                "invalid jaeger header format",
-                "JaegerPropagator",
-            )));
+            otel_error!(
```
cijothomas, line 84: I think warning or info is good enough here.

cijothomas (changes requested):
Use Warning/Info instead of Error severity.
https://github.com/open-telemetry/opentelemetry-rust/pull/2250/files#r1817682423

opentelemetry-jaeger-propagator/src/propagator.rs
```diff
@@ -82,10 +81,11 @@ impl Propagator {
 
         let parts = header_value.split_terminator(':').collect::<Vec<&str>>();
         if parts.len() != 4 {
-            global::handle_error(Error::Propagation(PropagationError::extract(
-                "invalid jaeger header format",
-                "JaegerPropagator",
-            )));
+            otel_warn!(
```
cijothomas, line 84: I haven't followed the entire code path, so unclear what is the impact of this Warning. For eg: if an end users see this log entry, what should they be doing? 
For all user facing logs, we should be very clear about the impact of this - will a span be dropped? will a span be using different parent ? will a span be created as root span? will we just ignore the header for tracestate and baggage too? or just ignore the traceparent header? something else?

I am not saying this should be fixed in this PR, as these are existing issues, nothing newly introduced in this PR. Since Metrics, Logs are where we are actively working towards Stable, we can probably merge this now, and come back when we start working towards Tracing Stable release?

cijothomas (approved):
I do have some concerns about whether the logs are actionable/informative enough for end users - https://github.com/open-telemetry/opentelemetry-rust/pull/2250/files#r1817692142

But I don't think this should be a blocker now, as Tracing signal is unlikely to be in stable state soon. (For Metrics, Logs, we need to make sure every log is informative/actionable, as it is getting closer to stable release now)

PR #2256, opened 2024-10-28
Global error handler cleanup - Trace SDK

lalitb:
Fixes #

## Changes


opentelemetry-sdk/src/trace/config.rs
```diff
@@ -129,9 +129,9 @@ impl Default for Config {
                     if let Some(r) = ratio {
                         Box::new(Sampler::TraceIdRatioBased(r))
                     } else {
-                        handle_error(
-                            Error::Other(String::from(
-                                "Missing or invalid OTEL_TRACES_SAMPLER_ARG value. Falling back to default: 1.0"))
+                        otel_warn!(
+                            name: "TracesSampler.TraceIdRatio.Error",
```
cijothomas, line 133: the name make it sound like there is something wrong with TraceSampler, but the issue is with invalid configuration.

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -129,9 +129,9 @@ impl Default for Config {
...
+                            message= "Missing or invalid OTEL_TRACES_SAMPLER_ARG env variable. Falling back to default: 1.0"
```
cijothomas, line 134: need more user actionable message.
OTEL_TRACES_SAMPLER env variable has the value traceidratio and hence ratio should be provided by OTEL_TRACES_SAMPLER_ARG env variable which is missing/invalid. TraceIdRatioBased sampler will still be enabled, but will a fallback value of 1.0 (i.e 100% traces are accepted) for ratio.

opentelemetry-sdk/src/trace/config.rs
lalitb: done

opentelemetry-sdk/src/trace/config.rs
lalitb: done

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -129,9 +129,9 @@ impl Default for Config {
                     if let Some(r) = ratio {
                         Box::new(Sampler::TraceIdRatioBased(r))
                     } else {
-                        handle_error(
-                            Error::Other(String::from(
-                                "Missing or invalid OTEL_TRACES_SAMPLER_ARG value. Falling back to default: 1.0"))
+                        otel_warn!(
+                            name: "Sampler.TraceIdRatio.InvalidArgument",
```
cijothomas, line 133: the name is misleading. This is not an issue occurring in the Sampler (it affects sampler), but rather a configuration issue.

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -149,37 +149,45 @@ impl Default for Config {
...
                     }
                 }
                 "parentbased_jaeger_remote" => {
-                    handle_error(
-                        Error::Other(String::from(
-                        "Unimplemented parentbased_jaeger_remote sampler. Falling back to default: parentbased_always_on"
-                    )));
+                    otel_warn!(
+                        name: "Sampler.ParentBasedJaegerRemote.Unsupported",
+                        message = "ParentBased JaegerRemote sampler is not implemented in this SDK version. Using fallback sampler: ParentBased(AlwaysOn). Configure an alternative sampler using OTEL_TRACES_SAMPLER"
```
cijothomas, line 164: OTEL_TRACES_SAMPLER env variable has the value jaeger-remote, which is unsupported. Will fallback to using ParentBased(AlwaysOn) sampler.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -134,7 +134,10 @@ impl SpanProcessor for SimpleSpanProcessor {
             .and_then(|mut exporter| futures_executor::block_on(exporter.export(vec![span])));
 
         if let Err(err) = result {
-            global::handle_error(err);
+            otel_error!(
+                name: "SimpleProcessor.OnEnd.Error",
+                reason = format!("{:?}", err)
```
cijothomas, line 139: do we need error severity? would the information here be duplicate of what an exporter might already be logging?
I am also unsure if the message is really actionable by end users. there is no indication of what happened or what is the impact of this..

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -354,7 +367,10 @@ impl<R: RuntimeChannel> BatchSpanProcessorInternal<R> {
                     let export_task = self.export();
                     let task = async move {
                         if let Err(err) = export_task.await {
-                            global::handle_error(err);
+                            otel_error!(
+                                name: "BatchSpanProcessor.Export.Error",
```
cijothomas, line 371: There is another log with the name "BatchProcessor.OnEnd.Error". Is this duplicate ?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -313,14 +319,21 @@ impl<R: RuntimeChannel> BatchSpanProcessorInternal<R> {
...
+                        name: "BatchSpanProcessor.Flush.SendResultError",
+                        reason = format!("{:?}", result)
+                    );
                 }
             } else if let Err(err) = result {
-                global::handle_error(err);
+                // If no channel is provided and the export operation encountered an error,
+                // log the error directly here.
+                otel_error!(
+                    name: "BatchSpanProcessor.Flush.ExportError",
```
cijothomas, line 333: unsure why we need Error level internal log for Flush operation? Aren't we returning Flush `Result` back to user already?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -246,7 +249,10 @@ impl<R: RuntimeChannel> SpanProcessor for BatchSpanProcessor<R> {
         let result = self.message_sender.try_send(BatchMessage::ExportSpan(span));
 
         if let Err(err) = result {
-            global::handle_error(TraceError::Other(err.into()));
+            otel_error!(
```
cijothomas, line 252: will this be triggered when channel is full? if yes, we shouldn't do such logging as that can overwhelm internal logs

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: It doesn't in existing code, can add a TODO. For tracing, the PR is doing search-replace for the global handler.

opentelemetry-sdk/src/trace/config.rs
lalitb: I didn't get this - isn't this the configuration for sampler which is causing the issue? In that case, it is related to sampler.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: > would the information here be duplicate of what an exporter might already be logging?

Not necessarily - exporter may choose to return the export status to be logged here. The status may be actionable in that case. For logs, we check the error type and then decide whether to log as `error` or `debug`. Similar logic needs to be done here. Will add a TODO here. Meanwhile, do you want to change this to `debug` ?

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: yes it will. Will make it debug in that case.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Good catch. We have `SimpleProcessor.OnEnd.Error`. To make it consistent, this can be modified to `BatchProcessor.OnEnd.Error`

opentelemetry-sdk/src/trace/config.rs
lalitb: rephrased this here, and also other places.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Added TODO, and changed to otel_debug.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: updated as below:

```rust
        let result = self.message_sender.try_send(BatchMessage::ExportSpan(span));

        if let Err(err) = result {
            otel_debug!(
                name: "BatchSpanProcessor.OnEnd.SendResultError",
                reason = format!("{:?}", TraceError::Other(err.into()))
            );
        }
```

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: added TODO.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: These are the event types:

`SimpleProcessor.OnEnd.Error`: export error
`BatchSpanProcessor.OnEnd.Error`: export error
`BatchSpanProcessor.OnEnd.SendResultError`: error sending export status/

opentelemetry-sdk/src/trace/config.rs
cijothomas: this is purely configuration issue, that *affects* sampler, not an issue inside sampler. Sampler's code is not even executed, so the name is misleading, as it seem to indicate some issue inside Sampler. If this log is emitted from inside a Sampler's `new()` method, then the name makes sense, but in this case, it is occurring outside Sampler.

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -149,37 +149,45 @@ impl Default for Config {
...
+                        message = "OTEL_TRACES_SAMPLER is set to 'xray'. AWS X-Ray sampler is not implemented in this SDK version. Using fallback sampler: ParentBased(AlwaysOn). Configure an alternative sampler using OTEL_TRACES_SAMPLER"
+                    );
                     Box::new(Sampler::ParentBased(Box::new(Sampler::AlwaysOn)))
                 }
                 s => {
-                    handle_error(
-                        Error::Other(format!("Unrecognised OTEL_TRACES_SAMPLER value: {}. Falling back to default: parentbased_always_on",
-                        s
-                    )));
+                    otel_warn!(
```
cijothomas, line 183: name: TracerProvider.Configuration.InvalidSampler
message : the why which includes the list of supported values, and also the value input by user. the impact - usually fallback part.

some thoughts above.

opentelemetry-sdk/src/trace/config.rs
lalitb: updated as below:
```rust
    name: "TracerProvider.Config.Sampler.InvalidArgument",
    message = "OTEL_TRACES_SAMPLER is set to 'traceidratio' but OTEL_TRACES_SAMPLER_ARG environment variable is missing or invalid. OTEL_TRACES_SAMPLER_ARG must be a valid float between 0.0 and 1.0 representing the desired sampling probability (0.0 = no traces sampled, 1.0 = all traces sampled, 0.5 = 50% of traces sampled). Falling back to default ratio: 1.0 (100% sampling)"
 );
```

opentelemetry-sdk/src/trace/config.rs
lalitb: upated as below:
```rust
otel_warn!(
    name: "TracerProvider.Config.Sampler.InvalidSamplerType",
    message = format!(
         "Unrecognized sampler type '{}' in OTEL_TRACES_SAMPLER environment variable. Valid values are: always_on, always_off, traceidratio, parentbased_always_on, parentbased_always_off, parentbased_traceidratio. Using fallback sampler: ParentBased(AlwaysOn)", s
     ),
     sampler = s
);
```

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -129,9 +129,9 @@ impl Default for Config {
                     if let Some(r) = ratio {
                         Box::new(Sampler::TraceIdRatioBased(r))
                     } else {
-                        handle_error(
-                            Error::Other(String::from(
-                                "Missing or invalid OTEL_TRACES_SAMPLER_ARG value. Falling back to default: 1.0"))
+                        otel_warn!(
+                            name: "TracerProvider.Config.Sampler.InvalidArgument",
+                            message = "OTEL_TRACES_SAMPLER is set to 'traceidratio' but OTEL_TRACES_SAMPLER_ARG environment variable is missing or invalid. OTEL_TRACES_SAMPLER_ARG must be a valid float between 0.0 and 1.0 representing the desired sampling probability (0.0 = no traces sampled, 1.0 = all traces sampled, 0.5 = 50% of traces sampled). Falling back to default ratio: 1.0 (100% sampling)"
```
ThomsonTan, line 134: Include the current (invalid) value of `OTEL_TRACES_SAMPLER_ARG` in the warning message?

opentelemetry-sdk/src/trace/config.rs
cijothomas: TracerProvider.Config.Sampler.InvalidSamplerType can be shortened, as it repeats Sampler twice.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -354,7 +369,10 @@ impl<R: RuntimeChannel> BatchSpanProcessorInternal<R> {
                     let export_task = self.export();
                     let task = async move {
                         if let Err(err) = export_task.await {
-                            global::handle_error(err);
+                            otel_error!(
+                                name: "BatchSpanProcessor.OnEnd.Error",
```
cijothomas, line 373: this doesn't look correct to me, as this is occurring in background thread/task, and *not* in BSP's OnEnd method.

opentelemetry-sdk/src/trace/config.rs
lalitb: Done:
```
otel_warn!(
    name: "TracerProvider.Config.Sampler.InvalidArgument",
    message = "OTEL_TRACES_SAMPLER is set to 'traceidratio' but OTEL_TRACES_SAMPLER_ARG environment variable is missing or invalid. OTEL_TRACES_SAMPLER_ARG must be a valid float between 0.0 and 1.0 representing the desired sampling probability (0.0 = no traces sampled, 1.0 = all traces sampled, 0.5 = 50% of traces sampled). Falling back to default ratio: 1.0 (100% sampling)",
    otel_traces_sampler_arg = format!("{:?}", sampler_arg)
);
```

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Good point. Changed it to `BatchSpanProcessor.Export.Error`, to be consistent with `BatchLogProcessor.Export.Error`, as the error occurs while processing export message.

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -143,43 +144,52 @@ impl Default for Config {
...
-                        Error::Other(format!("Unrecognised OTEL_TRACES_SAMPLER value: {}. Falling back to default: parentbased_always_on",
-                        s
-                    )));
+                    otel_warn!(
+                        name: "TracerProvider.Config.Sampler.InvalidSamplerType",
+                        message = format!(
+                            "Unrecognized sampler type '{}' in OTEL_TRACES_SAMPLER environment variable. Valid values are: always_on, always_off, traceidratio, parentbased_always_on, parentbased_always_off, parentbased_traceidratio. Using fallback sampler: ParentBased(AlwaysOn)",
+                            s
+                        ),
+                        sampler = s
```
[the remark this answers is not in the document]
lalitb: Just thought separate attributes are in general helpful as search option in backend queries, mayn't be in this case. Will remove it as attribute, as message looks incomplete without it.

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -143,43 +144,51 @@ impl Default for Config {
...
+                    );
                     Box::new(Sampler::ParentBased(Box::new(Sampler::AlwaysOn)))
                 }
                 s => {
-                    handle_error(
-                        Error::Other(format!("Unrecognised OTEL_TRACES_SAMPLER value: {}. Falling back to default: parentbased_always_on",
-                        s
-                    )));
+                    otel_warn!(
+                        name: "TracerProvider.Config.Sampler.InvalidSamplerType",
```
cijothomas, line 186: TracerProvider.Config.Sampler.InvalidSamplerType
TracerProvider.Config.Sampler.Unsupported
TracerProvider.Config.Sampler.InvalidArgument

Lets use consistent naming. The 1st one has Sampler word repeated, while others don't.
A shorter name like `TracerProvider.InvalidSamplerType` should be sufficient IMHO.
Alternatively, name can be `TracerProviderInvalidConfiguration`, and message can explain what went wrong, and additional attributes can be used to describe the component being configured incorrectly.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -246,7 +250,10 @@ impl<R: RuntimeChannel> SpanProcessor for BatchSpanProcessor<R> {
         let result = self.message_sender.try_send(BatchMessage::ExportSpan(span));
 
         if let Err(err) = result {
-            global::handle_error(TraceError::Other(err.into()));
+            otel_debug!(
+                name: "BatchSpanProcessor.Export.SendResultError",
```
cijothomas, line 254: this is confusing. This is the `on_end` method.. Given this is BatchProcessor, this will only queue up spans on `on_end` and export occurs asynchronously. The name "BatchSpanProcessor.Export.SendResultError" seem to indicate that some sort of Export is going on here, which is incorrect/misleading, as export never occurs on `on_end` call for BatchSpanProcessor. Also "SendResultError" is not reflecting anything going on here...

opentelemetry-sdk/src/trace/config.rs
lalitb: Have named it to:
TracerProvider.Config.InvalidSamplerArgument
TracerProvider.Config.UnsupportedSampler
TracerProvider.Config.InvalidSamplerType

Wanted to keep 3 levels of naming as done in other places. Let me know if this still doesn't look consistent.

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Good catch. Have named it to `BatchSpanProcessor.OnEnd.ExportQueueingFailed`.

cijothomas (approved):
Good to merge now,
Have to revisit all the logs here in the future to ensure we have the right balance between too much vs too little.

opentelemetry-sdk/src/trace/config.rs
cijothomas: From consistency, this is better. Though I am unsure if the name is easily followable by end users. User may not have ever touched/used `Config` at all, and when they see 'Config' in event name, it may not be obvious. Good to follow up later.

cijothomas:
@lalitb There are quite large number of unresolved conversations in the PR, hard to tell if they are still relevant. Could you resolve the ones that are addressed and merge?

opentelemetry-sdk/src/trace/config.rs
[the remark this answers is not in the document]
lalitb: this is done./

lalitb:
> @lalitb There are quite large number of unresolved conversations in the PR, hard to tell if they are still relevant. Could you resolve the ones that are addressed and merge?

I have addressed all the comments, so resolved them. Merging now. Thanks for the thorough review

PR #2260, opened 2024-10-28
Remove Global Error handler.

lalitb:
Fixes #2175

Needs #2259 #2257 #2256  to be merged first.

## Changes

- Removed `global::set_error_handler` and `global::handle_error`, as the `global_handle_error` usage inside the opentelemetry crates has already been replaced with `global::otel_info`, `otel_warn`, `otel_debug` and `otel_error`

## Migration Guide:

> 📌Note: This setup requires the internal-logs feature to be enabled for the OpenTelemetry crates. This feature is enabled by default, so application should ensure not to disable it explicitly for these crates in its `Cargo.toml`.

This guide explains how to migrate from `global::set_error_handler` to a new setup with `otel_*` macros and Tokio tracing. If your application previously used `global::set_error_handler`, remove it as part of this migration. If not, this configuration ensures errors previously going to `eprintln!` are now handled through structured tracing.

**Step 1: Remove set_error_handler (if used)**
If your code currently sets a custom error handler using global::set_error_handler, remove this line and any associated custom handler function.
```rust
// Remove any custom error handler
// global::set_error_handler(custom_error_handler).unwrap();
```

**Step 2: Internal OpenTelemetry Logs Filter and Layer:**
Set up a layer to handle OpenTelemetry internal logs and stop them from propagating to tracing bridge/appender.

1. ***Internal OpenTelemetry Logs Filter and Layer:***
    - This layer captures logs with the opentelemetry prefix and prints them to eprintln!.
 
```rust
use tracing_subscriber::filter::filter_fn;
use tracing_subscriber::fmt::Layer;

let opentelemetry_layer = Layer::new()
    .with_writer(std::io::stderr)
    .with_filter(filter_fn(|metadata| metadata.target().starts_with("opentelemetry")));
```

2. ***Bridge Layer for OpenTelemetry Tracing (Excludes OpenTelemetry Internal Logs)::*** 
    - To prevent logs with the opentelemetry prefix from propagating further, apply a filter_fn in this layer that excludes any logs with the opentelemetry target prefix.
 ```rust
use opentelemetry_appender_tracing::layer;

let non_opentelemetry_filter = filter_fn(|metadata| !metadata.target().starts_with("opentelemetry"));
let otel_bridge_layer = layer::OpenTelemetryTracingBridge::new(&cloned_provider)
    .with_filter(non_opentelemetry_filter);
```

**Step 3: Register Layers in the Tracing Registry**
```rust
use tracing_subscriber::prelude::*;

[...]

PR #2266, opened 2024-11-01
Move Logs and Metrics error constructs to SDK

### LINKED ISSUE #1042: SDK/Export concepts leaked into API.

lalitb:
Fixes #1042 

## Changes
~Needs https://github.com/open-telemetry/opentelemetry-rust/pull/2256 and https://github.com/open-telemetry/opentelemetry-rust/pull/2260 to be merged first. Also the PR is on top of #2260, so better to review this once that PR is merged.~

Summary of changes:

1. `ExportError` trait:
     - Moved from `opentelemetry/src/common.rs` to `opentelemetry-sdk/src/export/mod.rs`
     - Trace API now has its own ExportError trait in `opentelemetry/src/trace/mod.rs`. This will be eventually consolidated within SDK once we work on traces.

2. `LogError` enum and `LogResult` type alias :
     - Moved from `opentelemetry/src/logs/mod.rs` to `opentelemetry-sdk/src/logs/error.rs`

3. `MetricError` enum and `MetricResult` type alias:
    - Moved from `opentelemetry/src/metrics/mod.rs`  to `opentelemetry-sdk/src/metrics/error.rs`

4. `TraceError` enum and `TraceResult` type alias:
     - No changes. Added TODO to move in future.


opentelemetry-sdk/src/error.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[cfg(feature = "logs")]
+use crate::logs::LogError;
+#[cfg(feature = "metrics")]
+use crate::metrics::MetricError;
+use opentelemetry::propagation::PropagationError;
+#[cfg(feature = "trace")]
+use opentelemetry::trace::TraceError;
+
+/// Wrapper for error from both tracing and metrics part of open telemetry.
```
cijothomas, line 11: Though this is moving from just one crate to another, I'd like to first confirm that these are indeed required. I understand that it is desirable to return specialized errors from user-facing APIs, but most of the errors here are only for internal logging only. Is there any additional value provided by this?

cijothomas: See https://github.com/open-telemetry/opentelemetry-rust/pull/2241#discussion_r1815986714 for prior discussion. 
If we don't plan to keep this, then its best to just remove it, instead of moving from api to sdk and then removing it again.
(Note: These are still public APIs in the sdk crate)

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust/pull/2266/files#r1826889942
It maybe better off removing them altogether than moving from api to sdk crate.

lalitb:
> See https://github.com/open-telemetry/opentelemetry-rust/pull/2266/files#r1826889942 It maybe better off removing them altogether than moving from api to sdk crate.

I am not sure if we agreed on removing this based on that comment. This needs some more discussions. These error constructs were always used in the SDK, so it's logical to first move their definition to SDK. This would also enable us to proceed with the RC for the Log (bridge) API and then make a decision regarding the SDK.

cijothomas:
> > See https://github.com/open-telemetry/opentelemetry-rust/pull/2266/files#r1826889942 It maybe better off removing them altogether than moving from api to sdk crate.
> 
> I am not sure if we agreed on removing this based on that comment. This needs some more discussions. These error constructs were always used in the SDK, so it's logical to first move their definition to SDK. This would also enable us to proceed with the RC for the API and then make a decision regarding the SDK.

Please use this PR to drive a decision for the same.

lalitb:
> Please use this PR to drive a decision for the same.

Why do we need to do that here?  The API cleanup is logical, and this PR make sense to go first without any breaking anything. We can have separate issue/PR to discuss the breaking changes in terms of removing the constructs altogether.

cijothomas:
> > Please use this PR to drive a decision for the same.
> 
> Why do we need to do that here? The API cleanup is logical, and this PR make sense to go first without any breaking anything. We can have separate issue/PR to discuss the breaking changes in terms of removing the constructs altogether.

Because, as mentioned [here](https://github.com/open-telemetry/opentelemetry-rust/pull/2266#discussion_r1826890055), this PR is introducing new public API to the sdk crate, that we may need to remove in future again.

Totally agree with removing it from `opentelemetry` crate.

lalitb:
The Result/Error are used within SDK in these public APIs:

- LogExporterBuilder::build() -> Result<LogExporter, opentelemetry_sdk::logs::LogError> 
- LogExporter::export() -> LogResult<()> 
- LoggerProvider::force_flush(&self) -> Vec<LogResult<()>> 
- LoggerProvider::shutdown(&self) -> LogResult<()> 
Similarly for Metrics, probably while returning the error from View creation.

The question is - if we want to remove the Result/Error, what are we instead looking for - returning bool as status, or just the error string?
If the motive is to keep the SDK's public surface minimal, we can still achieve this with result/error construct minimal in future before SDK getting stable.

```rust
/// Describe the result of operations in log SDK.
pub type LogResult<T> = Result<T, LogError>;

#[derive(Error, Debug)]
#[non_exhaustive]
/// Errors returned by the log SDK.
pub enum LogError {
    /// A generic error message for errors in the log SDK.
    #[error("{0}")]
    Message(String),
}

```

cijothomas (changes requested):
From Community Meeting:
1. This is mandatory to declare opentelemetry as RC.
2. The APIs being moved to SDK might need a re-review in the future, but that is okay.
3. Make changelog as this is user breaking change.
4. Must merge for 0.27, so not worth trying to split into shorter ones.
5. Traces is untouched, lets leave it for a future release.

opentelemetry-sdk/src/export/mod.rs
```diff
@@ -8,4 +8,8 @@ pub mod logs;
 #[cfg_attr(docsrs, doc(cfg(feature = "trace")))]
 pub mod trace;
 
-pub use opentelemetry::ExportError;
+/// Marker trait for errors returned by exporters
```
utpilla, line 11: Not introduced in this PR but this is not a marker trait. Up to you if you want to fix this in this PR or later.

opentelemetry/src/trace/mod.rs
```diff
@@ -186,9 +186,17 @@ pub use self::{
     tracer::{SamplingDecision, SamplingResult, SpanBuilder, Tracer},
     tracer_provider::TracerProvider,
 };
-use crate::{ExportError, KeyValue};
+use crate::KeyValue;
 use std::sync::PoisonError;
 
+// TODO - Move ExportError and TraceError to opentelemetry-sdk
+
+/// Marker trait for errors returned by exporters
```
utpilla, line 194: Same as https://github.com/open-telemetry/opentelemetry-rust/pull/2266#discussion_r1830195663

opentelemetry/CHANGELOG.md
```diff
@@ -39,6 +39,17 @@ let counter = meter.u64_counter("my_counter").build();
   - The default behavior of `global::handle_error` was to log the error using `eprintln!`. With otel macro, the internal logs get emitted via `tracing` macros of matching severity. Users now need to configure the `tracing` layer to capture these logs.
   - Refer to this PR description for migration guide. Also refer to [self-diagnostics](https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/self-diagnostics) example on how to configure the tracing layer for internal logs.
 
+- [#2266](https://github.com/open-telemetry/opentelemetry-rust/pull/2266)
+   - Moved `ExportError` trait from `opentelemetry::ExportError` to `opentelemetry_sdk::export::ExportError`
+   - Created new trait `opentelemetry::trace::ExportError` for trace API. This would be eventually be consolidated with ExportError in the SDK.
+   - Moved `LogError` enum from `opentelemetry::logs::LogError` to `opentelemetry_sdk::logs::LogError`
+   - Moved `LogResult` type alias from `opentelemetry::logs::LogResult` to `opentelemetry_sdk::logs::LogResult`
+   - Renamed `opentelemetry::metrics::Result` type alias to `opentelemetry::metrics::MetricResult`
```
utpilla, line 47: I don't think we need an entry for renaming this as we have also moved it to SDK.

opentelemetry/src/global/mod.rs
```diff
@@ -130,15 +130,13 @@
 //! [`MeterProvider`]: crate::metrics::MeterProvider
 //! [`set_meter_provider`]: crate::global::set_meter_provider
 
-mod error_handler;
```
utpilla, line 133: Could we also remove the `error_handler.rs` file?

lalitb: This is already removed, and it's content moved to otel_sdk:

{91AE70A0-90DE-4DAE-B955-8F24899D4532}

utpilla: Ah yes. My bad! It would render as move in the diff.

utpilla (approved):

opentelemetry/CHANGELOG.md
```diff
@@ -39,6 +39,15 @@ let counter = meter.u64_counter("my_counter").build();
...
   - Refer to this PR description for migration guide. Also refer to [self-diagnostics](https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/self-diagnostics) example on how to configure the tracing layer for internal logs.
 
+- [#2266](https://github.com/open-telemetry/opentelemetry-rust/pull/2266)
+   - Moved `ExportError` trait from `opentelemetry::ExportError` to `opentelemetry_sdk::export::ExportError`
+   - Created new trait `opentelemetry::trace::ExportError` for trace API. This would be eventually be consolidated with ExportError in the SDK.
+   - Moved `LogError` enum from `opentelemetry::logs::LogError` to `opentelemetry_sdk::logs::LogError`
+   - Moved `LogResult` type alias from `opentelemetry::logs::LogResult` to `opentelemetry_sdk::logs::LogResult`
+   - Moved `MetricError` enum from `opentelemetry::metrics::MetricError` to `opentelemetry_sdk::metrics::MetricError`
+   - Moved `MetricResult` type alias from `opentelemetry::metrics::MetricResult` to `opentelemetry_sdk::metrics::MetricResult`
+  These changes shouldn't directly affect the users of OpenTelemetry crates, as these constructs are used in SDK and Exporters.
```
cijothomas, line 49: nit addition:
if you are an author of an sdk component/plug-in, like an exporter etc. please use these types from sdk. Add link to the sdk's changelog.md file.

cijothomas (approved):
Thanks for cleaning up the API crate! 
Lets follow up in the future to review these to make everything is needed in sdk also as public.

PR #2267, opened 2024-11-02
Add collect methods on ValueMap

fraillt:
Design discussion issue #2145

## Changes

Mostly extract `DataPoint` collection logic from specific metrics implementations into `ValueMap` under `collect_readonly` and `collect_and_reset` functions.
I tried to preserve existing logic with few exceptions:
* for cumulative temporality - acquire read lock on hashmap instead of write lock.
* for delta temporality - since hashmap is cleared anyway, reduce the time hashmap is locked. Instead of "lock->iterate(consume)->unlock" changed to "lock->consume->unlock->iterate".

There are a lot more things that could be improved, but with this change it will become more convenient/simpler to experiment/implement further improvements in the future.


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -22,10 +23,7 @@ use crate::metrics::AttributeSet;
 pub(crate) static STREAM_OVERFLOW_ATTRIBUTES: Lazy<Vec<KeyValue>> =
     Lazy::new(|| vec![KeyValue::new("otel.metric.overflow", "true")]);
 
-pub(crate) trait Aggregator<T>
-where
-    T: Number,
-{
+pub(crate) trait Aggregator {
```
cijothomas, line 26: is this change doable as a standalone, separate PR, to keep the PRs shorter and easier to review.?

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -135,6 +134,66 @@ where
             );
         }
     }
+
+    /// Iterate through all attribute sets and populate `DataPoints` in readonly mode.
```
cijothomas, line 138: nit: Add a comment here to state that this is used in Cumulative temporality mode where ValueMap is not cleared, and hence Read lock is sufficient.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -135,6 +134,66 @@ where
...
+    pub(crate) fn collect_readonly<Res, MapFn>(&self, dest: &mut Vec<Res>, mut map_fn: MapFn)
+    where
+        MapFn: FnMut(Vec<KeyValue>, &A) -> Res,
+    {
+        prepare_data(dest, self.count.load(Ordering::SeqCst));
+        if self.has_no_attribute_value.load(Ordering::Acquire) {
+            dest.push(map_fn(vec![], &self.no_attribute_tracker));
+        }
+
+        let Ok(trackers) = self.trackers.read() else {
```
cijothomas, line 148: good improvement here by removing the need for write lock!

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -135,6 +134,66 @@ where
...
+    {
+        prepare_data(dest, self.count.load(Ordering::SeqCst));
+        if self.has_no_attribute_value.swap(false, Ordering::AcqRel) {
+            dest.push(map_fn(
+                vec![],
+                self.no_attribute_tracker.clone_and_reset(&self.config),
+            ));
+        }
+
+        let trackers = match self.trackers.write() {
```
cijothomas, line 173: Something we can come back in the future - the current behavior of aggressively cleaning up valuemap during every collect() is not really needed - we only need to remove those trackers which did not have any new update since the last collect. Doing this requires additional bookkeeping costs, but that might be a better overall design.
Anyway, not for this PR!

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -135,6 +134,66 @@ where
             );
         }
     }
+
+    /// Iterate through all attribute sets and populate `DataPoints` in readonly mode.
+    pub(crate) fn collect_readonly<Res, MapFn>(&self, dest: &mut Vec<Res>, mut map_fn: MapFn)
```
cijothomas, line 139: Another issue worth exploring further: Instead of passing mutable Vec or owned Vec to Exporters, we can simply pass immutable refs, so the sdk can allocate the DataPoints once, and maintain it forever. Exporters just get a immutable ref to Datapoints only. For another PR definitely!

opentelemetry-sdk/src/metrics/internal/mod.rs
fraillt: Done, created separate revision for this #2269.
Updated this revision with git rebase, (I think it's ok, as it basically deleted code that shouldn't be there)

opentelemetry-sdk/src/metrics/internal/mod.rs
fraillt: I'm not sure if I follow this idea... you mean having immutable `&Vec<DataPoints>`, then how would you update value and time, and exemplars... ? (but maybe let's leave this for another time :))

opentelemetry-sdk/src/metrics/internal/mod.rs
fraillt: I probably understand your idea, but I'm afraid that this might require holding write lock for a long time. In this case, write lock is super fast, just replace variable basically.

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: > then how would you update value and time, and exemplars... ?

Exporters don't need to update them, so they don't need mutable one. (lets comeback to this in future)

opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -37,6 +36,14 @@ where
             buckets: Mutex::new(Buckets::<T>::new(*count)),
         }
     }
+
+    fn clone_and_reset(&self, count: &usize) -> Self {
+        let mut current = self.buckets.lock().unwrap_or_else(|err| err.into_inner());
+        let cloned = replace(current.deref_mut(), Buckets::new(*count));
```
utpilla, line 42: This would lead to unnecessary allocation. The replacement with `Buckets::new` is only needed for the no attribute case as there could be newer updates trying to update the same tracker. For trackers with attributes, we could simply `mem::take` the buckets instead of creating a new set of buckets for each tracker.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -130,6 +134,68 @@ where
...
+            dest.push(map_fn(
+                vec![],
+                self.no_attribute_tracker.clone_and_reset(&self.config),
+            ));
+        }
+
+        let trackers = match self.trackers.write() {
+            Ok(mut trackers) => {
+                self.count.store(0, Ordering::SeqCst);
+                take(trackers.deref_mut())
```
utpilla, line 178: This would lead to additional allocation as we would be replacing `trackers` with a new `Hashmap` (with capacity - 0) on each collect. So, after every collect, we would be allocating to add entries to the Hashmap. This was avoided earlier when we used `drain` on `Hashmap`, which retained the Hashmap's capacity.

One approach to solve this would be to add another `RwLock<HashMap>` field to the `ValueMap` struct and instead of calling `mem::take` here, we could `mem::swap` with current trackers with this newly added Hashmap.

opentelemetry-sdk/src/metrics/internal/histogram.rs
fraillt: Yes, you're absolutely right we should be able to simply 'take' it from trackers with attributes, but at the moment we cant (or rather it would be super inefficient), because hashmap can contain several instances of same tracker (one with sorted attrib set, an N with not unsorted).
I actually plan to create revision for that as well (was afraid to do this as part as this revision, to reduce scope):)
Idea is to split this hashmap into two: one with sorted attributes  and another all attributes for faster lookup. This change would greatly improve collection performance, and by accident would solve #2093 as well:)

In summary, this is temporary minor inefficiency that will be solved:)

opentelemetry-sdk/src/metrics/internal/mod.rs
fraillt: I tried to minimize holding write lock, as much as possible (i cannot release lock if I want to use 'drain').
In #2247 I have observed almost 4x improvement (when collect phase is running), with this change.
Regarding extra allocation, I don't think that it makes a huge difference, (given that they happen constantly, and we really don't notice them that much:)). 
Maybe I could improve this just a bit by doing something like `replace(trackers, Hashmap::with_capacity(n))`? and leave other potential improvements for the future.
Would that work for you?

opentelemetry-sdk/src/metrics/internal/histogram.rs
utpilla: > but at the moment we cant (or rather it would be super inefficient), because hashmap can contain several instances of same tracker

I didn't get that. Fixing this extra allocation here does not depend on the `Hashmap` containing two sets of the same attributes. There is a relatively simple fix to this:

1. Add a new method to the `Aggregator` trait and update `collect_and_reset` to call this new method for trackers with attributes

```rust
pub(crate) trait Aggregator {
    ...

    /// Return the current value
    fn get_value(&self, init: &Self::InitConfig) -> Self;
}

pub(crate) fn collect_and_reset<Res, MapFn>(&self, dest: &mut Vec<Res>, mut map_fn: MapFn)
where
    MapFn: FnMut(Vec<KeyValue>, A) -> Res,
{
    prepare_data(dest, self.count.load(Ordering::SeqCst));

    ...

    for (attrs, tracker) in trackers.drain() {
            dest.push(map_fn(attrs, tracker.get_value(&self.config)));
        }
    }
}
```

2. Implement the new method in the `HistogramTracker` by not calling `Buckets::new`. Notice that we call `mem::take` here and not `mem::replace`.

```rust
fn get_value(&self, _init: &Self::InitConfig) -> Self
{
    let mut current = self.buckets.lock().unwrap_or_else(|err| err.into_inner());
    Self {
        buckets: Mutex::new(take(&mut current)),
    }
}
```

This should work.

opentelemetry-sdk/src/metrics/internal/mod.rs
utpilla: I'm totally in favor of reducing the time held for write lock, so I'm not asking to go back to calling `drain` with the write lock. However, since we are optimizing collect anyway why not just avoid any amount of allocation that we can?

One simple way to fix this would be:

1. Add another field `trackers_for_collect` to the `ValueMap` struct

```rust
pub(crate) struct ValueMap<A>
where
    A: Aggregator,
{
    /// Trackers store the values associated with different attribute sets.
    trackers: RwLock<HashMap<Vec<KeyValue>, Arc<A>>>,

    /// Trackers store the values associated with different attribute sets.
    trackers_for_collect: RwLock<HashMap<Vec<KeyValue>, Arc<A>>>,

    ...
}

impl<A> ValueMap<A>
where
    A: Aggregator,
{
    fn new(config: A::InitConfig) -> Self {
        ValueMap {
            trackers: RwLock::new(HashMap::new()),
            trackers_for_collect: RwLock::new(HashMap::new()),

            ...
        }
    }
}
```

2. Update the `collect_and_reset` method to:
   - Use `mem::swap` to swap the trackers used for updates with the trackers used for collect.
   - `trackers_for_collect` would then have the aggregated data
   - Call `drain` instead of `into_iter` on trackers for collect to retain its capacity for subsequent collects

```rust
pub(crate) fn collect_and_reset<Res, MapFn>(&self, dest: &mut Vec<Res>, mut map_fn: MapFn)
    where
        MapFn: FnMut(Vec<KeyValue>, A) -> Res,
{
    prepare_data(dest, self.count.load(Ordering::SeqCst));

    ...
    
    let mut trackers = match self.trackers.write() {
        Ok(mut trackers_guard) => {
            // take(trackers_guard.deref_mut())
            let mut trackers_export = self.trackers_for_collect.write().unwrap();
            swap(trackers_guard.deref_mut(), trackers_export.deref_mut()); // swap the tracker instances
            trackers_export
        }
        Err(_) => todo!(),
    };

    for (attrs, tracker) in trackers.drain() { // calling `drain` here retains the capacity of `trackers_for_collect`
            ...
        }
    }
}
```

How does that look?

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: Resolving now. https://github.com/open-telemetry/opentelemetry-rust/pull/2267/files#r1828334335 will make it more non-trivial than before, but will come back to this.

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: resolving as this is no a blocker for this PR.

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: nit: if we go with the 2 hashmap approach, leave a code comment like "even though trackers_for_collect is RWLock guarded, only collect() thread ever need to accesses it, so the W lock on it has no impact on update() threads.".

utpilla (approved):
Great work!

There are two newer kinds of allocations happening with these changes as mentioned in https://github.com/open-telemetry/opentelemetry-rust/pull/2267#discussion_r1828280907 and https://github.com/open-telemetry/opentelemetry-rust/pull/2267#discussion_r1828334335. These can be avoided. I had a chat with @fraillt on Slack and he agreed to address these in his upcoming PRs.

opentelemetry-sdk/src/metrics/internal/histogram.rs
fraillt: Agreed to come back to this in next PR.

opentelemetry-sdk/src/metrics/internal/mod.rs
fraillt: Agreed to come back to this in next PR.

cijothomas (approved):
Thanks! Looking forward to the follow up improvements.

PR #2274, opened 2024-11-05
Simplified self-diagnostics example

cijothomas:
Simplifies self-diag example to use stdout exporter, and trigger internal log by using an invalid metric instrument name. The purpose is to demonstrate internal logging, so this seems the easiest setup to trigger one.

I kept the minimum filter example to show how to prevent the infinite, recursive logging problem. More complex examples might be needed in practice, but I think it is better instead of focus on fixing the [root cause](https://github.com/open-telemetry/opentelemetry-rust/issues/761) as priority than workarounds which are hacky. (But they are a necessary hack now)


cijothomas:
@lalitb We may actually be able to get-rid of the dedicated self-diag example, and instead demonstrate this in every examples as well.. Maybe something to iterate for next release, based on feedback, especially since we have now starting doing extensive internal logging.

examples/self-diagnostics/src/main.rs
```diff
@@ -92,41 +23,43 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
...
+    // Target name used by OpenTelemetry always start with "opentelemetry".
+    // Hence, one may use "add_directive("opentelemetry=off".parse().unwrap())"
+    // to turn off all logs from OpenTelemetry.
+
+    let filter = EnvFilter::new("info")
+        .add_directive("hyper=error".parse().unwrap())
+        .add_directive("tonic=error".parse().unwrap())
+        .add_directive("h2=error".parse().unwrap())
+        .add_directive("tower=error".parse().unwrap())
+        .add_directive("reqwest=error".parse().unwrap());
```
lalitb, line 47: we can remove this filter if none of these transport dependencies are used in the example.

lalitb (approved):
LGTM, agree on removing this example eventually.

examples/self-diagnostics/src/main.rs
cijothomas: let me move it to the otlp examples, so when users face this issue, they have something to easily refer to, to mitigate the infinite logs problem.

cijothomas: will add this shortly in a follow up PR.

Issue #2246, opened 2024-10-25
Instrumentation Scope and cloning


cijothomas:
Follow up to https://github.com/open-telemetry/opentelemetry-rust/pull/2220

I think we should remove the Cow, and make scope take owned versions. Scopes are created only very few times in the app, so there should not be perf concerns.
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/common.rs#L422

if perf is a concern, we can make Instrumentation Scope be an Arc of InstrumentationScopeInner to make cloning cheap.

cijothomas:
@stormshield-fabs I think you mentioned this in the PR but i am not able to locate that comment now! Can you help check if this cleanup is needed?

cijothomas:
tagging for 0.27 as we want to avoid breaking changes after 0.27

stormshield-fabs:
This was brought up in https://github.com/open-telemetry/opentelemetry-rust/pull/2220#discussion_r1810075412

cijothomas:
> This was brought up in [#2220 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/2220#discussion_r1810075412)

Thanks! I was searching for it, but wasn't able to locate easily!!
Would you be able to address this?

PR #2275, opened 2024-11-05
Remove Cow from InstrumentationScope

stormshield-fabs:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/2246

## Changes

Removed the `InstrumentationScope` attributes from the public API and replaced them with getters.

### DIFF
--- a/opentelemetry/src/common.rs
+++ b/opentelemetry/src/common.rs
@@ -419,18 +419,18 @@ pub struct InstrumentationScope {
     /// The library name.
     ///
     /// This should be the name of the crate providing the instrumentation.
-    pub name: Cow<'static, str>,
+    name: Cow<'static, str>,
 
     /// The library version.
-    pub version: Option<Cow<'static, str>>,
+    version: Option<Cow<'static, str>>,
 
     /// [Schema URL] used by this library.
     ///
     /// [Schema URL]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
-    pub schema_url: Option<Cow<'static, str>>,
+    schema_url: Option<Cow<'static, str>>,
 
     /// Specifies the instrumentation scope attributes to associate with emitted telemetry.
-    pub attributes: Vec<KeyValue>,
+    attributes: Vec<KeyValue>,
 }
 
 // Uniqueness for InstrumentationScope does not depend on attributes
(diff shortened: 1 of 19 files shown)


opentelemetry/src/common.rs
```diff
@@ -419,15 +419,15 @@ pub struct InstrumentationScope {
     /// The library name.
     ///
     /// This should be the name of the crate providing the instrumentation.
-    pub name: Cow<'static, str>,
+    pub name: String,
```
cijothomas, line 422: From SIG CalL
1. Make these private, and offer pub getters

opentelemetry/src/common.rs
```diff
@@ -496,7 +493,7 @@ impl InstrumentationScopeBuilder {
     ///     .with_version("v0.1.0")
     ///     .build();
     /// ```
-    pub fn with_version(mut self, version: impl Into<Cow<'static, str>>) -> Self {
+    pub fn with_version(mut self, version: impl Into<String>) -> Self {
```
cijothomas, line 496: Cow allows us to avoid alloc. Maynot be a concern since scope creation is done at most few times in an app.

cijothomas (changes requested):
From Community Call:
1. Lets keep builder as-is (i.e accept Cow.. )
2. Hide fields from InstrumentationScope and offer pub getters.

opentelemetry/src/common.rs
```diff
@@ -462,6 +462,28 @@ impl InstrumentationScope {
             attributes: None,
         }
     }
+
+    /// Instrumentation library name.
```
utpilla, line 466: nit: Use the word "return" in the documentation for all the getters.

```suggestion
    /// Returns the instrumentation library name.
```

utpilla (approved):
Looks good with a CHANGELOG entry.

opentelemetry/src/common.rs
```diff
@@ -462,6 +462,28 @@ impl InstrumentationScope {
...
+    pub fn name(&self) -> &str {
+        &self.name
+    }
```
lalitb, line 469: nit - make all these getters #[inline], just to be explicit. Even though compiler should be smart enough to do that.

opentelemetry/src/common.rs
```diff
@@ -462,6 +462,28 @@ impl InstrumentationScope {
...
+
+    /// Instrumentation library version.
+    pub fn version(&self) -> Option<&str> {
+        self.version.as_ref().map(AsRef::as_ref)
```
lalitb, line 473: `self.version.as_deref()` would be more idiomatic. Similarly for schema too.

opentelemetry/src/common.rs
```diff
@@ -462,6 +462,28 @@ impl InstrumentationScope {
...
+    ///
+    /// [Schema URL]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
+    pub fn schema_url(&self) -> Option<&str> {
+        self.schema_url.as_ref().map(AsRef::as_ref)
+    }
+
+    /// Instrumentation scope attributes to associate with emitted telemetry.
+    pub fn attributes(&self) -> &[KeyValue] {
+        &self.attributes
+    }
```
lalitb, line 486: Should we consider returning an iterator instead? This would give us flexibility to modify the internal storage later, for example, switching to non-contiguous storage or using something like [GrowableArray](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/growable_array.rs).

lalitb (approved):

cijothomas (approved):

PR #2276, opened 2024-11-05
Add force_flush to LogExporter

### LINKED ISSUE #2261: Exporter::force_flush() interface missing for Logs
The logs specification requires that the exporter implements force_flush as described here: - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#forceflush-2. This function can be asynchronous, but if so, the exporter should offer a way to inform the provider of its status.

Apart from that, the force_flush() interface are different for traces and metrics:

**Traces:**

https://github.com/open-telemetry/opentelemetry-rust/blob/a5e2061162abf6d86137f8b22b057bfb0aca9203/opentelemetry-sdk/src/export/trace.rs#L63

  
**Metrics:**

https://github.com/open-telemetry/opentelemetry-rust/blob/a5e2061162abf6d86137f8b22b057bfb0aca9203/opentelemetry-sdk/src/metrics/exporter.rs#L21

Looking closer, both interfaces are indeed asynchronous. For traces, the interface returns a `Future`, making it naturally `async`, while for metrics, it’s defined using `async fn`, which simplifies things a bit. Although the difference isn’t major, the `async fn` approach in metrics does look cleaner, more straightforward interface.

ThomsonTan:
Fixes #2261

## Changes

Added the force_flush interface to LogExporter and call it in the LogProcessor, right after flushing the export queue.


opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -91,6 +92,25 @@ pub trait LogExporter: Send + Sync + Debug {
         // By default, all logs are enabled
         true
     }
+
+    /// This is a hint to ensure that the export of any Spans the exporter
```
lalitb, line 96:
```suggestion
    /// This is a hint to ensure that the export of any logs the exporter
```

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -91,6 +92,25 @@ pub trait LogExporter: Send + Sync + Debug {
...
+    /// has received prior to the call to this function SHOULD be completed
+    /// as soon as possible, preferably before returning from this method.
+    ///
+    /// This function SHOULD provide a way to let the caller know
+    /// whether it succeeded, failed or timed out.
+    ///
+    /// This function SHOULD only be called in cases where it is absolutely necessary,
+    /// such as when using some FaaS providers that may suspend the process after
+    /// an invocation, but before the exporter exports the completed spans.
```
lalitb, line 105:
```suggestion
    /// an invocation, but before the exporter exports the completed logs.
```

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -91,6 +92,25 @@ pub trait LogExporter: Send + Sync + Debug {
...
+    ///
+    /// This function SHOULD complete or abort within some timeout. This function can be
+    /// implemented as a blocking API or an asynchronous API which notifies the caller via
+    /// a callback or an event. OpenTelemetry client authors can decide if they want to
+    /// make the flush timeout configurable.
+    fn force_flush(&mut self) -> BoxFuture<'static, ExportResult> {
```
lalitb, line 111: should we instead have `async fn force_flush(&mut self) -> ExportResult` - it's more cleaner, without boxing overhead

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -91,6 +92,25 @@ pub trait LogExporter: Send + Sync + Debug {
...
+    ///
+    /// This function SHOULD only be called in cases where it is absolutely necessary,
+    /// such as when using some FaaS providers that may suspend the process after
+    /// an invocation, but before the exporter exports the completed logs.
+    ///
+    /// This function SHOULD complete or abort within some timeout. This function can be
+    /// implemented as a blocking API or an asynchronous API which notifies the caller via
+    /// a callback or an event. OpenTelemetry client authors can decide if they want to
+    /// make the flush timeout configurable.
+    fn force_flush(&mut self) -> BoxFuture<'static, ExportResult> {
```
cijothomas, line 111: can we do normal method instead of async? 
`fn force_flush(&mut self) {}`

lalitb: I think this was discussed in community meeting, and there were suggestions to keep it async. cc @TommyCpp

lalitb: Specifically for scenario's where it is called from any async task, it will monopolizes the thread for the duration of that blocking operation, preventing other tasks to run on that thread.

cijothomas: I understand we want to offer a async version of shutdown, but not sure if there is a need for force_flush. What scenario even requires user to call flush at all?

lalitb: >  What scenario even requires user to call flush at all?

It's part of the specs, so good to provide. Application can decide when to call - say for some high critical events when application want them to logged immediately, without waiting for the periodic flush. Also the specs talks about FaaS scenairo, where serverless function (acting as application) can get's suspended when there are no request, and which will also suspend the periodic flush of batch processor. The function may want to flush everything before getting suspended.

cijothomas: makes sense!

opentelemetry-sdk/src/export/logs/mod.rs
cijothomas: +1

ThomsonTan: Updated it to be async.

lalitb:
Can we also add a unit-test for batch processor, with mock-exporter, to validate that the exporter's force flush get's called.

opentelemetry-sdk/src/export/logs/mod.rs
utpilla: > What scenario even requires user to call flush at all?

I was thinking the same. We already have `force_flush` on `LoggerProvider`. That should work for users, right? Do we really need to add it for exporters as well? Is it just because the spec mentions it?

lalitb: Just to be clear - This is not to be called by user. But by `LoggerProvider::force_flush`.

`LoggerProvider::force_flush -> Processor::force_flush -> (for batch) exporter::export() for all events -> exporter::force_flush()`

I believe specs mentions it for a reason. We don't need it for OTLP exporters because of the way they are implemented. And also not required for the synchronous exporters like - etw and user_events. But we don't know how custom exporter are implemented - the successful export for them may be storing events initially to persistent storage. And this is then periodically flushed to some backend in background. This force_flush would acts as trigger for them to complete any such pending tasks if they want.

utpilla: Good point!

hdost (approved):

lalitb (approved):

lalitb:
We may need to add both sync and async versions for provider / processors and exporter for shutdown() and force_flush(), and so different pipelines based on what is invoked. I am looking into those aspects, but this would be still required. @cijothomas @utpilla @TommyCpp  if anyone has concern here.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -91,6 +91,25 @@ pub trait LogExporter: Send + Sync + Debug {
...
+    /// as soon as possible, preferably before returning from this method.
+    ///
+    /// This function SHOULD provide a way to let the caller know
+    /// whether it succeeded, failed or timed out.
+    ///
+    /// This function SHOULD only be called in cases where it is absolutely necessary,
+    /// such as when using some FaaS providers that may suspend the process after
+    /// an invocation, but before the exporter exports the completed logs.
+    ///
+    /// This function SHOULD complete or abort within some timeout. This function can be
```
utpilla, line 106: Let's update the comments here to reflect the actual implementation instead of putting the spec guidelines for this method.

We could put something like:
```rust
/// This method will block the current thread until all pending log records are exported. If the export was not
/// successful, an error is returned.
```

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -91,6 +91,25 @@ pub trait LogExporter: Send + Sync + Debug {
...
+    /// implemented as a blocking API or an asynchronous API which notifies the caller via
+    /// a callback or an event. OpenTelemetry client authors can decide if they want to
+    /// make the flush timeout configurable.
+    fn force_flush(&mut self) -> ExportResult {
+        Ok(())
```
utpilla, line 111: I think we should not have a default implementation for this method. Otherwise, we could run into a situation where an exporter user goes through the documentation and calls `force_flush` with certain expectations that haven't been agreed upon by the exporter author.

It looks like we have a default implementation for a few other methods as well such as `shutdown` and `set_resource`. We should revisit them based on what we decide.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -278,7 +282,8 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                             &timeout_runtime,
                             logs.split_off(0),
                         )
-                        .await;
+                        .await
+                        .and(exporter.as_mut().force_flush());
```
utpilla, line 286: Do we need this? `export_with_timeout` already calls `exporter.export` which is what is expected from `force_flush`.

opentelemetry-sdk/src/export/logs/mod.rs
cijothomas: _Otherwise, we could run into a situation where an exporter user goes through the documentation and calls force_flush with certain expectations that haven't been agreed upon by the exporter author._

If the exporter author does not provide an implementation for flush, then they can document that right? Or the doc for Provider can be updated to merely state that it'll invoke flush on the processor, and the same for processor can state they will invoke the flush for exporter. 

Default implementation seems reasonable to me, as not every exporter need to do something for flush.

lalitb:
@ThomsonTan can we address the comments, and resolve the conflicts, to be eventually merged.

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan: The default `force_flush` is empty, but the user exporter could override it and do more than export, so I think we need make the call to `force_flush` here.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust/pull/2276/files#r2001245989

The processor/exporter have gone through more changes since this PR was opened. We don't need mutable self in flush. Could you adjust to the new way and rebase from main?

ThomsonTan:
> See [#2276 (files)](https://github.com/open-telemetry/opentelemetry-rust/pull/2276/files#r2001245989)
> 
> The processor/exporter have gone through more changes since this PR was opened. We don't need mutable self in flush. Could you adjust to the new way and rebase from main?

Yes, a lot of changes added after the change. Pickup the change into a new PR seems easier than merge all the conflicts.

PR #2277, opened 2024-11-05
Add few debug level internal logs to PeriodicReader

cijothomas:
Debug level, only for troubleshooting deep.
Need to find a common style like whether every Message should end with a period etc. in the future.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -132,7 +132,7 @@ where
         };
 
         otel_debug!(
-            name: "PeriodicReaderBuilt",
+            name: "PeriodicReader.BuildCompleted",
             message = "Periodic reader built.",
             interval_in_secs = self.interval.as_secs(),
         );
@@ -258,7 +258,7 @@ impl<RT: Runtime> PeriodicReaderWorker<RT> {
         match message {
             Message::Export => {
                 otel_debug!(
-                    name: "PeriodicReader.ExportMessageReceived",
+                    name: "PeriodicReader.ExportTriggered",
                     message = "Export message received.",
                 );
                 if let Err(err) = self.collect_and_export().await {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -131,6 +131,12 @@ where
             }));
         };
 
+        otel_debug!(
+            name: "PeriodicReaderBuilt",
```
utpilla, line 135: We can be consistent with the other event names and follow this format: `{componentName}.{action specified in past tense form}`

```suggestion
            name: "PeriodicReader.BuildCompleted",
```

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -251,6 +257,10 @@ impl<RT: Runtime> PeriodicReaderWorker<RT> {
     async fn process_message(&mut self, message: Message) -> bool {
         match message {
             Message::Export => {
+                otel_debug!(
+                    name: "PeriodicReader.ExportMessageReceived",
```
utpilla, line 261: `Message` and channels are implementation related terms. I think we should use more generic and intuitive terms for event names such as these:
- `PeriodicReader.ExportTriggered`
- `PeriodicReader.ForceFlushCalled`
- `PeriodicReader.ShutdownCalled`

lalitb: Not related to this comment, but just wondering if it should be fine to include implementation-specific terms in `debug` macros, as long as they're not intended for the end user?

cijothomas: I think its inevitable that we need to use implementation specific terms in some places. My only goal is to restrict them to TRACE/DEBUG level logs only.

utpilla: Let's use user-friendly names wherever possible for any log level? It would make the collected logs more readable. It would also save the effort of renaming event names, if at all we change the underlying implementation details. 

We can use implementation details in the event name, when it's indeed inevitable.

cijothomas: Yes, that's the plan.

cijothomas: This PR is just adding some debug logs, so I can see some logs without having to "trigger" an error condition to see a log, and see if self-diag example can be removed. (https://github.com/open-telemetry/opentelemetry-rust/pull/2274#issuecomment-2456179200)

lalitb (approved):

utpilla (approved):

PR #2280, opened 2024-11-06
Update documentation for Metrics API

utpilla:
## Changes
- Update documentation for Metrics API


opentelemetry/src/global/metrics.rs
```diff
@@ -11,6 +11,8 @@ static GLOBAL_METER_PROVIDER: Lazy<RwLock<GlobalMeterProvider>> =
 
 /// Sets the given [`MeterProvider`] instance as the current global meter
 /// provider.
+///
+/// **NOTE:** This function should be called before getting [`Meter`] instances via [`meter()`] or [`meter_with_scope()`]. Otherwise, you could get no-op [`Meter`] instances.
```
cijothomas, line 15: can we mention something like - do this at the earliest stage of your application/similar? Or is this sufficient?
I am also wondering what happens if a library statically creates Meter - is that going to be no-ops always, as we won't have an opportunity to set the global provider beforehand?

cijothomas (approved):

opentelemetry/src/global/metrics.rs
TommyCpp: Library author should just get meter from provider without the need to know the whether the provider is set or not. Application owner is responsible to instrument the provider before collectiong metrics.

cijothomas: Yes! This API is for application owners only (this is the set one), so we should include doc comment targeting app owners. In fact, I think we should also mention that "if you are library author, *do not* use this API.

PR #2281, opened 2024-11-06
Update code documentation for Metric instruments

utpilla:
## Changes
- Update code documentation for Metric instruments

### DIFF
--- a/opentelemetry/src/metrics/instruments/counter.rs
+++ b/opentelemetry/src/metrics/instruments/counter.rs
@@ -5,6 +5,8 @@ use std::sync::Arc;
 use super::SyncInstrument;
 
 /// An instrument that records increasing values.
+///
+/// [`Counter`] can be cloned to create multiple handles to the same instrument. Avoid creating duplicate [`Counter`]s for the same instrument.
 #[derive(Clone)]
 #[non_exhaustive]
 pub struct Counter<T>(Arc<dyn SyncInstrument<T> + Send + Sync>);
@@ -31,6 +33,8 @@ impl<T> Counter<T> {
 }
 
 /// An async instrument that records increasing values.
+///
+/// [`ObservableCounter`] can be cloned to create multiple handles to the same instrument. Avoid creating duplicate [`ObservableCounter`]s for the same instrument.
 #[derive(Clone)]
 #[non_exhaustive]
 pub struct ObservableCounter<T> {
--- a/opentelemetry/src/metrics/instruments/counter.rs
+++ b/opentelemetry/src/metrics/instruments/counter.rs
@@ -33,8 +33,6 @@ impl<T> Counter<T> {
 }
 
 /// An async instrument that records increasing values.
-///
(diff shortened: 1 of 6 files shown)


opentelemetry/src/metrics/instruments/counter.rs
```diff
@@ -31,6 +33,8 @@ impl<T> Counter<T> {
 }
 
 /// An async instrument that records increasing values.
+///
+/// [`ObservableCounter`] can be cloned to create multiple handles to the same instrument. Avoid creating duplicate [`ObservableCounter`]s for the same instrument.
```
cijothomas, line 37: given callbacks are already specified at creation time, is there a use case for cloning Observable instruments today?

opentelemetry/src/metrics/instruments/counter.rs
```diff
@@ -5,6 +5,8 @@ use std::sync::Arc;
 use super::SyncInstrument;
 
 /// An instrument that records increasing values.
+///
+/// [`Counter`] can be cloned to create multiple handles to the same instrument. Avoid creating duplicate [`Counter`]s for the same instrument.
```
cijothomas, line 9:
```suggestion
/// [`Counter`] can be cloned to create multiple handles to the same instrument. Avoid creating duplicate [`Counter`]s for the same metric.
```

Is this better?
Also, should we put a stronger statement about re-using this struct, instead of repeatedily created it over and over?

opentelemetry/src/metrics/instruments/counter.rs
utpilla: Good catch 😄

I'll remove this from the asynchronous instruments.

opentelemetry/src/metrics/instruments/counter.rs
utpilla: > Is this better?

Yes. I'll update them to use the term "metric".

> Also, should we put a stronger statement about re-using this struct, instead of repeatedily created it over and over?

We should do that if it's super bad for perf.  But I was thinking since we cache the aggregators and there is no increased contention for creating duplicate instruments (unlike .NET), this level of warning is enough?

cijothomas: its best to have that warning now itself, so that even if a future implementation change causes slower perf when re-creating existing metric, we have it covered by doc warnings.

utpilla: Okay. How does this look?

```text
/// [`Counter`] can be cloned to create multiple handles to the same instrument. If a [`Counter`] needs to be shared, 
users are recommended to clone the [`Counter`] instead of creating duplicate [`Counter`]s for the same metric. Creating 
duplicate [`Counter`]s for the same metric could lower SDK performance.
```

cijothomas: its good 👍

cijothomas (approved):

PR #2283, opened 2024-11-07
Histogram implementation cleanup

fraillt:
## Changes

Mostly refactoring by removing unnecessary code, with few minor improvements:
* remove `.clone()` on buckets in `delta` collection.
* in Histogram initialization first remove NaNs from bounds, and then get it's length.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/histogram.rs
@@ -9,11 +9,7 @@ use opentelemetry::KeyValue;
 use super::ValueMap;
 use super::{Aggregator, Number};
 
-struct HistogramTracker<T> {
-    buckets: Mutex<Buckets<T>>,
-}
-
-impl<T> Aggregator for HistogramTracker<T>
+impl<T> Aggregator for Mutex<Buckets<T>>
 where
     T: Number,
 {
@@ -22,27 +18,26 @@ where
     type PreComputedValue = (T, usize);
 
     fn update(&self, (value, index): (T, usize)) {
-        let mut buckets = match self.buckets.lock() {
-            Ok(guard) => guard,
-            Err(_) => return,
-        };
+        let mut buckets = self.lock().unwrap_or_else(|err| err.into_inner());
 
-        buckets.bin(index, value);
-        buckets.sum(value);
+        buckets.total += value;
+        buckets.count += 1;
+        buckets.counts[index] += 1;
+        if value < buckets.min {
+            buckets.min = value;
+        }
+        if value > buckets.max {
+            buckets.max = value
+        }
     }
 
     fn create(count: &usize) -> Self {
-        HistogramTracker {
-            buckets: Mutex::new(Buckets::<T>::new(*count)),
-        }
+        Mutex::new(Buckets::<T>::new(*count))
     }
 
     fn clone_and_reset(&self, count: &usize) -> Self {
-        let mut current = self.buckets.lock().unwrap_or_else(|err| err.into_inner());
-        let cloned = replace(current.deref_mut(), Buckets::new(*count));
-        Self {
-            buckets: Mutex::new(cloned),
-        }
+        let mut current = self.lock().unwrap_or_else(|err| err.into_inner());
+        Mutex::new(replace(current.deref_mut(), Buckets::new(*count)))
     }
 }
 
@@ -65,27 +60,12 @@ impl<T: Number> Buckets<T> {
             ..Default::default()
         }
     }
-
-    fn sum(&mut self, value: T) {
-        self.total += value;
-    }
-
-    fn bin(&mut self, idx: usize, value: T) {
-        self.counts[idx] += 1;
-        self.count += 1;
-        if value < self.min {
-            self.min = value;
-        }
-        if value > self.max {
-            self.max = value
-        }
-    }
 }
 
 /// Summarizes a set of measurements as a histogram with explicitly defined
 /// buckets.
 pub(crate) struct Histogram<T: Number> {
-    value_map: ValueMap<HistogramTracker<T>>,
+    value_map: ValueMap<Mutex<Buckets<T>>>,
     bounds: Vec<f64>,
     record_min_max: bool,
     record_sum: bool,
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -22,27 +18,26 @@ where
     type PreComputedValue = (T, usize);
 
     fn update(&self, (value, index): (T, usize)) {
-        let mut buckets = match self.buckets.lock() {
-            Ok(guard) => guard,
-            Err(_) => return,
-        };
+        let mut buckets = self.lock().unwrap_or_else(|err| err.into_inner());
```
cijothomas, line 21: perhaps a good idea to start using internal logs for these errors?

cijothomas: just to confirm I understand this right:
These locks are taken by user-threads when calling histogram.Record(..).
But the only code executed within the lock is these 4 lines to update min/max/sum/count, which has no chance of panic.
And due to that, we don't need to worry about poisoning here.

The other place where this lock is taken is the collect() thread, but that is only executing a mem-replace, with no possibility of panic.

And because of this, we don't ever expect a lock poison here?

Is there a possibility that the user thread after acquiring the lock gets somehow killed? I am not saying we should account for that here!

opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -65,62 +60,34 @@ impl<T: Number> Buckets<T> {
...
 }
 
 impl<T: Number> Histogram<T> {
-    pub(crate) fn new(boundaries: Vec<f64>, record_min_max: bool, record_sum: bool) -> Self {
-        // TODO fix the bug, by first removing NaN and only then getting buckets_count
-        // once we know the reason for performance degradation
-        let buckets_count = boundaries.len() + 1;
-        let mut histogram = Histogram {
+    pub(crate) fn new(mut bounds: Vec<f64>, record_min_max: bool, record_sum: bool) -> Self {
+        bounds.retain(|v| !v.is_nan());
```
cijothomas, line 77: There are only 3 ways bounds can come from:
1. Default.  - In this case, sdk provides it so we ensure its already sorted and validated.
2. HistogramCreation API - In this case, user provides it, but while creating the instrument, sdk *should* validate it, and ensure its sorted/validated. If the validation fails, sdk can return the defaults or noop instrument. Opened this issue to track it: https://github.com/open-telemetry/opentelemetry-rust/issues/2286
3. Views - In this case also, user provides it, but we should do same as 2.

I believe this method is the wrong place to validate/filter bounds as it should be taken care elsewhere.

Not a blocker for this PR, just added a comment to make sure have same understanding.

cijothomas (approved):
LGTM! 
Left some comments but not a blocker at all.

opentelemetry-sdk/src/metrics/internal/histogram.rs
utpilla: > I believe this method is the wrong place to validate/filter bounds as it should be taken care elsewhere.

💯

opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -245,68 +209,25 @@ impl<T: Number> Histogram<T> {
...
-//             assert_eq!(test.expected.max, res.max);
-//             assert_eq!(test.expected.min, res.min);
-//             assert_eq!(test.expected.sum, res.total);
-//             assert_eq!(test.expected.count, res.count);
-//         }
-//     }
-// }
+    #[test]
+    fn check_buckets_are_selected_correctly() {
+        let hist = Histogram::<i64>::new(vec![1.0, 3.0, 6.0], false, false);
```
utpilla, line 218: We don't need this test.

We already have tests for checking `Histogram` with custom bounds: https://github.com/open-telemetry/opentelemetry-rust/blob/52f6d9214eca58c32e3fc41e3829038b868c07e1/opentelemetry-sdk/src/metrics/mod.rs#L319-L325

opentelemetry-sdk/src/metrics/internal/histogram.rs
utpilla: > But the only code executed within the lock is these 4 lines to update min/max/sum/count, which has no chance of panic.
And due to that, we don't need to worry about poisoning here.

Updating sum or count could lead to overflows. Even though vverflows would cause a panic only in debug build, I think we should update these operations to always use `wrapping_add`.

> The other place where this lock is taken is the collect() thread, but that is only executing a mem-replace, with no possibility of panic.

Currently, the replace involves creating a new set of buckets which needs allocation. The worst case is we run into OOM doing that but if that happens, we have bigger problems than lock poisoning 😄

> Is there a possibility that the user thread after acquiring the lock gets somehow killed?

I'm not sure if that is possible.

cijothomas: > Currently, the replace involves creating a new set of buckets which needs allocation. The worst case is we run into OOM doing that but if that happens, we have bigger problems than lock poisoning 

Once we fix the logic to use pre-allocated one, this concern should be mitigated right?

I'll resolve this comment to not block.

utpilla: Yes, it should be fixable.

Using `wrapping_add` instead of regular arithmetic operations can be done outside of this PR.

fraillt: > Is there a possibility that the user thread after acquiring the lock gets somehow killed? I am not saying we should account for that here!

AFAIK this is not possible.

fraillt: > Yes, it should be fixable.
> 
> Using `wrapping_add` instead of regular arithmetic operations can be done outside of this PR.

Next PR is coming soon, which will make that 'clone_and_reset' will be used only by no_attributes_tracker.

Regarding wrapping_add, its nice idea:)

opentelemetry-sdk/src/metrics/internal/histogram.rs
fraillt: Thanks for pointing that out.
would you mind if I move it from 'mod.rs' to 'histogram.rs' (this file) in this PR?

utpilla: > those are more like "e2e" tests or integration tests, and is best kept in mod.rs itself.

I agree with that too.

@fraillt Up to you if you want to update the existing test or keep this one here. I'm going to resolve this for now.

utpilla (approved):

PR #2284, opened 2024-11-07
perf: Baggage insert & BaggageExt::with_baggage & updated constants to latest standard

fraillt:
## Changes

Improve few things:
*  `BaggageExt::with_baggage` now "clone" baggage, instead of iterate+collect.
* `Baggage::insert_with_metadata` (used everywhere) should be a bit faster, as it had redundant calculations:
  * access hashmap once (before it was twice)
  * calculate entry content length once (before it was three times, although it's so trivial that probably it's not noticeable anyway)
* updated incorrect comment on `Baggage::iter`

### DIFF
--- a/opentelemetry/src/baggage.rs
+++ b/opentelemetry/src/baggage.rs
@@ -566,4 +566,22 @@ mod tests {
         assert!(b.to_string().contains("bar=2;yellow"));
         assert!(b.to_string().contains("foo=1;red;state=on"));
     }
+
+    #[test]
+    fn replace_existing_key() {
+        let half_minus2: StringValue = (0..MAX_LEN_OF_ALL_PAIRS / 2 - 2)
+            .map(|_| 'x')
+            .collect::<String>()
+            .into();
+
+        let mut b = Baggage::default();
+        b.insert("a", half_minus2.clone()); // +1 for key
+        b.insert("b", half_minus2); // +1 for key
+        b.insert("c", StringValue::from(".")); // total of 2 bytes
+        assert!(b.get("a").is_some());
+        assert!(b.get("b").is_some());
+        assert!(b.get("c").is_some());
+        assert!(b.insert("c", StringValue::from("..")).is_none()); // exceeds MAX_LEN_OF_ALL_PAIRS
+        assert_eq!(b.insert("c", StringValue::from("!")).unwrap(), ".".into()); // replaces existing
+    }
 }
--- a/opentelemetry/src/baggage.rs
+++ b/opentelemetry/src/baggage.rs
@@ -15,14 +15,14 @@
 //!
 //! [W3C Baggage]: https://w3c.github.io/baggage
 use crate::{Context, Key, KeyValue, StringValue};
+use std::collections::hash_map::Entry;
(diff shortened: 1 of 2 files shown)


opentelemetry/src/baggage.rs
```diff
@@ -146,10 +149,41 @@ impl Baggage {
         S: Into<BaggageMetadata>,
     {
         let (key, value, metadata) = (key.into(), value.into(), metadata.into());
-        if self.insertable(&key, &value, &metadata) {
-            self.inner.insert(key, (value, metadata))
-        } else {
-            None
+        if !key.as_str().is_ascii() {
+            return None;
+        }
```
lalitb, line 154: I understand this is the existing code, but baggage specs only allow for the printable ASCII characters. Probably, good to also check for control characters.

opentelemetry/src/baggage.rs
```diff
@@ -146,10 +149,41 @@ impl Baggage {
...
+                );
+                let new_content_len = self.kv_content_len + entry_content_len - prev_content_len;
+                if new_content_len > MAX_LEN_OF_ALL_PAIRS {
+                    return None;
+                }
+                self.kv_content_len = new_content_len;
+                Some(occupied_entry.insert((value, metadata)))
+            }
+            Entry::Vacant(vacant_entry) => {
+                if entries_count == MAX_KEY_VALUE_PAIRS {
```
lalitb, line 176: Though this is existing code, the limit is 64 as per the specs, while we use 180 - good to update that.

opentelemetry/src/baggage.rs
fraillt: I'm not sure how to fix that in efficient way, so I would prefer to leave this to someone more competent.

opentelemetry/src/baggage.rs
fraillt: Ok, I can updated that, also I noticed that MAX_BYTES_FOR_ONE_PAIR is also removed form spec, so I'll remove this as well.

opentelemetry/src/baggage.rs
```diff
@@ -146,10 +147,38 @@ impl Baggage {
...
         let (key, value, metadata) = (key.into(), value.into(), metadata.into());
-        if self.insertable(&key, &value, &metadata) {
-            self.inner.insert(key, (value, metadata))
-        } else {
-            None
+        if !key.as_str().is_ascii() {
+            return None;
+        }
+        let entry_content_len =
+            key_value_metadata_bytes_size(key.as_str(), value.as_str().as_ref(), metadata.as_str());
```
cijothomas, line 154: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/baggage.rs#L59

Interesting..We accept Value for Baggage value, but the [spec only allows](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#overview) strings!
**_Baggage values are any valid UTF-8 strings_**

@fraillt should we fix the API to accept OTelString only instead of Value?

cijothomas (approved):

PR #2285, opened 2024-11-07
Remove Counter/Add benchmark for delta temporality

fraillt:
## Changes

Whether it's Cumulative or Delta temporality, there's no difference for in measurement stage, so I removed redudant benchmarks.


cijothomas (approved):
Okay with this cleanup, but i think we should still maintain a stress test+benchmark for delta+cumulative. The perf maybe same, but a benchmark is what users can use to validate that they are indeed same.

(There are scenarios where perf maybe diff. for delta and cumulative even in the update path)

PR #2289, opened 2024-11-08
Move Temporality enum opentelemetry_sdk::metrics::data:: to opentelemetry_sdk::metrics::

cijothomas:
Earlier this Enum was used mostly by exporter authors, but this is now used by end users when setting up `MeterProvider`, so it makes to move it to metrics module directly.


lalitb (approved):

examples/metrics-basic/src/main.rs
```diff
@@ -8,7 +8,7 @@ use std::vec;
 fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
     let exporter = opentelemetry_stdout::MetricExporterBuilder::default()
         // Build exporter using Delta Temporality (Defaults to Temporality::Cumulative)
-        // .with_temporality(data::Temporality::Delta)
+        // .with_temporality(Temporality::Delta)
```
utpilla, line 11: Should we use the fully qualified name here? That way, it would still work if the user were to uncomment this line. Otherwise, with just `Temporality::Delta`, they would have to import `opentelemetry_sdk::metrics::Temporality` after they uncomment this line.

```suggestion
        // .with_temporality(opentelemetry_sdk::metrics::Temporality::Delta)
```

utpilla (approved):
Left a suggestion regarding the user experience when using the example code. Not a blocker.

PR #2291, opened 2024-11-08
Rename `logs_level_enabled`  flag to `spec_unstable_logs_enabled`

lalitb:
## Changes
The flag was added when this feature was not there in specs, so named as `logs_level_enabled`. However, now the feature is there in specs under Dev state, so renamed it to `spec_unstable_logs_enabled`.


opentelemetry-appender-log/Cargo.toml
```diff
@@ -21,14 +21,14 @@ opentelemetry-semantic-conventions = { path = "../opentelemetry-semantic-convent
 ] }
 
 [features]
-logs_level_enabled = ["opentelemetry/logs_level_enabled"]
+experimental_logs_level_enabled = ["opentelemetry/experimental_logs_level_enabled"]
```
cijothomas, line 24: https://github.com/open-telemetry/opentelemetry-rust/pull/2291/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R156 Based on the wording here, the experimental spec based things should be feature flagged under `otel_unstable` ? Do we want to follow that approach or modify it?

lalitb: Yes, I thought using `otel_unstable`. But I think not good to use single feature flag for all the features. Will revisit the wording in separate PR.

lalitb: How about:

`experimental_<flag-name>`:

- Use for features that are NOT in any specification
- These are purely experimental/exploratory features
- When the feature's implementation becomes stable, rename to  `<flag-name>`


`unstable_<flag-name>`:

- Use for features that ARE in specifications but either:
   - The spec itself is still experimental, or
   - The spec is stable but our implementation is incomplete
- Remove this feature flag entirely once the feature is fully implemented and stable

lalitb: renamed to `unstable_logs_level_enabled` based on above comment.

cijothomas: the name logs_level_enabled is not very clear. enabled() takes more than level, so why specially mention level?
maybe `unstable_logs_enabled` is good enough.?

Also, I am wondering if we need separate flags for each feature or a single flag for all unstable spec ones? The previous discussion is here https://github.com/open-telemetry/opentelemetry-rust/pull/1410#discussion_r1414694845

lalitb: > maybe unstable_logs_enabled is good enough.?

Agree `unstable_logs_enabled` seems better

> Also, I am wondering if we need separate flags for each feature or a single flag for all unstable spec ones?

Yes, the agreement was to have `otel_unstable` as a single feature flag to enable/disable all features for which
 - specs is still experimental,
 - stable implementation as per the current specs (i.e, won't crash/panic, no perf issues)
 - can have breaking changes if the specs changes.
 
which means that it would be safe to enable/disable such features together without affecting the other existing performance. logs_level_enabled does fall in this category.

lalitb: Thinking more, I feel [this](https://github.com/open-telemetry/opentelemetry-rust/pull/2291#discussion_r1835170027) brings better clarity, with different flags for each feature. I can update the CONTRIBUTION.md guide is we agree.

cijothomas (approved):
Thanks!

PR #2292, opened 2024-11-08
Add PeriodicReaderWithOwnThread

cijothomas:
Originally part of https://github.com/open-telemetry/opentelemetry-rust/pull/2142/files, but this one leaves existing PeriodicReader as-is, and instead adds a new one `PeriodicReaderWithOwnThread`. While implementing 2142, we discovered several issues related to async runtimes + exporters requiring async runtimes(OTLPExporter) that will take some time to resolve. Hence, this is a separate reader, currently useable with stdoutexporter, etw or user-events exporter, as these exporters don't rely on any particular runtimes. Once the stability is verified, will modify this to allow usage with OTLP Exporters and eventually get merged with PeriodicReader itself.


examples/metrics-basic/src/main.rs
```diff
@@ -24,24 +33,16 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+tracing_subscriber::registry()
```
cijothomas, line 36: modified example just to demo the working on new reader. These changes will be reverted before merging, this is only to help anyone wanting to quickly test.

cijothomas: this is removed. Pick the changes from the first commit if one wants to try it out.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -50,6 +50,7 @@ pub(crate) mod meter;
 mod meter_provider;
 pub(crate) mod noop;
 pub(crate) mod periodic_reader;
+pub(crate) mod periodic_reader_with_own_thread;
```
lalitb, line 53: Would it make sense to add this behind a feature flag to indicate to end-users that it’s still experimental ?

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,796 @@
...
+            otel_debug!(name: "NoMetricsCollected");
+            return Ok(());
+        }
+
+        // TODO: substract the time taken for collect from the timeout. collect
+        // involves observable callbacks too, which are user defined and can
+        // take arbitrary time.
+        //
+        // Relying on futures executor to execute async call.
+        // TODO: Add timeout and pass it to exporter.
```
lalitb, line 358: We previously discussed this approach, though I don’t fully recall if we agreed on it or if we’re open to either option:
- Passing the timeout to the exporter
- Using manual polling within a loop, checking for timeout in each iteration.
If we're still considering both options, it might be best to add a TODO comment for clarity."

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,796 @@
...
+                    event_name = "PeriodReaderForceFlushError",
+                    error = format!("{:?}", e)
+                );
+                return Err(MetricError::Other(e.to_string()));
+            }
+        }
+
+        if let Ok(response) = response_rx.recv() {
+            if response {
+                Ok(())
```
lalitb, line 396: Add a TODO to also call the exporter::force_flush once ollect-export is done?

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,796 @@
...
+/// # {
+///
+/// let exporter = get_exporter(); // set up a push exporter
+///
+/// let reader = PeriodicReaderWithOwnThread::builder(exporter).build();
+/// # drop(reader);
+/// # }
+/// ```
+#[derive(Clone)]
+pub struct PeriodicReaderWithOwnThread {
```
utpilla, line 142: I think we should use a different name. Something more intuitive to the user (component author) like: `PeriodicReaderWithBlockingExporter` or `PeriodicReaderWithSynchronousExporter` or `PeriodicReaderWithNoRuntime`

cijothomas: As mentioned in the PR desc, this will be merged into PeriodicReader itself for stable, so don't have to worry about a perfect name.
I'll move this under a feature flag as suggested here: https://github.com/open-telemetry/opentelemetry-rust/pull/2292#discussion_r1835184394

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
cijothomas: added

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: added

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,798 @@
...
+                                otel_debug!(
+                                    name: "PeriodReaderThreadExportTookLongerThanInterval",
+                                    event_name = "PeriodReaderThreadExportTookLongerThanInterval"
+                                );
+                                // if export took longer than interval, do the
+                                // next export immediately.
+                                // Alternatively, we could skip the next export
+                                // and wait for the next interval.
+                                // Or enforce that export timeout is less than interval.
+                                // What is the desired behavior?
```
lalitb, line 253: Perhaps hybrid approach can be one option - enforce  that the export `timeout` is configured to be less than the `interval`. And then, if the delay is slightly more than the interval, we can do next export immediately. However, if the delay is in multiple of intervals, it signifies heavy load on system, and better to skip next cycle - Just a thought, but good to discuss for further PRs.

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,798 @@
...
+                i_clone.fetch_add(1, Ordering::Relaxed);
+            })
+            .build();
+
+        // Sleep for a duration 5X (plus liberal buffer to account for potential
+        // CI slowness) the interval to ensure multiple collection.
+        // Not a fan of such tests, but this seems to be the only way to test
+        // if periodic reader is doing its job.
+        // TODO: Decide if this should be ignored in CI
+        std::thread::sleep(interval * 5 * 20);
```
lalitb, line 577: yes sleep will slow down the CI. Instead of sleep, we can have a  loop that periodically checks this condition as below (not syntactically correct):
```
while i.load(Ordering::Relaxed) < 5 && elapsed_time < 5*20 ms) 
{
}
```

Or better to skip this test for the CI. Not a blocker for this PR.

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,798 @@
...
+        collection_triggered_by_interval_helper();
+        collection_triggered_by_flush_helper();
+        collection_triggered_by_shutdown_helper();
+    }
+
+    #[tokio::test(flavor = "current_thread")]
+    async fn collection_from_tokio_current() {
+        collection_triggered_by_interval_helper();
+        collection_triggered_by_flush_helper();
+        collection_triggered_by_shutdown_helper();
```
lalitb, line 738: nit - These 3 helper methods can be wrapped in a single method and called in all test cases to reduce redundancy

opentelemetry-sdk/src/metrics/periodic_reader_with_own_thread.rs
```diff
@@ -0,0 +1,798 @@
...
+    }
+
+    fn collection_triggered_by_interval_helper() {
+        collection_helper(|_| {
+            // Sleep for a duration longer than the interval to ensure at least one collection
+            // Not a fan of such tests, but this seems to be the only way to test
+            // if periodic reader is doing its job.
+            // TODO: Decide if this should be ignored in CI
```
lalitb, line 746: Either skip in CI, or conditional polling. Not a blocker here though - for separate PR.

lalitb (approved):
LGTM with nit comments. As this is under feature flag, it can be improved further after release. Thanks for working on it.

utpilla (approved):
Looks good overall!

cijothomas:
Merging to unblock release. This component is under experimental ff, and no impact on any existing code.

PR #2304, opened 2024-11-17
DRAFT: metrics sharding handwritten

fraillt:
As we discussed in our meeting, I'm creating a draft version with implemented sharding:
It mainly consist of these steps:
* Took #2288 as a base, to be able to implement `collect_and_reset`
* Merged #2296 , to be able to avoid recalculations of hash
* Used `rustc-hash` as hashing function (seam to be faster then `ahash` crate, on my machine)
* Implemented sharding by hand
* added few tweeks inspired by `dashmap`. 

Could probably be using `dashmap` directly as well, but it's quite trivial to write it our own, and we can avoid having extra dependency :)

Waiting for your test results :)
Would be good to know CPU and rust version.

### DIFF
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -1,5 +1,4 @@
-use opentelemetry::global;
-use opentelemetry::KeyValue;
+use opentelemetry::{global, KeyValue};
 use opentelemetry_sdk::metrics::{PeriodicReader, SdkMeterProvider};
 use opentelemetry_sdk::{runtime, Resource};
 use std::error::Error;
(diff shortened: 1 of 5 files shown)


fraillt:
I did some latency testing using #2323.
Wanted to compare it to #2305.
Flurry concurrent hashmap performs pretty well. However, it LOOSES SOMES UPDATES! So if want to move forward, we definitely need to fix this first, but for testing purposes I just ignored this "little" issue :)
The observation is as follows:
* for updates this PR performs a bit worse (compared to flurry), especially when it comes to p99 (1.5x slower on low load) (although on high loads this PR comes close)
* for inserts it performs much better (almost 3-8x better, especially at p99).
* for mixed loads (mostly updates with few inserts) are mixes as well: on low loads this PR wins, on high loads flurry takes over).

cijothomas:
Closing. This is linked to from the tracking issue to revisit later: https://github.com/open-telemetry/opentelemetry-rust/issues/2450

PR #2305, opened 2024-11-18
[Perf][Metrics] Use flurry's concurrent hashmap for 5x throughput

utpilla:
Towards #1740

## Changes
- Current implementation of Metrics SDK uses a `RwLock<Hashmap>` for aggregating measurements. `RwLock` brings in significant amount of contention even for concurrent reads
- This PR showcases how we could use [`flurry`](https://crates.io/crates/flurry) crate's concurrent Hashmap instead of a `RwLock<Hashmap>`.

The performance gains are huge!

#### Machine details:
RAM: 64.0 GB

- Counter throughput went up from **9M to 45M** iterations per sec
- Histogram throughput went up from **7.5 M to 35M** iterations per sec

## Benchmarks
- The benchmark results are comparable. There is no significant difference in the benchmark results.

## Note for reviewers

This PR is not meant for merging as-is. It's meant to show that we can utilize a more efficient concurrent data structure to our advantage. If we indeed decide to use `flurry`'s Hashmap, we have to address the following:

1. There is a potential race condition during collect for Delta temporality. This needs to be fixed to avoid losing measurements.
2. Provide meaningful `Ord` implementations for `KeyValue`. `Ord` implementation for the Hashmap's key type is a requirement from `flurry`. For this PR, I have added a very basic implementation just to unblock myself from testing the Hashmap.
3. We need to offer this under a feature flag. We don't want to add dependency on an external crate by default.

### DIFF
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -29,6 +29,7 @@ tokio = { workspace = true, features = ["rt", "time"], optional = true }
 tokio-stream = { workspace = true, optional = true }
 http = { workspace = true, optional = true }
 tracing = {workspace = true, optional = true}
+flurry = "0.5.1"
 
 [package.metadata.docs.rs]
 all-features = true
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -6,11 +6,10 @@ mod precomputed_sum;
 mod sum;
 
 use core::fmt;
-use std::collections::{HashMap, HashSet};
-use std::mem::take;
-use std::ops::{Add, AddAssign, DerefMut, Sub};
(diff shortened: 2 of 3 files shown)


lalitb:
Thank you for sharing the results! The performance improvements are impressive, as anticipated, given that `Flurry` is designed after Java's ConcurrentHashMap and provides more fine-grained concurrency compared to `DashMap`. This can be added under the feature flag, but we also need to see some concern raised on its stability on high-load - https://github.com/jonhoo/flurry/issues/127:

``
flurry has been a very interesting concurrent hash table experiment and has driven a lot of innovation in that space. However, at this point, flurry suffers from performance as well as https://github.com/jonhoo/flurry/issues/115 issues under load.
``

cijothomas:
Closing. This is linked to from the tracking issue to revisit later: https://github.com/open-telemetry/opentelemetry-rust/issues/2450

PR #2306, opened 2024-11-18
Make stress tests independent of batch size

utpilla:
### Current implementation:
- The current implementation of stress test reports the progress made in batches instead of reporting it on the fly.
- Each thread executes the function to be stress tested `1000` times (`BATCH_SIZE` is set to `1000`) and then updates its stats by updating the `AtomicU64` value in `WorkerStats` vec.
- The motivation to report the progress in batches is mainly to lower the number of `fetch_add` atomic operations so that it does not interfere with the measurements reported for the function under test.

### Motivation for this PR
- Reporting measurements in batches makes the stress test results dependent on the `BATCH_SIZE`. You have to choose an optimal `BATCH_SIZE` for the function being stress tested. If the `BATCH_SIZE` is too low, then you make a lot more `fetch_add` calls for `WorkerStats` which would influence the final throughput. If the `BATCH_SIZE` is too high, then you don't report the actual progress made in a timely manner.
- If the whole idea, is to not have the stress test's own bookkeeping affect the throughput results, then we can consider avoiding making `fetch_add` calls altogether.
- This PR makes use of some `unsafe` code to have the update threads report progress by making a simple `+` operation instead of the atomic `fetch_add` operation.
- Another change in the PR is to use scoped threads which simplifies things a bit by not having to join on the update thread handles.

### Note for reviewers:
Please hide whitespace from the diff for easier reviewing:

image

### DIFF
--- a/stress/src/throughput.rs
+++ b/stress/src/throughput.rs
@@ -1,6 +1,7 @@
 use num_format::{Locale, ToFormattedString};
+use std::cell::UnsafeCell;
 use std::env;
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::thread;
 use std::time::{Duration, Instant};
@@ -8,14 +9,13 @@ use std::time::{Duration, Instant};
 use sysinfo::{Pid, System};
 
 const SLIDING_WINDOW_SIZE: u64 = 2; // In seconds
-const BATCH_SIZE: u64 = 1000;
 
 static STOP: AtomicBool = AtomicBool::new(false);
 
 #[repr(C)]
 #[derive(Default)]
 struct WorkerStats {
-    count: AtomicU64,
+    count: u64,
     /// We use a padding for the struct to allow each thread to have exclusive access to each WorkerStat
     /// Otherwise, there would be some cpu contention with threads needing to take ownership of the cache lines
     padding: [u64; 15],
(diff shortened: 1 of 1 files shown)


stress/src/throughput.rs
```diff
@@ -58,91 +58,111 @@ where
...
 
-            thread::sleep(Duration::from_millis(5000));
+        for thread_index in 0..num_threads {
+            let func_arc_clone = Arc::clone(&func_arc);
+            s.spawn(move || loop {
+                func_arc_clone();
+                unsafe {
+                    shared_mutable_stats_slice.increment(thread_index);
+                }
+                if STOP.load(Ordering::SeqCst) {
```
cijothomas, line 130: Won't we be now forced to check this AtomicBool for each iteration? I think that still be avoided and instead can be checked every BATCH_SIZE?

stress/src/throughput.rs
```diff
@@ -58,91 +58,111 @@ where
...
+            #[cfg(feature = "stats")]
+            let pid = Pid::from(std::process::id() as usize);
+            #[cfg(feature = "stats")]
+            let mut system = System::new_all();
+
+            loop {
+                let current_time = Instant::now();
+                let elapsed = current_time.duration_since(last_collect_time).as_secs();
+                if elapsed >= SLIDING_WINDOW_SIZE {
+                    let total_count_u64 = shared_mutable_stats_slice.sum();
```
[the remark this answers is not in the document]
cijothomas: at worse, we'll underreport the numbers. ?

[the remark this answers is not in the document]
utpilla: At worst, we won't get the most up-to-date sum which is okay?

stress/src/throughput.rs
utpilla: Since the `STOP` is only ever changed when exiting the stress test, the cache line where `STOP` resides wouldn't have to be updated during the test. So, I don't expect it to have much of a perf implication. (don't think it suffers from false sharing here as `STOP` is `static`).

If we still want to avoid it, I'd prefer if remove the `STOP` variable altogether and simply exit the process on `Ctrl + C`. The existing code for using `STOP` doesn't add much value.

utpilla: If we still want to avoid it, I'd prefer if remove the `STOP` variable altogether and simply exit the process on `Ctrl + C`.

I ran the stress test with the above-mentioned change, and there isn't much of a difference in the results. I would prefer removing it as the existing code for using `STOP` doesn't add much value.

cijothomas: not a blocker for this PR, lets revisit removing it alltogether in a future pr, if needed.

cijothomas (approved):

cijothomas:
_perf numbers from my laptop (	11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz	4 cores with 48 GB RAM)_

_tracing with no subscriber                                       4B iterations/sec   (.NET is ~35  M/sec)_


^ Something I found in my personal note, a long time ago. 4B /sec was with `tracing` + no subscriber, which is pretty close to empty fn()!

PR #2309, opened 2024-11-20
feat: update otel versions for prometheus to 0.27

kudlatyamroth:
## Changes

- Update `opentelemetry` dependency version to 0.27
- Update `opentelemetry_sdk` dependency version to 0.27
- Update `opentelemetry-semantic-conventions` dependency version to 0.27

### DIFF
--- a/opentelemetry-prometheus/Cargo.toml
+++ b/opentelemetry-prometheus/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "opentelemetry-prometheus"
-version = "0.17.0"
+version = "0.27.0"
 description = "Prometheus exporter for OpenTelemetry"
 homepage = "https://github.com/open-telemetry/opentelemetry-rust"
 repository = "https://github.com/open-telemetry/opentelemetry-rust"
@@ -21,17 +21,20 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 once_cell = { workspace = true }
-opentelemetry = { version = "0.26", default-features = false, features = ["metrics"] }
-opentelemetry_sdk = { version = "0.26", default-features = false, features = ["metrics"] }
+opentelemetry = { version = "0.27", default-features = false, features = ["metrics"] }
+opentelemetry_sdk = { version = "0.27", default-features = false, features = ["metrics", "spec_unstable_metrics_views"] }
 prometheus = "0.13"
 protobuf = "2.14"
+tracing = {workspace = true, optional = true} # optional for opentelemetry internal logging
 
 [dev-dependencies]
-opentelemetry-semantic-conventions = { version = "0.26" }
+opentelemetry-semantic-conventions = { version = "0.27" }
 http-body-util = { workspace = true }
 hyper = { workspace = true, features = ["full"] }
 hyper-util = { workspace = true, features = ["full"] }
 tokio = { workspace = true, features = ["full"] }
 
 [features]
+default = ["internal-logs"]
 prometheus-encoding = []
+internal-logs = ["tracing"]
--- a/opentelemetry-prometheus/Cargo.toml
+++ b/opentelemetry-prometheus/Cargo.toml
@@ -22,7 +22,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 once_cell = { workspace = true }
 opentelemetry = { version = "0.27", default-features = false, features = ["metrics"] }
-opentelemetry_sdk = { version = "0.27", default-features = false, features = ["metrics", "spec_unstable_metrics_views"] }
+opentelemetry_sdk = { version = "0.27", default-features = false, features = ["metrics"] }
 prometheus = "0.13"
 protobuf = "2.14"
 tracing = {workspace = true, optional = true} # optional for opentelemetry internal logging
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -276,7 +276,7 @@ impl prometheus::core::Collector for Collector {
             Ok(guard) => guard,
             Err(err) => {
                 otel_error!(
-                    name: "OpenTelemetryPrometheus.AcquireLockError",
+                    name: "MetricScrapeFailed",
                     message = err.to_string(),
                 );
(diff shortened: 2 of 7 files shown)


wojciechbator (approved):

opentelemetry-prometheus/Cargo.toml
```diff
@@ -21,17 +21,20 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 once_cell = { workspace = true }
-opentelemetry = { version = "0.26", default-features = false, features = ["metrics"] }
-opentelemetry_sdk = { version = "0.26", default-features = false, features = ["metrics"] }
+opentelemetry = { version = "0.27", path = "../opentelemetry", default-features = false, features = ["metrics"] }
```
cijothomas, line 24: please remove the path="..". It is intentional until we hit 1.0

kudlatyamroth: ok, removed :)

opentelemetry-prometheus/src/lib.rs
```diff
@@ -281,7 +273,10 @@ impl prometheus::core::Collector for Collector {
         let mut inner = match self.inner.lock() {
             Ok(guard) => guard,
             Err(err) => {
-                global::handle_error(err);
+                otel_error!(
```
cijothomas, line 276: not sure of the impact of this error? is that going to fail entire metrics scraping? if yes, please use a more descriptive message. If not, use a lower severity. Either way, lets try to make sure the internal logs are descriptive enough.

opentelemetry-prometheus/src/lib.rs
```diff
@@ -291,7 +286,10 @@ impl prometheus::core::Collector for Collector {
             scope_metrics: vec![],
         };
         if let Err(err) = self.reader.collect(&mut metrics) {
-            global::handle_error(err);
+            otel_error!(
```
cijothomas, line 289: same comment as above.

cijothomas (changes requested):
Changes look good. I am requesting changes to fix the internal logs. These are now emitted via `tracing` and most users likely have a tracing subscriber already enabled.. So we need to be very careful about emitting logs, especially the ones with high severity.
Also, try and make the descriptions more self-explanatory.(Totally understand these are existing issues, but just want to make it easier for end users to navigate the internal logs)

opentelemetry-prometheus/src/lib.rs
```diff
@@ -152,30 +148,26 @@ pub struct PrometheusExporter {
...
         self.reader.force_flush()
     }
 
-    fn shutdown(&self) -> Result<()> {
+    fn shutdown(&self) -> MetricResult<()> {
         self.reader.shutdown()
     }
+
+    fn temporality(&self, kind: InstrumentKind) -> Temporality {
+        self.reader.temporality(kind)
```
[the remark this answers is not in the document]
kudlatyamroth: so, should i hard code it like this:
```
    fn temporality(&self, _kind: InstrumentKind) -> Temporality {
        Temporality::Cumulative
    }
```
?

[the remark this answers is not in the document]
cijothomas: yes, just to make it very obvious that prometheus needs cumulative.

opentelemetry-prometheus/src/lib.rs
```diff
@@ -152,30 +148,26 @@ pub struct PrometheusExporter {
     reader: Arc<ManualReader>,
 }
 
-impl TemporalitySelector for PrometheusExporter {
-    /// Note: Prometheus only supports cumulative temporality so this will always be
-    /// [Temporality::Cumulative].
```
hdost, line 157: We should keep this comment on the temporality function

hdost (changes requested):

opentelemetry-prometheus/src/lib.rs
kudlatyamroth: im not sure of the impact in these places, but i tried to update description and change two of them to warns :)

cijothomas: Looks like you haven't pushed the commit? Not seeing the change!

kudlatyamroth: i've changed otel error to otel warning in validate metrics, here i think should be error as it results in empty vec.
here i changed only name, end error as a message

cijothomas: > OpenTelemetryPrometheus.AcquireLockError

This name is not end user friendly. Could you use something like MetricScrapeFailed or something similar, avoiding any internal details like locks.

cijothomas: also, there is no need to prefix OpenTelemetryPrometheus - it'll be automatically added as target!

kudlatyamroth: k, changed :)

cijothomas (approved):

ttys3 (changes requested):

opentelemetry-prometheus/Cargo.toml
```diff
@@ -21,17 +21,20 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 once_cell = { workspace = true }
-opentelemetry = { version = "0.26", default-features = false, features = ["metrics"] }
-opentelemetry_sdk = { version = "0.26", default-features = false, features = ["metrics"] }
+opentelemetry = { version = "0.27", default-features = false, features = ["metrics"] }
+opentelemetry_sdk = { version = "0.27", default-features = false, features = ["metrics", "spec_unstable_metrics_views"] }
```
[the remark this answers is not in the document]
kudlatyamroth: done :)
thanks :)

rubenfiszel:
Hi, is there a published  opentelemetry-prometheus crate that one can depend on with this change?

kudlatyamroth:
i was hoping that merge will be followed by release... but it looks like we will need to wait

cijothomas:
the crate was released last week, sorry missed to update here!

cijothomas:
https://crates.io/crates/opentelemetry-prometheus/0.27.0

PR #2314, opened 2024-11-21
Make LogRecord fields private and add getters for encapsulation

lalitb:
## Changes

The PR refactors the `LogRecord` struct by making all its fields crate-private for better abstraction. Public getter methods have been added to provide controlled access to the fields. 

TBD - Will add changelog, if the changes are agreed.


opentelemetry-sdk/src/logs/record.rs
```diff
@@ -118,7 +118,56 @@ impl opentelemetry::logs::LogRecord for LogRecord {
...
+    #[inline]
+    pub fn event_name(&self) -> Option<&'static str> {
+        self.event_name
+    }
+
+    /// Returns the target
+    #[inline]
+    pub fn target(&self) -> Option<&Cow<'static, str>> {
+        self.target.as_ref()
+    }
```
lalitb, line 131: Returning &str would simplify the API, but want to preserve the ability for consumers to optimize based on whether the target is a static string.

utpilla: I think we could favor simplicity of the API here and use `&str` as the return type. The current exporters that we have cannot utilize the `static` str reference to their benefit anyway.

Only exporters that would benefit from `Cow<'static, str>` would be the ones that hold these values in memory for some reason as they could then avoid the cost to clone a static str.

utpilla (approved):

opentelemetry-sdk/src/logs/record.rs
utpilla: Not a strong opinion here as it's mainly used by exporters and not end users. We can expect exporter authors to work with `Cow<'static,str>`.

lalitb: > Not a strong opinion here as it's mainly used by exporters and not end users. We can expect exporter authors to work with Cow<'static,str>.

Yes, that's what I thought - keeping it for the custom export authors if they want to reap the benefit of static target.

cijothomas (approved):
LGTM with changelog as this is breaking export/processor authors.

lalitb:
> LGTM with changelog as this is breaking export/processor authors.

updated changelog.

PR #2316, opened 2024-11-21
Remove default logger name 

lalitb:
## Changes
As per the specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#logger-creation

``
In the case where an invalid name (null or empty string) is specified, a working Logger MUST be returned as a fallback rather than returning null or throwing an exception, its name SHOULD keep the original invalid value, and a message reporting that the specified value is invalid SHOULD be logged.
``

So changed the implementation to create the instrumentation scope with empty `name` if provided and log the warning.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -1,7 +1,7 @@
 use super::{BatchLogProcessor, LogProcessor, LogRecord, SimpleLogProcessor, TraceContext};
 use crate::{export::logs::LogExporter, runtime::RuntimeChannel, Resource};
 use crate::{logs::LogError, logs::LogResult};
-use opentelemetry::{otel_debug, trace::TraceContextExt, Context, InstrumentationScope};
+use opentelemetry::{otel_debug, otel_warn, trace::TraceContextExt, Context, InstrumentationScope};
 
 #[cfg(feature = "spec_unstable_logs_enabled")]
 use opentelemetry::logs::Severity;
@@ -44,17 +44,14 @@ pub struct LoggerProvider {
     inner: Arc<LoggerProviderInner>,
 }
 
-/// Default logger name if empty string is provided.
-const DEFAULT_COMPONENT_NAME: &str = "rust.opentelemetry.io/sdk/logger";
-
 impl opentelemetry::logs::LoggerProvider for LoggerProvider {
     type Logger = Logger;
 
     fn logger(&self, name: impl Into<Cow<'static, str>>) -> Self::Logger {
-        let mut name = name.into();
+        let name = name.into();
 
         if name.is_empty() {
-            name = Cow::Borrowed(DEFAULT_COMPONENT_NAME)
+            otel_warn!(name: "LoggerProvider.Logger.EmptyName",  message = "Logger name is empty; consider providing a meaningful name");
         };
 
         let scope = InstrumentationScope::builder(name).build();


opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -44,17 +44,14 @@ pub struct LoggerProvider {
...
-const DEFAULT_COMPONENT_NAME: &str = "rust.opentelemetry.io/sdk/logger";
-
 impl opentelemetry::logs::LoggerProvider for LoggerProvider {
     type Logger = Logger;
 
     fn logger(&self, name: impl Into<Cow<'static, str>>) -> Self::Logger {
-        let mut name = name.into();
+        let name = name.into();
 
         if name.is_empty() {
```
cijothomas, line 53: lets add a test for this?

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -44,17 +44,14 @@ pub struct LoggerProvider {
...
-            name = Cow::Borrowed(DEFAULT_COMPONENT_NAME)
+            otel_warn!(name: "LoggerProvider.Logger.EmptyName",  message = "Logger name is empty; consider providing a meaningful name");
```
cijothomas, line 54:
```suggestion
            otel_info!(name: "LoggerNameEmpty",  message = "Logger name is empty; consider providing a meaningful name. Logger will function normally and the provided name will be used as-is.");
```

lalitb: shouldn't this be otel_warn, as the end-user has provided *invalid* name?

cijothomas: Okay with warn as well. Since the telemetry is still going to flow, I though maybe Info is also okay. (if we were going to drop logs from this logger, definitely Warn)

cijothomas: Also - its not end-users who provided invalid information - its the appender authors, right?

lalitb: yes for tracing-appender. Maybe some custom appender allows user to configure it.

cijothomas: I think this should be Info only, as telemetry is flowing without issues, and is merely a suggestion to user to provide better names.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -45,17 +44,14 @@ pub struct LoggerProvider {
...
 impl opentelemetry::logs::LoggerProvider for LoggerProvider {
     type Logger = Logger;
 
     fn logger(&self, name: impl Into<Cow<'static, str>>) -> Self::Logger {
-        let mut name = name.into();
+        let name = name.into();
 
         if name.is_empty() {
-            name = Cow::Borrowed(DEFAULT_COMPONENT_NAME)
+            otel_info!(name: "LoggerNameEmpty",  message = "Logger name is empty; consider providing a meaningful name. Logger will function normally and the provided name will be used as-is.");
```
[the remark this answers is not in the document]
lalitb: Good point. Done.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Done.

opentelemetry-sdk/src/logs/log_emitter.rs
cijothomas, line 54: lets remove this , otherwise we'll print duplicate logs, one from here and another from inside the logger_with_scope too which this calls.

lalitb: oh yes, missed that.

cijothomas (approved):
Thanks!

PR #2322, opened 2024-11-22
Feat/add resource builder

### LINKED ISSUE #2320: Improve Resource related public APIs
`Resource` related public APIs need some improvement:
1. Remove the [`timeout`](https://github.com/open-telemetry/opentelemetry-rust/blob/465fcc2eda4e6e1f98ec5fdb74472f52cdc5207c/opentelemetry-sdk/src/resource/mod.rs#L265) parameter from the `detect` method signature of the trait `ResourceDetector`. This doesn't seem to add any value. The individual resource detectors implementing this trait can account for timeout in their implementation if required.
2. Use a builder pattern for creating resource to make it simpler for users to add custom detectors. Currently, this functionality is offered by [`from_detectors`](https://github.com/open-telemetry/opentelemetry-rust/blob/465fcc2eda4e6e1f98ec5fdb74472f52cdc5207c/opentelemetry-sdk/src/resource/mod.rs#L140) method.

pitoniak32:
fixes #2320

## Migration Guide
Creating a Resource with default detectors:
```rust
// old
let resource = Resource::default()
    .with_attributes([
        KeyValue::new("service.name", "example_service"),
        KeyValue::new("test1", "test_value"),
        KeyValue::new("test1", "test_value1"),
        KeyValue::new("test2", "test_value2"),
    ]);

// new
let resource = Resource::builder()
    .with_service_name("example_service")
    .with_attributes([
        KeyValue::new("test1", "test_value")
        KeyValue::new("test1", "test_value1"),
        KeyValue::new("test2", "test_value2"),
    ])
    .build();
```
Creating an empty Resource:
```rust
// old
let resource = Resource::empty();

// new
let resource = Resource::builder_empty().build();
```

## Changes
- Add `ResourceBuilder` to more easily create resources.

example usage:
```rust
let resource = Resource::builder()
    .with_service_name("gifting_service")
    .with_detector(Box::new(EnvResourceDetector::new()))
    .with_attribute(KeyValue::new("test1", "test_value"))
    .with_attributes([
        KeyValue::new("test1", "test_value1"),
        KeyValue::new("test2", "test_value2"),
    ])
    .build();
```


pitoniak32:
I still need to expose the new builder to the API, and would like feedback on https://github.com/open-telemetry/opentelemetry-rust/issues/2324 before going further since the changes are related :)

utpilla:
@pitoniak32 Could you split this into two PRs? Have one PR just for removing the `timeout` argument and one for resource builder. That way we can merge the `timeout` related PR while we review the builder related changes.

pitoniak32:
> @pitoniak32 Could you split this into two PRs? Have one PR just for removing the `timeout` argument and one for resource builder. That way we can merge the `timeout` related PR while we review the builder related changes.

split those changes into: https://github.com/open-telemetry/opentelemetry-rust/pull/2332

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /// Add multiple [ResourceDetector] to your resource.
+    pub fn with_detectors(mut self, detectors: Vec<Box<dyn ResourceDetector>>) -> Self {
+        self.resource = self.resource.merge(&Resource::from_detectors(detectors));
+        self
+    }
+
+    /// Add a [KeyValue] to the resource.
+    pub fn with_key_value(self, kv: KeyValue) -> Self {
```
cijothomas, line 38: interesting why this does not need `mut self`...

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -54,18 +56,31 @@ pub struct Resource {
...
+    /// Starts with a [Resource::empty()].
+    pub fn builder() -> ResourceBuilder {
+        ResourceBuilder::new_empty()
+    }
+
+    /// Creates a Builder that allows you to configure multiple aspects of the Resource.
+    ///
+    /// Starts with a [Resource::default()].
+    pub fn builder_default() -> ResourceBuilder {
+        ResourceBuilder::new_empty()
```
cijothomas, line 81: this is meant to call the default() one right?

stress/src/throughput.rs
```diff
@@ -155,7 +155,7 @@ impl<'a> UnsafeSlice<'a> {
     #[inline(always)]
     unsafe fn increment(&self, i: usize) {
         let value = self.slice[i].get();
-        (*value).count = (*value).count + 1;
```
cijothomas, line 158: lets remove all unrelated changes, to make it easier on reviewers.
(Its totally fine to send a short, separate PR for this.)

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: It might be since the with_detectors consumes the builder and its not a `&mut` reference? That's just a guess I'm actually not sure about that 😅

pitoniak32:
I was also thinking of doing something like this for with_key_value, I'm wondering what others think?

```rust
let resource = Resource::builder()
    .with_key_value(KeyValue::new("test1", "test_value"))
    .build();

// vs

let resource = Resource::builder()
    .with_key_value("test1", "test_value")
    .build();
```

```rust
/// Add a [KeyValue] to the resource.
pub fn with_key_value<K, V>(self, key: K, value: V) -> Self
where
    K: Into<Key>,
    V: Into<Value>,
{
    self.with_key_values(vec![KeyValue::new(key, value)])
}
```

currently its:
```rust
/// Add a [KeyValue] to the resource.
pub fn with_key_value(self, kv: KeyValue) -> Self {
    self.with_key_values(vec![kv])
}
```

examples/metrics-advanced/src/main.rs
```diff
@@ -49,13 +49,14 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
...
+        .build();
+
     let reader = PeriodicReader::builder(exporter, runtime::Tokio).build();
     let provider = SdkMeterProvider::builder()
         .with_reader(reader)
-        .with_resource(Resource::new([KeyValue::new(
-            "service.name",
-            "metrics-advanced-example",
-        )]))
+        .with_resource(resource)
```
cijothomas, line 59: wondering if we can do `with_servicename_resource("metrics-advanced-example")` ? This will be the most common use case.
Or just `Resource::builder().with_servicename("metrics-advanced-example")......build()` ?

pitoniak32: I like that, I was using the go implementation for Resource builder this weekend and they had something similar to this and it felt pretty nice
```go
sResource, err := resource.New(
  ctx,
  resource.WithTelemetrySDK(),
  resource.WithAttributes(semconv.ServiceName("trace-export-service")),
)
```

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,108 @@
...
+impl ResourceBuilder {
+    /// Create ResourceBuilder with an empty [Resource].
+    pub fn new_empty() -> Self {
+        ResourceBuilder {
+            resource: Resource::empty(),
+        }
+    }
+
+    /// Create ResourceBuilder with a default [Resource].
+    pub fn new_default() -> Self {
```
hdost, line 22:
```suggestion
    pub fn default() -> Self {
```
Probably should be done as an `impl Default for ResourceBuilder` to make it more generally compatible.

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// Builder to allow easy composition of a Resource
+#[derive(Debug, Default)]
+pub struct ResourceBuilder {
+    resource: Resource,
+}
+
+impl ResourceBuilder {
+    /// Create ResourceBuilder with an empty [Resource].
+    pub fn new_empty() -> Self {
```
hdost, line 15:
```suggestion
    pub fn new() -> Self {
```
This makes it easier to reason about. Since on `Resource` the new is the same as empty but it accepts key values. Since this is a builder most would assume it's empty.

hdost (changes requested):
Thank you for the contribution 😊

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: updated

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: updated

examples/metrics-advanced/src/main.rs
pitoniak32: added the option of `.with_service_name("name")`

TommyCpp: Hmm should we give special treatment to service name? I feel like it will open a can of worms that which attributes should get dedicate builder method. I think we should keep the API lean for now. 

On the long term we can use dedicate trait to enrich the builder. Image something like 
```rust
pub trait SemCov {
  fn with_service_name(&str);
  fn with_service_version(&str);
  ...
  // those method can probably generated in sem cov crate
}

impl SemCov for ResourceBuilder {
  fn with_service_name(&str) {
    ...
   // again we can generated those code in sem cov crate
  }
}

// main.rs
import {ResourceBuilder, SemCov}

let builder = ResourceBuilder::new().with_service_name("test")
```

pitoniak32: Sounds reasonable to me, I was kind of wondering the same thing where we would want to use the key from sem conv instead of the hardcoded one in the sdk. I will remove this, and it can become another discussion

opentelemetry-sdk/src/resource/builder.rs
lalitb: `with_key_value` doesn’t need `mut self` because it doesn’t directly modify the builder; it delegates to `with_key_values`, which takes mut self to perform the actual mutation.

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,102 @@
...
+        self
+    }
+
+    /// Add a [KeyValue] to the resource.
+    pub fn with_key_value(self, kv: KeyValue) -> Self {
+        self.with_key_values(vec![kv])
+    }
+
+    /// Add multiple [KeyValue]s to the resource.
+    pub fn with_key_values<T: IntoIterator<Item = KeyValue>>(mut self, kvs: T) -> Self {
```
lalitb, line 49: should the method be named `with_attribute` and `with_attributes` - Just that the spec uses the term "attribute" consistently when describing resource data.

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.resource = self.resource.merge(&Resource::new(kvs));
+        self
+    }
+
```
lalitb, line 53: should there also be a method `with_schema_url` to add schema url ?

lalitb:
> /// Add a [KeyValue] to the resource.
pub fn with_key_value<K, V>(self, key: K, value: V) -> Self
where
    K: Into<Key>,
    V: Into<Value>,
{
    self.with_key_values(vec![KeyValue::new(key, value)])
}

I vote for this, it's similar to what is followed for `LogRecord::add_attribute`:
https://github.com/open-telemetry/opentelemetry-rust/blob/a3c469b1a681d9a794f78aeab5f942e6b6527aa6/opentelemetry-sdk/src/logs/record.rs#L98-L102

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,102 @@
...
+use super::{Resource, ResourceDetector};
+
+/// Builder to allow easy composition of a Resource
+#[derive(Debug)]
+pub struct ResourceBuilder {
+    resource: Resource,
+}
+
+impl Default for ResourceBuilder {
+    /// Create ResourceBuilder with [Resource::default()].
```
cijothomas, line 14: can we mention what the default would be here also? i.e it'll contain the telemetry,env etc.?

opentelemetry-sdk/src/resource/builder.rs
cijothomas: +1

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    /// Add a single [ResourceDetector] to your resource.
+    pub fn with_detector(self, detector: Box<dyn ResourceDetector>) -> Self {
+        self.with_detectors(vec![detector])
+    }
+
+    /// Add multiple [ResourceDetector] to your resource.
+    pub fn with_detectors(mut self, detectors: Vec<Box<dyn ResourceDetector>>) -> Self {
```
TommyCpp, line 36: Can we use `&'a dyn ResourceDetector`? IIRC `detect` takes `&self` and we only need it for the lifetime of this functions

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// Add multiple [KeyValue]s to the resource.
+    pub fn with_key_values<T: IntoIterator<Item = KeyValue>>(mut self, kvs: T) -> Self {
+        self.resource = self.resource.merge(&Resource::new(kvs));
+        self
+    }
+
+    /// Create a [Resource] with the options provided to the [ResourceBuilder].
+    pub fn build(self) -> Resource {
```
TommyCpp, line 55: Can we implement `Into<Resource>` for `ResourceBuilder` to make it easier to use `ResourceBuilder` as `Resource` in parameters

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: yeah I can add this here, I was thinking the link to the default builder might be good to not have possible duplication / stale comments

examples/metrics-advanced/src/main.rs
pitoniak32: so do we want to add the option for service.name?

cijothomas: Yes let's do it. Its okay to do it as a separate follow up PR so this PR is not blocked.

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: added more info

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: I did the From<ResourceBuilder> because clippy was complaining since it gives Into for free

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: fixed this

examples/metrics-advanced/src/main.rs
pitoniak32: added it back because its small

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: How would we like the behavior of the `with_schema_url` to act? Do we want to follow the rules that `from_schema_url` has? Or override the current one with the specified one, or maybe use TypeState pattern to only allow the schema_url to be set once?

pitoniak32: this is my initial thought
```rust
/// This will merge the provided `schema_url` with the current state of the Resource being built. It
/// will use the following rules to determine which `schema_url` should be used.
///
/// ### [Schema url]
/// If both of the resources are not empty. Schema url is determined by the following rules, in order:
/// 1. If the provided resource has a schema url, it will be used.
/// 2. If the provided resource does not have a schema url, and the builders current resource has a schema url, it will be used.
/// 3. If both resources have a schema url and it's the same, it will be used.
/// 4. If both resources have a schema url and it's different, the schema url will be empty.
/// 5. If both resources do not have a schema url, the schema url will be empty.
///
/// [Schema url]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
pub fn with_schema_url<S>(mut self, schema_url: S) -> Self
where
    S: Into<Cow<'static, str>>,
{
    self.resource = Resource::from_schema_url([], schema_url).merge(&self.resource);
    self
}
```

pitoniak32: added this, would love to hear feedback on the implementation!

examples/metrics-advanced/src/main.rs
```diff
@@ -49,13 +49,14 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
         .with_temporality(Temporality::Delta)
         .build();
 
+    let resource = Resource::builder()
+        .with_attribute(KeyValue::new("service.name", "metrics-advanced-example"))
```
cijothomas, line 53: lets use `with_service_name` here.

pitoniak32: done

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -165,16 +189,24 @@ impl Resource {
     ///
     /// [Schema url]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
     pub fn merge<T: Deref<Target = Self>>(&self, other: T) -> Self {
+        println!("here1");
```
cijothomas, line 192: nit: remove these.

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn default() -> Self {
+        ResourceBuilder {
+            resource: Resource::default(),
+        }
+    }
+}
+
+impl ResourceBuilder {
+    /// Create ResourceBuilder with [Resource::empty()].
+    pub fn new() -> Self {
```
cijothomas, line 29: is this needed, given Resource already expose public methods to create a ResourceBuilder?

cijothomas: wondering what is the best option?
Should users have the ability to get Builder instance from Builder itself, or via the thing it is ultimately building or both? We need to review rest of the codebase to ensure we do this consistently across.

I think its best to offer just one way to create a builder, and that must be via a method on the thing we are ultimately  building.
so Resource::builder().....build() - only this should be required.
and not ResourceBuilder::new()....build().

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: opps!

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: It looks like the others are keeping new private, and exposing the builder through the api. In this case how would be want to handle starting with empty resource vs a default one? Would we want to always start with a empty one and then allow `with_default()` or something like that?

pitoniak32: the others are hiding the new method, I will update this, I agree with you.

pitoniak32: I think the reason that was public is to allow the creation of an empty resource vs a default resource. In this case would it make more sense to make it `pub(super)`? so the resource is still able to use it? Or only allow starting from an empty resource, and then expose a `with_default()`?

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: fixed

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -61,7 +63,29 @@ impl Default for Resource {
...
+    }
+}
+
 impl Resource {
+    /// Creates a Builder that allows you to configure multiple aspects of the Resource.
+    ///
+    /// If you want to start from a [Resource::default()] see [Resource::builder_default()].
+    ///
+    /// Starts with a [Resource::empty()].
+    pub fn builder() -> ResourceBuilder {
```
cijothomas, line 78: I am still debating about this.. It maybe better to name this builder_empty, to make it very obvious that this starts with an empty Resource, and not with OTel SDK defaults?

opentelemetry-sdk/src/resource/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+
+use super::{Resource, ResourceDetector, SERVICE_NAME};
+
+/// Builder to allow easy composition of a Resource
+#[derive(Debug)]
+pub struct ResourceBuilder {
+    resource: Resource,
+}
+
+impl Default for ResourceBuilder {
```
cijothomas, line 13: we can remove the Default implementation  for the builder, so as to force users to always go via Resource to get hold of a builder.

examples/metrics-advanced/src/main.rs
```diff
@@ -49,13 +49,14 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
...
+        .build();
+
     let reader = PeriodicReader::builder(exporter, runtime::Tokio).build();
     let provider = SdkMeterProvider::builder()
         .with_reader(reader)
-        .with_resource(Resource::new([KeyValue::new(
-            "service.name",
-            "metrics-advanced-example",
-        )]))
+        .with_resource(resource)
```
cijothomas, line 59: 1. If user never called with_resource.-  use SDK defaults.
2. with_resource(Resource::builder().with..with...build()) - Start with SDK defaults, and merge the rest.
3. with_resource(Resource::builder_empty().with..with...build()) - Start with empty, and merge the rest.
4. provider.append_resource(..) - merges with existing Resource(s)

^ Does this look the right approach @open-telemetry/rust-approvers please share your thoughts, this is important to get closure soon.

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -61,7 +63,29 @@ impl Default for Resource {
     }
 }
 
+impl From<ResourceBuilder> for Resource {
```
utpilla, line 66: I don't think we need to implement this conversion. I believe the intuitive thing to do is call `build()` on a builder type and not `into()`.

examples/metrics-advanced/src/main.rs
utpilla: After some more thinking into this, I believe there mainly two use-cases here:

1. Set resource to something very specific (remove default resources)
2. Append more resources to the existing ones

I believe a simple solution here could be:
- Introduce an API named `set_resource(resource)` which will replace the provider's resource with the value in the method argument. This would cover the 1st use case.
- Use the existing `with_resource(resource)` API (or rename it to `append_resource(resource)` to allow users to append/merge resources with the existing provider resource. The existing resource would either be the default ones or something else if the user called `set_resource(resource)`. This would cover the 2nd use case.

lalitb: Do we have the use-case where resource would be updated/merged in Provider after it's creation. Atleast, I don't see it.  The user needs to ensure that it creates the resource object by merging all the default/custom/required detectors before creating the Provider. And in that case, set_resource(), add_resource() and merge_resource() on Provider is not required. So,

These looks fine:
- If user never called with_resource.- use SDK defaults.

This is also fine:
- with_resource(Resource::builder().with..with...build()) - Start with SDK defaults, and merge the rest.
- with_resource(Resource::builder_empty().with..with...build()) - Start with empty, and merge the rest.

Or maybe another option could be:
`Resource::builder(start_with_default: bool) -> ResourceBuilder`
i.e,
- with_resource(Resource::builder(start_with_default: true).with..with...build()) - Start with SDK defaults, and merge the rest.
- with_resource(Resource::builder(start_with_default: false).with..with...build()) - Start with SDK defaults, and merge the rest.
And `ResourceBuilder::new()` and `ResourceBuilder::default()` can be made private/removed.

But _NOT_ this, as I don't see it's use-case:
- provider.append_resource(..) - merges with existing Resource(s). 

the user should instead do invoke
```rust
  let merged_resource = resource.merge(to_be_merged);
  let provider = Provider::builder().with_resource(merged_resource)... .buld();
```

pitoniak32: I like the idea of using defaults, and giving the option of overriding them with `with_resource(Resource::empty())`, im not sure if it would be redundant to also provide `Resource::builder_empty()`, but I also like that option.

I'm not partial to `Resource::builder(true)` because it doesn't feel idiomatic requiring an argument to toggle behavior with a bool.

lalitb: > I'm not partial to Resource::builder(true) because it doesn't feel idiomatic requiring an argument to toggle behavior with a bool.

Agree, this doesn't look idiomatic. 

> I like the idea of using defaults, and giving the option of overriding them with with_resource(Resource::empty()), im not sure if it would be redundant to also provide Resource::builder_empty(), but I also like that option.

I don't see `Resource::builder_empty()` redundant. It would be to used to start building the resource empty, how can we achieve this otherwise?

pitoniak32: The with_resource could be given a empty resource to override the default one

lalitb: > I like the idea of having builder_empty, it makes the option more obvious

Yes, I also prefer this. Also correct me if I am wrong, but I don't see use of `Resource::empty()`, `Resource::new_with_defaults()`, `Resource::new()`, `ResourceBuilder::new()`, `ResoureBuilder::defaults()` to be public. `Resource::builder()` and `Resource::builder_empty()` should be the only entry point to create resources.

cijothomas: > Resource::builder() and Resource::builder_empty() should be the only entry point to create resources.

This is good. Let us remove all other public APIs other than this.


_provider.append_resource(..)_
Agreed to *not* offer this in 1.0. If we chose to add it later, it is an additive change only. (There were scenarios which required this, especially from .NET feedbacks... But its best to be conservative in 1.0 release).

pitoniak32: Thank you all for the feedback! I should be able to apply it tonight after work

pitoniak32: so we want to remove `Resource::empty()` from public api? This seems to be used in quite a few places, just want to confirm before I remove it

pitoniak32: pushed up most of the suggestions other than removing the empty from public api

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: removed

opentelemetry-sdk/src/resource/builder.rs
pitoniak32: removed

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -62,6 +62,24 @@ impl Default for Resource {
 }
 
 impl Resource {
+    /// Creates a Builder that allows you to configure multiple aspects of the Resource.
+    ///
+    /// Starts with a [Resource::default()].
+    pub fn builder() -> ResourceBuilder {
```
cijothomas, line 68: shouldn't we remove all other pub methods on Resource, and ask users to user ResourceBuilder only?

pitoniak32: I had comments on our thread asking about this to confirm understanding before making the changes since it's used in quite a few places throughout the codebase

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -62,6 +62,24 @@ impl Default for Resource {
 }
 
 impl Resource {
+    /// Creates a Builder that allows you to configure multiple aspects of the Resource.
+    ///
+    /// Starts with a [Resource::default()].
```
utpilla, line 67: nit: It would be nice to list out the default resources that would be added for easy reference.

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -261,6 +280,65 @@ pub trait ResourceDetector {
...
+}
+
+impl ResourceBuilder {
+    /// Add a single [ResourceDetector] to your resource.
+    pub fn with_detector(self, detector: Box<dyn ResourceDetector>) -> Self {
+        self.with_detectors(vec![detector])
+    }
+
+    /// Add multiple [ResourceDetector] to your resource.
+    pub fn with_detectors(mut self, detectors: Vec<Box<dyn ResourceDetector>>) -> Self {
```
utpilla, line 296: Could we change the argument type from `Vec` to a slice? We don't have to force users to allocate a vec when they can create an array on stack.

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: updated

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: removed the public api options for creating resources

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: updated

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
     [#2338](https://github.com/open-telemetry/opentelemetry-rust/pull/2338)
   - `ResourceDetector.detect()` no longer supports timeout option.
   - `opentelemetry::global::shutdown_tracer_provider()` Removed from the API, should now use `tracer_provider.shutdown()` see [#2369](https://github.com/open-telemetry/opentelemetry-rust/pull/2369) for a migration example. "Tracer provider" is cheaply cloneable, so users are encouraged to set a clone of it as the global (ex: `global::set_tracer_provider(provider.clone()))`, so that instrumentations and other components can obtain tracers from `global::tracer()`. The tracer_provider must be kept around to call shutdown on it at the end of application (ex: `tracer_provider.shutdown()`)
-
+- *Feature*: Add `ResourceBuilder` for an easy way to create new `Resource`s
```
cijothomas, line 13: please note the breaking change in changelog and show the migration snippet.

examples/metrics-basic/src/main.rs
```diff
@@ -12,10 +12,11 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
     let reader = PeriodicReader::builder(exporter).build();
     let provider = SdkMeterProvider::builder()
         .with_reader(reader)
-        .with_resource(Resource::new([KeyValue::new(
-            "service.name",
-            "metrics-basic-example",
-        )]))
+        .with_resource(
+            Resource::builder_empty()
```
cijothomas, line 16: I think we can consistently use Resource::builder in all examples, as that would be normal path most users need. builder_empty() is only for the advanced users who want to avoid the default resources.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -24,7 +24,7 @@ fn noop_logger_provider() -> &'static LoggerProvider {
     NOOP_LOGGER_PROVIDER.get_or_init(|| LoggerProvider {
         inner: Arc::new(LoggerProviderInner {
             processors: Vec::new(),
-            resource: Resource::empty(),
+            resource: Resource::builder_empty().build(),
```
cijothomas, line 27: nit: no harm keeping Resource::empty() as pub(crate)

cijothomas (approved):
LGTM.
I think we can keep few methods on Resource like new() as pub(mod) to reduce the amount of internal changes.

Please update changelog to callout the breaking change/migration snippets.

Thanks for patiently working through this!

examples/metrics-basic/src/main.rs
pitoniak32: okay will update, I was keeping consistent to how it was before

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -51,20 +51,35 @@ pub struct Resource {
...
-    fn default() -> Self {
-        Self::from_detectors(vec![
-            Box::new(SdkProvidedResourceDetector),
-            Box::new(TelemetryResourceDetector),
-            Box::new(EnvResourceDetector::new()),
-        ])
+impl Resource {
+    /// Creates a [ResourceBuilder] Starting from [Resource::default()] that allows you to configure multiple aspects of the Resource.
+    ///
+    /// This Resource will contain the following [ResourceDetector]s:
```
utpilla, line 57:
```suggestion
    /// This [ResourceBuilder] will include the following [ResourceDetector]s by default:
```

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -51,20 +51,35 @@ pub struct Resource {
...
+                Box::new(SdkProvidedResourceDetector),
+                Box::new(TelemetryResourceDetector),
+                Box::new(EnvResourceDetector::new()),
+            ]),
+        }
+    }
+
+    /// Creates a [ResourceBuilder] starting from [Resource::empty()] that allows you to configure multiple aspects of the Resource.
+    ///
+    /// This Resource will contain no attributes, and no [ResourceDetector]s.
```
utpilla, line 73:
```suggestion
    /// This [ResourceBuilder] will not include any attributes or [ResourceDetector]s by default.
```

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -261,6 +269,65 @@ pub trait ResourceDetector {
...
+    resource: Resource,
+}
+
+impl ResourceBuilder {
+    /// Add a single [ResourceDetector] to your resource.
+    pub fn with_detector(self, detector: Box<dyn ResourceDetector>) -> Self {
+        self.with_detectors(&[detector])
+    }
+
+    /// Add multiple [ResourceDetector] to your resource.
```
utpilla, line 284:
```suggestion
    /// Add multiple [ResourceDetector]s to your resource.
```

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -261,6 +269,65 @@ pub trait ResourceDetector {
...
+    pub fn with_detectors(mut self, detectors: &[Box<dyn ResourceDetector>]) -> Self {
+        self.resource = self.resource.merge(&Resource::from_detectors(detectors));
+        self
+    }
+
+    /// Add a [KeyValue] to the resource.
+    pub fn with_attribute(self, kv: KeyValue) -> Self {
+        self.with_attributes(vec![kv])
```
utpilla, line 292: We can avoid this `vec!` allocation and use any array instead.

```suggestion
        self.with_attributes([kv])
```

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -10,7 +10,22 @@
...
+- *Breaking*: Remove `Resource::{new,empty,from_detectors,new_with_defaults,from_schema_url,merge,default}` from public api. To create Resources you should only use `Resource::builder()` or `Resource::builder_empty()`. See [#2322](https://github.com/open-telemetry/opentelemetry-rust/pull/2322) for a migration guide.
+  Example Usage:
+  ```rust
+  // old
+  Resource::default().with_attributes(vec![
```
utpilla, line 18: It would be nice to showcase good practices. We don't really need a `vec!` allocation here.

```suggestion
  Resource::default().with_attributes([
```

opentelemetry-sdk/src/resource/env.rs
```diff
@@ -22,7 +22,7 @@ impl ResourceDetector for EnvResourceDetector {
     fn detect(&self) -> Resource {
         match env::var(OTEL_RESOURCE_ATTRIBUTES) {
             Ok(s) if !s.is_empty() => construct_otel_resources(s),
-            Ok(_) | Err(_) => Resource::new(vec![]), // return empty resource
+            Ok(_) | Err(_) => Resource::empty(), // return empty resource
```
lalitb, line 25: We are not removing Resource::empty() ?

lalitb (approved):
LGTM! This looks great. Thank you for making all the updates and for your patience in addressing the comments

opentelemetry-sdk/src/resource/mod.rs
pitoniak32: updated, and updated the migration guide

opentelemetry-otlp/src/lib.rs
```diff
@@ -162,7 +162,7 @@
 //!                 .with_max_events_per_span(64)
 //!                 .with_max_attributes_per_span(16)
 //!                 .with_max_events_per_span(16)
-//!                 .with_resource(Resource::new(vec![KeyValue::new("service.name", "example")])),
+//!                 .with_resource(Resource::builder_empty().with_attributes(vec![KeyValue::new("service.name", "example")]).build()),
```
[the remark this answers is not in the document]
pitoniak32: yeah I can open another pr to convert these references

pitoniak32:
> LGTM! This looks great. Thank you for making all the updates and for your patience in addressing the comments
> 
> ps. Integration test is failing.

does it make sense to keep using an empty resource for integration tests, or should it get updated to have the default detectors?

cijothomas:
> > LGTM! This looks great. Thank you for making all the updates and for your patience in addressing the comments
> > ps. Integration test is failing.
> 
> does it make sense to keep using an empty resource for integration tests, or should it get updated to have the default detectors?

either is fine! Do what is easiest for this PR.

PR #2331, opened 2024-11-23
Emit internal log for global meter provider

cijothomas:
Continuing from https://github.com/open-telemetry/opentelemetry-rust/pull/2330.
Also realized the `opentelemetry` crate did not internal-logs ff enabled by default, and fixed it.

### DIFF
--- a/opentelemetry/src/global/metrics.rs
+++ b/opentelemetry/src/global/metrics.rs
@@ -1,5 +1,5 @@
 use crate::metrics::{self, Meter, MeterProvider};
-use crate::InstrumentationScope;
+use crate::{otel_error, otel_info, InstrumentationScope};
 use std::sync::{Arc, OnceLock, RwLock};
 
 type GlobalMeterProvider = Arc<dyn MeterProvider + Send + Sync>;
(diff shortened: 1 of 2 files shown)


opentelemetry/src/global/metrics.rs
```diff
@@ -19,10 +19,15 @@ pub fn set_meter_provider<P>(new_provider: P)
...
-        .expect("GLOBAL_METER_PROVIDER RwLock poisoned");
-    *global_provider = Arc::new(new_provider);
+    // Try to set the global meter provider. If the RwLock is poisoned, we'll log an error.
+    let mut global_provider = global_meter_provider().write();
+
+    if let Ok(ref mut provider) = global_provider {
+        **provider = Arc::new(new_provider);
+        otel_info!(name: "MeterProvider.GlobalSet", message = "Global meter provider is set. Meters can now be created using global::meter() or global::meter_with_scope().");
+    } else {
+        otel_error!(name: "MeterProvider.GlobalSetFailed", message = "Global meter provider is not set due to lock poison. Meters created using global::meter() or global::meter_with_scope() will not function.");
```
cijothomas, line 29: not sure if this (lock poison) is ever user-actionable... Maybe just state "OTel internal error occurred, report it at OpenTelemtry Github repo"?

lalitb: It's not user-actionable in its current form. If `set_meter_provider` returned a `PoisonError`, users could reinitialize the provider to Noop and retry. In current form, it's better to log the error and instruct users to report it to the repository.

lalitb (approved):
LGTM with nit comment to report error to Otel repo.

opentelemetry/src/global/metrics.rs
cijothomas: will change it in next PR

PR #2333, opened 2024-11-24
InternalLog fixes for GlobalMeterProvider

cijothomas:
Addresses https://github.com/open-telemetry/opentelemetry-rust/pull/2331#discussion_r1855278739
Adds log when Get fails due to posion.

### DIFF
--- a/opentelemetry/src/global/metrics.rs
+++ b/opentelemetry/src/global/metrics.rs
@@ -21,21 +21,24 @@ where
 {
     // Try to set the global meter provider. If the RwLock is poisoned, we'll log an error.
     let mut global_provider = global_meter_provider().write();
-
     if let Ok(ref mut provider) = global_provider {
         **provider = Arc::new(new_provider);
         otel_info!(name: "MeterProvider.GlobalSet", message = "Global meter provider is set. Meters can now be created using global::meter() or global::meter_with_scope().");
     } else {
-        otel_error!(name: "MeterProvider.GlobalSetFailed", message = "Global meter provider is not set due to lock poison. Meters created using global::meter() or global::meter_with_scope() will not function.");
+        otel_error!(name: "MeterProvider.GlobalSetFailed", message = "Setting global meter provider failed. Meters created using global::meter() or global::meter_with_scope() will not function. Report this issue in OpenTelemetry repo.");
     }
 }
 
 /// Returns an instance of the currently configured global [`MeterProvider`].
 pub fn meter_provider() -> GlobalMeterProvider {
-    global_meter_provider()
-        .read()
-        .expect("GLOBAL_METER_PROVIDER RwLock poisoned")
-        .clone()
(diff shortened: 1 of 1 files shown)


opentelemetry/src/global/metrics.rs
```diff
@@ -21,21 +21,24 @@ where
 {
     // Try to set the global meter provider. If the RwLock is poisoned, we'll log an error.
     let mut global_provider = global_meter_provider().write();
-
     if let Ok(ref mut provider) = global_provider {
         **provider = Arc::new(new_provider);
         otel_info!(name: "MeterProvider.GlobalSet", message = "Global meter provider is set. Meters can now be created using global::meter() or global::meter_with_scope().");
     } else {
-        otel_error!(name: "MeterProvider.GlobalSetFailed", message = "Global meter provider is not set due to lock poison. Meters created using global::meter() or global::meter_with_scope() will not function.");
+        otel_error!(name: "MeterProvider.GlobalSetFailed", message = "Setting global meter provider failed. Meters created using global::meter() or global::meter_with_scope() will not function. Report this issue in OpenTelemetry repo.");
```
cijothomas, line 28: in a future PR, i'll try to make this consistent, by providing an attribute "suggested_action" where the value can inform users about actions they can do. In this case there are 2
1. Report it at OpenTelemetry repro.
2. Use meters from other meter provider instance, if they have access to.

This requires some time to come up with a consistent way across entire repo to ensure every warn/err logs have a clear action for end users.

opentelemetry/src/global/metrics.rs
```diff
@@ -21,21 +21,24 @@ where
...
-        .read()
-        .expect("GLOBAL_METER_PROVIDER RwLock poisoned")
-        .clone()
+    // Try to get the global meter provider. If the RwLock is poisoned, we'll log an error and return a NoopMeterProvider.
+    let global_provider = global_meter_provider().read();
+    if let Ok(provider) = global_provider {
+        provider.clone()
+    } else {
+        otel_error!(name: "MeterProvider.GlobalGetFailed", message = "Getting global meter provider failed. Meters created using global::meter() or global::meter_with_scope() will not function. Report this issue in OpenTelemetry repo.");
+        Arc::new(crate::metrics::noop::NoopMeterProvider::new())
```
lalitb, line 40: This does great, to maintain continuity for application. Additionally, we can also consider trying to recover the poisoned provider, check if it's still intact (since corruption could happen if another thread panicked while holding the lock), and return it if everything is good. Of course, this would need to be handled on a case-by-case basis, and only where recovery makes sense. Something, to be tried separate of this PR, I will have a look if it is possible.

lalitb (approved):
Thanks for the fix.

opentelemetry/src/global/metrics.rs
cijothomas: agree. we absolutely cannot panic here so this is the minimal change to meet that goal. lets see if we can recover nicely.

Issue #1942, opened 2024-07-17
Eliminate Dynamic Dispatching in Log Pipeline for Performance Optimization


lalitb:
Currently, the log pipeline employs dynamic dispatching within its workflow:

```
                        (dynamic)                                      (dynamic)
Logger::emit(record)  ------------> LogRecordProcessor::emit(data)   ------------->   LogExporter::export(batch)
```

While it is possible to implement custom `LogRecordProcessor` to avoid the second dynamic dispatch, the first dynamic dispatch remains unavoidable. This overhead should be eliminated.

This issue has been created to track the necessary improvements.

Reference:

1. LogEmitter.rs: https://github.com/open-telemetry/opentelemetry-rust/blob/a9b8621cd1004e7f083002071b9f9a57748c6b99/opentelemetry-sdk/src/logs/log_emitter.rs#L140-L144

2. SimpleLogProcessor: https://github.com/open-telemetry/opentelemetry-rust/blob/a9b8621cd1004e7f083002071b9f9a57748c6b99/opentelemetry-sdk/src/logs/log_processor.rs#L78-L81

4. BatchLogProcessor: https://github.com/open-telemetry/opentelemetry-rust/blob/a9b8621cd1004e7f083002071b9f9a57748c6b99/opentelemetry-sdk/src/logs/log_processor.rs#L200-L202

lalitb:
Assuming [SimpleConcurrentProcessor](https://github.com/open-telemetry/opentelemetry-specification/pull/4163) is part of the specs. The simple solution could be 
 - Add static dispatch for `SimpleLogProcessor` and `SimpleConcurrentProcessor` because their types are known at compile time.
 - We require dynamic dispatch for `BatchLogProcessor` as its runtime type can vary based on configuration, making it difficult to use a static type.
 - Any other custom log processor would be dispatched at runtime.
 
```rust
enum LogProcessorEnum {
    SimpleLogProcessor(SimpleLogProcessor),
    SimpleConcurrentProcessor(SimpleConcurrentProcessor),
    Batch(BatchLogProcessor<Box<dyn RuntimeChannel>>),
    DynLogProcessor(Box<dyn LogProcessor>),

}

impl LogProcessor for LogProcessorEnum {
    fn emit(&self, data: &mut LogData) {
        match self {
            LogProcessorEnum::SimpleLogProcessor(p) => p.emit(data),
[...]

PR #2338, opened 2024-11-25
Use static dispatch in SimpleLogProcessor

cijothomas:
Part of https://github.com/open-telemetry/opentelemetry-rust/issues/1942
Given this is a breaking change, we should not merge until 0.27.1 is out.

Thanks @lalitb for helping with this. I will see if I can continue this to achieve the rest of ideas mentioned in 1942 and avoid dyn dispatch completely in Log hot path (not for Batch scenario+any custom)

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -77,13 +77,13 @@ pub trait LogProcessor: Send + Sync + Debug {
 /// debugging and testing. For scenarios requiring higher
 /// performance/throughput, consider using [BatchLogProcessor].
 #[derive(Debug)]
-pub struct SimpleLogProcessor {
-    exporter: Mutex<Box<dyn LogExporter>>,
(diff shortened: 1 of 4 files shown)


utpilla (approved):

lalitb (approved):

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -77,21 +77,21 @@ pub trait LogProcessor: Send + Sync + Debug {
 /// debugging and testing. For scenarios requiring higher
 /// performance/throughput, consider using [BatchLogProcessor].
 #[derive(Debug)]
-pub struct SimpleLogProcessor {
-    exporter: Mutex<Box<dyn LogExporter>>,
+pub struct SimpleLogProcessor<T: LogExporter> {
```
utpilla, line 80: Technically this is a breaking change, but I doubt it would impact anyone. It's probably okay to push it for `0.27.1`.

PR #2336, opened 2024-11-24
More consistent readme for msrv

cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-rust/issues/1306
This PR fixes some inconsistencies, but not all. There'll be further follow ups to make everything consistent.

PR #2341, opened 2024-11-25
Move docs from readme to docs.rs

cijothomas:
Continuing from https://github.com/open-telemetry/opentelemetry-rust/pull/2336.
Part of https://github.com/open-telemetry/opentelemetry-rust/issues/1306

Removes docs from the github readme.md file and moves everything into the docs.rs.
Readme.md is what gets rendered in crates.io, so I'd like to get some feedback from others on how much info we need to put there vs docs.rs... 
We want to avoid duplicating information as much as possible.


opentelemetry/src/lib.rs
```diff
@@ -161,6 +246,9 @@
 //! supported version is 1.70. The current OpenTelemetry version is not
 //! guaranteed to build on Rust versions earlier than the minimum supported
 //! version.
+//! This crate is built against the latest stable release. The minimum supported
```
cijothomas, line 249: lets put in crates.io also.

opentelemetry/src/lib.rs
```diff
@@ -5,6 +5,116 @@
 //! [`API`]: https://opentelemetry.io/docs/specs/otel/overview/#api
 //! [OpenTelemetry]: https://opentelemetry.io/docs/what-is-opentelemetry/
 //!
+//! [Jaeger]: https://www.jaegertracing.io/
+//! [Prometheus]: https://www.prometheus.io/
+//! 
+//! # Overview
```
cijothomas, line 11: put this in crates.io

hdost:
I feel like at least for the main `opentelemetry` it is good to have a more comprehensive page. Giving a good intro before needing to drop into the docs directly.

Issue #2328, opened 2024-11-23
Metrics Collect improvement for delta temporality


cijothomas:
During collect the hashmap is [cleared out entirely](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/internal/mod.rs#L176), forcing updates() to re-insert entries to the hashmap after each collect. Since inserts require Write lock and heap allocation, we should avoid unnecessarily causing insertion.

One rough idea how to achieve this:
1. Associate a status with each tracker indicating if it has a collect pending or not. Update need to set this.
2. During collect, remove those trackers which has no collect pending.
3. For trackers that have collect pending, it must be collected and status updated to no collect pending.

This does introduce some extra cost, but it'll help avoid insertion frequency for those trackers that are constantly updated (at least once in collect period), as their trackers will not be removed from hashmap. This is a desirable feature for Metrics, where once a tracker is reported (i.e a timeseries), as long as user is not providing new timeseries, there should be zero memory allocations and write locks. In other words, after initial hydration, the steady state should not require any memory allocations, and won't cause an update thread to get blocked trying to acquire a lock.

fraillt:
This has undesirable cost for cumulative temporality as well, because of the requirement to keep track of whether collect is pending or not.
However, it looks like temporality never changes for specific aggregation, which means that instead of having `Sum` aggregation with collection methods `cumulative` and `delta`, we could have `Sum<generic over temporality>` instead.
This approach would allow to have efficient update for cumulative, and efficient collect for delta.

cijothomas:
Removing from MetricSDK Stable milestone - this is not a blocker. The changes should be fully internal with no public API changes required.

cijothomas:
#3420 is merged, closing this.

PR #2343, opened 2024-11-26
Preallocate and keep memory for HashMap in Metric aggregation

cijothomas:
Discussed originally in https://github.com/open-telemetry/opentelemetry-rust/pull/2267#discussion_r1828595967
and in https://github.com/open-telemetry/opentelemetry-rust/pull/2288#discussion_r1834460750

2 changes
1. HashMap for aggregations are created with enough capacity to hold cardinality. (Its currently hardcoded, but will be configurable later)
2. Delta - during collect, the hashmap is swapped with a secondary hashmap., and the original is drained.

These ensures HashMap never requires re-sizing in its lifetime.

Incremental changes only, as this is not quite the final state - we'll need to likely store the overflow outside of hashmap and also tackle https://github.com/open-telemetry/opentelemetry-rust/issues/2328

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -12,7 +12,7 @@ use super::{
     precomputed_sum::PrecomputedSum, sum::Sum, Number,
 };
 
-pub(crate) const STREAM_CARDINALITY_LIMIT: u32 = 2000;
+pub(crate) const STREAM_CARDINALITY_LIMIT: usize = 2000;
 
 /// Checks whether aggregator has hit cardinality limit for metric streams
 pub(crate) fn is_under_cardinality_limit(size: usize) -> bool {
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -16,7 +16,7 @@ pub(crate) const STREAM_CARDINALITY_LIMIT: usize = 2000;
 
 /// Checks whether aggregator has hit cardinality limit for metric streams
 pub(crate) fn is_under_cardinality_limit(size: usize) -> bool {
-    size < STREAM_CARDINALITY_LIMIT as usize
+    size < STREAM_CARDINALITY_LIMIT
 }
 
 /// Receives measurements to be aggregated.
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -12,7 +12,7 @@ use super::{
     precomputed_sum::PrecomputedSum, sum::Sum, Number,
 };
 
-const STREAM_CARDINALITY_LIMIT: u32 = 2000;
+pub(crate) const STREAM_CARDINALITY_LIMIT: u32 = 2000;
 
 /// Checks whether aggregator has hit cardinality limit for metric streams
 pub(crate) fn is_under_cardinality_limit(size: usize) -> bool {
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -72,14 +72,10 @@ where
 {
     fn new(config: A::InitConfig) -> Self {
         ValueMap {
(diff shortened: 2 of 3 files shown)


opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -12,7 +12,7 @@ use super::{
     precomputed_sum::PrecomputedSum, sum::Sum, Number,
 };
 
-const STREAM_CARDINALITY_LIMIT: u32 = 2000;
+pub(crate) const STREAM_CARDINALITY_LIMIT: u32 = 2000;
```
utpilla, line 15: nit: Could we use `usize` to avoid this type casting in `ValueMap`: `STREAM_CARDINALITY_LIMIT as usize`?

```suggestion
pub(crate) const STREAM_CARDINALITY_LIMIT: usize = 2000;
```

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -170,16 +182,25 @@ where
             ));
         }
 
-        let trackers = match self.trackers.write() {
-            Ok(mut trackers) => {
+        let mut trackers = if let Ok(mut trackers_guard) = self.trackers.write() {
```
utpilla, line 185: I think we could make a few improvements here:
1. Take the write lock on `trackers_collect` before taking the write lock on `trackers`. That way we can reduce the amount of work done while holding the write lock on `trackers`.
2. I'm not sure how helpful or actionable the warnings on poisoned lock are. We could omit that and make the whole thing less verbose.
3. Use more intuitive variable names. For example, using `tracker_collect` instead of `trackers` to convey that collect is actually running on `tracker_collect`.

Here is an attempt to address the above-mentioned points:

```rust
if let Ok(mut trackers_collect) = self.trackers_for_collect.write() {
    if let Ok(mut trackers_current) = self.trackers.write() {
        swap(
            trackers_collect.deref_mut(),
            trackers_current.deref_mut(),
        );
    }

    for (attrs, tracker) in trackers_collect.drain() {
            dest.push(map_fn(attrs, tracker.clone_and_reset(&self.config)));
        }
    }
}
```
```

File: stress/Cargo.toml [hunk 3]
```diff
@@ -50,7 +50,7 @@ rand = { version = "0.8.4", features = ["small_rng"] }
 tracing = { workspace = true, features = ["std"]}
 tracing-subscriber = { workspace = true, features = ["registry", "std"] }
 num-format = "0.4.4"
-sysinfo = { version = "0.30.12", optional = true }
+sysinfo = { version = "0.32.0", optional = true }
```
lalitb, line 53: Seems this is not fixed in latest version too, though no hard me upgrading.

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: > Take the write lock on trackers_collect before taking the write lock on trackers. That way we can reduce the amount of work done while holding the write lock on trackers.

:100: thanks!

cijothomas: > I'm not sure how helpful or actionable the warnings on poisoned lock are. We could omit that and make the whole thing less verbose.

It is not user-actionable, but without the internal log, we'll be blind if the issue actually hits. We need to revisit all internal logs before stable, and this will be revisited along with many similar areas.
Similar to https://github.com/open-telemetry/opentelemetry-rust/pull/2331#discussion_r1855268026

cijothomas: made the changes suggested, and keeping internal logs for now. Resolving.

utpilla: > It is not user-actionable, but without the internal log, we'll be blind if the issue actually hits. We need to revisit all internal logs before stable, and this will be revisited along with many similar areas.

After thinking some more, since we know that the operations we perform under the write lock will not panic, there should be no scope of lock poisoning. So, it's okay to remove them entirely.

utpilla (approved):

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: Good point. We can use `.expect()` or `unwrap` in those places where we can sure of no poisoning.

PR #2346, opened 2024-11-26
Remove non-default feature from doc test

bantonsson:
## Changes

Running `cargo test --all` locally fails in the doc tests with:

```
---- opentelemetry-semantic-conventions/src/resource.rs - resource (line 16) stdout ----
error[E0425]: cannot find value `SERVICE_NAMESPACE` in module `semconv::resource`
    --> opentelemetry-semantic-conventions/src/resource.rs:25:42
     |
12   |         KeyValue::new(semconv::resource::SERVICE_NAMESPACE, "my-namespace"),
     |                                          ^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `SERVICE_NAME`
     |
```

This removes the `SERVICE_NAMESPACE` resource from the doc test, since it is now behind the non default feature `semconv_experimental`, making the doc test fail. It also updates the `metrics` template to use the `build()` method instead of the renamed `init()`

### DIFF
--- a/opentelemetry-semantic-conventions/src/resource.rs
+++ b/opentelemetry-semantic-conventions/src/resource.rs
@@ -21,7 +21,6 @@
 //! let _tracer = TracerProvider::builder()
 //!     .with_config(config().with_resource(Resource::new(vec![
 //!         KeyValue::new(semconv::resource::SERVICE_NAME, "my-service"),
-//!         KeyValue::new(semconv::resource::SERVICE_NAMESPACE, "my-namespace"),
 //!     ])))
 //!     .build();
 //! ```
--- a/opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
+++ b/opentelemetry-semantic-conventions/scripts/templates/registry/rust/metric.rs.j2
@@ -26,7 +26,7 @@
 //!     .u64_histogram(semconv::metric::HTTP_SERVER_REQUEST_DURATION)
 //!     .with_unit("By")
 //!     .with_description("Duration of HTTP server requests.")
-//!     .init();
+//!     .build();
 //! ```
 
 {% for root_ns in ctx %}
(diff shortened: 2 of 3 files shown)


opentelemetry-semantic-conventions/src/resource.rs
```diff
@@ -21,7 +21,6 @@
 //! let _tracer = TracerProvider::builder()
 //!     .with_config(config().with_resource(Resource::new(vec![
 //!         KeyValue::new(semconv::resource::SERVICE_NAME, "my-service"),
-//!         KeyValue::new(semconv::resource::SERVICE_NAMESPACE, "my-namespace"),
```
cijothomas, line 24: this was auto-generated after templates were changed, right?

cijothomas (approved):
Thanks. Need one more eyes to help review before merge.

lalitb (approved):

opentelemetry-semantic-conventions/src/resource.rs
lalitb: Any discrepancy in the committed and auto-generated files would have been caught in CI. So should be good.

bantonsson: Are you sure about that? The `init()` to `build()` change was a correction of a missed change in the `metrics` template, and that had not been caught by CI.

lalitb: Oh sorry, got confused with the proto generated file comparison, the semconv files discrepancy is not caught. @bantonsson  - Can you regenerate the files in new PR.

bantonsson: I did regenerate the files before submitting this PR. That’s how I caught the mismatch in the `metrics` files.

lalitb: Got it, yes directly modifying the `metrics.rs` was unfortunate and should have been caught in the CI (the way we do it for proto).

PR #2349, opened 2024-11-26
Deprecate logger::provider() and logger::instrumentation_scope()

lalitb:
## Changes

Deprecating these methods. The methods were meant to be used by appenders, and they can always keep hold to the `provider` handle to get this data. We can anyway have additive changes even after stable release, so safe to remove it for now. Thanks for noticing them @utpilla :)

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -240,7 +240,7 @@ impl Logger {
 
     #[deprecated(
         since = "0.27.1",
-        note = "This method is intended for appender developers, but has no defined use-case in typical workflows. It is deprecated and will be removed in the next major release."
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -238,11 +238,19 @@ impl Logger {
         Logger { scope, provider }
     }
 
+    #[deprecated(
+        since = "0.27.1",
+        note = "This method is intended for appender developers, but has no defined use-case in typical workflows. It is deprecated and will be removed in the next major release."
```
cijothomas, line 243:
```suggestion
        note = "This method was intended for appender developers, but has no defined use-case in typical workflows. It is deprecated and will be removed in the next major release."
```

cijothomas (approved):

utpilla (approved):

PR #2350, opened 2024-11-26
Add few internal logs for Metrics sdks

cijothomas:
Modified self-diag example to not do anything invalid to trigger logs, as we have sufficient logs otherwise to demonstrate!
Also few logs inside Metric sdk.

### DIFF
--- a/examples/self-diagnostics/src/main.rs
+++ b/examples/self-diagnostics/src/main.rs
@@ -1,6 +1,7 @@
 use opentelemetry::global;
 use opentelemetry::KeyValue;
 use opentelemetry_sdk::metrics::PeriodicReader;
+use opentelemetry_sdk::Resource;
 use std::error::Error;
 use tracing::info;
 use tracing_subscriber::fmt;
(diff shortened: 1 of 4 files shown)


examples/self-diagnostics/src/main.rs
```diff
@@ -13,6 +13,9 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
     let reader = PeriodicReader::builder(exporter, opentelemetry_sdk::runtime::Tokio).build();
 
     let provider = opentelemetry_sdk::metrics::SdkMeterProvider::builder()
+        .with_resource(opentelemetry_sdk::resource::Resource::new(vec![
```
utpilla, line 16: There is no need to allocate a `vec` here. Since this is an example, it's good to showcase good practices.

```suggestion
        .with_resource(opentelemetry_sdk::resource::Resource::new([
```

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -174,6 +174,10 @@ impl MeterProvider for SdkMeterProvider {
 
     fn meter_with_scope(&self, scope: InstrumentationScope) -> Meter {
         if self.inner.is_shutdown.load(Ordering::Relaxed) {
+            otel_debug!(
+                name: "MeterProvider.NoOpMeterReturned",
```
utpilla, line 178: Since this event is used at two different places, it would be nice to add a message field to distinguish between the two. Message for this one could be something like: "MeterProvider already shutdown." and for the other one it could be "Mutex for meters is poisoned."

utpilla (approved):
Left some suggestions.

PR #2351, opened 2024-11-26
Bugfix - add validation for custom buckets provided for Histograms

### LINKED ISSUE #2286: Validate user provided Histogram bounds at sdk during Instrument creation time
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/meter.rs#L385 is only validating name,unit etc. We should validate the buckets at this point, and if validation fails, we should either
1. Return no-op instrument as do when other validation fails.
2. Use default bounds, and ignore the user-provided one.

Emit WARN internal log in either case.

cijothomas:
Fixes #2286

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter.rs
+++ b/opentelemetry-sdk/src/metrics/meter.rs
@@ -397,6 +397,9 @@ impl SdkMeter {
         if let Some(ref boundaries) = builder.boundaries {
             let validation_result = validate_buckets(boundaries);
             if let Err(err) = validation_result {
+                // TODO: Include the buckets too in the error message.
+                // TODO: This validation is not done when Views are used to
+                // provide boundaries, and that should be fixed.
                 otel_error!(
                     name: "InstrumentCreationFailed",
                     meter_name = self.scope.name(),
--- a/opentelemetry-sdk/src/metrics/meter.rs
+++ b/opentelemetry-sdk/src/metrics/meter.rs
@@ -394,6 +394,20 @@ impl SdkMeter {
             return Histogram::new(Arc::new(NoopSyncInstrument::new()));
         }
 
+        if let Some(ref boundaries) = builder.boundaries {
+            let validation_result = validate_buckets(boundaries);
+            if let Err(err) = validation_result {
+                otel_error!(
+                    name: "InstrumentCreationFailed",
+                    meter_name = self.scope.name(),
+                    instrument_name = builder.name.as_ref(),
+                    message = "Measurements from this Histogram will be ignored.",
+                    reason = format!("{}", err)
+                );
+                return Histogram::new(Arc::new(NoopSyncInstrument::new()));
+            }
+        }
+
         match resolver
             .lookup(
                 InstrumentKind::Histogram,
@@ -531,6 +545,25 @@ impl InstrumentProvider for SdkMeter {
 
 fn validate_instrument_config(name: &str, unit: &Option<Cow<'static, str>>) -> MetricResult<()> {
     validate_instrument_name(name).and_then(|_| validate_instrument_unit(unit))
+}
+
+fn validate_buckets(buckets: &[f64]) -> MetricResult<()> {
+    if buckets.is_empty() {
+        return Err(MetricError::InvalidInstrumentConfiguration(
+            "Buckets must not be empty",
+        ));
+    }
+
+    // validate that buckets are sorted and non-duplicate
+    for window in buckets.windows(2) {
+        if window[0] >= window[1] {
+            return Err(MetricError::InvalidInstrumentConfiguration(
+                "Buckets must be sorted and non-duplicate",
+            ));
+        }
+    }
+
+    Ok(())
 }
 
 fn validate_instrument_name(name: &str) -> MetricResult<()> {
--- a/opentelemetry-sdk/src/metrics/meter.rs
+++ b/opentelemetry-sdk/src/metrics/meter.rs
@@ -555,7 +555,7 @@ fn validate_bucket_boundaries(boundaries: &[f64]) -> MetricResult<()> {
     for boundary in boundaries {
         if boundary.is_nan() || boundary.is_infinite() {
             return Err(MetricError::InvalidInstrumentConfiguration(
-                "Buckets must not contain NaN, +Inf, or -Inf",
+                "Bucket boundaries must not contain NaN, +Inf, or -Inf",
             ));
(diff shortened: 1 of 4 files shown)


lalitb (approved):

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -533,6 +550,25 @@ fn validate_instrument_config(name: &str, unit: &Option<Cow<'static, str>>) -> M
     validate_instrument_name(name).and_then(|_| validate_instrument_unit(unit))
 }
 
+fn validate_buckets(buckets: &[f64]) -> MetricResult<()> {
+    if buckets.is_empty() {
+        return Err(MetricError::InvalidInstrumentConfiguration(
```
utpilla, line 555: We should allow empty buckets for histogram, right?

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -533,6 +550,25 @@ fn validate_instrument_config(name: &str, unit: &Option<Cow<'static, str>>) -> M
...
+            "Buckets must not be empty",
+        ));
+    }
+
+    // validate that buckets are sorted and non-duplicate
+    for window in buckets.windows(2) {
```
utpilla, line 561: I think it's more readable (also a bit more efficient) to use a simple `for` loop here.

```rust
if let Some(boundaries) = builder.boundaries {
    for i in 1..boundaries.len() {
        if boundaries[i] <= boundaries[i - 1] {
            return Err(MetricError::InvalidInstrumentConfiguration(
                "Buckets must be sorted and non-duplicate",
            ));
        }
    }
}
```

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -533,6 +550,25 @@ fn validate_instrument_config(name: &str, unit: &Option<Cow<'static, str>>) -> M
     validate_instrument_name(name).and_then(|_| validate_instrument_unit(unit))
 }
 
+fn validate_buckets(buckets: &[f64]) -> MetricResult<()> {
```
utpilla, line 553: We should check for `f64::NAN`, `f64::INFINITY`, and `f64::NEG_INFINITY` as well.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -178,6 +178,27 @@ mod tests {
...
+            let test_context = TestContext::new(Temporality::Cumulative);
+            let histogram = test_context
+                .meter()
+                .f64_histogram("test")
+                .with_boundaries(buckets)
+                .build();
+            histogram.record(1.9, &[]);
+            test_context.flush_metrics();
+
+            // As buckets via Advisory params are invalid, no metrics should be
```
utpilla, line 198:
```suggestion
            // As buckets provided via Advisory params are invalid, no metrics should be
```

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -394,6 +394,23 @@ impl SdkMeter {
             return Histogram::new(Arc::new(NoopSyncInstrument::new()));
         }
 
+        if let Some(ref boundaries) = builder.boundaries {
```
utpilla, line 397: Let's remove the validation checks from the `Histogram` struct now that we would be checking it during instrument creation.

https://github.com/open-telemetry/opentelemetry-rust/blob/8e6b479684d550fb70b570d059080488b893aaf4/opentelemetry-sdk/src/metrics/internal/histogram.rs#L78-L79

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -178,6 +178,29 @@ mod tests {
             // As instrument name is invalid, no metrics should be exported
             test_context.check_no_metrics();
         }
+
+        let invalid_bucket_boundaries = vec![
+            vec![1.0, 1.0],                          // duplicate boundaries
+            vec![1.0, 2.0, 3.0, 2.0],                // duplicate non consequent boundaries
+            vec![1.0, 2.0, 3.0, 4.0, 2.5],           // unsorted boundaries
+            vec![1.0, 2.0, 3.0, f64::INFINITY, 4.0], // boundaries with NaNs
```
utpilla, line 186:
```suggestion
            vec![1.0, 2.0, 3.0, f64::INFINITY, 4.0], // boundaries with positive infinity
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -178,6 +178,29 @@ mod tests {
...
+            vec![1.0, 2.0, 3.0, f64::NAN],           // boundaries with NaNs
+            vec![f64::NEG_INFINITY, 2.0, 3.0],       // boundaries with NaNs
```
utpilla, line 188:
```suggestion
            vec![f64::NEG_INFINITY, 2.0, 3.0],       // boundaries with negative infinity
```

opentelemetry-sdk/src/metrics/meter.rs
cijothomas: The view route is unvalidated still I think. Something we need to comeback to and fix.

utpilla: Ah yes.

utpilla (approved):

opentelemetry/src/metrics/instruments/mod.rs
```diff
@@ -87,7 +87,26 @@ impl<'a, T> HistogramBuilder<'a, T> {
...
+    /// # Notes
+    /// - Boundaries must not contain `f64::NAN`, `f64::INFINITY` or
+    ///   `f64::NEG_INFINITY`
+    /// - Values must be in strictly increasing order (e.g., each value must be
+    ///   greater than the previous).
+    /// - Boundaries must not contain duplicate values.
+    ///
+    /// If invalid boundaries are provided, the instrument will not report
+    /// measurements.
+    /// Providing an empty `vec![]` means no bucket information will be
```
utpilla, line 102: With the current implementation, we would still count and export the bucket information. It would just be one bucket though. We should probably change the implementation to avoid bucket counting altogether in case of empty bounds.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2354

PR #2353, opened 2024-11-27
Few doc updates for opentelemetry

cijothomas:
Trying to see if this is better, where we provide most details in the `opentelemetry` crates' doc page.

### DIFF
--- a/opentelemetry/src/lib.rs
+++ b/opentelemetry/src/lib.rs
@@ -7,7 +7,6 @@
 //!
 //! # Getting Started with Traces
 //!
-//!
 //! The [`trace`] module includes types for tracking the progression of a single
 //! request while it is handled by services that make up an application. A trace
 //! is a tree of [`Span`]s which are objects that represent the work being done
--- a/opentelemetry/src/lib.rs
+++ b/opentelemetry/src/lib.rs
@@ -160,7 +160,7 @@
 //! * `logs`: Includes the logs bridge API.
 //! * `internal-logs`: Includes internal logging for the OpenTelemetry library via `tracing`.
 //!
-//! The default feature flags are ["trace", "metrics", "logs", "internal-loggs"].
+//! The default feature flags are ["trace", "metrics", "logs", "internal-logs"].
 //!
 //! The following feature flags provides additional configuration for `logs`:
 //! * `spec_unstable_logs_enabled`: Allow users to control the log level
--- a/opentelemetry/src/lib.rs
+++ b/opentelemetry/src/lib.rs
@@ -54,7 +54,7 @@
 //! 3. **Record Measurements:** Use the instruments to record measurement values
 //!    along with optional attributes.
 //!
-//! ## How Metrics Work in OpenTelemetry
+//! ## How Metrics work in OpenTelemetry
 //! In OpenTelemetry, raw measurements recorded using instruments are
 //! **aggregated in memory** to form metrics. These aggregated metrics are
 //! periodically exported by the [`opentelemetry_sdk`] at fixed intervals (e.g.,
@@ -91,6 +91,8 @@
 //! - An **ObservableUpDownCounter** can monitor the size of an in-memory queue
 //!   by reporting the size using queue's len() method within the callback
 //!   function.
+//! - An **ObservableGauge** can monitor the CPU temperature by using
+//!   temperature sensor APIs within the callback function.
 //!   
 //! For detailed guidance, refer to [OpenTelemetry Metrics API - Instrumentation
 //! Guidance](https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/#instrument-selection).
@@ -109,19 +111,19 @@
 //! let meter = global::meter("my_service");
 //!
 //! // Create an instrument (in this case, a Counter).
-//! let counter = meter.u64_counter("my_counter").build();
+//! let counter = meter.u64_counter("request.count").build();
 //!
 //! // Record a measurement by passing the value and a set of attributes.
 //! counter.add(1, &[KeyValue::new("http.client_ip", "83.164.160.102")]);
 //!
 //! // Create an ObservableCounter and register a callback that reports the measurement.
 //! let _observable_counter = meter
-//! .u64_observable_counter("my_observable_counter")
+//! .u64_observable_counter("bytes_received")
 //! .with_callback(|observer| {
 //!     observer.observe(
 //!         100,
 //!         &[
-//!             KeyValue::new("mykey1", "myvalue1"),
+//!             KeyValue::new("protocol", "udp"),
 //!         ],
 //!     )
 //! })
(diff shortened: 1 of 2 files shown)


opentelemetry/src/lib.rs
```diff
@@ -109,14 +158,15 @@
...
 //!
-//! The default feature flags are ["trace", "metrics", "logs"]
+//! The default feature flags are ["trace", "metrics", "logs", "internal-logs"].
 //!
 //! The following feature flags provides additional configuration for `logs`:
 //! * `spec_unstable_logs_enabled`: Allow users to control the log level
 //!
 //! The following feature flags enable APIs defined in OpenTelemetry specification that is in experimental phase:
-//! * `otel_unstable`: Includes unstable APIs.
+//! * `otel_unstable`: Includes unstable APIs. There are no features behind this flag at the moment.
```
lalitb, line 169: nit - not related to this PR, but line 171 should be under separate section, and not subsection of `logs`.

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
-//!
-//! The [`metrics`] module includes types for recording measurements about a
-//! service at runtime.
-//!
-//! ### Creating instruments and recording measurements
-//!
+//! # Getting Started with Metrics
+//!
+//! The [`metrics`] module provides types for recording measurements about a
+//! service at runtime. Below are the key steps to report measurements using
```
utpilla, line 48: "Service" is an overloaded term. We should remove "about a service" from the sentence here to just say: `The `[`metrics`]` module provides types for recording measurements about a service at runtime. `

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
+//! OpenTelemetry Metrics:
+//!
+//! 1. **Obtain a Meter:** Get a `Meter` from a `MeterProvider`.
+//! 2. **Create Instruments:** Use the `Meter` to create one or more instruments
+//!    (e.g., counters, histograms).
+//! 3. **Record Measurements:** Use the instruments to record measurement values
+//!    along with optional attributes.
+//!
+//! ## How Metrics Work in OpenTelemetry
```
utpilla, line 57:
```suggestion
//! ## How Metrics work in OpenTelemetry
```

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
+//! In OpenTelemetry, raw measurements recorded using instruments are
+//! **aggregated in memory** to form metrics. These aggregated metrics are
+//! periodically exported by the [`opentelemetry_sdk`] at fixed intervals (e.g.,
```
utpilla, line 60: Since this is a general introduction, should we talk about pull based exporters as well?

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
+//! - Use **Histograms** for measuring the distribution of a value, such as
+//!   response times or payload sizes.
+//!
+//! ### Observable Instruments
+//!
+//! Counters, UpDownCounters, and Gauges have Observable variants that allow
+//! values to be reported through a callback function. Observable instruments
+//! are ideal when the metric value is managed elsewhere and needs to be
+//! observed by OpenTelemetry instrumentation. The callbacks are automatically
+//! invoked by the OpenTelemetry SDK before every export (e.g., every 60
```
utpilla, line 85:
```suggestion
//! invoked by the OpenTelemetry SDK before every export, force_flush and shutdown.
```

opentelemetry/src/lib.rs
cijothomas: in future iteration, yes.

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
-//! # }
+//!
+//! // Create an ObservableCounter and register a callback that reports the measurement.
+//! let _observable_counter = meter
+//! .u64_observable_counter("my_observable_counter")
+//! .with_callback(|observer| {
+//!     observer.observe(
+//!         100,
+//!         &[
+//!             KeyValue::new("mykey1", "myvalue1"),
```
utpilla, line 124: It would be nice to stay consistent with the sample attributes: either use specific real-world examples (such as IP Address etc. used for Counter, Traces) or use generic examples such as `mykey1, myvalue1`. I would prefer using specific examples.

opentelemetry/src/lib.rs
cijothomas: Good point. i'll keep iterating to make it more thorough, in follow up PRs.

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
+//! observed by OpenTelemetry instrumentation. The callbacks are automatically
+//! invoked by the OpenTelemetry SDK before every export (e.g., every 60
+//! seconds).
+//!
+//! For example:
+//! - An **ObservableCounter** can monitor the number of page faults in a
+//!   process as reported by the operating system.
+//! - An **ObservableUpDownCounter** can monitor the size of an in-memory queue
+//!   by reporting the size using queue's len() method within the callback
+//!   function.
```
lalitb, line 93: Also add the example for `ObservableGauge` otherwise it will feel alienated :)

lalitb (approved):
Thanks. These are ongoing improvements, so we can keep revisiting them as needed.

opentelemetry/src/lib.rs
```diff
@@ -55,40 +35,109 @@
...
+//! [examples](https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples/metrics-basic)
+//! directory that show a runnable example with all type of instruments.
+//!
+//!
 //! See the [`metrics`] module docs for more information on creating and
 //! managing instruments.
 //!
 //!
-//! # Logs
+//! # Getting Started with Logs
```
utpilla, line 140: We could add a code sample for Logs as well similar to Metrics and Traces.

utpilla (approved):
It's a good start. Further improvements can be made in follow-up PRs.

opentelemetry/src/lib.rs
```diff
@@ -5,35 +5,15 @@
...
-//! }
-//! # }
-//! ```
-//!
-//! See the [examples] directory for different integration patterns.
-//!
-//! [examples]: https://github.com/open-telemetry/opentelemetry-rust/tree/main/examples
-//!
-//! # Traces
 //!
```
utpilla, line 10: This is an extra new line that can be removed.

opentelemetry/src/lib.rs
cijothomas: Agree. It is copied from otel official website. I'll address it in the future.

opentelemetry/src/lib.rs
cijothomas: we'll need to revisit how to do that, as for logs, we just ask people to use existing tracing or log apis....

PR #2356, opened 2024-11-27
Use separate names for examples named basic

bantonsson:
## Changes
Renames two examples both named `basic.rs`, since building and running the examples will give you warnings (and not even work) because there are two example targets named `basic`.

Running `cargo build --examples` gives this:
```
warning: output filename collision.
The example target `basic` in package `opentelemetry-stdout v0.27.0 (/.../opentelemetry-rust/opentelemetry-stdout)` has the same output filename as the example target `basic` in package `opentelemetry-appender-tracing v0.27.0 (/.../opentelemetry-rust/opentelemetry-appender-tracing)`.
Colliding filename is: /.../opentelemetry-rust/target/debug/examples/basic
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

And running `cargo run --example basic` fails with:

```
error: `cargo run` can run at most one executable, but multiple were specified
```
(


lalitb:
Thanks can we also modify the comments e.g,-https://github.com/open-telemetry/opentelemetry-rust/blob/53e892a2ca9f831fb09752f9ce223e26836163af/opentelemetry-stdout/examples/basic.rs#L1

cijothomas:
I am neutral to this...The [issue linked](https://github.com/rust-lang/cargo/issues/6313) is open from 2018.

When inside the directory, running `cargo run --example basic` is easy to remember than `cargo run --example stdout-basic`

Happy to hear other thoughts on this..

bantonsson:
This is not a big thing. I just thought that the name collisions were a bad thing based on the warning and behavior, and there was already a precedent with the `logs-basic` example not being named `basic`.

For me, my shell will tab complete the existing example names when I write `cargo run --example <TAB>`, so remembering the name is not much of an issue.

I'll update the comments if you think the name change is ok, otherwise just close this PR.

PR #2357, opened 2024-11-27
Track dropped spans and logs due to full buffer

### LINKED ISSUE #2273: Internal logging in BatchProcessor when buffer full
Opening a separate issue to specifically tracking internal logging when BatchProcessor gets full.
https://github.com/open-telemetry/opentelemetry-rust/issues/2175#issuecomment-2394816345

1. We certainly don't want to log each time the buffer is full
2. A good balance might be to keep a count of items-dropped-due-to-buffer-full.
3. This count can be exposed as a Metric itself.
4. And/Or this count can be emitted as a internal log upon shutdown/drop with severity WARN if non-zero items were dropped.

2,4 is what OTel .NET has implemented: See https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/BatchExportProcessor.cs#L102 and https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/BatchExportProcessor.cs#L207

scottgerring:
Fixes #2273

## Changes

As suggested in #2273, this change adds a counter and a metric for dropped spans and logs; the counter is used to log total drop count at exporter shutdown, the metric is used to export this information.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -156,7 +156,6 @@ pub struct BatchLogProcessor<R: RuntimeChannel> {
     message_sender: R::Sender<BatchMessage>,
 
     // Track dropped logs. We'll log this at shutdown and also emit
-    // as a metric.
     dropped_logs_count: AtomicUsize,
 
     // Track the maximum queue size that was configured for this processor
@@ -180,11 +179,11 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
 
         // TODO - Implement throttling to prevent error flooding when the queue is full or closed.
         if result.is_err() {
-            // Increment dropped logs counter and metric. The first time we have top drop a log,
+            // Increment dropped logs count. The first time we have to drop a log,
             // emit a warning.
             if self.dropped_logs_count.fetch_add(1, Ordering::Relaxed) == 0 {
                 otel_warn!(name: "BatchLogProcessor.LogDroppingStarted",
-                    message = "Beginning to drop log messages due to full exporter queue.");
+                    message = "BatchLogProcessor dropped a LogRecord due to queue full/internal errors. No further log will be emitted for further drops until Shutdown. During Shutdown time, a log will be emitted with exact count of total logs dropped.");
             }
         }
     }
@@ -208,7 +207,7 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
                 name: "BatchLogProcessor.LogsDropped",
                 dropped_logs_count = dropped_logs,
                 max_queue_size = max_queue_size,
-                message = "Logs were dropped due to a full or closed queue. The count represents the total count of lost logs in the lifetime of the BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
+                message = "Logs were dropped due to a queue being full or other error. The count represents the total count of lost dropped in the lifetime of this BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
             );
         }
 
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -207,7 +207,7 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
                 name: "BatchLogProcessor.LogsDropped",
                 dropped_logs_count = dropped_logs,
                 max_queue_size = max_queue_size,
-                message = "Logs were dropped due to a queue being full or other error. The count represents the total count of lost dropped in the lifetime of this BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
+                message = "Logs were dropped due to a queue being full or other error. The count represents the total count of log records dropped in the lifetime of this BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
             );
         }
 
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -11,9 +11,11 @@ use futures_util::{
 };
 #[cfg(feature = "spec_unstable_logs_enabled")]
 use opentelemetry::logs::Severity;
-use opentelemetry::{otel_debug, otel_error, otel_warn, InstrumentationScope};
+use opentelemetry::{
+    global, metrics::Counter, otel_debug, otel_error, otel_warn, InstrumentationScope,
+};
 
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::{cmp::min, env, sync::Mutex};
 use std::{
     fmt::{self, Debug, Formatter},
@@ -154,6 +156,11 @@ impl LogProcessor for SimpleLogProcessor {
 /// them at a pre-configured interval.
 pub struct BatchLogProcessor<R: RuntimeChannel> {
     message_sender: R::Sender<BatchMessage>,
+
+    // Track dropped logs. We'll log this at shutdown and also emit
+    // as a metric.
+    dropped_logs_count: AtomicUsize,
+    dropped_logs_metric: Counter<u64>,
 }
 
 impl<R: RuntimeChannel> Debug for BatchLogProcessor<R> {
@@ -172,11 +179,10 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
         )));
 
         // TODO - Implement throttling to prevent error flooding when the queue is full or closed.
-        if let Err(err) = result {
-            otel_error!(
-                name: "BatchLogProcessor.Export.Error",
-                error = format!("{}", err)
-            );
+        if result.is_err() {
+            // Increment dropped logs counter and metric
+            self.dropped_logs_count.fetch_add(1, Ordering::Relaxed);
+            self.dropped_logs_metric.add(1, &[]);
         }
     }
 
@@ -192,6 +198,15 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
     }
 
     fn shutdown(&self) -> LogResult<()> {
+        let dropped_logs = self.dropped_logs_count.load(Ordering::Relaxed);
+        if dropped_logs > 0 {
+            otel_warn!(
+                name: "BatchLogProcessor.Shutdown",
+                dropped_logs = dropped_logs,
+                message = "Logs were dropped due to a full buffer."
+            );
+        }
+
         let (res_sender, res_receiver) = oneshot::channel();
         self.message_sender
             .try_send(BatchMessage::Shutdown(res_sender))
@@ -296,8 +311,18 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                 }
             }
         }));
+
+        let dropped_logs_metric = global::meter("opentelemetry")
+            .u64_counter("dropped_logs")
+            .with_description("Number of logs dropped due to full buffer")
+            .build();
+
         // Return batch processor with link to worker
-        BatchLogProcessor { message_sender }
+        BatchLogProcessor {
+            message_sender,
+            dropped_logs_count: AtomicUsize::new(0),
+            dropped_logs_metric,
+        }
     }
 
     /// Create a new batch processor builder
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -296,8 +311,18 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                 }
             }
         }));
+
+        let dropped_logs_metric = global::meter("opentelemetry")
```
cijothomas, line 315: I'd suggest holding off from exposing these via OTel metrics.
1. Fetching meter from global only works if user has already set a global meter provider. There is no document guiding them to setup Metrics signal before logs/traces, so by default, this is not guaranteed to work.
2. We cannot guarantee the name of the metric nor the name of the attributes. OTel spec/convention is planning to add one for the same, and we need to wait until it lands (We can offer OTel Rust as one of the language to prototype this to support spec/semantic convention)

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -172,11 +179,10 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
         )));
 
         // TODO - Implement throttling to prevent error flooding when the queue is full or closed.
-        if let Err(err) = result {
-            otel_error!(
-                name: "BatchLogProcessor.Export.Error",
-                error = format!("{}", err)
-            );
+        if result.is_err() {
```
cijothomas, line 182: for the very first time, lets also emit an internal log informing the user about data drop.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -192,6 +198,15 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
     }
 
     fn shutdown(&self) -> LogResult<()> {
+        let dropped_logs = self.dropped_logs_count.load(Ordering::Relaxed);
+        if dropped_logs > 0 {
+            otel_warn!(
+                name: "BatchLogProcessor.Shutdown",
```
cijothomas, line 204:
```suggestion
                name: "BatchLogProcessor.LogsDropped",
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -192,6 +198,15 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                dropped_logs = dropped_logs,
```
cijothomas, line 205:
```suggestion
                dropped_logs_count= dropped_logs,
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -192,6 +198,15 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                message = "Logs were dropped due to a full buffer."
```
cijothomas, line 206:
```suggestion
                message = "Logs were dropped due to full buffer. The count represents the total count of lost logs in the lifetime of the BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -192,6 +198,15 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+            );
```
cijothomas, line 207: would be nice to also include an attribute for the max queue size the batchprocessor was operating with.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas, line 206: I am unsure if it is correct to state that logs were dropped due to Buffer full. It looks like there could be other reasons too?

cijothomas (changes requested):
Thanks for working on this! Left my comments in the PR.
OTel internal metrics is something we can add, but not right away:
https://github.com/open-telemetry/opentelemetry-rust/pull/2357/files#r1860843464

opentelemetry-sdk/src/logs/log_processor.rs
scottgerring: Added this to each processor:

https://github.com/open-telemetry/opentelemetry-rust/blob/b626d7f5b00c336b0d1c263df9e43521269d8bee/opentelemetry-sdk/src/logs/log_processor.rs#L185-L188

opentelemetry-sdk/src/logs/log_processor.rs
scottgerring: Applied

opentelemetry-sdk/src/logs/log_processor.rs
scottgerring: I think the only other reason is the queue is closed; i've updated the messages to reflect that

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -154,6 +154,13 @@ impl LogProcessor for SimpleLogProcessor {
 /// them at a pre-configured interval.
 pub struct BatchLogProcessor<R: RuntimeChannel> {
     message_sender: R::Sender<BatchMessage>,
+
+    // Track dropped logs. We'll log this at shutdown and also emit
```
cijothomas, line 158: nit: remove the comment about emit as metric that was removed.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -154,6 +154,13 @@ impl LogProcessor for SimpleLogProcessor {
...
+    // as a metric.
+    dropped_logs_count: AtomicUsize,
+
+    // Track the maximum queue size that was configured for this processor
+    max_queue_size: usize,
```
cijothomas, line 163: odd that we have to store this here just for logging purposes, but not an issue!

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -192,6 +201,17 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
 
     fn shutdown(&self) -> LogResult<()> {
+        let dropped_logs = self.dropped_logs_count.load(Ordering::Relaxed);
+        let max_queue_size = self.max_queue_size;
+        if dropped_logs > 0 {
+            otel_warn!(
+                name: "BatchLogProcessor.LogsDropped",
+                dropped_logs_count = dropped_logs,
+                max_queue_size = max_queue_size,
+                message = "Logs were dropped due to a full or closed queue. The count represents the total count of lost logs in the lifetime of the BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
```
cijothomas, line 211:
```suggestion
                message = "Logs were dropped due to a queue being full or other error. The count represents the total count of lost dropped in the lifetime of this BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
```

Given this is for end-users, I think we cannot use "closed queue", as that is strictly internal implementation that we use a channel which has the possibility of being closed. See if my above suggestion is good. Happy to get other feedback as well.
Also, we can improve the messaging in the future also based on feedbacks we get, so no need to block this PR.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -172,11 +179,13 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
         )));
 
         // TODO - Implement throttling to prevent error flooding when the queue is full or closed.
-        if let Err(err) = result {
-            otel_error!(
-                name: "BatchLogProcessor.Export.Error",
-                error = format!("{}", err)
-            );
+        if result.is_err() {
+            // Increment dropped logs counter and metric. The first time we have top drop a log,
```
cijothomas, line 183:
```suggestion
            // Increment dropped logs count. The first time we have to drop a log,
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -154,6 +154,13 @@ impl LogProcessor for SimpleLogProcessor {
 /// them at a pre-configured interval.
 pub struct BatchLogProcessor<R: RuntimeChannel> {
     message_sender: R::Sender<BatchMessage>,
+
+    // Track dropped logs. We'll log this at shutdown and also emit
+    // as a metric.
```
scottgerring, line 159:
```suggestion
```

cijothomas (approved):
Thanks! 
Left some suggestions for improved log messages.

opentelemetry-sdk/src/logs/log_processor.rs
scottgerring: LGTM. I defer entirely to your experience on the way we speak to users!

opentelemetry-sdk/src/logs/log_processor.rs
scottgerring:
```suggestion
// Track dropped logs - we'll log this at shutdown
```

scottgerring:
Changes applied - should be good to go once the builds are done !

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -192,6 +200,17 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
 
     fn shutdown(&self) -> LogResult<()> {
+        let dropped_logs = self.dropped_logs_count.load(Ordering::Relaxed);
+        let max_queue_size = self.max_queue_size;
+        if dropped_logs > 0 {
+            otel_warn!(
+                name: "BatchLogProcessor.LogsDropped",
+                dropped_logs_count = dropped_logs,
+                max_queue_size = max_queue_size,
+                message = "Logs were dropped due to a queue being full or other error. The count represents the total count of lost dropped in the lifetime of this BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
```
utpilla, line 210:
```suggestion
                message = "Logs were dropped due to a queue being full or other error. The count represents the total count of log records dropped in the lifetime of this BatchLogProcessor. Consider increasing the queue size and/or decrease delay between intervals."
```

cijothomas:
@lalitb @utpilla Could you take a look? If I get one more approval, I plan to include this for today's release (0.27.1)

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -249,11 +256,14 @@ impl<R: RuntimeChannel> SpanProcessor for BatchSpanProcessor<R> {
...
-                name: "BatchSpanProcessor.OnEnd.ExportQueueingFailed",
-                reason = format!("{:?}", TraceError::Other(err.into()))
-            );
+        // If the queue is full, and we can't buffer a span
+        if result.is_err() {
+            // Increment the number of dropped spans. If this is the first time we've had to drop,
+            // emit a warning.
+            if self.dropped_spans_count.fetch_add(1, Ordering::Relaxed) == 0 {
+                otel_warn!(name: "BatchSpanProcessor.SpanDroppingStarted",
+                    message = "Beginning to drop span messages due to full exporter queue.");
```
utpilla, line 265: Let's be consistent with the message in `BatchLogProcessor`.
```suggestion
                    message = "Beginning to drop span messages due to full/internal errors. No further log will be emitted for further drops until Shutdown. During Shutdown time, a log will be emitted with exact count of total spans dropped.");
```

utpilla (approved):
Left a suggestion to be consistent with the message content for both the batch processor.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: I believe we can avoid this by moving the logging of dropped logs (otel_warn!) from the shutdown() method to the worker's Shutdown message processing.

lalitb:
Would be good to include a unit test to validate that the dropped logs count is correctly reflected during shutdown. Since the field is private, we could consider adding a test-only accessor method to retrieve it. That said, if this is time-sensitive and needs to be included in today's release, can be done separately.

cijothomas:
> Would be good to include a unit test to validate that the dropped logs count is correctly reflected during shutdown. Since the field is private, we could consider adding a test-only accessor method to retrieve it. That said, if this is time-sensitive and needs to be included in today's release, can be done separately.

I think we have orchestrated such a test in OTel .NET, so we can copy some ideas.

cijothomas:
@scottgerring I edited the PR Title ("chore: Track dropped spans and logs due to full buffer"), to remove "chore" from it.

PR #2359, opened 2024-11-27
Add internal log ff to jaeger propagator

cijothomas:
CI is red, not sure why now!!


cijothomas:
this is not needed, sorry CI failure is different reason.

PR #2362, opened 2024-11-27
Bump version of sdk release 0.27.1

cijothomas:
I think only SDK crates need to be released to avoid doing empty-releases for other crates.
This has a lot of internal logs improvements.


lalitb (approved):

lalitb:
There are few nit changes in opentelemetry API too, if we want to have a patch for them.

PR #2367, opened 2024-11-30
Time is not optional in DataPoints

fraillt:
## Changes

Make `start_time` and `time` non optional in `DataPoint<T>` (used for `Sum` and `Gauge`).

### DIFF
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -97,9 +97,9 @@ pub struct DataPoint<T> {
     /// time series.
     pub attributes: Vec<KeyValue>,
     /// The time when the time series was started.
-    pub start_time: Option<SystemTime>,
+    pub start_time: SystemTime,
     /// The time when the time series was recorded.
-    pub time: Option<SystemTime>,
+    pub time: SystemTime,
     /// The value of this data point.
     pub value: T,
     /// The sampled [Exemplar]s collected during the time series.
@@ -338,8 +338,8 @@ mod tests {
     fn validate_cloning_data_points() {
         let data_type = DataPoint {
             attributes: vec![KeyValue::new("key", "value")],
-            start_time: Some(std::time::SystemTime::now()),
-            time: Some(std::time::SystemTime::now()),
+            start_time: std::time::SystemTime::now(),
+            time: std::time::SystemTime::now(),
             value: 0u32,
             exemplars: vec![Exemplar {
                 filtered_attributes: vec![],
--- a/opentelemetry-proto/src/transform/metrics.rs
+++ b/opentelemetry-proto/src/transform/metrics.rs
@@ -295,8 +295,8 @@ pub mod tonic {
                     .iter()
                     .map(|dp| TonicNumberDataPoint {
                         attributes: dp.attributes.iter().map(Into::into).collect(),
-                        start_time_unix_nano: dp.start_time.map(to_nanos).unwrap_or_default(),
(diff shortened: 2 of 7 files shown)


opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -97,9 +97,9 @@ pub struct DataPoint<T> {
     /// time series.
     pub attributes: Vec<KeyValue>,
     /// The time when the time series was started.
-    pub start_time: Option<SystemTime>,
+    pub start_time: SystemTime,
```
cijothomas, line 100: It'd be good to mention in changelog as this is breaking anyone writing own exporters. I expect a lot more changes in this area, so it is probably better to club them all into one changelog entry.

utpilla: Start time is marked optional by spec.

@cijothomas Currently, the SDK would always populate it, but do you think there could be a scenario where we wouldn't populate the start time. With MetricProducer or something else?

cijothomas: Yes, producer might not provide it. so keeping Option for startTime might be safer then?

utpilla: Yes.

cijothomas: @fraillt Could you keep Option<startTime> ? Sorry I merged this already, so needs to be a separate PR.

fraillt: I was reading https://opentelemetry.io/docs/specs/otel/metrics/data-model/ and got impression that these fields can be in fact not optional...
Basically everything is required, except for `Gauge` one field:
> (optional) A timestamp (start_time_unix_nano) which best represents the first possible moment a measurement could be recorded. This is commonly set to the timestamp when a metric collection system started.

In the proto [definition](https://github.com/open-telemetry/opentelemetry-proto/blob/c5c8b28012583fda55b0cb16f73a820722171d49/opentelemetry/proto/metrics/v1/metrics.proto#L388C6-L388C23) same field has comment (emphasis is mine):
>   // StartTimeUnixNano is optional but **strongly encouraged**, see the detailed comments above Metric.

Maybe I'm reading outdated/invalid documentation?

cijothomas: > (optional) A timestamp (start_time_unix_nano) which best represents the firs

^ This part is where startTime is mentioned to be optional. An external system interfacing via Producer to OTel may not have startTime.

fraillt: Just want to understand better... at the moment it's not possible that Gauge will have `None` using SdkMeterProvider.
So basically, if you want to support `Gauge` with optional start time, at minimum you would have to implement your own, `MeterProvider`, `InstrumentProvider`, and your own `SyncInstrument`. Which also means that you need to implement your own collect implementation as well.
I understand that this is part of `api` and user should be able to do so. I'm wondering if you know a single example where anyone actually does that ? (implement core functionality itself' instead of using SdkMeterProvider). 
Or in other words, do you have any example that `SdkMeterProvider` is not suitable and users write their own implementation of metrics ? (just curious)

cijothomas: When we start supporting Producers concept, where users aggregate metrics elsewhere and bridge it to OTel, startTime can be missing. 
The current sdk has no scenario for startTime being optional, but we just want to make sure our public API remains stable even when adding Producer support.

PR #2371, opened 2024-12-02
Directly implement Measure trait for metric aggregates

fraillt:
Prerequisite for #2328
This PR is similar to #2425, but it's for measurement instead of collection.

## Changes

* created `AttributeSetFilter` utility that (depending on provided filtering function) applies filter to attribute-sets before measurement. Similar logic as in `AggregateBuilder::filter`
* store `AttributeSetFilter` as part of aggregate 
* aggregates directly implement `Measure` trait.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -97,6 +97,31 @@ impl Default for AggregateTimeInitiator {
     }
 }
 
+type Filter = Arc<dyn Fn(&KeyValue) -> bool + Send + Sync>;
+
+/// Applies filter on provided attribute set
+/// No-op, if filter is not set
+#[derive(Clone)]
+pub(crate) struct AttributeSetFilter {
+    filter: Option<Filter>,
+}
+
+impl AttributeSetFilter {
+    pub(crate) fn new(filter: Option<Filter>) -> Self {
+        Self { filter }
+    }
+
+    pub(crate) fn apply(&self, attrs: &[KeyValue], run: impl FnOnce(&[KeyValue])) {
+        if let Some(filter) = &self.filter {
+            let filtered_attrs: Vec<KeyValue> =
+                attrs.iter().filter(|kv| filter(kv)).cloned().collect();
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -83,28 +94,24 @@ impl<T: Number> AggregateBuilder<T> {
...
-        let filter = self.filter.clone();
-        move |n, attrs: &[KeyValue]| {
-            if let Some(filter) = &filter {
-                let filtered_attrs: Vec<KeyValue> =
-                    attrs.iter().filter(|kv| filter(kv)).cloned().collect();
-                f.call(n, &filtered_attrs);
-            } else {
-                f.call(n, attrs);
-            };
+    fn maybe_filtered_measurement(&self, instrument: Arc<impl Measure<T>>) -> Arc<dyn Measure<T>> {
```
fraillt, line 97: Any advice on naming?
both for `maybe_filtered_measurement` function, and for `FilteredMeasureInstrument`...

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -97,77 +97,77 @@ impl Default for AggregateTimeInitiator {
...
+
+impl AttributeSetFilter {
+    pub(crate) fn new(filter: Option<Filter>) -> Self {
+        Self { filter }
+    }
+
+    pub(crate) fn apply(&self, attrs: &[KeyValue], run: impl FnOnce(&[KeyValue])) {
+        if let Some(filter) = &self.filter {
+            let filtered_attrs: Vec<KeyValue> =
+                attrs.iter().filter(|kv| filter(kv)).cloned().collect();
```
cijothomas, line 117: filtering is supposed to help reduce cost (including perf cost), but unfortunately, filtering path need heap allocations. Nothing introduced in this PR, but just mentioning as something we can fix in the future.

cijothomas (approved):
LGTM!
Wondering if there any measurable difference in performance?

PR #2374, opened 2024-12-02
Use native Rust support for async traits in LogExporter::export() method (11% improvement)

lalitb:
## Changes

Continuation to #2143, more to show feasibility, performance improvements, and have discussion before raising the PR for eventual review, this PR demonstrate using async traits support in Rust exporter. The support was added in Rust v1.75, while the msrv as of now is v1.70 for otel-sdk

The change is:
Existing:
```rust
#[async_trait]
  async fn export(&mut self, batch: LogBatch<'_>) -> LogResult<()>;
}
```

PR:
```rust
    fn export<'a>(
        &'a mut self,
        batch: &'a LogBatch<'a>,
    ) -> impl std::future::Future<Output = LogResult<()>> + Send + 'a;
}
```
There is 11%  improvement as seen from stress test:

Before PR:

Number of threads: 16
Throughput: 40,468,186 iterations/sec
Throughput: 40,792,480 iterations/sec
Throughput: 40,456,227 iterations/sec
Throughput: 40,455,461 iterations/sec

After this PR:

Throughput: 44,793,174 iterations/sec
Throughput: 45,321,168 iterations/sec
Throughput: 45,387,969 iterations/sec
Throughput: 45,629,648 iterations/sec
Throughput: 45,607,326 iterations/sec
Throughput: 45,303,331 iterations/sec
Throughput: 45,634,108 iterations/sec

Thanks,
Lalit


opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -82,7 +80,11 @@ pub trait LogExporter: Send + Sync + Debug {
     /// A `LogResult<()>`, which is a result type indicating either a successful export (with
     /// `Ok(())`) or an error (`Err(LogError)`) if the export operation failed.
     ///
-    async fn export(&mut self, batch: LogBatch<'_>) -> LogResult<()>;
+    fn export<'a>(
+        &'a mut self,
```
cijothomas, line 84: @lalitb I think it maybe better to first check how to avoid requiring `mut` reference for exporting, before evaluating the results of this perf test, as that requirement has forced the need of mutex and introduced contention and hence stress tests shows way low throughput than before.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -247,11 +247,16 @@ mod tests {
 
     #[async_trait]
```
cijothomas, line 248: we don't need async-trait anymore?

opentelemetry-sdk/src/export/logs/mod.rs
lalitb: @cijothomas If we remove the `mut` self-reference from the` LogExporter::export()` method, it will force the implementation to rely on interior mutability to modify its internal state. This would involve using synchronization mechanisms like locks or atomic operations, which add complexity and overhead. But in the case of `SimpleLogProcessor` and `BatchLogProcessor`, these methods are always called sequentially, so there’s no real benefit to introducing this extra layer of synchronization. It would just make the code more complicated and less performant without actually solving any problem.

The use-case for having a non-mutating export method would be in reentrant log processors, where a user thread could call `LogProcessor::emit()`, which in turn calls `LogProcessor::export()` concurrently from multiple threads. In such scenarios, interior mutability would be essential to safely handle concurrent access. However, this is not relevant for the current processors, which operate in a strictly sequential manner.


Actually, there are 4 different exporter scenarios to think about - 
 - sync/concurrent:
 ```rust
     fn export<'a>(&'a self, batch: &'a LogBatch<'a>) -> LogResult<()>;
```

 - sync/sequential, 
 ```rust
    fn export<'a>(&'a mut self, batch: &'a LogBatch<'a>) -> LogResult<()>;
```

 - async/concurrent
 ```rust
   fn export<'a>(&'a self, batch: &'a LogBatch<'a>) -> impl std::future::Future<Output = LogResult<()>> + Send + 'a;
```

 - async/sequential 
 ```rust
  fn export<'a>(&'a mut self, batch: &'a LogBatch<'a>) -> impl std::future::Future<Output = LogResult<()>> + Send + 'a;
```

Currently:

- The `SimpleLogProcessor` and `BatchLogProcessor` invoke the exporter in `async/sequential` mode.
- The `ReentrantProcessor`, used by ETW and user_events in the contrib repository, operates the exporter in `sync/concurrent` mode.

To ensure clarity and extensibility, we need to properly model these scenarios within the design.

opentelemetry-appender-tracing/src/layer.rs
lalitb: yes, this can be safely removed.

opentelemetry-sdk/src/export/logs/mod.rs
lalitb: Just to add, we don't have currently have specific use-case for `sync/sequential`, and `async/concurrent`, they are mentioned above for completeness of all the scenarios.

lalitb:
> @lalitb I think it maybe better to first check how to avoid requiring mut reference for exporting, before evaluating the results of this perf test, as that requirement has forced the need of mutex and introduced contention and hence stress tests shows way low throughput than before.

OK, have updated the PR after removing the `mut self` in #2380. This eliminates the need of mutex for exporter in stress test. The stress result are now similar to earlier i.e., earlier without the inclusion of exporter. Which means, adding `futures::block_on()` on native async-method doesn't introduce much of latency.

Before this PR (the regression introduced by #2380 because of inclusion of exporter):
```
Number of threads: 16
Throughput: 40,468,186 iterations/sec
Throughput: 40,792,480 iterations/sec
Throughput: 40,456,227 iterations/sec
Throughput: 40,455,461 iterations/sec
```

After this PR:
```
Throughput: 44,793,174 iterations/sec
Throughput: 45,321,168 iterations/sec
Throughput: 45,387,969 iterations/sec
Throughput: 45,629,648 iterations/sec
Throughput: 45,607,326 iterations/sec
Throughput: 45,303,331 iterations/sec
Throughput: 45,634,108 iterations/sec
```

The cons of introducing the native-async in trait would be
 - Bump the msrv to 1.75
 - The exporter interface is no longer object-safe. Which means, it can't be used in dynamic dispatch (e.g., via dyn Trait). This introduced slight complexity in otel-otlp, where dyn dispatch was used for http and tonic exporters.

Current export method could be simplified a bit for lifetime specifiers. I will look into that. But in general, open for discussion if we should bump the msrv.

cijothomas:
> Bump the msrv to 1.75
> The exporter interface is no longer object-safe. Which means, it can't be used in dynamic dispatch (e.g., via dyn Trait). This introduced slight complexity in otel-otlp, where dyn dispatch was used for http and tonic exporters.

Based on today's SIG discussion:
1.75 bump is acceptable.
the exporter interface not being object safe is an easy to fix issue as this PR already demonstrates fixing it for OTLP.

Awaiting more feedback.

lalitb:
This is ready for review now, with #2417 to bump msrv is merged

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -247,11 +247,17 @@ mod tests {
...
     impl LogExporter for ReentrantLogExporter {
-        async fn export(&self, _batch: LogBatch<'_>) -> LogResult<()> {
-            // This will cause a deadlock as the export itself creates a log
-            // while still within the lock of the SimpleLogProcessor.
-            warn!(name: "my-event-name", target: "reentrant", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
-            Ok(())
+        #[allow(clippy::manual_async_fn)]
```
lalitb, line 250: This is false positive - https://github.com/rust-lang/rust-clippy/issues/12664

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -34,8 +34,12 @@ struct NoopExporter {
 
 #[async_trait]
 impl LogExporter for NoopExporter {
-    async fn export(&self, _: LogBatch<'_>) -> LogResult<()> {
-        LogResult::Ok(())
+    #[allow(clippy::manual_async_fn)]
```
lalitb, line 37: This annotation is required for false positive - rust-lang/rust-clippy#12664

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -58,35 +58,40 @@ impl TonicLogsClient {
 
 #[async_trait]
```
cijothomas, line 59: we can remove async_trait from everywhere now.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: unresolving to remove async_trait

opentelemetry-appender-tracing/benches/logs.rs
TommyCpp: Maybe have it on the project level until #12664 is addressed?

opentelemetry-otlp/src/logs.rs
```diff
@@ -108,7 +108,16 @@ impl HasHttpConfig for LogExporterBuilder<HttpExporterBuilderSet> {
 /// OTLP exporter that sends log data
 #[derive(Debug)]
 pub struct LogExporter {
-    client: Box<dyn opentelemetry_sdk::export::logs::LogExporter>,
+    //client: Box<dyn opentelemetry_sdk::export::logs::LogExporter>,
```
TommyCpp, line 111: Can remove this now

opentelemetry-otlp/src/logs.rs
```diff
@@ -108,7 +108,16 @@ impl HasHttpConfig for LogExporterBuilder<HttpExporterBuilderSet> {
...
+    client: LogExporterInner,
+}
+
+#[derive(Debug)]
+enum LogExporterInner {
```
TommyCpp, line 116: The name can be better IMO. Maybe `SupportedTransportClient`. `Inner` just don't convey this is a enum clearly

opentelemetry-appender-tracing/benches/logs.rs
lalitb: Adding `#[allow(clippy::manual_async_fn)]` at the package level in `Cargo.toml` is not supported—it only works as an attribute in Rust source files. To suppress this lint effectively, it needs to be added in `lib.rs` or `mod.rs` for the main crate and separately for `benches` (and likely `tests` as well). It seems, adding at method level seems cleaner as of now. We can revisit this if similar implementation is done for Spans, and the issue is not fixed by then.

opentelemetry-otlp/src/exporter/tonic/logs.rs
lalitb: This is done, thanks.

opentelemetry-otlp/src/logs.rs
lalitb: Yes, removed. thanks

opentelemetry-otlp/src/logs.rs
lalitb: Good point. Have renamed it to `SupportedTransportClient`.

lalitb:
Have updated the description now to reflect the improvement seen through stress test, which is ~11%

opentelemetry-appender-tracing/src/layer.rs
cijothomas: @lalitb This needs to be cleaned up.

lalitb: oh missed it. Done now :)

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -81,7 +79,14 @@ pub trait LogExporter: Send + Sync + Debug {
     /// A `LogResult<()>`, which is a result type indicating either a successful export (with
     /// `Ok(())`) or an error (`Err(LogError)`) if the export operation failed.
     ///
-    async fn export(&self, batch: LogBatch<'_>) -> LogResult<()>;
+    /// Note:
+    /// The `Send` bound ensures the future can be safely moved across threads, which is crucial for multi-threaded async runtimes like Tokio.
+    /// Explicit lifetimes (`'a`) synchronize the lifetimes of `self`, `batch`, and the returned future.
+    fn export<'a>(
```
cijothomas, line 85: its not obvious what does this lifetimes tell.. 

I understand this much:
The batch is only valid for the duration of the call. If exporter want to store the batch after that, then it should clone it and store. Though none of the existing exporter has such need.

But unclear how is the self's lifetime and return value's lifetime same as that of the batch itself...

stress/src/logs.rs
```diff
@@ -37,14 +39,15 @@ pub struct MockLogProcessor {
 impl LogProcessor for MockLogProcessor {
     fn emit(&self, record: &mut opentelemetry_sdk::logs::LogRecord, scope: &InstrumentationScope) {
         let log_tuple = &[(record as &LogRecord, scope)];
-        let _ = futures_executor::block_on(self.exporter.export(LogBatch::new(log_tuple)));
+        let log_batch = LogBatch::new(log_tuple);
+        let _ = futures_executor::block_on(self.exporter.export(&log_batch));
```
cijothomas, line 43: The overall flow looks complex..
emit gets a mutable ref to LogRecord. It is ref by design (and mut by design.)
processor, if need to batch it and send, can clone and keep the copy. When it calls the exporter, can it pass ownership the exporter? Wondering why we only pass a ref to the exporter? The processor has no reason to retain ownership anyway, so why not give it up, to the exporter?

Thinking out loud to see if this can be simplified without sacrificing perf..

opentelemetry-sdk/src/export/logs/mod.rs
hdost: You can think about it this way:
- `self` has a life time 
- `batch` has a lifetime 
By tying `'a` to both the inputs and the outputs we are telling the caller that the future we return will only live as long as the shorter lifetime. 
Since lifetimes are like expiration dates.

lalitb: The rules which these lifetime specifier do:
- The `batch` is borrowed only for the duration of the asynchronous export operation.
- `self` is also temporarily borrowed for the same duration.
- The returned `Future` must complete before these borrows end.

This doesn't change the condition that if the exporter want to store the `batch` for later use, it would need to clone that data first. The key is that both `self` and `batch` are temporary borrows that only last for this specific export operation - neither is stored long-term by default.

lalitb: oh @hdost has more concise and better reply :)

stress/src/logs.rs
lalitb: This is good point. We are passing `LogBatch` as reference. `LogBatch` is a lightweight wrapper around the slice of tuple of `LogRecord` and `InstrumentationScope`. Passing it by value won't change the ownership of this borrowed data (record and scope). We can change batch to be passed as value (though record and scope would still remain as reference). This can simplify the design (remove 'a need). However, reference just shows the intent better, and need to test the perf (theoretically it shouldn't affect).

ps - Need to dig the reason for passing record and scope as reference, I remember we discussed and agreed for it in earlier discussion. Here, we are only taking about passing batch as reference or value.

lalitb: >  The processor has no reason to retain ownership anyway, so why not give it up, to the exporter?

Probably it was not very clear from my above message. Just to clarify - this PR is not changing the ownership model. The exporter always gets the reference to the `LogRecord`, and it clones if the `LogRecord` needs to be used beyond the export function lifetime. Are we now discussing the existing design?

cijothomas: > reference just shows the intent better,

Not so sure, since we are passing a custom structure `LogBatch`, which is defined to contains refs only.
Or we can pass slice of refs only, avoiding LogBatch structure entirely. This is better discussed separately.

> Are we now discussing the existing design?

This PR is changing existing design - previously owned LogBatch was passed, now we pass a &LogBatch only. It is not clear if that is required for using async native or it is unrelated to that.
If there is no need to change from passing LogBatch to &LogBatch to make async-native work, then lets revert that part.

lalitb: > If there is no need to change from passing LogBatch to &LogBatch to make async-native work, then lets revert that part.

This is reverted now.

stress/src/logs.rs
```diff
@@ -37,14 +39,15 @@ pub struct MockLogProcessor {
 impl LogProcessor for MockLogProcessor {
     fn emit(&self, record: &mut opentelemetry_sdk::logs::LogRecord, scope: &InstrumentationScope) {
         let log_tuple = &[(record as &LogRecord, scope)];
-        let _ = futures_executor::block_on(self.exporter.export(LogBatch::new(log_tuple)));
+        let log_batch = LogBatch::new(log_tuple);
```
cijothomas, line 42: nit: we can keep existing code to reduce the amount of diffs in PR.

opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -81,7 +79,11 @@ pub trait LogExporter: Send + Sync + Debug {
     /// A `LogResult<()>`, which is a result type indicating either a successful export (with
     /// `Ok(())`) or an error (`Err(LogError)`) if the export operation failed.
     ///
-    async fn export(&self, batch: LogBatch<'_>) -> LogResult<()>;
+    fn export(
```
cijothomas, line 82: please add a changelog entry as this affects custom exporter authors.

stress/src/logs.rs
lalitb: Sure. done.

cijothomas (approved):
LGTM.

The perf gains are only applicable for concurrently exporting log processors that exports without any batching (like etw exporting).
For components in this repo (like OTLP), export is not in hot path and won't see perf gains.

But irrespective, this still is an improvement and also eliminates one more dependency!

cijothomas:
@TommyCpp would you have time to another look?

cijothomas:
Merging to unblock other works that'd otherwise have to deal with merge conflicts. Please continue reviewing, and we can address feedbacks in followups.

Issue #1690, opened 2024-04-30
Broken trace example


cijothomas:
This issue is simple example to showcase broken traces when intermixing OTel Tracing API and Tokio-tracing API.

Cargo.toml

```toml
[package]
name = "otel-tracing-interop"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
opentelemetry = "0.22.0"
opentelemetry-stdout = { version = "0.3.0", features = ["trace"] }
opentelemetry_sdk = "0.22.1"
serde_json = "1.0.116"
tracing = "0.1.40"
tracing-opentelemetry = "0.23.0"
tracing-subscriber = "0.3.18"
```

main.rs
```rs
use opentelemetry::{
    trace::{Span, Tracer, TracerProvider as _},
    KeyValue,
};

use opentelemetry_sdk::{
    resource,
    trace::{config, Config, TracerProvider},
    Resource,
};
use opentelemetry_stdout::SpanExporterBuilder;
use tracing::span;
use tracing_subscriber::{layer::SubscriberExt, Registry};

fn main() {
    println!("Hello, world!");
    let stdout_exporter = SpanExporterBuilder::default()
        .with_encoder(|writer, data| {
            serde_json::to_writer_pretty(writer, &data).unwrap();
            Ok(())
        })
        .build();
    let provider = TracerProvider::builder()
        .with_config(Config::default().with_resource(Resource::new(vec![])))
        .with_simple_exporter(stdout_exporter)
        .build();
    let tracer = provider.tracer("tracing-opentelemetry");
    let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
    let subscriber = Registry::default().with(telemetry);

    tracing::subscriber::with_default(subscriber, || {
        let tracer = provider.tracer("otel-tracing");

        // this is the root span created using tokio tracing api
        let root = span!(tracing::Level::TRACE, "tokio-tracing-span-parent");
        let _enter = root.enter();

        // this is a child span created using tokio tracing api
        // this correctly parented to the root span
        let child = span!(tracing::Level::TRACE, "tokio-tracing-span-child");
        let _enter_child = child.enter();

        // this is another child, created using otel tracing api
        // but this is *NOT* parented to the above spans
        // demonstrating broken traces when mixing and matching tracing and opentelemetry apis
[...]

PR #2378, opened 2024-12-03
fix: Allow overlapping context scopes

### LINKED ISSUE #1887: [Bug]: Overlapping contexts
### What happened?

If `ContextGuard`s are dropped out of order, the state may be corrupted. Each `ContextGuard` restores the last context but that may cause a context different than the root one to be active after all `ContextGuards` are dropped.

The current code assumes that context guards will be dropped in reverse order of their creation. I didn't see anything in the spec saying that this should or shouldn't be assumed. I think this whole report mostly boils to what should be the behavior and whether this should be allowed or the API should be built so that these out-of-order operations were impossible to trigger.

The following is slightly changed `opentelemetry::context::tests::nested_contexts`.

```rust
    #[test]
    fn overlapping_contexts() {
        #[derive(Debug, PartialEq)]
        struct ValueA(&'static str);
        #[derive(Debug, PartialEq)]
        struct ValueB(u64);
        let outer_guard = Context::new().with_value(ValueA("a")).attach();

        // Only value `a` is set
        let current = Context::current();
        assert_eq!(current.get(), Some(&ValueA("a")));
        assert_eq!(current.get::<ValueB>(), None);

        let inner_guard = Context::current_with_value(ValueB(42)).attach();
        // Both values are set in inner context
        let current = Context::current();
        assert_eq!(current.get(), Some(&ValueA("a")));
        assert_eq!(current.get(), Some(&ValueB(42)));

        assert!(Context::map_current(|cx| {
            assert_eq!(cx.get(), Some(&ValueA("a")));
            assert_eq!(cx.get(), Some(&ValueB(42)));
            true
        }));

        drop(outer_guard);

        let current = Context::current();
        assert_eq!(current.get::<ValueA>(), None);
        assert_eq!(current.get::<ValueB>(), None);
        // `inner_guard` is still alive so both `ValueA` and `ValueB` should still be accessible? Not sure about this one.


        drop(inner_guard);

        let current = Context::current();
        assert_eq!(current.get(), Some(&ValueA("a")));
        assert_eq!(current.get::<ValueB>(), None);
        // Both guards are dropped and neither value should be accessible.
    }
```

### API Version

0.23.0, git master

### SDK Version

N/A

N/A

### Relevant log output

_No response_

bantonsson:
Fixes #1887

## Changes

This PR changes the single _current_ `Context` to a more robust stack that is resilient to out of order and overlapping `Context` scopes. This robustness is needed to be able to support better interop with [tokio-rs/tracing](https://github.com/tokio-rs/tracing).

There is also an optimization based on the fact that the `Map` in the `Context` is immutable, and these are the performance benchmarks  for all the context related benchmarks.

| name                                                  | baseDuration         | changesDuration     | difference % |
|-------------------------------------------------------|----------------------|---------------------|--------------|
| `context/has_active_span/in-cx/alt`                   | `8.6±1.15ns`         | **`8.4±0.15ns`**    | `-2.0`       |
| `context/has_active_span/in-cx/spec`                  | `16.5±0.06ns`        | `16.5±0.05ns`       | `0.0`        |
| `context/has_active_span/no-cx/alt`                   | `8.4±0.01ns`         | `8.4±0.02ns`        | `0.0`        |
| `context/has_active_span/no-cx/spec`                  | `15.6±0.05ns`        | **`15.2±0.07ns`**   | `-2.0`       |
| `context/has_active_span/no-sdk/alt`                  | `8.4±0.02ns`         | `8.4±0.04ns`        | `0.0`        |
| `context/has_active_span/no-sdk/spec`                 | `15.6±0.06ns`        | **`15.2±0.05ns`**   | `-2.0`       |
| `context/is_recording/in-cx/alt`                      | `4.7±0.05ns`         | `4.7±0.07ns`        | `0.0`        |
| `context/is_recording/in-cx/spec`                     | `18.1±0.12ns`        | **`17.8±0.13ns`**   | `-2.0`       |
| `context/is_recording/no-cx/alt`                      | `4.7±0.05ns`         | `4.7±0.06ns`        | `0.0`        |
| `context/is_recording/no-cx/spec`                     | `15.9±0.06ns`        | `15.9±0.12ns`       | `0.0`        |
| `context/is_recording/no-sdk/alt`                     | `4.7±0.05ns`         | `4.7±0.04ns`        | `0.0`        |
| `context/is_recording/no-sdk/spec`                    | `15.9±0.06ns`        | `15.9±0.08ns`       | `0.0`        |
| `context/is_sampled/in-cx/alt`                        | `8.7±0.18ns`         | `8.7±0.02ns`        | `0.0`        |
| `context/is_sampled/in-cx/spec`                       | `16.8±0.03ns`        | **`16.5±0.04ns`**   | `-2.0`       |
| `context/is_sampled/no-cx/alt`                        | `8.7±0.03ns`         | `8.7±0.06ns`        | `0.0`        |
| `context/is_sampled/no-cx/spec`                       | `15.6±0.11ns`        | **`15.3±0.14ns`**   | `-2.0`       |
| `context/is_sampled/no-sdk/alt`                       | `8.7±0.03ns`         | `8.7±0.04ns`        | `0.0`        |
| `context/is_sampled/no-sdk/spec`                      | `15.6±0.11ns`        | **`15.2±0.07ns`**   | `-2.0`       |
| `context_attach/nested_cx/empty_cx`                   | `48.6±0.31ns`        | **`29.4±0.26ns`**   | `-39`        |
| `context_attach/nested_cx/single_value_cx`            | `96.9±1.66ns`        | **`30.8±0.11ns`**   | `-68`        |
| `context_attach/nested_cx/span_cx`                    | `55.7±0.21ns`        | **`30.9±0.28ns`**   | `-44`        |
| `context_attach/out_of_order_cx_drop/empty_cx`        | `51.6±0.60ns`        | **`38.0±0.27ns`**   | `-26`        |
| `context_attach/out_of_order_cx_drop/single_value_cx` | `161.5±673.14ns`     | **`39.5±0.79ns`**   | `-76`        |
| `context_attach/out_of_order_cx_drop/span_cx`         | `61.0±0.37ns`        | **`39.6±0.29ns`**   | `-35`        |
| `context_attach/single_cx/empty_cx`                   | `29.3±0.23ns`        | **`16.1±0.09ns`**   | `-45`        |
| `context_attach/single_cx/single_value_cx`            | `43.5±1.09ns`        | **`17.6±0.09ns`**   | `-60`        |
| `context_attach/single_cx/span_cx`                    | `29.5±0.14ns`        | **`17.2±0.12ns`**   | `-42`        |

### DIFF
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -460,4 +460,46 @@ mod tests {
             true
         }));
     }
+
+    #[test]
+    #[ignore = "overlapping contexts are not supported yet"]
+    fn overlapping_contexts() {
+        #[derive(Debug, PartialEq)]
+        struct ValueA(&'static str);
+        #[derive(Debug, PartialEq)]
+        struct ValueB(u64);
+
+        let outer_guard = Context::new().with_value(ValueA("a")).attach();
+
+        // Only value `a` is set
+        let current = Context::current();
+        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get::<ValueB>(), None);
+
+        let inner_guard = Context::current_with_value(ValueB(42)).attach();
+        // Both values are set in inner context
+        let current = Context::current();
+        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueB(42)));
+
+        assert!(Context::map_current(|cx| {
+            assert_eq!(cx.get(), Some(&ValueA("a")));
+            assert_eq!(cx.get(), Some(&ValueB(42)));
+            true
+        }));
+
+        drop(outer_guard);
+
+        // `inner_guard` is still alive so both `ValueA` and `ValueB` should still be accessible
+        let current = Context::current();
+        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueB(42)));
+
+        drop(inner_guard);
+
+        // Both guards are dropped and neither value should be accessible.
+        let current = Context::current();
+        assert_eq!(current.get::<ValueA>(), None);
+        assert_eq!(current.get::<ValueB>(), None);
+    }
 }
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -79,8 +79,10 @@ thread_local! {
 pub struct Context {
     #[cfg(feature = "trace")]
     pub(super) span: Option<Arc<SynchronizedSpan>>,
-    entries: HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>,
+    entries: Option<Arc<EntryMap>>,
 }
+
+type EntryMap = HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>;
 
 impl Context {
     /// Creates an empty `Context`.
@@ -111,7 +113,7 @@ impl Context {
     /// do_work()
     /// ```
     pub fn current() -> Self {
-        Context::map_current(|cx| cx.clone())
+        Self::map_current(|cx| cx.clone())
     }
 
     /// Applies a function to the current context returning its value.
@@ -153,12 +155,7 @@ impl Context {
     /// assert_eq!(all_current_and_b.get::<ValueB>(), Some(&ValueB(42)));
     /// ```
     pub fn current_with_value<T: 'static + Send + Sync>(value: T) -> Self {
-        let mut new_context = Context::current();
-        new_context
-            .entries
-            .insert(TypeId::of::<T>(), Arc::new(value));
-
-        new_context
+        Self::map_current(|cx| cx.with_value(value))
     }
 
     /// Returns a reference to the entry for the corresponding value type.
@@ -184,8 +181,9 @@ impl Context {
     /// ```
     pub fn get<T: 'static>(&self) -> Option<&T> {
         self.entries
-            .get(&TypeId::of::<T>())
-            .and_then(|rc| rc.downcast_ref())
+            .as_ref()?
+            .get(&TypeId::of::<T>())?
+            .downcast_ref()
     }
 
     /// Returns a copy of the context with the new value included.
@@ -216,12 +214,20 @@ impl Context {
     /// assert_eq!(cx_with_a_and_b.get::<ValueB>(), Some(&ValueB(42)));
     /// ```
     pub fn with_value<T: 'static + Send + Sync>(&self, value: T) -> Self {
-        let mut new_context = self.clone();
-        new_context
-            .entries
-            .insert(TypeId::of::<T>(), Arc::new(value));
-
-        new_context
+        let entries = if let Some(current_entries) = &self.entries {
+            let mut inner_entries = (**current_entries).clone();
+            inner_entries.insert(TypeId::of::<T>(), Arc::new(value));
+            Some(Arc::new(inner_entries))
+        } else {
+            let mut entries = EntryMap::default();
+            entries.insert(TypeId::of::<T>(), Arc::new(value));
+            Some(Arc::new(entries))
+        };
+        Context {
+            entries,
+            #[cfg(feature = "trace")]
+            span: self.span.clone(),
+        }
     }
 
     /// Replaces the current context on this thread with this context.
@@ -327,7 +333,7 @@ impl Context {
 impl fmt::Debug for Context {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut dbg = f.debug_struct("Context");
-        let mut entries = self.entries.len();
+        let mut entries = self.entries.as_ref().map_or(0, |e| e.len());
         #[cfg(feature = "trace")]
         {
             if let Some(span) = &self.span {
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -1,3 +1,4 @@
+use crate::otel_warn;
 #[cfg(feature = "trace")]
 use crate::trace::context::SynchronizedSpan;
 use std::any::{Any, TypeId};
@@ -9,7 +10,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 
 thread_local! {
-    static CURRENT_CONTEXT: RefCell<Context> = RefCell::new(Context::default());
+    static CURRENT_CONTEXT: RefCell<ContextStack> = RefCell::new(ContextStack::default());
 }
 
 /// An execution-scoped collection of values.
@@ -122,7 +123,7 @@ impl Context {
     /// Note: This function will panic if you attempt to attach another context
     /// while the current one is still borrowed.
     pub fn map_current<T>(f: impl FnOnce(&Context) -> T) -> T {
-        CURRENT_CONTEXT.with(|cx| f(&cx.borrow()))
+        CURRENT_CONTEXT.with(|cx| cx.borrow().map_current_cx(f))
     }
 
     /// Returns a clone of the current thread's context with the given value.
@@ -298,12 +299,10 @@ impl Context {
     /// assert_eq!(Context::current().get::<ValueA>(), None);
     /// ```
     pub fn attach(self) -> ContextGuard {
-        let previous_cx = CURRENT_CONTEXT
-            .try_with(|current| current.replace(self))
-            .ok();
+        let cx_id = CURRENT_CONTEXT.with(|cx| cx.borrow_mut().push(self));
 
         ContextGuard {
-            previous_cx,
+            cx_pos: cx_id,
             _marker: PhantomData,
         }
     }
@@ -344,17 +343,19 @@ impl fmt::Debug for Context {
 }
 
 /// A guard that resets the current context to the prior context when dropped.
-#[allow(missing_debug_implementations)]
+#[derive(Debug)]
 pub struct ContextGuard {
-    previous_cx: Option<Context>,
-    // ensure this type is !Send as it relies on thread locals
+    // The position of the context in the stack. This is used to pop the context.
+    cx_pos: u16,
+    // Ensure this type is !Send as it relies on thread locals
     _marker: PhantomData<*const ()>,
 }
 
 impl Drop for ContextGuard {
     fn drop(&mut self) {
-        if let Some(previous_cx) = self.previous_cx.take() {
-            let _ = CURRENT_CONTEXT.try_with(|current| current.replace(previous_cx));
+        let id = self.cx_pos;
+        if id > ContextStack::BASE_POS && id < ContextStack::MAX_POS {
+            CURRENT_CONTEXT.with(|context_stack| context_stack.borrow_mut().pop_id(id));
         }
     }
 }
@@ -381,17 +382,116 @@ impl Hasher for IdHasher {
     }
 }
 
+/// A stack for keeping track of the [`Context`] instances that have been attached
+/// to a thread.
+///
+/// The stack allows for popping of contexts by position, which is used to do out
+/// of order dropping of [`ContextGuard`] instances. Only when the top of the
+/// stack is popped, the topmost [`Context`] is actually restored.
+///
+/// The stack relies on the fact that it is thread local and that the
+/// [`ContextGuard`] instances that are constructed using ids from it can't be
+/// moved to other threads. That means that the ids are always valid and that
+/// they are always within the bounds of the stack.
+struct ContextStack {
+    /// This is the current [`Context`] that is active on this thread, and the top
+    /// of the [`ContextStack`]. It is always present, and if the `stack` is empty
+    /// it's an empty [`Context`].
+    ///
+    /// Having this here allows for fast access to the current [`Context`].
+    current_cx: Context,
+    /// A `stack` of the other contexts that have been attached to the thread.
+    stack: Vec<Option<Context>>,
+    /// Ensure this type is !Send as it relies on thread locals
+    _marker: PhantomData<*const ()>,
+}
+
+impl ContextStack {
+    const BASE_POS: u16 = 0;
+    const MAX_POS: u16 = u16::MAX;
+    const INITIAL_CAPACITY: usize = 8;
+
+    #[inline(always)]
+    fn push(&mut self, cx: Context) -> u16 {
+        // The next id is the length of the `stack`, plus one since we have the
+        // top of the [`ContextStack`] as the `current_cx`.
+        let next_id = self.stack.len() + 1;
+        if next_id < ContextStack::MAX_POS.into() {
+            let current_cx = std::mem::replace(&mut self.current_cx, cx);
+            self.stack.push(Some(current_cx));
+            next_id as u16
+        } else {
+            // This is an overflow, log it and ignore it.
+            otel_warn!(
+                name: "Context.AttachFailed",
+                message = format!("Too many contexts. Max limit is {}. \
+                  Context::current() remains unchanged as this attach failed. \
+                  Dropping the returned ContextGuard will have no impact on Context::current().",
+                  ContextStack::MAX_POS)
+            );
+            ContextStack::MAX_POS
+        }
+    }
+
+    #[inline(always)]
+    fn pop_id(&mut self, pos: u16) {
+        if pos == ContextStack::BASE_POS || pos == ContextStack::MAX_POS {
+            // The empty context is always at the bottom of the [`ContextStack`]
+            // and cannot be popped, and the overflow position is invalid, so do
+            // nothing.
+            return;
+        }
+        let len: u16 = self.stack.len() as u16;
+        // Are we at the top of the [`ContextStack`]?
+        if pos == len {
+            // Shrink the stack if possible to clear out any out of order pops.
+            while let Some(None) = self.stack.last() {
+                _ = self.stack.pop();
+            }
+            // Restore the previous context. This will always happen since the
+            // empty context is always at the bottom of the stack if the
+            // [`ContextStack`] is not empty.
+            if let Some(Some(next_cx)) = self.stack.pop() {
+                self.current_cx = next_cx;
+            }
+        } else {
+            // This is an out of order pop.
+            if pos >= len {
+                // This is an invalid id, ignore it.
+                return;
+            }
+            // Clear out the entry at the given id.
+            _ = self.stack[pos as usize].take();
+        }
+    }
+
+    #[inline(always)]
+    fn map_current_cx<T>(&self, f: impl FnOnce(&Context) -> T) -> T {
+        f(&self.current_cx)
+    }
+}
+
+impl Default for ContextStack {
+    fn default() -> Self {
+        ContextStack {
+            current_cx: Context::default(),
+            stack: Vec::with_capacity(ContextStack::INITIAL_CAPACITY),
+            _marker: PhantomData,
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    #[derive(Debug, PartialEq)]
+    struct ValueA(u64);
+    #[derive(Debug, PartialEq)]
+    struct ValueB(u64);
+
     #[test]
     fn context_immutable() {
-        #[derive(Debug, PartialEq)]
-        struct ValueA(u64);
-        #[derive(Debug, PartialEq)]
-        struct ValueB(u64);
-
         // start with Current, which should be an empty context
         let cx = Context::current();
         assert_eq!(cx.get::<ValueA>(), None);
@@ -424,26 +524,22 @@ mod tests {
 
     #[test]
     fn nested_contexts() {
-        #[derive(Debug, PartialEq)]
-        struct ValueA(&'static str);
-        #[derive(Debug, PartialEq)]
-        struct ValueB(u64);
-        let _outer_guard = Context::new().with_value(ValueA("a")).attach();
+        let _outer_guard = Context::new().with_value(ValueA(1)).attach();
 
         // Only value `a` is set
         let current = Context::current();
-        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueA(1)));
         assert_eq!(current.get::<ValueB>(), None);
 
         {
             let _inner_guard = Context::current_with_value(ValueB(42)).attach();
             // Both values are set in inner context
             let current = Context::current();
-            assert_eq!(current.get(), Some(&ValueA("a")));
+            assert_eq!(current.get(), Some(&ValueA(1)));
             assert_eq!(current.get(), Some(&ValueB(42)));
 
             assert!(Context::map_current(|cx| {
-                assert_eq!(cx.get(), Some(&ValueA("a")));
+                assert_eq!(cx.get(), Some(&ValueA(1)));
                 assert_eq!(cx.get(), Some(&ValueB(42)));
                 true
             }));
@@ -451,39 +547,33 @@ mod tests {
 
         // Resets to only value `a` when inner guard is dropped
         let current = Context::current();
-        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueA(1)));
         assert_eq!(current.get::<ValueB>(), None);
 
         assert!(Context::map_current(|cx| {
-            assert_eq!(cx.get(), Some(&ValueA("a")));
+            assert_eq!(cx.get(), Some(&ValueA(1)));
             assert_eq!(cx.get::<ValueB>(), None);
             true
         }));
     }
 
     #[test]
-    #[ignore = "overlapping contexts are not supported yet"]
     fn overlapping_contexts() {
-        #[derive(Debug, PartialEq)]
-        struct ValueA(&'static str);
-        #[derive(Debug, PartialEq)]
-        struct ValueB(u64);
-
-        let outer_guard = Context::new().with_value(ValueA("a")).attach();
+        let outer_guard = Context::new().with_value(ValueA(1)).attach();
 
         // Only value `a` is set
         let current = Context::current();
-        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueA(1)));
         assert_eq!(current.get::<ValueB>(), None);
 
         let inner_guard = Context::current_with_value(ValueB(42)).attach();
         // Both values are set in inner context
         let current = Context::current();
-        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueA(1)));
         assert_eq!(current.get(), Some(&ValueB(42)));
 
         assert!(Context::map_current(|cx| {
-            assert_eq!(cx.get(), Some(&ValueA("a")));
+            assert_eq!(cx.get(), Some(&ValueA(1)));
             assert_eq!(cx.get(), Some(&ValueB(42)));
             true
         }));
@@ -492,7 +582,7 @@ mod tests {
 
         // `inner_guard` is still alive so both `ValueA` and `ValueB` should still be accessible
         let current = Context::current();
-        assert_eq!(current.get(), Some(&ValueA("a")));
+        assert_eq!(current.get(), Some(&ValueA(1)));
         assert_eq!(current.get(), Some(&ValueB(42)));
 
         drop(inner_guard);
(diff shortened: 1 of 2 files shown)


opentelemetry/benches/context_attach.rs
```diff
@@ -0,0 +1,119 @@
...
+                criterion::BatchSize::SmallInput,
+            );
+        },
+    );
+}
+
+#[inline(never)]
+fn nested_cx(cx1: Context, cx2: Context) {
+    let _ = black_box(cx1.attach());
+    let _ = black_box(cx2.attach());
```
paullegranddc, line 72: I don't this nest contexts. Binding a value to `_` immediately drops it ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=49ecb5881848e0ae967ce63a00da371c)) , to the opposite of binding to an unused variable

```suggestion
    let _cx1 = black_box(cx1.attach());
    let _cx2 = black_box(cx2.attach());
```

opentelemetry/src/context.rs
```diff
@@ -381,6 +389,75 @@ impl Hasher for IdHasher {
...
+    // TODO:ban wrap the whole id thing in its own type
+    id_count: usize,
+    // TODO:ban wrap the the tuple in its own type
+    stack: Vec<Option<(usize, Context)>>,
+}
+
+impl ContextStack {
+    #[inline(always)]
+    fn push(&mut self, cx: Context) -> usize {
+        self.id_count += 512; // TODO:ban clean up this
```
paullegranddc, line 404: Is this to make relying on the id value opaque?
It limits the context stack to 511 nested contexts, which is not that terrible since it's already a huge stack trace.
 But  I'm afraid that if we reach this number we will start looping over. So the 512th context will be mapped to id 0, so we won't do anything on popping so we won't change the current context and context stack will never be freed.
 
 Maybe an opaque wrapper is better?

opentelemetry/benches/context_attach.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/context.rs#L262-L264

cijothomas:
@bantonsson Can you add some details on how this fixes the issue of contexts being dropped out-of-order? Perhaps some code comments will help future readers.

opentelemetry/src/context.rs
```diff
@@ -381,6 +389,75 @@ impl Hasher for IdHasher {
...
+            while let Some(None) = self.stack.last() {
+                self.stack.pop();
+            }
+            // There is always the initial context at the bottom of the stack
+            if let Some(Some((next_id, next_cx))) = self.stack.pop() {
+                self.current_cx = next_cx;
+                self.current_id = next_id;
+            }
+        } else {
+            let pos = id & 511; // TODO:ban clean up this
```
lalitb, line 429: Can there be collision here - different id's mapping to same pos, and the logic mistakenly trying to remove the wrong entry?

lalitb: Seems the increment by 512 in push is trying to avoid the collision, but I am not sure if it prevents entirely.

opentelemetry/benches/context_attach.rs
bantonsson: Thanks for finding this. I'll make sure to update the benchmark.

opentelemetry/src/context.rs
bantonsson: Indeed an opaque wrapper is better, and as the description of the PR mentioned, the edge cases are ignored until ppl are ok with the general approach. I'm actually not sure if there needs to be any other `id` part to the `ContextStack` than the array position. I mean how would you mess up the `ContextGuard` since it's `!Send` and doesn't implement `Copy` or `Clone`.

opentelemetry/src/context.rs
bantonsson: Thinking some more about it, I actually think that there is no need for anything other than the stack position in the `ContextGuard`, since it's `!Send` and doesn't implement `Copy` or `Clone`. The `ContextStack` is thread local, and the `ContextGuard` is limited to the same thread. A stack position will not be reused until it is empty, so there shouldn't be any collisions.

bantonsson:
@cijothomas I'll clean this up a bit more and add comments describing the `ContextStack` and how it solves the out of order closing of contexts.

opentelemetry/src/context.rs
```diff
@@ -182,9 +179,13 @@ impl Context {
...
-        self.entries
-            .get(&TypeId::of::<T>())
-            .and_then(|rc| rc.downcast_ref())
+        if let Some(entries) = &self.entries {
+            entries
+                .get(&TypeId::of::<T>())
+                .and_then(|rc| rc.downcast_ref())
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
bantonsson: Almost works. Just needed an `as_ref()` in the mix. Thanks for the tip.

cijothomas:
> @cijothomas Do you know if the integration tests are flaky? The previous run [failed](https://github.com/open-telemetry/opentelemetry-rust/actions/runs/13286533085) but now it succeeded without any functional changes to the code.

You can ignore it for now. Integration tests are not marked *required* for merging PRs. It does has some instability due to spinning up OTel Collector in a docker and multiple tests using the same instance etc. That will be addressed separately.

shaun-cox:
Interesting. I too am curious to know a bit more about `tokio-rs/tracing` use cases that cause the context guards to be dropped out of order.

What is the expected behavior of the context stack if new contexts are pushed during out-of-order pops?

bantonsson:
@shaun-cox So this is not for a normal use case, but to ensure that the `Context` can handle abuse by broken code. The expected behavior is that the _current_ `Context` is the last entered and not exited `Context`. Any new push of a `Context` would be the new _current_ `Context`.

The same type of mechanism exists in the `Registry` in `tokio-rs/tracing-subscriber` that has a [stack](https://github.com/tokio-rs/tracing/blob/b4868674ba73f3963b125ec38b57efaadc714d90/tracing-subscriber/src/registry/stack.rs#L18) as well.

This might be obvious but just to be clear, the pushing and popping of a `Context` is only about what `Context` is active at this thread at this time, and does not enforce any relationship between this and the previous `Context`. The relationships are created by the `Span` instances and their parent information.

shaun-cox:
> @shaun-cox So this is not for a normal use case, but to ensure that the `Context` can handle abuse by broken code. The expected behavior is that the _current_ `Context` is the last entered and not exited `Context`. Any new push of a `Context` would be the new _current_ `Context`.
> 
> This might be obvious but just to be clear, the pushing and popping of a `Context` is only about what `Context` is active at this thread at this time and does not enforce any relationship between this and the previous `Context`. The relationships are created by the `Span` instances and their parent information.

@bantonsson Thank you for the clarification... it is helpful. I was originally interpretting the associated bug and this PR as an indication that ContextGuards could be dropped in the wrong order by *correctly* functioning application code. My understanding now is that this change is kind of "defense-in-depth" change to allow for buggy application code to keep executing? If that is the case, it kind of goes against the spirit of idiomatic Rust (IMO) which says that one should panic at the first sign of a bug, so that it can be fixed. Is it not sufficient to avoid this complexity and associated hiding of application bugs and just add a `# Panics` section to `ContextGuard` which explains that they will panic if dropped in the wrong order? (Assuming it will panic, or that its relatively simple to make them panic if dropped in the wrong order?)

cijothomas:
> > @shaun-cox So this is not for a normal use case, but to ensure that the `Context` can handle abuse by broken code. The expected behavior is that the _current_ `Context` is the last entered and not exited `Context`. Any new push of a `Context` would be the new _current_ `Context`.
> > This might be obvious but just to be clear, the pushing and popping of a `Context` is only about what `Context` is active at this thread at this time and does not enforce any relationship between this and the previous `Context`. The relationships are created by the `Span` instances and their parent information.
> 
> @bantonsson Thank you for the clarification... it is helpful. I was originally interpretting the associated bug and this PR as an indication that ContextGuards could be dropped in the wrong order by _correctly_ functioning application code. My understanding now is that this change is kind of "defense-in-depth" change to allow for buggy application code to keep executing? If that is the case, it kind of goes against the spirit of idiomatic Rust (IMO) which says that one should panic at the first sign of a bug, so that it can be fixed. Is it not sufficient to avoid this complexity and associated hiding of application bugs and just add a `# Panics` section to `ContextGuard` which explains that they will panic if dropped in the wrong order? (Assuming it will panic, or that its relatively simple to make them panic if dropped in the wrong order?)

[OTel Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles) explicitly prohibit crash/panic/exception except when at the initialization stage. However, this is probably a place where Rust idiomatic ways supersede specs, and panic, as there is no "correct" way for us to resolve this user issue.

bantonsson:
@shaun-cox I agree that dropping the `ContextGuard` out of order is something that should be an error/panic, but the main purpose of this PR is to allow OpenTelemetry to provide better interoperability with `tokio-rs/tracing`, and since [this code](https://github.com/tokio-rs/tracing/blob/b4868674ba73f3963b125ec38b57efaadc714d90/tracing-subscriber/src/registry/stack.rs#L33) in `tokio-rs/tracing-subscriber` explicitly allows for out of order closing of spans, we need to accomodate that on the OpenTelemetry side as well. It would be very bad if adding an OpenTelemetry compatibility layer to a _working_ application caused it to **panic**.

mladedav:
> If that is the case, it kind of goes against the spirit of idiomatic Rust (IMO) which says that one should panic at the first sign of a bug

I don't think logging/tracing code should panic except during explicit calls in the setup phase.

> and just add a `# Panics` section to `ContextGuard`

If possible, you [should not panic in `drop`](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics) which could make programs crash even when you try catch unwinding panics. That documentation will also be harder to discover than other panics since drops are implicit.

If you decide that this is unnecessary complexity added just for broken programs, I think the current behavior (i.e. just broken spans but no crashes) with additional `otel_error!` would be preferable.

bantonsson:
I think that if we want to have any chance of real interoperability with `tokio-rs/tracing` we will need to handle out of order context dropping (preferably in the same way), at least if we are to believe this old ticket https://github.com/tokio-rs/tracing/issues/495 and this old fix https://github.com/tokio-rs/tracing/commit/e2a38977f16dab847f883268110cd622c2960c77 that is introducing this robustness behavior in `tokio-rs/tracing`.

shaun-cox:
I might be missing the larger picture here too... what is meant by "real interoperability with `tokio-rs/tracing`"?

bantonsson:
> I might be missing the larger picture here too... what is meant by "real interoperability with `tokio-rs/tracing`"?

Thanks for pointing this out @shaun-cox. I see that my comment is lacking some context (pun intended).

Long story short, the existing `tokio-rs/tracing-opentelemetry` crate is not activating an OpenTelemetry `Span` to mimic the Tokio `Span` that is _current_ on the thread, but instead carries around a `SpanBuilder` and only quickly start/stop the OpenTelemetry `Span` when the Tokio `Span` is _closed_. This leads to unexpected behavior when mixing the two APIs, as described in #1690. I'm currently working on #2420 to fix this, but a _proper_ fix depends on being resilient to out of order context scope deactivation due to the behavior in `tokio-rs/tracing` introduced in https://github.com/tokio-rs/tracing/commit/e2a38977f16dab847f883268110cd622c2960c77 to fix tokio-rs/tracing#495.

opentelemetry/src/context.rs
```diff
@@ -346,15 +350,17 @@ impl fmt::Debug for Context {
 /// A guard that resets the current context to the prior context when dropped.
 #[allow(missing_debug_implementations)]
 pub struct ContextGuard {
-    previous_cx: Option<Context>,
-    // ensure this type is !Send as it relies on thread locals
+    // The position of the context in the stack. This is used to pop the context.
+    cx_pos: usize,
```
scottgerring, line 354: This'll be 8 bytes on a 64 bit system, but nonetheless is it worth asserting that it's less than max on `attach` ? It'd be self-documenting (yes, we've considered this), and also make it super clear what's happened if the user does anything particularly heinous to blow through contexts.

opentelemetry/src/context.rs
```diff
@@ -381,6 +387,87 @@ impl Hasher for IdHasher {
...
+        // The next id is the length of the `stack`, plus one since we have the
+        // top of the [`ContextStack`] as the `current_cx`.
+        let next_id = self.stack.len() + 1;
+        let current_cx = std::mem::replace(&mut self.current_cx, cx);
+        self.stack.push(Some(current_cx));
+        next_id
+    }
+
+    #[inline(always)]
+    fn pop_id(&mut self, pos: usize) {
```
scottgerring, line 425: I believe this isn't reentrant - there's some checking of IDs and then _subsequent action_ ™️  mutating the context and the current_cx pointer. 

I believe this doesn't matter, because this can never be called more than once concurrently in the same thread. Is this 100% the case? Is there some fun edge case here?

scottgerring: It'd be good to get coverage on the two bounds checks in here too

scottgerring (commented):
Some minor comments & queries inline @bantonsson !

Also to this:
> This PR is fully functional, but not complete with edge case checks, tests and comments. It's ment as a basis for discussions.

What's outstanding beyond what's open in the threads?

opentelemetry/benches/context_attach.rs
```diff
@@ -0,0 +1,122 @@
...
+};
+use opentelemetry::{
+    trace::{SpanContext, TraceContextExt},
+    Context,
+};
+
+// Run this benchmark with:
+// cargo bench --bench context_attach
+
+fn criterion_benchmark(c: &mut Criterion) {
```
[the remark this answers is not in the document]
scottgerring: Added #2705 and #2706 to do this. fairly straightforward!

opentelemetry/src/context.rs
```diff
@@ -78,9 +78,11 @@ thread_local! {
 pub struct Context {
     #[cfg(feature = "trace")]
     pub(super) span: Option<Arc<SynchronizedSpan>>,
-    entries: HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>,
+    entries: Option<Arc<EntryMap>>,
```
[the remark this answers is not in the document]
bantonsson: So the `Span` metadata is separately stored in the `pub(super) span: Option<Arc<SynchronizedSpan>>`, and this would mostly be `None`. I don't think that direct usage of the `Context` will be that common. But maybe the cost of always having this point to an empty `EntryMap` is ok.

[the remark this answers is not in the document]
cijothomas: For well known use cases (Span for now, a Suppression flag in the future, Baggage is another well known use case), it would be faster to store as top level fields instead of going through the HashMap. (Even Span used to be in the HashMap earlier, and was promoted to top-level field to gain perf).
https://github.com/open-telemetry/opentelemetry-rust/pull/1268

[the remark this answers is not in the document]
bantonsson: Just did some quick measurements and the creation of the `Arc` for the empty `Context` has a noticeable performance impact, and I think that will be a common case.

opentelemetry/src/context.rs
bantonsson: This is one of the undecided and not tested parts. We just need to decide on how large we want the context stack to be, and then we can adjust this type.

opentelemetry/src/context.rs
```diff
@@ -381,6 +387,87 @@ impl Hasher for IdHasher {
...
+    /// Ensure this type is !Send as it relies on thread locals
+    _marker: PhantomData<*const ()>,
+}
+
+impl ContextStack {
+    #[inline(always)]
+    fn push(&mut self, cx: Context) -> usize {
+        // The next id is the length of the `stack`, plus one since we have the
+        // top of the [`ContextStack`] as the `current_cx`.
+        let next_id = self.stack.len() + 1;
```
bantonsson, line 418: This is lacking any type of bounds check. The stack can grow very large, and we need to decide what the expected behavior should be if that happens.

opentelemetry/src/context.rs
bantonsson: Yes, more tests should be added. Also, it is safe. Or at least that's the promise of Rust since we can't share or create the `ContextGuard` instances in any other way.
```
/// The stack relies on the fact that it is thread local and that the
/// [`ContextGuard`] instances that are constructed using it can't be shared with
/// other threads.
```

opentelemetry/src/context.rs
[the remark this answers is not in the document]
bantonsson: Also, to avoid creating a new empty `HashMap` for every `Context::default()`, you would need to use some static lazily initialized HashMap instance.

opentelemetry/benches/context_attach.rs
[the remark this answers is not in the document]
bantonsson: Benchmarks are now here #2707

opentelemetry/src/context.rs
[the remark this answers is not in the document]
cijothomas: Does HashMap::new() allocate anything, until some entries are added to it?

opentelemetry/src/context.rs
scottgerring: @cijothomas any opinions? 

I reckon we'll need assertions / tests around the limits anyway. `usize` will be at least 16 bits. We could use `u16` so that it's constant width and 65k entries is still _a lot_.

opentelemetry/src/context.rs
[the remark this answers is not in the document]
bantonsson: As far as I can see, `HashMap::new()` allocates a small struct, but reuse a constant empty table. Maybe that part is not relevant.

bantonsson:
Since the github action can't post any benchmark results on fork PRs and there is a metric benchmark that fails I'm just going to post the results for `context_attach` for now.

| name                                                  | baseDuration       | changesDuration   | difference |
|-------------------------------------------------------|--------------------|-------------------|------------|
| 'context_attach/nested_cx/empty_cx'                   | '49.3±0.23ns'      | '**27.9±0.25ns**' | '-44'      |
| 'context_attach/nested_cx/single_value_cx'            | '97.0±2.27ns'      | '**29.6±0.35ns**' | '-69'      |
| 'context_attach/nested_cx/span_cx'                    | '55.5±0.37ns'      | '**29.3±0.29ns**' | '-47'      |
| 'context_attach/out_of_order_cx_drop/empty_cx'        | '51.1±0.39ns'      | '**37.5±0.10ns**' | '-26'      |
| 'context_attach/out_of_order_cx_drop/single_value_cx' | '164.2±699.46ns'   | '**39.1±0.19ns**' | '-76'      |
| 'context_attach/out_of_order_cx_drop/span_cx'         | '61.0±0.30ns'      | '**39.4±0.28ns**' | '-35'      |
| 'context_attach/single_cx/empty_cx'                   | '29.2±0.07ns'      | '**15.7±0.11ns**' | '-46'      |
| 'context_attach/single_cx/single_value_cx'            | '43.6±0.43ns'      | '**17.2±0.10ns**' | '-60'      |
| 'context_attach/single_cx/span_cx'                    | '29.4±0.18ns'      | '**16.9±0.11ns**' | '-43'      |

opentelemetry/src/context.rs
cijothomas: > We could use u16 so that it's constant width and 65k entries is still a lot.

This is plentiful, but wondering - In the event we hit the limit somewhere how do we communicate that to the caller of `Attach`? Is just internal log sufficient?

opentelemetry/src/context.rs
cijothomas, line 418:
```rust
let current_cx = std::mem::replace(&mut self.current_cx, cx);
self.stack.push(Some(current_cx));
self.stack.len()
```
Isn't this sufficient?

opentelemetry/src/context.rs
```diff
@@ -381,6 +387,87 @@ impl Hasher for IdHasher {
...
+    fn map_current_cx<T>(&self, f: impl FnOnce(&Context) -> T) -> T {
+        f(&self.current_cx)
+    }
+}
+
+impl Default for ContextStack {
+    fn default() -> Self {
+        ContextStack {
+            current_cx: Context::default(),
+            stack: Vec::with_capacity(64),
```
cijothomas, line 465: do we need to pre-allocate? It can just grow as needed, right? And once allocated, it won't shrink allocated capacity...

opentelemetry/src/context.rs
```diff
@@ -78,9 +78,11 @@ thread_local! {
 pub struct Context {
     #[cfg(feature = "trace")]
     pub(super) span: Option<Arc<SynchronizedSpan>>,
-    entries: HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>,
+    entries: Option<Arc<EntryMap>>,
 }
 
+type EntryMap = HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>;
```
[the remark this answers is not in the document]
bantonsson: The common case is not adding something to the `HashMap`, but entering/exiting spans, and since the `SpanRef` is on the side, you can reuse the same `HashMap` even though you change the `span`.

opentelemetry/src/context.rs
bantonsson: Yes, it is, but we need to add a bounds check here eventually, so I'll wait with changing this.

opentelemetry/src/context.rs
bantonsson: We don't need to preallocate, but at the same time, this is a thread local stack, and I think that preparing for a _reasonable_ number of spans/contexts is ok.

As for shrinking, I think that we should add some heuristics for doing that when the gap between that capacity and top of the stack becomes large. There is always the risk that the stack will inflate/deflate in a ping-pong fashion, so we need to decide what is more preferable.

opentelemetry/src/context.rs
bantonsson: I think that logging is the right option here, and the `attach` will become a *no-op*, and we return a special `ContextGuard` that is a *no-op* as well. We've effectively hit a hard limit and the application will not work as expected.

cijothomas: > I think that logging is the right option here, and the `attach` will become a _no-op_, and we return a special `ContextGuard` that is a _no-op_ as well. We've effectively hit a hard limit and the application will not work as expected.

Agree with this approach.

opentelemetry/src/context.rs
cijothomas: Got it. Context of 64 depth may also be a lot. Maybe stick with a lower number like 8 or so. (don't have any data to support any number, so just picked 8 to be conservative!)

The shrink part is not going to be common right? If user does attach/detach in the expected order, there should not be any gaps/holes

bantonsson: Yes, the shrink part should not be common, or if the depth of spans balloons up, and then down, it seems likely that it will do that again.

bantonsson:
I've implemented the discussed behavior now.

opentelemetry/src/context.rs
```diff
@@ -381,10 +388,107 @@ impl Hasher for IdHasher {
...
+            _marker: PhantomData,
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    #[derive(Debug, PartialEq)]
```
cijothomas, line 487: just a nit: `fn context_immutable()` below (added today) can also use this.
Also, I prefer both ValueA/B be just i64. Just makes it easy to focus on the Context logic! (Just a personal pref)

opentelemetry/src/context.rs
```diff
@@ -381,10 +388,107 @@ impl Hasher for IdHasher {
...
+        // The next id is the length of the `stack`, plus one since we have the
+        // top of the [`ContextStack`] as the `current_cx`.
+        let next_id = self.stack.len() + 1;
+        if next_id < ContextStack::MAX_POS.into() {
+            let current_cx = std::mem::replace(&mut self.current_cx, cx);
+            self.stack.push(Some(current_cx));
+            next_id as u16
+        } else {
+            // This is an overflow, log it and ignore it.
+            otel_warn!(name: "ContextStack.push", message = "Context stack overflow, context not pushed.");
```
cijothomas, line 430:
```suggestion
            otel_warn!(name: "Context.AttachFailed", message = "Too many contexts. Max limit is {}", ContextStack::MAX_POS);
```


^ Something like this. ContextStack is an internal implementation which can be changed anytime, so for user-facing error message, its best to avoid mentioning it.

cijothomas: Also good to mention that the Context::Current will be the context that was attempted to be pushed, so it must be dropped to return to previous Context. (Is my understanding correct?)

bantonsson: When this happens, the `Context` that we're trying to `attach` will not be pushed, so `Context::current()` will be the same `Context` as before this `attach` call. I'm not sure how to convey this in a concise manner in a log line.

cijothomas: Got it. Then we can state "Context::current() remains unchanged as this attach failed. Dropping the returned ContextGuard will also have no impact on Context::current()"

cijothomas (approved):
Looks good for me. There are generally a lack of tests for Context area, but we should address that separately from this.
Context is not announced as RC/Stable yet, needs to finish tests coverage before that.

Also - I wonder if we could do the perf improvement in its own PR, separate from supporting out-of-order drops? Or is the perf gain strictly tied to supporting out-of-order drops? (Not a blocker)

cijothomas:
@open-telemetry/rust-approvers PTAL. I could use 1-2 extra reviews before merging.

cijothomas:
@bantonsson Could you update PR desc. with the perf results from the most recent run. Also add a changelog entry for this.

cijothomas:
@mladedav Could you review this please?

opentelemetry/src/context.rs
```diff
@@ -381,17 +388,115 @@ impl Hasher for IdHasher {
...
+        // The next id is the length of the `stack`, plus one since we have the
+        // top of the [`ContextStack`] as the `current_cx`.
+        let next_id = self.stack.len() + 1;
+        if next_id < ContextStack::MAX_POS.into() {
+            let current_cx = std::mem::replace(&mut self.current_cx, cx);
+            self.stack.push(Some(current_cx));
+            next_id as u16
+        } else {
+            // This is an overflow, log it and ignore it.
+            otel_warn!(
```
lalitb, line 430: Will it make sense to also try to purge the stale entries when limit is reached before warning - not in this PR, probably future enhancement?

opentelemetry/src/context.rs
```diff
@@ -381,17 +388,115 @@ impl Hasher for IdHasher {
...
+/// A stack for keeping track of the [`Context`] instances that have been attached
+/// to a thread.
+///
+/// The stack allows for popping of contexts by position, which is used to do out
+/// of order dropping of [`ContextGuard`] instances. Only when the top of the
+/// stack is popped, the topmost [`Context`] is actually restored.
+///
+/// The stack relies on the fact that it is thread local and that the
+/// [`ContextGuard`] instances that are constructed using it can't be shared with
+/// other threads.
```
mladedav, line 400: Nit, I think "can't be shared with other threads" written like this usually means `!Sync` while here the guards are just `!Send`.

I think changing it to "can't be moved to other threads" would be clearer.

opentelemetry/src/context.rs
```diff
@@ -381,17 +388,115 @@ impl Hasher for IdHasher {
...
 
+/// A stack for keeping track of the [`Context`] instances that have been attached
+/// to a thread.
+///
+/// The stack allows for popping of contexts by position, which is used to do out
+/// of order dropping of [`ContextGuard`] instances. Only when the top of the
+/// stack is popped, the topmost [`Context`] is actually restored.
+///
+/// The stack relies on the fact that it is thread local and that the
+/// [`ContextGuard`] instances that are constructed using it can't be shared with
```
mladedav, line 399: This talks about guards constructed using the stack, but they're just stored there and not constructed using it? Or what does the "it" here refer to?

mladedav (approved):
Just small documentation nits but otherwise this looks good to me.

opentelemetry/src/context.rs
cijothomas: I think the current limit is already quite high, so the simple protection is good enough in my opinion.

lalitb: I agree that the limit is pretty high, but it would be better to still proactively clear out stale entries when it’s reached, especially for long-running applications. We could revisit this later as a future improvement after we gather more usage data or feedback. For now, I’m good with moving forward as-is. Thanks!

lalitb (approved):
LGTM. It would also be helpful to have a design doc for the current Context implementation here: https://github.com/open-telemetry/opentelemetry-rust/tree/main/docs/design. Something to consider for a separate PR.

opentelemetry/src/context.rs
bantonsson: There are no _stale_ entries in the stack. When an entry is removed out of order, it is replaced by a `None`, so there can be _holes_ in the stack. Since the `id` in the `ContextGuard` is the position in the stack, we can't resize it until the topmost element is removed.

opentelemetry/src/context.rs
bantonsson: Yes, that is not clear. This refers to the `id` in the `guards` coming from the use of the `stack`. I'll rephrase it.

PR #2381, opened 2024-12-04
chore: modify LogExporter and TraceExporter interfaces to support returning failure

### LINKED ISSUE #2262: Different Exporter::shutdown() interface across signals
Metrics:

https://github.com/open-telemetry/opentelemetry-rust/blob/a5e2061162abf6d86137f8b22b057bfb0aca9203/opentelemetry-sdk/src/metrics/exporter.rs#L30

Trace:

https://github.com/open-telemetry/opentelemetry-rust/blob/a5e2061162abf6d86137f8b22b057bfb0aca9203/opentelemetry-sdk/src/export/trace.rs#L46

Logs:

https://github.com/open-telemetry/opentelemetry-rust/blob/a5e2061162abf6d86137f8b22b057bfb0aca9203/opentelemetry-sdk/src/export/logs/mod.rs#L87

Logs and Trace are similar, Metrics return the status. Specs doesn't mandates returning status.

This should not be a big deal, just opening the issue to track.

scottgerring:
Fixes #2262 

## Changes


opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -47,8 +47,9 @@ impl LogExporter for OtlpHttpClient {
         Ok(())
     }
 
-    fn shutdown(&mut self) {
-        let _ = self.client.lock().map(|mut c| c.take());
+    fn shutdown(&mut self) -> LogResult<()> {
+        let _ = self.client.lock()?.take();
```
scottgerring, line 51: This could actually fail and return an error

scottgerring:
@lalitb 👋  

Here's a variant where i've simply propgated the errors through, like `MetricsExporter`. As I mentioned on the issue, I think triggering a lint in users projects when they upgrade ("do something with this unused return value") is likely preferable to multiplying out the interfaces to provide a `shutdown` and `shutdown_clean` variant. What do you think?

scottgerring:
@cijothomas  & @lalitb I think this should be good to go

opentelemetry-sdk/src/export/trace.rs
```diff
@@ -10,6 +10,9 @@ use std::time::SystemTime;
 /// Describes the result of an export.
 pub type ExportResult = Result<(), TraceError>;
 
+/// Describes the results of other operations on the trace API.
+pub type TraceResult<T> = Result<T, TraceError>;
```
cijothomas, line 14: Need to ensure consistency here. LogExporter uses same LogResult for export and shutdown both, but TraceExporter uses different Results.

Is this something we can fix too?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -590,7 +596,9 @@ mod tests {
             Ok(())
         }
 
-        fn shutdown(&mut self) {}
+        fn shutdown(&mut self) -> LogResult<()> {
```
cijothomas, line 599: we can remove this implementation as the trait's default body is good enough here.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
     [#2338](https://github.com/open-telemetry/opentelemetry-rust/pull/2338)
   - `ResourceDetector.detect()` no longer supports timeout option.
   - `opentelemetry::global::shutdown_tracer_provider()` Removed from the API, should now use `tracer_provider.shutdown()` see [#2369](https://github.com/open-telemetry/opentelemetry-rust/pull/2369) for a migration example. "Tracer provider" is cheaply cloneable, so users are encouraged to set a clone of it as the global (ex: `global::set_tracer_provider(provider.clone()))`, so that instrumentations and other components can obtain tracers from `global::tracer()`. The tracer_provider must be kept around to call shutdown on it at the end of application (ex: `tracer_provider.shutdown()`)
+  - The trait functions `LogExporter.shutdown` and `TraceExporter.shutdown` now explicitly return a result. The
```
cijothomas, line 13: nit: Could we add something for the custom authors as well?
If you are an exporter author, shutdown now requires returning a result. Also, implementing shutdown is optional, as trait already provides a default implementation that returns Ok().

^ something like this...

cijothomas:
> > As I mentioned on the issue, I think triggering a lint in users projects when they upgrade ("do something with this unused return value") is likely preferable to multiplying out the interfaces to provide a shutdown and shutdown_clean variant. What do you think?
> 
> I believe it's acceptable to introduce this as a BREAKING change at this stage, especially since we anticipate the SDK to undergo more breaking changes, at least until the next release. The CHANGELOG can include migration steps to guide users on handling the return status. Initially, I considered deprecating the existing `shutdown()` method and introducing a new `shutdown_new()` method with a plan to eventually rename it. However, I don't think that approach would be straightforward or practical.

Is this going to affect end users or only affecting the SDK itself (and custom exporter authors)? The exporter() shutdown() is only called from the SDK, so end user wouldn't have to deal with the return type from it, right?

opentelemetry-sdk/src/logs/log_processor.rs
scottgerring: 🤦

opentelemetry-sdk/src/export/trace.rs
scottgerring: We can use a type alias so that `export` is marked as returning `TraceResult<()>`, but the existing callers (all 38 of them in our codebase) don't have to change. 

I've pushed a change that works this way. I can also just pull of the band-aid and fix all our existing internal usages to use `TraceResult` directly. 

As we are looking at consistency here, another thing that jumps out is that we're using `TraceError` (which i've introduced `TraceResult` to match) , despite the trait itself being `SpanExporter`. Better would be `SpanError` and `SpanResult`. To give an indication of scale, we have 72 refs in this codebase to `TraceError` alone (plus external usages of course).

My suggestion is:

* Leave `TraceError` alone
* Change our internal usages of `ExportResult` to `TraceResult<()>` 
* Leave the `ExportResult` alias for any external usages but mark it as deprecated

Thoughts @cijothomas ?

opentelemetry-sdk/CHANGELOG.md
scottgerring: Based on your comment further down ("The exporter() shutdown() is only called from the SDK, so end user wouldn't have to deal with the return type from it") I think we _only_ need to address the exporter authors, right? Have changed lint accordingly.

opentelemetry-sdk/src/export/trace.rs
cijothomas: I'll need some time to think through this. I am unsure if the current state is what we finally want. I am thinking that we should have very specific Enums for each Result returned to user.
If shutdown() is called on an already shutdown provider, then the Result should an Error variant, but a specific one like Error::ShutdownAlreadyInvoked, so users know exactly what failed and why instead of relying on "Other("something went wrong")"

This will require more refactoring, but something we should be going before Stable. 
The "Other" should be truly for "others" - i.e something went wrong that is not accounted for, and should be rarely used, if not possible to avoid entirely.

cijothomas: Another quick thought - validate if it is considered breaking to expand the enum with new variants. Do we need to provide Into/From for std Error and that is sufficient?

scottgerring: Ok - so we're happy to take some breaking changes now and make things clean! 

I think what you'd typically see is errorType-per-function, with explicitly mapped errors where it makes sense (e.g. youe example with shutdown-already-invoked) and  [thiserror](https://docs.rs/thiserror/latest/thiserror/)  used to automatically map the rest (things like socket errors, etc. - where re-wrapping them makes no sense).

These days you see `anyhow` used a lot for application code (e.g., see how i've added it to the integration tests to clean up error handling) and `thiserror` used a lot in library code, or in situations where you care about being able to map on top of the error variants.

I think this would be pretty idiomatic (note - errors and logic for illustration purposes  only):

```rust

use thiserror::Error;

#[derive(Error, Debug)]
pub enum ExportError {
     #[error("Socket IO Error")]
    SocketError(#[from] io::Error),

    #[error("Exporter already shut down")]
    ExporterShutdown,
}

pub trait SpanExporter {
    async fn export(&mut self, batch: Vec<SpanData>) -> Result<(), ExportError> { 
          // explicit error
          if already_shutdown { 
              Err(ExportError::ExporterShutdown);
          }

          // by way of example, imagine this may return an IO error
          // thiserror has setup the error mapping for us, so we can just use ? 
          let ack = my_socket.send()?;
         
          Ok(())
     }
    async fn shutdown(&mut self) -> Result<(), ShutdownError>; 
}
```

Finally I don't think the extra typedefs add much value.

lalitb: Not sure if I understood this fully, but the code already uses `thiserror` nicely to get the error status from exporters, with below skeleton. Are you proposing to change this in better way?

```rust
#[derive(thiserror::Error, Debug)]
#[non_exhaustive]
pub enum TraceError {
    #[error("Exporter {0} encountered the following error(s): {name}", name = .0.exporter_name())]
    ExportFailed(Box<dyn ExportError>),
    // Rest of variants
}
/// Trait for errors returned by exporters
pub trait ExportError: std::error::Error + Send + Sync + 'static {
    /// The name of exporter that returned this error
    fn exporter_name(&self) -> &'static str;
}
// and then in OTLP  crate
/// Wrap type for errors from this crate.
#[derive(thiserror::Error, Debug)]
pub enum Error {
    // OTLP specific error variants
}
impl opentelemetry::trace::ExportError for Error {
    fn exporter_name(&self) -> &'static str {
        "otlp"
    }
}
``

Comment (scottgerring) [reply opentelemetry-sdk/src/export/trace.rs:11]: Whoops, so it does. i'll fix my suggestion so it doesn't include adding thiserror as its already there.

Comment (cijothomas) [reply opentelemetry-sdk/src/export/trace.rs:11]: // explicit error if already_shutdown { Err(ExportError::ExporterShutdown) }

^ I actually meant to suggest we use explicit ones like this, instead of "Other("Shutdown. already invoked"), consistently throughout.

File: opentelemetry-otlp/CHANGELOG.md [hunk 5]
```diff
@@ -2,6 +2,9 @@
 
 ## vNext
 
+- `OtlpHttpClient.shutdown` `TonicLogsClient.shutdown`, and `TonicTracesClient.shutdown` now explicitly return a result. The
+    semantics of the method have not changed, but you will have a new lint encouraging you to consume these results.
```
cijothomas, line 6: the 2nd sentence maybe omitted as these are anyway expected to be called from upstream sdk.

opentelemetry-sdk/src/export/trace.rs
scottgerring: @cijothomas @lalitb I think this makes sense, otherwise we're assuming a lot about what custom implementations might need in terms of thiserror-mapped-errors.

Are we happy for me to go through and apply this? Would look like:

 ```rust
shutdown(...) -> Result<(), ShutdownError>
export(...) -> Result<(), ExportError>
```

and an extension of the error types following the explicit-named-errors-style established already in `LogError`, e.g.:

```rust
pub enum LogError {
    /// Export failed with the error returned by the exporter.
    #[error("Exporter {0} encountered the following errors: {name}", name = .0.exporter_name())]
    ExportFailed(Box<dyn ExportError>),

    /// Export failed to finish after certain period and processor stopped the export.
    #[error("Exporter timed out after {} seconds", .0.as_secs())]
    ExportTimedOut(Duration),

    /// Processor is already shutdown
    #[error("{0} already shutdown")]
    AlreadyShutdown(String),

    /// Mutex lock poisoning
    #[error("mutex lock poisioning for {0}")]
    MutexPoisoned(String),
```

scottgerring:
@cijothomas i've separated out ShutdownResult for both trace and log as discussed. This was more fiddly than I expected!
imho It'd be good to get this in soon as it touches the interfaces (and I will be off for a few weeks 🎄 )

opentelemetry-sdk/src/logs/error.rs
```diff
@@ -61,3 +61,22 @@ impl<T> From<PoisonError<T>> for LogError {
...
 #[error("{0}")]
 struct Custom(String);
+
+/// Errors returned during shutdown
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum ShutdownError {
+    /// Mutex lock poisoning
+    #[error("mutex lock poisioning for {0}")]
+    MutexPoisoned(String),
```
cijothomas, line 71: unsure if we should make these public.. There are strictly internal details, and a mutex poison is not user actionable at all.

scottgerring: If we keep one uber-error and the caller wants to handle it, they are forced to handle variants that have nothing to do with the operation they are calling

opentelemetry-sdk/src/export/trace.rs
```diff
@@ -98,3 +105,32 @@ pub struct SpanData {
...
+pub enum ShutdownError {
+    /// The exporter has already been shut down.
+    #[error("Shutdown already performed")]
+    AlreadyShutdown,
+
+    /// Shutdown timed out before completing.
+    #[error("Shutdown timed out after {0:?}")]
+    Timeout(Duration),
+
+    /// The export client failed while holding the client lock. It is not
```
scottgerring, line 121: @cijothomas I reckon this is a reasonable "API user summary" of what's happened with the mutex poisoning, and advice on dealing with it

opentelemetry-sdk/src/logs/error.rs
```diff
@@ -54,10 +56,32 @@ impl From<&'static str> for LogError {
...
+    ClientFailed(String),
+
+    /// Other errors propagated from log SDK that weren't covered above.
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
+}
+
+impl<T> From<PoisonError<T>> for ShutdownError {
+    fn from(err: PoisonError<T>) -> Self {
+        ShutdownError::ClientFailed(format!("Mutex poisoned during shutdown: {}", err))
```
scottgerring, line 85: I think wrapping up the formatted inner error is plenty of context here given the narrow failure mode. I also considered wrapping up `Box<Err>` , but I think this is 1/ confusing in the interface and 2/ not at all actionable. 

There's some [good stuff on canonical's rust best practices doc](https://github.com/canonical/rust-best-practices/blob/main/best-practices.md#error-conversion) on this stuff

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -149,14 +149,10 @@ impl SpanProcessor for SimpleSpanProcessor {
     }
 
     fn shutdown(&self) -> TraceResult<()> {
```
scottgerring, line 151: I think we should (in another issue!) chase this distinction between `TraceError` and `ShutdownError` further down the rabbit hole. For the same reasons I don't think it makes sense to use an über-error in the exporter trait ...

opentelemetry-sdk/src/export/trace.rs
```diff
@@ -98,3 +105,32 @@ pub struct SpanData {
     /// Instrumentation scope that produced this span
     pub instrumentation_scope: InstrumentationScope,
 }
+
+/// Errors returned by shutdown operations in the Export API.
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum ShutdownError {
+    /// The exporter has already been shut down.
```
cijothomas, line 113: still unsure of this....I can understand that a Provider's shutdown method, if invoked more than once, can return AlreadyShutdown error. How would an Exporter's shutdown be invoked more than once, since provider protects the processor/exporter from entering that situation?

I am not even sure if processor/exporter should even have a variable like `is_shutdown` at all.. If its shutdown is invoked, it'll kill the transports etc. any further attempt to export will result in failure naturally as transports are killed.

(Sorry not really part of this PR, but this is something that I need some time to think through....)

lalitb: > I am not even sure if processor/exporter should even have a variable like is_shutdown at all.. If its shutdown is invoked, it'll kill the transports etc. any further attempt to export will result in failure naturally as transports are killed.

I feel processor should have this flag. It can't assume how the exporters are implemented. Maybe exporter's shutdown methods are just no-op, and in that case the export may continue happening if the processor doesn't have `is_shutdown` safeguard.

cijothomas: What is the need for processor to have the flag? If shutdown was already signaled, then no further shutdown signal will ever arrive at the LogProcessor, as the Provider takes care of that, and provider own Processors. 

How would log processor ever get a 2nd shutdown call? Or how does it ever get an emit() call after shutdown is done?

lalitb: > How would log processor ever get a 2nd shutdown call? Or how does it ever get an emit() call after shutdown is done?

Yes correct, is_shutdown at provider level will take care of that. I missed that. However, logger::emit() as of now doesn't check for the provider flag, which needs to be fixed.

lalitb: raised a PR to discuss this further - #2462

scottgerring: <FWIW>

I feel like it'd be cool if `shutdown` was idempotent - you call it, it shuts down, you call it again, it no-ops. 

I think the way i've got it here with a "AlreadyShutdown" error smells a bit and i'm happy to pick up whatever you folks decide in the other PR!

cijothomas: > I feel like it'd be cool if shutdown was idempotent - you call it, it shuts down, you call it again, it no-ops.

> you call it again, it no-ops.

This maybe okay. But if a user is calling shutdown more than once, it indicates there have an bug in the way they orchestrate telemetry pipelines, so isn't it best to let them know about it...?

scottgerring: Is logging it sufficient ?

scottgerring: I reckon there’s a clear line there between errors that actually reflect some failure of the system, and errors that have no impact and are just hinting that the caller has done something funny.

By way of example a flush that fails become the connections gone is very different to “probably you didn’t mean to call this twice”. 

That stdout error one is interesting. Is there a story there ?

scottgerring: Also and regardless - it feels like something we should have a consistent view on - when to error, when not to - that we can point at apply uniformly. That consistency is maybe even more important than what the policy itself is 🤷‍♂️

opentelemetry-sdk/src/export/trace.rs
```diff
@@ -110,10 +110,6 @@ pub struct SpanData {
 #[derive(Error, Debug)]
 #[non_exhaustive]
 pub enum ShutdownError {
-    /// The exporter has already been shut down.
-    #[error("Shutdown already performed")]
```
scottgerring: Removed this as discussed

opentelemetry-sdk/src/testing/logs/in_memory_exporter.rs
```diff
@@ -148,8 +148,17 @@ impl InMemoryLogExporter {
     /// ```
     ///
     pub fn get_emitted_logs(&self) -> LogResult<Vec<LogDataWithResource>> {
-        let logs_guard = self.logs.lock().map_err(LogError::from)?;
-        let resource_guard = self.resource.lock().map_err(LogError::from)?;
+        let logs_guard = self.logs.lock().map_err(|_| {
```
scottgerring, line 151: The mutex poisoned errors here are an implementation detail of `InMemoryLogExporter` - that is, they have nothing to do with the common shape of a `LogExporter`. 

So, I removed the automatic `From` for poisoned on the `LogError` type and explicitly map detailed errors through here.

lalitb:
@scottgerring Could `ShutdownError` be a common enum shared across all three signals, or perhaps a common enum wrapped within signal-specific shutdown enums? It seems like error types for shutdown are mostly common for all signals.

Maybe for now, move it to `sdk/error.rs`, so it can be shared across all signals? It could then be re-exported in signal-specific modules, allowing users to access it as `sdk::logs::ShutdownError`, `sdk::trace::ShutdownError`, etc.  Thinking out loud if it makes sense?

scottgerring:
> @scottgerring Could `ShutdownError` be a common enum shared across all three signals, or perhaps a common enum wrapped within signal-specific shutdown enums? It seems like error types for shutdown are mostly common for all signals.
> 
> Maybe for now, move it to `sdk/error.rs`, so it can be shared across all signals? It could then be re-exported in signal-specific modules, allowing users to access it as `sdk::logs::ShutdownError`, `sdk::trace::ShutdownError`, etc. Thinking out loud if it makes sense?

Good morning @lalitb 👋 !

I had a play with this just now. Good idea! I like it; it keeps it DRY, we've already got precedent with the common `Error` type in there, and with the re-export we maintain flexibility should we need to extend one signal's variant later. 

I extended the `Error` to contain a variant for this, too, but it is worth mentioning that there are no concrete usages of this type yet, and we'd really only want to use it when a more "targeted" error can't be used _and_ the impl emitting the error can't meaningfully handle the error itself. 

Relevant bits in the [last commit](https://github.com/open-telemetry/opentelemetry-rust/pull/2381/commits/26bbdc4c1c8edc2754ccacde4b7b7fac8010df8f)

lalitb:
@scottgerring - 
Apologies for the delay in reviewing this. I’m back from vacation and will be more active now. 🙂
As I understand, this PR updates the exporter's shutdown method to return a `ShutdownResult`. Specifically:
`LogExporter::shutdown() -> ShutdownResult`
`SpanExporter::shutdown() -> ShutdownResult`

And the key benefit of introducing a dedicated ShutdownResult is that it allows the SDK (e.g., `log_emitter.rs`) to avoid handling irrelevant error types. However, the end-user invoking the shutdown method will still need to handle all possible errors.
Are you planning to propose similar changes to the shutdown methods in the provider and processor?

scottgerring:
> @scottgerring - Apologies for the delay in reviewing this. I’m back from vacation and will be more active now. 🙂 As I understand, this PR updates the exporter's shutdown method to return a `ShutdownResult`. Specifically: `LogExporter::shutdown() -> ShutdownResult` `SpanExporter::shutdown() -> ShutdownResult`
> 
> And the key benefit of introducing a dedicated `ShutdownResult` is that it allows the SDK (e.g., `log_emitter.rs`) to avoid handling irrelevant error types. However, the end-user invoking the shutdown method will still need to handle all possible errors. Are you planning to propose similar changes to the shutdown methods in the provider and processor?

Ciao @lalitb welcome back! 
im still on holiday, I can look again next week. At first blush that sounds sensible but I’d suggest that we’d want to do that in another PR; I’ve spent more time chasing down merge conflicts on this than change at this point and am keen to bound the scope. 

Chat soon,

Scott

scottgerring:
### Public API Impact

> Are you planning to propose similar changes to the shutdown methods in the provider and processor?

I think the next step after this PR is to do a proper RFC on how we want to expose aggregate errors on public APIs to the caller. Let me explain!

As you suggested, I've moved the common `ShutdownError` to `sdk/src/error.rs`, and added a variant to the `Error` enum there (note - this type _is not currently used in the repo_): 

https://github.com/open-telemetry/opentelemetry-rust/blob/de49068095e2c0e7c7b209b0c6beb2642fca1dfd/opentelemetry-sdk/src/error.rs#L40-L42

... We could, if we so chose, simply use that `Error` in the APIs that can fail in this or myriad other ways. Having said that, I don't think I really like the sound of that ("handle every possible permutations of errors" vibes, again), and think this bit _in particular_ should have a separate RFC because it goes to "aggregate public facing API error handling" in general and we should have a good think! 

### This PR / Errors on 'inner' traits

I think we should treat the _big question_ as orthogonal to the PR at hand -  i've tried to make this as surgical as possible to keep this PR focused. For now, I simply log the errors accordingly in the callers, e.g.: 

https://github.com/open-telemetry/opentelemetry-rust/blob/de49068095e2c0e7c7b209b0c6beb2642fca1dfd/opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs#L191-L196

The immediate impact of this is that we can say 1/ The shape of the `shutdown` calls is stable for the next release, and 2/ we have given ourselves the ability to handle particular errors in the immediate parents.

opentelemetry-sdk/src/error.rs
```diff
@@ -44,3 +51,28 @@ impl<T> From<PoisonError<T>> for Error {
...
+pub enum ShutdownError {
+    /// Shutdown timed out before completing.
+    #[error("Shutdown timed out after {0:?}")]
+    Timeout(Duration),
+
+    /// The export client failed while holding the client lock. It is not
+    /// possible to complete the shutdown and a retry will not help.
+    /// This is something that should not happen and should likely emit some diagnostic.
+    #[error("export client failed while holding lock; cannot retry.")]
+    ClientFailed(String),
```
cijothomas, line 67: this is exposing internal details. "Client" may or may not be a thing, depending on the exporter.

Can we live with the following for Provider.ShutDown()
1. Timeout
2. ShutdownAlreadyInvoked (yes, provider need this as users can call shutdown() more than once so we need to return this if they do so).
3. Other(String explaining more details)

Provider.ShutDown() is expected to pass shutdown() calls to registered Processors/Readers, and clean up its own internal state, if any. There is no "export" notion in provider, as provider don't even have access to Exporters.

Also wondering if we should even have Result with custom Error types for processor.shutdown(), exporter.shutdown() reader.shutdown() ? Or can be return just Result with just String, and do internal logging? These are anyway not invoked by user (user only invoke provider.shutdown()), so is there value in returning Result with custom Errors?

PR #2390, opened 2024-12-06
[DRAFT] Add Custom Thread-Based Runtime for OpenTelemetry batch Processing

lalitb:
**_Disclaimer_** - Not for review or merge. If this seems useful, we can keep this PR open - for someone to take it further to production ready in future (post 1.0 era :) )

## Changes

This draft PR introduces a custom thread-based runtime for asynchronous task handling in the OpenTelemetry Rust SDK. It provides an alternative for users which don't want to take dependency on external frameworks like Tokio or Async-Std for batch processing, featuring a simple thread-based worker pool with interval and delay scheduling. Additionally, unit tests are included to demonstrate integration with common transport libraries - hyper, reqwest/reqwest-blocking. The implementation uses `futures_executor::block_on` and `async-compat` to marshal tasks into the required runtime. The latter is smaller crate, and relevant code can be incorporated inside the otel to remove the dependency. And also added an example for using this runtime impl within batch processor for logs.

The purpose of this draft is to demonstrate the concept and implementation of a thread-based runtime tailored for OpenTelemetry logging. It is not intended for immediate approval or merging but can serve as a foundation for future enhancements.


opentelemetry-sdk/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tokio = { workspace = true, features = ["rt", "time"], optional = true }
 tokio-stream = { workspace = true, optional = true }
 http = { workspace = true, optional = true }
 tracing = {workspace = true, optional = true}
+async-compat = { version = "0.2"}
```
lalitb, line 31: This package is for marshalling async events from futures runtime to tokio /async-std. Its dependency can be eliminated by integrating the necessary code from this package, as it appears to be relatively small.

cijothomas (commented):
I think this is the *right* long term direction - we expose the required methods in runtime  traits, and provide implementations using std::thread (like this), tokio etc, and default to std::thread.

Hoping to make this the case after 1.0, once we get good confidence in all the trait methods we need.

cijothomas:

Issue #2386, opened 2024-12-05
Use Dedicated Threads for BatchExportProcessor and PeriodicReader


cijothomas:
For the 1.0 release: 

Use dedicated threads for BatchExportProcessor and PeriodicReader. This is done to improve usability, telemetry reliability, simplify debugging, and eliminate issues stemming from asynchronous runtime dependencies and/or improper usage. Existing runtime integrations (eg: `rt-tokio` etc. will remain available under an **experimental** feature flag to ensure backward compatibility and allow time for further refinement.

1. **Dedicated Background Threads as Default:**
   - Both `BatchExportProcessor` and `PeriodicReader` will use dedicated background threads by default.

2. **Existing Runtime - Offer under Experimental Feature flag:**
   - For platforms with threading restrictions or users preferring OpenTelemetry to integrate with their existing async runtimes, we will continue to support current runtime-based configurations. These configurations will be moved under an _experimental_ feature flag. This approach ensures that existing users are not disrupted while providing time to polish the APIs and functionalities beyond 1.0. Post 1.0, these can be advanced to be offered under stable feature flags.

3. **Networking Library Issues:**
   - Networking libraries like `reqwest`, `hyper`, and `tonic` used by our core Exporter (OTLP) present [challenges when operating outside async runtimes](https://github.com/open-telemetry/opentelemetry-rust/pull/2142#discussion_r1792483957):
     - **`reqwest`:** Replace with `reqwest-blocking` for dedicated threads. Given that the thread being blocked is OpenTelemetry's own background thread, this should not be a concern.
     - **`hyper` and `tonic`:** Evaluate their compatibility with dedicated threads and find appropriate solutions. If certain clients cannot be supported, users requiring those clients can revert to the experimental runtime feature.

### Advantages of Dedicated Threads

1. **Improved Reliability:**
   - Telemetry processing becomes independent of shared thread pools (e.g., Tokio), ensuring it continues even when pools are saturated.

2. **Eliminate a class of bugs:**
   - Many shutdown and deadlock issues reported stem from incorrect use or combinations of async runtimes. Dedicated threads, combined with defaults that work for majority scenarios, eliminate this class of [issues/bugs](https://github.com/open-telemetry/opentelemetry-rust/labels/shutdown%26runtime).

3. **Simplified Debugging:**
   - Debugging issues like "telemetry not flowing" becomes easier when telemetry processing occurs on a known, dedicated thread.

4. **Ease of onboarding**
[...]

Issue #2188, opened 2024-10-09
SimpleExporter is not useable with OTLP/gRPC 


cijothomas:
SimpleExporter is not useable with OTLP/gRPC (Tonic client used for gRPC)
Minimal repro:

```rust
use tracing::info;
use tracing_subscriber::layer::SubscriberExt;
use tracing_subscriber::Registry;
use tracing_subscriber::util::SubscriberInitExt;
use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
use opentelemetry_otlp::WithExportConfig;

fn main() {
    let logger_provider =  opentelemetry_otlp::new_pipeline()
    .logging()
    .with_exporter(
        opentelemetry_otlp::new_exporter()
            .tonic()
            .with_endpoint("http://localhost:4317"),
    )
    .install_simple().unwrap();

    // Create a new OpenTelemetryTracingBridge using the above LoggerProvider.
    let layer = OpenTelemetryTracingBridge::new(&logger_provider);
    tracing_subscriber::registry()
        .with(layer)
        .init();

    info!(name: "my-event", target: "my-target", "hello from {}. My price is {}", "apple", 1.99);

    let _ = logger_provider.shutdown();
}
```

Panics with
```txt
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

If making the `main` as `tokio::main`, then export hangs forever.

Need to modify exporters in such a way that they don't require async runtimes, and can do normal blocking calls (assuming there are APIs to do that).

RichardChukwu:
is this open to be worked on @cijothomas ?

mzabaluev:
The problem is that libraries in the exporter network stack create tracing spans of their own, these go to the global subscriber, and with the simple exporter this leads to a worker thread calling OpenTelemetry tracer while a lock is held in the exporter by the main thread.
Try to prevent the reentrancy in tracing by filtering the OpenTelemetry layer with e.g.

```rust
use tracing_subscriber::filter;

let layer = OpenTelemetryTracingBridge::new(&logger_provider)
    .with_filter(
        filter::Targets::new().with_target("my-target")
    );
```

mzabaluev:
> @mzabaluev would you say the Tokio runtime needs to be initialized properly to handle async tasks then?

Yes, putting the initialization into the tokio runtime context will solve the panic.

lalitb:
[...]

PR #2394, opened 2024-12-07
OTLP Example fixes

cijothomas:
Fixes a few small, but important things
1. Modifies both OTLP Examples to filter out (completely) logs from OTel/known dependencies, from being routed to OTel itself. This prevents infinite logging and also deadlocks when using SimpleProcessors.
2. Both examples modified to show how to use the `tracing::fmt` layer along with above. This demonstrates how one can view OTel's own internal logs in stdout using `tracing::fmt`. We intend to eliminate the self-diagnostics example, as that is now redundant. (The actual removal is in a future PR, after making sure we have everything covered.)
3. Adds doc on how to use SimpleExportProcessor for logs, traces, closing https://github.com/open-telemetry/opentelemetry-rust/issues/2188

(Some of these are in prep to land https://github.com/open-telemetry/opentelemetry-rust/issues/2386 )

### DIFF
--- a/opentelemetry-otlp/examples/basic-otlp-http/README.md
+++ b/opentelemetry-otlp/examples/basic-otlp-http/README.md
@@ -18,6 +18,7 @@ making the `main()` a normal main and *not* `tokio::main`
 
 // TODO: Metrics does not work with non tokio main when using `reqwest-blocking-client` today, fix that when switching 
 // default to use own thread.
+// TODO: Document `hyper` feature flag when using SimpleProcessor.
 
 ## Usage
 
--- a/opentelemetry-otlp/examples/basic-otlp-http/README.md
+++ b/opentelemetry-otlp/examples/basic-otlp-http/README.md
@@ -1,10 +1,23 @@
-# Basic OTLP exporter Example
+# Basic OTLP Exporter Example
 
-This example shows how to setup OpenTelemetry OTLP exporter for logs, metrics
-and traces to export them to the [OpenTelemetry
+This example demonstrates how to set up an OpenTelemetry OTLP exporter for logs,
+metrics, and traces to send data to the [OpenTelemetry
 Collector](https://github.com/open-telemetry/opentelemetry-collector) via OTLP
-over selected protocol such as HTTP/protobuf or HTTP/json. The Collector then sends the data to the appropriate
-backend, in this case, the logging Exporter, which displays data to console.
+over HTTP (using `protobuf` encoding by default but can be changed to use
+`json`). The Collector then forwards the data to the configured backend, which
+in this case is the logging exporter, displaying data on the console.
+Additionally, the example configures a `tracing::fmt` layer to output logs
+emitted via `tracing` to `stdout`. For demonstration, this layer uses a filter
+to display `DEBUG` level logs from various OpenTelemetry components. In real
+applications, these filters should be adjusted appropriately.
+
+The example employs a `BatchExporter` for logs and traces, which is the
+recommended approach when using OTLP exporters. While it can be modified to use
+a `SimpleExporter`, this requires enabling feature flag `reqwest-blocking-client` and
+making the `main()` a normal main and *not* `tokio::main`
+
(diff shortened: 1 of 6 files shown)


opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -71,51 +71,48 @@ fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, Metric
...
-    let filter = EnvFilter::new("info")
-        .add_directive("hyper=error".parse().unwrap())
-        .add_directive("tonic=error".parse().unwrap())
-        .add_directive("reqwest=error".parse().unwrap());
-
+    // - Restrict `opentelemetry`, `hyper`, `tonic`, and `reqwest` completely.
+    // Note: This will also drop events from crates like `tonic` etc. even when
+    // they are used outside the OTLP Exporter. For more details, see:
+    // https://github.com/open-telemetry/opentelemetry-rust/issues/761
+    let filter_otel = EnvFilter::new("info")
```
cijothomas, line 95: Mostly copied from @lalitb examples earlier, with the most important difference being the filtering of these targets completely ("off" instead of "error") - which is to ensure there is no deadlock when simple exporter is used and collector is down. Of course, users can tweak filtering according to their needs.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -158,7 +155,7 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 
     tracer_provider.shutdown()?;
     logger_provider.shutdown()?;
-    meter_provider.shutdown()?;
+    // meter_provider.shutdown()?;
```
lalitb, line 158: why is this commented out ?

lalitb (approved):
thanks for the fixes. Some nit comments.

opentelemetry-otlp/examples/basic-otlp-http/README.md
```diff
@@ -1,10 +1,24 @@
...
+# Basic OTLP Exporter Example
 
-This example shows how to setup OpenTelemetry OTLP exporter for logs, metrics
-and traces to export them to the [OpenTelemetry
+This example demonstrates how to set up an OpenTelemetry OTLP exporter for logs,
+metrics, and traces to send data to the [OpenTelemetry
 Collector](https://github.com/open-telemetry/opentelemetry-collector) via OTLP
-over selected protocol such as HTTP/protobuf or HTTP/json. The Collector then sends the data to the appropriate
-backend, in this case, the logging Exporter, which displays data to console.
+over HTTP (using `protobuf` encoding by default but can be changed to use
```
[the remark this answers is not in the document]
cijothomas: omitted it intentionally, as I expect request-blocking will be the default when we make BatchProcessor with thread as default in sdk. Will make the changes as sdk changes are done.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -71,51 +71,48 @@ fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, Metric
...
-    let result = init_tracer_provider();
-    assert!(
-        result.is_ok(),
-        "Init tracer failed with error: {:?}",
-        result.err()
-    );
-
-    let tracer_provider = result.unwrap();
+    let tracer_provider = init_traces()?;
     global::set_tracer_provider(tracer_provider.clone());
```
utpilla, line 75: It might be better to set the global provider within the `init_traces` and `init_metrics` method itself. That could better convey to the users that setting global providers is required as part of the initialization setup.

cijothomas: I wanted to show explicitly that it is cheap and okay to clone the provider here. 
Maybe setup another help create_provider, which is invoked from init_traces so we can cover your recommendation, and also show the cloning aspect..?

PR #2397, opened 2024-12-08
Modify logs basic example to use tracing

cijothomas:
Need to include few wordings in the future about preventing infinite loops, how to setup OTel as a layer to user's existing tracing pipeline etc. This PR is just first step.

### DIFF
--- a/examples/logs-basic/README.md
+++ b/examples/logs-basic/README.md
@@ -1,15 +1,15 @@
 # OpenTelemetry Log Appender for log -  Example
 
-This example shows how to use the opentelemetry-appender-log crate, which is a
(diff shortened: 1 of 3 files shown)


examples/logs-basic/README.md
```diff
@@ -1,15 +1,15 @@
 # OpenTelemetry Log Appender for log -  Example
```
lalitb, line 1:
```suggestion
# OpenTelemetry Log Appender for tracing -  Example
```

lalitb (approved):
LGTM. Another option could be to have two separate examples. Existing one can be renamed to `logs-basic-log` (or better naming), and new one `logs-basic-tracing` for tracing. But this can be revisited later.

PR #2403, opened 2024-12-10
Make PeriodicReader with own Thread the default

cijothomas:
Metrics/PeriodicReader part of https://github.com/open-telemetry/opentelemetry-rust/issues/2386

1. Makes PeriodicReaderWithOwnThread as the default PeriodicReader
2. The existing PeriodicReader that required runtime is now enabled under flag "experimental_metrics_periodicreader_with_async_runtime". Existing functionality retained as-is, just under a feature flag and module name.
3. Modified examples to use the default PeriodicReader.
4. Otlp-Basic-Http does not yet work with PeriodicReader as `reqwest` client it uses needs tokio runtime. This example is retained as-is, demonstrating that existing functionality can be retained as-is by adding "experimental_metrics_periodicreader_with_async_runtime" feature.


opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -63,8 +63,18 @@ fn init_metrics() -> Result<opentelemetry_sdk::metrics::SdkMeterProvider, Metric
...
         .build()?;
 
+    #[cfg(feature = "experimental_metrics_periodicreader_with_async_runtime")]
+    let reader =
+        opentelemetry_sdk::metrics::periodic_reader_with_async_runtime::PeriodicReader::builder(
+            exporter,
+            runtime::Tokio,
+        )
+        .build();
+    #[cfg(not(feature = "experimental_metrics_periodicreader_with_async_runtime"))]
```
utpilla, line 73: Let's add a comment saying for now this would only work when called from a regular main method.

utpilla (approved):

cijothomas:
Merging to make progress, happy to address feedbacks post merger.

PR #2303, opened 2024-11-17
Refactor/deprecate trace config

pitoniak32:
resolves: #2294 

## Changes
Moving away from `trace::Config` and `.with_config(...)` to remain consistent with the MeterProvider, and `LoggerProvider`.

# Migration Guide

Moving away from `TracerProvider::builder().with_config(trace::Config::default())`.

## Simple Example
The provider interface should be similar to the old method. For example, a `TracerProvider` being configured.
```rust
// old
let tracer_provider: TracerProvider = TracerProvider::builder()
    .with_config(Config::default().with_resource(Resource::empty()))
    .build();

// new
let tracer_provider: TracerProvider = TracerProvider::builder()
    .with_resource(Resource::empty())
    .build();
```

## Advanced Example
```rust
// old
let tracer_provider = TracerProvider::builder()
    .with_config(
        Config::default()
            .with_resource(Resource::empty())
            .with_sampler(Sampler::AlwaysOn)
            .with_max_events_per_span(16),
    )
    .build();

// new
let tracer_provider = TracerProvider::builder()
    .with_resource(Resource::empty())
    .with_sampler(Sampler::AlwaysOn)
    .with_max_events_per_span(16)
    .build();
```

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #2405, opened 2024-12-10
sdk: set correct span limit in builder

djc:
#2303 refactored the `TracerProvider` builder to deprecate the `Config` API. I like the API changes this resulted in, but the `Builder::max_events_per_span()` method sets the incorrect `span_limits.max_attributes_per_span` field.


cijothomas (approved):
Thanks!

PR #2406, opened 2024-12-10
Upgrade to thiserror 2

djc:
The minimal version for thiserror (`1`) was too low, resulting in build failures with `-Z minimal-versions`. Upgrade to thiserror 2 instead of trying to pinpoint a better minimal version.


djc:
The markdown-link-check failure appears to be a fluke.

cijothomas (approved):

cijothomas:
@utpilla I think we discussed in the past about needing `thiserror` or not. Let us continue with this, and if we ever need to remove it, it can be done without breaking change.

PR #2408, opened 2024-12-10
fix: missing `thiserror` crate when building with `--no-default-features`

weichweich:
Building with `cargo check -p opentelemetry --no-default-features` fails since `thiserror` is optional.

## Changes

### DIFF
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -24,7 +24,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 futures-core = { workspace = true, optional = true }
 futures-sink = { version = "0.3", optional = true }
 pin-project-lite = { workspace = true, optional = true }
-thiserror = { workspace = true, optional = true}
+thiserror = { workspace = true }
 tracing = {workspace = true, optional = true} # optional for opentelemetry internal logging
 
 [target.'cfg(all(target_arch = "wasm32", not(target_os = "wasi")))'.dependencies]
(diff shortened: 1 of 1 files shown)


opentelemetry/Cargo.toml
```diff
@@ -24,15 +24,15 @@ rustdoc-args = ["--cfg", "docsrs"]
 futures-core = { workspace = true, optional = true }
 futures-sink = { version = "0.3", optional = true }
 pin-project-lite = { workspace = true, optional = true }
-thiserror = { workspace = true, optional = true}
+thiserror = { workspace = true }
```
lalitb, line 27: We want to keep optional dependency on `thiserror`. As of now, it is used only when `trace` feature flag is enabled.

weichweich: `thiserror` is used in [`opentelemetry::propagation`](https://github.com/open-telemetry/opentelemetry-rust/blob/257377315d3949bff73d627b5deccd7f6a85376d/opentelemetry/src/propagation/mod.rs#L23) which is not feature gated. Building with `--no-default-features` fails. Therefore `thiserror` is required even without the feature `trace`.

`cargo check -p opentelemetry --no-default-features` fails.

lalitb: propagators are only used in the context of traces. See if we can move it include if only if `trace` flag is enabled here - https://github.com/open-telemetry/opentelemetry-rust/blob/257377315d3949bff73d627b5deccd7f6a85376d/opentelemetry/src/lib.rs#L267

lalitb: and also, you need to sign the CLA in order for this to be merged :)

weichweich: If it's fine for you I would prefer to not sign the CLA and leave the fix to you? :sweat_smile: Thanks for getting back so quickly!

lalitb: Sure. Thanks for the issue :)

PR #2409, opened 2024-12-11
Minor fixes to PeriodicReader and examples

cijothomas:
Cleanup unwanted duplicate attributes from internal logs
Example - re-order initialization to ensure Logs are initialized first, so we can see internal logs from metric/trace provider initialization.
Remove unwanted feature flags.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -344,7 +344,9 @@ impl PeriodicReaderInner {
             return Ok(());
         }
 
-        let metrics_count = rm.scope_metrics.iter().fold(0, | count, scope_metrics | count + scope_metrics.metrics.len());
+        let metrics_count = rm.scope_metrics.iter().fold(0, |count, scope_metrics| {
+            count + scope_metrics.metrics.len()
+        });
         otel_debug!(name: "PeriodicReaderMetricsCollected", count = metrics_count);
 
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -175,12 +175,12 @@ impl PeriodicReader {
                 let mut remaining_interval = interval;
                 otel_info!(
                     name: "PeriodReaderThreadStarted",
-                    interval = interval.as_secs(),
-                    timeout = timeout.as_secs()
+                    interval_in_secs = interval.as_secs(),
```
utpilla, line 178: Let's use milliseconds here as well to be consistent.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -350,6 +344,8 @@ impl PeriodicReaderInner {
             return Ok(());
         }
 
+        otel_debug!(name: "PeriodicReaderMetricsCollected", count = rm.scope_metrics.len());
```
utpilla, line 347: This would return the count of meters, right? I think what we want is something like this:

```rust
// Count the total number of metrics collected.
let metrics_count = rm.scope_metrics.iter().fold(0, | count, scope_metrics | count + scope_metrics.metrics.len());
otel_debug!(name: "PeriodicReaderMetricsCollected", count = metrics_count);
```

utpilla (approved):

Issue #2373, opened 2024-12-02
tracing minimum version dependency


cijothomas:
Opening an issue to continue the discussion in https://github.com/open-telemetry/opentelemetry-rust/pull/2128#issuecomment-2478644843

lalitb:
I think relatively easy non-breaking solution would be to remove use of `name:` metadata from the otel macros, and introduce them later.
```rust
macro_rules! otel_info {
    (name: $name:expr $(,)?) => {
        #[cfg(feature = "internal-logs")]
        {
            tracing::info!( name: $name, target: env!("CARGO_PKG_NAME") ~, "");
        }
        #[cfg(not(feature = "internal-logs"))]
        {
            let _ = $name; // Compiler will optimize this out as it's unused.
        }
    };
```

cijothomas:
but is there anything preventing us from requiring the minimum version of tracing that supports name?

lalitb:
>  but is there anything preventing us from requiring the minimum version of tracing that supports name?

Yes I think this should be fine to go with.

PR #2418, opened 2024-12-12
Version pinning for tracing and tracing-core

lalitb:
Fixes: #2373 
## Changes

Pinning versions of tracing and tracing-core that are compatible with the version required to support name metadata in macros.


cijothomas (approved):
Please add changelog entry before merge
Also left a non blocking question/comment.

lalitb:
> Please add changelog entry before merge Also left a non blocking question/comment.

updated changelog. Regarding appender, I believe it can work at version 0.1. However things would be bit complex once we add otel internal logging in appender. Let me check if we can conditionally use the proper tracing version in appender as separate PR.

PR #2421, opened 2024-12-12
Small improvements to PeriodicReader

cijothomas:
Removed `Mutex` wrapping the channel, as channel.send/receive requires shared ref only.
Explicitly handle all outcome on try_recv.
Slightly improved internal logging.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -117,9 +117,6 @@ where
 /// return metric data to the user. It will not automatically send that data to
 /// the exporter outside of the predefined interval.
 ///
-/// As this spuns up own background thread, this is recommended to be used with push exporters
-/// that do not require any particular async runtime. As of now, this cannot be used with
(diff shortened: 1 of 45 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -408,20 +405,9 @@ impl PeriodicReaderInner {
...
-                    error = format!("{:?}", e)
-                );
-                return Err(MetricError::Other(e.to_string()));
-            }
-        }
+        self.message_sender
+            .send(Message::Shutdown(response_tx))
+            .map_err(|e| MetricError::Other(e.to_string()))?;
 
         if let Ok(response) = response_rx.recv() {
```
lalitb, line 412: not related - but should this be recv_timeout, to ensure that the main thread doesn't block indefinitely if the collect/export takes too much time? Similar for flush too.

lalitb (approved):
LGTM.

utpilla (approved):

PR #2422, opened 2024-12-13
Improve shutdown handling in PeriodicReader

cijothomas:
If shutdown is called from multiple threads, modify the logic so that only one enters the core logic of sending shutdown signal etc. Previously, both threads can send shutdown message to background thread, and after the 1st one, the thread exists, so the other shutdown fails. Though it is okay, the error message may not be super easy to follow.
This PR makes sure only one shutdown proceeds, the other one immediately returning a failure indicating that failure is due to shutdown already initiated. This also ensures, checking for shutdown flag is only done at user thread invoking shutdown/flush.

There is a similar race between Flush and Shutdown possible, added it as a TODO to improve log messaging.

Also added more test for drop.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -158,7 +158,7 @@ impl PeriodicReader {
         let reader = PeriodicReader {
             inner: Arc::new(PeriodicReaderInner {
                 message_sender: Arc::new(message_sender),
-                is_shutdown: AtomicBool::new(false),
+                shutdown_invoked: AtomicBool::new(false),
                 producer: Mutex::new(None),
                 exporter: Arc::new(exporter),
             }),
@@ -300,7 +300,7 @@ struct PeriodicReaderInner {
     exporter: Arc<dyn PushMetricExporter>,
     message_sender: Arc<mpsc::Sender<Message>>,
     producer: Mutex<Option<Weak<dyn SdkProducer>>>,
-    is_shutdown: AtomicBool,
+    shutdown_invoked: AtomicBool,
 }
 
 impl PeriodicReaderInner {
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -402,8 +412,13 @@ impl PeriodicReaderInner {
     }
 
     fn shutdown(&self) -> MetricResult<()> {
-        if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
-            return Err(MetricError::Other("Reader is already shut down".into()));
+        if self
```
utpilla, line 415: Could we just use a `Mutex` object to synchronize `shutdown` and `force_flush` calls? That would greatly simplify things by eliminating both the race conditions.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -402,8 +412,13 @@ impl PeriodicReaderInner {
...
+            .is_shutdown
```
utpilla, line 416: We should rename this variable from `is_shutdown` to `shutdown_invoked` to reflect its true meaning.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -402,8 +412,13 @@ impl PeriodicReaderInner {
...
+            .swap(true, std::sync::atomic::Ordering::Relaxed)
+        {
+            return Err(MetricError::Other(
+                "PeriodicReader shutdown already invoked.".into(),
```
utpilla, line 420: Update the error message in `force_flush` as well to say "PeriodicReader shutdown already invoked." instead of "reader is shut down".

https://github.com/open-telemetry/opentelemetry-rust/blob/ce550e36bb881f4195db9b397df45d3325f2b3b4/opentelemetry-sdk/src/metrics/periodic_reader.rs#L373-L375

utpilla (approved):
Left some suggestions.

lalitb (approved):

PR #2425, opened 2024-12-13
Directly implement ComputeAggregation

fraillt:
Prerequisite for #2328

Currently part of aggregate measuring and collecting logic is scattered in several places:
* in [AggregateBuilder](https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry-0.27.1/opentelemetry-sdk/src/metrics/internal/aggregate.rs#L60)
  * attribute set filtering is applied during measurement
  * collection method is selected for specified temporality
* in specific aggregate e.g. [Sum](https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry-0.27.1/opentelemetry-sdk/src/metrics/internal/sum.rs#L45)
  * actual measurement is calculations are performed
  * actual collection happens, depending on which method was called in `AggregateBuilder`.

Instead we should implement all logic in one place (aggregate implementation e.g. `Sum`), and `AggregateBuilder` would be responsible only for initializing it.
This is especially important for #2328, because we want to be able to select specific implementation (depending on temporality) at compile time. The goal is to be able to initialize/build aggregation like this:
```rust
/// Builds a sum aggregate function input and output.
pub(crate) fn sum(&self, monotonic: bool) -> Arc<dyn Aggregate> {
  match self.temporality {
    Temporality::Delta => Arc::new(Sum::new(DeltaCollector::new(), monotinic)),
    _ => Arc::new(Sum::new(CumulativeCollector::new(), monotinic)),
  }
}
```

As a side effect this approach will also has less indirection which simplifies code, and might improve performance.
  
## Changes

* `Temporality` is now passed in to aggregation during creation
* aggregations directly implements [ComputeAggregation](https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry-0.27.1/opentelemetry-sdk/src/metrics/internal/aggregate.rs#L38)


cijothomas:
Oops just merged the startTime PR! Could you resolve conflicts

opentelemetry-sdk/src/lib.rs
```diff
@@ -118,7 +118,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo.svg"
 )]
-#![cfg_attr(test, deny(warnings))]
+// #![cfg_attr(test, deny(warnings))]
```
cijothomas, line 121: nit: is this intentional?

cijothomas, line 121:
```suggestion
#![cfg_attr(test, deny(warnings))]
```

cijothomas (approved):
Good move!

PR #2431, opened 2024-12-15
otlp: spawn thread to create blocking reqwest client

### LINKED ISSUE #2400: OTLP Exporter - initialize Reqwest::Blocking client must be done in non async context
Related to https://github.com/open-telemetry/opentelemetry-rust/issues/1757
Currently, initializing OTLP Exporter causes the creation of Reqwest::blocking client creation. If the user's main thread is a tokio::main or other asyn runtime context, this panics.

Reqwest::blocking client must be created separately to avoid this panic. Options include spinning up a throwaway thread, just for the sake of initializing the blocking client, or delaying the creation of client until 1st export.

Required for https://github.com/open-telemetry/opentelemetry-rust/issues/2386

pitoniak32:
fixes #2400

## Changes
Using a new thread to create reqwest blocking client to avoid panics from async main

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -66,8 +66,11 @@ pub struct HttpConfig {
 impl Default for HttpConfig {
     fn default() -> Self {
         #[cfg(feature = "reqwest-blocking-client")]
-        let default_client =
-            Some(Arc::new(reqwest::blocking::Client::new()) as Arc<dyn HttpClient>);
+        let default_client = std::thread::spawn(|| {
+            Some(Arc::new(reqwest::blocking::Client::new()) as Arc<dyn HttpClient>)
+        })
+        .join()
+        .expect("creating reqwest::blocking::Client on a new thread not to fail");
         #[cfg(all(not(feature = "reqwest-blocking-client"), feature = "reqwest-client"))]
         let default_client = Some(Arc::new(reqwest::Client::new()) as Arc<dyn HttpClient>);
         #[cfg(all(


pitoniak32:
I still need to test this, just getting this up to make sure its not off track.

lalitb:
> 
> I still need to test this, just getting this up to make sure its not off track.

Thanks @pitoniak32. You can uncomment the integration test for [reqwest-blocking](https://github.com/open-telemetry/opentelemetry-rust/blob/main/scripts/integration_tests.sh#L13) to verify if it functions correctly with these changes? This mayn't necessarily work as there could be some issue with the integration test, I can check that separately in that case.

cijothomas:
@pitoniak32 @lalitb 
I can confirm this is working as expected. https://github.com/open-telemetry/opentelemetry-rust/compare/main...cijothomas:cijothomas/testreq-block?expand=1 is the change I tried. Obviously, had to turn off logs/traces as they are using tokio, and won't work. But once that is also addressed, this should achieve what we want - which is to use background Thread + req::Blocking by default

lalitb:
> @pitoniak32 @lalitb I can confirm this is working as expected. https://github.com/open-telemetry/opentelemetry-rust/compare/main...cijothomas:cijothomas/testreq-block?expand=1 is the change I tried. Obviously, had to turn off logs/traces as they are using tokio, and won't work. But once that is also addressed, this should achieve what we want - which is to use background Thread + req::Blocking by default

thanks @cijothomas. That makes me realize that the integration tests logs and traces for `reqwest-blocking` won't work till we run batch exporter as separate dedicated thread. @pitoniak32 - please ignore my earlier comment on enabling and validating integration test :)

lalitb:
I can confirm these changes also work for batch log exporter modeled as separate thread in #2436.

lalitb (approved):

PR #2424, opened 2024-12-13
chore: Add metric integration tests

scottgerring:
Contributes to #2401 

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #2432, opened 2024-12-15
chore: refactor integration tests and add metrics coverage

### LINKED ISSUE #2401: Use integration test to cover key OTLP scenarios
Pre-req to https://github.com/open-telemetry/opentelemetry-rust/pull/2142
We should improve the existing integration tests to add coverage for OTLP under various combinations of ExportProcessor/PeriodicReader and the networking library (reqwest/hyper/tonic).

The test infra already exists, but it is missing few things
1. Metric is untested - Done https://github.com/open-telemetry/opentelemetry-rust/pull/2432
2. Simple processor scenarios not covered.
3. tokio::main vs normal main
4. Any missing combinations.
5. reqwest-blocking support is not working, and commented out - https://github.com/open-telemetry/opentelemetry-rust/blob/f5137683052810c964a817c1a9680de21da8ef87/scripts/integration_tests.sh#L13. Need to be fixed.

scottgerring:
Fixes #2401

This is an alternate to #2424 , with a more of an opinionated take on how the integration tests could look. Essentially I am trying to make it into a more normal looking test suite, so it is easier to extend and reason about the results of. 

1. Decompose the single "mega unit test" in `integration_test.rs` into discrete unit tests
1. Pull the otel-collector container out into `test_utils.rs` and re-use it everywhere 
1. Introduce `anyhow` to make error handling cleaner and easier to follow panic output when it happens 
1. Removed the `#[ignore]` from all the integration tests - if we don't want them to run as part of `cargo test`, we can pass `--lib` in our CI scripts 
1. Modified the collector image so it outputs a separate log file for each signal
1. Upgraded `testcontainers` and added startup waits so that we don't have to sleep waiting for the container to start and potentially run over our startup time budget

This saves a fair bit of duplication too, which is nice! 

In metrics i've added test-per-meter and some supporting code to pluck the data out for each meter easily. IMHO this will make it easy to extend and easy to follow, rather than just having a enormous "total world" diff to pick through. 

I've adapted traces/logs so each has a single normal unit test in its own file, but have not decomposed them or introduced more tests there, yet. 

## Changes


scottgerring:
@cijothomas @lalitb let me know what you think about this; I sunk a bit of time into this this weekend to try organise the integration tests so they are a bit more "ordinary". 
If we're happy with this i'll fill in the gaps in the coverage off of here rather than the other branch!

opentelemetry-otlp/tests/integration_test/expected/metrics/test_up_down_meter.json
```diff
@@ -0,0 +1,46 @@
+{
```
scottgerring, line 1: Seperate test data per meter

opentelemetry-otlp/tests/integration_test/otel-collector-config-2.yaml
```diff
@@ -0,0 +1,28 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
+
+exporters:
+  file/traces:
```
scottgerring, line 10: Separate test result file per signal

opentelemetry-otlp/tests/integration_test/tests/logs.rs
```diff
@@ -34,14 +38,23 @@ fn init_logs() -> Result<sdklogs::LoggerProvider, LogError> {
         .build())
 }
 
-pub async fn logs() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+pub async fn test_logs() -> Result<()> {
```
scottgerring, line 42: These are regular tests now, they just need to make sure the container is running with the first line

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
```diff
@@ -1,23 +1,237 @@
...
+        let metrics_in_json: Value = serde_json::from_str(metrics_in)?;
+        let metrics_out_json: Value = serde_json::from_str(&metrics_out)?;
+
+        assert_eq!(metrics_in_json, metrics_out_json);
+
+        Ok(())
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    async fn test_u64_counter() -> Result<()> {
```
scottgerring, line 175: Test-per-meter.
More to do, but this shows the general idea!

opentelemetry-otlp/tests/integration_test/tests/logs.rs
cijothomas: can you add code comment or a md file on how one can run this locally? like listing the pre-requisites (only docker, I guess?)

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
```diff
@@ -1,23 +1,235 @@
...
```
cijothomas, line 173: @pitoniak32 I tried https://github.com/open-telemetry/opentelemetry-rust/pull/2431/files on top of this, and the reqwest::blocking test passed!
(As mentioned in the other PR, just have to run the metric one alone as logs,traces still use tokio::runtime, so won't work)

opentelemetry-otlp/tests/integration_test/otel-collector-config-2.yaml
cijothomas: this is nice.

cijothomas (commented):
I like this refactor! Thanks for working on this.

(I also tested locally and confirmed that https://github.com/open-telemetry/opentelemetry-rust/pull/2431/files is solving the issue it is intending to fix)

Not marking explicit approval as the PR is marked draft. Feel free to make it review ready.

lalitb:
> Removed the #[ignore] from all the integration tests - if we don't want them to run as part of cargo test, we can pass --lib in our CI scripts

The reason for ignoring integration tests by default is that they can take a long time to run, so `cargo test` should not execute them unless explicitly specified.

lalitb:
> opentelemetry-otlp/tests/integration_test/lcov.info

nit - Do you need this file, as it is bringing ~38K loc, this will increase the clone time for repo.?

opentelemetry-otlp/tests/integration_test/src/metrics_asserter.rs
```diff
@@ -1,40 +1,64 @@
...
+    // Parse the contents into a JSON Value
+    let metrics_data: Value = serde_json::from_str(&contents)?;
+    Ok(metrics_data)
+}
 
 pub struct MetricsAsserter {
-    results: Vec<ResourceMetrics>,
-    expected: Vec<ResourceMetrics>,
+    results: Value,
+    expected: Value,
```
lalitb, line 23: The earlier code was explicitly testing the conversion of JSON to proto structs, while it seems the current code no longer does this?

lalitb: The [grpc_build.rs](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-proto/tests/grpc_build.rs) adds the annotations for serialization/deserialization between json and proto structs, and this got tested in current setup.

cijothomas:
> > Removed the #[ignore] from all the integration tests - if we don't want them to run as part of cargo test, we can pass --lib in our CI scripts
> 
> The reason for ignoring integration tests by default is that they can take a long time to run, so `cargo test` should not execute them unless explicitly specified.

if it is executing in parallel to the main CI, and is taking less time than the longest CI (which I think is the windows ci), then l think we can let integration test run always.

lalitb:
> > > Removed the #[ignore] from all the integration tests - if we don't want them to run as part of cargo test, we can pass --lib in our CI scripts
> > 
> > 
> > The reason for ignoring integration tests by default is that they can take a long time to run, so `cargo test` should not execute them unless explicitly specified.
> 
> if it is executing in parallel to the main CI, and is taking less time than the longest CI (which I think is the windows ci), then l think we can let integration test run always.

Yes, should be fine to keep enabled in CI. The concern was that is this fast enough to keep it enabled for `cargo test` from command line.

opentelemetry-otlp/tests/integration_test/src/metrics_asserter.rs
scottgerring: > The earlier code was also testing the conversion of JSON to proto structs, while it seems the current code no longer does this?

I changed it because the roundtrip for metrics is broken and its a can of worms - [this test I added as `ignore`](https://github.com/open-telemetry/opentelemetry-rust/pull/2432/files#diff-60ed943e21eb7847d83c35cae40bdf2ae03d82942c3aa7f15190174145e32cc3R157-R170) demonstrates it. I will raise a separate issue later to track this. Some of the fields drop out when you roundtrip, including e.g. the value. This is happening on the deserialization side, not the serialization side, which I think is less of a worry for us!

I had a bit of a look but because it is all tied up into the protobuff and serde serialization magic I didn't want to pull that into this PR.

scottgerring:
Hey both, thanks for the quick turnaround!

## `cargo test` and integration tests
I changed this so that it's possible to skip tests in the crate; the round-trip example i've added that shows an issue with the JSON serialization caught me out, as I marked it skipped and had to work out why it was still running in the CI build. I expect this suite will grow and the need to have compiled-but-skip tests will increase.

 [It's normal behaviour](https://doc.rust-lang.org/cargo/commands/cargo-test.html) for `cargo test` to run the integration suite, and per the docs you can skip it from the CLI by adding `--lib`. I've modified the CI jobs here so that they maintain the same behaviour - e.g. the integration test suite runs the integration suite, and CI does not. I think it is nice to keep them separate as integration suites often end up a little flakier and separating them out in Github makes it easier for devs to reason about. 

For reference, the integration suite currently adds 10s on my MBP if I run `cargo test` from the root without `--lib`. 

## Serialization/Deserialization Tests
> The earlier code was also testing the conversion of JSON to proto structs, while it seems the current code no longer does this?

I changed to using Serde types because the roundtrip for metrics is broken (raised #2434) and its a can of worms - [this test I added as `ignore`](https://github.com/open-telemetry/opentelemetry-rust/pull/2432/files#diff-60ed943e21eb7847d83c35cae40bdf2ae03d82942c3aa7f15190174145e32cc3R157-R170) demonstrates it and validates the serialization/deserialization and can be unskipped when it's fixed. 

This value (at least) disappears on the deserialization side, not the serialization side, which I think is less of a worry for us! I had a bit of a look but because it is all tied up into the protobuff and serde serialization magic I didn't want to pull that into this PR. The test I've added catches this issue with the serialization 

## Mysterious build issue
I'm [getting a failure running on GitHub](https://github.com/open-telemetry/opentelemetry-rust/actions/runs/12351835991/job/34467597262?pr=2432#step:6:331) that seems to be pointing at code that is different to what is checked in: 

```
   ::: /home/runner/work/opentelemetry-rust/opentelemetry-rust/opentelemetry-sdk/src/resource/mod.rs:95:5
   |
95 |     pub(crate) fn new<T: IntoIterator<Item = KeyValue>>(kvs: T) -> Self {
   |     ------------------------------------------------------------------- private associated function defined here
```
... the actual definition:

https://github.com/open-telemetry/opentelemetry-rust/blob/5707bfe582bbb3bc8f4b4af080a492fda5b3e192/opentelemetry-sdk/src/resource/mod.rs#L80

. I assume there is something funny going on with the crate deps but it is not immediately obvious to me - if either of you have seen this before let me know what the trick is! Else I will dive into it.  I note that the same code appears in each of the tests and seems to only fail for metrics:

https://github.com/open-telemetry/opentelemetry-rust/blob/5707bfe582bbb3bc8f4b4af080a492fda5b3e192/opentelemetry-otlp/tests/integration_test/tests/metrics.rs#L52-L56

https://github.com/open-telemetry/opentelemetry-rust/blob/5707bfe582bbb3bc8f4b4af080a492fda5b3e192/opentelemetry-otlp/tests/integration_test/tests/logs.rs#L33-L35

## Remaining Work

I will add a test for `shutdown()` flush to metrics and I think that is it. I should be able to do this tomorrow!

opentelemetry-otlp/tests/integration_test/Cargo.toml
```diff
@@ -4,15 +4,17 @@ version = "0.1.0"
 edition = "2021"
 publish = false
 
-
 [dependencies]
-opentelemetry = { path = "../../../opentelemetry", features = ["metrics", "logs"] }
-opentelemetry_sdk = { path = "../../../opentelemetry-sdk", features = ["rt-tokio", "logs", "testing"] }
-opentelemetry-proto = { path = "../../../opentelemetry-proto", features = ["gen-tonic-messages", "trace", "logs", "with-serde"] }
+opentelemetry = { path = "../../../opentelemetry", features = [] }
```
cijothomas, line 8: thanks for cleanup here!

opentelemetry-otlp/tests/integration_test/README.md
```diff
@@ -0,0 +1,14 @@
...
+[testcontainers](https://testcontainers.com/) to start an instance of the OTEL collector using [otel-collector-config.yaml](otel-collector-config.yaml), which then uses a file exporter per signal to write the output it receives back to the host machine.
+
+The tests connect directly to the collector on `localhost:4317` and `localhost:4318`, push data through, and then check that what they expect 
+has popped back out into the files output by the collector.
+
+For this to work, you need a couple of things:
+
+* Docker, for the test container
+* TCP/4317 and TCP/4318 free on your local machine. If you are running another collector, you'll need to stop it for the tests to run 
+
```
cijothomas, line 12: nit: remove empty lines

opentelemetry-otlp/tests/integration_test/README.md
```diff
@@ -0,0 +1,14 @@
+# OTLP - Integration Tests
```
cijothomas, line 1: 💯 like this wording!

opentelemetry-otlp/tests/integration_test/actual/.gitignore
```diff
@@ -0,0 +1 @@
+*.json
```
cijothomas, line 1: not sure if there are best practices around maintaining multiple gitignore files? maybe move them all to the root one itself.

cijothomas:
> > > > Removed the #[ignore] from all the integration tests - if we don't want them to run as part of cargo test, we can pass --lib in our CI scripts
> > > 
> > > 
> > > The reason for ignoring integration tests by default is that they can take a long time to run, so `cargo test` should not execute them unless explicitly specified.
> > 
> > 
> > if it is executing in parallel to the main CI, and is taking less time than the longest CI (which I think is the windows ci), then l think we can let integration test run always.
> 
> Yes, should be fine to keep enabled in CI. The concern was that is this fast enough to keep it enabled for `cargo test` from command line.

Got it. Yes I think it is best to keep this ignored, and only triggered from the integration test CI.
Even if it is just few seconds, given it needs the port free I am inclined to keep existing way.

opentelemetry-otlp/tests/integration_test/src/metrics_asserter.rs
cijothomas: do we need to test that?

cijothomas: I am totally fine to just get an ACK from Collector that it accepted our metrics/logs/traces. That will give instant value (to validate lot of stuff we are manually validating now, like is the shutdown going to panic or do its job etc.!)

Validating the actual content - Its important, but I consider it as non-blocking now, and can be added later too.

scottgerring: Just to be clear: we're getting an ACK from the collector, and parsing all the results back out from the copy written by the collector to its file outputs, and comparing them to our expectation - we're validating the data that comes out of the collector is what we sent it now ✅ 

The only "thing" is, we're just using Serde models to deserialize for that validation, not our own proto-derived ones, because of the aforementioned fields-going-missing issue that would make a proto-based-deserialization  test succeed when data is actually lost.

So - i think this is in a pretty good state 🙏 !

scottgerring:
@cijothomas , I think this is in a pretty good state, except that I need to work out why the integration suite fails on CI and not locally 😱 I'll look again tomorrow.

cijothomas:
> @cijothomas , I think this is in a pretty good state, except that I need to work out why the integration suite fails on CI and not locally 😱 I'll look again tomorrow.

I am curious if we can enable tracing::fmt in the integration tests, and view internal logs? Not required in this PR, just sharing something that'd make our lives easier if things go wrong.

opentelemetry-otlp/tests/integration_test/src/metrics_asserter.rs
lalitb: > do we need to test that?

Yes it's important to test from _opentelemetry-proto_ prospective that we can successfully deserialize the metrics data written by collector. This is a standalone crate, and also being used by users for consumers for collector. And the only way to test it is with integration tests. This is not the blocker, but we should bring it back once the serde model is fixed for metrics.

scottgerring:
@cijothomas , this should be good to merge. I've rebased and squashed everything together to make a nicer merge history too. 

There are two outstanding issues: 

* #2434 
* **HTTP-client exporters have issues that cause test failures** I will raise an issue to resolve this after this PR is merged, and work on it; as it seems like it is likely happening in the exporter itself, and i've fairly significantly refactored the integration test code, i'm keen to get this PR merged and then look at this extra issue. For the moment the new test is simply skipping itself for hyper/reqwest until whatever issue lies in there is fixed.

opentelemetry-otlp/tests/integration_test/src/metrics_asserter.rs
lalitb: Please add a TODO (with reference to the issue), so that we know this would be fixed eventually.

lalitb (approved):
Thanks for the refactor. Nicely done.

scottgerring:
hey @lalitb , i've got a test over here and a link to the issue for the roundtripping the models:

https://github.com/open-telemetry/opentelemetry-rust/blob/21174e8dc04429320c51b4ddb087d6231ddc0c72/opentelemetry-otlp/tests/integration_test/tests/metrics.rs#L199-L221

I reckon, if we keep the MetricsAsserter stuff on Serde, and have separate "roundtrip the models" tests, then we won't accidentally miss cases where an integration test regresses and we don't notice it because of model mapping. What do you think?

cijothomas (approved):
LGTM. We can address some of the remaining issues in followups as needed.

opentelemetry-otlp/tests/integration_test/src/test_utils.rs
```diff
@@ -0,0 +1,145 @@
...
+pub static METRICS_FILE: &str = "./actual/metrics.json";
+pub static LOGS_FILE: &str = "./actual/logs.json";
+pub static TRACES_FILE: &str = "./actual/traces.json";
+
+static INIT_TRACING: Once = Once::new();
+
+fn init_tracing() {
+    INIT_TRACING.call_once(|| {
+        let subscriber = FmtSubscriber::builder()
+            .with_max_level(tracing::Level::DEBUG)
```
cijothomas, line 44: DEBUG is giving a ton of noise from the networking libraries in the CI logs.. probably okay for now, we can revisit if we find it too much noise.

lalitb:
> hey @lalitb , i've got a test over here and a link to the issue for the roundtripping the models:
> 
> https://github.com/open-telemetry/opentelemetry-rust/blob/21174e8dc04429320c51b4ddb087d6231ddc0c72/opentelemetry-otlp/tests/integration_test/tests/metrics.rs#L199-L221
> 
> I reckon, if we keep the MetricsAsserter stuff on Serde, and have separate "roundtrip the models" tests, then we won't accidentally miss cases where an integration test regresses and we don't notice it because of model mapping. What do you think?

Yes agree, this ensures serialization and deserialization are validated independently.

scottgerring:
Hurrah! Thanks @lalitb @cijothomas . 
Opening another issue for the rest of the HTTP clients now.

PR #2433, opened 2024-12-16
Move shutdown checks to MeterProvider

cijothomas:
Unsure why we need to duplicate this in PeriodicReader if the MeterProvider can take care of this.
PeriodicReader will still return Error when shutdown is invoked again, but won't be as descriptive. But not something to worry, as users should only be calling flush/shutdown on the Providers, not on reader/exporter directly.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -1,7 +1,6 @@
 use std::{
     env, fmt,
     sync::{
-        atomic::AtomicBool,
         mpsc::{self, Receiver, Sender},
         Arc, Mutex, Weak,
     },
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -414,15 +402,6 @@ impl PeriodicReaderInner {
     }
 
     fn shutdown(&self) -> MetricResult<()> {
-        if self
-            .shutdown_invoked
```
lalitb, line 418: Just thinking - we don't need any check in periodically executed `collect_and_export` to _not_ do export if shutdown is invoked?

lalitb (approved):

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: after shutdown is invoked, there is no way to invoke collect or export methods (they are not public).
I am still checking for ways to better architect this - reader is cloneable today, which I am not sure if we need to. but requires more refactoring/design..

Issue #2402, opened 2024-12-09
Validate tonic can be used in blocking scenarios/non-tokio context


cijothomas:
Pre-req to https://github.com/open-telemetry/opentelemetry-rust/issues/2386 which proposes to change default to use dedicated thread, and then make blocking calls to networking libraries.

However, tonic *requires* a tokio runtime to function.
But, based on our testing, it is seen that tonic client captures/stores the tokio runtime at creation time, and subsequently, gRPC methods can be called from non-tokio context as well (i.e from the normal std::thread)
See https://github.com/open-telemetry/opentelemetry-rust/pull/2142#discussion_r1803736632 

Have done some tests to validate this behavior in separate repo:
https://github.com/cijothomas/async-explore/blob/main/src/client-no-tokio.rs

This issue is opened to validate that the above is valid, and add integration tests to validate it, and also add doc comments/examples/readme as relevant.

An example content in readme/doc:
When using OTLP with gRPC/tonic, provider(tracer/meter/logger) creation must be done in a tokio runtime, which is required for tonic for function. If your app uses tokio::main, then no further action required. Otherwise, ensure providers are created inside a tokio runtime.
Example: https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-otlp/examples/basic-otlp#basic-otlp-exporter-example

PR #2436, opened 2024-12-16
Handle batch log processing in a dedicated background thread

### LINKED ISSUE #2066: [Feature]: remove async runtime dependency in batch processors

ThomsonTan:
Fixes #2066 . This will replace #2096, with the feedback taken there and ported the change from scratch to avoid merging conflicts.

The replaces `BatchLogProcessor` depending on async runtime will be extracted out to a separated type under feature flag, in case it is still needed in some cases. 

## Changes


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -39,6 +37,12 @@ const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE: &str = "OTEL_BLRP_MAX_EXPORT_BATCH_SIZE";
 /// Default maximum batch size.
 const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE_DEFAULT: usize = 512;
 
+/// Default timeout for forceflush and shutdown.
+const OTEL_LOGS_DEFAULT_FORCEFLUSH_TIMEOUT: Duration = Duration::from_secs(1);
+
+/// environment variable name for forceflush and shutdown timeout.
+const OTEL_LOGS_FORCEFLUSH_TIMEOUT_NAME: &str = "OTEL_LOG_EXPORT_INTERVAL";
```
cijothomas, line 44: no env variables defined for force_flush timeout. Lets not enforce any for now.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -211,14 +247,36 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                LogResult::Ok(())
+            })
+            .map_err(|err| {
+                match err {
+                    RecvTimeoutError::Timeout => {
+                        otel_error!(
+                            name: "BatchLogProcessor.Shutdown.Timeout",
+                            message = "BatchLogProcessor shutdown timed out."
+                        );
+                        LogError::ExportTimedOut(self.shutdown_timeout)
```
cijothomas, line 269: this is shutdown timing out.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -229,115 +287,127 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                    Ok(BatchMessage::SetResource(resource)) => {
                         exporter.set_resource(&resource);
                     }
+                    Err(RecvTimeoutError::Timeout) => {
+                        // FIXME handle result
+                        let _result = export_with_timeout_sync(remaining_time, exporter.as_mut(), logs.split_off(0), &mut last_export_time);
+                    }
+                    Err(err) => {
+                        otel_error!(
+                            name: "BatchLogProcessor.ReceiveError",
```
cijothomas, line 337: For high severity logs like this, we need to make them user friendly. Cannot use ReceiveError as it won't tell anything to end users.

cijothomas (changes requested):
Looks good overall, left few comments.
We need to make it similar to https://github.com/open-telemetry/opentelemetry-rust/pull/2403, and keep existing batch processor under separate feature flag before merge.

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan: Is there a category for internal or unknown error?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -189,17 +215,36 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
     }
 
     fn force_flush(&self) -> LogResult<()> {
-        let (res_sender, res_receiver) = oneshot::channel();
-        self.message_sender
-            .try_send(BatchMessage::Flush(Some(res_sender)))
+        if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
+            otel_warn!(
```
lalitb, line 219: nit - no need to emit warning here, as we are already returning result to the application. If required for debugging purpose, we can make it otel_debug.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -189,17 +215,36 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
...
+                name: "BatchLogProcessor.ForceFlush.ProcessorShutdown",
+                message = "BatchLogProcessor has been shutdown. No further logs will be emitted."
+            );
+            return LogResult::Err(LogError::Other("BatchLogProcessor is already shutdown".into()));
```
lalitb, line 223: nit -return [LogError ::AlreadyShutdown](https://github.com/open-telemetry/opentelemetry-rust/blob/9cf7a40b217cf8f30bf9cf867148342eebd8fe78/opentelemetry-sdk/src/logs/error.rs#L23)

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -211,14 +256,36 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
             );
         }
 
-        let (res_sender, res_receiver) = oneshot::channel();
-        self.message_sender
-            .try_send(BatchMessage::Shutdown(res_sender))
+        let (sender, receiver) = mpsc::sync_channel(1);
```
lalitb, line 259: I remember we discussed replacing oneshot with mpsc::sync_channel to remove dependency on external crate. However, I think oneshot is still lightweight option for point-to-point communication. Not something for this PR, something to evaluate again for future.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -229,115 +296,119 @@ impl<R: RuntimeChannel> LogProcessor for BatchLogProcessor<R> {
     }
 }
 
-impl<R: RuntimeChannel> BatchLogProcessor<R> {
-    pub(crate) fn new(mut exporter: Box<dyn LogExporter>, config: BatchConfig, runtime: R) -> Self {
+impl BatchLogProcessor {
+    pub(crate) fn new(mut exporter: Box<dyn LogExporter>, config: BatchConfig) -> Self {
```
lalitb, line 300: nit - this will have some conflict with #2374 depending on which is merged first as `dyn exporter` is not possible anymore with native async support. Just to mention, nothing to be done as of now in this PR.

lalitb:
This looks in good shape at first glance. Thanks for the PR. Will like to do one final round of review before approval today. Meanwhile we need to -
 - Ensure to bring back the existing runtime based batch processor as experimental feature.
 - Test this for OTLP scenarios, which is important factor. I believe @cijothomas has explored the feasibility with periodic reader, I can spend some time to test this for our batch scenarios.  
 - The internal optimizations (e.g. memory copy while pushing to queue) can be ongoing improvements beyond this PR.

opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -197,9 +197,8 @@ impl Builder {
     pub fn with_batch_exporter<T: LogExporter + 'static, R: RuntimeChannel>(
         self,
```
lalitb, line 198: This method `Builder::with_batch_exporter()` shouldn't take the runtime as generic argument anymore.

lalitb:
I tested the following OTLP export scenarios:

1. **reqwest-blocking client**: Works when invoked from both `non-tokio-main` and `tokio-main` contexts (with #2431 changes).
2. **reqwest client**: Does not work when invoked from `tokio-main` (expected behavior).
3. **hyper client**: Does not work when invoked from `tokio-main` (expected behavior).
4. **tonic client(tokio-main)**: Works when invoked from `tokio-main` (expected behavior).
5. **tonic client (non-tokio-main)**: Works when init_logs is invoked within the Tokio context, keeping the context active (consistent with #2402).

I believe this behavior is consistent with what we expect, and observed with the Periodic exporter, which is positive.

scottgerring:
We should be able to add this to the integration test suite easily enough!

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan: Removed warning.

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan: Yes, I think the issue is oneshot channel doesn't provide recv with timeout. Let's revisit it later as the assumption could change.

opentelemetry-sdk/src/logs/log_emitter.rs
ThomsonTan: Removed.

opentelemetry-sdk/src/logs/log_processor.rs
ThomsonTan: Changed it to internal error for now.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -859,11 +1193,24 @@ mod tests {
         assert_eq!(1, exporter.get_emitted_logs().unwrap().len())
     }
 
+    #[tokio::test(flavor = "current_thread")]
+    async fn test_batch_log_processor_shutdown_under_async_runtime_current_flavor_multi_thread() {
+        let exporter = InMemoryLogExporterBuilder::default().build();
+        let processor = BatchLogProcessor::new(Box::new(exporter.clone()), BatchConfig::default());
+
+        //
+        // deadloack happens in shutdown with tokio current_thread runtime
```
cijothomas, line 1202: is this true with new implementation?

ThomsonTan: Removed the comments. It doesn't apply and the test is also enabled for the new implementation to make sure it doesn't happen.

PR #2445, opened 2024-12-18
Enable metric integration test for req-blocking

cijothomas:
Reqwest-blocking-client should work for Metrics, so that is enabled.
Also changed the verbosity of internal logs to avoid overwhelming amount of debug logs. We still get debug level logs from OpenTelemetry.
Also enabled internal-logs from OTLP exporter itself.

### DIFF
--- a/opentelemetry-otlp/tests/integration_test/tests/metrics.rs
+++ b/opentelemetry-otlp/tests/integration_test/tests/metrics.rs
@@ -30,7 +30,7 @@ async fn init_metrics() -> SdkMeterProvider {
     let exporter = create_exporter();
 
     let reader = PeriodicReader::builder(exporter)
-        .with_interval(Duration::from_millis(100))
+        .with_interval(Duration::from_millis(500))
         .with_timeout(Duration::from_secs(1))
         .build();
 
@@ -146,7 +146,7 @@ async fn setup_metrics_test() -> Result<()> {
         println!("Running setup before any tests...");
         *done = true; // Mark setup as done
 
-        // Initialise the metrics subsystem
+        // Initialize the metrics subsystem
         _ = init_metrics().await;
     }
 
@@ -184,13 +184,13 @@ pub fn validate_metrics_against_results(scope_name: &str) -> Result<()> {
 }
 
 ///
-/// TODO - the HTTP metrics exporters do not seem to flush at the moment.
+/// TODO - the HTTP metrics exporters except reqwest-blocking-client do not seem
+/// to work at the moment.
 /// TODO - fix this asynchronously.
 ///
 #[cfg(test)]
 #[cfg(not(feature = "hyper-client"))]
 #[cfg(not(feature = "reqwest-client"))]
-#[cfg(not(feature = "reqwest-blocking-client"))]
 mod tests {
 
     use super::*;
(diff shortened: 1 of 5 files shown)


opentelemetry-otlp/tests/integration_test/tests/metrics.rs
```diff
@@ -293,7 +293,7 @@ mod tests {
         // Set up the exporter
         let exporter = create_exporter();
         let reader = PeriodicReader::builder(exporter)
-            .with_interval(Duration::from_millis(100))
+            .with_interval(Duration::from_secs(30))
```
cijothomas, line 296: putting a large interval here, to make sure the test is actually testing shutdown triggered exporting, not a timer triggered.

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
```diff
@@ -30,7 +30,7 @@ async fn init_metrics() -> SdkMeterProvider {
     let exporter = create_exporter();
 
     let reader = PeriodicReader::builder(exporter)
-        .with_interval(Duration::from_millis(100))
+        .with_interval(Duration::from_millis(500))
```
cijothomas, line 33: the sleep is around 5 seconds, so 500 msec interval is fine, and this reduces the amount of internal logs.
Also, I am thinking if we should rely on force_flush/shutdown for integration tests more, to avoid this sleep requirement.  We should still have one set of tests for testing the export triggered by interval.

lalitb (approved):

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
```diff
@@ -146,7 +146,7 @@ async fn setup_metrics_test() -> Result<()> {
         println!("Running setup before any tests...");
         *done = true; // Mark setup as done
 
-        // Initialise the metrics subsystem
+        // Initialize the metrics subsystem
```
scottgerring, line 149: US-English spelling strikes again 🤣

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
scottgerring: Good idea

opentelemetry-otlp/tests/integration_test/tests/metrics.rs
scottgerring: I think we'll still need a bit of a buffer because of the flushing on the otlp-collector side? What I think what would be _the nicest_ would be to setup a simple, short exponential backoff mechanism in here - the function knows what it is looking for for the particular test, so it's well positioned to decide to wait a bit and look again: 

https://github.com/open-telemetry/opentelemetry-rust/blob/9011f63ff653e1e1cde5e17cbcf3cdb71ab280ce/opentelemetry-otlp/tests/integration_test/tests/metrics.rs#L80

so that we can use a tighter timing in the best case.

I've also added this `rotation` thing for logs, [which I believe](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18251#issuecomment-1414497442) will decrease buffering on the collector side - doesn't help us with the other signals, though:

https://github.com/open-telemetry/opentelemetry-rust/blob/9011f63ff653e1e1cde5e17cbcf3cdb71ab280ce/opentelemetry-otlp/tests/integration_test/otel-collector-config.yaml#L12-L14

scottgerring (approved):
Nice - lgtm!

PR #2461, opened 2024-12-20
MeterProvider to return specific MetricError

cijothomas:
Related to #2381
Trying to play with better Error handling, happy to get feedback on the approach..

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -123,9 +123,7 @@ impl SdkMeterProviderInner {
             .shutdown_invoked
             .load(std::sync::atomic::Ordering::Relaxed)
         {
-            Err(MetricError::Other(
-                "Cannot perform flush as MeterProvider shutdown already invoked.".into(),
-            ))
+            Err(MetricError::AlreadyShutdown)
         } else {
             self.pipes.force_flush()
         }
@@ -137,9 +135,7 @@ impl SdkMeterProviderInner {
             .swap(true, std::sync::atomic::Ordering::SeqCst)
         {
             // If the previous value was true, shutdown was already invoked.
-            Err(MetricError::Other(
-                "MeterProvider shutdown already invoked.".into(),
-            ))
+            Err(MetricError::AlreadyShutdown)
         } else {
             self.pipes.shutdown()
         }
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -123,9 +123,7 @@ impl SdkMeterProviderInner {
             .shutdown_invoked
             .load(std::sync::atomic::Ordering::Relaxed)
         {
-            Err(MetricError::Other(
-                "Cannot perform flush as MeterProvider shutdown already invoked.".into(),
-            ))
+            Err(MetricError::AlreadyShutdown)
```
lalitb, line 126: Just to be clear, is this approach different from the one proposed in #2381, where we were using separate enum types for Shutdown, and possibly ForceFlush as well? In this case, the proposal is to use the same MetricError enum for the entire user-facing interface, correct?

scottgerring: I believe it is a best practice to have error-per-call when there is a clear disunity of potential errors, like there is between `shutdown` and `export` (but probably not between `export` and `forceFlush`). 

As it stands, `shutdown()` can return two errors - `AlreadyShutdown` and `ShutdownTimeout` (plus the common boxed one for `PoisonError`.

By adding these types to `MetricError` we're promising the caller that `shutdown()` may also return  any of `Config`, `ExportErr`, `InvalidInstrumentConfiguration` errors. This is simply not true of the interface, and places extra burden on the caller to handle, leaving the caller with these options:

* 1/ Treat all errors as unrecoverable - simply match error
* 2/ Extensively dig through our code, work out that `AlreadyShutdown` and `ShutdownTimeout` are two that they care about handling, and `_ => ` to ignore the rest
* 3/ Exhaustively match all the arms without realising that a bunch of them will never be invoked

cijothomas: > By adding these types to MetricError we're promising the caller that shutdown() may also return any of Config, ExportErr, InvalidInstrumentConfiguration errors.

Agree this PR is the wrong approach. Thanks for the feedback. Need to rework.

scottgerring: Hey @cijothomas I think I overindexed on the single error type - sorry about that. 

What you have done with the named-logical-errors makes sense to me! I stole it and lifted it into the other PR for metrics/traces :D

Issue #1660, opened 2024-04-11
[Bug]: OpenTelemetry log error occurred. cannot send message to batch processor as the channel is closed


NickLarsenNZ:
### What happened?

When Tokio shuts down, I receive a lot of errors about a channel being closed:
This happens whether or not I attempt to shutdown (unset) the trace/log providers.

Here is a brief explanation as code:

```rs
#[tokio::main]
async fn main() {
    // initialise tracing/logging here

    opentelemetry::global::shutdown_tracer_provider();
    opentelemetry::global::shutdown_logger_provider();

    // Without this sleep, I get a different error:
    // OpenTelemetry log error occurred. Exporter otlp encountered the following errors: the grpc server returns error (Unknown error): , detailed error message: transport error
    tokio::time::sleep(Duration::from_secs(2)).await;

    // errors are printed at this point.
}
```

For a full (hopefully) reproducable version of the code and dependencies, see: https://gist.github.com/NickLarsenNZ/fbbb477230e9992e9a4aa2603edcf54f

I initially raised it in the #otel-rust Slack channel: https://cloud-native.slack.com/archives/C03GDP0H023/p1712590806651679

### API Version

0.22.0

### SDK Version

0.22.1

### What Exporters are you seeing the problem on?

OTLP

### Relevant log output

```shell
OpenTelemetry log error occurred. cannot send message to batch processor as the channel is closed.
```

TommyCpp:
The batch log processor has a background task that exporting the logs. When it's being `drop`ed the receiver in the background task will gets dropped too. 

After this operation, every attempt to emit logs will result in such error.

ramgdev:
I am seeing this too. Is it safe not to call `opentelemetry::global::shutdown_tracer_provider();` when the application exits to avoid this error? What are the other implications of doing so?

Oliboy50:
so if we have this error, it means that the tracer provider has not been shutdown correctly before the application stopped?

I have this error even when I call `opentelemetry::global::shutdown_tracer_provider()` before ending my application, is it normal?

Oliboy50:
on my side, I realized that this error occurred after cloning the pipeline provider with something like:
```rust
let provider = otlp_pipeline
    .install_batch(opentelemetry_sdk::runtime::Tokio)
    .expect("Failed to build OTLP pipeline")
    .provider()
    .clone();
```

[...]

PR #2462, opened 2024-12-20
Remove is_shutdown flag from processors. And fix logger::emit() to check for the flag before emit.

lalitb:
## Changes

To discuss the change suggested be @cijothomas  here - https://github.com/open-telemetry/opentelemetry-rust/pull/2381#discussion_r1894079938

### DIFF
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -271,7 +271,7 @@ impl opentelemetry::logs::Logger for Logger {
         if self.provider.inner.is_shutdown.load(Ordering::Relaxed) {
             // Optionally, log a debug message indicating logs are being discarded due to shutdown.
             otel_debug!(
-                name: "Logger.Emit.Discarded",
+                name: "Logger.Emit.ProviderShutdown",
                 message = "Log discarded because the LoggerProvider is shut down."
             );
             return;
--- a/opentelemetry-sdk/src/logs/log_emitter.rs
+++ b/opentelemetry-sdk/src/logs/log_emitter.rs
@@ -268,6 +268,14 @@ impl opentelemetry::logs::Logger for Logger {
 
     /// Emit a `LogRecord`.
     fn emit(&self, mut record: Self::LogRecord) {
+        if self.provider.inner.is_shutdown.load(Ordering::Relaxed) {
+            // Optionally, log a debug message indicating logs are being discarded due to shutdown.
+            otel_debug!(
+                name: "Logger.Emit.Discarded",
+                message = "Log discarded because the LoggerProvider is shut down."
+            );
+            return;
+        }
         let provider = &self.provider;
         let processors = provider.log_processors();
 
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -9,7 +9,7 @@ use std::sync::mpsc::{self, RecvTimeoutError, SyncSender};
 use opentelemetry::logs::Severity;
 use opentelemetry::{otel_debug, otel_error, otel_warn, InstrumentationScope};
 
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::{cmp::min, env, sync::Mutex};
 use std::{
     fmt::{self, Debug, Formatter},
@@ -87,29 +87,18 @@ pub trait LogProcessor: Send + Sync + Debug {
 #[derive(Debug)]
 pub struct SimpleLogProcessor<T: LogExporter> {
     exporter: Mutex<T>,
-    is_shutdown: AtomicBool,
 }
 
 impl<T: LogExporter> SimpleLogProcessor<T> {
     pub(crate) fn new(exporter: T) -> Self {
         SimpleLogProcessor {
             exporter: Mutex::new(exporter),
-            is_shutdown: AtomicBool::new(false),
         }
     }
 }
 
 impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
     fn emit(&self, record: &mut LogRecord, instrumentation: &InstrumentationScope) {
-        // noop after shutdown
-        if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
-            // this is a warning, as the user is trying to log after the processor has been shutdown
-            otel_warn!(
-                name: "SimpleLogProcessor.Emit.ProcessorShutdown",
-            );
-            return;
-        }
-
         let result = self
             .exporter
             .lock()
@@ -141,8 +130,6 @@ impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
     }
 
     fn shutdown(&self) -> LogResult<()> {
-        self.is_shutdown
-            .store(true, std::sync::atomic::Ordering::Relaxed);
         if let Ok(mut exporter) = self.exporter.lock() {
             exporter.shutdown();
             Ok(())
@@ -182,7 +169,6 @@ pub struct BatchLogProcessor {
     handle: Mutex<Option<thread::JoinHandle<()>>>,
     forceflush_timeout: Duration,
     shutdown_timeout: Duration,
-    is_shutdown: AtomicBool,
 
     // Track dropped logs - we'll log this at shutdown
     dropped_logs_count: AtomicUsize,
@@ -201,15 +187,6 @@ impl Debug for BatchLogProcessor {
 
 impl LogProcessor for BatchLogProcessor {
     fn emit(&self, record: &mut LogRecord, instrumentation: &InstrumentationScope) {
-        // noop after shutdown
-        if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
-            otel_warn!(
-                name: "BatchLogProcessor.Emit.ProcessorShutdown",
-                message = "BatchLogProcessor has been shutdown. No further logs will be emitted."
-            );
-            return;
-        }
-
         let result = self
             .message_sender
             .try_send(BatchMessage::ExportLog(Box::new((
@@ -229,11 +206,6 @@ impl LogProcessor for BatchLogProcessor {
     }
 
     fn force_flush(&self) -> LogResult<()> {
-        if self.is_shutdown.load(std::sync::atomic::Ordering::Relaxed) {
-            return LogResult::Err(LogError::Other(
-                "BatchLogProcessor is already shutdown".into(),
-            ));
-        }
         let (sender, receiver) = mpsc::sync_channel(1);
         self.message_sender
             .try_send(BatchMessage::ForceFlush(sender))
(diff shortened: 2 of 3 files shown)


opentelemetry-sdk/src/logs/log_emitter.rs
```diff
@@ -268,6 +268,14 @@ impl opentelemetry::logs::Logger for Logger {
 
     /// Emit a `LogRecord`.
     fn emit(&self, mut record: Self::LogRecord) {
+        if self.provider.inner.is_shutdown.load(Ordering::Relaxed) {
```
cijothomas, line 271: The issue with this is, this can affect throughput due to the contention introduced here. (Logs so far has no contention when using etw/user_events)....
Can you check stress test before and after?

I am unsure of a solution. Maybe don't check shutdown anywhere except in stdout like non-prod processors, and rely on mechanisms like export client/ etw etc. returning errors..
what harm can it cause 🤔

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -201,15 +187,6 @@ impl Debug for BatchLogProcessor {
 
 impl LogProcessor for BatchLogProcessor {
     fn emit(&self, record: &mut LogRecord, instrumentation: &InstrumentationScope) {
```
cijothomas, line 189: if an emit is called after shutdown, even if we don't do is_shutdown check, the channel would be already closed, so the error it returned is good enough?

cijothomas: something like
1. Don't check for is_shutdown.
2. Just export as usual.
3. Since the channel is closed, it'll error out.
4. Log that error.

No contention/perf cost for normal path. If logs are still emitted after shutdown, it clearly indicates some issue with user managing the lifetimes.

opentelemetry-sdk/src/logs/log_emitter.rs
lalitb: Oh yes, it doesn’t make sense to introduce the contention in the hot path, even if the contention is at atomic level.

utpilla: Since we are only reading the atomic variable and the variable's value does not change for the most part of application lifetime, it should not have any visible effect on the throughput.

We should be able to confirm that with the stress test.

lalitb: Yes, the effect of uncontested atomic would be negligible enough to be noticed over stress test. It is slightly more than normal bool-check, but much less than the perf associated in case of contention. I added the benchmark over logger:emit() in this PR, which shows the latency of 1-2ns:
`main branch`: 
```
logger_emit             time:   [37.916 ns 37.977 ns 38.077 ns]
                        change: [-0.2733% -0.0648% +0.1328%] (p = 0.58 > 0.05)
```

`PR branch`:
```
logger_emit             time:   [38.941 ns 39.027 ns 39.172 ns]
                        change: [+2.6756% +2.9861% +3.3292%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

lalitb: > I am unsure of a solution. Maybe don't check shutdown anywhere except in stdout like non-prod processors, and rely on mechanisms like export client/ etw etc. returning errors..
what harm can it cause 

The custom exporter connected to reentrant and simple need to have the shutdown in that case. As of now, etw and user_events don't do anything, so they will continue to emit even after shutdown invoked by user.

utpilla: > Yes, the effect of uncontested atomic would be negligible enough to be noticed over stress test.

@lalitb Just to confirm, you agree that the change in this PR (reading `is_shutdown`) is not introducing any contention, right?

lalitb: Yes, for uncontested _relaxed_ atomic read for **AtomicBool**, the cost is close to regular bool read - I observed latenct of 1-2 ns by adding this check, which I believe should be acceptable.

cijothomas: Thanks for the discussion. Agree that the cost is actually small. However I still suggest to leave this check and let us rely on underlying transport to error out and let us know if we export something after shutdown.

(A similar idea is followed in Otel .NET's Prometheus serializer - we pre-allocate byte array, and write to it, without checking explicitly each time if the array has enough capacity. We simply rely on .NET runtime to throw OutOfBoundsException to detect that. I believe we should leverage checks already done elsewhere instead of duplicating it, unless there is a strong reason to repeat.)

lalitb: > Agree that the cost is actually small. However I still suggest to leave this check and let us rely on underlying transport to error out and let us know if we export something after shutdown.

The custom exporter which can be connected to reentrant and simple processor need to handle the shutdown properly in that case. As of now, etw and user_events don't do anything in shutdown, so they will continue to emit even after shutdown invoked by user.

cijothomas: > > Agree that the cost is actually small. However I still suggest to leave this check and let us rely on underlying transport to error out and let us know if we export something after shutdown.
> 
> The custom exporter which can be connected to reentrant and simple processor need to handle the shutdown properly in that case. As of now, etw and user_events don't do anything in shutdown, so they will continue to emit even after shutdown invoked by user.

Irrespective of this, exporters are expected to handle shutdown right, if they have something that needs to shutdown? if etw/user-events exporter is not doing that, then we can fix that.

utpilla: I think adding this `if check` in `Logger::emit()` is not only a more efficient short-circuiting operation but it also simplifies things. It lets us offer a more robust guarantee:
 _"SDK would not do any wasteful computation once its shutdown (regardless of what processors/exporters are used)"_

cijothomas: Don't think it simplifies things. How do we determine the best place to do the is_shutdown check? Logger::emit() is one place.. how about the processor? how about the exporter? 

> "SDK would not do any wasteful computation once its shutdown (regardless of what processors/exporters are used)"

I don't know if we need to make such a statement? If user want to achieve that, they can drop the providers. (even then, we need to redesign something in Metrics to make it truly loops

utpilla: > Don't think it simplifies things.

The simplification I'm referring to is the effort required to achieving the expected shutdown behavior. Why call `emit()` on each processor, if the provider is shutdown. With having the "if check" in `Loggger::emit()` we effectively make it a no-op SDK, when the provider is shutdown.

> How do we determine the best place to do the is_shutdown check? Logger::emit() is one place.. how about the processor? how about the exporter? or even at the appender itself?

Since the shutdown is an SDK concept, it would only make sense to have the check in an SDK component. Appenders do not depend on SDK, so they are out of question. Maybe we can update the `LoggerProvider` trait to add another method `is_shutdown` which the appenders can use to avoid creating log records altogether. That's a different discussion though as it involves updating the public trait from API.

With the current state of code, among the SDK components: `Logger`, `Processor`, and `Exporter`, I think the best place to check for shutdown would be `Logger` as that's the first SDK component that interacts with the appender. By checking it in the `Logger`, we can avoid making calls to processors altogether.

> I don't know if we need to make such a statement?

We should have some stance about shutdown behavior. It could either be
1. The effectiveness of the shutdown depends on the processors/exporters used
2.  Or that shutdown works regardless, of the configured pipeline.

I think it's nice to have #2 and having the "if check" is a very easy way to achieve that. I don't see any disadvantage to doing that.

> If user want to achieve that, they can drop the providers.

I'm not sure if it's as straightforward. Since the provider is essentially an `Arc<LoggerProviderInner>`, the inner value would not be dropped until there are no more references to it. The appender layer keeps a clone of the provider which then moved to the `tracing_sbuscriber::registry()`. So simply dropping the provider might not help.

lalitb: > We should have some stance about shutdown behavior. It could either be
The effectiveness of the shutdown depends on the processors/exporters used
Or that shutdown works regardless, of the configured pipeline.
I think it's nice to have # 2 and having the "if check" is a very easy way to achieve that. I don't see any disadvantage to doing that.

I am also inclined with #2, as it ensures consistent shutdown behavior regardless of pipeline configuration or internal component details. Relying on assumptions, such as the batch processor closing its channel during shutdown, creates unnecessary dependencies that could break with future changes. An explicit provider/logger level check provides a more reliable solution, more importantly since cost of adding it is small.

cijothomas: > The simplification I'm referring to is the effort required to achieving the expected shutdown behavior. Why call emit() on each processor, if the provider is shutdown. With having the "if check" in Loggger::emit() we effectively make it a no-op SDK, when the provider is shutdown.

Without further redesign/refactoring, this won't make it a no-op sdk. There is non-trivial amount of work done even before emit() is invoked, to prepare LogRecord, and that'll continue even after shutdown. Yes, we save some work, but not quite no-op.

> We should have some stance about shutdown behavior. It could either be
> 
> 1 The effectiveness of the shutdown depends on the processors/exporters used
> 2 Or that shutdown works regardless, of the configured pipeline.

So far, we have been mostly sticking with 1 - we expect components to behave as expected. Example: We expect ObservableCallbacks to return in reasonable time, and if they don't return, Metrics SDK never recovers. Similarly we expect Exporters to return in reasonable time.
Not a strict comparison, but if we want to make a stance that Otel SDK works predictable, irrespective of the components plugged in, then it need a lot of redesign.
If we generally expect other components to behave properly, then why not continue rely on them to deal with shutdown too?

Another interesting thing we can do here (Similar to what [Otel .NET does for its console exporter](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.Console/ConsoleLogRecordExporter.cs#L39-L42)). If we make shutdown a pure exporter problem, we can have Stdout exporter continue to work even after shutdown is invoked, but can do something in the Stdout exporter to let user know that they have incorrect lifecycle management. IMHO, this is better than short-circuiting at up-stream sdk.(Of course internal logs can help, but unless users take care of avoiding self-logging, it can cause stack overflow due to cyclic loop)


> I'm not sure if it's as straightforward. Since the provider is essentially an Arc<LoggerProviderInner>, the inner value would not be dropped until there are no more references to it. The appender layer keeps a clone of the provider which then moved to the tracing_sbuscriber::registry(). So simply dropping the provider might not help.

Excellent point! This is something we need to have some guidance. (Even more challenging is the Metric case referred earlier).
If user decided to setup a tracing::subscriber globally (via init()), then they won't be able to really drop anything, unless tracing::subscriber provides a way. In other words, user is signing up to a global subscriber, so they should not expect to have the ability to drop things. tracing has alternate solutions to set a subscriber for a scope only, and they can drop the scope which in-turn can drop the logger_provider.

(https://github.com/open-telemetry/opentelemetry-rust/issues/1660 demonstrates this issue already)

cijothomas:
Closing as the changes are done over several previous PRs.

PR #2463, opened 2024-12-21
Move BatchLogProcessor with runtime to separate module

cijothomas:
This makes the migration consistent with PeriodicReader as users need to enable feature-flag + add new module name.
Experimental BatchProcessor_with_runtime is moved to a separate module.

### DIFF
--- a/opentelemetry-sdk/src/logs/mod.rs
+++ b/opentelemetry-sdk/src/logs/mod.rs
@@ -13,7 +13,7 @@ pub use log_processor::{
 pub use record::{LogRecord, TraceContext};
 
 #[cfg(feature = "experimental_logs_batch_log_processor_with_async_runtime")]
-/// Module for periodic reader with async runtime.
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -13,9 +13,8 @@ pub use log_processor::{
 pub use record::{LogRecord, TraceContext};
 
 #[cfg(feature = "experimental_logs_batch_log_processor_with_async_runtime")]
-pub use log_processor::{
-    BatchLogProcessorWithAsyncRuntime, BatchLogProcessorWithAsyncRuntimeBuilder,
-};
+/// Module for periodic reader with async runtime.
```
cijothomas, line 16:
```suggestion
/// Module for BatchLogProcessor with async runtime.
```

lalitb (approved):

PR #2469, opened 2024-12-24
Documenting memory usage/management by BatchLogProcessor 

lalitb:
## Changes

Felt it would be good to document it for better understanding of the memory management by BatchLogProcessor. This is not the doc-comment so won't go the docs.rs. If there is better place to add this, please suggest.


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -167,6 +167,72 @@ enum BatchMessage {
 
 /// A [`LogProcessor`] that buffers log records and reports
 /// them at a pre-configured interval from a dedicated background thread.
+// **Memory Management in BatchLogProcessor**
```
cijothomas, line 170: a lot of these can be moved off to docs.rs itself, as they are good for users to be aware of.

cijothomas (approved):
The contents are very good. Lets plan to move them to user-facing docs before stable release.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -167,6 +167,72 @@ enum BatchMessage {
...
+//
+// The `BatchLogProcessor` manages memory through the following stages of log processing:
+//
+// 1. **Record Ingestion**:
+//    - Each `LogRecord` is **cloned** upon entering the processor.
+//    - `LogRecordAttributes` utilize a hybrid memory model:
+//      - Attributes up to `PREALLOCATED_ATTRIBUTE_CAPACITY` are **stack-allocated**.
```
utpilla, line 177: Since `PREALLOCATED_ATTRIBUTE_CAPACITY` is not a public `const`, lets write the exact number of attributes here. I would suggest something like this:
```md
- The first 5 attributes are allocated on the stack.
- Any additional attributes are allocated on the heap in a dynamically growing vector.
```

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -167,6 +167,72 @@ enum BatchMessage {
...
+//
+// 3. **Worker Thread Storage**:
+//    - The worker thread maintains a pre-allocated `Vec` of boxed record pairs:
+//      - The vector’s capacity is fixed at `max_export_batch_size`.
+//    - Records are **moved** (not cloned) from the queue to the vector for processing.
+//
+// 4. **Export Process**:
+//    - A temporary `Vec` is created during the export process to hold ownership of the boxed records:
+//        - The contents of the worker thread's buffer (`Vec<Box<(LogRecord, InstrumentationScope)>>`)
+//          are moved into this temporary vector using `logs.split_off(0)`.
```
utpilla, line 199: @lalitb Thanks for writing up the documentation. It's very insightful.

I was wondering if we could avoid the allocation caused by `split_off()`. We should be able to reuse the existing `logs` vec when calling `export_with_timeout_sync` method. I created #2483 as a proof of concept.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -167,6 +167,72 @@ enum BatchMessage {
...
+//        - The original boxed records are dropped, releasing heap memory.
+//    - Export is triggered in the following scenarios:
+//        - When the batch size reaches `max_export_batch_size`.
+//        - When the scheduled delay timer expires.
+//        - When `force_flush` is called.
+//        - During processor shutdown.
+//
+/// 5. **Memory Limits**:
+///    - **Worst-Case Memory Usage**:
+///      - **Queue Memory** = `max_queue_size * size of boxed (LogRecord + InstrumentationScope)`.
```
utpilla, line 217: It'd be nice to mention the actual size of `LogRecord` and `InstrumentationScope` in bytes to get a better idea of the memory consumption.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -167,6 +167,72 @@ enum BatchMessage {
...
+//      - The vector’s capacity is fixed at `max_export_batch_size`.
+//    - Records are **moved** (not cloned) from the queue to the vector for processing.
+//
+// 4. **Export Process**:
+//    - A temporary `Vec` is created during the export process to hold ownership of the boxed records:
+//        - The contents of the worker thread's buffer (`Vec<Box<(LogRecord, InstrumentationScope)>>`)
+//          are moved into this temporary vector using `logs.split_off(0)`.
+//        - Ownership of the boxed records is transferred to the new vector, ensuring efficient
+//          memory usage without additional cloning.
+//    - The temporary vector is then used to construct references passed to the exporter via `LogBatch`.
```
utpilla, line 202: It'd be good to mention that there two temporary vectors created during each export:
1. One vector is created by `split_off(0)` (we could avoid this if we pursue #2483)
2. The second vector is created by collecting references from the first vector mentioned above.

I think we should be able to avoid the second vector creation as well by updating `LogBatch` constructors.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: @utpilla, yes that's a good point - the original buffer is not getting used/updated during export process, so it's good idea to reuse it for export, and then clear it up.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Done

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Good point, have added some numbers of memory usage in worst case scenario, with average size of LogRecord and InstrumentationScope.

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Have updated the document as per the latest optimizations, where both these vectors usage is removed.

lalitb:
> The contents are very good. Lets plan to move them to user-facing docs before stable release.

Have moved them as doc comment.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -263,7 +263,94 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
 ///     .build();
+/// ```
+///
+/// **Memory Management in BatchLogProcessor**
+///
+/// The `BatchLogProcessor` manages memory through the following stages of log processing:
+///
+/// 1. **Record Ingestion**:
+///    - Each `LogRecord` is **cloned** upon entering the processor.
+///    - `LogRecordAttributes` utilize a hybrid memory model:
```
cijothomas, line 274: How attributes are stored inside LogRecord is not a concern of BatchLogProcessor. It can be documented (and should be!), but not in processor doc.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -263,7 +263,94 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
+///      - First 5 attributes are **stack-allocated**.
+///      - Adding additional attributes trigger **heap allocation** in a dynamically growing vector.
+///    - The `LogRecord` and its associated `InstrumentationScope` are **boxed together**
+///      to allocate them on the heap before entering the queue. This means:
+///      - The `LogRecord`'s inline attributes (if any) are moved to the heap as part of the boxed structure.
+///      - Any dynamically allocated data already on the heap (e.g., strings, overflow attributes) remains unaffected.
+///      - Ownership of the boxed data is transferred to the queue, ensuring it can be processed independently of the original objects.
```
cijothomas, line 281: Not sure I follow this.. The data inside box is already a clone, so it is already independent of originals right?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -263,7 +263,94 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
+///      - Stores log records as **heap-allocated** `Box<(LogRecord, InstrumentationScope)>`.
+///      - The queue size is configurable and defined by `max_queue_size`.
+///      - If the queue is full:
+///        - New log records are **dropped**, and a warning is logged the first time this happens.
+///        - Dropped records are counted for reporting during shutdown.
+///
+///    - **Control Message Queue**:
+///      - Stores control messages (`BatchMessage`) to manage operations like exporting, force flushing, setting resources, and shutting down.
+///      - The control message queue has a fixed size (e.g., 64 messages).
+///      - Control messages are processed with higher priority, ensuring operational commands are handled promptly.
```
cijothomas, line 298: is there anything we do to process them with higher priority? This does not match the implementation.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -263,7 +263,94 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
+///      - The use of a separate control queue ensures that critical commands, such as `Shutdown`, are not lost even if the log record queue is full.
+///      - Messages supported include:
+///        - `ExportLog`: Triggers an immediate export of log records.
+///        - `ForceFlush`: Flushes all buffered log records to the exporter.
+///        - `SetResource`: Updates the exporter with a new resource.
```
cijothomas, line 303: this is misleading. There is no ability to update a resource. This could be misinterpreted as Resource can be changed and processor will be informed of the same, which is not true.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -263,7 +263,94 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
+///        - `Shutdown`: Cleans up and flushes logs before terminating the processor.
+///
+/// 3. **Worker Thread Storage**:
+///    - The worker thread maintains a pre-allocated `Vec` of boxed record pairs:
+///      - The vector’s capacity is fixed at `max_export_batch_size`.
+///    - Records are **moved** (not cloned) from the log record queue to the vector for processing.
+///
+/// 4. **Export Process**:
```
cijothomas, line 311: this looks too much details for end users to care about!

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -263,7 +263,94 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
+///          Calculation: `(2048 + 512) * (300 + 50) = 2560 * 350 = 896000 bytes = 896 KB`.
+///
+/// 6. **Key Notes on Memory Behavior**:
+///    - Boxing a `LogRecord` and `InstrumentationScope` moves the record to the heap,
+///      including stack-allocated attributes.
+///    - During the export process, records are moved from the log record queue to the worker thread’s vector.
+///    - No additional cloning or copying occurs during the export process, minimizing memory overhead while ensuring efficient handling of log records.
 ///
+/// 7. **Control Queue Prioritization**:
+///    - Control messages take precedence over log record processing to ensure timely execution of critical operations.
```
cijothomas, line 351: This does not look like how BatchProcessor works.

cijothomas (changes requested):
Lot of open comments before merge.
We can move a lot of inner workings to the design doc and leave end user facing ones here.

cijothomas:

PR #2471, opened 2024-12-25
json: ser/de bytes as base64 strings not an array of bytes

adriangb:
I believe this is how the proto <-> json spec defines that binary data should be handled: https://protobuf.dev/programming-guides/json/

I confirmed by exporting data from Python:

```python
from opentelemetry.proto.collector.trace.v1.trace_service_pb2 import ExportTraceServiceRequest
from opentelemetry.proto.common.v1.common_pb2 import AnyValue, KeyValue
from opentelemetry.proto.resource.v1.resource_pb2 import Resource
from opentelemetry.proto.trace.v1.trace_pb2 import ResourceSpans
from google.protobuf.json_format import MessageToJson


export = ExportTraceServiceRequest(
    resource_spans=[
        ResourceSpans(
            resource=Resource(
                attributes=[
                    KeyValue(key="key", value=AnyValue(bytes_value=b"value"))
                ]
            )
        )
    ]
)

json = MessageToJson(export, use_integers_for_enums=True)

print(json)

with open("export.json", "w") as f:
    f.write(json)
```

```json
{
  "resourceSpans": [
    {
      "resource": {
        "attributes": [
          {
            "key": "key",
            "value": {
              "bytesValue": "dmFsdWU="
            }
          }
        ]
      }
    }
  ]
}
```

### DIFF
--- a/opentelemetry-proto/src/proto.rs
+++ b/opentelemetry-proto/src/proto.rs
@@ -56,6 +56,11 @@ pub(crate) mod serializers {
                 map.serialize_entry("intValue", &i.to_string());
                 map.end()
             }
+            Some(Value::BytesValue(b)) => {
+                let mut map = serializer.serialize_map(Some(1))?;
+                map.serialize_entry("bytesValue", &base64::encode(b));
+                map.end()
+            }
             Some(value) => value.serialize(serializer),
             None => serializer.serialize_none(),
         }
@@ -127,8 +132,10 @@ pub(crate) mod serializers {
                             value = Some(any_value::Value::KvlistValue(kv));
                         }
                         "bytesValue" => {
-                            let bytes = map.next_value()?;
-                            value = Some(any_value::Value::BytesValue(bytes));
+                            let base64: String = map.next_value()?;
+                            let decoded = base64::decode(base64.as_bytes())
+                                .map_err(|e| de::Error::custom(e))?;
+                            value = Some(any_value::Value::BytesValue(decoded));
                         }
                         _ => {
                             //skip unknown keys, and handle error later.
--- a/opentelemetry-proto/Cargo.toml
+++ b/opentelemetry-proto/Cargo.toml
@@ -46,7 +46,7 @@ testing = ["opentelemetry/testing"]
 # add ons
 internal-logs = ["tracing"]
 with-schemars = ["schemars"]
-with-serde = ["serde", "hex"]
+with-serde = ["serde", "hex", "base64"]
 
 [dependencies]
 tonic = { workspace = true, optional = true, features = ["codegen", "prost"] }
@@ -57,6 +57,7 @@ schemars = { version = "0.8", optional = true }
 serde = { workspace = true, optional = true, features = ["serde_derive"] }
 hex = { version = "0.4.3", optional = true }
 tracing = {workspace = true, optional = true} # optional for opentelemetry internal logging
+base64 = { version = "0.22.1", optional = true }
 
 [dev-dependencies]
 opentelemetry = { features = ["testing"], path = "../opentelemetry" }
--- a/opentelemetry-proto/tests/json_serde.rs
+++ b/opentelemetry-proto/tests/json_serde.rs
@@ -275,18 +275,20 @@ mod json_serde {
                                 start_time_unix_nano: 1544712660000000000,
                                 end_time_unix_nano: 1544712661000000000,
                                 attributes: vec![
-                                  KeyValue {
-                                    key: String::from("my.span.attr"),
-                                    value: Some(AnyValue {
-                                        value: Some(Value::StringValue(String::from("some value"))),
-                                    }),
-                                  },
-                                  KeyValue {
-                                    key: String::from("my.span.bytes.attr"),
-                                    value: Some(AnyValue {
-                                        value: Some(Value::BytesValue(vec![0x80, 0x80, 0x80])),
-                                    }),
-                                  },
+                                    KeyValue {
+                                        key: String::from("my.span.attr"),
+                                        value: Some(AnyValue {
+                                            value: Some(Value::StringValue(String::from(
+                                                "some value",
+                                            ))),
+                                        }),
+                                    },
+                                    KeyValue {
+                                        key: String::from("my.span.bytes.attr"),
+                                        value: Some(AnyValue {
+                                            value: Some(Value::BytesValue(vec![0x80, 0x80, 0x80])),
+                                        }),
+                                    },
                                 ],
                                 dropped_attributes_count: 1,
                                 events: vec![Event {
--- a/opentelemetry-proto/tests/json_serde.rs
+++ b/opentelemetry-proto/tests/json_serde.rs
@@ -274,12 +274,20 @@ mod json_serde {
                                 kind: 2,
                                 start_time_unix_nano: 1544712660000000000,
                                 end_time_unix_nano: 1544712661000000000,
-                                attributes: vec![KeyValue {
+                                attributes: vec![
+                                  KeyValue {
                                     key: String::from("my.span.attr"),
                                     value: Some(AnyValue {
                                         value: Some(Value::StringValue(String::from("some value"))),
                                     }),
-                                }],
+                                  },
+                                  KeyValue {
+                                    key: String::from("my.span.bytes.attr"),
+                                    value: Some(AnyValue {
+                                        value: Some(Value::BytesValue(vec![0x80, 0x80, 0x80])),
+                                    }),
+                                  },
+                                ],
                                 dropped_attributes_count: 1,
                                 events: vec![Event {
                                     time_unix_nano: 1544712660500000000,
@@ -368,6 +376,12 @@ mod json_serde {
                   "key": "my.span.attr",
                   "value": {
                     "stringValue": "some value"
+                  }
+                },
+                {
+                  "key": "my.span.bytes.attr",
+                  "value": {
+                    "bytesValue": "gICA"
                   }
                 }
               ],


adriangb:
I'm not familiar with the codebase so I took a guess as to where to apply the fix. I tried to add an integration test but I could not get a failure case to fail a test so I'm not sure if that's even in the right place either. Some pointers would be appreciated.

adriangb:
Hi @lalitb @TommyCpp would one of you mind reviewing or pinging an appropriate reviewer (sorry if this is not relevant to you, I just went off of git history)? Thanks

opentelemetry-proto/src/proto.rs
```diff
@@ -182,6 +189,17 @@ pub(crate) mod serializers {
         let s: String = Deserialize::deserialize(deserializer)?;
         s.parse::<i64>().map_err(de::Error::custom)
     }
+
+    pub fn serialize_vec_u8_as_base64_string<S: Serializer>(v: &Vec<u8>, s: S) -> Result<S::Ok, S::Error> {
```
lalitb, line 193: Are we planning to use these methods in the future, given that they are currently unused?

opentelemetry-otlp/tests/integration_test/expected/traces.json
```diff
@@ -57,6 +57,12 @@
                   "value": {
                     "stringValue": "yes"
                   }
+                },
+                {
+                  "key": "data",
+                  "value": {
+                    "bytesValue": "gICA"
+                  }
```
lalitb, line 65: To make it complete, can we also add these in the json samples for logs and metrics?

opentelemetry-proto/src/proto.rs
adriangb: Yes but I'm not sure where or how. I'm having some trouble grokking the codebase because of all of the codegen.

opentelemetry-otlp/tests/integration_test/expected/traces.json
adriangb: Sure yes!

opentelemetry-proto/src/proto.rs
lalitb: I don't see bytes <-> base64 conversion required outside of `AnyValue::Value`. The only other candidates could have been `traceid`, `spanid` byte arrays but they are encoded as hex-str as per the [specs](https://github.com/open-telemetry/opentelemetry-proto/blob/v1.5.0/docs/specification.md#json-protobuf-encoding). So it should be safe to remove these methods for now, and bring back later if required.

adriangb: ok removed

opentelemetry-otlp/tests/integration_test/expected/traces.json
adriangb: Also I wasn't sure how to add an example in a logs `body` field. Can you suggest the right way to do that?

lalitb: > Added. But if these are the outputs, shouldn't tests be failing?

Yes these would indeed fail. Please run ./scripts/integration_tests.sh locally to validate that. 

ps. - disable the failing tests from #2495 before that, so that you only get errors from your changes.

lalitb: > Also I wasn't sure how to add an example in a logs body field. Can you suggest the right way to do that?

Good question. The log appender doesn't [handle ](https://github.com/open-telemetry/opentelemetry-rust/blob/42b4f2fbd8c9be6abf749929f8b93299f077ddea/opentelemetry-appender-tracing/src/layer.rs#L121) bytes as of now. The good test probably would be adding binary data to [resource](https://github.com/open-telemetry/opentelemetry-rust/blob/42b4f2fbd8c9be6abf749929f8b93299f077ddea/opentelemetry-otlp/tests/integration_test/tests/logs.rs#L29).

adriangb: That's exactly what I did 😄 https://github.com/open-telemetry/opentelemetry-rust/pull/2471/files#diff-4629c7e6add9d0ba42d5f44077d079b4c6f85ac86bc326a42106d67146ae4dbc

adriangb:
@lalitb I can't seem to get tests to agree on the order of items in attributes arrays. I'm not sure if they're getting sorted in one place but not another or something. Otherwise all tests seem to be passing. I'll point out that I had to implement bytes support for the rust sdk itself, maybe that should be it's own PR but if so it will block this PR.

adriangb:
Hi, just checking if there's anything else I can do to move this along. Thanks!

adriangb:
We've got some conflicts now. I'm happy to resolve them, or split this PR up, but I need a bit of guidance on how you'd like to proceed. Thanks!

opentelemetry/src/common.rs
```diff
@@ -161,6 +189,8 @@ pub enum Array {
     F64(Vec<f64>),
     /// Array of strings
     String(Vec<StringValue>),
+    /// Array of bytes
+    Bytes(Vec<BytesValue>),
```
cijothomas, line 193: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute
ByteArrary is not supported in the Otel Spec for common attributes...

But it is supported for Logs data - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any and this repo also supports it already.

@lalitb @TommyCpp  also to check if I am missing something obvious.

adriangb: Good point, I've removed Array(Bytes), I don't think it was strictly necessary for this PR

lalitb: Good point @cijothomas. The common attributes (used for `resources`, `traces`, and `metrics`) are not supposed to support bytes altogether. The use-case is only for logs, but then tokio-tracing and log crate doesn't support bytes as attributes. Just wondering if the example in the description is valid in that case or not - otel-python shouldn't allow populating the resource attributes with binary values.

adriangb: I don't see why bytes values wouldn't be allowed in resources specifically. In any case, the original ask was for span attributes but see https://github.com/open-telemetry/opentelemetry-rust/pull/2471#discussion_r1902025537

lalitb: @adriangb - I think we can remove the json example with base64 values in Resource. And also remove adding the Value::Bytes variant. The SDK should allow deserialize the Logs attributes containing the binary data, probably we can have validation for that.

lalitb:
> Hi folks pinging again. I'm willing to do the work here and this seems like an important issue (invalid data being generated) but I need some input from the dev team, this is my first PR to the repo. cc @cijothomas as main commiter.

@adriangb Sorry on delay on this. I believe the CI failure is for the ordering of resource attributes. But see my comment [here](https://github.com/open-telemetry/opentelemetry-rust/pull/2471#discussion_r1941828813) if that makes sense.

opentelemetry/src/common.rs
lalitb: > I don't see why bytes values wouldn't be allowed in resources specifically. In any case, the original ask was for span attributes but see [#2471 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/2471#discussion_r1902025537)

Yes i didn’t realise both span and resource attributes are driven from common attributes, and so can’t have byte value.

lalitb: Probably I was not very clear why bytes values are not allowed in resources and span attributes. As this is driven by the specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute

```
The attribute value is either:
- A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.
- An array of primitive type values. The array MUST be homogeneous, i.e., it MUST NOT contain values of different types.
```

and for log attributes - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#type-any
```
```

lalitb: > I think that bit of the spec just needs updating: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-type-mapping.md#byte-sequences

At proto level, there is common structure `AnyValue` used for all kind of attributes (including logs attributes), so it supports Binary type. However at SDK level, it shouldn't be possible to populate the binary values from resouce and span-attributes into this AnyValue. Refer - https://github.com/open-telemetry/opentelemetry-proto/issues/501

adriangb: I'll open an issue to clarify the spec - I believe the movement is towards allowing bytes in trace attributes.

In any case, I made the change you've suggested 😄

lalitb: > I'll open an issue to clarify the spec - I believe the movement is towards allowing bytes in trace attributes.

I agree, it would be better to get clarity from specs.

> In any case, I made the change you've suggested.

Thanks. It would be useful for deserialization of the log attributes :)

lalitb (approved):

cijothomas:
@adriangb Thanks for helping! 
I am curious to know if you are impacted by this or you are pro-actively addressing this? Given the appenders maintained in this repo does not today produce value of type array of bytes, I am curious - are there other ways people are using the crates?

adriangb:
I use the crate to load telemetry exported as JSON by a browser SDK that follows the OTLP JSON spec, so yes I was hitting this in production.

adriangb:
And thank _you_ for getting this across the line!

cijothomas:
> https://pydantic.dev/logfire

Thanks for sharing more context. We were earlier considering removing publishing the opentelemetry-proto crate itself, as it was considered an internal dependency for the opentelemetry-otlp itself! But later decided to keep it!

adriangb:
I'm glad you did!

lalitb:
Yes, interestingly, this crate is being utilized more than we initially expected. Another example: #2487.

cijothomas:
Created an issue to track getting it to 1.0
https://github.com/open-telemetry/opentelemetry-rust/issues/2615

@adriangb If you have bandwidth, please do check and see any gaps before we can make it a stable crate. (ofcource, the opentelemetry and opentelemetry_sdk crates must be 1.0 as a pre-req, and that is scheduled by mid year.)

PR #2477, opened 2024-12-26
Fix PeriodicReader shutdown to invoke shutdown on exporter

cijothomas:
Was a miss in the PeriodicReader implementation that uses own thread. This was already handled in the old periodic reader.

More refactoring is expected in this area, but this addresses the gap.


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -525,6 +532,31 @@ mod tests {
         }
     }
 
+    #[derive(Debug, Clone, Default)]
+    struct MockMetricExporter {
```
cijothomas, line 536: in a future PR, I expect to replace this with InMemoryExporter itself.

utpilla (approved):

utpilla (approved):

PR #2479, opened 2024-12-27
Provide default implementation for MetricExporter trait methods

cijothomas:


TommyCpp (commented):
I think we should require exporter author to think through the expected behavior rather than have a blank implementation. The benifit I see is we reduce some duplication in our code base. 

Also adding a default implemtation is not breaking change but removing the default implementation is. Thus I prefer to not add default implemtation. Overall I don't think the benifit is worth the risk

cijothomas:
Closing inactive PRs. This can be discussed via an issue.

Issue #2410, opened 2024-12-11
Exporting EventName


cijothomas:
Now we have a proper place in OTLP payload to store the [EventName](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/record.rs#L27).
Opening this issue to track adding this to the right place, when new Proto is released.
https://github.com/open-telemetry/opentelemetry-proto/pull/600#event-15610485928

PR #2480, opened 2024-12-27
Remove unnecessary featureflag for eventname

cijothomas:
Fully fixes https://github.com/open-telemetry/opentelemetry-rust/issues/2410 by removing feature flags that are no longer relevant.


lalitb (approved):
Thanks. LGTM. nit - There is some reference of this flag in stdout exporter too (ref - https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-rust%20populate-logs-event-name&type=code), good to remove it here or separately.

PR #2482, opened 2024-12-27
PeriodicReader - alternate shutdown implementation

cijothomas:
An alternate way of handling shutdown - this just sends a message to the background thread to initiate shutdown, and wait for thread handle to join. 
The handle.join() has some risks - as shutdown could get block indefinitely if the background thread is stuck in a export, and we cannot enforce timeout :(

An alternative way would be keep the existing way, but enforce recv_timeout, so that shutdown always ends in a fixed time, no matter whether background is stuck or not. Will send a PR with this approach as well, but either way, if a rogue exporter is used, background thread is still stuck. All we can ensure is shutdown gets a timed exit. (esp. important since shutdown can be triggered automatically when Drop() occurs...)

Not for merge, but to have a discussion about various approaches.


cijothomas:
Closing as this approach is not good.

PR #2485, opened 2024-12-29
Use bytes::Bytes as the HTTP request body in HttpClient.

mstyura:
This will allow pooling of buffers on caller side compared to usage of Vec<u8>.

## Fixes

Lack of possibility to reuse buffers which are consumed by `HttpClient` after request has been sent.

## Design discussion issue
Should old function be kept with deprecated annotation and hence new function had to have a different name or should I just introduce breaking change by changing signature of a function?

## Changes

I propose to change type of body of http request accepted by `HttpClient` trait from `Vec<u8>` to `bytes::Bytes`.
This should unblock pooling of buffers which can be used both for serialization & being http request bodies.
Something like this (pseudo-code to get the idea):
```
async fn export(&mut self, batch: Vec<SpanData>) -> ExportResult  {
    let mut buffer = pool.get();
    let request = self.build_request(&mut buffer, batch);
    let res = send_request(client, request);
    pool.return(buffer);
    res
}
```


mstyura:
Hello @cijothomas
Could you please take a look at this PR once you have time? 
Thanks a lot in advance!

opentelemetry-http/src/lib.rs
```diff
@@ -55,13 +55,24 @@ pub type HttpError = Box<dyn std::error::Error + Send + Sync + 'static>;
...
+        self.send_bytes(request.map(Into::into)).await
+    }
+
+    /// Send the specified HTTP request with `Bytes` payload.
+    ///
+    /// Returns the HTTP response including the status code and body.
+    ///
+    /// Returns an error if it can't connect to the server or the request could not be completed,
+    /// e.g. because of a timeout, infinite redirects, or a loss of connection.
+    async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError>;
```
TommyCpp, line 75: Just so I understand you want to create a pool of `Bytes` and reuse them across different `send/send_bytes` call?

mstyura: Yes, exactly, serialization of spans to export and send to Datadog is one of the largest memory allocation source of backend service I'm working on. I want to reuse buffers to serialize and sending over http between calls to exports.

lalitb: @mstyura  - will this enable buffer reuse in opentelemetry-otlp as well? It doesn't seem to be part of this PR, so ensuring.

mstyura: @lalitb I didn't have such plans initially, as I see proposed API is used by `oltp` and `zipkin` inside this repo, so they could leverage as well from buffer pooling. But as I don't use them in production I don't understand how heavily they actually allocate memory compared to DataDog exported I initially wanted to fix.
Do you want me to implement pooling for `oltp` and `zipkin`? If so I think it's preferable to do it as separate MR.

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -28,7 +28,7 @@ impl LogExporter for OtlpHttpClient {
                 .method(Method::POST)
                 .uri(&self.collector_endpoint)
                 .header(CONTENT_TYPE, content_type)
-                .body(body)
+                .body(body.into())
```
lalitb, line 31: I believe this simply takes ownership of the Vec's buffer and wraps it in a Bytes object? Just want to ensure there is no copy operation here.

mstyura: As I see from source code the only allocation possible is new Box allocated in case of buffer length != capacity: https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#964-996. The ownership of buffer is just transferred.

lalitb (approved):
LGTM. I believe it is safe to have dependency on [Bytes ](https://crates.io/crates/bytes)crate (as of now, it was only used for header injection), as it well maintained, and widely used (eg, by tokio, reqwest, and hyper).

mstyura:
And actually `Bytes` was already used in as type of response body :)

cijothomas:
> Is there anything I can do to facilitate merging this?

Sorry for the delay! Missed this un-intentionally!

Would you be sending a PR to do buffer pooling for otlp?

mstyura:
I don't have prepared PR to implement pooling for oltp, I've planned to implement it for Datadog (in contrib repo), because  I have app which uses Datadog exporter and I see in profiler it will benefit if pooling. Regarding oltp I could try implementing it, but not very soon.

cijothomas:
I think Datadog now supports OTLP natively, so you should be able to leverage OTLP Exporter. We were discussing the possibility of sunsetting the DD exporter in favor of OTLP (can't find that discussion anymore :( )

@scottgerring @julianocosta89 shared this doc : https://www.datadoghq.com/blog/monitor-rust-otel/

julianocosta89:
> I think Datadog now supports OTLP natively, so you should be able to leverage OTLP Exporter. We were discussing the possibility of sunsetting the DD exporter in favor of OTLP (can't find that discussion anymore :( )
> 
> 
> 
> @scottgerring @julianocosta89 shared this doc : https://www.datadoghq.com/blog/monitor-rust-otel/ 

Hello hello, just for reference, this thread had the initial discussion: https://cloud-native.slack.com/archives/C03GDP0H023/p1737396437600899

PR #2488, opened 2024-12-30
Avoid vec allocation during each export for BatchLogProcessor - Part 2

utpilla:
For batch export, we create an additional vec just to hold the references to `LogRecord` and `InstrumentationScope`. This is needed because the constructor for `LogBatch` only accepts a slice of tuples that are references to `LogRecord` and `InstrumentationScope`. `BatchLogProcessor` already has a vec of owned `LogRecord` and `InstrumentationScope`. This PR is an attempt to reuse that vec.

## Changes
- Add a new constructor for `LogBatch` which accepts a shared reference of a vec of owned `LogRecord` and `InstrumentationScope`.
- No changes to the public API

### DIFF
--- a/opentelemetry-sdk/src/export/logs/mod.rs
+++ b/opentelemetry-sdk/src/export/logs/mod.rs
@@ -23,12 +23,12 @@ pub struct LogBatch<'a> {
 
 /// The `LogBatchData` enum represents the data field of a `LogBatch`.
 /// It can either be:
-/// - A mutable reference to a vector of tuples, where each tuple consists of a `LogRecord` and an `InstrumentationScope`.
+/// - A shared reference to a vector of tuples, where each tuple consists of a `LogRecord` and an `InstrumentationScope`.
 /// - Or it can be a slice of tuples, where each tuple consists of a reference to a `LogRecord` and a reference to an `InstrumentationScope`.
 #[derive(Debug)]
 #[allow(clippy::vec_box)] // TODO: Revisit this. Clippy complains about using Box in a Vec, but it's done here for performant moves of the data between channel and the vec.
 enum LogBatchData<'a> {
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/export/logs/mod.rs
```diff
@@ -18,9 +18,18 @@ use std::fmt::Debug;
...
-    data: &'a [(&'a LogRecord, &'a InstrumentationScope)],
+    data: LogBatchData<'a>,
+}
+
+/// The `LogBatchData` enum represents the data field of a `LogBatch`.
+/// It can either be:
+/// - A shared reference to a vector of tuples, where each tuple consists of a `LogRecord` and an `InstrumentationScope`.
+/// - Or it can be a slice of tuples, where each tuple consists of a reference to a `LogRecord` and a reference to an `InstrumentationScope`.
+#[derive(Debug)]
+#[allow(clippy::vec_box)] // Clippy complains about using Box in a Vec, but it's done here for performant moves of the data between channel and the vec.
```
utpilla, line 29: Clippy complains about using [`Box` inside a `vec`](https://rust-lang.github.io/rust-clippy/master/index.html#/vec_box). Since `vec` allocates its elements on heap, boxing introduces unnecessary redirection.

The boxing happens in [`LogProcessor::emit()`](https://github.com/open-telemetry/opentelemetry-rust/blob/9a8ad95ee8a9f777427ad3742b4e10fd6acc298c/opentelemetry-sdk/src/logs/log_processor.rs#L281-L286). I think we do this for faster movement of the data from the channel to the vec. We may need to revisit this. It's outside of the scope of this PR.

cijothomas (approved):
Okay to special case to let BatchProcessor avoid alloc. Given no public api change, lets merge.

PR #2489, opened 2024-12-31
[Metrics SDK] Reduce the scope of write lock

utpilla:
- Write lock is only needed to get hold of the right tracker
- We don't need to hold the write lock while updating the trackers. Updates for Histograms are more involved than other instruments. There is no need to hold the write lock for the duration of the update.

## Changes
- Drop the `RwLockWriteGuard` once we have the right tracker and then update the tracker


utpilla:
This won't work properly for Delta temporality.

Issue #2478, opened 2024-12-27
BatchProcessor shutdown issue


cijothomas:
BatchProcessor relies on a channel to pass messages (which includes Log/Span, and also control signals like Shutdown/Flush). If the channel is full, then Shutdown() signals will get lost. For example, the buffer got 500 items, and the thread is doing blocking export() call. During this time, 2000 items got into the channel. Any new items are dropped, which is by design. But if a shutdown is initiated during this time, that message is lost as well. When the thread is back to processing the queue, it'll drain the normal telemetry items, but shutdown signal is lost forever. Even when shutdown message is not lost, the message gets to the last position in the queue, so it'll be processed after all other export messages are processed (potential after multiple exports()).

May need to use a separate channel for control flow like shutdown/flush to mitigate the issue. This may not be easy to fully eliminate due to blocking call being made, and anything occurring during this time is hard to respond to.

PeriodicReader has a similar problem, but less severe as the Shutdown message will not be processed, but reacted to immediately after the blocking export is done.

lalitb:
Good point, I observed this while doing unit-tests for dedicated thread span processor. The separate control channel for shutdown/flush looks to be good idea - reading control channel first. However, there may now be some log events in data-queue arrived before shutdown message. We need to somehow ensure to also drain those message for export, apart from one existing in internal vector.

utpilla:
There is another issue with the current implementation of `BatchLogProcessor`. The background thread is woken up for every log record that gets emitted. This is not ideal. We want the background thread to have low utilization so that it doesn't aggressively competer with the user's application threads. We only want to wake up the background thread on these conditions:
1. The number of log records in the channel has reached `max_export_batch_size`.
2. The `scheduled_delay` interval has elapsed.
3. `force_flush` is called.
4. `shut_down` is called.

I have created #2494 to show one possible approach to handle both the shutdown issue and the background thread's utilization issue.

ThomsonTan:
For missing flush/forceflush, probably we can just use a CondVar instead of a separate channel? As it should be fine to aggregate multiple flushes into one.

Ideally, we could replace the channel here by our own circular buffer like C++/C# to avoid the potential excessive synchronization.

ThomsonTan:
[...]

PR #2494, opened 2025-01-03
Refactor BatchLogProcessor

utpilla:
Towards #2478 

## Changes
- Refactor `BatchLogProcessor` to wake up the background thread ONLY when needed https://github.com/open-telemetry/opentelemetry-rust/issues/2478#issuecomment-2568633280
- Address the shutdown issue

## Performance
There is a significant performance improvement mainly because of `mpsc:Receiver` not being excessively active (unlike how it is on main branch). It looks like there is a significant amount of synchronization overhead with senders and receivers actively working on the channel all the time. With this PR, the channel receivers are only woken up when needed so the logging threads incur a lesser synchronization overhead cost.

I updated the [`Logging_Comparable_To_Appender`](https://github.com/open-telemetry/opentelemetry-rust/blob/37d2e51c23afea4d318a31294f3229963c6e88dc/opentelemetry-sdk/benches/log.rs#L103-L127) benchmark and [logs stress test](https://github.com/open-telemetry/opentelemetry-rust/blob/main/stress/src/logs.rs) to use a `BatchProcessor` with a no-op exporter and ran the tests. Here are the results:

- **25% improvement in throughput** with stress test results going up from 12M (main) to 15 M
- **~46% improvement in benchmarks** with the latency measurement going down from 420ns (main) to 227 ns.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -199,8 +199,8 @@ impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 enum BatchMessage {
-    /// Export logs, called when the log is emitted.
-    ExportLog(Box<(LogRecord, InstrumentationScope)>),
+    /// This ONLY sent when the number of logs records in the data channel has reached `max_export_batch_size`.
+    ExportLog(Arc<AtomicBool>),
     /// ForceFlush flushes the current buffer to the exporter.
     ForceFlush(mpsc::SyncSender<ExportResult>),
     /// Shut down the worker thread, push all logs in buffer to the exporter.
@@ -208,6 +208,8 @@ enum BatchMessage {
     /// Set the resource for the exporter.
     SetResource(Arc<Resource>),
 }
+
+type LogsData = Box<(LogRecord, InstrumentationScope)>;
 
 /// The `BatchLogProcessor` collects finished logs in a buffer and exports them
 /// in batches to the configured `LogExporter`. This processor is ideal for
@@ -246,11 +248,15 @@ enum BatchMessage {
 ///     .build();
 ///
 pub struct BatchLogProcessor {
-    message_sender: SyncSender<BatchMessage>,
+    logs_sender: SyncSender<LogsData>, // Data channel to store log records and instrumentation scopes
+    message_sender: SyncSender<BatchMessage>, // Control channel to store control messages for the worker thread
     handle: Mutex<Option<thread::JoinHandle<()>>>,
     forceflush_timeout: Duration,
     shutdown_timeout: Duration,
     is_shutdown: AtomicBool,
+    export_log_message_sent: Arc<AtomicBool>,
+    current_batch_size: Arc<AtomicUsize>,
+    max_export_batch_size: usize,
 
     // Track dropped logs - we'll log this at shutdown
     dropped_logs_count: AtomicUsize,
@@ -279,11 +285,19 @@ impl LogProcessor for BatchLogProcessor {
         }
 
         let result = self
-            .message_sender
-            .try_send(BatchMessage::ExportLog(Box::new((
-                record.clone(),
-                instrumentation.clone(),
-            ))));
+            .logs_sender
+            .try_send(Box::new((record.clone(), instrumentation.clone())));
+
+        if self.current_batch_size.fetch_add(1, Ordering::Relaxed) >= self.max_export_batch_size {
+            // Check if the a control message for exporting logs is already sent to the worker thread.
+            // If not, send a control message to export logs.
+            // `export_log_message_sent` is set to false ONLY when the worker thread has processed the control message.
+            if !self.export_log_message_sent.swap(true, Ordering::Relaxed) {
+                let _ = self.message_sender.try_send(BatchMessage::ExportLog(
+                    self.export_log_message_sent.clone(),
+                )); // TODO: Handle error
+            }
+        }
 
         if result.is_err() {
             // Increment dropped logs count. The first time we have to drop a log,
@@ -388,8 +402,12 @@ impl BatchLogProcessor {
     where
         E: LogExporter + Send + Sync + 'static,
     {
-        let (message_sender, message_receiver) = mpsc::sync_channel(config.max_queue_size);
+        let (logs_sender, logs_receiver) = mpsc::sync_channel::<LogsData>(config.max_queue_size);
+        let (message_sender, message_receiver) = mpsc::sync_channel::<BatchMessage>(64); // Is this a reasonable bound?
         let max_queue_size = config.max_queue_size;
+        let max_export_batch_size = config.max_export_batch_size;
+        let current_batch_size = Arc::new(AtomicUsize::new(0));
+        let current_batch_size_for_thread = current_batch_size.clone();
 
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Logs.BatchProcessor".to_string())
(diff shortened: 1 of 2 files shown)


lalitb:
Excellent approach! Please make this ready for review once you feel it's ready :)

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -279,11 +285,19 @@ impl LogProcessor for BatchLogProcessor {
...
         let result = self
-            .message_sender
-            .try_send(BatchMessage::ExportLog(Box::new((
-                record.clone(),
-                instrumentation.clone(),
-            ))));
+            .logs_sender
+            .try_send(Box::new((record.clone(), instrumentation.clone())));
+
+        if self.current_batch_size.fetch_add(1, Ordering::Relaxed) >= self.max_export_batch_size {
```
lalitb, line 291: This will add small perf overhead with user thread. I believe this should be find to have, given the advantage this allow us to wake up the processor-thread only when export is required. But good to have some measure of the cost.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -388,8 +402,12 @@ impl BatchLogProcessor {
     where
         E: LogExporter + Send + Sync + 'static,
     {
-        let (message_sender, message_receiver) = mpsc::sync_channel(config.max_queue_size);
+        let (logs_sender, logs_receiver) = mpsc::sync_channel::<LogsData>(config.max_queue_size);
+        let (message_sender, message_receiver) = mpsc::sync_channel::<BatchMessage>(64); // Is this a reasonable bound?
```
lalitb, line 406: 64 for signal (flush, shutdown, and trigger export). I don't think user should be doing flush/shutdown too often. But could there be scenario if the code-generated signal to trigger export can ever get flooded.

utpilla: > But could there be scenario for the code-generated signal to trigger export can ever get flooded.

There can never be more than one code-generated signal in the channel. Threads calling `emit()` cannot send the `ExportLog` signal if there's one already in the channel.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: I ran the stress test and benchmarks, and the performance seems to have improved. I believe it's because the logger threads incur lesser synchronization costs as their receiver isn't actively working on the channel. I updated the PR description with the setup and results. Let me know what you think. It would be great if you could verify it on your machine as well.

lalitb: Yes, I see similar improvement with stress test with your branch - 17.2 iterations/sec (main) to 20.5 iterations/sec.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -292,6 +295,20 @@ impl LogProcessor for BatchLogProcessor {
                 otel_warn!(name: "BatchLogProcessor.LogDroppingStarted",
                     message = "BatchLogProcessor dropped a LogRecord due to queue full/internal errors. No further log will be emitted for further drops until Shutdown. During Shutdown time, a log will be emitted with exact count of total logs dropped.");
             }
+            return;
+        }
+
+        // At this point, sending the log record to the data channel was successful.
+        // Increment the current batch size and check if it has reached the max export batch size.
+        if self.current_batch_size.fetch_add(1, Ordering::Relaxed) >= self.max_export_batch_size {
```
lalitb, line 303: should this be changed to
```suggestion
        if self.current_batch_size.fetch_add(1, Ordering::Relaxed) + 1 >= self.max_export_batch_size {
```


as `fetch_add` returns value before increment. So as per the pr logic, if `max_export_batch_size` = 2000, the export signal would be sent when `current_batch_size` = 2001, which doesn't seem correct.
`

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -292,6 +295,20 @@ impl LogProcessor for BatchLogProcessor {
...
+            // Check if the a control message for exporting logs is already sent to the worker thread.
+            // If not, send a control message to export logs.
+            // `export_log_message_sent` is set to false ONLY when the worker thread has processed the control message.
+            if !self.export_log_message_sent.swap(true, Ordering::Relaxed) {
```
lalitb, line 307: should we use `compare_exchange` instead of `swap`. This will avoid redundant writes if the `export_log_message_sent` is already true?

opentelemetry-sdk/src/logs/log_processor.rs
lalitb: Make sense. Probably add a TODO to decide on the optimal value.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: Thanks for catching that!

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: I used `swap` because it's less verbose and `swap` uses `compare_exchange` internally anyway. I pushed some new changes. I have added a new if check using atomic load to avoid aggressively calling `swap`. We would now call `swap` only if the atomic load value returned `false`. Let me know what you think.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -402,6 +440,42 @@ impl BatchLogProcessor {
                 );
                 let mut last_export_time = Instant::now();
                 let mut logs = Vec::with_capacity(config.max_export_batch_size);
+                let current_batch_size = current_batch_size_for_thread;
+
+                // This method gets upto `max_export_batch_size` amount of logs from the channel and exports them.
+                // It returns the result of the export operation.
+                // It expects the logs vec to be empty when it's called.
```
cijothomas, line 447: we can get rid of `logs` vec and instead create a fresh array and pass its slice to export method.?

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -410,37 +484,44 @@ impl BatchLogProcessor {
...
+
+                            // Reset the export log message sent flag now it has has been processed.
+                            export_log_message_sent.store(false, Ordering::Relaxed);
                         }
                         Ok(BatchMessage::ForceFlush(sender)) => {
                             otel_debug!(name: "BatchLogProcessor.ExportingDueToForceFlush");
-                            let result = export_with_timeout_sync(
-                                config.max_export_timeout,
-                                &mut exporter,
+                            let result = get_logs_and_export(
```
cijothomas, line 506: this will not be sufficient as it'll only export upto max_export_batch_size logs. We need to export everything received upto flush invocation.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -410,37 +484,44 @@ impl BatchLogProcessor {
...
+                                &config,
                             );
                             let _ = sender.send(result);
                         }
                         Ok(BatchMessage::Shutdown(sender)) => {
                             otel_debug!(name: "BatchLogProcessor.ExportingDueToShutdown");
-                            let result = export_with_timeout_sync(
-                                config.max_export_timeout,
-                                &mut exporter,
+                            let result = get_logs_and_export(
```
cijothomas, line 518: same as flush - we need to ensure logs are continued to be exported until it finishes all logs received upto the point shutdown was invoked.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: For using array, we need to know the size of the array beforehand which we don't because we don't know how many items are present in the channel. It might be less than `max_export_batch_size`.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: > We need to export everything received upto flush invocation.

That should be taken care of by the signal sent by the logging thread whenever we have `max_export_batch_size` number of logs. That means that the background thread would be working on the signal sent by the logger thread before working on ForceFlush/Shutdown thread.

utpilla:
> Also, I am unsure of the perf testing, as any perf test would instantly fill up the channel, so majority of items would be taking the "drop_due_to_full" path. 

Since we are using a no-op exporter, the channel should be getting emptied really quick as well. But yes, it would be nice to verify that objectively instead of manual reasoning.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: Had an offline discussion about this with @cijothomas, and we agreed that there is a possibility of logs just residing in the channel without being exported. This should be fixable. I would address this in the next PR.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: we can allocate an array of max_export_batch_size, and slice it at the required length. Anyway, this is not a blocker.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: resolving for follow up.

opentelemetry-sdk/src/logs/log_processor.rs
utpilla: Always stack allocating a big array even to export very few logs seems wasteful.

What's the benefit of using an array instead of a `vec` here? We are reusing the same `vec` for each export anyway so allocation isn't a concern.

Issue #2442, opened 2024-12-17
Metrics SDK - can we cleanup after a shutdown?


cijothomas:
Though shutdown prevents metrics from ever getting collected/exported, they are still kept in the HashMaps/Aggregations. There is no check in hot path to see if the provider is already shutdown (it may be too expensive to do that check). This maybe okay, but opening an issue to discuss more.

Also, given the Counter/Meter from the API use a `Arc<dyn>` object pointing to something from the SDK, it is not easily possible to do much cleanup in the SDK, unless the user explicitly drops the Counter/Meters themselves.

fraillt:
As you have mentioned, I would prefer to avoid having this check in hot path as well.
However, we should probably clear `Pipeline` (more precisely `inner: Mutex<PipelineInner>`) during shutdown, so that memory can be cleaned up once user explicitly drops meters themselves.

utpilla:
> Also, given the Counter/Meter from the API use a `Arc<dyn>` object pointing to something from the SDK, it is not easily possible to do much cleanup in the SDK

We can use `Weak` instead of `Arc` in the instruments to address this issue. I created #2497 as a proof of concept to show that instruments can be updated to respect the `MeterProvider` shutdown. We can do the same for `Meter`s as well. In the PR, I have only updated `Counter` to use `Weak`.

This comes with a performance cost though as the `Weak` to `Arc` upgrade happens on the hot path. On my machine, the benchmarks saw a 5% decline and the stress test throughput went from 9M to 7M iterations per second.

> unless the user explicitly drops the Counter/Meters themselves.

This would not always be possible. Especially, if the users are using some instrumentation library and have no visibility and access to the instruments and Meters.

lalitb:
> This comes with a performance cost though as the Weak to Arc upgrade happens on the hot path. On my machine, the benchmarks saw a 5% decline and the stress test throughput went from 9M to 7M iterations per second.

This is not related, but just to add. The Arc upgrade indeed is perf heavy operation, which was the reason it was removed from Logger/LoggerProvider in #1455.

cijothomas:
TODO: Also check this for other signals like Logs. Add tests to confirm the current behavior and if they are behaving as expected or not.

cijothomas:
Removing from stable milestone. This is nice-to-have, but not a blocker for SDK stable release.

PR #2497, opened 2025-01-05
[Metrics][POC] Use Weak instead of Arc in Counter to respect MeterProvider shutdown

utpilla:
Towards #2442 

## Changes
- Use `Weak` instead of `Arc` in instruments to respect `MeterProvider` shutdown

Note: This also requires `MeterProvider` instance to be valid for metrics to be emitted.

### DIFF
--- a/opentelemetry/src/metrics/instruments/counter.rs
+++ b/opentelemetry/src/metrics/instruments/counter.rs
@@ -1,6 +1,6 @@
 use crate::KeyValue;
 use core::fmt;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 
 use super::SyncInstrument;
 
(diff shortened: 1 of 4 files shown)


opentelemetry/src/metrics/instruments/counter.rs
```diff
@@ -25,12 +25,16 @@ where
...
     /// Create a new counter.
     pub fn new(inner: Arc<dyn SyncInstrument<T> + Send + Sync>) -> Self {
-        Counter(inner)
+        Counter(Arc::downgrade(&inner))
     }
 
     /// Records an increment to the counter.
     pub fn add(&self, value: T, attributes: &[KeyValue]) {
-        self.0.measure(value, attributes)
+        if let Some(inner) = self.0.upgrade() {
```
cijothomas, line 33: how much is the perf cost for this upgrade as opposed to storing Arc itself?

utpilla: There is a considerable perf cost. Check this https://github.com/open-telemetry/opentelemetry-rust/issues/2478#issuecomment-2573491710.

cijothomas:
Closing this, as the perf cost is not acceptable in the hot path.

PR #2502, opened 2025-01-08
Update metrics benchmarks

utpilla:
## Changes
- Update Histogram benchmark to use `iter_batched` to exclude the setup time from the measurement. Similar to what #2233 did for Counter
- Add a benchmark to test attribute values that non-static string slices to both Counter and Histogram (Address https://github.com/open-telemetry/opentelemetry-rust/pull/2089#discussion_r1751155425)

### DIFF
--- a/opentelemetry-sdk/benches/metrics_counter.rs
+++ b/opentelemetry-sdk/benches/metrics_counter.rs
@@ -53,18 +53,11 @@ fn criterion_benchmark(c: &mut Criterion) {
     counter_add_sorted(c);
     counter_add_unsorted(c);
 
-    let attribute_values = [
-        "value1".to_owned(),
-        "value2".to_owned(),
-        "value3".to_owned(),
-        "value4".to_owned(),
-        "value5".to_owned(),
-        "value6".to_owned(),
-        "value7".to_owned(),
-        "value8".to_owned(),
-        "value9".to_owned(),
-        "value10".to_owned(),
-    ];
+    let attribute_values: [String; 10] = (1..=10)
+        .map(|i| format!("value{}", i))
+        .collect::<Vec<String>>()
+        .try_into()
+        .expect("Expected a Vec of length 10");
 
     counter_add_sorted_with_non_static_values(c, attribute_values);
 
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/benches/metrics_counter.rs
```diff
@@ -127,6 +144,54 @@ fn counter_add_unsorted(c: &mut Criterion) {
...
+                let index_first_attribute = rands[0];
+                let index_second_attribute = rands[1];
+                let index_third_attribute = rands[2];
+                let index_fourth_attribute = rands[3];
+                counter.add(
+                    1,
+                    &[
+                        KeyValue::new(
+                            "attribute1",
+                            attribute_values[index_first_attribute].as_str().to_owned(),
```
utpilla, line 173: With #2089, we would be able to avoid the string allocation
```suggestion
                            attribute_values[index_first_attribute].as_str(),
```

opentelemetry-sdk/benches/metrics_counter.rs
```diff
@@ -51,6 +52,22 @@ fn create_counter(name: &'static str) -> Counter<u64> {
...
+        "value2".to_owned(),
+        "value3".to_owned(),
+        "value4".to_owned(),
+        "value5".to_owned(),
+        "value6".to_owned(),
+        "value7".to_owned(),
+        "value8".to_owned(),
+        "value9".to_owned(),
+        "value10".to_owned(),
+    ];
```
TommyCpp, line 67: We can just range over 1 to 10, prefix with `value` and then `map(to_owned)` here?
```rust
let attribute_values: [String; 10] = (1..=10)
    .map(|i| format!("value{}", i))
    .collect::<Vec<String>>()
    .try_into()
    .expect("Expected a Vec of length 10");
```

utpilla: Thanks for the suggestion! I have updated the code.

lalitb (approved):

TommyCpp (approved):

PR #2504, opened 2025-01-10
feat: export SDK crate version as constant

the-wondersmith:
## Changes

PR adds a simple constant for the SDK crate's version (pulled directly from the crate's manifest [[ref](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates:~:text=of%20your%20package.-,CARGO_PKG_VERSION,-%E2%80%94%20The%20full%20version)]). 

This is specifically useful in conjunction with things like populating the value of [`telemetry.sdk.version`](https://docs.rs/opentelemetry-semantic-conventions/latest/opentelemetry_semantic_conventions/attribute/constant.TELEMETRY_SDK_VERSION.html) when building a [`Resource`](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/struct.Resource.html) (for example).

* ✅ [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* `N/A` Unit tests added/updated (if applicable)
* `N/A` Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* `N/A` Changes in public API reviewed (if applicable)


opentelemetry-sdk/src/lib.rs
```diff
@@ -150,3 +150,7 @@ pub mod util;
 pub use resource::Resource;
 
 pub mod error;
+
+/// The crate's version string, useful for populating [`Resource`] keys like
+/// [`telemetry.sdk.version`](https://docs.rs/opentelemetry-semantic-conventions/latest/opentelemetry_semantic_conventions/attribute/constant.TELEMETRY_SDK_VERSION.html)
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
```
cijothomas, line 156: we try to keep public API as lean as possible, so unless there is a strong reason to expose this, we should not.
Users rarely need this as this is automatically populated.

cijothomas (changes requested):
Thanks for your interest in contributing to the project!
I've requested changes. Please see https://github.com/open-telemetry/opentelemetry-rust/pull/2504/files#r1913817079

opentelemetry-sdk/src/lib.rs
the-wondersmith: FWIW, my specific usecase involves setting the value of `telemetry.sdk.version` (as defined in [`opentelemetry_semantic_conventions`](https://docs.rs/opentelemetry-semantic-conventions/0.27.0/opentelemetry_semantic_conventions/attribute/constant.TELEMETRY_SDK_VERSION.html)), e.x.:

Screenshot 2025-01-16 at 3 10 26 PM

Specifically, that's in a library crate intended for use in *other* "actual" crates to simplify setting up `tracing` with export to an OTLP collector.

Including this simple constant just makes it much easier to "stay in line" with the defined convention without having to constantly remember to update that static string whenever the SDK version changes. Plus, `cargo` *does* do a certain amount of fudging with version numbers, so unless I pin *explicitly* to something like `opentelemetry_sdk = "= 0.27.0"` the **_actual_** version of the SDK that ends up getting used could be anything in the `0.27.*` range. Much easier and more reliable to just... export a simple string constant, don't you think? 😅

Oh, and not to potentially talk past the close, but this simple constant won't meaningfully inflate the "weight" of the crate's public API at all - unless it's explicitly used in some downstream crate it won't be included in compiled binaries. In terms of compile-time overhead, cargo takes the time to determine and set that value unconditionally, so it's no extra overhead there either 🙂

the-wondersmith: @cijothomas ^bump 🙂

the-wondersmith: @cijothomas **_^bump_** 🙃

cijothomas: Sorry was busy with few other issues, will recheck today, thanks for reminder!

the-wondersmith: No problem at all! Thank you for taking the time 😅

lalitb: @the-wondersmith Can't you use [TelemetryResourceDetector](https://github.com/open-telemetry/opentelemetry-rust/blob/85206105f8368d4930c2e9437640bbf43ba133b6/opentelemetry-sdk/src/resource/telemetry.rs) to populate the otel-sdk version? As an example here - 
https://github.com/open-telemetry/opentelemetry-rust/blob/85206105f8368d4930c2e9437640bbf43ba133b6/opentelemetry-prometheus/tests/integration_test.rs#L365-L371

the-wondersmith: @lalitb **_Yup_**, sure can! 😁

the-wondersmith:
Closing in favor of @lalitb's [suggestion](https://github.com/open-telemetry/opentelemetry-rust/pull/2504#discussion_r1936042877)

PR #2506, opened 2025-01-11
Metrics aggregate collector generic over temporality

fraillt:
Prerequisite for #2328

## Changes

Provided several interfaces that allow to provide "ValueMap" implementation for specific temporality.
The key element is `Collector` (which implements `AggregateCollector`). It provides common functionality for all aggregations (Sum, LastValue, Histogram, etc...):
* provide attribute-set filtering functionality (for measurement).
* for specific temporality:
  * time initialization
  * aggregate data creation/initialization
  * collection of aggregation data
This allow to have very clean concrete implementations. I have refactored `ExponentialHistogram` to use these new interfaces as an example.

### DIFF
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -53,6 +53,14 @@ pub trait Aggregation: fmt::Debug + any::Any + Send + Sync {
     fn as_mut(&mut self) -> &mut dyn any::Any;
 }
 
+/// Allow to access data points of an [Aggregation].
+pub trait AggregationDataPoints {
+    /// The type of data point in the aggregation.
+    type Point;
+    /// The data points of the aggregation.
+    fn points(&mut self) -> &mut Vec<Self::Point>;
+}
+
 /// DataPoint is a single data point in a time series.
 #[derive(Debug, PartialEq)]
 pub struct GaugeDataPoint<T> {
@@ -225,6 +233,14 @@ impl<T: fmt::Debug + Send + Sync + 'static> Aggregation for ExponentialHistogram
     }
     fn as_mut(&mut self) -> &mut dyn any::Any {
         self
+    }
+}
+
+impl<T> AggregationDataPoints for ExponentialHistogram<T> {
+    type Point = ExponentialHistogramDataPoint<T>;
+
+    fn points(&mut self) -> &mut Vec<Self::Point> {
+        &mut self.data_points
     }
 }
 
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -54,11 +54,11 @@ pub trait Aggregation: fmt::Debug + any::Any + Send + Sync {
 }
 
 /// Allow to access data points of an [Aggregation].
-pub trait AggregationDataPoints {
+pub(crate) trait AggregationDataPoints {
     /// The type of data point in the aggregation.
-    type Point;
+    type DataPoint;
     /// The data points of the aggregation.
-    fn points(&mut self) -> &mut Vec<Self::Point>;
+    fn points(&mut self) -> &mut Vec<Self::DataPoint>;
 }
 
 /// DataPoint is a single data point in a time series.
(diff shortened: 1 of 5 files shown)


opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -53,6 +53,14 @@ pub trait Aggregation: fmt::Debug + any::Any + Send + Sync {
     fn as_mut(&mut self) -> &mut dyn any::Any;
 }
 
+/// Allow to access data points of an [Aggregation].
+pub trait AggregationDataPoints {
+    /// The type of data point in the aggregation.
+    type Point;
```
utpilla, line 59:
```suggestion
    type DataPoint;
```

opentelemetry-sdk/src/metrics/internal/collector.rs
```diff
@@ -0,0 +1,195 @@
...
+    where
+        MapFn: FnMut(Vec<KeyValue>, &Self::Aggr) -> DP;
+}
+
+/// Higher level abstraction (compared to [`AggregateMap`]) that also does the filtering and collection into aggregation data
+pub(crate) trait AggregateCollector: Send + Sync + 'static {
+    const TEMPORALITY: Temporality;
+    type Aggr: Aggregator;
+
+    fn measure(&self, value: <Self::Aggr as Aggregator>::PreComputedValue, attributes: &[KeyValue]);
```
utpilla, line 30: It's getting a bit too confusing with these new coupled traits. Could we separate the concerns here? Could we update `Collector` trait to only have collect related methods and `AggergateMap` trait to only have update related methods.

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
```diff
@@ -333,135 +330,55 @@ where
...
 /// measurements were made in.
-pub(crate) struct ExpoHistogram<T: Number> {
-    value_map: ValueMap<Mutex<ExpoHistogramDataPoint<T>>>,
-    init_time: AggregateTimeInitiator,
-    temporality: Temporality,
-    filter: AttributeSetFilter,
-    record_sum: bool,
-    record_min_max: bool,
+pub(crate) struct ExpoHistogram<AC> {
+    pub(crate) aggregate_collector: AC,
```
utpilla, line 352: It's better to use the new method. That way we cannot mistakenly update these fields outside of the struct creation.

opentelemetry-sdk/src/metrics/internal/collector.rs
fraillt: Not sure if it's possible to have this separation, basically this whole things is like an onion.
Starting from the center:
1. `trait AggregateMap` - implement the core data structure optimized for updating AND collecting aggregates for specific temporality (e.g. `DeltaValueMap`, and `CumulativeValueMap`)
2. `trait AggregateCollector` - provides filtering attributes AND collecting impl specific `DataPoints` into `dyn Aggregation` (including time initialization for specific temporality). There's only one implementation, - `Collector`. The reason we have this trait, is because `Collector` itself is also generic (currently only over `trait AggregateMap`, but it might also be generic over "aggregate-filter"), which makes trait bounds for implementations to be very verbose).
3. specific aggregate e.g. (Sum, LastValue, Histogram, etc..) - implements `Measure` and `ComputeAggregation` traits, which are used by SDK. Also implements `InitAggregationData` which is used by `AggregateCollector` to make whole collection phase reusable.

The thing that I like about this is that 1) and 2) (e.g. impls for `trait AggregateMap` and `trait AggregateCollector` ) is common code for absolutely all aggregates. The only aggregate specific logic is in three traits: `Measure`, `ComputeAggregation`, `InitAggregationData`.

Regarding splitting...:
* `trait AggregateMap` cannot be split, because it's implementation has optimized internal structure that only it knows how to update and collect
* since `trait AggregateCollector` depends on `trait AggregateMap`, it also cannot be split into few instances.
* same fields within specific implementations can be used in measure and collect phases, as well. (e.g. `Histogram::bounds` field), so splitting concrete implementation not optimal as well.
* so the `AggregateFns` implementation is probably most efficient way to do it. (e.g. Arc<specific impl> and clone, so one instance implements Measure,- another `ComputeAggregation`).

BTW, I agree that this gets a bit confusing, but I think better naming would solve the problem here... but that's the hardest part :)

fraillt: Mm... maybe I have one more idea.
Instead of having specific aggregation "the final thing" (which implements `Measure` and `ComputeAggregation`). I could make that `Collector` is the final thing, and specific aggregations would have a trait which will be used to provide specific functionality...
I will probably create separate revision for that, (so we could compare both at the same time). 
Ok?

fraillt: #2530 Same thing,- new design.
IMO it turned out very good, I like this new design much more.
If you have same opinion, then we can close this PR.

fraillt:
#2530 has similar functionality, with IMO better design.

PR #2510, opened 2025-01-14
Fix BatchLogProcessor

utpilla:
Addresses https://github.com/open-telemetry/opentelemetry-rust/pull/2494#discussion_r1913786362

## Changes
- For every export trigger, find the export target: the number of logs in the channel at the time the trigger is invoked
- Export logs until the export target is met

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -306,7 +306,7 @@ impl LogProcessor for BatchLogProcessor {
             // If not, send a control message to export logs.
             // `export_log_message_sent` is set to false ONLY when the worker thread has processed the control message.
 
-            if !self.export_log_message_sent.load(Ordering::Acquire) {
+            if !self.export_log_message_sent.load(Ordering::Relaxed) {
                 // This is a cost-efficient check as atomic load operations do not require exclusive access to cache line.
                 // Perform atomic swap to `export_log_message_sent` ONLY when the atomic load operation above returns false.
                 // Atomic swap/compare_exchange operations require exclusive access to cache line on most processor architectures.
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -457,23 +457,31 @@ impl BatchLogProcessor {
...
-                        config.max_export_timeout,
-                        exporter,
-                        logs,
-                        last_export_time,
-                    ); // This method clears the logs vec after exporting
+                        let count_of_logs = logs.len(); // Count of logs that will be exported
+                        total_exported_logs += count_of_logs;
 
-                    current_batch_size.fetch_sub(count_of_logs, Ordering::Relaxed);
+                        result = export_with_timeout_sync(
```
utpilla, line 476: If multiple exports happen within the same invocation of this method, we could have different results (for example if one of them fails).  Currently this method would only return the last result observed.

cijothomas: 512 items emitted, emit thread sends export-message signal.
background-thread-start-exporting the 512
513-2048 items got into channel while export was going on, but they didn't trigger export-message as one was already pending.
background-thread finishes 0-512. Won't export the 513-2048 until flush/shutdown/timer is triggered.

^ This is something we still need to fix, right? What this PR fixes is that, if any of the trigger occurs, it takes care of exporting 513-2048 instead of previously stopping at 513-1024.

Maybe we need to modify emit, so that every time current size hits the batch_size, send a message to background thread, irrespective of whether one is pending or not. With this, background thread will immediately start next batch when done with one export (if a export-message was pending)

utpilla: Good catch again! Yes, we would need to fix that issue.

> Maybe we need to modify emit, so that every time current size hits the batch_size, send a message to background thread, irrespective of whether one is pending or not.

This might lead to issues with missing out on Shutdown message. If the update threads aggressively send `ExportLog` message to the channel, we might run out of capacity to receive a Shutdown message.

I was thinking a simpler fix could be to set `export_log_message_sent` to `false` as soon as we receive `ExportLog` message. That is set it to `false` before exporting. That way, if the update threads send data while the export is going on, they can still send the `ExportLog` message. What do you think?

cijothomas (approved):

PR #2514, opened 2025-01-15
Remove redundant shutdown check in batchlogprocessor

cijothomas:
The check is already done by the LogProvider.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -353,7 +353,8 @@ impl LogProcessor for BatchLogProcessor {
 
     fn shutdown(&self) -> LogResult<()> {
         // Set is_shutdown to true
-        self.is_shutdown.store(true, std::sync::atomic::Ordering::Relaxed);
+        self.is_shutdown
+            .store(true, std::sync::atomic::Ordering::Relaxed);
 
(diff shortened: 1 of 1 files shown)


lalitb:
So, we don't go with the approach in #2462 - i.e., having `is_shutdown` check in logger::emit()?

cijothomas:
> So, we don't go with the approach in #2462 - i.e., having `is_shutdown` check in logger::emit()?

Irrespective of that, this check is redundant, as processor's shutdown() cannot be called without going through provider, which already checks for duplicate shutdown.

lalitb:
> Irrespective of that, this check is redundant, as processor's shutdown() cannot be called without going through provider, which already checks for duplicate shutdown.

yes, #2462 does same thing, by removing the `is_shutdown` flag altogether from the processors and let `logger::emit()` also check for the flag.

lalitb (approved):

cijothomas:
> > Irrespective of that, this check is redundant, as processor's shutdown() cannot be called without going through provider, which already checks for duplicate shutdown.
> 
> yes, #2462 does same thing, by removing the `is_shutdown` flag altogether from the processors and let `logger::emit()` also check for the flag.

Got it. I'll revisit 2462 again. We discussed this a bit in yesterday's SIG too. I will share more comments/thoughts on that PR.

PR #2515, opened 2025-01-15
feat(trace): impl ShouldSample for Box<dyn ShouldSample>

Oliboy50:
Fixes #

## Changes

Implement `ShouldSample` for `Box<dyn ShouldSample>` because: 
- users of this crate cannot do it themselves (because of the Rust orphan rule)
- while working on a small wrapper around `opentelemetry-sdk` but `tracer_provider_builder.with_sampler(...)` currently only accepts a `T: crate::trace::ShouldSample + 'static` generic parameter... but I can't introduce generic types in my wrapper (for some reasons) and therefore I would like to use a `dyn ShouldSample` instead... and I currently can't do that 

I'll update this PR to fix the "Merge requirement checklist" below, only if you are OK with it in the first place.

### DIFF
--- a/opentelemetry-sdk/src/trace/sampler.rs
+++ b/opentelemetry-sdk/src/trace/sampler.rs
@@ -102,6 +102,21 @@ impl Clone for Box<dyn ShouldSample> {
     }
 }
 
+impl ShouldSample for Box<dyn ShouldSample> {
+    fn should_sample(
+        &self,
+        parent_context: Option<&Context>,
+        trace_id: TraceId,
+        name: &str,
+        span_kind: &SpanKind,
+        attributes: &[KeyValue],
+        links: &[Link],
+    ) -> SamplingResult {
+        self.as_ref()
+            .should_sample(parent_context, trace_id, name, span_kind, attributes, links)
+    }
+}
+
 /// Default Sampling options
 ///
 /// The [built-in samplers] allow for simple decisions. For more complex scenarios consider


utpilla:
> ## Changes
> Implement `ShouldSample` for `Box<dyn ShouldSample>` because:
> 
> * users of this crate cannot do it themselves (because of the Rust orphan rule)
> * while working on a small wrapper around `opentelemetry-sdk` but `tracer_provider_builder.with_sampler(...)` currently only accepts a `T: crate::trace::ShouldSample + 'static` generic parameter... but I can't introduce generic types in my wrapper (for some reasons) and therefore I would like to use a `dyn ShouldSample` instead... and I currently can't do that

If I understand the issue correctly, your wrapper version of `with_sampler` method can only take trait objects and not impl traits.
If that's the case, could you use a wrapper struct to get around this issue? Something like this:

```rust
fn with_sampler_custom(sampler: Box<dyn ShouldSample>) -> opentelemetry_sdk::trace::Builder {
    opentelemetry_sdk::trace::TracerProvider::builder().with_sampler(WrapperSampler(sampler))
}

#[derive(Debug, Clone)]
struct WrapperSampler(Box<dyn ShouldSample>);

impl ShouldSample for WrapperSampler {
    fn should_sample(
        &self,
        parent_context: Option<&Context>,
        trace_id: opentelemetry::trace::TraceId,
        name: &str,
        span_kind: &SpanKind,
        attributes: &[KeyValue],
        links: &[opentelemetry::trace::Link],
    ) -> opentelemetry::trace::SamplingResult {
        let sampler = self.0.as_ref();
        let result =
            sampler.should_sample(parent_context, trace_id, name, span_kind, attributes, links);
        result
    }
}
```

Oliboy50:
@utpilla you are right, I don't know why I didn't think about using a NewType pattern to get around the orphan rule 🤦 

I close my PR then, thank you for your help ❤️

PR #2520, opened 2025-01-15
Integration test for non tokio main

### LINKED ISSUE #2402: Validate tonic can be used in blocking scenarios/non-tokio context

cijothomas:
Fixes #2402 

## Changes


cijothomas:
marking draft to fix and isolate the test.

opentelemetry-otlp/tests/integration_test/tests/logs.rs
```diff
@@ -113,29 +132,28 @@ mod logtests {
...
-
-    LogsAsserter::new(left, right).assert();
-
-    assert!(File::open(result).unwrap().metadata().unwrap().size() > 0);
+pub fn assert_logs_results(result: &str, expected_content: &str) -> Result<()> {
+    let file = File::open(result)?;
+    let mut contents = String::new();
+    let mut reader = std::io::BufReader::new(&file);
+    reader.read_to_string(&mut contents)?;
+    assert!(contents.contains(expected_content));
```
lalitb, line 156: This doesn't compare the complete JSON, but only the UUID attribute?

lalitb: I don't think this is the right approach in that case. The test should be capable to validate all the attributes of different types.

lalitb (approved):

opentelemetry-otlp/tests/integration_test/tests/logs.rs
lalitb: I added an alternate approach here - #2521, running the tests serially, and letting it validate the complete JSON. We can keep the existing tests as it is, and add new tests to only validate the uuid attribute. This way we have both the scenarios tested?

cijothomas: yes, we need that too. This one is to quickly validate that the shutdown issues are indeed fixed

PR #2524, opened 2025-01-17
Add OTLP Exporting support matrix

cijothomas:
Draft.
To be filled in with works vs not-works.
Opening to get early feedback.

### DIFF
--- a/opentelemetry-otlp/tests/integration_test/README.md
+++ b/opentelemetry-otlp/tests/integration_test/README.md
@@ -15,3 +15,42 @@ popped back out into the files output by the collector.
 * Docker, for the test container
 * TCP/4317 and TCP/4318 free on your local machine. If you are running another
   collector, you'll need to stop it for the tests to run.
+
+## Test Matrix
+
+| **Exporter Type**   | **Provider Creation**     | **Logging**           | **Network Clients**        | **Status** |
+|----------------------|---------------------------|------------------------|----------------------------|------------|
+| **Simple**           | Inside `tokio::Runtime`  | Inside `tokio::Runtime` | `tonic`                    | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Inside `tokio::Runtime` | `reqwest`                  | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Inside `tokio::Runtime` | `reqwest-blocking`         | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Inside `tokio::Runtime` | `hyper`                    | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Outside `tokio::Runtime` | `tonic`                   | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Outside `tokio::Runtime` | `reqwest`                 | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Outside `tokio::Runtime` | `reqwest-blocking`        | Works      |
+| **Simple**           | Inside `tokio::Runtime`  | Outside `tokio::Runtime` | `hyper`                   | Works      |
+| **Simple**           | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `tonic`                   | Works      |
+| **Simple**           | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `reqwest`                 | Works      |
+| **Simple**           | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `reqwest-blocking`        | Works      |
+| **Simple**           | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `hyper`                   | Works      |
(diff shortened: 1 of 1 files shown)


opentelemetry-otlp/tests/integration_test/README.md
```diff
@@ -15,3 +15,42 @@ popped back out into the files output by the collector.
 * Docker, for the test container
 * TCP/4317 and TCP/4318 free on your local machine. If you are running another
   collector, you'll need to stop it for the tests to run.
+
+## Test Matrix
+
+| **Exporter Type**   | **Provider Creation**     | **Logging**           | **Network Clients**        | **Status** |
+|----------------------|---------------------------|------------------------|----------------------------|------------|
+| **Simple**           | Inside `tokio::Runtime`  | Inside `tokio::Runtime` | `tonic`                    | Works      |
```
utpilla, line 23: nit: It would be nice to use check marks here similar to the test matrix for metrics: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/test.md

image


```suggestion
| **Simple**           | Inside `tokio::Runtime`  | Inside `tokio::Runtime` | `tonic`                    | :white_check_mark: |
```

lalitb:
Should we simplify this by grouping common scenarios. Eg,
```
| Simple / Batch   | Inside `tokio::Runtime`  | Outside `tokio::Runtime` | All clients          | Works      |   
```

opentelemetry-otlp/tests/integration_test/README.md
```diff
@@ -15,3 +15,42 @@ popped back out into the files output by the collector.
 * Docker, for the test container
 * TCP/4317 and TCP/4318 free on your local machine. If you are running another
   collector, you'll need to stop it for the tests to run.
+
+## Test Matrix
+
+| **Exporter Type**   | **Provider Creation**     | **Logging**           | **Network Clients**        | **Status** |
```
scottgerring, line 21: I find it difficult to parse the table for "exhaustiveness". It might be easier if we break it down along some other axis into multiple tables.

Perhaps by network client - as then we'll get 4 tables of 8 rows?

opentelemetry-otlp/tests/integration_test/README.md
```diff
@@ -15,3 +15,42 @@ popped back out into the files output by the collector.
 * Docker, for the test container
 * TCP/4317 and TCP/4318 free on your local machine. If you are running another
   collector, you'll need to stop it for the tests to run.
+
+## Test Matrix
+
```
scottgerring, line 20: Should we add:

```suggestion
We aim to support and provide integration test coverage for every permutation of [A/B/C]
```

... or something along these lines, so its clear what the desired state here is? It'd be helpful to identify gaps. I think optimally we'd link out to some architectural or design goal elsewhere, if we had one handy.

opentelemetry-otlp/tests/integration_test/README.md
```diff
@@ -15,3 +15,42 @@ popped back out into the files output by the collector.
...
+| **Batch**            | Inside `tokio::Runtime`  | Outside `tokio::Runtime` | `hyper`                  | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `tonic`                  | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `reqwest`                | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `reqwest-blocking`       | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Inside `tokio::Runtime` | `hyper`                  | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Outside `tokio::Runtime` | `tonic`                 | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Outside `tokio::Runtime` | `reqwest`               | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Outside `tokio::Runtime` | `reqwest-blocking`      | Works      |
+| **Batch**            | Outside `tokio::Runtime` | Outside `tokio::Runtime` | `hyper`                 | Works      |
+
```
scottgerring, line 55: This is not really related to the PR at hand, but I also remain unconvinced that we are gaining much by supporting hyper/tonic/reqwest all at once; we started discussing this elsewhere (but I can't find the PR anymore). I struggle to find a strong need for anything more than a single client for async and sync calling, as it is 1/ an implementation detail that does not concern the caller and 2/ likely won't save space in the app anyway because of the way rust manages dep resolution.

cijothomas: Similar to async part, Rust does not have a std library for making Http calls or asynchronous runtimes... So we need to provide some ability for users to bring-their-own, and provide a default implementation for at least 1 common one.

Issue #2493, opened 2025-01-02
BatchSpanProcessor optimization


lalitb:
Inline with the changes done for BatchLogProcessor -#2488, #2494

utpilla:
Add #2483 as well to the list.

cijothomas:
This can be closed with #2526

PR #2526, opened 2025-01-21
BatchSpanProcessor optimizations  - Separate control signal queue, and wake up background thread only when required.

lalitb:
Fixes: Towards #2493 

## Changes

Similar to the BatchLogProcessor in #2494 :
  - wake up the background thread ONLY when needed
  - Separate control signal queue  ( to ensure that the shutdown signal is not lost when message queue is full).

### DIFF
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -221,12 +221,15 @@ use futures_executor::block_on;
 use std::sync::mpsc::sync_channel;
 use std::sync::mpsc::RecvTimeoutError;
 use std::sync::mpsc::SyncSender;
+use std::sync::mpsc::Receiver;
+use crate::export::trace::ExportResult;
 
 /// Messages exchanged between the main thread and the background thread.
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -279,44 +292,71 @@ impl BatchSpanProcessor {
...
+                                let _ = Self::get_spans_and_export(
+                                    &span_receiver,
+                                    &mut exporter,
+                                    &mut spans,
+                                    &mut last_export_time,
+                                    &current_batch_size,
+                                    &config,
+                                );
+                                // Reset the export span message sent flag now it has has been processed.
+                                export_span_message_sent.store(false, Ordering::Relaxed);
```
cijothomas, line 308: this reset need to happen before the exporting. Just like logs.

cijothomas (approved):
LGTM, need to make this fix: https://github.com/open-telemetry/opentelemetry-rust/pull/2526/files#r1924235350 before merge

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Good point. Update accordingly.

PR #2529, opened 2025-01-22
Doc additions for Simple and Batch processors

### LINKED ISSUE #2189: SimpleExporter is not useable with OTLP/Http
SimpleExporter is not useable with OTLP/HTTP (Reqwest client and Hyper client both fails), as export just hangs forever.

Minimal repro when using Reqwest
```rust
use tracing::info;
use tracing_subscriber::layer::SubscriberExt;
use tracing_subscriber::Registry;
use tracing_subscriber::util::SubscriberInitExt;
use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
use opentelemetry_otlp::WithExportConfig;

fn main() {
     let exporter = opentelemetry_otlp::new_exporter().http();
    let logger_provider = opentelemetry_otlp::new_pipeline()
        .logging()
        .with_exporter(
            exporter
                .with_protocol(Protocol::HttpBinary)
                .with_endpoint("http://localhost:4318/v1/logs"),
        )
        .install_simple()
        .unwrap();

    // Create a new OpenTelemetryTracingBridge using the above LoggerProvider.
    let layer = OpenTelemetryTracingBridge::new(&logger_provider);
    tracing_subscriber::registry()
        .with(layer)
        .init();

    info!(name: "my-event", target: "my-target", "hello from {}. My price is {}", "apple", 1.99);

    let _ = logger_provider.shutdown();
}
```

Minimal repro when using Hyper
```
let exporter = opentelemetry_otlp::new_exporter().http();
    let exporter = exporter.with_http_client(hyper::HyperClient::default());
    let logger_provider = opentelemetry_otlp::new_pipeline()
        .logging()
        .with_exporter(
            exporter
                .with_protocol(Protocol::HttpBinary)
                .with_endpoint("http://localhost:4318/v1/logs"),
        )
        .install_simple()
        .unwrap();
``

### PR DESCRIPTION
**Pull request by cijothomas**
Fixes #2189

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -109,7 +109,7 @@ pub trait LogProcessor: Send + Sync + Debug {
 /// When using this processor with the OTLP Exporter, the following exporter
 /// features are supported:
 /// - `grpc-tonic`: This requires LoggerProvider to be created within a tokio
-///   runtime. Logs can emitted from any thread, including tokio runtime
+///   runtime. Logs can be emitted from any thread, including tokio runtime
 ///   threads.
 /// - `reqwest-blocking-client`: LoggerProvider may be created anywhere, but
 ///   logs must be emitted from a non-tokio runtime thread.
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -105,7 +105,7 @@ pub trait LogProcessor: Send + Sync + Debug {
 
 /// A [`LogProcessor`] designed for testing and debugging purpose, that immediately
 /// exports log records as they are emitted. Log records are exported synchronously
-/// in the same thread that emits the log record. 
+/// in the same thread that emits the log record.
 /// When using this processor with the OTLP Exporter, the following exporter
 /// features are supported:
 /// - `grpc-tonic`: This requires LoggerProvider to be created within a tokio
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -104,7 +104,18 @@ pub trait LogProcessor: Send + Sync + Debug {
 }
 
 /// A [`LogProcessor`] designed for testing and debugging purpose, that immediately
-/// exports log records as they are emitted.
+/// exports log records as they are emitted. Log records are exported synchronously
+/// in the same thread that emits the log record. 
+/// When using this processor with the OTLP Exporter, the following exporter
+/// features are supported:
+/// - `grpc-tonic`: This requires LoggerProvider to be created within a tokio
+///   runtime. Logs can emitted from any thread, including tokio runtime
+///   threads.
+/// - `reqwest-blocking-client`: LoggerProvider may be created anywhere, but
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: opentelemetry-sdk/src/trace/span_processor.rs [hunk 1]
```diff
@@ -102,6 +102,17 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
 /// `SpanExporter`, as soon as they are finished, without any batching. This is
 /// typically useful for debugging and testing. For scenarios requiring higher
 /// performance/throughput, consider using [BatchSpanProcessor].
+/// Log records are exported synchronously
```
lalitb, line 105:
```suggestion
/// Spans are exported synchronously
```

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -102,6 +102,17 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
...
+/// in the same thread that emits the log record.
+/// When using this processor with the OTLP Exporter, the following exporter
+/// features are supported:
+/// - `grpc-tonic`: This requires TracerProvider to be created within a tokio
+///   runtime. Spans can emitted from any thread, including tokio runtime
```
lalitb, line 110:
```suggestion
///   runtime. Spans can be emitted from any thread, including tokio runtime
```

lalitb (approved):
Nicely documented.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -104,7 +104,18 @@ pub trait LogProcessor: Send + Sync + Debug {
...
-/// exports log records as they are emitted.
+/// exports log records as they are emitted. Log records are exported synchronously
+/// in the same thread that emits the log record.
+/// When using this processor with the OTLP Exporter, the following exporter
+/// features are supported:
+/// - `grpc-tonic`: This requires LoggerProvider to be created within a tokio
+///   runtime. Logs can emitted from any thread, including tokio runtime
+///   threads.
+/// - `reqwest-blocking-client`: LoggerProvider may be created anywhere, but
+///   logs must be emitted from a non-tokio runtime thread.
```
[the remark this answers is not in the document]
cijothomas: lets improve this based on any feedback we get. I am neutral about this currently, since we talk about supported features flags in general, without talking about **why** something is not supported.

ratijas:
This is actually so very well hidden, considering that scattered examples mostly use builder method `with_simple_exporter` which doesn't really expose the issue with `SimpleSpanProcessor`; so you have to find this thing among the docs, or read through the source code

ratijas:
Also, what exactly does this mean?

> The following feature flags offer additional configurations on http:

As far as my reverse-engineering skills go, these feature flags offer no configuration. In fact, it's just a big `if { #[cfg(...)] #[cfg(...)] #[cfg(...)] }` in another crate which will kinda panic if you got the `cfg` part wrong. Which you _will_ get wrong on the first try, because you forgot to remove `default-features` before adding the right flavor of client. AFAICT nothing mentions those client features are supposed to be mutually exclusive either. You _will_ get it wrong on the second try too, because you just broke the build, so now you have to add all the default features back one by one — all but the `"reqwest-blocking-client"` obviously. Phew…

PR #2538, opened 2025-01-23
Move inmemory exporter for logs to diff namespace

cijothomas:
`opentelemetry_sdk::testing::logs::{InMemoryLogExporter, InMemoryLogExporterBuilder};`
to 
  `opentelemetry_sdk::logs::{InMemoryLogExporter, InMemoryLogExporterBuilder};`

This is already under a feature-flag "testing", so this avoids the unnecessary "testing" module name.


lalitb (approved):
Thanks, we need similar for other in-memory exporter too.

Issue #2527, opened 2025-01-17
[Feature]: A Dev Feature for Naming Convention used in Windows Performance Counters


anujnegi270:
### Related Problems?

Our team is an early adopter of OpenTelemetry Rust for ingesting metrics, but we face challenges when dealing with Windows Performance Counters (PerfC) due to differences in naming conventions between OpenTelemetry and PerfC. Specifically, performance counters like `\Processor Information(_Total)\% Processor Time` (just an example) cannot be directly sent using OpenTelemetry because its naming structure doesn't align with OpenTelemetry’s expected format.

### What component are you working with?

opentelemetry-user-events-metrics, N/A

### Describe the solution you'd like:

If there's a way to bypass the name check or instead add a Performance Counter support in opentelemetry as a feature, this can be done!

cijothomas:
Moving to otel-rust repo, as naming validation is part of Otel API/SDK, not exporter.

anujnegi270:
[like]  Anuj Negi reacted to your message:
________________________________

cijothomas:
Sharing one issue in the spec which is also asking to relax naming requirement: https://github.com/open-telemetry/opentelemetry-specification/issues/4371

While we wait spec movement, @lalitb is it okay to offer a "experimental_metrics_disable_name_validation" feature flag to bypass the validation to unblock users? I don't see any other way to report Windows Performance Counters which uses names disallowed by Otel....

lalitb:
@cijo i think it make sense to have it under experimental flag.

cijothomas:
Okay.

@anujnegi270 Feel free to send a PR to bypass the validation, under a feature flag "experimental_metrics_disable_name_validation".

anujnegi270:
[like]  Anuj Negi reacted to your message:
________________________________

PR #2543, opened 2025-01-23
Disabling the Instrument Name Validation under a new feature flag

anujnegi270:
A fix for the feature [[Feature]: A Dev Feature for Naming Convention used in Windows Performance Counters · Issue #2527 · open-telemetry/opentelemetry-rust ](https://github.com/open-telemetry/opentelemetry-rust/issues/2527) and a related [issue ](https://github.com/open-telemetry/opentelemetry-specification/issues/4371). 

Fixes #
This fix is done under a feature flag `experimental_metrics_disable_name_validation`, where all types of instrument name are acceptable except empty names and names exceeding the length limit.

## Changes
Defining the validation check function under the feature flag to bypass and return Ok(). 

* [✔️] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [✔️] Unit tests added/updated (if applicable)
* [✔️] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes


opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -572,6 +575,33 @@ fn validate_bucket_boundaries(boundaries: &[f64]) -> MetricResult<()> {
...
 
+// Regex pattern for Windows Performance Counter names
+// #[cfg(feature = "experimental_metrics_disable_name_validation")]
+// fn is_valid_windows_perf_counter_name(name: &str) -> bool {
+//     let pattern = r"^\\[A-Za-z0-9_ ]+\\[A-Za-z0-9_ /%-]+$";
+//     let re = Regex::new(pattern).unwrap();
+//     re.is_match(name)
+// }
+
+#[cfg(feature = "experimental_metrics_disable_name_validation")]
```
cijothomas, line 586: to stick with the meaning of the feature flag, which says "disable name validation", disable all validations when this feature flag is enabled, not selective.

anujnegi270: Sounds right! But what about empty names? Should I also remove the empty name validation?

anujnegi270: Removed all names check validation under feature flag - `experimental_metrics_disable_name_validation`

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -572,6 +583,21 @@ fn validate_bucket_boundaries(boundaries: &[f64]) -> MetricResult<()> {
     Ok(())
 }
 
+// Regex pattern for Windows Performance Counter names
+// #[cfg(feature = "experimental_metrics_disable_name_validation")]
+// fn is_valid_windows_perf_counter_name(name: &str) -> bool {
+//     let pattern = r"^\\[A-Za-z0-9_ ]+\\[A-Za-z0-9_ /%-]+$";
+//     let re = Regex::new(pattern).unwrap();
+//     re.is_match(name)
+// }
```
lalitb, line 592: nit - is this commented method intentional?

anujnegi270: @lalitb I can remove this code, but the reason I kept it was because, if in-case we want to add another feature flag where we use regex to match the naming pattern of Windows Performance Counter (instead of disabling name validation). 

Any Thoughts? 
Notify - @cijothomas

anujnegi270: Removed this code block.

lalitb (approved):

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- *Feature*: Introduced a new feature flag, `experimental_metrics_disable_name_validation`, under the `opentelemetry-sdk`, which allows disabling the Instrument Name Validation. This is useful in scenarios where you need to use *special characters*, *Windows Perf Counter Wildcard Path*, or similar cases. For more details, check [#2543](https://github.com/open-telemetry/opentelemetry-rust/pull/2543).
```
cijothomas, line 5: need stronger warning about using this, as this can caused undesired behavior with systems like Prometheus.

anujnegi270: Make sense. Will add some warning to it.

anujnegi270: Added a Warning Note

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -215,6 +216,68 @@ mod tests {
         }
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[cfg(feature = "experimental_metrics_disable_name_validation")]
+    async fn valid_instrument_config_noops_with_feature_experimental_metrics_disable_name_validation(
```
cijothomas, line 221: the name is confusing. When the ff is enabled, there are no "noops" anymore.

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -20,18 +20,29 @@ use crate::metrics::{
 use super::noop::NoopSyncInstrument;
 
 // maximum length of instrument name
+#[cfg(not(feature = "experimental_metrics_disable_name_validation"))]
```
cijothomas, line 23: nit: could we restrict the change to a single place? I believe we get warnings about unused constants, but we can suppress it for this page temporarily.

opentelemetry-sdk/src/metrics/meter.rs
```diff
@@ -694,6 +719,52 @@ mod tests {
...
+                }
+            };
+
+            assert(validate_instrument_name(name).map(|_| ()));
+        }
+    }
+
+    #[test]
+    #[cfg(feature = "experimental_metrics_disable_name_validation")]
+    fn instrument_name_validation() {
```
cijothomas, line 750: unsure of the test's intent? There is no validation, so no error is expected...

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -132,15 +132,16 @@ mod tests {
...
     // Note for all tests from this point onwards in this mod:
     // "multi_thread" tokio flavor must be used else flush won't
     // be able to make progress!
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[cfg(not(feature = "experimental_metrics_disable_name_validation"))]
     async fn invalid_instrument_config_noops() {
         // Run this test with stdout enabled to see output.
-        // cargo test invalid_instrument_config_noops --features=testing -- --nocapture
+        // cargo test invalid_instrument_config_noops --features=testing,spec_unstable_metrics_views -- --nocapture
```
cijothomas, line 144: Thanks for fixing this!

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -215,6 +216,67 @@ mod tests {
...
+                .meter()
+                .i64_observable_up_down_counter(name)
+                .with_callback(move |observer| {
+                    observer.observe(1, &[]);
+                })
+                .build();
+
+            test_context.flush_metrics();
+
+            // As instrument name is invalid, no metrics should be exported
```
cijothomas, line 275: comment is misleading as it says no metrics to be exported.

opentelemetry-sdk/src/metrics/meter.rs
anujnegi270: I added this test to demonstrate the feature flag in action, ensuring users understand that any instrument name is valid. This test exists to show what the functionality of feature flag.

opentelemetry-sdk/src/metrics/meter.rs
anujnegi270: Using `#![allow(dead_code)]` & `#[allow(unused_imports)]` to suppress the warnings now.

opentelemetry-sdk/src/metrics/meter.rs
cijothomas: test itself is useful..but the test name is not very clear on its intend... 

might be better to rename to this:
instrument_name_validation_disabled()
{
}

Not a blocker, just some aesthetic!

anujnegi270: Makes sense, @cijothomas! Test renamed. :)

cijothomas:
Merging to unblock users. This flag is expected to be removed in near future. I am working with OTel spec to relax the validation rules.

PR #2546, opened 2025-01-24
Remove unused Error enum

cijothomas:
While reviewing https://github.com/open-telemetry/opentelemetry-rust/pull/2381 I noted that there is a pub enum Error declared, that is not used anywhere in the codebase. I can't think of any reason why it exists, so removing it!

@scottgerring @lalitb Please see if you agree with this.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -268,7 +268,7 @@ hardcoded limit of 2000 and no ability to change it. This feature will be
 re-introduced in a future date, along with the ability to change the cardinality
 limit.
 
-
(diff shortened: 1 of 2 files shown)


lalitb (approved):
I agree to this, but better to have one more eye on it, in case we are missing any relevant use-case for it.

cijothomas:
> I agree to this, but better to have one more eye on it, in case we are missing any relevant use-case for it.

A lot of this and other error types were originally used by global error handler, which is replaced with internal logging. We should be able to do lot more cleanups similar to this.

cijothomas:
I'll merge this to unblock progress and look for more. @scottgerring Good chance this'll affect your PR - I'll leave more comments there soon, and we can discuss together.

PR #2549, opened 2025-01-24
Restrict LogRecord creation access outside opentelemetry_sdk crate. 

lalitb:
## Changes

The only way to create `LogRecord` outside the `opentelemetry_sdk` should be using `Logger::create_log_record()` method.
So,

- Removed the `Default` implementation for LogRecord.
- Added a crate-level constructor `LogRecord::new`.
- Ensured LogRecord creation is restricted to `Logger::create_log_record()`.
- Updated tests to use `LogRecord::new()` or `Logger::create_log_record()` for record creation where applicable.

Note: It is not possible to make Default implementation private, so it was removed, and crate private constructor was added instead.


opentelemetry-sdk/CHANGELOG.md
```diff
@@ -282,6 +282,10 @@ limit.
   now:
   `opentelemetry_sdk::logs::{ExportResult, LogBatch, LogExporter};`
 
+- *Breaking* `opentelemetry_sdk::LogRecord::default()` method is removed.
```
cijothomas, line 285: - *Breaking* `opentelemetry_sdk::LogRecord::default()` method is removed.
   Use `Logger::create_log_record()` method instead.

cijothomas (approved):

Issue #2571, opened 2025-01-28
Error handling ADR


scottgerring:
# Error handling patterns in public API interfaces
This issue follows the [mADR template](https://github.com/adr/madr/blob/4.0.0/template/adr-template-minimal.md?plain=1) following [the ADR structure](https://adr.github.io/) and is written in the style of an architectural decision record. It attempts to provide a top-level view on how we should handle errors in our public traits, something that's come up in a number of discrete issues and a PR now: 

* #2564 
* #2561 
* #2381

My hope is that we can collectively make a design decision and argue the details of that here, and then the actual PRs and implementations become straightforward. 

Additionally, If we like this format (ADRs!), we could begin capturing these in the repository itself for future reference, along with other architectural artefacts. 

## Context and Problem Statement
There is uncertainty around how to model errors in  in the `opentelemetry-rust` public API interfaces - that is, APIs facing the consumers. At the time of writing this is an important issue to resolve as moving beginning to move the signals towards RC and eventually a stable release is an urgent priority. 

The situation is as follows; a concrete example is given, but the issue holds across various public traits, in particular the exporters:

* A given public interface in `opentelemetry-sdk`,such as [trait LogExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/3ec4c186ad22944b208ae7c3d38435e735a85c8e/opentelemetry-sdk/src/logs/export.rs#L115) 
* ... exposes multiple discrete actions with logically disjoint error types (e.g. [export](https://github.com/open-telemetry/opentelemetry-rust/blob/3ec4c186ad22944b208ae7c3d38435e735a85c8e/opentelemetry-sdk/src/logs/export.rs#L133-L136) and [shutdown](https://github.com/open-telemetry/opentelemetry-rust/blob/3ec4c186ad22944b208ae7c3d38435e735a85c8e/opentelemetry-sdk/src/logs/export.rs#L139)  - that is, the class of errors returned for each of these actions are foreseeably very different, as is the callers reaction to them
* ... is implemented by multiple concrete types such as `InMemoryLogExporter`, `OtlpLogExporter`, `StdOutLogExporter` that have different error requirements - for instance, an `OtlpLogExporter` will experience network failures, an `InMemoryLogExporter` will not 
* Potentially has operations on the API that, either in the direct implementation, or in a derived utility that utilises the direct implementation, call _multiple_ API actions and therefore need to return an aggregated log type

[...]

Issue #2564, opened 2025-01-25
[Metrics] Revisit every public Error enums and Results


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/error.rs#L13 defines pub enums that consist of errors from a variety of scenarios. The same is used as return type for export(), flush(), shutdown() methods, and also for internal validation methods for Views.

Must redesign this to ensure specific Error types are used for each methods.

Already discussed similar idea in https://github.com/open-telemetry/opentelemetry-rust/pull/2381, but this issue is specifically for fixing Metrics.

cijothomas:
The MeterProvider/Reader/Exporter is modified to use the newly added https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/error.rs

Views uses different one, that need to addressed before stabilizing it.
Additionally OTLP Exporter builder() also need modifications to the Result it returns. 
Given the core work is completed, removing from 0.28 milestone..

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2905
https://github.com/open-telemetry/opentelemetry-rust/pull/2928
https://github.com/open-telemetry/opentelemetry-rust/pull/2906

These PRs did the most cleanup. Only remaining item is the struct passed to exporters, which is being refactored here: https://github.com/open-telemetry/opentelemetry-rust/pull/2921

Closing as there is no pending tasks for this.

PR #2573, opened 2025-01-29
Use dedicated ShutdownResult for Metric SDK shutdown

cijothomas:
Trying to check how the changes discussed in 
#2571 and #2564 would look like. 
This touches just Metrics's Shutdown path.

### DIFF
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -141,7 +141,7 @@ async fn main() {
     // however shutting down the MeterProvider here instantly flushes
     // the metrics, instead of waiting for the 30 sec interval.
     let shutdown_result = meter_provider.shutdown();
-    
+
     // Demonstrate handling the shutdown result.
     match shutdown_result {
         Ok(_) => println!("MeterProvider shutdown successfully"),
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -23,7 +23,7 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
 }
 
 #[tokio::main]
-async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+async fn main() {
     // Initialize the MeterProvider with the stdout Exporter.
     let meter_provider = init_meter_provider();
 
@@ -140,6 +140,32 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // Metrics are exported by default every 30 seconds when using stdout exporter,
     // however shutting down the MeterProvider here instantly flushes
     // the metrics, instead of waiting for the 30 sec interval.
-    meter_provider.shutdown()?;
-    Ok(())
+    let shutdown_result = meter_provider.shutdown();
+    
+    // Demonstrate handling the shutdown result.
+    match shutdown_result {
+        Ok(_) => println!("MeterProvider shutdown successfully"),
+        Err(e) => {
+            match e {
+                opentelemetry_sdk::error::ShutdownError::Failed(e) => {
+                    // This indicates some failure during shutdown.
+                    // Not much to do here, but log the error.
+                    // So users at least know something went wrong,
+                    // and possibly explain why some metrics were not exported.
+                    println!("MeterProvider shutdown failed: {}", e)
+                }
+                opentelemetry_sdk::error::ShutdownError::AlreadyShutdown => {
+                    // This indicates some user code tried to shutdown elsewhere.
+                    // user need to review their code to ensure shutdown is called only once.
+                    println!("MeterProvider already shutdown")
+                }
+                opentelemetry_sdk::error::ShutdownError::Timeout(e) => {
+                    // This indicates the shutdown timed out, and a good
+                    // hint to user to increase the timeout or even retry.
+                    // (Shutdown method does not allow custom timeout today, but that is temporary)
+                    println!("MeterProvider shutdown timed out after {:?}", e)
+                }
+            }
+        }
+    }
 }
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -147,7 +147,7 @@ async fn main() {
         Ok(_) => println!("MeterProvider shutdown successfully"),
         Err(e) => {
             match e {
-                opentelemetry_sdk::error::ShutdownError::Failed(e) => {
+                opentelemetry_sdk::error::ShutdownError::InternalFailure(e) => {
                     // This indicates some failure during shutdown.
                     // Not much to do here, but log the error.
                     // So users at least know something went wrong,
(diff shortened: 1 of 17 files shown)


examples/metrics-basic/src/main.rs
```diff
@@ -140,6 +140,32 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // Metrics are exported by default every 30 seconds when using stdout exporter,
     // however shutting down the MeterProvider here instantly flushes
     // the metrics, instead of waiting for the 30 sec interval.
-    meter_provider.shutdown()?;
-    Ok(())
+    let shutdown_result = meter_provider.shutdown();
```
cijothomas, line 143: @scottgerring @lalitb @TommyCpp I put this just to think out loudly, showcasing what a user can possibly do with shutdown result from MeterProvider. (reader/exporter is not directly used by end users)

opentelemetry-sdk/src/error.rs
```diff
@@ -1,7 +1,31 @@
...
+    AlreadyShutdown,
+
+    /// Shutdown timed out before completing.
+    #[error("Shutdown timed out after {0:?}")]
+    Timeout(Duration),
+
+    /// Shutdown failed with an error.
+    /// This error is returned when the shutdown process failed with an error.
+    #[error("Shutdown failed: {0}")]
+    Failed(String),
```
cijothomas, line 27: "Failed" is not a good name as it is already obvious that something failed given this is an Error!

Maybe "Other".

One more point to be discussed - is there a need to mark these enums non-exhaustive? I prefer to avoid marking as non-exhaustive, so we won't be able to add new variants after 1.0. Anything new, should be folded into "Other".

lalitb: `Other` looks fine, or else `Internal` indicating this is internal/unexpected error returned by Exporter.

lalitb: >  I prefer to avoid marking as non-exhaustive, so we won't be able to add new variants after 1.0. Anything new, should be folded into "Other".

Yes, I also prefer to have it exhaustive(as default) at-least before going 1.0

scottgerring: I reckon we should box a nested error into Failed. To paraphrase the ADR issue:

> Q: Trying to see if there are any actual advantages of letting processor be more adventurous and try to down_cast and get hold of actual Error?

 > A: I think it's foreseeable that an impl will have some ✨ special failure mode ✨ that could be retried by a savvy caller that depends on that particular impl directly. The trait doesn't have to foresee that exact interaction, it just has to provide an ergonomic escape hatch to let the trait impl and its dependent do their thing

Moreover, I don't think we _lose_ anything by boxing it up; you can still coerce it to a string and dump it to the console if that's all you want.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -131,15 +134,13 @@ impl SdkMeterProviderInner {
...
+    fn shutdown(&self) -> ShutdownResult {
         if self
             .shutdown_invoked
             .swap(true, std::sync::atomic::Ordering::SeqCst)
         {
             // If the previous value was true, shutdown was already invoked.
-            Err(MetricError::Other(
-                "MeterProvider shutdown already invoked.".into(),
-            ))
+            Err(crate::error::ShutdownError::AlreadyShutdown)
```
scottgerring, line 143: I'm still not convinced that this should be an error, rather than just asserting that `shutdown` should be idempotent. @cijothomas I feel like you might've had a good argument for this I've forgotten?

opentelemetry-sdk/src/error.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2571#issuecomment-2625145109
As discussed, will modify to use Box<dyn Error> here.

opentelemetry-sdk/src/metrics/meter_provider.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2571#issuecomment-2625145109 As discussed, we agree to keep this as-is. This is done so that user can realize that someone else in their codebase is calling shutdown before them, and that is a mistake/issue they need to rectify. Given this is user actionable we should keep this.

(Also, internal logging may not help always when we are at the shutdown phase, as ordering of shutdowns may cause internal logs to stop working when we get to shutdown!!)

opentelemetry-sdk/src/error.rs
cijothomas: @scottgerring @lalitb Had attempted to use Box<dyn Error>, but I am rethinking it.
1. It now forces every component to forcefully create something that implements Error, even when there is no Error but only a string exists.
2. Even when there is a dyn Error available, I am not sure if we should force components to return it. For example, the PeriodicReader or BatchProcessor can have many mpsc(channel) errors. It should not be returning those error types, as that is leaking implementation detail outside. It may change implementation to *not* use channels anymore etc. so there is no guarantee to the invoker about the dyn Error it returns.


I tried making Other to accept both String and Error like below. (crates like anyhow make these super easy to avoid boilerplates, but not so sure about depending on it for a library)
```
    #[error(transparent)]
    Other(#[from] OtherError),
#[derive(Debug)]
pub enum OtherError {
    Error(Box<dyn std::error::Error + Send + Sync>),
    Message(String),
}
```

But I am seeing that all the use cases within OTel are only going to use the String variant. (to avoid leaking my error info to my caller. Most OTel components are not like network clients where it is desired to return specific errors)

I am inclined to go back to String, but will wait to hear more opinions on this.

lalitb: My thoughts are - If Box-ing is supposed to be used to exposing third-party crate errors directly in the public API, it should be avoided as this can introduce breaking changes if dependencies change in the future.  Better to convert these internal errors into a structured variant (or a `String` as you mentioned above) rather than exposing them directly.

But @scottgerring had some valid thoughts in today's discussion, so good to hear from him :)

scottgerring: I'm struggling for lack of a concrete "additional error an exporter could meaningfully propagate through". That's not an evidence of absence, though. 

I see your point, though and don't want to push to over-engineer it. Here's my last question to the point:

If you imagine we add a stringy `OtherError` only and 6 months pass, can anyone imagine a situation where this leads to someone trying to string-match on the error and do something - perhaps experience from otel-dotnet or otel-cpp ?

cijothomas: I just pushed a commit with doc comments that communicates that the string inside the Other is not to be used for anything but logging.
We cannot prevent anyone from doing string-match on it, but not something we need to worry.

(OTel .NET returns Exceptions from initialization code liberally, but makes no promise about the Exception Message. Nothing prevents a user from string-matching on the message, but not recommended. In .NET world, users should
 match on the concrete Exception type. In Rust, match on the enum variant, and so on...)


(I wasn't able to come up with scenario where dyn Error would help in a telemetry library like OTel... A network client likely has stronger reasons to return underlying Error....)

scottgerring (approved):
This LGTM and follows from our extensive discussions ! 
if you manage to merge it before the weekend, I could quickly do trace/logger Monday morning.

cijothomas:
Thanks @scottgerring . @lalitb could you do another review?

lalitb (approved):
Nicely done. Thanks.

PR #2581, opened 2025-01-31
Better handling of shutdown in BatchLogProcessor

cijothomas:
Another attempt at https://github.com/open-telemetry/opentelemetry-rust/pull/2462
The key change/idea here is
1. Do not perform any unnecessary check in hot path that unnecessarily punishes normal users
2. Rely on alternative mechanisms to determine state.

Specifically, in this PR, BatchLogProcessor no longer keeps track of "is_shutdown" in the AtomicBool and check it in every *hot* path. (1 above).
Instead, it relies on the well known error from the Channel ("failed to send message to channel as no receiver exists") to determine that Shutdown is already performed (2 above).

No change in behavior, just avoid an unnecessary check in hot path. (the perf impact is [minimal (couple of ns)](https://github.com/open-telemetry/opentelemetry-rust/pull/2462#discussion_r1894582949) but every ns counts)

If this is generally agreed, we need to replicate this idea everywhere.

@scottgerring Unfortunately this touches Log::Shutdown area (these changes were sitting in me for quite a while!!). I hope the changes here will allow you to return better, targeted Result as we agreed. Currently it sticks with LogError for everything. If it turns out to be merge nightmare, I am happy to abandon this one, and re-attempt it after you fix Result handling. (This has no public API change, so can be done later too)

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -270,7 +270,6 @@ pub struct BatchLogProcessor {
     handle: Mutex<Option<thread::JoinHandle<()>>>,
     forceflush_timeout: Duration,
     shutdown_timeout: Duration,
-    is_shutdown: AtomicBool,
     export_log_message_sent: Arc<AtomicBool>,
     current_batch_size: Arc<AtomicUsize>,
     max_export_batch_size: usize,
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -292,87 +291,112 @@ impl Debug for BatchLogProcessor {
...
+                        LogError::ExportTimedOut(self.forceflush_timeout)
+                    } else {
+                        LogError::Other(err.into())
+                    }
+                })?,
+            Err(mpsc::TrySendError::Full(_)) => {
+                // If the control message could not be sent, emit a warning.
+                otel_debug!(
+                    name: "BatchLogProcessor.ForceFlush.ControlChannelFull",
+                    message = "Control message to flush the worker thread could not be sent as the control channel is full. This can occur if user repeatedily calls force_flush without finishing the previous call."
```
lalitb, line 379: I think can also happen if user repeatedly calls the shutdown message, and the first shutdown message is not yet processed, so channel is still intact :)

lalitb (approved):
LGTM. Nicely done.

opentelemetry-sdk/src/logs/log_processor.rs
cijothomas: that is a good point. Let me modify error message to include that.

cijothomas: will do in separate PR to keep this merged asap.

PR #2582, opened 2025-01-31
Add more metric integration tests

cijothomas:
Add few more tests to metrics. 
Tokio::Current is failing, will investigate that separately. -- This is fixed.https://github.com/open-telemetry/opentelemetry-rust/issues/2583#issuecomment-2629087126


cijothomas:
@lalitb Did some refactoring. In a follow up, I plan to split the metrics test into two - one testing the full payload, and another one just checking presence of the guid. The latter one is where I plan to add all testing permutations (tokio vs other vs etc...)

lalitb:
@cijothomas Seems tokio current test is still failing. IF so, Should we keep it ignored, and add as separate PR., to unblock this?

cijothomas:
> @cijothomas Seems tokio current test is still failing. If so, should we keep it ignored, and add as separate PR., to unblock this?

It is expected to pass. There seems to be occasional instability in the test, not related to tokio::current, but the general tension of running things in parallel. We'll need to make it more robust in future.

lalitb (commented):
Thanks.

lalitb (approved):

Issue #2576, opened 2025-01-30
Enforce timeouts


cijothomas:
BatchExportProcessor, PeriodicReader has been rewritten to use own thread. In the new implementation, they do not enforce timeouts and instead relies on Exporters to enforce timeouts.

Tonic currently enforce timeout - default 10, and respects both OTEL_EXPORTER_OTLP_TIMEOUT and the signal specific override of this.
Reqwest-Blocking does not enforce timeout, so the default timeout from Reqwest-Blocking client, which is 30 seconds, is the only one.
Need to fix this to also enforce timeout like the tonic crate.

We also need to document that OTEL_BLRP_EXPORT_TIMEOUT,OTEL_BSP_EXPORT_TIMEOUT, OTEL_METRIC_EXPORT_TIMEOUT are not honored by default in the SDK.

cijothomas:
#[cfg(not(target_arch = "wasm32"))]
    #[async_trait]
    impl HttpClient for reqwest::blocking::Client {
        async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
            otel_debug!(name: "ReqwestBlockingClient.Send");
            let mut request: reqwest::blocking::Request = request.try_into()?;
            let request_timeout = request.timeout_mut();
            if request_timeout.is_none() {
                request_timeout.replace(std::time::Duration::from_secs(30));
            }
            let mut response = self.execute(request)?.error_for_status()?;
            let headers = std::mem::take(response.headers_mut());
            let mut http_response = Response::builder()
                .status(response.status())
                .body(response.bytes()?)?;
            *http_response.headers_mut() = headers;

            Ok(http_response)
        }
    }

The above should work for Reqwest::Blocking.

cijothomas:
Its better to do it in the opentelemetry-otlp exporter rather than in the http crate, as it needs to get the time out vases from otlp specific env variables.

cijothomas:
Additional fix to be done
BatchProcessor currently has a timeout it is not enforced. It is better to remove it from API.

///             .with_max_export_timeout(Duration::from_secs(30))

darkraider01:
[...]

PR #2584, opened 2025-02-02
Enable reqwest and reqwest-blocking client creation with custom timeout

lalitb:
Fix: towards #2576 

## Changes

This PR defers the creation of http_client until the configured timeout is accessible. Previously, http_client was initialized within `HttpConfig::default()`, where the timeout value was unavailable.


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -13,6 +13,15 @@
...
   and `MetricExporter::new(..)` method. Use
   `MetricExporter::builder()...build()` to obtain `MetricExporter`.
+- The HTTP clients (reqwest, reqwest-blocking, hyper) now support the
+timeout internal configured in below order
+  - Signal specific env variable `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT`, 
+  `OTEL_EXPORTER_OTLP_LOGS_TIMEOUT` or `OTEL_EXPORTER_OTLP_TIMEOUT`.
+  - `OTEL_EXPORTER_OTLP_TIMEOUT` env variable.
+  - `with_http().with_timeout()` API method of
+`LogExporterBuilder` and `SpanExporterBuilder` and `MetricsExporterBuilder`.
+  - The default interval of 30sec is used if none is configured.
```
cijothomas, line 23: https://github.com/open-telemetry/opentelemetry-specification/blob/714b1ae1382ba57850f29ee94b854b2aa7d1be83/specification/protocol/exporter.md?plain=1#L67 should it be 10 instead?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -171,11 +125,56 @@ impl HttpExporterBuilder {
             },
             None => self.exporter_config.timeout,
         };
-        let http_client = self
-            .http_config
-            .client
-            .take()
-            .ok_or(crate::Error::NoHttpClient)?;
+
+        #[allow(unused_mut)] // TODO - clippy thinks mut is not needed, but it is
```
cijothomas, line 129: there is another issue about parsing env variables for timeout, I opened https://github.com/open-telemetry/opentelemetry-rust/issues/2591 to track it separate from this PR

cijothomas (approved):
LGTM. For metrics we'd need a follow up as we need to use a per-request timeout override, to account for the the time taken for observable callbacks. 

Also need to fix the changelog to state its 10 secs default, not 30.

opentelemetry-otlp/CHANGELOG.md
lalitb: Good catch. It's indeed 10sec. Have update the changelog accordingly.

https://github.com/open-telemetry/opentelemetry-rust/blob/dde68a073c5c960708b0bfc6c7cc2a8f5f069b17/opentelemetry-otlp/src/exporter/mod.rs#L55

PR #2586, opened 2025-02-02
Fix PeriodicReader panic due to timeout

cijothomas:
Gracefully handle error when PeriodicReader's thread is unable to send response back to user invoking shutdown/flush. Occurs when the caller times out.


lalitb (approved):
LGTM. I think similar changes are required for Spans and Logs too.

PR #2587, opened 2025-02-03
Remove export timeout configuration for BatchLogProcessor.

lalitb:
Fixes: towards #2576 

## Changes


lalitb:
> Are you planning to do the same for Span and PeriodicReader too?
PeriodicReader may need some discussion... As Observables are called during collect, so pushing timeout entirely to Exporter is not the best option. Unfortunately that is probably our only option too, so maybe we can cover it via good doc.

Yes, will raise the PR for both. We can discuss the Observables issue over the PR.

PR #2594, opened 2025-02-03
Doc additions to warn about tokio current

### LINKED ISSUE #2583: Cannot shutdown or force_flush in a Tokio::CurrentThread application when using Tonic
The rewritten BatchProcessor, PeriodicReader spawns own thread for background job. The shutdown/force_flush is a blocking call, which sends a message to the background thread and wait for a response. When using Tokio::CurrentThread, the export (tonic.export) methods will deadlock, as the "current_thread" it got while initialization, is now blocked on shutdown(). 

The integrations were passing due to export being triggered due to time interval (which is 1 seconds, and we sleep for much more than that to let Collector store results).

Opening an issue to discuss solutions. An async_shutdown() might help here, so the tokio::current_thread can do a non-blocking shutdown? Or the user of tokio:current must be advised to invoke shutdown in a separate thread. Or can continue to the existing tokio based async rt. (it is not known if this works there.)

cijothomas:
Fixes #2583
and few doc improvements.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -228,18 +228,32 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
 /// individually. It uses a **dedicated background thread** to manage and export logs
 /// asynchronously, ensuring that the application's main execution flow is not blocked.
 ///
-/// - This processor supports the following configurations:
-///     - **Queue size**: Maximum number of log records that can be buffered.
-///     - **Batch size**: Maximum number of log records to include in a single export.
-///     - **Scheduled delay**: Frequency at which the batch is exported.
+/// This processor supports the following configurations:
+/// - **Queue size**: Maximum number of log records that can be buffered.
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -228,19 +228,33 @@ type LogsData = Box<(LogRecord, InstrumentationScope)>;
...
+/// export is triggered when `max_export_batch_size` is reached or every
+/// `scheduled_delay` milliseconds. Users can explicitly trigger an export using
+/// the `force_flush` method. Shutdown also triggers an export of all buffered
+/// logs and is recommended to be called before the application exits to ensure
+/// all buffered logs are exported.
+///
+/// **Warning**: When using tokio's current-thread runtime, `shutdown()`, which
+/// is a blocking call ,should not be called from your main thread. This can
+/// cause deadlock. Instead, call `shutdown()` from a separate thread or use
+/// tokio's `spawn_blocking`.
```
lalitb, line 253: This is valid for force_flush() too.

lalitb (approved):
Nicely documented.

Issue #2557, opened 2025-01-24
Use opentelemetry::time::now() everywhere


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/lib.rs#L289 was introduced to support wasm (I think). However it is not used in all places requesting current time.
This issue is opened to track leveraging this throughout the repo.

cijothomas:
@TommyCpp Could you confirm if replacing all now() usage with opentelemetry::time::now() is the right thing to do to address this?

TommyCpp:
> [@TommyCpp](https://github.com/TommyCpp) Could you confirm if replacing all now() usage with opentelemetry::time::now() is the right thing to do to address this?

Yes. It helps support different platform. We can also use this to return constant time during unit test

PR #2595, opened 2025-02-03
Use opentelemetry:time::now instead of systemtime

mohammadVatandoost:
Part 2 for https://github.com/open-telemetry/opentelemetry-rust/issues/2557


cijothomas (approved):
Thanks!
Would be great if you can continue in follow up PRs to use this everywhere!

PR #2597, opened 2025-02-03
Add easier way to add periodicreader to meterprovider

cijothomas:
Trying to avoid the term "reader" as much, which may not be easily relatable for most users.

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -19,7 +19,8 @@ use crate::{
 };
 
 use super::{
-    meter::SdkMeter, noop::NoopMeter, pipeline::Pipelines, reader::MetricReader, view::View,
+    exporter::PushMetricExporter, meter::SdkMeter, noop::NoopMeter, pipeline::Pipelines,
+    reader::MetricReader, view::View, PeriodicReader,
 };
 
 /// Handles the creation and coordination of [Meter]s.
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -244,14 +245,37 @@ impl MeterProviderBuilder {
...
+    ///
+    /// To customize the export interval, set the
+    /// **"OTEL_METRIC_EXPORT_INTERVAL"** environment variable (in
+    /// milliseconds).
+    ///
+    /// Most users should use this method to attach an exporter. Advanced users
+    /// who need finer control over the export process can use
+    /// [`PeriodicReaderBuilder`] to configure a custom reader and attach it
+    /// using [`with_reader`].
+    pub fn with_periodic_exporter<T>(mut self, exporter: T) -> Self
```
cijothomas, line 270: it is intentional to pick with_periodic_exporter and not with_periodic_reader. This is what was already followed in Logs/Traces (with_batch_exporter, not with_batch_processor).
Given this is a convenience API, better to avoid "reader" which is irrelevant to most users.

lalitb (approved):

cijothomas:
@lalitb and I also discussed some redesign where MetricReader is not something a user ever need to create and manage, as they have no direct use with that. Will open separate issue for detailed discussion, but this PR does the minimal job of not forcing users to even be aware of a thing called "reader".

PR #2600, opened 2025-02-04
Metric Export to return dedicated ExportResult

cijothomas:
Continuing this https://github.com/open-telemetry/opentelemetry-rust/pull/2573/files for ExportResult. Only touching Metrics, but this should be reused for all signals.

### DIFF
--- a/opentelemetry-sdk/src/error.rs
+++ b/opentelemetry-sdk/src/error.rs
@@ -41,5 +41,30 @@ pub enum ShutdownError {
     InternalFailure(String),
 }
 
+#[derive(Error, Debug)]
+/// Errors that can occur during export.
+/// TODO: This should be ExportError, but that can be done after rest of repo changes are done.
+pub enum ExportErrorMetric {
+    /// Export timed out before completing.
+    ///
+    /// This does not necessarily indicate a failure—export may still be
+    /// complete. If this occurs frequently, consider increasing the timeout
+    /// duration to allow more time for completion.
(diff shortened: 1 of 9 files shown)


cijothomas:
@scottgerring @lalitb My attempt at Export error for Metrics. To ease merge conflicts, maybe we should get just the Enum/Result alone in a small PR and get it merged, so we can work without conflicts.
Let me know if you want to split the Enum to an short separate PR to be reviewed first.

opentelemetry-sdk/src/error.rs
```diff
@@ -41,5 +41,30 @@ pub enum ShutdownError {
     InternalFailure(String),
 }
 
+#[derive(Error, Debug)]
+/// Errors that can occur during export.
+/// TODO: This should be ExportError, but that can be done after rest of repo changes are done.
+pub enum ExportErrorMetric {
```
cijothomas, line 47: After checking more, this enum would also need AlreadyShutDown variant. Which means we should be able to re-use the same enum for Export,Shutdown,ForceFlush.

@scottgerring @lalitb Let me know if you agree. I'll make a small PR, just adding the common enum first, and then divide&conquer using it in all places.

lalitb: > I'll make a small PR, just adding the common enum first, and then divide&conquer using it in all places.

Looks good to me.

cijothomas:
Closing in favor of https://github.com/open-telemetry/opentelemetry-rust/pull/2604

PR #2603, opened 2025-02-04
Make a common Error enum for export shutdown and flush operations

cijothomas:
As discussed https://github.com/open-telemetry/opentelemetry-rust/pull/2600/files#r1941764809 this renames the ShutdownError to OTelSdkError
ShutdownResult to OTelSdkResult

In separate follow up PR, we can modify usages to move to the new one, and finally do the removal of existing Errors types

### DIFF
--- a/examples/metrics-basic/src/main.rs
+++ b/examples/metrics-basic/src/main.rs
@@ -1,5 +1,5 @@
 use opentelemetry::{global, KeyValue};
-use opentelemetry_sdk::error::ShutdownError;
+use opentelemetry_sdk::error::OTelSdkError;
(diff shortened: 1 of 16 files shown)


lalitb (approved):
Looks good to me.
int - OTelSdkResult -> SdkResult as opentelemetry is already part of namespace. But no strong opinion.

cijothomas:
> Looks good to me. nit - OTelSdk{Result}/Error -> Sdk{Result}/Error as opentelemetry is already part of namespace. But no strong opinion.

Will merge as-is. After every other change, lets us make a call on that and rename. (I like your suggestion to avoid repeating OTel given it is already in namespace.)

PR #2607, opened 2025-02-04
Use opentelemetry:time::now instead of systemtime (final part)

### LINKED ISSUE #2557: Use opentelemetry::time::now() everywhere

mohammadVatandoost:
Fixes #2557 
Final 2 for https://github.com/open-telemetry/opentelemetry-rust/issues/2557


cijothomas (approved):
Thanks!

PR #2609, opened 2025-02-05
Move traceid spanid traceflags to common

cijothomas:
Trying to move TraceId,SpanId,TraceFlags outside of `trace` to common module.


cijothomas:
@TommyCpp @lalitb Please review. I reexported from trace module too, so this should not be a breaking change...

lalitb (approved):
Nice. thanks.

PR #2618, opened 2025-02-05
ObservableGauge collect data points since previous collection

### LINKED ISSUE #2213: [Bug]: ObservableGauge no longer dropping data points from old Attributes values
### Related Problems?

_No response_

A change in behaviour is present between 0.24 & 0.26 for ObservableGauge & possibly other Async Instruments.

In 0.26 once any reading is made for a given combination of attributes that reading is always sent in subsequent collections. There's no way to remove any attributes that are no longer present in the data. This doesn't match the [SDK Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/0a78571045ca1dca48621c9648ec3c832c3c541c/specification/metrics/sdk.md#metricreader):

> For asynchronous instruments with Delta or Cumulative aggregation temporality, [MetricReader.Collect](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#collect) MUST only receive data points with measurements recorded since the previous collection.

Adapted one of the existing tests to show an error - <https://github.com/markdingram/opentelemetry-rust/commit/68d375b320c1e2cbcec5030a2f0269cd9553a7fb>

Comment from @cijothomas on CNCF Slack: 

> OTel Rust currently is likely violating the above part of spec, as it never forgets data points for Cumulative temporality, but the spec says it should forget data points for CUMULATIVE too, for Observable instruments.

### Considered Alternatives

_No response_

### Additional Context

_No response_

fraillt:
Fixes #2213

## Changes

Some explanation for implementation:
1. observable/asynchronous instruments ALWAYS report data points since previous collection
2. Temporality affects how measurements are aggregated AND when data points are reported
3. LastValue aggregation always use last measured value, no matter what temporality
* Hence: always use Delta temporality for observable gauge (which uses last value aggregation), in order to report data points from previous collection.

I think there are more issues with other observable instruments, but these will require more sophisticated solution to solve...
We probably need to utilize different value map implementations to make things efficient for all metrics, so #2530 will become essential to further improvements/fixes.


cijothomas (approved):
LGTM. Thank you!

Maybe a changelog for letting users know the issue is fixed for ObservableGauge and still open for other Observables.

PR #2622, opened 2025-02-06
0.28 migration guide and OTLP Example fixes

### LINKED ISSUE #2611: Migration guide for 0.28
Given the amount of breaking changes for 0.28, we need to write a migration guide, covering common scenario of using
otel/otel-sdk/otlp exporter.

cijothomas:
Fixes #2611 
Also fixed OTLP Examples a bit to panic when exporter creation fails. The exporter failures need to return proper Result, and until then its best to panic in the example application.

### DIFF
--- a/docs/migration_0.28.md
+++ b/docs/migration_0.28.md
@@ -18,7 +18,7 @@ Before (0.27):
 opentelemetry::global::shutdown_tracer_provider();
 ```
 
- • After (0.28):
+ After (0.28):
 
 ```rust
 let tracer_provider = opentelemetry_sdk::trace::SdkTracerProvider::builder()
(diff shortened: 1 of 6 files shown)


docs/migration_0.28.md
```diff
@@ -0,0 +1,155 @@
+# Migration guide from 0.27 to 0.28
+
+OpenTelemetry Rust 0.28 introduces a large number of breaking changes that
+impact all signals (logs/metrics/traces). This guide is intended to help with a
+smooth migration for the common use cases of using `opentelemetry`,
+`opentelemetry_sdk` `opentelemetry-otlp`, `opentelemetry-appender-tracing`
```
anujnegi270, line 6: nit - `opentelemetry-sdk`

docs/migration_0.28.md
```diff
@@ -0,0 +1,155 @@
+# Migration guide from 0.27 to 0.28
```
anujnegi270, line 1: Are you planning to add DataPoint change (to SumDataPoint & GuageDataPoint) here as well?

cijothomas: no. Changelog should cover that. This is for the most common scenario only.

docs/migration_0.28.md
cijothomas: it is actually "_"!! That one crate was named differently due to some historical reason...

lalitb (approved):
LGTM. Nicely written :)

PR #2634, opened 2025-02-07
Log - modify export/processor to not require mutable self in shutdown and set-resource

cijothomas:
And let the component that needs mutability rely on interior mutability. Most of it already uses interior mutability.

Metric already does not require mutable self on shutdown, this makes Logs consistent with that, and later we need to get Traces too.

DRAFT as I am still working through OTLP.


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -135,53 +135,31 @@ pub trait LogProcessor: Send + Sync + Debug {
...
-    is_shutdown: AtomicBool,
+    exporter: T,
 }
 
 impl<T: LogExporter> SimpleLogProcessor<T> {
     pub(crate) fn new(exporter: T) -> Self {
         SimpleLogProcessor {
-            exporter: Mutex::new(exporter),
-            is_shutdown: AtomicBool::new(false),
+            exporter: exporter
```
lalitb, line 144: I believe the purpose of Mutex lock was to ensure that the multiple exports are not invoked concurrently.

cijothomas: good point! We need to enforce that spec requirement.
Will see how to do it without Mutex.

PR #2641, opened 2025-02-11
feat: Refine runtime trait

martintmk:
Contributes to #2643

## Changes

This PR cleans-up and simplifies the `Runtime` trait.

Before:

```rust
pub trait Runtime: Clone + Send + Sync + 'static {
    type Interval: Stream + Send;

    type Delay: Future + Send + Unpin;

    fn interval(&self, duration: Duration) -> Self::Interval;

    fn spawn(&self, future: BoxFuture<'static, ()>);

    fn delay(&self, duration: Duration) -> Self::Delay;
}
```

After:

```rust
pub trait Runtime: Clone + Send + Sync + 'static {
    fn spawn<F>(&self, future: F) where F: Future<Output = ()> + Send + 'static;

    fn delay(&self, duration: Duration) -> impl Future<Output = ()> + Send + 'static;
}
```

The main changes is the drop of the interval, which can be moved to an internal helper method. 

**Future Improvements:** 

This trait could even made to be agnostic to single-threaded and multi-threaded runtimes and be simplified to:

```rust
pub trait Runtime {
    fn spawn<F>(&self, future: F) where F: Future<Output = ()> + Send + 'static;

    async fn delay(&self, duration: Duration);
}
```

The trait requirements would have to be specified on the consuming side:

```rust
impl<E, RT> PeriodicReaderBuilder<E, RT>
where
    E: PushMetricExporter,
    RT: Runtime + Send + Sync + 'static,
    RT::delay(..): send + 'static
{
   fn new(exporter: E, runtime: RT) -> Self {
      ...
   }
}
```

However, this requires the [return type notation support](https://blog.rust-lang.org/inside-rust/2024/09/26/rtn-call-for-testing.html) which is not yet stable.

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs
+++ b/opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs
@@ -126,7 +126,7 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
         let inner_runtime = runtime.clone();
 
         // Spawn worker process via user-defined spawn function.
-        runtime.spawn(Box::pin(async move {
+        runtime.spawn(async move {
             // Timer will take a reference to the current runtime, so its important we do this within the
             // runtime.spawn()
             let ticker = to_interval_stream(inner_runtime.clone(), config.scheduled_delay)
@@ -204,7 +204,7 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
                     }
                 }
             }
-        }));
+        });
         // Return batch processor with link to worker
         BatchLogProcessor {
             message_sender,
--- a/opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs
+++ b/opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs
@@ -17,7 +17,7 @@ use std::{
 };
 
 use super::{BatchConfig, LogProcessor};
-use crate::runtime::{RuntimeChannel, TrySend};
+use crate::runtime::{to_interval_stream, RuntimeChannel, TrySend};
 use futures_channel::oneshot;
 use futures_util::{
     future::{self, Either},
@@ -129,10 +129,10 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
         runtime.spawn(Box::pin(async move {
             // Timer will take a reference to the current runtime, so its important we do this within the
             // runtime.spawn()
-            let ticker = inner_runtime
-                .interval(config.scheduled_delay)
+            let ticker = to_interval_stream(inner_runtime.clone(), config.scheduled_delay)
                 .skip(1) // The ticker is fired immediately, so we should skip the first one to align with the interval.
                 .map(|_| BatchMessage::Flush(None));
+
             let timeout_runtime = inner_runtime.clone();
             let mut logs = Vec::new();
             let mut messages = Box::pin(stream::select(message_receiver, ticker));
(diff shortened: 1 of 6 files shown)


berkus (approved):

cijothomas:
This would be towards # 1 in https://github.com/open-telemetry/opentelemetry-rust/issues/2643

opentelemetry-sdk/src/runtime.rs
```diff
@@ -6,42 +6,57 @@
...
-    /// not important.
-    type Interval: Stream + Send;
-
-    /// A future, which resolves after a previously specified amount of time. The output type is
-    /// not important.
-    type Delay: Future + Send + Unpin;
-
-    /// Create a [futures_util::stream::Stream], which returns a new item every
-    /// [std::time::Duration].
-    fn interval(&self, duration: Duration) -> Self::Interval;
```
lalitb, line 30: Does the removal of `interval` now restrict the runtimes to bring their own optimized constructs. Eg, we were able to use the `tokio_stream::wrappers::IntervalStream` for tokio runtime. But now, all runtimes now must use the generic unfold+delay implementation provided by `to_interval_stream`? Do we need any performance comparison between the original runtime-specific interval implementations and the new unfold+delay approach ?

martintmk: My thinking here is that for the use-cases of OTel, the interval is relatively high, in case of metrics 60 seconds. While individual runtimes can provide their own optimized stream, it's unlikely this will bring any measurable perf benefit. 

the interval can really be replaced with the following to completely avoid streams:

```rust
loop {
   runtime.delay(...).await;
}
```

I can try to provide some benchmarks, but this is async code which is kinda difficult to measure correctly.

cijothomas: > for the use-cases of OTel, the interval is relatively high, in case of metrics 60 seconds

Logs export frequency is 1 seconds.

(Not saying it is an issue, just checking if that makes any difference)

opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs
```diff
@@ -126,13 +126,13 @@ impl<R: RuntimeChannel> BatchLogProcessor<R> {
         let inner_runtime = runtime.clone();
 
         // Spawn worker process via user-defined spawn function.
-        runtime.spawn(Box::pin(async move {
+        runtime.spawn(async move {
             // Timer will take a reference to the current runtime, so its important we do this within the
             // runtime.spawn()
-            let ticker = inner_runtime
-                .interval(config.scheduled_delay)
+            let ticker = to_interval_stream(inner_runtime.clone(), config.scheduled_delay)
```
cijothomas, line 132: we might be able to rewrite the batchlogprocessor to be similar to the one that spawns own thread, which do not rely on interval_steam, but use alternatives to achieve similar goal.

cijothomas (approved):
Thanks!

cijothomas:
@martintmk https://github.com/hyperium/hyper/blob/master/src/rt/mod.rs is what Hyper has. Maybe we should also need just that?

opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs
martintmk: Is this something that needs to be addressed in this PR?

cijothomas: No.

PR #2649, opened 2025-02-11
Refactor InMemoryMetricExporter

cijothomas:
Refactor so that InMemoryMetricExporter does not manage storage, instead user must provide one that it internally uses.
Removes the need for cloning the exporter, which is very unusual, and makes InMemoryExporter look odd/inconsistent with other exporters.

Draft now.


cijothomas:
@martintmk Please feel free to take over this/start new PR with similar idea.

PR #2654, opened 2025-02-12
OTLP Exporter builder to return more specific Error type

### LINKED ISSUE #2561: Should OTLP Exporter Error enum be public and visible to the processor?
More for discussion.

The OTLP Exporter crate defines a public [Error enum](https://github.com/open-telemetry/opentelemetry-rust/blob/ade08cb1a3876da219a47760fe45abd075b7cd38/opentelemetry-otlp/src/lib.rs#L298) that is currently accessible throughout the crate and to external consumers. The processor, however, interacts with these errors through the [ExportError](https://github.com/open-telemetry/opentelemetry-rust/blob/b2d3d74962c525bd5058b5248c9155581d443bae/opentelemetry-sdk/src/logs/error.rs#L15) trait, which abstracts away the specific details of the underlying Error enum.

- Should the `Error` enum and its variants remain public, given that they are primarily relevant to the OTLP Exporter crate itself?
- Is there a use case for the processor or other consumers to directly match on or handle specific variants of the Error enum, or is the abstraction provided by ExportError sufficient?
- Does the processor need to take any action based on specific OTLP Error enum variants?
- Should the processor be designed to react differently to specific errors (e.g., transport errors, invalid URIs, etc.), or does it suffice to treat all exporter errors uniformly?

In the current design, the processor wraps errors using `LogError::ExportFailed`, which stores a boxed `ExportError`:
```rust
LogError::ExportFailed(Box<dyn ExportError>)
```
To handle specific OTLP Error variants, the consumer(i.e, processor) must perform a manual downcast:
```rust
if let LogError::ExportFailed(export_error) = log_error {
    if let Some(otlp_error) = export_error.downcast_ref::<otlp::Error>() {
        match otlp_error {
            otlp::Error::Transport(err) => println!("Transport error: {}", err),
            otlp::Error::InvalidUri(err) => println!("Invalid URI: {}", err),
            otlp::Error::Status { code, message } => {
                println!("gRPC Status - Code: {:?}, Message: {}", code, message);
            }
            _ => println!("Other OTLP error"),
        }
    }
}
```

However, in that case, processor should handle such downcast and error handling for all other exporters (if they have any such error enum).

cc @scottgerring , @cijothomas

cijothomas:
Fixes #2561 

Also removes now unused Error enum from Log SDK. Another set of cleanup required to clean Metrics, Traces completely.
Also, the newly introduced enum `ExporterBuildError` is not fully refined, this can be done in follow ups.

Note: This is a *breaking* change for OTLP Exporter, however, the impact is minimal (most users/examples simply unwrap/expect the Result instead of matching on it.), as described in the changelog, so I propose to make the change even though OTLP Exporter is marked RC.


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -320,7 +318,7 @@ impl OtlpHttpClient {
     fn build_logs_export_body(
         &self,
         logs: LogBatch<'_>,
-    ) -> opentelemetry_sdk::logs::LogResult<(Vec<u8>, &'static str)> {
+    ) -> Result<(Vec<u8>, &'static str), Box<dyn std::error::Error>> {
```
scottgerring, line 321: Is this the right type for this?

it looks like:
1/ It is private
2/ It is only used in `OtlpHttpClient`'s `LogExport` impl, in `export` 

It seems like it can **only** be a serde_json Error - do we lose anything if we return this?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -348,9 +346,7 @@ impl OtlpHttpClient {
             #[cfg(feature = "http-json")]
             Protocol::HttpJson => match serde_json::to_string_pretty(&req) {
                 Ok(json) => Ok((json.into(), "application/json")),
-                Err(e) => Err(opentelemetry_sdk::metrics::MetricError::Other(
-                    e.to_string(),
-                )),
+                Err(e) => Err(Box::new(e)),
```
scottgerring, line 349: Same as above - should we return the serde error when it's the only failure that can shake out of this?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -388,13 +384,22 @@ fn resolve_http_endpoint(
         return Ok(endpoint);
     }
 
+    let provided_endpoint_clone = provided_endpoint.clone();
```
scottgerring, line 387: Is the clone here necessary?
If it is, could it be moved into the map_err body so its only done when it's needed?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -92,6 +93,42 @@ impl Default for ExportConfig {
     }
 }
 
+#[derive(Error, Debug)]
+/// Errors that can occur while building an exporter.
+// TODO: Refine and polish this.
```
scottgerring, line 98: 😇 
```suggestion
```

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -92,6 +93,42 @@ impl Default for ExportConfig {
...
+    /// Spawning a new thread failed.
+    #[error("Spawning a new thread failed. Unable to create Reqwest-Blocking client.")]
+    ThreadSpawnFailed,
+
+    /// Feature required to use the specified compression algorithm.
+    #[cfg(any(not(feature = "gzip-tonic"), not(feature = "zstd-tonic")))]
+    #[error("feature '{0}' is required to use the compression algorithm '{1}'")]
+    FeatureRequiredForCompressionAlgorithm(&'static str, Compression),
+
+    /// Unsupported compression algorithm.
```
scottgerring, line 109:
```suggestion
    /// HTTP client not specified
```

scottgerring (commented):
comments inline

opentelemetry-otlp/src/exporter/http/mod.rs
cijothomas: fixed to return serde_json::Error only

cijothomas: Refactored and modified to use dyn Error itself, as the serde::json is only applicable in json feature flag.

cijothomas:
Closing in favor of https://github.com/open-telemetry/opentelemetry-rust/pull/2790

PR #2656, opened 2025-02-12
Chang timeout value to miliseconds

### LINKED ISSUE #2591: Timeout and delay values from ENV variable should be in milliseconds
Following seem to treat as seconds which should be treated as a bug.
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/mod.rs#L90
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/tonic/mod.rs#L203

Other places are interpreting it correctly as milliseconds.

mohammadVatandoost:
Fixes #2591


lalitb:
@mohammadVatandoost Thanks for working on this. Can you have a look into the OTLP flow, as I belive there are some other places too. E,g
https://github.com/open-telemetry/opentelemetry-rust/blob/05920759303358cb979b7b5538da2b6b5c0be8ee/opentelemetry-otlp/src/exporter/http/mod.rs#L120-L123.

Also update the changelog as this is a breaking change. 

@cijothomas @TommyCpp  Should this be a breaking change for env-vars, or should we keep current behavior as seconds, and additive change to allow specifying in ms. So `10` => 10 secs, `10ms` => 10 millisecs. As of now, compile time configuration [with_timeout](https://github.com/open-telemetry/opentelemetry-rust/blob/05920759303358cb979b7b5538da2b6b5c0be8ee/opentelemetry-otlp/src/exporter/mod.rs#L195) accepts duration as an argument, which means it can take both secs and millisecs.

cijothomas:
> @cijothomas @TommyCpp - Should this be a breaking change for environment variables, or should we maintain the current behavior (interpreting values as seconds) while adding support for specifying milliseconds? For example, 10 (also 10s) would remain 10 seconds, while 10ms would represent 10 milliseconds. Currently, the compile-time configuration [with_timeout](https://github.com/open-telemetry/opentelemetry-rust/blob/05920759303358cb979b7b5538da2b6b5c0be8ee/opentelemetry-otlp/src/exporter/mod.rs#L195) accepts a duration argument, allowing both seconds and milliseconds. Trying to avoid breaking changes unless it significantly complicates the implementation, or adds to confusion.

I think this is a *bug* and must be treated accordingly. Every bug can be thought of as a breaking change for those users who were relying on the buggy behavior 🤣 !
10s,10ms are not something covered by spec, and won't be respected by other languages, so its best to stick with ENV variables spec as-is.

I agree that the impact is bad - A user who had set 1 sec timeout now instantly only has a 1 ms timeout (that can easily be hit much more frequently) with the upgrade - this can be covered in change-log and migration guide as a key warning/bullet point.

Happy to hear other thoughts.

TommyCpp:
> 10 (also 10s) would remain 10 seconds, while 10ms would represent 10 milliseconds

probably additional complexity we don't want to maintain on the long term. I think we should just note it in breaking changes

mohammadVatandoost:
> OTEL_EXPORTER_OTLP_TIMEOUT

The line, you mentioned, is already changed. Please see the PR file changes.
I couldn't find any other places than these three files. Please let me know if i miss sth.

lalitb:
> I couldn't find any other places than these three files. Please let me know if i miss sth.

 I think it looks good.

> Changelog is updated

The changes also affects signal specific env-variables - `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` `OTEL_EXPORTER_OTLP_LOGS_TIMEOUT`. Please mention them too in the changelog.

lalitb (approved):
LGTM, nit comment on updating changelog with signal specific env-variables.

mohammadVatandoost:
> The changes also affects signal specific env-variables - `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` `OTEL_EXPORTER_OTLP_LOGS_TIMEOUT`. Please mention them too in the changelog.

Added

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -120,7 +120,7 @@ impl HttpExporterBuilder {
             .or(env::var(OTEL_EXPORTER_OTLP_TIMEOUT).ok())
         {
             Some(val) => match val.parse() {
-                Ok(seconds) => Duration::from_secs(seconds),
+                Ok(seconds) => Duration::from_millis(seconds),
```
shaun-cox, line 123: This doesn't pass the sniff test, does it?
Why leave the variable name as `seconds` if its being passed to `from_millis`?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -52,7 +52,7 @@ const OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON: &str = "http/json";
 /// Max waiting time for the backend to process each signal batch, defaults to 10 seconds.
 pub const OTEL_EXPORTER_OTLP_TIMEOUT: &str = "OTEL_EXPORTER_OTLP_TIMEOUT";
 /// Default max waiting time for the backend to process each signal batch.
-pub const OTEL_EXPORTER_OTLP_TIMEOUT_DEFAULT: u64 = 10;
+pub const OTEL_EXPORTER_OTLP_TIMEOUT_DEFAULT: u64 = 10000;
```
shaun-cox, line 55: It's not obvious what the units are... can we not declare this to be a `const Duration` instead of a unit-less `u64`?

opentelemetry-otlp/src/exporter/http/mod.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/pull/2688

PR #2660, opened 2025-02-13
Simplifiy OTLP Example Resource creation

cijothomas:
Resource needs to be handed over to each provider, so it must be created or cloned anyway.


lalitb (approved):

cijothomas:
On second thought, this is not good, as it need to detect resources again and again. Better to use OnceLock and clone.
Let me change

lalitb (approved):

Issue #2517, opened 2025-01-15
Switch to Rust native async


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2374 did this for Logs. Need to replicate this in other places.

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/export.rs#L31 Export method in Tracing signal can be modified to avoid the Box ? @lalitb was exploring this..

lalitb:
> Export method in Tracing signal can be modified to avoid the Box ? @lalitb was exploring this..

Yes, I have the changes locally. Unless some-one wants to do it earlier, I can raise a PR next week.

cijothomas:
Closed with https://github.com/open-telemetry/opentelemetry-rust/pull/2685 and https://github.com/open-telemetry/opentelemetry-rust/pull/2672

PR #2662, opened 2025-02-13
leverage native async trait in PushMetricExporter

gruebel:
Relates #2517

## Changes

- also removed unused dependencies in sdk and stdout
- also migrated `LogExporterWithFuture` in the sdk bench to native async trait

### DIFF
--- a/opentelemetry-sdk/src/metrics/exporter.rs
+++ b/opentelemetry-sdk/src/metrics/exporter.rs
@@ -1,6 +1,4 @@
 //! Interfaces for exporting metrics
-use async_trait::async_trait;
-
 use crate::error::OTelSdkResult;
 
 use crate::metrics::data::ResourceMetrics;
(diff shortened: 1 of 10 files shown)


opentelemetry-sdk/src/metrics/in_memory_exporter.rs
```diff
@@ -263,7 +262,6 @@ impl InMemoryMetricExporter {
     }
 }
 
-#[async_trait]
```
lalitb, line 266: Is it not necessary to keep the `export` function definition consistent with the trait i.e., to return 
`impl std::future::Future<Output = OTelSdkResult<()>> + Send?`

gruebel: Not to my knowledge. You only have to return a future in the trait definition not where you implement it.

opentelemetry-sdk/src/metrics/exporter.rs
```diff
@@ -10,17 +8,19 @@ use super::Temporality;
...
     /// considered unrecoverable and will be logged.
-    async fn export(&self, metrics: &mut ResourceMetrics) -> OTelSdkResult;
+    fn export(
+        &self,
+        metrics: &mut ResourceMetrics,
+    ) -> impl std::future::Future<Output = OTelSdkResult> + Send;
 
     /// Flushes any metric data held by an exporter.
-    async fn force_flush(&self) -> OTelSdkResult;
+    fn force_flush(&self) -> impl std::future::Future<Output = OTelSdkResult> + Send;
```
cijothomas, line 23: just realized that this is something we need to remove, in a separate PR. Flush can be normal method, consistent with other signals.

cijothomas (approved):
LGTM! Thank you.

lalitb (approved):
Thanks

PR #2664, opened 2025-02-14
docs: Add ADR dir and error handling ADR

### LINKED ISSUE #2571: Error handling ADR

scottgerring:
Fixes #2571

Capturing decision record for error handling in repo with new docs. 

### DIFF
--- a/docs/adr/001_error_handling.md
+++ b/docs/adr/001_error_handling.md
@@ -3,6 +3,10 @@
 ## Date
 17 Feb 2025 
 
+## Accepted Option
+
+**Option 3** 
+
 ## Context and Problem Statement
 There is uncertainty around how to model errors in the `opentelemetry-rust` public API interfaces - that is, APIs that are exposed to users of the project's published crates. This is for example the case with the exporter traits - [SpanExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/trace/export.rs#L18), [LogExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/logs/export.rs#L115), and [PushMetricExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/metrics/exporter.rs#L11) which form part of the API surface of `opentelemetry-sdk`.
 
@@ -11,6 +15,7 @@ We will focus on the exporter traits in this example, but the outcome should be 
 There are various ways to handle errors on trait methods, including swallowing them and logging, panicing, returning a shared global error, or returning a method-specific error. We strive for consistency, and we want to be sure that we've put enough thought into what this looks like that we don't have to make breaking interface changes unecessarily in the future.
 
 This was discussed extensively in #2571.
+
 
 ## Related Work
 
@@ -94,6 +99,8 @@ pub trait SpanExporter {
 }
 
 ```
+
+### When to box custom errors
 
 Note above that we do not box anything into `InternalFailure`. Our rule here is that if the caller cannot reasonably be expected to handle a particular error variant, we will use a simplified interface that returns only a descriptive string. In the concrete example we are using with the exporters, we have a [strong signal in the opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#export) that indicates concretely that the error types are not actionable by the caller.
 
--- a/docs/adr/001_error_handling.md
+++ b/docs/adr/001_error_handling.md
@@ -14,14 +14,14 @@ We will focus on the exporter traits in this example, but the outcome should be 
 
 There are various ways to handle errors on trait methods, including swallowing them and logging, panicing, returning a shared global error, or returning a method-specific error. We strive for consistency, and we want to be sure that we've put enough thought into what this looks like that we don't have to make breaking interface changes unecessarily in the future.
 
-This was discussed extensively in #2571.
+This was discussed extensively in [#2571](https://github.com/open-telemetry/opentelemetry-rust/issues/2571).
 
 
 ## Related Work
 
-* #2564 
-* #2561 
-* #2381
+* [#2564](https://github.com/open-telemetry/opentelemetry-rust/issues/2564)
+* [#2561](https://github.com/open-telemetry/opentelemetry-rust/issues/2561)
+* [#2381](https://github.com/open-telemetry/opentelemetry-rust/issues/2381)
 
 ## Considered Options
 
@@ -45,7 +45,7 @@ Our preference for error types is thus:
 1. Devolves into error type per method of a particular trait type (e.g., `SdkShutdownResult`, `SdkExportResult`) _if the error types need to diverge_
 1. May alternatively devolve into error type per signal (e.g., `SpanExporter`) if the _signals diverge_
 
-This approach generalises across both **signals** and **trait methods**. For example, returning to our exporter traits, we have a trait that looks the same for each signal, with the same three methods. Upon closer inspection (#2600), the potential error set is the same both between the methods *and* between the signals; this means we can use a single shared error type across both axes:
+This approach generalises across both **signals** and **trait methods**. For example, returning to our exporter traits, we have a trait that looks the same for each signal, with the same three methods. Upon closer inspection ([#2600](https://github.com/open-telemetry/opentelemetry-rust/issues/2600)), the potential error set is the same both between the methods *and* between the signals; this means we can use a single shared error type across both axes:
 
 ```rust
 
--- a/docs/adr/001_error_handling.md
+++ b/docs/adr/001_error_handling.md
@@ -1,107 +1,104 @@
 # Error handling patterns in public API interfaces
-
 ## Date
-17 Feb 2025 
+27 Feb 2025 
 
-## Accepted Option
+## Summary
 
-**Option 3** 
-
-## Context and Problem Statement
-There is uncertainty around how to model errors in the `opentelemetry-rust` public API interfaces - that is, APIs that are exposed to users of the project's published crates. This is for example the case with the exporter traits - [SpanExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/trace/export.rs#L18), [LogExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/logs/export.rs#L115), and [PushMetricExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/metrics/exporter.rs#L11) which form part of the API surface of `opentelemetry-sdk`.
+This ADR describes the general pattern we will follow when modelling errors in public API interfaces - that is, APIs that are exposed to users of the project's published crates. . It summarises the discussion and final option from [#2571](https://github.com/open-telemetry/opentelemetry-rust/issues/2571); for more context check out that issue. 
 
 We will focus on the exporter traits in this example, but the outcome should be applied to _all_ public traits and their fallible operations. 
 
+These include [SpanExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/trace/export.rs#L18), [LogExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/logs/export.rs#L115), and [PushMetricExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/metrics/exporter.rs#L11) which form part of the API surface of `opentelemetry-sdk`.
+
 There are various ways to handle errors on trait methods, including swallowing them and logging, panicing, returning a shared global error, or returning a method-specific error. We strive for consistency, and we want to be sure that we've put enough thought into what this looks like that we don't have to make breaking interface changes unecessarily in the future.
 
-This was discussed extensively in [#2571](https://github.com/open-telemetry/opentelemetry-rust/issues/2571).
+## Design Guidance
 
+### 1. Panics are only appropriate at startup
+Failures _after_ startup should not panic, instead returning errors to the caller where appropriate, _or_ logging an error if not appropriate.
+Some of the opentelemetry SDK interfaces are dictated by the specification in way such that they may not return errors. 
 
-## Related Work
+### 2. Consolidate error types within a trait where we can, let them diverge when we can't**
 
-* [#2564](https://github.com/open-telemetry/opentelemetry-rust/issues/2564)
-* [#2561](https://github.com/open-telemetry/opentelemetry-rust/issues/2561)
-* [#2381](https://github.com/open-telemetry/opentelemetry-rust/issues/2381)
+We aim to consolidate error types where possible _without indicating a function may return more errors than it can actually return_. 
+Here's an example:
 
-## Considered Options
+```rust
+enum ErrorOne {
+  TooBig,
+  TooSmall,
+}
 
-**Option 1: Continue as is**
-Continue the status quo, returning a mix of either nothing or the trait-wide error type. This is inconsistent and limits the caller's ability to handle errors. 
+enum ErrorTwo {
+  TooLong,
+  TooShort
+}
 
-**Option 2: Extend trait-wide error type to all methods on trait**
-In this option we have an error type per trait regardless of the potential error paths for the individual methods. Concretely if `fn (a)` can return `Failure1` and `Failure2`,  and `fn (b)` can return `FailureC`, we have a unified error type that contains `Failure`, `Failure2`, and `Failure3`.
+trait MyTrait {
+  fn action_one() -> Result<(), ErrorOne>;
 
- This will mean that callers will have to know how and when to discard errors from a particular API call based on an understanding of which subset of errors that particular call can make. 
+  // Action two and three share the same error type. 
+  // We do not introduce a common error MyTraitError for all operations, as this would
+  // force all methods on the trait to indicate they return errors they do not return,
+  // complicating things for the caller.  
+  fn action_two() -> Result<(), ErrorTwo>;
+  fn action_three() -> Result<(), ErrorTwo>;
+}
+```
+## 3. Consolidate error types between signals where we can, let them diverge where we can't
 
-Conversely, it will reduce the number of error types in the code base. 
-
-**Option 3: Use shared errors where practical across signals, devolve into individual errors per operation if they need to diverge**
-
-Here we aim to consolidate error types where possible _without indicating a function may return more errors than it can actually return_. Conversely in **Option 2**, a caller of either of the example functions is forced to handle or discard all errors. In this case, we choose to sacrifice the single error and diverge into a separate error for each operation.
-
-Our preference for error types is thus:
-
-1. Consolidated error that covers all methods of a particular "trait type" (e.g., signal export) and method
-1. Devolves into error type per method of a particular trait type (e.g., `SdkShutdownResult`, `SdkExportResult`) _if the error types need to diverge_
-1. May alternatively devolve into error type per signal (e.g., `SpanExporter`) if the _signals diverge_
-
-This approach generalises across both **signals** and **trait methods**. For example, returning to our exporter traits, we have a trait that looks the same for each signal, with the same three methods. Upon closer inspection ([#2600](https://github.com/open-telemetry/opentelemetry-rust/issues/2600)), the potential error set is the same both between the methods *and* between the signals; this means we can use a single shared error type across both axes:
+Consider the `Exporter`s mentioned earlier. Each of them has the same failure indicators - as dicated by the OpenTelemetry spec  - and we will
+share the error types accordingly: 
 
 ```rust
 
+/// opentelemetry-sdk::error
+
 #[derive(Error, Debug)]
-
-// Errors that can occur during SDK operations export(), force_flush() and shutdown().
 pub enum OTelSdkError {
-
-    /// All error types in here may be returned by any of the operations
-    /// of the exporters, on any of their APIs.
-    /// If this were not the case, we would not be able to use a shared error.
     #[error("Shutdown already invoked")]
     AlreadyShutdown,
-
-    // ... Other errors ...
-
-    /// The error message is intended for logging purposes only and should not
-    /// be used to make programmatic decisions. It is implementation-specific
-    /// and subject to change without notice. Consumers of this error should not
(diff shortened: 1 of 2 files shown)


docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,97 @@
...
+
+Conversely, it will reduce the number of error types in the code base. 
+
+**Option 3: Introduce an error-type per fallible operation, aggregate these into a single trait-wide error type**
+
+For example, in the above we'd have something like:
+```rust
+pub trait LogExporter {
+        
+	fn export(...) -> Result<..., ExportError>;
```
cijothomas, line 40: lets capture the actual outcome. We ended up with same Error for Export/Flush/Shutdown.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,97 @@
...
+	// Allow impls to box up errors that can't be logically mapped
+	// back to one of the APIs errors 
+	#[error("Unknown error (should not occur): {source:?}")] 
+	Unknown { 
+		source: Box<dyn std::error::Error + Send + Sync>, 
+	},
+}
+
+// Aggregate error type for convenience 
+// **Note**: This will be added in response to need, not pre-emptively
```
cijothomas, line 62: this maybe moved to a "Future addition" section, so it is easy to follow what the actual outcome, and curious readers/contributors can look at Future section for potential improvements to be added.

scottgerring: I think this was a bit of a rabbit hole that feels unlikely with a reasonably obvious solution, so i've dropped it completely.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+	fn export(...) -> OtelSdkResult;
+	fn shutdown(...) -> OtelSdkResult;
+  fn force_flush(...) -> OTelSdkResult;
+}
+
+```
+
+### When to box custom errors
+
+Note above that we do not box anything into `InternalFailure`. Our rule here is that if the caller cannot reasonably be expected to handle a particular error variant, we will use a simplified interface that returns only a descriptive string. In the concrete example we are using with the exporters, we have a [strong signal in the opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#export) that indicates concretely that the error types are not actionable by the caller.
```
cijothomas, line 105: Love this phrasing!

scottgerring:
@cijothomas is this good to merge? It'd be great to have a _concrete example_ in place so we can start to follow the pattern - for instance for the tracing interop Björn is working on

docs/adr/README.md
```diff
@@ -0,0 +1,5 @@
+# Architectural Decision Records
+
+This directory contains architectural decision records made for the opentelemetry-rust project. These allow us to consolidate discussion, options, and outcomes, around key architectural decisions. You can read more about ADRs [here](https://adr.github.io/).
```
cijothomas, line 3: I'd avoid links to the adr.github.io/similar. We can simply document the designs here, without necessarily adhering to any particular version of it.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+
+If the caller may potentially recover from an error, we will follow [canonical's rust best practices](https://canonical.github.io/rust-best-practices/error-and-panic-discipline.html) and instead preserve the nested error.
```
cijothomas, line 107: I'd suggest to remove the link to canonical guides. It is not clear whether we'll always follow it or not.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+
+Our preference for error types is thus:
+
+1. Consolidated error that covers all methods of a particular "trait type" (e.g., signal export) and method
+1. Devolves into error type per method of a particular trait type (e.g., `SdkShutdownResult`, `SdkExportResult`) _if the error types need to diverge_
+1. May alternatively devolve into error type per signal (e.g., `SpanExporter`) if the _signals diverge_
+
+This approach generalises across both **signals** and **trait methods**. For example, returning to our exporter traits, we have a trait that looks the same for each signal, with the same three methods. Upon closer inspection ([#2600](https://github.com/open-telemetry/opentelemetry-rust/issues/2600)), the potential error set is the same both between the methods *and* between the signals; this means we can use a single shared error type across both axes:
+
+```rust
```
cijothomas, line 50: let us put this under a header - Example/something, so its easy to link to from other discussions.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Related Work
+
+* [#2564](https://github.com/open-telemetry/opentelemetry-rust/issues/2564)
+* [#2561](https://github.com/open-telemetry/opentelemetry-rust/issues/2561)
+* [#2381](https://github.com/open-telemetry/opentelemetry-rust/issues/2381)
+
+## Considered Options
+
+**Option 1: Continue as is**
```
cijothomas, line 28: I still suggest to only list the decision here to avoid long reads. At the end of the doc, we can mention considered-alternatives, and move this there.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
+# Error handling patterns in public API interfaces
+
+## Date
+17 Feb 2025 
+
+## Accepted Option
+
+**Option 3**
```
cijothomas, line 8: I suggest to make it super easy for future reader to know what is the design guidance here, without having to scan through rest of doc. i.e Let's put the followed design right here, and a why that was chosen just below it.

Everything else can be moved to bottom of the doc. https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/log/DESIGN.md#rejected-alternatives has something like this.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+Conversely, it will reduce the number of error types in the code base. 
+
+**Option 3: Use shared errors where practical across signals, devolve into individual errors per operation if they need to diverge**
+
+Here we aim to consolidate error types where possible _without indicating a function may return more errors than it can actually return_. Conversely in **Option 2**, a caller of either of the example functions is forced to handle or discard all errors. In this case, we choose to sacrifice the single error and diverge into a separate error for each operation.
+
+Our preference for error types is thus:
+
+1. Consolidated error that covers all methods of a particular "trait type" (e.g., signal export) and method
+1. Devolves into error type per method of a particular trait type (e.g., `SdkShutdownResult`, `SdkExportResult`) _if the error types need to diverge_
```
cijothomas, line 45: The final outcome is - there is no separate Export vs Shutdown result.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+// The actionable errors returned by the exporter traits are effectively
+// the same for all operations; we can use a single shared error.
+// 
+
+use opentelemetry_sdk::error::OTelSdkResult;
+
+pub trait LogExporter {
+	fn export(...) -> OtelSdkResult;
+	fn shutdown(...) -> OtelSdkResult; 
+  fn force_flush(...) -> OTelSdkResult;
```
lalitb, line 90: nit - indentation

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// ...
+
+pub trait SpanExporter {
+	fn export(...) -> OtelSdkResult;
+	fn shutdown(...) -> OtelSdkResult;
+  fn force_flush(...) -> OTelSdkResult;
```
lalitb, line 98: nit - indentation

docs/adr/001_error_handling.md
lalitb, line 107: Just to be clear - if `InternalError` variant is defined as `InternalFailure(String),` - it strictly holds the string, and no boxing is possible. It would be good to clarify that, or specify the example for variant with boxing.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,104 @@
...
+
+We will focus on the exporter traits in this example, but the outcome should be applied to _all_ public traits and their fallible operations. 
+
+These include [SpanExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/trace/export.rs#L18), [LogExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/logs/export.rs#L115), and [PushMetricExporter](https://github.com/open-telemetry/opentelemetry-rust/blob/eca1ce87084c39667061281e662d5edb9a002882/opentelemetry-sdk/src/metrics/exporter.rs#L11) which form part of the API surface of `opentelemetry-sdk`.
+
+There are various ways to handle errors on trait methods, including swallowing them and logging, panicing, returning a shared global error, or returning a method-specific error. We strive for consistency, and we want to be sure that we've put enough thought into what this looks like that we don't have to make breaking interface changes unecessarily in the future.
+
+## Design Guidance
+
+### 1. Panics are only appropriate at startup
```
cijothomas, line 17: mm.. I am not sure we are ready to make a decision on panic-ing at startup either. Let's discuss it separately.
For this PR, let us add that panic-ing from sdk operations are *not* allowed.

scottgerring: Fair enough - updated!
I think if we can avoid it, we should avoid it. I _thought_ we had some startup panics, but I couldn't find anything outside of the tests.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
...
+    fn force_flush(&self, /* ... */ ) -> OTelSdkResult;
+    fn shutdown(&self, /* ... */ ) -> OTelSdkResult;
+```
+
+If this were _not_ the case - if we needed to mark an extra error for instance for `LogExporter` that the caller could reasonably handle - 
+we would let that error traits diverge at that point. 
+
+### 4. Box custom errors where a savvy caller may be able to handle them, stringify them if not
+
+Note above that we do not box anything into `InternalFailure`. Our rule here is that if the caller cannot reasonably be expected to handle a particular error variant, we will use a simplified interface that returns only a descriptive string. In the concrete example we are using with the exporters, we have a [strong signal in the opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#export) that indicates that the error types _are not actionable_ by the caller.
```
cijothomas, line 86:
```suggestion
Note above that we do not box any `Error` into `InternalFailure`. Our rule here is that if the caller cannot reasonably be expected to handle a particular error variant, we will use a simplified interface that returns only a descriptive string. In the concrete example we are using with the exporters, we have a [strong signal in the opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#export) that indicates that the error types _are not actionable_ by the caller. 
```

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+If this were _not_ the case - if we needed to mark an extra error for instance for `LogExporter` that the caller could reasonably handle - 
+we would let that error traits diverge at that point. 
+
+### 4. Box custom errors where a savvy caller may be able to handle them, stringify them if not
+
+Note above that we do not box anything into `InternalFailure`. Our rule here is that if the caller cannot reasonably be expected to handle a particular error variant, we will use a simplified interface that returns only a descriptive string. In the concrete example we are using with the exporters, we have a [strong signal in the opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#export) that indicates that the error types _are not actionable_ by the caller. 
+
+If the caller may potentially recover from an error, we will follow the generally-accepted best practice (e.g., see [canonical's guide](https://canonical.github.io/rust-best-practices/error-and-panic-discipline.html) and instead preserve the nested error:
```
cijothomas, line 88: nit addition: Mention that there is no place in the repo *today* that needed preserving error.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
...
+## Design Guidance
+
+### 1. No panics from SDK APIs
+Failures during regular operation should not panic, instead returning errors to the caller where appropriate, _or_ logging an error if not appropriate.
+Some of the opentelemetry SDK interfaces are dictated by the specification in way such that they may not return errors. 
+
+### 2. Consolidate error types within a trait where we can, let them diverge when we can't**
+
+We aim to consolidate error types where possible _without indicating a function may return more errors than it can actually return_. 
+Here's an example:
```
cijothomas, line 24: Agree with the content, but code example is not intuitive. May I suggest a pattern like
"don't do this"
show bad code
"instead, do this"
show correct one.

I think such a pattern makes it easier to follow.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
...
+pub trait PushMetricExporter {
+    fn export(&self, /* ... */) -> OtelSdkResult;
+    fn force_flush(&self, /* ... */ ) -> OTelSdkResult;
+    fn shutdown(&self, /* ... */ ) -> OTelSdkResult;
+```
+
+If this were _not_ the case - if we needed to mark an extra error for instance for `LogExporter` that the caller could reasonably handle - 
+we would let that error traits diverge at that point. 
+
+### 4. Box custom errors where a savvy caller may be able to handle them, stringify them if not
```
cijothomas, line 84: lets also mention we use `thiserror` to manage Error types. As indicated in its readme, it won't appear in public API, and we may remove it in the future (if there is a need), without breaking public API.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
...
+// pub trait LogExporter {
+// pub trait SpanExporter {
+pub trait PushMetricExporter {
+    fn export(&self, /* ... */) -> OtelSdkResult;
+    fn force_flush(&self, /* ... */ ) -> OTelSdkResult;
+    fn shutdown(&self, /* ... */ ) -> OTelSdkResult;
+```
+
+If this were _not_ the case - if we needed to mark an extra error for instance for `LogExporter` that the caller could reasonably handle - 
+we would let that error traits diverge at that point.
```
cijothomas, line 82: lets be clear that we *cannot* change the enum post 1.0 (unless when we bump to 2.0), as it'll be a breaking change; adding a a new variant/ removing one, changing type are all breaking change.

cijothomas (approved):
LGTM. Great start to following this practice throughout.

Left some small comments. Would love to get one more eyes to review!

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+There are various ways to handle errors on trait methods, including swallowing them and logging, panicing, returning a shared global error, or returning a method-specific error. We strive for consistency, and we want to be sure that we've put enough thought into what this looks like that we don't have to make breaking interface changes unecessarily in the future.
+
+## Design Guidance
+
+### 1. No panics from SDK APIs
+Failures during regular operation should not panic, instead returning errors to the caller where appropriate, _or_ logging an error if not appropriate.
+Some of the opentelemetry SDK interfaces are dictated by the specification in way such that they may not return errors. 
+
```
lalitb, line 20: Is it possible that some of the dependencies can panic - which may be the right approach when it happens during SDK configuration time. Or do we expect the SDK APIs to catch all such panic and return as error or log them? Either is fine to me, but good to document that.

cijothomas: Handling startup failures is not covered here. It should be discussed and added additionally.
I don't think we should panic. Returning Result should be the behavior.

Do we have situation where a dependency throws panic, and we cannot prevent it?

Anyway, we can discuss this separately.

lalitb (approved):
LGTM with nit comments. Thanks for adding this ADR, definitely bring clarity in our error handling design.

docs/adr/001_error_handling.md
```diff
@@ -0,0 +1,103 @@
+# Error handling patterns in public API interfaces
+## Date
+27 Feb 2025 
+
+## Summary
+
+This ADR describes the general pattern we will follow when modelling errors in public API interfaces - that is, APIs that are exposed to users of the project's published crates. . It summarises the discussion and final option from [#2571](https://github.com/open-telemetry/opentelemetry-rust/issues/2571); for more context check out that issue.
```
TommyCpp, line 7:
```suggestion
This ADR describes the general pattern we will follow when modelling errors in public API interfaces - that is, APIs that are exposed to users of the project's published crates. It summarises the discussion and final option from [#2571](https://github.com/open-telemetry/opentelemetry-rust/issues/2571); for more context check out that issue. 
```

docs/adr/001_error_handling.md
TommyCpp: +1 on nt panic in opentelemetry components. We cannot control panic in our dependencies but let's make a note that prinpical is no panic from openetlemety unless with _very good_ reason

docs/adr/001_error_handling.md
TommyCpp: > as it'll be a breaking change;

if the enum has [#[non_exhaustive]](https://doc.rust-lang.org/reference/attributes/type_system.html) adding a variant won't be breaking changes. User will still need to modify the code to handle it but at least it will still compile

For error types I think it's worth adding #[non_exhaustive] attributes to force consumer to think about "future errors" when they implementing it

docs/adr/001_error_handling.md
TommyCpp: +1, also should we consider declear error message itself can change at anytime to leave us some room for improvement in the future?

docs/adr/001_error_handling.md
scottgerring: I'm kinda skeptical of this; adding `#[non_exhaustive]` forces the user to add a default match, which doesn't change whether or not a change breaks your code, it just changes whether or not the compiler tells you about it.

docs/adr/001_error_handling.md
scottgerring: Added just below on L168

docs/adr/001_error_handling.md
scottgerring: Added a section below about thiserror

@TommyCpp I reckon this goes unsaid with the structured-error handling in the rust community, but if other folks feel strongly about it we could change it

scottgerring:
@cijothomas went through all the comments again, should be good!

cijothomas (approved):
Thanks!

PR #2668, opened 2025-02-14
Fix OTel Appender for tracing from suppressing other layers

### LINKED ISSUE #1682: otel-appender-tracing can suppress other tracing layers
With `logs_level_enabled` feature, the appender-tracing returns [false on `event_enabled`](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L185-L195) if it has no interest in the event. But this causes *other* layers, like a `fmt` layer to also miss the events. Lack of interest for otel layer should not influence other layers.

I believe the solution should be to rewrite the `logs_level_enabled` to use https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/index.html#per-layer-filtering

cijothomas:
Fixes #1682 

No change in performance.
Also modified benchmarks to remove the need of a noop exporter, as noop processor is sufficient to test perf for enabled/disabled. (This change has no perf impact even without this cleanup.)

### DIFF
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -2,13 +2,14 @@
 
 ## vNext
 
+Fixes [1682](https://github.com/open-telemetry/opentelemetry-rust/issues/1682).
+"spec_unstable_logs_enabled" feature now do not suppress logs for other layers.
+
 ## 0.28.1
 
 Released 2025-Feb-12
 
-- Bump `tracing-opentelemetry` to 0.29
-- New experimental feature to use trace\_id & span\_id from spans created through the [tracing](https://crates.io/crates/tracing) crate (experimental_use_tracing_span_context) [#2438](https://github.com/open-telemetry/opentelemetry-rust/pull/2438)
-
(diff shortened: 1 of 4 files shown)


opentelemetry-appender-tracing/CHANGELOG.md
```diff
@@ -2,13 +2,14 @@
...
 ## vNext
 
+Fixes [1682](https://github.com/open-telemetry/opentelemetry-rust/issues/1682).
+"spec_unstable_logs_enabled" feature now do not suppress logs for other layers.
+
 ## 0.28.1
 
 Released 2025-Feb-12
 
-- Bump `tracing-opentelemetry` to 0.29
```
lalitb, line 9: this is removed, is it intended?

lalitb (approved):
Thanks for fixing this.

opentelemetry-appender-tracing/CHANGELOG.md
cijothomas: yes. we never released with old version!

lalitb: We released - https://crates.io/crates/opentelemetry-appender-tracing/0.28.1

cijothomas: I meant - we never released a version which uses tracing-opentelemetry 0.28 or old.  So there is no need to state in changelog that we updated tracing-opentelemetry to 0.29!

PR #2672, opened 2025-02-15
leverage native async trait in MetricsClient

gruebel:
Relates #2517 

## Changes

- I think this is the last `asyn_trait` we can remove. there is still some usage around `HttpClient` trait, but this one is used quite often with dynamic dispatch and this can't be handled by the native async capabilities.


cijothomas (approved):
Nice! Thank you.

PR #2674, opened 2025-02-16
use workspace lint configuration

gruebel:
## Changes

- Rust added in 1.74 (I think 😅) the possibility to configure the lints for a whole workspace in the `Cargo.toml`. Since `prometheus` doesn't belong to the workspace it is handled separately. Now you can easily configure lints for all workspace crates 🍻 

### DIFF
--- a/opentelemetry-appender-log/Cargo.toml
+++ b/opentelemetry-appender-log/Cargo.toml
@@ -34,3 +34,6 @@ opentelemetry-stdout = { path = "../opentelemetry-stdout", features = ["logs"] }
 log = { workspace = true, features = ["kv_serde"] }
 tokio = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"] }
+
+[lints]
+workspace = true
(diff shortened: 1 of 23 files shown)


opentelemetry-appender-log/Cargo.toml
```diff
@@ -34,3 +34,6 @@ opentelemetry-stdout = { path = "../opentelemetry-stdout", features = ["logs"] }
 log = { workspace = true, features = ["kv_serde"] }
 tokio = { workspace = true }
 serde = { workspace = true, features = ["std", "derive"] }
+
+[lints]
+workspace = true
```
lalitb, line 39: Just curious, since clippy lints are already defined at the workspace level, won't that get inherited here automatically? Or do we also need `workspace = true` ?

gruebel: Sadly not, it is similar to defining workspace dependency versions. Each crate needs to opt-in.https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table

lalitb (approved):

Issue #2670, opened 2025-02-15
BatchProcessor test must be modified to be reliable in CI


cijothomas:
Tests relying on sleeps is inevitable sometimes, but that must be limited.
Modify normal tests to rely on flush/shutdown, instead of timer trigger, as they are flaky, esp. when run in CI machines.

PR #2675, opened 2025-02-16
Modify batch processor test use flush instead of sleeptime

mohammadVatandoost:
For #2670


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -1263,7 +1263,7 @@ mod tests {
             processor.on_end(span);
         }
 
-        tokio::time::sleep(Duration::from_millis(1000)).await;
+        processor.force_flush().unwrap();
```
cijothomas, line 1266: since we are relying on force_flush, let us modify the BatchConfig to stick with defaults, instead of overriding to 50 msecs.

cijothomas: @mohammadVatandoost Just checking if you are planning to make this suggestion in this PR itself?

mohammadVatandoost: Done

cijothomas (approved):

PR #2677, opened 2025-02-18
Calls to `with_resource` for signal builders (Metrics, Logs, Traces) are now additive

### LINKED ISSUE #2324: [Feature]: add ability for with_resource to merge Resources on providers
### Related Problems?

Resulting from a conversation with @cijothomas about `ResourceBuilder` and Signal Providers

Would like to hear thoughts on this!

Instead of replacing the Resource on a Provider if `.with_resource` is called multiple times, it would be good if we could build up a resource from many calls on the provider.

A few on how this could work:

We could add something like `.with_resource_builder(…)` that will aggregate on top of the existing resource / other builders in the provider.

Or update the .with_resource to just merge with existing resources so it can be called multiple times.

Or maybe both

### Considered Alternatives

_No response_

### Additional Context

relates to: https://github.com/open-telemetry/opentelemetry-rust/issues/2320

inspired by:
- https://github.com/open-telemetry/opentelemetry-dotnet/issues/2909
- https://github.com/open-telemetry/opentelemetry-dotnet/pull/3307/files

martintmk:
Fixes #2324

## Changes

The calls to:

- `MeterProviderBuilder::with_resource`
- `TracerProviderBuilder::with_resource`
- `LoggerProviderBuilder::with_resource`

Are now additive and the previous resource is not discarded anymore. Additionally, I fixed a bug where setting the `schema_url` was discarded due to a bug how merging of resources was handled.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## vNext
 
-- *Breaking* Make `force_flush()` in `PushMetricExporter` synchronous
+- *Breaking*: Calls to `MeterProviderBuilder::with_resource` are now additive ([#2324](https://github.com/open-telemetry/opentelemetry-rust/issues/2324)).
+- *Breaking*: Make `force_flush()` in `PushMetricExporter` synchronous
 
 ## 0.28.0
 
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## vNext
 
-- *Breaking*: Calls to `MeterProviderBuilder::with_resource`, `TracerProviderBuilder::with_resource`, 
+- Calls to `MeterProviderBuilder::with_resource`, `TracerProviderBuilder::with_resource`, 
   `LoggerProviderBuilder::with_resource` are now additive ([#2677](https://github.com/open-telemetry/opentelemetry-rust/pull/2677)).
 - *Breaking*: Make `force_flush()` in `PushMetricExporter` synchronous
 
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## vNext
 
-- *Breaking*: Calls to `MeterProviderBuilder::with_resource` are now additive ([#2324](https://github.com/open-telemetry/opentelemetry-rust/issues/2324)).
+- *Breaking*: Calls to `MeterProviderBuilder::with_resource`, `TracerProviderBuilder::with_resource`, 
+  `LoggerProviderBuilder::with_resource` are now additive ([#2677](https://github.com/open-telemetry/opentelemetry-rust/pull/2677)).
 - *Breaking*: Make `force_flush()` in `PushMetricExporter` synchronous
 
 ## 0.28.0
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -232,10 +232,16 @@ impl MeterProviderBuilder {
     /// with all [Meter]s the [MeterProvider] will create.
     ///
     /// By default, if this option is not used, the default [Resource] will be used.
+    /// Note that the calls to this method are additive, each call merges the provided
+    /// resource with the previous one.
     ///
     /// [Meter]: opentelemetry::metrics::Meter
     pub fn with_resource(mut self, resource: Resource) -> Self {
-        self.resource = Some(resource);
+        self.resource = match self.resource {
+            Some(existing) => Some(existing.merge(&resource)),
+            None => Some(resource),
+        };
+
         self
     }
 
@@ -331,6 +337,7 @@ mod tests {
     use opentelemetry::{global, InstrumentationScope};
     use opentelemetry::{Key, KeyValue, Value};
     use std::env;
+    use crate::metrics::SdkMeterProvider;
 
     #[test]
     fn test_meter_provider_resource() {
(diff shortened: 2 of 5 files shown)


cijothomas (changes requested):
I agree with this change. Requesting changes to address below.
1. https://github.com/open-telemetry/opentelemetry-rust/issues/2324#issuecomment-2666129606 - Check this and see if this makes sense, given there are more than one ideas suggested in the linked issue. 
2. Please make the change consistent across signals. (Okay to do that in separate PR to keep scope smaller)

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -232,10 +232,16 @@ impl MeterProviderBuilder {
     /// with all [Meter]s the [MeterProvider] will create.
     ///
     /// By default, if this option is not used, the default [Resource] will be used.
+    /// Note that the calls to this method are additive, each call merges the provided
+    /// resource with the previous one.
     ///
     /// [Meter]: opentelemetry::metrics::Meter
     pub fn with_resource(mut self, resource: Resource) -> Self {
```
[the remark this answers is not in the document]
martintmk: Ok, will adopt this to other providers too

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -232,10 +232,16 @@ impl MeterProviderBuilder {
...
-        self.resource = Some(resource);
+        self.resource = match self.resource {
+            Some(existing) => Some(existing.merge(&resource)),
+            None => Some(resource),
```
lalitb, line 242: Not sure if it is only me, but this is causing confusion. A Resource object represents a single instance with multiple attributes, so `with_resource()` should ideally set the resource rather than merging attributes. If we want to support additive or cooperative resource setting, we could introduce a separate method like `with_merge_resource` or `with_append_resource` (or similiar) to make the merging behavior explicit. This differs from `with_processor`, where multiple processors can be configured.

martintmk: In the end, this is solving the same problem as in:
https://github.com/open-telemetry/opentelemetry-dotnet/issues/2909

Do we want to deviate from a solved problem?

Personally, I was surprised that `with_resource` had replace behavior, but it's purely subjective opinion.

cijothomas: OTel .NET was different, as it named its API "Set-Resource", so it was easy to tell that it follows "Set" semantics, and it overrides any existing. It introduced a *new* API called `ConfigureResource`, which operates on the `ResourceBuilder`, and *not* Resource.
In other words, what is proposed in this PR is not same as what .NET had/has.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2324#issuecomment-2666129606 see my comment on the issue. We can discuss there, and once agreed on the API share, we can come back to this PR.

opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -169,11 +169,11 @@ impl Resource {
     /// 5. If both resources do not have a schema url, the schema url will be empty.
     ///
     /// [Schema url]: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.9.0/specification/schemas/overview.md#schema-url
-    fn merge<T: Deref<Target = Self>>(&self, other: T) -> Self {
-        if self.is_empty() {
+    pub(crate) fn merge<T: Deref<Target = Self>>(&self, other: T) -> Self {
```
lalitb, line 172: This is also a breaking change, and good to explicitly mention in the changelog.

lalitb: And this should be made crate-private only once `with_resource` is made additive for all the signals.

martintmk: The `merge` function was private before and now I just made it internal to the crate. This should not cause any breaking changes in public API, or am I missing something?

lalitb: Yes, sorry for confusion.

lalitb (approved):

opentelemetry-sdk/src/metrics/meter_provider.rs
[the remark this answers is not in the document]
martintmk: Implemented for other signals too.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## vNext
 
-- *Breaking* Make `force_flush()` in `PushMetricExporter` synchronous
+- *Breaking*: Calls to `MeterProviderBuilder::with_resource` are now additive ([#2324](https://github.com/open-telemetry/opentelemetry-rust/issues/2324)).
```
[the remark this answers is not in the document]
martintmk: updated

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## vNext
 
-- *Breaking* Make `force_flush()` in `PushMetricExporter` synchronous
+- *Breaking*: Calls to `MeterProviderBuilder::with_resource`, `TracerProviderBuilder::with_resource`,
```
cijothomas, line 5:
```suggestion
- Calls to `MeterProviderBuilder::with_resource`, `TracerProviderBuilder::with_resource`, 
```

cijothomas: My suggestion is to just mention the change, and not call it "breaking" to avoid scaring people unnecessarily!

cijothomas (approved):

cijothomas:
@martintmk Edited the PR Desc slightly to mark the associated issue to be resolved with this PR.
Nice work, thanks for exploring and weighing all options and driving closure!
(I removed "breaking" wording from changelog.)

PR #2682, opened 2025-02-18
Fix 2638: Use absolute URL for LICENSE/CHANGELOG in README

### LINKED ISSUE #2638: Changelog in crates.io points to 404
https://crates.io/crates/opentelemetry-otlp - link to changelog need to be fixed.
(All crates)

Shunpoco:
Fixes #2638 

## Changes
Use absolute URL (for example, `https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-proto/CHANGELOG.md`) instead of a relative URL (`./CHANGELOG.md`).

I found that cargo.io treats a relative URL like `<repository URL>/<commit and path in .cargo_vcs_info.json>/<the relative path>`. So the relative path is converted like `https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-otlp/blob/HEAD/opentelemetry-otlp/CHANGELOG.md`. This causes 404 errors.

To solve this, we have two options:
1. Use an absolute URL instead of a relative URL (this PR takes this option)
2. Change repository URL to the root in each Cargo.toml (`https://github.com/open-telemetry/opentelemetry-rust`)
    - cargo.io treats the relative path like `https://github.com/open-telemetry/opentelemetry-rust/blob/HEAD/opentelemetry-otlp/CHANGELOG.md`, it is a valid URL
    - Because ./opentelemetry and ./opentelemetry-sdk do so, their crate page doesn't have 404s.

At first I thought to take the option 2, but I found this issue: https://github.com/rust-lang/crates.io/issues/9939.
The issue is not the exact same as our pattern, but they also suffer that cargo.io doesn't treat a relative path as they expected. And the cargo.io maintainer says that developer should use an absolute URL. If we take the option 2, for now it fixes the problem but in the future if cargo.io team does some breaking changes, our URL may be broken again. So I took the first option.

NOTE: In my second commit, I changed homepage/repository URL on opentelemetry/opentelemetry-http/opentelemetry-sdk's Cargo.toml since they are inconsistent against other sub-dir. But technically this change is not related to #2638 .


cijothomas (approved):
Thank you!

Issue #2150, opened 2024-09-26
[Feature]: Implement error recording on opentelemetry-appender-tracing


Yamakaky:
### Describe the solution you'd like:

Currently, `opentelemetry-appender-tracing` records errors using their `Debug` implementation. The cause chain has to be added manually, which is a pain to do at every log point. 
I believe this would be done in https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L72.

Otel conventions for errors: https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/
`tracing-opentelemetry` implementation for reference: https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L297

lalitb:
@Yamakaky That's a fair ask. Would you like to contribute?

Yamakaky:
Sure. Some things I'm not sure about:
```rust
let error = ...;
error!(error);
error!(error, "some error happened");
error!(error, "some error happened: {}", error);
```
- Which option do we recommend?
    - 1 would not have a log message, maybe have `error.to_string()` as default message value?
    - 3 makes it more verbose.
- New `exception.*` fields are added, cf semconv. Do we keep the `error` field as duplicate? I'd say to remove it
- What about if there are multiple errors fields, or if `exception.*` are manually added? I'd say to raise an error.

lalitb:
@Yamakaky . Regarding:

```rust
let error = ...;
error!(error);
error!(error, "some error happened");
error!(error, "some error happened: {}", error);
```
I am confused as none of the above you shared are the valid error calls. They won't compile.

I can think of below scenarios: 

`error!("An error occurred: {}", e);` ->  Uses `Display`, not triggering `record_error`. 
 `error!(error = ?e, "An error occurred");` -> Uses `Debug`, not triggering `record_error`.
` error!(error = &e as &dyn Error, "An error occurred");` -> This would trigger `record_error` (3)
` error!(error = &e as &dyn Error);` -> This would trigger `record_error` (4)

And in 4th case above, we don't need any default message. Just the `error` attribute is enough.


Also, we don't need to support semconv for exception for now, as they are still experimental.

Yamakaky:
I was doing something like that:
```rust
    let e = anyhow::anyhow!("pote").context("prout");
    error!(error = e.as_ref() as &dyn std::error::Error);
```
[...]

PR #2689, opened 2025-02-20
Appender-Tracing - extend spl treatment of message when recording str

cijothomas:
TODO: Tests in next PR, if this is merged.

Follow up from https://github.com/open-telemetry/opentelemetry-rust/pull/2001/files

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -88,15 +88,14 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
         if is_duplicated_metadata(field.name()) {
             return;
         }
-        //TODO: Consider special casing "message" to populate body and document
-        // to users to use message field for log message, to avoid going to the
-        // record_debug, which has dyn dispatch, string allocation and
-        // formatting cost.
-
         //TODO: Fix heap allocation. Check if lifetime of &str can be used
         // to optimize sync exporter scenario.
-        self.log_record
-            .add_attribute(Key::new(field.name()), AnyValue::from(value.to_owned()));
+        if field.name() == "message" {
+            self.log_record.set_body(AnyValue::from(value.to_owned()));
+        } else {
+            self.log_record
+                .add_attribute(Key::new(field.name()), AnyValue::from(value.to_owned()));
+        }
     }
 
     fn record_bool(&mut self, field: &tracing_core::Field, value: bool) {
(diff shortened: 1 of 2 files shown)


cijothomas:
@mladedav Could you also review this?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -88,15 +88,14 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
-        //TODO: Consider special casing "message" to populate body and document
-        // to users to use message field for log message, to avoid going to the
-        // record_debug, which has dyn dispatch, string allocation and
-        // formatting cost.
-
         //TODO: Fix heap allocation. Check if lifetime of &str can be used
         // to optimize sync exporter scenario.
-        self.log_record
-            .add_attribute(Key::new(field.name()), AnyValue::from(value.to_owned()));
+        if field.name() == "message" {
```
lalitb, line 93: This would be a breaking change if user is expecting the message to go an attribute?

cijothomas: > This would be a breaking change if user is expecting the message to go an attribute?

There was no contract documented about this, so while this change behavior, I don't think it should be a major issue. (We were already special casing "message" for the record_debug, so missing this may also be considered a bug)

cijothomas: > And what should be the expected behavior if both message attribute and default body field is provided

We can add tests to make sure this is covered, but the behavior is really controlled by tracing itself. We can make some decision and document it in the event of conflict, as there can only be one body.

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/src/layer.rs#L121 Fixing these TODOs are also going to change the behavior, so is fixing error to be reported specially. It's questionable to that 1.0 (but maybe okay), but I don't see an issue doing it at this stage. Let me know your thoughts.

lalitb (approved):

opentelemetry-appender-tracing/src/layer.rs
cijothomas: We need to document the final behavior before 1.0.
https://github.com/open-telemetry/opentelemetry-rust/issues/2150 will also be related to this...

cijothomas:
> Sorry for the late response but yeah, it seems fine.

Thanks!

PR #2695, opened 2025-02-21
Remove `mut` self reference from SpanExporter::export() method.

lalitb:
## Changes

To make it consistent with `LogExporter` as done in #2380. This doesn't affect OTLP exporters, as their export are stateless.  If exporter need to modify its internal state during export process (e.g. batching etc.), it should rely on the interior mutability.


cijothomas (approved):
Nice!

PR #2696, opened 2025-02-21
Fix: BatchLogProcessor to invoke shutdown on exporter

cijothomas:
This area still requires more testing coverage and possibly better error logging. But this is a start!


lalitb (approved):
LGTM. Do we need to check for async code too ?

PR #2699, opened 2025-02-21
Fix `.with_headers` to support multiple k/v pairs

joshwlewis:
## Changes

Currently, `.with_headers` doesn't work consistently when passing in multiple k/v pairs. For instance, something like this only adds one of the headers:

```rust
HttpExporterBuilder::default()
    .with_headers(HashMap::from([
        ("k1".to_string(), "v1".to_string()),
        ("k2".to_string(), "v2".to_string()),
    ]))
```

This is quite surprising -- only one header is actually accepted, and it's a random choice. The others are lost. To work around this with the existing code, add one k/v at a time like this:

```rust
HttpExporterBuilder::default()
    .with_headers(HashMap::from([("k1".to_string(), "v1".to_string())]))
    .with_headers(HashMap::from([("k2".to_string(), "v2".to_string())]));
```

I've updated the tests to cover the surprising scenario, and adjusted the logic to support multiple k/v pairs.


cijothomas (approved):
Thanks! Could you add a one-line to the changelog for this bug fix.

lalitb (approved):

Issue #2697, opened 2025-02-21
Issues with new logging model


Sushisource:
I recently upgraded from `0.26 -> 0.28` and found that the global error handler has been removed: https://github.com/open-telemetry/opentelemetry-rust/pull/2260 in favor of Tokio tracing logging.

That's great, however there are at least a few pain points:

Various things logged at info level that don't need to be: Ex [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/periodic_reader.rs#L164) and [here](https://github.com/open-telemetry/opentelemetry-rust/blob/ee76205a8adff2a27bb214e1fdb7ba2819df84af/opentelemetry-sdk/src/metrics/meter_provider.rs#L314) (and many other places). I would argue there is next to no reason for a library like this to log at info level. [This example](https://github.com/open-telemetry/opentelemetry-rust/blob/ee76205a8adff2a27bb214e1fdb7ba2819df84af/opentelemetry/src/global/metrics.rs#L29) is maybe appropriate, but pretty much any consumer is not going to be interested in lifecycle events like the others.

The new thread for metrics export [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/periodic_reader.rs#L158) will never log properly unless the _global_ tracing subscriber is set. I am a library author consuming this library. I can't set a global subscriber, because my users might want to. However, I do want the subscriber my users tell me to use to be set as the subscriber my dependencies, like otel use. Normally this isn't a problem because I make sure all threads I spawn set the configured subscriber, but I can't control the subscriber for this (or any other) thread spawned internally by otel. It would be great if there was a way to configure the subscriber for exporters, so that it would be used by any spawned threads.

cijothomas:
Agree that some INFO level ones could be revisited to downgrade to DEBUG level. Would you mind sending a PR with your proposal, so we can review and adjust?

cijothomas:
> The new thread for metrics export [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/periodic_reader.rs#L158) will never log properly unless the global tracing subscriber is set.

True. Do you have a proposal in mind to help ease this? (It's totaly okay if you don't have!) Let us keep this issue open, and look for solutions...

Sushisource:
[...]

PR #2700, opened 2025-02-21
Reduce various info logs to debug

Sushisource:
Fixes #
Design discussion issue (if applicable) #2697

## Changes

Reduce various recently introduced info level logs to debug.

Many of these are lifecycle related and unlikely to be interesting to most users.

### DIFF
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -23,7 +23,7 @@ use crate::{
 };
 use std::sync::mpsc::{self, RecvTimeoutError, SyncSender};
 
-use opentelemetry::{otel_debug, otel_error, otel_info, otel_warn, InstrumentationScope};
+use opentelemetry::{otel_debug, otel_error, otel_warn, InstrumentationScope};
 
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::{cmp::min, env, sync::Mutex};
@@ -342,7 +342,7 @@ impl BatchLogProcessor {
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Logs.BatchProcessor".to_string())
             .spawn(move || {
-                otel_info!(
+                otel_debug!(
                     name: "BatchLogProcessor.ThreadStarted",
                     interval_in_millisecs = config.scheduled_delay.as_millis(),
                     max_export_batch_size = config.max_export_batch_size,
@@ -476,7 +476,7 @@ impl BatchLogProcessor {
                         }
                     }
                 }
-                otel_info!(
+                otel_debug!(
                     name: "BatchLogProcessor.ThreadStopped"
                 );
             })
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -160,7 +160,7 @@ impl LoggerProviderInner {
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         if !self.is_shutdown.load(Ordering::Relaxed) {
-            otel_info!(
+            otel_debug!(
                 name: "LoggerProvider.Drop",
                 message = "Last reference of LoggerProvider dropped, initiating shutdown."
             );
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -160,7 +160,7 @@ impl LoggerProviderInner {
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         if !self.is_shutdown.load(Ordering::Relaxed) {
-            otel_debug!(
+            otel_info!(
                 name: "LoggerProvider.Drop",
                 message = "Last reference of LoggerProvider dropped, initiating shutdown."
             );
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -110,7 +110,7 @@ impl SdkMeterProvider {
     /// There is no guaranteed that all telemetry be flushed or all resources have
     /// been released on error.
     pub fn shutdown(&self) -> OTelSdkResult {
-        otel_info!(
+        otel_debug!(
             name: "MeterProvider.Shutdown",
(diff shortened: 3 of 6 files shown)


opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -342,7 +342,7 @@ impl BatchLogProcessor {
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Logs.BatchProcessor".to_string())
             .spawn(move || {
-                otel_info!(
+                otel_debug!(
```
cijothomas, line 345: agree with this change.

opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -475,7 +475,7 @@ impl BatchLogProcessor {
                         }
                     }
                 }
-                otel_info!(
+                otel_debug!(
```
cijothomas, line 478: agree with this change.

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -160,7 +160,7 @@ impl LoggerProviderInner {
 impl Drop for LoggerProviderInner {
     fn drop(&mut self) {
         if !self.is_shutdown.load(Ordering::Relaxed) {
-            otel_info!(
+            otel_debug!(
```
cijothomas, line 163: would like to discuss more on this one. My original intention behind keeping this info was:
If a user forgets to keep the LoggerProvider around and its gets dropped, then Logs will not flow. This log message would have helped figure out why logs are not flowing in that scenario.

Given majority users do the right thing (i.e they keep LoggerProvider active), this should not cause a noise, right? And we recommend people to explicitly invoke shutdown(), so the else branch of this code is going to be executed in normal flow, which only does debug level.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -110,7 +110,7 @@ impl SdkMeterProvider {
     /// There is no guaranteed that all telemetry be flushed or all resources have
     /// been released on error.
     pub fn shutdown(&self) -> OTelSdkResult {
-        otel_info!(
+        otel_debug!(
```
cijothomas, line 113: agree.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -310,7 +310,7 @@ impl MeterProviderBuilder {
             }),
         };
 
-        otel_info!(
+        otel_debug!(
```
cijothomas, line 313: agree

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -153,7 +153,7 @@ impl Drop for SdkMeterProviderInner {
                 message = "MeterProvider was already shut down; drop will not attempt shutdown again."
             );
         } else {
-            otel_info!(
+            otel_debug!(
```
cijothomas, line 156: same comment as https://github.com/open-telemetry/opentelemetry-rust/pull/2700/files#r1966358902
worth discussing more before changing.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -317,7 +316,7 @@ impl BatchSpanProcessor {
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Traces.BatchProcessor".to_string())
             .spawn(move || {
-                otel_info!(
+                otel_debug!(
```
cijothomas, line 319: agree

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -413,7 +412,7 @@ impl BatchSpanProcessor {
                         }
                     }
                 }
-                otel_info!(
+                otel_debug!(
```
cijothomas, line 415: agree

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -312,7 +312,7 @@ impl<E: PushMetricExporter> PeriodicReader<E> {
                         }
                     }
                 }
-                otel_info!(
+                otel_debug!(
```
cijothomas, line 315: agree

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -251,16 +251,16 @@ impl<E: PushMetricExporter> PeriodicReader<E> {
                             // tells that it timed out.
                             if export_result.is_err() || shutdown_result.is_err() {
                                 if response_sender.send(false).is_err() {
-                                    otel_info!(
+                                    otel_debug!(
```
cijothomas, line 254: Would like to discuss more on this. The code comment above explains why this is Info. See if you agree with this.

Again, this should not occur in normal workflows so should not cause cogs issues.

Sushisource: My reading of the comment was that while I understand it's an unexpected situation, it's not entirely clear that it's actionable. If the shutdown-calling thread went away, presumably they no longer care about this?

opentelemetry-sdk/src/logs/logger_provider.rs
Sushisource: Yeah, that makes sense to me. Changed both of these ones back

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: This is actionable to service owners: If they see this log often, it is an indication to them that they need to increase their timeouts.

(Nothing they can do immediately as the shutdown() is already over, and there is no option to try-again)

cijothomas (changes requested):
Looks good except one place : https://github.com/open-telemetry/opentelemetry-rust/pull/2700/files#r1966362700 
Once this is resolved, we are good to merge.

(Also, there is merge conflict due to changelog. Please resolve that too to merge)

opentelemetry-sdk/src/metrics/periodic_reader.rs
Sushisource: Changed back

cijothomas (approved):

PR #2701, opened 2025-02-21
InstrumentationScope to include attributes in hash and eq check

### LINKED ISSUE #2225: Instrumentation Scope should be identifying
Spec newly clarification that Scope attributes are identifying. We need to reflect this for all signals.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md#logs-1

cijothomas:
Fixes #2225

For Metrics this affects aggregation (SDK), as now different Meter is returned for scope with different attributes.
Logs,Traces do not store Logger/Tracers, so this requires no change in SDKs.

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -563,6 +563,25 @@ mod tests {
     }
 
     #[test]
+    fn same_meter_reused_same_scope_attributes() {
+        let meter_provider = super::SdkMeterProvider::builder().build();
+        let make_scope = |attributes| {
+            InstrumentationScope::builder("test.meter")
+                .with_version("v0.1.0")
+                .with_schema_url("http://example.com")
+                .with_attributes(attributes)
+                .build()
+        };
+
+        let _meter1 =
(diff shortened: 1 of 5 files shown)


opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -562,6 +562,25 @@ mod tests {
...
+                .with_attributes(attributes)
+                .build()
+        };
+
+        let _meter1 =
+            meter_provider.meter_with_scope(make_scope(vec![KeyValue::new("key", "value1")]));
+        let _meter2 =
+            meter_provider.meter_with_scope(make_scope(vec![KeyValue::new("key", "value1")]));
+
+        assert_eq!(meter_provider.inner.meters.lock().unwrap().len(), 1);
```
lalitb, line 581: nit - add test with different attributes, just to ensure that existing behavior is not affected.

lalitb (approved):
LGTM

opentelemetry-sdk/src/metrics/meter_provider.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2702 Yes there are various combinations that need to have good coverage. Opening an issue to track them overall.

Issue #251, opened 2020-10-06
Research best way to track benchmarks across PRs


jtescher:
Currently benchmarks are only run manually, would be nice to have a history and possibly fail if there are significant regressions.

jtescher:
@awiede mentioned https://github.com/marketplace/actions/continuous-benchmark as a possible solution.

hdost:
https://github.com/benchmark-action/github-action-benchmark/blob/master/.github/workflows/rust.yml

cijothomas:
Removed from Tracing API stable milestone.

PR #2706, opened 2025-02-25
ci: Add criterion performance regressions to PR workflows

scottgerring:
This PR follows on from #2705, introducing a workflow that publishes a criterion regression comparison between `main` and the current branch.

It cannot be merged and the build will not succeed until #2705 is in place.


cijothomas:
Closes https://github.com/open-telemetry/opentelemetry-rust/issues/251

scottgerring:
@cijothomas this is good to go. 

I'm _not sure_ if the permissions are going to work for this to submit to the PR feedback from forks. I am hopeful that once the workflow is in main it will, but we'll see. If it doesn't I'll spend some more time looking at if we can set the perms up differently to achieve this (perhaps with repository settings) after this goes in.

Regardless, you always see the performance comparison in the job run:

https://github.com/open-telemetry/opentelemetry-rust/actions/runs/13538282948/job/37833792773?pr=2706

cijothomas (approved):

cijothomas:
> @cijothomas this is good to go.
> 
> I'm _not sure_ if the permissions are going to work for this to submit to the PR feedback from forks. I am hopeful that once the workflow is in main it will, but we'll see. If it doesn't I'll spend some more time looking at if we can set the perms up differently to achieve this (perhaps with repository settings) after this goes in.
> 
> Regardless, you always see the performance comparison in the job run:
> 
> https://github.com/open-telemetry/opentelemetry-rust/actions/runs/13538282948/job/37833792773?pr=2706

merged. let us see !

scottgerring:
@cijothomas `benchmark_collect_histogram intermittently` fails. I don't want to introduce a CI check that breaks the build. I'd rather comment out this assertion for now with a TODO so we can get this in. What do you think? 

```
thread 'main' panicked at opentelemetry-sdk/benches/metric.rs:348:36:
index out of bounds: the len is 0 but the index is 0
```

```rust
fn benchmark_collect_histogram(b: &mut Bencher, n: usize) {
    let r = SharedReader(Arc::new(ManualReader::default()));
    let mtr = SdkMeterProvider::builder()
        .with_reader(r.clone())
        .build()
        .meter("sdk/metric/bench/histogram");

    for i in 0..n {
        let h = mtr.u64_histogram(format!("fake_data_{i}")).build();
        h.record(1, &[]);
    }

    let mut rm = ResourceMetrics {
        resource: Resource::builder_empty().build(),
        scope_metrics: Vec::new(),
    };

    b.iter(|| {
        let _ = r.collect(&mut rm);
        assert_eq!(rm.scope_metrics[0].metrics.len(), n); // Fails here
    })
```

PR #2709, opened 2025-02-26
docs: Modify examples to show best practices - reuse tracer

cijothomas:
One of many with the goal of showcasing best practices in all examples maintained in this repo.
This PR shows how to re-use Tracer instead of repeatedly obtaining it from provider/global which involve at-least one atomic check and an Arc clone internally.
A simple benchmark was added to show the impact of Tracer creation.

If the direction is agreed, we need to review every examples and make sure they show best practices for all signals.


lalitb (approved):
Nice.

PR #2710, opened 2025-02-26
docs: Fix tracing grpc example

cijothomas:
This example was not building as required features were not enabled.
Also modified provider to use SimpleExporter, consistent with rest of examples to immediately see spans printed to console.
changed port from 50051 to 50052, as 50051 is what most tutorials showcase and hence likely to clash port.
Nit additions to span attributes. Eventually we need this to match the OTel conventions, but not this PR.

### DIFF
--- a/examples/tracing-grpc/Cargo.toml
+++ b/examples/tracing-grpc/Cargo.toml
@@ -19,7 +19,7 @@ opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["rt-tokio"] 
 opentelemetry-stdout = { workspace = true, features = ["trace"] }
 prost = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
-tonic = { workspace = true, features = ["server"] }
+tonic = { workspace = true, features = ["server", "codegen", "channel", "prost"] }
 
 [build-dependencies]
 tonic-build = { workspace = true }
--- a/examples/tracing-grpc/src/client.rs
+++ b/examples/tracing-grpc/src/client.rs
@@ -13,7 +13,7 @@ fn init_tracer() -> sdktrace::SdkTracerProvider {
     global::set_text_map_propagator(TraceContextPropagator::new());
     // Install stdout exporter pipeline to be able to retrieve the collected spans.
     let provider = sdktrace::SdkTracerProvider::builder()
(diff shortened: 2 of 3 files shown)


examples/tracing-grpc/src/client.rs
```diff
@@ -58,16 +62,23 @@ async fn greet() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static
 
     let response = client.say_hello(request).await;
 
+    let span = cx.span();
```
cijothomas, line 65: trying to get the span from context once, and re-use. Seeing the benchmarks for Context, I suspect asking for current span from Context repeatedly is unnecessarily burning CPU cycles.

examples/tracing-grpc/Cargo.toml
```diff
@@ -19,7 +19,7 @@ opentelemetry_sdk = { path = "../../opentelemetry-sdk", features = ["rt-tokio"]
 opentelemetry-stdout = { workspace = true, features = ["trace"] }
 prost = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
-tonic = { workspace = true, features = ["server"] }
+tonic = { workspace = true, features = ["server", "codegen", "channel", "prost"] }
```
lalitb, line 22: Curious - Why isn't there CI build failure because of these missing feature dependencies?

cijothomas: Mystery for me! I can confirm that this example is indeed built from CI and is having no errors!.

lalitb (approved):

examples/tracing-grpc/Cargo.toml
dzmitry-lahoda: do you really need "server" in release dependencies here?

dzmitry-lahoda: or may be move examples to other repo? so that examples do not need "server" which leads to "axum" which fails usage of tonic without "axum" (so tonic also not so great as it has no separate axum feature)

cijothomas: Only examples pull in server feature and hence Axum, right? Since it is not part of published crates, is that a concern?

dzmitry-lahoda: i guess you are right.

PR #2712, opened 2025-02-26
test: comment out intermittent failing assertion

scottgerring:
#2706 turns benchmarks on in the CI build. The bench in this PR fails intermittently. This comments it out to keep the build going smoothly so we can fix it async.

## Changes

### DIFF
--- a/opentelemetry-sdk/benches/metric.rs
+++ b/opentelemetry-sdk/benches/metric.rs
@@ -345,7 +345,8 @@ fn benchmark_collect_histogram(b: &mut Bencher, n: usize) {
 
     b.iter(|| {
         let _ = r.collect(&mut rm);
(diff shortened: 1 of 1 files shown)


scottgerring:
@cijothomas the ongoing error here, _I believe_, is when it regresses the performance against `main`. The panic is on :348, which is exactly the assert that has been removed here. 

So, getting this into `main` should fix for subsequent PRs running on top of main.

scottgerring:
@cijothomas , this removes the metric bench from the cargo configuration, so we can fix it async. I've validated it does not run locally.

The continued failure of the test here appears to be the running of the (still existent) bench against `main`, to form the baseline. Merging this down should fix this issue for subsequent PRs, as the bench will then be removed from main and not run.

PR #2719, opened 2025-02-26
test: Try fix a flaky test

cijothomas:
Occasionally seeing this failing in CI. I suspect it maybe capturing OTel's own logs. A quick fix is to filter for Error and above, so OTel Logs are unlikely to be captured. A more robust fix might be needed in the future, but this should unblock hopefully!

> ---- layer::tests::tracing_appender_inside_tracing_crate_context stdout ----
> 
> thread 'layer::tests::tracing_appender_inside_tracing_crate_context' panicked at opentelemetry-appender-tracing/src/layer.rs:550:9:
> assertion `left == right` failed
>   left: 3
>  right: 2
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -524,7 +524,7 @@ mod tests {
             .build();
         let tracer = tracer_provider.tracer("test-tracer");
 
-        let level_filter = tracing_subscriber::filter::LevelFilter::INFO;
+        let level_filter = tracing_subscriber::filter::LevelFilter::ERROR;
         let log_layer =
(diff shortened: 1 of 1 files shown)


lalitb:
> I suspect it maybe capturing OTel's own logs. 

Yes, I also suspected that. But I don't understand why it is not consistent.

lalitb (approved):

cijothomas:
> > I suspect it maybe capturing OTel's own logs.
> 
> Yes, I also suspected that. But I don't understand why it is not consistent.

next time it fails, we'll have the LogRecord in console logs. That should give some hint. (or this fix works and we never see failure again 🤣 )

PR #2722, opened 2025-02-26
ci: Run perf tests on demand

cijothomas:
It's taking a lot of time (60 mins of near full CPU usage!) and burning unnecessary resources (paid by CNCF) to run performance tests on every commit. So making it run on demand only.


lalitb (approved):
Thanks, make sense.

PR #2724, opened 2025-02-27
test : Add test to confirm Context expected behavior

cijothomas:
To assist reviewing https://github.com/open-telemetry/opentelemetry-rust/pull/2378/files I wrote some tests, which left me wondering with a unresolved question.

Context MUST be immutable, so adding value to context (with_value), should result in a *new* Context, leaving original untouched. A test for this is added and works as expected.
Then I tried this with Spans - Created an empty context, and added a Span to the Context. When I add another random value to that Context, it is cloning the random value, but the Span is stored via Arc, so clone is just pointing to the original span. Updates made to span from the new context updates the original span.
Not sure if this is the intended behavior, opening to discuss.

Edit: The expectation is Context itself is immutable, but the Span inside is mutable. Tests modified to assert this.

### DIFF
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -77,7 +77,7 @@ mod tests {
     };
 
     #[test]
-    fn span_context_immutable() {
+    fn span_modification_via_context() {
         let exporter = InMemorySpanExporterBuilder::new().build();
         let provider = SdkTracerProvider::builder()
             .with_span_processor(SimpleSpanProcessor::new(exporter.clone()))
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -76,6 +76,49 @@ mod tests {
         Context, KeyValue,
     };
 
+    #[test]
+    fn span_context_immutable() {
```
cijothomas, line 80: https://github.com/open-telemetry/opentelemetry-rust/pull/1268 is where we stored Span separate from the HashMap as an Arc. @shaun-cox can you share your understanding? Is it intentional that Span was Arc wrapped?

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -76,6 +76,49 @@ mod tests {
...
+        assert!(cx.has_active_span() == false);
+
+        // modify the span in the context
+        let span_ref = cx_with_span.span();
+        span_ref.set_attribute(KeyValue::new("attribute1", "value1"));
+
+        // create a new context, which should not affect the original
+        let cx_with_span_and_more = cx_with_span.with_value(ValueA(1));
+
+        // modify the span in the new context, which should not affect the original
```
cijothomas, line 108: but it affects the original, as the new Context is Arc pointer to the original Span itself...

opentelemetry-sdk/src/trace/mod.rs
bantonsson: @cijothomas that change does not change anything with regards to the `Span` being mutable. Before it was stored in the `HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>`, still an `Arc` pointing to the same `SynchronizedSpan`, and then it was moved to a separate field for performance reasons.

opentelemetry-sdk/src/trace/mod.rs
bantonsson: Isn't this how it's supposed to work? I don't understand what you think is wrong here. If you do something to a `Span` that should be reflected on the `Span` right? The `Span` API has explicitly mutating operations like `set_attribute`, and they should change the `Span`.

The `Context` being **immutable** doesn't mean that all things in the `Context` are **immutable**.

scottgerring:
I used this as an excuse to have a read of the spec 😄 My read is that the `Span` can and is designed to mutate internally without the surrounding `Context` caring. 

**A [Context is immutable](https://opentelemetry.io/docs/specs/otel/context/#overview)**: `Set Value` operation 
> The API MUST return a new Context containing the new value.

Also, **A [SpanContext is immutable](https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext)**, but this doesn't matter much for the present discussion: 
> SpanContexts are immutable.

Whereas, **[Spans are very mutable](https://opentelemetry.io/docs/specs/otel/trace/api/#set-attributes)** - "Set Attributes", "Add Events", etc. 

None of these operations nor the section [Context Interaction](https://opentelemetry.io/docs/specs/otel/trace/api/#context-interaction) mention anything about span mutation forcing a corresponding update/replace on the context. This would be in principal quite strange 2-way coupling - the `Span` having to tell the thing that happens to contain it that it had changed.

opentelemetry-sdk/src/trace/mod.rs
cijothomas: oh yes, that is correct!

opentelemetry-sdk/src/trace/mod.rs
cijothomas: Yes I expect Span modification to take effect on the Span, but the Context immutability part confused me. I'll modify the test so that it validates the expected behavior.

cijothomas:
@bantonsson @scottgerring Thanks for clearing up the confusion I had. I modified the tests now to test the expected behavior!

lalitb (approved):

Issue #2717, opened 2025-02-26
Baggage - Stabilization


cijothomas:
Opening a parent issue to track stabilizing `Baggage`. The spec for this is already stable, so it is just a matter of ensuring the APIs we have is good, enough tests and docs.

1. Fix compliance issue - https://github.com/open-telemetry/opentelemetry-rust/issues/2716
2. Add benchmarks - there is none today
3. Add test coverage and investigate all open issues/bugs. (eg: https://github.com/open-telemetry/opentelemetry-rust/discussions/2631)
4. Improve docs - make sure the docs are up-to-date.
5. Example.- Modify existing examples to show Baggage usage. A common misconception in OTel is many people assume Baggage attributes automatically becomes attributes in Logs/Traces etc. But that require explicit action. Examples should showcase that.

gruebel:
Related to 6:

I'm not sure how to interpret the specs in relation to the Rust implementation. Both `set` and `remove` should actually return the `Baggage` itself, but we are currently mimicking the HashMap behaviour instead. It is definitely valid and feels more Rust idiomatic. Should we change both to not return anything instead or keep it as is?

cijothomas:
There is something about Baggage should be immutable, so adding a new entry should result in a new Baggage?

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/baggage.rs#L322-L325
With Baggage should be just taking `Baggage` as input? Right now it takes `KeyValueMetadata` which is also hence forced to be a public API.

gruebel:
I think you are right with both. Adding/Removing an entry should return a copy of the baggage. The `Context` interaction should be with the `Baggage` instance and not by passing `KeyValueMetadata`. I checked the JS and Go implementation (I also checked Python, but it works directly with the `Context`) and they both work on an immutable object and return a copy. And they pass the `Baggage` instance to the `Context`.

My suggestion, we create a new `BaggageBuilder` and deprecate the direct manipulation of `Baggage`. This is also allowed and fits quite well into other areas, where we use a builder pattern.
For the Context, we either make it a breaking change or we create a new trait `BaggageContextExt` and mark the old one as deprecated. thoughts?

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/baggage.rs#L159
[...]

PR #2734, opened 2025-03-01
fix: align baggage.remove() implementation

gruebel:
Relates to #2717 point 6

## Changes

- also added a simple CRUD test to make sure, all the operations work as intended (also the conflict override)

### DIFF
--- a/opentelemetry/src/baggage.rs
+++ b/opentelemetry/src/baggage.rs
@@ -622,11 +622,11 @@ mod tests {
         assert_eq!(baggage.len(), 1);
 
         // get
-        assert_eq!(baggage.get("foo"), Some(&Value::from("1")));
+        assert_eq!(baggage.get("foo"), Some(&StringValue::from("1")));
 
         // update
         baggage.insert("foo", "2");
-        assert_eq!(baggage.get("foo"), Some(&Value::from("2")));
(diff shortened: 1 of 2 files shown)


opentelemetry/src/baggage.rs
```diff
@@ -161,8 +161,8 @@ impl Baggage {
 
     /// Removes a name from the baggage, returning the value
     /// corresponding to the name if the pair was previously in the map.
-    pub fn remove<K: Into<Key>>(&mut self, key: K) -> Option<(Value, BaggageMetadata)> {
-        self.inner.remove(&key.into())
+    pub fn remove<K: AsRef<str>>(&mut self, key: K) -> Option<(Value, BaggageMetadata)> {
```
cijothomas, line 164: would this be considered user-affecting? Lets include this in the changelog, along with the other Baggage changes.

cijothomas (approved):
Thanks! 
(Merged other PR so need to resolve conflict before merge.)

opentelemetry/src/baggage.rs
gruebel: Added a changelog entry, but I'm not sure, if it is really a breaking change. Better be safe than sorry 😅

PR #2736, opened 2025-03-02
test: Add  Scope attributes test cases for metrics

### LINKED ISSUE #2702: Extend test coverage for Scope attributes
In particular, metrics, as Scope is part of identity of the metric.

Shunpoco:
Fixes #2702 

## Changes

Add additional unit test to cover Scope attributes for metrics.
I don't add any test cases for Logger/Tracer because their with_scope() method creates new Logger/Tracer regardless of Scope attribute's identity (Will we change the methods in the future?).


cijothomas:
> I don't add any test cases for Logger/Tracer because their with_scope() method creates new Logger/Tracer regardless of Scope attribute's identity (Will we change the methods in the future?).

Not likely to change as it is not required. For Metrics this is important to get pre-aggregation correctly

cijothomas (approved):
Thanks!

PR #2737, opened 2025-03-02
test: add benchmark tests for baggage

gruebel:
Relates to #2717 point 2

## Changes

- I was not sure about the `MAX_KEY_VALUE_PAIRS` const, it is currently private so I just duplicated it for the benchmark test. let me know, if you want to make it public or not.

### DIFF
--- a/opentelemetry/Cargo.toml
+++ b/opentelemetry/Cargo.toml
@@ -62,6 +62,10 @@ name = "context_attach"
 harness = false
 required-features = ["tracing"]
 
+[[bench]]
+name = "baggage"
+harness = false
+
 [lib]
 bench = false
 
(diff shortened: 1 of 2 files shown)


gruebel:
@cijothomas could you add the `performance` label to validate the benchmark works in CI as expected.

cijothomas (approved):
LGTM.
Let us also add some measuring the cost of adding Baggage to Context. It might be required to make Baggage a top-level entry in Context instead of storing it inside Context's hash map to gain perf.

cijothomas:
> @cijothomas could you add the `performance` label to validate the benchmark works in CI as expected.

oops sorry missed this. We can run targeted benches locally for now. The CI runs entire set, which is quite expensive, and not sure if that is something CNCF encourages us to do...

PR #2741, opened 2025-03-03
perf: Add benchmarks to show the cost of attributes in Log

cijothomas:
Just benches to later measure improvements in the area.

### DIFF
--- a/opentelemetry-appender-tracing/benches/log-attributes.rs
+++ b/opentelemetry-appender-tracing/benches/log-attributes.rs
@@ -1,0 +1,271 @@
+/*
+// Run this benchmark with:
+ // cargo bench --bench log-attributes
+ // Adding results in comments for a quick reference.
+ // Apple M4 Pro
+ //     Total Number of Cores:	14 (10 performance and 4 efficiency)
+
+| Test                 | Average time | Increment |
+|----------------------|--------------|-----------|
+| otel_0_attributes    | 72 ns        | -         |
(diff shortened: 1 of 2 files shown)


opentelemetry-appender-tracing/benches/log-attributes.rs
```diff
@@ -0,0 +1,271 @@
...
+                            field8 = "field8",
+                            field9 = "field9",
+                            field10 = "field10",
+                            message = "Unable to process checkout."
+                        );
+                    }
+                }
+            });
+        });
+    });
```
lalitb, line 247: nit - courtesy AI:
```rust
let mut fields = vec![("name", "CheckoutFailed"), ("message", "Unable to process checkout.")];

for i in 1..=num_attributes {
    fields.push((format!("field{}", i).as_str(), format!("field{}", i).as_str()));
}

error!(fields.iter().map(|(k, v)| (*k, *v)).collect::<Vec<_>>());
```

lalitb (approved):
LTTM. Thanks for adding this :)

opentelemetry-appender-tracing/benches/log-attributes.rs
cijothomas: I don;t think that is doing what we want it to do. 

We could potentially create a macro to achieve it, but the current one (though involves copy-paste) is quite easy to read and reason about.

PR #2748, opened 2025-03-03
refactor: refactor Baggage with Context interaction

gruebel:
Relates to #2717 point 6

## Changes

- `with_baggage()` and `current_with_baggage()` now only accept a `Baggage` instance and override any existing `Baggage` in the `Context`
- added a private `BaggageContextValue` to hide the direct retrieval of `Baggage` from `Context` to make sure it needs to accessed via `context.baggage()`

### DIFF
--- a/opentelemetry-sdk/src/propagation/baggage.rs
+++ b/opentelemetry-sdk/src/propagation/baggage.rs
@@ -1,5 +1,5 @@
 use opentelemetry::{
-    baggage::{BaggageExt, KeyValueMetadata},
+    baggage::{Baggage, BaggageExt, KeyValueMetadata},
     otel_warn,
     propagation::{text_map_propagator::FieldIter, Extractor, Injector, TextMapPropagator},
     Context,
@@ -30,7 +30,7 @@ fn baggage_fields() -> &'static [String; 1] {
 /// # Examples
 ///
 /// ```
-/// use opentelemetry::{baggage::BaggageExt, KeyValue, propagation::TextMapPropagator};
+/// use opentelemetry::{baggage::{Baggage, BaggageExt}, propagation::TextMapPropagator};
 /// use opentelemetry_sdk::propagation::BaggagePropagator;
 /// use std::collections::HashMap;
 ///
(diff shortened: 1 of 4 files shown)


opentelemetry/src/baggage.rs
```diff
@@ -308,39 +308,40 @@ pub trait BaggageExt {
...
+    /// let _ = baggage.insert("my-name", "my-value");
+    ///
+    /// let cx = Context::current_with_baggage(baggage);
     ///
     /// assert_eq!(
     ///     cx.baggage().get("my-name"),
     ///     Some(&StringValue::from("my-value")),
     /// )
     /// ```
-    fn current_with_baggage<T: IntoIterator<Item = I>, I: Into<KeyValueMetadata>>(
```
cijothomas, line 341: KeyValueMetadata can be removed from public API now, can we?

opentelemetry-sdk/src/propagation/baggage.rs
```diff
@@ -48,14 +48,17 @@ fn baggage_fields() -> &'static [String; 1] {
 /// }
 ///
 /// // Add new baggage
-/// let cx_with_additions = cx.with_baggage(vec![KeyValue::new("server_id", 42)]);
+/// let mut baggage = Baggage::new();
+/// let _ = baggage.insert("server_id", "42");
+///
+/// let cx_with_additions = cx.with_baggage(baggage);
```
cijothomas, line 54: can you think any Rust magic like Into impl for a slice/vec of KeyValue for Baggage, so users can write
cx.with_baggage([KeyValue::new("f","b"),...]) or slice or vec...

gruebel: You are still able to do `cx.with_baggage(Baggage::from_iter([KeyValue::new("f","b"),...]))` which is basically the old behaviour.

opentelemetry/src/baggage.rs
gruebel: I probably would still keep it, because you can use in combination with `Baggage::from_iter(...)`.

opentelemetry-sdk/src/propagation/baggage.rs
TommyCpp: Can't we just take `Into<Baggage>` as parameter and then implement `Into<Baggage>` for `Vec<KeyValue>`?

opentelemetry/src/baggage.rs
cijothomas: Unless it is really required, we should remove it to keep API surface as slim as possible. But lets tackle it separately from this PR.

opentelemetry/src/baggage.rs
```diff
@@ -312,8 +322,17 @@ pub trait BaggageExt {
...
+    /// let mut baggage = Baggage::new();
+    /// let _ = baggage.insert("my-name", "my-value");
+    ///
+    /// let cx = Context::map_current(|cx| {
+    ///     cx.with_baggage(baggage)
+    /// });
     ///
+    /// // Passing an iterator
     /// let cx = Context::map_current(|cx| {
     ///     cx.with_baggage(vec![KeyValue::new("my-name", "my-value")])
```
cijothomas, line 337: nit: an array should be good enough, so no need for showing vec! in the doc.

cijothomas (approved):

TommyCpp (approved):

opentelemetry-sdk/src/propagation/baggage.rs
lalitb: > I think I found a workable solution.

Does it mean we should now change this example to use:
```rust
let cx_with_additions = cx.with_baggage([("server_id", "42")]);
```

opentelemetry/src/baggage.rs
gruebel: I'll check, if it is still needed or not in a follow-up

opentelemetry-sdk/src/propagation/baggage.rs
gruebel: @lalitb that's not possible for now. it could either be an iterator with `KeyValue` or `KeyValueMetadata` entries. Do we want to support this here, then we probably need to implement an other trait or so.

PR #2735, opened 2025-03-01
fix: Fix Logger scope in Tracing Appender

cijothomas:
Fixes #2658

PR #2756, opened 2025-03-04
feat: Allow event name to be provided to IsEnabled check in Logger

cijothomas:
Followup to https://github.com/open-telemetry/opentelemetry-rust/pull/2735, this PR adds an additional parameter (event name) to be passed to the `event_enabled` method, allowing consumer to leverage this additional information.
EventName is part of `tracing` metadata, so there is no extra cost to obtain it.

Modified SdkLogger to pass the name received to the processor, instead of passing scope.name. (Scope.name was empty due to 2735, so it no longer makes sense to pass it. Even without 2735, it was not useful, as scope-name was just the hardcoded crate name of the appender, not particularly useful!)

(Note: EventEnabled is still under experimental feature flag.)

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -55,12 +55,11 @@
   when its `shutdown` is invoked.
 
 - Reduced some info level logs to debug
-- **Breaking** for custom LogProcessor/Exporter authors:
-  Added an additional `name: Option<&str>` parameter to the `event_enabled`
-  method on the `LogProcessor` and `LogExporter` traits. This allows
-  implementations to leverage this additional information to determine if an
-  event is enabled. `SdkLogger` no longer passes its `scope` name but instead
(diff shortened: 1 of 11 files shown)


lalitb (approved):

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -52,6 +55,12 @@
   when its `shutdown` is invoked.
 
 - Reduced some info level logs to debug
+- **Breaking** for custom LogProcessor/Exporter authors:
+  Added an additional `name: Option<&str>` parameter to the `event_enabled`
```
utpilla, line 59: This isn't an addition of a new parameter. It's a change in the type of the existing parameter.

cijothomas: oh yes that is true. I'll rephrase the wording.

utpilla (approved):

Issue #1983, opened 2024-08-05
[Bug]:  OTEL_EXPORTER_OTLP_PROTOCOL env var is not supported eyt


3g0r:
### What happened?

Application with latest opentelemetry rust doesn't accept env var OTEL_EXPORTER_OTLP_PROTOCOL after [this commit]( https://github.com/open-telemetry/opentelemetry-rust/commit/f66e9626cb345f2e13558a40076a58841c2286ae#diff-13bdfa4891026a23503ebdbb92913eecb6984acb56eda1979037d919e0652fd6). Now protocol selection based on `fn default_protocol` or application code

### API Version

otlp 1.3.2

### SDK Version

after 0.14.0 otel otlp

lalitb:
@3g0r This looks to be valid issue. The protocol is selected based on the feature flag, or the ExportConfig provided by the application. Env variable is not considered. Would you like to contribute the fix?

TommyCpp:
Emmm good point. We removed it because back then all exporter only supports one protocol(http exporter only had protobuf back then). Now they can be either `json` or `protobuf`

tomDev5:
any updates on this?
I could contribute a fix on it as well if needed

PR #2758, opened 2025-03-05
feat: implement `FromStr` and `Display` for `Protocol`

davidhewitt:
## Changes

Implements `Display` and `FromStr` for `Protocol`, using the values documented in https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -221,6 +221,9 @@ mod metric;
 #[cfg(any(feature = "http-proto", feature = "http-json", feature = "grpc-tonic"))]
 mod span;
 
+use std::fmt::Display;
+use std::str::FromStr;
+
 pub use crate::exporter::Compression;
 pub use crate::exporter::ExportConfig;
 #[cfg(feature = "trace")]
@@ -257,6 +260,9 @@ pub use crate::exporter::{
     OTEL_EXPORTER_OTLP_TIMEOUT_DEFAULT,
 };
 
+use exporter::OTEL_EXPORTER_OTLP_PROTOCOL_GRPC;
+use exporter::OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON;
+use exporter::OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF;
 use opentelemetry_sdk::ExportError;
 
 /// Type to indicate the builder does not have a client set.
@@ -352,6 +358,10 @@ pub enum Error {
     #[cfg(any(not(feature = "gzip-tonic"), not(feature = "zstd-tonic")))]
     #[error("feature '{0}' is required to use the compression algorithm '{1}'")]
     FeatureRequiredForCompressionAlgorithm(&'static str, Compression),
+
+    /// Unsupported protocol.
+    #[error("unsupported protocol '{0}'")]
+    UnsupportedProtocol(String),
 }
 
 #[cfg(feature = "grpc-tonic")]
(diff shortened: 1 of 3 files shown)


davidhewitt:
Possibly interesting to note that the `Serialize` and `Deserialize` values are automatically derived and will not match these.

(Perhaps the serde implementations should be changed to use these constants? They seem to be universal across all OTEL SDKs. I am happy to do that as a follow-up.)

gruebel (commented):
nice work 🍻 can you also add a short changelog entry for it? Regarding the serde part,  it would be great to be consistent with other SDKs.

opentelemetry-otlp/src/lib.rs
```diff
@@ -396,7 +406,53 @@ pub enum Protocol {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Protocol::Grpc => write!(f, "{}", OTEL_EXPORTER_OTLP_PROTOCOL_GRPC),
+            Protocol::HttpBinary => write!(f, "{}", OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF),
+            Protocol::HttpJson => write!(f, "{}", OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON),
+        }
+    }
+}
+
+impl FromStr for Protocol {
```
utpilla, line 419: Is this necessary? Why would a user have this awkward construct:

```rust
.with_protocol("grpc".parse::<Protocol>().unwrap())
```

when they could simply do this:

```rust
.with_protocol(Protocol::Grpc)
```

opentelemetry-otlp/src/lib.rs
```diff
@@ -396,7 +406,53 @@ pub enum Protocol {
     HttpJson,
 }
 
+impl Display for Protocol {
```
utpilla, line 409: What's the motivation behind adding a `Display` implementation for `Protocol`?

opentelemetry-otlp/src/lib.rs
davidhewitt: Completely agree if hard coded then users should use the enum variant.

I am working on [a downstream SDK](https://github.com/pydantic/logfire-rust) which uses `opentelemetry`, `tracing` and `log`. We configure an otel exporter and are working to support setting the [protocol via environment](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol), which is [not yet supported by this library](https://github.com/open-telemetry/opentelemetry-rust/issues/1983) (although in future perhaps I can also contribute a solution to that issue).

Other downstream packages to set up an exporter, like `init-tracing-opentelemetry`, also have [similar code doing this conversion](https://github.com/davidB/tracing-opentelemetry-instrumentation-sdk/blob/5f787b3a2aaa1da1c0ad34365d8e2a501cdb5380/init-tracing-opentelemetry/src/otlp.rs#L24-L42).

It feels most correct to have the `FromStr` implementation here in the library for all downstream implementations to re-use. This is also [how the `Compression` type is already parsed](https://github.com/open-telemetry/opentelemetry-rust/blob/b33f0cc56cc8ec75c66439c7bfa7be7d435af60f/opentelemetry-otlp/src/exporter/tonic/mod.rs#L251-L252) from env within this library.

opentelemetry-otlp/src/lib.rs
davidhewitt: After seeing the `Compression` implementation within this library had both `FromStr` and `Display`, I felt it was most correct to add `Display` to match the `FromStr` here too.

If you have concerns, I am happy to remove it - the `FromStr` is the functionality I need at present (as per above comment).

davidhewitt:
> Regarding the serde part, it would be great to be consistent with other SDKs.

I opened #2765 for this as a separate PR, because it's potentially breaking and deserves its own visibility / changelog note.

PR #2760, opened 2025-03-05
test: Stress test for logs modified to show throughput when enabled()…

cijothomas:
Simple commented out code to enable/disable. (In future, we can make it customizable via cmd line etc.)

### DIFF
--- a/stress/src/logs.rs
+++ b/stress/src/logs.rs
@@ -7,6 +7,11 @@
 
     Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
     ~40 M /sec
+
+    Hardware: Apple M4 Pro
+    Total Number of Cores:	14 (10 performance and 4 efficiency)
+    ~40 M/sec
+    ~1.1 B/sec (when disabled)
 */
 
 use opentelemetry::InstrumentationScope;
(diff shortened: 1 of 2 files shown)


utpilla (approved):

stress/src/logs.rs
```diff
@@ -7,6 +7,11 @@
 
     Hardware: AMD EPYC 7763 64-Core Processor - 2.44 GHz, 16vCPUs,
     ~40 M /sec
+
+    Hardware: Apple M4 Pro
+    Total Number of Cores:	14 (10 performance and 4 efficiency)
+    ~40 M/sec
+    ~1.1 B/sec (when disabled)
```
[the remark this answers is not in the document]
utpilla: Could we update the `error!` statement to log integers instead of strings and compare the results? That would tell us if it's indeed the string allocation that's causing the perf hit.

stress/src/logs.rs
```diff
@@ -51,13 +57,26 @@ impl LogProcessor for MockLogProcessor {
...
+        _target: &str,
+        _name: Option<&str>,
+    ) -> bool {
+        self.enabled
+    }
 }
 
 fn main() {
+    // change this to false to test the throughput when enabled is false.
+    let enabled = true;
```
lalitb, line 73: nit - let it pass as command line argument.

lalitb (approved):

stress/src/logs.rs
cijothomas: yes I mentioned that in the description too.

PR #2762, opened 2025-03-05
test: Modify perf tests to pass event name correctly to avoid string allocation

cijothomas:
Follow up to https://github.com/open-telemetry/opentelemetry-rust/pull/2760

### DIFF
--- a/opentelemetry-appender-tracing/benches/logs.rs
+++ b/opentelemetry-appender-tracing/benches/logs.rs
@@ -10,7 +10,7 @@
     | noop_layer_disabled         | 12 ns       |
     | noop_layer_enabled          | 25 ns       |
     | ot_layer_disabled           | 19 ns       |
-    | ot_layer_enabled            | 196 ns      |
+    | ot_layer_enabled            | 167 ns      |
 */
 
 use criterion::{criterion_group, criterion_main, Criterion};
@@ -94,7 +94,7 @@ fn benchmark_no_subscriber(c: &mut Criterion) {
     c.bench_function("log_no_subscriber", |b| {
         b.iter(|| {
             error!(
-                name = "CheckoutFailed",
+                name : "CheckoutFailed",
                 book_id = "12345",
                 book_title = "Rust Programming Adventures",
                 message = "Unable to process checkout."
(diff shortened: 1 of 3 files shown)


lalitb (approved):

stress/src/logs.rs
```diff
@@ -88,7 +88,7 @@ fn main() {
 
 fn test_log() {
     error!(
-        name = "CheckoutFailed",
+        name : "CheckoutFailed",
```
cijothomas, line 91: in case anyone wondering what this is doing - previous "name" was just any attribute, so the string has to copied to new heap memory.
now, name is part of metadata of `tracing`, and is `static. Otel stores it as event_name, which is `static str, so allocation and copying is gone.

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -137,10 +147,10 @@ fn benchmark_with_noop_layer(c: &mut Criterion, enabled: bool, bench_name: &str)
         c.bench_function(bench_name, |b| {
             b.iter(|| {
                 error!(
-                    name = "CheckoutFailed",
+                    name : "CheckoutFailed",
                     book_id = "12345",
                     book_title = "Rust Programming Adventures",
-                    "Unable to process checkout."
+                    message = "Unable to process checkout."
```
utpilla, line 153: Why have we removed `message` here?

opentelemetry-appender-tracing/benches/logs.rs
```diff
@@ -10,7 +10,17 @@
     | noop_layer_disabled         | 12 ns       |
     | noop_layer_enabled          | 25 ns       |
     | ot_layer_disabled           | 19 ns       |
-    | ot_layer_enabled            | 196 ns      |
+    | ot_layer_enabled            | 167 ns      |
+
+    Hardware: Apple M4 Pro
+    Total Number of Cores:	14 (10 performance and 4 efficiency)
```
utpilla, line 16: It'd be great if you could also add the CPU speed as part of the details.

cijothomas: I need to find that detail. @lalitb do you know?

opentelemetry-appender-tracing/benches/logs.rs
cijothomas: we added message, to make it consistent with rest of benchmarks. 
(message is preferred as it allows one less dyn dispatch: https://github.com/open-telemetry/opentelemetry-rust/pull/2689/files)

PR #2763, opened 2025-03-05
refactor: make baggage::KeyValueMetadata private

gruebel:
Relates to https://github.com/open-telemetry/opentelemetry-rust/pull/2748#discussion_r1978292523

## Changes

@cijothomas let me know, if you had this in mind. if it is too inconvenient to use 2/3 value tuples, then we can also keep the current implementation

### DIFF
--- a/opentelemetry-sdk/src/propagation/baggage.rs
+++ b/opentelemetry-sdk/src/propagation/baggage.rs
@@ -1,8 +1,8 @@
 use opentelemetry::{
-    baggage::{BaggageExt, KeyValueMetadata},
+    baggage::{BaggageExt, BaggageMetadata},
     otel_warn,
     propagation::{text_map_propagator::FieldIter, Extractor, Injector, TextMapPropagator},
-    Context,
+    Context, Key, StringValue,
 };
 use percent_encoding::{percent_decode_str, utf8_percent_encode, AsciiSet, CONTROLS};
 use std::iter;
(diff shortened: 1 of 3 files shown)


cijothomas:
Can you share a code snippet of before/after that users have to do due to this change? That'll make it easy for us to make a decision. (Also, after looking at KeyValueMetadata struct, it not a concern to keep it public either.)

gruebel:
Instead of creating it directly, like
```rust
KeyValueMetadata::new("key1", "val1", "prop1")
```
you have to pass it as a tuple
```rust
(KeyValue::new("key1", "val1"), BaggageMetadata::from("prop1"))
// or
(Key::from("key1"), StringValue::from("val1"), BaggageMetadata::from("prop1"))
```

as said, if keeping `KeyValueMetadata` public is ok, then there is no real benefit to go with this approach

cijothomas:
> Instead of creating it directly, like
> 
> ```rust
> KeyValueMetadata::new("key1", "val1", "prop1")
> ```
> 
> you have to pass it as a tuple
> 
> ```rust
> (KeyValue::new("key1", "val1"), BaggageMetadata::from("prop1"))
> // or
> (Key::from("key1"), StringValue::from("val1"), BaggageMetadata::from("prop1"))
> ```
> 
> as said, if keeping `KeyValueMetadata` public is ok, then there is no real benefit to go with this approach

Agree to keep the existing and keep `KeyValueMetadata` public. Let us close this PR.
(Thanks for exploring and writing up this!)

PR #2764, opened 2025-03-05
fix: Remove mut ref requirement for shutdown LogExporter

cijothomas:
Redoing https://github.com/open-telemetry/opentelemetry-rust/pull/2634/files

`LogExporter` shutdown() no longer require `mut ref`. The component that needs mutability should rely on interior mutability. Most of the exporters already use interior mutability.

Metric already does not require mutable self on shutdown, this makes Logs consistent with that, and later we need to get Traces too.

Also LogProcessor, LogProvider does not require `mut ref` on shutdown, so this makes exporter aligned as well.

Additionally this also fixes https://github.com/open-telemetry/opentelemetry-rust/issues/2770


opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -7,21 +7,20 @@ use opentelemetry_sdk::error::{OTelSdkError, OTelSdkResult};
 use opentelemetry_sdk::logs::{LogBatch, LogExporter};
 use tonic::{codegen::CompressionEncoding, service::Interceptor, transport::Channel, Request};
 
-use opentelemetry_proto::transform::logs::tonic::group_logs_by_resource_and_scope;
-
 use super::BoxInterceptor;
-use tokio::sync::Mutex;
+use opentelemetry_proto::transform::logs::tonic::group_logs_by_resource_and_scope;
+use std::sync::Mutex;
```
utpilla, line 12: Shouldn't we be using tokio's `Mutex` as we are locking across an await point when calling `client.export()`? The std library's `Mutex` is not async aware so it might lead to a deadlock.

opentelemetry-sdk/src/logs/export.rs
```diff
@@ -136,7 +136,7 @@ pub trait LogExporter: Send + Sync + Debug {
     ) -> impl std::future::Future<Output = OTelSdkResult> + Send;
 
     /// Shuts down the exporter.
-    fn shutdown(&mut self) -> OTelSdkResult {
+    fn shutdown(&self) -> OTelSdkResult {
```
utpilla, line 139: Please add a CHANGELOG entry for this.

opentelemetry-otlp/src/exporter/tonic/logs.rs
cijothomas: Thanks for catching that. Yes, it is risky to use std::sync::Mutex here. In one of early attempts, compiler did warn about this and I switched to tokio::sync::Mutex, but that makes shutdown more complex. So I reverted to `std::sync::Mutex` (compiler won't catch these always it seems) because that is exactly what Metric client is doing. And it looks like Metrics have this deadlock risk.

Some options.
1. Use tokio::Mutex, and in shutdown(), try and find tokio::current_runtime, and if not found, create own. Then use that to acquire tokio::Mutex. I made a commit to do this in this PR. PTAL.
2. Ignore the issue in this PR - which is basically expanding an existing issue from Metrics to Logs too. And find a common solution and apply across. (We'll need the same for Traces too)
3. Something else

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -86,8 +84,12 @@ impl LogExporter for TonicLogsClient {
         Ok(())
     }
 
-    fn shutdown(&mut self) -> OTelSdkResult {
-        match self.inner.take() {
+    fn shutdown(&self) -> OTelSdkResult {
+        let handle = tokio::runtime::Handle::try_current()
```
cijothomas, line 88: this is not so clean. We already require users to create the provider from inside a tokio runtime. Maybe we can extend that guideline for shutdown also?

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp/README.md

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -5,23 +5,23 @@ use opentelemetry_proto::tonic::collector::logs::v1::{
 };
 use opentelemetry_sdk::error::{OTelSdkError, OTelSdkResult};
 use opentelemetry_sdk::logs::{LogBatch, LogExporter};
+use tokio::sync::Mutex;
```
gruebel, line 8: how about importing `use std::sync::Mutex;` instead, then you can basically use the same code as in `metrics` for `shutdown()` and `export()` and it should work at least for me it does 😅

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/pull/2764#discussion_r1985994302

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -86,8 +84,20 @@ impl LogExporter for TonicLogsClient {
...
+            Ok(handle) => handle,
+            Err(e) => {
+                return Err(OTelSdkError::InternalFailure(format!(
+                    "Shutdown must be called from a tokio runtime: {:?}",
+                    e
+                )));
+            }
+        };
+
+        let mut inner = handle.block_on(self.inner.lock());
```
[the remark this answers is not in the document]
utpilla: We could spawn a new thread to acquire the lock on `inner`. This should work regardless of the runtime used. Since, shutdown only happens once, the cost of spawning a thread should be acceptable.

```rust
    otel_debug!(name: "TonicsLogsClient.Shutdown");

    let inner = std::thread::scope(|s| {
        s.spawn(|| futures_executor::block_on(self.inner.lock()))
            .join() // We can enforce some kind of timeout here if needed.
    });

    match inner {
        Ok(mut inner) => {
            match inner.take() {
                Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
                None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
            }
        }
        Err(_) => Err(OTelSdkError::InternalFailure(
            "Failed to shutdown".to_string(),
        )),
    }
}
```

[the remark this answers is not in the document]
cijothomas: Yes spawning own thread would be a solution. For this PR, I am not doing tonic shutdown. This won't be a regression for tonic anyway, as the shutdown was not invoked at all.
I opened https://github.com/open-telemetry/opentelemetry-rust/issues/2777 to track adding proper shutdown to grpc/tonic, to unblock this PR.

opentelemetry-otlp/src/exporter/tonic/logs.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2777 Will use this for further discussion. I am unblocking this PR by removing tonic's shutdown logic. (it was never called anyway!)

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -86,11 +84,15 @@ impl LogExporter for TonicLogsClient {
...
 
-    fn shutdown(&mut self) -> OTelSdkResult {
-        match self.inner.take() {
-            Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
-            None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
-        }
+    fn shutdown(&self) -> OTelSdkResult {
+        // TODO: Implement actual shutdown
+        // Due to the use to tokio::sync::Mutex to guard
+        // the inner client, we need to lock the mutex
```
utpilla, line 90:
```suggestion
        // the inner client, we need to await the call to lock the mutex
```

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -60,6 +60,10 @@
   `LogProcessor` and `LogExporter` traits. `SdkLogger` no longer passes its
   `scope` name but instead passes the incoming `name` when invoking
   `event_enabled` on processors.
+- **Breaking** for custom LogExporter authors: `shutdown()` method in
+ `LogExporter` trait no longer requires a mutable ref to self. If the exporter
```
utpilla, line 64:
```suggestion
 `LogExporter` trait no longer requires a mutable ref to `self`. If the exporter
```

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -86,11 +84,15 @@ impl LogExporter for TonicLogsClient {
...
     }
 
-    fn shutdown(&mut self) -> OTelSdkResult {
-        match self.inner.take() {
-            Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
-            None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
-        }
+    fn shutdown(&self) -> OTelSdkResult {
+        // TODO: Implement actual shutdown
+        // Due to the use to tokio::sync::Mutex to guard
```
utpilla, line 89:
```suggestion
        // Due to the use of tokio::sync::Mutex to guard
```

utpilla (approved):
Looks good for now. The shutdown logic for tonic export can be implemented later.

cijothomas:
@gruebel Merged this to unblock progress! Please share any feedback on this, and I'll follow up.

PR #2765, opened 2025-03-06
refactor: change `Protocol`'s `Serialize` and `Deserialize` to use standard values

davidhewitt:
Follow up from #2758

## Changes

As noted in the above PR, the current implementation of `Serialize` and `Deserialize` for protocol will emit the enum name as it is read in source code. I propose it would be better to match [the standard values in the spec](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol), especially if #2758 is merged.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## vNext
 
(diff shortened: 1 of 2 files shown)


gruebel (approved):
@davidhewitt looks good. can you fix the merge conflicts?

PR #2301, opened 2024-11-16
fix: AtomicI64, AtomicU64 for build mipsel processor

TGLuong:
Fixes #
Design discussion issue (if applicable) #

## Changes
- Use portable-atomic for mipsel processor. Because when build for mipsel architecture, it don't have `AtomicI64`, `AtomicU64` in std.

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #2773, opened 2025-03-08
fix: AtomicI64, AtomicU64 for xtensa arch

brunobrolesi:
Fixes #

I'm working on a project that tries to add observability to an mqtt system with ESP32 microcontrollers. When I tried to use the open telemetry crate to export my program's traces, I encountered the following build error:

image

## Changes

After doing some analysis on the subject, I came across the following [PR](https://github.com/open-telemetry/opentelemetry-rust/pull/2301) that had the same problem, but for other architectures. Based on the discussion of the PR, I made the changes for the architecture I am using and was able to use the stdout exporter, as can be seen in the image below:

image

One thing to note, this change only allowed me to use the stout exporter, the others had problems due to a compilation error related to the dependencies hyper and socket 2, this problem can be seen in the issues below:

- [hyper](https://github.com/hyperium/hyper/issues/3684)
- [socket2](https://github.com/rust-lang/socket2/issues/516)


cijothomas:
@brunobrolesi Thanks! Can you check if this is only place where Atomics were used? It's okay to do it in a follow up, if you are okay with that. (Trying to make sure we do support extra architectures consistently across repo (wherever feasible))

hdost (approved):

hdost:
> @brunobrolesi Thanks! Can you check if this is only place where Atomics were used? It's okay to do it in a follow up, if you are okay with that. (Trying to make sure we do support extra architectures consistently across repo (wherever feasible))

https://sourcegraph.com/search?q=std%3A%3Async%3A%3Aatomic%20repo%3Aopentelemetry-rust&patternType=regex

PR #2776, opened 2025-03-10
chore: Move FutureExt into context 

### LINKED ISSUE #2754: Future extensions should be in common module, not under trace
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/trace/context.rs#L385-L475

These are extensions to help propagate `Context` for Futures. While tracing has a common need to propagate context across async functions, there are use-cases outside of tracing as well. (eg: Propagate Baggage or anything else in Context to a async method.)

This should be moved to common `Context` module, or re-exported.

scottgerring:
Fixes #2754

## Changes
* Switch `opentelemetry::context` from `context.rs` to `context/mod.rs` module style so that we can add `future_ext` without jamming everything together into the same file
* Move `FutureExt` from `opentelemetry::trace::context` into `opentelemetry::context::future_ext`
* Re-export `FutureExt` in trace to maintain backwards compatibility

I had to make a couple bits in `Context` `pub(crate)` to make this work.

I've switched to use the module style where the parent is in `src/context.rs` and the nested module is in `src/context/future_ext.rs`. I think this works nicely here, as the majority of the module is `Context` itself, and if we move that into `context/mod.rs` style, we'd need to either include `Context` in the mod.rs which is a bad practice, or invent some extra layer of indirection which doesn't fit. 


scottgerring:
@cijothomas is this what you had in mind?

opentelemetry/src/lib.rs
```diff
@@ -246,7 +246,7 @@ pub mod global;
 
 pub mod baggage;
 
-mod context;
+pub mod context;
```
cijothomas, line 249: is this needed?

cijothomas (approved):

bantonsson (approved):

Issue #2555, opened 2025-01-24
BatchSpanProcessor concurrent export support


cijothomas:
The original BSP did support this capability, but the new one (that spawns own thread) does not support this. Opening an issue to track demand for this feature and add this support to BatchSpanProcessor (and BatchLogProcessor. But Log one never had this capability to begin with.)

PR #2780, opened 2025-03-10
feat: Add experimental concurrent processor for logs

cijothomas:
1. Add a new processor under experimental feature flag. 
2. This is to facilitate discussions in OTel spec about adding such a component to the official spec itself.
3. This is *NOT* useful as-is with any existing exporters in this repo (like OTLP Log Exporter
4. This is an initial version. Next, I'll introduce ways to ensure that only those exporters that can support concurrent exporting will be paired with ConcurrentExporter. This is one possible way spec could evolve. Alternatively, existing processors can be modified to export concurrently if the exporter advertises that they are okay with that!. BatchLogProcessor also need to evolve to support [concurrent exporting in the future](https://github.com/open-telemetry/opentelemetry-rust/issues/2555).
5. Also modified Stress test to use the new experimental processor to compare the perf numbers with Concurrent vs Existing  SimpleProcessor.
6. SimpleProcessor did not have a public new method and it did not implement `event_enabled`. This addresses that also, as it was required to run stress test comparison. Can do this in a separate PR it helps review easier.

Spec discussion:
https://github.com/open-telemetry/opentelemetry-specification/issues/4231

### DIFF
--- a/opentelemetry-sdk/src/logs/concurrent_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/concurrent_log_processor.rs
@@ -1,4 +1,4 @@
-use opentelemetry::InstrumentationScope;
+use opentelemetry::{otel_info, InstrumentationScope};
 
 use crate::error::OTelSdkResult;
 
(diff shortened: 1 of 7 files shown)


opentelemetry-sdk/src/logs/concurrent_log_processor.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Creates a new `ConcurrentExportProcessor` with the given exporter.
+    pub fn new(exporter: T) -> Self {
+        Self { exporter }
+    }
+}
+
+impl<T: LogExporter> LogProcessor for SimpleConcurrentProcessor<T> {
+    fn emit(&self, record: &mut SdkLogRecord, instrumentation: &InstrumentationScope) {
+        let log_tuple = &[(record as &SdkLogRecord, instrumentation)];
+        let _ = futures_executor::block_on(self.exporter.export(LogBatch::new(log_tuple)));
```
lalitb, line 30: nit - handle export error/

opentelemetry-sdk/src/logs/concurrent_log_processor.rs
```diff
@@ -0,0 +1,53 @@
...
+/// A concurrent log processor calls exporter's export method on each emit. This
+/// processor does not buffer logs. Note: This invokes exporter's export method
+/// on the current thread without synchronization. i.e multiple export() calls
+/// can happen simultaneously from different threads. This is not a problem if
+/// the exporter is designed to handle that. As of now, exporters in the
+/// opentelemetry-rust project (stdout/otlp) are not thread-safe.
+/// This is intended to be used when exporting to operating system
+/// tracing facilities like Windows ETW, Linux TracePoints etc.
+#[derive(Debug)]
+pub struct SimpleConcurrentProcessor<T: LogExporter> {
```
lalitb, line 16: Should we name it as `SimpleConcurrentLogProcessor`  - to be consistent with `SimpleLogProcessor` and `BatchLogProcessor`.

lalitb (approved):
LGTM.

opentelemetry-sdk/src/logs/simple_log_processor.rs
```diff
@@ -131,6 +132,20 @@ impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
...
+    fn event_enabled(
+        &self,
+        level: opentelemetry::logs::Severity,
+        target: &str,
+        name: Option<&str>,
+    ) -> bool {
+        if let Ok(exporter) = self.exporter.lock() {
+            exporter.event_enabled(level, target, name)
+        } else {
+            true
```
ThomsonTan, line 146: Wondering why `true` is returned to enable events if the exporter cannot be locked?

cijothomas: Good point! Not sure what is the right behavior here. 
Since `true` is the default, I opted for `true`. It can be argued that if the lock got poisoned, no logs can be exported anyway, so returning false also is valid.

Good to revisit this. I'll probably revisit the need of Mutex lock in SimpleLogProcessor, and use atomics to achieve the goal is ensuring one export() is active at a time. With that, we won't face the lock poison issue.

PR #2781, opened 2025-03-11
test: Add test to confirm programmatic config wins over env in BatchConfig

cijothomas:
Part of https://github.com/open-telemetry/opentelemetry-rust/issues/1225
For BatchLogProcessor, we were already preferring code over env_variables, so there is no change required. I added tests to confirm the behavior, and included this information in the docs too.

This is just the start, we need to audit entire ENV variables in the repo, and make changes.

### DIFF
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -200,15 +200,22 @@ impl LoggerProviderBuilder {
         LoggerProviderBuilder { processors, ..self }
     }
 
-    /// Adds a [BatchLogProcessor] with the configured exporter to the pipeline.
+    /// Adds a [BatchLogProcessor] with the configured exporter to the pipeline,
+    /// using the default [super::BatchConfig].
+    ///
+    /// The following environment variables can be used to configure the batching configuration:
+    ///
+    /// * `OTEL_BSP_SCHEDULE_DELAY` - Corresponds to `with_scheduled_delay`.
+    /// * `OTEL_BSP_MAX_QUEUE_SIZE` - Corresponds to `with_max_queue_size`.
+    /// * `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` - Corresponds to `with_max_export_batch_size`.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -200,15 +200,22 @@ impl LoggerProviderBuilder {
...
 
-    /// Adds a [BatchLogProcessor] with the configured exporter to the pipeline.
+    /// Adds a [BatchLogProcessor] with the configured exporter to the pipeline,
+    /// using the default [super::BatchConfig].
+    ///
+    /// The following environment variables can be used to configure the batching configuration:
+    ///
+    /// * `OTEL_BSP_SCHEDULE_DELAY` - Corresponds to `with_scheduled_delay`.
+    /// * `OTEL_BSP_MAX_QUEUE_SIZE` - Corresponds to `with_max_queue_size`.
+    /// * `OTEL_BSP_MAX_EXPORT_BATCH_SIZE` - Corresponds to `with_max_export_batch_size`.
```
lalitb, line 210: The added env variables are for span processor, probably it should be:
```
    /// * `OTEL_BLRP_MAX_QUEUE_SIZE`
    /// * `OTEL_BLRP_SCHEDULE_DELAY`
    /// * `OTEL_BLRP_MAX_EXPORT_BATCH_SIZE`
    /// * `OTEL_BLRP_EXPORT_TIMEOUT`
```

But then these env-variable are already documented for **BatchConfigBuilder** - adding at multiple places can lead to inconsistency, and difficult to maintain.

lalitb (approved):
Thanks for adding the test. Similar to be added for Span Processor too. LGTM once this is fixed - https://github.com/open-telemetry/opentelemetry-rust/pull/2781/files#r1989799553

opentelemetry-sdk/src/logs/logger_provider.rs
cijothomas: I was thinking the same. One argument to keep it here (duplicated) is for convenience. When someone looks at `with_batch_exporter` they can know about configuration options right here. Alternate is to look at BatchConfig and then to BatchConfigBuilder....

PR #2784, opened 2025-03-11
test: fix otlp log integration tests failing with some set of features 

paullegranddc:
I like to run `cargo test --workspace --all-features` to run all tests in a workspace. This is currently failing:

* Running otlp logger integration tests with `reqwest-blocking-client` and the default features in the `integration_test_runner` package are failing two of the tests are not compatible with having both  `reqwest-blocking-client` and `tonic-client` enabled at the same time.
* Running otlp logger integration tests with `experimental_metadata_attributes` fails because of the extra attributes added to logs do not match the snapshot.

## Changes

* Skip tests on incompatible features
* Remove extra attributes before doing the snapshot comparison if the `experimental_metadata_attributes` flag is set

### DIFF
--- a/opentelemetry-otlp/tests/integration_test/tests/logs.rs
+++ b/opentelemetry-otlp/tests/integration_test/tests/logs.rs
@@ -255,6 +255,7 @@ mod logtests {
     // current thread
     #[test]
     #[cfg(feature = "reqwest-blocking-client")]
+    #[cfg(not(feature = "tonic-client"))]
     pub fn logs_batch_non_tokio_main_with_init_logs_outside_rt() -> Result<()> {
         logs_non_tokio_helper(false, false)
     }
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/tests/integration_test/tests/logs.rs
```diff
@@ -309,7 +311,7 @@ mod logtests {
         feature = "tonic-client",
         feature = "reqwest-client"
     ))]
-    pub fn logs_simple_non_tokio_main_with_init_logs_outsie_rt() -> Result<()> {
+    pub fn logs_simple_non_tokio_main_with_init_logs_outside_rt() -> Result<()> {
```
cijothomas, line 314: good catch!

cijothomas (approved):
Thanks!

lalitb (approved):

PR #2788, opened 2025-03-11
test: BatchProcessor stress testing

cijothomas:
Trying out one way to test overall throughput in batching scenarios. This could be another unit test?

### DIFF
--- a/opentelemetry-sdk/src/logs/export.rs
+++ b/opentelemetry-sdk/src/logs/export.rs
@@ -77,6 +77,14 @@ impl LogBatch<'_> {
             index: 0,
         }
     }
+
+    /// Returns the number of log records in the batch.
+    pub fn len(&self) -> usize {
+        match &self.data {
+            LogBatchData::SliceOfOwnedData(data) => data.len(),
+            LogBatchData::SliceOfBorrowedData(data) => data.len(),
+        }
+    }
 }
 
 struct LogBatchDataIter<'a> {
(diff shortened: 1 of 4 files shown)


scottgerring:
This is a cool idea!

I wonder if .... 
* there are similar examples of this sort of thing in other opentelem projects we can be inspired by (probably you already know this! ) 
* there's an idiomatic way ™️  of fitting this into rust's testing infrastructure. E.g. could we structure it to be more like a criterion test ("it took X ms to push through Y logs"), or is that total madness because of the granularity?

cijothomas:
> This is a cool idea!
> 
> I wonder if ....
> 
> * there are similar examples of this sort of thing in other opentelem projects we can be inspired by (probably you already know this! )
> * there's an idiomatic way ™️  of fitting this into rust's testing infrastructure. E.g. could we structure it to be more like a criterion test ("it took X ms to push through Y logs"), or is that total madness because of the granularity?

Inspired from .NET's https://github.com/open-telemetry/opentelemetry-dotnet/pull/1086 4 years ago!

We do have some tests similar to this in this repo already (metrics focused), and I expect to make a normal-test one for this.

The stress test is to test various additional scenarios, like what happens in network cable is pulled out, what if only single core is available.. what if we run this for a long time etc.


> ("it took X ms to push through Y logs")

I'll be working a lot on these kind of ideas for next few weeks/months.
(A specific example is: Today, BatchProcessor sequentially calls Exporter.export(). To achieve higher throughput, this may not be sufficient. We have spec discussion about improving this, so the tests here will prove/disprove the needs)

scottgerring:
> The stress test is to test various additional scenarios, like what happens in network cable is pulled out, what if only single core is available.. what if we run this for a long time etc.

Sounds useful!

Thinking out loud, it'd be amazing if we could track this over time somehow. I don't know if we could use e.g. Sonarqube or bencher, or some other SaasY code quality platform to do this automatically. Maybe that ties into a better way to track the micro-benchmarks too. We could be clever and run this once a week, say, and then bisect manually if there are big regressions.

Issue #2574, opened 2025-01-30
Add ability to add timeout duration to shutdown


cijothomas:
All Providers have shutdown() method now, and it internally uses a hardcoded 5sec timeout to perform shutdown. (usually passed on to reader/processor/exporter). 
Opening an issue to let user provide their own timeout.

Option1 
```
fn shutdown(&self, timeout: Option<Duration>)
```

Option2
```
fn shutdown(&self)
{
self.shutdown_with_timeout(5secs)
}

fn shutdown_with_timeout(&self, timeout: Duration);
```
I am inclined to do 2, as it keeps backward compatibility, and we can add this anytime, not necessarily before 1.0. But given https://github.com/open-telemetry/opentelemetry-rust/pull/2573/files and changes in this area, it'd be better to club them in together.

cijothomas:
Removing from 0.28 as option2 is additive change, and can be done post 0.28 as well.

mohammadVatandoost:
I would be happy to work on this. Please assign it to me

cijothomas:
@mohammadVatandoost Thanks! Assigned. To confirm - you'll be pursuing the option2 from the issue, right?

mohammadVatandoost:
> [@mohammadVatandoost](https://github.com/mohammadVatandoost) Thanks! Assigned. To confirm - you'll be pursuing the option2 from the issue, right?

Yes, backward compatibility is important.

mohammadVatandoost:
I have a question:
we should implement option2 for reader and option1 for exporter and processor. Since we only use reader shutdown from outside, and reader shutdown function runs the exporter/processor's shutdown function. This will not effect the backward compatibility,right?
I didn't find any usecase that we need to run exporter/processor's shutdown function directly.
For logger will be like this:
`
impl SdkLoggerProvider {
 fn shutdown(&self) -> OTelSdkResult{
  self.shutdown_with_timeout(Duration::from_secs(5))
 }
 fn shutdown_with_timeout(&self, timeout: Duration)  -> OTelSdkResult {
 .....
  }
}
// ====================
trait LogProcessor {
 fn shutdown(&self, timeout: Duration) -> OTelSdkResult;
}
// =========
trait LogExporter {
fn shutdown(&mut self, timeout: Duration) -> OTelSdkResult;
}

`
@cijothomas  What is your idea?

cijothomas:
Only providers need option2.
[...]

PR #2789, opened 2025-03-11
feat: Add timeout duration to shutdown

mohammadVatandoost:
#2574 Part 1: Adding shutdown_with_timeout for Log and metric providers. Others will be added in another PR


cijothomas (changes requested):
After reviewing, I realized its better we make additive change everywhere to accept timeout.
https://github.com/open-telemetry/opentelemetry-rust/issues/2574#issuecomment-2715989290

Also, please do one component at a time, to keep PR scope smaller - it is a lot easier to review small PRs.
I suggest starting with Exporter for Metrics. Then Reader+MeterProvider for Metrics. Then repeat for Logs, then traces and so on

mohammadVatandoost:
@cijothomas Ok I close this PR and I will create another PR

PR #2790, opened 2025-03-12
feat: OTLP Exporter builders to return specific Error type

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/2561

Also removes now unused Error enum from Log SDK. Another set of cleanup required to clean Metrics, Traces completely.
Also, the newly introduced enum ExporterBuildError is not fully refined, this can be done in follow ups. This is necessary to unblock Log SDK Stable Release.

Note: This is a breaking change for OTLP Exporter, however, the impact is minimal (most users/examples simply unwrap/expect the Result instead of matching on it.), as described in the changelog, so I propose to make the change even though OTLP Exporter is marked RC.


pitoniak32 (commented):
I just took a brief look, and overall I agree with these changes and think its a good direction to go.

the only small comment I have currently is asserting the errors are of the expected variant for the builder errors

cijothomas:
> comment I have currently is asserting the errors are of the expected variant for the builder errors

Thanks. I expect to add more tests in a follow up. Tests require some combinations of feature flags to ensure it validates what we want it to validate! 
(I just added a basic test in this PR).

examples/tracing-jaeger/src/main.rs
```diff
@@ -3,12 +3,14 @@ use opentelemetry::{
...
 };
-use opentelemetry_sdk::trace::{SdkTracerProvider, TraceError};
+use opentelemetry_otlp::ExporterBuildError;
+use opentelemetry_sdk::trace::SdkTracerProvider;
 use opentelemetry_sdk::Resource;
 
 use std::error::Error;
 
-fn init_tracer_provider() -> Result<opentelemetry_sdk::trace::SdkTracerProvider, TraceError> {
+fn init_tracer_provider() -> Result<opentelemetry_sdk::trace::SdkTracerProvider, ExporterBuildError>
```
TommyCpp, line 12: unintended change for the `{`?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -92,6 +93,44 @@ impl Default for ExportConfig {
...
+#[derive(Error, Debug)]
+/// Errors that can occur while building an exporter.
+// TODO: Refine and polish this.
+// Non-exhaustive to allow for future expansion without breaking changes.
+// This could be refined after polishing and finalizing the errors.
+#[non_exhaustive]
+pub enum ExporterBuildError {
+    /// Spawning a new thread failed.
+    #[error("Spawning a new thread failed. Unable to create Reqwest-Blocking client.")]
+    ThreadSpawnFailed,
```
TommyCpp, line 105: This will only happens if reqwest-blocking is used right? If so can we guard it in feature flag?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -92,6 +93,44 @@ impl Default for ExportConfig {
...
+
+    /// Feature required to use the specified compression algorithm.
+    #[cfg(any(not(feature = "gzip-tonic"), not(feature = "zstd-tonic")))]
+    #[error("feature '{0}' is required to use the compression algorithm '{1}'")]
```
TommyCpp, line 109: Hmm looks like a little bit inconsistent if the first letter is upper case across the variant?

TommyCpp (approved):
Nothing blocking. LGTM once the comments are addressed

opentelemetry-otlp/src/exporter/mod.rs
cijothomas: yes, thats why I put a big TODO on this. we definitely need to refine it and see which ones makes sense and remove others.

opentelemetry-otlp/src/exporter/mod.rs
cijothomas: yes will address them in follow ups.

lalitb (commented):
LGTM - Spent some time to review, and realized it is merged meanwhile :)

PR #2793, opened 2025-03-12
Fix OOM in `span_processor_with_async_runtime::BatchSpanProcessor`

### LINKED ISSUE #2787: [BUG]: OOM in BatchSpanProcessor
# What happened?

OOM in `trace::span_processor_with_async_runtime::BatchSpanProcessor`: https://github.com/open-telemetry/opentelemetry-rust/blob/dac8bd5cde97797be25cf82bd03b4859a1d9f03b/opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs#L236.

`BatchSpanProcessor` adds a span to already overloaded batch which leads to OOM.

Not tested on sync version, but seems like the problem related to both sync and async: https://github.com/open-telemetry/opentelemetry-rust/blob/dac8bd5cde97797be25cf82bd03b4859a1d9f03b/opentelemetry-sdk/src/trace/span_processor.rs#L464

# OpenTelemetry API Version / SDK version

Latest release:
```toml
opentelemetry = "0.28"
opentelemetry_sdk = "0.28"
opentelemetry-otlp = "0.28"
```

# What Exporter(s) are you seeing the problem on?

OTLP

# Relevant log output

Here's the MRE demonstrating the problem: https://github.com/50U10FCA7/otel-oom

# P.S. Please fix your "Bug Report" GitHub template:

Image

50U10FCA7:
Fixes #2787

## Changes

Fixes `BatchSpanProcessor` overflowing.

### DIFF
--- a/opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
+++ b/opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
@@ -234,14 +234,21 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
         match message {
             // Span has finished, add to buffer of pending spans.
             BatchMessage::ExportSpan(span) => {
-                if self.spans.len() == self.config.max_export_batch_size {
+                if self.spans.len() == self.config.max_queue_size {
                     // Replace the oldest span with the new span to avoid suspending messages
                     // processing.
                     self.spans.pop_front();
+
+                    otel_warn!(
+                        name: "BatchSpanProcessor.Export.Error",
+                        dropped_spans = 1,
+                        max_queue_size = self.config.max_queue_size,
+                        message = "Spans were dropped due to a full queue / slow export. The count represents the total count of span records dropped in the lifetime of the BatchSpanProcessor. Consider increasing the queue size and/or decrease delay between intervals."
+                    );
                 }
                 self.spans.push_back(span);
 
-                if self.spans.len() == self.config.max_export_batch_size {
+                if self.spans.len() >= self.config.max_export_batch_size {
                     // If concurrent exports are saturated, wait for one to complete.
                     if !self.export_tasks.is_empty()
                         && self.export_tasks.len() == self.config.max_concurrent_exports
@@ -314,7 +321,8 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
             return Ok(());
         }
 
-        let export = self.exporter.export(self.spans.drain(..).collect());
+        let count = self.spans.len().min(self.config.max_export_batch_size);
+        let export = self.exporter.export(self.spans.drain(..count).collect());
         let timeout = self.runtime.delay(self.config.max_export_timeout);
         let time_out = self.config.max_export_timeout;
 
--- a/opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
+++ b/opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
@@ -6,20 +6,19 @@ use crate::trace::Span;
 use crate::trace::SpanProcessor;
 use crate::trace::{SpanData, SpanExporter};
 use futures_channel::oneshot;
+use futures_util::pin_mut;
 use futures_util::{
     future::{self, BoxFuture, Either},
-    pin_mut, select,
+    select,
     stream::{self, FusedStream, FuturesUnordered},
-    FutureExt as _, StreamExt as _, TryFutureExt as _,
+    StreamExt as _,
 };
 use opentelemetry::Context;
 use opentelemetry::{otel_debug, otel_error, otel_warn};
 use std::collections::VecDeque;
 use std::fmt;
-use std::future::Future;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
-use tokio::sync::RwLock;
 
 /// A [`SpanProcessor`] that asynchronously buffers finished spans and reports
 /// them at a preconfigured interval.
@@ -190,17 +189,14 @@ struct BatchSpanProcessorInternal<E, R> {
     spans: VecDeque<SpanData>,
     export_tasks: FuturesUnordered<BoxFuture<'static, OTelSdkResult>>,
     runtime: R,
-    exporter: Arc<RwLock<E>>,
+    exporter: E,
     config: BatchConfig,
 }
 
-impl<E, R> BatchSpanProcessorInternal<E, R>
-where
-    E: SpanExporter + 'static,
-    R: RuntimeChannel,
-{
+impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
     async fn flush(&mut self, res_channel: Option<oneshot::Sender<OTelSdkResult>>) {
-        let task = self.export().map(|export_result| {
+        let export_result = self.export().await; // TODO: Move execution to `export_tasks`.
+        let task = Box::pin(async move {
             if let Some(channel) = res_channel {
                 // If a response channel is provided, attempt to send the export result through it.
                 if let Err(result) = channel.send(export_result) {
@@ -226,7 +222,7 @@ where
         if self.config.max_concurrent_exports == 1 {
             let _ = task.await;
         } else {
-            self.export_tasks.push(task.boxed());
+            self.export_tasks.push(task);
             while self.export_tasks.next().await.is_some() {}
         }
     }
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -233,27 +238,32 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
...
                         && self.export_tasks.len() == self.config.max_concurrent_exports
                     {
+                        // TODO: Refactor to avoid stopping message processing to not delay
+                        //       shutdown/resource set because of export saturation.
                         self.export_tasks.next().await;
                     }
 
-                    let export_result = self.export().await;
-                    let task = async move {
-                        if let Err(err) = export_result {
```
50U10FCA7, line 248: There's additional minor changes moving the export execution inside `FuturesUnordered`.

Before these changes the export suspend the `BatchSpanProcessor` execution (meaning the `max_concurrent_exports` setting not work).

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -233,27 +238,32 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
...
+                }
+                self.spans.push_back(span);
 
                 if self.spans.len() == self.config.max_export_batch_size {
                     // If concurrent exports are saturated, wait for one to complete.
                     if !self.export_tasks.is_empty()
                         && self.export_tasks.len() == self.config.max_concurrent_exports
                     {
+                        // TODO: Refactor to avoid stopping message processing to not delay
+                        //       shutdown/resource set because of export saturation.
```
50U10FCA7, line 254: Related to [previous comment](https://github.com/open-telemetry/opentelemetry-rust/pull/2793/files#r1991464196).

Queuing the export tasks should be refactored to avoid suspending incoming messages processing. This is not a minor change, so I only placed a TODO.

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -6,18 +6,20 @@ use crate::trace::Span;
...
+    FutureExt as _, StreamExt as _, TryFutureExt as _,
 };
 use opentelemetry::Context;
 use opentelemetry::{otel_debug, otel_error, otel_warn};
+use std::collections::VecDeque;
 use std::fmt;
+use std::future::Future;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
+use tokio::sync::RwLock;
```
lalitb, line 22: We can't assume that the Tokio runtime is always available in this context. Instead, the [runtime abstraction](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/runtime.rs) is used to ensure flexibility and avoid locking the code to a specific async runtime.

50U10FCA7: @lalitb Please read the following: https://docs.rs/tokio/latest/tokio/sync/index.html#runtime-compatibility

lalitb: Thanks for link. This makes sense. However, I was just thinking of the scenario where `experimental_async_runtime` and `rt-async-std` features are enabled for otel-sdk. I believe the compilation will fail in that case, as tokio being optional dependency. Or am I missing something.

lalitb: That makes me think - we don't have CI test for this :)

50U10FCA7: @lalitb Makes sense. We can make `experimental_async_runtime` to depend on `tokio/sync` (it contains only Tokio sync primitives) to avoid importing some other libraries containing asynchronous RwLock. The result will be
```toml
//Cargo.toml
experimental_async_runtime = ["dep:tokio", "tokio/sync"]
rt-tokio = ["dep:tokio", "dep:tokio-stream", "tokio/rt", "tokio/time", "experimental_async_runtime"]
rt-tokio-current-thread = ["dep:tokio", "dep:tokio-stream", "tokio/rt", "tokio/time", "experimental_async_runtime"]
rt-async-std = ["dep:async-std", "experimental_async_runtime"]
```

lalitb: Looks good to me. Also, another option could be to use the [async_std::sync::RwLock](https://docs.rs/async-std/latest/async_std/sync/struct.RwLock.html) for rt-async-std feature.

50U10FCA7: @lalitb

> Also, another option could be to use the [async_std::sync::RwLock](https://docs.rs/async-std/latest/async_std/sync/struct.RwLock.html)

This will bring additional `#[cfg]` attributes depending on runtime, which will increase the code complexity. As long Tokio's RwLock works for any runtime, consider using it.

Also, here's the demo of using `async-std`. Everything complies (as we discussed above), but it seems like `opentelemetry-otlp` crate relying on some Tokio stuff:
https://github.com/50U10FCA7/otel-oom/tree/debug-async-std

cijothomas (commented):
Haven't reviewed this in detail, but I suggest to describe the fix in the PR description to make it easy to review. This looks like a lot of changes, and unless they are directly related to fixing the OOM bug, I'd suggest separate PRs. (Really prefer small, focused PRs)

Additionally, I also suggest to fix the default BatchSpanProcessor (which spawn own thread). It is unknown how we'll support custom async runtimes - it could be enriching the BatchSpanProcessor itself to work better in tokio/other runtimes, and not necessarily in the span_processor_with_async_runtime.
(Not opposed to fixing bugs meanwhile, just sharing that the future of this struct is not decided)

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
lalitb: > Everything complies (as we discussed above), but it seems like opentelemetry-otlp crate relying on some Tokio stuff (opentelemetry-otlp uses hyper-util which relies on Tokio runtime). Not part of this PR, but just mentioning.

Yes, this is expected for otlp/gRPC and otlp/http with hyper.

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
50U10FCA7: Rollback. See https://github.com/open-telemetry/opentelemetry-rust/pull/2793#pullrequestreview-2679184631

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -233,17 +234,24 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
         match message {
             // Span has finished, add to buffer of pending spans.
             BatchMessage::ExportSpan(span) => {
-                self.spans.push(span);
+                if self.spans.len() == self.config.max_export_batch_size {
+                    // Replace the oldest span with the new span to avoid suspending messages
+                    // processing.
+                    self.spans.pop_front();
```
lalitb, line 240: This seems to be silently dropping the old span - Should we log a warning here?

50U10FCA7: @lalitb We should. Also, we probably need to bound on `self.config.max_queue_size` instead of `self.config.max_export_batch_size`. `BatchConfig` notifies the spans will be dropped only if `max_queue_size` reached, not `max_export_batch_size`.

utpilla (changes requested):
It's not clear how the current code is leading to OOM. https://github.com/open-telemetry/opentelemetry-rust/pull/2793#discussion_r1992398484

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -233,17 +234,31 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
         match message {
             // Span has finished, add to buffer of pending spans.
             BatchMessage::ExportSpan(span) => {
-                self.spans.push(span);
```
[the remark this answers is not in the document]
50U10FCA7: @utpilla Whoops, you right. The difference in code that actually fails with OOM and MRE I've trying to repro this bug is in custom `Runtime` using unbounded channel:

```markdown
// something like this.
impl RuntimeChannel for Runtime {
    type Receiver<T: Debug + Send> = UnboundedReceiverStream<T>;
    type Sender<T: Debug + Send> = UnboundedSender<T>;

    fn batch_message_channel<T: Debug + Send>(
        &self,
        _: usize,
    ) -> (Self::Sender<T>, Self::Receiver<T>) {
        let (sender, receiver) = mpsc::unbounded_channel();
        (UnboundedSender(sender), UnboundedReceiverStream::new(receiver))
    }
}
```

No-one mentioned previously that MRE is not reproducing so I didn't noticed there is additional abstraction in my own code.

Not an issue with OpenTelemetry, but my own bug, so closing the PR (and issue) as solved. Thanks!

PR #2796, opened 2025-03-12
fix: Change logger scope in log appender

### LINKED ISSUE #2747: Fix logger scope in log-appender
https://github.com/open-telemetry/opentelemetry-rust/pull/2735#pullrequestreview-2655144624

gruebel:
Fixes #2747 

## Changes

- I was not sure what exactly is the desired state for
https://github.com/open-telemetry/opentelemetry-rust/blob/1ddecb04d9f19dafab682be42c5a2579a8c8182d/opentelemetry-proto/src/transform/common.rs#L57
set all the values to empty strings/vec, if target is `None`?

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -173,7 +173,8 @@ where
 
         let mut log_record = self.logger.create_log_record();
 
-        log_record.set_target(target);
+        // TODO: Fix heap allocation
+        log_record.set_target(target.to_string());
         log_record.set_event_name(name);
         log_record.set_severity_number(severity);
         log_record.set_severity_text(metadata.level().as_str());
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -173,8 +173,7 @@ where
 
         let mut log_record = self.logger.create_log_record();
 
-        // TODO: Fix heap allocation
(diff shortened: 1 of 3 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -173,8 +173,7 @@ where
 
         let mut log_record = self.logger.create_log_record();
 
-        // TODO: Fix heap allocation
-        log_record.set_target(target.to_string());
+        log_record.set_target(target);
```
cijothomas, line 176: oh wait.. this works now?

cijothomas: @utpilla helped me understand better. We were unnecessarily copying the target before!

@gruebel Please do this change in a separate PR, with a before/after benchmark - it should show measurable improvement.

cijothomas:
> Fixes #2747
> 
> ## Changes
> * I was not sure what exactly is the desired state for
>   https://github.com/open-telemetry/opentelemetry-rust/blob/1ddecb04d9f19dafab682be42c5a2579a8c8182d/opentelemetry-proto/src/transform/common.rs#L57
>   
>   set all the values to empty strings/vec, if target is `None`?


If target is None, then get everything from InstrumentationScope itself. If target is there, then it should be used for name/ver/schema_url. But we should still copy the Attributes from Scope. (This requires some more investigation, so we can tackle it later)

cijothomas (approved):

PR #2797, opened 2025-03-13
experiment: prototype measurement processor

martintmk:
In this PR I am trying to experiment with dynamic metric enrichment. This API tries to do the enrichment in a naive way and should not affect he performance if this feature is not used.

I have experimented with using `Cow<'a, [KeyValue]>` as an argument to the processor. The processor might decide to not modify the attributes and, in that case, the original slice is returned without any additional allocations.

When measurement processors are not used, I have not detected any changes in throughput. 

Fixes #

## Changes


examples/metrics-advanced/src/main.rs
```diff
@@ -154,3 +159,36 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
...
+            UserType::User => "user",
+            UserType::Admin => "admin",
+        }
+    }
+}
+
+struct UserTypeMeasurementProcessor;
+
+impl MeasurementProcessor for UserTypeMeasurementProcessor {
+    fn process<'a>(&self, attributes: Cow<'a, [KeyValue]>) -> Cow<'a, [KeyValue]> {
```
cijothomas, line 182: could the trait be modified to accept slice of attributes, and return Option<Vec> ? If processor has no interest in updating, it can return None, at which point the caller uses the original slice itself.
If processor wants to change, it'll create a new Vec! and return it. 
Cost is paid only when one wants to enrich..

martintmk: I have updated the trait with this suggestion. Looks nicer indeed.

martintmk: FYI: One big advantage of using `Cow` is the re-use of owned `Vec` in case multiple processors are modifying the attributes:

```
impl MeasurementProcessor for UserTypeMeasurementProcessor {
    fn process<'a>(&self, attributes: Cow<'a, [KeyValue]>) -> Cow<'a, [KeyValue]> {
        match Context::current().get::<UserType>() {
            Some(user_type) => {
                let mut attrs = attributes.into_owned();
                attrs.push(KeyValue::new("user_type", user_type.as_str()));
                Cow::Owned(attrs)
            }

            // No changes to the attributes
            None => attributes,
        }
    }
}
```

The `into_owned` will either:

- Return already allocated vector: `Cow::Owned`
- Create a new vector in case the `Cow` contained the borrowed data.

Blinkuu:
Hey @martintmk, I love this experiment!

We have a formal proposal for the MeasurementProcessor concept, which you can find here: https://github.com/open-telemetry/opentelemetry-specification/pull/4318

It would be wonderful if you could slightly refactor this PoC to follow the proposal. This would help with getting the proposal merged.

From a functional standpoint, it would require redefining the MeasurementProcessor trait from:

```rust
pub trait MeasurementProcessor: Send + Sync + 'static {
    fn process<'a>(&self, attributes: &[KeyValue]) -> Option<Vec<KeyValue>>;
}
```

to something like:

```rust
pub trait MeasurementProcessor: Send + Sync + 'static {
    fn process<F>(&self, measurement: Measurement, next_processor: F)
    where
        F: FnOnce(Measurement);
}
```

> Note: Please forgive me if the above doesn't make sense - I don't write Rust :D

PR #2799, opened 2025-03-13
perf: Small perf optimization for logs

cijothomas:
Originally done in https://github.com/open-telemetry/opentelemetry-rust/pull/2796#discussion_r1992461998 by @gruebel 
Also nit improvement in trace_context.


lalitb (approved):
Thanks.

PR #2801, opened 2025-03-13
fix: LogProcessor.SetResource to require mutable self

cijothomas:
While attempting to measure throughput for `SimpleConcurrentLogProcessor`, I realized that it cannot pass the resource to exporter, as exporter's set_resource required `mut self` but processor only needed `self`. This PR addresses the issue, so that processor can now pass resource to the exporter.
(The stress test is modified to show that processor can pass resource to exporter without sacrificing performance. Without this PR, processor had to rely on interior mutability (i.e store exporter in a Mutex), affecting throughput.

This also opens up making Simple/BatchProcessor a lot more simpler. That'll be in a follow up PR.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -64,6 +64,9 @@
  `LogExporter` trait no longer requires a mutable ref to `self`. If the exporter
  needs to mutate state, it should rely on interior mutability.
  [2764](https://github.com/open-telemetry/opentelemetry-rust/pull/2764)
+- **Breaking** for custom LogProcessor/Exporter authors: Changed `set_resource`
+  to require mutable ref.
+  `fn set_resource(&mut self, _resource: &Resource) {}`
 
 ## 0.28.0
 
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -64,7 +64,7 @@
  `LogExporter` trait no longer requires a mutable ref to `self`. If the exporter
  needs to mutate state, it should rely on interior mutability.
  [2764](https://github.com/open-telemetry/opentelemetry-rust/pull/2764)
-- **Breaking** for custom LogProcessor/Exporter authors: Changed `set_resource`
+- **Breaking** for custom `LogProcessor` authors: Changed `set_resource`
   to require mutable ref.
   `fn set_resource(&mut self, _resource: &Resource) {}`
 
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -258,7 +258,7 @@ impl LoggerProviderBuilder {
 
         let logger_provider = SdkLoggerProvider {
             inner: Arc::new(LoggerProviderInner {
-                processors: processors,
+                processors
                 is_shutdown: AtomicBool::new(false),
             }),
         };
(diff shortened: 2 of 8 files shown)


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/1898 can be revisited after this is merged, as it should be a lot simpler.

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -346,9 +338,43 @@ mod tests {
...
+                resource: Arc::new(Mutex::new(Resource::empty())),
+            }
+        }
+        fn resource(&self) -> Resource {
+            self.resource.lock().unwrap().clone()
+        }
+    }
+
     #[test]
     fn test_logger_provider_default_resource() {
```
utpilla, line 376: This test should either include the assertions to check resources for exporter as well or let's rename the test to: 
```suggestion
    fn test_logger_provider_default_resource_for_log_processor() {
```

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -346,9 +338,43 @@ mod tests {
             Ok(())
         }
     }
+
+    #[derive(Debug, Clone)]
+    struct TestProcessorForResource {
+        resource: Arc<Mutex<Resource>>,
```
utpilla, line 344: Could we get rid of the `Arc<Mutex<T>>` wrapper now that you have updated the method to require a `&mut self`?

opentelemetry-sdk/src/logs/logger_provider.rs
utpilla: Basically, we need to test that both processors and exporter are able to get hold of the provider's `Resource`.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -64,6 +64,9 @@
  `LogExporter` trait no longer requires a mutable ref to `self`. If the exporter
  needs to mutate state, it should rely on interior mutability.
  [2764](https://github.com/open-telemetry/opentelemetry-rust/pull/2764)
+- **Breaking** for custom LogProcessor/Exporter authors: Changed `set_resource`
```
utpilla, line 67: It's only affecting `LogProcessor` authors, right?

```suggestion
- **Breaking** for custom `LogProcessor` authors: Changed `set_resource`
```

opentelemetry-sdk/src/logs/logger_provider.rs
cijothomas: Provider's responsibility ends when it hands over Resource to processor. Then it is upto processor to pass it to exporter. So this test is testing provider only.

I'll add tests to cover exporter too. Do you prefer to modify the existing test itself? or add one in process.rs file?

opentelemetry-sdk/CHANGELOG.md
cijothomas: true!

opentelemetry-sdk/src/logs/logger_provider.rs
cijothomas: added exporter coverage too to this test

utpilla (approved):
Left a comment about testing the right thing. Not blocking for this PR.

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -346,71 +338,190 @@ mod tests {
...
+        }
+
+        fn shutdown(&self) -> OTelSdkResult {
+            Ok(())
+        }
+
+        fn set_resource(&mut self, resource: &Resource) {
+            let mut res = self.resource.lock().unwrap();
+            *res = resource.clone();
+            self.exporter.set_resource(resource);
```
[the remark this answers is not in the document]
cijothomas: such tests exists in the corresponding modules : https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/simple_log_processor.rs#L198-L218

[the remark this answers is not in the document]
utpilla: Ah okay.

PR #2805, opened 2025-03-14
chore: Modify OTLP doc to show jaeger usage instead of a full example

cijothomas:
No need for jaeger specific example.


gruebel (approved):
always nice to remove stuff 😄

opentelemetry-otlp/src/lib.rs
```diff
@@ -1,33 +1,68 @@
...
+//! fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+//!     // Initialize OTLP exporter using HTTP binary protocol
+//!     let otlp_exporter = opentelemetry_otlp::SpanExporter::builder()
+//!         .with_http()
+//!         .with_protocol(Protocol::HttpBinary)
+//!         .build()?;
+//!
+//!     // Create a tracer provider with the exporter
+//!     let _ = opentelemetry_sdk::trace::SdkTracerProvider::builder()
+//!         .with_simple_exporter(otlp_exporter)
```
lalitb, line 43: I think better to document examples with batch processor.

opentelemetry-otlp/src/lib.rs
```diff
@@ -36,22 +71,41 @@
...
+//! # }
 //! }
 //! ```
 //!
+//! ## Using with Jaeger
+//!
+//! Jaeger natively supports the OTLP protocol, making it easy to send traces directly:
+//!
+//! ```shell
+//! $ docker run -p 16686:16686 -p 4317:4317 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:latest
```
lalitb, line 100: 16686 is the default port for OTLP/HTTP ?

lalitb (approved):

opentelemetry-otlp/src/lib.rs
cijothomas: slightly below, there is a section about performance which talks about Batch vs Span. I think that should be sufficient right? (Batch here means we now have to show shutdown too, making this example more lengthier.. But I don't really mind that either!)

opentelemetry-otlp/src/lib.rs
cijothomas: I think it supports both.

Issue #2723, opened 2025-02-26
Show best practices in all Examples


cijothomas:
Parent issue for tracking. All examples maintained in the repo should show the best practices.

1. Cache Tracer : https://github.com/open-telemetry/opentelemetry-rust/pull/2709
2. Add others here.

cijothomas:
Another showed up during community call today:
How to do shutdown correctly! The following does not proceed to call remaining shutdowns, if one before fails. That needs to be fixed
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp/src/main.rs#L159-L161

PR #2806, opened 2025-03-14
docs: OTLP Examples to shutdown all signals

cijothomas:
Current code will not invoke shutdown on the rest, if any provider fails.

Addressing https://github.com/open-telemetry/opentelemetry-rust/issues/2723#issuecomment-2715101190


lalitb (approved):
Thanks.

PR #2807, opened 2025-03-14
fix: Byte arrays represented as Anyvalue::Byte instead of string

cijothomas:
Not very performant due to the need to vec! allocate and box, but this is the best we can do right now.

### DIFF
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -46,7 +46,8 @@ transparent to most users.
 when conversion is feasible. Otherwise stored as
 `opentelemetry::logs::AnyValue::String`. This avoids unnecessary string
 allocation when values can be represented in their original types.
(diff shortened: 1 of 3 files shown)


lalitb (approved):

lalitb:
> Not very performant due to the need to vec! allocate and box, but this is the best we can do right now.

Agree. If really want to be performant, we can use the [Bytes](https://crates.io/crates/bytes):
```rust
use bytes::Bytes;

pub enum AnyValue {
   // ...
    Bytes(Bytes),
 // .. 
}
```

But then, external dependency, even if it is coming from Tokio Org. And eventually OTLP also translates to [Vec\<u8\>](https://github.com/open-telemetry/opentelemetry-rust/blob/6e58810080951d06f0646486141be763f27bfdc6/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.common.v1.rs#L45)

PR #2808, opened 2025-03-15
fix: Report error using OTel convention

### LINKED ISSUE #2150: [Feature]: Implement error recording on opentelemetry-appender-tracing

cijothomas:
Fixes #2150 
Only does "exception.message", as I don't see any way to report other attributes. This may be improved in the future.
Also updated to map u128 and i128.
This is the last change planned before stable release of the crate.


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -80,6 +80,18 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
     }
 
+    fn record_error(
+        &mut self,
+        _field: &tracing_core::Field,
+        value: &(dyn std::error::Error + 'static),
+    ) {
+        self.log_record.add_attribute(
+            Key::new("exception.message"),
+            AnyValue::from(value.to_string()),
```
lalitb, line 90: The consequence of this change is that users will no longer be able to record multiple error messages. Earlier it was possible as they were recorded as attribute under different keys, Eg. for 
```rust
error!(....,error1 = &OTelSdkError::AlreadyShutdown as &dyn std::error::Error, error2 = &OTelSdkError::InternalError as &dyn std::error::Error)
```
Only one of the errors will be retained (likely InternalError), while the other will be lost due to being overwritten.

cijothomas: Good point! Wondering if we should do something about it or not? (Note: Nothing is lost, as we don't override, so it'll result in duplicate keys which is also bad). OTel spec has provision to report one exception on one LogRecord anyway, so if user has multiple, they should make multiple calls rather than trying to put them all to one statement.

If there is feedback on improving this, we can introduce an opt-in option to *not* report exception as Otel way. Given opt-in, we can do it back-compat way, if there is a need in the future.

lalitb: > Note: Nothing is lost, as we don't override, so it'll result in duplicate keys which is also bad

Yes correct. So this is bad, but not bad enough to loose data. :) Does it make sense to keep on appending the errors to existing string value. Or else, value can be json-serialized string which means both key and values for the errors would be preserved.

opentelemetry-appender-tracing/src/lib.rs
```diff
@@ -112,7 +114,7 @@
 //! |----------------|-------------------------------|
 //! | `i64`          | `Int`                         |
 //! | `f32`, `f64`   | `Double`                      |
-//! | `u64`          | `Int` (if convertible without loss) else `String` |
+//! | `u64`,`u128` ,`i128`         | `Int` (if convertible without loss) else `String` |
```
utpilla, line 117:
```suggestion
//! | `u64`,`u128` ,`i128`         | `Int` (if convertible to `i64` without loss) else `String` |
```

opentelemetry-appender-tracing/src/layer.rs
cijothomas: I am reluctant to offer appending behavior now as that is going in un-chartered territories. I see two options
1. Do what this PR is attempting to do. If a user reports the need to report multiple Error in one log statement, we can tell them it is not supported. If there is huge demand, we can do an opt-in feature to support it via appending or keeping user provided key itself.
2. Do not do any special for dyn Error - report it as-is, and ignore OTel convention. If someone complains in the future, we can make an opt-in feature to follow OTel convention.

I prefer 1, but happy to hear more opinions.

lalitb: Option 1 looks fine to me.

lalitb (approved):

Issue #2777, opened 2025-03-10
Handle shutdown in OTLP/gRPC


cijothomas:
tonic does not handle shutdown properly today. The gRPC client is not dropped during shutdown. 

See discussion here: https://github.com/open-telemetry/opentelemetry-rust/pull/2764#discussion_r1986133497

stefanobaghino:
@scottgerring I'd like to work on this. 🙂

Issue #2770, opened 2025-03-08
LogExporter not invoking shutdown on the client.


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/logs.rs#L141-L158
Looks like LogExporter never call shutdown on the underlying client (tonic/http etc.)

Same for Traces. Metrics is doing this correctly.

PR #2812, opened 2025-03-16
fix: Adds shutdown function for spans

### LINKED ISSUE #2779: SpanExporter shutdown
Similar to https://github.com/open-telemetry/opentelemetry-rust/issues/2770

shivanshuraj1333:
Fixes: https://github.com/open-telemetry/opentelemetry-rust/issues/2779

### DIFF
--- a/opentelemetry-otlp/src/exporter/tonic/trace.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/trace.rs
@@ -90,6 +90,11 @@ impl SpanExporter for TonicTracesClient {
     }
 
     fn shutdown(&self) -> OTelSdkResult {
+        // ToDo: as part of https://github.com/open-telemetry/opentelemetry-rust/pull/2812
+        // self is no longer mutable due to trait change for span exporter
+        // the shutdown for gRPC needs to implemented and tracked in
+        // https://github.com/open-telemetry/opentelemetry-rust/issues/2777
+        //
         // match self.inner.take() {
         //     Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
         //     None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
--- a/opentelemetry-otlp/src/exporter/tonic/trace.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/trace.rs
@@ -89,11 +89,12 @@ impl SpanExporter for TonicTracesClient {
         Ok(())
     }
 
-    fn shutdown(&mut self) -> OTelSdkResult {
-        match self.inner.take() {
-            Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
-            None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
-        }
+    fn shutdown(&self) -> OTelSdkResult {
+        // match self.inner.take() {
+        //     Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
+        //     None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
+        // }
+        Ok(())
     }
 
     fn set_resource(&mut self, resource: &opentelemetry_sdk::Resource) {
--- a/opentelemetry-otlp/src/span.rs
+++ b/opentelemetry-otlp/src/span.rs
@@ -159,4 +159,12 @@ impl opentelemetry_sdk::trace::SpanExporter for SpanExporter {
             SupportedTransportClient::Http(client) => client.set_resource(resource),
         }
     }
+    fn shutdown(&self) -> OTelSdkResult {
+        match &self.client {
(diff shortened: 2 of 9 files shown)


opentelemetry-otlp/src/span.rs
```diff
@@ -159,4 +159,12 @@ impl opentelemetry_sdk::trace::SpanExporter for SpanExporter {
             SupportedTransportClient::Http(client) => client.set_resource(resource),
         }
     }
+    fn shutdown(&self) -> OTelSdkResult {
```
scottgerring, line 162: Hey @shivanshuraj1333 thanks for the interest in contributing! 

Looks like you need:

```suggestion
    fn shutdown(&mut self) -> OTelSdkResult {
```

gruebel: depending on the scope of the PR this needs much more changes to actually make `shutdown()` work without the need of mutability or this can be tackled via a follow-up PR. @cijothomas thoughts, because you created the issue? 😅

opentelemetry-otlp/src/span.rs
```diff
@@ -159,4 +159,12 @@ impl opentelemetry_sdk::trace::SpanExporter for SpanExporter {
...
+        match &self.client {
```
gruebel, line 163: additionally to @scottgerring suggestion you also need to make it mutable here
```suggestion
        match &mut self.client {
```

opentelemetry-otlp/src/span.rs
cijothomas: yes this might be more involved. Following what Log signal is doing would be a good start.

shivanshuraj1333: Can I get a one liner description of what is the expected output of the `shutdown` implementation of traces?
How can I decide if the self is mutable or not? Following the changes from https://github.com/open-telemetry/opentelemetry-rust/issues/2770, I need some inputs on how to test it end to end, if the shutdown for traces working as expected.

I'm new to rust, that's why some questions, to get started.

shivanshuraj1333:
After taking some inspirations from the logs PR https://github.com/open-telemetry/opentelemetry-rust/pull/2764/files
I have updated the trait, and not sure how to handle shutdown for `tonic/trace.rs`

opentelemetry-otlp/src/exporter/tonic/trace.rs
```diff
@@ -89,11 +89,12 @@ impl SpanExporter for TonicTracesClient {
         Ok(())
     }
 
-    fn shutdown(&mut self) -> OTelSdkResult {
-        match self.inner.take() {
-            Some(_) => Ok(()), // Successfully took `inner`, indicating a successful shutdown.
-            None => Err(OTelSdkError::AlreadyShutdown), // `inner` was already `None`, meaning it's already shut down.
-        }
+    fn shutdown(&self) -> OTelSdkResult {
+        // match self.inner.take() {
```
shivanshuraj1333, line 93: need to make self mutable for this to work, shutdown for tonic needs to be implemented according to https://github.com/open-telemetry/opentelemetry-rust/issues/2777

opentelemetry-otlp/src/span.rs
shivanshuraj1333: update: have made the necessary changes

opentelemetry-otlp/src/span.rs
shivanshuraj1333: just like other signals, have made the change in the trait to not requiring mutable self

shivanshuraj1333:
@cijothomas does this also require a change like this? https://github.com/open-telemetry/opentelemetry-rust/pull/2764/files#diff-7e030c09778110d005a4f1b3030b8318bffa2ca09a7d897877ec71a495a971f4R16

opentelemetry-otlp/src/exporter/tonic/trace.rs
cijothomas: can you add a TODO here linking to the issue.

cijothomas (approved):
LGTM.
Please add changelog entry marking breaking change for custom processor/exporter authors

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -17,6 +17,10 @@
  `Error` which contained many variants unrelated to building an exporter, the
  new one returns specific variants applicable to building an exporter. Some
  variants might be applicable only on select features.
+- *Breaking*
```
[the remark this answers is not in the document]
cijothomas: We need to list the breaking change in the SDK changelog. (You can check the previous entries there that did similar breaking changes)

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust/pull/2812/files

Blocking only to prevent merge while changelog is fixed.

lalitb (approved):
LGTM, once changelog is fixed.

scottgerring:
Hey @shivanshuraj1333 apologies this has slipped through our net!
Do you still have time to quickly resolve the diffs against main so we can try get this in?

PR #2817, opened 2025-03-17
test: Improve OTLP Builder test

cijothomas:
Edit: Ignore is the best option now. See https://github.com/open-telemetry/opentelemetry-rust/pull/2817#issuecomment-2730554440


cijothomas:
This helped identify the root cause. The failures are due to some env variables being set by tests running in parallel. Given env variable are treated as high priority, this test is fragile now. The best option is to ignore them, until https://github.com/open-telemetry/opentelemetry-rust/issues/2818 is resolved.

gruebel (approved):
👍

lalitb (approved):

PR #2821, opened 2025-03-18
feat: Add Suppression flag to context

cijothomas:
This **establishes the basic building block** required to suppress telemetry using Context. This is a pre-requisite for solving issues like https://github.com/open-telemetry/opentelemetry-rust/issues/1171 or https://github.com/open-telemetry/opentelemetry-rust/issues/761
This is not yet applied anywhere, that can be in follow ups.
(I did try in Logs Sdk, and there is a 1-2 ns additional overhead to check this. It may not look much, but given Logs SDK can check IsEnabled() currently under 2 ns, this is a measurable increase in terms of percentage!. But this should not be a major concern.)

Previous attempts: https://github.com/open-telemetry/opentelemetry-rust/pull/1330


opentelemetry/benches/context_suppression.rs
```diff
@@ -0,0 +1,66 @@
...
+// The benchmark results:
+// criterion = "0.5.1"
+// Hardware: Apple M4 Pro
+// Total Number of Cores:   14 (10 performance and 4 efficiency)
+// | Benchmark                  | Time      |
+// |----------------------------|-----------|
+// | enter_suppressed           | 9.0 ns    |
+// | normal_attach              | 9.0 ns    |
+// | is_current_suppressed_false| 750 ps    |
+// | is_current_suppressed_true | 750 ps    |
```
cijothomas, line 16: this is quite inexpensive!

opentelemetry/src/context.rs
```diff
@@ -328,19 +330,183 @@ impl Context {
...
+    ///
+    /// // Default context has telemetry suppression disabled
+    /// assert_eq!(Context::is_current_suppressed(), false);
+    ///
+    /// // Enter a suppressed scope
+    /// let _guard = Context::enter_suppressed();
+    /// assert_eq!(Context::is_current_suppressed(), true);
+    /// ```
+    #[inline]
+    pub fn is_current_suppressed() -> bool {
```
cijothomas, line 465: maybe I can start with just this method and `enter_unsuppressed` only in public API, and keep others non-public... That should do the job for now.

cijothomas: for the initial need, this can be in the SDK, not in the API.
There is a possible use case for this to be in the API, which is instrumentations of higher layer want to suppress underlying layer. (like reqwest wanting to suppress hyper..)..
But I am okay with start with SDK to keep the API lighter. And move to API when the actual need comes.

cijothomas: From community call today: will keep in API itself, to avoid any future movement.

opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## vNext
 
+Added the ability to prevent recursive telemetry generation through new
+context-based suppression mechanisms. This feature helps prevent feedback loops
+and excessive telemetry when OpenTelemetry components perform their own
+operations.
```
lalitb, line 8: is this adding support for suppressing internal telemetry, or also the telemetry generated by dependencies which are instrumented using otel api.

opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
...
+
+New methods added to `Context`:
+
+- `is_telemetry_suppressed()` - Checks if telemetry is suppressed in this
+  context
+- `with_telemetry_suppressed()` - Creates a new context with telemetry
```
utpilla, line 14: You were planning to update the PR to only have two public functions for now, right? https://github.com/open-telemetry/opentelemetry-rust/pull/2821#discussion_r2000016166

opentelemetry/CHANGELOG.md
cijothomas: all. To be precise, as long as 
1. Otel context is carried forward, 
2. and suppression flag not over-ridden. (`test_nested_suppression_scopes` test shows when suppression flows/not-flows)

opentelemetry/CHANGELOG.md
lalitb: cool. Thanks.

bantonsson (commented):
I think that the addition is good, but you should run the benchmarks on this change. Locally for me this gives between 63% and 173% slower execution of the `context_attach` benchmarks.

I have some ideas of how to remedy this.

opentelemetry/src/context.rs
```diff
@@ -897,4 +979,171 @@ mod tests {
...
+            assert!(Context::current().is_telemetry_suppressed());
+        }
+
+        // After guard is dropped, should be back to unsuppressed
+        assert!(!Context::is_current_telemetry_suppressed());
+        assert!(!Context::current().is_telemetry_suppressed());
+    }
+
+    #[test]
+    fn test_is_current_telemetry_suppressed() {
```
utpilla, line 1039: It looks like this test is redundant. `test_nested_suppression_scopes` is already covering what this test covers.

opentelemetry/src/context.rs
```diff
@@ -897,4 +979,171 @@ mod tests {
...
+    fn test_enter_telemetry_suppressed_scope() {
+        // Ensure we start with a clean context
+        let _reset_guard = Context::new().attach();
+
+        // Default context should not be suppressed
+        assert!(!Context::is_current_telemetry_suppressed());
+
+        // Enter a suppressed scope
+        {
+            let _guard = Context::enter_telemetry_suppressed_scope();
```
utpilla, line 1028: It'd be good to test that `enter_telemetry_suppressed_scope()` also clones the existing entries.

opentelemetry/src/context.rs
```diff
@@ -897,4 +979,171 @@ mod tests {
...
+                    Context::current().get::<ValueB>(),
+                    Some(&ValueB(24)),
+                    "Parent value should still be available after adding new value"
+                );
+                assert!(Context::is_current_telemetry_suppressed());
+
+                // Do some async work to simulate real-world scenario
+                sleep(Duration::from_millis(10)).await;
+
+                // Values should still be available after async work
```
utpilla, line 1120: `#[tokio::test]` would be using the current-threaded runtime. Could we test this for the multi-threaded runtime as well? I suspect we would have to update the sleep call with the context for that.
```rust
sleep(Duration::from_millis(10)).with_context(Context::current()).await;
```

cijothomas: since we are not doing anything *inside* sleep() call, what is the need to invoking it with_context? If it was any other async method, then we should invoke with with_context(), so that within the body of that method, we have correct context. I am not following why we need to sleep with context....

cijothomas:
> I think that the addition is good, but you should run the benchmarks on this change. Locally for me this gives between 63% and 173% slower execution of the `context_attach` benchmarks.
> 
> I have some ideas of how to remedy this.

Thanks! Yes it is somewhat surprising to me too that the attach benches regress (I do see 50-80% slow). 
We can keep improving the perf with pure internal optimizations, give we expose no internals.

Additionally, sharing what I believe must be the highest perf path:
1. Checking if current is in telemetry_suppressed -- critical, as every hot path (logger.emit, span.start etc.) has to check this.
2. Check if there is a Span in the current and is that Span is being recorded - critical as this will allow users to quickly short circuit code, if there is no active span and code has no intent to start new one.

Enter suppression_context - is done by background thread/exporters only and is less important to be as performant as others.
Baggage - important, but less important as Baggage is anyway propagated via HTTP headers, so there is non-trivial cost already for Baggage, so adding few nanoseconds should be okay.

cijothomas:
@utpilla and I discussed another thought: When one is entering suppression mode - is there a value in keeping/cloning existing context? What good is baggage or active span, if we are in suppressed mode anyway? 
Won't be exploring this idea for this PR, but will come back to this question after this PR's subsequent steps are also done throughout the repo.

lalitb:
Just an alternative idea specifically for suppression logic in the synchronous case:

log/span -> concurrent_processor -> exporter
or
log/span -> simple_exporter -> otlp (using reqwest_blocking)

Can we rely on a thread-local suppression flag to handle the logic, avoiding the need for context cloning?

cijothomas:
> Just an alternative idea specifically for suppression logic in the synchronous case:
> 
> log/span -> concurrent_processor -> exporter or log/span -> simple_processor -> otlp (using reqwest_blocking)
> 
> Can we rely on a thread-local suppression flag to handle the logic, avoiding the need for context cloning?

As mentioned in https://github.com/open-telemetry/opentelemetry-rust/pull/2821#issuecomment-2754802719 , I don't think entering_suppression is a hot path operation to heavily perf optimize.

lalitb:
> > Just an alternative idea specifically for suppression logic in the synchronous case:
> > log/span -> concurrent_processor -> exporter or log/span -> simple_processor -> otlp (using reqwest_blocking)
> > Can we rely on a thread-local suppression flag to handle the logic, avoiding the need for context cloning?
> 
> As mentioned in [#2821 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/2821#issuecomment-2754802719) , I don't think entering_suppression is a hot path operation to heavily perf optimize.

Got it. For,
 log/span -> concurrent_processor -> exporter or 
log/span -> simple_processor -> otlp (using reqwest_blocking)

Would it still be outside hot-path? Not a blocker, just curious how will it work. Also, once suppression is enabled, would there be situation where we need to switch it off?

opentelemetry/src/context.rs
utpilla: My bad. I was concerned about how context gets propagated across the sleep await point. I forgot about the `Future` implementation for `WithContext` which ensures that things work correctly https://github.com/open-telemetry/opentelemetry-rust/blob/f3e93a09ea9059487c4a8a7e402fce00cc56aa34/opentelemetry/src/context/future_ext.rs#L13-L18

opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## vNext
 
+Added the ability to prevent recursive telemetry generation through new
```
utpilla, line 5: Add PR link as well.

cijothomas:
> > > Just an alternative idea specifically for suppression logic in the synchronous case:
> > > log/span -> concurrent_processor -> exporter or log/span -> simple_processor -> otlp (using reqwest_blocking)
> > > Can we rely on a thread-local suppression flag to handle the logic, avoiding the need for context cloning?
> > 
> > 
> > As mentioned in [#2821 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/2821#issuecomment-2754802719) , I don't think entering_suppression is a hot path operation to heavily perf optimize.
> 
> Got it. For, log/span -> concurrent_processor -> exporter or log/span -> simple_processor -> otlp (using reqwest_blocking)
> 
> Would it still be outside hot-path? Not a blocker, just curious how will it work. Also, once suppression is enabled, would there be situation where we need to switch it off?

Yes you are totally right! For ConcurrentProcessor, it'll be in hot-path but I don't plan to use it for them at all. (You may have seen many PRs from me in the contrib repo, where I try to make the exporters used with ConcurrentProcessor safe from re-entrancy, so we don't have to pay the cost!)

The benchmarks in this PR shows there is a 10ns cost to just attach a context, whether or not it has suppression flag in it. That cost is not something we should be paying for the re-entrant ones. That limits observability, but there are alternate ideas to address them. 

I hope to uncover all these in the follow up PRs, where we put the suppression flag into actual use!

cijothomas:
> Also, once suppression is enabled, would there be situation where we need to switch it off?

I don't see the need myself for the purposes inside this repo. But if a component author wants to, they can turn it off. There is a test that shows this too.

PR #2823, opened 2025-03-18
perf: Inline enabled check in Logger

cijothomas:
Good boost to the enabled() check.

### DIFF
--- a/opentelemetry-sdk/benches/log_enabled.rs
+++ b/opentelemetry-sdk/benches/log_enabled.rs
@@ -5,9 +5,11 @@
     Total Number of Cores:   14 (10 performance and 4 efficiency)
     | Test                                         | Average time|
     |---------------------------------------------|-------------|
-    | exporter_disabled_concurrent_processor      |  1.9 ns     |
-    | exporter_disabled_simple_processor          |  5.0 ns     |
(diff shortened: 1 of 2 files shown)


shivanshuraj1333 (approved):

opentelemetry-sdk/benches/log_enabled.rs
```diff
@@ -5,10 +5,12 @@
     Total Number of Cores:   14 (10 performance and 4 efficiency)
     | Test                                         | Average time|
     |---------------------------------------------|-------------|
-    | exporter_disabled_concurrent_processor      |  1.9 ns     |
-    | exporter_disabled_simple_processor          |  5.0 ns     |
+    | exporter_disabled_concurrent_processor      |  1.0 ns     |
```
utpilla, line 8: Add the inline attribute to the processor and exporter's `event_enabled` methods as well?

lalitb (approved):

opentelemetry-sdk/benches/log_enabled.rs
cijothomas: I tried that but it didn't give any boost 🤷‍♂️

PR #2825, opened 2025-03-18
fix: SimpleProcessor for Logs simplified

cijothomas:
SimpleLogProcessor has no technical reason to Mutex wrap the exporter, as it does not need mutable ref to exporter. However, as per spec, only one Export() can be active at a time. This PR removes the Mutex wrapping of exporter, and uses a separate bool to sequence the export() calls. 
This simplifies other methods, which no longer has to lock the exporter.

There is minor perf gains, but that should be immaterial as SimpleProcessor for only for test/learn purposes.


opentelemetry-sdk/src/logs/simple_log_processor.rs
```diff
@@ -60,54 +59,47 @@ use std::sync::Mutex;
...
-            otel_warn!(
-                name: "SimpleLogProcessor.Emit.ProcessorShutdown",
-            );
-            return;
+        // export() does not require mutable self and can be called in parallel
+        // with other export() calls. However, OTel Spec requires that
+        // existing export() must be completed before the next export() call.
+        while !self
+            .is_exporting
+            .compare_exchange(false, true, Ordering::Acquire, Ordering::Relaxed)
```
lalitb, line 83: This looks inefficient to me - spin lock with busy loop. I understand the processor is not for production but still don't see if we really need to add this. If Mutex is not really the way, should we use the mpsc channel instead.

cijothomas: Yes, I can use mutex and block until lock is acquired. (All I wanted is to *not* wrap Exporter in mutex.)

cijothomas: this is done!

opentelemetry-sdk/src/logs/simple_log_processor.rs
```diff
@@ -116,21 +98,11 @@ impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
     }
 
     fn shutdown(&self) -> OTelSdkResult {
-        self.is_shutdown
```
utpilla, line 119: This code is not complicated. It's simple enough.

I don't find the `Mutex<()>` approach making things simpler. In fact, from a readability standpoint, the existing `export` code is much better than the one introduced in this PR. If you're using a `Mutex` to guard access to some resource, then it's best to wrap it with `Mutex` instead of having to manually ensure that we acquire a `Mutex`.

cijothomas: I don't disagree at all! 
It felt quite awkward to Mutex lock to do `EventEnabled()`, though it did not require mutability! 
Not worried about performance at all here, just does not feel right to do that.

Regd. Rust idiomatic way to prevent multiple export at the same time - I think it'd be better to model export() as requiring mutable self, but we know that will limit us from achieving higher perf when we need it!

If all agree *this* PR is wrong direction, I can abandon it. (Its harmless with/without this PR, as simple processor is just a learning/test purpose component only)

PR #2827, opened 2025-03-18
refactor: prioritize code-based config over env vars in OTLP exporter

### LINKED ISSUE #2818: OTLP Exporter - treat code based config as final
A specific issue is opened to track making [this change ](https://github.com/open-telemetry/opentelemetry-rust/issues/1225)for OTLP Exporters.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/http/mod.rs#L371 Looks like we are treating ENV variable as higher priority, so that need to be fixed everywhere.

gruebel:
Fixes #2818 

## Changes

- had to make the `timeout` field optional to properly handle code-based config prefer over env vars

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -18,10 +18,11 @@
  new one returns specific variants applicable to building an exporter. Some
  variants might be applicable only on select features.
 - **Breaking** `ExportConfig`'s `timeout` field is now optional(`Option<Duration>`)
-- **Breaking** Configuration done via code is final. ENV variables cannot be used to override the code config. Do not use code based config, if there is desire to control the settings via ENV variables.
- List of ENV variables and corresponding setting being affected by this change.
-  priority changed in favor of compile time values. Additionally the `endpoint`
-  field prioritizes compile time values over environment variables.
+- **Breaking** Export configuration done via code is final. ENV variables cannot be used to override the code config.
+  Do not use code based config, if there is desire to control the settings via ENV variables.
+  List of ENV variables and corresponding setting being affected by this change.
+  - `OTEL_EXPORTER_OTLP_ENDPOINT` -> `ExportConfig.endpoint`
+  - `OTEL_EXPORTER_OTLP_TIMEOUT` -> `ExportConfig.timeout`
 
 ## 0.28.0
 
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -76,19 +77,19 @@ pub struct ExportConfig {
     pub protocol: Protocol,
 
     /// The timeout to the collector.
-    pub timeout: Duration,
+    pub timeout: Option<Duration>,
```
cijothomas, line 80: strange that we exposed this struct without a builder!
Can you add doc comments to indicate configuring via this would be final, and cannot be overridden by env?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -250,7 +251,7 @@ impl<B: HasExportConfig> WithExportConfig for B {
     }
 
     fn with_timeout(mut self, timeout: Duration) -> Self {
-        self.export_config().timeout = timeout;
+        self.export_config().timeout = Some(timeout);
```
cijothomas, line 254: Please modify the doc comment for all these methods to indicate that they are final and won't be overrrideable via ENV.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
  `Error` which contained many variants unrelated to building an exporter, the
  new one returns specific variants applicable to building an exporter. Some
  variants might be applicable only on select features.
+- *Breaking* `ExportConfig`'s `timeout` field is now optional, and the order of
```
cijothomas, line 20:
```suggestion
- **Breaking** `ExportConfig`'s `timeout` field is now optional(`Option<Duration>`)
- **Breaking** Configuration done via code is final. ENV variables cannot be used to override the code config. Do not use code based config, if there is desire to control the settings via ENV variables.
 List of ENV variables and corresponding setting being affected by this change.
```

Wording suggestion to make it easy to spot the changes from changelog.

cijothomas:
@TommyCpp Could you also double check? I recall you suggested to add a `with_env_preference` like API to make it easier to migrate, but it looks like we preferred ENV over code only for OTLP, and that too for 2-3 settings, which make me thing there is no need of additional convenience method to revert to old behavior.. Please take a look and see if this is okay!

TommyCpp:
> but it looks like we preferred ENV over code only for OTLP, and that too for 2-3 settings, which make me think there is no need of additional convenience method to revert to old behavior

Yep if that's the case then sounds fair to me

PR #2813, opened 2025-03-17
ci: Use bare-metal OTEL runners for benchmark CI 

scottgerring:
Fixes #
Design discussion issue (if applicable) #

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #2832, opened 2025-03-19
ci: Use bare-metal OTEL runners for benchmark CI for pushes to main

scottgerring:
Supersedes #2813. This PR triggers our benchmark suite on pushes to main, regressing the performance against the previous commit on main. 

## Changes

Following on from https://github.com/open-telemetry/community/issues/2616 , this PR moves our criterion suite to use the self-hosted OTEL runners for pushes to `main`, whilst continuing to use the ubuntu shared runners for PRs, on request, when labeling a PR with `performance`. I've also renamed the workflow to `benchmark` as its no longer only tied to the PR workflow.

**Why?**
_Opting in_ to PR regressions against main gives us an easy way to check the performance of a PR we're interested in against main - e.g. if it looks performance-impacting, one way or another. This avoids the "works on my machine" ad-hoc re-running of the benchmark suite.

Adding performance regression on the dedicated runners for commits to main will give us a stable, long-term journal of performance impact of changes. 

**What else?**
This looks more intense than it is. Really there's just a bit of logic added to switch 1/ the worker pool and 2/ the "point of comparison" if we're running on a push to main. I've tested these bits work by inverting the logic and pushing. 


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,58 @@
...
+jobs:
+  runBenchmark:
+    name: run benchmark
+    permissions:
+      pull-requests: write
+
+    # If we're running on a PR, use ubuntu-latest - a shared runner. We can't use the self-hosted
+    # runners on arbitrary PRs, and we don't want to unleash that load on the pool anyway.     
+    # If we're running on main, use the OTEL self-hosted runner pool. 
+    runs-on: ${{ github.event_name == 'pull_request' && 'ubuntu-latest' || 'self-hosted' }}
```
scottgerring, line 25: Switch between the self-hosted (labelled PR) and ubuntu-latest (main)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,58 @@
...
+    if: |
+      ${{
```
scottgerring, line 27: This maintains our filter for the `performance` label being required to fire on a PR

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: ${{ github.event_name == 'pull_request' && 'ubuntu-latest' || 'self-hosted' }}
+    if: |
+      ${{ 
+        github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'performance') ||
+        github.event_name == 'push'
+      }}
+    env:
+      # For PRs, compare against the base branch - e.g., 'main'. 
+      # For pushes to main, compare against the previous commit
+      BRANCH_NAME: ${{ github.event_name == 'pull_request' && github.base_ref || github.event.before }}
```
scottgerring, line 34: This switches to compare to the previous commit if we are building for a push to main.

For PRs, we compare to the current state of main

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,58 @@
+# This workflow runs a Criterion benchmark on a PR and compares the results against the base branch.
```
scottgerring, line 1: Extra review facts:

* This is actually a rename-and-change of pr_criterion, but git has got confused
* For the branches that behave differently on push-to-main, i've tested they work by inverting the logic and pushing. I am confident the "template magic" is good.

cijothomas (approved):
Thanks for the detailed comments!

scottgerring:
> Thanks for the detailed comments!

I'm nothing if not wordy 😆

scottgerring:
I'll need to get the build rule updated on the runners on the issue, now, before this will work on main. On it.

PR #2833, opened 2025-03-19
chore: context stabilization tests & logging pieces

### LINKED ISSUE #2768: Context - Stabilization
A parent issue to track the stabilization of `Context`. We've just put a bunch of effort into stabilising the implementation so it can be used to support robust integration with tokio-tracing, and I reckon it makes sense to capitalize on this effort and "fresh knowledge" to make sure the impl matches the spec and there's nothing else in there that's going to jump out at us after we stabilize the API. 

An initial review of the implementation [against the spec](https://opentelemetry.io/docs/specs/otel/context/) will follow on this issue, and we'll fork out sub-issues for anything that needs more attention.

scottgerring:
Fixes #2768 

## Changes


opentelemetry/src/context.rs
```diff
@@ -460,6 +460,15 @@ impl ContextStack {
             // The empty context is always at the bottom of the [`ContextStack`]
             // and cannot be popped, and the overflow position is invalid, so do
             // nothing.
+            otel_warn!(
+                name: "Context.PopInvalidPosition",
```
cijothomas, line 464:
```suggestion
                name: "Context.OutOfOrderDrop",
```

Trying to make this more user-friendly, see if this is better. (push/pop is inner implementation mostly...)

opentelemetry/src/context.rs
```diff
@@ -659,14 +676,186 @@ mod tests {
...
+    /// Tests that:
+    /// 1. Parent context values are properly propagated to async operations
+    /// 2. Values added during async operations do not affect parent context
+    #[tokio::test]
+    async fn test_async_context_propagation() {
+        // Set up initial context with ValueA
+        let parent_cx = Context::new().with_value(ValueA(42));
+
+        // Create and run async operation with the parent context
+        async {
```
cijothomas, line 790: nit: maybe easier to read if this is a separate async method, and called like `method_async..with_context(parent_cx).await;`, so it is easy to spot that context is flowing to async due to explicit action, not automatic!

opentelemetry/src/context.rs
```diff
@@ -659,14 +676,186 @@ mod tests {
...
+        }
+        .with_context(parent_cx.clone()) // Propagate the parent context to the async operation
+        .await;
+
+        // After async operation completes:
+        // 1. Parent context should be unchanged
+        assert_eq!(
+            parent_cx.get::<ValueA>(),
+            Some(&ValueA(42)),
+            "Parent context should be unchanged"
```
cijothomas, line 841: nit: Make it more colorful, by having the async block insert ValueA(43), and assert here that those changes are not propagated back here.

opentelemetry/src/context.rs
scottgerring: I prefer this wording too; updated

opentelemetry/src/context.rs
scottgerring: done!

opentelemetry/src/context.rs
```diff
@@ -460,6 +460,15 @@ impl ContextStack {
             // The empty context is always at the bottom of the [`ContextStack`]
             // and cannot be popped, and the overflow position is invalid, so do
             // nothing.
+            otel_warn!(
+                name: "Context.OutOfOrderDrop",
+                position = pos,
+                message = if pos == ContextStack::BASE_POS {
+                    "Attempted to pop the base context which is not allowed"
```
scottgerring, line 467: I don't believe codecov is right; the test I added with out of order async drops hits this line in the debugger. Weird.

bantonsson: I think the issue is that the tests run with the logging macros evaluating to nothing, so there is nothing to run there.

opentelemetry/src/context.rs
```diff
@@ -653,20 +670,226 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_out_of_order_context_detachment_futures() {
+        // This function returns a future, but doesn't await it
+        // It will complete before the future that it creates.
+        async fn create_a_future() -> impl std::future::Future<Output = ()> {
+            // Create a future that will do some work, referencing our current
+            // context, but don't await it.
+            async {
```
bantonsson, line 872: I think this block could start with asserting that we have `ValueA(42)` from the outer context.

opentelemetry/src/context.rs
```diff
@@ -653,20 +670,226 @@ mod tests {
...
+        }
+
+        // Create our base context
+        let parent_cx = Context::new().with_value(ValueA(42));
+
+        // await our nested function, which will create and detach a context
+        let future = create_a_future().with_context(parent_cx).await;
+
+        // Execute the future. The completion of the future will result in an out-of-order drop,
+        // as the parent context created by create_a_future was already dropped earlier.
```
bantonsson, line 887: Are we really sure that this will be an out of order drop? I mean the _inner_ future creates a copy of the `Context` in the _outer_ future, that is pushed onto the stack of whichever thread is executing the _inner_ future, which doesn't have to be the same one as executed the _outer_ future, or?

The test is perfectly fine, but I think the comment is wrong.

bantonsson (commented):
Looks good to me, only some comments about the `test_out_of_order_context_detachment_futures`.

opentelemetry/src/context.rs
scottgerring: @bantonsson I'm now a bit less convinced than I was before - let me know if i'm missing something

opentelemetry/src/context.rs
scottgerring: oooh good catch.
I've jazzed up the macros so they print to stdout in test. You need to `--nocapture` cargo test to see this, so by default it goes to /dev/null unless you want to see it.

opentelemetry/src/global/internal_logging.rs
```diff
@@ -4,6 +4,9 @@
 /// **internally within OpenTelemetry code** or for **custom exporters, processors and other plugins**. They are not designed
```
scottgerring, line 4: Changes here cause the macros to print to stdout if you are running in a test. I find it super convenient to see internal logging when debugging tests.  You need to opt into this with `--nocapture` when you run them, otherwise the output goes to `/dev/null`.

opentelemetry/src/context.rs
bantonsson: Yes, but you have still made a copy of the _outer_ `Context` that _lives_ with the _inner_ future and gets activated when that future is awaited. There won't be any out of order here as far as I can tell.

scottgerring: 💡 Ah, I grok it now. I had a strange mental model of the relationship between the thread local and the `Context` instances. Looking at what `FutureExt` is doing made it clear. Thanks!

scottgerring: Actioned: i've left the test but made it clearer what it is testing - weird parent->child relationships with futures.

opentelemetry/src/context.rs
scottgerring: ... and as predicted this fixes the coverage lint.

bantonsson (approved):

opentelemetry/src/global/internal_logging.rs
```diff
@@ -25,7 +28,13 @@ macro_rules! otel_info {
         {
             tracing::info!( name: $name, target: env!("CARGO_PKG_NAME"), name = $name, "");
         }
-        #[cfg(not(feature = "internal-logs"))]
+
+        #[cfg(test)]
```
cijothomas, line 32: this is a good idea. Let us do this in a separate PR, as there are some tests which setup fmt layer already, and we need to discontinue that to avoid the duplication to stdout.

cijothomas (approved):

opentelemetry/src/global/internal_logging.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2841 opened to check this.

scottgerring: @cijothomas great - thanks! I'll chase this up.

PR #2834, opened 2025-03-19
ci: fix filter for benchmark pushes on branch 

scottgerring:
## Changes
Something wasn't quite write with the filter for the branch runs; simplifying it into a single line seems to do the trick.
Have tested with and without label and validated it fires and doesn't fire accordingly. 


scottgerring:
@cijothomas sorry about that! The previous PR seems to trigger for branches regardless of label. This slight reformatting of the condition into a single line addresses that.

cijothomas (approved):

PR #2836, opened 2025-03-19
test: validate code-based config is preferred over env vars in ZipkinExporterBuilder

### LINKED ISSUE #1225: Discussion: priority on environment variable vs compiling time value

gruebel:
Relates to #1225 

## Changes

- I couldn't find a way to validate the timeout env var prioritization

### DIFF
--- a/opentelemetry-zipkin/src/exporter/mod.rs
+++ b/opentelemetry-zipkin/src/exporter/mod.rs
@@ -93,6 +93,9 @@ impl ZipkinExporterBuilder {
     }
 
     /// Assign client implementation
+    ///
+    /// Note: Programmatically setting the timeout will override any value
+    /// set via the environment variable `OTEL_EXPORTER_ZIPKIN_TIMEOUT`.
     pub fn with_http_client<T: HttpClient + 'static>(mut self, client: T) -> Self {
         self.client = Some(Arc::new(client));
         self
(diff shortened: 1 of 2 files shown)


opentelemetry-zipkin/src/exporter/mod.rs
```diff
@@ -93,6 +93,9 @@ impl ZipkinExporterBuilder {
     }
 
     /// Assign client implementation
+    ///
+    /// Note: Programmatically setting the timeout will override any value
+    /// set via the environment variable `OTEL_EXPORTER_ZIPKIN_TIMEOUT`.
```
cijothomas, line 98: this comment is for timeout, but this method is for setting client?

gruebel: correct, you can only set the timeout manually, if you create a client, because the timeout is used during the client creation.

cijothomas: can you modify the comment to make it clearer?

_when using `with_http_client`, the timeout will depend on the provided client implementation, and may not respect timeout set via the environment variable `OTEL_EXPORTER_ZIPKIN_TIMEOUT`._

cijothomas (approved):
Left a non-blocking comment about improving doc comment.

PR #2838, opened 2025-03-20
fix: Error cleanups continued

cijothomas:


opentelemetry-sdk/src/logs/in_memory_exporter.rs
```diff
@@ -157,14 +155,9 @@ impl InMemoryLogExporter {
...
-    pub fn get_emitted_logs(&self) -> LogResult<Vec<LogDataWithResource>> {
-        let logs_guard = self
-            .logs
-            .lock()
-            .map_err(|e| OTelSdkError::InternalFailure(format!("Failed to lock logs: {}", e)))?;
-        let resource_guard = self.resource.lock().map_err(|e| {
-            OTelSdkError::InternalFailure(format!("Failed to lock resource: {}", e))
-        })?;
+    pub fn get_emitted_logs(&self) -> Result<Vec<LogDataWithResource>, String> {
+        let logs_guard = self.logs.lock().map_err(|e| e.to_string())?;
```
scottgerring, line 159: If we use `anyhow` and an error type we'd also be able to use `[From]` to get rid of the explicit  to-stringing here

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -287,7 +287,7 @@ impl OtlpHttpClient {
     fn build_trace_export_body(
         &self,
         spans: Vec<SpanData>,
-    ) -> opentelemetry_sdk::trace::TraceResult<(Vec<u8>, &'static str)> {
+    ) -> Result<(Vec<u8>, &'static str), String> {
```
scottgerring, line 290: I don't know how strongly I feel about this, but it smells a bit to me to have `Result<T, String>` and not `Result<T, SomeEnumErrorType` where the thing implements the `Error` trait (presumably with anyhow) , _even though the error type would wrap a single variant_.

This would of course make no difference in the behaviour in this instance, but it would look more consistent, and mean we continue to only return `Error`-implementing things in `Result`.

opentelemetry-sdk/src/logs/in_memory_exporter.rs
```diff
@@ -157,14 +155,9 @@ impl InMemoryLogExporter {
...
+        let resource_guard = self.resource.lock().map_err(|e| e.to_string())?;
```
scottgerring, line 160: If we use `anyhow` and an error type we'd also be able to use `[From]` to get rid of the explicit  to-stringing here

scottgerring (commented):
Overall I think this is a good idea - there's not much value in propagating 1 million error variants back to a caller that ignores them - but I wonder if we shouldn't try and use `Error` trait impls in everything that returns errors for consistency and clarity?

opentelemetry-otlp/src/exporter/http/mod.rs
cijothomas: This is just internal helper functions, so okay with this. But for public API, lets ensure Result<T, SomeError> will be returned always.

opentelemetry-sdk/src/logs/in_memory_exporter.rs
cijothomas: done in latest commit

cijothomas:
> Overall I think this is a good idea - there's not much value in propagating 1 million error variants back to a caller that ignores them - but I wonder if we shouldn't try and use `Error` trait impls in everything that returns errors for consistency and clarity?

done.

lalitb (approved):

PR #2839, opened 2025-03-20
fix: Cleanup zipkin build error

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2790 for Zipkin


gruebel (approved):
👍

scottgerring (approved):
LGTM.
We should probably reflect on `codecov`'s config, as we find ourselves ignoring it pretty regularly.

PR #2837, opened 2025-03-20
chore: Prepare docs before 0.29

cijothomas:

PR #2843, opened 2025-03-21
fix: KeyValueMetadata to not expose its fields public

cijothomas:
Don't think public was required, and this should be transparent to users anyway...
https://github.com/open-telemetry/opentelemetry-rust/pull/2837/files#r2008002156


lalitb (approved):

gruebel (approved):
👍

PR #2854, opened 2025-03-23
feat: Add shutdown with timeout for metric exporter

mohammadVatandoost:
https://github.com/open-telemetry/opentelemetry-rust/issues/2574 Part 1: Adding shutdown_with_timeout for metric exporter


opentelemetry-otlp/src/metric.rs
```diff
@@ -163,6 +164,10 @@ impl PushMetricExporter for MetricExporter {
     }
 
     fn shutdown(&self) -> OTelSdkResult {
+        self.shutdown_with_timeout(Duration::from_secs(5))
```
TommyCpp, line 167: Is it from spec? If so please link the document here

TommyCpp (approved):

opentelemetry-otlp/src/metric.rs
mohammadVatandoost: The 5 second is used in other functions as default timeout. like:
https://github.com/open-telemetry/opentelemetry-rust/blob/a071d8fc3906da56e089061927fbc1561e0e7677/opentelemetry-sdk/src/trace/span_processor.rs#L425

I don't know why we choose 5 seconds. As it is written in ToDo, I think we need to read it from configuration.
I think this is not related to issue of the PR. we can work on it at another PR.

cijothomas: There is no spec on what should be the default, so 5 seconds is our invention. We should recommend users to use the shutdown_with_timeout passing a Duration of their choice, but we also do shutdown() from Drop, where we need to pick some default. 5 secs seems reasonable to me as a default.

opentelemetry-sdk/src/metrics/exporter.rs
```diff
@@ -26,6 +28,9 @@ pub trait PushMetricExporter: Send + Sync + 'static {
     ///
     /// After Shutdown is called, calls to Export will perform no operation and
     /// instead will return an error indicating the shutdown state.
+    fn shutdown_with_timeout(&self, timeout: Duration) -> OTelSdkResult;
```
cijothomas, line 31: now that we have both, lets give a default implementation for the shutdown(), invoking the new one passing 5 seconds.

Also please update changelog to reflect this - this is breaking change for custom MetricExporter author.

opentelemetry-otlp/src/metric.rs
mohammadVatandoost: Done

opentelemetry-sdk/src/metrics/exporter.rs
mohammadVatandoost: Done

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ Released 2025-Mar-21
 
   Custom exporters will need to internally synchronize any mutable state, if applicable.
 
+- **Breaking** The `shutdown_with_timeout` method is added to MetricExporter trait. This is breaking change for a custom MetricExporter.
```
cijothomas, line 71:
```suggestion
- **Breaking** The `shutdown_with_timeout` method is added to MetricExporter trait. This is breaking change for custom `MetricExporter` authors.
```

cijothomas (approved):

lalitb (approved):

PR #2855, opened 2025-03-23
docs: Modify example to use logs, baggage

cijothomas:
Eventually, I expect that we should just need one example (or two - one for console apps, and one web app) that'd showcase each signals. This PR enhances the existing web examples to also use logs. It also showcases Baggage and composite propagators.
Only a small step - expecting many more changes to this examples next

### DIFF
--- a/examples/tracing-http-propagator/Cargo.toml
+++ b/examples/tracing-http-propagator/Cargo.toml
@@ -23,5 +23,8 @@ tokio = { workspace = true, features = ["full"] }
 opentelemetry = { path = "../../opentelemetry" }
 opentelemetry_sdk = { path = "../../opentelemetry-sdk" }
 opentelemetry-http = { path = "../../opentelemetry-http" }
-opentelemetry-stdout = { workspace = true, features = ["trace"] }
+opentelemetry-stdout = { workspace = true, features = ["trace", "logs"] }
 opentelemetry-semantic-conventions = { path = "../../opentelemetry-semantic-conventions" }
+opentelemetry-appender-tracing = { workspace = true }
+tracing = { workspace = true, features = ["std"]}
+tracing-subscriber = { workspace = true, features = ["env-filter","registry", "std", "fmt"] }
--- a/examples/tracing-http-propagator/src/client.rs
+++ b/examples/tracing-http-propagator/src/client.rs
@@ -3,13 +3,18 @@ use hyper_util::{client::legacy::Client, rt::TokioExecutor};
 use opentelemetry::{
     global,
     trace::{SpanKind, TraceContextExt, Tracer},
-    Context, KeyValue,
+    Context,
 };
+use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
 use opentelemetry_http::{Bytes, HeaderInjector};
-use opentelemetry_sdk::{propagation::TraceContextPropagator, trace::SdkTracerProvider};
-use opentelemetry_stdout::SpanExporter;
+use opentelemetry_sdk::{
+    logs::SdkLoggerProvider, propagation::TraceContextPropagator, trace::SdkTracerProvider,
+};
+use opentelemetry_stdout::{LogExporter, SpanExporter};
+use tracing::info;
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
-fn init_tracer() {
+fn init_tracer() -> SdkTracerProvider {
     global::set_text_map_propagator(TraceContextPropagator::new());
     // Install stdout exporter pipeline to be able to retrieve the collected spans.
     // For the demonstration, use `Sampler::AlwaysOn` sampler to sample all traces.
@@ -17,7 +22,23 @@ fn init_tracer() {
         .with_simple_exporter(SpanExporter::default())
         .build();
 
-    global::set_tracer_provider(provider);
+    global::set_tracer_provider(provider.clone());
+    provider
+}
+
+fn init_logs() -> SdkLoggerProvider {
+    // Setup tracerprovider with stdout exporter
+    // that prints the spans to stdout.
+    let logger_provider = SdkLoggerProvider::builder()
+        .with_simple_exporter(LogExporter::default())
+        .build();
+    let otel_layer = OpenTelemetryTracingBridge::new(&logger_provider);
+    tracing_subscriber::registry()
+        .with(otel_layer)
+        .with(tracing_subscriber::filter::LevelFilter::INFO)
+        .init();
+
+    logger_provider
 }
 
 async fn send_request(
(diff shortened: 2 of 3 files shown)


examples/tracing-http-propagator/src/client.rs
```diff
@@ -37,21 +58,22 @@ async fn send_request(
     global::get_text_map_propagator(|propagator| {
         propagator.inject_context(&cx, &mut HeaderInjector(req.headers_mut().unwrap()))
     });
+    req.headers_mut()
+        .unwrap()
+        .insert("baggage", "is_synthetic=true".parse().unwrap());
```
cijothomas, line 63: hardcoded instead of putting it to context and then letting the propagator do this job. In future, we can modify this.

examples/tracing-http-propagator/src/client.rs
```diff
@@ -3,21 +3,42 @@ use hyper_util::{client::legacy::Client, rt::TokioExecutor};
...
     let provider = SdkTracerProvider::builder()
         .with_simple_exporter(SpanExporter::default())
         .build();
 
-    global::set_tracer_provider(provider);
+    global::set_tracer_provider(provider.clone());
+    provider
+}
+
+fn init_logs() -> SdkLoggerProvider {
```
scottgerring, line 29: Semi-related - is there _prior art_ in other OTel's to provide opinionated "configure everything" helpers? 

E.g. if I common-case just want to send logs/traces/metrics to some OTLP endpoint. With some carefully thought-out builder and escape hatches this could make a much simpler getting-started experience.

examples/tracing-http-propagator/src/client.rs
```diff
@@ -37,21 +58,22 @@ async fn send_request(
     global::get_text_map_propagator(|propagator| {
         propagator.inject_context(&cx, &mut HeaderInjector(req.headers_mut().unwrap()))
     });
+    req.headers_mut()
```
scottgerring, line 61: Could use the `.header(...)` method on the builder returned on line 57

examples/tracing-http-propagator/src/client.rs
```diff
@@ -66,5 +88,11 @@ async fn main() -> std::result::Result<(), Box<dyn std::error::Error + Send + Sy
     )
     .await?;
 
+    tracer_provider
```
scottgerring, line 91: How do we feel about using anyhow in examples? On one hand it would be the typical way of doing this in the community, on the other it's an extra dep.

examples/tracing-http-propagator/src/server.rs
```diff
@@ -30,11 +41,11 @@ async fn handle_health_check(
     _req: Request<Incoming>,
 ) -> Result<Response<BoxBody<Bytes, hyper::Error>>, Infallible> {
     let tracer = get_tracer();
-    let mut span = tracer
+    let _span = tracer
```
scottgerring, line 44: I think it would be worth calling out why you haven't assigned this to `_` (presumably because you don't want it to be dropped until the end of the function?)

examples/tracing-http-propagator/src/server.rs
```diff
@@ -50,11 +61,11 @@ async fn handle_echo(
     req: Request<Incoming>,
 ) -> Result<Response<BoxBody<Bytes, hyper::Error>>, Infallible> {
     let tracer = get_tracer();
-    let mut span = tracer
+    let _span = tracer
```
scottgerring, line 64: and here (not using discard)

examples/tracing-http-propagator/src/server.rs
```diff
@@ -66,17 +77,20 @@ async fn router(
 ) -> Result<Response<BoxBody<Bytes, hyper::Error>>, Infallible> {
     // Extract the context from the incoming request headers
     let parent_cx = extract_context_from_request(&req);
+    for bag in parent_cx.baggage() {
+        println!("Baggage: {:?} = {:?}", bag.0, bag.1);
```
scottgerring, line 81: why not use the logger?

examples/tracing-http-propagator/src/server.rs
```diff
@@ -93,24 +107,86 @@ async fn router(
     response
 }
 
+#[derive(Debug)]
```
scottgerring, line 110:
```suggestion
///
/// This LogProcessor allows us to print all of our baggage
// alongside each log message, which is helpful for our demo.
///
#[derive(Debug)]
```

examples/tracing-http-propagator/Cargo.toml
```diff
@@ -23,5 +23,8 @@ tokio = { workspace = true, features = ["full"] }
 opentelemetry = { path = "../../opentelemetry" }
 opentelemetry_sdk = { path = "../../opentelemetry-sdk" }
 opentelemetry-http = { path = "../../opentelemetry-http" }
-opentelemetry-stdout = { workspace = true, features = ["trace"] }
+opentelemetry-stdout = { workspace = true, features = ["trace", "logs"] }
```
scottgerring, line 26: This isn't related to this PR, but I feel a bit funny about using workspace deps _in the examples_, as it prevents the user from simply copying the example and using it.

scottgerring (changes requested):
Couple of minor lints and some questions!

examples/tracing-http-propagator/Cargo.toml
cijothomas: once we move docs to opentelemetry.io, this would be proper examples that can be just copy-pasted.

examples/tracing-http-propagator/src/client.rs
cijothomas: Yes, the concept of "Distro" exists. A lot of it are vendor specific, but once we have all 3 signals ready, we should also have a one-package+one-liner to configure the most reasonable defaults for everything.

examples/tracing-http-propagator/src/server.rs
cijothomas: this was a leftover from my local testing, will remove!

examples/tracing-http-propagator/src/server.rs
cijothomas: `_` is fine here. modified.

examples/tracing-http-propagator/src/client.rs
cijothomas: neutral on that. I want to ensure we are consistent throughout the examples. (which we are not today, but I expect to get to that stage soon)

examples/tracing-http-propagator/src/server.rs
```diff
@@ -30,11 +41,11 @@ async fn handle_health_check(
     _req: Request<Incoming>,
 ) -> Result<Response<BoxBody<Bytes, hyper::Error>>, Infallible> {
     let tracer = get_tracer();
-    let mut span = tracer
+    let _ = tracer
```
TommyCpp, line 44: Doesn't this change means the span will be dropped immediately on this line? The span time will probably be off as it's no longer the execution time of the function

cijothomas: Thanks! fixed.

examples/tracing-http-propagator/src/client.rs
scottgerring: Is "export everything to OTLP" essentially a default distro then?

scottgerring (approved):
LGTM!

examples/tracing-http-propagator/src/client.rs
cijothomas: Could be.. We don't have that yet.

PR #2857, opened 2025-03-24
refactor: AggregatedMetrics as enum instead of dyn Aggregation

fraillt:
## Changes

* Introduced a new `AggregatedMetrics` enum, which store data reported by Instruments.
* Change `Metric::data` field from `Box<dyn Aggregation>` to `AggregatedMetrics`.
* `ResourceMetrics` is now `Clone`, because `AggregatedMetrics` is a simple enum that can be easily `cloned`.

### DIFF
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -9,7 +9,7 @@ use crate::Resource;
 use super::Temporality;
 
 /// A collection of [ScopeMetrics] and the associated [Resource] that created them.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct ResourceMetrics {
     /// The entity that collected the metrics.
     pub resource: Resource,
@@ -18,7 +18,7 @@ pub struct ResourceMetrics {
 }
 
 /// A collection of metrics produced by a meter.
-#[derive(Default, Clone, Debug)]
+#[derive(Default, Debug)]
 pub struct ScopeMetrics {
     /// The [InstrumentationScope] that the meter was created with.
     pub scope: InstrumentationScope,
@@ -29,7 +29,7 @@ pub struct ScopeMetrics {
 /// A collection of one or more aggregated time series from an [Instrument].
 ///
 /// [Instrument]: crate::metrics::Instrument
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Metric {
     /// The name of the instrument that created this data.
     pub name: Cow<'static, str>,
@@ -42,7 +42,7 @@ pub struct Metric {
 }
 
 /// Aggregated metrics data from an instrument
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum AggregatedMetrics {
     /// All metric data with `f64` value type
     F64(MetricData<f64>),
@@ -53,7 +53,7 @@ pub enum AggregatedMetrics {
 }
 
 /// Metric data for all types
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum MetricData<T> {
     /// Metric data for Gauge
     Gauge(Gauge<T>),
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -1,6 +1,6 @@
 //! Types for delivery of pre-aggregated metric time series data.
 
-use std::{any, borrow::Cow, fmt, time::SystemTime};
+use std::{borrow::Cow, time::SystemTime};
 
 use opentelemetry::{InstrumentationScope, KeyValue};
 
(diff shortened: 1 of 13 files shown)


opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -38,23 +38,77 @@ pub struct Metric {
...
-    fn as_any(&self) -> &dyn any::Any;
-    /// Support downcasting during aggregation
-    fn as_mut(&mut self) -> &mut dyn any::Any;
+/// Aggregated metrics data from an instrument
+#[derive(Debug, Clone)]
+pub enum AggregatedMetrics {
+    /// All metric data with `f64` value type
+    F64(MetricData<f64>),
+    /// All metric data with `u64` value type
+    U64(MetricData<u64>),
```
cijothomas, line 50: this shouldn't exist. Only f64 and i64 should ever come out after aggregation.

fraillt: .. but `InstrumentProvider` has `u64_counter`, `u64_gauge`, `u64_histogram`, etc...

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -9,7 +9,7 @@ use crate::Resource;
 use super::Temporality;
 
 /// A collection of [ScopeMetrics] and the associated [Resource] that created them.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
cijothomas, line 12: This could be refactored out! We should not have to pass Resource to exporters during each export. It can be sent just once at build time. Exporters just need scope+aggregated_metrics.
Logs already follow this model of sending Resource once. We can (and we should) follow it for Metrics and Traces too.
(a separate refactor not this PR, to keep scope contained)

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -18,7 +18,7 @@ pub struct ResourceMetrics {
 }
 
 /// A collection of metrics produced by a meter.
-#[derive(Default, Debug)]
+#[derive(Default, Clone, Debug)]
```
cijothomas, line 21: what is the need to do Clone? Is it just for InMemoryExporter's need ?

fraillt: For the most part yes :)
Previously we had `Box<dyn Aggregation>` which prevented other types from being cloneable as well.
Since AggregatedMetrics is simple enum, I made other types cloneable as well.
I'm not sure if it is valuable to have `Clone` on these types, but given that end users will use them, maybe someone will find it useful :)

cijothomas: I see. Then my suggestion is *not* to implement Clone now (or do it when testing feature is enabled only so only InMemoryExporter can get it). Adding clone in future is always possible, but reversing that decision is not allowed as it'll be breaking change.

(Exporters usually export the metric in the export() call itself, and even if it gets stored for retry etc. it'll be in a serialized state, so I don't think Clone is needed)

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -38,23 +38,77 @@ pub struct Metric {
...
-/// It will be one of: [Gauge], [Sum], or [Histogram].
-///
-/// [Instrument]: crate::metrics::Instrument
-pub trait Aggregation: fmt::Debug + any::Any + Send + Sync {
-    /// Support downcasting
-    fn as_any(&self) -> &dyn any::Any;
-    /// Support downcasting during aggregation
-    fn as_mut(&mut self) -> &mut dyn any::Any;
+/// Aggregated metrics data from an instrument
+#[derive(Debug, Clone)]
```
cijothomas, line 45: See if you like this better:
https://github.com/open-telemetry/opentelemetry-rust/compare/main...cijothomas:opentelemetry-rust:cijothomas/new-metrics#diff-041ac5cb4c3201f64e52be34cdba07e769e2b3637d605ddf59852a66ef104762R7-R70

opentelemetry-sdk/src/metrics/data/mod.rs
fraillt: There are no conversions happening at the moment, e.g. in case a user creates `u64_counter`, we'll return `MetricData<u64>`. I'm not sure at which point we should convert from `u64` to `i64`.
Maybe let's leave this for another discussion?

opentelemetry-sdk/src/metrics/data/mod.rs
fraillt: `Clone` removed.

opentelemetry-sdk/src/metrics/data/mod.rs
cijothomas: yes lets follow up separately.

cijothomas (approved):
LGTM. We can followup on the open comments (Moving resource out, removing i64 in aggregation etc.)

PR #2867, opened 2025-03-26
refactor: re-export tracing for internal-logs

gruebel:
## Changes

- also no need anymore to exclude `tracing` from `machete`


opentelemetry/src/lib.rs
```diff
@@ -299,3 +299,6 @@ pub mod time {
         SystemTime::UNIX_EPOCH + std::time::Duration::from_millis(js_sys::Date::now() as u64)
     }
 }
+
+#[cfg(feature = "internal-logs")]
+pub use tracing::{debug, error, info, warn}; // re-export
```
cijothomas, line 304: @lalitb can you share your thoughts on this? We tried to "hide" tracing, so that Otel is not tied to any particular logging library...Will this affect that ? Are there better alternatives to achieve the goals?

lalitb: Yes, the approach here will expose tracing macros directly, and make future changes difficult. I feel it is better to let the other crates take dependency on tracing instead.  I feel it is better to let the other crates take dependency on tracing instead. Or if we really want to use this approach, try to hide this under the _private module pattern, something like:
```rust
#[doc(hidden)]
pub mod _private {
    #[cfg(feature = "internal-logs")]
    pub use tracing;
}
#[macro_export]
macro_rules! log_error {
    ($($arg:tt)*) => {
        #[cfg(feature = "internal-logs")]
        ::opentelemetry::_private::tracing::error!($($arg)*);
    }
}
```

The other crates can still use `_private.tracing`, but this is still better than directly exposing the tracing macros. It at least communicates through naming that users shouldn't rely on this implementation detail.

opentelemetry-proto/Cargo.toml
```diff
@@ -44,7 +44,7 @@ zpages = ["trace"]
 testing = ["opentelemetry/testing"]
 
 # add ons
-internal-logs = ["tracing"]
+internal-logs = []
```
bantonsson, line 47: So there doesn't seem to be any mention of `internal-logs` anywhere in the crate. Why is there such a feature here?

lalitb: The "internal-logs" feature is used by internal logging macros defined in the "opentelemetry" crate. These macros expand at compile time, injecting the corresponding logging code into crates that use them and so this feature needs to be defined in those crates too. Ideally, it would be good if this can be avoided, while still allowing individual crates to independently enable or disable logging.

bantonsson: @lalitb I know exactly what the feature does in the `opentelemetry` crate, but the fact that it has no dependencies here and is not referenced in code makes me think that it’s not needed in this crate.

gruebel: I think the configuration is/was actually incorrect, it should include the reference to the identically named feature in the `opentelemetry` crate.

lalitb: @bantonsson i believe it is being referenced indirectly [here](https://github.com/open-telemetry/opentelemetry-rust/blob/99cb67d19ca6ad4df9a87cb31473a130827740c9/opentelemetry-proto/src/transform/metrics.rs#L10)

opentelemetry/src/lib.rs
```diff
@@ -299,3 +299,9 @@ pub mod time {
         SystemTime::UNIX_EPOCH + std::time::Duration::from_millis(js_sys::Date::now() as u64)
     }
 }
+
+#[doc(hidden)]
+pub mod _private {
+    #[cfg(feature = "internal-logs")]
+    pub use tracing; // re-export
```
cijothomas, line 306: should this expose entire tracing module or selected components we care about?

gruebel: I exported the macros we actually use now. at the end we can always change it. it is not meant to be used directly by someone else than us.

opentelemetry-proto/Cargo.toml
bantonsson: Indeed it is. Then the omission of `internal-logs = ["opentelemetry/internal-logs"]`  wasn't caught by CI before since we didn't test the features separately.

bantonsson (approved):

cijothomas (approved):

PR #2868, opened 2025-03-27
feat: Leverage Suppression Context in Sdk

cijothomas:
Fixes: https://github.com/open-telemetry/opentelemetry-rust/issues/761
Continuing on top of https://github.com/open-telemetry/opentelemetry-rust/pull/2821 , this PR modifies SDK's own components to suppress telemetry from its own operation using `Context::enter_telemetry_suppressed_scope`. This is done in the dedicated thread created by BatchProcessors, PeriodicReader.

Also modified Log SDK to respect this and ignore logs if under suppression.
I *did not* modify Trace SDK and Metrics SDK to do the same because: 
1. I am unsure if this is even needed in Metrics SDK at all, given there is some perf hit (only ~1 ns, which may not look much, but once we have bound instruments even 1 ns looks huge!)
2. ~Trace SDK (and API) need some refactoring to find the right spot to do this. This is better handled in its own PR.~ Edited: This is done in Sampler now. This could be moved even earlier after some refactoring outside of this PR.

Note: This does not prevent logs from external crates like tonic, hyper being fed back to Otel, as they don't propagate Otel context. That also can be solved, but this PR is not attempting to solve that.

Basic tests added, design doc updated.


opentelemetry-sdk/benches/log_enabled.rs
```diff
@@ -5,8 +5,8 @@
     Total Number of Cores:   14 (10 performance and 4 efficiency)
     | Test                                        | Average time|
     |---------------------------------------------|-------------|
-    | exporter_disabled_concurrent_processor      |  1.1 ns     |
-    | exporter_disabled_simple_processor          |  4.3 ns     |
+    | exporter_disabled_concurrent_processor      |  2.5 ns     |
```
cijothomas, line 8: @lalitb @utpilla This regression (paid by everyone, irrespective of which processor/exporter) is why we need to try and improve the cost of checking for suppression. @bantonsson showed some approach to make this faster, will explore that/other-ideas-too subsequently.

docs/design/logs.md
```diff
@@ -345,7 +345,25 @@ only meant for OTel components itself and anyone writing extensions like custom
 Exporters etc.
 
 // TODO: Document the principles followed when selecting severity for internal
-logs // TODO: Document how this can cause circular loop and plans to address it.
+logs
+
+When OpenTelemetry components generate logs that could potentially feed back
```
scottgerring, line 350: This is super clearly written 👍

docs/design/logs.md
```diff
@@ -345,7 +345,25 @@ only meant for OTel components itself and anyone writing extensions like custom
...
+checks this flag in the current `Context` and ignores logs if suppression is
+enabled.
+
+This mechanism relies on proper in-process propagation of the `Context`.
+However, external libraries like `hyper` and `tonic`, which are used by
+OpenTelemetry in its OTLP Exporters, do not propagate OpenTelemetry's `Context`.
+As a result, the suppression mechanism does not work out-of-the-box to suppress
+logs originating from these libraries.
+
+// TODO: Document how OTLP can solve this issue without asking external
```
scottgerring, line 365: ... do we know how to do this without context propagation?! Would love a TL;DR if so!

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -319,25 +319,26 @@ mod tests {
...
+            // Without the suppression above, this will cause a deadlock as the export itself creates a log
             // while still within the lock of the SimpleLogProcessor.
             warn!(name: "my-event-name", target: "reentrant", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
             Ok(())
         }
     }
 
     #[test]
-    #[ignore = "See issue: https://github.com/open-telemetry/opentelemetry-rust/issues/1745"]
+    #[ignore = "While this test runs fine, this uses global subscriber and does not play well with other tests and hence ignored in CI."]
```
scottgerring, line 331: Quite a picky nit, but why not leave the issue link in a comment as well?

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -52,6 +52,7 @@ impl opentelemetry::logs::LoggerProvider for SdkLoggerProvider {
     }
 
     fn logger_with_scope(&self, scope: InstrumentationScope) -> Self::Logger {
+        // TODO: Should this check suppression context and return a no-op logger?
```
scottgerring, line 55: To the comment: could a logger created here potentially be used outside of a suppressed context, when it is created within it?

I can't imagine a situation - if you create a logger in a suppressed context, _you are us - OTel_ - and there isn't really a reason you should pass that back out to a client of ours. On the other hand, if this ever happened it would be gnarly to debug the missing logs.

docs/design/logs.md
cijothomas: We can't fix without context propagation. And context propagation is only needed if the export task moves around threads. What we can try to do is - create a Tokio_multi_thread_runtime outselves, set its thread_count=1, and hand that runtime to OTLPExporter, instead of the current way of it grabbing the current runtime. 
Once we know OTLP (more specifically tonic/reqwest) always run in a given thread, we can somehow suppress context in that thread once, and it should work forever then.

Obviously, this needs to be tried out!

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -319,25 +319,26 @@ mod tests {
 
     impl LogExporter for ReentrantLogExporter {
         async fn export(&self, _batch: LogBatch<'_>) -> OTelSdkResult {
-            // This will cause a deadlock as the export itself creates a log
+            let _suppress = Context::enter_telemetry_suppressed_scope();
```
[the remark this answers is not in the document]
cijothomas: Telemetry suppression is *not just* for internal logs. I used that as as start only. We'll need to suppress logs and traces (metrics I am un-sure due to cost) too.
Eventually libraries gets instrumented (natively, or via instrumentation libraries) to produce spans too. (HttpClient can produce spans, but we want to suppress them when HTTPClient is executing in the context of OTLPExporter or JaegerRemoteSampler call etc.)

I don't see any perf concerns with EnteringContext. It's usually done once only. (This PR shows it's done in the background thread once, and that's all).

Additionally, we don't have much (maybe none actually) internal logs in happy hot path. Any such logs are at debug! level, which is rarely enabled in PROD systems, and tokio/tracing nicely optimizes them away.

(And yes, I am worried about the cost of checking if we are in suppressed context or not! https://github.com/open-telemetry/opentelemetry-rust/pull/2868/files#r2015295825)

opentelemetry-appender-tracing/src/layer.rs
cijothomas: There is no issue to track this, as there is no bug to be fixed. The test is written in such a way that it needs "global_subscriber", and such tests don't play well with others, so not running in CI.

(Irrespective of that, this test is better fit for the Otel sdk tests, will be moving it in follow ups)

opentelemetry-sdk/src/logs/logger_provider.rs
cijothomas: good point! You are right. I'll remove this comment, as we should not check suppression for creating Logger, but only while emitting Logs.

opentelemetry-appender-tracing/src/layer.rs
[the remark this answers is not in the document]
scottgerring: > Telemetry suppression is not just for internal logs. I used that as as start only. We'll need to suppress logs and traces (metrics I am un-sure due to cost) too.

The suggestion above generalizes to any _internal_ stuff we do(although the example macro name I suggested doesn't imply that 😆 ) 

Even if we don't care about the additional cost, I still reckon it's worth thinking about "how can we enforce our internal signals going through suppression", rather than relying on developer prudence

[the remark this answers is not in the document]
cijothomas: >  "how can we enforce our internal signals going through suppression", rather than relying on developer prudence

I don't quite follow this. SDK Authors (i.e us!) need to ensure suppression context is entered appropriately.
It is not meant for regular end users to know about.

If someone writes a custom exporter/sampler etc, they most likely would have to know about suppression. I don't think we can avoid this. Can you elaborate more, perhaps a code snippet so I can better follow the suggestion.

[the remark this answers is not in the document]
scottgerring: ... there's a good chance i'm trying to over-complicate things here, but I figure it's worth thinking through.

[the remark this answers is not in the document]
cijothomas: > pub fn my_func() {
>     info!("I'm about to export something!!1111");
> }
> This is wrong, right? The author missed entering suppressed mode

No, this is not wrong. We don't need to enter suppression mode for every function everywhere.. This PR already does it for they key entry points already, so I don't think everyone authoring exporters need to re-do the work, except when they jump threads.

opentelemetry-sdk/src/logs/simple_log_processor.rs
```diff
@@ -76,6 +76,7 @@ impl<T: LogExporter> SimpleLogProcessor<T> {
 
 impl<T: LogExporter> LogProcessor for SimpleLogProcessor<T> {
     fn emit(&self, record: &mut SdkLogRecord, instrumentation: &InstrumentationScope) {
+        let _suppress_guard = Context::enter_telemetry_suppressed_scope();
```
lalitb, line 79: Not related to the line of code, but still - we plan to _NOT_ add the suppression guard in the SimpleConcurrentLogProcessor, which means if the exporter attached to it is sending internal logs, it needs to add the suppression guard within the exporter? By default, ETW and user_events exporters will not be logging any debug/errors in their code?

cijothomas: that is correct. SimpleConcurrentLogProcessor may need a rename to "ReentrantLogProcessor"/somethign else, and make it clear that it is to be used with exporters that do not cause re-entrancy back to itself.

opentelemetry-appender-tracing/src/layer.rs
utpilla: > There is no issue to track this, as there is no bug to be fixed.

It's okay if we are following up on this right after the PR, if not we should create a tracking issue for this. I just don't want us to keep making further Context related changes and forget about this test not being run in CI.

> (Irrespective of that, this test is better fit for the Otel sdk tests, will be moving it in follow ups)

If it's simple enough, I would suggest you do it in this PR itself since you're already making changes to the SDK project in this PR.

opentelemetry-sdk/src/logs/logger.rs
```diff
@@ -29,6 +29,9 @@ impl opentelemetry::logs::Logger for SdkLogger {
 
     /// Emit a `LogRecord`.
     fn emit(&self, mut record: Self::LogRecord) {
+        if Context::is_current_telemetry_suppressed() {
```
utpilla, line 32: Do we need this check here as well? The appender would anyway call `SdkLogger::event_enabled` before calling `emit` method. So why check this twice?

```suggestion
        #[cfg(not(feature = "spec_unstable_logs_enabled"))]
        if Context::is_current_telemetry_suppressed() {
```

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -533,4 +533,30 @@ mod tests {
...
+        let tracer_provider = SdkTracerProvider::builder()
+            .with_span_processor(span_processor)
+            .build();
+
+        // Act
+        let tracer = tracer_provider.tracer("test");
+        {
+            let _suppressed_context = Context::enter_telemetry_suppressed_scope();
+            let span = tracer.span_builder("span_name").start(&tracer);
+            drop(span);
```
utpilla, line 551: Why drop the `span` explicitly? It would be dropped anyway since it's going out of scope in the next line, right?

opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -172,6 +172,25 @@ impl ShouldSample for Sampler {
         attributes: &[KeyValue],
         links: &[Link],
     ) -> SamplingResult {
+        // Checking suppression mode in sampler is more efficient than checking
+        // it in the span processor as this allows earlier short-circuiting.
+        // Also, it is currently *NOT* possible to check suppression mode or
+        // even access current context in the span processor's OnEnd method. See
+        // https://github.com/open-telemetry/opentelemetry-rust/issues/2871
+        // TODO: Can we move this even earlier, i.e. in the Tracer itself?
+        // TODO: Check and fix: why this methods gets Option<&Context> and not
```
utpilla, line 181: It's optional because the span might be a root span?

opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -172,6 +172,25 @@ impl ShouldSample for Sampler {
...
+        // Context?
+        if let Some(parent_context) = parent_context {
+            if parent_context.is_telemetry_suppressed() {
```
utpilla, line 184: I'm not so familiar with the trace SDK implementation but it looks like we [don't always](https://github.com/open-telemetry/opentelemetry-rust/blob/297146701d483397432323c6e6cdef86e9da9999/opentelemetry-sdk/src/trace/tracer.rs#L216-L229) call `should_sample`. I don't know if that `tracing-opentelemetry` pre-sampling scenario need the suppression check.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## vNext
 
+`SdkLogger` modified to respect telemetry suppression based on `Context`. In
+other words, if the current context has telemetry suppression enabled, then logs
+will be ignored. The flag is typically set by OTel components to prevent
+telemetry from itself being fed back into OTel. `BatchLogProcessor`,
+`BatchSpanProcessor`, and `PeriodicReader` modified to set the suppression flag
+in their dedicated thread, so that telemetry generated from those threads will
+not be fed back into OTel. Similarly, `SimpleSpanProcessor` and
```
utpilla, line 11: `SimpleSpanProcessor` was not updated.

opentelemetry-sdk/src/trace/sampler.rs
cijothomas: Good point. Those should be fixed/cleaned up. I also put a TODO to move this check even earlier than sampling.

opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -172,6 +172,25 @@ impl ShouldSample for Sampler {
         attributes: &[KeyValue],
         links: &[Link],
     ) -> SamplingResult {
+        // Checking suppression mode in sampler is more efficient than checking
+        // it in the span processor as this allows earlier short-circuiting.
+        // Also, it is currently *NOT* possible to check suppression mode or
+        // even access current context in the span processor's OnEnd method. See
+        // https://github.com/open-telemetry/opentelemetry-rust/issues/2871
+        // TODO: Can we move this even earlier, i.e. in the Tracer itself?
```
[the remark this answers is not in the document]
cijothomas: yes exactly! (That's why I put a TODO to move it to Tracer.)

utpilla:
> 1. I am unsure if this is even needed in Metrics SDK at all, given there is some perf hit (only ~1 ns, which may not look much, but once we have bound instruments even 1 ns looks huge!)

The regular instruments don't have to follow what bounded instrument does. They can easily accommodate the cost of doing the check, but we probably should have a consistent story for all metric instruments.

opentelemetry-sdk/src/trace/sampler.rs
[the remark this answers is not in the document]
utpilla: Why create a TODO for it? 😄 This is first time this check is getting added. Why not just put in Tracer to begin with?

opentelemetry-sdk/src/logs/logger.rs
cijothomas: Appenders are not obligated to leverage that path. They may check it, they may not check. We can recommend all to do it (and do it in the appenders maintained here), but since we cannot enforce, I think the extra check is okay.

The dual cost is true for the enabled() case - we check it twice, but if Log is enabled, adding 1 ns should be okay?

opentelemetry-appender-tracing/src/layer.rs
cijothomas: Moving to the SDK project tests, as that feels the right place to do that test.

opentelemetry-sdk/src/trace/sampler.rs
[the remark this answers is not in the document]
cijothomas: good point! moved to tracer.
Don't ask me why I didn't put it in Tracer to begin with 🤣

opentelemetry-sdk/src/logs/logger.rs
utpilla: > but since we cannot enforce, I think the extra check is okay.

Yes. Agreed.

opentelemetry-otlp/examples/basic-otlp-http/src/main.rs
```diff
@@ -84,7 +84,6 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // https://github.com/open-telemetry/opentelemetry-rust/issues/761
     let filter_otel = EnvFilter::new("info")
         .add_directive("hyper=off".parse().unwrap())
-        .add_directive("opentelemetry=off".parse().unwrap())
```
utpilla, line 87: Let's remove the code comments which mention about adding this filter.

examples/logs-basic/src/main.rs
```diff
@@ -27,7 +27,6 @@ fn main() {
     // https://github.com/open-telemetry/opentelemetry-rust/issues/761
     let filter_otel = EnvFilter::new("info")
         .add_directive("hyper=off".parse().unwrap())
-        .add_directive("opentelemetry=off".parse().unwrap())
```
utpilla, line 30: Same as https://github.com/open-telemetry/opentelemetry-rust/pull/2868#discussion_r2017584940

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -78,7 +78,6 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
     // https://github.com/open-telemetry/opentelemetry-rust/issues/761
     let filter_otel = EnvFilter::new("info")
         .add_directive("hyper=off".parse().unwrap())
-        .add_directive("opentelemetry=off".parse().unwrap())
```
utpilla, line 81: Update the `opentelemetry-appender-tracing` example as well.

https://github.com/open-telemetry/opentelemetry-rust/blob/a071d8fc3906da56e089061927fbc1561e0e7677/opentelemetry-appender-tracing/examples/basic.rs#L19-L34

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -212,4 +214,86 @@ mod tests {
...
+    impl LogProcessor for ReentrantLogProcessor {
+        fn emit(&self, _data: &mut SdkLogRecord, _instrumentation: &InstrumentationScope) {
+            let _suppress = Context::enter_telemetry_suppressed_scope();
+            // Without the suppression above, the logger.emit(log_record) below will cause a deadlock,
+            // as it emits another log, which will attempt to acquire the same lock that is
+            // already held by itself!
+            let logger = self.logger.lock().unwrap();
+            if let Some(logger) = logger.as_ref() {
+                let mut log_record = logger.create_log_record();
+                log_record.set_severity_number(Severity::Error);
```
utpilla, line 272: nit: Setting severity number is not required for this test.

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -212,4 +214,86 @@ mod tests {
...
+    fn log_does_not_deadlock_when_suppression_enabled() {
+        let processor: ReentrantLogProcessor = ReentrantLogProcessor::new();
+        let logger_provider = SdkLoggerProvider::builder()
+            .with_log_processor(processor.clone())
+            .build();
+        processor.set_logger(logger_provider.logger("processor-logger"));
+
+        let logger = logger_provider.logger("test-logger");
+        let mut log_record = logger.create_log_record();
+        log_record.set_severity_number(Severity::Error);
```
utpilla, line 296: Same as https://github.com/open-telemetry/opentelemetry-rust/pull/2868#discussion_r2017596194

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -212,4 +214,86 @@ mod tests {
...
+            Ok(())
+        }
+
+        fn shutdown(&self) -> OTelSdkResult {
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn log_does_not_deadlock_when_suppression_enabled() {
```
utpilla, line 287: nit: I think we could make the name more descriptive.

```suggestion
    fn processor_internal_log_does_not_deadlock_with_suppression_enabled() {
```

opentelemetry-sdk/src/logs/simple_log_processor.rs
```diff
@@ -438,4 +440,56 @@ mod tests {
...
+                log_record.set_severity_number(opentelemetry::logs::Severity::Error);
+                logger.emit(log_record);
+            }
+
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn does_not_deadlock_when_exporter_is_rentrant_simple_log_processor() {
```
utpilla, line 480:
```suggestion
    fn exporter_internal_log_does_not_deadlock_with_simple_processor() {
```

utpilla (approved):

opentelemetry-sdk/src/logs/mod.rs
```diff
@@ -212,4 +214,86 @@ mod tests {
...
+            .build();
+
+        // Act
+        let logger = logger_provider.logger("test-logger");
+        let mut log_record = logger.create_log_record();
+        log_record.set_severity_number(Severity::Error);
+
+        {
+            let _suppressed_context = Context::enter_telemetry_suppressed_scope();
+            logger.emit(log_record);
```
lalitb, line 233: nit - add a comment that this emitted log would get suppressed, even if it's obvious. It took time as I was searching for tokio macros to be suppressed :)

opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -533,4 +533,29 @@ mod tests {
...
+        let span_processor = SimpleSpanProcessor::new(exporter.clone());
+        let tracer_provider = SdkTracerProvider::builder()
+            .with_span_processor(span_processor)
+            .build();
+
+        // Act
+        let tracer = tracer_provider.tracer("test");
+        {
+            let _suppressed_context = Context::enter_telemetry_suppressed_scope();
+            let _span = tracer.span_builder("span_name").start(&tracer);
```
lalitb, line 550: nit - add comment that this span would get suppressed, even if it's obvious.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -316,6 +316,7 @@ impl BatchSpanProcessor {
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Traces.BatchProcessor".to_string())
             .spawn(move || {
+                let _suppress_guard = Context::enter_telemetry_suppressed_scope();
```
lalitb, line 319: Suppression for simple processor will be added as separate PR?

opentelemetry-sdk/benches/log_enabled.rs
lalitb: Should or do we have a tracking issue for this as you right mentioned it affects all exporters (which mayn't generate any logs, or else choose to filter the logs using tokio-filter).

lalitb (approved):
Thanks for working on this., LGTM, with nit comments.

opentelemetry-sdk/src/trace/span_processor.rs
cijothomas: Can't do it until https://github.com/open-telemetry/opentelemetry-rust/issues/2871 is resolved. "touching" context in On_End panics today!

opentelemetry-sdk/src/logs/mod.rs
cijothomas: cleaned up.!

opentelemetry-sdk/src/trace/span_processor.rs
lalitb: Thanks for context on the context issue :) add TODO with link to issue?

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2883

PR #2870, opened 2025-03-27
perf: Run all benchmarks for shorter time

bantonsson:
## Changes

Shortens the warmup time to 1 second and measurement time to 2 seconds for all benchmarks, and prepares them to be run in one go from the workspace root. Most benchmarks have millions or billions of iterations with a 5 second measurement, which seems a bit excessive.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -32,6 +32,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 10  # Fetch current commit and its parent
+      - name: "Show change commit"
+        run: git log -1
       - uses: arduino/setup-protoc@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
@@ -42,11 +44,15 @@ jobs:
         with:
           cwd: opentelemetry
           branchName: ${{ env.BRANCH_NAME }}
+      - name: "Checkout change commit"
+        run: git checkout $GITHUB_SHA
       - uses: boa-dev/criterion-compare-action@v3
         with:
           cwd: opentelemetry-appender-tracing
           features: spec_unstable_logs_enabled
           branchName: ${{ env.BRANCH_NAME }}
+      - name: "Checkout change commit"
+        run: git checkout $GITHUB_SHA
       - uses: boa-dev/criterion-compare-action@v3
         with:
           cwd: opentelemetry-sdk
--- a/opentelemetry-appender-tracing/benches/log-attributes.rs
+++ b/opentelemetry-appender-tracing/benches/log-attributes.rs
@@ -258,14 +258,19 @@ fn criterion_benchmark(c: &mut Criterion) {
 #[cfg(not(target_os = "windows"))]
 criterion_group! {
     name = benches;
-    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    config = Criterion::default()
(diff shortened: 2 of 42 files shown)


scottgerring:
Ooh! I'd love for this to be quick enough we can run it more often. Let's see how this run goes.

.github/workflows/benchmark.yml
```diff
@@ -40,15 +40,5 @@ jobs:
           toolchain: stable
       - uses: boa-dev/criterion-compare-action@v3
         with:
```
scottgerring, line 42: Great to combine this all into one

opentelemetry-appender-tracing/benches/log-attributes.rs
```diff
@@ -258,14 +258,19 @@ fn criterion_benchmark(c: &mut Criterion) {
 #[cfg(not(target_os = "windows"))]
 criterion_group! {
     name = benches;
-    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    config = Criterion::default()
```
scottgerring, line 261: is there an obvious place we can chuck this to have a "global default" across our crates?

bantonsson: Not that I know about. The other thing would be to have the runner specify these on the command line, but there doesn't seem to be any way to configure that in the action.

scottgerring: FWIW - I'd like to switch to something like https://bencher.dev/ anyway, but I don't want to scope creep that here. Let's keep in mind that we could factor this out later if we wanted, and leave it as you've done it.

bantonsson:
Also, since the `base` benchmarks are being run on `main` they will warm up and measure for 3 + 5 seconds, so the time gains won't be as pronounced.

bantonsson:
I just noticed that the github workflow is not switching back to the PR branch after the first invocation of `boa-dev/criterion-compare-action@v3`, so the two last measurements are not comparing the right things. They are comparing `main` to `main` 😱.

scottgerring:
> I just noticed that the github workflow is not switching back to the PR branch after the first invocation of `boa-dev/criterion-compare-action@v3`, so the two last measurements are not comparing the right things. They are comparing `main` to `main` 😱.

This is just since you've combined them all together though, right?

scottgerring:
> I just noticed that the github workflow is not switching back to the PR branch after the first invocation of `boa-dev/criterion-compare-action@v3`, so the two last measurements are not comparing the right things. They are comparing `main` to `main` 😱.

Yuck. Good catch. Fortunately we've only ever discussed changes from the first run - all the context stuff happens before the rest!

bantonsson:
> Yuck. Good catch. Fortunately we've only ever discussed changes from the first run - all the context stuff happens before the rest!

No @scottgerring this is a real issue with the original way of running the same `boa-dev/criterion-compare-action@v3` three times after each other.

There is a fix in this PR as well.

bantonsson:
This now runs the benchmarks in ~50 minutes instead of ~1:05. And that is with only the _changed_ side running faster while `main` is still warms up for 3 seconds and measures for 5 seconds. I expect that there will be another ~15 minutes decrease when this is merged to main.

scottgerring (approved):
This is just what we need, and all looks good to me. Thanks for the effort! Excited to see how fast it'll be once we have it in main too.

Issue #2871, opened 2025-03-27
Trace SDK and Context access issue


cijothomas:
Accessing current context on Span's `OnEnd` panics. `OnBegin` is fine, but OnBegin is given an explicit context, but OnEnd does not get any, so has to rely on current (which panics)

Related is the old spec issue: https://github.com/open-telemetry/opentelemetry-specification/issues/867

PR #2872, opened 2025-03-27
test: Add ignored code in test to prove known issue

cijothomas:
To demonstrate the issue https://github.com/open-telemetry/opentelemetry-rust/issues/2871
This PR is not investigating/fixing the issue, but rather adding a test to confirm the issue exists.


opentelemetry-sdk/src/trace/mod.rs
```diff
@@ -136,7 +136,11 @@ mod tests {
             }
         }
 
-        fn on_end(&self, _span: SpanData) {}
+        fn on_end(&self, _span: SpanData) {
+            // TODO: Accessing Context::current() will panic today and hence commented out.
+            // See https://github.com/open-telemetry/opentelemetry-rust/issues/2871
+            // let _c = Context::current();
```
lalitb, line 142: We can also add an ignored test using a TestSpanProcessor that accesses Context::current() in on_end(), to validate the expected behavior once it's safe.

lalitb (approved):

PR #2884, opened 2025-03-28
perf: Nit fix to benchmarks

cijothomas:
Some test were not run in benchmarks-CI due to missing feature flag.
Also did a nit cleanup - benchmarks didn't need `rt-tokio` so its cleaned up.


lalitb (approved):

bantonsson:
I've done a more future proof fix for missing feature flags in #2885

PR #2890, opened 2025-04-01
feat: add shutdown with timeout for metric reader and provider

mohammadVatandoost:
https://github.com/open-telemetry/opentelemetry-rust/issues/2574 Part 2: Adding shutdown_with_timeout for metric reader and provider


cijothomas (approved):
LGTM, need to move changelog entry to the unreleased part.

PR #2894, opened 2025-04-02
feat: allow retrieval of Context from ContextGuard

brettmc:
## Changes
When attaching a context, store the context in the ContextGuard so that it can be retrieved.

### Why?
I'm a maintainer of opentelemetry-php, and am [prototyping](https://github.com/brettmc/otel-php-rust) using opentelemetry-rust to do all of the heavy lifting for a PHP API/SDK. One of the features from PHP's context API that I want to implement is being able to store a guard (we call it a scope), then later fetching it and operating on the context's active span.

### Usage
In rust, what I want to do would look like this:

```
let guard = context.attach();
//store guard

//elsewhere
let retrieved_guard = /* get guard from storage */
let span = retrieved_guard.context().span();
//mutate and end span
```

I think this change works (see added test), but running the benchmarks locally, the `context_attach` tests have a regression which probably makes this PR a non-starter. I'm submitting just in case there's an optimization that I can't see, or the regression isn't as bad as it looks.

### Benchmarks
Running locally, compared to main:
```text
context_attach/single_cx/empty_cx                                                                            
                        time:   [26.208 ns 26.245 ns 26.284 ns]
                        change: [+49.500% +50.047% +50.608%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/nested_cx/empty_cx                                                                            
                        time:   [48.317 ns 48.408 ns 48.510 ns]
                        change: [+49.426% +49.904% +50.388%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/out_of_order_cx_drop/empty_cx                                                                            
                        time:   [37.119 ns 37.351 ns 37.644 ns]
                        change: [+26.830% +27.500% +28.154%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/single_cx/single_value_cx                                                                            
                        time:   [32.586 ns 32.660 ns 32.737 ns]
                        change: [+77.010% +77.527% +78.077%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/nested_cx/single_value_cx                                                                            
                        time:   [62.826 ns 62.945 ns 63.069 ns]
                        change: [+86.643% +87.316% +87.937%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/out_of_order_cx_drop/single_value_cx                                                                            
                        time:   [54.517 ns 54.585 ns 54.662 ns]
                        change: [+73.080% +73.636% +74.264%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/single_cx/span_cx                                                                            
                        time:   [32.646 ns 32.730 ns 32.816 ns]
                        change: [+76.620% +77.238% +77.926%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/nested_cx/span_cx                                                                            
                        time:   [62.843 ns 63.110 ns 63.422 ns]
                        change: [+85.580% +86.407% +87.284%] (p = 0.00 < 0.05)
                        Performance has regressed.

context_attach/out_of_order_cx_drop/span_cx                                                                            
                        time:   [52.952 ns 53.026 ns 53.098 ns]
                        change: [+72.477% +73.032% +73.663%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### DIFF
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -324,10 +324,11 @@ impl Context {
     /// assert_eq!(Context::current().get::<ValueA>(), None);
     /// ```
     pub fn attach(self) -> ContextGuard {
-        let cx_id = CURRENT_CONTEXT.with(|cx| cx.borrow_mut().push(self));
+        let cx_id = CURRENT_CONTEXT.with(|cx| cx.borrow_mut().push(self.clone()));
 
         ContextGuard {
             cx_pos: cx_id,
+            context: self,
             _marker: PhantomData,
         }
     }
@@ -456,8 +457,17 @@ impl fmt::Debug for Context {
 pub struct ContextGuard {
     // The position of the context in the stack. This is used to pop the context.
     cx_pos: u16,
+    // The current context.
+    context: Context,
     // Ensure this type is !Send as it relies on thread locals
     _marker: PhantomData<*const ()>,
+}
+
+impl ContextGuard {
+    /// Retrieve the Context
+    pub fn context(&self) -> &Context {
+        &self.context
+    }
 }
 
 impl Drop for ContextGuard {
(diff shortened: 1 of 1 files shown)


bantonsson (commented):
If we decide to add this type of API, I have another solution in mind that doesn't add any overhead.

scottgerring:
@brettmc you proposed an alternate solution that carries the context & guard together on the PHP side - what do you see the downside of this being?

I think _in general_ we should minimize additions to our public API surface, especially where they seem 'narrow' - in a sense that we don't expect them to be broadly used.

brettmc:
> you proposed an alternate solution that carries the context & guard together on the PHP side - what do you see the downside of this being?

Not terrible downsides, more complicated implementation for me but that is all, it will still work and do what I want.
If it's not likely to be useful for other folks, then I completely understand wanting to keep the public API small.

cijothomas:
Based on the SIG discussion, given there is an alternative available, we should not expose additional APIs - Closing.

@brettmc Let us know if we need to revisit this.

PR #2895, opened 2025-04-02
fix: Processor now gets passed in reference #2726 

### LINKED ISSUE #2726: Tracing SDK - Processors get clone of SpanData which is incorrect

Each processor gets a clone

Puffy1215:
Fixes #2726 

## Changes

See #2726. Previous implementation was not up to spec. I replaced the `on_end()` trait function parameter with a `&mut` reference to `SpanData`, and refactored from there. Cloning still happens inside `BatchSpanProcessor` where necessary. If there are any unnecessary clones that are still happening that I can fix, please let me know.

### DIFF
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -145,7 +145,7 @@ impl SpanProcessor for EnrichWithBaggageSpanProcessor {
         }
     }
 
-    fn on_end(&self, _span: opentelemetry_sdk::trace::SpanData) {}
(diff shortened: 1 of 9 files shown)


cijothomas:
@Puffy1215 Could you clear the EasyCLA check? The bot comment should have all instructions. You'd probably want to consult your employer if contributing on behalf. (This is a one time requirement, before we can accept contributions)
Thank you!

TommyCpp (commented):
Related to https://github.com/open-telemetry/opentelemetry-rust/issues/2726#issuecomment-2822165915 I am not sure if we want a mutable span for `on_end` method

PR #2898, opened 2025-04-03
feat(otlp): Re-export tonic crate

### LINKED ISSUE #882: enh: permit users with different versions of tonic
Right now tonic is used in two quite different ways: 1) it is an internal library that facilitates e.g. grpc handling and 2) tonic types are part of the public interface of opentelemetry-otlp.

For case (2) it would be really nice to have access to the version of tonic that opentelemetry-otlp expects as a re-export, rather than having to do lock-step upgrades of opentelemetry & tonic across our codebases.

something as simple as a pub use tonic in the root of opentelemetry-otlp might well be enough to fix this.

Example of (2) code today:
```
       TracingProtocol::Otlp(url, token) => {
            let mut map = tonic::metadata::map::MetadataMap::new();
            map.insert(ACCESS_TOKEN_HEADER, token.parse().unwrap());

                .tracing()
                .with_exporter(
                        .tonic()
                        .with_metadata(map)
                        .with_timeout(OTLP_TIMEOUT)
                        .with_protocol(Protocol::Grpc)
                        .with_endpoint(url),
                )
                .with_trace_config(trace_config)
                .install_batch(opentelemetry::runtime::Tokio)?
        }
```

The challenge here is that without reference to an exactly matching tonic version, this code will fail like so:
```
error[E0308]: mismatched types
Error:    --> src/tracing.rs:148:40
    |
148 |                         .with_metadata(map)
    |                          ------------- ^^^ expected struct `tonic::metadata::map::MetadataMap`, found struct `tonic::metadata::map::MetadataMap`
    |                          |
    |                          arguments to this function are incorrect
    |
    = note: perhaps two different versions of crate `tonic` are being used?
note: associated function defined here
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/opentelemetry-otlp-0.10.0/src/exporter/tonic.rs:44:12
    |
44  |     pub fn with_metadata(mut self, metadata: MetadataMap) -> Self {
    |            ^^^^^^^^^^^^^
```

MathieuTricoire:
Fixes #882

## Design discussion issue

I chose to re-export the entire `tonic` crate instead of cherry-picking types to avoid maintenance overhead. The re-export is gated behind the `grpc-tonic` feature.

Let me know your thoughts if re-exporting the whole `tonic` crate is too much and how we should tackle this in this situation.

## Changes

Re-exports the entire `tonic` crate from the `opentelemetry-otlp` crate, to allows downstream crates to use `tonic` types (such as `MetadataMap`) compatible with `opentelemetry-otlp` without needing to manually match the internal `tonic` version.

### DIFF
--- a/opentelemetry-otlp/allowed-external-types.toml
+++ b/opentelemetry-otlp/allowed-external-types.toml
@@ -13,6 +13,7 @@ allowed_external_types = [
     # prost is a pre 1.0 crate
     "prost::error::EncodeError",
     # tonic is a pre 1.0 crate
+    "tonic",
     "tonic::status::Code",
     "tonic::status::Status",
     "tonic::metadata::map::MetadataMap",
--- a/opentelemetry-otlp/allowed-external-types.toml
+++ b/opentelemetry-otlp/allowed-external-types.toml
@@ -3,23 +3,16 @@
 # This is used with cargo-check-external-types to reduce the surface area of downstream crates from
 # the public API. Ideally this can have a few exceptions as possible.
 allowed_external_types = [
-    "opentelemetry::*",
     "opentelemetry_http::*",
     "opentelemetry_sdk::*",
-    # http is a pre 1.0 crate
-    "http::uri::InvalidUri",
-    "http::header::name::InvalidHeaderName",
-    "http::header::value::InvalidHeaderValue",
-    # prost is a pre 1.0 crate
-    "prost::error::EncodeError",
+    # serde
+    "serde::de::Deserialize",
+    "serde::ser::Serialize",
     # tonic is a pre 1.0 crate
-    "tonic::status::Code",
-    "tonic::status::Status",
     "tonic::metadata::map::MetadataMap",
     "tonic::transport::channel::tls::ClientTlsConfig",
     "tonic::transport::tls::Certificate",
     "tonic::transport::tls::Identity",
     "tonic::transport::channel::Channel",
-    "tonic::transport::error::Error",
     "tonic::service::interceptor::Interceptor",
 ]
--- a/opentelemetry-otlp/allowed-external-types.toml
+++ b/opentelemetry-otlp/allowed-external-types.toml
@@ -13,10 +13,12 @@ allowed_external_types = [
     # prost is a pre 1.0 crate
     "prost::error::EncodeError",
     # tonic is a pre 1.0 crate
-    "tonic",
     "tonic::status::Code",
     "tonic::status::Status",
     "tonic::metadata::map::MetadataMap",
+    "tonic::transport::channel::tls::ClientTlsConfig",
+    "tonic::transport::tls::Certificate",
+    "tonic::transport::tls::Identity",
     "tonic::transport::channel::Channel",
     "tonic::transport::error::Error",
     "tonic::service::interceptor::Interceptor",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,7 @@ jobs:
         run: |
           cargo install cargo-check-external-types@0.1.13
           cd ${{ matrix.example }}
-          cargo check-external-types --config allowed-external-types.toml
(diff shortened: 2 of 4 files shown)


lalitb:
@MathieuTricoire Re-exporting the whole tonic crate works, but it exposes all of tonic’s API as part of our public API. If we only need a few types (like `MetadataMap`), we could selectively re-export just those (e.g., `pub use tonic::metadata::MetadataMap;). If more types are needed in the future, we can always add them as needed. Or do you see a  reason to expose the whole crate up front?

MathieuTricoire:
@latlib Thanks for the answer! That makes sense, and I'm totally good with cherry-picking, I actually thought about that but I didn't want to overthinking it right away if not necessary.

Going cherry-picking route, I see two options:

__Flat re-export:__

```rust
pub mod tonic_types {
    #[doc(no_inline)]
    pub use tonic::metadata::MetadataMap;
    #[doc(no_inline)]
    #[cfg(feature = "tls")]
    pub use tonic::transport::{Certificate, ClientTlsConfig, Identity};
}
```

__Grouped by module:__

```rust
pub mod tonic_types {
    pub mod metadata {
        #[doc(no_inline)]
        pub use tonic::metadata::MetadataMap;
    }

    #[cfg(feature = "tls")]
    pub mod transport {
        #[doc(no_inline)]
        pub use tonic::transport::{Certificate, ClientTlsConfig, Identity};
    }
}
```

I kind of like the module-based one, feels cleaner and easier to extend later to me, but maybe over-engineered? Happy to go with whichever makes more sense for you.

Also wondering what you think about the naming, would `tonic_types` work? Or maybe `tonic_compat`, `tonic_export`, or something else?

Also are you ok with the current re-export scope?

Let me know what direction you'd prefer and I'll adjust the PR 👍

MathieuTricoire:
I pushed the grouped by module proposal, let me know your thoughts.

lalitb:
thanks @MathieuTricoire - I agree - the module-based structure feels cleaner and more future-proof, especially if we add more exports later. So let's go with that.

opentelemetry-otlp/allowed-external-types.toml
```diff
@@ -16,6 +16,9 @@ allowed_external_types = [
     "tonic::status::Code",
     "tonic::status::Status",
     "tonic::metadata::map::MetadataMap",
+    "tonic::transport::channel::tls::ClientTlsConfig",
+    "tonic::transport::tls::Certificate",
+    "tonic::transport::tls::Identity",
     "tonic::transport::channel::Channel",
     "tonic::transport::error::Error",
     "tonic::service::interceptor::Interceptor",
```
bantonsson, line 24: I assume that all of these should be reexported and removed?

bantonsson (commented):
👍🏻 on the module based approach

opentelemetry-otlp/allowed-external-types.toml
MathieuTricoire: Hi @bantonsson, thanks for the review!
The added types are the ones now re-exported.
I also updated the CI script to check allowed external types with all features enabled, because the previous configuration was correct but wasn't actually enforced (some re-exported types are feature-gated).
I've also cleaned up the old types that are no longer re-exported.

bantonsson: Thanks for the explanation @MathieuTricoire. Didn't know how that plugin works.

bantonsson (approved):

MathieuTricoire:
CI has been fixed after the merge of #2945 and the PR should be ready for merging.

I have checked the `Appropriate CHANGELOG.md files updated for non-trivial, user-facing changes` requirement, but do you think a CHANGELOG update is needed?

cijothomas (approved):
Looks good, can you add a changelog entry too?

MathieuTricoire:
Hi @cijothomas, I've added the changelog entry, let me know if that works for you 👍

Issue #2451, opened 2024-12-18
Discontinue prometheus crate?


cijothomas:
Given more and more native support from Prometheus to ingest OTLP, wondering if we should keep publishing the prometheus crate?
https://prometheus.io/blog/2024/11/14/prometheus-3-0/#otlp-support

cijothomas:
Related: https://github.com/open-telemetry/opentelemetry-rust/issues/1551 and https://github.com/open-telemetry/opentelemetry-rust/issues/2769

I'd vote to mark this crate as deprecated in the coming release, and remove it fully. We can bring it back, when we have the bandwidth to rewrite it using the official prometheus crate and also avoiding the protobuf dependency which seem to be unmaintained (https://github.com/stepancheg/rust-protobuf/?tab=readme-ov-file#help)

cijothomas:
@open-telemetry/rust-maintainers Do you see any concerns with the above? Happy to discuss ideas, but we should not be shipping crates that cannot be easily maintained/patched due to dependencies on unmaintained (or less actively maintained) crates IMHO.

cijothomas:
Yes those changes were relatively minor ones to keep up with upstream. but changing the dependencies is much bigger in scope and I think it'd be possibly require rewriting from scratch, and will be breaking changes for end users too 
https://github.com/open-telemetry/opentelemetry-rust/issues/1551

sandhose:
So I've decided to have a shot at rewriting a prometheus exporter which doesn't depend on the prometheus crate. I've followed the [official guidelines for this](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/) as much as possible. Here is the crate: https://crates.io/crates/opentelemetry-prometheus-text-exporter

Note that I haven't tested it in production yet, but it looks alright in tests, and quite more performant than the `opentelemetry-prometheus` crate in some basic benchmarks.

I'd be happy to contribute that to this repo if it makes sense, and if the OTEL Rust project is willing to maintain that

PR #2899, opened 2025-04-03
refactor: Replace OnceCell with std::sync::OnceLock

FalkWoldmann:
Fixes #

## Changes

Replaces `once_cell::sync::OnceCell` with its`std` equivalent. We cannot yet remove the whole `once_cell` crate, since we still use `Lazy` and its `std` counterpart `LazyLock` requires rust 1.80


lalitb:
Changes look good, but I believe this crate won't accept changes any longer as mentioned here - https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-prometheus

-clip
Warning: This crate is no longer recommended for use.

Development of the Prometheus exporter has been discontinued. See the related https://github.com/open-telemetry/opentelemetry-rust/issues/2451. This crate depends on the unmaintained protobuf crate and has unresolved security vulnerabilities. Version 0.29 will be the final release.

For Prometheus integration, we strongly recommend using the [OTLP](https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/) exporter instead. Prometheus [natively supports OTLP](https://prometheus.io/docs/guides/opentelemetry/#enable-the-otlp-receiver), providing a more stable and actively maintained solution.

FalkWoldmann:
Ah, I missed this comment. Thanks for the feedback!

Issue #2878, opened 2025-03-28
[Bug]: Cardinality overflow metric should use a boolean value rather than a string


timbotnik:
### What happened?

I'd say this is a pretty minor issue, but we did encounter it when looking specifically for this metric value so maybe it will help someone and maybe it should get fixed before a 1.x.

When hitting the cardinality limit for metrics export, the library correctly produces a _string value_ attribute `otel.metric.overflow` on the metrics instrument that has an overflow.  However [the OTel specification](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#overflow-attribute) indicates this should be a _boolean value_.  

I think the fix is to just remove the quotes, e.g. `"true"` => `true`:

https://github.com/open-telemetry/opentelemetry-rust/blob/b2de6cc5a3dc271fffee45f050e894fba57508ba/opentelemetry-sdk/src/metrics/internal/mod.rs#L27

This is probably not an issue post v0.28 since there is no longer a cardinality limit in place!  But probably worth noting as limits with configuration support are coming back in soon.

### OpenTelemetry API Version (i.e version of `opentelemetry` crate)

0.24.0

### OpenTelemetry SDK Version (i.e version of `opentelemetry_sdk` crate)

0.24.1

### Relevant log output

```shell

```

cijothomas:
Thanks for reporting this! Yes it should be boolean true! We are bringing back cardinality limits soon, so will take care of fixing this.

cijothomas:
Cardinality cap and configuration ability is added. Will add this shortly.

PR #2901, opened 2025-04-04
feat: Add and enabled Metric cardinality capping by default

### LINKED ISSUE #1065: Implement cardinality limits for metrics streams

cijothomas:
Fixes #1065

Re-introduce Metric cardinality capping, but this time, with the ability to configure the limit using Views.
The default is 2000, following OTel spec. 
There is no ability to change the limit globally (at reader level) - this is not a MUST from the spec, and based on my experience, this is rarely used/needed.
Existing tests are re-enabled, and also enhanced to test customization ability and to test that limits are per cycle.

TODO (in next PR): is to add coverage for all instruments, as only Counters are currently covered.
TODO : doc additions. Depending on the next PR where I plan to introduce customization ability via Advisory, we can decide how best to document this.
TODO(next PR): Fix https://github.com/open-telemetry/opentelemetry-rust/issues/2878 too

I expect to send another PR which will enhance the Instrument Advisory parameters to accept cardinality limits via that option. This is not a thing from the spec (but spec does not prohibit offering additional capabilities either), but planned to be added to Rust as using Views is generally non-trivial and my own experience tells it is somewhat confusing.
(Also planning to make a spec enhancement proposal in parallel, but I don't think it is a blocker)

### DIFF
--- a/opentelemetry-sdk/src/metrics/instrument.rs
+++ b/opentelemetry-sdk/src/metrics/instrument.rs
@@ -203,6 +203,9 @@ pub struct Stream {
     /// dropped. If the set is empty, all attributes will be dropped, if `None` all
     /// attributes will be kept.
     pub allowed_attribute_keys: Option<Arc<HashSet<Key>>>,
+
+    /// Cardinality limit for the stream.
+    pub cardinality_limit: Option<usize>,
 }
 
 #[cfg(feature = "spec_unstable_metrics_views")]
@@ -243,6 +246,12 @@ impl Stream {
     pub fn allowed_attribute_keys(mut self, attribute_keys: impl IntoIterator<Item = Key>) -> Self {
         self.allowed_attribute_keys = Some(Arc::new(attribute_keys.into_iter().collect()));
 
+        self
+    }
+
+    /// Set the stream cardinality limit.
+    pub fn cardinality_limit(mut self, limit: usize) -> Self {
+        self.cardinality_limit = Some(limit);
         self
     }
 }
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -374,8 +374,8 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn counter_aggregation_overflow_cumulative() {
-        counter_aggregation_overflow_helper(Temporality::Delta);
-        counter_aggregation_overflow_helper_custom_limit(Temporality::Delta);
+        counter_aggregation_overflow_helper(Temporality::Cumulative);
+        counter_aggregation_overflow_helper_custom_limit(Temporality::Cumulative);
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
@@ -2550,7 +2550,7 @@ mod tests {
             assert_eq!(data_point.value, 100);
         } else {
             // For cumulative, overflow should still be there, and new points should not be added.
-            assert_eq!(sum.data_points.len(), 2002);
+            assert_eq!(sum.data_points.len(), cardinality_limit + 1 + 1);
             let data_point =
                 find_sum_datapoint_with_key_value(&sum.data_points, "otel.metric.overflow", "true")
                     .expect("overflow point expected");
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -370,12 +370,14 @@ mod tests {
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn counter_aggregation_overflow_delta() {
         counter_aggregation_overflow_helper(Temporality::Delta);
+        counter_aggregation_overflow_helper_custom_limit(Temporality::Delta);
     }
 
     #[ignore = "https://github.com/open-telemetry/opentelemetry-rust/issues/1065"]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn counter_aggregation_overflow_cumulative() {
         counter_aggregation_overflow_helper(Temporality::Cumulative);
+        counter_aggregation_overflow_helper_custom_limit(Temporality::Delta);
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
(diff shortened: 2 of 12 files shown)


opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -245,6 +248,12 @@ impl Stream {
 
         self
     }
+
+    /// Set the stream cardinality limit.
+    pub fn cardinality_limit(mut self, limit: usize) -> Self {
+        self.cardinality_limit = Some(limit);
```
cijothomas, line 254: todo: Add validation for usize

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2412,6 +2414,123 @@ mod tests {
             empty_attrs_data_point.value, 6,
             "Empty attributes value should be 3+3=6"
         );
+
+        // Phase 2 - for delta temporality, after each collect, data points are cleared
+        if temporality == Temporality::Delta {
```
utpilla, line 2419: You can also check that these measurements continue to go to overflow for Cumulative temporality.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -370,12 +370,14 @@ mod tests {
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn counter_aggregation_overflow_delta() {
         counter_aggregation_overflow_helper(Temporality::Delta);
+        counter_aggregation_overflow_helper_custom_limit(Temporality::Delta);
     }
 
     #[ignore = "https://github.com/open-telemetry/opentelemetry-rust/issues/1065"]
```
utpilla, line 376: Remove the ignore attribute from `counter_aggregation_overflow_cumulative` and `counter_aggregation_overflow_delta` tests now that you're adding cardinality limit.

cijothomas: yes good catch!

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2412,6 +2414,123 @@ mod tests {
...
+            let data_point =
+                find_sum_datapoint_with_key_value(&sum.data_points, "A", "yet_another")
+                    .expect("point expected");
+            assert_eq!(data_point.value, 100);
+        }
+    }
+
+    fn counter_aggregation_overflow_helper_custom_limit(temporality: Temporality) {
+        // Arrange
+        let cardinality_limit = Arc::new(2300);
```
utpilla, line 2450: There is no need to use `Arc` and related cloning, dereferencing etc. here. A simple `usize` typed variable would work.

```suggestion
        let cardinality_limit: usize = 2300;
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2412,6 +2414,123 @@ mod tests {
...
+            empty_attrs_data_point.attributes.is_empty(),
+            "Non-empty attribute set"
+        );
+        assert_eq!(
+            empty_attrs_data_point.value, 6,
+            "Empty attributes value should be 3+3=6"
+        );
+
+        // Phase 2 - for delta temporality, after each collect, data points are cleared
+        if temporality == Temporality::Delta {
```
utpilla, line 2507: Same as https://github.com/open-telemetry/opentelemetry-rust/pull/2901#discussion_r2029438174

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -385,7 +386,11 @@ where
                 .clone()
                 .map(|allowed| Arc::new(move |kv: &KeyValue| allowed.contains(&kv.key)) as Arc<_>);
 
-            let b = AggregateBuilder::new(self.pipeline.reader.temporality(kind), filter);
+            let b = AggregateBuilder::new(
+                self.pipeline.reader.temporality(kind),
+                filter,
+                stream.cardinality_limit.unwrap_or(2000),
```
utpilla, line 392: I think we can make the code more readable / self-explanatory by adding a `const CARDINALITY_LIMIT_DEFAULT` in the file and use that instead of directly using `2000`.

utpilla, line 392: Do we have any logging for instrument creation? We should add some information about the cardinality limit to the logs.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2412,6 +2412,160 @@ mod tests {
             empty_attrs_data_point.value, 6,
             "Empty attributes value should be 3+3=6"
         );
+
+        // Phase 2 - for delta temporality, after each collect, data points are cleared
```
lalitb, line 2416: Is this test (phase 2) related to cardinality limit?

cijothomas: yes. it is validating that cardinality limit is per cycle for delta.

lalitb: ok please ignore. Seems its checking for overflow.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -366,16 +366,16 @@ mod tests {
...
     async fn counter_aggregation_overflow_delta() {
         counter_aggregation_overflow_helper(Temporality::Delta);
+        counter_aggregation_overflow_helper_custom_limit(Temporality::Delta);
     }
 
-    #[ignore = "https://github.com/open-telemetry/opentelemetry-rust/issues/1065"]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn counter_aggregation_overflow_cumulative() {
         counter_aggregation_overflow_helper(Temporality::Cumulative);
+        counter_aggregation_overflow_helper_custom_limit(Temporality::Delta);
```
utpilla, line 378:
```suggestion
        counter_aggregation_overflow_helper_custom_limit(Temporality::Cumulative);
```

opentelemetry-sdk/src/metrics/pipeline.rs
cijothomas: added cardinality info to the existing log.

lalitb (approved):

opentelemetry-sdk/src/metrics/mod.rs
utpilla: It looks like you have changed both of the methods to use `Delta` instead of `Cumulative`.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2412,6 +2412,160 @@ mod tests {
...
+                .expect("point expected");
+            assert_eq!(data_point.value, 100);
+
+            let data_point =
+                find_sum_datapoint_with_key_value(&sum.data_points, "A", "yet_another")
+                    .expect("point expected");
+            assert_eq!(data_point.value, 100);
+        } else {
+            // For cumulative, overflow should still be there, and new points should not be added.
+            assert_eq!(sum.data_points.len(), 2002);
```
utpilla, line 2553: This should be `cardinality_limit + 2`.

opentelemetry-sdk/src/metrics/instrument.rs
cijothomas: nothing in spec for this, better not to do anything now.

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -385,12 +384,25 @@ where
...
+                cardinality_limit,
+            );
             let AggregateFns { measure, collect } = match aggregate_fn(b, &agg, kind) {
                 Ok(Some(inst)) => inst,
                 other => return other.map(|fs| fs.map(|inst| inst.measure)), // Drop aggregator or error
             };
 
+            otel_debug!(
+                name : "Metrics.InstrumentCreated",
+                instrument_name = stream.name.as_ref(),
```
utpilla, line 402: nit: Add `Aggregation` information as well to understand what kind of instrument was created.

utpilla (approved):

PR #2905, opened 2025-04-06
fix: Cleanup MetricError and use OTelSdkResult instead

cijothomas:
Continuing from https://github.com/open-telemetry/opentelemetry-rust/pull/2604
In another small follow up, I should be able to remove `pub enum MetricError` completely!


gruebel (approved):
noice

PR #2906, opened 2025-04-07
fix: cleanup MetricError

cijothomas:
Continuing from https://github.com/open-telemetry/opentelemetry-rust/pull/2905

There will be further cleanups for `MetricError` to the point it'll be completely removed. All places where it is required has been modified to use the `OTelSdkResult`. The remaining usages are only required for "Views", which is experimental features, so the `MetricError` can remain public only when that feature is enabled, otherwise non-public.
This helps cleanup unnecessary public API from the Metrics Sdk, and gets one more step closer to declaring Metrics Sdk as stable.

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -325,7 +325,7 @@ impl OtlpHttpClient {
     fn build_metrics_export_body(
         &self,
         metrics: &mut ResourceMetrics,
-    ) -> opentelemetry_sdk::metrics::MetricResult<(Vec<u8>, &'static str)> {
+    ) -> Option<(Vec<u8>, &'static str)> {
         use opentelemetry_proto::tonic::collector::metrics::v1::ExportMetricsServiceRequest;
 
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -4,6 +4,8 @@ use super::{
 };
 use crate::{ExportConfig, Protocol, OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_HEADERS};
 use http::{HeaderName, HeaderValue, Uri};
+#[cfg(feature = "http-json")]
```
gruebel, line 7: shouldn't this be `internal-logs` feature?

gruebel (approved):
nice

opentelemetry-otlp/src/exporter/http/mod.rs
cijothomas: no, this is only used when json serialization failed, and it was complaining about unused imports without this!

PR #2916, opened 2025-04-08
fix: Cardinality overflow to use bool value instead of string

### LINKED ISSUE #2878: [Bug]: Cardinality overflow metric should use a boolean value rather than a string

cijothomas:
Fixes #2878

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -2833,9 +2833,7 @@ mod tests {
         })
     }
 
-    fn find_overflow_sum_datapoint<'a, T>(
-        data_points: &'a [SumDataPoint<T>],
-    ) -> Option<&'a SumDataPoint<T>> {
+    fn find_overflow_sum_datapoint<T>(data_points: &[SumDataPoint<T>]) -> Option<&SumDataPoint<T>> {
         data_points.iter().find(|&datapoint| {
             datapoint.attributes.iter().any(|kv| {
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2838,6 +2833,16 @@ mod tests {
         })
     }
 
+    fn find_overflow_sum_datapoint<'a, T>(
+        data_points: &'a [SumDataPoint<T>],
```
utpilla, line 2837: There is a lint error asking to remove lifetime annotations here.

Issue #2917, opened 2025-04-08
MetricReader redesign


cijothomas:
MetricReader is an important aspect of the OTel Metric Specification. It is currently part of the public API in OTel Rust, though it is rarely required by end users to directly deal with MetricReaders. (100% of examples in the repo - no need of learning about MetricReader at all, as it is handled transparently)

The design of MetricReader has some challenges, especially in locking the Public API. Opening an issue to track improvements to the same.
Few items from top of mind:
1. Consult OTel TC/GC to get an exemption from having to build Prometheus Exporter in this repo. While we already deprioritized it, we need an official blessing to remove it completely. The "pull" model used by Prometheus has some design challenges, and if we can avoid that, it'll make MetricReader design a lot more simple.
2. See if there is any issue marking MetricReader under an experimental feature, allowing us to make breaking changes to it in the future. If this is feasible, this can unblock the stable release of Metrics Sdk.
3. See if it is possible to mark 2 methods in MetricReader trait as experimental. (Collect, register_pipeline are the ones questionable.) This is a smaller version of 2 from above.

I don't expect any disruption to end-users as they don't really have a need to deal with MetricReader. Ability to write custom MetricExporters will be continued as is. 

cc: @fraillt

fraillt:
I would also want to emphasis that with current `MetricReader` design we're not able to make some changes to `PushMetricExporter` (e.g. add `set_resource(&mut self, ...)`, to make all exporters more consistent).

cijothomas:
Questionable APIs are now moved to experimental_feature flags, and that is the extent of what we plan for the coming release.

PR #2921, opened 2025-04-09
refactor: PushMetricExporter interface

fraillt:
## Changes

Changed `PushMetricExporer::export` signature. `fn export(&self, metrics: ResourceMetricsRef<'_>) -> ...`. It is very similar to `LogExporter::export`, with one key difference: I do not expose `iter()` method on items (metrics) (to create iterator), but directly provide Iterator over data. This means, that it's only possible to iterate everything once.
However, since I implement standard `Iterator` trait we still need to have all the items stored somewhere, so I'm not sure how much we can optimize internal implementation...

If I would change `Iterator` to [LendingIterator](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#what-are-gats), it would be possible to collect/export all metrics without any memory allocation at all!

So I have few questions for this PR:
* Do we want to be able to implement collect/export without memory allocation (e.g. have one `Metric` instance on the stack (iterator itself) ?)
* Regarding naming... if we optimize internal implementation it might be that `ResourceMetrics` and `ScopeMetrics` will not be present anymore, so I could use these names here (it sounds better than `ResourceMetricRef`...). In general, any suggestion for better naming is welcome :)

P.S. I wasn't able to provide `set_resource(&mut self, resource: &Resource)` on the `PushMetricExporter` (without making `PeriodicReader` very inefficient, and disabling async version of it) before MetricReader is refactored/improved #2917.

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/metrics.rs
+++ b/opentelemetry-otlp/src/exporter/http/metrics.rs
@@ -4,12 +4,12 @@ use crate::metric::MetricsClient;
 use http::{header::CONTENT_TYPE, Method};
 use opentelemetry::otel_debug;
 use opentelemetry_sdk::error::{OTelSdkError, OTelSdkResult};
-use opentelemetry_sdk::metrics::exporter::ResourceMetricsRef;
+use opentelemetry_sdk::metrics::exporter::ResourceMetrics;
 
 use super::OtlpHttpClient;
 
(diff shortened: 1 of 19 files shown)


fraillt:
This PR contains is a bit old and contains too many code changes, so it's hard to review.
The essence of this PR is moved into #2957

PR #2924, opened 2025-04-10
fix: fix trace id in logs when using set_parent nested in a trace span

### LINKED ISSUE #2824: [Bug]: traceId not correctly set for logs when using set_parent with opentelemetry-appender-tracing
### What happened?

I'm using the tracing crate, tracing_opentelemetry, opentelemetry crates along with opentelemetry-appender-tracing (with experimental_use_tracing_span_context enabled). And noticed an issue when creating a new span and then calling set_parent, the logs exported will have the old traceId created with span! instead of the traceId from parent_context set with [set_parent](https://docs.rs/tracing-opentelemetry/0.29.0/src/tracing_opentelemetry/span_ext.rs.html#43). 

```
let span = span!(Level::INFO, SPAN_NAME);
span.set_parent(parent_context);
let _enter = span.enter();

// the following logging will have the trace ID created from `span!` instead of the parent_context
// when exporting with `opentelemetry-appender-tracing`
tracing::info!("next message in nats");
// if printing span.context().span().span_context().trace_id(), the traceId would be the one from parent_context
```

0.28.0

0.28.0

OTLP

### Relevant log output

```shell

```

SF-Zhou:
Fixes #2824.

## Changes

When using distributed tracing and calling `set_parent` to set the context, the currently obtained `trace_id` was incorrect. This PR fixes this issue.

This fix has been validated in online production environments. For users with the same requirements, they can temporarily opt to use the patched `opentelemetry-appender-tracing-patched` crate.


scottgerring:
Hey @SF-Zhou thanks for opening the PR!
Can you give a few more details on what exactly led to this?

SF-Zhou:
> Hey @SF-Zhou thanks for opening the PR! Can you give a few more details on what exactly led to this?

The description of the PR has been updated.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -246,21 +246,30 @@ where
...
-                root_span
-                    .extensions()
-                    .get::<OtelData>()
-                    .and_then(|otd| otd.builder.trace_id)
-            });
-
-            if let Some((trace_id, span_id)) = opt_trace_id.zip(opt_span_id) {
-                log_record.set_trace_context(trace_id, span_id, None);
+            if let Some(otd) = span.extensions().get::<OtelData>() {
+                if let Some(attributes) = &otd.builder.attributes {
```
cijothomas, line 252: attaching span attributes to log should not be done by default. Let's discuss this in an issue and come up with a design before adding this.

I think this is a very valid capability, and should be offered by the appender natively (on opt-in basis of course). But we need to design this carefully to avoid duplications.
(Another important question is - do we only append the attributes from the immediate parent Span? What about nested span scenarios?)

cijothomas (changes requested):
Thanks!

Requesting changing to address the comment below:
See https://github.com/open-telemetry/opentelemetry-rust/pull/2924/files#r2044837776

opentelemetry-appender-tracing/src/layer.rs
SF-Zhou: Sure, I'll remove the feature for now.

SF-Zhou:
> Thanks!
> 
> Requesting changing to address the comment below: See https://github.com/open-telemetry/opentelemetry-rust/pull/2924/files#r2044837776

Done.

hdost (approved):

SF-Zhou:
@cijothomas could you please review the updates when convenient? If everything looks good, we can proceed to merge.

bantonsson (approved):

PR #2928, opened 2025-04-11
feat: Hide MetricReader and friends

cijothomas:
Option2 from https://github.com/open-telemetry/opentelemetry-rust/issues/2917

Trimming public API, by hiding MetricReader+friends under feature flag. This won't affect normal usage of Metrics, but only custom MetricReader authors. It is unclear what scenario requires custom MetricReaders.....
Ability to write custom Exporter is still maintained.
Also removed `Aggregation` enum to existing View feature, as that was the only use for Aggregation.

Hoping to trim the questionable public API before declaring Metrics SDK as stable.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -23,15 +23,22 @@ also modified to suppress telemetry before invoking exporters.
   - Fixed the overflow attribute to correctly use the boolean value `true`
     instead of the string `"true"`.
     [#2878](https://github.com/open-telemetry/opentelemetry-rust/issues/2878)
-
-- *Breaking* change for custom `MetricReader` authors.
-  The `shutdown_with_timeout` method is added to `MetricReader` trait.
-  `collect` method on `MetricReader` modified to return `OTelSdkResult`.
-  [#2905](https://github.com/open-telemetry/opentelemetry-rust/pull/2905)
 - *Breaking* `MetricError`, `MetricResult` no longer public (except when
   `spec_unstable_metrics_views` feature flag is enabled). `OTelSdkResult` should
   be used instead, wherever applicable.
+- *Breaking* change, affecting custom MetricReader authors:   The
+  `shutdown_with_timeout` method is added to `MetricReader` trait. `collect`
+  method on `MetricReader` modified to return `OTelSdkResult`.
   [#2905](https://github.com/open-telemetry/opentelemetry-rust/pull/2905)
+  [#2905](https://github.com/open-telemetry/opentelemetry-rust/pull/2905)
+- *Breaking* change, affecting custom MetricReader authors: `MetricReader`
+  trait, `ManualReader` struct, `Pipeline` struct, `InstrumentKind` enum moved
+  behind feature flag "experimental_metrics_custom_reader". These were only
+  required for writing custom readers.
+  [2928](https://github.com/open-telemetry/opentelemetry-rust/pull/2928)
+- *Breaking* `Aggregation` enum moved behind feature flag
+  "spec_unstable_metrics_views". This was only required when using Views.
+  [2928](https://github.com/open-telemetry/opentelemetry-rust/pull/2928)
 
 ## 0.29.0
 
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -53,7 +53,7 @@ rt-tokio-current-thread = ["tokio", "tokio-stream", "experimental_async_runtime"
 internal-logs = ["opentelemetry/internal-logs"]
 experimental_metrics_periodicreader_with_async_runtime = ["metrics", "experimental_async_runtime"]
 spec_unstable_metrics_views = ["metrics"]
-experimental_custom_metric_reader = ["metrics"]
+experimental_metrics_custom_reader = ["metrics"]
 experimental_logs_batch_log_processor_with_async_runtime = ["logs", "experimental_async_runtime"]
 experimental_logs_concurrent_log_processor = ["logs"]
 experimental_trace_batch_span_processor_with_async_runtime = ["trace", "experimental_async_runtime"]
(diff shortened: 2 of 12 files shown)


fraillt:
I have looked a bit deeper into this and here's my understanding **(maybe not correct :))** about the situation.
If we ignore code for a bit and look philosophically, the purpose of Metrics SDK is to allow a user to specify **how** and **when** metrics should be exported.
How - is solved by `PushMetricExporter`.
When - is solved by `MetricReader`.
The *how* part is done nicely: it has `export`, `temporality` and few more to deal with technical stuff ("flush" and "shutdown").
The *when* part looks a bit more complex (for my taste at least) than it should be. 
Ideally, user should accept some sort of  `CollectTrigger` instance and provide implementation which would call `fn collect(&mut self)` either in the loop (for PeriodicReader) or some other way (e.g. pull based reader for Prometheus).

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -51,6 +52,7 @@ where
     ///
     /// If this option is not used or `interval` is equal to zero, 60 seconds is
     /// used as the default.
+    #[allow(unused)]
```
utpilla, line 55: It's common for users to override the default export interval. This should be made available to users by default without having to opt-in for any feature.

cijothomas:
> I have looked a bit deeper into this and here's my understanding **(maybe not correct :))** about the situation. If we ignore code for a bit and look philosophically, the purpose of Metrics SDK is to allow a user to specify **how** and **when** metrics should be exported. How - is solved by `PushMetricExporter`. When - is solved by `MetricReader`. The _how_ part is done nicely: it has `export`, `temporality` and few more to deal with technical stuff ("flush" and "shutdown"). The _when_ part looks a bit more complex (for my taste at least) than it should be. Ideally, user should accept some sort of `CollectTrigger` instance and provide implementation which would call `fn collect(&mut self)` either in the loop (for PeriodicReader) or some other way (e.g. pull based reader for Prometheus).

Yes, Users should be in control of export interval - https://github.com/open-telemetry/opentelemetry-rust/pull/2928#discussion_r2039662058
i.e, they can control how often collect/export should get triggered automatically.
Additionally, users can trigger export/collect ad-hoc too, by invoking meter_provider.flush() themselves.
Most users don't need to even know that there is a thing/concept called `MetricReader` at all.

cijothomas:
> I would suggest to rename PushMetricExporter to MetricExporter (because Push part is actually controlled by collector), and would rename MetricReader to MetricCollector.

Agree with removing "Push" from MetricExporter.
The other one - not so sure. Spec calls it MetricReader. But Reader is being proposed to be hidden from end-users via feature flag in this PR, so we can alter it later too.

opentelemetry-sdk/src/metrics/periodic_reader.rs
cijothomas: Based on offline discussions, I kept periodic_reader back as public, so existing behaviors would continue to work as-is.

opentelemetry-sdk/Cargo.toml
```diff
@@ -53,6 +53,7 @@ rt-tokio-current-thread = ["tokio", "tokio-stream", "experimental_async_runtime"
 internal-logs = ["opentelemetry/internal-logs"]
 experimental_metrics_periodicreader_with_async_runtime = ["metrics", "experimental_async_runtime"]
 spec_unstable_metrics_views = ["metrics"]
+experimental_custom_metric_reader = ["metrics"]
```
utpilla, line 56: We have been following this naming convention: `experimental-{signal}-{feature}`

```suggestion
experimental_metrics_custom_reader = ["metrics"]
```

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -23,15 +23,22 @@ also modified to suppress telemetry before invoking exporters.
...
     [#2878](https://github.com/open-telemetry/opentelemetry-rust/issues/2878)
-
-- *Breaking* change for custom `MetricReader` authors.
-  The `shutdown_with_timeout` method is added to `MetricReader` trait.
-  `collect` method on `MetricReader` modified to return `OTelSdkResult`.
-  [#2905](https://github.com/open-telemetry/opentelemetry-rust/pull/2905)
 - *Breaking* `MetricError`, `MetricResult` no longer public (except when
   `spec_unstable_metrics_views` feature flag is enabled). `OTelSdkResult` should
   be used instead, wherever applicable.
+- *Breaking* change, affecting custom MetricReader authors:   The
```
utpilla, line 29:
```suggestion
- *Breaking* change, affecting custom `MetricReader` authors:   The
```

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -23,15 +23,22 @@ also modified to suppress telemetry before invoking exporters.
...
+  `shutdown_with_timeout` method is added to `MetricReader` trait. `collect`
+  method on `MetricReader` modified to return `OTelSdkResult`.
+  [#2905](https://github.com/open-telemetry/opentelemetry-rust/pull/2905)
```
utpilla, line 32: This PR link is added twice.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -39,12 +39,15 @@
...
+#[allow(unreachable_pub)]
+#[allow(unused)]
 pub(crate) mod aggregation;
 pub mod data;
 mod error;
 pub mod exporter;
 pub(crate) mod instrument;
 pub(crate) mod internal;
+#[cfg(feature = "experimental_metrics_custom_reader")]
 pub(crate) mod manual_reader;
```
utpilla, line 51: We should probably just delete `manual_reader.rs` file instead of providing it under a feature. There doesn't seem to be any use case for it.

utpilla (approved):
Left some suggestions.

PR #2936, opened 2025-04-21
ci: Add cargo deny license check

utpilla:
Follow-up to #2935 

## Changes
- Add cargo deny license check

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -7,7 +7,30 @@ allow = [
     "Apache-2.0",
     "ISC",
     "BSD-3-Clause",
-    "OpenSSL"
+]
+
+exceptions = [
+    { allow = ["MPL-2.0"], crate = "webpki-roots" }, # This crate is a dependency of `reqwest`.
+    { allow = ["Unicode-3.0"], crate = "icu_collections" }, # This crate gets used transitively by `reqwest`.
+    { allow = ["Unicode-3.0"], crate = "icu_locid" }, # This crate gets used transitively by `reqwest`.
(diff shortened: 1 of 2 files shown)


deny.toml
```diff
@@ -7,7 +7,30 @@ allow = [
     "Apache-2.0",
     "ISC",
     "BSD-3-Clause",
-    "OpenSSL"
+]
+
+exceptions = [
+    { allow = ["MPL-2.0"], crate = "webpki-roots" }, # This crate is a dependency of `reqwest`.
```
utpilla, line 13: Need to check if these licenses are compatible with our licensing model.

lalitb: I recall seeing it somewhere in CNCF exception list - but good to check.

lalitb (approved):

deny.toml
utpilla: These two are not listed under the allow list: https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist

lalitb: Ok, I think I saw some components with MPL-2 license in the exception list, like this - https://github.com/cncf/foundation/blob/main/license-exceptions/cncf-exceptions-2023-08-31.json

btw, the latest version of webpki-roots seems to be Apache licensed - https://github.com/rustls/webpki-roots/tree/v/0.26.8

utpilla: Also, we don't expose any of these crates' (API or data) through this repo's components, so we should be good.

PR #2959, opened 2025-05-06
fix: allow span links to be added to a SpanRef

bantonsson:
## Changes

This adds the ability to add a `Link` to a `Span` via a `SpanRef`. Since you can only access a `SpanRef` from the current `Context`, there was no way to add a new `Link` to that span.

### DIFF
--- a/opentelemetry/src/trace/context.rs
+++ b/opentelemetry/src/trace/context.rs
@@ -185,6 +185,30 @@ impl SpanRef<'_> {
         self.with_inner_mut(move |inner| inner.update_name(new_name))
     }
 
+    /// Adds a [`Link`] to another [`SpanContext`].
+    ///
+    /// This method allows linking the current span to another span, identified by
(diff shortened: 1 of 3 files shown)


opentelemetry/src/trace/context.rs
```diff
@@ -185,6 +185,26 @@ impl SpanRef<'_> {
         self.with_inner_mut(move |inner| inner.update_name(new_name))
     }
 
+    /// Adds a [`Link`] to another [`SpanContext`].
```
cijothomas, line 188: while you are at it, could you also add a note stating
1. Links added via this mechanism is not accessible for Samplers.
2. It is recommended to add Links at Span creation time, rather than adding it after Span creation.

Relevant spec part:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#link

cijothomas (approved):
LGTM.
Nit: add a changelog entry too as this is user facing addition!

PR #2960, opened 2025-05-06
fix: remove unnecessary thread spawning

therustmonk:
## Changes

Removed an unnecessary thread that was spawned just to create an instance of the blocking client.
Also made similar changes to the other implementations.


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2431 See the original PR which spun up new thread, just to create the blocking::client. It is done to address an issue - so it should not be removed unless the original issue is solved via some other means..

therustmonk:
@cijothomas Thanks!

PR #2963, opened 2025-05-14
fix: PeriodicReader to reuse data structures across collect

cijothomas:
PeriodicReader was unnecessarily creating a new empty `ResourceMetrics` during every collect. This PR modifies periodic reader to create once at startup, and re-use it throughout. Of course, the actual contents change for each collect, but the backing arrays for vectors will be re-used.

### DIFF
--- a/opentelemetry-sdk/src/metrics/periodic_reader.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader.rs
@@ -409,7 +409,7 @@ impl<E: PushMetricExporter> PeriodicReaderInner<E> {
 
         // Relying on futures executor to execute async call.
         // TODO: Pass timeout to exporter
-        futures_executor::block_on(self.exporter.export(&rm))
+        futures_executor::block_on(self.exporter.export(rm))
     }
 
     fn force_flush(&self) -> OTelSdkResult {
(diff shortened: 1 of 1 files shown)


hdost (approved):
Please fix the lint otherwise good 👍

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -378,16 +383,9 @@ impl<E: PushMetricExporter> PeriodicReaderInner<E> {
...
-        // owned data structures to be passed to exporters.
-        let mut rm = ResourceMetrics {
-            resource: Resource::empty(),
-            scope_metrics: Vec::new(),
-        };
-
+    fn collect_and_export(&self, rm: &mut ResourceMetrics) -> OTelSdkResult {
         let current_time = Instant::now();
-        let collect_result = self.collect(&mut rm);
+        let collect_result = self.collect(rm);
```
utpilla, line 388: I guess this still works without the need to clear the `scope_metrics` vec within the `ResourceMetrics`. I'm not sure if we have tests for this already but we need to ensure that clearing is not required by verifying that subsequent collects don't continue to show the old data from the previous collect (in case of Delta).

utpilla (approved):

PR #2968, opened 2025-05-14
fix: Avoid exposing implementation detail in public API for PushMetricExporter

cijothomas:
Continuing from https://github.com/open-telemetry/opentelemetry-rust/pull/2965

PR #2969, opened 2025-05-14
fix: Avoid exposing HistogramBuckets and bounds

cijothomas:
Last part of https://github.com/open-telemetry/opentelemetry-rust/pull/2968


opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -250,9 +250,9 @@ pub struct HistogramDataPoint<T> {
...
     ///
     /// Because the last boundary is +infinity this one is implied.
-    pub bounds: Vec<f64>,
+    pub(crate) bounds: Vec<f64>,
     /// The count of each of the buckets.
-    pub bucket_counts: Vec<u64>,
+    pub(crate) bucket_counts: Vec<u64>,
 
     /// The minimum value recorded.
     pub min: Option<T>,
```
utpilla, line 258: Should we make all struct fields internal and provide getters for a consistent experience?

utpilla (approved):

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -275,6 +275,16 @@ impl<T> HistogramDataPoint<T> {
     pub fn exemplars(&self) -> impl Iterator<Item = &Exemplar<T>> {
         self.exemplars.iter()
     }
+
+    /// Returns an iterator over the bucket boundaries in [HistogramDataPoint].
+    pub fn bounds(&self) -> impl Iterator<Item = &f64> {
```
utpilla, line 280: We could make the API more ergonomic by returning `impl Iterator<Item = f64>` instead of `impl Iterator<Item = &f64>`:

```rust
/// Returns an iterator over the bucket boundaries in [HistogramDataPoint].
pub fn bounds(&self) -> impl Iterator<Item = f64> + '_ {
    self.bounds.iter().copied()
}

/// Returns an iterator over the bucket counts in [HistogramDataPoint].
pub fn bucket_counts(&self) -> impl Iterator<Item = u64> + '_ {
    self.bucket_counts.iter().copied()
}
```

cijothomas: yes makes sense. Made this change and updated exporters accordingly.

PR #2981, opened 2025-05-20
fix: Metrics Views - fix a bug that causes unit, description to be lost when applying views that influence other aspects

cijothomas:
unit, desc were lost when the View did not explicitly specify them. This PR fixes this bug. If any aspect is not specified in the View's Stream configuration, it'll use the default (if advisory is available, that is considered first before default)

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -42,6 +42,8 @@ also modified to suppress telemetry before invoking exporters.
   behind feature flag "experimental_metrics_custom_reader".
   [#2928](https://github.com/open-telemetry/opentelemetry-rust/pull/2928)
 
+- TODO: Placeholder for View related changelog.
+
 - *Breaking* `Aggregation` enum moved behind feature flag
   "spec_unstable_metrics_views". This was only required when using Views.
   [#2928](https://github.com/open-telemetry/opentelemetry-rust/pull/2928)
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -42,7 +42,9 @@ also modified to suppress telemetry before invoking exporters.
   behind feature flag "experimental_metrics_custom_reader".
   [#2928](https://github.com/open-telemetry/opentelemetry-rust/pull/2928)
 
-- TODO: Placeholder for View related changelog.
+- TODO: Placeholder for View related changelog. Polish this after all
+  changes are done.
+  Hide public fields from `Stream` struct.
 
 - *Breaking* `Aggregation` enum moved behind feature flag
   "spec_unstable_metrics_views". This was only required when using Views.
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -2702,6 +2702,7 @@ mod tests {
             unreachable!()
         };
 
+
         // Expecting (cardinality_limit + 1 overflow + Empty attributes) data points.
         assert_eq!(sum.data_points.len(), cardinality_limit + 1 + 1);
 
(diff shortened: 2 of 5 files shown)


opentelemetry-sdk/CHANGELOG.md
```diff
@@ -42,6 +42,10 @@ also modified to suppress telemetry before invoking exporters.
   behind feature flag "experimental_metrics_custom_reader".
   [#2928](https://github.com/open-telemetry/opentelemetry-rust/pull/2928)
 
+- TODO: Placeholder for View related changelog. Polish this after all
```
cijothomas, line 45: There will be another change in a followup PR to change Stream to use a proper builder pattern and return Result. the changelog will be finally updated after all changes are in.
@utpilla

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1463,6 +1463,144 @@ mod tests {
         );
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn view_test_rename() {
```
utpilla, line 1467: These tests and the helper method need not be async.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1463,6 +1463,144 @@ mod tests {
...
+            },
+            "my_counter",
+            "my_unit_new",
+            "my_description",
+        )
+        .await;
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn view_test_match_none() {
```
utpilla, line 1535: Add tests cases for:
- Updating all the fields
- Updating multiple fields (for example, name and unit but not description)

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -1463,6 +1463,144 @@ mod tests {
...
+            .build();
+
+        counter.add(1.5, &[KeyValue::new("key1", "value1")]);
+        meter_provider.force_flush().unwrap();
+
+        // Assert
+        let resource_metrics = exporter
+            .get_finished_metrics()
+            .expect("metrics are expected to be exported.");
+        assert!(!resource_metrics.is_empty());
```
utpilla, line 1585: Also, verify that count is one to ensure that we only create one stream for this.

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -403,16 +413,16 @@ where
 
             otel_debug!(
                 name : "Metrics.InstrumentCreated",
-                instrument_name = stream.name.as_ref(),
+                instrument_name = stream.name.clone().unwrap_or_default().as_ref(),
```
utpilla, line 416: We are being forced to introduce a lot of `unwrap_or_default()` method calls even though we know that the stream provided to `cached_aggregator` method would always have some name, unit, and description.

We should ideally create another internal struct that would have the same fields as `Stream` but they wouldn't be optional. Using such a struct instead of `Stream` would allow to get rid of these `unwrap_or_default()` calls which have weakened the contract for `cached_aggregator` method.

opentelemetry-sdk/src/metrics/view.rs
```diff
@@ -110,7 +110,7 @@ pub fn new_view(criteria: Instrument, mask: Stream) -> MetricResult<Box<dyn View
     let contains_wildcard = criteria.name.contains(['*', '?']);
 
     let match_fn: Box<dyn Fn(&Instrument) -> bool + Send + Sync> = if contains_wildcard {
-        if !mask.name.is_empty() {
+        if mask.name.is_some() {
```
utpilla, line 113: We still might have to check for empty string. A user could still provide an empty string as `Some(String::new())`.

utpilla (approved):
Left some suggestions.

opentelemetry-sdk/src/metrics/pipeline.rs
cijothomas: good point! we can modify this separately, as we use the unwrap_default for buckets, attribute_keys too.

PR #2982, opened 2025-05-21
feat: Promote subset of Metric Views to stable

cijothomas:
Ability to change name, desc, unit and cardinality limit is promoted to not require experimental opt-in.

`Stream` is expected to be moved to a builder pattern with validation in the next PR, which will also eliminate the need for `new_view` helper method.

### DIFF
--- a/opentelemetry-sdk/src/metrics/pipeline.rs
+++ b/opentelemetry-sdk/src/metrics/pipeline.rs
@@ -363,6 +363,9 @@ where
         kind: InstrumentKind,
         mut stream: Stream,
     ) -> MetricResult<Option<Arc<dyn internal::Measure<T>>>> {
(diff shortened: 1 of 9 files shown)


opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -363,6 +363,9 @@ where
         kind: InstrumentKind,
         mut stream: Stream,
     ) -> MetricResult<Option<Arc<dyn internal::Measure<T>>>> {
+        // TODO: Create a separate pub (crate) Stream struct for the pipeline,
+        // as Stream will not have any Optional fields as None at this point and
```
utpilla, line 367: nit:
```suggestion
        // as Stream will not have any optional fields as None at this point and
```

utpilla (approved):

PR #2988, opened 2025-05-21
feat: View cleanups

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/2985

Continuing from https://github.com/open-telemetry/opentelemetry-rust/pull/2984 and https://github.com/open-telemetry/opentelemetry-rust/pull/2982

This removes the `new_view` way of creating a View. Instead users are expected to pass anything implementing `View` trait. For convenience, `View` is implemented for `Fn(&Instrument) -> Option<Stream>`, so this can be used to add a view.

This allowed a lot of other cleanups. If there is a need, more additive changes can be done to make it easier to apply view for the common scenario.

Example of future additions (not part of this PR!):
```rust
meter_provider_builder().with_view_on_instrument(name: "Foo", steam: Stream::builder().name("foo-new"));
meter_provider_builder().with_view_on_instrument(name: "Foo*", steam: Stream::builder().unit("unit-new"));
meter_provider_builder().with_view_on_instrument(name: "*", steam: Stream::builder().aggregation(Aggregation::Drop));
```

PR #2989, opened 2025-05-22
fix: Further trim public API on views

cijothomas:
Leftover cleanup from https://github.com/open-telemetry/opentelemetry-rust/pull/2988

`View` trait itself is not required to be public - no functionality change.
Plus few other leftover nits from PR comments earlier.

### DIFF
--- a/opentelemetry-sdk/src/metrics/instrument.rs
+++ b/opentelemetry-sdk/src/metrics/instrument.rs
@@ -64,12 +64,12 @@ impl InstrumentKind {
         }
     }
 }
-
-/// Describes properties an instrument is created with, used for filtering in
-/// [View](crate::metrics::View)s.
+/// Describes the properties of an instrument at creation, used for filtering in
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -185,8 +185,8 @@ impl StreamBuilder {
     /// Set the stream allowed attribute keys.
     ///
     /// Any attribute recorded for the stream with a key not in this set will be
-    /// dropped. If the set is empty, all attributes will be dropped, if `None` all
-    /// attributes will be kept.
+    /// dropped. If the set is empty, all attributes will be dropped.
+    /// If not set, all attributes will be kept.
```
utpilla, line 189: I think we can make it more clear without overloading the word "set". Even the line above mentions "set".
```suggestion
    /// If this method is not used, all attributes will be kept.
```

utpilla (approved):

Issue #2986, opened 2025-05-21
[Bug] Metric Views allow bypassing validations.


cijothomas:
Views can be used to alter name, unit, aggregation etc. However, the validations are currently applied to the values provided at instrument creation time, and not after the view has been applied.

cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/pull/5094 I actually misunderstood the spec! This issue can be closed and there is a PR already (draft until spec merges) to fix the spec-drift.

PR #2991, opened 2025-05-22
fix: Fix validation in Metric stream

cijothomas:
Most copy-pasted from instrument creation validation for now. A future refactor can move this to a common place and fully address #2986


opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -207,27 +212,53 @@ impl StreamBuilder {
...
+
+        // Validate cardinality limit
         if let Some(limit) = self.cardinality_limit {
             if limit == 0 {
                 return Err("Cardinality limit must be greater than 0".into());
             }
         }
 
-        // If the aggregation is set to ExplicitBucketHistogram, validate the bucket boundaries.
+        // Validate bucket boundaries if using ExplicitBucketHistogram
```
utpilla, line 261: Now that we are doing the validation while building the stream, we could get rid of this code:

https://github.com/open-telemetry/opentelemetry-rust/blob/8c29ca7e21cf0c041695a50d683504f9cbeca515/opentelemetry-sdk/src/metrics/internal/histogram.rs#L95-L100

utpilla (approved):

opentelemetry-sdk/src/metrics/instrument.rs
cijothomas: good point, will tackle it soon.

PR #2997, opened 2025-05-22
fix: Remove pub fields and replace with getter method consistently across …

cijothomas:
We previously hid the implementation details (eg: Vec!), and exposed getter returning slices. To ensure consistency, this PR changes all fields to be accessible via getter methods only.

### DIFF
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -383,11 +383,6 @@ impl<T> HistogramDataPoint<T> {
     pub fn count(&self) -> u64 {
         self.count
     }
-
-    /// Returns the sum of the values recorded.
-    pub fn sum(&self) -> &T {
-        &self.sum
-    }
 }
 
 impl<T: Copy> HistogramDataPoint<T> {
@@ -399,6 +394,11 @@ impl<T: Copy> HistogramDataPoint<T> {
     /// Returns the maximum value recorded.
     pub fn max(&self) -> Option<T> {
         self.max
+    }
+
+    /// Returns the sum of the values recorded.
+    pub fn sum(&self) -> T {
+        self.sum
     }
 }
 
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -395,19 +494,78 @@ impl<T> ExponentialHistogramDataPoint<T> {
...
         self.exemplars.iter()
     }
+
+    /// Returns the number of updates this histogram has been calculated with.
+    pub fn count(&self) -> usize {
+        self.count
+    }
+
+    /// Returns the sum of the values recorded.
+    pub fn sum(&self) -> &T {
```
utpilla, line 504: Return `T` instead of `&T`.

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -395,19 +494,78 @@ impl<T> ExponentialHistogramDataPoint<T> {
...
+}
+
+impl ExponentialBucket {
+    /// Returns the bucket index of the first entry in the counts vec.
+    pub fn offset(&self) -> i32 {
+        self.offset
+    }
+
+    /// Returns a reference to the counts vec.
+    pub fn counts(&self) -> &[u64] {
```
utpilla, line 566: Use `impl Iterator<Item = u64>` as the return type?

opentelemetry-sdk/src/metrics/data/mod.rs
```diff
@@ -317,6 +378,28 @@ impl<T> HistogramDataPoint<T> {
...
         self.bucket_counts.iter().copied()
     }
+
+    /// Returns the number of updates this histogram has been calculated with.
+    pub fn count(&self) -> u64 {
+        self.count
+    }
+
+    /// Returns the sum of the values recorded.
+    pub fn sum(&self) -> &T {
```
utpilla, line 388: Return `T` instead of `&T`.

opentelemetry-sdk/src/metrics/data/mod.rs
cijothomas: done

utpilla (approved):

PR #2998, opened 2025-05-23
Add doc on how to set right cardinality limit

cijothomas:
Overall doc is still WIP, but adding some more useful information about setting right cardinality limit.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -594,23 +594,42 @@ Consider these guidelines when determining the appropriate limit:
   above), use the same calculation approach as cumulative temporality.
 * For dynamic scenarios where not all combinations appear in every export cycle,
   base the limit on expected total measurements within a single interval.
-* **Example 1:** With a 60-second export interval and 1,000 measurements per
-  interval, set the cardinality limit to 1,000. Delta temporality allows the SDK
-  to reset after each export, accommodating different attribute combinations
-  across intervals without accumulating state.
+* **Example 1:** If your application generates at most 1,000 distinct attribute
+  combinations per export interval (regardless of the interval duration), set
+  the cardinality limit to 1,000. Delta temporality allows the SDK to reset
+  after each export, accommodating different attribute combinations across
+  intervals without accumulating state.
 * **Example 2:** For web applications with known Request Per Second (RPS) rates,
-  calculate the maximum measurements per interval: `RPS × Export Interval`. With
-  500 RPS and a 60-second interval: `500 × 60 = 30,000` measurements per cycle.
-  Set the cardinality limit to 30,000.
+  calculate the maximum measurements per interval: `RPS × Export Interval`
+  (assuming one measurement per request). With 500 RPS and a 60-second interval:
+  `500 × 60 = 30,000` measurements per cycle. Set the cardinality limit to
+  30,000.
 * **High-Cardinality Attributes:** Delta temporality excels with attributes like
-  `user_id` where not all values appear simultaneously. Base the limit on
-  concurrent active users within an interval rather than total possible users.
-  Using the same calculation (`500 RPS × 60 seconds = 30,000`), this
-  accommodates realistic concurrent user activity.
+  `user_id` where not all values appear simultaneously. Due to delta temporality's
+  state-clearing behavior and the fact that not all users are active within a
+  single interval, you can set a cardinality limit much lower than the total
+  possible cardinality. For example, even with millions of registered users, if
+  only 30,000 are active per interval (based on `500 RPS × 60 seconds`), the
+  cardinality limit can be set to 30,000 rather than millions.
 * **Export Interval Tuning:** Reducing export intervals lowers cardinality
   requirements. With 30-second intervals: `500 × 30 = 15,000` measurements,
   allowing a lower limit. However, balance this against increased serialization
   and network overhead from more frequent exports.
+
+**3. Backend Compatibility Considerations:**
+
+While delta temporality offers significant advantages for cardinality
+management, your choice may be constrained by backend support:
+
+* **Backend Restrictions:** Some metrics backends only support cumulative
+  temporality. For example, Prometheus requires cumulative temporality and
+  cannot directly consume delta metrics.
+* **Collector Conversion:** To leverage delta temporality's memory advantages
+  while maintaining backend compatibility, configure your SDK to use delta
+  temporality and deploy an OpenTelemetry Collector with a delta-to-cumulative
+  conversion processor. This approach pushes the memory overhead from your
+  application to the collector, which can be more easily scaled and managed
+  independently.
 
 TODO: Add the memory cost incurred by each data points, so users can know the
 memory impact of setting a higher limits.
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -560,7 +560,63 @@ The exported metrics would be:
   words, attributes used to create `Meter` or `Resource` attributes are not
   subject to this cap.
 
-// TODO: Document how to pick cardinality limit.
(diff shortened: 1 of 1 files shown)


docs/metrics.md
```diff
@@ -560,7 +560,82 @@ The exported metrics would be:
...
+* **Limit too low**: Measurements will be folded into the overflow bucket
+  (`{"otel.metric.overflow": true}`), losing granular attribute information and
+  making attribute-based queries unreliable.
+
+Consider these guidelines when determining the appropriate limit:
+
+**1. For Cumulative Temporality:**
+
+* Cumulative metrics retain their state across export cycles, so the total
+  cardinality equals the Cartesian product of all attribute combinations.
```
utpilla, line 583: We want the count of items in the Cartesian product not the product itself.

color = {"red", "green", "yellow"}
name = { "apple", banana"}

Cartesian product = {("red", "apple"), ('green", "apple"), ...}

```suggestion
  cardinality equals the cardinality of the Cartesian product of all attribute sets.
```

docs/metrics.md
```diff
@@ -560,7 +560,82 @@ The exported metrics would be:
...
+
+* Delta temporality resets aggregations after each export cycle, enabling more
+  flexible cardinality management based on actual usage patterns rather than
+  theoretical maximums.
+* When all possible attribute combinations are known (eg: Fruit example from
+  above), use the same calculation approach as cumulative temporality.
+* For dynamic scenarios where not all combinations appear in every export cycle,
+  base the limit on expected total measurements within a single interval.
+* **Example 1:** If your application generates at most 1,000 distinct attribute
+  combinations per export interval (regardless of the interval duration), set
```
utpilla, line 598: I think saying "per export interval regardless of the interval" makes it a bit confusing.

```suggestion
  combinations throughout the application lifetime (regardless of the export interval duration), set
```

docs/metrics.md
```diff
@@ -560,7 +560,82 @@ The exported metrics would be:
...
+  the cardinality limit to 1,000. Delta temporality allows the SDK to reset
+  after each export, accommodating different attribute combinations across
+  intervals without accumulating state.
+* **Example 2:** For web applications with known Request Per Second (RPS) rates,
```
utpilla, line 602: I have some concerns about using this example. It might be confusing to people who rely on metrics to calculate RPS in the first place.

docs/metrics.md
```diff
@@ -560,7 +560,82 @@ The exported metrics would be:
...
+**2. For Delta Temporality:**
+
+* Delta temporality resets aggregations after each export cycle, enabling more
+  flexible cardinality management based on actual usage patterns rather than
+  theoretical maximums.
+* When all possible attribute combinations are known (eg: Fruit example from
+  above), use the same calculation approach as cumulative temporality.
+* For dynamic scenarios where not all combinations appear in every export cycle,
+  base the limit on expected total measurements within a single interval.
+* **Example 1:** If your application generates at most 1,000 distinct attribute
```
utpilla, line 597: How about putting the examples along with their respective scenarios?

- When all possible attribute combinations are known... Use Example1 here
- For dynamic scenarios where not all combinations appear in every export cycle, ... Use Example 2 here

docs/metrics.md
lalitb: I think this would be more accurate:

``so the total cardinality equals the size of the Cartesian product of all attribute combinations seen so far.``

or even simpler

``so the total cardinality equals the number of unique attribute value combinations observed so far.``

docs/metrics.md
```diff
@@ -560,7 +560,85 @@ The exported metrics would be:
   words, attributes used to create `Meter` or `Resource` attributes are not
   subject to this cap.
 
-// TODO: Document how to pick cardinality limit.
+#### Cardinality Limits - How to Choose the Right Limit
```
scottgerring, line 563: How about we link to some of the very detailed OTel docs about cardinality limits for more info? 

https://opentelemetry.io/docs/specs/otel/metrics/sdk/#cardinality-limits

https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/ (covers cardinality explosion)

cijothomas: this is already added in the cardinality limit section above. this PR is specifically adding section about setting the right limit. 
(I'll add supplementary guidelines link at the end too)

docs/metrics.md
cijothomas: removed this.

docs/metrics.md
cijothomas: rewrote the section.

docs/metrics.md
```diff
@@ -560,7 +560,101 @@ The exported metrics would be:
...
+normal operation.**
+
+###### Export Interval Tuning
+
+Shorter export intervals further reduce the required cardinality:
+
+* If your interval is halved (e.g., from 60s to 30s), the number of unique
+  attribute sets seen per interval may also be halved.
+
+But beware: more frequent exports increase CPU/network overhead due to
```
utpilla: Use "NOTE" or "CAUTION"

```suggestion
NOTE: More frequent exports increase CPU/network overhead due to
```

docs/metrics.md
```diff
@@ -560,7 +560,101 @@ The exported metrics would be:
...
+  attribute.
+
+###### Example - Fruit Sales Scenario
+
+Attributes:
+
+* `name` can be "apple" or "lemon" (2 values)
+* `color` can be "red", "yellow", or "green" (3 values)
+
+You may observe up to 2 × 3 = 6 unique attribute sets.
```
utpilla: It would be nice to call out the difference between theoretical max and practical max here.  It would be natural for someone reading this example to question why the limit should be 6 when we know that we would never emit metrics for a "red lemon". 

We could mention something like the simplest way is to use the theoretical max (which is the product of each attribute's cardinality), however, if you know that certain combinations of the attributes will not be valid then you can reduce that number from the limit.

docs/metrics.md
```diff
@@ -560,7 +560,101 @@ The exported metrics would be:
...
+* For scenarios with predictable attribute combinations (as in the fruit
+  example), you can use the same calculation method as with cumulative
+  temporality.
+* Delta temporality particularly excels when dealing with high-cardinality
+  dimensions like `user_id`, where the active set changes dynamically over time
+  and only a subset is active during any given export interval.
+
+###### Example - High Cardinality Attribute Scenario
+
+Attributes:
```
utpilla, line 622: Add export interval as well the prerequisite for this example to make it very clear.

```suggestion
Export interval: 60 sec

Attributes:
```

docs/metrics.md
```diff
@@ -560,7 +560,101 @@ The exported metrics would be:
...
+
+You may observe up to 2 × 3 = 6 unique attribute sets.
+
+**Set the cardinality limit to 6.**
+
+##### Choosing the Right Limit for Delta Temporality
+
+Delta metrics reset their aggregation state after every export interval. This
+approach enables more efficient memory utilization by focusing on attributes
+observed during each interval rather than tracking all possible combinations.
```
utpilla:
```suggestion
observed during each export interval rather than tracking all possible combinations.
```

docs/metrics.md
```diff
@@ -560,7 +560,101 @@ The exported metrics would be:
...
+
+* For scenarios with predictable attribute combinations (as in the fruit
```
utpilla: It could be "predictable" even for the `user_id` example. Maybe something like this:

```suggestion
* For scenarios with a reasonable number (<10,000) of attribute combinations (as in the fruit
```

docs/metrics.md
cijothomas: good callout, made a tweak to call this out.

PR #3003, opened 2025-05-23
fix: add panic error when multiple HTTP features are enabled

### LINKED ISSUE #2994: [Bug]: NoHttpClient error when reqwest-client feature is enabled
### What happened?

When the `reqwest-client` feature is enabled it produces an `ExporterBuildError::NoHttpClient` when a `SpanExporter` (potentially other exporters as well, I've only tested the span version) is built.

I've found that the `reqwest-client` feature conflicts with the default `reqwest-blocking-client` in `opentelemetry-otlp/src/exporter/http/mod.rs` as none of the `cfg` blocks end up being included. The block you would expect to be included because of the `reqwest-client` feature is excluded due to the `not(feature = "reqwest-blocking-client")` and vice versa for the `reqwest-blocking-client` block.

## Reproduction
Cargo.toml:
```toml
[package]
name = "repro-no-http-client"
version = "0.1.0"
edition = "2024"

[dependencies]
opentelemetry = "0.29.1"
opentelemetry-otlp = { version = "0.29.0", features = ["reqwest-client"] }
```

main.rs:
```rust
use opentelemetry_otlp::SpanExporter;

fn main() {
    let exporter = SpanExporter::builder().with_http().build().unwrap();

    println!("Exporter: {:#?}", exporter);
}
```

0.29.1

0.29.0

OTLP

### Relevant log output

```shell
thread 'main' panicked at src/main.rs:4:64:
called `Result::unwrap()` on an `Err` value: NoHttpClient
```

gruebel:
Fixes #2994 

## Changes

- I didn't add it to the changelog, because it fails already, but with a not very helpful error message 😅 

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -161,6 +161,14 @@ impl HttpExporterBuilder {
                     .unwrap(), // TODO: Return ExporterBuildError::ThreadSpawnFailed
                 ) as Arc<dyn HttpClient>);
             }
+            #[cfg(any(
+                all(feature = "hyper-client", feature = "reqwest-client"),
+                all(feature = "hyper-client", feature = "reqwest-blocking-client"),
+                all(feature = "reqwest-client", feature = "reqwest-blocking-client")
+            ))]
+            {
+                panic!("Can't enable more than one HTTP client features simultaneously. Please choose only one: hyper-client, reqwest-client, or reqwest-blocking-client (default feature)");
+            }
         }
 
         let http_client = http_client.ok_or(ExporterBuildError::NoHttpClient)?;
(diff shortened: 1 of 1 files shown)


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -161,6 +161,14 @@ impl HttpExporterBuilder {
                     .unwrap(), // TODO: Return ExporterBuildError::ThreadSpawnFailed
                 ) as Arc<dyn HttpClient>);
             }
+            #[cfg(any(
+                all(feature = "hyper-client", feature = "reqwest-client"),
+                all(feature = "hyper-client", feature = "reqwest-blocking-client"),
+                all(feature = "reqwest-client", feature = "reqwest-blocking-client")
+            ))]
+            {
+                panic!("Can't enable more than one HTTP client features simultaneously. Please choose only one: hyper-client, reqwest-client, or reqwest-blocking-client (default feature)");
```
cijothomas, line 170: could we return a Err here (https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/mod.rs#L107-L137) instead of panic?

gruebel: I changed it to a normal error. `compile_error!` would work, but it would also fail clippy linting with all features enabled 😅

PR #3006, opened 2025-05-24
fix: with_cleared_baggage

gabotechs:
Fixes `with_cleared_baggage` method in `opentelemtry::Context`. Without this change, the function is just not clearing the baggage as the new empty baggage is not being wrapped with the appropriate `BaggageContextValue`

## Changes

- wraps the `Baggage` struct with a `BaggageContextValue` in the `with_cleared_baggage`  method
- adds a unit test that fails on main but succeeds in this branch

* N/A Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* N/A Changes in public API reviewed (if applicable)

### DIFF
--- a/opentelemetry/src/baggage.rs
+++ b/opentelemetry/src/baggage.rs
@@ -412,7 +412,7 @@ impl BaggageExt for Context {
     }
 
     fn with_cleared_baggage(&self) -> Self {
-        self.with_value(Baggage::new())
+        self.with_baggage(Baggage::new())
     }
 
     fn baggage(&self) -> &Baggage {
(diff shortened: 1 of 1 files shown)


opentelemetry/src/baggage.rs
```diff
@@ -412,7 +412,7 @@ impl BaggageExt for Context {
     }
 
     fn with_cleared_baggage(&self) -> Self {
-        self.with_value(Baggage::new())
+        self.with_value(BaggageContextValue(Baggage::new()))
```
paullegranddc, line 415: nit: this looks slightly simpler

```suggestion
        self.with_baggage(Baggage::new())
```

paullegranddc (approved):
LGTM

Issue #2940, opened 2025-04-22
[Parent issue]: SpanProcessor API refactor


paullegranddc:
Parent issue for `SpanProcessor` API related issues

paullegranddc:
@cijothomas I don't think I have the permissions to link sub-issues,  but here are the ones I think make sense:

* [Tracing SDK - Processors get clone of SpanData which is incorrect](https://github.com/open-telemetry/opentelemetry-rust/issues/2726)  
* [[Feature]: Reading spans through the public API requires cloning everything to SpanData](https://github.com/open-telemetry/opentelemetry-rust/issues/2939)

PR #3010, opened 2025-05-30
feat(span_processor): add on_ending callback

paullegranddc:
## Issues

SpanProcessor API refactor  https://github.com/open-telemetry/opentelemetry-rust/issues/2940

### Add  `on_ending` method on thr `SpanProcessor` trait. [Spec link](https://github.com/open-telemetry/opentelemetry-specification/blob/e5bc8e18e647a47b25d264b0c67ce6c0d0e1ec93/specification/trace/sdk.md#onending)

This method is called during span ending, before `on_end`. it is given a mutable span and is very useful to implement obfuscation for instance.
It makes possible the use cases that were the motivation for passing mutable data to `on_end` while respecting the spec.

The feature is still marked as Development in the spec so it's kept behind an experimental feature flag.

This PR cherry picked from #2962


opentelemetry-sdk/src/trace/span.rs
```diff
@@ -198,26 +198,33 @@ impl opentelemetry::trace::Span for Span {
...
-        // skip if data has already been exported
-        let mut data = match self.data.take() {
-            Some(data) => data,
-            None => return,
-        };
-
-        let provider = self.tracer.provider();
         // skip if provider has been shut down
-        if provider.is_shutdown() {
+        if self.tracer.provider().is_shutdown() {
```
bantonsson, line 202: Is there a particular reason to re-read the `provider`?

bantonsson (approved):

opentelemetry-sdk/src/trace/span.rs
paullegranddc: Yes
We need to pass an `&mut Span` to `on_ending`. But if we grab the `provider` at the beginning of the function it's a immutable borrow that lasts for the whole function.

So we need to:
* Borrow the span immutably to read the provider. Not keep the provider around to release the borrow
* Borrow the span mutably for on_ending
* Borrow the span immutably again to grab the provider span processors

Calling to `self.tracer.provider()` is essentially zero cost. It just takes a reference to a field on the `Span`

examples/tracing-http-propagator/src/server.rs
```diff
@@ -105,6 +105,49 @@ async fn router(
...
+    fn on_start(&self, _span: &mut opentelemetry_sdk::trace::Span, _cx: &Context) {}
+
+    fn on_ending(&self, span: &mut opentelemetry_sdk::trace::Span) {
+        let mut obfuscated_attributes = Vec::new();
+        let Some(span) = span.exported_data() else {
+            return;
+        };
+        for KeyValue { key, value, .. } in span.attributes {
+            if let Some(redacted_uri) = obfuscate_http_auth_url(value.as_str().as_ref()) {
+                obfuscated_attributes.push((key.clone(), KeyValue::new(key.clone(), redacted_uri)));
```
cijothomas, line 143: mm...doesn't looks like the `span` is mutated with the updated attribute?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -79,6 +79,24 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
     /// synchronously on the thread that started the span, therefore it should
     /// not block or throw exceptions.
     fn on_start(&self, span: &mut Span, cx: &Context);
+
+    #[cfg(feature = "experimental_span_processor_on_ending")]
+    /// `on_ending` is called when a `Span` is ending. The end timestamp has already
+    /// been computed.
+    /// This method is called synchronously within the `Span::end` API, therefore it
+    /// should not block or throw an exception.
```
cijothomas, line 87: nit: remove the "throw exception" part as it won't fit Rust...

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -725,4 +732,70 @@ mod tests {
...
+
+        let exporter = crate::trace::span_processor::tests::MockSpanExporter::new();
+        let spans = exporter.exported_spans.clone();
+
+        let provider = crate::trace::SdkTracerProvider::builder()
+            .with_span_processor(FirstProcessor)
+            .with_span_processor(SecondProcessor)
+            .with_simple_exporter(exporter)
+            .build();
+        provider.tracer("test").start("test_span");
```
cijothomas, line 790: test suggestion:
start the span with some attributes.
then add few more attributes to it.
validate that on_start sees only those attributes added at start time.
validate that on_ending sees attributes added at start time, attributes added via span.setattribute, attributes added by any previous processor.

paullegranddc: The test has been updated with your suggestions

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -79,6 +79,24 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
...
+    ///
+    /// If multiple span processors are registered, their on_ending methods are invoked
+    /// in the order the span processors have been registered, and mutations to the span
+    /// will be visible to the next processor.
+    ///
+    /// The tracer will call `on_ending` for all span processors before calling `on_end`
```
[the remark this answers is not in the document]
paullegranddc: I reworked the comment a bit

examples/tracing-http-propagator/src/server.rs
```diff
@@ -105,6 +105,49 @@ async fn router(
     response
 }
 
+fn obfuscate_http_auth_url(s: &str) -> Option<String> {
```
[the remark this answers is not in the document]
paullegranddc: Sure, I added a processor that tags the span as error if it has a status code attribute `>=400`
I feel like this is both realistic and fairly simple

bantonsson (approved):

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -966,7 +983,7 @@ impl BatchConfigBuilder {
 }
 
 #[cfg(all(test, feature = "testing", feature = "trace"))]
-mod tests {
+pub(crate) mod tests {
```
cijothomas, line 986: we need to follow up on this and move shared utils to common wherever possible. (not a blocker for this PR)

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -80,6 +80,23 @@ pub trait SpanProcessor: Send + Sync + std::fmt::Debug {
     /// not block or throw exceptions.
     fn on_start(&self, span: &mut Span, cx: &Context);
 
+    /// `on_ending` is called after a `Span` is ended before [`SpanProcessor::on_end`] is invoked on any processor.
+    /// The end timestamp has already been set.
+    ///
+    /// Unlike [`SpanProcessor::on_start`], which only observes attributes provided at span creation,
```
cijothomas, line 86: this is a good place to clarify when to use on_ending vs on_ended too.
Also on_ended has some notes about accessing current context, which is applicable here too - lets add that too here.

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -200,26 +200,33 @@ impl opentelemetry::trace::Span for Span {
...
-        match provider.span_processors() {
+        #[cfg(feature = "experimental_span_processor_on_ending")]
+        {
+            let provider = self.tracer.provider().clone();
+            for processor in provider.span_processors() {
+                processor.on_ending(self);
+            }
+        }
+
+        let Some(data) = self.data.take() else { return };
```
cijothomas, line 227: Trying to understand this logic - I know it is existing, but why! It's like an SpanProcessor can block an item from reaching the next processor by consuming `data`.... which is not something spec has....

paullegranddc: Not it cannot, because data is not a pub field.
So a span processor defined outside of this module can only interact with the span through the pub methods which do not consume the data

opentelemetry-sdk/src/trace/span_processor.rs
paullegranddc: Ok, would you be happy with this wording?

```
    /// `on_ending` is called after a `Span` is ended before [`SpanProcessor::on_end`] is invoked on any processor.
    /// The end timestamp has already been set.
    ///
    /// Unlike [`SpanProcessor::on_start`], which only observes attributes provided at span creation,
    /// `on_ending` observes all attributes accumulated during the span's lifetime,
    /// including those added via `Span::set_attribute`.
    ///
    /// If multiple processors are registered, their `on_ending` methods are called
    /// in registration order and mutations are visible to subsequent processors.
    ///
    /// `on_ending` receives a mutable finished span, which processors can modify
    /// and the modifications are reflected in all exporters, whereas `on_end` receives a frozen span which is
    /// not shared between processors
    ///
    /// Should not block or panic.
    ///
    /// # Accessing Context
    ///
    /// **Important**: Do not rely on [`Context::current()`] in `on_ending`. When `on_ending`
    /// is called during span cleanup, `Context::current()` returns whatever context
    /// happens to be active at that moment, which is typically unrelated to the span
    /// being ended. Contexts can be activated in any order and are not necessarily
    /// hierarchical.
```

cijothomas: yes. Minor - observes all attributes (and links) accumulated during the span's lifetime

PR #2685, opened 2025-02-19
Span Exporter async native

lalitb:
## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #3028, opened 2025-06-18
fix(async-processor): concurrent exports actually serialised

alexbrt:
https://github.com/open-telemetry/opentelemetry-rust/pull/2685 unintentionally broke parallel exports by awaiting the `export()` future directly in `opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs`, rather than passing it to the runtime for concurrent polling. As a result, `OTEL_BSP_MAX_CONCURRENT_EXPORTS` became ineffective, serialising all exports and increasing the risk of dropped spans under load.

This PR restores true parallelism by respecting `max_concurrent_exports`, and adds tests to verify:
- Exports run in parallel when `max_concurrent_exports` > 1
- Exports are serialised when `max_concurrent_exports` == 1


opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -188,13 +190,19 @@ struct BatchSpanProcessorInternal<E, R> {
     spans: Vec<SpanData>,
     export_tasks: FuturesUnordered<BoxFuture<'static, OTelSdkResult>>,
     runtime: R,
-    exporter: E,
+    exporter: Arc<RwLock<E>>,
     config: BatchConfig,
 }
 
-impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
+impl<E: SpanExporter + Send + Sync + 'static, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
```
lalitb: nit - is `Send + Sync` bound here redundant, as the SpanExporter trait is already defined as:

```rust
pub trait SpanExporter: Send + Sync + Debug {
```

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -188,13 +190,19 @@ struct BatchSpanProcessorInternal<E, R> {
     spans: Vec<SpanData>,
     export_tasks: FuturesUnordered<BoxFuture<'static, OTelSdkResult>>,
     runtime: R,
-    exporter: E,
+    exporter: Arc<RwLock<E>>,
```
lalitb: Going forward, the `SpanExporter` trait should be redesigned to use immutable references (&self) for all methods. This would allow us to remove the `RwLock` and use just `Arc<E>` for sharing the exporter across concurrent tasks - similar to how `LogExporter` is implemented.

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
alexbrt: Good catch, thanks!

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
alexbrt: Agreed — my goal was mainly to minimise the scope of the change, but you're right that using a RwLock is... questionable.

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -243,9 +251,15 @@ impl<E: SpanExporter, R: RuntimeChannel> BatchSpanProcessorInternal<E, R> {
                         self.export_tasks.next().await;
                     }
 
-                    let export_result = self.export().await;
+                    let batch = self.spans.split_off(0);
+                    let exporter = self.exporter.clone();
+                    let runtime = self.runtime.clone();
```
lalitb, line 259: I believe these clones should be cheap - `Arc` cloning for exporter, and ZST cloning for runtime.

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
lalitb: Can you add a TODO comment for this so someone can work on it? I'll also file an issue. Thanks.

opentelemetry-sdk/Cargo.toml
```diff
@@ -55,7 +55,7 @@ spec_unstable_metrics_views = ["metrics"]
 experimental_metrics_custom_reader = ["metrics"]
 experimental_logs_batch_log_processor_with_async_runtime = ["logs", "experimental_async_runtime"]
 experimental_logs_concurrent_log_processor = ["logs"]
-experimental_trace_batch_span_processor_with_async_runtime = ["trace", "experimental_async_runtime"]
+experimental_trace_batch_span_processor_with_async_runtime = ["rt-tokio", "trace", "experimental_async_runtime"]
```
cijothomas: is it intentional to include `rt-tokio`?

alexbrt: I included it to wrap the exporter in a `tokio::sync::RwLock` and allow obtaining a mutable reference to `self` during shutdown.

lalitb: Didn't realize this. This would be an issue if the user wants to use non-tokio runtime.

alexbrt: I believe the `sync` module is compatible with any runtime as per tokio's [documentation](https://docs.rs/tokio/latest/tokio/sync/index.html#runtime-compatibility):
> All synchronization primitives provided in this module are runtime agnostic. You can freely move them between different instances of the Tokio runtime or even use them from non-Tokio runtimes.

I updated the `Cargo.toml` file to be more explicit and only pull in `tokio/sync`.

opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs
```diff
@@ -188,13 +190,22 @@ struct BatchSpanProcessorInternal<E, R> {
     spans: Vec<SpanData>,
     export_tasks: FuturesUnordered<BoxFuture<'static, OTelSdkResult>>,
     runtime: R,
-    exporter: E,
     config: BatchConfig,
+    // TODO: Redesign the `SpanExporter` trait to use immutable references (`&self`)
+    // for all methods. This would allow us to remove the `RwLock` and just use `Arc<E>`,
+    // similar to how `crate::logs::LogExporter` is implemented.
+    exporter: Arc<RwLock<E>>,
```
TommyCpp, line 197: Emmm if we need to add RwLock to support concurrent export maybe we should consider offer two favor of span processor so users that don't use concurrent export don't have to pay the cost

TommyCpp (approved):
> Redesign the `SpanExporter` trait to use immutable references (`&self`)

OK to merge this PR as it is but we should fix this before next release so we don't cause a preformance regression

lalitb (approved):
LGTM. RwLock (and so the dependency on tokio/sync) should be removed in subsequent iteration. Will create a tracking issue for same.

PR #3031, opened 2025-06-18
Remove spec_unstable_logs_enabled feature -copilot try1 #3029

### LINKED ISSUE #3020: Stabilize logger.enabled in sdk
once https://github.com/open-telemetry/opentelemetry-specification/pull/4536 is merged and released in the spec.

The "spec_unstable_logs_enabled" feature flag needs to be removed completely. All methods/code should be enabled always.
Do this for opentelemetry, opentelemetry_sdk, opentelemetry-appender-tracing and opentelemetry-appender-log crates.
Also update changelog.md file for all above crates to inform of the above change.

Once done, do a thorough search across repo to find any "spec_unstable_logs_enabled" references, and remove them. If there are places where it cannot be removed, create issues to track as follow ups.

cijothomas:
Fixes #3020

### DIFF
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -89,9 +89,8 @@
 //!
 //! # Feature Flags
 //!
-//! This library provides the following Cargo features:
+//! All log features are now always enabled. The previous `spec_unstable_logs_enabled` feature flag has been removed.
 //!
-//! - `spec_unstable_logs_enabled`: Allow users to control the log level.
 //! - `with-serde`: Support complex values as attributes without stringifying them.
 //!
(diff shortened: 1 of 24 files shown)


opentelemetry-appender-log/src/lib.rs
```diff
@@ -89,9 +89,8 @@
 //!
 //! # Feature Flags
 //!
-//! This library provides the following Cargo features:
+//! All log features are now always enabled. The previous `spec_unstable_logs_enabled` feature flag has been removed.
```
utpilla, line 92: That doesn't sound right. Only the  event_enabled feature is being made always enabled. Users would still have to opt in for other optional features such as `with-serde` in this case.

cijothomas: Correct. Silly mistake by Copilot!

PR #3057, opened 2025-07-04
fix: Support HttpClient implementation for HyperClient with custom connectors

### LINKED ISSUE #3056: [Feature]: Support HttpClient implementation for HyperClient with custom connectors
### Related Problems?

The current implementation of `HttpClient` for `HyperClient` in `opentelemetry-http/src/lib.rs` is limited to only work with the default `HttpConnector`. This prevents users from using custom connectors such as:
- Unix domain socket connectors
- Custom TLS connectors
- Proxy connectors
- Any other custom connection implementations

This limitation makes it impossible to use the OpenTelemetry HTTP client with non-standard connection types, which is a common requirement in production environments.

For example, Datadog officially recommends using Unix Domain Socket endpoints for their trace collector in Kubernetes environments (https://docs.datadoghq.com/containers/kubernetes/apm/?tab=datadogoperator). The current limitation prevents OpenTelemetry users from following this best practice when exporting traces to Datadog.

The `HttpClient` implementation should be generic over the connector type `C`, allowing `HyperClient<C>` to implement `HttpClient` for any `C` that satisfies the necessary trait bounds (Connect + Clone + Send + Sync + 'static).

The implementation should change from:
```rust
impl HttpClient for HyperClient
```

To:
```rust
impl<C> HttpClient for HyperClient<C>
where
    C: Connect + Clone + Send + Sync + 'static,
    HyperClient<C>: Debug,
```

The change is backward compatible as the existing `HyperClient` type alias defaults to using `HttpConnector`, so existing code will continue to work without modification.

### Considered Alternatives

Creating wrapper types for each custom connector type - This would lead to code duplication and maintenance burden.

### Additional Context

_No response_

palindrom615:
Fixes #3056

## Changes

The `HttpClient` implementation becomes generic over the connector type `C`, allowing `HyperClient<C>` to implement `HttpClient` for any `C` that satisfies the necessary trait bounds (Connect + Clone + Send + Sync + 'static).

### DIFF
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## vNext
 
 - Implementation of `Extractor::get_all` for `HeaderExtractor`
+- Support `HttpClient` implementation for `HyperClient<C>` with custom connectors beyond `HttpConnector`, enabling Unix Domain Socket connections and other custom transports
 
 ## 0.30.0
 
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -179,7 +179,11 @@ pub mod hyper {
     }
 
     #[async_trait]
-    impl HttpClient for HyperClient {
+    impl<C> HttpClient for HyperClient<C>
+    where
+        C: Connect + Clone + Send + Sync + 'static,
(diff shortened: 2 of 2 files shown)


lalitb (approved):
Thanks.

bantonsson (approved):

PR #3063, opened 2025-07-11
docs: Add OTLP arch docs

### LINKED ISSUE #3062: Document current OTLP architecture

scottgerring:
Fixes #3062 , documenting the architecture of the OTLP subsystem. I've written this as I dive into OTLP as part of reviewing it against the specification and figure it is helpful context for others working on opentelemetry-rust. 

## Changes


docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Everything Else
+
+* **OTLP** – [design doc](./otlp.md) — the OTLP exporter subsystem.
+
+## 5 Extensibility Hooks
+| Layer | Customisation points |
+|-------|---------------------|
+| API | Provide alternative `TracerProvider`, `MeterProvider`, `LoggerProvider` |
+| SDK | Plug-in **samplers**, **metric readers**, **log processors** |
```
TommyCpp: You can alos plugin your own span processor?

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+
+## 2. Overall Structure of `opentelemetry-rust` and OTLP
+```mermaid
+graph TD
+    A[Application code] --> B[opentelemetry-api]
+    B --> C[opentelemetry-sdk processors]
+    C --> D[opentelemetry-otlp exporters]
+    D -->|gRPC / HTTP| E[OTel Collector / Backend]
+    %% Auxiliary exporters
+    C --> F[opentelemetry-jaeger]
```
TommyCpp: I thought we deprecated jaeger already. Do we still want to mention it?

TommyCpp (approved):
Nice document. The only concern I have with those kind of doc is they become stale over time if we don't actively maintain them. Could we add the commit that this is based on? At least so that we have some reference to what's the last update time

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+* **Logs** – [design doc](./logs.md) — timestamped events bridged from existing logging frameworks (`log`, `tracing`).
+
+### Everything Else
+
+* **OTLP** – [design doc](./otlp.md) — the OTLP exporter subsystem.
+
+## 5 Extensibility Hooks
+| Layer | Customisation points |
+|-------|---------------------|
+| API | Provide alternative `TracerProvider`, `MeterProvider`, `LoggerProvider` |
```
cijothomas: API does not offer customization points...Once can swap out the official SDK with another one, but there is no such option for API...

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+Key points:
+1. The **API** crates are dependency-free facades that application code instruments against.
+2. The **SDK** crate supplies concrete implementations, batching, aggregation, and lifecycle management.  
+   • **Processors / readers** live inside the SDK and adapt buffering, temporality, and semantics.
+3. **Exporters** translate OTel data into wire formats (OTLP, Prometheus, etc.) and handle transport.
+
+## 3 Cross-cutting Components
+* **Resource & Attributes** – common schema for describing process, host, and service; see the [resource spec](https://opentelemetry.io/docs/specs/otel/resource/).
+* **Context & Propagation** – context management and carrier injection; see the [context propagation spec](https://opentelemetry.io/docs/specs/otel/context/).
+* **Runtime model** – public APIs are runtime-agnostic; heavy I/O lives behind optional Tokio-based exporters; see the [SDK design guidelines](https://opentelemetry.io/docs/specs/otel/).
```
cijothomas: not so sure of the statement here..This seems quite specific to SDK. I/O/exports can occur without any runtimes..

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+    E -->|"protocols (OTLP, Jaeger, Zipkin, Prometheus, stdout)"| F[Back-end / Collector]
+```
+Key points:
+1. The **API** crates are dependency-free facades that application code instruments against.
+2. The **SDK** crate supplies concrete implementations, batching, aggregation, and lifecycle management.  
+   • **Processors / readers** live inside the SDK and adapt buffering, temporality, and semantics.
+3. **Exporters** translate OTel data into wire formats (OTLP, Prometheus, etc.) and handle transport.
+
+## 3 Cross-cutting Components
+* **Resource & Attributes** – common schema for describing process, host, and service; see the [resource spec](https://opentelemetry.io/docs/specs/otel/resource/).
```
cijothomas: Resource and Attributes are different, better to separately list them.
Also "A Resource is a representation of the entity producing telemetry" is the wording from spec, lets stick with that too.

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+## 1 Purpose
+This document provides a stable, high-level description of how the OpenTelemetry Rust implementation is put together.  Detailed per-signal design notes live in their own files; this overview ties the pieces together.
+
+> Reference: the [OpenTelemetry Specification](https://opentelemetry.io/docs/specs/otel/) and its [GitHub repository](https://github.com/open-telemetry/opentelemetry-specification/)
+
+## 2 Layering
+```mermaid
+graph TD
+    A[Application code] --> B[opentelemetry API]
+    B --> C[opentelemetry-sdk core]
```
cijothomas: "core"  - there is no official thing called "core".

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+    C --> D[Processors / Readers]
+    D --> E[Exporters]
+    E -->|"protocols (OTLP, Jaeger, Zipkin, Prometheus, stdout)"| F[Back-end / Collector]
+```
+Key points:
+1. The **API** crates are dependency-free facades that application code instruments against.
```
cijothomas: what does "dependency-free" mean here? 
Also, API is used not just by application code, but library too

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+2. The **SDK** crate supplies concrete implementations, batching, aggregation, and lifecycle management.  
+   • **Processors / readers** live inside the SDK and adapt buffering, temporality, and semantics.
```
cijothomas: not sure what is "semantics" refer to here?

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+* **Resource & Attributes** – common schema for describing process, host, and service; see the [resource spec](https://opentelemetry.io/docs/specs/otel/resource/).
+* **Context & Propagation** – context management and carrier injection; see the [context propagation spec](https://opentelemetry.io/docs/specs/otel/context/).
+* **Runtime model** – public APIs are runtime-agnostic; heavy I/O lives behind optional Tokio-based exporters; see the [SDK design guidelines](https://opentelemetry.io/docs/specs/otel/).
+* **Error taxonomy** – see [ADR 001 Error Handling](../adr/001_error_handling.md).
+
+## Detailed Design
+
+### Signals
+
+* **Traces** – [design doc](./traces.md) — trees of _spans_ capturing distributed work.
```
cijothomas: I suggest *not* giving a definition/description to the signals. We can just state OTel Traces,Logs, Metrics and link to the respective spec doc.

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+* **Metrics** – [design doc](./metrics.md) — numerical time-series data.
+* **Logs** – [design doc](./logs.md) — timestamped events bridged from existing logging frameworks (`log`, `tracing`).
+
+### Everything Else
```
cijothomas:
```suggestion
### Exporters
```

docs/design/architecture.md
```diff
@@ -0,0 +1,46 @@
...
+
+* **OTLP** – [design doc](./otlp.md) — the OTLP exporter subsystem.
```
cijothomas:
```suggestion
* **OTLP** – [design doc](./otlp.md) — the OTLP exporter.
```

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
+# OpenTelemetry Rust – OTLP Architecture
+
+This document describes how the **OTLP exporters** are organised and how they integrate with the wider OpenTelemetry Rust SDK.  Review findings and compliance tracking live in `opentelemetry-otlp/arch-review.md`.
```
cijothomas: > Review findings and compliance tracking live in `opentelemetry-otlp/arch-review.md`.

Not sure what this is referring to?

docs/design/otlp.md
cijothomas: this is OTLP design doc, so no need to mention any other exporters..

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+    B --> C[opentelemetry-sdk processors]
+    C --> D[opentelemetry-otlp exporters]
+    D -->|gRPC / HTTP| E[OTel Collector / Backend]
+    %% Auxiliary exporters
+    C --> F[opentelemetry-jaeger]
+    C --> G[opentelemetry-zipkin]
+    C --> H[opentelemetry-prometheus]
+```
+Key points:
+* `opentelemetry-sdk` owns batching, aggregation, and lifecycle; **exporters only handle serialization + transport**.
```
cijothomas: exporter may also do retry policy too.

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+* OTLP exporters live in a separate crate to keep big deps (`tonic`, `reqwest`) optional.
```
cijothomas: don't think we need to state that separate crate is to keep big deps optional.

cijothomas: even exporters like stdout, that has no deps are separate crate.

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+---
+
+## 10. Key Architectural Decisions
+| Decision | Rationale |
+|----------|-----------|
+| *Builder pattern with marker types* | Compile-time guarantee that exactly one transport is chosen. |
+| *Transport-specific modules* | Keep heavy deps (`tonic`, `reqwest`) behind feature-flags to minimise compile times. |
+| *Env-vars hierarchy* | Matches OTLP spec; makes container/K8s configuration straightforward. |
```
cijothomas: Is there a spec on hierarchy? Lets link to the specific section to be precise

docs/design/otlp.md
scottgerring: I think it is helpful to see the extension point by way of example - "bolt your stuff on here if you want". But I can certainly change it to be generic!

docs/design/otlp.md
scottgerring: ... this is an accidental preview of the OTLP review-against-spec doc i've been writing in parallel  😆  i'll remove it.

docs/design/architecture.md
scottgerring: fair!

docs/design/otlp.md
scottgerring: added!

scottgerring:
> Nice document. The only concern I have with those kind of doc is they become stale over time if we don't actively maintain them. Could we add the commit that this is based on? At least so that we have some reference to what's the last update time

Good idea! Added the date ✅ Can always get more granular with `git blame`. 
I am hopeful this is written at the right level to provide 1/ a useful arch overview that is harder to tease out of the code and 2/ not have to change very often.

docs/design/architecture.md
```diff
@@ -0,0 +1,50 @@
...
+1. The **API** crate is a lightweight facade that library and application code instrument against.
+2. The **SDK** crate supplies concrete implementations, batching, aggregation, and lifecycle management.  
+   • **Processors / readers** live inside the SDK and adapt buffering and temporality.
+3. **Exporters** translate OTel data into wire formats (OTLP, Prometheus, etc.) and handle transport.
+
+## 3 Cross-cutting Components
+* **Resource** – a representation of the entity producing telemetry - see the [resource spec](https://opentelemetry.io/docs/specs/otel/resource/).
+* **Attributes** - a key-value pair used to annotate telemetry data - see the [common spec](https://opentelemetry.io/docs/specs/otel/common/#attributes).
+* **Context & Propagation** – context management and carrier injection; see the [context propagation spec](https://opentelemetry.io/docs/specs/otel/context/).
+* **Runtime model** — public APIs are agnostic to execution model (synchronous or asynchronous); implementations may support optional async exporters (e.g., Tokio) or blocking I/O, depending on configuration. See the [SDK design guidelines](https://opentelemetry.io/docs/specs/otel/).
```
cijothomas: See the [SDK design guidelines](https://opentelemetry.io/docs/specs/otel/)

^ Wrong link?

docs/design/architecture.md
```diff
@@ -0,0 +1,50 @@
...
+
+* **OTLP** – [design doc](./otlp.md) — the OTLP exporter.
+
+## 5 Extensibility Hooks
+| Layer | Customisation points |
+|-------|---------------------|
+| SDK | Provide complete alternative SDK |
+| SDK | Plug-in **samplers**, **metric readers**, **log processors**, and **providers** |
+| Exporters | Support additional wire-protocols |
+| Processors | Provide alternative `LogProcessor` or `SpanProcessor` |
```
cijothomas: its unclear what we are conveying here. What does "alternative" Processor mean?

cijothomas: this can be removed as its already covered above - plug-in

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+OpenTelemetry Protocol (OTLP) is the vendor-agnostic wire format used by OpenTelemetry to transport telemetry signals (traces, metrics, logs) from instrumented code to a backend (OTel Collector or compatible vendor). It supports:
+
+* **Encodings**  
+  * *Protobuf* – canonical format  
+  * *JSON* (optional)
+* **Transports**  
+  * *gRPC* (default port `4317`)  
+  * *HTTP/1.1*, *HTTP/2* (default port `4318`, binary protobuf or JSON payloads)
+* **Signals**  
+  * Traces, Metrics, Logs – transported independently but share common envelope (`Resource`, `Scope` etc.)
```
cijothomas: nit: Envelop types are common. The actual envelop isn't shared.

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+
+## 3. Crate Layout
+```text
+src/
+  lib.rs                 # Re-exports + feature flags + protocol enum
+  span.rs | metric.rs | logs.rs   # Signal-specific builders/exporters
+  exporter/
+     ├── mod.rs          # Common builder traits, env var parsing, compression
+     ├── http/           # Reqwest blocking/async clients, body encoder
+     └── tonic/          # Tonic clients, TLS/compression helpers
```
cijothomas:
```suggestion
     └── tonic/          # Tonic client, TLS/compression helpers
```

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+| `gzip-tonic` `zstd-tonic` | gRPC message compression | `exporter/tonic` |
+| `reqwest-client` / `reqwest-blocking-client` *(default)* | Choose async vs blocking Reqwest HTTP client |
+| `hyper-client` | Use Hyper HTTP transport *(requires both Reqwest features disabled)* |
+| TLS helpers (`tls-roots`, `tls-webpki-roots`, `reqwest-rustls`) | Supply trust roots for TLS |
+
+Because **only one transport is valid per exporter**, `protocol` is currently a *hint* – unsupported values are ignored by the concrete builder.
+
+### 4.1 Transport-specific Client Options
+
+Exporters must function both in binaries that start no async runtime and in services already running Tokio.  Consequently the crate provides **blocking** and **non-blocking** HTTP clients behind mutually-exclusive feature-flags.  The blocking variant is the default to ensure out-of-the-box operability; applications that already depend on Tokio can opt into an async client to avoid spawning extra threads.
```
cijothomas: nit: instead of stating "Tokio", could we say "async runtimes like Tokio".

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+
+| Implementation | Feature flag | Blocking? | Underlying crate | Transports | Typical use case |
```
cijothomas, line 75: Unclear what we mean by "Blocking" - this could be quite misleading as users might think OTel blocks user-threads..

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+|---------------|--------------|-----------|------------------|-----------|------------------|
+| Reqwest (blocking) **(default)** | `reqwest-blocking-client` | Yes | `reqwest::blocking` | HTTP/1.1 → auto-upgrades to HTTP/2 over TLS | CLI tools, synchronous binaries, or hybrid apps where introducing Tokio is undesirable. A helper thread is spawned once at builder time.
+| Reqwest (async) | `reqwest-client` | No (Tokio) | `reqwest` | HTTP/1.1 → auto-upgrades to HTTP/2 over TLS | Services that already run a Tokio runtime and prefer fully non-blocking exports.
+| Hyper | `hyper-client` | No (Tokio) | `hyper` | HTTP/1.1 → auto-upgrades to HTTP/2 over TLS | Lean dependency footprint when both Reqwest features are disabled.
+| Tonic (gRPC) | `grpc-tonic` | No (Tokio) | `tonic` | HTTP/2 (gRPC) | Environments standardising on gRPC or collectors exposing only port 4317.
```
cijothomas: The "typical use case" header is not adding value IMHO. If a user want to use gRPC, then they use tonic/gRPC.

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+    Exp->>Tx: serialize + send
+    Tx--)Collector: OTLP request
+    note right of Collector: 2xx = success,<br/>non-2xx → error handling
+    Tx--)Exp: Future resolves
+    Exp--)SDK: OTelSdkResult
+```
+Notes: 
+
+* Serialization happens **inside the transport client** module to keep exporters thin.
+* `export` is `async` for `tonic` / `reqwest` clients, surfaced via `SpanExporter` implementing `opentelemetry_sdk::trace::SpanExporter`.
```
cijothomas: nit: should we be mention all 3 exporter traits instead of just Span?

cijothomas: is it correct to state exporter is "async" for only tonic/reqwest? the trait method for export returns a Future...

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+
+Builder fails fast, runtime exporter surfaces errors through `export()` futures so processors (or Retry logic) can decide whether to back-off, drop, or escalate.
+
+---
+
+## 8. Extension & Customisation Points
+1. **Custom headers / metadata** – `.with_metadata(map)` (gRPC) or `.with_headers()` (HTTP).
+2. **Compression** – `.with_compression(Compression::Gzip | Compression::Zstd)` behind feature-flags.
+3. **TLS** – via `TonicConfig` or `HttpConfig`; root-store helper features embed common CA bundles.
+4. **Alternate HTTP client** – inject any `HttpClient` implementation via `.with_http_client()`.
```
cijothomas, line 134: this is something missing from docs I think! Do we actually support bring-your-own-client capability!

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+5. **Protocol JSON** – switch to JSON payloads at build-time with the `http-json` feature.
+
+---
+
+## 9. Interactions with Other Exporters
```
cijothomas: don't think this is worth mentioning in architecture...

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+---
+
+## 10. Key Architectural Decisions
+| Decision | Rationale |
+|----------|-----------|
+| *Builder pattern with marker types* | Compile-time guarantee that exactly one transport is chosen. |
+| *Transport-specific modules* | Keep heavy deps (`tonic`, `reqwest`) behind feature-flags to minimise compile times. |
+| *Separate crate* | Avoids pulling heavy deps into projects that don’t need OTLP. |
```
cijothomas: not sure if its worth mentioning here. even stdout exporter is own crate!

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+| *Non-exhaustive error enums* | Allows adding new failure modes without breaking semver. |
```
cijothomas: this is still open/not-finalized!

docs/design/otlp.md
```diff
@@ -0,0 +1,171 @@
...
+    .with_simple_exporter(jaeger)
+    .build();
+```
+
+---
+
+## 10. Key Architectural Decisions
+| Decision | Rationale |
+|----------|-----------|
+| *Builder pattern with marker types* | Compile-time guarantee that exactly one transport is chosen. |
```
cijothomas, line 142: we do have issues about users accidentally enable multiple feature flags...

docs/design/architecture.md
```diff
@@ -0,0 +1,50 @@
+# OpenTelemetry Rust – Architecture Overview
+
+> Last Updated: 2025-07-16
+
+## 1 Purpose
+This document provides a stable, high-level description of how the OpenTelemetry Rust implementation is put together.  Detailed per-signal design notes live in their own files; this overview ties the pieces together.
```
cijothomas: given OTLP exporters are not stable, should we call this stable yet?

docs/design/architecture.md
scottgerring: As far as I can see there's actually nothing in the spec that has much to say about impls using async exporters. It's an internal detail of the exporter itself. Removing the link altogether.

docs/design/otlp.md
scottgerring: fair; I'll use 'synchronous' which makes more sense in the context of the subsequent rows

docs/design/otlp.md
scottgerring: Doesn't it? Aside from "I want to use gRPC so I should use the only gRPC option" I don't think choosing between the others is obvious. It might be something that belongs [here](https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/) alternatively/as-well even!

docs/design/otlp.md
scottgerring: yeah, no 🤣  even the sync ones are yielding a future (although presumably it'll block a thread while it runs). Fixed.

docs/design/otlp.md
scottgerring: The code's there and it's a public trait and method so it should work!

docs/design/architecture.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Exporters
+
+* **OTLP** – [design doc](./otlp.md) — the OTLP exporter.
+
+## 5 Extensibility Hooks
+| Layer | Customisation points |
+|-------|---------------------|
+| SDK | Provide complete alternative SDK |
+| SDK | Plug-in **samplers**, **metric readers**, **log processors**, and **providers** |
```
cijothomas: nit: there is no plugin model for provider itself

```suggestion
| SDK | Plug-in **samplers**, **metric readers**, **log processors** |
```

docs/design/architecture.md
```diff
@@ -0,0 +1,50 @@
+# OpenTelemetry Rust – Architecture Overview
+
+> Status:Development 
+> Last Updated: 2025-07-16
```
cijothomas, line 4: nit: We have git blame for this!

cijothomas (approved):
Thanks for putting this together. We can iterate and improve it as we work on OTLP towards stability!

PR #3077, opened 2025-07-18
feat: generate protos for profiles collector

korniltsev:
## Changes

This change allows setting up a profiles collector client. This will allow to send profiles to collectors from Rust.


opentelemetry-proto/src/proto.rs
```diff
@@ -295,7 +302,7 @@ pub mod tonic {
     #[path = ""]
     pub mod profiles {
         #[path = "opentelemetry.proto.profiles.v1development.rs"]
-        pub mod v1;
+        pub mod v1development;
```
korniltsev, line 305: Note for reviewers: it renames the package which is a breaking change, but the package should be `v1development` and it is unstable and breaking changes are expected in the `v1development`

korniltsev: cc @cijothomas

korniltsev: cc @adriangb tagging you as you did the initial bootstrap

lalitb: Can we mention this specifically as breaking change in changelog. The renaming looks fine to me to clearly mark the API as unstable.

adriangb: seems reasonable to me apologies for not doing it this way in the initial pr

lalitb (approved):

opentelemetry-proto/src/proto.rs
korniltsev: Thanks. Applied a note on breaking change in the changelog.

PR #3078, opened 2025-07-19
refactor: Align global trace with metrics

andrzejressel:
## Changes

This pull requests align setting global tracer with settings global metrics - mem::replace was replaced with mutable assignment and old tracer is no longer returned.

In addition failure to set global tracer does not crash program anymore - it will return noop and log error - like with global metrics


opentelemetry/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## vNext
 
+- Change return type of `opentelemetry::global::set_tracer_provider` to Unit to align with metrics counterpart
```
cijothomas: please call it as a *Breaking Change*

andrzejressel: Fixed

PR #3079, opened 2025-07-23
perf: Compare different span creation scenarios

cijothomas:
Left existing benches untouched for now. Will consolidate all in a future date.
This is primarily comparing the costs of various span creation. Given there are multiple ways, we need to document the right choices for each scenario (or eliminate the choices). These benches are just starting point.

### DIFF
--- a/opentelemetry-stdout/src/trace/exporter.rs
+++ b/opentelemetry-stdout/src/trace/exporter.rs
@@ -94,18 +94,28 @@ fn print_spans(batch: Vec<SpanData>) {
             });
 
         println!();
-        println!("\tName        : {}", &span.name);
-        println!("\tTraceId     : {}", &span.span_context.trace_id());
-        println!("\tSpanId      : {}", &span.span_context.span_id());
-        println!("\tTraceFlags  : {:?}", &span.span_context.trace_flags());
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/benches/span.rs
```diff
@@ -0,0 +1,179 @@
...
+        span.set_attribute(KeyValue::new("key4", "world"));
+        span.set_attribute(KeyValue::new("key5", 123));
+        span.end();
+    });
+
+    trace_benchmark_group(c, "span-creation-tracer-in-span", |tracer| {
+        // This is similar to the simple span creation, but also does the job of activating
+        // the span in the current context.
+        // It is slower than other approaches of activation due to the fact that
+        // context activation is done, irrespective of sampling decision.
```
lalitb, line 72: Also, I remember you mentioned cost of mutex lock invoked here? Do we need to mention that in comment?

lalitb (approved):

opentelemetry-sdk/benches/span.rs
cijothomas: will do in the next iteration. Everytime span is retrieved from Context, this mutex gets invoked in all subsequent Span operations.

Issue #2877, opened 2025-03-27
Fix telemetry-induced-telemetry in OTLP Exporter


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2868/files fixes the issue of Otel's own operations generating telemetry and feeding that to OTel itself. However, this works only if Otel's Context is properly propagated.

This is *not* done by libraries like `hyper`, `tonic` etc. when they cross thread-boundaries. This issue is opened to address this.

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/issues/761 linking to old issue for the same problem.

PR #3084, opened 2025-07-25
fix: Explore how to achieve telemetry suppression with OTLP

cijothomas:
One way of addressing https://github.com/open-telemetry/opentelemetry-rust/issues/2877

This PR does not introduce a “fix” inside the OTLP Exporters themselves, but instead demonstrates how users can address the issue without requiring changes in OpenTelemetry.

**Background**

OpenTelemetry provides a mechanism to suppress telemetry based on the current Context. However, this suppression only works if every component involved properly propagates OpenTelemetry’s Context. Libraries like tonic and hyper are not aware of OTel’s Context and therefore do not propagate it across threads.

As a result, OTel’s suppression can fail, leading to telemetry-induced-telemetry—where the act of exporting telemetry (e.g., sending data via tonic/hyper) itself generates additional telemetry. This newly generated telemetry is then exported again, triggering yet more telemetry in a loop, potentially overwhelming the system.

**What this PR does**

OTLP/gRPC exporters rely on the tonic client, which captures the current runtime at creation time and uses it to drive futures. Instead of reusing the application’s existing runtime, this PR creates a dedicated Tokio runtime exclusively for the OTLP Exporter.

In this dedicated runtime:
	1.	We intercept on_start / on_stop events.
	2.	Sets OTel’s suppression flag in the context.

This ensures that telemetry generated by libraries such as hyper/tonic will be suppressed **only** within the exporter’s dedicated runtime. If those same libraries are used elsewhere for application logic, they continue to function normally and emit telemetry as expected.

Depending on the feedback, we could either address this purely through documentation and examples, or we could enhance the OTLP Exporter itself to expose a feature flag that, when enabled, would automatically create the tonic client within its own dedicated runtime.

### DIFF
--- a/opentelemetry-otlp/examples/basic-otlp/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp/src/main.rs
@@ -61,7 +61,7 @@ fn init_logs() -> SdkLoggerProvider {
 }
 
 thread_local! {
-    static SUPPRESS_GUARD: RefCell<Option<opentelemetry::ContextGuard>> = RefCell::new(None);
+    static SUPPRESS_GUARD: RefCell<Option<opentelemetry::ContextGuard>> = const { RefCell::new(None) };
 }
 
 // #[tokio::main]
--- a/opentelemetry-otlp/examples/basic-otlp/src/main.rs
+++ b/opentelemetry-otlp/examples/basic-otlp/src/main.rs
@@ -1,12 +1,13 @@
 use opentelemetry::trace::{TraceContextExt, Tracer};
-use opentelemetry::KeyValue;
 use opentelemetry::{global, InstrumentationScope};
+use opentelemetry::{Context, KeyValue};
 use opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge;
 use opentelemetry_otlp::{LogExporter, MetricExporter, SpanExporter};
 use opentelemetry_sdk::logs::SdkLoggerProvider;
 use opentelemetry_sdk::metrics::SdkMeterProvider;
 use opentelemetry_sdk::trace::SdkTracerProvider;
 use opentelemetry_sdk::Resource;
+use std::cell::RefCell;
 use std::error::Error;
 use std::sync::OnceLock;
 use tracing::info;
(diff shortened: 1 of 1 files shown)


opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -59,33 +60,37 @@ fn init_logs() -> SdkLoggerProvider {
...
 }
 
-#[tokio::main]
-async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
-    let logger_provider = init_logs();
+thread_local! {
+    static SUPPRESS_GUARD: RefCell<Option<opentelemetry::ContextGuard>> = const { RefCell::new(None) };
+}
+
+// #[tokio::main]
```
cijothomas, line 67: If user's application relies on tokio, then they can create a Rt themselves, and wrap their main method inside rt.blockon({...app code..}), ensuring telemetry initialization is outside of it.

scottgerring:
Although I can see that this works, it feels like a big leak of impl details into the user's domain. What would it look like as a helper in OTel itself? E.g. something like `withTelemetrySuppression(_ => { /* setup otel here  */ } )` ? 

I expect this would still require the user to not use a `tokio_main` but rather explicitly create their runtime after setting up OTel using this helper to wrap, but this would still go a ways to make it feel a bit less leaky.

My other question is - do we impact any of our users by requiring a separate tokio runtime in this case, for instance, folks in resource-constrained environments?

scottgerring:
It feels like our first suggestion to users should be:

> If you're not using tonic/hyper as a HTTP client in your app, simply tune your tracing subscriber to suppress their telemetry

Then this is is only necessary for a subset of users

cijothomas:
> Although I can see that this works, it feels like a big leak of impl details into the user's domain. What would it look like as a helper in OTel itself? E.g. something like `withTelemetrySuppression(_ => { /* setup otel here */ } )` ?
> 
> I expect this would still require the user to not use a `tokio_main` but rather explicitly create their runtime after setting up OTel using this helper to wrap, but this would still go a ways to make it feel a bit less leaky.
> 
> My other question is - do we impact any of our users by requiring a separate tokio runtime in this case, for instance, folks in resource-constrained environments?

Good point. This is already the case even without this PR! See https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/lib.rs#L112-L113
OTLP/gRPC Exporter already *requires* a tokio runtime - either it captures current one if user has tokio::main, or we explicitly ask users to create an RT and do OTLP instantiation inside it.

Exposing a helper/feature in OTLP Exporter bloats public API, and it'll be less flexible than users giving a runtime to us. (they can do other things inside thread_start/stop apart from just the suppression etc.)

At some point in the future, we could work with tokio-tracing maintainers and see if we can agree on a mutual `Context` field for suppression, but this requires a lot of research and co-ordination. The approach shown in this PR is just a way for users to unblock themselves right now, without OTel/OTLP doing anything extra.

opentelemetry-otlp/examples/basic-otlp/src/main.rs
cijothomas, line 67: can we ask tokio::main macro's to provide on_start/on_end callback..similar to the way they offer on_panic..?

cijothomas, line 67: todo: cleanup http one and confirm it does not need this technique by default.
todo: see if the client authors offer a way to opt-out of telemetry.

opentelemetry-otlp/examples/basic-otlp/src/main.rs
```diff
@@ -59,33 +60,37 @@ fn init_logs() -> SdkLoggerProvider {
...
+fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
+    let rt = tokio::runtime::Builder::new_multi_thread()
```
cijothomas, line 69: check if this can be wrapped inside std::thread

bryantbiggs:
While reviewing the telemetry suppression approach here, I noticed that `SimpleConcurrentLogProcessor::emit()` in `opentelemetry-sdk/src/logs/concurrent_log_processor.rs` does not enter a telemetry-suppressed scope before calling the exporter, unlike every other processor in the SDK:

- `SimpleLogProcessor::emit()` has `let _suppress_guard = Context::enter_telemetry_suppressed_scope();`
- `BatchLogProcessor` suppresses in both its export and batch-processing paths
- `SimpleConcurrentLogProcessor::emit()` calls `futures_executor::block_on(self.exporter.export(...))` with no suppression at all

If an exporter (or its underlying transport) emits logs during export, and those logs flow back through a `SimpleConcurrentLogProcessor`, there is nothing to break the recursion.

This is likely a straightforward fix -- adding `let _suppress_guard = Context::enter_telemetry_suppressed_scope();` at the top of `SimpleConcurrentLogProcessor::emit()`, consistent with how `SimpleLogProcessor` does it. Might be worth folding into this PR or a follow-up.

cijothomas:
> While reviewing the telemetry suppression approach here, I noticed that `SimpleConcurrentLogProcessor::emit()` in `opentelemetry-sdk/src/logs/concurrent_log_processor.rs` does not enter a telemetry-suppressed scope before calling the exporter, unlike every other processor in the SDK:
> 
> * `SimpleLogProcessor::emit()` has `let _suppress_guard = Context::enter_telemetry_suppressed_scope();`
> * `BatchLogProcessor` suppresses in both its export and batch-processing paths
> * `SimpleConcurrentLogProcessor::emit()` calls `futures_executor::block_on(self.exporter.export(...))` with no suppression at all
> 
> If an exporter (or its underlying transport) emits logs during export, and those logs flow back through a `SimpleConcurrentLogProcessor`, there is nothing to break the recursion.
> 
> This is likely a straightforward fix -- adding `let _suppress_guard = Context::enter_telemetry_suppressed_scope();` at the top of `SimpleConcurrentLogProcessor::emit()`, consistent with how `SimpleLogProcessor` does it. Might be worth folding into this PR or a follow-up.

SimpleConcurrentLogProcessor intentionally does not suppress telemetry. It is designed to be used exclusively with OS-native exporters (e.g., ETW on Windows, user_events on Linux) that are known to be reentrant-safe. These exporters don't use network transports like tonic/hyper, so there's no risk of telemetry-induced-telemetry loops. Adding suppression there would be unnecessary overhead for its intended use case. 

Having said that, we can probably remove `SimpleConcurrentLogProcessor`, given there is no spec support.

Issue #2972, opened 2025-05-15
[Bug]: Impossible to disable `internal-logs` if you import opentelemetry_sdk


joeshaw:
### What happened?

A service was spamming local logs with the message:

```
2025-05-15T18:21:42.793642Z  INFO opentelemetry:  name="MeterProvider.GlobalSet" Global meter provider is set. Meters can now be created using global::meter() or global::meter_with_scope().
```

I should be able to suppress this by disabling the `internal-logs` feature, which is on by default.  In my `Cargo.toml` file, I have under `workspace.dependencies`:

```
opentelemetry = { version = "0.27.1", default-features = false, features = ["metrics", "trace"] }
opentelemetry-otlp = { version = "0.27.0", default-features = false, features = ["metrics", "trace", "http-json"] }
opentelemetry-proto = { version = "0.27.0", default-features = false, features = ["metrics", "trace", "gen-tonic-messages", "with-serde"] }
opentelemetry_sdk = { version = "0.27.1", default-features = false, features = ["metrics", "trace"] }
```

So I would expect this to not be on.  However, running `cargo tree -e features --no-default-features` shows this, which I've trimmed and pruned for readability:

```
│       │       ├── opentelemetry-otlp feature "http-json"
│       │       │   ├── opentelemetry-otlp v0.27.0
│       │       │   │   ├── opentelemetry-proto v0.27.0
│       │       │   │   │   ├── opentelemetry_sdk v0.27.1
│       │       │   │   │   │   ├── opentelemetry feature "default"
│       │       │   │   │   │   │   ├── opentelemetry v0.27.1 (*)
│       │       │   │   │   │   │   ├── opentelemetry feature "internal-logs"
```

Looking in the `Cargo.toml` file for `opentelemetry_sdk`, we have:

```
[dependencies]
opentelemetry = { version = "0.29", path = "../opentelemetry/" }
opentelemetry-http = { version = "0.29", path = "../opentelemetry-http", optional = true }
```

I am not a Cargo expert, but my interpretation of this is that any import of `opentelemetry_sdk` will require the defaults of `opentelemetry` and thus `internal-logs`.  (And transitively also for `opentelemetry-http`, if it is included.)

It seems to me that _every_ dependency of an opentelemetry crate should be `default-features = false` in all `Cargo.toml` files across the repository, if you want it to be possible to turn off internal logging.

### OpenTelemetry API Version (i.e version of `opentelemetry` crate)

[...]

PR #3085, opened 2025-07-26
chore: Consolidate internal crate dependencies to use workspace dependencies

cijothomas:
Consolidates all internal OpenTelemetry crate dependencies to use `workspace = true` instead of explicit path dependencies.
Added missing crates to workspace, and set default-feature=false to all, letting individual crates pick the features they want to.

### DIFF
--- a/opentelemetry-appender-tracing/Cargo.toml
+++ b/opentelemetry-appender-tracing/Cargo.toml
@@ -23,7 +23,7 @@ tracing-opentelemetry = { workspace = true, optional = true }
 [dev-dependencies]
 log = { workspace = true }
 opentelemetry-stdout = { workspace = true, features = ["logs"] }
-opentelemetry_sdk = { workspace = true, features = ["logs", "testing"] }
+opentelemetry_sdk = { workspace = true, features = ["logs", "testing", "internal-logs"] }
 tracing = { workspace = true, features = ["std"]}
 tracing-subscriber = { workspace = true, features = ["env-filter","registry", "std", "fmt"] }
 tracing-log = { workspace = true }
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/Cargo.toml
```diff
@@ -42,7 +42,7 @@ pprof = { version = "0.14", features = ["flamegraph", "criterion"] }
 default = ["trace", "metrics", "logs", "internal-logs"]
 trace = ["opentelemetry/trace", "rand", "percent-encoding"]
 jaeger_remote_sampler = ["trace", "opentelemetry-http", "http", "serde", "serde_json", "url", "experimental_async_runtime"]
-logs = ["opentelemetry/logs", "serde_json"]
+logs = ["opentelemetry/logs", "opentelemetry/trace", "serde_json"]
```
scottgerring, line 45: this seems odd - does logs need trace?

cijothomas: yes. The current implementation requires it. It should be treated as a bug and fixed separately. Log just need couple of things from trace (traceid, ability to retrieve current span from context, but currently it brings in entire `traces` feature.

opentelemetry-stdout/Cargo.toml
```diff
@@ -28,12 +28,12 @@ logs = ["opentelemetry/logs", "opentelemetry_sdk/logs", "opentelemetry_sdk/spec_
...
-opentelemetry = { version = "0.30", path = "../opentelemetry" }
-opentelemetry_sdk = { version = "0.30", path = "../opentelemetry-sdk" }
+opentelemetry = { workspace = true }
+opentelemetry_sdk = { workspace = true }
 
 [dev-dependencies]
-opentelemetry = { path = "../opentelemetry", features = ["metrics"] }
-opentelemetry_sdk = { path = "../opentelemetry-sdk", features = ["metrics"] }
+opentelemetry = { workspace = true, features = ["metrics"] }
+opentelemetry_sdk = { workspace = true, features = ["metrics"] }
```
lalitb: These two packages are already available as regular dependencies, re-declaring them in dev-dependencies seems unnecessary. Though not coming from this PR.

cijothomas:
> Update the CHANGELOG as the breaking change - specifically since internal-logs is no longer enabled feature now?

There is no breaking change. internal-logs is still enabled by default. The change is.- it can be disabled now, addressing #2972 
Can you recheck and see if this makes sense?

lalitb (approved):

opentelemetry-sdk/Cargo.toml
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/3096 for tracking

PR #3088, opened 2025-07-28
feat: add support for `OTEL_SDK_DISABLED` environment variable

### LINKED ISSUE #1936: [Feature]: add support for environment variable OTEL_SDK_DISABLED
### Related Problems?

_No response_

According to the [general SDK configuration](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration), `OTEL_SDK_DISABLED` should make SDK a no-op implementation for all telemetry signals.

### Considered Alternatives

_No response_

### Additional Context

_No response_

iamsauravsharma:
Fixes #1936

## Changes

Add support for `OTEL_SDK_DISABLED` environment variable using no-op SDK implementation as explained [Opentelemetry SDK environment variable docs](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration). When set to `true`, the SDK will provide no-op implementations for all API components (Tracer, Logger, and Meter).

Implemented no-op provider classes that mimic the shutdown no-op providers for logger and tracer with only difference is shutdown variable which is set to false for disabled no op. Where as for metrics existing noop is re used


iamsauravsharma:
Since overwriting `OTEL_SDK_DISABLED` environment variable would effects other test when running parallel test. What should be done? Ignore test and run test at last stage of `scripts/test.sh` using command `cargo test --package opentelemetry_sdk --lib --all-features -- otel_sdk_disabled_env --ignored`?

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -12,19 +12,33 @@ use std::{
...
         inner: Arc::new(LoggerProviderInner {
             processors: Vec::new(),
             is_shutdown: AtomicBool::new(true),
         }),
     })
 }
+// a no op logger provider used as placeholder when sdk is disabled with
+// help of environment variable `OTEL_SDK_DISABLED`
+// TODO - replace it with LazyLock once it is stable
+static DISABLED_LOGGER_PROVIDER: OnceLock<SdkLoggerProvider> = OnceLock::new();
```
lalitb: Can we reuse the existing NOOP_LOGGER_PROVIDER for both shutdown and disable ?

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -286,7 +309,10 @@ impl opentelemetry::trace::TracerProvider for SdkTracerProvider {
 
     fn tracer_with_scope(&self, scope: InstrumentationScope) -> Self::Tracer {
         if self.inner.is_shutdown.load(Ordering::Relaxed) {
-            return SdkTracer::new(scope, noop_tracer_provider().clone());
+            return SdkTracer::new(scope, shutdown_tracer_provider().clone());
+        }
+        if std::env::var("OTEL_SDK_DISABLED").is_ok_and(|var| var.to_lowercase() == "true") {
```
cijothomas: what is the intended use case of "OTEL_SDK_DISABLED" env variable? Are we expected to re-read the env variable each time a tracer/logger/meter is requested? Or we read the env variable once at startup and not again?

It looks like the current PR is re-reading it each time. This can cause inconsistencies- tracers from before will continue to work vs new ones won't.. That is not the intended behavior in my opinion.

My suggestion:
Read the env variable **once**, at provider construction time, and store the value inside.
Any time tracer/logger/meter is requested, return NoOp if disabled. The log message should indicate NoOp is being returned due to the env-variable setting.

Happy to discuss other ideas.

cijothomas (changes requested):
Requesting changes to discuss and address comment: https://github.com/open-telemetry/opentelemetry-rust/pull/3088/files#r2237333415

opentelemetry-sdk/src/logs/logger_provider.rs
iamsauravsharma: setting `is_shutdown` as `true` also for disabled one have no effect currently but to me `true` value looks wrong since logger provider is not in shutdown state but disabled by user so i have created new logger provider

opentelemetry-sdk/src/trace/provider.rs
iamsauravsharma: implemented similar to python implementation https://github.com/open-telemetry/opentelemetry-python/pull/3648 after reading discussion at https://github.com/open-telemetry/opentelemetry-specification/issues/4332

iamsauravsharma:
Implemented some changes as requested

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -196,6 +197,10 @@ impl MeterProvider for SdkMeterProvider {
             );
             return Meter::new(Arc::new(NoopMeter::new()));
         }
+        if self.inner.is_disabled {
+            otel_warn!(name: "MeterProvider.NoOpMeterReturned", message = "Returned NoOpMeter. SDK is disabled");
```
cijothomas:
```suggestion
            otel_debug!(name: "MeterProvider.NoOpMeterReturned", message = "Returned NoOpMeter. SDK is disabled");
```

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -53,13 +69,18 @@ impl opentelemetry::logs::LoggerProvider for SdkLoggerProvider {
...
             otel_debug!(
                 name: "LoggerProvider.NoOpLoggerReturned",
                 logger_name = scope.name(),
             );
-            return SdkLogger::new(scope, noop_logger_provider().clone());
+            return SdkLogger::new(scope, shutdown_logger_provider().clone());
+        }
+        // If the provider is disabled, new logger will refer a disabled no-op logger provider.
+        if self.inner.is_disabled {
+            otel_warn!(name: "LoggerProvider.NoOpLoggerReturned", message = "Returned NoOpLogger. SDK is disabled");
```
cijothomas: lets use consistent logging as shutdown case.

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -53,13 +69,18 @@ impl opentelemetry::logs::LoggerProvider for SdkLoggerProvider {
     }
 
     fn logger_with_scope(&self, scope: InstrumentationScope) -> Self::Logger {
-        // If the provider is shutdown, new logger will refer a no-op logger provider.
+        // If the provider is shutdown, new logger will refer a shutdown no-op logger provider.
         if self.inner.is_shutdown.load(Ordering::Relaxed) {
             otel_debug!(
-                name: "LoggerProvider.NoOpLoggerReturned",
+                name: "LoggerProvider.ShutdownNoOpLoggerReturned",
```
iamsauravsharma: I have renamed old `NoOpLoggerReturned` to `ShutdownNoOpLoggerReturned` should i change it back to original name. Renamed so we can distinguish between shutdown and disabled case

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -267,10 +289,18 @@ impl LoggerProviderBuilder {
             processor.set_resource(&resource);
         }
 
+        let is_disabled =
+            std::env::var("OTEL_SDK_DISABLED").is_ok_and(|var| var.to_lowercase() == "true");
+
+        if is_disabled {
+            otel_warn!(name: "LoggerProvider.Disabled", message = "SDK is disabled through environment variable");
```
iamsauravsharma: Using warn! here since reference python implementation also used warn

iamsauravsharma (commented):
@cijothomas I have implemented changes suggested with some comment below

opentelemetry-sdk/src/logs/logger_provider.rs
cijothomas: we can keep the original name for this and for the env_variable disabled scenario both. Just add another attribute "reason", with value being "already_shutdown" and "disabled_via_env_variable".

iamsauravsharma: I have used original name and added reason attribute

cijothomas:
@iamsauravsharma 

Thanks for continuing to work on this.  Given this feature is supposed to disable entire SDK, I'd like to see a benchmark added which indeed proves that the perf cost when this env var is set is comparable to the API only perf. This will allow us to confirm that we are short-circuiting early enough.

iamsauravsharma:
@cijothomas I have added benchmark and below are result when running in my machine
| Test    | api-only | sdk-disabled | sdk-shutdown | sdk-enabled |
|---------|----------|--------------|--------------|-------------|
| trace   |  20.4 ns |  38.9 ns     |  38.5 ns     |  178 ns     |
| log     |  0.32 ns |  2.09 ns     |  2.09 ns     |  37 ns      |
| metrics |  6.11 ns |  6.11 ns     |  6.10 ns     |  66 ns      |

cijothomas:
> @cijothomas I have added benchmark and below are result when running in my machine
> 
> Test	api-only	sdk-disabled	sdk-shutdown	sdk-enabled
> trace	20.4 ns	38.9 ns	38.5 ns	178 ns
> log	0.32 ns	2.09 ns	2.09 ns	37 ns
> metrics	6.11 ns	6.11 ns	6.10 ns	66 ns

Thanks! Could you include the benchmark as part of this PR ? It definitely looks like there is room for improvement in traces and logs - are we running samplers/generating-trace-id etc when SDK-DISABLED? That might explain why its double the api-only...
Please see if you can optimize it more.

iamsauravsharma:
@cijothomas benchmark code is already included in PR.

Regarding the performance since `sdk-disabled` and `sdk-shutdown` yielded almost identical result. And, `sdk-shutdown` was already implemented I thought performance was acceptable. Anyway I would look if I can optimize further.

iamsauravsharma:
@cijothomas  Performance of trace and logs were improved further
https://github.com/iamsauravsharma/opentelemetry-rust/blob/4e437db2c682b4bab38107b4f7698d3725b2e306/opentelemetry-sdk/benches/sdk_disabled.rs Latest benchmark result and code performing benchmark
| Test    | api-only | sdk-disabled | sdk-shutdown | sdk-enabled |
|---------|----------|--------------|--------------|-------------|
| trace   |  20.3 ns |  31.9 ns     |  31.4 ns     |  185 ns     |
| log     |  0.32 ns |  1.28 ns     |  1.27 ns     |   38 ns     |
| metrics |  6.11 ns |  6.11 ns     |  6.10 ns     |   66 ns     |

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -295,10 +322,18 @@ impl LoggerProviderBuilder {
             processor.set_resource(&resource);
         }
 
+        let is_disabled =
+            std::env::var("OTEL_SDK_DISABLED").is_ok_and(|var| var.to_lowercase() == "true");
```
scottgerring: The Boolean parsing and logging behaviour should align with the [SDK environment-variable specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#boolean).

The spec says only case-insensitive `"true"` enables a Boolean setting. False, empty, and unset values should silently resolve to false, while other invalid values SHOULD produce a warning explaining that false is being used as the fallback.

Currently invalid values silently fall back to false, while valid `"true"` values emit a Warn-level message from each provider. Since disabling the SDK is intentional configuration, that state should be logged at Debug, if at all. The warning should instead be reserved for invalid values and explain the impact and corrective action.

Could we centralise this parsing so traces, metrics, and logs consistently:

- accept case-insensitive `"true"`;
- silently accept `"false"`, empty, or unset;
- warn and fall back to false for any other value; and
- avoid Warn-level logging for a valid disabled configuration?

scripts/test.sh
```diff
@@ -65,3 +65,7 @@ EOF
 echo "Running ignored tests for opentelemetry-appender-tracing package (global logger tests)"
 cargo test --manifest-path=opentelemetry-appender-tracing/Cargo.toml --all-features layer::tests::tracing_appender_standalone_with_tracing_log -- --ignored --exact
 cargo test --manifest-path=opentelemetry-appender-tracing/Cargo.toml --all-features layer::tests::tracing_appender_inside_tracing_context_with_tracing_log -- --ignored --exact
+
+# Run test which set environment variable `OTEL_SDK_DISABLED`
+echo "Running ignored test for opentelemetry-sdk package with OTEL_SDK_DISABLED environment variable set"
+cargo test --package opentelemetry_sdk --lib --all-features -- otel_sdk_disabled_env --ignored
```
scottgerring: This still runs the three `otel_sdk_disabled_env` tests concurrently. Since all three use `temp_env` to modify the same process-wide environment variable, one test can restore `OTEL_SDK_DISABLED` while another test is still running.

Could we serialise this invocation with `--test-threads=1`, or combine the three signal checks into one isolated test? For example:

```suggestion
cargo test --package opentelemetry_sdk --lib --all-features -- otel_sdk_disabled_env --ignored --test-threads=1
```

opentelemetry-sdk/src/logs/logger_provider.rs
scottgerring: I agree with Lalit: this is an internal no-op placeholder, not the original user-visible provider, so `is_shutdown = true` does not incorrectly describe the original provider’s state. I think we could reuse the existing no-op provider and avoid the extra `OnceLock` and duplicated state, with the disabled reason remaining at the point where the logger is returned?

opentelemetry-sdk/benches/sdk_disabled.rs
```diff
@@ -0,0 +1,239 @@
...
+    let mut span = tracer.start("span-name");
+    if span.is_recording() {
+        span.set_attribute(KeyValue::new("key1", "value1"));
+        span.set_attribute(KeyValue::new("key2", 123));
+    }
+    span.end();
+}
+
+fn emit_log<L: Logger>(logger: &L) {
+    if logger.event_enabled(Severity::Info, "benchmark", None) {
```
scottgerring, line 101: This only benchmarks the `event_enabled()` fast path, so `logger.emit()` is never called when the SDK is disabled. Could we also benchmark direct emission and make the disabled `emit()` path return before reading context/time or updating self-observability metrics?

opentelemetry-sdk/src/logs/logger_provider.rs
iamsauravsharma: Done — new env.rs with a shared `check_bool_env_var`, used by all three providers

scripts/test.sh
iamsauravsharma: Done, added --test-threads=1.

opentelemetry-sdk/benches/sdk_disabled.rs
iamsauravsharma: Added a log-emit group that calls emit() unconditionally

scripts/test.sh
```diff
@@ -65,3 +65,7 @@ EOF
 echo "Running ignored tests for opentelemetry-appender-tracing package (global logger tests)"
 cargo test --manifest-path=opentelemetry-appender-tracing/Cargo.toml --all-features layer::tests::tracing_appender_standalone_with_tracing_log -- --ignored --exact
 cargo test --manifest-path=opentelemetry-appender-tracing/Cargo.toml --all-features layer::tests::tracing_appender_inside_tracing_context_with_tracing_log -- --ignored --exact
+
+# Run test which set environment variable `OTEL_SDK_DISABLED`
```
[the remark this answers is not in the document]
iamsauravsharma: Fixed

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -165,23 +165,18 @@ impl opentelemetry::trace::Tracer for SdkTracer {
...
-                None,
-                self.clone(),
-                SpanLimits::default(),
-            );
+            return Span::new_noop(SpanContext::empty_context());
+        }
+
+        // no point starting a span if the SDK is disabled via OTEL_SDK_DISABLED.
+        if provider.is_disabled() {
+            return Span::new_noop(SpanContext::empty_context());
```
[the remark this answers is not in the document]
iamsauravsharma: Fixed, for the shutdown path too. The spec is explicit that a no-op API "MUST return a non-recording Span with the SpanContext in the parent Context. Rust was the only SDK dropping the context

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -401,6 +411,13 @@ impl MeterProviderBuilder {
             builder = format!("{:?}", &self),
         );
 
+        let is_disabled =
+            std::env::var("OTEL_SDK_DISABLED").is_ok_and(|var| var.to_lowercase() == "true");
+
+        if is_disabled {
+            otel_warn!(name: "MeterProvider.Disabled", message = "SDK is disabled through environment variable");
```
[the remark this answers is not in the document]
iamsauravsharma: Fixed no background thread

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -494,11 +539,19 @@ impl TracerProviderBuilder {
             p.set_resource(config.resource.as_ref());
         }
 
+        let is_disabled =
+            std::env::var("OTEL_SDK_DISABLED").is_ok_and(|var| var.to_lowercase() == "true");
+
+        if is_disabled {
+            otel_warn!(name: "TracerProvider.Disabled", message = "SDK is disabled through environment variable");
```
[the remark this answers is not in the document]
iamsauravsharma: Fixed no background thread

iamsauravsharma:
Pushed an update addressing latest review comment

- Boolean parsing for OTEL_SDK_DISABLED centralised in a new env.rs helper
- Disabled/shutdown tracers now carry the parent span context, as the trace API spec requires of a no-op API.
- no batch worker or periodic-reader thread is started.
- Added a log-emit benchmark group that calls emit() directly.

opentelemetry-sdk/src/env.rs
```diff
@@ -0,0 +1,63 @@
...
+/// Checks the value of a boolean environment variable.
+///
+/// Returns `true` if the environment variable is set to a case-insensitive `"true"`,
+/// returns `false` if it is unset, empty, or set to a case-insensitive `"false"`,
+/// and logs a warning and returns `false` for any other value.
+fn check_bool_env_var(var_name: &str) -> bool {
+    let Ok(value) = std::env::var(var_name) else {
+        return false;
+    };
+    let value = value.trim();
```
scottgerring: Trimming here means `OTEL_SDK_DISABLED=" true "` disables the SDK. The [Boolean environment-variable specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#boolean) says only the case-insensitive string `"true"` enables a Boolean setting, and implementations MUST NOT accept additional representations.

Could we remove the trim and update the test so whitespace-padded values are treated as invalid, warned about, and resolved to false? The current test explicitly expects `" true "` to enable the SDK.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -305,6 +327,9 @@ impl MeterProviderBuilder {
     where
         T: PushMetricExporter,
     {
+        if self.is_disabled {
+            return self;
+        }
```
scottgerring, line 335: This fixes the convenience paths such as `with_periodic_exporter()`, but the generic registration paths, including `with_reader()`, `with_log_processor()`, and `with_span_processor()`, still retain and invoke components while the SDK is disabled. The same issue applies across all three signals.

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -821,6 +875,32 @@ mod tests {
...
+            assert_eq!(logger_provider.log_processors().len(), 0);
+
+            let logger = logger_provider.logger("disabled");
+            assert!(!logger.event_enabled(Severity::Info, "test", None));
+            for _ in 0..3 {
+                let mut record = logger.create_log_record();
+                record.set_body("Testing sdk disabled logger".into());
+                logger.emit(record);
+            }
+            assert_eq!(exporter.get_emitted_logs().unwrap().len(), 0);
```
scottgerring, line 885: Could we add a regression assertion for the direct-emission side effect here? Checking that the exporter remains empty would also pass the previous impl because the disabled logger had no processors. But a direct `emit()` still updated `otel.sdk.log.created` and read context/time before discovering there were no processors. A self-observability test asserting that `otel.sdk.log.created` remains zero would cover this

scottgerring (changes requested):
couple other small changes but looks close!

opentelemetry-sdk/src/env.rs
iamsauravsharma: done

opentelemetry-sdk/src/metrics/meter_provider.rs
iamsauravsharma: done

opentelemetry-sdk/src/logs/logger_provider.rs
iamsauravsharma: added `log_created_not_counted_when_sdk_disabled` for testing

opentelemetry-sdk/src/logs/logger_provider.rs
scottgerring: @iamsauravsharma I believe this one is still unresolved

scottgerring (changes requested):
Commented inline; there's an open comment from Lalit that's still unresolved. Otherwise the rest of my feedback looks addressed!

opentelemetry-sdk/src/logs/logger_provider.rs
iamsauravsharma: @scottgerring done reused `NOOP_LOGGER_PROVIDER`

PR #3089, opened 2025-07-28
fix: batch size accounting in BatchSpanProcessor when queue is full

dojiong:
Fixes #

- When the queue is full, spans are dropped but `current_batch_size` is  __incremented__
- However, when spans are later exported, `current_batch_size` is __still decremented__ based on the actual number of spans exported

## Changes

moving the current_batch_size.fetch_add to only execute when try_send succeeds


lalitb (approved):
Good catch. Thanks for fixing. Just verified the approach in log processor - which looks fine.

please update the Changelog.

PR #3090, opened 2025-07-28
feat: add support for OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT

msierks-pcln:
### **User description**
Fixes https://github.com/open-telemetry/opentelemetry-rust/discussions/2214

## Changes

Add support for the `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` environment variable as explained in [Opentelemetry SDK environment variable doc](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#attribute-limits). This will truncate `Value::String` and `Array::String` attribute values if they exceed the limit.

- Adds `TracerProviderBuilder::with_max_attribute_value_length()` 

### DIFF
--- a/opentelemetry-sdk/src/trace/config.rs
+++ b/opentelemetry-sdk/src/trace/config.rs
@@ -43,6 +43,13 @@ impl Default for Config {
             config.span_limits.max_attributes_per_span = max_attributes_per_span;
         }
 
+        if let Some(max_attribute_value_length) = env::var("OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT")
+            .ok()
+            .and_then(|count_limit| i32::from_str(&count_limit).ok())
+        {
+            config.span_limits.max_attribute_value_length = max_attribute_value_length;
+        }
+
         if let Some(max_events_per_span) = env::var("OTEL_SPAN_EVENT_COUNT_LIMIT")
             .ok()
             .and_then(|max_events| u32::from_str(&max_events).ok())
--- a/opentelemetry-sdk/src/trace/provider.rs
+++ b/opentelemetry-sdk/src/trace/provider.rs
@@ -378,6 +378,15 @@ impl TracerProviderBuilder {
         self
     }
 
+    /// Specify the maximum allowed length of attribute values.
+    ///
+    /// This limit applies to [Value::String](opentelemetry::Value::String) and [Array::String](opentelemetry::Array::String)
+    /// attributes only. When the limit is exceeded, the value is truncated without any indication that truncation has occurred.
+    pub fn with_max_attribute_value_length(mut self, max_attribute_value_length: u32) -> Self {
+        self.config.span_limits.max_attribute_value_length = max_attribute_value_length as i32;
+        self
+    }
+
     /// Specify the number of events to be recorded per span.
     pub fn with_max_links_per_span(mut self, max_links: u32) -> Self {
         self.config.span_limits.max_links_per_span = max_links;
--- a/opentelemetry-sdk/src/trace/span.rs
+++ b/opentelemetry-sdk/src/trace/span.rs
@@ -104,7 +104,6 @@ impl opentelemetry::trace::Span for Span {
                 let dropped_attributes_count =
                     attributes.len().saturating_sub(event_attributes_limit);
                 attributes.truncate(event_attributes_limit);
-
                 data.events.add_event(Event::new(
                     name,
                     timestamp,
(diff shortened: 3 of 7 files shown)


opentelemetry/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - *Breaking* Change return type of `opentelemetry::global::set_tracer_provider` to Unit to align with metrics counterpart
 - Add `get_all` method to `opentelemetry::propagation::Extractor` to return all values of the given propagation key and provide a default implementation.
+- Add support for max span attribute value length; Applies to `Value::String` and `Array::String` only.
```
cijothomas: Attribute length enforcements are not part of `opentelemetry` crate. Its part of `opentelemetry_sdk` crate, so changelog should be in that crate.

opentelemetry-sdk/src/trace/config.rs
```diff
@@ -43,6 +43,19 @@ impl Default for Config {
             config.span_limits.max_attributes_per_span = max_attributes_per_span;
         }
 
+        if let Some(max_attribute_value_length) = env::var("OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT")
+            .ok()
+            .and_then(|count_limit| i32::from_str(&count_limit).ok())
+        {
+            config.span_limits.max_attribute_value_length = max_attribute_value_length;
+        } else if let Some(max_attribute_value_length) =
+            env::var("OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT")
```
cijothomas: supporting OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT looks reasonable, but OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT is not something we can add right now, as Logs don't respect this at all.

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -378,6 +378,12 @@ impl TracerProviderBuilder {
         self
     }
 
+    /// Specify the maximum allowed length of attribute values.
```
cijothomas, line 381: could you enhance the doc to make it clear that this is applicable to string values or array of strings.
and also mention that the trimming behavior is simply trimming!
(some people might expect "..." being added as the last 3 characters to indicate trimming has occurred.. But we are not doing that, and there is no way a backend can tell if trimming has occurred. So if a user opts in to this, they should be made aware of the implications)

opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -378,6 +378,12 @@ impl TracerProviderBuilder {
...
+    pub fn with_max_attribute_value_length(mut self, max_attribute_value_length: i32) -> Self {
```
cijothomas: we expect positive values only right? i32 is probably the wrong choice.

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -566,6 +572,58 @@ mod tests {
...
+                ),
+            ),
+        ];
+        let span_builder = SpanBuilder::from_name("test_span").with_attributes(initial_attributes);
+
+        let mut span = tracer.build(span_builder);
+        span.set_attribute(KeyValue::new("third", "test data, after span builder"));
+
+        span.with_data(|data| {
+            assert_eq!(
```
cijothomas: instead of picking attributes at each index(0,1,..), could we do a general validation - for each string or array of string, validate that length is <max_attribute_value_length?

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -73,6 +73,14 @@ impl SdkTracer {
             .saturating_sub(span_attributes_limit);
         attribute_options.truncate(span_attributes_limit);
         let dropped_attributes_count = dropped_attributes_count as u32;
+        let span_attribute_value_limit = span_limits.max_attribute_value_length;
```
cijothomas, line 76: `set_attribute` already does the truncation...Is it needed to be done again on this method?

cijothomas (changes requested):
Looks reasonable. left some small comments.

One is blocking - https://github.com/open-telemetry/opentelemetry-rust/pull/3090/files#r2237524061

opentelemetry-sdk/src/trace/tracer.rs
msierks-pcln: From what I can tell, this method is used by the `SpanBuilder` when a span is initially built. The `set_attribute` method is used for setting single attributes after the span has been created.

cijothomas:
@msierks-pcln Can you sign the CLA, so we can consider contributions? It's a one-time process and is mandatory. If contributing on behalf of your employer, please do consult them before signing.

bryantbiggs:
@msierks-pcln is this something you still plan on pursuing and signing the CLA?

Issue #3083, opened 2025-07-24
OTLP Stabilization: HTTP exporter should support `gzip` and `zstd`


scottgerring:
Per the spec.

PR #3092, opened 2025-07-31
Feat: add zstd and gzip compression support to HTTP OTLP exporter

scottgerring:
Fixes #
#3083 

## Changes

Straightforward enough! None of our clients support this natively, so we have to do the compressing ourselves. Have copied the pattern used by the gRPC exporter in terms of features and API, and moved a bit of code around so we can share it. 


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -261,26 +277,59 @@ pub(crate) struct OtlpHttpClient {
...
+                use std::io::Write;
+
+                let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+                encoder.write_all(&body).map_err(|e| e.to_string())?;
+                let compressed = encoder.finish().map_err(|e| e.to_string())?;
+                Ok((compressed, Some("gzip")))
+            }
+            #[cfg(not(feature = "gzip-http"))]
+            Some(crate::Compression::Gzip) => {
+                Err("gzip compression requested but gzip-http feature not enabled".to_string())
```
cijothomas, line 328: could this be detected and fail-fast at exporter build time itself?

opentelemetry-otlp/src/lib.rs
```diff
@@ -281,7 +281,9 @@
 //! use opentelemetry_sdk::{trace::{self, RandomIdGenerator, Sampler}, Resource};
 //! # #[cfg(feature = "metrics")]
 //! use opentelemetry_sdk::metrics::Temporality;
-//! use opentelemetry_otlp::{Protocol, WithExportConfig, WithTonicConfig};
+//! use opentelemetry_otlp::{Protocol, WithExportConfig};
+//! # #[cfg(feature = "grpc-tonic")]
+//! use opentelemetry_otlp::WithTonicConfig;
```
cijothomas, line 288: good catch! Wondering how our CI didn't catch this before!

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -14,14 +14,20 @@ impl LogExporter for OtlpHttpClient {
...
 
-        let mut request = http::Request::builder()
+        let mut request_builder = http::Request::builder()
             .method(Method::POST)
             .uri(&self.collector_endpoint)
-            .header(CONTENT_TYPE, content_type)
+            .header(CONTENT_TYPE, content_type);
+
+        if let Some(encoding) = content_encoding {
+            request_builder = request_builder.header("Content-Encoding", encoding);
```
cijothomas: nit: Move Content-Encoding to const string, or is it available from http crate already?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -50,6 +49,9 @@ pub struct HttpConfig {
 
     /// Additional headers to send to the collector.
     headers: Option<HashMap<String, String>>,
+
+    /// The compression algorithm to use when communicating with the collector.
```
cijothomas:
```suggestion
    /// The compression algorithm to use when communicating with the OTLP endpoint.
```

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -289,59 +338,75 @@ impl OtlpHttpClient {
...
+                Ok(json) => (json.into_bytes(), "application/json"),
+                Err(e) => return Err(e.to_string()),
             },
-            _ => Ok((req.encode_to_vec(), "application/x-protobuf")),
-        }
+            _ => (req.encode_to_vec(), "application/x-protobuf"),
+        };
+
+        let (compressed_body, content_encoding) = self.compress_body(body)?;
+        Ok((compressed_body, content_type, content_encoding))
```
cijothomas: nit: The variable gives the wrong impression that compressed_body is always compressed. But if Compression is None, then its just returning existing body as-is.
May be:
if (self.compression is None)
{
 Ok(body,content_type,None)
}
else
{
call self.compress_body(...),
}

opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
```diff
@@ -15,6 +15,8 @@ bench = false
 [features]
 default = ["reqwest-blocking"]
 reqwest-blocking = ["opentelemetry-otlp/reqwest-blocking-client"]
+gzip = ["opentelemetry-otlp/gzip-http"]
```
cijothomas: examples are currently kept basic.-showing absolute minimum required to work. I suggest to keep it that way in this PR. If we make a separate decision about improving these examples to show more scenarios, we can do it for this+ the gRPC examples too, in a separate PR.

cijothomas (changes requested):
Looks good! Requesting changes just for the below:
1. See if we can detect invalid/incorrect compression setting at build_exporter time itself and not in actual export. Eg; user enabled compression via env_variable, but the feature-flag is not enabled.
2. Add a changelog as this is a new capability
3. Lets keep example as-is in this PR.

opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/lib.rs#L269-L271 The kitchensink would be a good place to showcase this.

opentelemetry-otlp/src/lib.rs
scottgerring: I'm not sure TBH. The combination of feature flags and lint scripts in the build is a lot to hold in your head!

opentelemetry-otlp/src/exporter/http/mod.rs
scottgerring: Absolutely 🤦  We even have an existing error for it

opentelemetry-otlp/src/exporter/http/mod.rs
scottgerring: There's a lot of 'unnecessary concretisation' in this fashion scattered around; i've just fixed this and the obvious adjacent ones

opentelemetry-otlp/src/exporter/http/mod.rs
scottgerring: I've renamed it to "process_body"; this way we don't have confusion about whether or not the caller needs to also check if the body needs compressing, we just always pass the body through the call!

opentelemetry-otlp/examples/basic-otlp-http/Cargo.toml
scottgerring: Done!

scottgerring:
Comments addressed!

PR #3093, opened 2025-07-31
chore: update tonic and prost to 0.14

acrrd:
## Changes
update tonic to 0.14.1 (we need the .1 to avoid msrv bump)
prost to 0.14.
tonic split out prost code to tonic-prost and tonic-prost-build.

~~it bump msrv of -otlp and -proto crates to 1.86.0 because of the same bump in tonic.~~

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Make repo safe for Git inside container
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 10 # Fetch a bit of history so we can do perf diffs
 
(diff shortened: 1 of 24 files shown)


acrrd:
atm this is incompatible with the msrv policy that requires to support stable - 3 releases. fortunately a new release will be available on the 7th of August.

cijothomas:
> it bump msrv of -otlp and -proto crates to 1.86.0 because of the same bump in tonic.

https://github.com/hyperium/tonic?tab=readme-ov-file#rust-version states 1.75.

cijothomas:
> > > it bump msrv of -otlp and -proto crates to 1.86.0 because of the same bump in tonic.
> > 
> > 
> > https://github.com/hyperium/tonic?tab=readme-ov-file#rust-version states 1.75.
> 
> In Cargo.toml 1.86 is specified and that's why `cargo machete` fails https://github.com/hyperium/tonic/blob/edd281b6b562101fa2a04742879713006d7bb360/Cargo.toml#L32

Got it. It looks like that was not intentional, as I see an issue opened for it - https://github.com/hyperium/tonic/issues/2361

cijothomas:
> Some bump will be necessary anyway. Do you prefer to keep it confined to just the crates that are involved like now or we bump all of them?

Let's wait to see how much we have to bump MSRV. Generally we prefer to bump OTLP only, as that's the only one needing tonic etc., so no need to force bump all crates.

acrrd:
@cijothomas no bump needed!

emilk:
I would love to have a release with this 🙏

Issue #3021, opened 2025-06-04
[Bug]: OTEL logs raw Authentication token to std-out on auth error


pulasthibandara:
### What happened?

OTEL logs the raw bearer token to stdout when an error happens on a processor during authentication. The log processor returns an Authentication error due to a misconfiguration, and in-return OTEL logs the the bearer token to stdout.

### OpenTelemetry API Version (i.e version of `opentelemetry` crate)

"0.29.1"

### OpenTelemetry SDK Version (i.e version of `opentelemetry_sdk` crate)

 "0.29.0"

### What Exporter(s) are you seeing the problem on?

OTLP

### Relevant log output

```shell
ERROR opentelemetry_sdk: name="BatchLogProcessor.ExportError" error="Operation failed: export error: Status { code: Unauthenticated, message: \"scheme or token does not match: Bearer <redacted-token>\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\"} }, source: None }"
```

cijothomas:
Thanks for reporting. I think we need to change the exporters to stop including the token in the err string instead of fixing sdk's logging.

PR #3108, opened 2025-08-02
fix: Improved logging for OTLP Tonic Exporter

cijothomas:
Improved logging
1. Interceptor failures are logged separately from export failures. Built-in interceptor simply adds Metadata, and has no scope of failure, so if this fails, it indicates user has added custom interceptors which is failing.
2. Log individual fields in the OTLP exporter, and then return Result contains just the minimal info about failure as OTLP's own log already has sufficient information, and no need to duplicate it by upstream component like batchexportprocessor/readers.

Related to https://github.com/open-telemetry/opentelemetry-rust/issues/3021

### DIFF
--- a/opentelemetry-otlp/src/exporter/tonic/logs.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/logs.rs
@@ -63,7 +63,15 @@ impl LogExporter for TonicLogsClient {
                 let (m, e, _) = inner
                     .interceptor
                     .call(Request::new(()))
-                    .map_err(|e| OTelSdkError::InternalFailure(format!("error: {e:?}")))?
+                    .map_err(|e| {
+                        otel_debug!(
+                            name: "TonicLogsClient.InterceptorFailed",
+                            grpc_code = format!("{:?}", e.code()),
+                            grpc_message = e.message(),
+                            grpc_details = format!("{:?}", e.details())
+                        );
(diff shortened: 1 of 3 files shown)


opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -63,7 +63,15 @@ impl LogExporter for TonicLogsClient {
...
                     .call(Request::new(()))
-                    .map_err(|e| OTelSdkError::InternalFailure(format!("error: {e:?}")))?
+                    .map_err(|e| {
+                        otel_debug!(
+                            name: "TonicLogsClient.InterceptorFailed",
+                            grpc_code = format!("{:?}", e.code()),
+                            grpc_message = e.message(),
+                            grpc_details = format!("{:?}", e.details())
+                        );
+                        OTelSdkError::InternalFailure("Logs export failed in interceptor".into())
```
lalitb, line 73: should we at least add the grpc_code here ?

cijothomas: neutral to it..we already log it here. And upstream SDK would log again... I can add just the `grpc_code` alone to the failure message.

lalitb: Also, add TODO to do selective error handling - i.e. return detailed info for non-auth errors

lalitb (approved):
LGTM with nit comments

PR #3110, opened 2025-08-02
fix: Change user-agent in OTLP Exporter to match spec

cijothomas:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

Changes default user-agent from "OTel OTLP Exporter Rust/version" to "OTel-OTLP-Exporter-Rust/version". Spec does not mandate this particular format, but the example given in the spec and some implementations (Java/.NET) follows this. Golang follows different format.

We could ask spec to clarify, but for now, I think its best to change to the ones shown in spec examples.

### DIFF
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Implementation of `Extractor::get_all` for `HeaderExtractor`
 - Support `HttpClient` implementation for `HyperClient<C>` with custom connectors beyond `HttpConnector`, enabling Unix Domain Socket connections and other custom transports
+- Add `reqwest` and `reqwest-blocking` features to enable async and blocking
+  reqwest HTTP clients
 
 ## 0.30.0
 
--- a/opentelemetry-http/Cargo.toml
+++ b/opentelemetry-http/Cargo.toml
@@ -13,8 +13,10 @@ autobenches = false
 [features]
 default = ["internal-logs"]
 hyper = ["dep:http-body-util", "dep:hyper", "dep:hyper-util", "dep:tokio"]
-reqwest-rustls = ["reqwest", "reqwest/rustls-tls-native-roots"]
-reqwest-rustls-webpki-roots = ["reqwest", "reqwest/rustls-tls-webpki-roots"]
+reqwest = ["dep:reqwest"]
+reqwest-blocking = ["dep:reqwest", "reqwest/blocking"]
+reqwest-rustls = ["dep:reqwest", "reqwest/rustls-tls-native-roots"]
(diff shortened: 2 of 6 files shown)


lalitb (approved):

lalitb:
FYI - there is PR in specs to make user-agent field configurable - https://github.com/open-telemetry/opentelemetry-specification/pull/4560

utpilla (approved):

cijothomas:
> FYI - there is PR in specs to make user-agent field configurable - [open-telemetry/opentelemetry-specification#4560](https://github.com/open-telemetry/opentelemetry-specification/pull/4560)

yes that is to append something to user-agent right, not outright replacement? replacement can be done today by passing extra header with the same name, to replace the existing one.

cijothomas:
> > replacement can be done today by passing extra header with the same name, to replace the existing one.
> 
> Are we validating the header values in such cases? If not, we should add some validation at least for length and character set.
> 
> https://medium.com/perimeterx/user-agent-based-attacks-are-a-low-key-risk-that-shouldnt-be-overlooked-1a5d487a1aa0

No we are not doing validations on _any_ additional headers users add. Should we be doing that? There isn't anything about this in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specifying-headers-via-environment-variables), so we have to come up with our own rules.

We can do some sanity check when headers are provided at build time. But users can add interceptors, and add/modify headers on each request - so we need to add validation here too after all user-interceptors are done. Even then users can still bring their own http client or gRPC channel, at which point we (OTel) has minimal control?

Anyway, this is better discussed in a separate issue.

cijothomas:
> > > replacement can be done today by passing extra header with the same name, to replace the existing one.
> > 
> > 
> > Are we validating the header values in such cases? If not, we should add some validation at least for length and character set.
> > https://medium.com/perimeterx/user-agent-based-attacks-are-a-low-key-risk-that-shouldnt-be-overlooked-1a5d487a1aa0
> 
> No we are not doing validations on _any_ additional headers users add. Should we be doing that? There isn't anything about this in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specifying-headers-via-environment-variables), so we have to come up with our own rules.
> 
> We can do some sanity check when headers are provided at build time. But users can add interceptors, and add/modify headers on each request - so we need to add validation here too after all user-interceptors are done. Even then users can still bring their own http client or gRPC channel, at which point we (OTel) has minimal control?
> 
> Anyway, this is better discussed in a separate issue.

https://github.com/open-telemetry/opentelemetry-rust/issues/3115 opened

PR #3116, opened 2025-08-05
fix: BatchSpanProcessor implementation to match that of logs

cijothomas:
Mostly copying logic from Logs implementation to Span. In a future time, I can look into making this generic and keep single implementation for Logs/Spans.

This also exposed a bug that shutdown() was not invoked to exporter. Added tests to cover this, which requiring enhancing `InMemorySpanExporter` too.

### DIFF
--- a/opentelemetry/benches/context_attach.rs
+++ b/opentelemetry/benches/context_attach.rs
@@ -98,7 +98,7 @@ fn dummy_work() -> i32 {
     black_box(1 + 1)
 }
 
-fn group(c: &mut Criterion) -> BenchmarkGroup<WallTime> {
+fn group(c: &mut Criterion) -> BenchmarkGroup<'_, WallTime> {
     c.benchmark_group("context_attach")
 }
 


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -460,29 +459,37 @@ impl BatchSpanProcessor {
...
-        // Get upto `max_export_batch_size` amount of spans from the channel and push them to the span vec
-        while let Ok(span) = spans_receiver.try_recv() {
-            spans.push(span);
-            if spans.len() == config.max_export_batch_size {
-                break;
+        let target = current_batch_size.load(Ordering::Relaxed); // `target` is used to determine the stopping criteria for exporting spans.
+        let mut result = OTelSdkResult::Ok(());
+        let mut total_exported_spans: usize = 0;
+
+        while target > 0 && total_exported_spans < target {
```
TommyCpp, line 466: `current_batch_size` might be changed by other tasks so we are using `target` which is a snapshot of  `current_batch_size` here.

If we won't run this funcion concurrently then we should consider not passing in a `AtomicUsize` as `current_batch_size`

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -493,14 +500,17 @@ impl BatchSpanProcessor {
         let export = exporter.export(batch.split_off(0));
         let export_result = futures_executor::block_on(export);
 
+        // Clear the batch vec after exporting
+        batch.clear();
```
TommyCpp: `batch.split_off(0)` should already clear the batch?

opentelemetry-sdk/src/trace/span_processor.rs
cijothomas: It can be changed by other threads, but this method only exports up to snapshot. If more items get into the queue which it is working, that'd be handled by another call to this method. IOW, this method has a target it set out to achieve - it'll work until that target is achieved and nothing more.

Also, this function is not run concurrently - but we still need `current_batch_size` as `AtomicSize` as this function updates its value based on how much it exports.

opentelemetry-sdk/src/trace/span_processor.rs
cijothomas: Good catch! This is not required for Span. (It was required in Logs case, as it does not give up ownership of batch to export only a slice is passed)

TommyCpp (approved):

PR #3122, opened 2025-08-05
docs: start doc for distributed tracing and logs guidance

cijothomas:
Adds initial guidance docs for logs and traces in the OpenTelemetry Rust ecosystem.

- `docs/logs.md` — marked **Stable**.
- `docs/traces.md` — marked **Work-In-Progress** (OpenTelemetry Tracing in this repo is not yet stable).


docs/traces.md
```diff
@@ -0,0 +1,56 @@
...
+   applies to a group of logs. The `otel-appender-tracing` crate will be
+   enhanced to extract span attributes and attach them to logs automatically.
+
+4. **If using tracing::span! to create spans**
+
+   This is not directly supported by OpenTelemetry. Use the
+   `tracing-opentelemetry` bridge to convert tracing spans into OTel spans.
+
+   There are some limitations with this approach arising due to `tracing`s lack of support for
+   creating Spans following OpenTelemetry specification. Examples include
```
davidbarsky:
```suggestion
   There are some limitations with this approach arising due to `tracing`s lack of support for
   creating Spans following OpenTelemetry specification. For example, `tracing` is unable to:
```

docs/traces.md
```diff
@@ -0,0 +1,56 @@
...
+   will be converted to OTel LogRecords via opentelemetry-appender-tracing and
+   will be automatically correlated with the current active OTel trace context
+   as well.
+
+3. **In-proc contextual enrichment for logs/events**
+
+   This is not something OTel has a spec-ed out solution for. This is very
+   specific to the logging library (tracing) and its bridge.
+
+   Use `tracing::span!` macros to add contextual metadata (e.g., filename) that
```
davidbarsky: Small nit: make this the *first* sentence, _then_ explain that OTel does not have a spec'd out solution for this use-case, by-design.

docs/traces.md
```diff
@@ -0,0 +1,56 @@
...
+## Introduction
+
+This document provides comprehensive guidance on leveraging OpenTelemetry traces
+in Rust applications.
+
+## Instrumentation Guidance
+
+1. **Use OTel API for distributed traces**
+
+   Use the `opentelemetry::trace` API to create spans. This supports context
```
davidbarsky: Do I have this right?

```suggestion
Use the `opentelemetry::trace` API to create that should be propagated across a process boundary. This supports context
```

docs/traces.md
```diff
@@ -0,0 +1,56 @@
+# OpenTelemetry Rust Traces
+
+Status: **Work-In-Progress**
+
+## Introduction
```
davidbarsky, line 5: I'd have the first sentence have a high-level overview as to what the guidance *is*—e.g., say something like "use `tokio-rs/tracing` within a process, use OpenTelemetry at the process boundaries", and then go into detail as below.

docs/traces.md
```diff
@@ -0,0 +1,56 @@
...
+   - Cannot specify span kind (e.g., server/client)
+   - Cannot add span links
+
+   The bridge offers extension APIs to support some of these, but they are not
+   standard and are maintained outside the OpenTelemetry and Tracing project and
+   within the bridge itself.
+
+   TODO: Should we make a recommendation about
+   avoiding this extension APIs for instrumentation?
+
```
alessandrobologna: I actually find the OpenTelemetrySpanExt (I assume you are referring to it) quite useful, and functional in its intended goal of bridging the gap. You can use it effectively for [context propagation](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/opentelemetry-remote-context.rs#L25-L45), setting span kind, adding links and and even dynamic attributes to a span (that are not supported by the pure tracing::span! macro). 

I understand that is not standard for open telemetry, but then also the event!, info!  etc macros are not standard interfaces in opentelemetry (and also suffer from the same limitation that they don't support dynamic attributes). In my humble opinion, to recommend avoiding altogether the OpenTelemetrySpanExt, which I suspect is widely used in the wild (i certainly use it),  is a little premature at this point.

docs/traces.md
cijothomas: The intention here was to recommend OTel API for *all* spans, not just the ones on edge.
and recommend using `tracing:span` for contextual log enrichment only. 

Of course, this is just the basic scenario, and one most people can just follow if starting from scratch. And this will be just like any other OTel clients like Java or Go etc.

 For users who already use `tracing:span` and want to see them as OTel Spans, then # 4 applies.

In my mind, the ideal state is "For non-edge spans (aka internal spans), users should be able to pick between `tracing:span` or `OTel Tracing API` with nearly same effect.

docs/traces.md
cijothomas: fixed

docs/traces.md
cijothomas: Thanks for sharing your thoughts!

> OpenTelemetrySpanExt (I assume you are referring to it) quite useful,

Of course they are useful and mostly works. I don't think we (OTel) can recommend using it - it's not a standard API, not bound by OTel specs. (I totally understand users use them a lot which is why we need to come up with formal guidance)

> but then also the event!, info! etc macros are not standard interfaces in opentelemetry

This is by-design. OTel Rust does not intend to make an end-user facing Logging API, and recommends exiting logging libraries. This is the case in every OTel language implementations. (though some languages have started offering end-user facing logging API (eg: C++), OTel Rust has no such plans.)

alessandrobologna: got it, so the direction is to keep a dependency with tokio tracing only for logging, and to add the tracing layer (with tracing-opentelemetry) only if we intend to surface tokio tracing spans (maybe created in other crates) in our traces, otherwise just stick with straight otel apis? 

Maybe it would be useful to add a link to examples in this context, I didn't realize that now they are [actually good quality and provide a great starting point]( https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/tracing-http-propagator/src/server.rs#L80-L103).

docs/logs.md
```diff
@@ -0,0 +1,55 @@
...
+## OTel Log Bridge API
+
+The OpenTelemetry Log Bridge API (part of the `opentelemetry` crate) is **not intended
+for direct use by application developers**. It is provided for authoring log
+appenders that bridge existing logging systems with OpenTelemetry. Bridges for
+`tracing` and `log` crates are already available.
+
+## Instrumentation Guidance
+
+1. **Use the `tracing` crate**: We strongly recommend using the
```
scottgerring, line 37: Aware I sound like a broken record, but i'm still not sure about this, as using the tracing crate while The Great Tracing Bridge Integration Issue remains _can_ lead users down a frustrating path, depending on what they're doing.

Having said that, once Björns PR is merged, I think this is good guidance.

docs/traces.md
```diff
@@ -0,0 +1,63 @@
...
+   If you are creating spans to track in-proc work (what OTel calls "internal" spans),
+   `tracing:span` API is sufficient with `tracing-opentelemetry` bridge converting the
+   `tracing` Span to OTel Span, and properly activating/de-activating OTel's context,
```
scottgerring: > properly activating/de-activating OTel's context

Is this the right level of detail for a user-facing doc?

docs/logs.md
cijothomas: this is specifically for logs, not spans.

cijothomas:
Marked ready for review. @davidbarsky @scottgerring — would appreciate another look; the open threads from your earlier review should now be addressed (intro now leads with a thesis, level-of-detail trimmed). @alessandrobologna — courtesy ping; the `tracing-opentelemetry` framing is now neutral and I added the `tracing-http-propagator` example to See Also per your suggestion.

docs/logs.md
```diff
@@ -0,0 +1,106 @@
...
+## Terminology
+
+OpenTelemetry defines Events as Logs with an EventName. When you follow the guidance
+above and explicitly set the `name` field on every `tracing` log, each log maps to
+an OpenTelemetry Event. (Without an explicit `name`, the synthesized source-location
+string is technically present but is not a meaningful EventName.)
+
+**Note**: These are **not** mapped to Span Events. OpenTelemetry is deprecating
+Span Events in favor of Events (Logs with an EventName), so use Events as
+described above rather than Span Events.
```
scottgerring: The spec is only deprecating the **API** for span events, right? Assuming we are talking about [OTEP-4430](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md). 

I note also the [event-to-span bridge in the spec in development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md?plain=1#L556-L595) which seems to have been touched in the last 2 months.

Maybe here just something like ...


```suggestion
**Note**: These are **not** mapped to Span Events. If you specifically need
Span Events today, record them with the OpenTelemetry tracing API
(`Span::add_event` / `add_event_with_timestamp`). Otherwise, prefer Events
(Logs with an EventName) as described above.
```

?

docs/logs.md
```diff
@@ -0,0 +1,106 @@
...
+## Introduction
+
+This document provides guidance on leveraging OpenTelemetry logs in Rust
+applications.
+
+In short: use [`tracing`] for logs and events; OpenTelemetry plugs in via the
+[`opentelemetry-appender-tracing`] appender. Existing `tracing` (or `log`)
+instrumentation continues to work as-is — adopting OpenTelemetry for logs is
+a setup change, not a code rewrite. For span guidance, see
+[traces.md](traces.md).
```
scottgerring: If we talk about 'application logging' we can remove the use of logs-and-the-very-overloaded-events and make it a bit less ambiguous I reckon:

 ```suggestion
 In short: for application logging, use [`tracing`] with the
 [`opentelemetry-appender-tracing`] appender. Existing `tracing` or `log`
 instrumentation can continue to work as-is; adopting OpenTelemetry for logs is
primarily a setup change, not a code rewrite. For span guidance, see
[traces.md](traces.md).
 ```

docs/traces.md
```diff
@@ -0,0 +1,79 @@
...
+   parent — concepts central to the OpenTelemetry specification, particularly
+   for *edge* spans (see #3 below for nuance). Use it primarily for
+   logs/events (see [logs.md](logs.md)).
+
+3. **Bridging from `tracing::span!` to OpenTelemetry spans.** If you are
+   already using `tracing::span!` and want those spans surfaced as
+   OpenTelemetry spans, the third-party [`tracing-opentelemetry`] crate
+   provides a bridge. It is maintained outside the OpenTelemetry project and
+   is not part of this repo; we point to it so users are aware of the option
+   in the broader ecosystem.
```
scottgerring: I think we can be more succinct here:

```suggestion
   provides a bridge. It is maintained outside the OpenTelemetry project and is not part of this repo.
```

docs/traces.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Instrumentation Guidance
+
+1. **Use the OpenTelemetry Tracing API to create spans.** The
+   `opentelemetry::trace` API is designed around the OpenTelemetry
+   specification, with first-class support for span kind
+   (server/client/producer/consumer/internal), links, remote parents, and
+   context propagation across process boundaries.
+
+2. **[`tracing`] is not a complete substitute for the OpenTelemetry Tracing API.**
```
scottgerring: I think it would be helpful to explain the conceptual difference between the two so folks can form a mental model. I hesitate to call tracing a "structured logging library" as that undersells it a bit, but maybe something like this? 

```suggestion
2. **[`tracing`] is designed to collect structured, event-based diagnostic information and is not a complete substitute for the OpenTelemetry Tracing API, which focusses on distributed tracing**
```

docs/traces.md
```diff
@@ -0,0 +1,79 @@
...
+   For *internal* spans (spans that represent in-process work and never cross
+   a process boundary), `tracing::span!` through this bridge produces a
+   result nearly identical to using the OpenTelemetry Tracing API directly —
+   span kind, links, and remote parent are not relevant for internal spans.
+   The `tracing` limitations matter primarily for *edge* spans (e.g.,
+   incoming/outgoing HTTP, messaging), where span kind, links, and remote
+   parents are central to the OpenTelemetry data model. The bridge offers
+   extension APIs to express these concepts.
+
+4. **Prefer instrumentation libraries.** For framework-level spans (HTTP
```
scottgerring: Perhaps this should be the _first_ point - you'd generally plug these in to get the overall shape of the request/response/downstreams, then add your own custom spans to flavour internally

docs/traces.md
scottgerring: I think this could also be a subheading of 1

scottgerring (commented):
Some nits!

cijothomas:
@scottgerring Addressed your comments, thank you!

scottgerring (approved):
🚢

cijothomas:
@open-telemetry/rust-maintainers Please review. I'd like to get more eyes before merging this.

TommyCpp (approved):
Overall LGTM as the first version

docs/traces.md
```diff
@@ -0,0 +1,84 @@
...
+4. **Bridging from `tracing::span!` to OpenTelemetry spans.** If you are
+   already using `tracing::span!` and want those spans surfaced as
+   OpenTelemetry spans, the third-party [`tracing-opentelemetry`] crate
+   provides a bridge. It is maintained outside the OpenTelemetry project
+   and is not part of this repo; we mention it here for completeness.
+
+   For *internal* spans (spans that represent in-process work and never cross
+   a process boundary), `tracing::span!` through this bridge produces a
+   result nearly identical to using the OpenTelemetry Tracing API directly —
+   span kind, links, and remote parent are not relevant for internal spans.
```
lalitb: I don’t think links are safe to classify as “not relevant” for internal spans.  Could we narrow this to span kind/remote parent, or say links are less common but still possible?

docs/logs.md
```diff
@@ -0,0 +1,110 @@
...
+
+2. **Lean on the `tracing` ecosystem**: OpenTelemetry doesn't replace what
+   `tracing` already offers. The appender is a standard `tracing-subscriber`
+   `Layer`, so it composes with `fmt::Layer`, `EnvFilter`, and any other
+   existing layer — for example, sending logs to stdout via `tracing`'s
+   `fmt::Layer` while exporting the same logs to an OTLP endpoint via
+   OpenTelemetry, or filtering what reaches the OpenTelemetry pipeline. Use
+   `tracing`'s ecosystem directly; OpenTelemetry just plugs into it.
+
+3. **Explicitly provide `name` and `target` fields**: These map to OpenTelemetry's
```
lalitb, line 49: This seems to overstates the `target` mapping. Only OTLP exporter overrides scope name from `target. But this doc reads like `target` is always the instrumentation scope. Could we make that distinction explicit?

docs/traces.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Introduction
+
+This document provides guidance on leveraging OpenTelemetry traces in Rust
+applications.
+
+In short: prefer instrumentation libraries that use the OpenTelemetry Tracing
+API for framework-level spans, and use the OpenTelemetry Tracing API
+directly to create your own custom spans. [`tracing`] is for logs and events,
+not spans. See [logs.md](logs.md) for log guidance.
```
lalitb: > [`tracing`] is for logs and events, not spans.

 Could we soften this? As written, it sounds like `tracing` is not for spans at all, even though `tracing` has its own span concept. Maybe say it is not a complete substitute for the OpenTelemetry Tracing API instead.

docs/logs.md
```diff
@@ -0,0 +1,110 @@
...
+OpenTelemetry defines Events as Logs with an EventName. When you follow the guidance
+above and explicitly set the `name` field on every `tracing` log, each log maps to
+an OpenTelemetry Event. (Without an explicit `name`, the synthesized source-location
+string is technically present but is not a meaningful EventName.)
+
+**Note**: These are **not** mapped to Span Events. If you specifically need
+Span Events today, record them with the OpenTelemetry tracing API
+(`Span::add_event` / `add_event_with_timestamp`). Otherwise, prefer Events
+(Logs with an EventName) as described above. The OpenTelemetry specification
+is deprecating the *Span Events API* (see [OTEP-4430]) in favor of Events.
```
lalitb: Could we soften this wording? The current Trace API spec still lists AddEvent as a stable Span API, while OTEP-4430 is a deprecation plan/ proposal. Maybe say “OTEP-4430 proposes deprecating the Span Events API” instead of “the specification is deprecating”?

lalitb (approved):
LGTM - few comments, good to resolve them in this or separate PR.

docs/logs.md
cijothomas: It is not overstating - target *is the* scope. If an exporter does not treat it that way, it's a bug. I'll follow up with a doc update on `target` to make this stated more clearly.

docs/traces.md
cijothomas: reworded in latest commit.

docs/traces.md
cijothomas: Good point. addressed in latest commit

cijothomas:
@lalitb @scottgerring @TommyCpp I reworded a bit more, could use another review before merging.

cijothomas:
@bantonsson Please review!

TommyCpp (approved):

scottgerring:
I think this is good to merge eh @cijothomas ?

cijothomas:
> I think this is good to merge eh @cijothomas ?

Yes adding to merge-queue.

PR #3126, opened 2025-08-12
feat: support backoff/retry in OTLP

### LINKED ISSUE #3081: OTLP Stabilization: Throttling
Exporter MUST honour `RetryInfo` / `Retry‑After` for both generic retryable errors and throttling signals (HTTP 429 / gRPC resource_exhausted), applying exponential back‑off. gRPC & HTTP paths ignore this today.  

This can likely be done leveraging existing work in #2727

scottgerring:
Fixes #3081, building on the work started by @AaronRM 🤝  

### A new **retry** module added to `opentelemetry-sdk` 
Models the sorts of retry an operation may request (retry / can't retry / throttle), and provides a helper `retry_with_backoff` mechanism that can be used to wrap up a retryable operation and retry it. The helper relies on `experimental_async_runtime` for its runtime abstraction, to provide the actual pausing. It also takes a lambda to classify the error, so the caller can inform the retry mechanism if a retry is required.

### A new **retry_classification** module added to `opentelemetry-otlp`
This bit takes the actual error responses that we get back over OTLP and maps them back to the retry model. Because this is OTLP-specific stuff it belongs here rather than alongside the retry code. 

### Retry binding
... happens in each one of the concrete exporters to tie it all together. 

Also ...

* Extended exporter builders to allow the user to customise default retry policy
* Added new feature flags experimental-http-retry and experimental-grpc-retry which pull in the experimental-async-runtime dep and set everything up. This way we can get going with this _now_ without having to stabilise the experimental-async-runtime feature.

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -388,7 +388,6 @@ impl OtlpHttpClient {
         #[cfg(feature = "experimental-http-retry")]
         {
             use crate::retry::retry_with_backoff;
-            use opentelemetry_sdk::runtime::Tokio;
 
             // Build request body once before retry loop
             let (body, content_type, content_encoding) = build_body_fn(self, data)
@@ -400,8 +399,17 @@ impl OtlpHttpClient {
                 endpoint: self.collector_endpoint.to_string(),
             });
 
+            // Select runtime based on HTTP client feature - if we're using
+            // one without Tokio, we don't need or want the Tokio async blocking
+            // behaviour.
+            #[cfg(feature = "reqwest-blocking-client")]
+            let runtime = opentelemetry_sdk::runtime::NoAsync;
+
+            #[cfg(not(feature = "reqwest-blocking-client"))]
+            let runtime = opentelemetry_sdk::runtime::Tokio;
+
             retry_with_backoff(
-                Tokio,
+                runtime,
                 self.retry_policy.clone(),
                 classify_http_export_error,
                 operation_name,


opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -58,40 +63,65 @@ impl TonicLogsClient {
...
-            Ok(_) => {
-                otel_debug!(name: "TonicLogsClient.ExportSucceeded");
-                Ok(())
-            }
-            Err(e) => {
-                let error = format!("export error: {e:?}");
-                otel_debug!(name: "TonicLogsClient.ExportFailed", error = &error);
-                Err(OTelSdkError::InternalFailure(error))
-            }
+        let batch = Arc::new(batch);
```
scottgerring, line 74: see if we can avoid this

opentelemetry-otlp/Cargo.toml
```diff
@@ -35,6 +35,7 @@ tracing = {workspace = true, optional = true}
 
 prost = { workspace = true, optional = true }
 tonic = { workspace = true, optional = true }
+tonic-types = { workspace = true, optional = true }
```
scottgerring, line 38: Needed for gRPC error type

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -4,9 +4,124 @@ use http::{header::CONTENT_TYPE, Method};
...
+use super::{classify_http_export_error, HttpExportError, HttpRetryData};
+#[cfg(feature = "http-retry")]
+use opentelemetry_sdk::retry::{retry_with_backoff, RetryPolicy};
+#[cfg(feature = "http-retry")]
+use opentelemetry_sdk::runtime::Tokio;
+
 impl LogExporter for OtlpHttpClient {
+    #[cfg(feature = "http-retry")]
+    async fn export(&self, batch: LogBatch<'_>) -> OTelSdkResult {
+        let policy = RetryPolicy {
```
scottgerring: Should we create a default retry policy and share it between all exporters?

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -4,9 +4,124 @@ use http::{header::CONTENT_TYPE, Method};
...
+
+                otel_debug!(name: "HttpLogsClient.ExportSucceeded");
+                Ok(())
+            },
+        )
+        .await
+        .map_err(|e| OTelSdkError::InternalFailure(e.message))
+    }
+
+    #[cfg(not(feature = "http-retry"))]
```
scottgerring: This is a massive duplication of code.

If we decide that http export always has retry behaviour, which means we include the unstable runtime feature, then we can remove all of this.

Alternatively we can provide `export` once, with all the extra pomp and fanfare to support retry, and then just use the stub "don't actually retry" impl. There would be some slight runtime overhead to this, but the codebase would be much simpler.

scottgerring: Comment above applies to the other HTTP exporters also.

opentelemetry-otlp/src/retry.rs
```diff
@@ -0,0 +1,461 @@
...
+// Generates a random jitter value up to max_jitter
+#[cfg(feature = "experimental_async_runtime")]
+fn generate_jitter(max_jitter: u64) -> u64 {
+    let now = SystemTime::now();
+    let nanos = now
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
+        .subsec_nanos();
+    nanos as u64 % (max_jitter + 1)
+}
```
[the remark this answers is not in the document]
scottgerring: This is for jitter and I don't see how this could matter.

opentelemetry-otlp/src/exporter/tonic/trace.rs
```diff
@@ -60,42 +65,68 @@ impl TonicTracesClient {
...
-            Ok(_) => {
-                otel_debug!(name: "TonicTracesClient.ExportSucceeded");
-                Ok(())
-            }
-            Err(e) => {
-                let error = e.to_string();
-                otel_debug!(name: "TonicTracesClient.ExportFailed", error = &error);
-                Err(OTelSdkError::InternalFailure(error))
-            }
+        let batch = Arc::new(batch);
```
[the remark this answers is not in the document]
scottgerring: This is cloning the Arc not the batch which is a cheap operation, especially at the relatively slow cadence a retry will run at

opentelemetry-otlp/src/exporter/http/logs.rs
bantonsson: Wouldn't it be possible to use the same code for the logic (broken out into a method) and construct the `HttpError` instances but just have different variants of `retry_with_backoff` and the `HttpError` _classification_ depending on if we have the unstable runtime feature enabled?

opentelemetry-otlp/src/exporter/http/logs.rs
```diff
@@ -4,9 +4,124 @@ use http::{header::CONTENT_TYPE, Method};
...
+            Tokio,
+            policy,
+            classify_http_export_error,
+            "HttpLogsClient.Export",
+            || async {
+                // Get client
+                let client = self
+                    .client
+                    .lock()
+                    .map_err(|e| HttpExportError {
```
bantonsson: I think the code would be easier to read if there were two constructors for `HttpExportError`. Something like `HttpExportError::new(500, format!("Mutex lock failed: {e}"))` and `HttpExportError::with_retry_after(...)`. It would cut down on the boiler plate.

opentelemetry-otlp/src/exporter/http/logs.rs
bantonsson: I think a default retry policy makes sense.

opentelemetry-otlp/src/exporter/http/logs.rs
scottgerring: Good catch - have implemented

opentelemetry-otlp/src/exporter/http/logs.rs
scottgerring: @bantonsson something like this commit? --> https://github.com/open-telemetry/opentelemetry-rust/pull/3126/commits/ba1f9e2f6179d450e9ef4fea246a52e6e57ce61f

I've factored the retry behaviour out into `otlp::exporter::http`, and then switch out the implementation based on the `retry-retry` flag. This removes huge volumes of code duplication and means that we could get this in without stabilising the async runtime feature.

> and the HttpError classification depending on if we have the unstable runtime feature enabled?

I'm not sure the classification needs to change independently of the "do we use the retry variant or not" bit - but I might be missing something?

scottgerring: I've also not touched this for the tonic exporters yet, but once we have the HTTP ones in shape I can continue the pattern over there.

bantonsson: > I'm not sure the classification needs to change independently of the "do we use the retry variant or not" bit - but I might be missing something?

You're not missing anything. That refactoring is cleaner than what I was thinking about.

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -291,19 +342,62 @@ impl TonicExporterBuilder {
...
 
+/// Wrapper for retry functionality in tonic exporters.
+/// Provides a unified call path that either uses retry_with_backoff when experimental-grpc-retry
+/// feature is enabled, or executes the operation once when it's not.
+#[cfg(all(
+    feature = "grpc-tonic",
+    feature = "experimental-grpc-retry",
+    any(feature = "trace", feature = "metrics", feature = "logs")
+))]
+async fn tonic_retry_with_backoff<R, F, Fut, T>(
```
scottgerring, line 366: Because of the way the tonic path works, it is much harder to cleanly extract as much from the signal-specific exporters as we managed with HTTP.

I've done this as a kind of middleground - this way the feature flagging we use to sub between retry-available / no-retry is all in one place - as a thin wrapper around the global retry function `retry_with_backoff`.

opentelemetry-otlp/src/exporter/http/logs.rs
scottgerring: Because of the way the tonic exporters are structured its harder to cleanly pull it apart. I've struck what I think is a sensible compromise over there, adding a wrapper for the retry function that subs in retry/no-op based on feature flags. I think this is better than going too wild with traits/generics/lambdas to squeeze the last little bit of code re-use out in terms of clarity.

opentelemetry-otlp/src/exporter/http/logs.rs
scottgerring: Added!

opentelemetry-otlp/src/exporter/tonic/mod.rs
bantonsson: I keep trying to slice and dice this in some nice way, but the generated clients do not implement a common interface, so it's complicated to do common code without changing them.

bantonsson (approved):
👍🏼 for the HTTP code. I can't see a clear way to reuse more of the Tonic code.

opentelemetry-otlp/src/exporter/tonic/mod.rs
scottgerring: Yeah, I know what you mean. I came up with _something_ that saved a few lines by forcing some intermediate abstractions but it didn't reduce line count and certainly didn't improve clarity.

I feel like the way it is strikes a reasonable balance even if it isn't particularly satisfying.

opentelemetry-sdk/src/lib.rs
```diff
@@ -167,3 +167,6 @@ impl<T> From<std::sync::PoisonError<T>> for InMemoryExporterError {
         InMemoryExporterError::InternalFailure(format!("Mutex poison error: {err}"))
     }
 }
+
+/// Retry logic for exporting telemetry data.
+pub mod retry;
```
lalitb: While this is a useful, reusable component, do we really expect it to be used outside of `opentelemetry-otlp`? Making it a public API in the SDK would increase our maintenance burden - especially if users begin depending on it for non-OTel use cases. A safer approach would be to keep it crate-private within `opentelemetry-otlp` for now, and only promote it into the SDK if we later see a clear need in other areas.

opentelemetry-otlp/src/retry.rs
```diff
@@ -0,0 +1,461 @@
...
+        Self
+    }
+}
+
+// Generates a random jitter value up to max_jitter
+#[cfg(feature = "experimental_async_runtime")]
+fn generate_jitter(max_jitter: u64) -> u64 {
+    let now = SystemTime::now();
+    let nanos = now
+        .duration_since(SystemTime::UNIX_EPOCH)
```
lalitb, line 91: Wondering if using current system time right approach to calculate jitter. Two clients that happen to call this around the same instant will get the same value (especially if their clocks are synced by NTP or they’re on the same host). That means their “jitter” is identical or nearly identical, so they still retry in sync.

bantonsson: You could always send the value through the `DefaultHasher` to get some spread.

lalitb: > You could always send the value through the DefaultHasher to get some spread.

how will that solve the synchronization problem, since we would still be hashing the same input values.

scottgerring: I would expect hashed nanoseconds should be sufficient ? Likelihood of getting the same ns twice at once negligible. Happy to take any other suggestions too!

opentelemetry-sdk/src/lib.rs
scottgerring: I would expect any exporter would be interested in it, but I appreciate your point about stretching the API. I can move it 👌

opentelemetry-otlp/src/retry.rs
bantonsson: Indeed. The risk of getting the same ns twice is negligible.

lalitb: Makes sense. I was probably thinking of using a proper RNG, and also SystemTime syscall perf - but the current implementation works fine for now. We can revisit if needed, not a blocker.​​​​​​​​​​​​​​​​

opentelemetry-sdk/src/lib.rs
lalitb: My thoughts are: if only OTLP uses this retry logic and the other core exporters in main repo don’t use it, we should keep it scoped to the OTLP module. Will leave it to you to decide :)

scottgerring: I do agree with you - better minimise scope when it can be minimised! I’ll clean this up and the rand and the missing date parsing once I’m back from the break.

bantonsson (approved):

scottgerring:
@lalitb, i've moved the retry down into otlp (rather than sdk), implemented the date parsing, and added a hash around the nanos for jitter. I think this should be ready to go! Would you like to give it another skim?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -298,16 +363,148 @@ impl HttpExporterBuilder {
...
+        build_body_fn: F,
+        operation_name: &'static str,
+    ) -> opentelemetry_sdk::error::OTelSdkResult
+    where
+        F: Fn(&Self, T) -> Result<(Vec<u8>, &'static str, Option<&'static str>), String>,
+    {
+        #[cfg(feature = "experimental-http-retry")]
+        {
+            use crate::retry::retry_with_backoff;
+            use opentelemetry_sdk::runtime::Tokio;
```
lalitb: This adds a `Tokio` dependency even for setups using the default (thread-based) BatchProcessor with `reqwest-blocking`, which previously didn’t require a runtime. Not a blocker since it’s feature-gated, but just checking if this dependency is intentional for the retry path.

lalitb (approved):
LGTM overall - implementation looks solid and well-structured. I left a comment on the Tokio dependency; not a blocker, but something we can revisit or refine once the feature flag is removed.

opentelemetry-otlp/src/exporter/http/mod.rs
scottgerring: Hey @lalitb this is a great catch. I think we need: 

1. A 'NoAsync' runtime, that uses sync sleep 
2. To select the correct runtime when we construct the OtlpHttpClient, defaulting to Tokio if we're not behind the reqwest-blocking flag (so we also get a runtime if a user specifies a custom client)

I think this is important to get into this change and will set it up.

scottgerring: Hey @lalitb i've gone ahead and done the above ⏫ . I think this looks sensible. Two things:

1/ I noticed whilst doing it was the `TokioCurrentThread` runtime; for the new features i've introduced e.g. `experimental-http-backoff` , we're depending directly on `rt-tokio` here and we don't have a way of using `rt-tokio-current-thread`. My gut feeling is we should _keep it simple_ for now (and effectively preclude using rt-tokio-current-thread with the retrying) rather than adding more feature sprawl. I am missing a bit of context, though, on the general usage of the two features here. What do you think?

2/ For your comment above (added tokio dep even when using reqwest-blocking), i've solved it in the code, but have not adjusted the feature deps; this means we're still pulling `rt-tokio` into the build. Again here, I am hesitant to add more feature variants, and would love your opinion!

lalitb: 1. Agree to keep it simple for now with multithread Tokio-only for the experimental retry/backoff initially, and open a tracking issue to discuss for current--thread-based support.
2. I think we should avoid pulling `rt-tokio` for non-async runtime scenario. Even if that needs adding feature flag. Let's handle it as separate PR.

scottgerring: Happy to do it in a separate PR; this one is getting big.
Reckon we're done then!

opentelemetry-otlp/src/retry.rs
```diff
@@ -0,0 +1,495 @@
...
+
+                match error_type {
+                    RetryErrorType::NonRetryable => {
+                        otel_warn!(name: "OtlpRetry", message = format!("Operation {:?} failed with non-retryable error: {:?}", operation_name, err));
+                        return Err(err);
+                    }
+                    RetryErrorType::Retryable if attempt < policy.max_retries => {
+                        attempt += 1;
+                        // Use exponential backoff with jitter
+                        otel_warn!(name: "OtlpRetry", message = format!("Retrying operation {:?} due to retryable error: {:?}", operation_name, err));
```
cijothomas, line 153: this should be `info` level, as we are not yet giving up and losing data.

opentelemetry-otlp/src/retry.rs
```diff
@@ -0,0 +1,495 @@
...
+    loop {
+        match operation().await {
+            Ok(result) => return Ok(result), // Return the result if the operation succeeds
+            Err(err) => {
+                // Classify the error
+                let error_type = error_classifier(&err);
+
+                match error_type {
+                    RetryErrorType::NonRetryable => {
+                        otel_warn!(name: "OtlpRetry", message = format!("Operation {:?} failed with non-retryable error: {:?}", operation_name, err));
```
cijothomas, line 147: Let's stick with structured logging as much as possible instead of stringifying. i.e Operation and Error should be own fields.

cijothomas, line 147: name:"OtlpRetry" - this is not the best usage of event names - as its reused many times in this file itself, each time with different event. We need to do dedicated event name for each distinct type of logs, and ensure the schema (fields/etc) is same for a given event name.

opentelemetry-otlp/src/retry.rs
```diff
@@ -0,0 +1,495 @@
...
+                        let delay_with_jitter = std::cmp::min(delay + jitter, policy.max_delay_ms);
+                        runtime
+                            .delay(Duration::from_millis(delay_with_jitter))
+                            .await;
+                        delay = std::cmp::min(delay * 2, policy.max_delay_ms); // Exponential backoff
+                    }
+                    RetryErrorType::Throttled(server_delay) if attempt < policy.max_retries => {
+                        attempt += 1;
+                        // Use server-specified delay (overrides exponential backoff)
+                        otel_warn!(name: "OtlpRetry", message = format!("Retrying operation {:?} after server-specified throttling delay: {:?}", operation_name, server_delay));
```
cijothomas, line 164: please downgrade to info level.

opentelemetry-otlp/src/retry.rs
```diff
@@ -0,0 +1,495 @@
...
+    pub max_retries: usize,
+    /// Initial delay in milliseconds before the first retry.
+    pub initial_delay_ms: u64,
+    /// Maximum delay in milliseconds between retries.
+    pub max_delay_ms: u64,
+    /// Maximum jitter in milliseconds to add to the delay.
+    pub jitter_ms: u64,
+}
+
+/// A runtime stub for when experimental_async_runtime is not enabled.
```
cijothomas, line 64: Not sure I follow the use case for this. This PR already adds a runtime implementation for the dedicated thread case in the Sdk crate. Should the OTLP exporter be aware of anything more than letting the Sdk Runtime implementation do its own delay - either using asynchronous delay or blocking sleeps?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -298,16 +363,156 @@ impl HttpExporterBuilder {
...
+            let (body, content_type, content_encoding) = build_body_fn(self, data)
+                .map_err(opentelemetry_sdk::error::OTelSdkError::InternalFailure)?;
+
+            let retry_data = Arc::new(HttpRetryData {
+                body,
+                headers: self.headers.clone(),
+                endpoint: self.collector_endpoint.to_string(),
+            });
+
+            // Select runtime based on HTTP client feature - if we're using
```
cijothomas, line 402: this might need to be made more robust..OTLP exporter having to pick the runtime feels flaky - it won't know of all possible runtime implementations.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Released 2025-Sep-25
 
 - Update `opentelemetry-proto` and `opentelemetry-http` dependency version to 0.31.0
 - Add HTTP compression support with `gzip-http` and `zstd-http` feature flags
+- Add retry with exponential backoff and throttling support for HTTP and gRPC exporters
```
cijothomas, line 11: lets add some more details here so the user reading the changelog would know how to use this feature. (given this is experimental and opt-in).

opentelemetry-otlp/src/retry.rs
scottgerring: ack

opentelemetry-otlp/src/retry.rs
scottgerring: Yeah, this is vestigial and slipped through. We can remove it.

opentelemetry-otlp/src/exporter/http/mod.rs
scottgerring: open to suggestions; there was a brief discussion around this.

opentelemetry-otlp/CHANGELOG.md
scottgerring: sure

PR #3131, opened 2025-08-20
feat: allow Resource transformation in TracerProviderBuilder

bantonsson:
## Changes

Adds the ability to transform a `Resource` in the `TracerProviderBuilder` so it's possible to add configuration from other sources in a safe way.

This has come up when trying to add resource configuration based on vendor specific environment variables while not overriding the resources configured by the OpenTelemetry ones.


opentelemetry-sdk/src/trace/provider.rs
```diff
@@ -422,6 +422,25 @@ impl TracerProviderBuilder {
...
+    /// The transformed [Resource] represents the entity producing telemetry and
+    /// is associated with all [Tracer]s the [SdkTracerProvider] will create.
+    ///
+    /// By default, if this option is not used, the default [Resource] will be used.
+    ///
+    /// *Note*: Calls to this method are not additive, the returned resource will
+    /// completely replace the existing [Resource].
+    ///
+    /// [Tracer]: opentelemetry::trace::Tracer
+    pub fn transform_resource<F>(self, f: F) -> Self
```
scottgerring, line 436: The type isn't so long ... would it be cleaner inline? Something like

```suggestion
    pub fn transform_resource(self, f: impl FnOnce(Option<&Resource>) -> Resource) -> Self {
```

scottgerring: Should we add `transform_resource` to the other signal builders also?

PR #3133, opened 2025-08-20
make `shutdown_with_timeout` required on `LogProcessor`

### LINKED ISSUE #3132: [Bug]: experimental async log exporter does not shutdown properly
### What happened?

The async `BatchLogProcessor` implements `shutdown` instead of `shutdown_with_timeout`. This means `shutdown_with_timeout` is defaulted to just return `Ok(())`, which is the default implementation in the `LogProcessor` trait.

`LoggerProvider` routes all calls to `shutdown_with_timeout`, so this means shutdown does nothing and no flushing happens.

Possibly a root cause bug is that `shutdown_with_timeout` has a default at all? The `SpanProcessor` trait does not have one.

`opentelemetry` 0.30.0

`opentelemetry-sdk` 0.30.0

_No response_

### Relevant log output

```shell

```

### Tip

davidhewitt:
Fixes #3132 

## Changes

Removes the default implementation of `shutdown_with_timeout` from `LogProcessor`, to match `SpanProcessor` which also requires this method.

This fixes two `LogProcessor` implementations which had implemented `shutdown` but not `shutdown_with_timeout`, meaning that they were doing a no-op shutdown as identified in #3132:
 - `SimpleLogProcessor`
 - async `BatchLogProcessor`

### DIFF
--- a/opentelemetry-sdk/src/logs/log_processor.rs
+++ b/opentelemetry-sdk/src/logs/log_processor.rs
@@ -57,9 +57,7 @@ pub trait LogProcessor: Send + Sync + Debug {
     /// Shuts down the processor.
     /// After shutdown returns the log processor should stop processing any logs.
     /// It's up to the implementation on when to drop the LogProcessor.
-    fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult {
-        Ok(())
-    }
+    fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult;
     /// Shuts down the processor with default timeout.
     fn shutdown(&self) -> OTelSdkResult {
         self.shutdown_with_timeout(Duration::from_secs(5))
@@ -140,6 +138,10 @@ pub(crate) mod tests {
         fn force_flush(&self) -> OTelSdkResult {
             Ok(())
         }
+
+        fn shutdown_with_timeout(&self, _timeout: std::time::Duration) -> OTelSdkResult {
+            Ok(())
+        }
     }
 
     #[derive(Debug)]
@@ -164,6 +166,10 @@ pub(crate) mod tests {
         }
 
         fn force_flush(&self) -> OTelSdkResult {
+            Ok(())
+        }
+
+        fn shutdown_with_timeout(&self, _timeout: std::time::Duration) -> OTelSdkResult {
             Ok(())
         }
     }
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -121,6 +121,10 @@ impl LogProcessor for EnrichWithBaggageLogProcessor {
     fn force_flush(&self) -> OTelSdkResult {
         Ok(())
     }
+
+    fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult {
+        Ok(())
+    }
 }
 
 /// A custom span processor that enriches spans with baggage attributes. Baggage
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -931,6 +931,10 @@ mod tests {
         fn force_flush(&self) -> OTelSdkResult {
             Ok(())
         }
+
+        fn shutdown_with_timeout(&self, _timeout: std::time::Duration) -> OTelSdkResult {
+            Ok(())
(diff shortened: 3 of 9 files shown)


opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -57,9 +57,7 @@ pub trait LogProcessor: Send + Sync + Debug {
     /// Shuts down the processor.
     /// After shutdown returns the log processor should stop processing any logs.
     /// It's up to the implementation on when to drop the LogProcessor.
-    fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult {
-        Ok(())
-    }
+    fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult;
```
cijothomas: we cannot make this change, as this is *breaking*.

davidhewitt: I documented it in the changelog as breaking, am fully aware. 

I defend it as a good change because as evidenced here there are existing bugs in production where implementations got that wrong. 

The span processor trait does not have this implementation, so removing it improves consistency too.

But if you prefer I can close the PR.

cijothomas: Logs (and Metrics) SDK were declared stable, so we should not make any breaking change.
(trace/span sdk is beta, so we can fix anything there.)

> there are existing bugs in production where implementations got that wrong.

Given we can't make breaking change, only option is to better document this so custom processor authors' won't make the mistake! 
The existing bug (in this repo) is in an experimental component only right?


> But if you prefer I can close the PR.

Let's fix the bug by implementing the right shutdown method in log_processor - that should be a good fix and no breaking change issues.

Agree it's not great. IIRC, we did have some discussions about providing default implementations before declaring stable, and ended up with a less-than-ideal choice :(

davidhewitt: Understood. I'll aim to come back to this PR probably on Monday.

> The existing bug (in this repo) is in an experimental component only right?

Unfortunately when running this change I found it also affects `SimpleLogProcessor`.

https://github.com/open-telemetry/opentelemetry-rust/pull/3133/files#diff-834c43f0e4972291eea6782a3ebf19e9b93f150d0b699dd5991f2ff52dc9c1fcR120

> Agree it's not great. IIRC, we did have some discussions about providing default implementations before declaring stable, and ended up with a less-than-ideal choice :(

Ouch.

I guess we can document that `shutdown` usually should not be implemented and `shutdown_with_timeout` should be? Can we change the default implementation of `shutdown_with_timeout` to log a warning? Or is that still too breaking?

cijothomas: adding a log is fine. It is not common for users to write processors that require implementing any of these.. Most use cases are around enrichment (adding more attributes), and they don't need to worry about shutdown method at all.

davidhewitt: I pushed a commit which re-adds this default implementation with an internal warning log.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -57,10 +57,21 @@ pub trait LogProcessor: Send + Sync + Debug {
     /// Shuts down the processor.
     /// After shutdown returns the log processor should stop processing any logs.
     /// It's up to the implementation on when to drop the LogProcessor.
+    ///
+    /// All implementors should implement this method.
```
cijothomas, line 61: This is incorrect statement. It is entirely upto the log processor to decide if they need to do anything in shutdown or not.

opentelemetry-sdk/src/logs/log_processor.rs
```diff
@@ -57,10 +57,21 @@ pub trait LogProcessor: Send + Sync + Debug {
...
     fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult {
+        // It would have been better to make this method required, but that ship
+        // sailed when the logs API was declared stable.
+        otel_warn!(
+            name: "LogProcessor.DefaultShutdownWithTimeout",
+            message = format!("LogProcessor::shutdown_with_timeout should be implemented by all LogProcessor types")
```
cijothomas, line 67: This is giving the wrong message - it is not required by all implementations to have shutdown methods.

PR #3134, opened 2025-08-21
feat: implement observable histograms

thomaseizinger:
## Changes

Adds observable histograms as a metric, analogous to the other observable metrics.

### DIFF
--- a/opentelemetry-sdk/src/metrics/instrument.rs
+++ b/opentelemetry-sdk/src/metrics/instrument.rs
@@ -26,6 +26,8 @@ pub enum InstrumentKind {
     /// A group of instruments that record a distribution of values synchronously with
     /// the code path they are measuring.
     Histogram,
+    /// A group of instruments that record a distribution of values in an asynchronously callback.
+    ObservableHistogram,
     /// A group of instruments that record increasing values in an asynchronous
     /// callback.
     ObservableCounter,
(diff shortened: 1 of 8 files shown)


thomaseizinger:
Hmm that is a shame, it seems kind of obvious that this one is missing.

My usecase is to observe the depths of queues over time. I'll see if a gauge is sufficient for my usecase. I did want to see the load pattern on it, like how often it is almost full or almost empty and not just the current value.

Is there a way to make an observable metric that operates on its own timer and not just when it gets exported? I'd like a more fine-granular recording interval than the exporting interval.

cijothomas:
> Hmm that is a shame, it seems kind of obvious that this one is missing.
> 
> My usecase is to observe the depths of queues over time. I'll see if a gauge is sufficient for my usecase. I did want to see the load pattern on it, like how often it is almost full or almost empty.

Queue sizes are often tracked via ObservableGauge or ObservableUpDownCounter.

> Is there a way to make an observable metric that operates on its own timer and not just when it gets exported? I'd like a more fine-granular recording interval than the exporting interval.

No. You have to create the timer mechanisms yourself, and then report Gauge/histogram when the timer is triggered. OTel's observable instruments are only invoked at export interval - there were discussions around keeping them separate, but those are not in the specification yet.

Please explore if either of the above helps your needs.

thomaseizinger:
> Queue sizes are often tracked via ObservableGauge or ObservableUpDownCounter.

Thanks, I'll try to use that. While we are at the topic: Is there a standard metric name for queue sizes? I couldn't find anything.

PR #3162, opened 2025-09-10
fix: Fix stack-overflow inducing infinite feedback loop when calling `tracing::event!` after `SdkLoggerProvider::shutdown`

### LINKED ISSUE #3161: [Bug]: Stack overflow when calling unfiltered `tracing::event!` after `SdkLogProvider` is shutdown.
### What happened?

When integrating the opentelemetry crate with the `tracing_subscriber` crate, calling `tracing::event!` and its derivates after the `SdkLoggerProvider` has been called leads to a stack overflow due to `BatchLogProcessor::emit` using `otel_warn!` when trying to report the `mpsc::TrySendError:Disconnected` error.

It seems that under the hood, `otel_warn!` may call `tracing::warn!`, which will cause the warning reporting the BatchLogProcessor being shutdown to get fed back into the `BatchLogProcessor` due to the tracing subscriber being none the wiser, which in turn causes the same warning to be emitted again, leading to infinite recursion.

I have created a PR that fixes this issue by simply not calling `otel_warn!` in this edge case if the log record itself is emitted by the same `otel_warn!`. I will publish this PR after creating this issue.

0.30.0

0.30.0

N/A

### Relevant log output

```shell
thread 'test_no_stack_overflow_when_event_is_emitted_after_shutdown' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

### Tip

lucamuscat:
Fixes #3161

## Changes

When integrating the opentelemetry crate with the tracing_subscriber crate, calling tracing::event! and its derivates after the SdkLoggerProvider has been called leads to a stack overflow due to BatchLogProcessor::emit using otel_warn! when trying to report the mpsc::TrySendError:Disconnected error.

This PR prevents `otel_warn!` from being called when handling the `mpsc::TrySendError:Disconnected` error in `BatchLogProcessor::emit` if the name of the log record is the same as that emitted by the `otel_warn!` in the same branch.

### DIFF
--- a/opentelemetry-sdk/src/logs/logger.rs
+++ b/opentelemetry-sdk/src/logs/logger.rs
@@ -33,7 +33,7 @@ impl opentelemetry::logs::Logger for SdkLogger {
 
     /// Emit a `LogRecord`.
     fn emit(&self, mut record: Self::LogRecord) {
-        if Context::is_current_telemetry_suppressed() {
+        if Context::is_current_telemetry_suppressed() || self.provider.is_shutdown() {
             return;
         }
         let provider = &self.provider;
--- a/opentelemetry-appender-tracing/tests/test_no_stack_overflow_after_shutdown.rs
+++ b/opentelemetry-appender-tracing/tests/test_no_stack_overflow_after_shutdown.rs
@@ -1,0 +1,20 @@
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_sdk::logs::SdkLoggerProvider;
+use tracing::info;
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+#[test]
+fn test_no_stack_overflow_when_event_is_emitted_after_shutdown() {
+    let exporter = opentelemetry_stdout::LogExporter::default();
+    let provider: SdkLoggerProvider = SdkLoggerProvider::builder()
+        .with_batch_exporter(exporter)
+        .build();
+
+    let otel_layer = layer::OpenTelemetryTracingBridge::new(&provider);
+
+    tracing_subscriber::registry().with(otel_layer).init();
+
+    provider.shutdown().unwrap();
+
+    info!("Don't crash")
+}
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -129,6 +129,10 @@ impl SdkLoggerProvider {
     pub fn shutdown(&self) -> OTelSdkResult {
         self.shutdown_with_timeout(Duration::from_secs(5))
     }
+
+    pub(crate) fn is_shutdown(&self) -> bool {
+        self.inner.is_shutdown.load(Ordering::SeqCst)
+    }
 }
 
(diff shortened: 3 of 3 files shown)


lucamuscat:
I am aware that this is likely not the most optimal solution, as the bug ultimately lies in the integration between the otel_sdk and the tracing subscriber.

An alternative would be to add a public method `is_shutdown` which returns whether the LogProvider is shutdown (or in the process of being shutdown), which will ultimately be used by the tracing layer to discard any logs produced after shutdown. Of course, this alternative would require adding new public APIs and would require more changes in general.

opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -207,12 +207,20 @@ impl LogProcessor for BatchLogProcessor {
...
-                otel_warn!(
-                    name: "BatchLogProcessor.Emit.AfterShutdown",
-                    message = "Logs are being emitted even after Shutdown. This indicates incorrect lifecycle management of OTelLoggerProvider in application. Logs will not be exported."
-                );
+                // Re-emitting the following warning here may lead to a stack overflow due to the
+                // warning produced by this branch being reconsumed by this same branch, producing a
+                // cycle. This may occur when integrating with the `tracing` crate due to
+                // `otel_warn!` possibly calling `tracing::warn!` under the hood, which will emit
+                // the warning back into the otel log processor.
+                if matches!((record.event_name, record.target()), (Some(event_name), Some(target_name)) if !(event_name == Self::AFTER_SHUTDOWN_WARNING_NAME && target_name == env!("CARGO_CRATE_NAME")))
```
cijothomas: don't think we can hardcode our way out. We need to find why Context based suppression is not working and fix that.

lucamuscat: Thanks for the pointer, it seems that unlike SdkTracer, SdkLogger does not discard logs when the provider is shutdown. The PR has been adjusted accordingly

opentelemetry-sdk/src/logs/logger.rs
```diff
@@ -33,7 +33,7 @@ impl opentelemetry::logs::Logger for SdkLogger {
 
     /// Emit a `LogRecord`.
     fn emit(&self, mut record: Self::LogRecord) {
-        if Context::is_current_telemetry_suppressed() {
+        if Context::is_current_telemetry_suppressed() || self.provider.is_shutdown() {
```
lucamuscat, line 36: It also suppresses the warning logs emitted when a log is sent after the LoggerProvider is shutdown, effectively hiding the log.

lucamuscat:
@lalitb @cijothomas The PR is ready for you guys to have another look.

@cijothomas Is this what you had in mind when you mentioned "Context based suppression"?

Do note that since logs sent to a BatchLogProcessor with a shutdown provider are no longer propagated to the underlying loggers, the original log warning of a log being emitted after shutdown is no longer emitted. Is this an acceptable trade off for you guys?

lucamuscat:
Closing in favour of #3172, which follows the maintainer's suggestions of fixing context suppression for `BatchLogProcessor::emit`

PR #3172, opened 2025-09-17
fix: Suppress telemetry emitted inside of `BatchLogProcessor::emit`

### LINKED ISSUE #3161: [Bug]: Stack overflow when calling unfiltered `tracing::event!` after `SdkLogProvider` is shutdown.

lucamuscat:
Fixes #3161

## Changes
The `BatchLogProcessor` background thread itself runs with a telemetry suppressed context, however, `BatchLogProcessor::emit` does not use a telemetry suppressed context as it is executed in the caller's thread, as opposed to the suppressed background thread.

Using a telemetry suppressed context whilst calling `BatchLogProcessor::emit` fixes the issue where telemetry-induced-telemetry is generated whilst calling `emit` on a shutdown `BatchLogProcessor`, avoiding a stack overflow.

### DIFF
--- a/opentelemetry-appender-tracing/tests/test_no_stack_overflow_after_shutdown.rs
+++ b/opentelemetry-appender-tracing/tests/test_no_stack_overflow_after_shutdown.rs
@@ -1,0 +1,23 @@
+use opentelemetry_appender_tracing::layer;
+use opentelemetry_sdk::logs::SdkLoggerProvider;
+use tracing::info;
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+#[test]
+fn test_logging_after_shutdown_does_not_cause_telemetry_induced_telemetry() {
+    //! Reproduces [#3161](https://github.com/open-telemetry/opentelemetry-rust/issues/3161)
+    let exporter = opentelemetry_stdout::LogExporter::default();
+    let provider: SdkLoggerProvider = SdkLoggerProvider::builder()
+        .with_batch_exporter(exporter)
+        .build();
+
+    let otel_layer = layer::OpenTelemetryTracingBridge::new(&provider);
+
+    tracing_subscriber::registry().with(otel_layer).init();
+
+    provider.shutdown().unwrap();
+
+    // If logging causes telemetry-induced-telemetry after shutting down the provider, then a stack
+    // overflow may occur.
+    info!("Don't crash")
+}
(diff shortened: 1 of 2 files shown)


lalitb (changes requested):
as discussed here - https://github.com/open-telemetry/opentelemetry-rust/pull/3172#issuecomment-3305182555

Better to have perf numbers for the change.

lucamuscat:
> as discussed here - [#3172 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/3172#issuecomment-3305182555)
> 
> Better to have perf numbers for the change.

I had initially prevented the call to `otel_warn!` from causing the infinite feedback loop by checking if the warning about to be emitted using `otel_warn!` is the same as the log record in the process of being emitted, however, I was told by a maintainer that we cannot hardcode our way out of the bug.

If you wish, I may move the context suppression in the same branch as the `otel_warn!` causing the stack overflow to keep the context suppression out of the hot loop. I don't think that you would mind if an error case had slightly more overhead.

bantonsson:
@lucamuscat I think that moving the suppression around the offending `otel_warn!` sounds like a good strategy.

lucamuscat:
Telemetry suppression has been moved out of the happy path and right before the offending `otel_warn!`. The happy path no longer bears the overhead of telemetry suppression.

@lalitb

cijothomas:
> @lucamuscat I think that moving the suppression around the offending `otel_warn!` sounds like a good strategy.

+1. That was the plan for other places hitting (potentially hitting) this issue.
https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-logs/src/logs/processor.rs#L53-L54

cijothomas (approved):
Thanks!

lalitb (approved):

PR #3180, opened 2025-09-25
build: replicate tonic tls features

jdrouet:
## Changes

`ring` is not maintained anymore and could lead to compilation issues. We need a way to use the tonic exporter with https without relying on `ring`. This PR replicates the way `tonic` handles TLS by separating the `ring` feature from the others.
This PR also introduces a `tls-aws-lc` feature so that users can rely on it.

### DIFF
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -80,11 +80,14 @@ experimental-grpc-retry = ["grpc-tonic", "opentelemetry_sdk/experimental_async_r
 # http compression
 gzip-http = ["flate2"]
 zstd-http = ["zstd"]
-tls = []
-tls-aws-lc = ["tls", "tonic/tls-aws-lc"]
-tls-ring = ["tls", "tonic/tls-ring"]
-tls-roots = ["tls", "tonic/tls-native-roots"]
-tls-webpki-roots = ["tls", "tonic/tls-webpki-roots"]
+
+# tls
+_tls-any = []  # Internal feature for TLS guards
+tls = ["tls-ring"]
+tls-aws-lc = ["_tls-any", "tonic/tls-aws-lc"]
+tls-ring = ["_tls-any", "tonic/tls-ring"]
+tls-roots = ["_tls-any", "tonic/tls-native-roots"]
+tls-webpki-roots = ["_tls-any", "tonic/tls-webpki-roots"]
 
 # http binary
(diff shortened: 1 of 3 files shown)


opentelemetry-otlp/Cargo.toml
```diff
@@ -76,7 +76,9 @@ zstd-tonic = ["tonic/zstd"]
 # http compression
 gzip-http = ["flate2"]
 zstd-http = ["zstd"]
-tls = ["tonic/tls-ring"]
+tls = []
```
scottgerring: The tonic [feature flags impl referenced](https://github.com/hyperium/tonic/blob/54f0139f655b3f2f7490de81d2f4d30ea1a0b370/tonic/Cargo.toml#L28-L30):

```
_tls-any = ["dep:tokio", "tokio?/rt", "tokio?/macros", "tls-connect-info"] # Internal. Please choose one of `tls-ring` or `tls-aws-lc`
tls-ring = ["_tls-any", "tokio-rustls/ring"]
tls-aws-lc = ["_tls-any", "tokio-rustls/aws-lc-rs"]
```

I believe the intention is that "_tls-any" is an "internal" feature, not something the user is intended to rely on, and is what's then used to guard the backend-independent TLS code.

With this in mind, and with an eye to maintaining backward compatibility, how about: 

```
_tls-any = []  # Internal feature for TLS guards
tls-ring = ["_tls-any", "tonic/tls-ring"]    # User opts into ring
tls-aws-lc = ["_tls-any", "tonic/tls-aws-lc"] # User opts into aws-lc
tls = ["tls-ring"]  # Default to existing behaviour 
```

This would need code changes too to move from `tls` to `_tls-any` in the feature guards - e.g. a bunch of stuff in exporter/tonic/mod: 

 https://github.com/open-telemetry/opentelemetry-rust/blob/51eac97243cebe75ef6eed67e7c66dde5a2afaf4/opentelemetry-otlp/src/exporter/tonic/mod.rs#L11-L12

scottgerring: As this suggestion stretches the scope a little here, it would be great if one of the others piped in -  @lalitb @bantonsson , does this seem reasonable to you?

jdrouet: 👋 Sorry for the late reply,
I've addressed your feedback 😉

GabrielBianconi:
We'd love support for `aws-lc-rs` - thanks!

PR #3183, opened 2025-09-29
chore: avoid alloc or fmt when possible

tisonkun:
TRIVIAL AS IS.


tisonkun:
@lalitb @cijothomas shall we merge this patch?

PR #3198, opened 2025-10-14
feat: add get_ref to retrieve Resource values by reference

### LINKED ISSUE #3163: `Resource::get` always cloning the returned `Value`
# Issue

I noticed the the `opentelemetry_sdk::Resource::get(&self, &Key) -> Value` needs to returned an owned `Value`, and thus always clones it.

The operation is not that expensive when the underlying value is a number, of a static/Arc String. But it does require and allocation for reading values that are owned string, and array values.
This is a bit annoying for instance:
* when writing span processors that need to check the presence of a field on the Resource frequently without storing them. 
* when writing span exporters that just need a reference to the data for serialization

# Proposal

I think a more flexible API would be to return a reference and modify the signature to `Resource::get(&self, &Key) -> &Value`.
Although this is a breaking change, existing users of that API would just do `get(key).cloned()` instead if they really need to have a owned value.

Another option if we don't want to break the API is to keep the old `get` as before and add a `get_ref` method.

leghadjeu-christian:
Fixes #3163

## Changes

Adds a new `get_ref` method to `opentelemetry_sdk::Resource` that returns a reference to the stored value instead of cloning it. This improves performance and avoids unnecessary allocations for read-only access, while keeping the existing `get` method for backward compatibility.

### DIFF
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -428,7 +428,7 @@ mod tests {
             assert_eq!(
                 processor
                     .resource()
-                    .get_ref(&Key::from_static_str(resource_key))
+                    .get(&Key::from_static_str(resource_key))
                     .map(|v| v.to_string()),
                 expect.map(|s| s.to_string())
             );
@@ -436,7 +436,7 @@ mod tests {
             assert_eq!(
                 exporter
                     .resource()
-                    .get_ref(&Key::from_static_str(resource_key))
+                    .get(&Key::from_static_str(resource_key))
                     .map(|v| v.to_string()),
                 expect.map(|s| s.to_string())
             );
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/resource/mod.rs
```diff
@@ -228,6 +228,11 @@ impl Resource {
     pub fn get(&self, key: &Key) -> Option<Value> {
         self.inner.attrs.get(key).cloned()
     }
+
+    /// Returns a reference to the value for the given key without cloning.
```
scottgerring:
```suggestion
    /// Returns a reference to the value for the resource associated with the given key without cloning.
```

leghadjeu-christian: Thanks for catching that @scottgerring 👍

scottgerring (approved):
Thanks - LGTM!

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -424,44 +424,44 @@ mod tests {
             assert_eq!(
                 processor
                     .resource()
-                    .get(&Key::from_static_str(resource_key))
+                    .get_ref(&Key::from_static_str(resource_key))
```
cijothomas: I have requested for some additional information on the need of this new method [here](https://github.com/open-telemetry/opentelemetry-rust/issues/3163#issuecomment-3424254764).

Irrespective of that, I am fine adding the new method. One comment - unless really needed, let's keep all existing usage unmodified in this PR. If needed, I'd prefer to have a follow up PR updating all usages to the new method. This helps keep PRs short and easy to see that this is purely additive change.

PR #3200, opened 2025-10-20
fix: add timestame for log appender

songfei:
Fixes #
The collector cannot display the log correctly because there is no timestamp in the record.

## Changes

Add a timestamp to the record

* [x ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed

### DIFF
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -151,6 +151,7 @@ where
             log_record.set_severity_number(severity_of_level(record.level()));
             log_record.set_severity_text(record.level().as_str());
             log_record.set_body(AnyValue::from(record.args().to_string()));
+            log_record.set_timestamp(SystemTime::now());
 
             #[cfg(feature = "experimental_metadata_attributes")]
             {


opentelemetry-appender-log/src/lib.rs
```diff
@@ -151,6 +151,7 @@ where
             log_record.set_severity_number(severity_of_level(record.level()));
             log_record.set_severity_text(record.level().as_str());
             log_record.set_body(AnyValue::from(record.args().to_string()));
+            log_record.set_timestamp(SystemTime::now());
```
lalitb, line 154: As per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-timestamp), this field should come from source (in this case `log` library), and should be empty otherwise.

cijothomas: +1. OTel Sdk sets ObservedTimeStamp only. If you need to have timestamp also set, you can write a `LogProcessor` that can set it (either from now(), or to same value as ObservedTimeStamp)

songfei: Thank you very much for your reply, I understand. 
It seems that the collector I use does not follow the protocol specifications.

PR #3204, opened 2025-10-21
chore: OTLP retry post-review cleanup

scottgerring:
Fixes #
Follow on from #3126 

## Changes

### DIFF
--- a/opentelemetry-otlp/src/retry.rs
+++ b/opentelemetry-otlp/src/retry.rs
@@ -11,7 +11,7 @@
     feature = "experimental-grpc-retry",
     feature = "experimental-http-retry"
 ))]
-use opentelemetry::otel_info;
+use opentelemetry::{otel_debug, otel_info};
 
 #[cfg(any(
     feature = "experimental-grpc-retry",
@@ -138,7 +138,7 @@ where
                         // Use exponential backoff with jitter
                         let jitter = generate_jitter(policy.jitter_ms);
                         let delay_with_jitter = std::cmp::min(delay + jitter, policy.max_delay_ms);
-                        otel_info!(name: "Export.InProgress.Retrying",
+                        otel_debug!(name: "Export.InProgress.Retrying",
                             operation = operation_name,
                             attempt = attempt,
                             delay_ms = delay_with_jitter,


scottgerring:
Open for suggestions for where we can put the runtime selection. This strikes me as a broader question about the stabilisation of the runtime mechanism. 

https://github.com/scottgerring/opentelemetry-rust/blob/9ba3a06df952d4888009764bebe665281f9e45b7/opentelemetry-otlp/src/exporter/http/mod.rs#L402-L409

opentelemetry-otlp/src/retry.rs
```diff
@@ -144,13 +128,13 @@ where
 
                 match error_type {
                     RetryErrorType::NonRetryable => {
-                        otel_warn!(name: "OtlpRetry", message = format!("Operation {:?} failed with non-retryable error: {:?}", operation_name, err));
+                        otel_warn!(name: "OtlpRetryNonRetryable", operation = operation_name, error = format!("{:?}", err));
```
cijothomas: operation here would always be "export", so this is not a quite useful. I guess you meant to use something like signal=logs/traces/metrics, transport=http/grpc ?

opentelemetry-otlp/src/retry.rs
```diff
@@ -144,13 +128,13 @@ where
...
                         return Err(err);
                     }
                     RetryErrorType::Retryable if attempt < policy.max_retries => {
                         attempt += 1;
                         // Use exponential backoff with jitter
-                        otel_warn!(name: "OtlpRetry", message = format!("Retrying operation {:?} due to retryable error: {:?}", operation_name, err));
+                        otel_info!(name: "OtlpRetryRetrying", operation = operation_name, error = format!("{:?}", err));
```
cijothomas: the `error` here is going to be the stringified/formatted one, likely coming from the underlying transport library, partially defeating structured logging purpose.

cijothomas: also, this is a retry attempt  and need to log the delay/jitter etc. to be fully useful.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Released 2025-Sep-25
 - Update `opentelemetry-proto` and `opentelemetry-http` dependency version to 0.31.0
 - Add HTTP compression support with `gzip-http` and `zstd-http` feature flags
 - Add retry with exponential backoff and throttling support for HTTP and gRPC exporters
+  This behaviour is opt in via the `experimental-grpc-retry` and `experimental-http-retry flags` on this crate.
```
cijothomas:
```suggestion
  This behaviour is opt in via the `experimental-grpc-retry` and `experimental-http-retry` flags on this crate.
```

cijothomas: also mention that there is no ability to customize retry policy.

opentelemetry-otlp/src/retry.rs
cijothomas: otel_warn!(
                            name: "Export.Failed.NonRetryable",
                            signal: "Logs",
                            transport: "HTTP",
                            message = "OTLP export failed with non-retryable error - telemetry data will be lost"
                        );

^ Suggestion. We should probably include minimum info like erro_code too, but detailed error message should be done at an inner level, not by the retry module.

scottgerring: The operation includes both the client type and signal, e.g.

https://github.com/open-telemetry/opentelemetry-rust/blob/16d08637029f6f2cc42f554a67c2f8f03c6f0020/opentelemetry-otlp/src/exporter/http/logs.rs#L8-L12

https://github.com/open-telemetry/opentelemetry-rust/blob/16d08637029f6f2cc42f554a67c2f8f03c6f0020/opentelemetry-otlp/src/exporter/tonic/trace.rs#L78-L85

We could plumb the things through seperately as `&str` to structure this, but for logging this feels a bit overwrought imho

opentelemetry-otlp/CHANGELOG.md
scottgerring: Fixed code formatting. 
You _can_ customize the retry policy, e.g. something like this:

```rust
 use opentelemetry_otlp::{SpanExporter, RetryPolicy};
  use opentelemetry_otlp::WithTonicConfig; // or WithHttpConfig

  let custom_policy = RetryPolicy {
      max_retries: 5,
      initial_delay_ms: 200,
      max_delay_ms: 3200,
      jitter_ms: 50,
  };

      .with_tonic()  // or .with_http()
      .with_retry_policy(custom_policy)
      .build()?;
```

Updated changelog accordingly.

opentelemetry-otlp/src/retry.rs
scottgerring: I've taken the `name` style from here, and left the signal/transport out, favouring `operation`, based on my comment above (it tells you this stuff already). 

I am leery of adding `error_code` here, as it's generally logged anyway and it would make the interfaces around retry more complicated just for extra logging, but it sounds like you agree with that take ("detailed error message should be done at an inner level") ?

opentelemetry-otlp/src/retry.rs
scottgerring: > the error here is going to be the stringified/formatted one, likely coming from the underlying transport library, partially defeating structured logging purpose.

Do you have a suggestion here?

cijothomas: > > the error here is going to be the stringified/formatted one, likely coming from the underlying transport library, partially defeating structured logging purpose.
> 
> Do you have a suggestion here?

Yes. Don't log the inner error in the retry module.

opentelemetry-otlp/src/retry.rs
cijothomas: got it. Yes `operation` sounds good enough.

opentelemetry-otlp/src/retry.rs
```diff
@@ -144,15 +128,15 @@ where
 
                 match error_type {
                     RetryErrorType::NonRetryable => {
-                        otel_warn!(name: "OtlpRetry", message = format!("Operation {:?} failed with non-retryable error: {:?}", operation_name, err));
+                        otel_warn!(name: "Export.Failed.NonRetryable", operation = operation_name, error = format!("{:?}", err));
```
cijothomas:
```rust
otel_warn!(
                            name: "Export.Failed.NonRetryable",
                            operation = operation_name,
                            message = "OTLP export failed with non-retryable error - telemetry data will be lost"
                        );
```

opentelemetry-otlp/src/retry.rs
```diff
@@ -161,13 +145,13 @@ where
...
                         // Use server-specified delay (overrides exponential backoff)
-                        otel_warn!(name: "OtlpRetry", message = format!("Retrying operation {:?} after server-specified throttling delay: {:?}", operation_name, server_delay));
+                        otel_info!(name: "Export.InProgress.Throttled", operation = operation_name, error = format!("{:?}", err), attempt = attempt, delay_ms = server_delay.as_millis());
                         runtime.delay(server_delay).await;
                         // Don't update exponential backoff delay for next attempt since server provided specific timing
                     }
                     _ => {
                         // Max retries reached
-                        otel_warn!(name: "OtlpRetry", message = format!("Operation {:?} failed after {} attempts: {:?}", operation_name, attempt, err));
+                        otel_warn!(name: "Export.Failed.Exhausted", operation = operation_name, error = format!("{:?}", err), retries = attempt);
```
cijothomas: otel_warn!(
                        name: "Export.Retry.Exhausted",
                        operation = operation_name,
                        retries = attempt,
                        message = "OTLP export failed after all retry attempts - telemetry data will be lost"
                    );

opentelemetry-otlp/src/retry.rs
cijothomas: operation sounds good here. thanks for clarifying.

opentelemetry-otlp/src/retry.rs
scottgerring: Should be good; removed the error inline from each of these.

opentelemetry-otlp/src/retry.rs
scottgerring: Done

bantonsson (approved):

opentelemetry-otlp/src/retry.rs
```diff
@@ -144,15 +128,23 @@ where
...
+                            message = "OTLP export failed with non-retryable error - telemetry data will be lost");
                         return Err(err);
                     }
                     RetryErrorType::Retryable if attempt < policy.max_retries => {
                         attempt += 1;
                         // Use exponential backoff with jitter
-                        otel_warn!(name: "OtlpRetry", message = format!("Retrying operation {:?} due to retryable error: {:?}", operation_name, err));
                         let jitter = generate_jitter(policy.jitter_ms);
                         let delay_with_jitter = std::cmp::min(delay + jitter, policy.max_delay_ms);
+                        otel_info!(name: "Export.InProgress.Retrying",
```
cijothomas: this could be too much... lets use debug instead of info?

cijothomas (approved):
LGTM- left a comment on lowering severity for retry attempts.

PR #3206, opened 2025-10-21
chore(otlp): Handle partial success

### LINKED ISSUE #865: Handle partial success responses from OTLP export services
### Problem Statement

Following the [specification change to add partial success responses](https://github.com/open-telemetry/opentelemetry-specification/pull/2696), each OTel SDK is encouraged to handle the resulting error message string in an appropriate way, considering existing norms. This requires OTLP v0.19.

### Proposed Solution

For each of the `Export*ServiceRequest` methods used by OTLP Trace and Metrics exporters, construct an error and call the appropriate handler with the error message string and the number of spans/points(/logs) dropped. Future OTel specifications may call for optional treatment of the number of dropped items, but presently that is just additional information to include in the handled error.

scottgerring:
Fixes #865, handling partial success for HTTP and tonic OTLP responses by logging the portion of the request that was rejected by the server and not retrying. 

The relevant bit of the spec is [here](https://github.com/open-telemetry/opentelemetry-rust/issues/865#issuecomment-3424744339). 

## Changes

Straightforward enough! The testing is rather limited because all we're doing is parsing and logging responses; let's make sure we don't panic for invalid or empty data, as well as a minimal smoke test. 

### DIFF
--- a/opentelemetry-otlp/src/retry_classification.rs
+++ b/opentelemetry-otlp/src/retry_classification.rs
@@ -163,13 +163,12 @@ pub mod grpc {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-    use std::time::Duration;
-
     // Tests for HTTP error classification
+    #[cfg(feature = "experimental-http-retry")]
     mod http_tests {
-        use super::*;
+        use crate::retry::RetryErrorType;
         use crate::retry_classification::http::*;
+        use std::time::Duration;
 
         #[test]
         fn test_http_429_with_retry_after_seconds() {


opentelemetry-otlp/tests/smoke.rs
```diff
@@ -136,3 +136,107 @@ async fn smoke_tracer() {
     let first_event = first_span.events.first().unwrap();
     assert_eq!("my-test-event", first_event.name);
 }
+
+// Test for partial success response handling
```
cijothomas: not sure if we are actually validating anything in this test about partial-success.

scottgerring: @cijothomas 
It's just to make sure things don't catch fire, but you are right, there is not much in the way of asserts - apart from it doesn't panic, and spans come out the end.

I changed the comment to reflect this, but if your feeling is it is overwrought it may well be and I can happily remove it. The use case here in general is straightforward enough - de-serialize and log partial success - that I don't think this adds much beyond the 'fail path' unit tests tbh.

PR #3211, opened 2025-10-24
propagate trace processor errors to users instead of internal logging

### LINKED ISSUE #3210: Library crates should propagate errors instead of silently logging them
### Problem

The OpenTelemetry Rust SDK currently logs errors on behalf of applications, which is inappropriate for library crates. While [ADR-001](docs/adr/001_error_handling.md) provides error handling guidance, it includes a problematic allowance:

> Failures during regular operation should not panic, instead returning errors to the caller where appropriate, **_or_ logging an error if not appropriate**.

**This guidance needs to be updated.** For library crates, it is _never_ appropriate to log errors. Per [CONTRIBUTING.md](CONTRIBUTING.md), the SDK should either return errors to callers or delegate to a global error handler registered by the application. However, many codepaths are logging directly instead, leaving applications unable to respond to failures.

**Example from `span_processor.rs`:**

```rust
fn on_end(&self, span: SpanData) {
    let result = self.exporter.lock().map(|mut exporter| {
        exporter.export(vec![span])
    });

    if let Err(err) = result {
        otel_error!(
            name: "BatchSpanProcessor.Export.Error",
            error = format!("{:?}", err)
        );
    }
}
```

**Why this is problematic:**

Library crates logging on behalf of applications creates several problems:

1. **No error visibility**: Applications cannot detect, count, or respond to failures
2. **No integration**: Errors cannot be integrated with the application's monitoring, alerting, or metrics systems
3. **Inconsistent formatting**: Library logs don't match the application's logging format, style, or context (request IDs, etc.), causing confusion for operators and breaking log ingestion pipelines
4. **Policy violations**: The library makes policy decisions (what to log, when, how) that belong to the application

Standard Rust library crates (`std`, `tokio`, `serde`, etc.) return errors and let applications decide how to handle them. OpenTelemetry Rust should follow the same pattern.

### Proposed Solution

**For synchronous operations with a direct caller:**

- Return `OTelSdkResult` or appropriate error types defined in `opentelemetry-sdk::error`
- Let callers decide whether to log, retry, or propagate errors
- Aligns with existing `SpanExporter`, `LogExporter`, and `PushMetricExporter` traits which already return `OTelSdkResult`

**For background/asynchronous operations without a direct caller:**

- Implement an **error callback mechanism** via `with_error_handler()` on processor builders
- The callback is invoked when background tasks fail
- Users can then log, emit metrics, trigger alerts, or implement custom strategies

#### Traces (High Priority)

- [x] `opentelemetry-sdk/src/trace/span_processor.rs` - Remove error logging in batch/simple processors
- [x] `opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs` - Add error callback for background exports
- [x] `opentelemetry-sdk/src/trace/provider.rs` - Remove redundant error logging in shutdown

#### Metrics (High Priority)

- [ ] `opentelemetry/src/metrics/instruments.rs` - `InstrumentProvider` trait methods should return `Result`
- [ ] `opentelemetry-sdk/src/metrics/meter.rs` - Return errors instead of logging and creating no-op instruments
- [ ] `opentelemetry-sdk/src/metrics/meter_provider.rs` - Propagate shutdown errors per ADR-001 patterns
- [ ] Periodic reader implementations - Expose background export errors via error callback

#### Logs (High Priority)

- [ ] `opentelemetry-sdk/src/logs/log_processor.rs` - Make `LogProcessor::emit()` fallible
- [ ] `opentelemetry-sdk/src/logs/simple_log_processor.rs` - Return errors from emit operations
- [ ] `opentelemetry-sdk/src/logs/log_processor_with_async_runtime.rs` - Add error callback for background processing

#### Other

- [ ] `opentelemetry-zipkin/src/exporter/env.rs` - Replace `eprintln!` with proper error returns
- [ ] Update examples to demonstrate proper error handling
- [ ] Update tests to verify error propagation

### Implementation Strategy

1. **Phase 1: Traces** ✅
   - Remove all `otel_error!`, `otel_warn!`, `otel_debug!` calls that mask export failures
   - Add `with_error_handler()` to `BatchSpanProcessorBuilder`
   - Background export errors invoke user-provided callback
   - Synchronous operations return `OTelSdkResult`

2. **Phase 2: Metrics**
   - Update trait definitions to return `Result` types per ADR-001 guidance
   - Implement error callbacks for periodic readers
   - Update meter implementation to propagate errors from instrument creation

3. **Phase 3: Logs**
   - Make `LogProcessor::emit()` fallible where appropriate
   - Add error callbacks for async log processors
   - Update log appenders to propagate errors

4. **Phase 4: Documentation & Examples**
   - **Update ADR-001** to remove the allowance for logging errors in library crates
   - Update all examples to demonstrate proper error handling
   - Add migration guide documenting breaking changes
   - Document error callback patterns and best practices

### Backward Compatibility

This is a **breaking change** that will require:

- Minor version bump (0.x -> 0.y, as the crate is pre-1.0)
- Migration guide for users updating from previous versions
- Updated examples and documentation
- **Update to ADR-001** clarifying that library crates must never log errors on behalf of applications

However, the benefits justify the breaking change:

- Proper library design following Rust best practices and standard library conventions
- Better error visibility and control for applications
- Enables custom error handling strategies (retry, metrics, alerting)
- Improved debuggability and observability in production

### Additional Context

**Why ADR-001 allows logging:**

The allowance for logging "where errors cannot be returned" likely stems from background operations where there's no direct caller. However, the solution is **not** to log, but rather to:

- Use error callbacks that applications can register
- Delegate to a global error handler if one is registered
- Return errors wherever possible

**Technical feasibility:**

- The OpenTelemetry specification requires operations like `on_end()` to be fast and non-blocking, but **does not mandate void return types**
- Returning an error does not violate the non-blocking requirement
- Error callbacks provide a way to handle background failures without blocking the hot path
- This change aligns with the existing exporter traits (`SpanExporter`, `LogExporter`, `PushMetricExporter`) which already return `OTelSdkResult` from their methods

---

**Related:** This aligns with Rust's error handling best practices, the Rust standard library's patterns, and the principle that libraries should be "honest" about failures, letting applications make all policy decisions about logging, retrying, and error handling.

svencowart:
Remove internal logging of errors in span processors and tracer provider in favor of returning errors to callers and providing an error handler callback mechanism for background operations.

## Changes

- Add error_handler callback to BatchSpanProcessor for background errors (span drops, export failures during timer-based flushes)
- Return errors from shutdown_with_timeout when spans were dropped
- Remove otel_debug/otel_warn/otel_error calls from span processors
- Simplify export error handling by propagating instead of logging
- Update BatchSpanProcessor builder with with_error_handler() method
- Change dropped_spans_count to Arc<AtomicUsize> for shared ownership
- Remove shutdown error logging from TracerProviderInner

This gives users explicit control over error handling rather than relying on internal SDK logging that may be missed or not actionable.

Fixes #3210 


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -151,6 +151,12 @@ impl<T: SpanExporter> SpanProcessor for SimpleSpanProcessor<T> {
                 reason = format!("{:?}", err)
             );
         }
+
+        // TODO: on_end() currently has a void return type, so errors are silently discarded.
+        // This should be changed to return Result so users can be notified of export failures.
+        // The OpenTelemetry spec requires on_end to be fast/non-blocking, but doesn't mandate
+        // a void return - returning an error doesn't violate that requirement.
+        // Until the trait is changed, we cannot return this error to the caller.
```
lalitb, line 159: Most spans end via `Drop`, which can't propagate errors. We should use the similar pattern as for batch processor here too.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -290,18 +295,37 @@ pub struct BatchSpanProcessor {
     export_span_message_sent: Arc<AtomicBool>,
     current_batch_size: Arc<AtomicUsize>,
     max_export_batch_size: usize,
-    dropped_spans_count: AtomicUsize,
+    dropped_spans_count: Arc<AtomicUsize>,
```
lalitb, line 298: Does it need to be Arc?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -576,18 +602,38 @@ impl SpanProcessor for BatchSpanProcessor {
...
+                let previous_count = self.dropped_spans_count.fetch_add(1, Ordering::Relaxed);
+                if previous_count == 0 {
+                    otel_warn!(
+                        name: "BatchSpanProcessor.OnEnd.AfterShutdown",
+                        message = "Spans are being emitted even after Shutdown. This indicates incorrect lifecycle management of TracerProvider in application. Spans will not be exported."
+                    );
+                }
+
+                // Also invoke error handler if registered
+                if let Some(ref handler) = self.error_handler {
```
cijothomas, line 631: what are some possible actions a user can do in this error handler? it does not have much information to really do anything about, except perhaps logging it?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -632,14 +678,21 @@ impl SpanProcessor for BatchSpanProcessor {
...
         let dropped_spans = self.dropped_spans_count.load(Ordering::Relaxed);
-        let max_queue_size = self.max_queue_size;
         if dropped_spans > 0 {
+            // Log warning for observability (always happens)
             otel_warn!(
-                name: "BatchSpanProcessor.SpansDropped",
-                dropped_span_count = dropped_spans,
-                max_queue_size = max_queue_size,
-                message = "Spans were dropped due to a queue being full. The count represents the total count of spans dropped in the lifetime of this BatchSpanProcessor. Consider increasing the queue size and/or decrease delay between intervals."
+                name: "BatchSpanProcessor.Shutdown",
```
cijothomas, line 684: why change event name? anything wrong with the existing name? Happy to improve it.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -632,14 +678,21 @@ impl SpanProcessor for BatchSpanProcessor {
...
+                dropped_spans = dropped_spans,
+                max_queue_size = self.max_queue_size,
+                message = "Spans were dropped due to a full queue. The count represents the total count of span records dropped in the lifetime of the BatchSpanProcessor. Consider increasing the queue size and/or decrease delay between intervals."
             );
+
+            // Also return error so user code can handle it programmatically
```
cijothomas, line 690: this feels incorrect. This is the `shutdown()` call, and that method is not failing here to warrant returning an `Err` - having items dropped due to buffer full is not a failure of `shutdown` itself.

cijothomas (changes requested):
Requesting changes as we don't want to proceed as-is.

Adding error callbacks per-processor doesn't scale — the same pattern would need to be replicated in places like exporters, samplers, log processor, metric reader, etc. If we pursue this, a global error handler (provider-level or SDK-wide) is the right abstraction. The existing otel_error!/otel_warn! macros already centralize error reporting and could serve as the hook point for a user-registered callback. 

There is also a fundamental design challenge around what gets passed to the callback.The current OTelSdkError enum is too coarse — it tells you something like "export failed" but nothing much more. On the other hand, we can't pass component-specific details either, because a global handler needs a uniform signature that works across all components (span processors, log processors, metric readers, exporters, samplers, etc.). Designing an error type that is both general enough for a single callback signature and specific enough to be actionable is a non-trivial problem that needs to be solved before implementation.

PR #3223, opened 2025-11-05
feat: add forceflush_timeout configurability

kovaszab:
## Changes

Adds configurability of force flushing timeout for batch processors (BatchSpanProcessor, BatchLogProcessor).

### DIFF
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -54,6 +54,10 @@ pub(crate) const OTEL_BLRP_MAX_QUEUE_SIZE_DEFAULT: usize = 2_048;
 pub(crate) const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE: &str = "OTEL_BLRP_MAX_EXPORT_BATCH_SIZE";
 /// Default maximum batch size.
 pub(crate) const OTEL_BLRP_MAX_EXPORT_BATCH_SIZE_DEFAULT: usize = 512;
+/// Force flush timeout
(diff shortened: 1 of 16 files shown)


opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -682,6 +692,17 @@ impl BatchConfigBuilder {
         self
     }
 
+    /// Set forceflush_timeout for [`BatchConfigBuilder`].
+    /// The default value is 5000 milliseconds.
+    ///
+    /// Corresponding environment variable: `OTEL_BLRP_FORCE_FLUSH_TIMEOUT`.
+    ///
+    /// Note: Programmatically setting this will override any value set via the environment variable.
+    pub fn with_forceflush_timeout(mut self, forceflush_timeout: Duration) -> Self {
```
cijothomas: for shutdown, we allow users to pass timeout in each shutdown call itself. Could we stick with the same for flush also?

cijothomas: Its best to be consistent with shutdown, if at all we are adding timeout to flush.
i.e Add this all the way from [provider](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/logs/logger_provider.rs#L86) itself.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust/pull/3223/files#r2495071383
Want to see if there is a reason for this be done differently than Shutdown timeouts.

opentelemetry-sdk/src/logs/batch_log_processor.rs
kovaszab: You're right, I'll refactor the code to mirror the propagation of the provider's timeout like we do for `shutdown_with_timeout`.

PR #3226, opened 2025-11-06
Add declarative configuration initial version

andborja:
## Changes

Add declarative configuration initial version.


opentelemetry-declarative-config/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+opentelemetry = { version = "0.31.0" }
+opentelemetry_sdk = { version = "0.31.0", features = ["experimental_metrics_custom_reader"] }
+opentelemetry-stdout = { version = "0.31.0" }
+opentelemetry-otlp = { version = "0.31.0" }
+serde = { workspace = true, features = ["derive"] }
+serde_yaml = "0.9.34"
+tokio = { workspace = true, features = ["sync", "rt"], optional = true }
```
lalitb: nit - do we need this as non-dev dependency? Don't see it being used.

opentelemetry-declarative-config/README.md
```diff
@@ -0,0 +1,29 @@
...
+[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
+
+Declarative configuration for applications instrumented with [`OpenTelemetry`]. 
+
+## OpenTelemetry Overview
+
+OpenTelemetry is an Observability framework and toolkit designed to create and
+manage telemetry data such as traces, metrics, and logs. OpenTelemetry is
+vendor- and tool-agnostic, meaning that it can be used with a broad variety of
+Observability backends, including open source tools like [Jaeger] and
```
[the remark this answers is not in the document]
andborja: Added below

opentelemetry-declarative-config/Cargo.toml
andborja: Good catch. No. Removed.

lalitb:
Thanks @andborja for the PR - great to see contribution coming in for the [opentelemetry-configuration](https://github.com/open-telemetry/opentelemetry-configuration/tree/main) implementation.

This isn’t specific to the PR content itself, but more of a broader question with other maintainers/approvers -

The OTel Configuration spec is currently at v1.0.0-rc.2 (stable target seems to be March 2026). Should this implementation start in `opentelemetrt-rust-ccontrib`,  and then migrate to main once config is stable? Or would we prefer to start it directly in main but mark it as experimental until the spec stabilizes?

opentelemetry-declarative-config/README.md
```diff
@@ -0,0 +1,30 @@
+# OpenTelemetry Declarative Configuration
+
+![OpenTelemetry — An observability framework for cloud-native software.][splash]
+
+[splash]: https://raw.githubusercontent.com/open-telemetry/opentelemetry-rust/main/assets/logo-text.png
+
+Declarative configuration for applications instrumented with [`OpenTelemetry`]. 
+
+## OpenTelemetry Overview
```
utpilla, line 9: This README should talk about OpenTelemetry Declarative Configuration and not general OpenTelemetry overview.

opentelemetry-declarative-config/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+opentelemetry = { version = "0.31.0" }
+opentelemetry_sdk = { version = "0.31.0", features = ["experimental_metrics_custom_reader"] }
+opentelemetry-stdout = { version = "0.31.0" }
+opentelemetry-otlp = { version = "0.31.0" }
+opentelemetry-http = { workspace = true, optional = true, default-features = false }
+serde = { workspace = true, features = ["derive"] }
+reqwest = { workspace = true, optional = true }
```
utpilla, line 28: I don't think `opentelemetry-http` and `reqwest` need to be added as direct dependencies. You already have `opentelemetry-otlp` dependency which should transitively pull these two dependencies.

Once you remove these two, you can also remove the `cargo-machete` section as well.

opentelemetry-declarative-config/examples/declarative_basic.rs
```diff
@@ -0,0 +1,44 @@
+use opentelemetry_declarative_config::Configurator;
+
+/// Example of configuring OpenTelemetry telemetry using declarative YAML configuration.
+
+#[tokio::main]
+pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
```
utpilla, line 6: Make this a regular main method. We don't need this to be `async`.
1. That would clearly convey that this crate does not have any asynchronous APIs
2. You can also remove the dependency on `tokio`.

opentelemetry-declarative-config/examples/declarative_basic.rs
```diff
@@ -0,0 +1,44 @@
...
+                - batch:
+                    exporter:
+                        stdout:
+                        otlp:
+                            protocol: http/protobuf
+                            endpoint: https://backend:4318
+        resource:
+            service.name: sample-service
+            service.version: "1.0.0"
+    "#;
```
utpilla, line 29: Could we move this to an actual `.yaml` instead of using raw string literal? That would make the example more helpful.

opentelemetry-declarative-config/src/metrics/reader_config.rs
```diff
@@ -0,0 +1,27 @@
...
+//! metrics data and exporting it to various backends or systems.
+
+use serde::Deserialize;
+use serde_yaml::Value;
+
+/// Configuration for Metrics Readers
+#[derive(Deserialize)]
+pub struct ReaderConfig {
+    pub pull: Option<PullReaderConfig>,
+    pub periodic: Option<PeriodicReaderConfig>,
```
utpilla, line 14: We want to use an `enum` here. A `MetricReader` would either be periodic (push-based) or pull-based. It can't be both.

opentelemetry-declarative-config/examples/declarative_basic.rs
andborja: It looks like the OTLP implementation requires it. Let's chat on how to avoid it.

opentelemetry-declarative-config/examples/declarative_basic.rs
andborja: As this will be embedded most likely in other configurations, the most common use case is actually a string or the object composition itself. I'll add another one with a file for more coverage.

utpilla: The examples are for our users. Wouldn't the common use case for users be to provide the config in a file? For our tests, we can always use a string.

opentelemetry-declarative-config/src/telemetry_config.rs
```diff
@@ -0,0 +1,86 @@
...
+pub struct TelemetryConfig {
+    /// Metrics telemetry configuration
+    pub metrics: Option<MetricsConfig>,
+
+    /// Logs telemetry configuration
+    pub logs: Option<LogsConfig>,
+
+    /// Resource attributes to be associated with all telemetry data
+    #[serde(default)]
+    pub resource: HashMap<String, String>,
```
lalitb, line 23: `HashMap<String, String>` restricts resource attributes to string values only. According to the specs,  resource attributes must support multiple data types including integers, booleans, doubles, and arrays. If we plan to fix this in future iteration, lets add a TODO, as this would be critical requirement.

opentelemetry-declarative-config/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+                        }
+                    }
+                } else {
+                    return Err("Pull exporter configuration must be defined".into());
+                }
+            }
+        }
+
+        let provider = provider_builder.build();
+        global::set_meter_provider(provider.clone());
```
lalitb, line 136: This should be left to the user - the configurator should just build providers, not register them globally.

opentelemetry-declarative-config/src/lib.rs
```diff
@@ -0,0 +1,353 @@
...
+        file_path: &str,
+    ) -> Result<TelemetryProviders, Box<dyn std::error::Error>> {
+        let config = TelemetryConfig::from_yaml_file(file_path)?;
+        self.configure_telemetry(config)
+    }
+
+    pub fn configure_telemetry(
+        &self,
+        telemetry_config: TelemetryConfig,
+    ) -> Result<TelemetryProviders, Box<dyn std::error::Error>> {
```
lalitb, line 50: Not for this iteration, but good to see if we can return more structured errors to the caller.

opentelemetry-declarative-config/src/logs.rs
```diff
@@ -0,0 +1,76 @@
...
+pub trait LogsBatchExporterFactory {
+    /// Creates a LogsExporter based on the provided configuration
+    fn create_logs_batch_exporter(
+        &self,
+        config: &Value,
+    ) -> Result<impl LogExporter + 'static, Box<dyn std::error::Error>>;
+}
+
+/// Enum representing different Batch Exporter Factories
+pub enum BatchExporterFactory {
```
cijothomas, line 37: might be better to separate processor and exporter.

opentelemetry-declarative-config/src/logs/processor_config.rs
```diff
@@ -0,0 +1,18 @@
...
+/// Configuration for Log Processors
+#[derive(Deserialize)]
+pub struct ProcessorConfig {
+    pub batch: Option<ProcessorBatchConfig>,
+}
+
+/// Configuration for Batch Log Processor
+#[derive(Deserialize)]
+pub struct ProcessorBatchConfig {
+    pub exporter: Value,
```
lalitb, line 17: This is more generic comment for all the config's - Storing the exporter/processor/reader as `serde_yaml::Value` isn’t ideal - it bypasses all type validation and defers errors to runtime. This should use a typed struct or enum so schema and type issues are caught during parsing, not during exporter creation.

opentelemetry-declarative-config/src/metrics.rs
```diff
@@ -0,0 +1,118 @@
...
+        &self,
+        config: &Value,
+    ) -> Result<Self::Exporter, Box<dyn std::error::Error>>;
+}
+
+/// Factory enum to create different types of exporters
+pub enum PeriodicExporterFactory {
+    Stdout(StdoutMetricsPeriodicExporterFactory),
+    Otlp(OtlpMetricsPeriodicExporterFactory),
+    //TODO: Add other variants as needed
```
lalitb, line 51: This is not very flexible. The external custom crates can't be added as new variants to the enum without adding them as dependency to `otel-declarative-config` crate. This should be more flexible to be extended at runtime.

One out of other approaches could be to replace the enums with trait + registry pattern:

```rust
pub trait MetricsExporterFactory: Send + Sync {
    fn create_exporter(&self, config: &Value) 
        -> Result<Box<dyn PushMetricExporter>, Error>;
}

pub struct Configurator {
    exporters: HashMap<String, Box<dyn MetricsExporterFactory>>,
}

impl Configurator {
    pub fn new() -> Self {
        let mut exporters = HashMap::new();
        // Built-in exporters
        exporters.insert("stdout".into(), Box::new(StdoutFactory));
        exporters.insert("otlp".into(), Box::new(OtlpFactory));
        Self { exporters }
    }
    
    // Allow external registration
    pub fn register_exporter(&mut self, name: String, factory: Box<dyn MetricsExporterFactory>) {
        self.exporters.insert(name, factory);
    }
}
```

PR #3229, opened 2025-11-10
chore(opentelemetry-otlp): Make invalid combinations of transport & encoding unrepresentable

### LINKED ISSUE #3082: OTLP Stabilization: Protocol enum ambiguity in API
**OpenTelemetry Protocol Requirements:**
The [OTLP specification](https://opentelemetry.io/docs/specs/otlp/) supports multiple transport protocols: gRPC, HTTP/protobuf, and HTTP/JSON. 

We model this using a fluent builder style API - first to select the transport protocol (note the feature gates): 

https://github.com/open-telemetry/opentelemetry-rust/blob/c5fde17e7d17af9f2987b07f9d3f80c03291b7cc/opentelemetry-otlp/src/span.rs#L60-L61

https://github.com/open-telemetry/opentelemetry-rust/blob/c5fde17e7d17af9f2987b07f9d3f80c03291b7cc/opentelemetry-otlp/src/span.rs#L68-L70

Then, to select the encoding (rather confusingly called "protocol"):

https://github.com/open-telemetry/opentelemetry-rust/blob/c5fde17e7d17af9f2987b07f9d3f80c03291b7cc/opentelemetry-otlp/src/exporter/mod.rs#L258

https://github.com/open-telemetry/opentelemetry-rust/blob/c5fde17e7d17af9f2987b07f9d3f80c03291b7cc/opentelemetry-otlp/src/lib.rs#L440-L450

This means we can select impossible combinations which are silently remapped at runtime:

**HTTP Transport + gRPC Protocol**: `with_http().with_protocol(Protocol::Grpc)` uses HTTP transport with protobuf instead of actual gRPC wire format, silently accepting semantically incorrect configuration. 

[exporter/http/mod.rs:303](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/http/mod.rs#L303) - `Protocol::Grpc` falls through to `_` wildcard and gets HTTP+protobuf instead of failing

**gRPC Transport + HTTP Protocol**: `with_tonic().with_protocol(Protocol::HttpJson)` ignores protocol setting entirely, always uses gRPC regardless of specified protocol  

[exporter/tonic/mod.rs:140](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/tonic/mod.rs#L140) - TonicExporterBuilder hard-codes `protocol: Protocol::Grpc` regardless of user configuration

**Feature-Gated Runtime Behavior**: `Protocol::HttpJson` compiles successfully but silently falls back to protobuf when `http-json` feature is disabled
[exporter/http/mod.rs:298-299](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/http/mod.rs#L298-L299) - `#[cfg(feature = "http-json")]` arm disappears at compile time without the feature

scottgerring:
# Really A Draft: Trying a different approach; please hold off on reviews!

Fixes #3082 
This PR removes the ability to specify invalid combinations of transport (HTTP/gRPC) and encoding (JSON/Protobuf) by making the invalid combinations unrepresentable. 

For completeness, I [staged a separate branch](https://github.com/open-telemetry/opentelemetry-rust/compare/main...scottgerring:opentelemetry-rust:chore/protocol-cleanup-1?expand=1) that simply errors out when a user specifies an invalid combination. This has less impact on the API surface and is a smaller change, but doesn't _prevent_ the user from specifying the invalid combinations. I prefer this variant here where we use the type system! 

## Changes

We do this by:

* Removing `with_protocol` from the `WithExportConfig`  shared between both gRPC and HTTP exporters
* Introducing `HttpEncodingProtocol`, and supporting methods on `WithHttpConfig` to configure JSON or Protobuf encoding

I wanted to draw a clear line between 'protocol' and 'encoding' as this feels like where the confusion has crept in, but [the spec itself](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol) uses the term protocol to refer to both application layer protocol and body encoding.

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -1,8 +1,8 @@
 use super::{
-    default_headers, default_protocol, parse_header_string, resolve_timeout, ExporterBuildError,
+    default_headers, parse_header_string, resolve_timeout, ExporterBuildError,
     OTEL_EXPORTER_OTLP_HTTP_ENDPOINT_DEFAULT,
 };
-use crate::{ExportConfig, Protocol, OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_HEADERS};
+use crate::{ExportConfig, OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_HEADERS};
 use http::{HeaderName, HeaderValue, Uri};
 use opentelemetry::otel_debug;
 use opentelemetry_http::{Bytes, HttpClient};
@@ -100,6 +100,32 @@ mod trace;
 ))]
 use opentelemetry_http::hyper::HyperClient;
 
+/// The protocol to used for HTTP requests
+///
+/// This combination of transport (HTTP) and encoding (protobuf / JSON) is
+/// used in the OTel spec itself.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Protocol {
+    /// Binary protobuf encoding over HTTP
+    HttpProtobuf,
+    /// JSON encoding over HTTP
+    HttpJson,
+}
+
+#[allow(clippy::derivable_impls)]
+impl Default for Protocol {
+    fn default() -> Self {
+        #[cfg(feature = "http-json")]
+        {
+            Protocol::HttpJson
+        }
+        #[cfg(not(feature = "http-json"))]
+        {
+            Protocol::HttpProtobuf
+        }
+    }
+}
+
 /// Configuration of the http transport
 #[derive(Debug, Default)]
 pub struct HttpConfig {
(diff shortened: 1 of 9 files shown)


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -115,6 +138,23 @@ pub struct HttpConfig {
     /// The retry policy to use for HTTP requests.
     #[cfg(feature = "experimental-http-retry")]
     retry_policy: Option<RetryPolicy>,
+
+    /// The encoding format for HTTP payloads.
+    encoding: HttpEncoding,
+}
+
+#[allow(clippy::derivable_impls)]
+impl Default for HttpConfig {
```
bantonsson: I don't see anything in this that can't be derived.

scottgerring: Good catch!

PR #3230, opened 2025-11-10
chore(otlp): prevent invalid combinations of exporter and protocol

### LINKED ISSUE #3082: OTLP Stabilization: Protocol enum ambiguity in API

### LINKED ISSUE #3208: [Bug]:  OTEL_EXPORTER_OTLP_PROTOCOL exists but is never used
### What happened?

I'd expect that with this version set and depending on all features to be able to initialise either tonic or http just by leveraging this env var and explecitely set all protocol options.
The defaulting and behaviour dependending on support is pretty clear in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specify-protocol)

Should there just be a `SpanExporter::default()` option?

I instead have to do:

```
        let exporter = match env::var(OTEL_EXPORTER_OTLP_PROTOCOL).unwrap_or(OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT.to_owned()).as_str() { 
            "http/json" => {
                opentelemetry_otlp::SpanExporter::builder().with_http().with_protocol(Protocol::HttpJson).build()
            }
            "grpc" => {
                opentelemetry_otlp::SpanExporter::builder().with_tonic().build()
            }
            _ => {
                opentelemetry_otlp::SpanExporter::builder().with_http().with_protocol(Protocol::HttpBinary).build()
            }
        }?;

```

opentelemetry-otlp = { version = "0.31", features = ["metrics", "grpc-tonic", "http-json", "http-proto"] }

opentelemetry_sdk = { version = "0.31.0", features = ["metrics", "rt-tokio", "experimental_metrics_periodicreader_with_async_runtime", "trace"] }

OTLP

### Relevant log output

```shell

```

### Tip

scottgerring:
Fixes #3082 
Fixes #3208 

## Changes
Make invalid combinations of configured exporter features and exporter protocols unrepresentable by using feature flags, and adds support for `OTEL_EXPORTER_OTLP_PROTOCOL`

Alternative to #3229. In #3229, I've taken the extra step of moving the configurable protocol on the builders down into the HTTP client _only_, as it is not actually configurable on the tonic exporter - there's no choice.

After speaking to @bantonsson I reckon this is a lighter touch change which feels appropriate.

### DIFF
--- a/opentelemetry-otlp/src/exporter/mod.rs
+++ b/opentelemetry-otlp/src/exporter/mod.rs
@@ -29,6 +29,13 @@ pub const OTEL_EXPORTER_OTLP_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_PROTOCOL";
 /// Compression algorithm to use, defaults to none.
 pub const OTEL_EXPORTER_OTLP_COMPRESSION: &str = "OTEL_EXPORTER_OTLP_COMPRESSION";
 
+/// Protocol value for HTTP with protobuf encoding
+pub const OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_PROTOBUF: &str = "http/protobuf";
+/// Protocol value for gRPC
+pub const OTEL_EXPORTER_OTLP_PROTOCOL_GRPC: &str = "grpc";
+/// Protocol value for HTTP with JSON encoding
+pub const OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON: &str = "http/json";
+
 /// Max waiting time for the backend to process each signal batch, defaults to 10 seconds.
 pub const OTEL_EXPORTER_OTLP_TIMEOUT: &str = "OTEL_EXPORTER_OTLP_TIMEOUT";
 /// Default max waiting time for the backend to process each signal batch.
@@ -66,7 +73,7 @@ pub struct ExportConfig {
 #[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
 impl Default for ExportConfig {
     fn default() -> Self {
-        let protocol = default_protocol();
+        let protocol = Protocol::default();
 
         Self {
             endpoint: None,
@@ -170,25 +177,34 @@ fn resolve_compression_from_env(
     }
 }
 
-/// Returns the default protocol based on enabled features.
+/// Returns the default protocol based on environment variable or enabled features.
 ///
 /// Priority order (first available wins):
-/// 1. http-json (if enabled)
-/// 2. http-proto (if enabled)
-/// 3. grpc-tonic (if enabled)
+/// 1. OTEL_EXPORTER_OTLP_PROTOCOL environment variable (if set and feature is enabled)
+/// 2. http-json (if enabled)
+/// 3. http-proto (if enabled)
+/// 4. grpc-tonic (if enabled)
 #[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
-fn default_protocol() -> Protocol {
-    #[cfg(feature = "http-json")]
-    return Protocol::HttpJson;
+impl Default for Protocol {
+    fn default() -> Self {
+        // Check environment variable first
+        if let Some(protocol) = Protocol::from_env() {
+            return protocol;
+        }
 
-    #[cfg(all(feature = "http-proto", not(feature = "http-json")))]
-    return Protocol::HttpBinary;
+        // Fall back to feature-based defaults
+        #[cfg(feature = "http-json")]
+        return Protocol::HttpJson;
 
-    #[cfg(all(
-        feature = "grpc-tonic",
-        not(any(feature = "http-proto", feature = "http-json"))
-    ))]
-    return Protocol::Grpc;
+        #[cfg(all(feature = "http-proto", not(feature = "http-json")))]
+        return Protocol::HttpBinary;
+
+        #[cfg(all(
+            feature = "grpc-tonic",
+            not(any(feature = "http-proto", feature = "http-json"))
+        ))]
+        return Protocol::Grpc;
+    }
 }
 
 /// default user-agent headers
@@ -451,10 +467,7 @@ mod tests {
             not(any(feature = "grpc-tonic", feature = "http-proto"))
         ))]
         {
-            assert_eq!(
-                crate::exporter::default_protocol(),
-                crate::Protocol::HttpJson
-            );
+            assert_eq!(crate::Protocol::default(), crate::Protocol::HttpJson);
         }
 
         #[cfg(all(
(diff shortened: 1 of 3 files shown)


scottgerring:
@bantonsson , as you looked at the Bigger One, if you have a second!

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -29,27 +29,6 @@ pub const OTEL_EXPORTER_OTLP_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_PROTOCOL";
 /// Compression algorithm to use, defaults to none.
 pub const OTEL_EXPORTER_OTLP_COMPRESSION: &str = "OTEL_EXPORTER_OTLP_COMPRESSION";
 
-#[cfg(feature = "http-json")]
-/// Default protocol, using http-json.
-pub const OTEL_EXPORTER_OTLP_PROTOCOL_DEFAULT: &str = OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON;
```
bantonsson, line 34: These constants could still be available

scottgerring: It's not clear to me what they're for aside from the default mechanism; they aren't used otherwise in the codebase, but I note they _are_ public.

scottgerring: Actually I'll need these to close #3208

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -190,14 +170,25 @@ fn resolve_compression_from_env(
...
 
-/// default protocol based on enabled features
+/// Returns the default protocol based on enabled features.
+///
+/// Priority order (first available wins):
+/// 1. http-json (if enabled)
+/// 2. http-proto (if enabled)
+/// 3. grpc-tonic (if enabled)
+#[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
 fn default_protocol() -> Protocol {
```
bantonsson: Can't this be changed to `impl Default for Protocol`?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -190,13 +177,33 @@ fn resolve_compression_from_env(
...
+/// Priority order (first available wins):
+/// 1. OTEL_EXPORTER_OTLP_PROTOCOL environment variable (if set and feature is enabled)
+/// 2. http-json (if enabled)
+/// 3. http-proto (if enabled)
+/// 4. grpc-tonic (if enabled)
+#[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
+impl Default for Protocol {
+    fn default() -> Self {
+        // Check environment variable first
+        if let Some(protocol) = Protocol::from_env() {
```
scottgerring, line 191: Let's close #3208 too seeing as its related; this lets us us the environment variables to pick exporter, and then falls back based on feature flag.

opentelemetry-otlp/src/lib.rs
```diff
@@ -450,13 +451,46 @@ use serde::{Deserialize, Serialize};
...
     /// HTTP protocol with JSON payload
+    #[cfg(feature = "http-json")]
     HttpJson,
 }
 
+#[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
+impl Protocol {
+    /// Attempts to parse a protocol from the `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable.
+    ///
+    /// Returns `None` if:
```
scottgerring, line 468: I've made a decision here to return `None` and _not panic_ if a user asks for a valid protocol that we have not compiled in. Open to discuss.

lalitb: If the env variable points to a protocol that wasn’t compiled in (e.g. http/protobuf but only gRPC is built), seems we currently fall back silently to the compiled one. Might be worth logging a warning so users know their requested protocol wasn’t used.

scottgerring: Good idea; folks might be accidentally relying on the existing behaviour

lalitb (approved):

opentelemetry-otlp/src/lib.rs
scottgerring: Added!

Issue #2985, opened 2025-05-21
View allows instrument selection via Description, which is not allowed


cijothomas:
`Instrument` currently allows users to select an instrument based on description, which is not allowed in the spec. Only name, kind, unit, scope should be exposed.

Can be handled as part of removing public fields and instead expose public getter like methods.

PR #3245, opened 2025-11-14
feat: Add description getter to `metrics::Instrument` in SDK

ongardie-atomix:
Note: This change is in conflict with <https://github.com/open-telemetry/opentelemetry-rust/issues/2985> by @cijothomas:

> Instrument currently allows users to select an instrument based on description, which is not allowed in the spec. Only name, kind, unit, scope should be exposed.

The spec at <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#instrument-selection-criteria> says:

> The SDK MAY accept additional criteria.

So it seems like the description MAY be exposed.

---

I could imagine accessing the description in a View for valid reasons, like wanting to append a "." (or remove a ".") from the end of the descriptions. I admit my motivation for doing this PR is part of a larger hack, however. I want my application to be able to list all of its metrics with descriptions, even for metrics that have been registered but not used. I think the only way I can get access to those currently is with a View, but I know that's not its intended use.

---

## Changes

- Adds a getter for the description field of `metrics::Instrument` in the SDK.
- Updates docs based on the field definitions, which seem more descriptive, and cleans up some others.
- Does this need a `CHANGELOG.md` entry?

### DIFF
--- a/opentelemetry-sdk/src/metrics/instrument.rs
+++ b/opentelemetry-sdk/src/metrics/instrument.rs
@@ -32,12 +32,10 @@ pub enum InstrumentKind {
     /// A group of instruments that record increasing and decreasing values in an
     /// asynchronous callback.
     ObservableUpDownCounter,
-
-    /// a group of instruments that record current value synchronously with
+    /// A group of instruments that record current value synchronously with
     /// the code path they are measuring.
     Gauge,
-    ///
-    /// a group of instruments that record current values in an asynchronous callback.
+    /// A group of instruments that record current values in an asynchronous callback.
     ObservableGauge,
 }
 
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -97,33 +95,38 @@ impl InstrumentKind {
...
 impl Instrument {
-    /// Instrument name.
+    /// The human-readable identifier of the instrument.
     pub fn name(&self) -> &str {
         self.name.as_ref()
     }
 
-    /// Instrument kind.
+    /// Describes the purpose of the instrument.
+    pub fn descripton(&self) -> &str {
```
ongardie-atomix: Oops, there's a typo here, should be `description` not `descripton`

ongardie-atomix:
I'll elaborate about my real motivation in another issue soon. I'm just trying to understand things a little better first.

PR #3246, opened 2025-11-15
feat: Add `#[must_use]` to AsyncInstrumentBuilder

ongardie-atomix:
This prevents code like the following from compiling:

```rust
meter
    .i64_observable_up_down_counter("my_counter")
    .with_callback(|observer| {
        observer.observe(1, &[]);
    });
```

This code is missing the `.build()` at the end. Before this commit, it would compile, but the callback would never be called. After this commit, it will cause a mostly helpful compile error:

```
warning: unused `AsyncInstrumentBuilder` that must be used
  --> examples/metrics-basic/src/main.rs:71:5
   |
71 | /     meter
72 | |         .i64_observable_up_down_counter("my_counter")
73 | |         .with_callback(|observer| {
74 | |             observer.observe(1, &[]);
75 | |         });
   | |__________^
   |
   = note: Callbacks will not be invoked unless you call .build() on this async instrument builder.
   = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
   |
71 |     let _ = meter
   |     +++++++
```

Most builders need to be built to be useful. Normal metrics are built and then observations are reported to them, where a failure to call `build()` would be discovered at compile-time. However, observable metrics (async instruments) don't have that natural use point, as the result of building an `AsyncInstrumentBuilder` isn't typically needed. This makes `AsyncInstrumentBuilder` especially error-prone without the `#[must_use]` declaration.

Fixes #

## Changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       with:
         toolchain: stable
         components: rustfmt, clippy
-    - uses: taiki-e/install-action@47be02f2de8a32619316956f6117e150bdc6763f # v2.62.44
+    - uses: taiki-e/install-action@6f9c7cc51aa54b13cbcbd12f8bbf69d8ba405b4b # v2.62.47
       with:
           tool: cargo-hack
     - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
@@ -93,7 +93,7 @@ jobs:
           # https://github.com/awslabs/cargo-check-external-types/releases
           toolchain: nightly-2025-05-04
           components: rustfmt
-      - uses: taiki-e/install-action@47be02f2de8a32619316956f6117e150bdc6763f # v2.62.44
+      - uses: taiki-e/install-action@6f9c7cc51aa54b13cbcbd12f8bbf69d8ba405b4b # v2.62.47
         with:
           tool: cargo-check-external-types@0.2.0
       - name: external-type-check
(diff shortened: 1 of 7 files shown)


opentelemetry/src/metrics/instruments/mod.rs
```diff
@@ -241,6 +241,7 @@ impl<T> fmt::Debug for HistogramBuilder<'_, T> {
 pub type Callback<T> = Box<dyn Fn(&dyn AsyncInstrument<T>) + Send + Sync>;
 
 /// Configuration for building an async instrument.
+#[must_use = "AsyncInstrumentBuilder must be built to receive callbacks."]
```
cijothomas:
```suggestion
#[must_use = "Callbacks will not be invoked unless you call .build() on this async instrument builder."]
```

opentelemetry/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - **Breaking** Removed the following public fields and methods from the `SpanBuilder` [#3227][3227]:
   - `trace_id`, `span_id`, `end_time`, `status`, `sampling_result`
   - `with_trace_id`, `with_span_id`, `with_end_time`, `with_status`, `with_sampling_result`
+- **Breaking** Added `#[must_use]` to `opentelemetry::metrics::AsyncInstrumentBuilder`, since unbuilt async instruments will not have their callbacks called. This may cause compile errors for buggy or suspicious code.
```
cijothomas: https://doc.rust-lang.org/cargo/reference/semver.html#new-lints

We cannot make breaking changes, but this should be acceptable based on rust guidelines. I suggest to avoid "breaking" work in changelog to avoid scaring users.

My suggestion:

- **Added** `#[must_use]` attribute to `opentelemetry::metrics::AsyncInstrumentBuilder` to add compile time warning when `.build()` is not called on observable instrument builders, preventing silent failures where callbacks are never registered and metrics are never reported.

cijothomas (changes requested):
Change itself is good (thanks for this nice improvement!).
Requesting change to make a better changelog and message (suggestions given as comments already).

cijothomas:
> My editor also removed some trailing whitespace in the file, sorry. I've left that in since it's technically an improvement, but let me know if you'd prefer me to revert that

@ongardie-atomix If you have interest/bandwidth, we could add a CI check to prevent such things from ever getting checked in. Example in other OTel repos: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/.github/workflows/sanitycheck.yml#L34

PR #3247, opened 2025-11-15
chore: replace `cargo-machete` with `cargo-shear`, removed unused deps

CathalMullan:
## Changes

Replaced `cargo-machete` with [`cargo-shear`](https://github.com/boshen/cargo-shear) for detecting unused dependencies.

`cargo-shear` uses syn to parse Rust code to find unused deps, rather than relying on regex like `cargo-machete`. It also has some additional features, like parsing doc blocks for imports, and running against expanded code (requiring nightly). Tends to give fewer false positives as a result. It is slower to run, though.

Had to bump `taiki-e/install-action` in order to get a newer version of `cargo-shear`, which has a needed fix around doc block parsing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
       with:
         toolchain: stable
         components: rustfmt, clippy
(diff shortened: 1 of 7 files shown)


lalitb (approved):

cijothomas (approved):
Thanks!

PR #3248, opened 2025-11-16
fix: ObservableCounter and ObservableUpDownCounter remove stale attribute combinations that are no longer observed

### LINKED ISSUE #3213: [Bug]: observable counters and up-down counters always report all historical label set
### What happened?

This seems very similar to https://github.com/open-telemetry/opentelemetry-rust/issues/2213.

I have an observable counter for bytes sent/received for my TCP server, with a dynamic label for the client IP/port. As with the linked bug, the last value for each ip/port gets reported, instead of only the current observed values.

Note that the MR for the linked bug seems to be working, since I can't reproduce the bug with observable gauges.

0.31

0.31

OTLP

### Relevant log output

```shell

```

### Tip

taisho6339:
Fixes #3213

## Changes

ObservableCounter and ObservableUpDownCounter now correctly report only data points from the current measurement cycle, removing stale attribute combinations that are no longer observed.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
@@ -116,8 +116,10 @@ impl<T: Number> PrecomputedSum<T> {
         s_data.temporality = Temporality::Cumulative;
         s_data.is_monotonic = self.monotonic;
 
+        // Use collect_and_reset to remove stale attributes (not observed in current callback)
+        // For cumulative, report absolute values (no delta calculation needed)
         self.value_map
-            .collect_readonly(&mut s_data.data_points, |attributes, aggr| SumDataPoint {
+            .collect_and_reset(&mut s_data.data_points, |attributes, aggr| SumDataPoint {
                 attributes,
                 value: aggr.value.get_value(),
                 exemplars: vec![],
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -17,11 +17,13 @@
 - Fix panics and exploding memory usage from large cardinality limit [#3290][3290]
 - Fix Histogram boundaries being ignored in the presence of views [#3312][3312]
 - `TracerProviderBuilder::with_sampler` allows to pass boxed instance of `ShouldSample` [#3313][3313]
+- Fix ObservableCounter and ObservableUpDownCounter now correctly report only data points from the current measurement cycle, removing stale attribute combinations that are no longer observed. [#3248][3248]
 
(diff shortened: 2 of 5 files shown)


opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
```diff
@@ -116,8 +116,10 @@ impl<T: Number> PrecomputedSum<T> {
         s_data.temporality = Temporality::Cumulative;
         s_data.is_monotonic = self.monotonic;
 
+        // Use collect_and_reset to remove stale attributes (not observed in current callback)
+        // For cumulative, report absolute values (no delta calculation needed)
         self.value_map
-            .collect_readonly(&mut s_data.data_points, |attributes, aggr| SumDataPoint {
+            .collect_and_reset(&mut s_data.data_points, |attributes, aggr| SumDataPoint {
```
cijothomas, line 122: I don't think this is the right fix as it causes other issue.
Take an ObservableCounter for example, and SDK is configured with Delta temporality.

T0-T1 {callback reports value=100 for {dimension=A} }
SDK exports {100, dimension=A}

T1-T2{callback reports 150 for dimension=A} //
SDK exports {50, dimension=A}

Inorder for us to report {50, dimension=A}, we need to remember the previous state, and subtract (150-100 = 50). Since this PR switches to `collect_and_reset`, we forget the previous state. When callback reports 150 for dimension=A, we won't have the prior state, so we'd export 150, dimension=A, which violates delta setting.


Let me know if this makes sense. If not, I can try to explain by adding a test case.
(This is a hard bug to fix it seems!)

cijothomas (changes requested):
Apologies for the delay in getting to review this. This is somewhat complex problem/bug, and the current solution is introducing another issue, unfortunately.
Please see https://github.com/open-telemetry/opentelemetry-rust/pull/3248/changes#r2695269659

opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
cijothomas: I can't think of an easy way out of this, so maybe we can accept your fix and live with the bug above! I don't think it is feasible to do both competing things: "Forget stale" vs "Remember to do delta calculation"!
This might be a spec issue actually, that I'll follow up and update here.

cijothomas (approved):
Removing the block. There is an open issue, that is better clarified with specification. This fixes the more important issue.

opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-specification/issues/4861 opened for clarifying. Lets proceed with this fix now.

PR #3251, opened 2025-11-19
feat(tracer): add more convenience methods for enterings spans

bantonsson:
## Changes

This adds 3 convenience methods to the `Tracer` trait that allows users to create, start, and _enter_ a scope with an active `Span` with parent contexts and based on a `SpanBuilder`.

Right now, the easiest way to _enter_ a `Span` is to use the convenience method on the tracer that only takes a `name` and then add tags to the span afterwards. This is problematic in the sense that samplers and processors that take attributes into account will not see these attributes at creation time.


opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -372,4 +372,145 @@ mod tests {
...
+
+        tracer.in_span_with_builder(
+            tracer
+                .span_builder("child")
+                .with_kind(SpanKind::Client)
+                .with_attributes(vec![KeyValue::new("test_key", "test_value")]),
+            |cx| {
+                let span = cx.span();
+                child_trace_id = Some(span.span_context().trace_id());
+                span.add_event("test_event", vec![]);
```
cijothomas: nit: no need to add event here, its not asserted anywhere.

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -372,4 +372,145 @@ mod tests {
...
+        );
+
+        let child_context = child_span_context.unwrap();
+
+        // Verify the child has the same trace_id as parent
+        assert_eq!(child_context.trace_id(), parent_trace_id);
+        // Verify the child has a different span_id than parent
+        assert_ne!(child_context.span_id(), parent_span_id);
+        // Verify both are sampled
+        assert!(child_context.is_sampled());
```
cijothomas: to truly validate the relationship is maintained, should we assert that the the child's parent_span_id is same as parent's span_id?

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -372,4 +372,145 @@ mod tests {
...
+                span.add_event("test_event", vec![]);
+            },
+        );
+
+        // Verify child span inherited parent's trace_id
+        assert_eq!(child_trace_id, Some(parent_trace_id));
+    }
+
+    #[test]
+    fn in_span_with_builder_and_context_uses_provided_context() {
```
cijothomas, line 473: nit: Add a variant of this test where these is an *active* context already and we don't pick that.

cijothomas (commented):
Looks good. lets add a changelog.md entry before merge.

opentelemetry/src/trace/tracer.rs
```diff
@@ -207,11 +207,126 @@ pub trait Tracer {
...
+    ///             .with_kind(SpanKind::Client)
+    ///             .with_attributes(vec![KeyValue::new("key", "value")]),
+    ///         &parent_cx,
+    ///         |_cx| {
+    ///             // child span is active here with configured attributes
+    ///         }
+    ///     )
+    /// }
+    /// ```
+    fn in_span_with_builder_and_context<T, F>(
```
cijothomas, line 315: based on our benchmarks (https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/span.rs#L71) the `in_span` is not the most performant way to span creation/activation. 

Not opposed to this change, just noting the perf implications. Maybe we can improve the implementation to make it better, and/or warn users about perf implication in the doc in follow ups.

bantonsson: I'll take a look at the benchmarks and extend them with versions for these variants.

bantonsson: So the benchmarks are really comparing apples and oranges. Do we want comparable numbers or what are we trying to measure in the benchmarks? Yes, if you don't attach a `Span` to the `Context` it will be faster than attaching a `Span` to the `Context`. If you don't set an `Attribute` on a `Span` it will always be faster than setting an `Attribute` on a `Span`.

The bigger question is what type of API do we want to have here. Should we have a method that maybe attaches the `Span` to the `Context` and gives you a ref to the `Span` instead of the `Context`? This would mean that `Context::current()` would return a `Context` that doesn't have this `Span` attached. This feels weird and I can't really see why you would want that.

cijothomas: > Should we have a method that maybe attaches the Span to the Context and gives you a ref to the Span instead of the Context? This would mean that Context::current() would return a Context that doesn't have this Span attached. This feels weird and I can't really see why you would want that.

I don't think we need such a change.

The benchmarks merely shows that it is not possible for users to avoid paying the context activation cost based on sampling_decision, when using the `tracer.in_span` option, though that is quite convenient to use.

As mentioned in my comment, they are not blockers. We can document this, so users who prefer convenience can accept the slight perf cost.
Or we can look into `tracer.in_span` implementation, to see if we can avoid context activation for unfavorable sampling decisions.

cijothomas (approved):
LGTM, we have a non-blocking conversation about perf implications.

opentelemetry-sdk/benches/span.rs
```diff
@@ -75,11 +75,32 @@ fn criterion_benchmark(c: &mut Criterion) {
             span.set_attribute(KeyValue::new("key1", false));
             span.set_attribute(KeyValue::new("key2", "hello"));
             span.set_attribute(KeyValue::new("key3", 123.456));
-            span.set_attribute(KeyValue::new("key4", "world"));
-            span.set_attribute(KeyValue::new("key5", 123));
+            if span.is_recording() {
```
cijothomas, line 78: don't think this matters much - the set_attribute internal checks this and return early. But yes, we still save the cost of KeyValue creation, which should save 10ns for 2 KVP. We should add this same check to the simple scenario too, to make it more comparable..

Issue #2881, opened 2025-03-28
[Feature]: Allow tower layers for OTLP GRPC exporter


valkum:
### Related Problems?

Google [just released](https://cloud.google.com/trace/docs/release-notes#March_25_2025) their tracing OTLP endpoint. This endpoint [needs](https://cloud.google.com/trace/docs/migrate-to-otlp-endpoints#telemetry_auth_tokens-python) authorization metadata.
Tonic provides interceptors to inject auth metadata and those are exposed via `WithTonicConfig::with_interceptor`. In contrast to tonic, this interceptor is maintained outside of tonic and just allows returning metadata that is then set in the `export` functions.
Getting a valid bearer token normally requires an async operation, and interceptors are currently non-async only.
Currently, you would need to duplicate a lot of code to have blocking counterparts.

### Describe the solution you'd like:

To support this kind of auth, `opentelemetry-otlp` should support adding async interceptors via `WithTonicConfig`.

The tonic native way would be using tower layers to add async interceptor functionality. There currently isn't a helper to do so. There is a standalone crate tonic-async-interceptor, but this also needs a tower layer. 

With the current integration of interceptors in `TonicTracesClient`, the least invasive solution would provide an async interceptor variant that is just awaited in the returned future in `export` and would add the metadata to the request similarly to the current interceptor.

### Additional Context

One common way would be to use `gcp_auth` which provides bearer tokens for use in an async manner.

reimarstier:
Hey, we were tackling the same problem in [OpenDuT](https://opendut.eclipse.dev/) and found another approach with the latest OpenTelemetry version, which has lifted the async runtime requirement for the batch processor. We're using OpenTelemetry with gRPC/Tonic. It shouldn't be complicated to do the same with the HTTP endpoint, though.

- [Interceptor](https://github.com/eclipse-opendut/opendut/blob/74db79a34752af43ca8ef7a74e218c8609ff820d/opendut-util/opendut-auth/src/confidential/client.rs#L195-L264)
- [Create interceptor with tokio runtime handle](https://github.com/eclipse-opendut/opendut/blob/74db79a34752af43ca8ef7a74e218c8609ff820d/opendut-util/src/telemetry/mod.rs#L96)
- [Use interceptor in LogExporter](https://github.com/eclipse-opendut/opendut/blob/74db79a34752af43ca8ef7a74e218c8609ff820d/opendut-util/src/telemetry/logging.rs#L36)
[...]

PR #3255, opened 2025-11-22
fix: handle shutdown in logs exporter

### LINKED ISSUE #2777: Handle shutdown in OTLP/gRPC

stefanobaghino:
Fixes #2777

## Changes

This follows the approach suggested [here](https://github.com/open-telemetry/opentelemetry-rust/issues/2777#issuecomment-3566350234) by @scottgerring:

> The issue is in `TonicLogsClient` which uses `tokio::sync::Mutex` (async) but the `LogExporter::shutdown_with_timeout()` trait method is synchronous. This prevents calling `.lock().await`, so the current implementation has a TODO and just returns `Ok()` without actually shutting down the gRPC client.
> 
> Comparing shutdown on `TonicMetricsClient` , and in particular the type of the mutex used for the inner client and locked in the shutdown, seems to show a workable pattern.

### DIFF
--- a/.clomonitor.yml
+++ b/.clomonitor.yml
@@ -1,0 +1,4 @@
+# see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions
+exemptions:
+  - check: artifacthub_badge
+    reason: "Artifact Hub doesn't support rust packages"
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -34,7 +34,7 @@ jobs:
       GIT_DISCOVERY_ACROSS_FILESYSTEM: 1
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+        uses: step-security/harden-runner@95d9a5deda9de15063e7595e9719c11c38c90ae2 # v2.13.2
         with:
           egress-policy: audit
 
@@ -46,7 +46,7 @@ jobs:
       - name: Make repo safe for Git inside container
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           fetch-depth: 10 # Fetch a bit of history so we can do perf diffs
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,11 @@ jobs:
     continue-on-error: ${{ matrix.rust == 'beta' }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
+      uses: step-security/harden-runner@95d9a5deda9de15063e7595e9719c11c38c90ae2 # v2.13.2
       with:
         egress-policy: audit
 
-    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
       with:
         submodules: true
     - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9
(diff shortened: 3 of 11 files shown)


opentelemetry-otlp/src/exporter/tonic/metrics.rs
```diff
@@ -93,7 +93,7 @@ impl MetricsClient for TonicMetricsClient {
                             Ok((inner.client.clone(), m, e))
                         }
                         None => Err(tonic::Status::failed_precondition(
-                            "exporter is already shut down",
+                            "metrics exporter is already shut down",
```
scottgerring, line 96: I think this is helpful; I don't think the user would necessarily see which failed otherwise

scottgerring (approved):
Thanks! LGTM

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 - Add partial success response handling for OTLP exporters (traces, metrics, logs) per OTLP spec. Exporters now log warnings when the server returns partial success responses with rejected items and error messages. [#865](https://github.com/open-telemetry/opentelemetry-rust/issues/865)
 - Refactor `internal-logs` feature in `opentelemetry-otlp` to reduce unnecessary dependencies[3191](https://github.com/open-telemetry/opentelemetry-rust/pull/3192)
+- Fixed
+  [#2777](https://github.com/open-telemetry/opentelemetry-rust/issues/2777)
+  to properly handle `shutdown_with_timeout()` when using `tonic`.
```
scottgerring: nit: No need to wrap, and we can reference the feature name `grpc-tonic` 

```suggestion
- Fixed [#2777](https://github.com/open-telemetry/opentelemetry rust/issues/2777)  to properly handle `shutdown_with_timeout()` when using `grpc-tonic`.
```

scottgerring:
@stefanobaghino that was a quick turnaround :D

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -5,9 +5,8 @@ use opentelemetry_proto::tonic::collector::logs::v1::{
 };
 use opentelemetry_sdk::error::{OTelSdkError, OTelSdkResult};
 use opentelemetry_sdk::logs::{LogBatch, LogExporter};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
```
cijothomas, line 8: are there any repercussions of moving to std Mutex instead of tokio one inside the async export call?

scottgerring: I think it's fine, and its the same pattern used in `TonicsMetricsClient` for the last couple years, which as far as I can tell we've not had problems with ([linky to blame](https://github.com/open-telemetry/opentelemetry-rust/blame/df412fecf79350f5ccc303c494767d826ed8c979/opentelemetry-otlp/src/exporter/tonic/metrics.rs#L19)).

In the export, we hold the lock only so long as to clone the client and call the interceptor - and significantly, not across any await points, which is the main contra-indicator for std mutexes:

```rust
                let (mut client, metadata, extensions) = self
                    .inner
                    .lock()
                    .map_err(|e| tonic::Status::internal(format!("Failed to acquire lock: {e:?}")))
                    .and_then(|mut inner| match &mut *inner {
Some(inner) => {
let (m, e, _) = inner
.interceptor
.call(Request::new(()))
.map_err(|e| {
// Convert interceptor errors to tonic::Status for retry classification
tonic::Status::internal(format!("interceptor error: {e:?}"))
})? // lock released
```

I _think_ if you called `shutdown_with_timeout` on a single-threaded tokio runtime while the runtime is still trying to do an export, _and_ the export was at that exact moment in the code I highlighted above, _then_ you could potentially deadlock it, but this relies on amazing timing, and I think we already warn against doing exactly this:

https://github.com/open-telemetry/opentelemetry-rust/blob/df412fecf79350f5ccc303c494767d826ed8c979/opentelemetry-sdk/src/logs/batch_log_processor.rs#L99-L102

I note there are some other suggestions (e.g. spawning a thread) to do this particular thing in the linked issue, but this feels a bit overwrought. 

What are you thinking @cijothomas ?

stefanobaghino:
I'd be happy to take any additional feedback on this. If this is not a satisfactory approach, I also have nothing against closing this. Please let me know how to progress. Thanks.

opentelemetry-otlp/src/exporter/tonic/logs.rs
cijothomas: okay given we drop the lock guard before the async export call, this should be fine.

If we do something for
https://github.com/open-telemetry/opentelemetry-rust/issues/2881 we should make sure its not going to cause any issues.

PR #3257, opened 2025-11-22
feat: Redesigns the `InMemoryMetricExporter`

### LINKED ISSUE #2725: Redesign InMemoryExporter
1. Remove the idea of cloning the exporter as shown in https://github.com/open-telemetry/opentelemetry-rust/pull/2649 
2. Make it easier to construct InMemoryExporter easily and retrieve the items from it, instead of forcing users to hand create Mutex wrapped structs etc.
3. Need this for all signals, but would be easy to start with one and expand.

emadpres:
# Redesigned `InMemoryMetricExporter`

Fixes #2725.

## Changes

Updates `InMemoryMetricExporter` (similar to what was suggested in #2649) to accept and use a user-specified data collection to store the metrics data, rather than creating one internally and returning it eventually. This makes the exporter more consistent with the 'production' exporters, which are given their sink by the user (e.g., by being passed a gRPC channel to write to) - rather than owning where the data goes. This also means we no longer need to be able to clone the exporter simply to access the data back out in unit tests. 

If we are happy with this approach, we can extend it to the other in memory exporters - `InMemoryLogExporter` and `InMemorySpanExporter`. 

API Changes
- The `InMemoryMetricExporter` now must be instanciated through `InMemoryMetricExporterBuilder` and it no longer has `Default` and `Clone` implementations.
- `InMemoryMetricExporterBuilder` has the mandatory `metrics: Arc<Mutex<Vec<_>>>` field to be set.

* [x] Unit tests ~~added/~~ updated

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Added `Resource::get_ref(&self, key: &Key) -> Option<&Value>` to allow retrieving a reference to a resource value without cloning.
 - **Breaking** Removed the following public hidden methods from the `SdkTracer` [#3227][3227]:
   - `id_generator`, `should_sample`
+- **Breaking** Removed `Default` and `Clone` implementation from `InMemoryMetricExporter`.
+- **Breaking** `InMemoryMetricExporterBuilder` requires mandatory `metrics` field to be set via
+  `.with_metrics` method.
 
 [3227]: https://github.com/open-telemetry/opentelemetry-rust/pull/3227
 
--- a/opentelemetry-sdk/src/metrics/in_memory_exporter.rs
+++ b/opentelemetry-sdk/src/metrics/in_memory_exporter.rs
@@ -114,7 +114,10 @@ impl Default for InMemoryMetricExporterBuilder {
 impl InMemoryMetricExporterBuilder {
     /// Creates a new instance of the `InMemoryMetricExporterBuilder`.
     pub fn new() -> Self {
-        Self { temporality: None, metrics: None }
+        Self {
+            temporality: None,
+            metrics: None,
+        }
     }
 
     /// Set the [Temporality] of the exporter.
--- a/opentelemetry-sdk/src/metrics/in_memory_exporter.rs
+++ b/opentelemetry-sdk/src/metrics/in_memory_exporter.rs
@@ -16,7 +16,7 @@ type InMemoryMetrics = Vec<ResourceMetrics>;
 /// An in-memory metrics exporter that stores metrics data in memory.
 ///
 /// This exporter is useful for testing and debugging purposes. It stores
-/// metric data in a user-provided `Vec<ResourceMetrics>`, from which the 
+/// metric data in a user-provided `Vec<ResourceMetrics>`, from which the
 /// exported data can be retrieved as well.
 ///
 /// # Panics
--- a/opentelemetry-sdk/src/metrics/in_memory_exporter.rs
+++ b/opentelemetry-sdk/src/metrics/in_memory_exporter.rs
@@ -28,6 +28,7 @@ type InMemoryMetrics = Vec<ResourceMetrics>;
 /// # Example
 ///
 /// ```
+///# use std::sync::{Arc, Mutex};
 ///# use opentelemetry_sdk::metrics;
 ///# use opentelemetry::{KeyValue};
 ///# use opentelemetry::metrics::MeterProvider;
@@ -36,15 +37,15 @@ type InMemoryMetrics = Vec<ResourceMetrics>;
 ///
 ///# #[tokio::main]
 ///# async fn main() {
-/// // Create an InMemoryMetricExporter
-///  let exporter = InMemoryMetricExporter::builder();
+///  // Create an InMemoryMetricExporter
 ///  let metrics = Arc::new(Mutex::new(Vec::new()));
-///  let exporter = InMemoryMetricExporterBuilder::new()
-///                .with_metrics(metrics.clone()).build();
+///  let exporter = InMemoryMetricExporter::builder()
+///      .with_metrics(metrics.clone())
+///      .build();
 ///
 ///  // Create a MeterProvider and register the exporter
 ///  let meter_provider = metrics::SdkMeterProvider::builder()
-///      .with_reader(PeriodicReader::builder(exporter.clone()).build())
+///      .with_reader(PeriodicReader::builder(exporter).build())
 ///      .build();
 ///
 ///  // Create and record metrics using the MeterProvider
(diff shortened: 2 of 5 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -564,43 +568,44 @@ mod tests {
...
-            if let Temporality::Cumulative = temporality {
-                assert_eq!(
-                    sum.temporality,
-                    Temporality::Cumulative,
-                    "Should produce cumulative"
-                );
-            } else {
-                assert_eq!(sum.temporality, Temporality::Delta, "Should produce delta");
-            }
+            test_context.with_aggregation::<u64>("my_observable_counter", None, |metric| {
```
scottgerring, line 567: You'll have to run cargo fmt on this!

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -3263,39 +3390,30 @@ mod tests {
...
+        fn with_aggregation<T: Number>(
             &mut self,
             counter_name: &str,
             unit_name: Option<&str>,
-        ) -> &MetricData<T> {
-            self.resource_metrics = self
-                .exporter
-                .get_finished_metrics()
-                .expect("metrics expected to be exported");
+            f: impl FnOnce(&MetricData<T>),
```
scottgerring, line 3405: Also generally I think the callback here is nice way of dealing with borrowing access and avoiding cloning to the stored data, but perhaps there is an option i'm missing. We eat a little bit of complexity in the impl here, but the usage is clean.

opentelemetry-sdk/src/metrics/in_memory_exporter.rs
```diff
@@ -121,48 +122,34 @@ impl InMemoryMetricExporterBuilder {
...
+        self.metrics = Some(metrics);
+        self
+    }
+
     /// Creates a new instance of the `InMemoryMetricExporter`.
     ///
     pub fn build(self) -> InMemoryMetricExporter {
         InMemoryMetricExporter {
-            metrics: Arc::new(Mutex::new(VecDeque::new())),
+            metrics: self.metrics.expect("Metrics must be provided"),
```
scottgerring, line 135: Should we return a `Result` here rather than paninicng?

opentelemetry-sdk/src/metrics/in_memory_exporter.rs
```diff
@@ -4,19 +4,20 @@ use crate::metrics::data::{
...
+// Not a user-facing type, just a type alias for clarity within this module.
+type InMemoryMetrics = Vec<ResourceMetrics>;
+
 /// An in-memory metrics exporter that stores metrics data in memory.
 ///
 /// This exporter is useful for testing and debugging purposes. It stores
-/// metric data in a `VecDeque<ResourceMetrics>`. Metrics can be retrieved
-/// using the `get_finished_metrics` method.
+/// metric data in a user-provided `Vec<ResourceMetrics>`, from which the
+/// exported data can be retrieved as well.
```
scottgerring, line 20: I think we should be more explicit here. Something like this perhaps?

```suggestion
  /// This exporter is useful for testing and debugging purposes. It stores
  /// metric data in a user-provided `Arc<Mutex<Vec<ResourceMetrics>>>`. Users
  /// retain a clone of this `Arc` and can access the exported metrics directly
  /// by locking the mutex.
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -271,11 +271,7 @@ mod tests {
             test_context.flush_metrics();
 
             // As instrument name are valid because of the feature flag, metrics should be exported
-            let resource_metrics = test_context
-                .exporter
-                .get_finished_metrics()
-                .expect("metrics expected to be exported");
-
+            let resource_metrics = test_context.resource_metrics.lock().unwrap();
```
scottgerring, line 274:
```suggestion
            let resource_metrics = test_context.resource_metrics.lock().expect("lock poisoned");
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -674,11 +681,10 @@ mod tests {
         meter_provider.force_flush().unwrap();
 
         // Assert
-        let resource_metrics = exporter
-            .get_finished_metrics()
-            .expect("metrics are expected to be exported.");
-        assert!(
-            resource_metrics[0].scope_metrics[0].metrics.len() == 1,
+        let resource_metrics = metrics.lock().unwrap();
```
scottgerring, line 684: I suggest using `expect("lock poisoned")` for the rest of these too - it captures intent and gives a clearer error message if it ever fails

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -3263,39 +3390,30 @@ mod tests {
...
 
-            assert!(
-                !self.resource_metrics.is_empty(),
-                "no metrics were exported"
-            );
+            assert!(!resource_metrics.is_empty(), "no metrics were exported");
 
-            assert!(
-                self.resource_metrics.len() == 1,
+            assert_eq!(
```
scottgerring, line 3411: Nice catch!

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -3263,39 +3390,30 @@ mod tests {
...
-                .get_finished_metrics()
-                .expect("metrics expected to be exported"); // TODO: Need to fix InMemoryMetricExporter to return None.
-
+            let resource_metrics = self.resource_metrics.lock().unwrap();
             assert!(resource_metrics.is_empty(), "no metrics should be exported");
         }
 
-        fn get_aggregation<T: Number>(
+        fn with_aggregation<T: Number>(
             &mut self,
```
scottgerring, line 3402: I don't think this needs a mutable ref anymore

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - Added `Resource::get_ref(&self, key: &Key) -> Option<&Value>` to allow retrieving a reference to a resource value without cloning.
 - **Breaking** Removed the following public hidden methods from the `SdkTracer` [#3227][3227]:
   - `id_generator`, `should_sample`
+- **Breaking** Removed `Default` and `Clone` implementation from `InMemoryMetricExporter`.
+- **Breaking** `InMemoryMetricExporterBuilder` requires mandatory `metrics` field to be set via
+  `.with_metrics` method.
```
scottgerring, line 10: I think we should phrase this so the user knows what the impact on them is. Something like this perhaps?:

```suggestion
 - **Breaking** Redesigned `InMemoryMetricExporter` to accept user-provided storage instead of managing it internally.
  Users now pass an `Arc<Mutex<Vec<ResourceMetrics>>>` via `InMemoryMetricExporterBuilder::with_metrics()` and access
exported data directly through that reference. This removes the need to clone the exporter and aligns with how
  production exporters receive their sink. `Default` and `Clone` implementations have been removed.
  [#3257](https://github.com/open-telemetry/opentelemetry-rust/pull/3257)


```

scottgerring (changes requested):
Hey @emadpres thanks for the PR! 🤝 
Some comments inline; lmk what you think

PR #3259, opened 2025-11-22
feat: Improve support for exponential histogram

### LINKED ISSUE #2111: [Feature]: Provide an example of exponential histogram.

The examples include explicit bucket histogram but not exponential histogram. It's not clear if one needs to create a metric and then an aggregation, or if directly creating the aggregation is feasible. @jtescher

thomasstauffer666:
Fixes #2111 and implements exponential histogram support in `opentelemetry-stdout`. 

## Changes


scottgerring:
@thomasstauffer666 let us know when this is ready for review!

scottgerring:
hey @thomasstauffer666 great - cheers! I'll check it out tomorrow 💪 

Apropos the stdout coverage; we try and be pragmatic with coverage, and my initial thought is that this is a case where we don't need to go wild chasing it, but I will form proper opinions shortly.

examples/metrics-advanced/src/main.rs
```diff
@@ -33,6 +33,22 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
         }
     };
 
+    // for example 3
```
scottgerring, line 36: It would be cool to be a bit more descriptive with the examples; "This is an exponential histogram, this is great for bucketed data that is exponential in nature" or something, I reckon. The example text on the others above is not so detailed, but this feels like it is introducing _nuance_, in that we have two examples of histograms now, and its not going to be obvious to all folks which to use when.

Likewise on the aggregation front down around L40-43, some brief comments on the tunables would be great; ultimately we want this all to feel overly descriptive / obvious, so there's less room for folks to waste time guessing at things!

opentelemetry-sdk/src/metrics/aggregation.rs
```diff
@@ -130,12 +130,12 @@ impl Aggregation {
             Aggregation::Base2ExponentialHistogram { max_scale, .. } => {
                 if *max_scale > EXPO_MAX_SCALE {
                     return Err(MetricError::Config(format!(
-                        "aggregation: exponential histogram: max scale ({max_scale}) is greater than 20",
+                        "aggregation: exponential histogram: max scale ({max_scale}) is greater than {}", EXPO_MAX_SCALE
```
scottgerring, line 133: nice catch!

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -266,6 +286,60 @@ fn print_hist_data_points<'a, T: Debug + Copy + 'a>(
...
+        let negative_bucket = data_point.negative_bucket();
+        let negative_offset = negative_bucket.offset();
+        println!("\t\t\tNegativeOffset : {}", negative_offset);
+        for (i, count) in negative_bucket.counts().enumerate() {
+            let from = -base.powf(i as f64 + negative_offset as f64);
+            let to = -base.powf(i as f64 + negative_offset as f64 + 1.0f64);
+            println!(
+                "\t\t\t\t ->  Bucket {} (> {:.3} to <= {:.3}) : {}",
+                i, from, to, count
+            );
```
scottgerring: I think this gets the labels backwards for negative `i` -  such that from > to

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -193,6 +193,26 @@ fn print_histogram<T: Debug + Copy>(histogram: &Histogram<T>) {
     print_hist_data_points(histogram.data_points());
 }
 
+fn print_exponential_histogram<T: Debug + Copy>(histogram: &ExponentialHistogram<T>) {
```
scottgerring, line 196: Could you update `opentelemetry-stdout/CHANGELOG.md` with a brief note about this?

scottgerring (changes requested):
Hey @thomasstauffer666 thanks for the PR! Couple of minor notes inline

p.s. if you would like to chat about this or other otel-rust _stuff_, you can join us in #opentelemetry-rust in the [CNCF slack](https://communityinviter.com/apps/cloud-native/cncf) 🤝

opentelemetry-stdout/src/metrics/exporter.rs
```diff
@@ -266,6 +286,66 @@ fn print_hist_data_points<'a, T: Debug + Copy + 'a>(
...
+            .collect::<Vec<_>>()
+            .into_iter()
+            .enumerate()
+            .rev()
+        {
+            let from = -base.powf(i as f64 + negative_offset as f64);
+            let to = -base.powf(i as f64 + negative_offset as f64 + 1.0f64);
+            println!(
+                "\t\t\t\t ->  Bucket {} (<= {:.3} to > {:.3}) : {}",
+                i, from, to, count
```
scottgerring: I struggle a bit with the format, but it might just be me:

> Bucket 0 (<= -2.000 to > -4.000) : 5"

Maybe something like is clearer?

> Bucket 0 ( -4.000 < x < -2.00) 

Also i'm not sure of the inclusivity ( x > y <= z ) - when it goes negative - did you have a look at the spec here?

thomasstauffer666: I looked up

https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram-bucket-inclusivity

But I will document it and put a reference to into the code and add an additional data point to the example where this is better visible.

It probably would be even better idea to access the boundaries via API. The official documentation goes into details how implementation may differ and even the reference implementation may not be fully accurate.

What about using mathematical notation?

			NegativeOffset : 0
				 ->  Bucket 4 (-5.657, -4.000] : 1
				 ->  Bucket 3 (-4.000, -2.828] : 1
				 ->  Bucket 2 (-2.828, -2.000] : 0
				 ->  Bucket 1 (-2.000, -1.414] : 0
				 ->  Bucket 0 (-1.414, -1.000] : 1
			PositiveOffset : -3
				 ->  Bucket 0 (0.354, 0.500] : 1
				 ->  Bucket 1 (0.500, 0.707] : 0
				 ->  Bucket 2 (0.707, 1.000] : 0
				 ->  Bucket 3 (1.000, 1.414] : 0
				 ->  Bucket 4 (1.414, 2.000] : 0
				 ->  Bucket 5 (2.000, 2.828] : 0
				 ->  Bucket 6 (2.828, 4.000] : 0
				 ->  Bucket 7 (4.000, 5.657] : 0
				 ->  Bucket 8 (5.657, 8.000] : 1

more obvious?

scottgerring: I think it's more obvious; although if we change it here we should also change the stdout formatter for the regular histogram for consistency. I don't know if there is some reason to _not_ do this - e.g. we expect folks to use this in bash-style grep-y pipelines and this breaks - feels unlikely, and i'd err for clarity. @cijothomas wdyt before we change more?

thomasstauffer666: For the regular histogram the example yields

				 0 to 5 : 2
				 5 to 10 : 0
				 10 to 25 : 0
				 25 to 50 : 0

IMHO somewhat ambiguous where e.g. a 5.0 would end up, but consistency between both histogram types is a good idea

cijothomas: stdout output is not formal/spec-ed out, and we can freely change them.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/stdout.md

examples/metrics-advanced/src/main.rs
```diff
@@ -112,6 +143,23 @@ async fn main() -> Result<(), Box<dyn Error + Send + Sync + 'static>> {
 
     histogram2.record(1.8, &[KeyValue::new("mykey1", "v7")]);
 
+    // Example 3 - Use exponential histogram.
+    let histogram3 = meter
+        .f64_histogram("my_third_histogram")
+        .with_description("My histogram example description")
+        .build();
+    histogram3.record(-1.3, &[KeyValue::new("mykey1", "v1")]);
```
cijothomas, line 151: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#record

-ve values are not expected to be provided, though we dont validate that!

See https://github.com/open-telemetry/opentelemetry-rust/pull/3260 for open PR/discussions.

Issue #3068, opened 2025-07-15
[Bug]: Jaeger shows OTLPResourceNoServiceName despite setting service.name resource


yasin-bozkurt:
### What happened?

After upgrading opentelemetry crates from version 0.23 to 0.30, we noticed that the service.name is no longer correctly propagated to Jaeger. Instead of the expected service.name, Jaeger displays the default OTLPResourceNoServiceName.

We define our tracer provider as follows:
```rust
use opentelemetry::{global, KeyValue};
use opentelemetry_sdk::{
    trace::SdkTracerProvider,
    Resource,
};
use opentelemetry_semantic_conventions::resource::{CONTAINER_IMAGE_TAGS, HOST_NAME, SERVICE_NAME};

impl<
        Runtime: opentelemetry_sdk::runtime::Runtime
            + opentelemetry_sdk::runtime::RuntimeChannel
            + std::fmt::Debug,
    > MySpanProcessor<Runtime>
{
    fn new<Exporter: SpanExporter + 'static>(
        exporter: Exporter,
        runtime: Runtime,
    ) -> Self {
        Self {
            otel_batch_span_processor: BatchSpanProcessor::builder(exporter, runtime).build(),
        }
    }
}

let exporter = opentelemetry_otlp::SpanExporter::builder()
    .with_tonic()
    .with_endpoint(endpoint)
    .build()?;

let my_span_processor = MySpanProcessor::new(exporter, Tokio);

let tracer_provider = SdkTracerProvider::builder()
    .with_span_processor(my_span_processor)
    .with_resource(
        Resource::builder_empty()
            .with_attributes(vec![
                KeyValue::new(SERVICE_NAME, service_name.clone()),
                KeyValue::new(HOST_NAME, host_name),
                KeyValue::new(CONTAINER_IMAGE_TAGS, image_tag),
            ])
            .build(),
    )
    .build();
let tracer_name = format!("{service_name}-tracer");
let tracer = tracer_provider.tracer(tracer_name);
let _ = global::set_tracer_provider(tracer_provider);
```

We have tried a couple of alternatives:

- Using Resource::builder().with_service_name(...) instead of a custom attribute list.
- Adding SdkProvidedResourceDetector (and setting the proper env) to the tracer provider builder.
- Explicitly logging the built resource to confirm service.name is present in the final resource.

### OpenTelemetry API Version (i.e version of `opentelemetry` crate)

0.30

[...]

PR #3266, opened 2025-11-25
docs: logs enrichment

### LINKED ISSUE #2759: Document (with example) how EnrichmentProcessor must be authored without sacrificing performance in Log pipeline
We recently tweaked [Logger.IsEnabled()](https://github.com/open-telemetry/opentelemetry-rust/pull/2756) capabilities to allow filtering decisions to be made based on name/scope(target),level.
This can greatly improve performance when us OS-native exporting (etw/user-events), as it allow us to determine that a particular Log is un-interesting, and can be skipped.

However, if a EnrichmentProcessor was also added, it will by default return `true` for the Enabled() check, forcing the Logger to anyway create the LogRecord, have it enriched by the EnrichmentProcessor, and then given to exporting processor, only to get it dropped there. To make this work without losing performance, the EnrichmentProcessor must wrap the exporting processor, and delegate calls to Enabled() to the underlying one.

Opening an issue to show an example on how to do this.

(Note: For users who want utmost performance, it can be argued that the enrichment can be done out-of-process, but sometimes, it is not feasible, if `Context` information is required.)

stefanobaghino:
Fixes #2759

## Changes

The approach suggested [here](https://github.com/open-telemetry/opentelemetry-rust/issues/2759#issuecomment-3547536632) by @scottgerring has been implemented:

> Something along these lines perhaps:
> ```
>  #[derive(Debug)]
>   pub struct EnrichmentProcessor<P: LogProcessor> {
>       /// The wrapped processor that will receive enriched log records
>       next: P,
>   }
> 
> impl<P: LogProcessor> LogProcessor for EnrichmentProcessor<P> {
>      #[cfg(feature = "spec_unstable_logs_enabled")]
>       fn event_enabled(&self, level: Severity, target: &str, name: Option<&str>) -> bool {
>           // Delegate to wrapped processor
>           self.next.event_enabled(level, target, name)
>       }
> 
> }
> ```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -16,7 +16,7 @@ Check this example if you want to understand *how to instrument logs using opent
 
 This example uses following crates from this repo:
 
--- opentelemetry(log)
+- opentelemetry(log)
 - opentelemetry-appender-tracing
 - opentelemetry-stdout
 


examples/logs-advanced/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct SlowEnrichmentLogProcessor<P: LogProcessor> {
+    /// The wrapped processor that will receive enriched log records
+    delegate: P,
+}
+
+impl<P: LogProcessor> SlowEnrichmentLogProcessor<P> {
```
cijothomas: I am not sure of the need for "slow" here. Just a normal EnrichmentProcessor is good enough.

examples/logs-advanced/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    fn force_flush(&self) -> OTelSdkResult {
+        self.delegate.force_flush()
+    }
+
+    #[cfg(feature = "spec_unstable_logs_enabled")]
+    fn event_enabled(&self, level: Severity, target: &str, name: Option<&str>) -> bool {
+        // It is important to call the delegate's event_enabled method to ensure that
+        // any filtering or logic implemented by downstream processors is respected.
+        // Skipping this call could result in logs being emitted that should have been filtered out
```
cijothomas: the key reason why this is important is performance.

examples/logs-enrichment/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "logs-enrichment"
```
cijothomas: this is better fit for a `logs-advanced` as this is not just enrichment...

stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/dfc718a651c98614f0b53a74f6aa30815ce4271e

examples/logs-advanced/src/main.rs
stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/e23c21bfe90cd53a881cddda8a7c56db0a984bd3

examples/logs-advanced/src/main.rs
stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/757a802a3f530bb18ed5e0516d710ff6c25d898b

stefanobaghino:
> Hey @stefanobaghino thanks for raising this! ACK Cijo's comments above; you'll also need to format, but otherwise seems good to me

Thanks for pointing that out to me, formatting addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/a446807bf178d22a1ce77bd3701232301a4110d6 (and I checked again after addressing @cijothomas's comments).

examples/logs-advanced/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+
+impl<P: LogProcessor> EnrichmentLogProcessor<P> {
+    pub fn new(delegate: P) -> EnrichmentLogProcessor<P> {
+        EnrichmentLogProcessor { delegate }
+    }
+}
+
+impl<P: LogProcessor> LogProcessor for EnrichmentLogProcessor<P> {
+    fn emit(&self, data: &mut SdkLogRecord, instrumentation: &InstrumentationScope) {
+        // Simulate an expensive enrichment step (e.g., fetching from a DB or service)
```
cijothomas: no need of this artificial slowness. processor are never designed to do anything like fetching from db, as they are run in hot path.

stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/50dba6da2c64830df2268674fedc31cfcddb6ca8

examples/logs-advanced/src/main.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    for i in 0..10 {
+        info!(name: "my-event-name", target: "my-system", event_id = 20 + i, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");
+    }
+    error!(name: "my-event-name", target: "my-system", event_id = 50, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");
+    let _ = provider.shutdown();
+}
+
+#[derive(Debug)]
+pub struct FilteringLogProcessor<P: LogProcessor> {
```
cijothomas: I am not sure if we should show FilteringProcessor. Filtering is best done in the the `tracing` layers itself, instead of relying on OTel. For now, it may be easier to show just the Enrichment part, which is something not super easy to achieve with just `tracing` and also closes the original issue.

stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/3c7623f269caf5e4b78bc805f418b9f249c3f0d1

examples/logs-advanced/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+[features]
+spec_unstable_logs_enabled = ["opentelemetry/spec_unstable_logs_enabled", "opentelemetry_sdk/spec_unstable_logs_enabled"]
+
+[dependencies]
+opentelemetry_sdk = { workspace = true, features = ["logs", "spec_unstable_logs_enabled"] }
+opentelemetry-stdout = { workspace = true, features = ["logs"] }
+opentelemetry-appender-tracing = { workspace = true, features = ["spec_unstable_logs_enabled"] }
+tracing = { workspace = true, features = ["std"]}
+tracing-subscriber = { workspace = true, features = ["env-filter","registry", "std", "fmt"] }
+opentelemetry = { version = "0.31.0", features = ["spec_unstable_logs_enabled"] }
```
cijothomas: lets do workspace version?

examples/logs-advanced/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+```shell
+cargo run --features="spec_unstable_logs_enabled"
+```
+
+Notice:
+1. only the error log is enriched with the `enriched: true` attribute
+2. the enrichment process only happens for the error log, without unnecessary work
+
+## Extra credit
```
cijothomas: misleading title?

examples/logs-advanced/README.md
```diff
@@ -0,0 +1,26 @@
...
+downstream processor's filtering is captured, avoiding unnecessary work.
+
+## Usage
+
+```shell
+cargo run --features="spec_unstable_logs_enabled"
+```
+
+Notice:
+1. only the error log is enriched with the `enriched: true` attribute
```
cijothomas: this is no longer true as we enrich all logs.

examples/logs-advanced/Cargo.toml
stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/90b9752749a2ac828e8c4a067d434c00d9a7c3fe

examples/logs-advanced/README.md
stefanobaghino: I don't understand the question or how the title would be misleading. Anyway, given the several rounds of feedback this section doesn't seem relevant anymore, so I just removed it in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/28f3cec088b4b948aaebbf3a637dfe1f213d9fb4.

examples/logs-advanced/README.md
stefanobaghino: Absolutely true, I removed the section since it's not relevant anymore. Done in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/e95656e64e96a334e7bbabc64a5d60cbcc4d1f36.

examples/logs-advanced/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl<P: LogProcessor> LogProcessor for EnrichmentLogProcessor<P> {
+    fn emit(&self, data: &mut SdkLogRecord, instrumentation: &InstrumentationScope) {
+        data.add_attribute("enriched", true);
+        self.delegate.emit(data, instrumentation);
+    }
+
+    fn force_flush(&self) -> OTelSdkResult {
+        self.delegate.force_flush()
```
cijothomas, line 96: just realized that we need to delegate all other calls as well. shutdown, set-resource etc. else the underlying processor won't know about it.

examples/logs-advanced/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+        .with(fmt_layer)
+        .init();
+
+    info!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");
+    error!(name: "my-event-name", target: "my-system", event_id = 50, user_name = "otel", user_email = "otel@opentelemetry.io", message = "This is an example message");
+    let _ = provider.shutdown();
+}
+
+#[derive(Debug)]
+pub struct EnrichmentLogProcessor<P: LogProcessor> {
```
cijothomas, line 78: this would be a good place to explain why Enrichment is done via a wrapping processor, instead of a normal processor. if done using normal processor (chaining), then performance advantages of IsEnabled check by the next processor is nullified as EnrichmentProcessor does not know how to deal with IsEnabled and returns true by default.

By wrapping it, we ensure that the underlying processor makes the important decision about Enabled, and it is honored. But wrapping means, this processor has to delegate every call to the underlying processor, including shutdown/flush etc.

(Some quick writing, need to polish.)

cijothomas: this would be a good follow up to also address some concerns raised in https://github.com/open-telemetry/opentelemetry-rust/issues/3068#issuecomment-3084275740

Let's keep the scope of this PR minimal and work on this in a future iteration. i.e lets make sure we delegate all calls to the delegate processor including shutdown and set_resource.

examples/logs-advanced/src/main.rs
stefanobaghino: Addressed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/00c08978772524abab38223dd1370b742d822f20

examples/logs-advanced/src/main.rs
stefanobaghino: I removed the comment specifically on `event_enabled` and added some documentation more in general about the implementation at the struct level in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/8da14103e96372b30252a6c1754092e80444881d

examples/logs-advanced/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+license = "Apache-2.0"
+rust-version = "1.75.0"
+publish = false
+autobenches = false
+
+[[bin]]
+name = "logs-advanced"
+path = "src/main.rs"
+bench = false
+required-features = ["spec_unstable_logs_enabled"]
```
cijothomas: we just merged a PR that removes this feature flag. The feature is stable and now does not require feature flags. Could you update to reflect it?

stefanobaghino: Merged with main in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/013c15f717e288bb78e4d5ab118f081b277c63e2, feature flag removed in https://github.com/open-telemetry/opentelemetry-rust/pull/3266/commits/4ee58a3ae0c94e8b1b8b794bc725393f19276a0f.

PR #3267, opened 2025-11-29
refactor: avoid span_data and  instrumentation_scope clone for every single span processor

### LINKED ISSUE #1125: Performance: send instrumentation library data to span processors once, not for every span when it ends
Similar to #1124, [sdk::trace::Span::build_export_data](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/trace/span.rs#LL236C9-L236C70) clones the tracer's instrumentation library for every span processor and every span before sending it to the span processor.

There is an opportunity to optimize this by assigning identifiers to each instrumentation library, unique within the provider associated with each tracer, and communicating (id, instrumentation library) to each span processor when the tracer is created.  Then, when each span ends, only the identifier of the instrumentation library needs to be sent to the span processor, rather than the full 120 bytes of instrumentation library data.

(I can work on a PR for this but wanted to hear of any objections here before starting that work.)

taisho6339:
Fixes #1125

## Changes

Changed `SpanProcessor::on_end()` signature to accept `&SpanData` and `&InstrumentationScope` by reference. This eliminates unnecessary `SpanData` clones when multiple `SpanProcessor`s are configured. Performance improvement scales with the number of processors. Processors that need to store the data asynchronously should clone it explicitly.

### DIFF
--- a/opentelemetry-sdk/src/trace/span.rs
+++ b/opentelemetry-sdk/src/trace/span.rs
@@ -219,24 +219,13 @@ impl Span {
             data.end_time = opentelemetry::time::now();
         }
 
-        match provider.span_processors() {
-            [] => {}
-            [processor] => {
-                processor.on_end(build_export_data(
-                    data,
-                    self.span_context.clone(),
-                    &self.tracer,
-                ));
-            }
-            processors => {
-                for processor in processors {
-                    processor.on_end(build_export_data(
-                        data.clone(),
-                        self.span_context.clone(),
-                        &self.tracer,
-                    ));
-                }
-            }
+        // Build export data once to avoid cloning for multiple processors
+        let export_data = build_export_data(data, self.span_context.clone(), &self.tracer);
+
+        let instrumentation_scope = self.tracer.instrumentation_scope();
+
+        for processor in provider.span_processors() {
+            processor.on_end(&export_data, instrumentation_scope);
         }
     }
 }
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -146,7 +146,12 @@ impl SpanProcessor for EnrichWithBaggageSpanProcessor {
         }
     }
 
-    fn on_end(&self, _span: opentelemetry_sdk::trace::SpanData) {}
+    fn on_end(
+        &self,
+        _span: &opentelemetry_sdk::trace::SpanData,
+        _instrumentation_scope: &opentelemetry::InstrumentationScope,
+    ) {
+    }
 }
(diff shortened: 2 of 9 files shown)


lalitb:
Thanks for the PR @taisho6339. Could you check if we can use a pattern similar to the one in the Logs SDK? In that implementation, the `InstrumentationScope` is passed by reference from the Logger to the Processor, allowing the Processor to decide whether to clone it or pass the reference to the exporter. I believe this approach was chosen to ensure the synchronous pipeline remains lightweight by avoiding cloning (even Arc overhead). This may lead to cloning of InstrumentationScope in case of batch processor, but that should also be cheap if the instrumentation-name is static string (which is the common use-case), and there are no attributes.

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -219,24 +219,13 @@ impl Span {
...
-                for processor in processors {
-                    processor.on_end(build_export_data(
-                        data.clone(),
-                        self.span_context.clone(),
-                        &self.tracer,
-                    ));
-                }
-            }
+        // Build export data once to avoid cloning for multiple processors
+        let export_data = build_export_data(data, self.span_context.clone(), &self.tracer);
```
cijothomas, line 223: `build_export_data` still clones the scope unnecessarily?

taisho6339: Yes, build_export_data() still clones InstrumentationScope once, but this is a deliberate trade-off to maintain SpanData's public API compatibility. The real performance gain comes from eliminating (N-1) clones of the entire SpanData (~120+ bytes plus variable-length fields) in Span::end() when multiple processors are configured. Removing the instrumentation_scope field from SpanData or passing the reference entirely would be a larger breaking change that could be considered in a future PR.

cijothomas: We are okay to make breaking changes to Span/Tracing given its not marked stable. However, let's have a full analysis to this area before making the changes. There are few discussions in https://github.com/open-telemetry/opentelemetry-rust/pull/2962 to be considered here.

(In general, yes, passing immutable ref to on_end makes total sense- just want to make sure we have the desired end state in mind before the implementation)

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/2726 related issue.

bantonsson: Yes, please take a look at #2962 @taisho6339 and come with input there.

PR #3268, opened 2025-11-30
chore: add clo monitor exemption for artifact hub

thompson-tomo:
## Changes

This defines a .clomonitor.yml file which is used by [clo monitor](https://clomonitor.io/projects/cncf/open-telemetry).

This file contains an exemption for artifact hub as artifact's produced by this library are not supported by artifact hub. By including the exemption we can improve the score for this project and the organisation. File contents have been copied from other repos where the exemption has been applied.

### DIFF
--- a/.clomonitor.yml
+++ b/.clomonitor.yml
@@ -1,0 +1,4 @@
+# see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions
+exemptions:
+  - check: artifacthub_badge
+    reason: "Artifact Hub doesn't support rust packages"


lalitb:
> Seems like the FOSSA badge should be straightforward enough but i'm not particularly familiar with this!

I think it would be easier to fix for OpenSSF best practices badge by embedding the badge in README.md as mentioned here - https://www.bestpractices.dev/en/projects/10394.

But this PR should hold good for artifact hub if that doesn't support Rust. I approved a similar PR for otel-cpp :)

thompson-tomo:
I was going to follow up with a pr to add additional badges to the readme ie the best practices badge which will further improve the score.

I would suggest looking at the license scanning topic seperately by seeing what other repos have done that way all but the security area is 100% which is a common state.

PR #3280, opened 2025-12-05
chore(opentelemetry-otlp): establish priority order for otlp clients

### LINKED ISSUE #2994: [Bug]: NoHttpClient error when reqwest-client feature is enabled

scottgerring:
Fixes #2994 using Paul's suggestion to implement a priority ordering for the HTTP clients when multiple are enabled.

## Changes


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -212,22 +212,14 @@ impl HttpExporterBuilder {
         #[allow(unused_mut)] // TODO - clippy thinks mut is not needed, but it is
         let mut http_client = self.http_config.client.take();
 
+        // When multiple HTTP client features are enabled, we use a priority order
+        // to select the client. This follows Rust's feature unification principle
+        // where features should be additive. Priority (highest to lowest):
+        // 1. reqwest-client (async)
```
cijothomas, line 218: how is this priority decided? The default we use is the blocking-client (since the detail is the dedicated thread model..). But this order is reverse...

scottgerring: My thinking is that when you explicitly turn on another client feature, that feature should win, to match the rust principal that features are additive. 

Because blocking is the default feature if it appears first in the fallthrough here users wouldn’t be able to simply toggle on an alternative client feature, they’d have to also turn off default. As Leo points out in the issue this’d be counterintuitive. WDYT?

PR #3282, opened 2025-12-05
feat(appender-tracing): add experimental span attributes enrichment

### LINKED ISSUE #3221: Add tracing span support to tracing events in opentelemetry-appender-tracing
`opentelemetry-appender-tracing` crate currently converts `tracing` events to OpenTelemetry `LogRecord`s but ignores span context. We should include span attributes in the `LogRecord` attributes when events are emitted within an active span (`tracing` span).

**Current behavior:**

```rust
let span = info_span!("request", user_id = 123, endpoint = "/api/users");
let _guard = span.enter();
info!(status = 200, "Request completed");
// LogRecord only contains: status = 200, message = "Request completed"
```

**Desired behavior:**
```rust
let span = info_span!("request", user_id = 123, endpoint = "/api/users");
let _guard = span.enter();
info!(status = 200, "Request completed");
// LogRecord should contain: status = 200, message = "Request completed", user_id = 123, endpoint = "/api/users"
```

This is distinct from the `tracing-opentelemetry` crate which converts tracing spans to OTel spans. This issue is specifically about enriching log records with span context from `tracing`.

leghadjeu-christian:
Fixes #3221

## Changes
Add experimental `experimental_span_attributes` feature to enrich log records with attributes from active tracing spans.
When enabled, span attributes are automatically included in log records. Event attributes take precedence on conflicts.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -15,10 +15,12 @@
   - `SamplingDecision`, `SamplingResult`
   - These types are SDK implementation details and should be imported from `opentelemetry_sdk::trace` instead.
 - Fix panics and exploding memory usage from large cardinality limit [#3290][3290]
+- Fix Histogram boundaries being ignored in the presence of views [#3312][3312]
 
 [3227]: https://github.com/open-telemetry/opentelemetry-rust/pull/3227
 [3277]: https://github.com/open-telemetry/opentelemetry-rust/pull/3277
 [3290]: https://github.com/open-telemetry/opentelemetry-rust/pull/3290
+[3312]: https://github.com/open-telemetry/opentelemetry-rust/pull/3312
 
 - "spec_unstable_logs_enabled" feature flag is removed. The capability (and the
   backing specification) is now stable and is enabled by default.
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -438,6 +438,14 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn histogram_aggregation_with_custom_bounds_and_view() {
+        // Run this test with stdout enabled to see output.
+        // cargo test histogram_aggregation_with_custom_bounds_and_view --features=testing -- --nocapture
+        histogram_aggregation_with_custom_bounds_and_view_helper(Temporality::Delta);
+        histogram_aggregation_with_custom_bounds_and_view_helper(Temporality::Cumulative);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn updown_counter_aggregation_cumulative() {
         // Run this test with stdout enabled to see output.
         // cargo test updown_counter_aggregation_cumulative --features=testing -- --nocapture
@@ -2533,6 +2541,77 @@ mod tests {
         assert_eq!(data_point.sum, 15);
         assert!(data_point.bounds.is_empty());
         assert!(data_point.bucket_counts.is_empty());
+    }
+
+    fn histogram_aggregation_with_custom_bounds_and_view_helper(temporality: Temporality) {
+        for specify_boundaries_in_view in [false, true] {
+            let view = move |_: &Instrument| {
+                let mut builder = Stream::builder();
+                if specify_boundaries_in_view {
+                    builder = builder.with_aggregation(Aggregation::ExplicitBucketHistogram {
+                        boundaries: vec![1.5, 4.2, 6.7],
+                        record_min_max: true,
+                    });
+                }
+                Some(builder.build().unwrap())
+            };
+            let mut test_context = TestContext::new_with_view(temporality, view);
+            let histogram = test_context
+                .meter()
+                .u64_histogram("test_histogram")
+                .with_boundaries(vec![1.0, 2.5, 5.5])
+                .build();
+            histogram.record(1, &[KeyValue::new("key1", "value1")]);
+            histogram.record(2, &[KeyValue::new("key1", "value1")]);
+            histogram.record(3, &[KeyValue::new("key1", "value1")]);
+            histogram.record(4, &[KeyValue::new("key1", "value1")]);
+            histogram.record(5, &[KeyValue::new("key1", "value1")]);
+
+            test_context.flush_metrics();
+
+            let MetricData::Histogram(histogram_data) =
+                test_context.get_aggregation::<u64>("test_histogram", None)
+            else {
+                unreachable!()
+            };
+            assert_eq!(histogram_data.data_points.len(), 1);
+            if let Temporality::Cumulative = temporality {
+                assert_eq!(
+                    histogram_data.temporality,
+                    Temporality::Cumulative,
+                    "Should produce cumulative"
+                );
+            } else {
+                assert_eq!(
+                    histogram_data.temporality,
+                    Temporality::Delta,
+                    "Should produce delta"
+                );
+            }
+
+            // find and validate key1=value1 datapoint
+            let data_point = find_histogram_datapoint_with_key_value(
+                &histogram_data.data_points,
+                "key1",
+                "value1",
+            )
+            .expect("datapoint with key1=value1 expected");
+
+            assert_eq!(data_point.count, 5);
+            assert_eq!(data_point.sum, 15);
+
+            // Check the bucket counts
+            if specify_boundaries_in_view {
+                // If boundaries are specified in the view, they should take precedence
+                assert_eq!(vec![1.5, 4.2, 6.7], data_point.bounds);
+                assert_eq!(vec![1, 3, 1, 0], data_point.bucket_counts);
+            } else {
+                // If boundaries are not specified in the view, the ones from the instrument
+                // should be used
+                assert_eq!(vec![1.0, 2.5, 5.5], data_point.bounds);
+                assert_eq!(vec![1, 1, 3, 0], data_point.bucket_counts);
+            }
+        }
     }
 
     fn gauge_aggregation_helper(temporality: Temporality) {
--- a/opentelemetry-sdk/src/metrics/pipeline.rs
+++ b/opentelemetry-sdk/src/metrics/pipeline.rs
@@ -279,6 +279,16 @@ where
                 stream.unit = Some(inst.unit.clone());
             }
 
+            // Override default histogram boundaries if provided.
+            if let Some(boundaries) = boundaries {
+                if kind == InstrumentKind::Histogram && stream.aggregation.is_none() {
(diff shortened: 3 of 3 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -246,6 +344,28 @@ where
         //emit record
         self.logger.emit(log_record);
     }
+    #[cfg(feature = "experimental_span_attributes")]
+    fn on_new_span(
```
cijothomas, line 375: do we need to do anything to remove this from the span extensions, when span is being exited? Or its auto removed by `tracing` itself?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -246,6 +344,28 @@ where
...
+        &self,
+        attrs: &tracing::span::Attributes<'_>,
+        id: &tracing::span::Id,
+        ctx: tracing_subscriber::layer::Context<'_, S>,
+    ) {
+        let span = ctx.span(id).expect("Span not found; this is a bug");
```
cijothomas, line 381: is it a bug in OTel side? or `tracing` ?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -821,4 +941,173 @@ mod tests {
         // The validation is done in the LogProcessorWithIsEnabled struct.
         error!(name: "my-event-name", target: "my-system", event_id = 20, user_name = "otel", user_email = "otel@opentelemetry.io");
     }
+
+    #[test]
+    #[cfg(feature = "experimental_span_attributes")]
+    fn test_span_context_enrichment_enabled() {
```
cijothomas: can you add a test for nested spans? I am curious on how that works as well...

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -237,6 +320,21 @@ where
...
         log_record.set_severity_number(severity);
         log_record.set_severity_text(metadata.level().as_str());
+
+        // Extract span attributes if feature is enabled
+        #[cfg(feature = "experimental_span_attributes")]
+        {
+            if let Some(span) = ctx.event_span(event) {
+                let extensions = span.extensions();
+                if let Some(stored) = extensions.get::<StoredSpanAttributes>() {
+                    // Add span attributes first (event attributes will overwrite on conflict). TODO: Consider conflict resolution strategies.
```
cijothomas: Can you point me to where is the overwrite logic? OTel SDK itself does not do any overwrite...

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -178,6 +178,92 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
+    }
+
+    fn record_str(&mut self, field: &tracing::field::Field, value: &str) {
+        self.fields.insert(
+            field.name().to_string(),
+            AnyValue::String(value.to_owned().into()),
+        );
+    }
+
+    fn record_i64(&mut self, field: &tracing::field::Field, value: i64) {
```
cijothomas, line 231: are the field names `&'static str` ? Can we leverage that and avoid `String` which forces a heap allocation for each attribute?

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/common.rs#L16-L49
We'd avoid allocation if `&'static str` is used. Let's do it. Its already done in the event visitor.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -178,6 +178,92 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
 
+#[cfg(feature = "experimental_span_attributes")]
+use std::collections::HashMap;
+#[cfg(feature = "experimental_span_attributes")]
+use std::sync::Arc;
+
+/// Visitor to extract fields from a tracing span
+#[cfg(feature = "experimental_span_attributes")]
+struct SpanFieldVisitor {
+    fields: HashMap<String, AnyValue>,
```
cijothomas: Instead of using `String` as the key, you should use OTel's `Key`.. that should avoid the heap allocations.

opentelemetry-appender-tracing/src/layer.rs
leghadjeu-christian: No, the Span extensions are automatically cleaned up when the span is closed. The `tracing-subscriber` Registry uses a pooled slab allocator for span storage, and when a span is removed from the pool [here](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/registry/sharded.rs#L514-L523) in [`DataInner::clear`], its `DataInner::clear()` method is called, which explicitly clears the extensions HashMap (dropping all stored data).


 The cleanup is triggered by the `CloseGuard::drop` implementation in the same file ([here](https://github.com/tokio-rs/tracing/blob/master/tracing-subscriber/src/registry/sharded.rs#L387-L409)), which calls `self.registry.spans.clear(id_to_idx(&self.id))`, invoking the `Clear` trait implementation above.

opentelemetry-appender-tracing/src/layer.rs
leghadjeu-christian: It is a bug on the `tracing` side, the `on_new_span` is only called after the span has been successfully allocated in the Registry, hence attributing it an `id`. This is ensured  by the implementation [here](https://github.com/tokio-rs/tracing/blob/9feb241133e55e70c7d4399689b8ef72f71d070f/tracing-subscriber/src/layer/layered.rs#L129C1-L134C6) in `Layered::new_span` , where `Registry::new_span()` completes before `Layer::on_new_span()` is invoked.

   ```rust
   fn new_span(&self, span: &span::Attributes<'_>) -> span::Id {
       let id = self.inner.new_span(span);        // 1. Registry allocates
       self.layer.on_new_span(span, &id, ...);     // 2. Layer notified
       id
   }
   ```
So the `on_new_span()` function is only called if a new span is created(hence having an `id`).
The only way this could panic is if there's a bug in the `tracing` or `tracing-subscriber` crate itself.

opentelemetry-appender-tracing/src/layer.rs
leghadjeu-christian: Thanks for the catch! You're absolutely right. I had a moment of confusion thinking the SDK used a HashMap for attributes, but it actually uses a 
GrowableArray (Vec-like structure).

opentelemetry-appender-tracing/src/layer.rs
leghadjeu-christian: I added a new test  `test_nested_spans_use_innermost_attributes`
 that verifies nested span behavior. It confirms that events correctly associate with the innermost active span and only include its attributes, ensuring proper scoping.

cijothomas: why we restrict to innermost active span? That'd mean we lose everything from non-innermost span, partially defeating the purpose. We should get attributes from all spans.

We can look at how `fmt` subscriber handles this, and mimic its behavior closely.

leghadjeu-christian: Yeah @cijothomas thanks for the clarification, I  updated the implementation to collect attributes from all parent spans using `ctx.event_scope(event)` instead of `ctx.event_span(event)` that was just for the current innermost span context and `scope.from_root()`(to iterate over all the  spans from root to leaf), matching the behavior of the `fmt`. Added `tracing_appender_nested_spans_collect_all_parent_attributes` test to verify attributes from the entire span hierarchy are included.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -237,6 +337,31 @@ where
...
+        {
+            // Collect attributes from all parent spans (root to leaf), including current span
+            if let Some(scope) = ctx.event_scope(event) {
+                for span_ref in scope.from_root() {
+                    // Clone Arc to release span extensions lock before processing attributes
+                    let attrs = {
+                        let extensions = span_ref.extensions();
+                        extensions
+                            .get::<StoredSpanAttributes>()
+                            .map(|stored| stored.attributes.clone())
```
bantonsson: nit:
```suggestion
                        span_ref
                            .extensions()
                            .get::<StoredSpanAttributes>()
                            .map(|stored| stored.attributes.clone())
```

opentelemetry-appender-tracing/benches/span-attributes.rs
```diff
@@ -0,0 +1,260 @@
+/*
+// Run this benchmark with:
+ // cargo bench --bench span-attributes --features experimental_span_attributes
+ // The benchmark results:
+ // Hardware: Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz (4 cores)
+
+| Test                      | Average time | Increment |
+|---------------------------|--------------|-----------|
+| span_0_attributes         | 146 ns       | -         |
```
cijothomas: we just need test with 4 and 8 attributes each. More importantly, let's add one with nesting (1,2 and 3 level depth, each with maybe 4 attributes)

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -246,6 +371,28 @@ where
...
+    #[cfg(feature = "experimental_span_attributes")]
+    fn on_new_span(
+        &self,
+        attrs: &tracing::span::Attributes<'_>,
+        id: &tracing::span::Id,
+        ctx: tracing_subscriber::layer::Context<'_, S>,
+    ) {
+        let span = ctx.span(id).expect("Span not found; this is a bug");
+
+        let mut visitor = SpanFieldVisitor::new();
```
cijothomas, line 383: `attrs` has metadata which can tell us how many fields are there maximum - we can leverage that and allocate HashMap with correct capacity to avoid reallocation.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -178,6 +178,109 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
+#[cfg(feature = "experimental_span_attributes")]
+struct SpanFieldVisitor {
+    fields: HashMap<Key, AnyValue>,
+}
+
+#[cfg(feature = "experimental_span_attributes")]
+impl SpanFieldVisitor {
+    fn new() -> Self {
+        Self {
+            fields: HashMap::new(),
```
cijothomas, line 196: we can do a vector here instead of hash map. And let sdk handles duplicate handling. (which it does not do today, and that is a violation we need to address separately)

cijothomas (approved):
LGTM. I left couple of suggestions, which should improve performance, and can be followed up separately.

Thanks for amazing work getting this in!

lalitb:
Also as a followup - The layer only captures attrs in `on_new_span`; any later `span.record(...)` calls never reach us because `on_record` isn’t implemented.

PR #2491, opened 2025-01-01
[WIP] fix serialization/deserialization for u64 and "NaN" float fields in Metrics

lalitb:
Fixes #2487 
Design discussion issue (if applicable) #

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #3284, opened 2025-12-11
fix: serialization/deserialization for u64, f64, and "NaN" float fields in Metrics

### LINKED ISSUE #2487: Serialization Fails for u64 Fields Due to Incoming String Values in OpenTelemetry Metrics
I am using the opentelemetry-proto crate to serialize incoming OpenTelemetry metrics data. The crate defines certain fields (`count`, `bucket_counts` under `HistogramDataPoint`) as u64. However, the OpenTelemetry exporter sends these fields as strings, which leads to serialization failures.

This mismatch between the data types in the crate and the incoming data prevents successful serialization of metrics.

Sample JSON 
[otel-metrics-histogram.json](https://github.com/user-attachments/files/18275602/otel-metrics-histogram.json)

Similar issue is found in the `ExponentialHistogramDataPoint`

parmesant:
Fixes #2487 

## Changes

This PR adds to the changes made by @lalitb in his PR #2491 

For metric type `Summary`, `quantileValues.quantile` gets treated as u64 and not f64 when its value is 1 

Also, `flags` comes optionally in the payload hence added default to it (and some other fields)

### DIFF
--- a/opentelemetry-proto/src/proto.rs
+++ b/opentelemetry-proto/src/proto.rs
@@ -170,8 +170,39 @@ pub(crate) mod serializers {
     where
         D: Deserializer<'de>,
     {
-        let s: String = Deserialize::deserialize(deserializer)?;
-        s.parse::<u64>().map_err(de::Error::custom)
+        struct U64Visitor;
+
+        impl<'de> de::Visitor<'de> for U64Visitor {
+            type Value = u64;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("a u64 integer or a string containing a u64 integer")
+            }
+
+            fn visit_u64<E>(self, value: u64) -> Result<u64, E>
+            where
+                E: de::Error,
+            {
+                Ok(value)
+            }
+
+            fn visit_i64<E>(self, value: i64) -> Result<u64, E>
+            where
+                E: de::Error,
+            {
+                u64::try_from(value)
+                    .map_err(|_| E::custom(format!("i64 value {} is out of range for u64", value)))
+            }
+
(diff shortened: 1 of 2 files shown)


parmesant:
@lalitb let me know if there is anything else required for this PR

opentelemetry-proto/src/proto.rs
```diff
@@ -213,6 +211,95 @@ pub(crate) mod serializers {
         let s: String = Deserialize::deserialize(deserializer)?;
         s.parse::<i64>().map_err(de::Error::custom)
     }
+    pub fn serialize_vec_u64_to_strings<S>(vec: &[u64], serializer: S) -> Result<S::Ok, S::Error>
```
lalitb: @parmesant  - I understand this could be coming from the previous PR I raised, but see if we can use the already existing methods:

  - serialize_vec_u64_to_string                                                                                                                                    
  - deserialize_vec_string_to_vec_u64

parmesant: @lalitb thanks for the suggestion. These helper functions seem to be working just fine.

parmesant:
@lalitb I have made the changes that you suggested

lalitb:
@parmesant  Sorry for late reply. We are close, with two remaining issues:

- The u64 deserializer (`deserialize_string_to_u64`) only accepts quoted strings - a bare "count": 2 will fail because serde_json can't deserialize a JSON number into String. The proto3 JSON spec says parsers must accept both 2 and "2" for 64-bit integers. This is actually a regression since count previously worked with bare numbers via default serde.

- Also worth noting - the f64 `visit_str` only matches "NaN", "Infinity", and "-Infinity", so a quoted number like "0.5" would be rejected. Not strictly required by the spec, but might be worth handling for robustness.

PR #3290, opened 2025-12-20
fix: Panics and exploding memory usage from large cardinality limit

ArniDagur:
### Problem

Disabling the cardinality limit by setting it to `usize::MAX` caused the following issues for us:

1. Integer overflow:
```
thread '<unnamed>' (8159738) panicked at /Users/arni/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/opentelemetry_sdk-0.31.0/src/metrics/internal/mod.rs:82:58:
attempt to add with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_add_overflow
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:175:17
   3: opentelemetry_sdk::metrics::internal::ValueMap<A>::new
             at /Users/arni/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/opentelemetry_sdk-0.31.0/src/metrics/internal/mod.rs:82:58
   4: opentelemetry_sdk::metrics::internal::last_value::LastValue<T>::new
             at /Users/arni/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/opentelemetry_sdk-0.31.0/src/metrics/internal/last_value.rs:59:24
   5: opentelemetry_sdk::metrics::internal::aggregate::AggregateBuilder<T>::last_value
             at /Users/arni/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/opentelemetry_sdk-0.31.0/src/metrics/internal/aggregate.rs:148:9
```

2. Hash table capacity overflow
```

thread 'main' (8171305) panicked at /rust/deps/hashbrown-0.15.5/src/raw/mod.rs:37:40:
Hash table capacity overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
```

3. Exploding memory usage if we always pre-allocate for maximum cardinality limit usage

### Solution

Only preallocate a portion, based on the default cardinality limit, and let the hashmaps resize if needed

### Merge requirement checklist

* [x] Changes in public API reviewed

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -6,6 +6,7 @@ mod precomputed_sum;
 mod sum;
 
 use core::fmt;
+use std::cmp::min;
 use std::collections::{HashMap, HashSet};
 use std::mem::swap;
 use std::ops::{Add, AddAssign, DerefMut, Sub};
@@ -80,7 +81,9 @@ where
 {
     fn new(config: A::InitConfig, cardinality_limit: usize) -> Self {
         ValueMap {
-            trackers: RwLock::new(HashMap::with_capacity(1 + DEFAULT_CARDINALITY_LIMIT)),
+            trackers: RwLock::new(HashMap::with_capacity(
+                1 + min(DEFAULT_CARDINALITY_LIMIT, cardinality_limit),
+            )),
             trackers_for_collect: OnceLock::new(),
             has_no_attribute_value: AtomicBool::new(false),
             no_attribute_tracker: A::create(&config),
@@ -92,8 +95,11 @@ where
 
     #[inline]
     fn trackers_for_collect(&self) -> &RwLock<HashMap<Vec<KeyValue>, Arc<A>>> {
-        self.trackers_for_collect
-            .get_or_init(|| RwLock::new(HashMap::with_capacity(1 + DEFAULT_CARDINALITY_LIMIT)))
+        self.trackers_for_collect.get_or_init(|| {
+            RwLock::new(HashMap::with_capacity(
+                1 + min(DEFAULT_CARDINALITY_LIMIT, self.cardinality_limit),
+            ))
+        })
     }
 
     /// Checks whether aggregator has hit cardinality limit for metric streams


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -79,7 +80,7 @@ where
 {
     fn new(config: A::InitConfig, cardinality_limit: usize) -> Self {
         ValueMap {
-            trackers: RwLock::new(HashMap::with_capacity(1 + cardinality_limit)),
+            trackers: RwLock::new(HashMap::with_capacity(1 + DEFAULT_CARDINALITY_LIMIT)),
```
cijothomas: this causes overallocation is user sets a low cardinality limit like 100 and we reserve for 2000. We could take the min(cardinality_limit, DEFAULT_LIMIT)

ArniDagur: Fixed

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -91,8 +95,11 @@ where
 
     #[inline]
     fn trackers_for_collect(&self) -> &RwLock<HashMap<Vec<KeyValue>, Arc<A>>> {
-        self.trackers_for_collect
-            .get_or_init(|| RwLock::new(HashMap::with_capacity(1 + self.cardinality_limit)))
+        self.trackers_for_collect.get_or_init(|| {
+            RwLock::new(HashMap::with_capacity(
+                1 + min(DEFAULT_CARDINALITY_LIMIT, self.cardinality_limit),
```
utpilla, line 100: @cijothomas @ArniDagur 

The problem mentioned in the PR is about code panicking when cardinality limit is set to `usize::MAX`. The solution offered in this PR seems problematic:
1. The core issue is not pre-allocation of `usize::MAX` elements in the `HashMap`. In fact, the input `usize::MAX` is invalid since we reserve one spot in the HashMap capacity for tracking overflows. We should either remove that reservation and track overflows outside of the HashMap similar to zero attributes tracker or we should validate the user input during `Stream` build so that we can disallow this input upfront.
2. The changes in this PR violate our design principle of avoiding allocation on hot-path when possible.

cijothomas: Excellent point! 
Agree. I was thinking that for new time-series, allocation is anyway required (to clone the key-value pairs) so adding slightly more is acceptable. But I totally missed that the resizing of hash-map could be very very expensive (much higher than the allocation needed to clone kv pairs).
As the hash-map is retained, the resize operations become increasing rare and reaches zero once we reach enough capacity to hold cardinality_limit.

Thinking more on this, I think we should revert to the original behavior of preallocation. (and of course add stream validation to avoid the overflow.)

For folks affected by the continuous use of memory accounting for the worst case scenario - a better fix would be to have an opt-in mode, where the hash map can grow (and also shrink) dynamically - This'd require some more rearchitecting. (its best to tackle it along with https://github.com/open-telemetry/opentelemetry-rust/issues/2328)

(I can't find the discussion where some users were concerned about the current strategy of allocating cardinality_limit memory all the time, as users added high cardinality limit, to account for extremely rare spikes only)

ArniDagur: > As the hash-map is retained, the resize operations become increasing rare and reaches zero once we reach enough capacity to hold cardinality_limit.

This was my reasoning as well. In that case, why is it an issue to pay the one time cost of resizes at the start of the program, while it discovers a meter's true cardinality?

Preallocating for the entire cardinality limit is a footgun IMO. I appreciate you addressing our use case and adding the option of both static and dynamic allocation, but I question the wisdom of the added (maintainer- and user-) complexity of maintaining two different options, for what I'm not sure is any added benefit.

PR #3293, opened 2025-12-31
build: upgrade reqwest to 0.13

tisonkun:
See https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0.

Prepare for any upstream usage :D


houseme (approved):

joshka-oai:
Thanks for getting this merged, really appreciate the work here 🙏

I am a downstream user of opentelemetry-rust (not the PR author), and this unblocks us from moving to reqwest 0.13 on our side. I just wanted to ask, purely for planning, do you have a rough sense of whether this is likely to make it into the next release, or if the release cadence is a bit longer right now?

No pressure at all. We are happy to wait either way. Just trying to get a feel for when we might be able to consume it.

PR #3307, opened 2026-01-09
perf: Benchmarks for OTLP conversion 

cijothomas:
Mostly to get a baseline of where is time being spent. That should help guide future perf optimizations.


opentelemetry-sdk/src/logs/export.rs
```diff
@@ -52,7 +52,18 @@ impl<'a> LogBatch<'a> {
...
+    /// # Arguments
+    ///
+    /// * `data` - A slice of boxed tuples, where each tuple consists of owned `SdkLogRecord`
+    ///   and `InstrumentationScope`. These tuples represent the log records and their
+    ///   associated instrumentation scopes to be exported.
+    ///
+    /// # Returns
+    ///
+    /// A `LogBatch` instance containing the provided log records and instrumentation scopes.
+    pub fn new_with_owned_data(
```
scottgerring: Moving this to `pub` seems notable; I note also that the `new` just above suggests a reduction of scope in the future:

```rust
    /// Note - this is not a public function, and should not be used directly. This would be
    /// made private in the future.
```

Can we restructure the benchmarks to avoid this?

opentelemetry-otlp/benches/logs_export.rs
```diff
@@ -0,0 +1,568 @@
...
+/// - Emits 511 logs per iteration to avoid auto-export, making force_flush the sole trigger
+///
+/// Despite these differences, this provides a reasonable approximation of the
+/// end-to-end cost for a single batch of logs through the full pipeline.
+fn bench_e2e_with_batch_processor(c: &mut Criterion) {
+    // Create runtime for async operations
+    let rt = Runtime::new().unwrap();
+
+    // Start fake OTLP server on a different port to avoid conflicts
+    let port = 14319;
```
scottgerring: Bit nit-y, but could we not bind to 0/give-me-a-port, then extract the bound port?

opentelemetry-proto/benches/logs.rs
```diff
@@ -0,0 +1,276 @@
...
+use opentelemetry_proto::transform::common::tonic::ResourceAttributesWithSchema;
+
+#[cfg(feature = "gen-tonic-messages")]
+use opentelemetry_sdk::logs::LogBatch;
+
+#[cfg(feature = "gen-tonic-messages")]
+use prost::Message;
+
+#[allow(clippy::vec_box)]
+fn create_log_batch(
```
scottgerring, line 58: Can we remove the copy-pasta between `logs_export.rs` and here?

opentelemetry-otlp/Cargo.toml
```diff
@@ -56,6 +61,8 @@ futures-util = { workspace = true }
 temp-env = { workspace = true }
 tonic = { workspace = true, features = ["router", "server"] }
 async-trait = { workspace = true }
+criterion = { version = "0.5", features = ["async_tokio", "html_reports"] }
+hyper = { version = "0.14", features = ["server", "http1", "tcp"] }
```
scottgerring: Hyper 0.14 is rather old; if this intentional we should probably document it

scottgerring (changes requested):

opentelemetry-sdk/src/logs/export.rs
cijothomas: The ship has sailed for making the above private. Both are okay to be public, as that is required for custom exporting processors to prepare data in format expected by exporter trait.

I'll update the comment and add changelog entry for this change.

opentelemetry-proto/benches/logs.rs
cijothomas: both are in different crates, so not super straightforward to avoid duplication.

opentelemetry-otlp/Cargo.toml
cijothomas: let me bump to newest and re-run benches.

scottgerring:
Hey @cijothomas had a quick look at this because of the benchmark failures - I _think_ what is going on is that `opentelemetry-prometheus` is using its own benchmark harness, which is not criterion but rather the normal rust one. So it doesn't understand `--save-baseline`. I think we should line that crate up with the criterion usage otherwise to get it going again. I think this is also going to break on `main` for the period checking as it stands.

PR #3308, opened 2026-01-10
feat: Metrics SDK to offer dynamic memory resizing for delta aggregations

cijothomas:
Addresses https://github.com/open-telemetry/opentelemetry-rust/pull/3290#discussion_r2677338361 , reverting the original fix.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -6,7 +6,6 @@ mod precomputed_sum;
 mod sum;
 
 use core::fmt;
-use std::cmp::min;
 use std::collections::{HashMap, HashSet};
 use std::mem::swap;
 use std::ops::{Add, AddAssign, DerefMut, Sub};
@@ -18,7 +17,6 @@ pub(crate) use exponential_histogram::{EXPO_MAX_SCALE, EXPO_MIN_SCALE};
 use opentelemetry::{otel_warn, KeyValue};
 
 use super::data::{AggregatedMetrics, MetricData};
-use super::pipeline::DEFAULT_CARDINALITY_LIMIT;
 
 // TODO Replace it with LazyLock once it is stable
 pub(crate) static STREAM_OVERFLOW_ATTRIBUTES: OnceLock<Vec<KeyValue>> = OnceLock::new();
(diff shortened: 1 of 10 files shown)


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -73,32 +71,66 @@ where
     /// Configuration for an Aggregator
     config: A::InitConfig,
     cardinality_limit: usize,
+    /// When true, HashMaps start with minimal capacity and grow dynamically.
+    /// When false (default), HashMaps pre-allocate up to cardinality_limit.
+    /// This is useful when cardinality_limit is set high for rare spikes
+    /// but typical usage has far fewer attribute sets.
+    dynamic_memory_allocation_delta: bool,
```
utpilla, line 78: I have a concern about the current approach. While I understand this helps users who face higher memory usage just to accommodate rare spikes in volume, I'm worried this could be used as an escape hatch to effectively disable cardinality capping altogether. Users could set `cardinality_limit` to `usize::MAX` and enable dynamic allocation to bypass the intended safeguards.

We want users to think carefully about the cardinality of their metrics rather than resorting to workarounds that avoid that important design consideration.

That said, we also want to solve the legitimate problem of over-allocating memory for rare spikes. One option to balance these concerns would be to include "unsafe" (or similar warning language) in the public config option name, e.g., `with_lazy_allocation_delta_unsafe()` combined with clear documentation about:

- The intended use case (high cardinality limit with typically low actual cardinality)
- Why this shouldn't be used to bypass cardinality limits

This would make it explicit that this is an advanced option that requires careful consideration and discourage casual misuse.

ArniDagur: >  I'm worried this could be used as an escape hatch to effectively disable cardinality capping altogether. Users could set cardinality_limit to usize::MAX and enable dynamic allocation [...]

This is exactly what I was trying to achieve when I filed the [original bug report / fix](https://github.com/open-telemetry/opentelemetry-rust/pull/3290). Please let users disable the cardinality limit if they want to.

> [...] to bypass the intended safeguards

Perhaps users don't want safeguards :)

---

In our case, we're careful not to add labels which can have an arbitrary number of possible values (and thus explode). Some of the metrics have a cardinality of a few million, but the rest a normal cardinality of maybe 1-1000.

Since setting the cardinality limit is [only a program level configuration option](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/?search=cardinality_limit), setting the cardinality limit for individual meters is not possible. Even if it was, it would be a pain to configure, and you could possibly get the configuration wrong (which would be a disaster since you'd be silently losing metrics).

cijothomas: > Since setting the cardinality limit is [only a program level configuration option](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/?search=cardinality_limit), setting the cardinality limit for individual meters is not possible.

There is no program level configuration - cardinality limit is per metric.

ArniDagur: > There is no program level configuration - cardinality limit is per metric.

Interesting. Can you show me how to use the cardinality limit without using a view?

For example
```
let meter = opentelemetry::global::meter("my_library");
let counter = meter.u64_counter("my_metric").with_description("My metric").build();
```
I don't see any `with_cardinality_limit` or similar on [`InstrumentBuilder`](https://docs.rs/opentelemetry/latest/opentelemetry/metrics/struct.InstrumentBuilder.html)

cijothomas: > Interesting. Can you show me how to use the cardinality limit without using a view?

Cardinality limit is set *using* view only.
example: https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/metrics-advanced/src/main.rs#L25-L30

ArniDagur: Views are a program level configuration, hence my point.

The only option from what I can tell is matching on the instrument name, but that obviously doesn't scale/work for most users.

cijothomas: I am not sure what do you mean by that. CardinalityLimit can be set for each metric individually (yes, by matching on instrument name), not just a program wide limit. If you can share some pain points, I'd be happy to explore improvements.

ArniDagur: It creates a dependency between one part of your code (where you set up the view), to another part of your code (where the instrument is created). That's problematic enough.

Furthermore, as the creator of an instrument, for example in a library, you have no control over the cardinality limit. You have to rely on each of the programs relying on that library to know each and every name of every instrument, and the proper value to set the cardinality limit to. If you change the name of the instrument, you have to remember to change every program's view as well.

cijothomas: The library author has no control over the limit and this is by design. Cardinality is affected by temporality, export frequency etc., and instrumentation author should not be concerned with these.

Changing the name of the instrument is a *breaking* change. If libraries randomly keep changing metrics name, you have bigger problems than just view not matching the name!

ArniDagur: > Cardinality is affected by temporality, export frequency etc

Not necessarily? For my use case, it's known with 100% precision beforehand what the cardinality will be, and it's way higher than the default of 2000. At instrument creation time, I should be able to specify at least a minimum cardinality limit.

cijothomas: Thanks for the detailed explanation of your use case — it makes a lot of sense. I have opened #3507 to track adding an explicit opt-in to disable the cardinality limit. Appreciate your patience here.

cijothomas:
Closing in favor of #3506 (reverts the original change in #3290 and validates input to prevent the panic) and #3507 (tracks the opt-in to disable cardinality capping).

PR #3309, opened 2026-01-13
perf: Nit fix to span creation benches

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/3251 nit followups after this PR

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -9,7 +9,7 @@
 # In the future, we might like to consider using bencher.dev or the framework used by otel-golang here. 
 on: 
   pull_request:
-    types: [opened, synchronize, reopened, labeled]
+    types: [labeled, synchronize]
   push:
     branches:
       - main


.github/workflows/benchmark.yml
```diff
@@ -9,6 +9,7 @@
 # In the future, we might like to consider using bencher.dev or the framework used by otel-golang here. 
 on: 
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
utpilla: Is this needed?

cijothomas: `labeled` was not there, so the PR didn't trigger perf run when I added the label.!

utpilla: Ah I see. The label would have to be added while creating the PR otherwise you would have to either push some commit or reopen the PR, if you added the label after the PR creation. That's probably how it worked for existing open PRs in the repo: https://github.com/open-telemetry/opentelemetry-rust/pulls?q=is%3Aopen+is%3Apr+label%3Aperformance

lalitb (approved):

PR #3312, opened 2026-01-16
fix: Histogram boundaries being ignored in the presence of views

ArniDagur:
I noticed all of my histograms were reporting the default bucket boundaries, even though I specified explicit boundaries when building the meters. That's how I discovered this bug (see test). Essentially, if you have a view, all the explicit histogram boundaries are ignored.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -15,10 +15,12 @@
   - `SamplingDecision`, `SamplingResult`
   - These types are SDK implementation details and should be imported from `opentelemetry_sdk::trace` instead.
 - Fix panics and exploding memory usage from large cardinality limit [#3290][3290]
+- Fix Histogram boundaries being ignored in the presence of views [#3312][3312]
 
 [3227]: https://github.com/open-telemetry/opentelemetry-rust/pull/3227
 [3277]: https://github.com/open-telemetry/opentelemetry-rust/pull/3277
 [3290]: https://github.com/open-telemetry/opentelemetry-rust/pull/3290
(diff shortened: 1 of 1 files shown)


opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2535,6 +2543,60 @@ mod tests {
...
+
+        // Check the bucket counts
+        // -∞ to 1.0: 1
+        // 1.0 to 2.5: 1
+        // 2.5 to 5.5: 3
+        // 5.5 to +∞: 0
+
+        assert_eq!(vec![1.0, 2.5, 5.5], data_point.bounds);
+        assert_eq!(vec![1, 1, 3, 0], data_point.bucket_counts);
+    }
```
cijothomas, line 2615: 💯 Nicely tested! Thank you!

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2535,6 +2543,60 @@ mod tests {
         assert!(data_point.bucket_counts.is_empty());
     }
 
+    fn histogram_aggregation_with_custom_bounds_and_view_helper(temporality: Temporality) {
+        let view = |_: &Instrument| Some(Stream::builder().build().unwrap());
```
cijothomas: do we have a test that validates that if both advice and view is used, the one from view wins?

ArniDagur: Done

maxwindiff:
I'd be interested in this fix as well (thanks @ArniDagur!). Please let me know if there's anything I can help.

PR #3318, opened 2026-01-20
test: BatchLogProcessor test improvement

cijothomas:


opentelemetry-sdk/src/logs/export.rs
```diff
@@ -62,6 +62,19 @@ impl<'a> LogBatch<'a> {
 }
 
 impl LogBatch<'_> {
+    /// Returns the number of log records in the batch.
+    pub fn len(&self) -> usize {
```
lalitb: These methods can be crate level private, if they are only used for tests.

cijothomas: good point! moved them to compile under test cfg only. we may expose it as public api if needed in future.

PR #3319, opened 2026-01-20
chore: Nit improvemenst to docs and test

cijothomas:
InMemoryExporter is used in docs, but it is only available when `testing` feature is enabled. This PR adds conditional build.

### DIFF
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -195,45 +195,44 @@ impl<T: SpanExporter> SpanProcessor for SimpleSpanProcessor<T> {
 /// to manage the export process.
 ///
 /// ```rust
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -195,45 +195,44 @@ impl<T: SpanExporter> SpanProcessor for SimpleSpanProcessor<T> {
...
+/// let tracer = global::tracer("example-tracer");
+/// let mut span = tracer.start("example-span");
+/// span.end(); // Mark the span as completed.
 ///
-///     // Step 5: Ensure all spans are flushed before exiting.
-///     provider.shutdown();
-/// }
+/// // Step 5: Ensure all spans are flushed before exiting.
+/// provider.shutdown();
+/// # }
```
lalitb, line 235: nit - leftover ?

PR #3322, opened 2026-01-22
feat: Stabilize more views - allow changing aggregation

cijothomas:
Was initially kept under feature flag to keep scope under control. Stabilizing more and more things now!
Also added additional tests as users can now change Aggregations  - 
1. Added ExponentialHistogram to coverage
2. Added test coverage to validate that we restrict certain aggregations (Counters can be aggregated using LastValue, and Gauges cannot be aggregated as Sum etc.) - There is no real spec behind this, but this is to avoid producing semantically correct aggregation. We could remove this restriction, if there are actual use cases behind them. Don't think it'd be a breaking change.

### DIFF
--- a/opentelemetry-sdk/src/metrics/mod.rs
+++ b/opentelemetry-sdk/src/metrics/mod.rs
@@ -621,6 +621,133 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn observable_counter_delta_attribute_set_reappears_after_gap() {
+        // Run this test with stdout enabled to see output.
+        // cargo test observable_counter_delta_attribute_set_reappears_after_gap --features=testing -- --nocapture
+
+        // This test verifies the behavior when an attribute set is not reported
+        // for one collection cycle and then reappears.
+        // See: https://github.com/open-telemetry/opentelemetry-specification/issues/4861
+        //
+        // Scenario (Observable Counter with Delta temporality):
+        // | Collection | Callback Reports  | Expected Delta Export  |
+        // |------------|-------------------|------------------------|
+        // | 1          | A=100, B=50       | A=100, B=50            |
+        // | 2          | A=150 (B missing) | A=50 (B not exported)  |
+        // | 3          | A=200, B=80       | A=50, B=80             |
+        //
+        // Current implementation: When B reappears, its delta is calculated from zero
+        // (fresh start), not from the last known value. This is Option 1 from the spec issue.
+
+        let mut test_context = TestContext::new(Temporality::Delta);
+
+        // Shared state for callback: (collection_cycle, value_a, value_b_option)
+        // value_b_option is None when B should not be reported
+        let callback_state = Arc::new(Mutex::new((0u32, 0u64, Option::<u64>::None)));
+        let callback_state_clone = callback_state.clone();
+
+        let _observable_counter = test_context
+            .meter()
+            .u64_observable_counter("my_observable_counter")
+            .with_callback(move |observer| {
+                let state = callback_state_clone.lock().unwrap();
+                let (_cycle, value_a, value_b_option) = *state;
+
+                observer.observe(value_a, &[KeyValue::new("key", "A")]);
+                if let Some(value_b) = value_b_option {
+                    observer.observe(value_b, &[KeyValue::new("key", "B")]);
+                }
+            })
+            .build();
+
+        // Collection 1: A=100, B=50
+        {
+            *callback_state.lock().unwrap() = (1, 100, Some(50));
+            test_context.flush_metrics();
+
+            let MetricData::Sum(sum) =
+                test_context.get_aggregation::<u64>("my_observable_counter", None)
+            else {
+                unreachable!()
+            };
+
+            assert_eq!(sum.data_points.len(), 2);
+            assert_eq!(sum.temporality, Temporality::Delta);
+
+            let dp_a = find_sum_datapoint_with_key_value(&sum.data_points, "key", "A")
+                .expect("datapoint for A expected");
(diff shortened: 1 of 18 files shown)


opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -306,15 +306,24 @@ where
...
         }
 
-        // Apply implicit default view if no explicit matched.
+        // Log warning if views matched but all failed
+        if matched && !errs.is_empty() {
+            otel_warn!(
+                name: "ViewAggregationIncompatible",
+                message = "All matching views have incompatible aggregation. Falling back to default aggregation.",
+                instrument_name = inst.name.as_ref(),
+                errors = format!("{errs:?}").as_str(),
```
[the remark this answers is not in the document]
cijothomas: This is not hot path, just one time at instrument creation.

opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -83,22 +83,14 @@ pub(crate) struct Histogram<T: Number> {
...
         record_sum: bool,
         cardinality_limit: usize,
     ) -> Self {
-        #[cfg(feature = "spec_unstable_metrics_views")]
-        {
-            // TODO: When views are used, validate this upfront
-            bounds.retain(|v| !v.is_nan());
-            bounds.sort_by(|a, b| a.partial_cmp(b).expect("NaNs filtered out"));
-        }
-
```
utpilla, line 101: We should remove this TODO from `metrics.rs`



https://github.com/open-telemetry/opentelemetry-rust/blob/0fd55fdf9b3d3b069d667c2e9831160353a5e83c/opentelemetry-sdk/src/metrics/meter.rs#L405-L406

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2614,6 +2723,108 @@ mod tests {
...
 
+    fn exponential_histogram_aggregation_with_view_helper(temporality: Temporality) {
+        // Arrange: Create a view that converts a regular histogram to Base2ExponentialHistogram
+        let view = |i: &Instrument| {
+            if i.name == "test_histogram" {
+                Some(
+                    Stream::builder()
+                        .with_aggregation(Aggregation::Base2ExponentialHistogram {
+                            max_size: 160,
+                            max_scale: 20,
```
utpilla, line 3664: We need validation for these parameters during `StreamBuilder::build()` similar to how it's done for regular Histogram.
- `max_size` should be greater than `0`
- `max_scale` should be between `-10` and `20`

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -2614,6 +2723,108 @@ mod tests {
...
+        let total_negative_count: u64 = negative_counts.iter().sum();
+        assert_eq!(
+            total_negative_count, 0,
+            "Negative bucket should be empty for positive-only values"
+        );
+
+        // Verify the attribute is present
+        let attrs: Vec<_> = data_point.attributes().collect();
+        assert_eq!(attrs.len(), 1);
+        assert_eq!(attrs[0].key.as_str(), "key1");
```
utpilla, line 3755: The tests for other instruments do another collect to verify Delta vs Cumulative behavior. We should add it for this one as well.

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -933,6 +938,110 @@ mod tests {
         );
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_with_lastvalue_aggregation_uses_default() {
+        // LastValue aggregation is only valid for Gauge instruments.
+        // When applied to a Counter via a view, the view is ignored and
+        // the default aggregation (Sum) is used per the spec:
+        // "proceed as if the View did not [match]"
```
utpilla:
```suggestion
        // "proceed as if the View did not match"
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -933,6 +938,110 @@ mod tests {
...
+            "Counter should use default Sum aggregation when LastValue is incompatible."
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn gauge_with_sum_aggregation_uses_default() {
+        // Sum aggregation is not valid for Gauge instruments.
+        // When applied to a Gauge via a view, the view is ignored and
+        // the default aggregation (LastValue/Gauge) is used per the spec:
+        // "proceed as if the View did not [match]"
```
utpilla:
```suggestion
        // "proceed as if the View did not match"
```

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -933,6 +938,110 @@ mod tests {
         );
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn counter_with_lastvalue_aggregation_uses_default() {
```
utpilla, line 1069: We need tests for valid combinations as well such as Histogram -> Sum, Gauge -> Histogram etc.

opentelemetry-sdk/src/metrics/internal/histogram.rs
cijothomas: done!

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: done

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: added

PR #3323, opened 2026-01-22
refactor(otlp): deprecate tls feature in favor of explicit tls-ring and tls-aws-lc

noahdormann:
The generic tls feature currently enables tonic/tls-ring, which implicitly selects the ring rustls crypto provider.
This makes it difficult or impossible for downstream users to use alternative rustls providers such as aws-lc, and can lead to feature unification conflicts when combined with other TLS-enabled dependencies.

This PR deprecates the generic tls feature and introduces explicit tls-ring and tls-aws-lc features, exposing both underlying tonic TLS options and allowing consumers to consistently select a single rustls crypto provider.

## Motivation

Rustls 0.23 requires a single, unambiguous crypto provider (ring or aws-lc).
Because Cargo features are additive, the existing tls feature could silently force ring even when users explicitly opted into aws-lc elsewhere, resulting in runtime panics during TLS initialization.

Providing explicit TLS feature flags avoids this class of failure and aligns with tonic’s provider model.

## Changes
- Deprecate the existing tls feature
- Introduce:
  - tls-ring → enables tonic/tls-ring
  - tls-aws-lc → enables tonic/tls-aws-lc
- Update documentation to recommend selecting exactly one TLS feature

## Migration

Users currently relying on tls should switch to one of:
```toml
features = ["tls-ring"]
```

or

```toml
features = ["tls-aws-lc"]
```

The deprecated tls feature remains temporarily for compatibility but should not be used in new configurations.

## Caveat
The feature `tls-roots` and `tls-webpki-roots` can no longer explicitly enable `tls`. This could lead to issues with users that relied on this impicit feature map.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Add partial success response handling for OTLP exporters (traces, metrics, logs) per OTLP spec. Exporters now log warnings when the server returns partial success responses with rejected items and error messages. [#865](https://github.com/open-telemetry/opentelemetry-rust/issues/865)
 - Refactor `internal-logs` feature in `opentelemetry-otlp` to reduce unnecessary dependencies[3191](https://github.com/open-telemetry/opentelemetry-rust/pull/3192)
 - Fixed [#2777](https://github.com/open-telemetry/opentelemetry rust/issues/2777)  to properly handle `shutdown_with_timeout()` when using `grpc-tonic`.
+- Deprecate `tls` feature in favor of explicit `tls-ring` and `tls-aws-lc` features.
+  **Migration**: Replace `tls` with `tls-ring` (or `tls-aws-lc`). Users of `tls-roots` or `tls-webpki-roots` must now also enable one of these.
 
 ## 0.31.0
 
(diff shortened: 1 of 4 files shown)


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Add partial success response handling for OTLP exporters (traces, metrics, logs) per OTLP spec. Exporters now log warnings when the server returns partial success responses with rejected items and error messages. [#865](https://github.com/open-telemetry/opentelemetry-rust/issues/865)
 - Refactor `internal-logs` feature in `opentelemetry-otlp` to reduce unnecessary dependencies[3191](https://github.com/open-telemetry/opentelemetry-rust/pull/3192)
 - Fixed [#2777](https://github.com/open-telemetry/opentelemetry rust/issues/2777)  to properly handle `shutdown_with_timeout()` when using `grpc-tonic`.
+- Refactor `tls` feature in `opentelemetry-otlp` to allow selecting either `ring` or `aws-lc` as crypto provider.
```
lalitb: I think changelog should be more explicit about deprecation and migration.
```
- Deprecate `tls` feature in favor of explicit `tls-ring` and `tls-aws-lc` features.
 **Migration**: Replace `tls` with `tls-ring` (or `tls-aws-lc`). Users of `tls-roots` or `tls-webpki-roots` must now also enable one of these.
```

lalitb (approved):

PR #3329, opened 2026-01-26
fix(proto): Permissive integers in protobuf JSON parsing

### LINKED ISSUE #3328: [Bug]: timestamps encoded as `u64` are not accepted (JSON)
### What happened?

We've been seeing an issue with incoming OpenTelemetry logs, encoded as JSON. It seems that these are due to integers being encoded as `u64` integers instead of strings.

Looking at the [OpenTelemetry spec](https://opentelemetry.io/docs/specs/otlp):

> Note that according to [Protobuf specs](https://developers.google.com/protocol-buffers/docs/proto3#json) 64-bit integer numbers in JSON-encoded payloads are encoded as decimal strings, and either numbers or strings are accepted when decoding.

However, `deserialize_string_to_u64` only accepts strings, rejecting valid JSON fragments like:

```json
{"timeUnixNano": 1544712660300000000}
```

with `error: invalid type: integer, expected a string`.

v0.31.0, also verified on latest `main`

marc-pydantic:
Fixes: #3328

## Changes

Ensure that we accept both `u64` and `String` as inputs for large numbers, as the protobuf spec dictates (see #3328).

Note: This deviates in pattern slightly from established code, using a visitor for `IntOrString` -- I hope this is okay, but there's potential to unify here.

(Technically it is an API change, since more values are now accepted?)

### DIFF
--- a/opentelemetry-proto/tests/json_serde.rs
+++ b/opentelemetry-proto/tests/json_serde.rs
@@ -773,6 +773,20 @@ mod json_serde {
                 let expected: Event = value();
                 assert_eq!(actual, expected);
             }
+
+            /// OTLP spec: 64-bit integers may be encoded as either strings or integers.
+            #[test]
+            fn deserialize_integer_timestamp() {
+                // language=json
+                const JSON: &str = r#"{
+  "timeUnixNano": 1234567890,
+  "name": "my_event"
+}"#;
+                let actual: Event =
+                    serde_json::from_str(JSON).expect("deserialization must succeed");
+                let expected: Event = value();
+                assert_eq!(actual, expected);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


opentelemetry-proto/tests/json_serde.rs
```diff
@@ -773,6 +773,20 @@ mod json_serde {
                 let expected: Event = value();
                 assert_eq!(actual, expected);
             }
+
+            /// OTLP spec: 64-bit integers may be encoded as either strings or integers.
+            #[test]
+            fn deserialize_integer_timestamp() {
```
scottgerring, line 779: Could be good to also cover deserializing a metrics payload with integrator timestamps - e.g. `startTimeUnixNano: 123456789`.

scottgerring (approved):
LGTM

adriangb:
Hi @lalitb  could you please take a look here before this bit rots more? Thanks!

adriangb:
I think this has been subsumed by other work that fixed the issue. @lalitb IMO we can close this issue.

PR #3334, opened 2026-01-31
test: Add test for observablecounter stale and reappear

cijothomas:
Small test only followup to https://github.com/open-telemetry/opentelemetry-rust/pull/3248
Added for discussion along https://github.com/open-telemetry/opentelemetry-specification/issues/4861


cijothomas:
@taisho6339 If you have a moment, please review this, as it is related to the discussion on https://github.com/open-telemetry/opentelemetry-rust/pull/3248

cijothomas:
incase anyone wondering why not `ObservableUpDownCounter` tests - this is because the issue is only applicable when exporting in "Delta" mode. And there is no real "delta' for ObservableUpDownCounter as they are always exported as cumulative.

Issue #3288, opened 2025-12-19
Un-deprecate Prometheus


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/issues/2451#issuecomment-2730144212 

We have multiple people ([1](https://github.com/open-telemetry/opentelemetry-rust/issues/2451#issuecomment-3193959474), [2](https://github.com/open-telemetry/opentelemetry-rust/pull/3287) & [3](https://github.com/open-telemetry/opentelemetry-rust/pull/3286))who offered to help rewrite the prometheus exporter without external dependencies. Opening an issue to add official support for the prometheus exporter.

PR #3341, opened 2026-02-04
feat: remove unused feature from prometheus dep

reneca:
Another way of #3286
Try to participate of #3288

## Changes

As many others, I still use opentelemetry-prometheus (even though it’s not maintained anymore) because some cloud providers only provide Prometheus to scrape metrics.
I see that you want to get rid of the Prometheus dependency.
But the main issue for me is the protobuf dependency outside of Prometheus.
It isn’t used in this project, and it takes a weekend to compile.

That’s why I propose this small change to avoid using the default dependencies of Prometheus. Thanks to this, we remove those dependencies from Prometheus:
```
< │   ├── protobuf v3.7.2
< │   │   ├── once_cell v1.21.3
< │   │   ├── protobuf-support v3.7.2
< │   │   │   └── thiserror v1.0.69
< │   │   │       └── thiserror-impl v1.0.69 (proc-macro)
< │   │   │           ├── proc-macro2 v1.0.106 (*)
< │   │   │           ├── quote v1.0.44 (*)
< │   │   │           └── syn v2.0.114 (*)
< │   │   └── thiserror v1.0.69 (*)
```

I’m looking forward to whether maintainership in this repo will return :)

### DIFF
--- a/opentelemetry-appender-log/CHANGELOG.md
+++ b/opentelemetry-appender-log/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - "spec_unstable_logs_enabled" feature flag is removed. The capability (and the
   backing specification) is now stable and is enabled by default.
+  [3278](https://github.com/open-telemetry/opentelemetry-rust/pull/3278)
 
 ## 0.31.0
 
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 - "spec_unstable_logs_enabled" feature flag is removed. The capability (and the
   backing specification) is now stable and is enabled by default.
+  [3278](https://github.com/open-telemetry/opentelemetry-rust/pull/3278)
 
 ## 0.31.1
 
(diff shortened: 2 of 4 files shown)


opentelemetry-prometheus/Cargo.toml
```diff
@@ -21,13 +21,13 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 once_cell = { version = "1.13" }
-opentelemetry = { version = "0.31.0", default-features = false, features = ["metrics", "internal-logs"] }
-opentelemetry_sdk = { version = "0.31.0", default-features = false, features = ["metrics", "experimental_metrics_custom_reader"] }
-prometheus = "0.14"
+opentelemetry = { version = "0.31", default-features = false, features = ["metrics", "internal-logs"] }
+opentelemetry_sdk = { version = "0.31", default-features = false, features = ["metrics", "experimental_metrics_custom_reader"] }
+prometheus = { version = "0.14", default-features = false }
```
cijothomas, line 26: nice! 

https://crates.io/crates/cargo-ensure-no-default-features might be something we want to include in this repo!

reneca:
I added a line in changelog and increase version to 0.31.1.

opentelemetry-prometheus/Cargo.toml
reneca: I’m not sure if everything needs to be like that, but at least the dependencies should be checked.
I like [cargo-machete](https://github.com/bnjbvr/cargo-machete) to do this.

Issue #3331, opened 2026-01-29
[Bug]: Improve context in BatchSpanProcessor/BatchLogProcessor error logs when otlp export fails


krisztianfekete:
### What happened?

Currently, when `otlp` export fails due to transport errors, e.g. misconfigured endpoint, the error level logs are not helpful finding the root cause of the problem.

## Current Behavior
With a misconfigured endpoint (missing schema, e.g., `localhost:4317` instead of `http://localhost:4317`):

**Error-level log:**
```
2026-01-29T12:12:15.755056Z error opentelemetry_sdk
  name="BatchSpanProcessor.ExportError"
  error="Operation failed: status: 'Unknown error', self: \"transport error\""
```

**Debug-level log:**
```
2026-01-29T12:33:01.335273Z debug tonic::transport::channel::service::reconnect
  reconnect::poll_ready: ConnectError(ConnectError("invalid URL, scheme is missing"))
```

Looks like we lose the context when the original error gets wrapped into `OTelSdkError::InternalFailure` and the processors log it as a generic error without the original context.

I propose preserving the original error in `InternalFailure` in error messages. What do you all think?

### OpenTelemetry API Version (i.e version of `opentelemetry` crate)

main

### OpenTelemetry SDK Version (i.e version of `opentelemetry_sdk` crate)

main

### What Exporter(s) are you seeing the problem on?

OTLP

### Relevant log output

See above.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

krisztianfekete:
Hey @cijothomas, thanks for taking a look!
The PR above doesn't really help with my problem where one have to enable DEBUG level logging to see that the problem is a user-side misconfiguration that is easily fixable. I am proposing surfacing the `ConnectError` from the above example at an elevated log level. In this case, further sanitization isn't even required as that message doesn't even contain the actual connection string.

scottgerring:
PR #3360 (merged) closed sibling #3343 but does not address this - in fact strips more info from the ERROR-level log. This seems like a legitimate concern to me, but a fix must distinguish pre-flight local transport errors from server-response-bearing errors.

krisztianfekete:
[...]

PR #3343, opened 2026-02-04
fix: OTLP logging improvements

cijothomas:
Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/3021
Supersedes https://github.com/open-telemetry/opentelemetry-rust/pull/3108

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -488,15 +488,13 @@ impl OtlpHttpClient {
 
         // Send request
         let response = client.send_bytes(request).await.map_err(|e| {
+            // Connection errors (e.g., "Connection refused", DNS failures) typically
+            // indicate user-side misconfigurations and don't contain sensitive data,
+            // so it's safe to log the error message at WARN level.
             otel_warn!(
                 name: "HttpClient.ExportFailed",
-                url = request_uri.as_str()
-            );
-            // Error details may contain sensitive information,
-            // so log them at debug level only.
-            otel_debug!(
-                name: "HttpClient.ExportFailedDetails",
-                error = format!("{e:?}")
+                url = request_uri.as_str(),
+                error = format!("{e}")
             );
             HttpExportError::new(0, "HTTP export failed".to_string())
         })?;
--- a/opentelemetry-otlp/src/exporter/tonic/logs.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/logs.rs
@@ -152,20 +152,44 @@ impl LogExporter for TonicLogsClient {
         {
             Ok(_) => Ok(()),
             Err(tonic_status) => {
-                otel_warn!(
-                    name: "TonicLogsClient.ExportFailed",
-                    grpc_code = format!("{:?}", tonic_status.code())
+                // For connection-related errors (Unavailable, Unknown, etc.), the message
+                // typically contains safe, actionable information (e.g., "Connection refused").
+                // For auth errors (Unauthenticated, PermissionDenied), the message may contain
+                // sensitive information, so we only log the code at WARN level.
+                let code = tonic_status.code();
+                let is_connection_error = matches!(
+                    code,
+                    tonic::Code::Unavailable
+                        | tonic::Code::Unknown
+                        | tonic::Code::DeadlineExceeded
+                        | tonic::Code::ResourceExhausted
+                        | tonic::Code::Aborted
+                        | tonic::Code::Cancelled
                 );
-                // grpc_message and grpc_details may contain sensitive information,
-                // so log them at debug level only.
-                otel_debug!(
-                    name: "TonicLogsClient.ExportFailedDetails",
-                    grpc_message = tonic_status.message(),
-                    grpc_details = format!("{:?}", tonic_status.details())
-                );
+
+                if is_connection_error {
+                    otel_warn!(
+                        name: "TonicLogsClient.ExportFailed",
+                        grpc_code = format!("{:?}", code),
+                        grpc_message = tonic_status.message()
+                    );
+                } else {
+                    // For potentially sensitive errors (Unauthenticated, PermissionDenied, etc.),
+                    // only log the code at WARN level.
+                    otel_warn!(
+                        name: "TonicLogsClient.ExportFailed",
+                        grpc_code = format!("{:?}", code)
+                    );
+                    // Log message and details at debug level for sensitive error types.
+                    otel_debug!(
+                        name: "TonicLogsClient.ExportFailedDetails",
+                        grpc_message = tonic_status.message(),
+                        grpc_details = format!("{:?}", tonic_status.details())
+                    );
+                }
                 Err(OTelSdkError::InternalFailure(format!(
                     "Logs export failed with gRPC code: {:?}",
-                    tonic_status.code()
+                    code
                 )))
             }
         }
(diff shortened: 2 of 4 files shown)


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## vNext
 
+- Improved error logging for gRPC and HTTP exporters. The status code is now
+  logged at WARN level, while potentially sensitive details (message, response
+  body) are logged at DEBUG level only. This prevents sensitive information
+  (like authentication tokens) from being logged at higher log levels while
+  still providing actionable error information in production.
```
utpilla: I think we could add an explicit note that's more actionable from user's perspective.

```suggestion
  (like authentication tokens) from being logged at higher log levels.
  **Note:** If you enable `DEBUG` level logging in production, these details 
  will still be visible. Use `WARN` or higher for production centralized logging.
```

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -488,7 +488,17 @@ impl OtlpHttpClient {
 
         // Send request
         let response = client.send_bytes(request).await.map_err(|e| {
-            HttpExportError::new(0, format!("Network error: {e:?}")) // Network error
+            otel_warn!(
+                name: "HttpClient.ExportFailed",
```
utpilla, line 495: We are using the same event name `HttpClient.ExportFailed` for two different kinds of errors:
- Network error from `client.send_bytes`
- HTTP status error when `response.status()` is not successful.

Maybe we could use different event names for better diagnosability: `HttpClient.NetworkError` and `HttpClientStatusError`.

cijothomas: I did consider that and ended up with same one. I am not 100% happy with this either.

I tried to distinguish between failures where we don't get a response at all and the ones where we get response, but it has status code indicating failure. But then realized that `request` client don't expose us a way to obtain `StatusCode` from its Error. Without having firmer control over the client, this is a decent tradeoff.
(The otel-http crate could offer additional logging, as it knows internals of the client it uses, but OTLP exporter does not know that much).

let me know if we should try to make this better or this is acceptable tradeoff.

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -137,9 +148,20 @@ impl LogExporter for TonicLogsClient {
...
+                    grpc_code = format!("{:?}", tonic_status.code())
+                );
+                // grpc_message and grpc_details may contain sensitive information,
+                // so log them at debug level only.
+                otel_debug!(
+                    name: "TonicLogsClient.ExportFailedDetails",
+                    grpc_message = tonic_status.message(),
+                    grpc_details = format!("{:?}", tonic_status.details())
+                );
+                Err(OTelSdkError::InternalFailure("Logs export failed".into()))
```
utpilla: We could include the status code here.

```suggestion
                Err(OTelSdkError::InternalFailure(format!(
                    "Logs export failed with gRPC code: {:?}",
                    tonic_status.code()
                )))
```

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -94,8 +94,19 @@ impl LogExporter for TonicLogsClient {
...
+                                    // grpc_message and grpc_details may contain sensitive information,
+                                    // so log them at debug level only.
+                                    otel_debug!(
+                                        name: "TonicLogsClient.InterceptorFailedDetails",
+                                        grpc_message = e.message(),
+                                        grpc_details = format!("{:?}", e.details())
+                                    );
                                     // Convert interceptor errors to tonic::Status for retry classification
-                                    tonic::Status::internal(format!("interceptor error: {e:?}"))
+                                    tonic::Status::internal("Logs export failed in interceptor")
```
utpilla: We could include the status code here.

```suggestion
                                    tonic::Status::internal(&format!(
                                        "Logs export failed in interceptor with gRPC code: {:?}",
                                        e.code()
                                    ))
```

opentelemetry-otlp/CHANGELOG.md
utpilla: Probably something that should be added to these documents as well:
- https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/design/logs.md#internal-logs
- `opentelemetry-otlp` README

cijothomas: ya its time to make a doc section for internal logging covering the overall internal logs capability of the project. Will add in a follow up PR soon.

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -137,9 +151,47 @@ impl LogExporter for TonicLogsClient {
...
-            Err(tonic_status) => Err(OTelSdkError::InternalFailure(format!(
-                "export error: {tonic_status:?}"
-            ))),
+            Err(tonic_status) => {
+                // For connection-related errors (Unavailable, Unknown, etc.), the message
+                // typically contains safe, actionable information (e.g., "Connection refused").
+                // For auth errors (Unauthenticated, PermissionDenied), the message may contain
+                // sensitive information, so we only log the code at WARN level.
+                let code = tonic_status.code();
+                let is_connection_error = matches!(
```
lalitb, line 160: nit - this ~35-40 lines of code seems to be repeated for all the signals - can have a shared helper method.

opentelemetry-otlp/src/exporter/tonic/metrics.rs
```diff
@@ -127,9 +136,46 @@ impl MetricsClient for TonicMetricsClient {
...
+                    );
+                    // Log message and details at debug level for sensitive error types.
+                    otel_debug!(
+                        name: "TonicMetricsClient.ExportFailedDetails",
+                        grpc_message = tonic_status.message(),
+                        grpc_details = format!("{:?}", tonic_status.details())
+                    );
+                }
+                Err(OTelSdkError::InternalFailure(
+                    "Metrics export failed".into(),
```
lalitb, line 176: `grpc_code` should also be added here?

opentelemetry-otlp/src/exporter/tonic/trace.rs
```diff
@@ -96,8 +96,19 @@ impl SpanExporter for TonicTracesClient {
...
+                                    // grpc_message and grpc_details may contain sensitive information,
+                                    // so log them at debug level only.
+                                    otel_debug!(
+                                        name: "TonicTracesClient.InterceptorFailedDetails",
+                                        grpc_message = e.message(),
+                                        grpc_details = format!("{:?}", e.details())
+                                    );
                                     // Convert interceptor errors to tonic::Status for retry classification
-                                    tonic::Status::internal(format!("interceptor error: {e:?}"))
+                                    tonic::Status::internal("Traces export failed in interceptor")
```
lalitb, line 111: `grpc_code` here too.

krisztianfekete:
> > I like this direction, but as discussed in #3331 I think we can improve the UX where one have to enable DEBUG level logging to see that the problem is a user-side misconfiguration that is easily fixable. I am proposing surfacing the `ConnectError` from the above example at an elevated log level. In this case, further sanitization isn't even required as that message doesn't even contain the actual connection string. cc. @cijothomas
> 
> can you recheck now? I made some changes to help.

@cijothomas It looks like the source error is still not logged at WARN/ERROR level, although we have some additional gRPC context for some codes. Do you think we can surface the source error message at a higher level without losing it?

PR #3350, opened 2026-02-14
feat(tracing): allow span processor to keep span reference

### LINKED ISSUE #602: Allow span processor to keep reference of spans
The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart) noted that we should allow span processor to keep a reference to the span passed as the parameter of `on_start` method
> It SHOULD be possible to keep a reference to this span object and updates to the span SHOULD be reflected in it

samugi:
Fixes #602 

## Changes

Adds a `get_handle` method to `Span` that allows obtaining an instance of `SpanHandle`: a structure that shares the same fields as the span, where the SpanData is protected by a Mutex.

This allows holding a reference to a Span, which enables use cases such as span processors keeping track of active spans, as mentioned in [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart).

Example:

```rust
impl SpanProcessor for MyProcessor {
    fn on_start(&self, span: &mut Span, _cx: &Context) {
        let handle = span.get_handle();
        // Store handle for later use
    }
}
```

### Notes

First contribution here. The change ended up being a bit more invasive than I would've hoped, if there's a cleaner or more idiomatic way to achieve the goal, I'm very happy to adjust the approach and keep iterating on it. Thanks!

### DIFF
--- a/opentelemetry-sdk/src/trace/mod.rs
+++ b/opentelemetry-sdk/src/trace/mod.rs
@@ -39,7 +39,7 @@ pub use id_generator::{IdGenerator, RandomIdGenerator};
 pub use links::SpanLinks;
 pub use provider::{SdkTracerProvider, TracerProviderBuilder};
 pub use sampler::{Sampler, SamplingDecision, SamplingResult, ShouldSample};
-pub use span::Span;
+pub use span::{Span, SpanHandle};
 pub use span_limit::SpanLimits;
 pub use span_processor::{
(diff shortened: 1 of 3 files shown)


scottgerring:
@cijothomas @bantonsson @paullegranddc - do you folks have an opinion on path forward here? This one has been sitting open for a while now

Issue #3283, opened 2025-12-10
[Feature]: Document supported and unsupported OTEL variables


BatmanAoD:
### Related Problems?

As far as I can tell, the documentation for the `opentelemetry_sdk` crate doesn't say anywhere what environment variables it does or does not recognize. The only authoritative documentation appears to be this table, which is somewhat difficult to read given its width: https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables

My team believed that the Rust SDK supports all the standard OTEL environment variables (possibly because we didn't notice the "note" on [this page](https://opentelemetry.io/docs/languages/sdk-configuration/general/), but I don't recall specifically how we came to this conclusion). We therefore relied on `OTEL_TRACES_SAMPLER` to configure our sampling, and consequently emitted 100x more traces than intended.

### Describe the solution you'd like:

The `opentelemetry_sdk` crate itself should clearly document which environment variables are respected, and which are not.

### Considered Alternatives

* The crate could link to the [compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables) somewhere prominent in the documentation. It seems odd for me that this information would be part of the _specification_ repo, though.
* The crate could support all environment variables. I think this should be the goal eventually.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

codefromthecrypt:
#3351 implements `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` env var support.

Issue #3061, opened 2025-07-11
Stabilization - OTLP


scottgerring:
This is a ticket to capture stabilization efforts around the OTLP crate. It will likely be composed of a review of the current state and some architectural documentation thereof, some proposed changes, and then the implementation of those changes.

We'll cover off everything in the [compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#exporters) for exporters here too.

scottgerring:
##  Open Issues & PRs related to OTLP as of 2025-07-10

### Issues (latest updated first)
| # | Title (abridged) | Updated | Area |
|---|------------------|---------|-------|
| [#3045](https://github.com/open-telemetry/opentelemetry-rust/issues/3045) | Simplify `opentelemetry-proto` (SDK decoupling, gRPC separation) | 2025-06-30 | Architecture |
| [#2881](https://github.com/open-telemetry/opentelemetry-rust/issues/2881) | Allow **async interceptors** for OTLP gRPC exporter | 2025-06-17 | Extensibility |
| [#2802](https://github.com/open-telemetry/opentelemetry-rust/issues/2802) | OTLP `MetricExporter` deadlock | 2025-05-19 | Metrics / runtime |
| [#1987](https://github.com/open-telemetry/opentelemetry-rust/issues/1987) | Broken-pipe errors when exporting via OTLP | 2025-03-31 | Stability |
| [#2877](https://github.com/open-telemetry/opentelemetry-rust/issues/2877) | Prevent *telemetry-induced telemetry* in OTLP exporter | 2025-03-27 | Performance |
| [#2777](https://github.com/open-telemetry/opentelemetry-rust/issues/2777) | Handle **shutdown** in OTLP/gRPC | 2025-03-11 | Lifecycle |
| [#2602](https://github.com/open-telemetry/opentelemetry-rust/issues/2602) | OTLP **File Exporter** | 2025-02-27 | Feature request |
| [#1592](https://github.com/open-telemetry/opentelemetry-rust/issues/1592) | OTLP Exporter **builder** ergonomics | 2024-12-14 | API |

### PRs (latest updated first)
| # | Title | Updated | Purpose |
|---|-------|---------|---------|
| [PR #3046](https://github.com/open-telemetry/opentelemetry-rust/pull/3046) | SDK decoupling & gRPC separation | 2025-07-06 | Follow-up to issue #3045 |
[...]

PR #3351, opened 2026-02-15
feat(otlp): support OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE

codefromthecrypt:
Relates to #3283, #3061

## Changes

Add support for `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
per the [OTLP Metrics Exporter spec](https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/).

This env var is supported by Go, Python, .NET, JS, PHP, and Ruby SDKs but was missing
from the Rust SDK, requiring users to manually parse it.

Accepted values (case-insensitive): `cumulative` (default), `delta`, `lowmemory`.
Programmatic `.with_temporality()` takes priority over the env var.

- `FromStr` impl for `Temporality` in `opentelemetry-sdk`
- `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` constant and `resolve_temporality()` in `opentelemetry-otlp`
- `MetricExporterBuilder` resolves temporality from env when not set programmatically
- `ExporterBuildError::InvalidConfig` variant for invalid env var values

### DIFF
--- a/opentelemetry-otlp/src/exporter/mod.rs
+++ b/opentelemetry-otlp/src/exporter/mod.rs
@@ -114,10 +114,10 @@ pub enum ExporterBuildError {
     #[error("invalid URI {0}. Reason {1}")]
     InvalidUri(String, String),
 
-    /// Invalid configuration from environment variable or builder.
+    /// Invalid configuration.
     #[error("{name}: {reason}")]
     InvalidConfig {
-        /// The configuration name (e.g. environment variable).
+        /// The configuration name.
         name: String,
         /// The reason the configuration is invalid.
         reason: String,
--- a/opentelemetry-otlp/src/metric.rs
+++ b/opentelemetry-otlp/src/metric.rs
@@ -37,7 +37,6 @@ pub const OTEL_EXPORTER_OTLP_METRICS_COMPRESSION: &str = "OTEL_EXPORTER_OTLP_MET
 /// Note: this is only supported for HTTP.
 pub const OTEL_EXPORTER_OTLP_METRICS_HEADERS: &str = "OTEL_EXPORTER_OTLP_METRICS_HEADERS";
 /// Temporality preference for metrics, defaults to cumulative.
-/// Accepted values: `cumulative`, `delta`, `lowmemory` (case-insensitive).
 pub const OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE: &str =
     "OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE";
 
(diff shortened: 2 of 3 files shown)


opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -376,7 +376,7 @@ fn parse_header_key_value_string(key_value_string: &str) -> Option<(&str, String
 
 #[cfg(test)]
 #[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
-mod tests {
+pub(crate) mod tests {
```
cijothomas: can we avoid this? Whatever necessitated this change can be moved to a shared utils module, or if it is simple enough, copy/duplicate for now.

opentelemetry-otlp/src/metric.rs
```diff
@@ -71,19 +75,41 @@ impl<C> MetricExporterBuilder<C> {
...
+///
+/// Priority: programmatic config > env var > default (`Cumulative`).
+#[cfg(any(feature = "http-proto", feature = "http-json", feature = "grpc-tonic"))]
+fn resolve_temporality(provided: Option<Temporality>) -> Result<Temporality, ExporterBuildError> {
+    if let Some(temporality) = provided {
+        return Ok(temporality);
+    }
+    if let Ok(val) = std::env::var(OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE) {
+        return val.parse::<Temporality>().map_err(|_| {
+            ExporterBuildError::InternalFailure(format!(
```
cijothomas: better to add a specific variant for this instead of bundling with internal-failure - this is an obvious user config error.
The enum has a big TODO already; for now, lets add a `InvalidConfig` variant with name and reason?

ExporterBuildError::InvalidConfig {
    name: OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE.to_string(),
    reason: format!("Invalid value '{val}'. Expected: cumulative, delta, or lowmemory"),
}

Not perfect and has room for future improvement, but would be better than bundling into InternalError.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## vNext
 
+- Add support for `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
```
cijothomas, line 5: 💯 nicely written!

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- Add `FromStr` implementation for `Temporality` enum, accepting `cumulative`, `delta`,
```
cijothomas: this can be removed as it feels noise to most SDK usages. The real change is covered via OTLP changelog.

cijothomas (approved):
LGTM

PR #3352, opened 2026-02-19
feat: Bump MSRV to `1.85` and replace 3rd party crates with std library APIs

bryantbiggs:
Fixes #

## Changes
- By bumping the MSRV from 1.75 to 1.85 we are able to remove the following 3rd party crates for rust std library APIs:
  - `once_cell`
  - `lazy_static`
  - `num-format`
  - `num_cpus`
- Resolves 6 TODO comments in opentelemetry-sdk that were waiting for LazyLock stabilization (stable since Rust 1.80). Each OnceLock paired with a getter function is now a single LazyLock static declaration

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,16 +143,7 @@ Registry](https://opentelemetry.io/ecosystem/registry/?language=rust).
 
 ## Supported Rust Versions
 
-OpenTelemetry is built against the latest stable release. The minimum supported
-version is 1.75. The current OpenTelemetry version is not guaranteed to build
-on Rust versions earlier than the minimum supported version.
-
-The current stable Rust compiler and the three most recent minor versions
-before it will always be supported. For example, if the current stable compiler
-version is 1.49, the minimum supported version will not be increased past 1.46,
-three minor versions prior. Increasing the minimum supported compiler version
-is not considered a semver breaking change as long as doing so complies with
-this policy.
+[![MSRV](https://img.shields.io/crates/msrv/opentelemetry)](https://crates.io/crates/opentelemetry)
 
 ## Contributing
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ http-body-util = "0.1"
 hyper = { version = "1.3", default-features = false }
 hyper-util = "0.1"
 log = "0.4.21"
-once_cell = "1.13"
 pin-project-lite = "0.2"
 prost = "0.14"
 rand = { version = "0.9", default-features = false }
@@ -57,9 +56,6 @@ ctrlc = "3.2.5"
 futures-channel = "0.3"
 futures-sink = "0.3"
 const-hex = "1.14.1"
-lazy_static = "1.4.0"
-num-format = "0.4.4"
-num_cpus = "1.15.0"
 opentelemetry = { path = "opentelemetry", version = "0.32", default-features = false }
 opentelemetry_sdk = { path = "opentelemetry-sdk", version = "0.32", default-features = false }
 opentelemetry-appender-log = { path = "opentelemetry-appender-log", version = "0.32", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,9 @@ members = [
 ]
 resolver = "2"
 
+[workspace.package]
+rust-version = "1.80.0"
+
 [profile.bench]
 # https://doc.rust-lang.org/cargo/reference/profiles.html#debug
 # See function names in profiling reports.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-rust-version = "1.80.0"
+rust-version = "1.85.0"
 
 [profile.bench]
 # https://doc.rust-lang.org/cargo/reference/profiles.html#debug
--- a/deny.toml
+++ b/deny.toml
@@ -7,6 +7,7 @@ allow = [
     "Apache-2.0",
     "ISC",
     "BSD-3-Clause",
+    "Unicode-3.0",
     "Zlib",
 ]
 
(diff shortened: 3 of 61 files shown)


README.md
```diff
@@ -147,16 +147,7 @@ Registry](https://opentelemetry.io/ecosystem/registry/?language=rust).
 
 ## Supported Rust Versions
 
-OpenTelemetry is built against the latest stable release. The minimum supported
-version is 1.75. The current OpenTelemetry version is not guaranteed to build
-on Rust versions earlier than the minimum supported version.
-
-The current stable Rust compiler and the three most recent minor versions
```
cijothomas, line 150: lets keep the policy documented in the main readme.md

bryantbiggs: sure, I will re-add. I will say, its a rather odd and confusing statement though. MSRV means the min version specified is the minimum - not the minimum -3 versions

cijothomas: This is just listing our policy - we don't require rust version newer than the current-3.
Can use a rewrite for conveying the intent clearly, but in a separate PR.

bryantbiggs:
looks like I missed some TODOs that were waiting for LAZYLOCK support - addressed in last commit.

I understand the resistance to disrupt users, theres always a natural tension between staying current and the least friction for users. I used CLAUDE to summarize the additional gains that could be realized if 1.85 and rust edition 2024 were supported (again, edition 2024 would be a separate PR)

## What MSRV 1.85 unlocks next

### `#[expect(lint)]` (1.81)

Works like `#[allow]` but warns you when the suppressed lint stops firing. We
have 60+ `#[allow]` attributes — converting the temporary ones (`dead_code`,
`unused`) to `#[expect]` means stale suppressions surface automatically instead
of accumulating silently.

### `use<>` precise capturing (1.82)

`opentelemetry-appender-tracing` currently has:

```rust
#[allow(impl_trait_overcaptures)] // can only be fixed with Rust 1.82+
```

This can now be properly fixed instead of suppressed.

### `core::error::Error` (1.81)

The `Error` trait is available in `core`, not just `std`. We already use
`thiserror` v2 with `default-features = false` for `no_std` error types — an
MSRV of 1.85 guarantees all downstream consumers have `core::error::Error`
available, which strengthens the `no_std` story.

### Stable `cfg(target_has_atomic)` (1.80)

PR #3345 added 32-bit platform support using `cfg(target_has_atomic = "64")`
with `portable-atomic`. This cfg was stabilized in 1.80 — bumping MSRV means it
works on stable without question.

### `async-trait` removal (evaluation)

The only first-party trait still using the `async-trait` proc-macro is
`HttpClient`. The SDK exporter traits (`SpanExporter`, `LogExporter`,
`PushMetricExporter`) already moved away from it. `HttpClient` is used as
`Arc<dyn HttpClient>`, so native async fn in traits doesn't work directly, but
the proc-macro can be replaced with explicit `Pin<Box<dyn Future>>` return
types — which is what `#[async_trait]` generates anyway, minus the compile-time
cost.

## What edition 2024 adds

### MSRV-aware cargo resolver (resolver v3)

This is the biggest practical win.

Edition 2024 defaults to resolver v3, which is MSRV-aware. When resolving
dependencies, cargo prefers versions whose `rust-version` is compatible with
ours. If a transitive dependency publishes a version requiring a newer Rust than
our MSRV, cargo picks the older compatible version instead of breaking the build.

**For maintainers:** No more surprise CI breakages from `cargo update` pulling in
a dependency that bumped its own MSRV. Less time spent pinning versions.

**For users:** Anyone using resolver v3 themselves gets correct resolution against
our declared `rust-version`. The field is already in our Cargo.toml files —
resolver v3 makes it actually meaningful during resolution.

### Stricter compiler defaults

- **`unsafe_op_in_unsafe_fn`** changes from `allow` to `warn` — requires
  explicit `unsafe {}` blocks inside `unsafe fn` bodies. We have one `unsafe fn`
  in the stress tests; it already has inner `unsafe` blocks.
- **Lifetime capture rules** — `-> impl Trait` return types now capture all
  in-scope lifetimes by default, preventing accidental lifetime escapes.
  `cargo fix --edition` handles the migration automatically.
- **Tail expression scoping** — temporaries in tail expressions drop sooner.
  Our `.lock().unwrap().clone()` patterns all produce owned values, so no
  breakage, but the stricter scoping is safer by default.
- **Never type `!` fallback** — changes from `()` to `!`, preventing a class of
  subtle type inference bugs.

### Cleanup

- `#![warn(rust_2018_idioms)]` in 7 crate roots becomes redundant (edition 2024
  enforces these by default) and can be removed.
- `edition` can move to `[workspace.package]` — one source of truth instead of 22
  individual Cargo.toml entries.

bryantbiggs:
I am using and most interested in the OTLP exporter, so if there are areas where I might be able to help 

> > @open-telemetry/rust-approvers Please comment if there are any major concerns about this bump. We can leave this open for 2-3 days to get feedbacks.
> 
> I appreciate the dependency cleanup, but I have a timing concern.
> 
> Traces and the OTLP exporters are still Beta/RC. If we bump MSRV now, users pinned to older toolchains will be stranded on a release that isn't yet a complete solution across all signals. I'd suggest we stabilize those first, so anyone who can't follow the toolchain bump at least has something solid to stay on.
> 
> Can we revisit this after the Traces/OTLP stabilization milestone?

I can certainly try to help with the stabilization since the OTLP exporter is what I am using and most interested in improving 
- https://github.com/open-telemetry/opentelemetry-rust/pull/3356

cijothomas:
> > > Could we have bumped to 1.80 and still used the std library APIs?
> > 
> > 
> > that was the original intent, but `opentelemetry-prometheus` looks like it requires a MSRV of 1.81 https://github.com/open-telemetry/opentelemetry-rust/actions/runs/22163623736/job/64085949878?pr=3352
> > and given that edition 2024 requires 1.85+ I thought the jump was worthwhile. Changing to edition 2024 was not included since that will probably have a few clippy lint changes for the syntax and would be better suited for a follow up PR
> 
> Historically , this project was quite defensive about bumping rust version. Officially we only promised current/current-3, but we were quite conservative and support much older rust version. I am okay with this bump, given it allows us to reduce external crate dependency.

`opentelemetry-prometheus` is 1.75 currently, so wondering why it needs 1.81... If this is just for prometheus crate, we can exclude it (it was excluded from main toml due to planned discontinuation of that, but that is being revisited to be included as part of core crates). There are plans to rewrite it to avoid its external dependencies.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/Cargo.toml#L16C1-L16C24

Issue #2802, opened 2025-03-14
OTLP MetricExporter deadlock issue


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/tonic/metrics.rs#L58

Tonic MetricExporter uses std::Mutex to protect inner client, and the lock is held across await points. This can cause deadlock. The fix would be to use tokio::Mutex, but then shutdown/cleanup cannot be easily done, like already faced by LogExporter. But that is better than having code that deadlocks.

Opening an issue to track fixing this.

cijothomas:
Looks like same issue in HTTP also https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/src/exporter/http/mod.rs#L271

fraillt:
I might be missing something, but current code doesn't hold the lock across the await point...
Given that `inner` is accessed in two places (export and shutdown), I pretty sure that code is correct and it's not possible to deadlock...
What do I miss here?

cijothomas:
> .and_then(|g| match &*g {
>                 Some(client) => Ok(Arc::clone(client)),
>                 _ => Err(OTelSdkError::AlreadyShutdown),
>             })?;

if the guard is dropped here itself, then yes, this looks correct. (I was incorrectly assuming the lock guard is not dropped until export() method ends..

brendanhay:
I've run into the same issue as @jakubdyszkiewicz with `shutdown()` completely blocking the program exit, not only in tests, but anywhere the `tokio` current thread runtime is used. You can omit `shutdown()` and `Drop`ping the `SdkMeterProvider` will also block program exit.

FWIW `#[tokio::test]` uses the `current_thread` runtime by default. If you use `#[tokio::test(flavor = "multi_thread")]` with the appropriate `tokio` features `shutdown()` will make progress and exit.

@jakubdyszkiewicz tangentially, due to how the `MetricExporter::builder()` works the timeout specified to `with_timeout` will not be passed to the http client specified in `with_http_client`. The timeout is only used if one of the internal feature gated http clients is created/used.

cijothomas:
Can you share a minimal repro of the shutdown issue? (do include the cargo.toml file as well so we can which the feature flags that are active.)

scottgerring:
Have confirmed the behaviour from @jakubdyszkiewicz's reproduce above; this is what's happening because the test is using the single-threaded tokio runtime:

[...]

PR #3356, opened 2026-02-19
fix(sdk): resolve OTLP exporter deadlock on single-threaded tokio runtimes

bryantbiggs:
## Problem

OTLP processors deadlock when `shutdown()`/`force_flush()` is called on single-threaded tokio runtimes (or multi-thread with 1 worker, e.g. 1-vCPU k8s pods).

Two root causes:

1. The `experimental_*_with_async_runtime` processor modules spawn tasks on the user's tokio runtime, then call `futures_executor::block_on(oneshot_receiver)` to wait for the response. On single-threaded runtimes this blocks the only available thread — deadlock.

2. The thread-based processors (PeriodicReader, BatchLogProcessor, BatchSpanProcessor) call `futures_executor::block_on(exporter.export(...))` on dedicated OS threads. When the exporter uses tonic/gRPC, the export future needs tokio runtime context — bare `futures_executor` doesn't provide that context, causing hangs or panics depending on the exporter.

Fixes https://github.com/open-telemetry/opentelemetry-rust/issues/2802
Refs: #2643, #2539, #2715, #2071

## Why this approach

Every other OpenTelemetry SDK uses dedicated OS threads for background processing and none of them expose async runtime configuration to users:

- **Go**: spawns a goroutine for `BatchSpanProcessor` and `PeriodicReader`. No async runtime concept exists. Shutdown uses `sync.Once` + `sync.WaitGroup`.
- **Java**: creates dedicated daemon threads (via `DaemonThreadFactory`) or uses `ScheduledExecutorService`. No async runtime exposure.
- **Python**: creates a dedicated daemon `threading.Thread` for batch processing. The SDK is entirely synchronous internally — it does not use `asyncio` at all.
- **.NET**: creates dedicated background threads with `AutoResetEvent`/`ManualResetEvent` for signaling. Despite .NET having native `async/await`, the OTel SDK deliberately uses OS threads to avoid sync-over-async deadlocks.

The Rust SDK is the only OTel implementation that has this deadlock problem because it's the only one where exporters are async (tonic, reqwest) while the SDK's background threads need to call them synchronously. The `experimental_*_with_async_runtime` modules attempted to solve this by integrating with the user's async runtime, but this created the deadlock path described above.

This PR aligns the Rust SDK with every other language implementation: dedicated OS threads for background work, with the tokio runtime context entered via `Handle::enter()` before calling `futures_executor::block_on()`. This makes tokio types (spawn, timers, IO resources) available on the worker threads without taking ownership of the reactor — IO continues to be driven by the runtime's own threads. This avoids the "Cannot drop a runtime in a context where blocking is not allowed" panic that `Handle::block_on()` can trigger when the runtime's lifecycle doesn't perfectly match the worker thread's.

## Changes

**New: `BlockingStrategy` utility** (`util.rs`)
- Captures the tokio runtime handle at construction time (when called from within a tokio context)
- Uses `Handle::enter()` + `futures_executor::block_on()` on worker threads to provide tokio context
- Falls back to plain `futures_executor::block_on()` when no tokio runtime is available

**Updated processors to use `BlockingStrategy`:**
- `BatchSpanProcessor` — created at construction, passed to worker thread
- `BatchLogProcessor` — same pattern
- `PeriodicReader` — stored in `PeriodicReaderInner`, used in `collect_and_export`

**Merged `Tokio`/`TokioCurrentThread`:**
- `Tokio::spawn` now auto-detects runtime flavor via `Handle::try_current()` + `runtime_flavor()`
- Multi-thread: spawns via `tokio::spawn`
- Current-thread: spawns a separate OS thread with its own runtime
- Deleted `TokioCurrentThread` struct

**Removed experimental async runtime modules and features:**
- `experimental_metrics_periodicreader_with_async_runtime` feature + `periodic_reader_with_async_runtime.rs`
- `experimental_logs_batch_log_processor_with_async_runtime` feature + `log_processor_with_async_runtime.rs`
- `experimental_trace_batch_span_processor_with_async_runtime` feature + `span_processor_with_async_runtime.rs`
- `rt-tokio-current-thread` feature
- `runtime_tests.rs`

**Not changed:**
- `SimpleSpanProcessor`/`SimpleLogProcessor` — these run on the caller's thread (possibly a tokio worker) where `Handle::block_on()` would panic, so they keep `futures_executor::block_on()`. This is an inherent limitation of synchronous-on-every-event processors.
- `NoAsync` runtime type — still used by OTLP retry logic
- `opentelemetry-otlp`, `opentelemetry-proto`, or other crates

## Breaking changes

All removed items were behind `experimental_*` feature flags, not stable API.

| Removed | Migration |
|---------|-----------|
| `experimental_metrics_periodicreader_with_async_runtime` feature | Use default thread-based `PeriodicReader` |
| `experimental_logs_batch_log_processor_with_async_runtime` feature | Use default thread-based `BatchLogProcessor` |
| `experimental_trace_batch_span_processor_with_async_runtime` feature | Use default thread-based `BatchSpanProcessor` |
| `rt-tokio-current-thread` feature | Use `rt-tokio` (now auto-detects) |
| `runtime::TokioCurrentThread` struct | Use `runtime::Tokio` |

## Test results

- `cargo check -p opentelemetry_sdk --all-features` — pass
- `cargo clippy -p opentelemetry_sdk --no-default-features -- -Dwarnings` — pass
- `cargo clippy -p opentelemetry_sdk --all-features -- -Dwarnings` — pass
- `cargo test -p opentelemetry_sdk --features="testing"` — 295 passed, 0 failed, 3 ignored (pre-existing)
- `cargo check -p opentelemetry-otlp --all-features` — pass
- `cargo test -p opentelemetry-otlp` — 43 passed

## Merge conflict risk

PRs #3223, #3267, #3257, #3211, #3139, #2962 touch some of the same files. May need coordination on merge order.

### DIFF
--- a/opentelemetry-otlp/tests/integration_test/src/metric_helpers.rs
+++ b/opentelemetry-otlp/tests/integration_test/src/metric_helpers.rs
@@ -3,7 +3,7 @@ use crate::test_utils;
 use anyhow::Result;
 use anyhow::{Context, Ok};
 use opentelemetry_otlp::MetricExporter;
-use opentelemetry_sdk::metrics::{MeterProviderBuilder, SdkMeterProvider};
+use opentelemetry_sdk::metrics::{MeterProviderBuilder, SdkMeterProvider, Temporality};
 use opentelemetry_sdk::Resource;
 use serde_json::Value;
 use std::fs;
@@ -34,6 +34,7 @@ fn create_exporter() -> MetricExporter {
     let exporter_builder = exporter_builder.with_http();
 
     exporter_builder
+        .with_temporality(Temporality::Cumulative)
         .build()
         .expect("Failed to build MetricExporter")
 }
--- a/opentelemetry-proto/tests/json_serde.rs
+++ b/opentelemetry-proto/tests/json_serde.rs
@@ -104,7 +104,8 @@ mod json_serde {
             }
           }
         ],
-        "droppedAttributesCount": 0
+        "droppedAttributesCount": 0,
+        "entityRefs": []
       },
       "scopeSpans": [
         {
@@ -349,7 +350,8 @@ mod json_serde {
             }
           }
         ],
-        "droppedAttributesCount": 1
+        "droppedAttributesCount": 1,
+        "entityRefs": []
       },
       "scopeSpans": [
         {
@@ -911,7 +913,8 @@ mod json_serde {
             }
           }
         ],
-        "droppedAttributesCount": 0
+        "droppedAttributesCount": 0,
+        "entityRefs": []
       },
       "scopeMetrics": [
         {
@@ -1298,7 +1301,8 @@ mod json_serde {
             }
           }
         ],
-        "droppedAttributesCount": 0
+        "droppedAttributesCount": 0,
+        "entityRefs": []
       },
       "scopeLogs": [
         {


scottgerring:
I'd suggest getting the first (the bug) one as a PR and having a chat about the second (what to do with runtime abstractions); it probably needs an ADR as there is a fair bit involved. 

You can also find us in the CNCF slack https://communityinviter.com/apps/cloud-native/cncf and #opentelemetry-rust if you like!

I'm off for the weekend, but I will have cycles next week. Have a good one!

bryantbiggs:
Split the bug fix into a standalone PR as suggested: #3380

That PR contains only the `BlockingStrategy` piece — 4 files, +68/-6 lines. The experimental async runtime removal is excluded and can be discussed separately (likely needs an ADR per Scott's suggestion).

Closing this in favor of #3380.

PR #3360, opened 2026-02-20
fix(otlp): prevent auth tokens from leaking in export error messages

### LINKED ISSUE #3021: [Bug]: OTEL logs raw Authentication token to std-out on auth error

bryantbiggs:
## Summary

Closes #3021
Closes #3108
Closes #3343

When gRPC/HTTP errors occur, `tonic::Status` was Debug-formatted into error messages, leaking Bearer tokens echoed back by the server. The HTTP exporter similarly included full response bodies in error messages.

### Changes

**gRPC exporters** — Two `macro_rules!` macros (`handle_tonic_export_error!`, `handle_interceptor_error!`) defined once in `tonic/mod.rs` and used by all three signal exporters (traces, metrics, logs):
- All export errors: gRPC code, message, and details logged at DEBUG only
- Interceptor errors: always treated as potentially sensitive (DEBUG only)
- `OTelSdkError` returned to callers contains only the gRPC code, never the message
- Signal-specific event names via `concat!()` (e.g., `TonicLogsClient.ExportFailed`)
- No WARN logging in exporters — SDK processors already log returned errors via `otel_error!`

**HTTP exporter** — DEBUG-only logging:
- Network errors: DEBUG with `HttpClient.NetworkError`
- Status errors: status code + response body at DEBUG (`HttpClient.StatusError`)
- No WARN logging in exporters — SDK processors already log returned errors via `otel_error!`

### Review feedback addressed
From #3343:
- Shared macros to eliminate duplicated error handling across signal exporters (@lalitb)
- gRPC code included in all `InternalFailure` and interceptor error messages (@utpilla, @lalitb)
- Distinct HTTP event names for network vs status errors (@utpilla)
- DEBUG caveat in CHANGELOG (@utpilla)

From #3360:
- `Unknown` moved to potentially sensitive errors — too broad, could carry sensitive server responses (@lalitb)
- Removed WARN logging from exporters to avoid duplicate log entries with SDK processor `otel_error!` (@lalitb)
- Merged connection-related and other error paths into single DEBUG log — distinction added complexity without benefit since both use same severity (@cijothomas)
- Fixed comment typo: "don't log here" → "don't log at WARN here" (@cijothomas)

### Test plan
- [x] 5 unit tests for error sanitization macros (token redaction, exporter name, gRPC codes)
- [x] `cargo clippy --all-features -- -D warnings` passes
- [x] `cargo test --all-features` — all tests pass
- [x] `cargo fmt -- --check` passes

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -492,7 +492,7 @@ impl OtlpHttpClient {
         let response = client.send_bytes(request).await.map_err(|e| {
             // Connection errors (e.g., "Connection refused", DNS failures) typically
             // indicate user-side misconfigurations and don't contain sensitive data.
-            // We don't log here because SDK processors (BatchLogProcessor,
+            // We don't log at WARN here because SDK processors (BatchLogProcessor,
             // BatchSpanProcessor, PeriodicReader) already log the returned error
             // via otel_error!.
             otel_debug!(
--- a/opentelemetry-otlp/src/exporter/tonic/mod.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/mod.rs
@@ -421,14 +421,8 @@ where
 #[cfg(any(feature = "trace", feature = "metrics", feature = "logs"))]
 /// Log and convert a `tonic::Status` from a failed export into an `OTelSdkError`.
 ///
-/// For connection-related errors (`Unavailable`, `DeadlineExceeded`,
-/// `ResourceExhausted`, `Aborted`, `Cancelled`), the message typically contains
-/// safe, actionable information (e.g., "Connection refused"), so it is included
-/// alongside the gRPC code in the debug log.
-///
-/// For all other errors — including `Unknown`, `Unauthenticated`, and
-/// `PermissionDenied` — the message and details are logged at DEBUG level only,
-/// since they may contain sensitive information such as authentication tokens
+/// The gRPC code, message, and details are logged at DEBUG level only, since
+/// the message may contain sensitive information such as authentication tokens
 /// echoed back by the server.
 ///
 /// The returned `OTelSdkError` never contains the gRPC message, only the code.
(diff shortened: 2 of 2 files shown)


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -488,7 +488,15 @@ impl OtlpHttpClient {
 
         // Send request
         let response = client.send_bytes(request).await.map_err(|e| {
-            HttpExportError::new(0, format!("Network error: {e:?}")) // Network error
+            // Connection errors (e.g., "Connection refused", DNS failures) typically
+            // indicate user-side misconfigurations and don't contain sensitive data,
+            // so it's safe to log the error message at WARN level.
+            otel_warn!(
```
lalitb: We emit `otel_warn!` here on export failure, but the SDK processors (BatchLogProcessor, BatchSpanProcessor, PeriodicReader, etc.) already log the returned error via `otel_error!`. This would result in duplicate log entries for the same export failure. Should we avoid logging in the macros and rely on the existing SDK-level logging instead?

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -398,6 +398,96 @@ where
...
+/// `$client_name` must be a string literal so that `concat!` can produce
+/// compile-time event names consistent with the codebase naming convention.
+macro_rules! handle_tonic_export_error {
+    ($client_name:literal, $tonic_status:expr) => {{
+        let status = &$tonic_status;
+        let code = status.code();
+        let is_connection_error = matches!(
+            code,
+            tonic::Code::Unavailable
+                | tonic::Code::Unknown
```
lalitb: I think `Unknown` is pretty broad and could carry sensitive server responses. Might be safer to treat it as potentially sensitive.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -490,7 +490,17 @@ impl OtlpHttpClient {
 
         // Send request
         let response = client.send_bytes(request).await.map_err(|e| {
-            HttpExportError::new(0, format!("Network error: {e:?}")) // Network error
+            // Connection errors (e.g., "Connection refused", DNS failures) typically
+            // indicate user-side misconfigurations and don't contain sensitive data.
+            // We don't log here because SDK processors (BatchLogProcessor,
```
cijothomas: typo I guess.. You probably meant we don't log at high severity level here?

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -418,6 +418,94 @@ where
...
+#[cfg(any(feature = "trace", feature = "metrics", feature = "logs"))]
+/// Log and convert a `tonic::Status` from a failed export into an `OTelSdkError`.
+///
+/// For connection-related errors (`Unavailable`, `DeadlineExceeded`,
+/// `ResourceExhausted`, `Aborted`, `Cancelled`), the message typically contains
+/// safe, actionable information (e.g., "Connection refused"), so it is included
+/// alongside the gRPC code in the debug log.
+///
+/// For all other errors — including `Unknown`, `Unauthenticated`, and
+/// `PermissionDenied` — the message and details are logged at DEBUG level only,
```
cijothomas: since both connection-related and others are logged only at DEBUG level, do we need to distinguish them? We can avoid the complexity by merging them to single path.

opentelemetry-otlp/src/exporter/http/mod.rs
bryantbiggs: Good catch — fixed. Changed to "We don't log at WARN here" since we do log at DEBUG right below.

opentelemetry-otlp/src/exporter/tonic/mod.rs
bryantbiggs: Agreed — merged into a single path that always logs code + message + details at DEBUG. The connection vs. sensitive distinction was adding complexity without benefit since both used the same severity level.

PR #3361, opened 2026-02-20
docs(adr): propose 12-month trailing window MSRV policy

bryantbiggs:
## Summary

Proposes replacing the current N-3 version-count MSRV policy with a 12-month trailing window via a new ADR (`docs/adr/002_msrv_policy.md`).

- The MSRV may be raised to any stable Rust version that is at least 12 months old
- A single uniform MSRV applies across all published workspace crates
- MSRV bumps are not semver-breaking and are shipped in minor releases only

## Context

This was requested during the discussion on #3352. The ADR includes:

- Ecosystem research showing tokio, hyper, and tower all use a 6-month rolling window — our 12-month proposal is twice as conservative
- A survey of how other OTel language SDKs (Go, Python, Node.js, .NET, Java) handle minimum version policies, with links to each project's documented policy
- Two alternatives considered (6-month window and the current N-3 policy) with rationale for why 12 months is the better fit for an infrastructure crate

## Practical impact

Under this policy the MSRV could immediately move from 1.75 (December 2023) to 1.85 (February 2025), enabling removal of `once_cell`, `lazy_static`, `num-format`, and `num_cpus` in favor of std APIs, plus eligibility for Edition 2024 and the MSRV-aware dependency resolver.

### DIFF
--- a/docs/adr/002_msrv_policy.md
+++ b/docs/adr/002_msrv_policy.md
@@ -1,0 +1,175 @@
+# MSRV (Minimum Supported Rust Version) Policy
+
+## Date
+
+20 Feb 2026
+
+## Summary
+
+This ADR proposes replacing the current "N-3" version-count MSRV policy with a
+12-month trailing window. Under this policy, the MSRV may be raised to any
+stable Rust release whose release date is at least 12 months prior to the date
+the MSRV bump PR is opened. This applies uniformly across all published
+workspace crates. MSRV bumps are not considered semver-breaking changes.
+
+## Motivation
+
+The current MSRV policy states that "the current stable Rust compiler and the
+three most recent minor versions before it will always be supported." Given
+Rust's 6-week release cadence, N-3 equates to roughly 4.5 months — but the
+actual MSRV (1.75, released December 2023) is over two years old as of this
+writing. The stated policy and practice are misaligned.
+
+A time-based policy addresses this in a few ways:
+
+- It is easier to reason about. Users and maintainers can determine eligibility
+  from a release date without needing to know Rust's release cadence or the
+  current stable version number.
+- It is more predictable for downstream consumers. A 12-month window gives
+  library authors and end users a full year to upgrade their toolchain before
+  they are affected.
+- It aligns with the broader Rust ecosystem. The most widely-adopted formal
+  MSRV policies in the Rust ecosystem are time-based (see
+  [Ecosystem Context](#ecosystem-context) below).
+- It is consistent with OpenTelemetry cross-language guidance. The OpenTelemetry
+  specification states: "Each language implementation SHOULD define how the
+  removal of a supported language/runtime version affects its versioning. As a
+  rule of thumb, it SHOULD follow the conventions in the given ecosystem."
+  ([source](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#language-version-support))
+
+## Policy
+
+### The rule
+
+The MSRV may be raised to any stable Rust version whose release date is at
+least 12 months prior to the date the MSRV bump PR is opened.
+
+### Scope
+
+A single, uniform MSRV applies to all published crates in the workspace.
+Examples, stress tests, and internal test crates are excluded.
+
+### Semver treatment
+
+MSRV bumps are not semver-breaking changes. They are shipped in minor releases
+only, never in patch releases.
+
+The `rust-version` field in each crate's `Cargo.toml` is the source of truth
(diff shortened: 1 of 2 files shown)


scottgerring:
Hey @bantonsson and @paullegranddc ; for one _helpful downstream anecdote_ - do we have a view on the Datadog side on MSRV? I note that [dd-trace-rs](https://github.com/DataDog/dd-trace-rs/) is up at 1.84 vs the current MSRV of 1.75. 

Also @open-telemetry/rust-approvers I think we also need to carefully consider what our MSRV policy should be. 

On the one hand it feels like we are quite out of date and as @bryantbiggs has pointed out this has led to us missing out on some pieces on `std` we could otherwise leverage, on the other, I expect that it is healthy for us to aim to drag our MSRV window out for as long as possible. Said otherwise, perhaps rather than promising to _advance_ it, we should promise _not to advance_ it, outside of reasons $X,Y,Z ? 

I will also share this in the slack to try solicit feedback !

docs/adr/002_msrv_policy.md
```diff
@@ -0,0 +1,175 @@
...
+
+| Date | Current Stable | 12-Month Window Allows |
+|------|---------------|------------------------|
+| Feb 2026 (today) | 1.93 | 1.85 (Feb 2025) |
+| Aug 2026 | ~1.98 | 1.89 (Aug 2025) |
+| Feb 2027 | ~1.102 | 1.93 (Jan 2026) |
+
+## Ecosystem context
+
+### Rust ecosystem
```
cijothomas, line 122: despite their documented policy, they are all actually quite conservative and some are on 1.6* today.

bryantbiggs: that is correct!

lalitb:
The ADR is well-researched, but I think we should be careful about the gap between policy and practice.

   The ADR cites tokio (6-month policy), hyper (6-month), and tower (6-month) as precedent. But their actual MSRVs tell a different story:

   | Crate | Stated Policy | Actual MSRV | Age |
   |-------|--------------|-------------|-----|
   | tokio | 6 months | 1.71 | ~2 years |
   | hyper | 6 months | 1.63 | ~3 years |
   | tower | 6 months | 1.64 | ~3 years |
   | tonic | — | 1.75 | ~2 years |

   Even projects with aggressive policies bump conservatively in practice. We should learn from that.

   I'm supportive of documenting a guideline, but I'd prefer it to be framed as a **recommendation rather than a policy that gets mechanically enforced.** The 12-month window should communicate *what's  permissible* to the community, not create an expectation that bumps will happen on a schedule. Maintainers should still exercise judgment - bumps should be motivated by concrete need (blocking dependency, std API replacing an external crate, security fix), not just because the window allows it.

I'd also suggest separating the guideline from the actual bump. There's no urgency to bump right now.

lalitb:
You're right that the ADR says "may" not "will" - I was reacting more to the "Practical impact" section which seems to be read as "here's what we'd do immediately." That framing made it feel like the policy and the first bump were being proposed together. I'd suggest reframing that section to read as illustrative rather than prescriptive. Something like:

   > **Practical impact**
   >
   > As an example, if this policy were adopted today, the 12-month window would permit raising the MSRV up to 1.85 (February 2025). Some capabilities that would become available at various MSRV levels
  include:
   >
   > - **1.80+**: `std::sync::LazyLock` — could replace `once_cell` / `lazy_static`
   > - **1.81+**: `#[expect(lint)]`, `core::error::Error`
   > - **1.85+**: Edition 2024 eligibility (MSRV-aware resolver, stricter `unsafe_op_in_unsafe_fn` defaults)
   >
   > Any actual MSRV bump would be proposed and evaluated separately based on concrete need at the time.

   This keeps the section useful as context without implying an immediate bump to 1.85. It also avoids the `num_cpus`/`num-format` claims — `num_cpus` could already be replaced at 1.75
  (`available_parallelism` was stabilized in 1.59), and `num-format` may not be in the dependency tree (not sure through).

>    "what is not clear is what is holding that up... So what we should take away is - we can see what others have for a policy, but they are not truly following it. Why?"


   On the "why not follow the policy" question - I think the answer for tokio et al. is simply that the cost of bumping (downstream churn) usually outweighs the benefit unless there's a concrete motivator. The policy exists as a social contract ("we won't go newer than X"), not as a schedule. I think we're actually aligned here  - I just want to make sure the ADR text reads that way to future contributors who might interpret a 12-month window as encouragement to bump proactively.

 The 12-month window as a ceiling is reasonable. No objections to the policy itself.

PR #3363, opened 2026-02-20
feat(otlp): support per-signal protocol environment variables

bryantbiggs:
## Summary

- Add support for `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`, `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL`, and `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL` environment variables per the [OTLP exporter specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/)
- Refactor `ExportConfig.protocol` from `Protocol` to `Option<Protocol>` to distinguish explicit configuration from defaults (consistent with `endpoint` and `timeout` fields)
- Add `Protocol::parse_from_env_var()` and `Protocol::feature_default()` for generalized protocol resolution
- Add `resolve_protocol()` with proper priority: **code config > signal env var > generic env var > feature default**
- Auto-select `build()` on `SpanExporterBuilder`, `LogExporterBuilder`, and `MetricExporterBuilder` now uses `resolve_protocol()` with the signal-specific env var, so `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc` correctly routes to tonic transport
- Transport/protocol mismatch validation in both directions:
  - HTTP builder returns `InvalidConfig` when gRPC protocol is configured
  - gRPC/tonic builder returns `InvalidConfig` when an HTTP protocol is configured
- **Breaking**: `Protocol::default()` no longer consults env vars — returns only the feature-based default. Use `Protocol::from_env()` for env var resolution.

### PR Series

This is **PR 1 of 3** for full OTLP exporter spec compliance:
1. **Per-signal protocol env vars** (this PR) — 3 new env vars
2. TLS certificate env vars — 12 new env vars (supersedes stalled #2465)
3. INSECURE env var support — 4 new env vars

Partially addresses #774, #984.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p opentelemetry-otlp --all-features -- -D warnings`
- [x] `cargo test -p opentelemetry-otlp --all-features` — 87 lib tests + 21 doc tests pass
- [x] Compiles cleanly with individual transport features (`grpc-tonic`, `http-proto`, `http-json`)
[...]

PR #2465, opened 2024-12-23
feat: configure TLS with environment variables.

jvanz:
Updates the opentelemetry-otlp crate to allow users to configure TLS using environment variables. Removing the need to crating the TLS config object and defining it with the `with_tls_config` method. In the same way other OTLP libraries does (e.g. go lang).

Partially fixes #774 

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

Issue #984, opened 2023-03-07
Support more env vars for exporter configuration


brennan-airtime:
It would be nice to configure things like `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_CERTIFICATE` and `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE` like in the spec.

https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/
https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md

brennan-airtime:
For sure. Just wanted to reach out first to make sure I'm not barking up the wrong tree or I didn't miss something obvious.

cijothomas:
@brennan-airtime Just checking if you are planning to work on this?

brennan-airtime:
I do still want to do it, but time is a limiting factor...

kellpossible:
It looks like `OTEL_EXPORTER_OTLP_HEADERS` is now supported?

Issue #774, opened 2022-04-06
Support all configuration of otlp exporter


TommyCpp:
Add support for all configurations in [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#opentelemetry-protocol-exporter)

- [ ] Endpoint (OTLP/HTTP) 
      - It was partially supported with #773 missing
- [x] Endpoint (OTLP/gRPC)
- [ ] Insecure
- [ ] Certificate File
- [ ] Client key file
- [ ] Client certificate file
- [ ] Headers
- [ ] Compression
- [ ] Timeout
- [ ] Protocol

TommyCpp:
> Does `opentelemtry-otlp` already add the signal path automatically when using gRPC?

gRPC doesn't really require a path, right? The service/client stub should take care of that

TylerHelmuth:
You're totally right, my bad

atanunq:
Hey all! I'm interested in trying to implement one of the items in the list as a way to get more familiar with the repository. `Certificate file` looks like a good candidate. However, I am struggling with seeing the configuration flow in general. Could you please share some pointers on how to go about implementing it?

For example, I am testing locally with `tonic`. I managed to connect over TLS to a collector using `new_exporter().tonic().with_tls_config(ClientTlsConfig::new(...))`. However, this is `tonic`-specific. Based on the linked spec above, I assume the new way of configuring should be taken into account within all protocols (i.e different third party crates). Is that correct? 

If that is the case, the `ExportConfig` struct looks like a good place to add it. Then, though, `ExportConfig` and `TonicConfig.tls_config` may have different values and one needs to take precedence over the other. How/where would that be decided?

I do hope any of that makes sense. Thanks!

hdost:
Related https://github.com/open-telemetry/opentelemetry-rust/issues/984

scottgerring:
Hey @TommyCpp , I'm trying to lock down what we need to mark OTLP stable and have picked through this again. I think we can close this tracking issue in favour of some other bits and pieces:

   ### Complete
   - **Endpoint (HTTP + gRPC)** — implemented, including signal-specific variants and auto-appended signal paths
   - **Headers** — implemented via `OTEL_EXPORTER_OTLP_HEADERS` + signal-specific variants
   - **Compression** — implemented via `OTEL_EXPORTER_OTLP_COMPRESSION` (gzip + zstd)

[...]

PR #3364, opened 2026-02-20
feat(otlp): support TLS certificate environment variables

bryantbiggs:
## Summary

- Add support for 12 TLS-related environment variables per the [OTLP exporter specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/):
  - `OTEL_EXPORTER_OTLP_CERTIFICATE` / `_CLIENT_KEY` / `_CLIENT_CERTIFICATE` (generic)
  - `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` / `_CLIENT_KEY` / `_CLIENT_CERTIFICATE`
  - `OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE` / `_CLIENT_KEY` / `_CLIENT_CERTIFICATE`
  - `OTEL_EXPORTER_OTLP_LOGS_CERTIFICATE` / `_CLIENT_KEY` / `_CLIENT_CERTIFICATE`
- These env vars point to PEM-encoded files for server TLS verification and mTLS client authentication
- Signal-specific vars take precedence over generic vars; programmatic `.with_tls_config()` takes full precedence over all env vars
- For tonic (gRPC): builds `ClientTlsConfig` from env vars when no programmatic TLS config is set
- For HTTP/reqwest: applies CA cert and mTLS identity to the reqwest client builder (both async and blocking)
- Partial mTLS config (only key or only cert set) logs a warning and skips mTLS

**PR 2 of 3** for OTLP exporter spec compliance. Merge order:
1. #3363 — per-signal protocol env vars
2. **This PR** — TLS certificate env vars
3. PR 3 (upcoming) — INSECURE env vars

Supersedes #2465 (stalled for 10+ months with merge conflicts).

## Test plan

- [x] `cargo clippy -p opentelemetry-otlp --all-targets --all-features -- -D warnings` passes
- [x] `cargo test -p opentelemetry-otlp --all-features` — 106 unit + 2 integration + 13 doc-tests pass
- [ ] Manual verification with actual TLS certificates against an OTLP collector
- [ ] Verify env var resolution priority: code config > signal env var > generic env var

Refs: #774, #984

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -46,6 +46,13 @@ Released 2026-May-08
   grpc-tonic). Protocol resolution from environment variables is handled internally by the
   exporter builders. Users who relied on `Protocol::default()` to read env vars should use
   `Protocol::from_env()` instead.
+- Add support for TLS certificate environment variables per the OTLP exporter spec:
+  `OTEL_EXPORTER_OTLP_CERTIFICATE`, `OTEL_EXPORTER_OTLP_CLIENT_KEY`,
+  `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE` (generic), plus per-signal variants for traces,
+  metrics, and logs. These env vars point to PEM-encoded files for server TLS verification and
+  mTLS client authentication. Programmatic `.with_tls_config()` takes precedence over env vars.
+  [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
+  [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
 - Add support for `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
   to configure metrics temporality. Accepted values: `cumulative` (default), `delta`,
   `lowmemory` (case-insensitive). Programmatic `.with_temporality()` overrides the env var.
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## vNext
 
+- Add support for TLS certificate environment variables per the OTLP exporter spec:
+  `OTEL_EXPORTER_OTLP_CERTIFICATE`, `OTEL_EXPORTER_OTLP_CLIENT_KEY`,
+  `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE` (generic), plus per-signal variants for traces,
+  metrics, and logs. These env vars point to PEM-encoded files for server TLS verification and
+  mTLS client authentication. Programmatic `.with_tls_config()` takes precedence over env vars.
+  For HTTP, the env vars are only applied to auto-created `reqwest` clients; hyper and
+  user-supplied clients (via `.with_http_client()`) must configure TLS themselves.
+  [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
+  [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
+
 ## 0.32.0
 
 Released 2026-May-08
@@ -46,13 +56,6 @@ Released 2026-May-08
   grpc-tonic). Protocol resolution from environment variables is handled internally by the
   exporter builders. Users who relied on `Protocol::default()` to read env vars should use
   `Protocol::from_env()` instead.
-- Add support for TLS certificate environment variables per the OTLP exporter spec:
-  `OTEL_EXPORTER_OTLP_CERTIFICATE`, `OTEL_EXPORTER_OTLP_CLIENT_KEY`,
-  `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE` (generic), plus per-signal variants for traces,
-  metrics, and logs. These env vars point to PEM-encoded files for server TLS verification and
-  mTLS client authentication. Programmatic `.with_tls_config()` takes precedence over env vars.
-  [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
-  [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
 - Add support for `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
   to configure metrics temporality. Accepted values: `cumulative` (default), `delta`,
   `lowmemory` (case-insensitive). Programmatic `.with_temporality()` overrides the env var.
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -184,17 +184,14 @@ struct ReqwestTlsCerts {
 ))]
 impl ReqwestTlsCerts {
     /// Apply TLS certs to an async reqwest client builder.
-    fn apply(
-        &self,
-        mut builder: reqwest::ClientBuilder,
-    ) -> Result<reqwest::ClientBuilder, ExporterBuildError> {
+    fn apply(&self, mut builder: reqwest::ClientBuilder) -> reqwest::ClientBuilder {
         if let Some(ref cert) = self.ca_cert {
             builder = builder.add_root_certificate(cert.clone());
         }
         if let Some(ref identity) = self.identity {
             builder = builder.identity(identity.clone());
         }
-        Ok(builder)
+        builder
     }
 
     /// Apply TLS certs to a blocking reqwest client builder.
@@ -298,7 +295,7 @@ impl HttpExporterBuilder {
 
                 #[cfg(any(feature = "reqwest-rustls", feature = "reqwest-rustls-webpki-roots"))]
                 if let Some(ref certs) = tls_certs {
-                    builder = certs.apply(builder)?;
+                    builder = certs.apply(builder);
                 }
 
                 http_client =
@@ -400,8 +397,7 @@ impl HttpExporterBuilder {
             OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE, OTEL_EXPORTER_OTLP_CLIENT_KEY,
         };
 
-        let ca_pem =
-            resolve_tls_env_and_read(tls_vars.cert_var, OTEL_EXPORTER_OTLP_CERTIFICATE)?;
+        let ca_pem = resolve_tls_env_and_read(tls_vars.cert_var, OTEL_EXPORTER_OTLP_CERTIFICATE)?;
         let client_key_pem =
             resolve_tls_env_and_read(tls_vars.client_key_var, OTEL_EXPORTER_OTLP_CLIENT_KEY)?;
         let client_cert_pem = resolve_tls_env_and_read(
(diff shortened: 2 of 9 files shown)


fredrik-jansson-se:
Cool, looking forward to this, needed in our setup. Thanks!!

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -73,6 +73,29 @@ pub(crate) struct TonicConfig {
     pub(crate) retry_policy: Option<RetryPolicy>,
 }
 
+impl std::fmt::Debug for TonicConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
scottgerring, line 77: This looks like it addresses the CodeQL leak, so from a security perspective, I think this specific concern is resolved.

nit: this leaves logs slightly inconsistent with `SpanExporterBuilt` and `MetricExporterBuilt`; do we need to log the result in this case?

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -46,6 +46,13 @@ Released 2026-May-08
   grpc-tonic). Protocol resolution from environment variables is handled internally by the
   exporter builders. Users who relied on `Protocol::default()` to read env vars should use
   `Protocol::from_env()` instead.
+- Add support for TLS certificate environment variables per the OTLP exporter spec:
```
scottgerring: Can we move this into `vNext`

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -317,13 +341,24 @@ impl TonicExporterBuilder {
...
             Some(tls_config) => endpoint
                 .tls_config(tls_config)
                 .map_err(|er| ExporterBuildError::InternalFailure(er.to_string()))?,
-            None if is_https => endpoint
-                .tls_config(ClientTlsConfig::new())
-                .map_err(|er| ExporterBuildError::InternalFailure(er.to_string()))?,
-            None => endpoint,
+            None => {
+                // Try to build TLS config from environment variables
+                let tls_config_from_env = Self::build_tls_config_from_env(&signal_tls_vars)?;
```
scottgerring, line 350: This compiles when any of `tls`, `tls-ring`, `tls-aws-lc`, `tls-provider-agnostic` are set, but if we jump over to `build_tls_config_from_env`, we're only including it for 

`tls`, `tls-ring`, or `tls-aws-lc`. So this combo fails:

```
  cargo check --manifest-path=opentelemetry-otlp/Cargo.toml \
     --no-default-features \
     --features grpc-tonic,tls-provider-agnostic,logs
 ```

I think either the helper cfg needs to include tls-provider-agnostic too, or the call site needs to avoid env-based TLS config for provider-agnostic builds.

File: opentelemetry-otlp/src/lib.rs [hunk 4]
```diff
@@ -224,6 +224,9 @@
 //! | `OTEL_EXPORTER_OTLP_TIMEOUT` | Maximum wait time (in milliseconds) for the backend to process each batch. | `10000` |
 //! | `OTEL_EXPORTER_OTLP_HEADERS` | Key-value pairs for request headers. Format: `key1=value1,key2=value2`. Values are URL-decoded. | (none) |
 //! | `OTEL_EXPORTER_OTLP_COMPRESSION` | Compression algorithm. Valid values: `gzip`, `zstd`. | (none) |
+//! | `OTEL_EXPORTER_OTLP_CERTIFICATE` | Path to PEM-encoded TLS certificate for server verification. | (none) |
```
scottgerring, line 227: Should we mention the caveats under which these are usable here? E.g. HTTP support is reqwest-only; custom/hyper clients must configure TLS themselves.

scottgerring: I also note that the feature prioritisation can make things surprising.

E.g. - if a user has default features and adds hyper-client, they get hyper selected over the default blocking reqwest client and TLS env vars are not applied. In that path, TLS env vars are not applied. The docs mention async HTTP clients shouldn’t be used with default SDK processors, which is good, but i'm wondering what we can do to reduce making this more "obvious" ?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -220,6 +268,13 @@ impl HttpExporterBuilder {
 
         let timeout = resolve_timeout(signal_timeout_var, self.exporter_config.timeout.as_ref());
 
+        // Resolve and parse TLS certificates from env vars (only when TLS features and a reqwest client are enabled)
+        #[cfg(all(
+            any(feature = "reqwest-rustls", feature = "reqwest-rustls-webpki-roots"),
+            any(feature = "reqwest-client", feature = "reqwest-blocking-client"),
+        ))]
+        let tls_certs = Self::resolve_reqwest_tls_certs(&signal_tls_vars)?;
```
scottgerring: I think this happens before we even know if we're going to need to resolve TLS env vars. For example, a custom .with_http_client(...) or a selected hyper-client won't use these certs, but invalid TLS env vars would still  make build() fail here. 

Could we move this resolution into the reqwest auto-client branches where the certs are actually applied?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -229,15 +284,22 @@ impl HttpExporterBuilder {
...
+                #[allow(unused_mut)] // mut needed when TLS features are enabled
+                let mut builder = reqwest::Client::builder().timeout(timeout);
+
+                #[cfg(any(feature = "reqwest-rustls", feature = "reqwest-rustls-webpki-roots"))]
+                if let Some(ref certs) = tls_certs {
+                    builder = certs.apply(builder);
+                }
+
+                http_client =
+                    Some(Arc::new(builder.build().unwrap_or_default()) as Arc<dyn HttpClient>);
```
scottgerring: This fallback was pre-existing, but with TLS env vars applied to the builder it gets a bit more surprising: if `builder.build()` fails after adding the configured certs, we silently fall back to a default reqwest client and lose the TLS config. Perhaps we should return an `ExporterBuildError` instead so invalid TLS/client config is visible? Same for the blocking path below.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -308,6 +383,67 @@ impl HttpExporterBuilder {
...
+                Some(reqwest::Identity::from_pem(&identity_pem).map_err(|e| {
+                    ExporterBuildError::InternalFailure(format!(
+                        "Failed to parse client identity: {e}"
+                    ))
+                })?)
+            }
+            (Some(_), None) | (None, Some(_)) => {
+                opentelemetry::otel_warn!(
+                    name: "TlsConfig.PartialMtls",
+                    message = "mTLS requires both CLIENT_KEY and CLIENT_CERTIFICATE to be set; only one was provided, skipping mTLS configuration"
```
scottgerring: nit: do we want to be more specific?

scottgerring (changes requested):
Hey @bryantbiggs thanks for this! The march towards OTLP stability progresses. 
Couple minor comments inline; it will be good to get this and the other env var one in soon :)

opentelemetry-otlp/src/exporter/tonic/mod.rs
bryantbiggs: Dropped the `result` field from `LogExporterBuilt` so it matches Span/Metric. Left the redacting `Debug` impl in place as defense-in-depth.

opentelemetry-otlp/CHANGELOG.md
bryantbiggs: Moved to vNext.

opentelemetry-otlp/src/exporter/tonic/mod.rs
bryantbiggs: Added `tls-provider-agnostic` to the helper's cfg and the shared `resolve_tls_env_and_read`. Your repro compiles now; covered by `cargo hack --each-feature`.

opentelemetry-otlp/src/lib.rs
bryantbiggs: Added a "TLS certificate caveats" section covering gRPC vs HTTP and the reqwest-only / self-configure behavior and a build-time otel_warn! (TlsConfig.EnvVarsIgnored) when env vars are set but the active client can't apply them, plus a note in the docs.

opentelemetry-otlp/src/exporter/http/mod.rs
bryantbiggs: Moved cert resolution into the reqwest async/blocking branches, so a custom or hyper client no longer fails the build on certs it won't use.

opentelemetry-otlp/src/exporter/http/mod.rs
bryantbiggs: Both paths now return `ExporterBuildError` on build failure instead of falling back to a default client.

opentelemetry-otlp/src/exporter/http/mod.rs
bryantbiggs: Now names the specific variable that was set (key-only vs cert-only).

bryantbiggs:
Addressed the inline comments — pushed fixes for all of them, plus a few related correctness fixes to the TLS env handling (empty signal-var now falls back to the generic var, PEM cert/key boundary safety, gRPC TLS applied only to `https://` endpoints, and a warning when an auto-created reqwest client is built without a `reqwest-rustls` feature).

While making these changes I noticed a couple of separate issues. I've put each in its own commit so they can be carried along with this PR as-is, or I can move them into separate PR(s) if you'd rather keep this one tightly scoped — your call:

- [`94aea5d`](https://github.com/open-telemetry/opentelemetry-rust/pull/3364/commits/94aea5db62edbf1cd2e358bb6553ffdd2db84255) — a pre-existing compile break (already on `main`, unrelated to TLS): `cargo build --features hyper-client` fails with E0433 because the `HyperClient` import's cfg didn't match its use-site. One-line cfg fix.
- [`9ad33d1`](https://github.com/open-telemetry/opentelemetry-rust/pull/3364/commits/9ad33d19d25a9b4a5b1075e9e980ec8563324d12) — minor: a partial mTLS pair (only key or only cert set) returned `Some(ReqwestTlsCerts { ca_cert: None, identity: None })` instead of `None`. Benign, just tidier.

One more I'll flag without changing (no code): `OTEL_EXPORTER_OTLP_CERTIFICATE` has different trust semantics per transport — reqwest's `add_root_certificate` is additive (system roots stay trusted), while tonic's `ClientTlsConfig` makes the provided cert the sole root. Making reqwest match tonic isn't small on reqwest 0.13 (no `tls_built_in_root_certs` toggle — it'd need `use_preconfigured_tls` with a hand-built rustls config). I can open a follow-up issue, or is additive-on-HTTP acceptable to you?

tj-cisco:
Does this mean this support has been abandoned?

bryantbiggs:
yes, I have abandoned this - it simply takes too long between cycles and very challenging to get anything merged. however, anyone else is free to take up the challenge and carry it forward if they have the time

tj-cisco:
@bryantbiggs I understand, working with OpenSource can be frustrating.  Would you be able to write up a brief summary of this PR and the outstanding items so that myself or someone else can pickup from where you left off?

PR #3365, opened 2026-02-20
feat(otlp): support INSECURE environment variables for gRPC

### LINKED ISSUE #774: Support all configuration of otlp exporter

### LINKED ISSUE #984: Support more env vars for exporter configuration

bryantbiggs:
## Summary

Add support for 4 INSECURE environment variables per the [OTLP exporter specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/) (SHOULD-level):

- `OTEL_EXPORTER_OTLP_INSECURE` (generic)
- `OTEL_EXPORTER_OTLP_TRACES_INSECURE`
- `OTEL_EXPORTER_OTLP_METRICS_INSECURE`
- `OTEL_EXPORTER_OTLP_LOGS_INSECURE`

Per the spec, INSECURE only applies to gRPC (not HTTP — HTTP security is determined by URL scheme). There is intentionally no programmatic builder method — this is env-var-only per spec.

### Behavior

When an endpoint has **no explicit scheme** (e.g., `collector.example.com:4317`):
- `INSECURE=true` → prepends `http://` (plaintext)
- `INSECURE=false` (default) → prepends `https://` and auto-applies `ClientTlsConfig::new()` when TLS features are enabled

Endpoints with an explicit `http://` or `https://` scheme are **unaffected** (scheme detection is case-insensitive).

Signal-specific env var takes precedence over generic; `"true"` (case-insensitive) = insecure, anything else = secure.

### Breaking behavior change

Schemeless endpoints (e.g., `collector.example.com:4317`) now default to `https://` instead of being passed as-is to `Channel::from_shared`. Users with schemeless endpoints connecting to plaintext gRPC servers should set `OTEL_EXPORTER_OTLP_INSECURE=true`. Endpoints with an explicit `http://` scheme (including the default `http://localhost:4317`) are unaffected.

## Test plan

- [x] `cargo clippy -p opentelemetry-otlp --all-targets --all-features -- -D warnings` passes
- [x] `cargo test -p opentelemetry-otlp --all-features` — 117 unit + 2 integration + 21 doc-tests pass
- [x] 5 unit tests for `resolve_insecure()`: signal overrides generic, default is false, case-insensitive, non-true values return false, falls back to generic
- [x] 4 integration tests for `build_channel` with INSECURE: schemeless+secure errors without TLS, schemeless+insecure succeeds, schemeless defaults to https, explicit http:// ignores INSECURE env
- [x] All tests use `temp_env::with_var_unset` for signal-specific var to prevent parallel test interference

Resolves: #774
Resolves: #984


opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -279,12 +280,29 @@ impl TonicExporterBuilder {
...
+            .is_some_and(|p| p.eq_ignore_ascii_case("https://"))
+            || endpoint_str
+                .get(..7)
+                .is_some_and(|p| p.eq_ignore_ascii_case("http://"));
+        if !has_scheme {
+            if insecure {
+                endpoint_str = format!("http://{endpoint_str}");
+            } else {
+                endpoint_str = format!("https://{endpoint_str}");
+            }
```
scottgerring: I think this should check for any explicit scheme, not only `http://`/`https://`. We already support `unix:///tmp/test`, and this PR currently rewrites it to `https://unix:///tmp/test`.

For this purpose, checking for `://` before prepending seems safer than special-casing just HTTP(S), while still treating `collector.example.com:4317` as schemeless.

There's an existing test that covers this, but it seems to be excluded by the CI suite:

https://github.com/open-telemetry/opentelemetry-rust/blob/f7b0dd999d3d6434a6e52291efccc98e9dfcea81/opentelemetry-otlp/src/exporter/tonic/mod.rs#L1233-L1247

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -46,6 +46,16 @@ Released 2026-May-08
...
+- Add support for INSECURE environment variables for gRPC (env-var-only, no builder method, per spec):
+  `OTEL_EXPORTER_OTLP_INSECURE` (generic), `OTEL_EXPORTER_OTLP_TRACES_INSECURE`,
+  `OTEL_EXPORTER_OTLP_METRICS_INSECURE`, `OTEL_EXPORTER_OTLP_LOGS_INSECURE`.
+  Per the spec, these only apply to gRPC connections. When an endpoint has no explicit scheme,
+  `INSECURE=true` uses `http://`, `INSECURE=false` (default) uses `https://` with auto-TLS.
+  **Note:** Schemeless endpoints (e.g., `collector.example.com:4317`) now default to `https://`
+  instead of being passed as-is. Set `OTEL_EXPORTER_OTLP_INSECURE=true` for plaintext connections.
+  Endpoints with an explicit `http://` or `https://` scheme are unaffected.
+  [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
+  [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
```
scottgerring: Could you move this changelog entry to `vNext`?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -38,6 +38,13 @@ pub const OTEL_EXPORTER_OTLP_PROTOCOL_HTTP_JSON: &str = "http/json";
 
 /// Max waiting time for the backend to process each signal batch, defaults to 10 seconds.
 pub const OTEL_EXPORTER_OTLP_TIMEOUT: &str = "OTEL_EXPORTER_OTLP_TIMEOUT";
+
+/// Whether to enable client transport security for the exporter's gRPC connection.
+/// Per the spec, this only applies to gRPC. For HTTP, security is determined by the URL scheme.
+/// There is intentionally no programmatic builder method — this is env-var-only per the
+/// [OTLP exporter spec](https://opentelemetry.io/docs/specs/otel/protocol/exporter/).
+/// Default: `false` (TLS is used).
```
scottgerring, line 48: Nit: this wording reads backwards for an `INSECURE` variable. Could we align each instance of it with the clearer phrasing in `lib.rs`, e.g. "Whether to disable TLS for gRPC?" 


Perhaps something like: 

 ```suggestion
 /// Whether to disable TLS for the exporter's gRPC connection.
 /// Per the [OTLP exporter spec](https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options),
 /// this only applies to gRPC when the endpoint has no explicit `http` or `https` scheme.
 /// HTTP security is determined by the endpoint URL scheme.
 /// Default: `false` (TLS is used).
 ```

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -1068,6 +1089,127 @@ mod tests {
...
     }
 
+    #[test]
+    #[cfg(not(any(
+        feature = "tls",
+        feature = "tls-ring",
+        feature = "tls-aws-lc",
+        feature = "tls-provider-agnostic"
+    )))]
+    fn test_schemeless_endpoint_insecure_false_errors_without_tls() {
```
scottgerring: This needs a tokio runtime and fails on my machine (aside: why is this and the other one not running in CI - should be using `cargo hack` for tests too?) 

```suggestion
     #[tokio::test]
    fn test_schemeless_endpoint_insecure_false_errors_without_tls() {
```

opentelemetry-otlp/src/logs.rs
```diff
@@ -32,6 +32,9 @@ pub const OTEL_EXPORTER_OTLP_LOGS_TIMEOUT: &str = "OTEL_EXPORTER_OTLP_LOGS_TIMEO
 pub const OTEL_EXPORTER_OTLP_LOGS_HEADERS: &str = "OTEL_EXPORTER_OTLP_LOGS_HEADERS";
 /// Protocol to use for log exports. Valid values: `grpc`, `http/protobuf`, `http/json`.
 pub const OTEL_EXPORTER_OTLP_LOGS_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_LOGS_PROTOCOL";
+/// Whether to enable client transport security for gRPC log exports.
```
scottgerring: Another instance where I think we could invert the phrasing!

opentelemetry-otlp/src/exporter/tonic/mod.rs
scottgerring: Looking more into this I think non-blocking for this PR: `cargo hack test -p opentelemetry-otlp --each-feature --lib` fails on `--features experimental-grpc-retry` because the new tests reach for the trace-gated crate-root re-export `crate::OTEL_EXPORTER_OTLP_TRACES_INSECURE`. This is a pre-existing class of issue (other `OTEL_EXPORTER_OTLP_TRACES_*` consts and `SpanExporter`/`LogExporter` fail the same way on `main`). Happy to file a separate issue to track. 

To future-proof these new tests, you could either gate them on `feature = "trace"` or reference them via `crate::span::OTEL_EXPORTER_OTLP_TRACES_INSECURE`.

scottgerring (commented):
Hey @bryantbiggs thanks for this one and sorry for very slow turnaround - it'll help with OTLP stabilisation! 

Some comments inline but I reckon it's almost there 🚢

opentelemetry-otlp/src/exporter/tonic/mod.rs
bryantbiggs: Good call. Switched to checking for `://`, so any explicit scheme (including `unix://`) is left as-is now. I pulled the logic into a small `apply_insecure_scheme()` helper with plain unit tests — no env, runtime, or TLS feature needed — so the `unix://` case is guarded going forward.

You were also right that the existing test was excluded from CI; I addressed that in the thread below.

opentelemetry-otlp/CHANGELOG.md
bryantbiggs: Done — moved to `vNext`. I also tweaked the wording to say any explicit scheme (`http`/`https`/`unix`) is left untouched, to match the fix above.

opentelemetry-otlp/src/exporter/mod.rs
bryantbiggs: Done — used your "Whether to disable TLS…" phrasing here, and aligned the per-signal const docs (traces/metrics/logs) to match.

opentelemetry-otlp/src/logs.rs
bryantbiggs: Done — inverted here, and in the metrics/traces const docs too.

opentelemetry-otlp/src/exporter/tonic/mod.rs
bryantbiggs: All addressed:

- The tests are `#[tokio::test]` now. (The one that actually panics without a runtime is the `insecure_true` case, since it builds a channel — I made all the `build()`-calling tests async to be safe.)
- You were right these weren't running in CI: the no-TLS tests get compiled out under `--all-features`. I added a curated set of feature combos to `scripts/test.sh` (gRPC without TLS, http-proto, http-json) so they execute now, and gated the new tests on `trace` so they're safe under `cargo hack`.

I also took a swing at the broader "cargo hack for tests" gap in a separate commit. It came down to a few pre-existing things that only break when a feature is built in isolation — the `http-proto`-only `HttpBinary` tests, a couple of `build_with_default_transport` tests that need a runtime under a gRPC-only build, and a latent `default_protocol()` typo. I used curated real-world combos rather than raw `--each-feature` (which would force transport-without-signal builds to compile), and kept it as its own commit so it's easy to peel into a separate PR if you'd prefer.

scottgerring: I think this makes sense for now; fwiw I tried the `hack` route quickly locally too and it looked like a pain. I think this should also make us reflect a bit about our feature flag sprawl, but that's a story for another day!

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## vNext
 
+- Add support for INSECURE environment variables for gRPC (env-var-only, no builder method, per spec):
+  `OTEL_EXPORTER_OTLP_INSECURE` (generic), `OTEL_EXPORTER_OTLP_TRACES_INSECURE`,
+  `OTEL_EXPORTER_OTLP_METRICS_INSECURE`, `OTEL_EXPORTER_OTLP_LOGS_INSECURE`.
+  Per the spec, these only apply to gRPC connections. When an endpoint has no explicit scheme,
+  `INSECURE=true` uses `http://`, `INSECURE=false` (default) uses `https://` with auto-TLS.
+  **Note:** Schemeless endpoints (e.g., `collector.example.com:4317`) now default to `https://`
```
scottgerring: Sorry I just noticed this - I think we should call this out explicitly @bryantbiggs as it will trip some folks up. Once this is done we should merge it! 

```suggestion
  **Breaking:** Schemeless endpoints (e.g., `collector.example.com:4317`) now default to `https://`  
```

scottgerring (approved):
LGTM pending build completing!
@cijothomas it would be great to get this in in the quest for OTLP stabilisation

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -573,6 +580,26 @@ pub(crate) fn render_source_chain(err: &(dyn std::error::Error + 'static)) -> St
...
+/// [OTLP exporter spec]: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options
+fn apply_insecure_scheme(endpoint: String, insecure: bool) -> String {
+    if endpoint.contains("://") {
+        endpoint
+    } else if insecure {
+        format!("http://{endpoint}")
+    } else {
+        format!("https://{endpoint}")
+    }
+}
```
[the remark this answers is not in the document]
bryantbiggs: Fixed in 1aa6a66d. An endpoint is now treated as schemed only when a `://` precedes any path/query/fragment delimiter, so `collector:4317/path://foo` is correctly prefixed instead of passed through to fail URI parsing.

Implemented with `split_once("://")` rather than the offset comparison — same behavior, but it drops the `unwrap_or(len)` sentinel and reads as the intent directly:

```rust
let has_scheme = endpoint
    .split_once("://")
    .is_some_and(|(scheme, _)| !scheme.contains(['/', '?', '#']));
```

Added a unit test covering `collector:4317/path://foo` for both `insecure` values.

PR #3366, opened 2026-02-20
ci: split lint into parallel jobs and remove redundant checks

bryantbiggs:
## Summary

The CI `lint` job is the most frequent source of PR failures. After investigating ~20 recent failures, **every single one was a legitimate clippy error** (dead code, unused variables, etc.) caught by `cargo hack --each-feature` — not resource exhaustion or flakiness.

The root cause: PRs introduce code that compiles fine with `--all-features` but triggers warnings when individual features are tested in isolation (e.g., a function only used with `grpc-tonic` enabled but missing `#[cfg(feature = "grpc-tonic")]`). Contributors don't know to run `cargo hack` locally, so they only discover the issue after pushing.

This PR improves the situation by:

- **Splitting lint into two parallel CI jobs** for faster feedback:
  - `lint`: format, workspace clippy, and multi-feature combo checks (~1-2 min)
  - `lint-feature-matrix`: per-feature clippy via `cargo hack --each-feature` (~5-6 min)
  
  Previously all ~150 clippy invocations ran sequentially in a single job. Now the quick checks give immediate feedback while the thorough feature matrix runs in parallel.

- **Removing 6 redundant `cargo_feature` checks** from `lint.sh` that are already covered by `cargo hack --each-feature`:
  - `opentelemetry-otlp "default"`, `"http-proto"`, `"metrics"` (single-feature, already tested individually by cargo hack)
  - `opentelemetry-jaeger-propagator "default"` (same)
  - `opentelemetry-proto "default"`, `"full"` (same)

- **Updating CONTRIBUTING.md** to document the `cargo hack` lint command so contributors can catch per-feature issues locally before pushing.

- **Updating `scripts/precommit.sh`** to include the feature matrix check.

## Test plan

- [x] Verify `lint` job passes (format + workspace clippy + multi-feature combos)
- [x] Verify `lint-feature-matrix` job passes (cargo hack per-feature checks)
- [x] Confirm both jobs run in parallel, not sequentially
- [x] Confirm no regressions in lint coverage (all multi-feature combos still checked)

### DIFF
--- a/scripts/lint_feature_matrix.sh
+++ b/scripts/lint_feature_matrix.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -eu
+
+# Run clippy for each feature individually across all workspace crates.
+# This catches dead code, unused imports, and other warnings that only
+# surface when features are tested in isolation (e.g., a function gated
+# behind feature A but missing #[cfg(feature = "A")]).
+
+if rustup component add clippy && \
+  ((cargo --list | grep -q hack) || cargo install cargo-hack); then
+  cargo hack --each-feature --no-dev-deps clippy -- -Dwarnings
+fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,6 @@ jobs:
       with:
         toolchain: stable
         components: rustfmt, clippy
-    - uses: taiki-e/install-action@51cd0b8c0499559d9a4d75c0f5c67bec3a894ec8 # v2.75.28
-      with:
-          tool: cargo-hack
     - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
       with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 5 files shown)


scripts/lint_feature_matrix.sh
```diff
@@ -0,0 +1,13 @@
...
+set -eu
+
+# Run clippy for each feature individually across all workspace crates.
+# This catches dead code, unused imports, and other warnings that only
+# surface when features are tested in isolation (e.g., a function gated
+# behind feature A but missing #[cfg(feature = "A")]).
+
+if rustup component add clippy && \
+  ((cargo --list | grep -q hack) || cargo install cargo-hack); then
+  cargo hack --each-feature --no-dev-deps clippy -- -Dwarnings
```
scottgerring, line 12: I _think_ we need `--all-targets` ? 
```suggestion
  cargo hack --each-feature --no-dev-deps clippy --all-targets -- -Dwarnings
```

cargo-hack still checks each feature combination, but unlike cargo_feature calls it does not pass --all-targets, so clippy coverage for tests/examples/benches under those feature combinations is reduced.

Issue #2643, opened 2025-02-11
Stabilize support for async runtimes


cijothomas:
Parent issue to track:
1. Move the `runtime` and associated trait from behind experimental feature flag. This requires thoroughly reviewing the public API and trim any excess.
2. Provide implementation of thread-based using `runtime` trait. (And keep it default)
3. Provide implementation of runtime for tokio,async-std. (It exists today, but under experimental ff.)
4. Add tests for each implementation. (integration test too)
5. Provide ability to bring your-own-runtime (again, this is available but behind experimental ff)

lalitb:
Thanks for raising the issue. I had some work done for thread based runtime in #2390. Will take this forward.

martintmk:
The #2641 PR cleans up the the `Runtime` trait and it should help with the step 1.

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2641#discussion_r1970792780 Ping here. We might want to further trim it.. A trait with just one method is much easier to stabilize for 1.0. Need to explore and see if this is actually possible or not.

trait runtime
spawn_background_job(fn);

thread implementation:
Create dedicated thread

tokio implementation:
tokio:spawn_blocking


We also need to consider https://github.com/open-telemetry/opentelemetry-rust/issues/2697#issuecomment-2675933396 otherwise, internal-logs won't flow unless global_tracing_subscriber is enabled. @Sushisource's suggestion to accept a thread-start hook is reasonable, and can be achieved with the above model - user decides how the background job is done, and they can decide to setup thread-local tracing_subscriber or global or do nothing.

scottgerring:
@cijothomas @lalitb @bantonsson we've all talked about stabilising the runtime a bit, and as its now becoming apparent _this_ is going to the blocker for OTLP stabilisation, I thought i'd share some notes I cranked out on this a while ago. It's a bit overly verbose but I think should at least share as a reasonable summary of where we are.

The options I see to go forward are at the bottom; I think this and whatever else you folks see is where we should focus our efforts!

---

# `Runtime` and `RuntimeChannel` traits

This document explores improvements and stabilisation around the `Runtime` and `RuntimeChannel` traits. 

[...]

PR #3380, opened 2026-02-20
fix(sdk): resolve exporter deadlock on constrained tokio runtimes

### LINKED ISSUE #2802: OTLP MetricExporter deadlock issue

bryantbiggs:
## Summary

- Add `BlockingStrategy` that captures the tokio runtime handle at construction and enters the runtime context via `Handle::enter()` before calling `futures_executor::block_on()` on dedicated background threads
- Update `BatchSpanProcessor`, `BatchLogProcessor`, and `PeriodicReader` to use `BlockingStrategy` instead of bare `futures_executor::block_on()`
- Falls back to plain `futures_executor::block_on()` when no tokio runtime is available

## Problem

The default thread-based processors call `futures_executor::block_on(exporter.export(...))` on their dedicated worker threads. When the exporter uses tonic/gRPC, the export future depends on tokio tasks (e.g. tonic's `Buffer` worker spawned via `tokio::spawn`) that can only be polled by tokio worker threads. If all tokio worker threads are blocked — single-threaded runtime (`current_thread`), or `multi_thread` with 1 worker (common in 1-vCPU k8s pods) — this creates a circular wait: the worker thread waits for the export to complete, but the export can't complete because no tokio thread is available to poll the Buffer worker.

Reproduction and detailed analysis in #3356 (comment): https://github.com/open-telemetry/opentelemetry-rust/pull/3356#issuecomment-3935459655

Minimal repro gist: https://gist.github.com/bryantbiggs/62737e105525fe341090d0ad97de2178

| Scenario | `force_flush()` | `shutdown()` |
|----------|-----------------|--------------|
| `current_thread` | Hangs forever | Returns `Err(Timeout(5s))`, worker thread stays stuck |
| `multi_thread(1)` + `tokio::spawn` | Hangs forever (entire runtime freezes) | Same pattern |
| `multi_thread(default workers)` | Returns immediately | Returns immediately |

## Solution

`BlockingStrategy::new()` is called during processor construction (while inside the tokio runtime context). It calls `Handle::try_current()` to capture the runtime handle. On the dedicated background thread, `blocking_strategy.block_on(future)` enters the runtime context via `Handle::enter()` before calling `futures_executor::block_on()`. This makes tokio types (spawn, timers, IO) available without taking ownership of the reactor — IO continues to be driven by the runtime's own threads.

When no tokio runtime is available (e.g., non-tokio environments), it falls back to plain `futures_executor::block_on()` — preserving existing behavior.

## Scope

This is the scoped-down version of #3356, containing only the bug fix as suggested by @scottgerring in https://github.com/open-telemetry/opentelemetry-rust/pull/3356#issuecomment-3935741213. The experimental async runtime removal is intentionally excluded and can be discussed separately.

Fixes #2802

## Additional shutdown improvements

While responding to @thalesfragoso's review, the mock exporters were strengthened to also exercise the timer and IO drivers (not just `tokio::spawn`), and test coverage expanded to shutdown, `current_thread`, multi-worker, and empty-buffer scenarios. That work uncovered three latent bugs in the shutdown path. Each is addressed in a follow-on commit so the area isn't left in a known-poor state.

The first is that the metric shutdown chain ignored its timeout argument. `SdkMeterProvider::shutdown_with_timeout` and the downstream `Pipelines` / `Pipeline` / `PeriodicReader` chain dropped the caller's value and used a hardcoded 5s wait inside `PeriodicReaderInner::shutdown`, where the `// TODO: Make this timeout configurable.` comment lives. The follow-on commit threads the timeout through each layer.

The second is that `PeriodicReader` detached its worker thread. `PeriodicReader::new` was dropping the `JoinHandle` returned from `thread::Builder::spawn`. The follow-on commit stores the handle and joins it on successful shutdown, matching the pattern already used by `BatchSpanProcessor` and `BatchLogProcessor`. On timeout the handle is left in place because the worker is hung in `block_on(export_future)` and joining would deadlock the caller.

The third is that `BatchSpanProcessor` and `BatchLogProcessor` left a stale `JoinHandle` in `self.handle` on the timeout and disconnected branches of `shutdown_with_timeout`. The follow-on commit takes the handle out on the error arm so subsequent shutdown calls don't see stale state. The underlying worker-thread leak when hung in `block_on` is fundamental and would need cancellation primitives in the export path; a follow-up issue will track that.

## Test plan

- [x] `cargo check -p opentelemetry_sdk --all-features`
- [x] `cargo check -p opentelemetry_sdk --no-default-features`
- [x] `cargo clippy -p opentelemetry_sdk --all-features -- -Dwarnings`
- [x] `cargo test -p opentelemetry_sdk --all-features --lib` — 391 passed, 0 failed, 13 ignored (pre-existing)
- [x] `cargo check -p opentelemetry-otlp --all-features`

### DIFF
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -17,6 +17,7 @@
 
 use crate::error::{OTelSdkError, OTelSdkResult};
 use crate::logs::log_processor::LogProcessor;
+use crate::util::BlockingStrategy;
 use crate::{
     logs::{LogBatch, LogExporter, SdkLogRecord},
     Resource,
@@ -380,6 +381,7 @@ impl BatchLogProcessor {
         let max_export_batch_size = config.max_export_batch_size;
         let current_batch_size = Arc::new(AtomicUsize::new(0));
         let current_batch_size_for_thread = current_batch_size.clone();
+        let blocking_strategy = BlockingStrategy::new();
 
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Logs.BatchProcessor".to_string())
@@ -406,6 +408,7 @@ impl BatchLogProcessor {
                     last_export_time: &mut Instant,
                     current_batch_size: &AtomicUsize,
                     max_export_size: usize,
+                    blocking_strategy: &BlockingStrategy,
                 ) -> OTelSdkResult
                 where
                     E: LogExporter + Send + Sync + 'static,
@@ -431,13 +434,15 @@ impl BatchLogProcessor {
                         }
                         total_exported_logs += count_of_logs;
 
-                        result = export_batch_sync(exporter, logs, last_export_time); // This method clears the logs vec after exporting
+                        result =
+                            export_batch_sync(exporter, logs, last_export_time, blocking_strategy); // This method clears the logs vec after exporting
 
                         current_batch_size.fetch_sub(count_of_logs, Ordering::AcqRel);
                     }
                     result
                 }
 
+                let blocking_strategy = blocking_strategy;
                 loop {
                     let remaining_time = config
                         .scheduled_delay
@@ -460,6 +465,7 @@ impl BatchLogProcessor {
                                 &mut last_export_time,
                                 &current_batch_size,
                                 max_export_batch_size,
+                                &blocking_strategy,
                             );
                         }
                         Ok(BatchMessage::ForceFlush(sender)) => {
@@ -471,6 +477,7 @@ impl BatchLogProcessor {
                                 &mut last_export_time,
                                 &current_batch_size,
                                 max_export_batch_size,
+                                &blocking_strategy,
                             );
                             let _ = sender.send(result);
                         }
@@ -483,6 +490,7 @@ impl BatchLogProcessor {
                                 &mut last_export_time,
                                 &current_batch_size,
                                 max_export_batch_size,
+                                &blocking_strategy,
                             );
                             let _ = exporter.shutdown();
                             let _ = sender.send(result);
@@ -511,6 +519,7 @@ impl BatchLogProcessor {
                                 &mut last_export_time,
                                 &current_batch_size,
                                 max_export_batch_size,
+                                &blocking_strategy,
                             );
                         }
                         Err(RecvTimeoutError::Disconnected) => {
@@ -609,6 +618,7 @@ fn export_batch_sync<E>(
     exporter: &E,
     batch: &mut Vec<Box<(SdkLogRecord, InstrumentationScope)>>,
     last_export_time: &mut Instant,
+    blocking_strategy: &BlockingStrategy,
 ) -> OTelSdkResult
 where
     E: LogExporter + ?Sized,
@@ -620,7 +630,7 @@ where
     }
 
     let export = exporter.export(LogBatch::new_with_owned_data(batch.as_slice()));
-    let export_result = futures_executor::block_on(export);
+    let export_result = blocking_strategy.block_on(export);
 
     // Clear the batch vec after exporting
     batch.clear();
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -1122,6 +1122,66 @@ mod tests {
         processor.shutdown().unwrap();
     }
 
+    // Mock exporter that uses tokio::spawn internally, simulating tonic/gRPC
+    // exporters where the export future depends on tokio tasks. Without
+    // BlockingStrategy, this deadlocks on constrained runtimes because
+    // futures_executor::block_on() cannot drive tokio::spawn-ed tasks.
+    #[derive(Debug, Clone)]
+    struct TokioSpawnLogExporter {
+        exported_count: Arc<AtomicUsize>,
+    }
+
+    impl TokioSpawnLogExporter {
+        fn new() -> Self {
+            Self {
+                exported_count: Arc::new(AtomicUsize::new(0)),
+            }
+        }
+    }
+
+    impl LogExporter for TokioSpawnLogExporter {
+        async fn export(&self, batch: LogBatch<'_>) -> OTelSdkResult {
+            let count = batch.len();
+            // Simulate tonic/gRPC: the export future depends on a tokio::spawn-ed task.
+            let result = tokio::spawn(async move { count }).await.unwrap();
+            assert_eq!(result, count);
+            self.exported_count.fetch_add(count, Ordering::Relaxed);
+            Ok(())
+        }
+
+        fn shutdown(&self) -> OTelSdkResult {
+            Ok(())
+        }
+    }
+
+    // Regression test for deadlock on constrained tokio runtimes (#2802, #3356).
+    // Uses TokioSpawnLogExporter which internally calls tokio::spawn(),
+    // simulating tonic/gRPC exporters where the export future depends on
+    // tokio-spawned tasks. Without BlockingStrategy, this deadlocks because
+    // futures_executor::block_on() cannot drive tokio::spawn-ed tasks
+    // without the runtime context.
+    //
+    // Note: current_thread runtime is not tested here because it has a
+    // fundamental limitation — the single runtime thread is blocked by
+    // force_flush()'s recv(), so no thread is available to drive spawned
+    // tasks. The multi_thread(1) scenario (1-vCPU k8s pods) is the primary
+    // target of this fix.
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_batch_log_processor_multi_thread_1_worker_with_tokio_spawn_exporter() {
+        let exporter = TokioSpawnLogExporter::new();
+        let exported_count = exporter.exported_count.clone();
+        let processor = BatchLogProcessor::new(exporter, BatchConfig::default());
+
+        let mut record = SdkLogRecord::new();
+        let instrumentation = InstrumentationScope::default();
+        processor.emit(&mut record, &instrumentation);
+
+        processor.force_flush().unwrap();
+
+        assert_eq!(exported_count.load(Ordering::Relaxed), 1);
+    }
+
     #[tokio::test(flavor = "multi_thread")]
     async fn test_batch_log_processor_shutdown_with_async_runtime_multi_flavor_current_thread() {
         let exporter = InMemoryLogExporterBuilder::default().build();
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -317,27 +317,28 @@ impl LogProcessor for BatchLogProcessor {
                         }
                         OTelSdkResult::Ok(())
                     })
-                    .map_err(|err| match err {
-                        RecvTimeoutError::Timeout => {
-                            // TODO: When shutdown times out, log records still
-                            // in the queue or mid-export are silently lost. The
-                            // background thread is not joined and may continue
-                            // running. Consider: (1) recording the lost count
-                            // in the self-diagnostics counter, (2) joining the
-                            // thread with a best-effort wait, or (3) signalling
-                            // the thread to abort the current export.
-                            otel_error!(
-                                name: "BatchLogProcessor.Shutdown.Timeout",
-                                message = "BatchLogProcessor shutdown timing out."
-                            );
-                            OTelSdkError::Timeout(timeout)
-                        }
-                        _ => {
-                            otel_error!(
-                                name: "BatchLogProcessor.Shutdown.Error",
-                                error = format!("{}", err)
-                            );
-                            OTelSdkError::InternalFailure(format!("{err}"))
+                    .map_err(|err| {
+                        // The worker thread is unreachable from here: Timeout
+                        // means it is hung in `block_on(export_future)` and
+                        // joining would deadlock; Disconnected means it has
+                        // already exited. Take the handle out so subsequent
+                        // shutdown attempts do not see a stale `Some(_)`.
+                        drop(self.handle.lock().unwrap().take());
+                        match err {
+                            RecvTimeoutError::Timeout => {
+                                otel_error!(
+                                    name: "BatchLogProcessor.Shutdown.Timeout",
+                                    message = "BatchLogProcessor shutdown timing out."
+                                );
+                                OTelSdkError::Timeout(timeout)
+                            }
+                            _ => {
+                                otel_error!(
+                                    name: "BatchLogProcessor.Shutdown.Error",
+                                    error = format!("{}", err)
+                                );
+                                OTelSdkError::InternalFailure(format!("{err}"))
+                            }
                         }
                     })?
             }
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -846,7 +846,7 @@ mod tests {
     };
     #[cfg(feature = "experimental_logs_batch_log_processor_with_async_runtime")]
     use super::{OTEL_BLRP_EXPORT_TIMEOUT, OTEL_BLRP_EXPORT_TIMEOUT_DEFAULT};
-    use crate::error::OTelSdkResult;
+    use crate::error::{OTelSdkError, OTelSdkResult};
     use crate::logs::log_processor::tests::MockLogExporter;
     use crate::logs::SdkLogRecord;
     use crate::logs::{LogBatch, LogExporter};
@@ -1142,8 +1142,14 @@ mod tests {
     impl LogExporter for TokioSpawnLogExporter {
         async fn export(&self, batch: LogBatch<'_>) -> OTelSdkResult {
             let count = batch.len();
-            // Simulate tonic/gRPC: the export future depends on a tokio::spawn-ed task.
-            let result = tokio::spawn(async move { count }).await.unwrap();
+            // Simulate tonic/gRPC: the export future needs tokio::spawn plus
+            // live timer and IO drivers on the worker thread.
+            let result = tokio::spawn(async move {
+                crate::util::exercise_tokio_drivers().await;
+                count
+            })
+            .await
+            .unwrap();
             assert_eq!(result, count);
             self.exported_count.fetch_add(count, Ordering::Relaxed);
             Ok(())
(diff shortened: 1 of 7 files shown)


opentelemetry-sdk/src/util.rs
```diff
@@ -7,3 +7,44 @@ pub fn tokio_interval_stream(
...
+        }
+        Self::FuturesExecutor
+    }
+
+    pub(crate) fn block_on<F: std::future::Future>(&self, future: F) -> F::Output {
+        match self {
+            #[cfg(feature = "rt-tokio")]
+            Self::TokioHandle(handle) => {
+                let _guard = handle.enter();
+                futures_executor::block_on(future)
```
thalesfragoso, line 45: It makes more sense to use `Hanlder::block_on` instead.

However, this still will cause problems with single-threaded tokio runtimes if `block_on` is called on the tokio thread:


> When this is used on a current_thread runtime, only the Runtime::block_on method can drive the IO and timer drivers, but the Handle::block_on method cannot drive them. This means that, when using this method on a current_thread runtime, anything that relies on IO or timers will not work unless there is another thread currently calling Runtime::block_on on the same runtime.

This also applies to `futures_executor::block_on`. However, it seems that this is only used in the batch processing and that they always seem to spawn their own threads, is that correct?

I would change your tests to include `tokio::time::sleep` in the mock exporters, to be sure the timer and IO drivers can really run.

bryantbiggs: Thanks for catching this @thalesfragoso. The mocks now exercise both the timer and IO drivers via a small `exercise_tokio_drivers` helper in `util.rs`.

That expansion uncovered three latent bugs in the shutdown path which are addressed in the follow-on commits, with a summary in the new "Additional shutdown improvements" section of the PR body. Appreciate the careful review.

opentelemetry-sdk/src/util.rs
```diff
@@ -7,3 +7,56 @@ pub fn tokio_interval_stream(
...
+/// types (spawn, timers, IO resources) available on dedicated background
+/// threads without taking ownership of the reactor — IO continues to be
+/// driven by the runtime's own threads.
+///
+/// Falls back to plain [`futures_executor::block_on()`] when no tokio runtime
+/// is available (e.g., non-tokio environments).
+#[cfg(any(feature = "trace", feature = "logs", feature = "metrics"))]
+#[derive(Clone, Debug)]
+pub(crate) enum BlockingStrategy {
+    #[cfg(feature = "rt-tokio")]
```
scottgerring, line 25: This gates behind `rt-tokio`, but that feature isn't enabled for `grpc-tonic`:

https://github.com/open-telemetry/opentelemetry-rust/blob/25717768b2b94822b5e54b7950555bbbd632cbf2/opentelemetry-otlp/Cargo.toml#L73

If a user uses OTLP gRPC exporters plus the default SDK batch processors they will still get `FuturesExecutor`.

opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -1112,6 +1123,173 @@ mod tests {
...
+    async fn test_batch_log_processor_multi_thread_1_worker_with_tokio_spawn_exporter() {
+        let exporter = TokioSpawnLogExporter::new();
+        let exported_count = exporter.exported_count.clone();
+        let processor = BatchLogProcessor::new(exporter, BatchConfig::default());
+
+        let mut record = SdkLogRecord::new();
+        let instrumentation = InstrumentationScope::default();
+        processor.emit(&mut record, &instrumentation);
+
+        processor.force_flush().unwrap();
```
scottgerring, line 1187: I _believe_ in this test `force_flush()` is being called from the test/runtime `block_on` thread, not from the single Tokio worker thread. That leaves the worker available to run the task spawned inside the mock exporter, so this may not demonstrate the constrained-runtime deadlock case completely.

 What happens if we call `force_flush()` from a Tokio task instead?

   ```rust
   tokio::spawn(async move {
       processor.force_flush().unwrap();
   })
   .await
   .unwrap();
 ```

 With worker_threads = 1, that spawned task may occupy the only Tokio worker while waiting for the processor response; then the exporter’s own tokio::spawn(...).await has no
 worker left to make progress.

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -711,20 +720,28 @@ impl SpanProcessor for BatchSpanProcessor {
...
-                                error = format!("{}", err)
-                            );
-                            OTelSdkError::InternalFailure(format!("{err}"))
+                    .map_err(|err| {
+                        // The worker thread is unreachable from here: Timeout
+                        // means it is hung in `block_on(export_future)` and
+                        // joining would deadlock; Disconnected means it has
+                        // already exited. Take the handle out so subsequent
+                        // shutdown attempts do not see a stale `Some(_)`.
+                        drop(self.handle.lock().unwrap().take());
```
scottgerring, line 729: Dropping here makes it impossible to surface the worker panic message or join error. in #3538 you're trying to improve this and it would be a pity to regress it here.

For the `Timeout` branch, dropping the handle seems to make sense as a best-effort detach because joining could block forever. But for  Disconnected`, the worker has already dropped the response sender, so it has likely exited or panicked. In that branch, perhaps we join the handle instead of dropping it, so we don't mask the original worker error?

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -1599,6 +1657,160 @@ mod tests {
...
+        }
+
+        processor.force_flush().unwrap();
+
+        let exported_spans = exporter_shared.lock().unwrap();
+        assert_eq!(exported_spans.len(), 4);
+    }
+
+    // Exercises BatchMessage::Shutdown's drain path, which has a 5s timeout
+    // wrapper instead of force_flush's infinite recv().
```
scottgerring, line 1697: I don't think we should normalize this as "force_flush's infinite recv" - in `BatchLogProcessor` force flush uses `recv_timeout(self.forceflush_timeout)`, though that timeout is hardcoded today. Maybe just:

   ```suggestion
       // Exercises BatchMessage::Shutdown's drain path, which has a timeout around
       // the response wait.
 ```

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -110,12 +110,12 @@ impl SdkMeterProvider {
     ///
     /// There is no guaranteed that all telemetry be flushed or all resources have
```
scottgerring, line 111:
```suggestion
    /// There is no guarantee that all telemetry be flushed or all resources have
```

scottgerring: Not from your change but it jumped out!

opentelemetry-sdk/src/logs/batch_log_processor.rs
scottgerring, line 840: The new exporter tests below are compiled under `logs,testing`, but the fix only works when `opentelemetry-sdk/rt-tokio` is enabled. Locally, `cargo test -p opentelemetry_sdk --no-default-features --features 'logs,testing' --lib test_batch_log_processor_multi_thread_1_worker_with_tokio_spawn_exporter` fails with `there is no
 reactor running`.

Similar to my other comment about feature gating

opentelemetry-sdk/src/metrics/periodic_reader.rs
scottgerring, line 449: This still waits with `response_rx.recv()` and can block forever if the worker is stuck in export. That seems relevant here because the PR is about constrained-runtime deadlocks, and the PR body calls out `force_flush()` hanging forever. Can we fix this by adding a bounded wait in this PR, or do we just want to document it?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -611,6 +623,214 @@ mod tests {
...
+        let reader = PeriodicReader::builder(exporter)
+            .with_interval(LONG_INTERVAL)
+            .build();
+
+        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
+        let meter = meter_provider.meter("test");
+        let counter = meter.u64_counter("test.counter").build();
+        counter.add(1, &[]);
+
+        meter_provider.force_flush().unwrap();
```
scottgerring, line 700: This is another one where we might consider moving the force_flush into the runtime with a spawn

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -450,29 +458,33 @@ impl<E: PushMetricExporter> PeriodicReaderInner<E> {
...
+        };
+
+        // On success the worker has acknowledged shutdown and exited its loop,
+        // so joining is near-instant. On timeout the worker is hung in
+        // `block_on` of an export future; we leave the handle in place so a
+        // subsequent shutdown attempt or `Drop` can retry rather than risk
+        // deadlocking the caller here.
+        if result.is_ok() {
+            if let Some(handle) = self.handle.lock().unwrap().take() {
+                let _ = handle.join();
```
scottgerring, line 484: Can we avoid masking join errors here?

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -450,29 +458,33 @@ impl<E: PushMetricExporter> PeriodicReaderInner<E> {
...
-            }
-            Err(mpsc::RecvTimeoutError::Timeout) => {
-                Err(OTelSdkError::Timeout(Duration::from_secs(5)))
-            }
+        let result = match response_rx.recv_timeout(timeout) {
+            Ok(true) => Ok(()),
+            Ok(false) => Err(OTelSdkError::InternalFailure("Failed to shutdown".into())),
+            Err(mpsc::RecvTimeoutError::Timeout) => Err(OTelSdkError::Timeout(timeout)),
             Err(mpsc::RecvTimeoutError::Disconnected) => {
                 Err(OTelSdkError::InternalFailure("Failed to shutdown".into()))
```
scottgerring, line 473: Should/could we join here? If it already exited or panicked, joining should be safe and would let us surface the panic.

opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -611,6 +623,214 @@ mod tests {
...
+            .build();
+
+        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
+        let meter = meter_provider.meter("test");
+        let counter = meter.u64_counter("test.counter").build();
+        counter.add(1, &[]);
+
+        meter_provider.force_flush().unwrap();
+
+        assert!(exported_count.load(Ordering::Relaxed) >= 1);
```
scottgerring, line 702: If this should just be one, can we make it just assert ==1 like the other tests, and if it is sometimes more, can we add a comment explaining why?

opentelemetry-sdk/src/metrics/pipeline.rs
scottgerring, line 720: It feels kinda unfortunate that if we have a single error and its a Timeout we're losing typing here; given the overall theme of the PR perhaps we should preserve it in the simple case?

opentelemetry-sdk/src/util.rs
```diff
@@ -7,3 +7,56 @@ pub fn tokio_interval_stream(
 ) -> tokio_stream::wrappers::IntervalStream {
     tokio_stream::wrappers::IntervalStream::new(tokio::time::interval(period))
 }
+
+/// Strategy for blocking on async futures from synchronous contexts.
+///
+/// When constructed within a tokio runtime, captures the runtime handle
```
scottgerring, line 13: I think we should be clearer about this - this only happens if you've opted into `rt-tokio`, which matters because `opentelemetry-otlp/grpc-tonic` pulls in tokio but doesn't toggle on this feature

opentelemetry-sdk/src/util.rs
scottgerring, line 4: This is compiled for _either_ feature, but `BlockingStrategy::TokioHandle` is only compiled for `rt-tokio`. This might be intentional but then we should probably document it.

opentelemetry-sdk/src/util.rs
```diff
@@ -7,3 +7,56 @@ pub fn tokio_interval_stream(
...
+            Self::FuturesExecutor => futures_executor::block_on(future),
+        }
+    }
+}
+
+/// Exercise the tokio timer and IO drivers from a spawned task. Mock
+/// exporters call this so a processor that stops keeping the runtime
+/// drivers alive on its worker thread shows up as a test failure rather
+/// than silently regressing the deadlock fix.
+#[cfg(test)]
```
scottgerring, line 56: Should this helper be cfg-gated the same way as the tests that need it? It uses `tokio::time` and `tokio::net`, but is currently available under plain `#[cfg(test)]`. Given the `TokioHandle` path is feature-gated, the helper/tests should probably be gated the same.

opentelemetry-sdk/src/util.rs
scottgerring: I don't know if `rt-tokio` is the right gate to use here. It doesn't seem like something users should have to enable to get correct behavior from the standard sync processors, especially when the common `opentelemetry-otlp/grpc-tonic` path already brings in Tokio but does not enable `opentelemetry-sdk/rt-tokio`.

scottgerring (changes requested):
Some comments inline! I think the big thing is the choice of feature gate

bryantbiggs:
@scottgerring you're right that `rt-tokio` is the wrong gate — and digging into *why* it has to be gated convinced me this is better fixed one layer down, in the exporter. Sharing where I landed, plus a prototype, since I think it changes the approach for this PR.

## Why the gate isn't really avoidable in the SDK

`opentelemetry-sdk` only links `tokio` when `rt-tokio` (or `rt-tokio-current-thread`/`testing`) is enabled — it's an optional dependency. So `BlockingStrategy::TokioHandle(tokio::runtime::Handle)` *has* to be `#[cfg(feature = "rt-tokio")]`; the type doesn't exist in the crate otherwise. That's also why `opentelemetry-otlp/grpc-tonic` doesn't help — it brings tokio into `opentelemetry-otlp`, not into the SDK. Any SDK-side `Handle::enter()` fix is therefore structurally stuck behind a feature the default gRPC user never sets. The gate isn't a wrong choice we can swap — it's forced by the SDK's (deliberate) runtime-agnostic design.

## Root cause: tonic borrows the caller's runtime; other OTel SDKs own their transport threads

The deadlock isn't really SDK-thread behavior — it's that **tonic drives its transport (the tower `Buffer` worker, the hyper connection task) on the *ambient* tokio runtime**. On a `current_thread` runtime, or `multi_thread` with one worker (1-vCPU pods), once the processor's worker thread blocks in `block_on(export)` there's no free tokio thread left to poll tonic's tasks → circular wait.

The other SDKs don't hit this because their gRPC transport owns its own I/O threads, independent of the application:

| SDK | Worker | Transport I/O driven by |
|---|---|---|
| Go | goroutine blocks on the RPC | runtime netpoller |
| Java | worker thread blocks on `result.join(timeout)` | OkHttp/Netty dispatcher |
| .NET | dedicated thread, sync export | client I/O threads |
| Python | daemon thread, blocking gRPC stub | gRPC C-core completion-queue threads |

There's no runtime-free path for tonic specifically (its "blocking" mode is just *owning a runtime + `block_on`*), and the in-progress official `grpc` crate is also tokio/hyper-based, so it shares the same property.

## Proposal: give the OTLP/tonic exporter its own transport runtime

Have the tonic exporter own a small dedicated multi-thread tokio runtime and run the export on it; the SDK processor thread just blocks waiting for the result over a channel. The export becomes "spawn the RPC onto the owned runtime, hand the result back" instead of "`block_on` the RPC on the caller's thread." This is the pure-Rust equivalent of what C-core gives Python/Go — the transport owns its threads.

Consequences:

- **No SDK feature gate.** The fix lives in `opentelemetry-otlp`, where tokio is already a dependency under `grpc-tonic`. `opentelemetry-sdk` stays runtime-agnostic — `BlockingStrategy`/`rt-tokio` aren't needed for this.
- **Correct on every runtime flavor**, including `current_thread`, which `Handle::enter()` fundamentally can't fix (on `current_thread`, `Handle::enter` + `block_on` still leaves no thread to drive the spawned tonic tasks).
- **Self-contained** to the transport layer — no exporter-trait or processor-architecture change.

## Prototype / evidence

I prototyped this against a **real tonic channel** (a genuine `LogsServiceClient` over `connect_lazy()` to a real `LogsServiceServer`, so the full Buffer-worker + hyper lazy-connect + HTTP/2 path runs — not a mock). Spawning the export onto an exporter-owned multi-thread runtime:

| Caller's runtime | `block_on(export)` on caller's thread (today) | export on exporter-owned runtime |
|---|---|---|
| `current_thread`, its only thread blocked | hangs | ✅ completes |
| no tokio runtime (plain thread) | panics on lazy-connect | ✅ completes |
| `multi_thread(1)` | works only if a worker is free | ✅ completes unconditionally |

The hyper lazy-connection concern (driver spawns on first request) didn't materialize as long as the *whole* export call is spawned onto the owned runtime rather than driven via `futures_executor` directly.

## Also resolves #3432

#3432 (`BatchSpanProcessor` panics with `experimental-grpc-retry`) is the same root cause: the retry backoff's `tokio::time::sleep` (via `Runtime::delay` on `Tokio`) is polled on the sync processor's reactor-less worker thread. I verified it — reproducing with the real `retry_with_backoff` + `Tokio` driven by `futures_executor::block_on` (the sync path) panics with exactly the message #3432 reports:

```
```

Running the same retry on the exporter-owned runtime (timers enabled) completes normally. So once the export — backoff included — runs on the owned runtime, the panic is gone and retry works with the default sync processors. This fixes the *panic*; the broader retry-flag stabilization (#3526 / #2643) is a separate concern this doesn't pre-empt.

## Trade-offs / open questions

- **Resource cost:** a dedicated runtime means extra threads. Probably want a single shared transport runtime across the trace/log/metric exporters with a configurable (small default) worker count, rather than one per exporter.
- **Lifecycle:** the owned runtime needs clean shutdown on exporter shutdown.
- **Non-tokio / no-runtime callers:** work (the owned runtime is self-sufficient) — strictly better than today.

## Relationship to the Runtime stabilization work (#2643 / #3412 / #3413)

I don't see this as competing with the `Runtime`-trait direction — those restructure SDK processor threading and shutdown lifecycle, which is the right long-term home for a unified model, and I want that to land. This is narrower: the transport-layer fix for the constrained-runtime deadlock specifically, which on #3542 you noted can be pulled as its own parallel workstream. I'm also conscious you paused a similar exporter-local change in #3432 pending Runtime stabilization — the difference I'd argue here is that owning the transport runtime isn't a workaround "around the edges," it's the same model every other SDK uses, and it makes export correct independent of any SDK feature or runtime flavor.

## Suggested path for this PR

If this direction makes sense, I'll rework the deadlock fix into `opentelemetry-otlp` along these lines and drop `BlockingStrategy` / the `rt-tokio` gating from the SDK. The shutdown-path fixes currently bundled here are independent and I'll move them to #3538 where they belong. Keen on your read before I rebuild.

cc @cijothomas / @lalitb

scottgerring:
Hey @bryantbiggs thanks for thinking hard about this!
I'm travelling currently and need a bit of focus to re-grok this but I will get to it in the next couple days and write back. Cleaning up more adjacent issues too would certainly be nice 💪

scottgerring:
Still thinking through this but wanted to get some thoughts down so we can iterate. 

**Hot Take**
To re-baseline: the root cause of all this misery is the mixing of threads (ambient tokio runtime supplied by the user, or not) and vanilla threads owned by some of the processors - combined with blocking shutdown operations. We're also half-pregnant with the `Runtime` trait and its async-centric ways which after the other big PR will solve this thread mixing ... for the async processors, but not these ones. Long term I think we should commit to the async stuff and the `Runtime` trait, but maybe @cijothomas has other opinions here 🤷 

**Threads**
We're going to end up with something like:

* **Batch[Span|Log]Processor** / **PeriodicReader** - have their own vanilla (i.e. not-tokio) threads to drive
* **Each tonic exporter** (or, maybe all shared) - self-managed tokio threads

I think as a user i'd expect 1/ more frugality with threading and 2/ coherence. 

To 1/,  I think we should avoid one runtime per tonic exporter if possible. A shared OTLP tonic transport runtime seems more reasonable. But then we need to be careful about lifecycle: trace/log/metric exporters may share it, so one exporter shutdown must not tear it down while others still need it. We also need to avoid reintroducing deadlocks by blocking from a runtime worker while waiting for work scheduled onto the same runtime.

By 2/ - coherence - I mean - I wouldn't expect disparate parts of the otel stack to launch disparate runtimes. Concretely if we do go down this path, I think when folks use the `Runtime` based processors the exporters should share the same runtime / worker pool. For the sync processors we may need an otel-owned transport fallback. 

**Other Languages**
I had a bit of a poke around here, and I think this is more nuanced than the the gRPC transport _owns_ its thread (this is also to some extent a function of the unique-ness of tokio's position in the rust ecosystem, I think):

* **go** - exporters just running on the global go runtime; I don't think there's really a useful comparison here because go's scheduler makes the shape of the problem very different
* **java** - `OkHttp` dispatcher fires up its own `ThreadPoolExecutor` unless the user provides an `ExecutorService`;  this matches the pattern you're proposing here 
*  **.NET**: not really exporter-owned transport threads; it blocks through `HttpClient`, with progress handled by .NET/runtime/OS I/O machinery.
* **Python**: not really SDK-managed transport threads; blocking grpcio stubs rely on grpcio/C-core internals.

   So Java is the closest precedent for an exporter-owned executor. The others mostly support the broader point that transport progress is independent of the currently-blocked application thread, but they don’t map as
 directly to “exporter owns a runtime”.

## TL;DR 

I think this could make sense as a tactical fix assuming we share a tokio runtime between gRPC exporters carefully and make sure we don't introduce any new and exciting races, but I defer to @cijothomas on whether or not we are happy to take the default extra tokio pool here. Based on the Java impl, I think we should - if he is happy too I think we move forward with this. 

Long term I would like to see us consolidate around the `Runtime` based trait and shave off a bunch of the complexity that we've picked up along the way, but in the meantime, we should absolutely do something.

@bryantbiggs if you'd like feel free to start a thread or DM with @cijothomas and I in slack to discuss so we can move forward on this!

PR #3383, opened 2026-02-23
chore: Added devcontainer for local development

vqrf:
## Changes

Added `devcontainer` for local development.

### Notes

Devcontainer capabilities:
- `docker in docker`, with `moby` disabled (not present in `debian 13`)
- `Rust` `1.93` with `rust-analyzer`, `rust-src`, `rustfmt`, `clippy`, `rust-docs`
- protobuf installed

#### Justification

Use new popular way to develop code with easy and reproducible environment for everyone.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,23 @@
+{
+    "name": "Rust Dev Container",
+    "build": {
+        "dockerfile": "containerfile"
+    },
+    "features": {
+        "ghcr.io/devcontainers/features/rust:1": {
+            "version": "1.93",
+            "components": "rust-analyzer,rust-src,rustfmt,clippy,rust-docs"
+        },
+        "ghcr.io/devcontainers/features/docker-in-docker:2": {
+            "moby": false
+        }
+    },
+    "remoteUser": "vscode",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "rust-lang.rust-analyzer"
+            ]
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "Rust Dev Container",
+    "build": {
+        "dockerfile": "containerfile"
+    },
+    "features": {
+        "ghcr.io/devcontainers/features/rust:1": {
+            "version": "1.93",
```
scottgerring, line 8: I think `latest` would probably be better? Otherwise we will be chasing our tails here

scottgerring:
We should reference this in CONTRIBUTING.md too, somewhere around here - if you prefer, you can get started quickly with a complete devcontainers setup!

https://github.com/open-telemetry/opentelemetry-rust/blob/146376fde8463a40b807aa68c3d54eab373b3730/CONTRIBUTING.md?plain=1#L56-L59

scottgerring:
I tried this out! It seems like we've got all the deps we need, but the "open in devcontainer" experience on my desktop (colima & docker, Mac) OOMs on build and test, in particular running the linker:

```
[ 1394.396063] Out of memory: Killed process 44829 (ld) total-vm:734740kB, anon-rss:685776kB, file-rss:104kB, shmem-rss:0kB, UID:1000 pgtables:1472kB oom_score_adj:0
```

The VM itself is setup with 2 GiB of RAM which I think is fairly typical. I have tried using `CARGO_BUILD_JOBS=1` to constrain the RAM impact but it still OOMs, which isn't surprising as it is the linker step that is failing.

I think if we use devcontainers they should optimally just work as that's a big part of the sell, but i'm not sure there is a way to advertise to the OS/container engine that you _need_ X GiB of RAM.  Any ideas @vqrf ?

PR #3384, opened 2026-02-24
feat: Cleanup OTLP Exporter public API

cijothomas:
ExportConfig, HasExportConfig, and the with_export_config() method were publicly exported but served no real purpose for external users. No examples, external crates, or documentation ever constructed ExportConfig directly — all user-facing configuration goes through the WithExportConfig trait's builder methods (.with_endpoint(), .with_protocol(), .with_timeout()), which remain public and unchanged.

Trying to trim off unnecessary public API.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## vNext
 
 - **Breaking** Removed `ExportConfig`, `HasExportConfig`, and `with_export_config()` from public API.
-  These were internal plumbing not intended for external use. Use the `WithExportConfig` trait
+Use the `WithExportConfig` trait
   methods (`.with_endpoint()`, `.with_protocol()`, `.with_timeout()`) instead, which remain public.
 - Add support for `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
   to configure metrics temporality. Accepted values: `cumulative` (default), `delta`,


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## vNext
 
+- **Breaking** Removed `ExportConfig`, `HasExportConfig`, and `with_export_config()` from public API.
+  These were internal plumbing not intended for external use. Use the `WithExportConfig` trait
```
scottgerring: Nit: if I was using something marked `pub`, upgraded, and was told it was never really intended to be used, i'd be a bit miffed. Let's just provide guidance:

```suggestion
Use the `WithExportConfig` trait
```

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -228,7 +228,7 @@ fn default_headers() -> std::collections::HashMap<String, String> {
 }
 
 /// Provide access to the [ExportConfig] field within the exporter builders.
-pub trait HasExportConfig {
+pub(crate) trait HasExportConfig {
```
scottgerring, line 231: I think we could make these `pub(crate)` too to slim down the API surface: 

https://github.com/open-telemetry/opentelemetry-rust/blob/09b85b5572d774eb820668416d1e3c6802aa8b14/opentelemetry-otlp/src/exporter/tonic/mod.rs#L438

https://github.com/open-telemetry/opentelemetry-rust/blob/09b85b5572d774eb820668416d1e3c6802aa8b14/opentelemetry-otlp/src/exporter/http/mod.rs#L730

cijothomas: Will follow up

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/pull/3386 here comes the PR!

PR #3385, opened 2026-02-24
docs: elaborate on contributing guide

scottgerring:
Our contributing guide is rather open ended and leaves a lot of room for interpretation on how best to interact with the project. This can lead to a frustrating experience both for contributors and maintainers. This PR takes some of the experience from the opentelemetry-collector and [their contributing guide]( https://github.com/open-telemetry/opentelemetry-collector/blob/main/CONTRIBUTING.md) and adds it here. 

## Changes

- Added "Target Audiences" section - defines priority order (application developers > library authors > SDK developers) - as an axiom to drive both 1/ what we should work on 2/ and to help make design tradeoffs
- Added "Before You Start" section - suggesting contributions should follow from accepted issues. This will help us align effort around what is important for the project, and save time on contribution side by helping to avoid effort going into PRs that are unlikely for whatever reason to be accepted
 - Added "Pull Request Size and Scope" section -  recommends PRs stay under 500 lines, gives a pattern for splitting large
  features into incremental PRs, and requires refactoring to be separated from behaviour changes
  - Added "Issue Lifecycle and Labels" section — documents the full label taxonomy (triage, area, module, priority,
  versioning, platform), explains the triage workflow (triage:todo → triage:accepted), and makes clear that contributors
  should not start work on untriaged issues. Directs users to file issues via templates so labels are automatically applied.

And a nit:

  - Fixed meeting schedule to reflect new cadence

## Follow-Up 

* I think we can do better at triage; we have a lot of stuff that is poorly labelled or that hasn't been triaged at all. I've started adding `triage:todo` to issues that need attention so we can be more mechanical here
* On the PR side, we have a bunch of PRs that are notionally approved but unmerged. It would be great to clear this backlog
* We should consider using the label bot work that the collector is using too, to reduce toil and bring our focus as maintainers and reviewers to where it should be

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## vNext
 
+- **Breaking** Removed `ExportConfig`, `HasExportConfig`, `with_export_config()`, `HasTonicConfig`, `HasHttpConfig`, `TonicConfig`, and `HttpConfig` from public API.
+  Use the public `WithExportConfig`, `WithTonicConfig`, and `WithHttpConfig` trait methods instead, which remain unchanged.
+- The gRPC/tonic OTLP exporter's build method now returns an error for all signals (traces, metrics, logs) when
+  an `https://` endpoint is configured but no TLS feature (`tls-ring` or `tls-aws-lc`) is enabled, instead of
+  silently sending unencrypted traffic. When a TLS feature is enabled and an `https://` endpoint is used without
+  an explicit `.with_tls_config()`, a default `ClientTlsConfig` is automatically applied.
+  [#3182](https://github.com/open-telemetry/opentelemetry-rust/issues/3182)
 - Add support for `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable
   to configure metrics temporality. Accepted values: `cumulative` (default), `delta`,
   `lowmemory` (case-insensitive). Programmatic `.with_temporality()` overrides the env var.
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -115,6 +171,93 @@ maintainers may decide to wait for more than one approval for certain PRs,
...
+  before it can be acted on.
+
+**Don't start work on an issue that hasn't been triaged.** An issue with
+`triage:todo` may be a duplicate, out of scope, or need a different approach
+than what's described. Wait for `triage:accepted` or ask in the issue if you're
+interested.
+
+### Finding Work
+
+- **`good first issue`** — scoped issues suitable for newcomers to the
```
cijothomas, line 193: there are good starting points, but we should welcome if anyone is willing to work on issues outside these labels.

CONTRIBUTING.md
```diff
@@ -28,6 +28,48 @@ and
 
 ## Pull Requests
 
+### Target Audiences
```
cijothomas: Don't think we need this section for this project. I can't recall situations where we had to deal with this conflict.

CONTRIBUTING.md
```diff
@@ -98,6 +140,20 @@ for us to generate a changelog  standard.
 - Include benchmarks (before/after) in the summary, for contributions that are
   performance enhancements.
 
+### Pull Request Size and Scope
```
bryantbiggs, line 126: what about elaborating a bit more on scope - i.e changes that cut across crates or signals?

scottgerring:
> FYI - there is a somewhat common workflow used on other active projects to auto label all new issues/PRs with something like `triage:todo` or `needs-triage`. that would be easy an easy, low hanging addition to automatically mark those that haven't been seen by a maintainer where they could then apply the more appropriate label (manually). there are other similar avenues such as:
> 
> * Auto label based on affected code (i.e. - highlight that the issue or PR affects the sdk or core which usually carries more scrutiny given how vital they are to everything else)

Yep - I call that out in the issue text. We can steal what collector's doing basically for the common labelling work it looks like, but I will do it in a subsequent PR. 

> * Auto label based on breaking change - use something like `cargo-semver-checks` to apply a `breaking-change` label (would only work on PRs - harder to infer from an issue in an automated manner)

this would be helpful and rust specific 💪

PR #3387, opened 2026-02-24
feat(appender-tracing): add with_span_attribute_filter to experimental_span_attributes

codefromthecrypt:
## Changes

Add `with_span_attribute_filter` to `OpenTelemetryTracingBridge` (`experimental_span_attributes` feature). Only named span attributes are stored and propagated to log records. Without it, all span attributes are copied.

```rust
let bridge = OpenTelemetryTracingBridge::new(&provider)
    .with_span_attribute_filter(["session.id"]);
```

Filtering happens at storage time in `SpanFieldVisitor`, so non-matching attributes are never allocated.

[Goose](https://github.com/block/goose) built a 60-line custom `SessionIdBridge` just to propagate `session.id` from spans to logs. This replaces that workaround. .NET has the same gap ([dotnet#5239](https://github.com/open-telemetry/opentelemetry-dotnet/issues/5239)). Java's `captureMdcAttributes` allow-list is prior art.

### DIFF
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -6,7 +6,6 @@
   tracing spans (`experimental_span_attributes`). Use
   `OpenTelemetryTracingBridge::builder()` with `with_span_attribute_allowlist`
   to control which span attributes are copied to log records.
-  [#3387](https://github.com/open-telemetry/opentelemetry-rust/pull/3387)
 
 - Remove the `experimental_use_tracing_span_context` since
   `tracing-opentelemetry` now supports [activating][31901]  the OpenTelemetry


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -179,82 +181,114 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
 #[cfg(feature = "experimental_span_attributes")]
-struct SpanFieldVisitor<'a>(&'a mut Vec<(Key, AnyValue)>);
+struct SpanFieldVisitor<'a> {
+    attributes: &'a mut Vec<(Key, AnyValue)>,
+    filter: Option<&'a HashSet<Key>>,
+}
+
+#[cfg(feature = "experimental_span_attributes")]
+impl SpanFieldVisitor<'_> {
+    fn allowed(&self, field: &tracing::field::Field) -> bool {
```
cijothomas, line 197: the `Key` is constructed here, and again on the caller if we chose to include it. Would be faster to create the Key at caller and pass a ref to this method, to avoid dual creation.

Also - do we have to use `Key` as key in HashSet? Could a `&'static str` or `Cow` work? Might be irrelevant as OTel Key has special casing for static strings, and IIRC, all field names from `tracing` are 'static, so this should be just 1 ns only.
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/benches/attributes.rs#L6

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -292,8 +328,20 @@ where
             // See https://github.com/open-telemetry/semantic-conventions/issues/1550
             logger: provider.logger(""),
             _phantom: Default::default(),
+            #[cfg(feature = "experimental_span_attributes")]
+            span_attribute_filter: None,
         }
     }
+
+    /// Only copy span attributes with the given keys to log records.
```
cijothomas: there would be other options in this area, so maybe its time to introduce a proper builder pattern here?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -292,8 +328,20 @@ where
...
+    #[cfg(feature = "experimental_span_attributes")]
+    pub fn with_span_attribute_filter(
```
cijothomas:
```suggestion
    pub fn with_span_attribute_allowlist(
```

WDYT of this? "filter" was confusing to me as it was not clear if it is allow or deny list. We may in future have a with_span_attribute_deny_list as well.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -179,82 +181,114 @@ impl<LR: LogRecord> tracing::field::Visit for EventVisitor<'_, LR> {
...
+#[cfg(feature = "experimental_span_attributes")]
+struct SpanFieldVisitor<'a> {
+    attributes: &'a mut Vec<(Key, AnyValue)>,
+    allowlist: Option<&'a HashSet<&'static str>>,
+}
+
 #[cfg(feature = "experimental_span_attributes")]
-struct SpanFieldVisitor<'a>(&'a mut Vec<(Key, AnyValue)>);
+impl SpanFieldVisitor<'_> {
+    fn allowed(&self, field: &tracing::field::Field) -> bool {
```
cijothomas, line 197: inline candidate?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1233,4 +1324,153 @@ mod tests {
...
+    #[test]
+    #[cfg(feature = "experimental_span_attributes")]
+    fn tracing_appender_span_attribute_allowlist_empty_copies_none() {
+        let exporter = InMemoryLogExporter::default();
+        let provider = SdkLoggerProvider::builder()
+            .with_simple_exporter(exporter.clone())
+            .build();
+
+        let layer = layer::OpenTelemetryTracingBridge::builder(&provider)
+            .with_span_attribute_allowlist(std::iter::empty::<&str>())
```
cijothomas, line 1423: this is worth discussing more..If a user did not want any attributes, they should be disabling the feature flag. or when we remove ff, it'll still be an opt-in mechanism, so by not opting in, they get the effect of "dont add anything from span".

Given that, what would be the reason a user provide an empty list? I guess they didn't knew about the better option or by mistake. Is it better to validate and reject it during build() itself?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -284,14 +320,57 @@ where
...
+where
+    P: LoggerProvider<Logger = L> + Send + Sync,
+    L: Logger + Send + Sync,
+{
+    /// Only copy span attributes whose keys are in the given allowlist to log
+    /// records. When no allowlist is set, all span attributes are copied.
+    #[cfg(feature = "experimental_span_attributes")]
+    pub fn with_span_attribute_allowlist(
+        mut self,
+        keys: impl IntoIterator<Item = &'static str>,
```
cijothomas: Given `tracing` field names are always `&'static str`, this might be okay. But I think this will prevent users from providing this list from a file or someplace? 
Not blocking this PR, but something to consider and decide before stabilizing this feature.

opentelemetry-appender-tracing/src/layer.rs
codefromthecrypt: done, good call

opentelemetry-appender-tracing/src/layer.rs
codefromthecrypt: good point.. build() now normalizes an empty allowlist to None (treat as not set). if someone doesn't want span attributes they should disable the feature flag like you said

opentelemetry-appender-tracing/src/layer.rs
codefromthecrypt: good catch.. switched to `Cow<'static, str>` which is the pattern used throughout the codebase (e.g. `with_name`, `with_description`, `with_unit`). zero-cost for static strings but also works if someone loads keys from a config file

opentelemetry-appender-tracing/CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
 
 ## vNext
 
-- New *experimental* feature to enrich log records with attributes from active tracing spans (`experimental_span_attributes`)
+- New *experimental* feature to enrich log records with attributes from active
```
codefromthecrypt, line 5: @cijothomas since this is unreleased, I adjusted the existing changelog to highlight there's a builder. hope this makes sense and if not you can ask me to adjust it or you can.

cijothomas (approved):

codefromthecrypt:
@cijothomas Fyi my goose PR hasn't merged yet. Before I rebase that.. any rough ETA for a release here? It affects if I copy the code vs upgrade

Issue #3371, opened 2026-02-20
Metrics ValueMap: evaluate sorted-only storage vs dual-insert for attribute handling


bryantbiggs:
## Problem

In `opentelemetry-sdk/src/metrics/internal/mod.rs`, `ValueMap::measure` inserts **two** entries into the `trackers` HashMap for every new attribute set: one keyed by the original (caller-provided) order and one keyed by the sorted order:

```rust
trackers.insert(attributes.to_vec(), new_tracker.clone());
trackers.insert(sorted_attrs, new_tracker);
```

This doubles the memory footprint of the map and doubles the work during collection (`collect_readonly` / `collect_and_reset`), which already has to deduplicate via a `HashSet<*const A>` because it knows duplicates exist.

The dual insertion is intentional — it avoids sorting on every incoming measurement by caching the unsorted order as a fast-path lookup. The tradeoff is memory and collection overhead vs per-measurement sort cost.

## Proposal

Evaluate whether storing only the canonical (sorted) representation, combined with a fast `is_sorted` check to skip sorting for pre-sorted callers, outperforms the current dual-insert approach.

An `is_sorted` check is O(n) with no allocation. For callers who always pass attributes in the same or sorted order, this would be nearly free. The question is whether the memory savings and simplified collection path offset the cost of the `is_sorted` check (and occasional sort) on the hot path.

`ValueMap` is `pub(crate)` and the `trackers` field and `sort_and_dedup` function are not part of the public API — this is an internal implementation detail, so any change here is non-breaking.

**This should only be changed if benchmarking demonstrates a meaningful improvement.** The goal is to gather data, not to push a specific solution.

## Out of scope

The following were discussed but are low return or premature to address:

- **Thread-local sort buffer in `sort_and_dedup`** — The common case is existing attributes (cache hit), so the allocation only happens on cache miss. Low return until #2328 is resolved.
- **Sorting `InstrumentationScope` attributes at construction time** — Instrument creation and scope lookups are not on the hot path with typical scope counts (1-5 per application).

## Context

- ValueMap sharding: #2297
- Bound instruments: #1374
- Metrics API usage pattern: #2328

PR #3388, opened 2026-02-24
feat(sdk): support bring-your-own hasher for metrics via BuildHasher

bryantbiggs:
## Summary

Following the discussion below, this PR pivots from the original
`metrics-use-foldhash` feature flag to a **bring-your-own-hasher** design, as
suggested by @utpilla and @scottgerring.

The metrics hot path (`counter.add()` / `histogram.record()`) looks up an
aggregation bucket for each attribute set in an internal `HashMap` inside
`ValueMap`. That map is now parameterized over a user-chosen
[`BuildHasher`], defaulting to the standard library's `RandomState`
(SipHash-1-3). Users who want more throughput on high-cardinality workloads can
supply their own hasher via the new `MeterProviderBuilder::with_hasher`:

```rust
    .with_hasher(foldhash::fast::RandomState::default())
    .with_reader(reader)
    .build();
```

No new SDK feature flags, and **no new SDK runtime dependencies** — users bring
their own hasher crate (`foldhash`, `ahash`, or anything implementing
`BuildHasher`).

## Why this approach

This addresses the concerns raised in review:

- **Safe by default** (@cijothomas, @utpilla): the default stays `RandomState`
  (SipHash), which is HashDoS-resistant. The faster, non-resistant path is
  strictly opt-in. Existing behavior is unchanged.
- **No feature-flag / dependency sprawl** (@utpilla): rather than adding one
  feature flag per hasher (each a new dependency to vet), users pick any hasher.
  The `foldhash` dependency, the `metrics-use-foldhash` feature, and the
  `deny.toml` exception from the previous revision are gone.
- **Configurable via the public API + user code** (@scottgerring): the choice
  lives entirely in user code through `with_hasher`.

## Design / public API

The chosen hasher is monomorphized into the per-instrument `ValueMap` storage
and then **type-erased** at the existing `Arc<dyn Measure>` boundary, via an
internal aggregate-function factory. As a result:

- `SdkMeterProvider` and `MeterProviderBuilder` stay **non-generic** —
  `with_hasher` returns the same builder type, and the provider keeps working
  through the type-erased `dyn MeterProvider` path used by
  `opentelemetry::global`. The public API adds **exactly one method and no type
  parameters**; code that does not call `with_hasher` is entirely unaffected.
- The measurement hot path is unchanged: `ValueMap` is concrete behind the
  existing `Arc<dyn Measure>`, so the only added indirection is a single
  `dyn` dispatch at instrument-creation time (cold path), never on
  `add`/`record`.
- Only `std` types (`BuildHasher`, `RandomState`) appear in the public API, so
  no `allowed-external-types.toml` change is required.

*Alternatives considered:* exposing the hasher as a public generic type
parameter on `SdkMeterProvider<S = RandomState>` (threaded through the meter
provider) was prototyped, but it leaks a viral, hard-to-reverse type parameter
into the SDK's most central public type for a niche, set-once performance knob.
The type-erased approach keeps that cost out of the public API.

## Benchmark

`benches/metrics_hasher.rs` compares the counter hot path using the default
SipHash against `foldhash` installed via `with_hasher`. Both providers are
identical except for the hash builder. 1600 time series, 4 attributes per
measurement.

| Hasher (counter hot path) | Time per `add` | vs. default |
|---|---|---|
| `RandomState` (SipHash-1-3, default) | ~213 ns | — |
| `foldhash` via `with_hasher` | ~151 ns | **−29%** |

<sub>AMD Ryzen Threadripper 1900X, x86_64, rustc 1.93.0, criterion 0.5. Single
representative run; absolute numbers are hardware-dependent.</sub>

Run with:

```shell
cargo bench --bench metrics_hasher --features metrics,experimental_metrics_custom_reader
```

Refs: #3371

### DIFF
--- a/examples/metrics-advanced/src/main.rs
+++ b/examples/metrics-advanced/src/main.rs
@@ -72,7 +72,19 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
         .with_service_name("metrics-advanced-example")
         .build();
 
+    // Example 4 - Use a custom hasher for the internal metric trackers maps.
+    //
+    // Every measurement hashes its attribute set to look up an aggregation
+    // bucket in an internal hash map, so the hasher's cost is paid on each
+    // recording regardless of cardinality. By default this uses the standard
+    // library's SipHash, which is resistant to hash-flooding (HashDoS) attacks.
+    // When attribute values come from trusted sources, a faster
+    // non-cryptographic hasher (here `foldhash`) can measurably improve
+    // recording throughput, at the cost of HashDoS resistance.
+    //
+    // The default (no `with_hasher` call) remains the safe SipHash choice.
     let provider = SdkMeterProvider::builder()
+        .with_hasher(foldhash::fast::RandomState::default())
         .with_periodic_exporter(exporter)
         .with_resource(resource)
         .with_view(my_view_rename_and_unit)
--- a/opentelemetry-sdk/Cargo.toml
+++ b/opentelemetry-sdk/Cargo.toml
@@ -35,6 +35,9 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dev-dependencies]
 criterion = { workspace = true, features = ["html_reports"] }
+# Used only by the `metrics_hasher` benchmark to exercise a custom `BuildHasher`
+# via `MeterProviderBuilder::with_hasher`. Not a runtime dependency of the SDK.
+foldhash = { workspace = true }
 rstest = { workspace = true }
 temp-env = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
@@ -118,6 +121,11 @@ required-features = ["testing"]
 name = "metric"
 harness = false
 required-features = ["metrics", "spec_unstable_metrics_views", "experimental_metrics_custom_reader"]
+
+[[bench]]
+name = "metrics_hasher"
+harness = false
+required-features = ["metrics", "experimental_metrics_custom_reader"]
 
 [[bench]]
 name = "log"
--- a/opentelemetry-sdk/benches/metrics_hasher.rs
+++ b/opentelemetry-sdk/benches/metrics_hasher.rs
@@ -1,0 +1,86 @@
+//! Compares the metrics hot path (`Counter::add`) using the default
+//! DoS-resistant `RandomState` (SipHash-1-3) against a user-supplied
+//! non-cryptographic hasher (`foldhash`) installed via
+//! [`MeterProviderBuilder::with_hasher`].
+//!
+//! A single fixed attribute set is recorded repeatedly, so the only difference
+//! between the two benchmarks is the hash builder threaded into the internal
+//! `ValueMap` trackers map. This isolates the per-measurement hashing cost.
+//!
+//! Run with:
+//! ```text
+//! cargo bench --bench metrics_hasher --features metrics,experimental_metrics_custom_reader
+//! ```
+//!
+//! Results (Criterion median per `Counter::add`, AMD Ryzen Threadripper 1900X):
+//!
+//! ```text
+//! Counter_Add_SipHash_Default   ~153 ns
+//! Counter_Add_Foldhash          ~108 ns   (~30% faster)
+//! ```
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use opentelemetry::{
+    metrics::{Counter, MeterProvider as _},
+    KeyValue,
+};
+use opentelemetry_sdk::metrics::{ManualReader, SdkMeterProvider};
+use std::hash::BuildHasher;
+
+fn counter_with_hasher<S>(name: &'static str, hash_builder: S) -> Counter<u64>
+where
+    S: BuildHasher + Clone + Send + Sync + 'static,
+{
+    let meter_provider = SdkMeterProvider::builder()
+        .with_hasher(hash_builder)
+        .with_reader(ManualReader::builder().build())
+        .build();
+    meter_provider.meter("benchmarks").u64_counter(name).build()
+}
+
+fn bench_counter(c: &mut Criterion, label: &str, counter: Counter<u64>) {
+    // A single fixed attribute set: every iteration hits the same tracker, so
+    // the measured delta is purely the hasher cost on a steady-state lookup.
+    let attributes = [
+        KeyValue::new("attribute1", "value1"),
+        KeyValue::new("attribute2", "value2"),
+        KeyValue::new("attribute3", "value3"),
+        KeyValue::new("attribute4", "value4"),
+    ];
+    c.bench_function(label, |b| {
+        b.iter(|| {
+            counter.add(1, &attributes);
+        });
+    });
+}
+
+fn criterion_benchmark(c: &mut Criterion) {
+    // Default: standard library `RandomState` (SipHash-1-3, DoS-resistant).
+    bench_counter(
+        c,
+        "Counter_Add_SipHash_Default",
+        counter_with_hasher(
+            "Counter_Add_SipHash_Default",
+            std::collections::hash_map::RandomState::new(),
+        ),
+    );
+
+    // Bring-your-own: foldhash (fast, non-DoS-resistant).
+    bench_counter(
+        c,
+        "Counter_Add_Foldhash",
+        counter_with_hasher(
+            "Counter_Add_Foldhash",
+            foldhash::fast::RandomState::default(),
+        ),
+    );
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default()
+        .warm_up_time(std::time::Duration::from_secs(1))
+        .measurement_time(std::time::Duration::from_secs(2));
+    targets = criterion_benchmark
+}
+criterion_main!(benches);
--- a/opentelemetry-sdk/src/metrics/internal/aggregate.rs
+++ b/opentelemetry-sdk/src/metrics/internal/aggregate.rs
@@ -1,4 +1,6 @@
 use std::{
+    collections::hash_map::RandomState,
+    hash::BuildHasher,
     marker,
     mem::replace,
     ops::DerefMut,
@@ -122,7 +124,7 @@ impl Default for AggregateTimeInitiator {
     }
 }
 
-type Filter = Arc<dyn Fn(&KeyValue) -> bool + Send + Sync>;
+pub(crate) type Filter = Arc<dyn Fn(&KeyValue) -> bool + Send + Sync>;
 
 /// Applies filter on provided attribute set
 /// No-op, if filter is not set
@@ -147,8 +149,12 @@ impl AttributeSetFilter {
     }
 }
 
-/// Builds aggregate functions
-pub(crate) struct AggregateBuilder<T> {
+/// Builds aggregate functions.
+///
+/// `S` is the [`BuildHasher`] for each aggregation's trackers map. Defaults to
+/// [`RandomState`] (SipHash-1-3), which is HashDoS-resistant; install a faster
+/// hasher via [`MeterProviderBuilder::with_hasher`](crate::metrics::MeterProviderBuilder::with_hasher).
+pub(crate) struct AggregateBuilder<T, S = RandomState> {
     /// The temporality used for the returned aggregate functions.
     temporality: Temporality,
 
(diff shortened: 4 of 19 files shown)


opentelemetry-sdk/Cargo.toml
```diff
@@ -11,6 +11,7 @@ rust-version = "1.75.0"
 autobenches = false
 
 [dependencies]
+ahash = "0.8"
```
cijothomas: We try to keep dependencies absolute minimum, so if we chose to do it, it must be via a opt-in feature flag, so users can knowingly opt-in to this.

cijothomas (changes requested):
Thanks for striving to improve metrics perf. As noted in my comments, I am okay with this change if we feature gate it, so users are explicitly opting into this.

@utpilla might have considered/tried this before, but we didn't quite end up adding it - Would want to get his thoughts too.

opentelemetry-sdk/Cargo.toml
bryantbiggs: perfectly reasonable! if we do go forward, any thoughts on feature name? `alt_hasher`? `ahash`? (naming is hard 😅 )

cijothomas: Tagging @utpilla for thoughts as he might have explored this before. 

Any prior art in similar crates to steal feature names from ? `metrics-use-ahash` (to indicate this is specifically for metrics)

bryantbiggs:
> A cleaner approach would be to make the internal storage generic over [BuildHasher](https://doc.rust-lang.org/std/hash/trait.BuildHasher.html), similar to how HashMap itself works. The default stays as RandomState (SipHash), keeping the safe behavior for users who don't opt in, but users who want to bring their own hasher: foldhash, ahash, or anything else can do so without us needing to take on additional dependencies or feature flags. This does mean the generic parameter would need to thread through to MeterProvider and related types, which is a non-trivial change, but I think it's worth exploring. @bryantbiggs would you be interested in checking the feasibility of this approach?

Yes, I can take a look and propose something.

cijothomas:
@bryantbiggs Do you have bandwidth to check how easy it is to support bring-your-own-hash to the Metric sdk? If it is too much changes, I am okay with the feature-flag option for now. There is a risk of more users asking for different hash approaches, but lets tackle it when we actually reach there.

bryantbiggs:
@utpilla @cijothomas @scottgerring — thanks for the thoughtful feedback here. I've reworked the PR to follow your suggestion: instead of a per-hasher feature flag, it now supports bring-your-own-hasher.

- The metrics storage accepts any `std::hash::BuildHasher` via a new `MeterProviderBuilder::with_hasher(...)`. The default stays `RandomState` (SipHash-1-3), so it remains safe-by-default and existing behavior is unchanged.
- No new feature flags and no new SDK runtime dependencies — users bring their own hasher crate (`foldhash`, `ahash`, etc.). The previous `metrics-use-foldhash` feature, the `foldhash` dependency, and its `deny.toml` exception are gone.
- The hasher is type-erased internally (at the existing `Arc<dyn Measure>` boundary), so `SdkMeterProvider` and `MeterProviderBuilder` stay non-generic — `with_hasher` returns the same builder type and the global `dyn MeterProvider` path keeps working. The public API adds exactly one method and no type parameters; code that doesn't opt in is unaffected.

I also prototyped the explicit-generic variant (`SdkMeterProvider<S = RandomState>` threaded through the provider, as @utpilla originally described). I ultimately went with the type-erased approach to keep a viral type parameter out of the SDK's most central public type for what is a set-once, niche performance knob — but I'm very happy to switch to the explicit generic if you'd prefer that tradeoff.

Updated design rationale and a benchmark (~29% faster counter hot path with `foldhash` vs. the SipHash default) are in the PR description. Thanks again for the guidance!

examples/metrics-advanced/src/main.rs
```diff
@@ -72,7 +72,18 @@ fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
         .with_service_name("metrics-advanced-example")
         .build();
 
+    // Example 4 - Use a custom hasher for the internal metric trackers maps.
+    //
+    // On the measurement hot path the SDK looks up an aggregation bucket for
+    // each attribute set in an internal hash map. By default this uses the
+    // standard library's SipHash, which is resistant to hash-flooding (HashDoS)
+    // attacks. For high-cardinality workloads whose attribute values come from
```
cijothomas: does it matter how high the cardinality is? Hashing cost if still paid irrespective of whether attributes has high cardinality or not.

opentelemetry-sdk/benches/metrics_hasher.rs
```diff
@@ -0,0 +1,106 @@
...
+//! [`MeterProviderBuilder::with_hasher`].
+//!
+//! Both providers are byte-identical except for the hash builder threaded into
+//! the internal `ValueMap` trackers map, so the delta isolates the cost of the
+//! hasher on the lookup/insert path.
+//!
+//! Run with:
+//! ```text
+//! cargo bench --bench metrics_hasher --features metrics,experimental_metrics_custom_reader
+//! ```
```
cijothomas, line 13: nit: can you share the numbers as a comment here for quick reference. (similar to how we do for most other benchmarks)

opentelemetry-sdk/src/metrics/internal/aggregate.rs
```diff
@@ -148,7 +150,12 @@ impl AttributeSetFilter {
 }
 
 /// Builds aggregate functions
-pub(crate) struct AggregateBuilder<T> {
+///
+/// The `S` type parameter selects the [`BuildHasher`] threaded into the
+/// internal trackers map of every aggregation this builder produces. It
+/// defaults to [`RandomState`] (SipHash-1-3), preserving the DoS-resistant
+/// default; users can opt into a faster hasher via the meter provider builder.
+pub(crate) struct AggregateBuilder<T, S = RandomState> {
```
cijothomas, line 157: /// Defaults to `RandomState` (SipHash-1-3), which is HashDoS-resistant.
 /// Use [`MeterProviderBuilder::with_hasher`] to install a faster hasher.

^ lets use comments like this in code docs throughout. Else this reads as PR-motivation than code documentation. (they can still be in changlog/ PR desc)

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -399,7 +413,12 @@ where
                 return;
             };
             self.count.store(0, Ordering::SeqCst);
-            std::mem::take(&mut *trackers)
+            // Replace with a fresh empty map that reuses the same hash builder,
+            // cloned from the current map. Using `mem::replace` (rather than
+            // `mem::take`) avoids requiring `S: Default` and preserves the
```
cijothomas: Just realized that we had a *gap* in this area - we used to `take` leaving a default HashMap in place which didn't start with the required capacity... That is an existing gap. We could fix that while we are here, or leave it for another short follow up.

cijothomas: Its simple enough - lets fix that in this PR itself.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -101,17 +108,24 @@ where
...
     pub(crate) fn config(&self) -> &A::InitConfig {
         &self.config
     }
 
-    fn new(config: A::InitConfig, cardinality_limit: usize) -> Self {
+    fn new(config: A::InitConfig, cardinality_limit: usize, hash_builder: S) -> Self {
         ValueMap {
-            trackers: RwLock::new(HashMap::with_capacity(1 + cardinality_limit)),
+            // Move the hash builder into the map; it can be recovered later via
+            // `HashMap::hasher` (e.g. in `drain_and_reset`), so there's no need
```
cijothomas:
```suggestion
            // `HashMap::hasher` (used in `drain_and_reset`), so there's no need
```

opentelemetry-sdk/benches/metrics_hasher.rs
```diff
@@ -0,0 +1,106 @@
...
+                    [
+                        rng.random_range(0..4),
+                        rng.random_range(0..4),
+                        rng.random_range(0..10),
+                        rng.random_range(0..10),
+                    ]
+                })
+            },
+            |rands| {
+                counter.add(
```
cijothomas: for demonstrating the gain of improved hashing, lets use a single attribute-set and remove the random etc, to get a pure feel of the gains by using faster hash.

opentelemetry-sdk/src/metrics/meter_provider.rs
```diff
@@ -41,6 +42,9 @@ struct SdkMeterProviderInner {
     pipes: Arc<Pipelines>,
     meters: Mutex<HashMap<InstrumentationScope, Arc<SdkMeter>>>,
     shutdown_invoked: AtomicBool,
+    /// The per-numeric-type aggregate-function builders, with the chosen hasher
+    /// already erased in. Threaded into every meter this provider creates.
+    builders: AggregateFnsBuilders,
```
cijothomas, line 47: we use the same seed for every instrument inside this provider right? Should we keep the existing behavior of generating fresh seed for each value map (i.e each instrument?)

cijothomas (commented):
This is awesome. Left few minor comments. Not yet marking as approved - will do another review once the comments are addressed. 
Would also need another pair of eyes (anything touching metric aggregation is complex and benefits from multiple reviewers!)

examples/metrics-advanced/src/main.rs
bryantbiggs: You're right — the hash is computed on every measurement regardless of cardinality, so tying the benefit to "high-cardinality" was misleading. Reworded in https://github.com/open-telemetry/opentelemetry-rust/pull/3388/commits/070956d375fce0e6fc26190e22f6a4bb2368b2d9 to frame it as a per-measurement cost: a faster hasher helps recording throughput on any hot instrument, independent of how many distinct attribute sets exist.

opentelemetry-sdk/benches/metrics_hasher.rs
bryantbiggs: Added in https://github.com/open-telemetry/opentelemetry-rust/pull/3388/commits/070956d375fce0e6fc26190e22f6a4bb2368b2d9. With the single-attribute-set bench (see below), median `Counter::add` is ~153 ns with the default SipHash vs ~108 ns with foldhash (~30% faster) on a Threadripper 1900X.

opentelemetry-sdk/src/metrics/internal/aggregate.rs
bryantbiggs: Agreed. Trimmed the `AggregateBuilder` doc to the terse code-doc style you suggested in https://github.com/open-telemetry/opentelemetry-rust/pull/3388/commits/070956d375fce0e6fc26190e22f6a4bb2368b2d9.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Fixed in https://github.com/open-telemetry/opentelemetry-rust/pull/3388/commits/070956d375fce0e6fc26190e22f6a4bb2368b2d9: `drain_and_reset` now replaces the trackers with `HashMap::with_capacity_and_hasher(1 + cardinality_limit, …)` instead of a zero-capacity map, so the next collection cycle refills without reallocating. Added a regression test (`drain_and_reset_preallocates_tracker_capacity`) that I confirmed fails on the old `with_hasher` version (capacity 0).

opentelemetry-sdk/benches/metrics_hasher.rs
bryantbiggs: Done in https://github.com/open-telemetry/opentelemetry-rust/pull/3388/commits/070956d375fce0e6fc26190e22f6a4bb2368b2d9. Dropped the RNG and the 1600-series spread; the bench now records one fixed attribute set repeatedly, so the delta is purely the hasher cost on a steady-state lookup. That's where the ~153 ns vs ~108 ns numbers above come from.

opentelemetry-sdk/src/metrics/meter_provider.rs
bryantbiggs: Good catch — fixed in https://github.com/open-telemetry/opentelemetry-rust/pull/3388/commits/070956d375fce0e6fc26190e22f6a4bb2368b2d9 to preserve the per-instrument behavior for the default. The builder now stores a hasher *factory* (`Arc<dyn Fn() -> S>`) instead of one stored instance: the default uses `RandomState::new`, so each instrument's `ValueMap` gets a freshly-seeded `RandomState` (matching std's per-`HashMap` seeding and the prior behavior), while `with_hasher(h)` clones the user's instance so a configured/fixed-seed hasher is preserved across instruments. Added `default_hasher_is_reseeded_per_instrument` and `custom_hasher_seed_is_preserved_across_instruments` tests (the former I confirmed fails if the default falls back to a shared cloned seed).

For what it's worth on the threat model: a single per-provider seed wouldn't have been a HashDoS regression here (the seed is process-random and never exposed; std's per-map seeding is itself just a counter increment off a shared per-thread key, and it mainly guards against iteration-order leakage, which these internal maps don't expose). But restoring fresh-per-instrument for the default is essentially free and keeps behavior identical to `main`, so it's the cleaner choice.

PR #3389, opened 2026-02-25
docs: OTLP Exporter full config in docs

cijothomas:
We never had a full doc on all the possible configuration options, so this is fixing that gap.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -35,7 +35,7 @@ jobs:
       GIT_DISCOVERY_ACROSS_FILESYSTEM: 1
     steps:
       - name: Harden the runner (Audit all outbound calls)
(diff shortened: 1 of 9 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -463,7 +682,10 @@ pub use crate::exporter::{
     OTEL_EXPORTER_OTLP_TIMEOUT_DEFAULT,
 };
 
-#[cfg(feature = "experimental-http-retry")]
+#[cfg(any(
```
cijothomas, line 685: @scottgerring this was a miss I guess, so fixing it!

scottgerring (approved):
LGTM, pending adding the doc link syntax copilot suggested.
I like the feature flag call-outs; this'll make it much easier to get started I reckon.

PR #3391, opened 2026-02-25
perf: Additional perf behches for tracing appender

cijothomas:
Fixes #

## Changes

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -35,7 +35,7 @@ jobs:
       GIT_DISCOVERY_ACROSS_FILESYSTEM: 1
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
+        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
         with:
           egress-policy: audit
 
@@ -47,7 +47,7 @@ jobs:
       - name: Make repo safe for Git inside container
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
 
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 10 # Fetch a bit of history so we can do perf diffs
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,11 @@ jobs:
     continue-on-error: ${{ matrix.rust == 'beta' }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
       with:
         egress-policy: audit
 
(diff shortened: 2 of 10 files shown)


opentelemetry-appender-tracing/benches/log-attributes.rs
```diff
@@ -4,22 +4,24 @@
...
-| otel_4_attributes    | 240 ns       | +44 ns    |
-| otel_5_attributes    | 278 ns       | +38 ns    |
-| otel_6_attributes    | 346 ns       | +68 ns    | // Array is full. 6th attribute causes vec! to be allocated
-| otel_7_attributes    | 390 ns       | +44 ns    |
-| otel_8_attributes    | 431 ns       | +41 ns    |
-| otel_9_attributes    | 480 ns       | +49 ns    |
-| otel_10_attributes   | 519 ns       | +39 ns    |
-| otel_11_attributes   | 625 ns       | +106 ns   | // vec! initial capacity is 5. 11th attribute causes vec! to be reallocated
-| otel_12_attributes   | 676 ns       | +51 ns    |
+| otel_0_attributes    | 51 ns        | -         |
```
scottgerring, line 12: Any idea why these have changed so much?

opentelemetry-appender-tracing/benches/span-attributes.rs
```diff
@@ -215,8 +229,101 @@ fn benchmark_nested_spans(c: &mut Criterion, depth: usize) {
...
+            });
+        });
+    });
+}
+
+/// Emit an error log with 1 attribute while inside a span that carries 2 attributes
+/// (total = 3 attrs on the log record, within the SDK's 5-attr inline threshold).
+/// With `experimental_span_attributes` the 2 span attributes are propagated onto the
+/// log record, so the delta vs `log_1_attr_no_span` shows the pure span-attribute cost.
+fn benchmark_log_1_attr_in_span_2_attr(c: &mut Criterion) {
```
scottgerring, line 269: This all looks well and good to me, but I don't have a good feeling about whether or not this particular usage is meaningful (2 atts on span, 1 on log).

I wonder if @bantonsson has any deep insight?

opentelemetry-appender-tracing/benches/log-attributes.rs
cijothomas: No code changes done to boost perf specifically. But rust compiler is new. It is entirely possible that my machine instability caused this. (After all, these raw numbers are just for easy reference, it does not really mean anything much.)

opentelemetry-appender-tracing/benches/span-attributes.rs
cijothomas: I'll explain my thinking:
we have reasonably optimized the perf for log path. The moment span_attribute enrichment is enabled, we have to do a bunch of work. These tests how much worse we are with span_attributes feature.
It helps answer questions like:
I am instrumenting my app with logs, and I need to add some common attribute to all log. I can either modify all callsite to add that common attribute to each log, or start a span with that attribute so logs get it automatically. While span approach makes coding easier, it comes with perf cost.

This is done to help compare the cost. We need to write a guide once I'm done with all measurements. So users can evaluate if the convenience is worth the perf. (or is the perf worth the inconvenience)

(These are real problems we are already facing in other OTel projects!)

opentelemetry-appender-tracing/benches/log-attributes.rs
scottgerring: Yeah; it just worries me a bit they've moved _so_ much! 

I think we should invest some time in getting the benchmark data continuously collected into a dashboard platform like [the collector does](https://opentelemetry.io/docs/collector/benchmarks/) for `main`. As long as we're ad-hoc running these with the `benchmark` tag and on our local machines we're not really reaping the full rewards of the benchmark suite. 

This is just an observation, not a call to action for this change

cijothomas: Agreed! Tracking trends over time is definitely the way to do.
(We still use benchmarks for before/after when targeting perf fixes.)

Issue #1374, opened 2023-11-15
[Feature]: Metrics - Bound instrument to achieve extremely high performance


cijothomas:
### Related Problems?

Performance of Metrics SDK is not good enough to be used in critical services.

### Describe the solution you'd like:

Majority of the time is spent in locating the time-series to aggregate to. If the key and value are known ahead of time, we can avoid this lookup. This used to exist in the spec as "bound instrument". but was removed from initial stable spec, due to lack of demand for such a feature.

https://github.com/open-telemetry/opentelemetry-specification/blob/12b5b27d4b6a4c0197560f508f0be3c38d190dc4/specification/metrics/old_api.md#bound-instrument-calling-convention

### Considered Alternatives

I am not aware of any other solution. Based on SDK benchmarks in OTel Rust and other languages (.NET in particular was used for comparison), the only way to really get blazing fast perf is by avoiding the manipulation/massaging+lookup of timeseries.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/metric.rs#L82-L87 shows that highest perf is achieved for no attributes, and perf decreases significantly with each attribute.

cijothomas:
Similar benchmark from OTel.NET metrics, showing the same conclusion : https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/Benchmarks/Metrics/MetricsBenchmarks.cs#L34-L36

jtescher:
Worth re-opening a spec issue? Seems like all language implementations will run into this same problem with the current APIs

cijothomas:
Will open spec issue for gauging interest from others. I am not very sure if every other language has such rigorous perf needs, so a likely outcome is OTel Rust offering this under an opt-in feature flag to begin with. And then trying to push this through spec.

lalitb:
Agree. I feel there is value in exploring and implementing this feature at-least for the languages used in high-throughout, low-latency environments. I will open a similar support request for otel-cpp.

KallDrexx:
I'm going to start exploring adding bound instrumentation support to the rust library, at least to give us a fast path for high throughput metric paths with relatively stable attribute sets.  

For our application when we generate a counter, that counter is always incremented with a stable set of dynamic attributes.  Those attributes can change for a particular counter, but it's infrequent for the life of that instance of a counter (and once it changes, it won't change for a while after).  

As a data point, under a load test of our application
[...]

PR #3392, opened 2026-02-25
feat(metrics): add bound instruments behind experimental feature flag

bryantbiggs:
Refs #1374, #2328

## Summary

Adds `BoundCounter` and `BoundHistogram` to the public API behind the `experimental_metrics_bound_instruments` feature flag. These types cache the resolved aggregator reference for a fixed attribute set, allowing subsequent measurements to bypass per-call sort, dedup, hash, and HashMap lookup entirely.

```rust
let counter = meter.u64_counter("requests").build();
let bound = counter.bind(&[KeyValue::new("method", "GET")]);
bound.add(1); // ~1.9ns — no attribute lookup
```

### TrackerEntry

```rust
pub(crate) struct TrackerEntry<A: Aggregator> {
    pub(crate) aggregator: A,
    pub(crate) has_been_updated: AtomicBool,
    pub(crate) bound_count: AtomicUsize,
}
```

`has_been_updated` tracks whether any measurement occurred since the last collect. `bound_count` tracks how many live `BoundCounter`/`BoundHistogram` handles reference this entry.

### Delta collection (`collect_and_reset`)

In-place iteration under a read lock (no write lock in steady state):
1. Export only entries where `has_been_updated` was true, then reset the flag
2. Entries with `bound_count > 0` are **never evicted** — they persist across idle cycles but produce no export when there are no updates
3. Stale unbound entries (`has_been_updated == false && bound_count == 0`) are evicted under a write lock with a TOCTOU re-check

### Cardinality overflow handling

`ValueMap::bind()` returns `Option<Arc<TrackerEntry>>` — `None` when at the cardinality limit. Bound handle types use a `Direct`/`Fallback` enum:

- **Direct**: normal case — dedicated tracker, single-digit ns, no map lookup per call
- **Fallback**: at overflow — delegates every call to the unbound `Measure::call()` path

This gives correct overflow attribution (same behavior as unbound at overflow) with automatic recovery when delta collect frees cardinality space. An `otel_debug!` log fires when fallback occurs.

Aggregator types that don't support direct binding (ExponentialHistogram, LastValue, PrecomputedSum) also use the fallback path gracefully instead of panicking.

### Feature flag

All public API, traits, noop impls, and internal plumbing are gated behind `experimental_metrics_bound_instruments`. The `testing` feature enables it by default.

## Benchmark Results

Apple M4 Max, 16 cores (12 performance + 4 efficiency), macOS 15.4:

| Benchmark | Time | vs Unbound |
|-----------|------|------------|
| Counter Unbound | 53.2 ns | — |
| **Counter Bound** | **1.87 ns** | **~28x faster** |
| Histogram Unbound | 58.6 ns | — |
| **Histogram Bound** | **6.57 ns** | **~8.9x faster** |

EC2 results across 9 instance configurations (c7i/c7a/c7g × large/xlarge/4xlarge) show 5-28x counter speedup and 5-9x histogram speedup depending on architecture. Full results in the expandable sections below.

<details>
<summary>EC2 Counter Results</summary>

| Instance | Unbound | Bound | Speedup |
|----------|---------|-------|---------|
| c7i.large | 86.2ns | 7.47ns | 11.5x |
| c7i.xlarge | 88.9ns | 7.62ns | 11.7x |
| c7i.4xlarge | 90.1ns | 7.65ns | 11.8x |
| c7a.large | 77.4ns | 2.71ns | 28.6x |
| c7a.xlarge | 75.0ns | 2.71ns | 27.7x |
| c7a.4xlarge | 75.0ns | 2.71ns | 27.7x |
| c7g.large | 111.8ns | 6.87ns | 16.3x |
| c7g.xlarge | 111.4ns | 6.84ns | 16.3x |
| c7g.4xlarge | 112.9ns | 6.73ns | 16.8x |

</details>

<details>
<summary>EC2 Histogram Results</summary>

| Instance | Unbound | Bound | Speedup |
|----------|---------|-------|---------|
| c7i.large | 97.1ns | 17.1ns | 5.7x |
| c7i.xlarge | 99.0ns | 17.9ns | 5.5x |
| c7i.4xlarge | 98.5ns | 18.4ns | 5.4x |
| c7a.large | 92.5ns | 13.7ns | 6.8x |
| c7a.xlarge | 92.4ns | 13.7ns | 6.7x |
| c7a.4xlarge | 92.5ns | 13.7ns | 6.8x |
| c7g.large | 133.2ns | 27.7ns | 4.8x |
| c7g.xlarge | 133.5ns | 27.6ns | 4.8x |
| c7g.4xlarge | 133.8ns | 27.2ns | 4.9x |

</details>

<details>
<summary>Previous experiments (EXP-002, EXP-003) and why bound instruments are the right approach</summary>

### EXP-002: Micro-optimizations on existing architecture
- **Changes**: Relaxed atomic ordering, stale entry eviction during delta collect
- **Result**: <2% improvement — the HashMap bottleneck is untouched

### EXP-003: Sharded HashMap with pre-hashing
- **Changes**: 16 Mutex-guarded shards, pre-computed foldhash, PassthroughHasher
- **Result**: **+10-33% regression** across all architectures. `HashedAttributes::from_raw()` on every `measure()` call is more expensive than the existing two-lookup pattern.

### Key insight
Profiling shows 78% of CPU in HashMap key hashing (52%) and key comparison (26%). The only way to meaningfully improve measure() throughput is to **skip the lookup entirely**. Bound instruments achieve this by moving the cost to bind-time (once) rather than measure-time (every call).

</details>

## Test Coverage

15 tests covering:
- Cumulative and delta temporality for both Counter and Histogram
- Bound + unbound sharing the same data point (same attribute set)
- Idle delta cycles (no update = no export, but handle persists)
- Cardinality overflow fallback to unbound path (Counter and Histogram)
- Recovery after delta eviction frees cardinality space
- Overflow fallback handles working across multiple delta cycles
- Drop enabling eviction of stale entries
- Multiple bound handles sharing the same tracker (ref-counting)
- Binding with empty attributes

## Impact on existing code

- Zero regression on unbound measurement path
- No changes to collect path semantics for unbound instruments
- All existing tests pass
- Additive public API behind feature flag — no breaking changes

## Related

- OpenTelemetry Specification discussion: https://github.com/open-telemetry/opentelemetry-specification/issues/3521
- Issue: #1374 (bound instruments)
- Issue: #2328 (delta temporality collect improvement)


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -221,7 +293,11 @@ where
             let mut seen = HashSet::new();
             for (attrs, tracker) in trackers_collect.drain() {
                 if seen.insert(Arc::as_ptr(&tracker)) {
-                    dest.push(map_fn(attrs, tracker.clone_and_reset(&self.config)));
+                    dest.push(map_fn(
+                        attrs,
+                        tracker.aggregator.clone_and_reset(&self.config),
+                    ));
+                    tracker.evicted.store(true, Ordering::Release);
```
cijothomas: thinking out loud:
do we need to evict bound instruments after delta collect()? Can we define the behavior as bound instruments remains forever, unless user explicitly call an unbind()?
if there are no new updates in a collect() cycle, then we need to *not* export anything, but value map keeps the tracker.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -145,27 +167,77 @@ where
...
+            return Arc::clone(tracker);
+        }
+
+        if self.is_under_cardinality_limit() {
+            let new_tracker = Arc::new(TrackerEntry::<A>::new(&self.config));
+            trackers.insert(sorted_attrs, new_tracker.clone());
+            self.count.fetch_add(1, Ordering::SeqCst);
+            new_tracker
+        } else {
+            // Over cardinality limit — bind to the overflow tracker
```
cijothomas: this is tricky...when user called bind(), we may have been over limit and it gets bound to overflow. But after a delta collect, unused points are reclaimed, but the bound instrument still goes to overflow? 
Or the delta collect set evicted to true, and it fallsback to slow path?

We may have to see if bound instruments should be exempt from the cardinality limit? (that is not good and will violate spec too). Open for ideas.

opentelemetry-sdk/benches/bound_instruments.rs
```diff
@@ -0,0 +1,88 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use opentelemetry::{metrics::MeterProvider as _, KeyValue};
+use opentelemetry_sdk::metrics::{ManualReader, SdkMeterProvider, Temporality};
+
+fn create_provider(temporality: Temporality) -> SdkMeterProvider {
```
cijothomas, line 19: @bryantbiggs If you can include the bench results in code comment (similar to what we do in many other benches), that'll make it easy to get a quick feel of things. PR description has it, let's just copy it here. (do include hardware info if you are comfortable with that).

The numbers shared meets our expectation: single digit ns for counters!

There is contention still, but solving that is a separate effort. (Already solved in OpenTelemetry Arrow project, we can steal it when the time comes)

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Addressed in 936993fa — replaced the swap-and-drain `collect_and_reset` with in-place delta collection that supports bound instrument persistence.

**What changed:**

`TrackerEntry` now has two status fields instead of `evicted`:
- `has_been_updated: AtomicBool` — set on every `measure()`/`bound.add()` call, checked and reset during collect. Only entries updated since last collect are exported. **No updates in a cycle = no export, but the tracker stays in the map.**
- `bound_count: AtomicUsize` — incremented on `bind()`, decremented on `Drop` of the bound handle. Entries with `bound_count > 0` are **never evicted**, even if they had no updates in a cycle.

**Delta collect behavior:**
1. Iterate trackers under a **read lock** (no write lock needed in steady state)
2. Export only entries where `has_been_updated` was true, then reset the flag
3. Collect stale candidates: entries where `has_been_updated=false` AND `bound_count=0` AND not overflow sentinel
4. If stale candidates exist, acquire write lock, re-check `has_been_updated` under the lock (TOCTOU guard against concurrent `measure()`), then evict

**Bound handles** no longer check an `evicted` flag or have a fallback path — they always go direct to the aggregator. The `Drop` impl decrements `bound_count` so the entry becomes eligible for eviction after all handles are dropped.

The `drain_and_reset` path is preserved for async/Observable instruments (PrecomputedSum) that need true staleness detection via map clearing.

This design was informed by a previous experiment branch (`feat/delta-collect-optimization`, part of #2328) which validated the `has_been_updated` + in-place iteration pattern on EC2 benchmarks. The `bound_count` mechanism extends it to prevent eviction of bound entries.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Addressed in 7b531c41 with an overflow fallback to the unbound path.

**Solution: `bind()` returns `Option`, `None` at overflow**

`ValueMap::bind()` now returns `Option<Arc<TrackerEntry>>` — `None` when at/over the cardinality limit. The bound handle types (`BoundSumHandle`, `BoundHistogramHandle`) use a `Direct`/`Fallback` enum:

- **`Direct`**: normal case — dedicated tracker, single-digit ns, no map lookup on each call
- **`Fallback`**: at overflow — delegates every `.call()` to the unbound `Measure::call()` path

This gives us:
1. **Correct overflow attribution** — same behavior as unbound measurements at overflow, no data merging between distinct bound handles sharing an overflow tracker
2. **Automatic recovery** — when delta collect evicts stale entries and frees cardinality space, `measure.call()` creates a proper entry for those attrs on the next invocation
3. **No `bound_count` inflation** on the overflow tracker (was a problem with the previous approach of binding directly to it)

**Trade-off**: a `Fallback` handle is no faster than unbound (full HashMap lookup each time). But if you're at the cardinality limit when calling `bind()`, you're already in degraded mode — the user still gets correct data, just not the fast-path benefit. The guidance remains: bind early during initialization before cardinality fills up.

**Benchmarks confirm zero regression** on the `Direct` hot path from the enum `match`:

| Benchmark | Before | After |
|---|---|---|
| Counter Bound | 1.81 ns | 1.83 ns |
| Histogram Bound | 6.57 ns | 6.57 ns |
| Counter Unbound | 51.40 ns | 51.75 ns |
| Histogram Unbound | 55.91 ns | 57.35 ns |

Added 4 new tests covering overflow + delta interaction scenarios.

opentelemetry-sdk/benches/bound_instruments.rs
bryantbiggs: Added benchmark results as a comment block at the top of `benches/bound_instruments.rs` with Apple M4 Max hardware info and all benchmark numbers (Counter/Histogram unbound vs bound, plus multi-threaded bound counter at 2/4/8 threads).

Single-digit ns for bound counters as expected (~1.87 ns bound vs ~53 ns unbound = ~28x speedup). Histogram bound is ~6.6 ns vs ~59 ns unbound (~8.9x).

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -4680,4 +4703,656 @@ mod tests {
...
+        };
+
+        let overflow_dp =
+            find_overflow_sum_datapoint(&sum.data_points).expect("overflow point expected");
+        assert_eq!(
+            overflow_dp.value, 35,
+            "All overflow-bound measurements should accumulate in overflow bucket"
+        );
+
+        // Cycle 2: bound handles still work after delta collect
```
cijothomas, line 5039: This is the only part I am not so sure of. If bind() occurred when we were in overflow, we will always remain in overflow, even though we got space cleaned up. Should not be an issue in practice, as bound instruments have very specialized use case, and the bind occur at the very start of application typically, and hence unlikely they'd go into overflow.

Leaving this comment for us to revisit in future, not a blocker for this PR.

Some options to discuss in future:
1. bind() can return err/none if overflowing, so users know upfront they are not using bind() properly. This return approach is not done in OTel world so far, but bound is special and might call for exceptions.
2. See if we can internally adjust bindings from overflow to normal during a delta collect.
3. Exclude bind() from cardinality capping - users leveraging bind() knows that no matter what happens elsewhere, the bound ones always reports proper and never overflows. 
4. Other ideas.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -52,6 +55,34 @@ pub(crate) trait Aggregator {
...
+    pub(crate) aggregator: A,
+    pub(crate) has_been_updated: AtomicBool,
+    pub(crate) bound_count: AtomicUsize,
+}
+
+impl<A: Aggregator> TrackerEntry<A> {
+    fn new(config: &A::InitConfig) -> Self {
+        TrackerEntry {
+            aggregator: A::create(config),
+            has_been_updated: AtomicBool::new(true),
```
cijothomas: wondering why we start with `true`? Shouldn't it be `false` until the first recording come.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -185,47 +276,108 @@ where
...
+                .swap(false, Ordering::AcqRel)
+        {
+            dest.push(map_fn(vec![], &self.no_attribute_tracker.aggregator));
+        }
+
+        let overflow_attrs = stream_overflow_attributes();
+        let mut stale_entries: Vec<Arc<TrackerEntry<A>>> = Vec::new();
+
+        {
+            let Ok(trackers) = self.trackers.read() else {
```
cijothomas, line 320: the previous design of maintaining 2 trackers allowed us to require a lock for a single instruction swap. But now, we need read() lock held thoughout the iteration through the entries, and then write() lock throughout the stale_entry cleanups.

Given we are not clearing entires for delta  aggressively in each collect, this might be less impactful than it appears. The typical pattern is to re-use majority of trackers, so they don't get evicted, and hence, update/hot-path won't need to get write lock. The new attribute set (or the ones that went unused and making a comeback), would slightly suffer as its competing for a write lock() with collect, and the time it waits is directly proportional to the overall count of trackers.

This maybe a reasonable tradeoff to accept, as previous situation was definitely bad due to force clearing after every collect(), and thereby forcing write lock() on every attributeset after every collect().


Need some more time to think through this fully, but feel free to comment if my observations are correct.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Good catch — fixed in b7e91972. `has_been_updated` now initializes to `false`.

The two `measure()` creation paths (normal entry and overflow entry) now explicitly `.store(true, Relaxed)` after calling `update(value)`. The `bind()` creation path intentionally does *not* set it — a bound slot with no recording should produce no export, which is the correct semantic.

opentelemetry-sdk/src/metrics/mod.rs
bryantbiggs: Captured this as a TODO comment directly in the code at the overflow `None` return in `bind_sorted()` (b7e91972), linking back to this discussion thread so it doesn't get lost (if/when) the PR merges.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Your observations are correct. Here's how the locking trade-off breaks down:

**Old design (swap-and-drain):** Write lock on `trackers` held only for `mem::swap` (pointer swap — near instant), then iteration happens on a second map. But every delta collect clears *all* entries, so the *next* `measure()` for every attribute set must take a write lock to re-insert — `O(n)` write lock acquisitions on the hot path per collection cycle.

**New design (in-place iteration):** Read lock held throughout iteration, then write lock only if there are stale entries to evict.

**Steady state (the common case):** Attribute sets are reused across cycles → zero stale entries → no write lock during collect at all. Both `measure()` and `collect_and_reset` hold only read locks and run concurrently. This is strictly better than the old design where every attr set had to re-acquire a write lock after every collect.

**Edge case — new/returning attrs during collect:** As you noted, a `measure()` that needs to *insert* a new entry competes for the write lock with the read lock held by collect. Wait time is proportional to total tracker count. But this only affects never-before-seen attribute sets arriving *during* the iteration window. Once inserted, they're back on the read-lock fast path.

**Stale eviction write lock:** Duration proportional to the number of stale entries (the `retain()` call), not total tracker count. In steady state this never fires.

**Bound instruments:** Completely unaffected — they hold `Arc<TrackerEntry>` directly, bypassing the map and all locking entirely.

**Atomics safety during concurrent read-lock iteration:** `has_been_updated.swap(false)` and the aggregator read/reset use atomics internally (`AtomicI64`/`AtomicU64` for sums, `Mutex` for histograms). If a concurrent `measure()` sets `has_been_updated` back to `true` after the swap, that measurement is captured in the next cycle — not lost.

Net: the new design trades a brief read-lock hold during collect for eliminating `O(n)` write-lock acquisitions on the hot path every cycle. The only regression is the narrow window where brand-new attribute sets must wait for collect's read lock to release before inserting.

cijothomas: Thanks for confirming! My conclusion is same - this is a trade-off, and I think this is a reasonable balance. 

Briefly chatted with @utpilla on this as well (who has most experience with all perf things we have in this area!).

cijothomas (changes requested):
Repeating my earlier sentiment - the bound instruments design is solid and the benchmark results are impressive!

After reviewing the full diff, I think we should spliot this into two PRs:
PR 1: In-place delta collection (avoids draining HashMap)
PR 2: Bound instruments (builds on PR 1)

The main reason - delta collection refactor changes locking semantics and aggregator reset behavior for every existing user (it's not behind the feature flag, and we don't really want to guard it behind flag either), so it deserves focused review on its own. 

Splitting also makes each PR easier to review and means any issues found in the collect path refactor won't block the bound instruments design discussion

What do you think?

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/3421 Closing in favor of

Issue #3402, opened 2026-03-03
sync BatchSpanProcessor silently ignores OTEL_BSP_MAX_CONCURRENT_EXPORTS / max_concurrent_exports


glaziermag:
## Summary
The sync `BatchSpanProcessor` accepts `max_concurrent_exports` / `OTEL_BSP_MAX_CONCURRENT_EXPORTS`, but does not honor it at runtime.

This looks like a **misleading no-op configuration** (docs/config-contract mismatch), not a newly introduced semantic bug in the sync redesign.

## Why this is confusing
From a user perspective:
1. They set `with_max_concurrent_exports(...)` or `OTEL_BSP_MAX_CONCURRENT_EXPORTS`.
2. The SDK accepts the value.
3. Behavior stays serial, with no warning/error.

## Repro
I used a custom exporter that tracks concurrent in-flight `export()` calls via atomics and sleeps briefly inside `export()`.

Config used on sync BSP:
- `max_export_batch_size = 1`
- `max_queue_size = 32`
- `scheduled_delay = 3600s` (to avoid timer-triggered noise)
- `max_concurrent_exports = 4`

Then emit several spans quickly and `force_flush()`.

Observed: max overlap stayed `1` (serial).

## Differential evidence
The async-runtime BSP path does honor this knob and has behavior tests for both concurrent and serial modes:
- `trace::span_processor_with_async_runtime::tests::test_concurrent_exports_expected`
- `trace::span_processor_with_async_runtime::tests::test_exports_serial_when_max_concurrent_exports_1`

## Intent vs contract mismatch
The changelog for the redesign says sync BSP does not support multiple concurrent exports, which implies the current sync behavior is intentional.

However, the config/docs surface still broadly advertises `OTEL_BSP_MAX_CONCURRENT_EXPORTS` and accepts values in sync path configuration.

That combination makes the setting effectively a silent no-op on sync BSP.

## Suggested resolution
Either (or both):
1. Document clearly that `max_concurrent_exports` / `OTEL_BSP_MAX_CONCURRENT_EXPORTS` is async-runtime BSP only.
2. In sync BSP, reject or warn (or clamp with warning) when `max_concurrent_exports > 1` instead of silently accepting it.

glaziermag:
[...]

PR #3403, opened 2026-03-03
docs(sdk): clarify BSP max_concurrent_exports scope for thread-based processor

glaziermag:
## Summary
Clarifies `max_concurrent_exports` / `OTEL_BSP_MAX_CONCURRENT_EXPORTS` scope to match current BSP behavior after the thread-based redesign.

## Changes
- Clarified crate docs (`opentelemetry-sdk/src/lib.rs`) that `OTEL_BSP_MAX_CONCURRENT_EXPORTS` is honored by `span_processor_with_async_runtime::BatchSpanProcessor`.
- Updated `span_processor.rs` docs for:
  - `OTEL_BSP_MAX_CONCURRENT_EXPORTS`
  - `BatchConfig::max_concurrent_exports`
  - `BatchConfigBuilder::with_max_concurrent_exports`
  so they consistently state the thread-based `BatchSpanProcessor` exports serially and that concurrent exports require the async-runtime processor + feature flag.
- Added a sync BSP regression test:
  - `batchspanprocessor_sync_ignores_max_concurrent_exports`
  - Verifies no overlap occurs and confirms exports actually ran (`export_calls == 3`).
- Added `max_concurrent_exports` field to startup debug telemetry context in thread-based BSP.

## Why
This addresses a docs/config-contract mismatch: the sync/thread-based BSP intentionally exports serially, while docs/config surface could imply concurrent-export tuning is effective there.

Attempts to address #3402.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ permissions:
   contents: read
 on:
   pull_request:
+  merge_group:
   push:
     branches:
     - main
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -8,6 +8,7 @@ permissions:
 
 on:
   pull_request:
+  merge_group:
   push:
     branches: [main]
   workflow_dispatch:
--- a/.github/workflows/markdown-link-check.yml
+++ b/.github/workflows/markdown-link-check.yml
@@ -2,6 +2,7 @@ name: Markdown link check
 
 on:
   pull_request:
+  merge_group:
   push:
     branches:
     - main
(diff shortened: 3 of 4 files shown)


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -67,6 +67,13 @@ pub(crate) const OTEL_BSP_EXPORT_TIMEOUT: &str = "OTEL_BSP_EXPORT_TIMEOUT";
 pub(crate) const OTEL_BSP_EXPORT_TIMEOUT_DEFAULT: Duration = Duration::from_millis(30_000);
 /// Environment variable to configure max concurrent exports for batch span
 /// processor.
+///
```
cijothomas: this won't show up anywhere right? this is not a pubic const

PR #3409, opened 2026-03-06
fix: `service.name` fallback

lahabana:
## Changes

The fallback introduced in #3302 doesn't work properly as the fallback didn't
read from `process.executable.name` as indicated in the [spec](https://opentelemetry.io/docs/specs/semconv/resource/service/#service).

We now read from the resource when it exists, we also do the fallback when building the Resource so that any possible override is done before falling back

I added a flag so that `builder_empty` still returns a properly empty resource but I'm happy to discuss other options.

        ^ Doesn't need updating because the current changelog item is correct.

### DIFF
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -482,9 +482,9 @@ mod tests {
                 .get(&Key::from_static_str(SERVICE_NAME))
                 .map(|v| v.to_string())
                 .unwrap();
-            assert!(
-                service_name.starts_with("unknown_service:opentelemetry_sdk-"),
-                "Expected service name to start with 'unknown_service:opentelemetry_sdk-', got: {}",
+            assert_eq!(
+                service_name, "unknown_service",
+                "Expected service name to be 'unknown_service', got: {}",
                 service_name
             );
             assert_telemetry_resource(&processor_with_resource, &exporter_with_resource);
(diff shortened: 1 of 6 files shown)


cijothomas:
> @cijothomas I'd love to hear what you think about this :)

> The fallback introduced in https://github.com/open-telemetry/opentelemetry-rust/pull/3302 doesn't work properly as the fallback didn't
> read from process.executable.name as indicated in the [spec](https://opentelemetry.io/docs/specs/semconv/resource/service/#service).

Can you elaborate more why the current behavior is not compliant with the spec? 

"fallback didn't read from process.executable.name" - If you meant an attribute named "process.executable.name", then we are not expected to do that! Process env variables are experimental, and SDKs won't take a dependency on experimental conventions. Even today, spec is not asking to pull value from that attribute. (We could modify the doc to make it more obvious)

cijothomas (changes requested):
Not required. The current behavior is compliant with the spec already.

cijothomas:
@lahabana I think there is some confusion around "process.executable.name" part. Maybe it's something we can help fix in the spec/sem-conventions repo. Let me know if you think this PR is still required. Happy to discuss more.

cijothomas:
> @lahabana I think there is some confusion around "process.executable.name" part. Maybe it's something we can help fix in the spec/sem-conventions repo. Let me know if you think this PR is still required. Happy to discuss more.

https://github.com/open-telemetry/semantic-conventions/pull/3637 Opened this in upstream semantic conventions repo.

lahabana:
Thanks this clarifies this significantly in the docs. I'm closing this PR as you are right this isn't the intended behaviour! Thanks for taking the time.

Issue #2939, opened 2025-04-22
[Feature]: Reading spans through the public API requires cloning everything to `SpanData`


paullegranddc:
### Describe the solution you'd like:

When creating a new span, the tracer calls `SpanExporter::on_start` method and passes it a `&mut Span`.
In this spec, this span parameter is described as taking a being read/write [opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onstart).

Currently, the only way to read data on this value is to call `span.exported_data() -> Option<crate::trace::SpanData>`, which clones all fields in the span, including things that require allocation such as the span links attributes, events and span links.

Other opentelemetry SDK implementation usually expose a ReadableSpan interface
[go](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L33) [java](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/ReadableSpan.java#L19) providing a view over the span without mutations, and I believe something similar could be implemented over the rust Span.

This is the API I am proposing:

```rust
impl Span {
    /// Returns a read-only view of the span data
    /// This is None if the span is not recording.
    pub fn read<'a>(&'a self) -> Option<ReadableSpan<'a>> {
        Some(ReadableSpan(self.data.as_ref()?))
    }
}

#[derive(Clone, Debug)]
pub struct ReadableSpan<'a>(&'a SpanData);

impl<'a> ReadableSpan<'a> {
    pub fn parent_span_id(&self) -> SpanId {..}
    pub fn span_kind(&self) -> SpanKind {..}

    pub fn name(&self) -> &str {..}

    pub fn start_time(&self) -> SystemTime {..}

    pub fn end_time(&self) -> SystemTime {..}

    pub fn attributes(&self) -> &'a [KeyValue] {..}

    pub fn dropped_attributes_count(&self) -> u32 {..}

    pub fn events(&self) -> &'a [Event] {..}

    pub fn dropped_events_count(&self) -> u32 {..}

    pub fn links(&self) -> &'a [Link] {..}

    pub fn dropped_links_count(&self) -> u32 {..}

    pub fn status(&self) -> &Status {..}
}
```

[POC here](https://github.com/open-telemetry/opentelemetry-rust/compare/main...paullegranddc:opentelemetry-rust:paullgdc/sdk/readable_span?expand=1) 

[...]

PR #3410, opened 2026-03-06
refactor(sdk): change SpanProcessor::on_end to take &mut FinishedSpan

### LINKED ISSUE #2726: Tracing SDK - Processors get clone of SpanData which is incorrect

bryantbiggs:
## Summary

Supersedes #2962 (original work by @paullegranddc). Refactors `SpanProcessor::on_end` from `fn on_end(&self, span: SpanData)` to `fn on_end(&self, span: &mut FinishedSpan)`, enabling zero-copy span inspection and ownership semantics for multi-processor chains.

### Breaking Change

`SpanProcessor::on_end` now takes `&mut FinishedSpan` instead of `SpanData`.

### Key Changes

- **`ReadableSpan` trait**: New trait implemented by both `Span` (live, for `on_start`) and `FinishedSpan` (ended, for `on_end`). Allows processors to inspect span fields (`name()`, `attributes()`, `span_kind()`, `instrumentation_scope()`, etc.) without cloning.
- **`FinishedSpan` type**: Wraps finished span data with consume semantics. `consume()` gives the last processor zero-copy ownership via `take()`; earlier processors receive a `clone()`. `is_consumed()` lets processors check post-consume state.
- **`end_time_set` flag**: Replaces the fragile `start_time == end_time` sentinel comparison. `end_with_timestamp()` now correctly preserves user-provided timestamps even when equal to start time.
- **Zero-copy `span_context`**: Uses `mem::replace` instead of `clone()` in `end_and_export` to avoid heap allocation for `TraceState`.
- **`end_and_export` rewrite**: Takes span data before shutdown check (fixes potential data leak on re-entrant Drop), sets end time via explicit flag.

### Migration Guide

```rust
// Before
impl SpanProcessor for MyProcessor {
    fn on_end(&self, span: SpanData) {
        export(span);
    }
}

// After — read without cloning
impl SpanProcessor for MyProcessor {
    fn on_end(&self, span: &mut FinishedSpan) {
        if span.name() == Some("skip_me") { return; }
        if let Some(data) = span.consume() {
            export(data);
        }
    }
}
```

### Files Changed

- `opentelemetry-sdk/src/trace/span.rs` — `FinishedSpan`, `ReadableSpan` trait, `end_and_export` rewrite, `end_time_set` flag
- `opentelemetry-sdk/src/trace/span_processor.rs` — Updated `SpanProcessor` trait + built-in impls
- `opentelemetry-sdk/src/trace/span_processor_with_async_runtime.rs` — Async batch processor updated
- `opentelemetry-sdk/src/trace/tracer.rs` — `end_time_set: false` in span construction
- `examples/tracing-http-propagator/src/server.rs` — New `RouteConcurrencyCounterSpanProcessor` example demonstrating `ReadableSpan` in both `on_start` and `on_end`
- `opentelemetry-sdk/benches/span_processor_api.rs` — New benchmark (0/1/2/4 processors, ~191 ns uniform on M4 Max)
- `stress/src/traces.rs` — Updated `NoOpSpanProcessor`

### Benchmark Results (Apple M4 Max)

| Processors | Time |
|---|---|
| 0 | 190.82 ns |
| 1 | 192.53 ns |
| 2 | 191.67 ns |
| 4 | 193.48 ns |

Uniform across processor counts — the new API adds negligible overhead.

## Test Plan

- [x] All existing span/span_processor tests pass (47 lib tests + 38 doc tests)
- [x] New tests: `test_readable_span_after_consume_returns_defaults`, `test_end_with_timestamp_equal_to_start_time`
- [x] clippy clean, fmt clean
- [x] Example (`tracing-http-propagator`) and stress test compile
- [x] Async runtime (`rt-tokio`) compiles
- [x] Benchmarks compile and run

Closes #2726, #2939. Relates to #2940.

### DIFF
--- a/opentelemetry-sdk/src/trace/span.rs
+++ b/opentelemetry-sdk/src/trace/span.rs
@@ -37,6 +37,8 @@ pub(crate) struct SpanData {
     pub(crate) start_time: SystemTime,
     /// Span end time
     pub(crate) end_time: SystemTime,
+    /// Whether end_time was explicitly set via `end_with_timestamp`
+    pub(crate) end_time_set: bool,
     /// Span attributes
     pub(crate) attributes: Vec<KeyValue>,
     /// The number of attributes that were above the configured limit, and thus
@@ -194,38 +196,45 @@ impl opentelemetry::trace::Span for Span {
 
     /// Finishes the span with given timestamp.
     fn end_with_timestamp(&mut self, timestamp: SystemTime) {
-        if let Some(data) = self.data.as_mut() {
+        if let Some(ref mut data) = self.data {
             data.end_time = timestamp;
+            data.end_time_set = true;
         }
         self.end_and_export();
     }
 }
 
 impl Span {
-    /// Span ending logic
+    /// Exports the span to all registered processors.
     ///
-    /// The end timestamp of the span has to be set before calling this function
+    /// Takes ownership of span data, sets end time if not already set,
+    /// and passes the finished span to each processor.
     fn end_and_export(&mut self) {
-        if self.tracer.provider().is_shutdown() {
+        // Take data first so Drop won't re-enter if provider is shut down
+        let mut data = match self.data.take() {
+            Some(data) => data,
+            None => return,
+        };
+
+        let provider = self.tracer.provider();
+        if provider.is_shutdown() {
             return;
         }
 
-        let Span {
-            data,
-            tracer,
-            span_context,
-            ..
-        } = self;
+        // Set end time to now if not explicitly set via end_with_timestamp
+        if !data.end_time_set {
+            data.end_time = opentelemetry::time::now();
+        }
 
-        // Grab the span data from the span and leave it empty
-        // This way we don't have to clone the data
-        let Some(data) = data.take() else { return };
-        let span_context: SpanContext =
-            std::mem::replace(span_context, SpanContext::empty_context());
+        // Move span_context out of the span. After data.take() above, the span
+        // is effectively dead (all further operations no-op), so replacing the
+        // context with an empty one avoids cloning the TraceState.
+        let span_context = std::mem::replace(&mut self.span_context, SpanContext::empty_context());
 
-        let mut finished_span = FinishedSpan::new(build_export_data(data, span_context, tracer));
+        let mut finished_span =
+            FinishedSpan::new(build_export_data(data, span_context, &self.tracer));
 
-        let span_processors = tracer.provider().span_processors();
+        let span_processors = provider.span_processors();
         for (i, processor) in span_processors.iter().enumerate() {
             finished_span.reset(i == span_processors.len() - 1);
             processor.on_end(&mut finished_span);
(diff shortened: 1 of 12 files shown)


opentelemetry-sdk/src/trace/span.rs
```diff
@@ -194,60 +196,394 @@ impl opentelemetry::trace::Span for Span {
...
+        Some(&self.data.as_ref()?.name)
+    }
+
+    fn start_time(&self) -> Option<SystemTime> {
+        self.data.as_ref().map(|data| data.start_time)
+    }
+
+    fn end_time(&self) -> Option<SystemTime> {
+        self.data.as_ref().map(|data| data.end_time)
+    }
```
bantonsson, line 281: Shouldn't this return `None` in a number of scenarios?

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -729,4 +1118,147 @@ mod tests {
...
+            }
+        }
+
+        let provider = crate::trace::SdkTracerProvider::builder()
+            .with_span_processor(TestSpanProcessor)
+            .with_span_processor(TestSpanProcessor)
+            .build();
+        drop(make_test_span(&provider.tracer("test")));
+        let res = provider.shutdown();
+        println!("{:?}", res);
```
bantonsson: leftover?

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -194,60 +196,394 @@ impl opentelemetry::trace::Span for Span {
...
+    /// Returns the `SpanContext` of the span.
+    fn context(&self) -> &SpanContext;
+
+    /// Returns the `SpanId` of the parent span.
+    fn parent_span_id(&self) -> SpanId;
+
+    /// Returns the `SpanKind` of the span.
+    ///
+    /// Returns `SpanKind::Internal` if the span is not recording.
+    fn span_kind(&self) -> SpanKind;
```
bantonsson: Nit: It feels like there is an inconsistency between `SpanKind` and `&Status`.

bantonsson (changes requested):
Thanks for moving this forward @bryantbiggs. Only some minor things.

opentelemetry-sdk/src/trace/span.rs
bryantbiggs: Good question! Since this is the `FinishedSpan` impl, the span has always been ended by the time you have one — `end_and_export` sets the end time before constructing the `FinishedSpan`. So the only way `data` is `None` here is after `consume()` has been called, which `data.as_ref()` already handles by returning `None`.

The trait-level doc on `end_time()` mentions returning `None` when "the span has not been ended" — but that case applies to the `Span` (live) impl, not `FinishedSpan`. Happy to add an inline comment here clarifying that if it would help!

opentelemetry-sdk/src/trace/span.rs
bryantbiggs: Yep, leftover debug print — removed. There was a second one a bit further down in `test_finished_span_consume_twice` from the same original commit, so I removed that one too.

opentelemetry-sdk/src/trace/span.rs
bryantbiggs: Good catch — changed `span_kind()` to return `&SpanKind` so it matches `status() -> &Status`. This also avoids the `.clone()` on `SpanKind` in the implementations, which is a small win since `SpanKind` doesn't derive `Copy`.

opentelemetry-sdk/src/trace/span.rs
bantonsson: Thanks for the explanation. I think it's fine to leave it as is.

bantonsson (approved):

bantonsson:
Can we get some maintainer 👀 on this?
/cc @cijothomas @hdost @lalitb @utpilla @TommyCpp

bantonsson:
Any of the maintainers have some time for this?
/cc @cijothomas @hdost @lalitb @utpilla @TommyCpp

cijothomas:
Thanks for the PR! Let's hold off on this one until after the current release ships — that way we can look at it together with the other pending tracing changes and batch all the breaking changes into the next release. Will pick it up right after — appreciate the patience!

PR #3413, opened 2026-03-11
chore: (For Discussion): Dynamic runtime detection & `Joinable` handle abstraction

### LINKED ISSUE #2643: Stabilize support for async runtimes

scottgerring:
Fixes #2643, stabilising Runtime trait.

Builds on all of the dynamic runtime detection changes from #3412 and goes further by introducing a `Joinable<T>` type - a unified join handle that wraps either a `std::thread::JoinHandle` or a `tokio::task::JoinHandle`.

Key changes:
- `Runtime::spawn` now returns `Joinable<T>` instead of `()`, capturing the handle to the spawned work
- `Joinable::join()` blocks until the task completes — using `thread::join()` for OS threads or `futures_executor::block_on(handle)` for Tokio tasks
- Removes `block_on` from the `Runtime` trait (no longer needed since shutdown can just `join()` the spawned task)
- Processors use the returned `Joinable` to **directly join on their background task at shutdown**, replacing the current `oneshot` channel pattern

This is a deeper refactor that simplifies shutdown logic across all three signal processors (traces, metrics, logs) by making the spawn-and-join lifecycle explicit rather than relying on side-channel signalling.

**Other alternatives?**
We could look at removing the processors that are explicitly threaded (e.g. `BatchLogProcessor`, `BatchSpanProcessor`) in favour of the `NoAsync` runtime and the 'regular' processors, but I think this is something that would happen downstream to this approach shaking out well .


opentelemetry-sdk/src/runtime.rs
```diff
@@ -6,9 +6,86 @@
...
+/// to provide a uniform interface for waiting on spawned work.
+#[cfg(feature = "experimental_async_runtime")]
+pub struct Joinable<T>(JoinableInner<T>);
+
+#[cfg(feature = "experimental_async_runtime")]
+enum JoinableInner<T> {
+    Thread(std::thread::JoinHandle<T>),
+    #[cfg(feature = "rt-tokio")]
+    TokioTask(tokio::task::JoinHandle<T>),
+}
```
lalitb, line 25: Wondering if this should be associated type instead of enum, for extensibility - to allow users to bring their own `Runtime` implementation.

opentelemetry-sdk/src/runtime.rs
```diff
@@ -23,18 +100,17 @@ use thiserror::Error;
...
-    /// finish when the caller shuts down.
+    /// # Note
     ///
-    /// At the moment, the shutdown happens by blocking the
-    /// current thread. This means runtime implementations need to make sure they can still execute
-    /// the given future even if the main thread is blocked.
-    fn spawn<F>(&self, future: F)
+    /// This is mainly used to run batch span processing in the background. The mechanism used
+    /// to provide the spawn may be relatively heavyweight.
+    fn spawn<F, T>(&self, future: F) -> Joinable<T>
```
lalitb: Should spawn return an associated handle type instead of `Joinable<T>`? As written, the concrete enum makes the public Runtime trait effectively non-extensible for external runtimes.

scottgerring: Good catch - I think we should decide if we want public runtime extensibility or not - probably actually yes?

opentelemetry-sdk/src/runtime.rs
scottgerring: I've set this up in this commit --> https://github.com/open-telemetry/opentelemetry-rust/pull/3413/changes/133ae038c7a39b929defa5169fa82d96bd07e20d 

I've added a public `JoinHandle<T>` trait. `Runtime::spawn` now returns an associated type `SpawnHandle<T>` bounded on `JoinHandle<T>`, so external runtime implementors can bring their own handle type rather than being coupled to the built-in `Joinable` enum. 

`Joinable` itself is still there as the concrete impl used by `Tokio` and `NoAsync`, but it's `#[doc(hidden)]` (can't actually make it `pub(crate)` it turns out). 

 I've also kept `spawn` generic over `T` rather than fixing `Output = ()`, since the flexibility is easy enough now and avoids a future breaking change if callers ever want to retrieve a result from a spawned task.

opentelemetry-sdk/src/runtime.rs
scottgerring: See comment above - i've added this to the change to maintain extensibility.

PR #3414, opened 2026-03-11
fix(otlp): use Endpoint::from_shared so unix sockets are supported

bouk:
Workaround for https://github.com/hyperium/tonic/issues/2546, currently you can't do `OTEL_EXPORTER_OTLP_ENDPOINT=unix:///tmp/test` because you get the error: `invalid URI unix:///tmp/test. Reason invalid format`

## Changes

Switch to `Endpoint::from_shared` which also returns an `Endpoint` but it supports `unix://` endpoints

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -730,10 +730,12 @@ Follow these guidelines when deciding where to attach metric attributes:
 
     ```rust
     // Example: Setting resource-level attributes
-    let resource = Resource::new(vec![
-        KeyValue::new("service.name", "payment-processor"),
-        KeyValue::new("deployment.environment", "production"),
-    ]);
+    // Use Resource::builder() to preserve SDK-provided defaults
+    // (telemetry.sdk.*, service.name).
+    let resource = Resource::builder()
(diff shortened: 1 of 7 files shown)


opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -1054,4 +1054,21 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_unix_socket_endpoint_succeeds() {
+        use crate::SpanExporter;
+        use crate::WithExportConfig;
+
+        let result = SpanExporter::builder()
+            .with_tonic()
+            .with_endpoint("unix:///tmp/test")
+            .build();
```
lalitb, line 1066: Non-blocking: the added unit test covers the parsing/building regression, but it would be useful to add a Unix-only smoke test in opentelemetry-otlp/tests/smoke.rs that runs a tonic server over UnixListenerStream and verifies an actual export over UDS. That feels reasonable as a separate PR.

PR #3420, opened 2026-03-14
refactor(metrics): replace swap-and-drain delta collection with in-place iteration

bryantbiggs:
Splits from #3392, per [maintainer feedback](https://github.com/open-telemetry/opentelemetry-rust/pull/3392#pullrequestreview-3949240401). Refs #2328.

## Summary

- Replaces the two-HashMap swap-and-drain pattern in `collect_and_reset` with in-place iteration using `TrackerEntry` status tracking
- `TrackerEntry` wraps each aggregator with `has_been_updated` (AtomicBool) to track whether measurements occurred since the last collection cycle
- Removes `has_no_attribute_value` — the `no_attribute_tracker.has_been_updated` flag now serves the same purpose
- `collect_and_reset` now iterates under a **read lock** (no write lock in steady state), exports only updated entries, and evicts stale entries under a write lock with TOCTOU re-check
- Adds `drain_and_reset` for Observable/async instruments that need true staleness detection via map clearing
- Eliminates O(n) write-lock acquisitions on the hot path per collection cycle when attribute sets are reused (the common case)

## User-facing behavior change

Recovery from cardinality overflow now requires **2 collect cycles** instead of 1 — the first cycle marks entries as stale (no updates), the second cycle evicts them. This trade-off enables the in-place iteration pattern that avoids write-lock contention in steady state.

## Locking trade-offs (discussed with @cijothomas and @utpilla in #3392)

| Scenario | Old (swap-and-drain) | New (in-place) |
|----------|---------------------|----------------|
| Steady state (reused attrs) | O(n) write locks per cycle | Read locks only |
| New/returning attrs during collect | Write lock for insert | Write lock for insert (competes with read lock held by collect) |
| Stale eviction | Implicit (map cleared) | Write lock proportional to stale count |

## Test plan

- [x] All 149 existing metrics tests pass
- [x] `counter_aggregation_overflow_delta` tests updated for two-phase eviction semantics
- [x] No changes to cumulative collection path (`collect_readonly`)
- [x] `cargo clippy -p opentelemetry_sdk --features metrics -- -D warnings` passes

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ## vNext
 
-- Refactored delta metrics collection to use in-place iteration instead of
-  swap-and-drain. Entries are now tracked via a `has_been_updated` flag and only
-  exported when updated since the last collection cycle. Stale unbound entries are
-  evicted under a write lock with TOCTOU re-check. This eliminates O(n) write-lock
-  acquisitions on the hot path per collection cycle in steady state.
+- Delta metrics collection now uses in-place eviction instead of draining the
+  HashMap on every collect cycle. Stale attribute sets that received no measurements
+  since the last collection are evicted. Note: recovery from cardinality overflow
+  now requires 2 collect cycles — the first marks entries as stale, the second
+  evicts them.
 - **Breaking** The SDK `testing` feature is now runtime agnostic. [#3407][3407]
   - `TokioSpanExporter` and `new_tokio_test_exporter` have been renamed to `TestSpanExporter` and `new_test_exporter`.
   - The following transitive dependencies and features have been removed: `tokio/rt`, `tokio/time`, `tokio/macros`, `tokio/rt-multi-thread`, `tokio-stream`, `experimental_async_runtime`
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- Refactored delta metrics collection to use in-place iteration instead of
+  swap-and-drain. Entries are now tracked via a `has_been_updated` flag and only
+  exported when updated since the last collection cycle. Stale unbound entries are
+  evicted under a write lock with TOCTOU re-check. This eliminates O(n) write-lock
+  acquisitions on the hot path per collection cycle in steady state.
 - **Breaking** The SDK `testing` feature is now runtime agnostic. [#3407][3407]
   - `TokioSpanExporter` and `new_tokio_test_exporter` have been renamed to `TestSpanExporter` and `new_test_exporter`.
   - The following transitive dependencies and features have been removed: `tokio/rt`, `tokio/time`, `tokio/macros`, `tokio/rt-multi-thread`, `tokio-stream`, `experimental_async_runtime`
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -51,19 +51,14 @@ pub(crate) trait Aggregator {
     fn clone_and_reset(&self, init: &Self::InitConfig) -> Self;
 }
 
-/// Wraps an aggregator with status tracking for delta collection and bound instruments.
+/// Wraps an aggregator with status tracking for delta collection.
 ///
 /// `has_been_updated` tracks whether the aggregator received measurements since the last
 /// collection cycle. This enables in-place delta collection: only updated entries are exported,
-/// and stale unbound entries are evicted to prevent unbounded memory growth.
-///
-/// `bound_count` tracks how many bound instrument handles reference this entry. Entries with
-/// bound_count > 0 are never evicted from the map, even if they had no updates in a cycle
-/// (they simply produce no export). This ensures bound handles always point to a live tracker.
+/// and stale entries are evicted to prevent unbounded memory growth.
 pub(crate) struct TrackerEntry<A: Aggregator> {
     pub(crate) aggregator: A,
     pub(crate) has_been_updated: AtomicBool,
-    pub(crate) bound_count: AtomicUsize,
 }
 
 impl<A: Aggregator> TrackerEntry<A> {
@@ -71,7 +66,6 @@ impl<A: Aggregator> TrackerEntry<A> {
         TrackerEntry {
             aggregator: A::create(config),
             has_been_updated: AtomicBool::new(false),
-            bound_count: AtomicUsize::new(0),
         }
     }
 }
@@ -220,9 +214,7 @@ where
 
     /// Iterate through all attribute sets in-place, populate `DataPoints` and reset.
     /// Only entries updated since the last collection (tracked via `has_been_updated`)
-    /// are exported. Stale unbound entries are evicted to prevent unbounded memory growth.
-    /// Bound entries (bound_count > 0) are never evicted — they persist until explicitly
-    /// unbound, but produce no export when they have no updates.
+    /// are exported. Stale entries are evicted to prevent unbounded memory growth.
     ///
     /// Used for synchronous instruments (Counter, Histogram, etc.) in Delta temporality mode.
     pub(crate) fn collect_and_reset<Res, MapFn>(&self, dest: &mut Vec<Res>, mut map_fn: MapFn)
@@ -252,10 +244,8 @@ where
                 if seen.insert(Arc::as_ptr(tracker)) {
                     if tracker.has_been_updated.swap(false, Ordering::Relaxed) {
                         dest.push(map_fn(attrs.clone(), &tracker.aggregator));
-                    } else if attrs.as_slice() != overflow_attrs.as_slice()
-                        && tracker.bound_count.load(Ordering::Relaxed) == 0
-                    {
-                        // Stale and not bound — candidate for eviction
+                    } else if attrs.as_slice() != overflow_attrs.as_slice() {
+                        // Stale — candidate for eviction
                         stale_entries.push(Arc::clone(tracker));
                     }
                 }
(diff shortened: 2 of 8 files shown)


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -52,6 +51,34 @@ pub(crate) trait Aggregator {
     fn clone_and_reset(&self, init: &Self::InitConfig) -> Self;
 }
 
+/// Wraps an aggregator with status tracking for delta collection and bound instruments.
+///
+/// `has_been_updated` tracks whether the aggregator received measurements since the last
+/// collection cycle. This enables in-place delta collection: only updated entries are exported,
+/// and stale unbound entries are evicted to prevent unbounded memory growth.
+///
+/// `bound_count` tracks how many bound instrument handles reference this entry. Entries with
```
cijothomas: lets add it in the PR adding bound_count, not in this one.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -83,37 +105,34 @@ impl<A> ValueMap<A>
...
-
     /// Checks whether aggregator has hit cardinality limit for metric streams
     fn is_under_cardinality_limit(&self) -> bool {
         self.count.load(Ordering::SeqCst) < self.cardinality_limit
     }
 
     fn measure(&self, value: A::PreComputedValue, attributes: &[KeyValue]) {
         if attributes.is_empty() {
-            self.no_attribute_tracker.update(value);
+            self.no_attribute_tracker.aggregator.update(value);
```
cijothomas, line 123: the `self.has_no_attribute_value` in the next line can now be removed I guess, given we are tracking it for all entries anyway via the new has_been_updated

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Done — removed `bound_count` from this PR. It's now added in #3421 alongside the bound instruments feature.

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- Refactored delta metrics collection to use in-place iteration instead of
```
cijothomas: this is more like internal details, not end-user friendly changelog. The change that is really worth mentioning is the requirement of 2 collect cycles to cleanup entries - if a user hit cardinality capping, and then they'd need to wait for 2nd cycle to free up space and start reporting non-overflow.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Good call — removed `has_no_attribute_value` entirely. The `no_attribute_tracker.has_been_updated` flag now serves the same purpose across all three collection paths (`collect_readonly`, `collect_and_reset`, `drain_and_reset`).

opentelemetry-sdk/CHANGELOG.md
bryantbiggs: Rewritten to focus on the user-facing behavior change: cardinality overflow recovery now requires 2 collect cycles instead of 1.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -185,47 +207,107 @@ where
...
+
+        if !stale_entries.is_empty() {
+            if let Ok(mut trackers) = self.trackers.write() {
+                // Re-check under write lock to avoid TOCTOU race: a measure() call between
+                // dropping the read lock and acquiring the write lock could have updated
+                // an entry we marked as stale.
+                stale_entries.retain(|entry| !entry.has_been_updated.load(Ordering::Relaxed));
+
+                if !stale_entries.is_empty() {
+                    let stale_pointers: HashSet<*const TrackerEntry<A>> =
```
cijothomas, line 263: nicely handled. we probably need to add a test to target this specifically - when removing a stale entry, we remove the entry+its sorted entry, if exists.

bryantbiggs: Added in 1f5fd4cd — inserts attrs in non-sorted order so both HashMap keys get created, verifies `trackers.len() == 2` before eviction, then confirms both drop to 0 after the second collect cycle.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -185,47 +207,107 @@ where
...
+
+        {
+            let Ok(trackers) = self.trackers.read() else {
+                return;
+            };
+
+            let mut seen = HashSet::new();
+            for (attrs, tracker) in trackers.iter() {
+                if seen.insert(Arc::as_ptr(tracker)) {
+                    if tracker.has_been_updated.swap(false, Ordering::Relaxed) {
```
lalitb: There is still a race in the in-place delta path I think - `collect_and_reset()` clears has_been_updated before the aggregator is actually read/reset via the callback. So if a concurrent `measure()` sneaks in between, its value gets consumed by the current `get_and_reset_value()`/`clone_and_reset()`, but it also flips `has_been_updated` back to true. Next collection cycle then sees an "updated" entry with an already-reset aggregator and emits a zero/empty datapoint.

The old swap-and-drain didn't have this problem since updates after the swap just went into the next cycle's map.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -185,47 +207,107 @@ where
...
+    /// Used for synchronous instruments (Counter, Histogram, etc.) in Delta temporality mode.
     pub(crate) fn collect_and_reset<Res, MapFn>(&self, dest: &mut Vec<Res>, mut map_fn: MapFn)
+    where
+        MapFn: FnMut(Vec<KeyValue>, &A) -> Res,
+    {
+        prepare_data(dest, self.count.load(Ordering::SeqCst));
+        if self
+            .no_attribute_tracker
+            .has_been_updated
+            .swap(false, Ordering::AcqRel)
```
lalitb, line 228: [Same](https://github.com/open-telemetry/opentelemetry-rust/pull/3420/changes#r2972629932) pattern exists here too.

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: Thanks Lalit for the review. Agree with the assessment — the race is real. A few notes:

1. This same race already exists today on `main` for the `no_attribute_tracker`. This PR extends the pattern to all tracked attribute sets.
2. I had put the same comment in [OTel .NET SDK](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MetricPoint/MetricPoint.cs#L1062-L1077) — It suffers the exact same problem, and we couldn't fix it in 5 years🤣 

The core issue is that the flag and the aggregator value are two separate atomics — each operation is atomic individually, but together they aren't, and making them atomic as a pair would require heavier locking that defeats the purpose of this optimization. The worst case is a spurious zero/empty data point in a subsequent delta cycle, with no data loss. 

From my time at .NET, it was never prioritized as the impact of an extra 0 value point was negligible for SUM and HISTOGRAM. It is more serious issue in Gauge - as we export a gauge value of 0, and it is incorrect (Gauge instrument was added much later to OTel world)

I am willing to accept this edge case for the overall gains this PR brings. What do you think? 

Tagging @utpilla also for his thoughts. IIRC, we may have discussed this during .NET metrics implementation time..

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/pull/2629/changes#r751506721 Maybe it was me and Reiley who discussed this. Since OTel didn't have sync. Gauge at the time, the impact was negligible! But Gauge issue is more serious.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -124,14 +133,16 @@ where
 
         // Try to retrieve and update the tracker with the attributes in the provided order first
         if let Some(tracker) = trackers.get(attributes) {
-            tracker.update(value);
+            tracker.aggregator.update(value);
+            tracker.has_been_updated.store(true, Ordering::Relaxed);
```
cijothomas, line 137: for the no_attribute_tracker, we correctly use the `Release` ordering, but this one uses `Relaxed`, potentially losing an update. Shouldn't we use `Release` here too?

bryantbiggs: Good catch — you're right, this should be `Release`. Fixed all the per-attribute stores to use `Release`, and updated the collector-side swap and TOCTOU re-check load to use `Acquire` to match.

Dug into this a bit: the reason `Relaxed` is wrong here is that the producer and collector can both hold **read** locks on the `RwLock` at the same time, and two concurrent readers don't synchronize with each other. So the flag is the only ordering point between `aggregator.update()` and the collector reading the aggregator. Without `Release`/`Acquire`, on weakly-ordered architectures (ARM) the collector could see the flag as `true` but read a stale aggregator value — the `fetch_add` from the producer hasn't become visible yet.

For the Mutex-based aggregators (Histogram, ExponentialHistogram) the Mutex itself provides the ordering, so `Relaxed` would technically be fine there. But since `ValueMap` is generic over the aggregator type, using `Release`/`Acquire` everywhere is the simple and correct approach. Zero cost on x86 (compiles to the same instructions), negligible on ARM.

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/pull/3420#discussion_r2972629932 Using this to discuss.

cijothomas (approved):
Thanks!

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: https://github.com/open-telemetry/opentelemetry-rust/issues/3470 opened to track fixing Gauge in future.

PR #3421, opened 2026-03-14
feat(metrics): add bound instruments behind experimental feature flag

bryantbiggs:
Splits from #3392, per [maintainer feedback](https://github.com/open-telemetry/opentelemetry-rust/pull/3392#pullrequestreview-3949240401). Built on top of #3420 (in-place delta collection, now merged). Refs #1374.

## Summary

Adds `BoundCounter` and `BoundHistogram` to the public API behind the `experimental_metrics_bound_instruments` feature flag. These types cache the resolved aggregator reference for a fixed attribute set, allowing subsequent measurements to bypass per-call sort, dedup, hash, and HashMap lookup entirely.

```rust
let counter = meter.u64_counter("requests").build();
let bound = counter.bind(&[KeyValue::new("method", "GET")]);
bound.add(1); // ~1.9ns — no attribute lookup
```

### TrackerEntry bound_count

`bound_count` (introduced in #3420) tracks how many live `BoundCounter`/`BoundHistogram` handles reference an entry. Entries with `bound_count > 0` are never evicted during delta collection.

### Cardinality overflow handling

`ValueMap::bind()` returns `Option>` — `None` when at the cardinality limit. Bound handle types use a `Direct`/`Fallback` enum:

- **Direct**: normal case — dedicated tracker, single-digit ns, no map lookup per call
- **Fallback**: at overflow — delegates every call to the unbound `Measure::call()` path

### Feature flag

All public API, traits, noop impls, and internal plumbing are gated behind `experimental_metrics_bound_instruments`.

## Benchmark Results

Apple M4 Max, 16 cores (12 performance + 4 efficiency), macOS 15.4:

| Benchmark | Time | vs Unbound |
|-----------|------|------------|
| Counter Unbound | 53.2 ns | — |
| **Counter Bound** | **1.87 ns** | **~28x faster** |
| Histogram Unbound | 58.6 ns | — |
| **Histogram Bound** | **6.57 ns** | **~8.9x faster** |

<details>
<summary>EC2 Counter Results (from #3392)</summary>

| Instance | Unbound | Bound | Speedup |
|----------|---------|-------|---------|
| c7i.large | 86.2ns | 7.47ns | 11.5x |
| c7i.xlarge | 88.9ns | 7.62ns | 11.7x |
| c7i.4xlarge | 90.1ns | 7.65ns | 11.8x |
| c7a.large | 77.4ns | 2.71ns | 28.6x |
| c7a.xlarge | 75.0ns | 2.71ns | 27.7x |
| c7a.4xlarge | 75.0ns | 2.71ns | 27.7x |
| c7g.large | 111.8ns | 6.87ns | 16.3x |
| c7g.xlarge | 111.4ns | 6.84ns | 16.3x |
| c7g.4xlarge | 112.9ns | 6.73ns | 16.8x |
</details>

<details>
<summary>EC2 Histogram Results (from #3392)</summary>

| Instance | Unbound | Bound | Speedup |
|----------|---------|-------|---------|
| c7i.large | 97.1ns | 17.1ns | 5.7x |
| c7i.xlarge | 99.0ns | 17.9ns | 5.5x |
| c7i.4xlarge | 98.5ns | 18.4ns | 5.4x |
| c7a.large | 92.5ns | 13.7ns | 6.8x |
| c7a.xlarge | 92.4ns | 13.7ns | 6.7x |
| c7a.4xlarge | 92.5ns | 13.7ns | 6.8x |
| c7g.large | 133.2ns | 27.7ns | 4.8x |
| c7g.xlarge | 133.5ns | 27.6ns | 4.8x |
| c7g.4xlarge | 133.8ns | 27.2ns | 4.9x |
</details>

## Test Coverage

15 tests covering:
- Cumulative and delta temporality for both Counter and Histogram
- Bound + unbound sharing the same data point (same attribute set)
- Idle delta cycles (no update = no export, but handle persists)
- Cardinality overflow fallback to unbound path (Counter and Histogram)
- Recovery after delta eviction frees cardinality space
- Overflow fallback handles working across multiple delta cycles
- Drop enabling eviction of stale entries
- Multiple bound handles sharing the same tracker (ref-counting)
- Binding with empty attributes

## Test plan

- [x] All 164 metrics tests pass (149 existing + 15 new)
- [x] Compiles with and without the feature flag
- [x] Zero regression on unbound measurement path


cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/3420 is merged. @bryantbiggs  can you rebase? The spec is progressing in parallel (https://github.com/open-telemetry/opentelemetry-specification/pull/5050), so this is perfect time to get this merged!

opentelemetry-sdk/Cargo.toml
```diff
@@ -48,7 +48,7 @@ trace = ["opentelemetry/trace", "rand", "percent-encoding"]
 jaeger_remote_sampler = ["trace", "opentelemetry-http", "http", "serde", "serde_json", "url", "experimental_async_runtime"]
 logs = ["opentelemetry/logs"]
 metrics = ["opentelemetry/metrics"]
-testing = ["opentelemetry/testing", "trace", "metrics", "logs", "tokio/sync"]
+testing = ["opentelemetry/testing", "trace", "metrics", "logs", "rt-tokio", "rt-tokio-current-thread", "tokio/macros", "tokio/rt-multi-thread", "experimental_metrics_bound_instruments"]
```
cijothomas: unclear why we need to modify testing feature to bring bound instruments to it?

opentelemetry/src/metrics/instruments/mod.rs
```diff
@@ -28,6 +28,21 @@ pub trait AsyncInstrument<T>: Send + Sync {
 pub trait SyncInstrument<T>: Send + Sync {
     /// Records a measurement synchronously.
     fn measure(&self, measurement: T, attributes: &[KeyValue]);
+
+    /// Binds this instrument to a fixed set of attributes, returning a handle
+    /// that records measurements without per-call attribute lookup.
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    fn bind(&self, _attributes: &[KeyValue]) -> Box<dyn BoundSyncInstrument<T> + Send + Sync> {
+        unimplemented!("bound instruments not supported by this implementation")
```
cijothomas: could we do noops instead of panics?

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -184,6 +194,78 @@ where
...
+    /// The caller can then call `tracker.aggregator.update()` directly, bypassing the
+    /// full lookup path on subsequent measurements.
+    ///
+    /// Returns `None` if the cardinality limit has been reached. The caller should
+    /// fall back to the unbound `Measure::call()` path, which handles overflow
+    /// correctly and enables automatic recovery when space opens up after delta
+    /// collection evicts stale entries.
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    fn bind(&self, attributes: &[KeyValue]) -> Option<Arc<TrackerEntry<A>>> {
+        if attributes.is_empty() {
```
cijothomas, line 213: for empty attributes, the perf was already good as it does not do any lookups in hashmap. Do we need this special handling for them in the bound instrument scenario? They can always do the fallback to original path, which would give same perf as binding, but keeps code simpler.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -51,21 +55,27 @@ pub(crate) trait Aggregator {
...
     pub(crate) has_been_updated: AtomicBool,
+    pub(crate) bound_count: AtomicUsize,
 }
 
 impl<A: Aggregator> TrackerEntry<A> {
     fn new(config: &A::InitConfig) -> Self {
         TrackerEntry {
             aggregator: A::create(config),
             has_been_updated: AtomicBool::new(false),
+            bound_count: AtomicUsize::new(0),
```
cijothomas, line 78: Could this be simplified: replace `bound_count: AtomicUsize` with `is_bound: AtomicBool`. Set it to `true` on `bind()`, never clear it. i.e no `Drop` impl needed. Eviction just entires with `is_bound == true`. Once bound, stays forever. We can offer a UnBind() method to flip the flag to false, and then next cycle will clean it up.

If a user binds same attributes more than once - that is fine - we return the same tracker anyway, just that we won't count how many are bound to a given bound - we just know it is bounded, so can't just remove during eviction.

What do you think? (The spec is still silent on the UnBind part, so I am kinda making some advanced guesses...)

opentelemetry-sdk/src/metrics/mod.rs
```diff
@@ -4692,4 +4703,665 @@ mod tests {
         assert!("".parse::<Temporality>().is_err());
         assert!("cumulativ".parse::<Temporality>().is_err());
     }
+
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
cijothomas, line 4721: we need to have a test for the scenario:
1. Counter is created.
2. Bound counter is created with k1=v1,k2=v2,k3=v3
3. User has a View to only allow k1,k2.
4. Validate that bind occurs to k1=v1,k2=v2. (i.e k3=v3 is ignored, even though provided at bind time)

Also a bench to confirm that the view resolution/filtering is done at bind() time itself, and hot path perf is as good as without views.

opentelemetry-sdk/Cargo.toml
bryantbiggs: Reverted. CI exercises bound instruments via `--all-features` already, so `testing` doesn't need it.

opentelemetry/src/metrics/instruments/mod.rs
bryantbiggs: Done. Replaced with a `NoopBoundSyncInstrument` plus an `otel_debug!` so unoverridden `bind()` calls drop measurements instead of crashing the host.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Good catch, and turns out there was actually a divergence bug here. `bind(&[])` was creating its own HashMap entry while `add(_, &[])` writes to `no_attribute_tracker` (a separate field), so mixed bound + unbound empty-attr calls produced two data points for the same logical aggregation. Fixed by wrapping `no_attribute_tracker` in an `Arc` and having `bind(&[])` return a handle to that same tracker, so both paths converge. Regression tests added: `bound_counter_empty_attributes_shares_with_unbound` and a histogram analog.

opentelemetry-sdk/src/metrics/internal/mod.rs
bryantbiggs: Mulled this over and I'd like to stick with ref-counting if you're open to it. RAII is pretty much the default for handle types in Rust (`File`, `MutexGuard`, `Arc`, etc.), so a dropped handle that permanently consumes a cardinality slot feels off. The bigger thing for me is that per-request `bind()`, which people *will* write whether we want them to or not, degrades to allocation churn under ref-counting but becomes a real cardinality leak under `is_bound`. The first self-heals, the second eventually breaks the metric.

We can still add a public `unbind()` (or similar) later as the spec firms up; the current design supports both. Happy to switch if there's a cross-SDK alignment angle I'm not seeing.

opentelemetry-sdk/src/metrics/mod.rs
bryantbiggs: Added. `bound_counter_view_filters_attributes_at_bind_time` and a histogram analog cover the scenario, and `Counter_Bound_With_View_Delta` confirms the hot path stays at ~bound speed (filter resolved at `bind()`, not per call). Heads up that criterion doesn't fail CI on regression by itself, so the bench documents the invariant rather than enforcing it. Happy to wire up a `cargo criterion --baseline` gate if you want hard enforcement.

bryantbiggs:
Pushed a round of fixes addressing all five comments; reply inline on the `bound_count` one with my reasoning for keeping ref-counting.

Two extras worth flagging that aren't tied to a comment: `ValueMap::bind()` now inserts both unsorted and sorted keys for new entries (mirrors `measure()`'s pattern so unbound calls hit the fast path regardless of which side created the tracker), and `BoundCounter`/`BoundHistogram` are now `Arc`-backed and `Clone` so a single bound state can be shared across threads without re-binding.

opentelemetry-sdk/src/metrics/mod.rs
cijothomas: this is good! Just wanted to make sure that the PR's code is correctly resolving view/filtering at bind time itself. It is important, as bind() is giving a contract that the performance of it will be same irrespective of filters being present or not!

opentelemetry-sdk/src/metrics/internal/mod.rs
cijothomas: Agree. Lets continue with bound_count now, and if spec makes a firmer requirement later, we can revisit.

opentelemetry-sdk/src/metrics/internal/sum.rs
```diff
@@ -37,6 +43,51 @@ where
     }
```
cijothomas, line 43: **Suggestion: bind directly to overflow tracker, eliminate the Fallback path**

The current `Fallback` variant delegates every `bound.add()` call through the full unbound `measure()` path (sort → hash → RwLock → HashMap lookup). This creates several problems:

1. **No performance guarantee.** A `bind()` call made before cardinality overflow gets ~2ns direct path. The same call made after overflow gets ~53ns — the user has no way to know which they got. The whole point of `bind()` is a consistent performance guarantee.

2. **Unpredictable attribution.** The Fallback handle calls `measure(value, attrs)` each time, which means the data could end up in the overflow bucket, *or* in a dedicated entry if space frees up after delta eviction — and can flip between the two across collection cycles depending on what other unbound callers are doing. That's not "bound" to anything.

3. **Code complexity.** The `Fallback` variant, `BoundFallbackHandle`, and the `fallback: Arc<dyn Measure<T>>` parameter threaded through `Measure::bind()` all exist solely for this path.

**Proposed alternative:** When `ValueMap::bind()` hits the cardinality limit, bind directly to the overflow `TrackerEntry` instead of returning `None`. This gives the handle the same ~2ns direct write path, consistently lands in the overflow bucket, and removes the `Fallback` variant entirely. The trade-off is that data stays attributed to overflow permanently (even if space frees up), but:

- `bind()` is typically done at startup before cardinality fills up — hitting overflow at bind time is already a degenerate case
- If a user wants to recover, they can drop and re-`bind()`, which will get a dedicated tracker if space is available
- Consistent, predictable behavior is better than accidental self-healing that depends on other callers' patterns

bryantbiggs: Pushed two commits.

1. `fix(metrics): bind() at cardinality limit binds directly to overflow` (commit 1)

`ValueMap::bind_attrs` now looks up (or lazily creates) the overflow `TrackerEntry` at the limit and returns a direct handle. The `Fallback` variants in `BoundSumInner` / `BoundHistogramInner` are removed. Bound handles hold `Arc<TrackerEntry>` and write directly for their lifetime — same atomic-update hot path as a normal bind. Drop-and-rebind to recover after delta eviction frees space. Spec-aligned with #5050: overflow attribution preserved (MUST), per-call lookup bypassed (SHOULD).

Bench (M4 Max, 3-attr key set):

| | Before | After |
|---|---:|---:|
| Counter_Bound_Delta | 1.80 ns | 1.84 ns |
| Counter_Bound_AtOverflow_Delta | ~50 ns | 1.84 ns |
| Histogram_Bound_Delta | 6.50 ns | 6.69 ns |
| Histogram_Bound_AtOverflow_Delta | ~58 ns | 6.59 ns |

Resolves points 1 and 2. New tests `bound_*_overflow_persists_across_eviction_cycles` lock in the predictability invariant.

2. `refactor(metrics): extend direct-bind to all aggregators` (commit 2)

Removes `BoundFallbackHandle` and the `fallback` parameter on `Measure::bind` entirely. `LastValue`, `PrecomputedSum`, and `ExpoHistogram` now have direct-bind impls matching the Sum/Histogram pattern. `PrecomputedSum::bind` is unreachable from user code today (async-only) but kept so the trait is uniform and future `Gauge` / `Observable` bind() extensions are mechanical. The rare RwLock-poisoned case yields a `NoopBoundMeasure` (drops silently, same as `measure()`'s own poison handling).

Tests added: bound `ExponentialHistogram` via View (delta, cumulative, NaN/inf filter, view filter at bind time, overflow attribution, persistence across delta eviction); cumulative bind-at-overflow for Counter and Histogram; `LastValue` / `PrecomputedSum` unit tests through the `Measure` trait; lock-poison test on `ValueMap::bind`.

The second commit is the full point-3 cleanup. Happy to keep it here or split into a follow-up if you'd rather land the correctness fix first

cijothomas: Looks good.

cijothomas (approved):
LGTM!
Thanks for working on this. Spec PR is still pending, but give this is gated under feature-flag, okay to merge. I
ts aligned with current spec PR.

PR #3332, opened 2026-01-30
fix(proto): preserve InstrumentationScope version and attributes

djvcom:
## Summary

When logs have a target set, the `InstrumentationScope`'s version and attributes were being discarded during OTLP export. The target parameter now correctly overrides only the scope name, preserving other metadata.

- Fix `From` implementations for `InstrumentationScope` to preserve version and attributes when target is provided
- Add unit tests for the `From` implementations
- Add integration test demonstrating the fix with `group_logs_by_resource_and_scope`

## Context

This behaviour was introduced in PR #1869 to avoid exporting the appender crate's version when the scope name was overridden by target. However, PRs #2735 and #2796 subsequently changed both appenders to use empty scopes, so this concern no longer applies. Additionally, users who explicitly create scopes via `logger_with_scope()` would reasonably expect their metadata to be preserved.

## Test plan

- [x] Unit tests for both owned and reference `From` implementations
- [x] Integration test for `group_logs_by_resource_and_scope` with target set
- [x] `cargo test -p opentelemetry-proto --all-features --lib` passes
- [x] `cargo clippy -p opentelemetry-proto --all-features -- -Dwarnings` passes

Fixes #3276

PR #3428, opened 2026-03-18
feat(appender-tracing): add builder support for custom instrumentation scope

### LINKED ISSUE #3415: [Feature]: Allow passing InstrumentationScope when creating an instance of OpenTelemetryTracingBridge
### Related Problems?

#3332 

I would like to be able to pass an `InstrumentationScope` when creating an `OpenTelemetryTracingBridge`.  
For example:

```rust
OpenTelemetryTracingBridge::with_scope(provider, scope);
```

This implies that inside the constructor we would use:

```rust
 { ..., logger: provider.logger_with_scope(scope), ... } 
```

instead of:

```rust
{ ..., logger: provider.logger(""), ... }
```

This will allow passing attributes at the `scopeLogs` level:

```yaml
scopeLogs:
  - scope:
      name: ""
      version: ""
      attributes: [ __scope_attributes__ ]
```

### Considered Alternatives

_No response_

### Additional Context

_No response_

### Tip

DmitryAstafyev:
## Changes

This PR adds `OpenTelemetryTracingBridgeBuilder::with_scope()` to allow configuring the OpenTelemetry `InstrumentationScope` used by the appender logger.

The implementation also updates `OpenTelemetryTracingBridgeBuilder` to hold the logger provider until `build()`, so the builder can expose a set of configuration methods:
- `with_scope()`
- `with_span_attribute_allowlist()`
-  and potentially additional builder options in the future

## Design Notes

This change introduces a lifetime parameter on `OpenTelemetryTracingBridgeBuilder<'a>`. That was done to keep the builder aligned with the existing style, where the bridge is constructed from a provider reference rather than from a pre-created logger. 

In return, this keeps the API shape cleaner and gives us a more natural builder surface with related configuration methods grouped together.

## Compatibility

Technically, adding the lifetime to the public builder type is a breaking API change.

However, this builder API has not yet shipped in a released version, so this does not affect production users of published crates. In practice, the change is limited to the unreleased API surface.

## Testing

Added a unit test (`tracing_appender_with_custom_scope`) covering `with_scope()` and verifying that the configured `InstrumentationScope` is propagated to emitted logs.

## Related

* Fix: #3415

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -395,7 +395,7 @@ where
             logger: provider.logger(""),
             _phantom: Default::default(),
             #[cfg(feature = "experimental_span_attributes")]
-            span_attribute_allowlist: None,
+            span_attributes: None,
         }
     }
 
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -378,15 +378,10 @@ where
     /// Creates a bridge using the default OpenTelemetry logger scope.
     ///
     /// The default scope uses an empty scope name for the appender logger.
-    /// Use [`Self::with_scope`] or [`Self::builder_with_scope`] to provide a
+    /// Use [`Self::builder_with_scope`] to provide a
     /// custom [`InstrumentationScope`].
     pub fn new(provider: &P) -> Self {
         Self::builder(provider).build()
-    }
-
-    /// Creates the bridge with a custom OpenTelemetry [`InstrumentationScope`].
-    pub fn with_scope(provider: &P, scope: InstrumentationScope) -> Self {
-        Self::builder_with_scope(provider, scope).build()
     }
 
     /// Creates a builder using the default OpenTelemetry logger scope.
@@ -956,48 +951,35 @@ mod tests {
             .with_schema_url("https://opentelemetry.io/schemas/1.0.0")
             .with_attributes([KeyValue::new("scope-key", "scope-value")])
             .build();
-        let subscribers = [
-            tracing_subscriber::registry().with(
-                layer::OpenTelemetryTracingBridge::builder_with_scope(
-                    &logger_provider,
-                    scope.clone(),
-                )
+        let subscriber = tracing_subscriber::registry().with(
+            layer::OpenTelemetryTracingBridge::builder_with_scope(&logger_provider, scope.clone())
                 .build(),
-            ),
-            tracing_subscriber::registry().with(layer::OpenTelemetryTracingBridge::with_scope(
-                &logger_provider,
-                scope,
-            )),
-        ];
+        );
 
-        for subscriber in subscribers {
-            let _guard = tracing::subscriber::set_default(subscriber);
+        let _guard = tracing::subscriber::set_default(subscriber);
 
-            error!(name: "scoped-event", target: "my-system", event_id = 20);
-            assert!(logger_provider.force_flush().is_ok());
+        error!(name: "scoped-event", target: "my-system", event_id = 20);
+        assert!(logger_provider.force_flush().is_ok());
 
-            let exported_logs = exporter
-                .get_emitted_logs()
-                .expect("Logs are expected to be exported.");
-            assert_eq!(exported_logs.len(), 1);
-            let log = exported_logs
-                .first()
-                .expect("At least one log is expected to be present.");
+        let exported_logs = exporter
+            .get_emitted_logs()
+            .expect("Logs are expected to be exported.");
+        assert_eq!(exported_logs.len(), 1);
+        let log = exported_logs
+            .first()
+            .expect("At least one log is expected to be present.");
 
-            let instrumentation_scope = &log.instrumentation;
-            assert_eq!(instrumentation_scope.name(), "test.scope");
-            assert_eq!(instrumentation_scope.version(), Some("1.2.3"));
-            assert_eq!(
-                instrumentation_scope.schema_url(),
-                Some("https://opentelemetry.io/schemas/1.0.0")
-            );
-            assert!(instrumentation_scope.attributes().eq([KeyValue::new(
-                "scope-key",
-                "scope-value"
-            )]
-            .iter()));
-            exporter.reset();
-        }
+        let instrumentation_scope = &log.instrumentation;
+        assert_eq!(instrumentation_scope.name(), "test.scope");
+        assert_eq!(instrumentation_scope.version(), Some("1.2.3"));
+        assert_eq!(
+            instrumentation_scope.schema_url(),
+            Some("https://opentelemetry.io/schemas/1.0.0")
+        );
+        assert!(instrumentation_scope
+            .attributes()
+            .eq([KeyValue::new("scope-key", "scope-value")].iter()));
+        exporter.reset();
     }
 
     #[test]
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -956,33 +956,48 @@ mod tests {
             .with_schema_url("https://opentelemetry.io/schemas/1.0.0")
             .with_attributes([KeyValue::new("scope-key", "scope-value")])
             .build();
+        let subscribers = [
+            tracing_subscriber::registry().with(
+                layer::OpenTelemetryTracingBridge::builder_with_scope(
+                    &logger_provider,
+                    scope.clone(),
+                )
+                .build(),
+            ),
+            tracing_subscriber::registry().with(layer::OpenTelemetryTracingBridge::with_scope(
+                &logger_provider,
+                scope,
+            )),
+        ];
 
-        let subscriber = tracing_subscriber::registry().with(
-            layer::OpenTelemetryTracingBridge::builder_with_scope(&logger_provider, scope).build(),
-        );
-        let _guard = tracing::subscriber::set_default(subscriber);
+        for subscriber in subscribers {
+            let _guard = tracing::subscriber::set_default(subscriber);
 
-        error!(name: "scoped-event", target: "my-system", event_id = 20);
-        assert!(logger_provider.force_flush().is_ok());
+            error!(name: "scoped-event", target: "my-system", event_id = 20);
+            assert!(logger_provider.force_flush().is_ok());
 
-        let exported_logs = exporter
-            .get_emitted_logs()
-            .expect("Logs are expected to be exported.");
-        assert_eq!(exported_logs.len(), 1);
-        let log = exported_logs
-            .first()
-            .expect("At least one log is expected to be present.");
+            let exported_logs = exporter
+                .get_emitted_logs()
+                .expect("Logs are expected to be exported.");
+            assert_eq!(exported_logs.len(), 1);
+            let log = exported_logs
+                .first()
+                .expect("At least one log is expected to be present.");
 
-        let instrumentation_scope = &log.instrumentation;
-        assert_eq!(instrumentation_scope.name(), "test.scope");
-        assert_eq!(instrumentation_scope.version(), Some("1.2.3"));
-        assert_eq!(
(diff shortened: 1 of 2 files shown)


DmitryAstafyev:
Hello @cijothomas, could you please take a look at this PR? It's my first time contributing here, and I'm ready to correct/change if it's required. Many thanks in advance for review.

lalitb (approved):

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -322,10 +322,21 @@ where
...
+    /// Use [`Self::with_scope`] or [`Self::builder_with_scope`] to provide a
+    /// custom [`InstrumentationScope`].
     pub fn new(provider: &P) -> Self {
         Self::builder(provider).build()
     }
 
+    /// Creates the bridge with a custom OpenTelemetry [`InstrumentationScope`].
+    pub fn with_scope(provider: &P, scope: InstrumentationScope) -> Self {
+        Self::builder_with_scope(provider, scope).build()
+    }
```
utpilla: I don't think need `with_scope` as a separate public method? Calling `builder_with_scope(provider, scope).build()` is not too much of an inconvenience, and it keeps the API surface smaller.

lalitb: @DmitryAstafyev  - can you accommodate this change, and we can merge afterwards.

DmitryAstafyev: Hello @lalitb, @utpilla 
Method `with_scope` has been removed

cijothomas:
@DmitryAstafyev Can you share bit more about the intended use-case? Since OTLP Exporter override scope-name with target, this may not have the intended effect, so want to make sure we have a good grasp of the scenario.

dmitryastafyev-dyna:
> @DmitryAstafyev Can you share bit more about the intended use-case? Since OTLP Exporter override scope-name with target, this may not have the intended effect, so want to make sure we have a good grasp of the scenario.

Thanks for your question @cijothomas . I hope I've gotten your question in right way. 

We are interested in not loosing attributes, overwriting scope-name with target is okey. The use case is simple: we need to attach per-invocation metadata to every log record's scope in OTLP output. This metadata is dynamic and cannot live on the Resource level, so it belongs on the `InstrumentationScope`. Without `builder_with_scope`, there was no API to pass a custom scope to the bridge, which forced us to work around it by wrapping the provider.

If the concern is that scope attributes were being discarded by the OTLP exporter when a target is present - that was indeed a real issue, but it appears to have been fixed by [3332](https://github.com/open-telemetry/opentelemetry-rust/pull/3332) which was merged recently. With both fixes in place the full use case should work as intended for us.

cijothomas:
> > @DmitryAstafyev Can you share bit more about the intended use-case? Since OTLP Exporter override scope-name with target, this may not have the intended effect, so want to make sure we have a good grasp of the scenario.
> 
> Thanks for your question @cijothomas . I hope I've gotten your question in right way.
> 
> We are interested in not loosing attributes, overwriting scope-name with target is okey. The use case is simple: we need to attach per-invocation metadata to every log record's scope in OTLP output. This metadata is dynamic and cannot live on the Resource level, so it belongs on the `InstrumentationScope`. Without `builder_with_scope`, there was no API to pass a custom scope to the bridge, which forced us to work around it by wrapping the provider.
> 
> If the concern is that scope attributes were being discarded by the OTLP exporter when a target is present - that was indeed a real issue, but it appears to have been fixed by [3332](https://github.com/open-telemetry/opentelemetry-rust/pull/3332) which was merged recently. With both fixes in place the full use case should work as intended for us.

Thanks for the explanation @DmitryAstafyev! One clarification — the InstrumentationScope is set once when the bridge is created and stays fixed for its lifetime. So by "per-invocation metadata," do you mean per-process-invocation (e.g., a CLI/batch job that runs and exits)? In a long-running server, the scope attributes wouldn't change across requests — per-request data would need to go as log record attributes instead. Once you share more details, I can tell if there is better way to do it.!

If you need per-request metadata in a long-running server, `tracing` span attributes would be a better fit — we recently added support for propagating span attributes to log records via the `experimental_span_attributes` feature (it won't be stored as Scope level, but at LogRecord level)

dmitryastafyev-dyna:
> > > @DmitryAstafyev Can you share bit more about the intended use-case? Since OTLP Exporter override scope-name with target, this may not have the intended effect, so want to make sure we have a good grasp of the scenario.
> > 
> > 
> > Thanks for your question @cijothomas . I hope I've gotten your question in right way.
> > We are interested in not loosing attributes, overwriting scope-name with target is okey. The use case is simple: we need to attach per-invocation metadata to every log record's scope in OTLP output. This metadata is dynamic and cannot live on the Resource level, so it belongs on the `InstrumentationScope`. Without `builder_with_scope`, there was no API to pass a custom scope to the bridge, which forced us to work around it by wrapping the provider.
> > If the concern is that scope attributes were being discarded by the OTLP exporter when a target is present - that was indeed a real issue, but it appears to have been fixed by [3332](https://github.com/open-telemetry/opentelemetry-rust/pull/3332) which was merged recently. With both fixes in place the full use case should work as intended for us.
> 
> Thanks for the explanation @DmitryAstafyev! One clarification — the InstrumentationScope is set once when the bridge is created and stays fixed for its lifetime. So by "per-invocation metadata," do you mean per-process-invocation (e.g., a CLI/batch job that runs and exits)? In a long-running server, the scope attributes wouldn't change across requests — per-request data would need to go as log record attributes instead. Once you share more details, I can tell if there is better way to do it.!
> 
> If you need per-request metadata in a long-running server, `tracing` span attributes would be a better fit — we recently added support for propagating span attributes to log records via the `experimental_span_attributes` feature (it won't be stored as Scope level, but at LogRecord level)

Thanks @cijothomas for your question. Yes, our use case is per-invocation, not per-request in a long-running server. Sorry for confusion in my prev-reply. For each function invocation we create a fresh OpenTelemetryTracingBridge with invocation-scoped metadata (e.g. invocation ID  etc.). Span attributes on individual log records wouldn't be a fit here, since the metadata describes the bridge's operational context rather than individual log events. A native builder_with_scope would let us drop the wrapper-workaround we currently rely on.

dmitryastafyev-dyna:
Hello @cijothomas, any news about the progress?

dmitryastafyev-dyna:
Hello @cijothomas, @lalitb, I've rebased PR to master. Could you please take a look. Thanks in advance!

cijothomas:
> Hello @cijothomas, any news about the progress?

Apologies for the delay. Before going further on the API, can you share a minimal example of the per-invocation use case? Specifically:

Is the process short-lived (one invocation per process), or long-running with many invocations in-process?
The right solution could be - Resource, tracing span attributes onto each LogRecord, or something else.

The reason this matters: as written, builder_with_scope(provider, scope) looks like the supplied scope is used directly, but the bridge uses a single placeholder Logger and the scope name on every record comes from `target`. 
Only the scope's attributes / version / schema_url are kept.

Renaming to `with_scope_attributes` is clearer about what's accepted but doesn't fix the per-target replication, and gets harder if we ever move to one Logger per target. Hence wanting to understand the scenario before deciding whether this should be a bridge-level API at all.

https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/design/logs.md#performance, section 5 explains why we are in this tricky situation. We certainly want to make sure the scenarios is well understood before expanding the API here else this would get more complex.

DmitryAstafyev:
@cijothomas  thanks for your question.

I read our previous discussion, and I think I may have confused you and a little messed up things. Sorry if that happened.

Let me clarify the lifecycle first:

- What is an invocation? - it's one run of user code.
- How many invocations live in one process? - one invocation per process.
- Is a new `LoggerProvider` created per invocation? - one per process/run.
- Is a new `OpenTelemetryTracingBridge` created per invocation? - one per
  process/run.
- Can two invocations log through the same subscriber/provider at the same
  time? - not in our case.
- Can logs from different invocations end up in the same OTLP batch?- not in our case.

I also wasn't correct when talking about log record attributes. We do not want to copy these attributes into every log record. What we need is scope attributes under `ResourceLogs.ScopeLogs.scope.attributes`. 

The attributes contain some important metadata for us. With the current `builder(P)`, the bridge creates the logger with `provider.logger("")`. That creates an empty scope, so `scope.attributes` are empty. After export, we do not see the attributes on the endpoint.

With `provider.logger_with_scope(scope)` inside `builder_with_scope`, the logger keeps the provided scope. This gives us scope attributes, including after OTLP export.

I agree that the method name could be better. `with_scope_attributes`, like you suggested, sounds reasonable for me . Please let me know if that would be enough.

I hope I've answered your question more clearly this time :) 

addon:

> The right solution could be - Resource

I will back with an answer tomorrow. Have to check.

cijothomas:
@DmitryAstafyev 

Given the lifecycle you described (one process per invocation, metadata constant for the whole run), I think Resource might actually be a cleaner fit than Scope here - it's process-scoped, set once at startup, and flows through as ResourceLogs.resource.attributes on every export, which sounds like exactly what you need.

(also - what backend are you using? Some might not support scope attributes, and might require Collector massaging to move attributes around to the right place)

dmitryastafyev-dyna:
Thanks @cijothomas for your response. I took time to check the proposed change and the alternatives against our system. Sorry, I can't share the code, but I'll try to explain.

In our model one process = one invocation. On the Resource we keep only stable identity that does not change between invocations (service name, version, etc.). We do not put per-invocation data (like the invocation id and others) there. The Resource is shared by traces and metrics in the same process, so anything we add to it also lands on our metrics and traces. Metrics are aggregated across many invocations on the backend, and the backend groups them by Resource identity. Per-invocation data in the Resource would make every invocation a separate series and affect aggregation. 

One way around this would be separate Resources per signal. But I would not do it. Resource is the identity of the entity that produces the telemetry, and backends correlate logs, metrics, and traces by matching that identity. In other words, using different Resources for what is actually the same entity does not look right. This is also reflected in the [spec](https://opentelemetry.io/docs/specs/semconv/resource/faas/), where `faas.instance` / `faas.name` / `faas.version` are Resource attributes, but `faas.invocation_id` is a span attribute.

Another work-around is the per-record option (span attributes on each LogRecord). I mean copying the some data (attrs) into every record. We can do this with a custom logger, but it isn't efficient: the value is constant for the whole run, so the right way is one copy at `ScopeLogs.scope.attributes`. It's what `builder_with_scope_attributes` (I renamed the method) gives us.

> what backend are you using? 

This is a data-model question (scope vs Resource vs per-record), independent of the backend. Our backend reads scope-level attributes, so it's not a blocker for us.

@cijothomas, with the rename and the use-case clarified, is there any remaining blocker, or are we good to merge?

cijothomas:
@dmitryastafyev-dyna Thanks. It's more clear now. Let me re-review the changes with this context.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -392,6 +398,20 @@ where
             span_attributes: None,
         }
     }
+
+    /// Creates a builder that uses a custom OpenTelemetry [`InstrumentationScope`]
+    /// for the appender logger.
+    pub fn builder_with_scope_attributes(
```
cijothomas, line 404: the name is good. But the API should be narrowed down to just accept attributes alone, not the full InstrumentationScope.

cijothomas:
> @dmitryastafyev-dyna Thanks. It's more clear now. Let me re-review the changes with this context.

https://github.com/open-telemetry/opentelemetry-rust/pull/3428/changes#r3473102342 - Once this is addressed, we can merge.

dmitryastafyev-dyna:
> > @dmitryastafyev-dyna Thanks. It's more clear now. Let me re-review the changes with this context.
> 
> https://github.com/open-telemetry/opentelemetry-rust/pull/3428/changes#r3473102342 - Once this is addressed, we can merge.

Thanks @cijothomas . Done.

cijothomas (approved):
Agree

cijothomas:
@DmitryAstafyev Queue to merge. Thanks for patiently working on this!

PR #3438, opened 2026-03-26
fix: timeout being ignored in MeterProvider shutdown_with_timeout

### LINKED ISSUE #2574: Add ability to add timeout duration to shutdown

partg952:
Fixes #2574 

## Changes

Passed down the ignored timeout argument in shutdown_with_timeout in MeterProvider

When calling the shutdown_with_timeout in MeterProvider the timeout was being ignored and not being passed down the incoming inner function calls
This PR fixes that by calling shutdown_with_timeout at each level and passing timeout along with it.

### DIFF
--- a/opentelemetry-sdk/src/metrics/meter_provider.rs
+++ b/opentelemetry-sdk/src/metrics/meter_provider.rs
@@ -110,12 +110,12 @@ impl SdkMeterProvider {
     ///
     /// There is no guaranteed that all telemetry be flushed or all resources have
     /// been released on error.
-    pub fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult {
+    pub fn shutdown_with_timeout(&self, timeout: Duration) -> OTelSdkResult {
         otel_debug!(
             name: "MeterProvider.Shutdown",
             message = "User initiated shutdown of MeterProvider."
         );
-        self.inner.shutdown()
+        self.inner.shutdown_with_timeout(timeout)
     }
 
     /// shutdown with default timeout
@@ -136,7 +136,7 @@ impl SdkMeterProviderInner {
         }
     }
 
-    fn shutdown_with_timeout(&self, _timeout: Duration) -> OTelSdkResult {
+    fn shutdown_with_timeout(&self, timeout: Duration) -> OTelSdkResult {
         if self
             .shutdown_invoked
             .swap(true, std::sync::atomic::Ordering::SeqCst)
@@ -144,7 +144,7 @@ impl SdkMeterProviderInner {
             // If the previous value was true, shutdown was already invoked.
             Err(crate::error::OTelSdkError::AlreadyShutdown)
         } else {
-            self.pipes.shutdown()
+            self.pipes.shutdown_with_timeout(timeout)
         }
     }
 
--- a/opentelemetry-sdk/src/metrics/pipeline.rs
+++ b/opentelemetry-sdk/src/metrics/pipeline.rs
@@ -3,6 +3,7 @@ use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use opentelemetry::{otel_debug, otel_warn, InstrumentationScope, KeyValue};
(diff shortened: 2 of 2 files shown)


scottgerring:
Hey @partg952 , thanks for opening the PR!
I think you need to plumb the timeout further still; for instance, for `PeriodicReader` the timeout is discarded:

https://github.com/open-telemetry/opentelemetry-rust/blob/146376fde8463a40b807aa68c3d54eab373b3730/opentelemetry-sdk/src/metrics/periodic_reader.rs#L495-L497

and a 5 second duration is used instead:

https://github.com/open-telemetry/opentelemetry-rust/blob/146376fde8463a40b807aa68c3d54eab373b3730/opentelemetry-sdk/src/metrics/periodic_reader.rs#L593-L595

(`ManualReader` is fine as is).

I reckon with _that_ fixed, we'd also want a changelog entry in opentelemetry-sdk, because the observed behaviour of the shutdown will change - we'll honour the API!

Could you also add a regression test that verifies the timeout is passed through?

scottgerring (changes requested):
left a comment above with some suggestions!

partg952:
Hey @scottgerring ,Thanks for taking a look , I will definitely edit it further and then add the regression test for it.

Thanks!

partg952:
> Hey @partg952 , thanks for opening the PR! I think you need to plumb the timeout further still; for instance, for `PeriodicReader` the timeout is discarded:
> 
> https://github.com/open-telemetry/opentelemetry-rust/blob/146376fde8463a40b807aa68c3d54eab373b3730/opentelemetry-sdk/src/metrics/periodic_reader.rs#L495-L497
> 
> and a 5 second duration is used instead:
> 
> https://github.com/open-telemetry/opentelemetry-rust/blob/146376fde8463a40b807aa68c3d54eab373b3730/opentelemetry-sdk/src/metrics/periodic_reader.rs#L593-L595
> 
> (`ManualReader` is fine as is).
> 
> I reckon with _that_ fixed, we'd also want a changelog entry in opentelemetry-sdk, because the observed behaviour of the shutdown will change - we'll honour the API!
> 
> Could you also add a regression test that verifies the timeout is passed through?

Hey @scottgerring , I have a little unclarity here, in periodic reeader, the shutdown with timeout is calling inner.shutdown normally without respecting the timeout , so do you want me to implement the whole functionality or just pass it down through the consequent function calls?

scottgerring:
> Hey @scottgerring , I have a little unclarity here, in periodic reeader, the shutdown with timeout is calling inner.shutdown normally without respecting the timeout , so do you want me to implement the whole functionality or just pass it down through the consequent function calls?

@partg952 i’d implement the behavior end-to-end, not just pass the value through. 

I don't think this should require anything super gnarly like changing public APIs, only the internal shutdown message/plumbing. For the final export during shutdown, I think it is okay that `PeriodicReader` can’t directly cancel an in-flight export; the important part is that the caller stops waiting after the provided timeout, and the exporter shutdown receives the timeout/remaining budget. Let us know if you get stuck or if something seems weird!

bryantbiggs:
@partg952 @scottgerring — heads up on some overlap worth reconciling. PR #3380 also threads the shutdown timeout down through `Pipelines` → `Pipeline` → `PeriodicReader` (the same plumbing requested in the review above), and additionally stores/joins the `PeriodicReader` worker `JoinHandle` (currently detached at construction) and fixes the constrained-Tokio-runtime export deadlock (#2802).

Since both PRs touch the metric shutdown-timeout path, it'd be good to align so the work isn't duplicated or conflicting. Whenever you have time, could you take a look at #3380 and share any thoughts? Happy to rescope #3380 around whatever lands here. For broader context, the timeout / worker-lifecycle picture across all the processors and readers is now collected in the umbrella #3542.

partg952:
I would still like to work on it , but if it is a high priority fix that needs to be taken care of immediately would be more than happy to hand it off.

PR #3451, opened 2026-04-04
feat: decouple prost from OTLP/JSON

### LINKED ISSUE #3419: [Feature]: Decouple `prost` from OTLP/JSON
### Related Problems?

Related to OTLP refactoring work: #3045

Enabling the OTLP "http-json" feature results in `prost` being pulled in and compiled, even though it isn't needed for OTLP/JSON.

This results in a much slower compile time:

```shell
> cd opentelemetry-otlp
> cargo clean
> cargo build --release --no-default-features --features "http-proto"
  ...
  Finished `release` profile [optimized] target(s) in 8.22s
```

```shell
> cd opentelemetry-otlp
> cargo clean
> cargo build --release --no-default-features --features "http-json"
  ...
  Finished `release` profile [optimized] target(s) in 15.19s
```

Since the OTLP types are generated through `prost-build`, there's no straightforward way to make `prost` itself optional. 

### Considered Alternatives

_No response_

### Additional Context

Given that `prost` is only passively maintained now, it might be a concern being tied to it for type generation. 

Decoupling the types would also likely make it easier to support the [official `protobuf` crate](https://crates.io/crates/protobuf/4.34.0-release) in the future too.

### Tip

jay-dee7:
Fixes #3419

> http-json currently drags in `prost` because the only OTLP message types we have are the prost-generated ones. Following the direction from @lalitb on the issue, this adds a second, serde-only type tree to opentelemetry-proto and points http-json at it. No cfg_attr on the prost tree.

```bash
$ cargo tree -p opentelemetry-otlp --no-default-features --features http-json --invert prost
error: package ID specification `prost` did not match any packages
```

## PR brief

The diff is big but almost all of it is generated or moved.

- `src/proto/json/*.rs` is newly generated & `src/proto/tonic/*.rs` is regenerated with new `tonic` path modifier, ~2k LoC
- `src/proto/serializers.rs` is the old serializers module from `proto.rs`, just moved for reusability
- `src/transform/{common,trace,logs,metrics}.rs` lost one level of indentation. The real changes are the use header and the enum conversions mentioned below.

What's actually new is around ~400 lines: 
- `transform/mod.rs`
- `tests/grpc_build.rs`
- the json module in proto.rs
- `opentelemetry-otlp/src/exporter/http/*.rs`

## Implementation detail

**The json tree comes from the same `prost-build` config as the tonic tree, with client/server generation off and prost references stripped afterwards (build_json in tests/grpc_build.rs).** Only five prost tokens ever appear in the output, so a text replacement is enough. The two trees are identical apart from that and the serde cfg_attr gating.

The serde helpers, the transforms and the conformance tests are written once and mounted twice with **`#[path]`**. Each mounting module declares:
```rust
use crate::proto::tonic as proto; // or json
```

and the file uses proto::... paths. 

See `transform/mod.rs`, `proto.rs` and `tests/json_serde/main.rs`. No macros, traits or new dependencies. `transform::*::tonic` paths are unchanged; `transform::*::json` is the same code against the other tree, and every transform test and conformance test now runs for both.

The HTTP exporter picks request and response types inside the existing `match self.protocol` arms. `OtlpHttpClient` stores the SDK Resource (an Arc, so set_resource is a refcount bump) and converts it for whichever tree the arm needs.

## A call I made (might be wrong)

The JSON enums don't get prost's `Default` / `From<Enum> for i32` / `TryFrom<i32>`. Enum fields are still `i32` so nothing needs them, but the transforms had six **`.into()`** calls that relied on the **`From impl`** and four **`DataPointFlags::default()`** calls. Those are now as `i32` (which is literally what prost's From impl does) and `DataPointFlags::DoNotUse` (the zero variant prost's Default returned). The tonic tree keeps all prost impls.

`#[allow(clippy::duplicate_mod)]` appears twice. It's the lint for mounting one file as two modules, which is the whole point here.

gen-json is not in full/default, same as profiles and with-schemars, so nothing changes for default users. http-json no longer enables opentelemetry-proto/gen-tonic-messages or with-serde; a downstream that relied on feature unification for those will get a build error rather than a silent behaviour change.

> The last commit is the one place shared code changes: the bucketCounts vec serializer becomes collect_seq (the old version built a throwaway Vec<String> and stringified each element twice) and the vec deserializer is composed from the scalar u64 one. Accepted inputs are the same; only the expecting message on malformed input differs.

Checked

- JSON output is byte-identical to main for traces, logs with every AnyValue variant, and sum/histogram/gauge/up-down-counter metrics, comparing main's tonic tree with both new trees.
- cargo hack --each-feature clippy -- -Dwarnings on both crates. lint.sh gains http-json on its own plus gen-json,{trace,metrics,logs}; with prost out of the graph, any prost use under http-json is a compile error.
- MSRV 1.75 for gen-json and http-json.
- Not done: tracez and profiles stay tonic-only and the json tree has no with-schemars. Neither has a JSON consumer.

Merge requirement checklist

- [x] CONTRIBUTING guidelines followed
- [x] Unit tests added/updated (if applicable)
- [x] Appropriate CHANGELOG.md files updated for non-trivial, user-facing changes
- [x] Changes in public API reviewed (if applicable)


scottgerring:
Hey @jay-dee7 thanks for the contribution 👋 

Just flagging that as this is quite a big diff and we are currently focussing on the OTLP stabilisation efforts we probably won't get to this for a bit

jay-dee7:
Sure @scottgerring. If you need a hand on anything higher priority, do let me know. Would love to contribute back in anyway possible :)

scottgerring:
> Sure @scottgerring. If you need a hand on anything higher priority, do let me know. Would love to contribute back in anyway possible :)

@jay-dee7 hey thanks for offering! I think we are about done; it's just a bunch of random PRs left to merge. We need to do a better job managing work on this repo so it is easy for folks to pick up and contribute, and we can turn our minds back to this once the OTLP piece is done 💪

akshatagarwl:
We hit the `with-serde` failure described in #3527 in a downstream OTLP/HTTP production service.

A canonical metrics payload containing a gauge with `asDouble: 1.5` and a sum with `asInt: "7"` returns `Ok(ExportMetricsServiceRequest)`, but the sum point's flattened `NumberDataPoint.value` becomes `None`; the gauge survives. We reproduced the same behavior against `opentelemetry-proto` 0.31, 0.32, and 0.33. The inner `i64` deserialization error is swallowed by the flattened `Option` rather than propagated.

It would be useful for this PR's JSON conformance suite to pin these cases before merge:

- `NumberDataPoint` and `Exemplar` accept canonical string-encoded `asInt` and retain `Some(AsInt(...))`.
- `aggregationTemporality` accepts names such as `AGGREGATION_TEMPORALITY_CUMULATIVE`.
- A mixed request containing a valid `asDouble` gauge and string-encoded `asInt` sum retains both metric data oneofs and both datapoints.
- The same request round-trips through `ExportMetricsServiceRequest`, not only the individual datapoint types, so flattened-oneof failures cannot pass silently.

Our downstream regression additionally queries persisted rows because an HTTP 200 alone did not reveal the loss. I can provide a smaller standalone Rust reproducer if that would help, though #3527 already contains the essential one.

PR #3457, opened 2026-04-09
Add public constructors to metrics data types and make SdkLogRecord::…

### LINKED ISSUE #2992: [Feature]: Add support for custom error handlers + make  `SdkLogRecord::new` public again
### Related Problems?

Hi,

We just migrated [Stalwart](https://github.com/stalwartlabs/stalwart) from opentelemetry `0.25` to `0.29` and noticed two issues that I briefly explained in #2260:

- We use our custom logging backend rather than tracing and `global::set_error_handler` has been removed.
- We also use run own own span, log and metrics export task and the `Default` trait has been removed from `SdkLogRecord`. In order to be able to generate a `SdkLogRecord` we now need to do `SdkLoggerProvider::builder().build().logger("stalwart").create_log_record()` which is not awful but it would be much nicer if `SdkLogRecord::new` was made public or the `Default` trait was added.

I propose to:

- Re-implement `global::set_error_handler` (or similar) and place it behind a Cargo feature, which will be disabled by default. I agree that most users will use `tracing`.
- Make  `SdkLogRecord::new` public or add the `Default` trait to `SdkLogRecord`.

I'm not familiar with the internals of this crate but if you prefer I can submit a PR for these.

### Considered Alternatives

_No response_

### Additional Context

_No response_

mdecimus:
Fixes #2992

## Changes

[Stalwart Mail Server](https://github.com/stalwartlabs/stalwart/) uses its own metrics collector and export pipeline rather than the SDK's periodic reader. Since 0.30 made all fields on metrics data types `pub(crate)`, it's no longer possible to construct `Metric`, `Sum`, `Gauge`, `Histogram`, `ResourceMetrics`, `ScopeMetrics`, or their data points from outside the crate. This adds `new()` constructors to each type so that custom exporters can build `ResourceMetrics` to pass to `PushMetricExporter::export()`.

Similarly, `SdkLogRecord::new()` is made public and `Default` is implemented, so consumers with custom logging backends can create log records directly without having to instantiate a `SdkLoggerProvider`.

* [X] Changes in public API reviewed (see #2992 comments)

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Fix panic when `SpanProcessor::on_end` calls `Context::current()` ([#3262][3262]).
   - Updated `SpanProcessor::on_end` documentation to clarify that `Context::current()` returns the parent context, not the span's context
 - Fix `traceparent` headers with unknown flags (e.g. W3C random-trace-id flag `0x02`) being incorrectly rejected. Unknown flags are now accepted and zeroed out as required by the W3C trace-context spec. [#3435][3435]
+- Add public constructors to metrics data types and make SdkLogRecord::new() public. [#2992][2992]
 
 [3227]: https://github.com/open-telemetry/opentelemetry-rust/pull/3227
 [3277]: https://github.com/open-telemetry/opentelemetry-rust/pull/3277
--- a/opentelemetry-sdk/src/logs/record.rs
+++ b/opentelemetry-sdk/src/logs/record.rs
@@ -119,8 +119,8 @@ impl opentelemetry::logs::LogRecord for SdkLogRecord {
 }
 
 impl SdkLogRecord {
-    /// Crate only default constructor
-    pub(crate) fn new() -> Self {
+    /// Creates a new empty `SdkLogRecord`.
+    pub fn new() -> Self {
         SdkLogRecord {
             event_name: None,
             target: None,
@@ -201,6 +201,12 @@ impl SdkLogRecord {
             .iter()
             .flatten()
             .any(|(k, v)| k == key && v == value)
+    }
+}
+
+impl Default for SdkLogRecord {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
--- a/opentelemetry-sdk/src/metrics/data/mod.rs
+++ b/opentelemetry-sdk/src/metrics/data/mod.rs
@@ -27,6 +27,14 @@ impl Default for ResourceMetrics {
 }
 
 impl ResourceMetrics {
+    /// Creates a new [ResourceMetrics] with the given [Resource] and [ScopeMetrics].
+    pub fn new(resource: Resource, scope_metrics: Vec<ScopeMetrics>) -> Self {
+        Self {
+            resource,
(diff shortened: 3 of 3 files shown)


scottgerring:
Hey @mdecimus ! 
Comparing this to the linked issue I think a few things are in play:

* Custom error log / handler callback -> I can't see that in this PR, but it _is_ something Cijo mentioned seemed reasonable
* Making it possible to create `SdkLogRecord`s directly --> in this PR and in the issue, but there's outstanding questions around the use case here
* Making metrics constructors public --> not in the issue (I think?) but in here. This seems to tie back to some changes made previously to cut down on the public API 

I think the first point could be raised easily as a standalone PR if you are still interested in this!

To the second two points to provide some more context for you, we are very cautious of expanding public API; it would be great to work through the reasoning for this on the linked issue. Depending on where that lands we'd probably also ask to break out the `SdkLogRecord` changes from the metrics API changes.

mdecimus:
Thanks for the review @scottgerring!

To give some context: Stalwart Mail Server has its own telemetry collector that aggregates spans, logs, and metrics from across the codebase using a custom event system. We don't use `tracing`, the SDK's periodic reader, batch processor, or any of the SDK's pipeline machinery; we just need the OTLP exporters as a transport to ship our already-aggregated data to a collector. Our flow looks like:

1. Stalwart's collector aggregates events
2. We translate them into `SdkLogRecord` / `SpanData` / `ResourceMetrics`
3. We hand them to `LogExporter::export()` / `SpanExporter::export()` / `PushMetricExporter::export()` directly

This means we're a downstream consumer of the SDK *types* without using the SDK's *processors*, which is why we hit these construction limitations.

Regarding **`SdkLogRecord`:** the use case is exactly the above; we need to build `SdkLogRecord`s in our own export task to feed to a `LogExporter`. Today the only way to do this is `SdkLoggerProvider::builder().build().logger("name").create_log_record()`, which forces us to instantiate a full provider just to get an empty record. Making `new()` public (or adding `Default`) removes that ceremony with no real change in surface area, since the struct is already `#[non_exhaustive]` and all fields are set through the trait.

Regarding the **metrics constructors**: since 0.30 made `Metric`, `Sum`, `Gauge`, `Histogram`, `ResourceMetrics`, `ScopeMetrics`, and the data point types `pub(crate)`, it's now impossible for any external consumer to call `PushMetricExporter::export()` at all, since the trait requires you to pass a `&ResourceMetrics` you cannot construct. The exporter trait is public but inert without a way to build its argument. The constructors here are minimal and additive; they don't expose any new fields.

Regarding the **error handler:** this isn't included in this PR. Adding it back cleanly would mean threading a global callback through the `otel_error!`/`otel_warn!` macros (currently they delegate to `tracing` when `internal-logs` is enabled, or compile to nothing otherwise), which is a bigger design change that probably warrants its own discussion. We ended up not needing it for Stalwart anyway, since we manually check `Result` at every `export()` call site.

Happy to split this into two PRs (one for `SdkLogRecord`, one for metrics constructors) if that makes review easier. Just let me know which direction you'd prefer.

mdecimus:
Hi, are you planning to merge this PR? Stalwart is several commits ahead.

PR #3460, opened 2026-04-14
feat: Implement process context publishing (OTEP-4719)

### LINKED ISSUE #3045: Simplify opentelemetry-proto: SDK decoupling, gRPC separation
Related issue #2508

This issue proposes a refactor of the `opentelemetry-proto` crate to improve clarity, reduce unnecessary dependencies, and simplify usage for consumers of just the OTLP types.

### 1. Decouple `opentelemetry-proto` from SDK
- Remove `opentelemetry-sdk` as a dependency of `opentelemetry-proto`.
- Move transform logic (e.g., `impl From<Resource> for proto::*`) to `opentelemetry-otlp`, so `opentelemetry-proto` contains only wire format definitions.

### 2. Logical Re-exports for Protobuf Types
- Organize generated Protobuf messages (e.g., `LogsData`, `LogRecord`, etc.) under logical module paths like:
  ```rust
  proto::opentelemetry::proto::logs::v1::*
  ```
- This avoids consumers needing to depend directly on the file layout under `src/proto/tonic`.

### 3. Isolate gRPC Service Definitions
- Keep gRPC service definitions (e.g., `LogsServiceClient`, `LogsServiceServer`) in the same physical location (`src/proto/tonic`), but expose them only behind the `gen-tonic` feature via:
  ```rust
  proto::tonic::opentelemetry::proto::collector::logs::v1::*
  ```

### 4. Rename `tonic` Folder for Clarity (Optional)
- To reduce confusion, consider renaming `src/proto/tonic` to `src/proto/generated` since it contains both tonic and prost-generated code.
- Maintain tonic-specific exports in a `mod tonic` gated by the `gen-tonic` feature.

### 5. Remove `gen-tonic-messages` Feature
- This flag is no longer needed now that grpcio support has been dropped. All prost-based message generation should be unconditional.

---

## ✅ Benefits

- Consumers can use Protobuf messages (for file/HTTP ingestion) without pulling in tonic or SDK dependencies.
- Reduces compile times and simplifies the dependency tree.
- Improves modularity and aligns with crate responsibilities:
  - `opentelemetry-proto`: types only
  - `opentelemetry-otlp`: SDK + transform logic
  - `gen-tonic`: gRPC transport

scottgerring:
## Summary

Implements the publisher side of [OTEP-4719 (Process Context Sharing)](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4719-process-ctx.md) behind the `process-context` feature flag on `opentelemetry-proto`. This is helpful for signal correlation - and together with the  [thread context sharing (OTEP-4947)](https://github.com/open-telemetry/opentelemetry-specification/pull/4947), will allow the Rust community to have request-correlated profiling support. This is a linux-only mechanism by design and the API to support it provides a no-op impl for other targets. 

On Linux, this publishes SDK resource attributes to a named memory mapping (`OTEL_CTX`) so external readers (such as the OpenTelemetry eBPF Profiler) can discover process metadata without direct integration or process activity.

This is a precursor for OTEP-4947 which will enable per-thread resource attribution building on top of the process-level mechanism implemented here.

The implementation is adapted from the implementation used in [libdatadog](https://github.com/DataDog/libdatadog/blob/main/libdd-library-config/src/otel_process_ctx.rs), a shared library we use underneath many of our own datadog libraries.  

### Usage

```rust

let resource = Resource::builder()
    .with_service_name("my-service")
    .build();

// Publish resource as process context (no-op on non-Linux)
opentelemetry_proto::process_context::publish(&resource);

// Later, during shutdown:
opentelemetry_proto::process_context::unpublish();
```

### Decisions
**Why `opentelemetry-proto`?**

I reckon this probably belongs in `opentelemetry-sdk`, but because `opentelemetry-proto` depends on _that_ for its transformation layer we can't reference our proto types there (see #3045). 

We have options: 

1. Publish it from the proto crate 
2. Create a new crate just for this (in the future perhaps it would also contain the thread-local context publishing mechanism) 
3. Fix #3045 first, put it in sdk

I'd like to do (3) (and am happy to impl it myself), but as I poked this issue recently and heard nothing back I am hesitant to put it in the dependency path for this. 

**Publishing mechanism**
I've made this _explicit_ - the user must explicitly use this API to publish - look at the updated example code. I'd prefer if this happened automatically, but because each signal stands alone and is provided its own resource, it is difficult to envisage how an automatic publishing mechanism should work. If anyone has any better ideas lmk. 


opentelemetry-proto/src/process_context/mod.rs
```diff
@@ -0,0 +1,139 @@
+//! Process context sharing via memory-mapped regions.
```
scottgerring, line 1: I've pulled this up to represent the public API, subbing in either a linux-64 or a no-op impl based on the target

opentelemetry-proto/Cargo.toml
```diff
@@ -61,12 +62,14 @@ schemars = { workspace = true, optional = true }
...
 const-hex = { workspace = true, optional = true }
 base64 = { workspace = true, optional = true }
+rustix = { workspace = true, optional = true, features = ["mm", "fs", "process", "time"] }
 
 [dev-dependencies]
 opentelemetry = { workspace = true, features = ["testing"] }
 tonic-prost-build = { workspace = true }
 tempfile = { workspace = true }
 serde_json = { workspace = true }
+serial_test = { workspace = true }
```
scottgerring, line 72: Makes it easier for testing

scottgerring:
@cijothomas @lalitb it would be great to get alignment on _where_ this should live (see my idea of the options in the PR text) first!

opentelemetry-proto/src/process_context/linux.rs
```diff
@@ -0,0 +1,611 @@
+//! Linux-specific implementation of the publisher part of the [OTEL process
+//! context](https://github.com/open-telemetry/opentelemetry-specification/pull/4719)
+//!
+//! # A note on race conditions
+//!
+//! Process context sharing implies concurrently writing to a memory area that another process
+//! might be actively reading. However, reading isn't done as direct memory accesses but goes
+//! through the OS, so the Rust definition of race conditions doesn't really apply. We also use
+//! atomics and fences, see MappingHeader's documentation.
```
yannham: So, I initially wrote this comment, but I think the wording is a bit misleading. I would maybe reword as:

```suggestion
//! might be actively reading. Reading isn't done as direct memory accesses but
//! goes through the OS, and is done by a separate process, which lies outside
//! of the Rust/C++ memory model. Our best bet is to use atomics and fences
//! defensively as if the reader was another thread of the same process. See
//! MappingHeader's documentation.
```

(If you accept the suggestion, I will propagate back to libdatadog)

opentelemetry-proto/src/process_context/linux.rs
```diff
@@ -0,0 +1,611 @@
...
+/// unpublished since, we follow the Update protocol.
+///
+/// # Fork safety
+///
+/// If we're a forked children of the original publisher, we are extremely restricted in the
+/// set of operations that we can do (we must be async-signal-safe). On paper, heap allocation
+/// is Undefined Behavior, for example. We assume that a forking runtime (such as Python or
+/// Ruby) that doesn't follow with an immediate `exec` is already "taking that risk", so to
+/// speak (typically, if no thread is ever spawned before the fork, things are mostly fine).
+pub(crate) fn publish_raw_payload(payload: Vec<u8>) -> Result<(), Error> {
```
yannham, line 346: I wonder if the whole fork business should really be here? I suppose support for languages with a forking runtime isn't really Datadog-specific, but one could also leave that out for a first implementation. I don't really have an opinion myself, just asking.

opentelemetry-proto/src/process_context/linux.rs
```diff
@@ -0,0 +1,611 @@
...
+    mem::ManuallyDrop,
+    ptr::{self, addr_of_mut},
+    sync::{
+        atomic::{fence, AtomicU64, Ordering},
+        Mutex, MutexGuard,
+    },
+    time::Duration,
+};
+
+use rustix::{
```
yannham, line 24: The latest version of the process context in `libdatadog` got rid of `rustix` to use `libc`. Rustix has a nicer interface (the switch was made for different technical reasons), but if by any chance you'd like to avoid a new dependency, the latest version in libdatadog is pure `libc`-based.

opentelemetry-proto/src/process_context/linux.rs
scottgerring: Accepting; I like the way this ties it back to _what the programmer should care about_

opentelemetry-proto/src/process_context/linux.rs
scottgerring: ACK; as it sounds like those reasons don't apply here I will stay with the higher level interface for the moment !

opentelemetry-proto/src/process_context/linux.rs
scottgerring: Good catch; this is certainly not a common pattern in Rust! I will remove it.

opentelemetry-proto/src/process_context/linux.rs
```diff
@@ -0,0 +1,611 @@
...
+/// Otherwise, if a context has been previously published from the same process and hasn't been
+/// unpublished since, we follow the Update protocol.
+///
+/// # Fork safety
+///
+/// If we're a forked children of the original publisher, we are extremely restricted in the
+/// set of operations that we can do (we must be async-signal-safe). On paper, heap allocation
+/// is Undefined Behavior, for example. We assume that a forking runtime (such as Python or
+/// Ruby) that doesn't follow with an immediate `exec` is already "taking that risk", so to
+/// speak (typically, if no thread is ever spawned before the fork, things are mostly fine).
```
scottgerring:
```suggestion
```

opentelemetry-proto/src/process_context/linux.rs
yannham: I also meant the whole `pid` business in the code. That is, I think we can simplify the code a bit if we don't care about forks (basically removing the `pid` field and the case `Some(handle) =>` below when `pid != getpid()`). We would also get away from the whole "memory allocation is UB after a fork" trap: we don't pretend to handle arbitrary forks anymore.

scottgerring:
<a href="https://easycla.lfx.linuxfoundation.org/#/?version=2">CLA Signed</a><br >The committers listed above are authorized under a signed CLA.<ul><li>:white_check_mark: login: scottgerring / name: Scott Gerring (377eda6a1dfd20164ba4433d8a925f9e8b02beec, 3bdd25c7d0e7010d5c84418af39ff15e575768df, 3e1cc9504efa56d99e86c8b6606202c245077773, 7085416b7f5e823eaceca7e1d34ad32edb012c85, a5acfd05bcb47f39c73293ea10f4143e5ecbb0c6, a95d00809b09d2fc1fdd291c36fcfb3298a4f7fd, c007c7a2407f853672fb60332dfd10ad97720cc4, c6ed75d547c78da6de9a8a247f709c78a2968736, d62ab331fd542f82df12fb76ec676b0e0a7e340f, 82ea43d501e30131f762f893909f281731718377)</li></ul>

opentelemetry-context-sharing/src/process_context/linux.rs
```diff
@@ -0,0 +1,611 @@
...
+/// unpublished since, we follow the Update protocol.
+///
+/// # Fork safety
+///
+/// If we're a forked children of the original publisher, we are extremely restricted in the
+/// set of operations that we can do (we must be async-signal-safe). On paper, heap allocation
+/// is Undefined Behavior, for example. We assume that a forking runtime (such as Python or
+/// Ruby) that doesn't follow with an immediate `exec` is already "taking that risk", so to
+/// speak (typically, if no thread is ever spawned before the fork, things are mostly fine).
+pub(crate) fn publish_raw_payload(payload: Vec<u8>) -> Result<(), Error> {
```
scottgerring: Gah missed this. Let me go over it again in more detail and make sure i've got it all. Thanks!

lalitb:
@scottgerring  - Thanks for calling out the placement tradeoff in the PR description. I agree that the process-context proto message itself is related to opentelemetry-proto, but I am less sure about adding publish/unpublish there as the public API. 

That API owns process-global runtime behavior, Linux mmap/memfd/prctl details, update semantics, and SDK Resource lifecycle. With thread-context sharing also being discussed, this area may grow beyond just process-level context. Option 2 from the description, a small dedicated context-sharing crate, feels cleaner to me than expanding `opentelemetry-proto` into this runtime surface.

One related question: should `process_context.proto` first live in open-telemetry/opentelemetry-proto? From what I can tell, this PR adds the proto locally in opentelemetry-rust rather than consuming it from the upstream proto definitions. If the message shape is still part of the OTEP/spec work, I would prefer not to make opentelemetry-rust
look like the canonical owner of that proto.

I do not want to block this PR only on crate placement, but I think it is worth aligning on the intended long-term home before this becomes a public API we have to support.

scottgerring:
Hey @lalitb thanks for taking a look!

> Option 2 from the description, a small dedicated context-sharing crate, feels cleaner to me than expanding opentelemetry-proto into this runtime surface.

Agreed 100% it is weird in the proto repo! I think there's an argument to be made it could go in `opentelemetry-sdk`, but for that we need to fix the proto<->sdk coupling issue. I have a local branch that does that but that's also a big piece of work in and of itself. What's your feeling on fixing that and putting it in the sdk vs. new crate? 

P.S. i've added two commits on the top to show what it looks like as a standalone crate.

> One related question: should `process_context.proto` first live in open-telemetry/opentelemetry-proto? From what I can tell, this PR adds the proto locally in opentelemetry-rust rather than consuming it from the upstream proto definitions. If the message shape is still part of the OTEP/spec work, I would prefer not to make opentelemetry-rust look like the canonical owner of that proto.

There's [some discussion](https://github.com/open-telemetry/opentelemetry-proto/pull/783) ongoing about this but it does look like it's going to end up in the proto repo. I think in the meantime we could get it going with a local copy with heavy caveats ("this is going to move!!1") - what do you think?

opentelemetry-context/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "opentelemetry-context"
```
lalitb: One naming thought, not strongly held: is `opentelemetry context` a bit broad here? In OpenTelemetry, “Context” already has a specific meaning around in-process propagation. Since this PR is specifically about process context, `opentelemetry-process-context` might be clearer. If the intent is to keep process and future thread context sharing together, maybe something like `opentelemetry-profiler-context` could also work. Mainly wondering if we should pick a name that avoids confusion with the existing OTel Context concept.

opentelemetry-proto/src/proto/process_context.proto
```diff
@@ -0,0 +1,51 @@
+// Copyright The OpenTelemetry Authors
```
lalitb: Now that open-telemetry/opentelemetry-proto#783 has merged, do we still need to check in `process_context.proto` directly in this repo? I would prefer this to come from the upstream `opentelemetry-proto` definitions, like the other upstream `proto` files, instead of adding a separate Rust-local proto copy here.

scottgerring: Yep - missed that one went in! Done.

opentelemetry-context/Cargo.toml
scottgerring: I agree!  Naming is hard. After a bunch of back and forth with my esteemed colleague @ivoanjo, i've renamed it to "opentelemetry-context-sharing"; this will give us a home for the TLS subscriber too when we get there, and is nicely disambiguated from "context"

scottgerring:
@lalitb I think this should be good to go; we're on the proto repo now

opentelemetry-context-sharing/src/process_context/linux.rs
```diff
@@ -0,0 +1,577 @@
...
+    /// it will fail. It is ok for naming to fail - we must only make sure that at least we
+    /// tried, as per the
+    /// [spec](https://github.com/open-telemetry/opentelemetry-specification/pull/4719).
+    fn set_name(&mut self) -> Result<(), Error> {
+        // Safety: the invariants of `MemMapping` ensures that `start` is non null and comes
+        // from a previous call to `mmap` of size `mapping_size()`
+        set_virtual_memory_region_name(
+            unsafe { std::slice::from_raw_parts(self.start_addr as *const u8, mapping_size()) },
+            Some(
+                std::ffi::CString::new(MAPPING_NAME)
```
yannham: Nit: if MAPPING_NAME is used only as a `CStr`, you can build it at const time like `const FOO: &core::ffi::CStr = c"Hello, world!";`. It's not really worth it just to avoid a `CString` allocation once in the process lifetime, but one little bonus is that it gets rid of this `Err` that can actually never happen. But it's a detail

opentelemetry-context-sharing/src/process_context/linux.rs
```diff
@@ -0,0 +1,577 @@
...
+/// based synchronization requires the use of atomics to have any effect (see [Mandatory
+/// atomic](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#mandatory-atomic))
+///
+/// We use `monotonic_published_at_ns` for synchronization with the reader. Ideally, it should
+/// be an `AtomicU64`, but this is incompatible with `#[repr(C, packed)]` by default, as it
+/// could be misaligned. In our case, given the page size and the layout of `MappingHeader`, it
+/// is actually 8-bytes aligned: we use [`AtomicU64::from_ptr`] to create an atomic view when
+/// synchronization is needed.
+#[repr(C, packed)]
+struct MappingHeader {
```
yannham, line 55: FYI, the representation [has changed](https://github.com/DataDog/libdatadog/blob/main/libdd-library-config/src/otel_process_ctx/writer.rs) a bit in libdatadog. There is a parametrization by traits but I think we shouldn't care about it here (it's for same-process context reading, which the spec isn't really made for).

But for `MappingHeader`, the `packed` is gone. Since the header is put at the beginning of a page-aligned address, we don't need it. Which in turn allow to make `monotonic_published_at_ns` an `AtomicU64` directly, instead of using `Atomic::from_ptr`. This is a bit of a detail to be honest, and the implementation is totally fine as it is.

Note: more surprisingly maybe, `payload_size` and `payload_ptr` have been made atomic as well: this is a way to circumvent the whole [seqlock is UB in Rust](https://github.com/m-ou-se/rfcs/blob/atomic-memcpy/text/3301-atomic-memcpy.md). My understanding is that it is for readers only, and libdatadog prepares for in-process reader. eBPF reader are not concerned, and writers/publishers are not concerned either, so I think it can safely be ignored.

opentelemetry-context-sharing/src/process_context/linux.rs
```diff
@@ -0,0 +1,577 @@
...
+}
+
+/// Unmaps the region used to share the process context. If no context has ever been published,
+/// this is a no-op.
+///
+/// A call to [`publish_raw_payload`] following an [`unpublish`] will create a new mapping.
+pub(crate) fn unpublish() -> Result<(), Error> {
+    let mut guard = lock_context_handle()?;
+
+    if let Some(ProcessContextHandle { mapping, .. }) = guard.take() {
```
yannham: There's also been some [hardening about unpublishing](https://github.com/DataDog/libdatadog/blob/8ac892e2c78323d0d2046bf4996585f16fa1cfd5/libdd-library-config/src/otel_process_ctx/writer.rs#L289-L294): we now first set `monotonic_published_at_ns` to zero before anything, with a fence, then we unmap, and then we drop the payload. The idea being that this limit races between a reader looking at the payload and a publisher freeing the memory at the same time

opentelemetry-context-sharing/src/process_context/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            opentelemetry::otel_warn!(
+                name: "process_context.publish.failed",
+                message = format!("{e}")
+            );
+        }
+    }
+
+    #[cfg(not(all(target_os = "linux", target_has_atomic = "64")))]
+    {
+        let _ = payload;
```
yannham: Very nit, but we could avoid encoding the payload at all if process context isn't supported. I guess in fact you could have

```
#[cfg(not(all(target_os = "linux", target_has_atomic = "64")))]
pub fn publish(_resource: &Resource) { }
```

Or, in the style of `unpublish` below,

```
pub fn publish(resource: &Resource) {
    #[cfg(all(target_os = "linux", target_has_atomic = "64"))]
    {
        let payload = encode_process_context(resource);
        if let Err(e) = linux::publish_raw_payload(payload) {
            opentelemetry::otel_warn!(
                name: "process_context.publish.failed",
                message = format!("{e}")
            );
        }
    }
}
```

yannham (approved):

opentelemetry-context-sharing/src/process_context/linux.rs
scottgerring: oh, that's cool! avoiding `packed` is a nice upshot here; have taken this one

opentelemetry-context-sharing/src/process_context/mod.rs
scottgerring: easy enough to implement, and rust programs running on non-64-bit-linux isn't unforeseen 👍

Issue #2673, opened 2025-02-15
there is no reactor running, must be called from the context of a Tokio 1.x runtime


maxwellflitton:
I'm trying to get the batch exporter with `reqwest` working in `Tokio` but I'm getting the following error:

```
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```
I've read through docs, pull requests, and other issues and I can't find any reference on how to solve this. 

I've got the following dependencies for my logging crate:

```toml
tracing = { version = "0.1", optional = true}
opentelemetry-appender-tracing = { version = "0.28.1", optional = true}
tracing-subscriber = { version = "0.3", features = ["std", "env-filter", "fmt", "registry"], optional = true}
tracing-opentelemetry = { version = "0.29.0", optional = true}
opentelemetry = { version = "0.28.0", features = ["trace"], optional = true }
opentelemetry_sdk = { version = "0.28.0", features = ["rt-tokio", "metrics", "spec_unstable_metrics_views"], optional = true }
opentelemetry-stdout = { version = "0.28.0", optional = true }
tokio = { version = "1.43.0", features = ["full"] }  # at this point we will be using Tokio for logs for different tracing providers 
opentelemetry-otlp = { version = "0.28.0", features = ["http-json", "reqwest-client"], optional = true }
reqwest = { version = "0.12.12", features = ["json"], optional = true}
tracing-attributes = { version = "0.1.24", optional = true }
```

And below is how I `init` the log provider:

```rust
use opentelemetry_appender_tracing::layer;
use opentelemetry_otlp::{WithExportConfig, WithHttpConfig};
use opentelemetry_sdk::logs::SdkLoggerProvider;
use opentelemetry_sdk::Resource;
use std::{error::Error, sync::OnceLock};
use tracing_subscriber::{prelude::*, EnvFilter};

pub fn init_log_provider<X, T>(service_name: X, endpoint: T) -> SdkLoggerProvider
where
    X: ToString + Into<opentelemetry::Value>,
    T: Into<String>,
{
    let client = reqwest::Client::new();
    let exporter = opentelemetry_otlp::LogExporter::builder()
        .with_http()
        .with_endpoint(endpoint)
        .with_protocol(opentelemetry_otlp::Protocol::HttpJson)
        .with_http_client(client)
        .build()
        .unwrap();

    let provider = SdkLoggerProvider::builder()
        .with_batch_exporter(exporter)
[...]

PR #3462, opened 2026-04-16
feat(sdk)!: provider builder build() returns Result instead of panicking

### LINKED ISSUE #2690: When invalid combination of clients are used return appropriate Result instead of panic
Suggested here:
https://github.com/open-telemetry/opentelemetry-rust/issues/2673#issuecomment-2672364656

It might be possible for the provider construction to check if an invalid pair of Processor+Exporter client is used and return an Error...As of now, wrong combinations are not detected until too late, where things panic.

### LINKED ISSUE #3375: Shutdown paths panic on thread join failures instead of returning errors
## Summary

Several places in the batch processors panic via `.unwrap()` or `.expect()` on thread operations (spawning and joining). These panics occur in shutdown/cleanup code and in constructors, where graceful error handling is expected. If the worker thread panicked for any reason, calling `handle.join().unwrap()` during shutdown will cause a second panic, masking the original error and potentially causing cascading failures.

### 1. `BatchSpanProcessor::new()` -- thread spawn

**File:** `opentelemetry-sdk/src/trace/span_processor.rs`, line 456

```rust
.expect("Failed to spawn thread"); //TODO: Handle thread spawn failure
```

The existing TODO comment acknowledges this should be handled. If thread creation fails (e.g., resource exhaustion), the entire process panics instead of returning an error.

### 2. `BatchSpanProcessor::shutdown()` -- thread join

**File:** `opentelemetry-sdk/src/trace/span_processor.rs`, lines 688-689

```rust
if let Some(handle) = self.handle.lock().unwrap().take() {
    handle.join().unwrap();
}
```

Two panics here: `lock().unwrap()` panics if the mutex is poisoned (which happens if another thread panicked while holding the lock), and `join().unwrap()` panics if the worker thread panicked.

### 3. `BatchLogProcessor::new()` -- thread spawn

**File:** `opentelemetry-sdk/src/logs/batch_log_processor.rs`, line 488

```rust
```

Same issue as the span processor constructor, with the same TODO comment.

### 4. `BatchLogProcessor::shutdown()` -- thread join

**File:** `opentelemetry-sdk/src/logs/batch_log_processor.rs`, lines 283-284

```rust
if let Some(handle) = self.handle.lock().unwrap().take() {
    handle.join().unwrap();
}
```

Same dual-panic issue as the span processor shutdown path.

## Why This Is Problematic

1. **Masks original errors:** If the worker thread panicked due to a bug, calling `join().unwrap()` in shutdown causes a *second* panic that obscures the root cause.
2. **Cascading failures:** Panicking in shutdown paths can abort the process before other cleanup (e.g., flushing other processors) completes.
3. **Violates shutdown contract:** `shutdown()` already returns `OTelSdkResult`, so callers expect errors to be returned, not to trigger panics.
4. **Constructor panics are unrecoverable:** Panicking in `new()` on thread spawn failure gives callers no chance to fall back to an alternative (e.g., a simpler processor or a no-op).

## Existing Precedent

`PeriodicReader` in `opentelemetry-sdk/src/metrics/periodic_reader.rs` (lines 327-335) already handles thread spawn failure gracefully:

```rust
// TODO: Should we fail-fast here and bubble up the error to user?
#[allow(unused_variables)]
if let Err(e) = result_thread_creation {
    otel_error!(
        name: "PeriodReaderThreadStartError",
        message = "Failed to start PeriodicReader thread. Metrics will not be exported.",
        error = format!("{:?}", e)
    );
}
```

Although it silently degrades (which has its own issues per #3210), it at least does not panic.

### For constructors (`new()`)

Return `Result<Self, OTelSdkError>` from the constructors (or provide a fallible builder `.build()` method) so callers can handle thread spawn failures. This is a breaking API change.

### For shutdown paths

- Replace `handle.join().unwrap()` with proper error handling that converts a thread panic into an `OTelSdkError`:
  ```rust
  if let Some(handle) = self.handle.lock()
      .map_err(|e| OTelSdkError::InternalFailure(format!("Mutex poisoned: {e}")))?
      .take()
  {
      handle.join().map_err(|_| {
          OTelSdkError::InternalFailure("Worker thread panicked".into())
      })?;
  }
  ```
- This ensures the `OTelSdkResult` return type is used as intended.

## Related

- #3210 -- broader effort to propagate errors instead of silently logging them
- #1963 -- previous report of panic during shutdown (different root cause, now closed)

lazureykis:
Fixes #2690
Design discussion issue (if applicable) #2673

## Changes

Makes `build()` return `Result` for all SDK provider and batch processor builders, eliminating
two categories of panics at initialization time:

**Thread spawn failures** (`ProviderBuildError::ThreadSpawnFailed`) — `BatchSpanProcessor` and
`BatchLogProcessor` both spawn a background OS thread. On resource-exhausted systems this
`thread::Builder::spawn()` call can fail; previously it would `expect()` and crash the process.

**Async-runtime/processor mismatch** (`ProviderBuildError::AsyncRuntimeRequired`) — The batch
processor runs on a dedicated OS thread using blocking calls (since 0.28). Async HTTP clients
like `reqwest::Client` and `HyperClient` require a Tokio reactor, which isn't available on that
thread — causing a panic at export time with *"there is no reactor running"*. This is a common
pitfall because `reqwest::Client` (async) is the default in many Rust applications.

This PR surfaces that misconfiguration at build time. `SpanExporter` and `LogExporter` gain a
`requires_async_runtime() -> bool` method (default: `false`); OTLP exporters propagate this
flag from their underlying `HttpClient`. The check is best-effort: third-party exporters that
do not override the method continue to return `false` and may still panic at export time.

Build-time detection is preferred over simply documenting "don't use async clients" because:
- The async `reqwest::Client` is what most Rust developers reach for first
- The panic surfaces deep in the export path, often minutes after startup, making it hard to
  diagnose
- Third-party `HttpClient` implementations can opt into the same safety net by overriding
  `requires_async_runtime()`
- A clear `ProviderBuildError::AsyncRuntimeRequired` is better DX than a tokio reactor panic

**Removed `with_batch_exporter()` convenience method** — from both `TracerProviderBuilder` and
`LoggerProviderBuilder`. Users now build the batch processor explicitly and pass it via
`with_span_processor()` / `with_log_processor()`. This eliminates the design tension of having
a fallible operation hidden inside an infallible builder method (see design rationale below).

### API changes (breaking)

| Before | After |
|---|---|
| `TracerProviderBuilder::build() -> SdkTracerProvider` | `-> Result<SdkTracerProvider, ProviderBuildError>` |
| `LoggerProviderBuilder::build() -> SdkLoggerProvider` | `-> Result<SdkLoggerProvider, ProviderBuildError>` |
| `BatchSpanProcessorBuilder::build() -> BatchSpanProcessor` | `-> Result<BatchSpanProcessor, ProviderBuildError>` |
| `BatchLogProcessorBuilder::build() -> BatchLogProcessor` | `-> Result<BatchLogProcessor, ProviderBuildError>` |
| `TracerProviderBuilder::with_batch_exporter()` | **Removed** — use `with_span_processor()` |
| `LoggerProviderBuilder::with_batch_exporter()` | **Removed** — use `with_log_processor()` |

`ProviderBuildError` is `#[non_exhaustive]` to allow future expansion (e.g. detecting additional
invalid client combinations as noted in #2690).

### Design rationale: removing `with_batch_exporter()`

`with_batch_exporter()` was a convenience wrapper that internally constructed a
`BatchSpanProcessor` — a fallible operation (thread spawn, async-runtime validation). This
created a design problem: how should a builder method that returns `Self` report errors?

Three approaches were considered:

1. **Deferred error (`pending_error` field)** — Store the error in the builder and surface it
   at `build()`. This is what `reqwest::ClientBuilder` does. The problem: hidden state, and if
   `with_batch_exporter()` is called twice with two invalid exporters, only the last error is
   reported (the first is silently dropped).

2. **Factory closure (`PendingProcessor` enum)** — Store a closure that captures the exporter
   and defer construction to `build()`. Correct and order-preserving, but novel — no major Rust
   crate uses this pattern — and adds complexity (closure boxing, manual `Debug` impls).

3. **Remove the convenience method entirely** — Require users to build the batch processor
   explicitly via `BatchSpanProcessor::builder(exporter).build()?` and pass it to
   `with_span_processor()`. This follows the `tracing-subscriber` pattern (`.with(layer)` takes
   pre-built layers) and the broader Rust ecosystem consensus: builder setters are infallible,
   `build()` is the single fallible call site.

**Option 3** was chosen because it matches the dominant pattern across the Rust ecosystem
(`tracing-subscriber`, `axum`, `tower`, `tokio`, `rdkafka` — all take pre-built components
rather than wrapping fallible construction in convenience methods). The two-line migration is
straightforward and the explicit processor construction gives users access to `BatchConfigBuilder`
for customization — something `with_batch_exporter()` did not expose.

### Provider `build()` infallibility note

Both `TracerProviderBuilder::build()` and `LoggerProviderBuilder::build()` are currently
infallible at the provider level — the `Result` return type is prophylactic for future
validation. The only errors today come from batch processor construction
(`BatchSpanProcessor::builder().build()` / `BatchLogProcessor::builder().build()`), which
happens *before* the provider `build()` call.

`SdkTracerProvider::Default` uses `.expect()` which is safe because the default builder
has no batch processors.

### Scope

This PR covers both `TracerProvider` and `LoggerProvider` in a single change so the two providers
stay consistent and the breaking change is delivered atomically.

### Migration

```rust
// Before
let provider = SdkTracerProvider::builder()
    .with_batch_exporter(exporter)
    .build();

// After
let processor = BatchSpanProcessor::builder(exporter).build()?;
let provider = SdkTracerProvider::builder()
    .with_span_processor(processor)
    .build()?;
```

```rust
// Before (logs)
let provider = SdkLoggerProvider::builder()
    .with_batch_exporter(exporter)
    .build();

// After (logs)
let processor = BatchLogProcessor::builder(exporter).build()?;
let provider = SdkLoggerProvider::builder()
    .with_log_processor(processor)
    .build()?;
```

### DIFF
--- a/examples/logs-advanced/src/main.rs
+++ b/examples/logs-advanced/src/main.rs
@@ -17,7 +17,8 @@ fn main() {
                 .build(),
         )
         .with_log_processor(enriching_processor)
-        .build();
+        .build()
+        .expect("Failed to build logger provider");
 
     // To prevent a telemetry-induced-telemetry loop, OpenTelemetry's own internal
     // logging is properly suppressed. However, logs emitted by external components
--- a/examples/logs-basic/src/main.rs
+++ b/examples/logs-basic/src/main.rs
@@ -13,7 +13,8 @@ fn main() {
                 .build(),
         )
         .with_simple_exporter(exporter)
-        .build();
+        .build()
+        .expect("Failed to build logger provider");
 
     // To prevent a telemetry-induced-telemetry loop, OpenTelemetry's own internal
     // logging is properly suppressed. However, logs emitted by external components
--- a/examples/tracing-grpc/src/client.rs
+++ b/examples/tracing-grpc/src/client.rs
@@ -14,7 +14,8 @@ fn init_tracer() -> sdktrace::SdkTracerProvider {
     // Install stdout exporter pipeline to be able to retrieve the collected spans.
     let provider = sdktrace::SdkTracerProvider::builder()
         .with_simple_exporter(SpanExporter::default())
-        .build();
+        .build()
+        .expect("Failed to build tracer provider");
 
     global::set_tracer_provider(provider.clone());
     provider
--- a/examples/tracing-grpc/src/server.rs
+++ b/examples/tracing-grpc/src/server.rs
@@ -15,7 +15,8 @@ fn init_tracer() -> SdkTracerProvider {
     // Install stdout exporter pipeline to be able to retrieve the collected spans.
     let provider = SdkTracerProvider::builder()
         .with_simple_exporter(SpanExporter::default())
-        .build();
+        .build()
+        .expect("Failed to build tracer provider");
 
     global::set_tracer_provider(provider.clone());
     provider
--- a/examples/tracing-http-propagator/src/client.rs
+++ b/examples/tracing-http-propagator/src/client.rs
@@ -20,7 +20,8 @@ fn init_tracer() -> SdkTracerProvider {
     // For the demonstration, use `Sampler::AlwaysOn` sampler to sample all traces.
     let provider = SdkTracerProvider::builder()
         .with_simple_exporter(SpanExporter::default())
-        .build();
+        .build()
+        .expect("Failed to build tracer provider");
 
     global::set_tracer_provider(provider.clone());
     provider
@@ -31,7 +32,8 @@ fn init_logs() -> SdkLoggerProvider {
     // that prints the spans to stdout.
     let logger_provider = SdkLoggerProvider::builder()
         .with_simple_exporter(LogExporter::default())
-        .build();
+        .build()
+        .expect("Failed to build logger provider");
     let otel_layer = OpenTelemetryTracingBridge::new(&logger_provider);
     tracing_subscriber::registry()
         .with(otel_layer)
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -164,7 +164,8 @@ fn init_tracer() -> SdkTracerProvider {
     let provider = SdkTracerProvider::builder()
         .with_span_processor(EnrichWithBaggageSpanProcessor)
         .with_simple_exporter(SpanExporter::default())
-        .build();
+        .build()
+        .expect("Failed to build tracer provider");
 
     global::set_tracer_provider(provider.clone());
     provider
@@ -176,7 +177,8 @@ fn init_logs() -> SdkLoggerProvider {
     let logger_provider = SdkLoggerProvider::builder()
         .with_log_processor(EnrichWithBaggageLogProcessor)
         .with_simple_exporter(LogExporter::default())
-        .build();
+        .build()
+        .expect("Failed to build logger provider");
     let otel_layer = OpenTelemetryTracingBridge::new(&logger_provider);
     tracing_subscriber::registry().with(otel_layer).init();
 
--- a/opentelemetry-appender-log/examples/logs-basic.rs
+++ b/opentelemetry-appender-log/examples/logs-basic.rs
@@ -24,7 +24,8 @@ async fn main() {
                 .build(),
         )
         .with_simple_exporter(exporter)
-        .build();
+        .build()
+        .expect("Failed to build logger provider");
 
     // Setup Log Appender for the log crate.
     let otel_log_appender = OpenTelemetryLogBridge::new(&logger_provider);
--- a/opentelemetry-appender-log/src/lib.rs
+++ b/opentelemetry-appender-log/src/lib.rs
@@ -17,8 +17,9 @@
 //! let exporter = opentelemetry_stdout::LogExporter::default();
 //!
 //! let logger_provider = SdkLoggerProvider::builder()
-//!     .with_log_processor(BatchLogProcessor::builder(exporter).build())
-//!     .build();
+//!     .with_log_processor(BatchLogProcessor::builder(exporter).build().unwrap())
+//!     .build()
+//!     .unwrap();
 //! # }
 //! ```
 //!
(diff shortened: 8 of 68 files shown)


opentelemetry-otlp/tests/integration_test/src/trace_asserter.rs
```diff
@@ -133,14 +133,8 @@ impl SpanForest {
 
     fn get_root_spans(&self) -> Vec<&SpanTreeNode> {
         self.spans
-            .iter()
```
lazureykis, line 136: clippy fix here

scottgerring:
Hey @lazureykis thanks for opening this PR! 
#3375 talks about doing this; as its a breaking change and reasonably sprawling we need to make sure the main maintainer @cijothomas is happy with the approach before moving forward. I think structured errors on this API make sense and the only controversy is likely to be the breaking-ness, but we're still pre-release so ...

Anyway, to that end, it would be great if you could add some text to that tracking issue about how you came across this and the misery it caused to help make it clear this needs addressing.

lazureykis:
> Hey @lazureykis thanks for opening this PR! #3375 talks about doing this; as its a breaking change and reasonably sprawling we need to make sure the main maintainer @cijothomas is happy with the approach before moving forward. I think structured errors on this API make sense and the only controversy is likely to be the breaking-ness, but we're still pre-release so ...
> 
> Anyway, to that end, it would be great if you could add some text to that tracking issue about how you came across this and the misery it caused to help make it clear this needs addressing.

Thanks Scott! I played with the API to eliminate the panic using 3 different approaches and this PR is what I ended up with. Curious what @cijothomas thinks.

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust#project-status 
API and SDK are marked stable for Logs and Metrics, so no breaking changes allowed there. OTLP Exporter is the perfect candidate to apply this change to, and that is also where we see most of the issues this is trying to address.

lazureykis:
Adding helpers to the OTLP crate would only cover OTLP users and the panic in BatchLogProcessor remains reachable through the SDK.

cijothomas:
> Adding helpers to the OTLP crate would only cover OTLP users and the panic in BatchLogProcessor remains reachable through the SDK.

You are fully correct. However, my position still remains the same - we cannot take breaking change to stable components, so OTLP Exporter builder is the only place we can address now. I think that should address majority of user feedback. Fixing BatchProcessor has to wait for v2.0 where we can make breaking changes.

PR #3463, opened 2026-04-17
fix(otlp): surface transport errors at ERROR level

### LINKED ISSUE #3331: [Bug]: Improve context in BatchSpanProcessor/BatchLogProcessor error logs when otlp export fails

krisztianfekete:
Fixes #3331 

## Problem

When OTLP export fails because of a misconfigured endpoint (missing URL scheme, connection refused, wrong host), the ERROR log looks like this:

`error="Operation failed: TonicTracesClient export failed with gRPC code: Unknown"`

Which tells you nothing actionable. To find out what actually went wrong, e.g. that your URL is missing a scheme you currently have to enable DEBUG logging, where the real message finally shows up.

## After this change:

We have
`error="Operation failed: TonicTracesClient export failed with gRPC code: Unknown: transport error: invalid URI"`

Same for connect failures:

 `error="... gRPC code: Unavailable: transport error: tcp connect error: Connection refused (os error 111)"`

`tonic::Status::source()` already tells us whether the error came from the local transport stack (safe to show) or from the server (might echo auth tokens, per #3360), so we just append the source chain in the first case and keep the second case DEBUG-only. 


krisztianfekete:
The linter failure seems to be unrelated to my changes.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -500,7 +500,7 @@ impl OtlpHttpClient {
                 url = request_uri.as_str(),
                 error = format!("{e}")
             );
-            HttpExportError::new(0, "HTTP export failed: network error".to_string())
+            HttpExportError::new(0, format!("HTTP export failed: network error: {e}"))
```
lalitb: This line looks a bit risky to me. On the HTTP side, `e` can be a client error whose message includes the full URL, so this might end up surfacing secrets from the endpoint at ERROR level. Could we avoid returning the raw error text here, or at least sanitize it first?

krisztianfekete: Makes sense, I reverted this, and kept it tonic-only.

krisztianfekete:
@lalitb, can you please re-review once you have a chance? Thank you!

lalitb (approved):
LGTM. Nit to update the PR desc as it still mentions HTTP side.

krisztianfekete:
Thanks @lalitb, just updated the description! I think it's ready to be merged.

PR #3465, opened 2026-04-19
fix: prevent logging of header values in tonic debug logs

### LINKED ISSUE #3367: Auth tokens leaked via otel_debug! header logging in OTLP tonic exporter
Related to #3021 and PR #3360.

PR #3360 does a good job redacting auth tokens from export error messages and interceptor error chains. But there's a separate code path it doesn't cover: the `otel_debug!` call in `build_channel()` that logs all headers in plaintext.

In `opentelemetry-otlp/src/exporter/tonic/mod.rs`, line 256:

```rust
otel_debug!(name: "TonicChannelBuilt", endpoint = endpoint_clone, timeout_in_millisecs = timeout.as_millis(), compression = format!("{:?}", compression), headers = format!("{:?}", headers_for_logging));
```

The `headers_for_logging` variable is a `Vec<(String, String)>` returned from `parse_headers_from_env()` (line 187, 415-435). It contains the raw key-value pairs from `OTEL_EXPORTER_OTLP_HEADERS` and signal-specific header env vars, with no redaction applied. If a user sets something like `OTEL_EXPORTER_OTLP_HEADERS=authorization=Bearer my-secret-token`, the full token ends up in debug output.

This is a real concern because:
- Users often enable debug logging when troubleshooting connectivity issues, which is exactly when auth problems show up
- Debug logs frequently end up in log aggregation systems, CI output, or support tickets
- The variable is even named `headers_for_logging`, suggesting it was intended to be logged, but it was never sanitized

Suggested fix: redact header values for known sensitive header names (authorization, x-api-key, etc.) before logging, or only log header names without values. Something like:

```rust
let redacted: Vec<_> = headers_for_logging.iter()
    .map(|(k, v)| {
        let lower = k.to_lowercase();
        if lower == "authorization" || lower.contains("api-key") || lower.contains("token") {
            (k.clone(), "[REDACTED]".to_string())
        } else {
            (k.clone(), v.clone())
        }
    })
    .collect();
otel_debug!(name: "TonicChannelBuilt", ..., headers = format!("{:?}", redacted));
```

This could be addressed as a follow-up to PR #3360

rornic:
Fixes #3367 

## Changes

Prevents logging of sensitive header values in OTLP tonic exporter by only logging header keys.

### DIFF
--- a/opentelemetry-otlp/src/exporter/tonic/mod.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/mod.rs
@@ -231,7 +231,9 @@ impl TonicExporterBuilder {
 
         let compression = self.resolve_compression(signal_compression_var)?;
 
-        let (headers_from_env, headers_for_logging) = parse_headers_from_env(signal_headers_var);
+        let (headers_from_env, all_headers) = parse_headers_from_env(signal_headers_var);
+        let headers_for_logging = redact_sensitive_headers(all_headers);
+
         let metadata = merge_metadata_with_headers_from_env(
             self.tonic_config.metadata.unwrap_or_default(),
             headers_from_env,
@@ -608,6 +610,25 @@ fn parse_headers_from_env(signal_headers_var: &str) -> (HeaderMap, Vec<(String, 
     )
 }
 
+fn redact_sensitive_headers(headers: Vec<(String, String)>) -> Vec<(String, String)> {
+    headers
+        .iter()
+        .map(|(k, v)| {
+            let header_name = k.to_lowercase();
+            let value = v.to_lowercase();
+            if header_name.contains("auth")
+                || header_name.contains("api-key")
+                || header_name.contains("token")
+                || value.contains("bearer")
+            {
+                (k.clone(), "[REDACTED]".to_string())
+            } else {
+                (k.clone(), v.clone())
+            }
+        })
+        .collect()
+}
+
 /// Expose interface for modifying [TonicConfig] fields within the exporter builders.
 pub(crate) trait HasTonicConfig {
     /// Return a mutable reference to the export config within the exporter builders.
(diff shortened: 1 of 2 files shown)


opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -550,6 +552,25 @@ fn parse_headers_from_env(signal_headers_var: &str) -> (HeaderMap, Vec<(String,
...
+                || header_name.contains("api-key")
+                || header_name.contains("token")
+                || value.contains("bearer")
+            {
+                (k.clone(), "[REDACTED]".to_string())
+            } else {
+                (k.clone(), v.clone())
+            }
+        })
+        .collect()
```
lalitb: Can we also add redact for `cookie` and `set-cookie` header, as they do Carry sensitive data. And also make the test coverage better.

lalitb:
Add a Changelog entry, as this is user visible fix.

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -199,7 +199,9 @@ impl TonicExporterBuilder {
     > {
         let compression = self.resolve_compression(signal_compression_var)?;
 
-        let (headers_from_env, headers_for_logging) = parse_headers_from_env(signal_headers_var);
+        let (headers_from_env, all_headers) = parse_headers_from_env(signal_headers_var);
+        let headers_for_logging = redact_sensitive_headers(all_headers);
```
cijothomas: stepping back question - do we actually need to include header keys *and values* in the debug logs at all? If we switch it to *just* have the header names, won't that be sufficient? (since the value is coming from env variable, operators can query for it anyway, once they know which keys are picked up by the application)..

if we do that, then there is no need of the redaction at all, and it is never possible catch all sensitive headers anyway. (what is sensitive to a user might be required for another user etc.)

rornic: Yes this seems reasonable since `headers_for_logging` is not modified at all `before` it's logged - so the values must be coming straight from environment.

It was mentioned as an alternative fix in the original issue.

I'll update the PR soon when I get a chance.

cijothomas: thanks!

cijothomas (changes requested):
https://github.com/open-telemetry/opentelemetry-rust/pull/3465#discussion_r3143308297 - better to skip logging values, as discussed here.

cijothomas (approved):
Thanks!

PR #3466, opened 2026-04-21
feat(appender-tracing): propagate span name to logs

SuperFluffy:
## Changes

Propagates the tracing span name to OTEL logs.

Similar to the experimental span attributes, the span containing the event is valuable to identify where and what happened.

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -368,6 +368,11 @@ where
     // - `Some(Allowlist(set))` => copy only attributes whose keys are in `set`.
     #[cfg(feature = "experimental_span_attributes")]
     span_attributes: Option<TracingSpanAttributesInner>,
+    // Tracing-span name enrichment configuration:
+    // - `None` => disabled (default). No scope walk unless span attributes are enabled.
+    // - `Some(key)` => copy the current tracing span name using `key`.
+    #[cfg(feature = "experimental_span_attributes")]
+    span_name: Option<Key>,
 }
 
 impl<P, L> OpenTelemetryTracingBridge<P, L>
@@ -390,6 +395,8 @@ where
             _phantom: Default::default(),
             #[cfg(feature = "experimental_span_attributes")]
             span_attributes: None,
+            #[cfg(feature = "experimental_span_attributes")]
+            span_name: None,
         }
     }
 }
@@ -403,6 +410,8 @@ where
     _phantom: std::marker::PhantomData<P>,
     #[cfg(feature = "experimental_span_attributes")]
     span_attributes: Option<TracingSpanAttributesInner>,
+    #[cfg(feature = "experimental_span_attributes")]
+    span_name: Option<Key>,
 }
 
 impl<P, L> OpenTelemetryTracingBridgeBuilder<P, L>
@@ -432,12 +441,37 @@ where
         self
     }
 
+    /// Store the current tracing span's name on each emitted log record using
+    /// the given log attribute key.
+    ///
+    /// "Current" means the leaf/innermost [`tracing::span!`] active when the
+    /// event is emitted. `span_name` is the log attribute key where that
+    /// tracing span name will be stored, with the span name being the attribute
+    /// value.
+    ///
+    /// This option is independent from [`Self::with_tracing_span_attributes`]:
+    /// callers can enable tracing-span attributes only, tracing-span name only,
+    /// or both. By default, span-name enrichment is disabled and no span-name
+    /// attribute is added.
+    ///
+    /// Calling this method multiple times replaces any prior span-name key -
+    /// the last call wins.
+    ///
+    /// [`tracing::span!`]: https://docs.rs/tracing/latest/tracing/macro.span.html
+    #[cfg(feature = "experimental_span_attributes")]
+    pub fn with_tracing_span_name(mut self, span_name: &'static str) -> Self {
+        self.span_name = Some(Key::from_static_str(span_name));
+        self
+    }
+
     pub fn build(self) -> OpenTelemetryTracingBridge<P, L> {
         OpenTelemetryTracingBridge {
             logger: self.logger,
             _phantom: self._phantom,
             #[cfg(feature = "experimental_span_attributes")]
             span_attributes: self.span_attributes,
+            #[cfg(feature = "experimental_span_attributes")]
+            span_name: self.span_name,
         }
     }
 }
@@ -476,19 +510,30 @@ where
         log_record.set_severity_number(severity);
         log_record.set_severity_text(metadata.level().as_str());
 
-        // Extract tracing-span attributes if enrichment is enabled.
+        // Extract tracing-span enrichment if enabled.
         #[cfg(feature = "experimental_span_attributes")]
-        if self.span_attributes.is_some() {
+        if self.span_attributes.is_some() || self.span_name.is_some() {
             // Collect attributes from all parent spans (root to leaf), including current span
             if let Some(scope) = ctx.event_scope(event) {
+                let mut current_span_name = None;
                 for span_ref in scope.from_root() {
-                    // Access extensions inline - each span has its own extension lock
-                    let extensions = span_ref.extensions();
-                    if let Some(stored) = extensions.get::<StoredSpanAttributes>() {
-                        for (key, value) in stored.attributes.iter() {
-                            log_record.add_attribute(key.clone(), value.clone());
+                    if self.span_name.is_some() {
+                        current_span_name = Some(span_ref.metadata().name());
+                    }
+
+                    if self.span_attributes.is_some() {
+                        // Access extensions inline - each span has its own extension lock
+                        let extensions = span_ref.extensions();
+                        if let Some(stored) = extensions.get::<StoredSpanAttributes>() {
+                            for (key, value) in stored.attributes.iter() {
+                                log_record.add_attribute(key.clone(), value.clone());
+                            }
                         }
                     }
+                }
+
+                if let (Some(key), Some(value)) = (self.span_name.as_ref(), current_span_name) {
+                    log_record.add_attribute(key.clone(), AnyValue::from(value));
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -414,15 +415,21 @@ where
...
                     let extensions = span_ref.extensions();
                     if let Some(stored) = extensions.get::<StoredSpanAttributes>() {
+                        current_span_name = Some(stored.span_name);
                         for (key, value) in stored.attributes.iter() {
                             log_record.add_attribute(key.clone(), value.clone());
                         }
                     }
                 }
+                if let Some(name) = current_span_name {
+                    log_record.add_attribute(Key::new("span.name"), AnyValue::from(name));
```
cijothomas: "span.name" - Its tricky! There is no established convention, so any name we pick can cause issues. Need to see if we can find any alternate solution.

Also, what happens when nesting of spans is there?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -1520,4 +1530,71 @@ mod tests {
...
+        let inner = tracing::info_span!("inner_span");
+        let _inner_guard = inner.enter();
+        tracing::error!("test message");
+
+        provider.force_flush().unwrap();
+        let logs = exporter.get_emitted_logs().unwrap();
+        assert_eq!(logs.len(), 1);
+        let log = &logs[0];
+
+        // The span.name should be the current (leaf/innermost) span
```
cijothomas: How do we determine what innermost is what user wants?

opentelemetry-appender-tracing/src/layer.rs
SuperFluffy: > Also, what happens when nesting of spans is there?

Related to your other comment for the test - this selects the inner-most span, so the span that immediately contains the event. I think that's the most natural span to select?

Alternatively we can follow tracing-subscriber's formatting subscriber which essentially contains a list of all spans.

> "span.name" 

I agree. None of this is great. Maybe just `name`?

opentelemetry-appender-tracing/src/layer.rs
SuperFluffy: Innermost is the most natural. Alternatively I can see collecting a list of all spans (probably not ideal).

opentelemetry-appender-tracing/src/layer.rs
cijothomas: Given there is no way we can ensure conflict avoidance, could we avoid making that decision ourselves, and let user make that?
`use_span_name("whatever name.user wants")` ? 
If user does not do the above, then we ignore the span-name. If the do it, then they gives us the attribute name.

WDYT?

opentelemetry-appender-tracing/src/layer.rs
cijothomas: seems reasonable. Though I can see some users wanting ability to customize it. Probably we can start with this, and based on feedback, allow advanced capabilities.

cijothomas: could you add this as a TODO

cijothomas (changes requested):
Agree we need to provide a way to attach span name, but need to also avoid picking an attribute name ourselves. See
https://github.com/open-telemetry/opentelemetry-rust/pull/3466#discussion_r3171206775

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -368,13 +375,24 @@ where
         self
     }
 
+    /// Propagate the current (leaf) span's name to log records using the given
+    /// attribute key. Setting this enables the propagation; leaving it unset
+    /// (the default) disables it.
+    #[cfg(feature = "experimental_span_attributes")]
+    pub fn with_span_name(mut self, key: impl Into<Key>) -> Self {
+        self.span_name_key = Some(key.into());
```
SuperFluffy: Note sure in how far this matters, but `key` here should be set once and be alive until the end of the program. Does it make sense to `Box::leak::<'static>` the key/string?

opentelemetry-appender-tracing/src/layer.rs
SuperFluffy: Added this as an inline TODO in the loop under `OpenTelemetryTracingBridge::on_event`.

opentelemetry-appender-tracing/src/layer.rs
SuperFluffy: Yup, that simplifies this a lot. Added a `OpenTelemetryTracingBridgeBuilder::with_span_name` to toggle injecting spans into the attribute list.

SuperFluffy:
> Agree we need to provide a way to attach span name, but need to also avoid picking an attribute name ourselves. See [#3466 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/3466#discussion_r3171206775)

Did the changes. I guess this leaves open the question what to do if span and attribute names clash - but I don't think this crate should handle. Last time I checked the tracing crate itself just uses the last value it finds for a given key, so this is consistent.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: given this is `tracing` related, and `tracing` only has `&`static` keys, lets also accept that here?

SuperFluffy: > given this is tracing related, and tracing only has `&static` keys, lets also accept that here?

Can you clarify? Do you mean we should only have
```rust
        pub fn with_span_name(mut self, key: &'static str) -> Self {
```

cijothomas: yes. We can internally make a Key cheaply using the static str.

cijothomas:
> > Agree we need to provide a way to attach span name, but need to also avoid picking an attribute name ourselves. See [#3466 (comment)](https://github.com/open-telemetry/opentelemetry-rust/pull/3466#discussion_r3171206775)
> 
> Did the changes. I guess this leaves open the question what to do if span and attribute names clash - but I don't think this crate should handle. Last time I checked the tracing crate itself just uses the last value it finds for a given key, so this is consistent.

The clash is not handled by the OTel api/sdk, so it is left to users. (Something to be revisited, but definitely outside scope of this PR)

opentelemetry-appender-tracing/src/layer.rs
SuperFluffy: Done. It's using static string refs now. Figured you don't want to `Box::leak` any arguments to make them `'static`.

SuperFluffy:
Updated the PR and removed the feature flags since span attribute propagation was stabilized.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -418,11 +421,22 @@ where
         self
     }
 
+    /// Propagate the current (leaf) span's name to log records using the given
+    /// attribute key. Setting this enables the propagation; leaving it unset
+    /// (the default) disables it.
+    ///
+    /// This does nothing if `Self::with_tracing_span_attributes` is not set.
```
SuperFluffy: The interaction with `TracingSpanAttributes` is odd now.

Should we scrap `with_span_name` in favor of extending `TracingSpanAttributes` (kinda like `OpenOptions`)?

cijothomas: @SuperFluffy This requires some discussion to make sure we do a holistic approach. Given we are doing a release today, I reverted and kept the attribute enrichment gated behind feature flag (https://github.com/open-telemetry/opentelemetry-rust/pull/3505).

I'll proceed with the release today, and after that, will get back to this. (Sorry we'll miss span-name enrichment in this release, but we can do ad-hoc release of this crate soon, so we don't have to wait for next full release).

SuperFluffy:
@cijothomas Is this a good time to revisit this? Happy to create a new patch.

cijothomas:
I’d suggest making span-name enrichment a separate builder option from span-attribute enrichment:

```rust
let layer = OpenTelemetryTracingBridge::builder(&provider)
    .with_tracing_span_attributes(TracingSpanAttributes::all())
    .with_tracing_span_name("span.name")
    .build();
```

`with_tracing_span_name` should be independent, so users can enable span attributes only, span name only, or both. The argument should be documented as the log attribute key where the current tracing span name will be stored, not the span name itself. If users do not call this method, we should not add any span-name attribute.

SuperFluffy:
@cijothomas Let's give this another shot. :-) The change is simple enough.

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -476,20 +510,31 @@ where
         log_record.set_severity_number(severity);
         log_record.set_severity_text(metadata.level().as_str());
 
-        // Extract tracing-span attributes if enrichment is enabled.
+        // Extract tracing-span enrichment if enabled.
         #[cfg(feature = "experimental_span_attributes")]
-        if self.span_attributes.is_some() {
+        if self.span_attributes.is_some() || self.span_name.is_some() {
```
cijothomas, line 515: if just span_name is required, do we need to walk entire scope? Any easier way to grab the span name from current ?

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -432,12 +441,37 @@ where
         self
     }
 
+    /// Store the current tracing span's name on each emitted log record using
```
cijothomas, line 444: @SuperFluffy 
Thanks for continuing to work on this. This is looking better. One thing I want to discuss more:

Pretty sure we'd need to expand the support for this to let use pick inner-most or outer-most (or all 🤣 ). Could you consider that future need in mind, and see if how could we evolve?

I haven't dug into details yet, but I think we should consider a design that can accommodate future expansion without looking awkward.

An example use-case I am aware of is: When a service receives incoming request, it starts a tracing::span with name indicating overall operation. Then there are sub spans too. For the log, it may be desired to store the outermost span name, instead of inner most. I understand we agreed to start with inner-most, but could you outline how this could be expanded without ending up with confusing API ?

cijothomas:
@SuperFluffy I will also ask in OTel semantic convention group if they have thought about blessing/reserving an attribute for stamping span-name to Logs. If such a thing lands, then we can follow that.

Issue #1386, opened 2023-11-21
[Feature]: Lock-free Metrics Counting API


shaun-cox:
### Related Problems?

The general problem is one of minimizing the performance overheads of instrumentation and delivering on the [mission](https://github.com/open-telemetry/community/blob/main/mission-vision-values.md#we-value-performance).

#1374 

### Describe the solution you'd like:

If the API could take advantage of Rust type-system features to distinguish when in a "thread-local" context where things are `!Send`, then modifying metrics could be even more performant than they are today.  i.e. not baking in the assumptions of `Arc<Mutex<State>>` when `Rc<RefCell<State>>` would do.

### Considered Alternatives

<TODO>

jtescher:
👍 would be great to experiment with a `LocalMeter` or similar that could expose instruments that are not send nor sync. Could still have them get picked up by global metric readers through channels or other means of communication, would just need to be careful to aggregate them across threads as instruments could have similar names/attributes.

KallDrexx:
Ok here's a crazy thought.  Keep in mind I'm still working my way through the code base and OpenTelemetry terminology, so feel free to tell me I'm way off.

Today when you create a `Counter`, it's ultimately given the `Sum::measure()` call as it's `Measure<T>`.  `Sum::measure()` in turn calls it's internal `ValueMap::measure()`.  The `ValueMap` is ultimately what holds all the attribute sets and their associated value, and the locking mechanism.  

Right now we heavily rely on locking  because `ValueMap` is the choke point for aggregation. Bounded instruments can help alleviate this but that only works for stable attributes.  So while it may not be completely lock free, what if we can utilize atomics to reduce how often we have to lock?  

How does this look in practice?  My current thought is along the lines of:

* `Counter<T>` is created.  This creates it's own `HashMap<AttributeSet, Arc<Atomic<T>>`.  
* `Counter<T>.add()` is changed to take a mutable reference to self (meaning you must have an exclusive reference to that specific counter instance in order to call `add()`).
* Calling `Counter::add()` looks up the passed in attribute sets against it's local hash map.  
  * If we have an entry for this attribute set, call `fetch_add()` on the atomic
  * if we do not have an entry for it
    * Call `Sum::get_atomic()` which calls `ValueMap::get_atomic()`
    * This acquires a lock, looks up the attribute set in the hashmap, and returns the appropriate atomic (or creates one and returns it if one does not already exist)
[...]

PR #3473, opened 2026-04-28
perf(metrics): partition ValueMap into per-thread shards to reduce lock contention

souyang-dev:
Metrics ValueMap is protected by a RwLock. It can be a source of contention under high concurrency. What's more, the cost to update a histogram is even higher because you need to grab a mutex lock after the read lock before you can update a bucket.
There is a previous attempt to shard the ValueMap (https://github.com/open-telemetry/opentelemetry-rust/pull/1564),  but that PR partitions metrics updates based on attribute-sets,  therefore updates to the same attribute-set still land on the same shard, so it's not effective in reducing locking contention when most updates are focused on a small set of attributes.

Fixes #
This PR partitions the single HashMap into shards.  Each thread update different shards based on thread id to minimize locking contention.  If the number of shards >= number of CPUs you won't see locking contentions when updating metrics.

Here are Benchcmark results (metrics_histogram, metrics) with AMD EPYC 9555 CPUs (255 CPU threads).
The numbers in the tables are updates/second in millions.  "shards=16" is this PR with 16 shards, "shards=32" is this PR with 32 shards. 

Counter update throughput (cargo run --release --package stress --bin metrics -- N)
|number of threads | 1 | 2 | 4 | 8 | 16 |
|-- | -- | -- | -- | -- | -- |
|upstream   | 10 | 5.7 | 6.8 | 6 | 5.5|
|shards=16 | 10 | 22 | 32 | 40 | 43|
|shards=32 | 10 | 22 | 31 | 48 | 100|

Histogram update throughput (cargo run --release --package stress --bin metrics_histogram -- N)
|number of threads | 1 | 2 | 4 | 8 | 16 |
|-- | -- | -- | -- | -- | -- |
|upstream| 9.5 | 4.8 | 5.2 | 6 | 5.5 |
|shards=16| 9.5 | 19 | 28 | 40 | 42|
|shards=32 | 9.5 | 19 | 28 | 45 | 84|

## Changes

Two changes are made:
1, Partition ValueMap into shards and let different threads update different shards to minimize locking contention. 
2, Implement merge function for Aggregators, so collector can merge the shards.

* [* ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [* ] Unit tests added/updated (if applicable)

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
+++ b/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
@@ -204,6 +204,31 @@ fn scale_change(max_size: i32, bin: i32, start_bin: i32, length: i32) -> u32 {
     count
 }
 
+/// Returns the number of doubling-shrinks (downscale delta) needed so the union bin range
+/// of two bucket sets fits within `max_size` after downscaling.
+/// This is used during merge to avoid allocating huge sparse bucket vectors that's beyond
+/// the "max_size" limit when two histograms at the same scale cover widely different ranges.
+fn scale_change_for_union(max_size: i32, start1: i32, len1: i32, start2: i32, len2: i32) -> u32 {
+    if len1 <= 0 || len2 <= 0 {
+        return 0;
+    }
+    let end1 = start1 + len1 - 1;
+    let end2 = start2 + len2 - 1;
+    let mut low = start1.min(start2);
+    let mut high = end1.max(end2);
+    let mut count = 0u32;
+    while high - low >= max_size {
+        low >>= 1;
+        high >>= 1;
+        count += 1;
+
+        if count > (EXPO_MAX_SCALE - EXPO_MIN_SCALE) as u32 {
+            return count;
+        }
+    }
+    count
+}
+
 // TODO - replace it with LazyLock once it is stable
 static SCALE_FACTORS: OnceLock<[f64; 21]> = OnceLock::new();
 
@@ -265,7 +290,7 @@ fn frexp(x: f64) -> (f64, i32) {
 }
 
 /// A set of buckets in an exponential histogram.
-#[derive(Default, Debug, PartialEq)]
+#[derive(Default, Debug, PartialEq, Clone)]
 struct ExpoBuckets {
     start_bin: i32,
     counts: Vec<u64>,
@@ -373,6 +398,135 @@ where
         let mut current = self.lock().unwrap_or_else(|err| err.into_inner());
         let cloned = replace(current.deref_mut(), ExpoHistogramDataPoint::new(init));
         Mutex::new(cloned)
+    }
+
+    fn merge_to(&self, target: &Self) {
+        let src = self.lock().unwrap_or_else(|err| err.into_inner());
+        // Source is empty: nothing to merge.
+        if src.count == 0 {
+            return;
+        }
+
+        let mut dst = target.lock().unwrap_or_else(|err| err.into_inner());
+        // dest is empty: just copy the src.
+        if dst.count == 0 {
+            *dst = ExpoHistogramDataPoint {
+                max_size: dst.max_size,
+                count: src.count,
+                min: src.min,
+                max: src.max,
+                sum: src.sum,
+                scale: src.scale,
+                pos_buckets: src.pos_buckets.clone(),
+                neg_buckets: src.neg_buckets.clone(),
+                zero_count: src.zero_count,
+            };
+            return;
+        }
+
+        dst.count += src.count;
+        dst.sum += src.sum;
+        dst.zero_count += src.zero_count;
+        if src.min < dst.min {
+            dst.min = src.min;
+        }
+        if src.max > dst.max {
+            dst.max = src.max;
+        }
+
+        // Unify the scales of both sides to smaller scale before merging.
+        let target_scale = src.scale.min(dst.scale);
+
+        if dst.scale > target_scale {
+            let delta = (dst.scale - target_scale) as u32;
+            dst.pos_buckets.downscale(delta);
+            dst.neg_buckets.downscale(delta);
+        }
+        dst.scale = target_scale;
+
+        // Prepare src buckets (cloned) at the common target_scale.
+        let mut src_pos = src.pos_buckets.clone();
+        let mut src_neg = src.neg_buckets.clone();
+        if src.scale > target_scale {
+            let delta = (src.scale - target_scale) as u32;
+            src_pos.downscale(delta);
+            src_neg.downscale(delta);
+        }
+
+        // If the two (now same-scale) bucket ranges are far apart (different start_bin
+        // because the original value ranges were widely different), their union at the
+        // current resolution would exceed max_size. Pre-downscale both so the union fits.
+        let pos_delta = scale_change_for_union(
+            dst.max_size,
+            dst.pos_buckets.start_bin,
+            dst.pos_buckets.counts.len() as i32,
+            src_pos.start_bin,
+            src_pos.counts.len() as i32,
+        );
+        let neg_delta = scale_change_for_union(
+            dst.max_size,
+            dst.neg_buckets.start_bin,
+            dst.neg_buckets.counts.len() as i32,
+            src_neg.start_bin,
+            src_neg.counts.len() as i32,
+        );
+        let delta = pos_delta.max(neg_delta);
+        if delta > 0 {
+            dst.pos_buckets.downscale(delta);
+            dst.neg_buckets.downscale(delta);
+            src_pos.downscale(delta);
+            src_neg.downscale(delta);
+            dst.scale -= delta as i8;
+        }
+
+        merge_expo_buckets(&src_pos, &mut dst.pos_buckets);
+        merge_expo_buckets(&src_neg, &mut dst.neg_buckets);
+
+        // Final safety net (in case alignment after downscale adds an extra bucket).
+        while dst.scale > EXPO_MIN_SCALE
+            && (dst.pos_buckets.counts.len() as i32 > dst.max_size
+                || dst.neg_buckets.counts.len() as i32 > dst.max_size)
+        {
+            dst.pos_buckets.downscale(1);
+            dst.neg_buckets.downscale(1);
+            dst.scale -= 1;
+        }
+    }
+}
+
+/// Merges `src` bucket counts into `dst`, aligning by bin index.
+/// Assumes both have the same scale (no downscaling needed).
+fn merge_expo_buckets(src: &ExpoBuckets, dst: &mut ExpoBuckets) {
+    if src.counts.is_empty() {
+        return;
+    }
+
+    if dst.counts.is_empty() {
+        dst.start_bin = src.start_bin;
+        dst.counts = src.counts.clone();
+        return;
+    }
+
+    // Compute the union bin range
+    let src_end = src.start_bin + src.counts.len() as i32 - 1;
+    let dst_end = dst.start_bin + dst.counts.len() as i32 - 1;
+    let new_start = src.start_bin.min(dst.start_bin);
+    let new_end = src_end.max(dst_end);
+    let new_len = (new_end - new_start + 1) as usize;
+
+    // Shift dst counts to align with new_start
+    let dst_offset = (dst.start_bin - new_start) as usize;
+    if dst_offset > 0 || new_len > dst.counts.len() {
+        let mut new_counts = vec![0u64; new_len];
+        new_counts[dst_offset..dst_offset + dst.counts.len()].copy_from_slice(&dst.counts);
+        dst.counts = new_counts;
+        dst.start_bin = new_start;
+    }
+
+    // Add src counts at their aligned offsets
+    let src_offset = (src.start_bin - new_start) as usize;
+    for (i, &c) in src.counts.iter().enumerate() {
+        dst.counts[src_offset + i] += c;
     }
 }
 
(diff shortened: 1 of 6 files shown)


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -74,23 +78,63 @@ impl<A: Aggregator> TrackerEntry<A> {
...
+}
+
+/// Number of shards = min(num_cpu, 32);
+#[cold]
+fn init_num_shards() -> usize {
+    let n = std::thread::available_parallelism()
+        .map(|n| n.get())
+        .unwrap_or(4)
+        .min(32);
+    NUM_SHARDS.store(n, Ordering::Relaxed);
```
ArniDagur: There is a benign race condition here where multiple threads will attempt to store `NUM_SHARDS` concurrently. Most of the time, `std::thread::available_parallelism` will return the same value, making this okay, however the docs say that `std::thread::available_parallelism` may technically change its value at runtime (in practice it never does).

I suggest a `OnceLock`. Something like
```
use std::sync::atomic::{AtomicUsize, Ordering};
use std::sync::OnceLock;

static THREAD_ID_COUNTER: AtomicUsize = AtomicUsize::new(0);
static NUM_SHARDS: OnceLock<usize> = OnceLock::new();

/// Number of shards = min(num_cpu, 32)
#[inline(always)]
fn num_shards() -> usize {
    *NUM_SHARDS.get_or_init(|| {
        std::thread::available_parallelism()
            .map(|n| n.get())
            .unwrap_or(4)
            .min(32)
    })
}

/// Returns a shard index for the current thread that round-robins across all shards,
/// so it's unlikely that two threads will write to the same shard.
#[inline]
fn shard_index() -> usize {
    thread_local! {
        static THREAD_SHARD_ID: usize =
            THREAD_ID_COUNTER.fetch_add(1, Ordering::Relaxed) % num_shards();
    }
    THREAD_SHARD_ID.with(|id| *id)
}
```

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -127,7 +173,10 @@ where
             return;
         }
 
-        let Ok(trackers) = self.trackers.read() else {
+        let shard_idx = shard_index();
+        let shard = &self.shards[shard_idx];
```
ArniDagur:
```suggestion
        let shard = &self.shards[shard_index()];
```
Unneccesary assignment

opentelemetry-sdk/src/metrics/internal/mod.rs
souyang-dev: Great comment!  Have updated the code.

opentelemetry-sdk/src/metrics/internal/mod.rs
souyang-dev: done.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -102,20 +135,22 @@ where
...
-                1 + min(DEFAULT_CARDINALITY_LIMIT, cardinality_limit),
-            )),
+            shards: (0..n)
+                .map(|_| RwLock::new(HashMap::with_capacity(per_shard_capacity)))
+                .collect(),
             no_attribute_tracker: TrackerEntry::new(&config),
             count: AtomicUsize::new(0),
             config,
-            cardinality_limit,
+            all_shards_cardinality_limit: per_shard_capacity.saturating_mul(n),
```
scottgerring: Pinging @cijothomas because metrics is really his area :) ! 

I'm not sure about this part but it might be a gap in my mental model. The metrics SDK spec defines `aggregation_cardinality_limit` as "the maximum number of data points allowed to be emitted in a collection cycle by a single instrument" and says that, for a given metric, the cardinality limit is a "hard limit" on the number of metric points collected during a collection cycle:

   https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#cardinality-limits

 It looks like with this change, the effective limit becomes roughly `configured_limit * num_shards` where `num_shards` is the smaller of CPU count or 64, so a configured limit of 2000 can produce many more than 2000 points. 

I think sharding needs to remain an internal implementation detail and preserve the configured stream cardinality limit. The user-visible impact is that a limit chosen to cap SDK memory usage and downstream time-series cardinality would no longer provide that cap; the same configuration could produce very different numbers of exported
 series depending on CPU count.

cijothomas: Previous shard attempts were also blocked on this same problem. It is non-trivial to solve with sharding approach, unfortunately. We'd need to solve contention issue differently where consciously opts in to this model. (I'll write what I have in my mind soon)

souyang-dev: Thank you @scottgerring for the comments.
I have updated the patch to enforce cardinality limit across all shards.

opentelemetry-sdk/src/metrics/internal/last_value.rs
```diff
@@ -39,6 +39,10 @@ where
             value: T::new_atomic_tracker(self.value.get_and_reset_value()),
         }
     }
+
+    fn merge_to(&self, target: &Self) {
```
cijothomas, line 73: what does this mean for the semantics? If 2 gauge.set() calls were done, we currently keep the last_value. Would that semantic be preserved when each shard stores its own value - do we guarantee the last_value semantic is preserved?

souyang-dev: Good point. 

Some aggregators are order dependent (gauge), i.e. the last update wins.  I plan to allocate only 1 shard in the ValueMap,  so all updates land in the only shard and they are serialized by the mutex.
For many aggregators (counter, histogram)  the merge() order doesn't matter.  My existing code works well.

I will upload a patch soon to address your comment.

souyang-dev: Have uploaded a new commit to address the in-order-update of guage.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -231,42 +331,68 @@ where
...
+                };
+                for (attrs, tracker) in trackers.iter() {
+                    if !seen.insert(Arc::as_ptr(tracker)) {
+                        continue; // skip duplicated attribute-sets within a shard.
+                    }
                     if tracker.has_been_updated.swap(false, Ordering::Acquire) {
-                        dest.push(map_fn(attrs.clone(), &tracker.aggregator));
+                        // Use sorted order so the same attribute-sets from different shards can be merged.
+                        let sorted = sort_and_dedup(attrs);
+                        let cloned = tracker.aggregator.clone_and_reset(&self.config);
```
lalitb: I think this reset can race with concurrent `measure()` calls.

  Simple example for a delta counter:

  1. counter value is 10, and `has_been_updated` is true
  2. collector changes `has_been_updated` (line 347 above) from true to false
  3. another thread records 1 more measurement, so the counter becomes 11 and `has_been_updated` becomes true again
  4. collector reaches this line, exports 11, and resets the counter to 0
  5. next collection still sees `has_been_updated = true`, but the counter is already 0, so it can export a fake zero delta

  So the measurement is probably not lost, but it can be counted in the wrong collection interval, and the next interval can get a phantom zero/empty datapoint. Since this PR now resets the live tracker while
  measurements can still update it, I think this needs stronger synchronization around the flag + reset operation.

  One possible fix direction would be to avoid resetting a tracker that `measure()` can still update. For example, move/swap the shard contents under the shard write lock for delta collection, or otherwise hold synchronization that makes the `has_been_updated` check and `clone_and_reset()` act as one atomic collection step.

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
```diff
@@ -338,6 +338,68 @@ where
...
+        let mut dst = target.lock().unwrap_or_else(|err| err.into_inner());
+
+        if src.count == 0 {
+            return;
+        }
+
+        // Assumes src and dst have the same max_size.
+        // The dst is the target of merge. It starts with max scale (finest resolution).
+        // If src is downscaled (smaller scale, coarse resolution) we should use the min scale.
+        dst.scale = dst.scale.min(src.scale);
```
lalitb: I think this merge needs to rescale bucket indexes before adding counts. Right now `dst.scale` is set to the coarser scale, but `merge_expo_buckets` still treats `src.start_bin` / `dst.start_bin` as if both
  bucket sets are already expressed at the same scale.

  That means if one shard downscaled and another shard stayed at a finer scale, the merged histogram can put counts into the wrong buckets, and possibly emit more buckets than `max_size`. The total count may
  still be right, but the bucket distribution can be wrong.

  A test with two shard-local histograms for the same attrs where one has downscaled before merge would probably catch this.

  One possible fix direction is to downscale both bucket sets to the target scale before merging, and then enforce the configured `max_size` on the merged bucket range. In other words, the bucket offsets/counts
  need to be converted to the same scale, not just the `scale` field.

opentelemetry-sdk/src/metrics/internal/mod.rs
souyang-dev: Have fixed this issue by grabbing a shard write lock.

opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs
souyang-dev: Yes, both src and dest histogram should have the same scale before merging.
I have updated the code, and added more unittest to make sure the scale consistency.

cijothomas:
Will hold off to finish the current release. and Review when the current release is out (this week)

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -231,41 +331,50 @@ where
...
-                    let stale_pointers: HashSet<*const TrackerEntry<A>> =
-                        stale_entries.iter().map(Arc::as_ptr).collect();
-                    trackers.retain(|_, tracker| !stale_pointers.contains(&Arc::as_ptr(tracker)));
-                    self.count.fetch_sub(stale_entries.len(), Ordering::SeqCst);
-                }
-            }
+        // Rebuild cardinality from the merged result, which are active attr-sets after eviction.
+        let attrs = merged.keys().cloned().collect();
+        let mut guard = self.cardinality.lock().unwrap_or_else(|e| e.into_inner());
+        guard.known_attrs = attrs;
```
lalitb: I think the cardinality guard can still lose live attrs here.

The shard write lock protects each shard while it is being scanned, but after that shard is unlocked, a concurrent `measure()` can insert a new attr set into that shard. That tracker remains live, but it is not in `merged`, and then this line replaces `known_attrs` with only `merged.keys()`.

Could we avoid rebuilding `known_attrs` from `merged` here unless new inserts are blocked for all shards? Another option is to rebuild it from the actual shard maps while holding synchronization that prevents concurrent inserts from being forgotten.

souyang-dev: Thanks for the comment.  Have updated the patch to fix this issue.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -81,24 +92,66 @@ impl<A: Aggregator> TrackerEntry<A> {
 /// Map from attribute sets to their aggregator tracker entries.
 type TrackerMap<A> = HashMap<Vec<KeyValue>, Arc<TrackerEntry<A>>>;
 
-/// The storage for sums.
+static THREAD_ID_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+static NUM_SHARDS: OnceLock<usize> = OnceLock::new();
+
+/// Number of shards = min(num_cpu, 64);
```
cijothomas, line 99: Trying to understand the reason why we minimum use 64? Aggregators pre-allocate memory for each Metric upto its cardinality limit. So this would cause it to be done 64 times. If a user has a simple MeterProvider/low volume, we'd be taking up a lot of memory?

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -329,46 +441,71 @@ where
...
-                return;
-            };
+        let mut merged: HashMap<Vec<KeyValue>, A> = HashMap::new();
 
+        for shard in &self.shards {
             let mut seen = HashSet::new();
+            let mut stale_pointers: HashSet<*const TrackerEntry<A>> = HashSet::new();
+            // Take a write lock so that no concurrent measure() can update a
+            // tracker between the has_been_updated swap and clone_and_reset().
+            let Ok(mut trackers) = shard.write() else {
```
cijothomas, line 451: previously, we were taking read lock only and its upgraded to write lock only if we have stale entries. This allows updates to continue to work on existing attribute set.
it looks like we now have a write lock here, and it held while iterating entire hashmap, blocking updates to even existing attribute set. This would cause jitter around collection intervals.

cijothomas:
@souyang-dev 
Thanks for the work here — the contention problem and the numbers are both real! As discussed earlier/offline, I'd like to really make sure if this is the right direction for us to pursue.
 
I think we should revisit the thread-local instrument idea from #1386 (likely as an evolution of bound instruments, added recently): an opt-in !Send handle that owns its own aggregate, merged centrally at collect time. The merge_to and expo-histogram merge logic from this PR would carry over directly, so this exploration helps. (OTel Arrow project already implemented a version of this for its internal metrics! They cannot use OTel Rust sdk today due to this exact problem!)
 
Related — @martintmk did some benchmarks in #1386 It showed ~92M/s with a per-thread MeterProvider vs ~8M/s shared, with no SDK changes at all. That's comparable to the gains in this PR, achievable today by end users. I think we should also document this pattern for users who need the throughput now, while we work on the proper local-instrument API.
 
 Would you be open to redirecting the effort in that direction?

(Apologies, I haven't fully review this PR yet. These are quite complex PRs and need lot of time to thoroughly review. Which is why I want to do a quick glance, and confirm our direction before moving too deep. Hope that makes sense)

Issue #3254, opened 2025-11-22
Improving Starting Page


thomasstauffer666:
### What happened?

The start pages states:

`If you are new to OpenTelemetry, start with the [Stdout Example](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-stdout/examples/basic.rs)`

1. In this folder is a readme which does not explain how to run the example
2. you cannot just run cargo run to execute this example

Either it would be nice to fix both of this or use another example as a first example on the start page.

### OpenTelemetry API Version (i.e version of `opentelemetry` crate)

v0.31.0

### OpenTelemetry SDK Version (i.e version of `opentelemetry_sdk` crate)

v0.31.0

### Relevant log output

```shell
v0.31.0
```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #3477, opened 2026-04-30
docs: Improve metrics-basic and tracing-grpc getting started READMEs

cijothomas:
Follow-up to #3299. Expands the `examples/metrics-basic/README.md` and `examples/tracing-grpc/README.md` files using the same template adopted in #3299 for `logs-basic`, so users landing on them from the front-page Getting Started section get a consistent introduction:

- Intro
- Understanding OpenTelemetry {Metrics,Tracing}
- What This Example Does
- Note on Exporters
- Usage

For `tracing-grpc`, also clarifies the role of the `TraceContextPropagator` and explains why a cross-process example was kept as the front-page tracing entry point (the propagator setup is the *whole point* of demonstrating distributed tracing realistically).

No code changes. Towards #3254.

### DIFF
--- a/examples/metrics-basic/README.md
+++ b/examples/metrics-basic/README.md
@@ -1,13 +1,65 @@
-# Metric Instrument Usage Example
+# Getting started with OpenTelemetry Rust Metrics
 
-This example shows how to use the various metric instruments to report metrics. The
-example setups a MeterProvider with stdout exporter, so metrics can be seen on
-the stdout.
+This example demonstrates the basics of recording metrics with OpenTelemetry in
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
LGTM! Couple things we might try to increase discoverability but not blocking:

* Updating [examples/README.md](https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/README.md) which feels like its now under-selling these
* Linking more prominently to them from the docs.rs docs? E.g. [here](https://docs.rs/opentelemetry/latest/opentelemetry/) the per signal bits could link out really explicitly

PR #3479, opened 2026-04-30
docs: Improve getting-started example discoverability

cijothomas:
Follow-up to #3299 / #3477 addressing [@scottgerring's review feedback](https://github.com/open-telemetry/opentelemetry-rust/pull/3477#pullrequestreview-3402041739):

> Couple things we might try to increase discoverability but not blocking:
> - Updating `examples/README.md` which feels like its now under-selling these
> - Linking more prominently to them from the docs.rs docs?

Changes:

- **`examples/README.md`**: New top-of-page **Getting Started** section with a signal/example/description table calling out `logs-basic`, `metrics-basic`, and `tracing-grpc` as the recommended entry points, plus a pointer to the OTLP examples for production. The full per-example list is kept below as an "All examples" reference, with the redundant "this example uses following crates" boilerplate replaced by a single `Uses:` line and a real one-sentence description of what each example shows.
- **`opentelemetry/src/lib.rs`** (docs.rs landing page):
  - *Tracing* section: replaced the vague "see examples directory" with an explicit pointer to `examples/tracing-grpc` (kept the broader examples link too).
  - *Metrics* section: promoted the existing `examples/metrics-basic` link to right after the 3-step intro; removed the duplicate at the bottom.
  - *Logs* section: added a `examples/logs-basic` link (was previously missing entirely).

Also drops a stale `tracing-jaeger` entry from `examples/README.md` - the directory no longer exists.

No code changes. `cargo doc --no-deps --all-features` passes with `-D warnings`.

Towards #3254.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,78 +1,68 @@
 # Examples
 
-This directory contains some examples that should help you get start crates from `opentelemetry-rust`.
+This directory contains runnable examples for the `opentelemetry-rust` crates.
 
-## logs-basic
+## Getting Started
 
-This example uses following crates from this repo:
+If you are new to OpenTelemetry Rust, start with one of the **Getting Started**
+examples below. Each one focuses on a single signal, uses a stdout exporter,
+and can be run with `cargo run` so you can immediately see telemetry on your
+console:
 
-- opentelemetry(log)
-- opentelemetry-appender-tracing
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
LGTM 🚢

Issue #3481, opened 2026-04-30
Stabilize span attributes customization: builder API shape


cijothomas:
## Context

The `experimental_span_attributes` feature in `opentelemetry-appender-tracing`
is ready to be stabilized. Today its only API is:

```rust
.with_span_attribute_allowlist(["session.id"])  // requires cargo feature
```

When we drop the cargo feature, we also need a way to **enable** the
behavior at runtime — not every user wants to pay per-span overhead.

Two builder shapes under consideration. Pick one.

---

### Option A — two methods, allowlist implies enable

```rust
.with_span_attributes(true)                       // enable, copy all
.with_span_attribute_allowlist(["session.id"])    // implies enable + filter
```

- Two methods; subtle interaction if both called.
- `bool` argument enables conditional config: `b.with_span_attributes(cfg.enable)`.

### Option B — one method, allowlist semantics

```rust
.with_span_attributes([])                         // enable, copy all
.with_span_attributes(["session.id"])             // enable, allowlist
// (not called)                                   // disabled
```

- Single method, no boolean trap.
- Empty iterator = "enable all" sentinel — slight cognitive load.
- Conditional enable requires `if cfg.enable { b = b.with_span_attributes(...) }`.

---

In both options:
- Default is **disabled** (no per-span work).
- Last call wins if invoked twice.

cijothomas:
@codefromthecrypt Can you share your thoughts.

cijothomas:
@leghadjeu-christian Can you share you thoughts?

codefromthecrypt:
I like option b! I also do not think this should be enabled by default.

cijothomas:
Going with a variation of Option B (suggested by @utpilla ) — single `with_span_attributes` method that accepts a `SpanAttributes` value type with named constructors:

```rust
use opentelemetry_appender_tracing::layer::SpanAttributes;

// Copy all tracing-span attributes:
let layer = OpenTelemetryTracingBridge::builder(&provider)
    .with_span_attributes(SpanAttributes::all())
    .build();

// Copy only an allowlist:
let layer = OpenTelemetryTracingBridge::builder(&provider)
    .with_span_attributes(SpanAttributes::allowlist(["session.id"]))
    .build();
```

This avoids the "empty iterator = copy all" sentinel from Option B and eliminates the two-knob misuse surface from Option A (no invalid states are expressible). One setter, so "what if both are called?" simply cannot happen.

PR #3482 has been updated with this design.

[...]

Issue #3376, opened 2026-02-20
Graduate stable spec features out of experimental feature flags before 1.0


bryantbiggs:
## Summary

Several features in the `opentelemetry-sdk` (and other crates) are gated behind `experimental_*` or `spec_unstable_*` feature flags. Before the 1.0 stable release, we should audit these flags and graduate the ones that correspond to stable specification features into the default or always-available API surface. Features that genuinely correspond to unstable/experimental spec areas should remain gated but be clearly documented.

Related issues:
- #1411 — Proposal: Naming for features
- #2643 — Stabilize support for async runtimes
- #1678 — Path to Stable Release: Roadmap
- #2917 — MetricReader redesign
- #3061 — Stabilization — OTLP

### `opentelemetry-sdk`

- [ ] **`spec_unstable_metrics_views`** — Gates View-related tests and the `Stream` instrument resolution in `instrument.rs`. PR #3322 already stabilized core view functionality (changing aggregation), but this flag still exists and gates some code paths. The OTel spec marks core View functionality (customizing instrument output, aggregation selection, attribute filtering) as **Stable**. This flag should be evaluated for removal or renaming.

- [ ] **`experimental_metrics_custom_reader`** — Gates public visibility of `MetricReader` trait, `ManualReader`, and `Pipeline`. Without this flag, `reader` module is `pub(crate)` only. The Prometheus exporter (`opentelemetry-prometheus`) already depends on this flag unconditionally. The underlying spec MetricReader concept is stable, but the Rust API surface may still need redesign (#2917). Decision: graduate the reader trait or stabilize a subset of it.

- [ ] **`experimental_async_runtime`** — Gates the entire `runtime` module (`Runtime` trait, `Tokio`, `TokioCurrentThread` implementations). Also implicitly enabled by `rt-tokio` and `rt-tokio-current-thread`. Tracked in #2643. This is a Rust-specific design concern (not spec-driven) but is fundamental infrastructure. Needs resolution before 1.0.

- [ ] **`experimental_metrics_periodicreader_with_async_runtime`** — Gates the async-runtime-based `PeriodicReader`. Depends on `experimental_async_runtime`. Should be resolved alongside #2643.

- [ ] **`experimental_logs_batch_log_processor_with_async_runtime`** — Gates the async-runtime-based `BatchLogProcessor`. Depends on `experimental_async_runtime`. Should be resolved alongside #2643.

[...]

PR #3482, opened 2026-04-30
feat(appender-tracing): stabilize span attribute propagation

cijothomas:
Implements the API shape discussed in #3481.

Note: throughout this PR, **"span" refers to a [`tracing::span!`](https://docs.rs/tracing/latest/tracing/macro.span.html) from the `tracing` crate** (the source the appender consumes), **not** an `opentelemetry::trace::Span`.

## Changes

- **Removes the `experimental_span_attributes` cargo feature.** Tracing-span attribute *enrichment* (copying attributes from active `tracing` spans onto each emitted log record) is now part of the stable API surface of `opentelemetry-appender-tracing`. Note: the experimental feature was never released, so this is effectively a new feature.
- **Disabled by default.** Users who don't enable enrichment pay no per-span overhead (`on_new_span` / `on_record` / scope walk all return early).
- **Single builder method with a value type** to configure runtime behavior:

  ```rust
  use opentelemetry_appender_tracing::layer::TracingSpanAttributes;

  // Copy all tracing-span attributes onto log records:
  let layer = OpenTelemetryTracingBridge::builder(&provider)
      .with_tracing_span_attributes(TracingSpanAttributes::all())
      .build();

  // Or restrict to an allowlist:
  let layer = OpenTelemetryTracingBridge::builder(&provider)
      .with_tracing_span_attributes(TracingSpanAttributes::allowlist(["session.id"]))
      .build();
  ```

  One setter means no invalid combinations are expressible — no runtime warnings or extra feature plumbing needed.

## Terminology

The PR uses **"enrichment"** rather than "propagation". Propagation is already an overloaded term in OpenTelemetry (W3C TraceContext, baggage, etc.), and "enrichment" matches the term used by the Java and .NET SDKs for this exact pattern.

## Tests

All existing tests have been updated. Tests cover:

- `tracing_appender_span_attributes_disabled_by_default` — enrichment off by default.
- `tracing_appender_span_attributes_all_copies_all` — `TracingSpanAttributes::all()` copies all span attributes.
- `tracing_appender_span_attribute_allowlist_includes_only_named` — `TracingSpanAttributes::allowlist(...)` filters correctly.
- `tracing_appender_span_attribute_allowlist_nested_spans` — allowlist filtering across nested spans.
- `tracing_appender_span_attribute_allowlist_with_on_record` — allowlist with late-recorded fields.
- `tracing_appender_empty_allowlist_copies_nothing` — empty allowlist means "allow nothing".

## Out of Scope

- `span.name` propagation (#3466) is intentionally not part of this PR. Discussion continues separately.
- `experimental_metadata_attributes` cargo feature is unchanged; that's a separate stabilization decision.

## Refs

- Design discussion: #3481
- Tracking issue: #3376
- Original feature: #3282, #3387, #3314

* [x] Changes in public API reviewed (adds `TracingSpanAttributes` type and `with_tracing_span_attributes` builder method, removes `experimental_span_attributes` feature)

### DIFF
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -311,34 +311,34 @@ struct StoredSpanAttributes {
 ///
 /// [`tracing`]: https://crates.io/crates/tracing
 /// [`tracing::span!`]: https://docs.rs/tracing/latest/tracing/macro.span.html
-pub struct SpanAttributes(SpanAttributesInner);
+pub struct TracingSpanAttributes(TracingSpanAttributesInner);
 
-enum SpanAttributesInner {
+enum TracingSpanAttributesInner {
     All,
     Allowlist(HashSet<Cow<'static, str>>),
 }
 
-impl SpanAttributes {
+impl TracingSpanAttributes {
     /// Copy **all** tracing-span attributes onto log records.
     pub fn all() -> Self {
-        Self(SpanAttributesInner::All)
+        Self(TracingSpanAttributesInner::All)
     }
 
     /// Copy only the tracing-span attributes whose keys are in the given
     /// allowlist.
     pub fn allowlist(keys: impl IntoIterator<Item = impl Into<Cow<'static, str>>>) -> Self {
-        Self(SpanAttributesInner::Allowlist(
+        Self(TracingSpanAttributesInner::Allowlist(
             keys.into_iter().map(Into::into).collect(),
         ))
     }
 }
 
-impl std::fmt::Debug for SpanAttributes {
+impl std::fmt::Debug for TracingSpanAttributes {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self.0 {
-            SpanAttributesInner::All => f.write_str("SpanAttributes::all()"),
-            SpanAttributesInner::Allowlist(set) => {
-                f.debug_tuple("SpanAttributes::allowlist").field(set).finish()
+            TracingSpanAttributesInner::All => f.write_str("TracingSpanAttributes::all()"),
+            TracingSpanAttributesInner::Allowlist(set) => {
+                f.debug_tuple("TracingSpanAttributes::allowlist").field(set).finish()
             }
         }
     }
@@ -404,8 +404,8 @@ where
     /// By default, enrichment is **disabled** and no per-span work is
     /// performed.
     ///
-    /// Use [`SpanAttributes::all()`] to copy every tracing-span attribute, or
-    /// [`SpanAttributes::allowlist(keys)`](SpanAttributes::allowlist) to copy
+    /// Use [`TracingSpanAttributes::all()`] to copy every tracing-span attribute, or
+    /// [`TracingSpanAttributes::allowlist(keys)`](TracingSpanAttributes::allowlist) to copy
     /// only named attributes.
     ///
     /// Calling this method multiple times replaces any prior configuration —
@@ -413,10 +413,10 @@ where
     ///
     /// [`tracing`]: https://crates.io/crates/tracing
     /// [`tracing::span!`]: https://docs.rs/tracing/latest/tracing/macro.span.html
-    pub fn with_span_attributes(mut self, span_attributes: SpanAttributes) -> Self {
+    pub fn with_span_attributes(mut self, span_attributes: TracingSpanAttributes) -> Self {
         self.span_attributes = Some(match span_attributes.0 {
-            SpanAttributesInner::All => HashSet::new(),
-            SpanAttributesInner::Allowlist(set) => set,
+            TracingSpanAttributesInner::All => HashSet::new(),
+            TracingSpanAttributesInner::Allowlist(set) => set,
         });
         self
     }
@@ -574,7 +574,7 @@ const fn severity_of_level(level: &Level) -> Severity {
 #[cfg(test)]
 mod tests {
     use crate::layer;
-    use crate::layer::SpanAttributes;
+    use crate::layer::TracingSpanAttributes;
     use opentelemetry::logs::Severity;
     use opentelemetry::trace::TracerProvider;
     use opentelemetry::trace::{TraceContextExt, TraceFlags, Tracer};
@@ -1190,7 +1190,7 @@ mod tests {
             .build();
 
         let layer = layer::OpenTelemetryTracingBridge::builder(&provider)
-            .with_span_attributes(SpanAttributes::all())
+            .with_span_attributes(TracingSpanAttributes::all())
             .build()
             .with_filter(tracing_subscriber::filter::filter_fn(|meta| {
                 // Allow spans at any level (needed for on_new_span to store span attributes),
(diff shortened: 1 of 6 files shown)


opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -357,9 +355,41 @@ where
...
+    ///
+    /// `with_span_attribute_allowlist` only takes effect if enrichment is
+    /// also explicitly enabled via this method. Calling
+    /// `with_span_attribute_allowlist` without also calling
+    /// `enable_span_attributes(true)` is a no-op (a warning is emitted via
+    /// `otel_warn!` from `build()`).
+    ///
+    /// [`tracing`]: https://crates.io/crates/tracing
+    /// [`tracing::span!`]: https://docs.rs/tracing/latest/tracing/macro.span.html
+    pub fn enable_span_attributes(mut self, enable: bool) -> Self {
```
utpilla: I think we can simplify the current two-method design (`enable_span_attributes(bool)` + `with_span_attribute_allowlist(keys)`) and make the builder harder to misuse.

The design issue (#3481) laid out two options, both of which share the property that calling the allowlist method *implies* enable, so there's no invalid "allowlist without enable" state. This PR took a different direction by making the two methods fully orthogonal, which introduced that invalid state. I think we'd be better served by something closer to what was originally proposed in the issue.

**Why the current shape feels like it could be simpler:**

There are only three valid configurations: off (default), on-copy-all, on-with-filter. The current API exposes two independent knobs whose combination produces a fourth state ("allowlist set but enrichment not enabled") that doesn't do anything useful. To handle that, the PR:

1. Adds a runtime `otel_warn!` in `build()` to catch the misconfiguration.
2. Introduces `internal-logs` cargo feature plumbing on this crate just to surface that warning.
3. Adds dedicated tests for the no-op case and for order-independence, both of which exist only because the API allows a state that ideally wouldn't be reachable.

Generally, needing a runtime warning to catch a builder misconfiguration is a hint that we might be able to reshape the API so that the invalid state simply can't be expressed.

Similarly, every call site in the PR is `.enable_span_attributes(true)`. The `false` argument would mainly be useful for undoing a prior `true` on the same builder chain, which isn't a pattern we'd expect in practice. A no-arg opt-in method might feel more natural here.

Here are two alternatives that are closer to the spirit of what was discussed in #3481. Either would be an improvement over the current shape:

**Option 1 (my preference): two self-descriptive methods, no bool, no warning needed**

This is a refinement of Option A from #3481, where calling the allowlist method implies enable:

```rust
// Off by default. Opt in by calling one of:

// Copy ALL tracing-span attributes:
let layer = OpenTelemetryTracingBridge::builder(&provider)
    .with_all_span_attributes()
    .build();

// Copy only an allowlist:
let layer = OpenTelemetryTracingBridge::builder(&provider)
    .with_span_attribute_allowlist(["session.id"])
    .build();
```

- Calling either method enables enrichment, no separate enable call needed.
- If both are called, last-call-wins (standard Rust builder convention; `tokio`, `reqwest`, etc. all work this way).
- No invalid combinations, no runtime warning, no `internal-logs` plumbing needed, no order-independence concerns.
- Migration text becomes simpler: "drop the cargo feature, add `.with_all_span_attributes()` (or keep your existing `.with_span_attribute_allowlist(...)`)."

**Option 2: single method with a value type**

This is similar to Option B from #3481, but uses named constructors instead of an empty-vs-non-empty iterator sentinel:

```rust
pub struct SpanAttributes(/* private */);

impl SpanAttributes {
    pub fn all() -> Self { ... }
    pub fn allowlist(keys: impl IntoIterator<Item = impl Into<Cow<'static, str>>>) -> Self { ... }
}

let layer = OpenTelemetryTracingBridge::builder(&provider)
    .with_span_attributes(SpanAttributes::all())
    .build();
```

- One method, so "what if both?" can't happen. There's only one setter.
- Avoids the "empty iterator = copy all" sentinel from issue Option B, which had some cognitive load.
- Slightly more verbose at the call site and introduces one small public type.

I'd lean toward **Option 1**. It's the smallest delta from the current PR, the most ergonomic at the call site, and removes an entire class of misconfiguration without extra plumbing. But happy to hear other perspectives. What do you think?

cijothomas: I picked option2, as it is relatively easy to expand it for future needs. PTAL.

cijothomas:
@codefromthecrypt @leghadjeu-christian Please review if you have some bandwidth!

leghadjeu-christian (approved):
@cijothomas  LGTM, thanks for this

TommyCpp (approved):

opentelemetry-appender-tracing/src/layer.rs
```diff
@@ -409,9 +456,8 @@ where
         log_record.set_severity_number(severity);
         log_record.set_severity_text(metadata.level().as_str());
 
-        // Extract span attributes if feature is enabled
-        #[cfg(feature = "experimental_span_attributes")]
-        {
+        // Extract tracing-span attributes if enrichment is enabled.
+        if self.span_attributes.is_some() {
```
lalitb, line 460: Since this feature is moving from experimental to stable, should we be defining or controlling what happens when attribute keys conflict?

For example, a span and the event can both have `user_id`, or parent and child spans can use the same key. This block adds span attributes first, and `event.record(&mut visitor)` adds event attributes later. Since `add_attribute` only appends, the exported log can contain duplicate keys.

lalitb (approved):
LGTM, once we agree on the collision-policy.

opentelemetry-appender-tracing/src/layer.rs
cijothomas: Good point.

The appender does not de-duplicate and relies on the SDK to do it. Unfortunately the SDK does not do this either today, so duplicates will be present. This is a pre-existing problem — even a single `tracing` macro can specify the same field twice, and we just append them all.

That said, you're right that span-attribute enrichment makes collisions meaningfully more likely than the existing event-only path: now a key set anywhere in the active span scope can collide with the event's own field, instead of only colliding within a single macro invocation.

Two things I see us doing in the future:

1. **OTel SDK should offer dedup.** It's required by spec, and users should be able to opt into paying that cost in the SDK so *all* appenders/instrumentations benefit, not just this one.
2. **Per-appender policy for enrichment.** Give this appender an option to control dedup behavior specifically for span-attribute enrichment, since that's the place where duplicates are most likely to be introduced.

Do we consider these blockers? If yes, we should keep the feature behind `experimental_*` until (1) and/or (2) land. My lean is to track them as follow-ups rather than block stabilization, but happy to go either way.

lalitb: Yes, not a blocker as it's pre-existing problem - just the surface area / probability is increased with making this stable - good to track this as follow-up as solution would be independent on this feature.

cijothomas: Filed follow-ups:

- #3497 — Logs SDK: deduplicate `LogRecord` attribute keys
- #3498 — appender-tracing: configurable collision policy for span-attribute enrichment

Resolving this thread.

codefromthecrypt: I'm a bit biased being former zipkin guy who didn't like the idea of attributes being allowed to repeat ;) last in wins is the zipkin policy.

Issue #2505, opened 2025-01-10
[Feature]: Implement MetricProducer


divergentdave:
### Related Problems?

This was removed in #2105 to reduce scope prior to 1.0.

### Describe the solution you'd like:

I wanted to file a feature request for the eventual re-implementation of this part of the spec, and share my use case for it.

I'm using `MetricProducer` from `opentelemetry_sdk` v0.24.1 to adapt runtime metrics from the Tokio async runtime library, and to export them alongside application-defined metrics that use the OpenTelemetry SDK in a more traditional way. My main reason for using a `MetricProducer` is that it lets me create a `HistogramDataPoint` from already-aggregated histogram counter values produced by the Tokio library. Each Tokio worker thread has an array of histogram counters representing different buckets, and these counters are incremented during performance-sensitive operations by the library. When `MetricProducer::produce()` is called, my implementation reads those per-thread bucket counts, and outputs a `HistogramDataPoint`. I cannot use SDK histogram instruments, because Tokio does not expose the elapsed time underlying this histogram to user code, only the already-aggregated histogram, and the only API that SDK histogram instruments provide is `record()`.

For reference, here's the source: https://github.com/divviup/janus/blob/e21f8663ad33902b03f5cf627b9cb6223646ad1f/aggregator/src/metrics/tokio_runtime.rs

### Additional Context

MetricProducer is defined in the [specification](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#metricproducer).

cijothomas:
Yes this was only removed to keep scope in control for Metrics SDK stable release. We are almost ready to declare Metrics SDK stable. @fraillt has some changes being worked on related to Export structures, and will impact producer implementation.
Once this is done, we can re-add Producer support back.
(If anyone has bandwidth to contribute, we are happy to help reviews etc. Please wait 1-3 weeks when we expect to close the pending discussion and declare Metrics Sdk stable)

PR #3489, opened 2026-05-03
feat(metrics): add public constructors for metric data types

### LINKED ISSUE #3469: Add public constructors for metric data types
# Add public constructors for metric data types

Hello everyone,

The `ResourceMetrics` struct and its child structs lack any public constructor, preventing users from using `PushMetricExporter::export`. `SpanData` has public fields, and `SdkLogRecord` has public trait setters, allowing users to use their corresponding exporters. Metrics is the only signal where external construction is blocked, preventing users from using a trait that is part of the public API.

Is there a specific reason for this inconsistency?

If this is to avoid breaking encapsulation, adding a public constructor or a builder would not expose any implementation details.

In my project I would like to call that trait, but not having this feature forces me to bypass the SDK entirely and reimplement the export path using `opentelemetry-proto` directly.

I am offering to implement this change. My idea is to implement this with a builder pattern, something like:

```rust
let data_point = GaugeDataPoint::new(42.0)
    .with_attributes(vec![KeyValue::new("host", "localhost")]);

let gauge = Gauge::new(vec![data_point]);

let metric = Metric::new("my_metric", MetricData::Gauge(gauge))
    .with_description("A custom gauge")
    .with_unit("ms");

let scope_metrics = ScopeMetrics::new(InstrumentationScope::builder("my.library").build())
    .with_metrics(vec![metric]);

let resource_metrics = ResourceMetrics::new(Resource::builder().build())
    .with_scope_metrics(vec![scope_metrics]);

exporter.export(&resource_metrics).await?;
```

darklight3it:
Fixes #3469

## Changes

Added public constructors to metric data types, allowing users to use the public `PushMetricExporter::export`.

Rationale:
- Types with one or more optional fields use the `builder` pattern, where the builder explicitly lists mandatory fields. This matches the established pattern in `opentelemetry-sdk`, `opentelemetry-otlp`, and other crates.
- Types with only required fields use a `new` method.

Summary:

| Type | Constructor | Required fields | Optional (`with_*`) fields |
|---|---|---|---|
| `ResourceMetrics` | `builder()` | — | `resource`, `scope_metrics` |
| `ScopeMetrics` | `builder()` | — | `scope`, `metrics` |
| `Metric` | `builder(name, data)` | `name`, `data` | `description`, `unit` |
| `Gauge<T>` | `builder(data_points, time)` | `data_points`, `time` | `start_time` |
| `GaugeDataPoint<T>` | `builder(value)` | `value` | `attributes`, `exemplars` |
| `Sum<T>` | `new(...)` | `data_points`, `temporality`, `is_monotonic`, `start_time`, `time` | — |
| `SumDataPoint<T>` | `builder(value)` | `value` | `attributes`, `exemplars` |
| `Histogram<T>` | `new(...)` | `data_points`, `temporality`, `start_time`, `time` | — |
| `HistogramDataPoint<T>` | `builder(count, sum, bounds, bucket_counts)` | `count`, `sum`, `bounds`, `bucket_counts` | `attributes`, `min`, `max`, `exemplars` |
| `ExponentialHistogram<T>` | `new(...)` | `data_points`, `temporality`, `start_time`, `time` | — |
| `ExponentialHistogramDataPoint<T>` | `builder(count, sum, scale, zero_count, positive_bucket, negative_bucket)` | `count`, `sum`, `scale`, `zero_count`, `positive_bucket`, `negative_bucket` | `attributes`, `min`, `max`, `zero_threshold`, `exemplars` |
| `ExponentialBucket` | `new(offset, counts)` | `offset`, `counts` | — |
| `Exemplar<T>` | `builder(value, time)` | `value`, `time` | `filtered_attributes`, `span_id`, `trace_id` |

Example usage:

```rust
use opentelemetry::{InstrumentationScope, KeyValue};
use opentelemetry::time::now;
use opentelemetry_sdk::metrics::data::*;
use opentelemetry_sdk::metrics::{Temporality, PushMetricExporter};

let gauge_dp = GaugeDataPoint::builder(42.0_f64)
    .with_attributes(vec![KeyValue::new("host", "localhost")])
    .build();

let gauge = Gauge::builder(vec![gauge_dp], now()).build();

let metric = Metric::builder("cpu.usage", AggregatedMetrics::F64(MetricData::Gauge(gauge)))
    .with_description("CPU usage percentage")
    .with_unit("%")
    .build();

let scope_metrics = ScopeMetrics::builder()
    .with_scope(InstrumentationScope::builder("my-library").build())
    .with_metrics(vec![metric])
    .build();

let resource_metrics = ResourceMetrics::builder()
    .with_resource(Resource::builder().with_service_name("my-service").build())
    .with_scope_metrics(vec![scope_metrics])
    .build();

exporter.export(&resource_metrics).await?;
```


lalitb:
@darklight3it  -   Can you clarify the intended use case for constructing `ResourceMetrics` and calling `PushMetricExporter::export()` directly?

In the normal metrics pipeline, instrumentation libraries create measurements through instruments, and the SDK/MetricReader produces `ResourceMetrics` for the exporter. So the example shared in the desc looks like a lower-level/exporter use case rather than normal instrumentation.

 Is the target use case pre-aggregated metrics from an external source, or directly sending measurements to an exporter while bypassing SDK aggregation? If it is the latter, I am not sure `ResourceMetrics` constructors are the right API shape, since these types represent already-aggregated metric points, not raw measurements. The spec direction for third-party/pre-processed metric sources seems closer to a [MetricProducer](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricproducer) plugged into a `MetricReader`.

darklight3it:
Yes, the target use case is pre-aggregated metrics from an external source. The aggregation is already done, the Rust side only needs to serialize and transport the resulting `ResourceMetrics` via OTLP.

The challenge I'm running into is that `MetricExporter::export()` is public and intended to be callable externally, but its argument (`ResourceMetrics`) can't be constructed outside the crate. 

This makes it difficult to use the exporter trait independently of the SDK's internal reader pipeline. I think opening up `ResourceMetrics` would better support the modularity that motivated having `opentelemetry-otlp` as a separate crate and would bring metrics in line with spans and logs, where `SpanData` and `LogRecord` already have public fields and work well for custom exporter implementations.

lalitb:
Thanks, that use case makes sense.

 I think this is closer to the spec’s `MetricProducer` path, though. MetricProducer is meant for third-party/pre processed metric sources, and it plugs into a MetricReader. So the shape would be more like: external aggregated source -> MetricProducer -> MetricReader -> exporter.

That makes me hesitant to expose constructors for the full ResourceMetrics tree and encourage calling `PushMetricExporter::export()` directly. It feels like we would be making a lot of SDK metric data internals part of the stable public construction API. I’d prefer we discuss a smaller API for this use case, maybe `MetricProducer`-style, instead of opening up all these data types as-is.

darklight3it:
Thanks @lalitb  for the thoughtful feedback, I appreciate the discussion.

On `MetricProducer`:

-  **It's not currently available**  Removed in #2105 for 1.0 scope (#1719), planned post-stable but not yet reintroduced (#2505, draft #3025).
- **It's a pull model, my use case is push.** My metrics arrive already aggregated and need to be exported immediately. There's no Rust-side `MetricReader` — adding one would introduce buffering and timer latency for what is a direct pass-through.

On the concern about opening up the full `ResourceMetrics` tree:

- **The builder doesn't expose fields directly.** It keeps internal data behind public methods, the same pattern as the existing getters (and in other types throughout the repo). The stability surface is symmetric with what's already committed: `export()` is public, the getters are public, `ResourceMetrics::Default()` already allows external construction. The builder just lets you construct a populated instance.

- **The PR does touch nested types (ScopeMetrics, Metric, etc.) with the same builder pattern.** But the same argument applies uniformly: each of these types already has public getters that expose their shape. The builders don't add new surface area, they provide the construction counterpart to what's already readable. New fields and new subtypes should already be maintained like this for the reading part.

- If maintenance is a concern, a derive macro (like `typed-builder`) could auto-generate the builders from the struct definitions, keeping them in sync automatically. I considered that approach but chose hand-written builders to **avoid adding a proc-macro dependency to the crate**, which I think is the right choice for a foundational library like this.

lalitb:
I think `MetricProducer` is still the closer conceptual fit - and it would be better to invest in bringing this feature back.

The “pull” part there is only inside the SDK: the MetricReader asks the producer for already-aggregated data. It does not mean the final export has to be pull-based. The final exporter can still be a push exporter sending OTLP.

So the shape would be: external pre-aggregated source -> MetricProducer -> MetricReader -> PushMetricExporter.

For the immediate-export case, we may need an API that lets the reader collect/export on demand, but I do not think that means `PushMetricExporter::export` should become the public ingestion path or that we should expose builders for the whole `ResourceMetrics` tree.

darklight3it:
I understand the preference for an internal path, but if I'm reading the proposed workflow correctly, the changes in this PR are still necessary for that to work.

```rust
// simplified MetricProducer shape to illustrate the point
struct MyProducer { buffer: Mutex<Vec<ScopeMetrics>> }

impl MetricProducer for MyProducer {
    fn produce(&self) -> Vec<ScopeMetrics> {
         self.buffer.lock().unwrap().drain(..).collect()
     }
 }

fn receive_external_metrics(data: ExternalData) {
     let scope_metrics: Vec<ScopeMetrics> = data.into(); // needs constructors for
                                                          // ScopeMetrics, Metric,
                                                          // Sum/Gauge/Histogram,
                                                          // DataPoint
     PRODUCER.buffer.lock().unwrap().extend(scope_metrics);
     READER.force_collect(); // on-demand collect to control when data is flushed
 }
```

Either way, users will need a supported way to create these types from external data, which means exposing some form of construction contract. We could introduce a parallel set of public DTO types and map them to the `ResourceMetrics` tree internally, but given that these types are already partially public (getters, `Default`, trait signatures), creating a separate anticorruption layer won't help without breaking the contract and closing these types.

The same reasoning applies to the `PushMetricExporter`.  It's public but currently unusable since its argument can't be constructed. If the intent is for it not to be called directly, it may be worth reconsidering whether it should remain public in a separate crate.

cijothomas:
> The same reasoning applies to the PushMetricExporter. It's public but currently unusable since its argument can't be constructed. If the intent is for it not to be called directly, it may be worth reconsidering whether it should remain public in a separate crate.


The `PushMetricExporter` trait is public because users need to **implement** it (write custom exporters), not because they are expected to call `export()` directly. The SDK's `PeriodicReader` constructs `ResourceMetrics` internally and calls `export()` on the exporter. There are already several implementations across crates (`opentelemetry-otlp`, `opentelemetry-stdout`, `InMemoryMetricExporter`) — the trait living in a separate crate is by design, not an oversight. Implementors receive `ResourceMetrics`; they don't need to construct it.

That said, I do acknowledge that when we bring back `MetricProducer`, we may well end up needing public constructors for some of these metric data types — a producer would need to build `ScopeMetrics`, `Metric`, data points, etc. It's just too early to commit to that API surface right now before `MetricProducer` itself is designed and landed.

Would you have bandwidth to help bring `MetricProducer` back? That would be the right foundation for your use case and would naturally inform which constructors are actually needed.

darklight3it:
Hello @cijothomas,

thanks for the answer and sorry for the delay,  my last two weeks were very intense.

> That said, I do acknowledge that when we bring back MetricProducer, we may well end up needing public constructors for some of these metric data types — a producer would need to build ScopeMetrics, Metric, data points, etc. It's just too early to commit to that API surface right now before MetricProducer itself is designed and landed.
> Would you have bandwidth to help bring MetricProducer back? That would be the right foundation for your use case and would naturally inform which constructors are actually needed.

I'm happy to help bring `MetricProducer` back if it unblocks my use case. But before I commit, I want to understand the intended architecture better. Unless you're thinking of duplicating those types or introducing a breaking change, the moment those types get public constructors `PushMetricExporter` also becomes automatically a fully usable extension point, falling back to the case I originally presented. What options were you thinking of?

Secondly, my use case requires that all of those types are constructable. As I said before, I already have the entire payload constructed and aggregated, I only need to map it to those types and export it.

michaelvanstraten:
I would like to use the [opentelemetry-otlp](https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/) exporters outside of the SDK context (e.g. not have them attached to a `Log/Metrics` processor). 

Going this way I would need to make `SdkLogRecord` publicly construct-able, feels kind of weird to break the `Logger` API that way. 

@darklight3it would it work for you if we added a `proto_export` methods to the `opentelemetry_otlp` exporters which would accept a type from [opentelemetry-proto](https://docs.rs/opentelemetry-proto/latest/opentelemetry_proto/) (e.g. https://docs.rs/opentelemetry-proto/latest/opentelemetry_proto/tonic/metrics/v1/struct.ResourceMetrics.html)?

darklight3it:
Hello @michaelvanstraten,

Thanks for your answer ❤️.

That would work, however if a user needs to re-derive the entire model from `opentelemetry-proto` types, it becomes more convenient to just reimplement a custom OTLP exporter from scratch rather than using the one provided out of the box. 

At that point the `proto_export` method doesn't save much, you're already doing half of the job rather than using a pluggable component and its interface.

michaelvanstraten:
> That would work, however if a user needs to re-derive the entire model from `opentelemetry-proto` types, it becomes more convenient to just reimplement a custom OTLP exporter from scratch rather than using the one provided out of the box.

Reimplementing a custom OTLP exporter is what I ended up doing (HTTP + GRPC under 1000 lines clean). However, the modeling from `opentelemetry-proto` types is not necessary since there are predefined transforms in the crate that already do this for you e.g. you can just call `.into()` on a [`ResourceMetric`](https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/metrics/data/struct.ResourceMetrics.html) to get the proto type.

ChihweiLHBird:
This would unblock [Spin](https://github.com/spinframework/spin/issues/3598), where WASI guests submit pre-aggregated `ResourceMetrics` that the host must forward through `PushMetricExporter`. Spin is currently stuck on OTel 0.29 because newer versions expose `export()` but no supported way to construct its input.

A better long-term solution might be a resource-aware `MetricProducer` API though, which is a producer that can return externally generated metrics together with their original `Resource`, rather than attaching all produced metrics to the host `MeterProvider`’s resource.

PR #3495, opened 2026-05-06
docs(metrics): document experimental bound instruments

cijothomas:
Documents the experimental bound instruments feature in docs/metrics.md.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -30,7 +30,7 @@ Status: **Work-In-Progress**
 * [Metrics Correlation](#metrics-correlation)
 * [Modelling Metric Attributes](#modelling-metric-attributes)
 * [Common Issues Leading to Missing
-  Metrics](#common-issues-that-lead-to-missing-metrics)a
+  Metrics](#common-issues-that-lead-to-missing-metrics)
 
 </details>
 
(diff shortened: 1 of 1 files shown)


docs/metrics.md
```diff
@@ -29,7 +30,7 @@ Status: **Work-In-Progress**
 * [Metrics Correlation](#metrics-correlation)
 * [Modelling Metric Attributes](#modelling-metric-attributes)
 * [Common Issues Leading to Missing
-  Metrics](#common-issues-that-lead-to-missing-metrics)
+  Metrics](#common-issues-that-lead-to-missing-metrics)a
```
lalitb:
```suggestion
  Metrics](#common-issues-that-lead-to-missing-metrics)
```

docs/metrics.md
```diff
@@ -241,6 +242,97 @@ fn setup_metrics(meter: &opentelemetry::metrics::Meter) {
...
+
+// On the hot path, the call site already knows which handle to use —
+// no attribute lookup is performed.
+tcp_packets.add(1);
+udp_packets.add(1);
+```
+
+`BoundCounter` and `BoundHistogram` are cheap to clone, so a single bound
+handle can be shared across threads or modules without re-binding. The
+underlying tracker is reclaimed when the last clone is dropped.
```
lalitb: This part may be a bit too strong. Dropping the last bound handle does not always reclaim the tracker right away. It only drops the handle; the SDK may clean up the tracker later depending on temporality/collection. Could we soften this wording?

lalitb (approved):

PR #3496, opened 2026-05-06
feat(otlp): Improved support for OTLP JSON metrics serialization (NaN,Infinity,-Infinity)

LorenzoTettamanti:
Updates #3442 

## Changes

- Added Serialize/Deserialization for NaN,Infinity,-Infinity to the following metrics fields:

| Source | Field |Type
| ----- | ----- | ----- | 
| `HistogramDataPoint` | `sum`, `min`, `max` | `Option<f64>` |
| `HistogramDataPoint` | `explicit_bounds` | `Vec<f64>` | 
| `ExponentialHistogramDataPoint` | `sum`, `min`, `max` | `Option<f64>` | 
| `ExponentialHistogramDataPoint` | `zero_threshold` | `f64` | 
| `SummaryDataPoint` | `sum` | `f64` | 
| `NumberDataPoint::Value` | `AsDouble` | `f64` | 
| `Exemplar::Value` | `AsDouble` | `f64` | 
| `AnyValue` | `DoubleValue` | `f64` | 

- Added regression tests covering NaN/Infinity serialization/deserialization for each affected field


lazureykis:
Hi @LorenzoTettamanti — while reading the `Option<f64>` path I hit what looks like a round-trip regression for histograms and wanted to check it with you.

`HistogramDataPoint`/`ExponentialHistogramDataPoint` `sum`/`min`/`max` are `Option<f64>` with no `skip_serializing_if`, so when they're `None` the output contains `"min":null` (etc.). On serialize that's fine — [`serialize_option_f64_special`](https://github.com/open-telemetry/opentelemetry-rust/blob/6fc8fb292b4a30c08a9c4e4f5f009a2a738583ee/opentelemetry-proto/src/proto.rs#L463) emits `null` for `None`, same as `main`. But [`deserialize_option_f64_special`](https://github.com/open-telemetry/opentelemetry-rust/blob/6fc8fb292b4a30c08a9c4e4f5f009a2a738583ee/opentelemetry-proto/src/proto.rs#L488) calls `deserialize_any`, and its [`OptionF64Visitor`](https://github.com/open-telemetry/opentelemetry-rust/blob/6fc8fb292b4a30c08a9c4e4f5f009a2a738583ee/opentelemetry-proto/src/proto.rs#L492) only implements `visit_f64`/`visit_u64`/`visit_i64`/`visit_str` — there's no `visit_unit`/`visit_none`. serde_json routes JSON `null` to `visit_unit`, whose default impl errors, so the value fails to parse.

The effect is that the crate can no longer deserialize its own serialized output whenever a histogram leaves `min`/`max`/`sum` unset (a common case — min/max are optional and `sum` is omitted for non-monotonic histograms). On `main` this round-trips because the default `Option<f64>` deserializer handles `null → None`.

Minimal repro (serialize a `HistogramDataPoint` with `min`/`max`/`sum` = `None`, then deserialize the result): https://github.com/lazureykis/opentelemetry-rust/blob/58640d270b674ee03d9808ef210c454c379cf0fc/opentelemetry-proto/tests/histogram_none_f64_roundtrip_evidence.rs

It serializes to `{...,"sum":null,...,"min":null,"max":null}` and then panics on deserialize:
`Error("invalid type: null, expected a float or a string representing NaN, Infinity, or -Infinity", line: 1, column: 82)`. The same test passes on `main`.

Would it make sense to have the visitor accept null — e.g. add `visit_unit`/`visit_none` returning `Ok(None)` (using `deserialize_option` so `visit_none` is reached)? Adding `skip_serializing_if = "Option::is_none"` to those fields would also stop emitting `null`, though that's a separate, pre-existing point and wouldn't cover external input that sends explicit `null`. What do you think?

LorenzoTettamanti:
Hi @lazureykis , that's a great catch. I think that's a great idea to implement the `visit_unit ` and `visit_none` in the `OptionF64Visitor` (using `deserialize_option`) so visit_none is reached for null values. Also adding `skip_serializing_if = "Option::is_none"` it's a great idea and would help to reduce the JSON

PR #3501, opened 2026-05-07
fix(otlp-http): limit response body size to prevent memory exhaustion

### LINKED ISSUE #3439: OTLP HTTP exporter buffers unbounded response bodies into memory
### What happened?

The OTLP HTTP exporter currently reads successful HTTP response bodies fully into memory before parsing partial-success responses. In the optional hyper client, non-2xx bodies are also fully collected before status handling. Because the default build uses reqwest-blocking, a malicious or compromised OTLP endpoint can return a very large response and cause memory exhaustion / DoS in the application process. The HTTP transport should enforce a small response-body limit and avoid reading non-success bodies entirely.

Refer discussion - https://github.com/open-telemetry/opentelemetry-go/security/advisories/GHSA-w8rr-5gcm-pp58#advisory-comment-180536

0.31.1

0.31.1

OTLP

### Relevant log output

```shell

```

### Tip

SATVIKsynopsis:
Fixes #3439

 ## Changes

 Bounds HTTP response body reads in all three built-in `opentelemetry-http` clients (async reqwest, blocking reqwest, hyper) to a fixed 4 MiB limit, and classifies oversized responses as non-retryable errors in the OTLP exporter.

 **Implementation:**

 - Each client streams/reads the response body incrementally and returns a `ResponseBodyTooLarge` error if the limit is exceeded.
 - `Content-Length`-based preallocation is capped at 4 MiB to prevent a malicious header from causing large allocations.
 - Non-success HTTP status codes and headers are preserved (no `error_for_status()`) so OTLP can classify 4xx/5xx and honor `Retry-After`.
 - In `opentelemetry-otlp`, `ResponseBodyTooLarge` is downcast and mapped to `NonRetryable`, satisfying the [OTLP/HTTP response
 specification](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-response) requirement that clients MUST NOT retry when unable to decode the response body.

 **Tests:**

 - Per-client body limit tests (within-limit accepted, over-limit rejected) for all three transports.
 - Status/header preservation tests for all three transports.
 - OTLP retry integration test proving exactly one attempt is made for an oversized response.


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -30,6 +30,10 @@ use crate::retry::{RetryErrorType, RetryPolicy};
 #[cfg(feature = "experimental-http-retry")]
 use crate::retry_classification::http::classify_http_error;
 
+///  This represents the maximum bytes read from an oltp hhtp response body.
```
cijothomas:
```suggestion
///  This represents the maximum bytes read from an OTLP Http response body.
```

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -30,6 +30,10 @@ use crate::retry::{RetryErrorType, RetryPolicy};
...
+/// It guards against memory exhaustion from a malicious endpoint.
+const MAX_RESPONSE_BODY_BYTES: usize = 10 * 1024 * 1024;
```
cijothomas: wondering what is a good limit to put here! 10MB feel high (just out of my intuition!) 
Any prior arts to follow?

opentelemetry-otlp/src/exporter/http/mod.rs
SATVIKsynopsis: resolved

opentelemetry-otlp/src/exporter/http/mod.rs
SATVIKsynopsis: I initially though to use a small figure like 5 mb but then thought for easier flow 10 mb could be used.  any suggestions you have on this matter?

SATVIKsynopsis: i am thinking 4-5 mb could be good to use.

cijothomas: .NET picked 4 MB. Let's pick that! 
@lalitb do you know if there is a good max limit to pick?

SATVIKsynopsis: 4 MB would be nice to go with

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -549,7 +552,17 @@ impl OtlpHttpClient {
         otel_debug!(name: "HttpClient.ExportSucceeded");
 
         // Return the response, consuming the body to save a copy
-        Ok(response.into_body())
+        let body = response.into_body();
+        if body.len() > MAX_RESPONSE_BODY_BYTES {
```
lalitb: This fix does not prevent memory exhaustion yet. By the time this code checks `body.len()`, the HTTP client has already read the full response into `Response<Bytes>`. So a malicious endpoint can still force the large allocation; this only skips parsing after the memory is already used.

A better fix is to apply the limit while reading the response body in the HTTP client implementations, like the `reqwest` and `hyper` clients in `opentelemetry-http`. They should read the body in chunks and stop once it crosses the max size. Then `OtlpHttpClient` can safely consume the already-limited Bytes.

SATVIKsynopsis: Thanks. That makes sense. I will fix it according to the suggestion

SATVIKsynopsis:
The current fix checks the size after the full body is already buffered. To truly prevent the allocation, the body needs to be read in chunks and aborted early as suggested earlier as well. I have an approach using `bytes_stream()` for reqwest async, `Read::take()` for reqwest blocking and `body.frame()` for hyper. should i go ahead with this or is there a preferred streaming pattern in this codebase i should follow?

lalitb:
> The current fix checks the size after the full body is already buffered. To truly prevent the allocation, the body needs to be read in chunks and aborted early as suggested earlier as well. I have an approach using `bytes_stream()` for reqwest async, `Read::take()` for reqwest blocking and `body.frame()` for hyper. should i go ahead with this or is there a preferred streaming pattern in this codebase i should follow?

Yes, that approach sounds right. The key requirement is that the limit is enforced while the body is being read, before it becomes `Response<Bytes>`. Using `bytes_stream()` for reqwest async, `Read::take()` for reqwest blocking, and `body.frame()` for hyper sounds like the right direction here. Please make sure each path aborts as soon as the accumulated bytes exceed `MAX_RESPONSE_BODY_BYTES`, so `OtlpHttpClient` only receives already-bounded Bytes.

Can we try to accommodate this in next release - #3508  -  @cijothomas if it can be fixed soon

SATVIKsynopsis:
Updated the fix to use chunked reading in all three HTTP client implementations as in reqwest async, reqwest blocking and hyper so the limit is enforced while the body is being read before it becomes `Response<Bytes>`. adding unit tests for the body limit in `opentelemetry-http` directly would require a mock HTTP server. happy to add one if you can suggest a preferred test crate for this codebase.

opentelemetry-http/src/lib.rs
```diff
@@ -200,11 +219,23 @@ pub mod hyper {
                     .insert(http::header::AUTHORIZATION, authorization.clone());
             }
             let mut response = time::timeout(self.timeout, self.inner.request(request)).await??;
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
-
+            let mut body_bytes = bytes::BytesMut::new();
+            let mut body = response.into_body();
+            while let Some(frame) = body.frame().await {
+                if let Ok(frame) = frame {
```
lalitb: I think we should not ignore the `Err` case here. Before this change, `collect().await?` would return the body read error to the caller. With `if let Ok(frame) = frame`, a mid-body read failure can be silently skipped and we may return a truncated body as if it succeeded.

opentelemetry-http/src/lib.rs
```diff
@@ -102,10 +103,18 @@ mod reqwest {
             otel_debug!(name: "ReqwestClient.Send");
             let request = request.try_into()?;
             let mut response = self.execute(request).await?.error_for_status()?;
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
+            let mut body_bytes = bytes::BytesMut::new();
+            while let Some(chunk) = response.chunk().await? {
+                body_bytes.extend_from_slice(&chunk);
```
lalitb: nit - can we do check before extending - 

 `body_bytes.len() + chunk.len() > MAX_RESPONSE_BODY_BYTES`

SATVIKsynopsis:
I have added the `Err` case in hyper client and resolved the nitpick.

SATVIKsynopsis:
Hi, just checking in. let me know if any further changes are needed or if there's anything blocking the review. Happy to address any feedback.

opentelemetry-http/src/lib.rs
```diff
@@ -200,11 +219,22 @@ pub mod hyper {
                     .insert(http::header::AUTHORIZATION, authorization.clone());
             }
             let mut response = time::timeout(self.timeout, self.inner.request(request)).await??;
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
-
+            let mut body_bytes = bytes::BytesMut::new();
+            let mut body = response.into_body();
+            while let Some(frame) = body.frame().await {
```
lalitb, line 260: Non-blocking/future follow-up: could we avoid reading the response body for non-success HTTP statuses? The current cap prevents unbounded memory use, but for error responses we only need the status/headers, so skipping the body would avoid extra memory and latency.

lalitb (approved):
LGTM. Since this is a behavior change in the built-in `opentelemetry-http` clients, we should add a short `vNext` changelog entry for `opentelemetry-http` before merging/releasing.

opentelemetry-http/src/lib.rs
SATVIKsynopsis: Yeah, I agree we can avoid reading bodies for non-success responses. I’ll treat that as a follow-up and add a todo/issue and implement a non-blocking fix in a separate PR once this lands.

SATVIKsynopsis:
> LGTM. Since this is a behavior change in the built-in `opentelemetry-http` clients, we should add a short `vNext` changelog entry for `opentelemetry-http` before merging/releasing.

Thanks. will add a short vNext changelog entry for before merging/releasing.

SATVIKsynopsis:
I have added the vNext changelog.

opentelemetry-http/src/lib.rs
```diff
@@ -133,6 +151,7 @@ mod reqwest {
 
 #[cfg(feature = "hyper")]
 pub mod hyper {
+    const MAX_RESPONSE_BODY_BYTES: usize = 4 * 1024 * 1024;
```
cijothomas: nit: could be declared one level to avoid duplication.

opentelemetry-http/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 Released 2026-May-08
 
 - `reqwest`'s crypto backend has changed from `ring` to `aws-lc-sys`.
+- Limit HTTP response body reads to 4 MiB in built-in HTTP clients (`reqwest` async/blocking and `hyper`) by enforcing the cap while streaming response chunks and reads that exceed the limit are aborted to prevent unbounded memory allocation.
```
cijothomas: changelog should be in vnext section.

cijothomas:
>  I also added a test with a mock client that returns an oversized body to verify the cap works.

Is that included in the PR? I am not seeing tests added.?

opentelemetry-http/src/lib.rs
```diff
@@ -200,11 +219,22 @@ pub mod hyper {
                     .insert(http::header::AUTHORIZATION, authorization.clone());
             }
             let mut response = time::timeout(self.timeout, self.inner.request(request)).await??;
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
-
+            let mut body_bytes = bytes::BytesMut::new();
```
cijothomas: could we start with some capacity other than 0 to avoid reallocation?

SATVIKsynopsis:
> > I also added a test with a mock client that returns an oversized body to verify the cap works.
> 
> Is that included in the PR? I am not seeing tests added.?

I had added a test earlier in `opentelemetry-otlp` but forgot to move it when the logic was moved to `opentelemetry-http`. I'll add coverage in `opentelemetry-http` using a small Axum test server.

opentelemetry-http/CHANGELOG.md
SATVIKsynopsis: resolved

opentelemetry-http/src/lib.rs
SATVIKsynopsis: Good point. Let me check whether we can leverage the response body size hint or content length to preallocate an appropriate capacity rather than starting from 0.

opentelemetry-http/src/lib.rs
SATVIKsynopsis: i have resolved this

opentelemetry-http/src/lib.rs
SATVIKsynopsis: i checked both implementations and was able to leverage the response size information for pre allocation and in both cases the capacity is capped at `MAX_RESPONSE_BODY_BYTES` before allocating.

opentelemetry-http/src/lib.rs
```diff
@@ -117,13 +129,22 @@ mod reqwest {
...
             let request = request.try_into()?;
             let mut response = self.execute(request)?.error_for_status()?;
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
+            let mut body_bytes = Vec::new();
+            response
+                .take(MAX_RESPONSE_BODY_BYTES as u64 + 1)
+                .read_to_end(&mut body_bytes)?;
+            if body_bytes.len() > MAX_RESPONSE_BODY_BYTES {
+                return Err("response body too large".into());
```
cijothomas: I don't think this is correct. If we return Err from here, I think this would cause OTLP Exporter to consider the request as failed, and likely initiate retry waves.
If the response is too big, we don't need to retry (we shouldn't).

opentelemetry-http/src/lib.rs
cijothomas: `let mut body_bytes = bytes::BytesMut::new();`

I don't see pre-allocation still?

opentelemetry-http/src/lib.rs
SATVIKsynopsis: good point. returning an Err here would indeed make the exporter treat it as a request failure and potentially retry so then i will take a look at how the response body is consumed by the oltp exporter and adjust the handling so oversized responses don't trigger retries.

cijothomas (changes requested):
Thanks for working on this. We need to do a more holistic approach to handle this - when response size is bigger, this PR returns Err and the OTLP Exporter would retry.
We should prevent that. Most likely, this'd mean conditionally reading the response body (we only need the body for partial response case, and then for loggging purposes.)

(Also - small note on pre-allocating the capacity for reading response).

opentelemetry-http/src/lib.rs
SATVIKsynopsis: you're right, that's still showing the previous version. the pre allocation change is currently only in my local branch using the response size hint/content length); I'll push it shortly.

cijothomas: Thanks!

SATVIKsynopsis:
Thanks for the feedback. looking through the oltp http flow, successful response bodies appear to be consumed primarily for partial success parsing so for oversized responses, would you prefer that we treat them similarly to an empty response body like skip partial success parsing and continue without retrying or would you rather move/condition the response body reading so it's only performed when needed for partial success or logging scenarios?

SATVIKsynopsis:
I have added the pre-allocation step in all three, added the http axum mock server, and the oltp exporter retry fix. All tests pass now. happy to address any feedback.

opentelemetry-http/src/lib.rs
```diff
@@ -117,13 +138,24 @@ mod reqwest {
     #[async_trait]
     impl HttpClient for reqwest::blocking::Client {
         async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
+            use std::io::Read;
             otel_debug!(name: "ReqwestBlockingClient.Send");
             let request = request.try_into()?;
             let mut response = self.execute(request)?.error_for_status()?;
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
+            let mut body_bytes = Vec::new();
```
cijothomas: this can also be pre-allocated right?

opentelemetry-http/Cargo.toml
```diff
@@ -30,6 +30,10 @@ opentelemetry = { workspace = true, features = ["trace"] }
 reqwest = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["time"], optional = true }
 
+[dev-dependencies]
+axum = { version = "0.8" }
```
cijothomas: I understand its dev-dev only, but wondering if we can avoid this ?

SATVIKsynopsis: I'll switch to a minimal raw TCP test server instead so we don't need to add axum.

SATVIKsynopsis:
I have added the pre-allocation on `reqwest-blocking` and added a minimal TCP test instead of axum.

scottgerring:
@cijothomas it looks like this one was almost over the line; is there anything we are waiting on here apart from the CHANGELOG conflict?

SATVIKsynopsis:
Thanks @scottgerring. The feedback from @cijothomas was mainly around avoiding retries on oversized response bodies, conditional body reading, pre allocation, and avoiding adding axum as a dev dependency.

I addressed those points in the follow up commits:
1.  handled the oversized response case without triggering exporter retries
2. added pre-allocation while reading response bodies
3. replaced the axum mock server with a minimal TCP test server

The remaining thing was just getting another review pass after those changes. Happy to make any further adjustments needed.

SATVIKsynopsis:
I have also resolved the merge conflict of Changelog.md file. Happy to adjust anything else if needed.

opentelemetry-http/src/lib.rs
```diff
@@ -90,22 +90,43 @@ pub trait HttpClient: Debug + Send + Sync {
...
+
+            let mut body_bytes = bytes::BytesMut::with_capacity(capacity);
+
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
+            while let Some(chunk) = response.chunk().await? {
+                if body_bytes.len() + chunk.len() > MAX_RESPONSE_BODY_BYTES {
+                    let mut http_response = Response::builder().status(status).body(Bytes::new())?;
+                    *http_response.headers_mut() = headers;
+                    return Ok(http_response);
```
scottgerring: The client has returned more than the max response size bytes - [according to the spec]( https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-response
), this should be treated as a non-retryable error:

> The client MUST limit the size of the response body when parsing it, including after decompression, to mitigate possible excessive memory usage caused by a misconfigured or malicious server. It is RECOMMENDED to use 4 MiB as the default limit. Implementations SHOULD allow this limit to be configured. **If the limit is exceeded, the client MUST treat the response as a non-retryable error and SHOULD record the fact that the response was discarded.**

I note the existing discussion here about returning `Err` triggering retries, and I can see that's how we've ended up returning an Ok here, but I don't think this is right either. We need some mechanism to communicate outwards to `export_http_once` that this _failed_, but was non-retryable. 

One option would be to introduce a new public error type `ResponseBodyTooLarge` or so, return this from here, and then try downcast in the caller; this would be a focussed change and avoid changing the signature. 

@cijothomas thoughts?

opentelemetry-http/src/lib.rs
```diff
@@ -357,3 +415,64 @@ mod tests {
         assert!(new_carrier.capacity() >= 5);
     }
 }
+
+#[cfg(all(test, feature = "reqwest"))]
```
scottgerring: The logic implemented in this PR is particular to the different HTTP clients, but the test only tests reqwest-async. We should generalise this out here:

 ```suggestion
   #[cfg(all(test, any(feature = "reqwest", feature = "hyper")))]
 ```

 then we could add some helpers like:

 ```rust
   async fn assert_exceeds_limit(client: &dyn HttpClient, addr: SocketAddr) {
       let request = Request::builder()
           .method("POST")
           .uri(format!("http://{}/", addr))
           .body(Bytes::new())
           .unwrap();
       assert!(client.send_bytes(request).await.is_err());
   }
 ```

 and use these in our feature-gated tests:

 ```rust
   #[cfg(feature = "reqwest")]
   #[tokio::test]
   async fn reqwest_blocking_body_exceeds_limit() {
       let addr = start_server(MAX_RESPONSE_BODY_BYTES + 1).await;
       assert_exceeds_limit(&reqwest::blocking::Client::new(), addr).await;
   }
 ```

opentelemetry-http/src/lib.rs
```diff
@@ -200,11 +238,31 @@ pub mod hyper {
...
-
+            let mut body = response.into_body();
+            while let Some(frame) = body.frame().await {
+                let frame = frame?;
+                if let Ok(chunk) = frame.into_data() {
+                    if body_bytes.len() + chunk.len() > MAX_RESPONSE_BODY_BYTES {
+                        let mut http_response =
+                            Response::builder().status(status).body(Bytes::new())?;
+                        *http_response.headers_mut() = headers;
+                        return Ok(http_response);
```
scottgerring: Need to do whatever we decide above here too to signal an error

scottgerring (commented):
Hey @SATVIKsynopsis sorry we've lagged on this, especially as it is important to get over the line for OTLP stabilisation.

I've put some comments on here, but I think for the big one (communicating the error), we should wait for @cijothomas to weigh in with his opinion before we implement

SATVIKsynopsis:
I have addressed the test feedback, expanded the test module to any `(feature = "reqwest", feature = "hyper")`, added `assert_within_limit` and `assert_exceeds_limit` helpers and added feature gated tests for reqwest. for the error signalling approach, I am waiting on @cijothomas 's input before implementing the `ResponseBodyTooLarge` error type as suggested.

opentelemetry-http/src/lib.rs
```diff
@@ -90,22 +90,43 @@ pub trait HttpClient: Debug + Send + Sync {
...
+        async_trait, Bytes, HttpClient, HttpError, Request, Response, MAX_RESPONSE_BODY_BYTES,
+    };
 
     #[async_trait]
     impl HttpClient for reqwest::Client {
         async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
             otel_debug!(name: "ReqwestClient.Send");
             let request = request.try_into()?;
-            let mut response = self.execute(request).await?;
+            let mut response = self.execute(request).await?.error_for_status()?;
```
cijothomas: this reverts the bug fix added in #3627 . Must be by mistake conflict resolution?

opentelemetry-http/src/lib.rs
cijothomas: `ResponseBodyTooLarge` approach seems reasonable.

opentelemetry-http/src/lib.rs
SATVIKsynopsis: yeah happened due to confilcts resolution. will revert this back.

SATVIKsynopsis:
I have fixed the merge resolution change by reverting the changes back and added the `ResponseBodyTooLarge` implementation. Let me know if any changes are required.

opentelemetry-http/src/lib.rs
```diff
@@ -90,24 +90,56 @@ pub trait HttpClient: Debug + Send + Sync {
...
+                .unwrap_or(0)
+                .min(MAX_RESPONSE_BODY_BYTES as u64) as usize;
+
+            let mut body_bytes = bytes::BytesMut::with_capacity(capacity);
+
+            let status = response.status();
             let headers = std::mem::take(response.headers_mut());
+            while let Some(chunk) = response.chunk().await? {
+                if body_bytes.len() + chunk.len() > MAX_RESPONSE_BODY_BYTES {
+                    return Err(Box::new(ResponseBodyTooLarge));
```
scottgerring, line 135: Great, the `ResponseBodyTooLarge` error type looks good. The remaining step is to recognise it in the OTLP caller before all client errors are converted to the generic network error:

https://github.com/open-telemetry/opentelemetry-rust/blob/846e6ed2d75c617bfff427b815a6f9ef03937a58/opentelemetry-otlp/src/exporter/http/mod.rs#L459-L472

Rather than adding a separate kind field and leaving `status_code` and `retry_after` meaningless for this case, I suggest changing the existing crate-private `HttpExportError` struct into an enum:

```rust
#[derive(Debug)]
pub(crate) enum HttpExportError {
    ClassifyByStatus {
        status_code: u16,
        retry_after: Option<String>,
        message: String,
    },
    ResponseBodyTooLarge {
        message: String,
    },
}
```

The existing `new` and `with_retry_after` constructors can continue to hide this change from their call sites by constructing `ClassifyByStatus`. The `send_bytes` error mapping can downcast to `opentelemetry_http::ResponseBodyTooLarge` and construct the dedicated variant while preserving `e.to_string()` as the message.

Then classification can be super explicit:

```rust
match error {
    HttpExportError::ResponseBodyTooLarge { .. } => RetryErrorType::NonRetryable,
    HttpExportError::ClassifyByStatus {
        status_code,
        retry_after,
        ..
    } => classify_http_error(*status_code, retry_after.as_deref()),
}
```

A small `into_message()` method can replace the current direct `e.message` access. This keeps the change focused and avoids inventing a HTTP status code for a client-side response parsing failure.

Could you also add a regression test with retries enabled where the mock client returns `ResponseBodyTooLarge`, asserting that only one request attempt occurs?

As this is a bit more involved, let's just make sure @cijothomas doesn't see a simpler way of doing this before getting started.  @cijothomas - wdyt?

opentelemetry-http/src/lib.rs
```diff
@@ -117,23 +149,40 @@ mod reqwest {
     #[async_trait]
     impl HttpClient for reqwest::blocking::Client {
         async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
+            use std::io::Read;
             otel_debug!(name: "ReqwestBlockingClient.Send");
             let request = request.try_into()?;
-            let mut response = self.execute(request)?;
+            let mut response = self.execute(request)?.error_for_status()?;
```
scottgerring: It looks like you've accidentally reverted #3627

opentelemetry-http/src/lib.rs
```diff
@@ -90,24 +90,56 @@ pub trait HttpClient: Debug + Send + Sync {
...
+        async_trait, Bytes, HttpClient, HttpError, Request, Response, MAX_RESPONSE_BODY_BYTES,
+    };
 
     #[async_trait]
     impl HttpClient for reqwest::Client {
         async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
             otel_debug!(name: "ReqwestClient.Send");
             let request = request.try_into()?;
-            let mut response = self.execute(request).await?;
+            let mut response = self.execute(request).await?.error_for_status()?;
```
scottgerring: It looks like you've accidentally reverted #3627

opentelemetry-http/src/lib.rs
```diff
@@ -450,3 +421,67 @@ Connection: close\r\n\r\n",
...
+        assert_eq!(response.body().len(), 100);
+    }
+
+    async fn assert_exceeds_limit(client: &dyn HttpClient, addr: SocketAddr) {
+        let request = Request::builder()
+            .method("POST")
+            .uri(format!("http://{}/", addr))
+            .body(Bytes::new())
+            .unwrap();
+        assert!(client.send_bytes(request).await.is_err());
```
scottgerring: Thanks for generalizing these helpers. At the moment the tests still exercise only the async `reqwest::Client`; the reqwest-blocking and hyper implementations each have different body-reading logic but are not invoked by these tests.

Could you follow the same within-limit / exceeds-limit pattern for each built-in client implementation?

For example, the blocking oversized case could look like this:

```rust
#[cfg(feature = "reqwest-blocking")]
#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
async fn reqwest_blocking_body_exceeds_limit() {
    let addr = start_server(MAX_RESPONSE_BODY_BYTES + 1).await;
    assert_exceeds_limit(&reqwest::blocking::Client::new(), addr).await;
}
```

The multi-thread runtime is important for blocking reqwest because the client blocks one worker while the test server runs on another.

I think we should end up with something like ... 

- reqwest async: within limit and exceeds limit
- reqwest blocking: within limit and exceeds limit
- hyper: within limit and exceeds limit

Please also include `reqwest-blocking` explicitly in the test module's feature gate.

Finally, `assert_exceeds_limit` should verify that the failure is specifically `ResponseBodyTooLarge`, rather than allowing an unrelated connection or server failure to satisfy `is_err()`:

```rust
let error = client.send_bytes(request).await.unwrap_err();
assert!(error.downcast_ref::<ResponseBodyTooLarge>().is_some());
```

That gives us confidence that all three client-specific bounded-read paths are exercised and fail for the intended reason.

opentelemetry-http/src/lib.rs
SATVIKsynopsis: Thanks Scott. I checked this against the commit history as well and I am a little confused by this one. The duplicate execute(request) lines shown in this comment were not present in the version I pushed after your earlier review either, the current file has only the .error_for_status()? call in both clients. also verified with grep that there is only one execute(request) occurrence for each client. could this comment be pointing at a stale version of the diff?

opentelemetry-http/src/lib.rs
SATVIKsynopsis: I implemented the blocking client body read using `spawn_blocking` but the exact test structure suggested above still panics when the temporary `reqwest::blocking::Client` is dropped inside the tokio async test. 

The async reqwest tests pass while both blocking tests fail only with this runtime drop panic. i am checking the appropriate test structure for `reqwest::blocking::Client` rather than changing the client behavior solely to accommodate the test.

SATVIKsynopsis:
I implemented the blocking reqwest fix with `spawn_blocking` and added the required tests. @cijothomas, could you take a look at the proposed retry classification approach and let me know?

opentelemetry-http/src/lib.rs
```diff
@@ -384,85 +367,6 @@ mod tests {
...
-            .enable_all()
-            .build()
-            .unwrap();
-        let response = runtime.block_on(client.send_bytes(request)).unwrap();
-
-        server.join().unwrap();
-        assert_eq!(response.status(), http::StatusCode::TOO_MANY_REQUESTS);
-        assert_eq!(response.headers().get("retry-after").unwrap(), "7");
-    }
-
```
scottgerring, line 404: This was also removed in the merge resolution in https://github.com/open-telemetry/opentelemetry-rust/pull/3501/changes/d6da39001a3150aaf8dce16641100f552c06dda6 but should not have been

opentelemetry-http/src/lib.rs
scottgerring: Hey @SATVIKsynopsis it's definitely gone! I reckon in your merge resolution in https://github.com/open-telemetry/opentelemetry-rust/pull/3501/changes/d6da39001a3150aaf8dce16641100f552c06dda6 you clicked "take mine" a bunch and that removed this as well as the other part I flagged below. 

To fix this, you can create a new commit that does this:

1. Remove `.error_for_status()?` from the async reqwest client (the `self.execute(request).await?` line)
2. Remove `.error_for_status()?` from the blocking reqwest client (the `client.execute(request)?` line)
3. Remove the unused `use crate::ResponseExt;` from the hyper module (this also fixes the lint failure)
4. Restore the three `_preserves_error_response_status_and_headers` regression tests and their `spawn_error_response_server` helper from main - these guard against this exact regression

... at the end of the day, we squash into main, so it doesn't matter if the intermediate merge commit is wrong as long as the final state of the change looks good - appending a new commit to fix to your branch is fine. 

That should also fix the cargo-shear failure (the tests use `futures-executor`, which is currently unused).

Once you've done this and you look at the diff here, you should see only your changes!

opentelemetry-http/src/lib.rs
SATVIKsynopsis: @scottgerring .the restored regression tests pass now which has been reverted and I had to use a dedicated thread for the blocking reqwest implementation because `futures_executor::block_on` doesn't provide a tokio runtime. also i also approached it as keeping the `reqwest-blocking` as `spawn_blocking` and changed the test `reqwest_blocking_preserves_error_response_status_and_headers` to to use a tokio runtime and both os thread in `send_bytes` and this tokio runtime approach in the test made the tests pass. so which approach should i go with?

opentelemetry-http/src/lib.rs
cijothomas: Thanks.

opentelemetry-http/src/lib.rs
scottgerring: Please use the direct blocking implementation, not `tokio::task::spawn_blocking`.

The fact that `futures_executor::block_on` does not provide a Tokio runtime is intentional: the default SDK processors already run the blocking exporter on their own dedicated OS threads, so there is no need to spawn more threads for blocking things, and we have (intentionally!) no need for Tokio in this setup. Calling `spawn_blocking` from those threads currently causes the integration-test failures you can see --> 

```text
```

Changing `reqwest_blocking_preserves_error_response_status_and_headers` to provide a Tokio runtime makes it pass by removing the behavior that the regression test is meant to verify. The test should continue to confirm that the default blocking client works without Tokio.

You'll need to:

1. Revert the `spawn_blocking` wrapper and per-request client clone.
2. Do the bounded synchronous `Read::take(MAX_RESPONSE_BODY_BYTES + 1)` directly in `send_bytes`, as before commit `d3dd7585`.
3. Remove Tokio from the `reqwest-blocking` feature.
4. Restore `reqwest_blocking_preserves_error_response_status_and_headers` using `futures_executor::block_on`.

This is getting a bit involved! If you'd like, you can reach out in #opentelemetry-rust  in the CNCF slack - you can [join here](https://join.slack.com/t/cloud-native/shared_invite/zt-fyy3b8up-qHeDNVqbz1j8HDY6g1cY4w).

SATVIKsynopsis:
Got it thanks. I will revert back the `send_bytes` to use the direct blocking implementation and do the other required changes. Also the invitation show revoked. can you resend it?

scottgerring:
Hrm i'm not sure how it works tbh; can you try this link? https://cncf-slack.netlify.app/

SATVIKsynopsis:
i implemented the direct blocking approach, removed tokio from `reqwest-blocking` and restored the regression test with futures_executor::block_on.
and also added a synchronous test server for the blocking body limit tests

SATVIKsynopsis:
> Hrm i'm not sure how it works tbh; can you try this link? https://cncf-slack.netlify.app/

maybe i don't have access to opentelemetry-rust channel that's why i can't join.

scottgerring:
> i implemented the direct blocking approach, removed tokio from `reqwest-blocking` and restored the regression test with futures_executor::block_on. and also added a synchronous test server for the blocking body limit tests

@SATVIKsynopsis Thanks! I'll take a look again soon.

for slack, there's some links on the README for our repo here - can you try those? It's all public including the otel channels

opentelemetry-http/src/lib.rs
```diff
@@ -450,3 +422,216 @@ Connection: close\r\n\r\n",
         assert!(new_carrier.capacity() >= 5);
     }
 }
```
scottgerring:
```suggestion
```

this closes the nested tests module too early; everything below here should remain in tests

opentelemetry-http/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 use async_trait::async_trait;
 use std::fmt::Debug;
+use std::io::{Read, Write};
+use std::net::{SocketAddr, TcpListener};
+use std::thread::JoinHandle;
```
scottgerring: I think these were mistakenly pulled in by the shifting boundary of the `tests` module; they can go back down into the tests module and be removed from here once you fix that!

opentelemetry-http/src/lib.rs
scottgerring: This should be the last bit:

https://github.com/SATVIKsynopsis/opentelemetry-rust/compare/oltp-http...scottgerring:sgg/handle-response-size?expand=1#diff-2a409f8034e20e9ca28de0386d76e4026d45be847f371160aefb861f427459b9R78

@SATVIKsynopsis, if you like you can cherry pick this in here, or if it's easier I think we can merge your PR _without_ this last little bit, then I can raise this directly on main. It is still a net improvement - we're classifying large response bodies as errors, we're just missing the last little bit to skip the retry - so I don't think its a tragedy to do it in two parts.

opentelemetry-http/src/lib.rs
scottgerring: this bit looks good now!

opentelemetry-http/src/lib.rs
SATVIKsynopsis: resolved

opentelemetry-http/src/lib.rs
SATVIKsynopsis: resolved

SATVIKsynopsis:
Thanks Scott! cherry picked this commit into my branch and pushed the changes. all `opentelemetry-otlp` tests are passing. this should now include the non retryable handling for oversized HTTP response bodies.

SATVIKsynopsis:
Let me know if any changes are needed. Happy to address!

opentelemetry-http/src/lib.rs
```diff
@@ -90,24 +90,56 @@ pub trait HttpClient: Debug + Send + Sync {
     async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError>;
 }
 
+#[cfg(any(feature = "reqwest", feature = "hyper"))]
+const MAX_RESPONSE_BODY_BYTES: usize = 4 * 1024 * 1024;
+
+#[derive(Debug)]
```
scottgerring, line 97:
```suggestion
/// Error returned when an HTTP response body exceeds the configured size limit.
#[derive(Debug)]
```

scottgerring:
I believe the lint failure is because of the new Rust release and not this change. I've raised #3641 to fix it; no need to action here.

opentelemetry-http/src/lib.rs
scottgerring: Thanks for adding this!

opentelemetry-http/src/lib.rs
```diff
@@ -257,20 +320,8 @@ impl<T> ResponseExt for Response<T> {
 mod tests {
     use super::*;
     use http::HeaderValue;
-    #[cfg(all(
-        any(feature = "hyper", feature = "reqwest", feature = "reqwest-blocking"),
-        not(target_arch = "wasm32")
-    ))]
     use std::io::{Read, Write};
```
scottgerring, line 329: Can we restore the guards around these includes? As it stands, we can break the build by turning off default features - this will fail: 

```bash
   cargo +1.98.0 clippy -p opentelemetry-http \
     --all-targets --no-default-features -- -Dwarnings
``` 

We just need to add the guards back:

```suggestion
    #[cfg(all(
        any(feature = "hyper", feature = "reqwest", feature = "reqwest-blocking"),
        not(target_arch = "wasm32")
    ))]
    use std::io::{Read, Write};
```

opentelemetry-http/src/lib.rs
```diff
@@ -257,20 +320,8 @@ impl<T> ResponseExt for Response<T> {
...
-    #[cfg(all(
-        any(feature = "hyper", feature = "reqwest", feature = "reqwest-blocking"),
-        not(target_arch = "wasm32")
-    ))]
     use std::net::{SocketAddr, TcpListener};
```
scottgerring, line 335: and here: 
```suggestion
    #[cfg(all(
        any(feature = "hyper", feature = "reqwest", feature = "reqwest-blocking"),
        not(target_arch = "wasm32")
    ))]
    use std::net::{SocketAddr, TcpListener};
```

opentelemetry-http/src/lib.rs
```diff
@@ -257,20 +320,8 @@ impl<T> ResponseExt for Response<T> {
...
-    #[cfg(all(
-        any(feature = "hyper", feature = "reqwest", feature = "reqwest-blocking"),
-        not(target_arch = "wasm32")
-    ))]
     use std::thread::JoinHandle;
```
scottgerring, line 341: and here:


```suggestion
    #[cfg(all(
        any(feature = "hyper", feature = "reqwest", feature = "reqwest-blocking"),
        not(target_arch = "wasm32")
    ))]
    use std::thread::JoinHandle;
```

opentelemetry-http/src/lib.rs
```diff
@@ -402,51 +501,142 @@ Connection: close\r\n\r\n",
...
-            HeaderExtractor(&carrier).get("test-header"),
-            Some("test-value")
-        );
 
-        // Test reserve with zero capacity
-        {
-            let mut injector = HeaderInjector(&mut carrier);
-            injector.reserve(0);
-            injector.set("another-header", "another-value".to_string());
+        fn start_blocking_server(body_size: usize) -> SocketAddr {
```
scottgerring, line 576: this needs a gate - otherwise this code is unused when we compile without `reqwest-blocking`, which fails our lints: 

```suggestion
        #[cfg(feature = "reqwest-blocking")]
        fn start_blocking_server(body_size: usize) -> SocketAddr {
```

SATVIKsynopsis:
I have addressed the all 4 suggestions and ran the specified commands.

scottgerring (approved):
LGTM!
@cijothomas I reckon we should get this in for the upcoming OTLP stabilisation release.

scottgerring:
MacOS failures are flakey timing tests from my #3637. PR to fix here; issue not introduced by this branch
https://github.com/open-telemetry/opentelemetry-rust/pull/3643

opentelemetry-http/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- Limit HTTP response body reads to 4 MiB in built-in HTTP clients (`reqwest` async/blocking and `hyper`) by enforcing the cap while streaming response chunks and reads that exceed the limit are aborted to prevent unbounded memory allocation.
```
cijothomas, line 5: do include this too: Responses exceeding this limit now return `ResponseBodyTooLarge`

Also, we need a changelog entry for OTLP Exporter too, given that is what end users directly use and this PR adds a new limitation to OTLP HTTP Exporter.

opentelemetry-http/src/lib.rs
```diff
@@ -90,24 +90,57 @@ pub trait HttpClient: Debug + Send + Sync {
...
     #[async_trait]
     impl HttpClient for reqwest::Client {
         async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
             otel_debug!(name: "ReqwestClient.Send");
             let request = request.try_into()?;
             let mut response = self.execute(request).await?;
+            let capacity = response
+                .content_length()
+                .unwrap_or(0)
+                .min(MAX_RESPONSE_BODY_BYTES as u64) as usize;
```
cijothomas, line 127: If content-length is present, and is above litmit could we short circuit right here, instead of allocating the capacity to match it, and then do work of reading upto the limit and still return error?

cijothomas (approved):
Thanks for patiently working through fixing this! 
I left couple of comments, which are okay to do as follow ups - change entry for OTLP, and early short-circuit when content-length already tells us we are above limit.

scottgerring:
Cheers @SATVIKsynopsis ! I've added #3644 to allow this to be configured and to quickly address Cijo's comments above; if you have a second to take a look that would be great 💪 !

PR #3506, opened 2026-05-08
fix(metrics): reject usize::MAX as cardinality limit

cijothomas:
Reverts #3290 and instead rejects `usize::MAX` in `StreamBuilder::build()`. The original integer-overflow panic is fixed at the input boundary, keeping the SDK's no-allocation-on-hot-path design intact.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -47,7 +47,8 @@
 - **Breaking** Moved the following SDK sampling types from `opentelemetry::trace` to `opentelemetry_sdk::trace` [#3277][3277]:
   - `SamplingDecision`, `SamplingResult`
   - These types are SDK implementation details and should be imported from `opentelemetry_sdk::trace` instead.
-- Fix panics and exploding memory usage from large cardinality limit [#3290][3290]
+- `StreamBuilder::build()` now rejects `usize::MAX` as a cardinality limit
+  with a validation error.
 - Fix Histogram boundaries being ignored in the presence of views [#3312][3312]
 - `TracerProviderBuilder::with_sampler` allows to pass boxed instance of `ShouldSample` [#3313][3313]
 - Fix ObservableCounter and ObservableUpDownCounter now correctly report only data points from the current measurement cycle, removing stale attribute combinations that are no longer observed. [#3248][3248]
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/CHANGELOG.md
```diff
@@ -47,7 +47,8 @@
 - **Breaking** Moved the following SDK sampling types from `opentelemetry::trace` to `opentelemetry_sdk::trace` [#3277][3277]:
   - `SamplingDecision`, `SamplingResult`
   - These types are SDK implementation details and should be imported from `opentelemetry_sdk::trace` instead.
-- Fix panics and exploding memory usage from large cardinality limit [#3290][3290]
+- `StreamBuilder::build()` now rejects `usize::MAX` as a cardinality limit
+  with a validation error.
```
utpilla, line 51: Add this PR's link.

utpilla (approved):

opentelemetry-sdk/src/metrics/instrument.rs
```diff
@@ -544,8 +549,24 @@ mod tests {
...
+        );
+        assert_eq!(
+            result.err().unwrap().to_string(),
+            "Cardinality limit must be less than usize::MAX",
+            "Expected cardinality limit usize::MAX error message"
+        );
+
         // Test valid cardinality limits
-        let valid_limits = vec![1, 10, 100, 1000];
+        let valid_limits = vec![1, 10, 100, 1000, usize::MAX - 1];
```
lalitb, line 569: Marking `usize::MAX - 1` as valid here is misleading. It avoids the `1 + cardinality_limit` overflow, but `ValueMap::new()` then calls `HashMap::with_capacity(usize::MAX)`, which still fails at allocation. The PR fixes the arithmetic-overflow case correctly, but any sufficiently large limit will still fail loudly when the stream is first used.

Should we either tighten the validation to a sane upper bound, or decouple the initial HashMap capacity from the configured cardinality limit so a high limit doesn't force a huge upfront allocation?

ArniDagur: Exactly this. The fix that was reverted was not centered around the integer overflow. The fix was to avoid the huge (and wasteful) up-front allocation. By reverting my PR, a very real bug is being re-introduced.

utpilla: The discussion about disabling cardinality limit (modifying upfront allocation logic) is being tracked in this https://github.com/open-telemetry/opentelemetry-rust/issues/3507. It would be worked upon after the 0.32 release.

PR #3512, opened 2026-05-11
chore: decouple opentelemetry-proto from opentelemetry-sdk

scottgerring:
## Summary

This refactors `opentelemetry-proto` so it only contains generated OTLP protobuf types. The diff is substantial because of the orphan rules - opentelemetry-otlp cannot implement the conversion traits for foreign types into foreign types. My feeling is that we just need to pull off the bandaid here! 

The SDK/api-aware transform helpers have been moved into `opentelemetry-otlp`, where they are used by the exporters. This removes the direct `opentelemetry` / `opentelemetry_sdk` dependency from `opentelemetry-proto` and makes the crate usable by consumers that only need OTLP wire-format types. This will also provide other options to structure changes like #3460 , where we _could_ benefit from having the dependency the other way around (although this is by no means a done deal).

## Changes

- Remove the public `opentelemetry_proto::transform` module.
- Move SDK/API → OTLP protobuf transformation logic into a private `opentelemetry-otlp::transform` module.
- Update OTLP exporters to use the local transform helpers.
- Update `opentelemetry-proto` docs and changelog to describe the crate as generated protobuf types only.

## Prior art / related work

This implements the SDK-decoupling part of:

- #3045  - “Simplify opentelemetry-proto: SDK decoupling, gRPC separation”
- #2508  - earlier proto layout / generated-code cleanup discussion; later folded into #3045
- #3046 - earlier attempted implementation of proto simplification / SDK decoupling
- #3353 - earlier attempted implementation that made SDK dependencies optional rather than fully moving transforms out
- #3419 / #3451 - adjacent work to reduce unnecessary proto/exporter dependencies for OTLP JSON/prost usage


scottgerring:
@lalitb I know you've worked on this before; it would be great to get your high-level opinions on this so I can push it in the right direction. I'm not sure if i've missed some important nuance given this has come up a few times but never landed!

opentelemetry-otlp/src/exporter/tonic/logs.rs
```diff
@@ -152,6 +152,7 @@ impl LogExporter for TonicLogsClient {
     }
 
     fn set_resource(&mut self, resource: &opentelemetry_sdk::Resource) {
-        self.resource = resource.into();
+        self.resource =
```
TommyCpp, line 155: More of a style suggestion. But I wonder if it would be better to package this as a extension trait against SDK resource, instead of `resource_to_attributes_with_schema` function

scottgerring: @TommyCpp thanks for taking a look, and good point!

I think this conversion is one of the few places where a method-style API could be reasonable, since it’s a high-level boundary used directly by the exporters.

That said, I’m a bit hesitant to make it look like an operation that inherently belongs on `Resource`: this is really an OTLP/proto-specific projection of a `Resource`, including how we carry the schema URL alongside attributes. Since we already have a local target type here, `From<&Resource>` gives us a short call site without adding an extension trait or implying this is a general `Resource` capability; maybe something like this? --> 

   ```suggestion
           self.resource = resource.into();
 ```

 For the broader transform module I’d probably keep the explicit *_to_proto functions, since most are lower-level implementation details or external-to-external mappings where `From` isn’t available due to orphan rules.

scottgerring:
This isn't stale, it just needs a review :D

cijothomas:
@lalitb Can you review this one?

scottgerring:
@lalitb do you have time to look at this?

Issue #3354, opened 2026-02-19
[Bug]: OpenTelemetry exporting via HTTP fails with a hyper Parse(Version) error


V0ldek:
### What happened?

I updated my old project to newest OpenTelemetry version. It used the `opentelemetry-oltp` exporter to send logs to a local Jaeger container and it used to work fine. Now it fails to send any logs to Jaeger with the following error:

```
2026-02-19T12:53:22.091417Z ERROR ThreadId(36) opentelemetry_sdk:  name="BatchSpanProcessor.ExportError" error="Operation failed: reqwest::Error { kind: Request, url: \"http://localhost:4317/\", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Parse(Version))) }"
  2026-02-19T12:53:22.091435Z ERROR opentelemetry_sdk: , name: "BatchSpanProcessor.ExportError", error: "Operation failed: reqwest::Error { kind: Request, url: \"http://localhost:4317/\", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Parse(Version))) }"
    at /home/mat/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/opentelemetry_sdk-0.31.0/src/trace/span_processor.rs:512
```

This happens in my big project, but I was able to do a MRE. Note that it happens for both http-proto and http-json, I checked on both.

<details>
<summary>MRE code</summary>

```sh
cargo init jaeger-oltp-repro
```

In `Cargo.toml`:

```toml
[package]
name = "jaeger-oltp-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
opentelemetry = "0.31.0"
opentelemetry-http = { version = "0.31.0", default-features = false, features = [
    "internal-logs",
    "reqwest-blocking",
    "reqwest-rustls",
]}
opentelemetry-otlp = { version = "0.31.0", default-features = false, features = [
    "http-proto",
    "reqwest-blocking-client",
    "reqwest-rustls",
] }
opentelemetry_sdk = { version = "0.31.0", features = [
    "rt-tokio",
] }
reqwest = { version = "0.13.2", features = ["rustls", "http2"] }
tokio = { version = "1.49.0", features = ["full"] }
tracing = { version = "0.1.40", features = ["release_max_level_warn"] }
tracing-opentelemetry = "0.32.1"
tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
```

In `main.rs`:

```rs
[...]

PR #3513, opened 2026-05-11
docs: clarify OTLP HTTP client runtime compatibility

scottgerring:
We have a bunch of HTTP client features and some unintuitive rules around their usage; this PR tries to make the rules explicit.

I think long run we can probably do better than this but as we've had [some feedback](https://github.com/open-telemetry/opentelemetry-rust/issues/3354#issuecomment-4025535409) on this I think we should make it clearer immediately.

The advice is:

* **Just getting started?** --> Use `reqwest-blocking-client` - this is the default behaviour with the stable processors. The async clients should not be used here.
* **Using `experimental_async_runtime` ? Use `reqwest-client`, or `hyper-client`

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -287,7 +287,8 @@
 //! * `gzip-http`: Use gzip compression for HTTP transport.
 //! * `zstd-http`: Use zstd compression for HTTP transport.
 //! * `reqwest-blocking-client`: Use reqwest blocking http client. This feature is enabled by default.
-//! * `reqwest-client`: Use reqwest http client.
+//! * `reqwest-client`: Use reqwest async http client.
+//! * `hyper-client`: Use hyper async http client.
 //! * `reqwest-rustls`: Use reqwest with TLS with system trust roots via `rustls-native-certs` crate.
 //! * `reqwest-rustls-webpki-roots`: Use reqwest with TLS with Mozilla's trust roots via `webpki-roots` crate.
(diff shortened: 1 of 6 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -510,6 +511,21 @@
...
+//! SDK-owned background threads, so OTLP/HTTP should use the default
+//! `reqwest-blocking-client` with them.
+//!
+//! HTTP client selection is based on enabled crate features and is not aware of
+//! the SDK processor or metric reader that will drive the exporter. If you enable
+//! async HTTP clients such as `reqwest-client` or `hyper-client`, do not use them
+//! with the default batch processors or periodic reader.
+//!
+//! Async HTTP clients are only appropriate when export futures are driven by a
+//! Tokio runtime, for example by custom code that polls the exporter future on
```
cijothomas: what do we mean when we say "custom code that polls the exporter" ? Do we mean users writing own exporting processors instead of using one of the built-in one like batch processor?

scottgerring: Exactly. Likely not so clear - could either 1/ drop this (wrong level of detail for here?) or 2/ clarify "for folks building their own processors"

cijothomas: drop is fine. (its very rare for people to write custom export processors. Most processor authors are for enrichment ones!)

TommyCpp (approved):

PR #3514, opened 2026-05-11
feat(sdk): add otel.sdk.processor.log.processed self-diagnostics metric

cijothomas:
Adds the `otel.sdk.processor.log.processed` counter to `BatchLogProcessor`, following the [OTel SDK metrics semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/otel/sdk-metrics.md#metric-otelsdkprocessorlogprocessed).

Gated behind the existing `experimental_metrics_bound_instruments` feature flag — bound instruments make this practical on the hot path (~1.8 ns per emit vs ~50 ns with unbound).

**Changes:**
- `BatchLogProcessor` emits a counter on every `emit()` call with three attribute variants: success, `queue_full`, and `already_shutdown`
- OTLP examples reordered to set up MeterProvider first (so the processor gets a real meter)
- Shutdown order corrected to tracer → logger → meter
- Design doc added at `docs/design/observability.md`
- Ignored test added (runs in isolation in CI via `test.sh`)

**Ordering requirement:** `global::set_meter_provider()` must be called before creating the `LoggerProvider`. Rust's `global::meter()` returns a snapshot, not a delegate. See the design doc for details.

### DIFF
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -25,6 +25,9 @@ use std::sync::mpsc::{self, RecvTimeoutError, SyncSender};
 
 use opentelemetry::{otel_debug, otel_error, otel_warn, Context, InstrumentationScope};
 
+#[cfg(feature = "experimental_metrics_bound_instruments")]
+use opentelemetry::KeyValue;
+
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::{cmp::min, env, sync::Mutex};
 use std::{
@@ -142,6 +145,17 @@ pub struct BatchLogProcessor {
 
     // Track the maximum queue size that was configured for this processor
     max_queue_size: usize,
+
+    // Self-diagnostics: otel.sdk.processor.log.processed counter.
+    // Gated behind experimental_metrics_bound_instruments so the hot-path
+    // `add` is a single atomic increment (~1.8 ns) with no per-call
+    // attribute resolution.
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    processed_success: opentelemetry::metrics::BoundCounter<u64>,
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    processed_queue_full: opentelemetry::metrics::BoundCounter<u64>,
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    processed_after_shutdown: opentelemetry::metrics::BoundCounter<u64>,
 }
 
 impl Debug for BatchLogProcessor {
@@ -161,6 +175,10 @@ impl LogProcessor for BatchLogProcessor {
         // match for result and handle each separately
         match result {
             Ok(_) => {
+                // Record successful processing in self-diagnostics
+                #[cfg(feature = "experimental_metrics_bound_instruments")]
+                self.processed_success.add(1);
+
                 // Successfully sent the log record to the data channel.
                 // Increment the current batch size and check if it has reached
                 // the max export batch size.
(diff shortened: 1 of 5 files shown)


opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -497,6 +530,48 @@ impl BatchLogProcessor {
...
             .expect("Thread spawn failed."); //TODO: Handle thread spawn failure
 
+        // Self-diagnostics: create the otel.sdk.processor.log.processed counter
+        #[cfg(feature = "experimental_metrics_bound_instruments")]
+        let (processed_success, processed_queue_full, processed_after_shutdown) = {
+            static INSTANCE_COUNTER: AtomicUsize = AtomicUsize::new(0);
+            let instance_id = INSTANCE_COUNTER.fetch_add(1, Ordering::Relaxed);
+            let component_name = format!("batching_log_processor/{instance_id}");
+
+            let meter = opentelemetry::global::meter("otel.sdk");
```
utpilla, line 540: **Follow-up suggestion (not a blocker for this PR): detect & warn when bound counters resolve to noop**

If the user constructs the `LoggerProvider` before `global::set_meter_provider`, the three `BoundCounter`s permanently bind to `NoopBoundSyncInstrument`, `add(1)` becomes a no-op, and no metric is ever emitted. The user gets no signal — just `total == 0`. The OTLP examples work around it by reordering setup, but the failure mode is silent.

The current public API can't detect this: `MeterProvider` is type-erased, doesn't extend `Any`, and has no `is_noop()` method. Detecting "was `set_meter_provider` called?" via a flag isn't enough either — `NoopMeterProvider` is `pub` and a valid `MeterProvider`, so a user can legitimately install it explicitly. We need to ask the actual provider whether it's a no-op.

### Proposed addition: defaulted `is_noop` on `MeterProvider`

```rust
// opentelemetry/src/metrics/meter.rs
pub trait MeterProvider {
    // existing methods...

    /// Returns `true` if this provider is a no-op implementation that
    /// silently discards all measurements. Defaults to `false`; the
    /// built-in `NoopMeterProvider` overrides to `true`. SDK components
    /// can use this to warn at construction when self-diagnostics metrics
    /// would be silently dropped.
    fn is_noop(&self) -> bool { false }
}

// opentelemetry/src/metrics/noop.rs
impl MeterProvider for NoopMeterProvider {
    // existing methods...
    fn is_noop(&self) -> bool { true }
}
```

Backwards-compatible — existing `MeterProvider` impls don't need to change.

### Usage in `BatchLogProcessor::new`

```rust
let provider = opentelemetry::global::meter_provider();
if provider.is_noop() {
    otel_warn!(
        name: "BatchLogProcessor.SelfDiagnosticsUnavailable",
        message = "BatchLogProcessor obtained a no-op MeterProvider from the global \
                   registry; otel.sdk.processor.log.processed will not record. Call \
                   global::set_meter_provider with a real MeterProvider before \
                   constructing the LoggerProvider to enable self-diagnostics."
    );
}
let meter = provider.meter("otel.sdk");
// ... build counter, bind() three times as today
```

### Benefits

- Catches both the "forgot to set the provider" and "set it after the LoggerProvider" cases — the warning fires whenever the bound counters would silently drop measurements.
- One virtual call at construction; zero impact on `add()` hot path.
- Works on injected providers too, in case any SDK component later accepts a `MeterProvider` parameter instead of reaching for the global.
- Reusable by `BatchSpanProcessor`, `PeriodicReader`, and any future component that emits self-diagnostics.

Does **not** fix the late-rebind case (user replaces the global provider with a different real one after construction) — that needs the late-binding `global::meter()` work already noted in `docs/design/observability.md`.

utpilla (approved):

cijothomas:
@bantonsson PTAL! thanks in advance.

TommyCpp (approved):

opentelemetry-sdk/src/logs/batch_log_processor.rs
cijothomas: agree. I think we need to solve the fundamental underlying issue - that is the proper solution!

PR #3517, opened 2026-05-14
feat(sdk): relax instrument name validation; remove experimental_metrics_disable_name_validation feature

cijothomas:
**Draft** — depends on [opentelemetry-specification#5092](https://github.com/open-telemetry/opentelemetry-specification/pull/5092). Will be ready for review once the spec PR is merged.

**Description**

Relaxes the SDK instrument-name validation to match the proposed expanded character set in [opentelemetry-specification#5092](https://github.com/open-telemetry/opentelemetry-specification/pull/5092):

- Adds `:`, `\`, `(`, `)`, `%`, `*`, `#`, and space to the allowed character set.
- Drops the requirement that the first character be alphabetic.
- Space is restricted to non-leading and non-trailing positions.
- ASCII-only, the 255-byte maximum, and the `name must be non-empty` rule are unchanged.

This enables Windows performance counter style names (e.g. `\Processor(_Total)\% Processor Time`, `\.NET CLR Memory(*)\# Bytes in all Heaps`) to be used directly as instrument names without workarounds.

**Removes the `experimental_metrics_disable_name_validation` Cargo feature** (introduced in #2538 specifically for these use cases). With the relaxed validation, the workaround is no longer needed and the feature can be retired.

**Backwards compatibility**

Strict superset of the existing rule. Every name valid under the old validator remains valid; some previously-rejected names are now accepted. Consumers that enabled `experimental_metrics_disable_name_validation` should drop the feature from their `Cargo.toml`.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## vNext
 
+- **Relaxed instrument name validation** to accept the expanded character set
+  defined by the OpenTelemetry specification. The allowed character set now
+  includes `:`, `\`, `(`, `)`, `%`, `*`, `#`, and space (with space restricted
+  to non-leading, non-trailing positions), and the first character is no
+  longer required to be alphabetic. This enables names shaped like Windows
+  performance counters (`\Processor(_Total)\% Processor Time`,
+  `\.NET CLR Memory(*)\# Bytes in all Heaps`) to be used directly without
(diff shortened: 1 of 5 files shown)


cijothomas:
Closing — open-telemetry/opentelemetry-specification#5092 was closed in favor of using Views for these cases (per spec clarification open-telemetry/opentelemetry-specification#5094). The Meter-level instrument name relaxation is no longer needed. The View-level fix is tracked in #3518.

PR #3521, opened 2026-05-21
feat: add prototype for EnvVarsCarrier (env var propagation)

### LINKED ISSUE #3285: [Feature]: Support the environment variable propagation spec in Rust
### Related Problems?

With the [environment variable context propagation spec](https://opentelemetry.io/docs/specs/otel/context/env-carriers/) being added, we'd like to have its support added to Javascript. This is part of a larger effort to adding support in each of the currently supported languages.

I'd like the specification implemented in Rust.

### Considered Alternatives

n/a

### Additional Context

_No response_

### Tip

jsurany:
Fixes #3285 
Design discussion issue (if applicable) N/A?

## Changes

- Added `EnvVarsCarrier` type that implements `Extractor` and `Injector` traits, with the normalization described in the spec.
- `EnvVarsCarrier` pulls the environment in `from_env`, with normalization
- The `Extractor` trait seemed straightforward, but `Injector` felt up to interpretation. In this design, the injection only modifies the internal map on the `EnvVarsCarrier` type (keys are always normalized).
- added environment variable to `.cargo/config.toml` for testing the carrier's initialization. This seemed better than adding `unsafe` code to the tests.

Would like some feedback on if this is the correct way to have implemented this feature, and advice on managing the changelog.


opentelemetry-sdk/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub fn new() -> Self {
+        let map = std::env::vars().map(|(k, v)| (normalize(&k), v)).collect();
+
+        Self { map }
+    }
+
+    /// Create a new `EnvVarsCarrier` object, internally empty. Useful for
+    /// testing and for setting up environment mapping for subprocesses from
+    /// scratch.
+    pub fn new_empty() -> Self {
```
TommyCpp:
```suggestion
    pub fn empty() -> Self {
```

opentelemetry-sdk/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,151 @@
...
+impl Default for EnvVarsCarrier {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl EnvVarsCarrier {
+    /// Create a new `EnvVarsCarrier` object, built from environment variables.
+    /// Environment variables are fetched and normalized at construction time.
+    pub fn new() -> Self {
```
TommyCpp:
```suggestion
    pub fn from_env() -> Self {
```

opentelemetry/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,151 @@
...
+/// ```
+///
+/// [POSIX.1-2024]: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html
+#[derive(Debug, Clone)]
+pub struct EnvVarsCarrier {
+    map: HashMap<String, String>,
+}
+
+impl Default for EnvVarsCarrier {
+    fn default() -> Self {
```
TommyCpp, line 45: Hmm what's the benifit to have Default scan the env vars? General speaking Default should be no-op

TommyCpp (commented):
Most implementation leave this in API and the carrier is SDK agnostic, so we should move it to api create(`opentelemetry/src/propagation`)

opentelemetry-sdk/src/propagation/env_vars_carrier.rs
jsurany: updated

opentelemetry-sdk/src/propagation/env_vars_carrier.rs
jsurany: updated

opentelemetry/src/propagation/env_vars_carrier.rs
jsurany: for me it was what made sense to make the default, but I'm not attached to it. Updated to map `default` to `empty()`

jsurany:
> Most implementation leave this in API and the carrier is SDK agnostic, so we should move it to api create(`opentelemetry/src/propagation`)

moved the change.

Would it be OK if I rebase? I didn't update my git config for this PR and as a result the EasyCLA thinks that the early commits are coming from a different account (one that I don't use anymore).

opentelemetry/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn empty() -> Self {
+        Self {
+            map: HashMap::new(),
+        }
+    }
+}
+
+impl Injector for EnvVarsCarrier {
+    /// Set the value for the normalized key in the carrier mapping.
+    /// Does NOT set environment variables, but may
```
pellared: Is the sentence broken?

opentelemetry/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn set(&mut self, key: &str, value: String) {
+        self.map.insert(normalize(key), value);
+    }
+}
+
+impl Extractor for EnvVarsCarrier {
+    /// Get the value for the normalized key from the carrier mapping.
+    fn get(&self, key: &str) -> Option<&str> {
+        self.map.get(&normalize(key)).map(|s| s.as_str())
```
pellared, line 79: It does not look correct. Please see https://github.com/open-telemetry/opentelemetry-specification/pull/5102 which clarifies the behavior. I also added a a test which should pass in https://github.com/open-telemetry/opentelemetry-rust/pull/3521#discussion_r3339606842.

PS. I would also welcome a review of https://github.com/open-telemetry/opentelemetry-specification/pull/5102

opentelemetry/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    #[test]
+    fn test_env_vars_carrier_extractor() {
+        let mut carrier = EnvVarsCarrier::empty();
+        carrier
+            .map
+            .insert("FOO_BAR".to_string(), "value".to_string());
+
+        assert_eq!(Extractor::get(&carrier, "foo.bar"), Some("value"));
+    }
```
pellared, line 181:
```rust
    #[test]
    fn test_env_vars_carrier_extractor_bis() {
        let mut carrier = EnvVarsCarrier::empty();
        carrier
            .map
            .insert("foo.bar".to_string(), "value".to_string());

        assert_eq!(Extractor::get(&carrier, "FOO_BAR"), Some("value"));
    }
```

this should also pass

pellared:
EasyCLA is failing.

opentelemetry/src/propagation/env_vars_carrier.rs
jsurany: fixed 317829f

opentelemetry/src/propagation/env_vars_carrier.rs
jsurany: Are you referring specifically to this rule?

> Get normalizes both the requested propagator key and carrier keys before matching.

This is fulfilled in that we normalize the propagator key when they call `get`, and that all keys stored in the carrier itself are already normalized; all writes to the internal map are protected by `normalize`.

Can you elaborate on what specifically is not correct? I responded to the test that you added below.

opentelemetry/src/propagation/env_vars_carrier.rs
jsurany: I don't think that the test you posted should pass, because it's doing a direct write to the inner map, which is a private field and bypasses normalization. I _think_ I understand your intent and I've added this test.

```rust
#[test]
    fn test_env_vars_carrier_inject_and_extract() {
        let mut carrier = EnvVarsCarrier::empty();
        Injector::set(&mut carrier, "foo.bar", "value".to_string());
        assert_eq!(Extractor::get(&carrier, "foo.bar"), Some("value"));
        assert_eq!(carrier.keys(), vec!["FOO_BAR"]);
    }
```

jsurany:
Sorry, I had to rebase in order to change the contact info for past commits for EasyCLA to pass

opentelemetry/src/propagation/env_vars_carrier.rs
pellared: I have missed `let map = std::env::vars().map(|(k, v)| (normalize(&k), v)).collect();`

LGTM. Feel free to resolve the comment (if you can).

opentelemetry/src/propagation/env_vars_carrier.rs
pellared: https://github.com/open-telemetry/opentelemetry-rust/pull/3521#discussion_r3351921419

Feel free to resolve the comment (if you can).

jsurany:
@pellared @TommyCpp I think I've fixed everything. I don't _think_ I'm responsible for the failing integration test, looks like it's from metrics. Let me know what you'd like me to do.

TommyCpp (approved):

pellared:
It might make sense to keep this as a draft until the following issue is resolved: https://github.com/open-telemetry/opentelemetry-specification/issues/5143. PTAL when you have a chance, and feel free to share your thoughts there. Waiting could help us avoid extra PRs and save reviewers some time.

jsurany:
Converted to draft per above. Waiting for https://github.com/open-telemetry/opentelemetry-specification/issues/5143

pellared:
> Converted to draft per above. Waiting for [open-telemetry/opentelemetry-specification#5143](https://github.com/open-telemetry/opentelemetry-specification/issues/5143)

https://github.com/open-telemetry/opentelemetry-specification/pull/5144 is going to be merged soon. Feel free to adopt necessary changes in this PR.

jsurany:
> [open-telemetry/opentelemetry-specification#5144](https://github.com/open-telemetry/opentelemetry-specification/pull/5144) is going to be merged soon. Feel free to adopt necessary changes in this PR.

@pellared the `Set` behavior is a bit weird now, isn't it?

- when built from the environment, we only include already-normalized strings into the carrier
- when `Get`ting a value, we return no value if the request key is not normalized
- when `Set`ing a value, we normalize the key before writing the value.

It feels like the overall design would be more consistent if `Set` only works if the key is already normalized. Otherwise the user can set a value with a key, then try to use that same key to fetch the value later, only for the `Get` to reject it on the grounds that it's not normalized. 

If we make it so `Set` fails for non-normalized keys, then the `Keys` API doesn't need to filter out non-normalized keys. It just reports on all keys, with the invariant that all keys in the map are already normalized.

EDIT - correction is that `Get`ting a value still applies normalization, the only `is_normalized` filter application is when we are building the carrier from the existing environment. All other interactions use `normalize` on the key.

kamphaus:
@jsurany `Get` must normalize its argument key before using the normalized key name for lookups.
That way using the same non-normalized key for both Get and Set operations works.

opentelemetry/src/propagation/env_vars_carrier.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn default() -> Self {
+        Self::empty()
+    }
+}
+
+impl EnvVarsCarrier {
+    /// Create a new `EnvVarsCarrier` object, built from environment variables.
+    /// Environment variables are fetched and normalized at construction time.
+    pub fn from_env() -> Self {
+        let map = std::env::vars().map(|(k, v)| (normalize(&k), v)).collect();
```
kamphaus: Following https://github.com/open-telemetry/opentelemetry-specification/pull/5144 this should filter to only include normalized key names.

jsurany:
Ok I think I understand now.

opentelemetry/src/propagation/env_vars_carrier.rs
jsurany: updated [714680e](https://github.com/open-telemetry/opentelemetry-rust/pull/3521/commits/714680ee6a1209a92c3c5bd5f31f5191fff818f3)

kamphaus (approved):

pellared:
Is this a draft on purpose?

jsurany:
Hey, sorry the effort under which I made this PR has wrapped up and I now need to wait for a different set of approvals to go through to continue work on this PR. It could take up to 6 weeks. I'll finish up the remaining comments when I have approval again.

pellared:
This is only a prototype and at this point of time we rather need an actual experimental implementation. I suggest closing.

kamphaus:
Superseded by https://github.com/open-telemetry/opentelemetry-rust/pull/3574

PR #3525, opened 2026-05-27
fix(sdk): skip current_exe() under Miri in default service.name fallback

### LINKED ISSUE #3523: [Bug]: SdkMeterProvider::builder().build() aborts under Miri since 0.32 (regression from current_exe call in SdkProvidedResourceDetector)
## What happened?

Starting with `opentelemetry_sdk` 0.32.0, calling
`SdkMeterProvider::builder().build()` (or any other code path that triggers
default `Resource` construction) aborts under [Miri](https://github.com/rust-lang/miri).
The same code worked under Miri on 0.31.0.

Miri intercepts `std::env::current_exe()` because it cannot be emulated in
isolated mode (and "real" emulation would be questionable anyway). When that
call is intercepted, Miri **aborts the program** rather than returning an
`Err`, so the existing `.ok().and_then(...)` fallback in
`SdkProvidedResourceDetector::detect` is unreachable.

Error on Windows:

```
error: unsupported operation: `GetModuleFileNameW` not available when isolation is enabled
  --> .../std/src/sys/paths/windows.rs:99:39
   |
99 |     fill_utf16_buf(|buf, sz| unsafe { c::GetModuleFileNameW(ptr::null_mut(), buf, sz) }, os2path)
   |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unsupported operation occurred here
...
   4: <opentelemetry_sdk::resource::SdkProvidedResourceDetector as opentelemetry_sdk::resource::ResourceDetector>::detect::{closure#2}
       at .../opentelemetry_sdk-0.32.1/src/resource/env.rs:92:25
```

Equivalent error on Linux (`readlink "/proc/self/exe"`). Both platforms abort
at the same call site.

### Bisection

The regression was introduced by
[#3302 — *fix: ServiceName default should include executable name*](https://github.com/open-telemetry/opentelemetry-rust/pull/3302),
shipped in 0.32.0.

Diff in `opentelemetry-sdk/src/resource/env.rs`,
`SdkProvidedResourceDetector::detect`:

```rust
// 0.31.0:
.unwrap_or_else(|| "unknown_service".into()),

// 0.32.0:
.unwrap_or_else(|| {
    env::current_exe()
        .ok()
        .and_then(|path| {
            path.file_name()
                .and_then(|name| name.to_str())
                .map(|name| format!("unknown_service:{}", name))
        })
        .unwrap_or_else(|| "unknown_service".to_string())
        .into()
}),
```

### Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "otel-miri-repro"
version = "0.0.0"
edition = "2024"
publish = false

[dependencies]
opentelemetry = { version = "0.32.0", default-features = false }
opentelemetry_sdk = { version = "0.32.1", default-features = false, features = ["metrics"] }
```

`tests/build_provider.rs`:

```rust
use opentelemetry_sdk::metrics::SdkMeterProvider;

#[test]
fn build_default_provider() {
    let provider = SdkMeterProvider::builder().build();
    drop(provider);
}
```

Run: `cargo +nightly miri test`

* **0.31.0**: `test build_default_provider ... ok`
* **0.32.0 / 0.32.1**: aborts with the error above on both Windows and Linux.

### What did you expect to happen?

The default `SdkMeterProvider` construction should remain Miri-compatible.

This matters because downstream crates that integrate with `opentelemetry_sdk`
use Miri as their primary correctness validation tool for memory safety,
data races, and aliasing rules. A test that constructs a default
`SdkMeterProvider` was previously a useful unit-level integration check;
after 0.32, every such test has to be marked `#[cfg_attr(miri, ignore)]`,
which silently shrinks Miri's coverage of downstream code.

### Suggested fixes (any one of these would resolve it)

1. **`cfg!(miri)` short-circuit** — skip the `current_exe()` call under Miri and
   fall through to `"unknown_service"`. The `miri` cfg is set automatically by
   the toolchain; no extra dependency or feature required. This is the
   smallest, most surgical fix.

   ```rust
   .unwrap_or_else(|| {
       if cfg!(miri) {
           return "unknown_service".to_string().into();
       }
       env::current_exe()
           .ok()
           .and_then(|path| { /* ... */ })
           .unwrap_or_else(|| "unknown_service".to_string())
           .into()
   }),
   ```

2. **Compute the default service name lazily once** behind a `OnceLock`/`OnceCell`,
   and expose a way to seed it (e.g. via an env var or a builder option) so users
   can pre-populate it under Miri and skip the syscall entirely.

3. **Make `SdkProvidedResourceDetector` opt-out** via a builder method on
   `MeterProviderBuilder` / `Resource::builder`. Today the user has to either
   set `OTEL_SERVICE_NAME` (which requires `temp_env` or process-wide env
   mutation in tests — itself problematic) or accept the syscall.

Option (1) is the lowest-risk, highest-value fix and would unblock every
downstream Miri user immediately.

## OpenTelemetry API Version (i.e version of `opentelemetry` crate)

0.32.0

## OpenTelemetry SDK Version (i.e version of `opentelemetry_sdk` crate)

0.32.0 and 0.32.1 (regression introduced in 0.32.0)

## What Exporter(s) are you seeing the problem on?

N/A — the issue is in the SDK's default resource construction, not in any
exporter. It reproduces with no exporters configured.

## Relevant log output

```
error: unsupported operation: `GetModuleFileNameW` not available when isolation is enabled
  --> .../std/src/sys/paths/windows.rs:99:39

stack:
  std::env::current_exe
  <opentelemetry_sdk::resource::SdkProvidedResourceDetector as ResourceDetector>::detect::{closure#2}
      at opentelemetry_sdk-0.32.1/src/resource/env.rs:92:25
  <opentelemetry_sdk::resource::SdkProvidedResourceDetector as ResourceDetector>::detect
      at opentelemetry_sdk-0.32.1/src/resource/env.rs:80:17
  opentelemetry_sdk::Resource::from_detectors
      at opentelemetry_sdk-0.32.1/src/resource/mod.rs:144:32
  opentelemetry_sdk::Resource::builder
      at opentelemetry_sdk-0.32.1/src/resource/mod.rs:64:23
  opentelemetry_sdk::metrics::MeterProviderBuilder::build
      at opentelemetry_sdk-0.32.1/src/metrics/meter_provider.rs:407:45
  build_default_provider
      at tests/build_provider.rs:5:20
```

scottgerring:
Fixes #3523.

`std::env::current_exe()` aborts under Miri isolation instead of returning Err, so the existing .ok() fallback in `SdkProvidedResourceDetector` was unreachable, causing any test that constructed a default SDK provider to abort under Miri.

Extract the fallback into default_service_name() and short-circuit to "unknown_service" when `cfg!(miri)` is set, preserving the normal `unknown_service:<process.executable.name>` behaviour outside Miri.

Also adds a targeted Miri smoke test (tests/miri_smoke.rs) and a minimal CI job that runs only that test under nightly Miri.

## Changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,24 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
     - name: Test
       run: bash ./scripts/test.sh
+  miri:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Harden the runner (Audit all outbound calls)
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+      with:
+        egress-policy: audit
+
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        submodules: true
+    - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9
+      with:
+        toolchain: nightly
+        components: miri
+    - name: Miri smoke test
+      run: cargo +nightly miri test --manifest-path=opentelemetry-sdk/Cargo.toml --no-default-features --features metrics --test miri_smoke
+
   lint:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,8 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
     - name: Test
       run: bash ./scripts/test.sh
+  # Targeted Miri regression check for SDK default resource construction.
+  # Broader Miri support and CI policy are tracked in #3554.
   miri:
     runs-on: ubuntu-latest
     steps:
--- a/opentelemetry-sdk/src/resource/env.rs
+++ b/opentelemetry-sdk/src/resource/env.rs
@@ -72,6 +72,24 @@ fn construct_otel_resources(s: String) -> Resource {
 #[derive(Debug)]
 pub struct SdkProvidedResourceDetector;
 
+fn default_service_name() -> Value {
+    // Fallback to unknown_service:<process.executable.name> per spec.
+    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#sdk-provided-resource-attributes
+    if cfg!(miri) {
+        return "unknown_service".into();
+    }
+
+    env::current_exe()
+        .ok()
+        .and_then(|path| {
+            path.file_name()
+                .and_then(|name| name.to_str())
+                .map(|name| format!("unknown_service:{}", name))
+        })
+        .unwrap_or_else(|| "unknown_service".to_string())
+        .into()
+}
+
 impl ResourceDetector for SdkProvidedResourceDetector {
     fn detect(&self) -> Resource {
         Resource::builder_empty()
@@ -86,19 +104,7 @@ impl ResourceDetector for SdkProvidedResourceDetector {
                             .detect()
                             .get(&Key::new(super::SERVICE_NAME))
                     })
-                    .unwrap_or_else(|| {
-                        // Fallback to unknown_service:<process.executable.name> per spec
-                        // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#sdk-provided-resource-attributes
-                        env::current_exe()
-                            .ok()
-                            .and_then(|path| {
-                                path.file_name()
-                                    .and_then(|name| name.to_str())
-                                    .map(|name| format!("unknown_service:{}", name))
-                            })
-                            .unwrap_or_else(|| "unknown_service".to_string())
-                            .into()
-                    }),
+                    .unwrap_or_else(default_service_name),
             )])
             .build()
     }
@@ -155,11 +161,15 @@ mod tests {
             .map(|v| v.to_string())
             .unwrap();
 
-        assert!(
-            service_name.starts_with("unknown_service:opentelemetry_sdk-"),
-            "Expected service name to start with 'unknown_service:opentelemetry_sdk-', got: {}",
-            service_name
-        );
+        if cfg!(miri) {
+            assert_eq!(service_name, "unknown_service");
+        } else {
+            assert!(
+                service_name.starts_with("unknown_service:opentelemetry_sdk-"),
+                "Expected service name to start with 'unknown_service:opentelemetry_sdk-', got: {}",
+                service_name
+            );
+        }
 
         temp_env::with_var(OTEL_SERVICE_NAME, Some("test service"), || {
             let with_service = SdkProvidedResourceDetector.detect();
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- Default SDK Resource construction now falls back to `unknown_service` under
+  Miri instead of calling `std::env::current_exe()`, avoiding an abort in Miri
+  isolation mode while preserving the normal
+  `unknown_service:<process.executable.name>` fallback outside Miri.
+
 ## 0.32.1
 
 Released 2026-May-23
--- a/opentelemetry-sdk/tests/miri_smoke.rs
+++ b/opentelemetry-sdk/tests/miri_smoke.rs
@@ -1,0 +1,6 @@
+#[cfg(feature = "metrics")]
+#[test]
+fn default_meter_provider_builds() {
+    let provider = opentelemetry_sdk::metrics::SdkMeterProvider::builder().build();
+    drop(provider);
+}
--- a/opentelemetry-sdk/tests/miri_smoke.rs
+++ b/opentelemetry-sdk/tests/miri_smoke.rs
@@ -1,3 +1,5 @@
+// Targeted Miri regression check for SDK default resource construction.
+// This does not imply general opentelemetry-rust Miri support; see #3554.
 #[cfg(feature = "metrics")]
 #[test]
 fn default_meter_provider_builds() {


opentelemetry-sdk/src/resource/env.rs
```diff
@@ -72,6 +72,24 @@ fn construct_otel_resources(s: String) -> Resource {
 #[derive(Debug)]
 pub struct SdkProvidedResourceDetector;
 
+fn default_service_name() -> Value {
```
scottgerring, line 75: this was getting a bit long so factored it out

.github/workflows/ci.yml
```diff
@@ -51,6 +51,25 @@ jobs:
...
+
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        submodules: true
+    - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9
+      with:
+        toolchain: nightly
+        components: miri
+    - name: Miri smoke test
+      run: cargo +nightly miri test --manifest-path=opentelemetry-sdk/Cargo.toml --no-default-features --features metrics --test miri_smoke
```
scottgerring, line 72: super basic 'run miri at all' smoke test to try catch these in the future

.github/workflows/ci.yml
```diff
@@ -51,6 +51,25 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
     - name: Test
       run: bash ./scripts/test.sh
+  miri:
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
lalitb: If the goal is to keep this path working under Miri, I think we should let this job fail CI?

lalitb (approved):
LGTM. Thanks.

.github/workflows/ci.yml
cijothomas: lets run all crates, not just sdk?

.github/workflows/ci.yml
scottgerring: 🤦  agreed

.github/workflows/ci.yml
scottgerring: i've not used miri much before, so I had a look at doing this; here's what i've found trying to run `cargo miri test` against each crate in the workspace using the same arguments as `scripts/test.sh`:

| Crate | Tests | Pass | Result | Failure reason |
|---|---|---|---|---|
| `opentelemetry` | 55 | 17 | ❌ aborts | `kqueue` - single `#[tokio::test]` for async context propagation |
| `opentelemetry-sdk` | 387 | 13 | ❌ aborts | `clock_gettime` - batch processor tests touch `std::time` |
| `opentelemetry-http` | 5 | 5 | ✅ clean | - |
| `opentelemetry-appender-log` | 6 | 1 | ❌ aborts | `clock_gettime` - first test reads system time |
| `opentelemetry-semantic-conventions` | 0 | 0 | ✅ clean | - |
| `opentelemetry-stdout` | 0 | 0 | ✅ clean | - |
| `opentelemetry-appender-tracing` | 15 | 0 | ❌ aborts | `clock_gettime` - hits on first test |
| `opentelemetry-jaeger-propagator` | 18 | 18 | ✅ clean | - |
| `opentelemetry-otlp` | 131 | 6 | ❌ aborts | `llvm.aarch64.crc32b` - gzip CRC32 uses an aarch64 hardware intrinsic Miri won't execute |
| `opentelemetry-proto` | 14 | 2 | ❌ aborts | `clock_gettime` |
| `opentelemetry-zipkin` | 13 | 7 | ❌ aborts | `clock_gettime` |
| `opentelemetry-prometheus` | 2 | 2 | ✅ clean | - |

The 8 ignored SDK tests (tokio runtime + self-diagnostics) and the `opentelemetry --no-default-features` run all fail with the same `clock_gettime` / `kqueue` errors.

### TL;DR - things that break

1. **`clock_gettime` - Miri's default isolation blocks all system clock calls. Anything that creates a `std::time::Instant`, `SystemTime`, or uses tokio's timer hits this. `-Zmiri-disable-isolation` unblocks the syscall but causes batch/async tests to hang indefinitely, because they also require real thread scheduling.

2. **`kqueue`** - One `#[tokio::test]` in `opentelemetry` for async context propagation. Same root cause as the `smoke.rs` failure in `opentelemetry-otlp`.

3. **`llvm.aarch64.crc32b`** - The OTLP gzip compression test uses a CRC32 hardware intrinsic on aarch64. Architecture-specific; might pass on x86_64 CI! 

The crates that pass cleanly are those that deal with pure data transformation: header/propagation format parsing, semantic convention constants. The crates with the most test value - SDK processors, appenders, OTLP - all depend on `std::time` or tokio, which are fundamentally incompatible with Miri's isolation model.

The `miri_smoke` test in `opentelemetry-sdk` avoids time and async entirely, exercising only the initialisation path where the original `current_exe()` bug lived. To expand Miri coverage to other crates we'd need to write tests specifically to avoid time, async, and I/O.

scottgerring: Effectively stuff that doesn't seem like it's going to benefit from miri is testable (logic, mapping - like `opentelemetry-jaeger-propagator` ). The crates where Miri would add genuine value (processors, exporters) are blocked by clock_gettime and tokio's I/O driver before they can run a single meaningful test.

If someone else has more experience here with miri i'd love to hear about it, but for now I suggest we stick with a basic smoke test that catches this regression and keep an eye on what we might improve later.

scottgerring:
@cijothomas this should be good to go!

cijothomas:
We've never explicitly committed to Miri support in this project. The Miri guard makes sense as a tactical fix, but the CI job implicitly makes Miri a supported use case going forward, which means future contributors would need to be careful about any new syscall-touching code too.

Should we make that commitment explicitly, or just land the tactical fix for now and revisit Miri CI as a separate decision?

scottgerring:
> We've never explicitly committed to Miri support in this project. The Miri guard makes sense as a tactical fix, but the CI job implicitly makes Miri a supported use case going forward, which means future contributors would need to be careful about any new syscall-touching code too.
> 
> Should we make that commitment explicitly, or just land the tactical fix for now and revisit Miri CI as a separate decision?

@cijothomas supporting miri _properly_ would be a big deal and we should absolutely defer that for the moment - there's a bunch more pressing stuff to land imho - i've created #3525 to track that. 

In the meantime ACK that the CI job might send a funny signal - that having any "miri looking" stuff might seem like a commitment. To that end, i've added some comments around the CI job and the test itself pointing at that issue. We could always drop the test/CI completely, but I kinda like having this regression catch as a beachhead for future miri support.

LMK what you think

cijothomas:
> > We've never explicitly committed to Miri support in this project. The Miri guard makes sense as a tactical fix, but the CI job implicitly makes Miri a supported use case going forward, which means future contributors would need to be careful about any new syscall-touching code too.
> > Should we make that commitment explicitly, or just land the tactical fix for now and revisit Miri CI as a separate decision?
> 
> @cijothomas supporting miri _properly_ would be a big deal and we should absolutely defer that for the moment - there's a bunch more pressing stuff to land imho - i've created #3525 to track that.
> 
> In the meantime ACK that the CI job might send a funny signal - that having any "miri looking" stuff might seem like a commitment. To that end, i've added some comments around the CI job and the test itself pointing at that issue. We could always drop the test/CI completely, but I kinda like having this regression catch as a beachhead for future miri support.
> 
> LMK what you think

Thanks. I'll open a similar issue to track related things (like wasm, no-std etc)

Issue #3528, opened 2026-06-01
Express AnyValue::Empty?


tisonkun:
The opentelemetry spec allows AnyValue to be empty - https://opentelemetry.io/docs/specs/otel/common/#anyvalue

And protobuf can represent it with tag = 0:

https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/common/v1/common.proto#L29-L31

However, `prost` and our `AnyValue` type seem to require `AnyValue` must not be None:

```rust
#[derive(Debug, Clone, PartialEq)]
#[non_exhaustive]
pub enum AnyValue {
    /// An integer value
    Int(i64),
    /// A double value
    Double(f64),
    /// A string value
    String(StringValue),
    /// A boolean value
    Boolean(bool),
    /// A byte array
    Bytes(Box<Vec<u8>>),
    /// An array of `Any` values
    ListAny(Box<Vec<AnyValue>>),
    /// A map of string keys to `Any` values, arbitrarily nested.
    Map(Box<HashMap<Key, AnyValue>>),
}

    #[cfg_attr(feature = "with-schemars", derive(schemars::JsonSchema))]
    #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
    #[cfg_attr(feature = "with-serde", serde(rename_all = "camelCase"))]
    #[derive(Clone, PartialEq, ::prost::Oneof)]
    pub enum Value {
        #[prost(string, tag = "1")]
        StringValue(::prost::alloc::string::String),
        #[prost(bool, tag = "2")]
        BoolValue(bool),
        #[prost(int64, tag = "3")]
        IntValue(i64),
        #[prost(double, tag = "4")]
        DoubleValue(f64),
        #[prost(message, tag = "5")]
        ArrayValue(super::ArrayValue),
        #[prost(message, tag = "6")]
        KvlistValue(super::KeyValueList),
        #[prost(bytes, tag = "7")]
        BytesValue(::prost::alloc::vec::Vec<u8>),
        /// Reference to the string value in ProfilesDictionary.string_table.
        ///
        /// Note: This is currently used exclusively in the Profiling signal.
        /// Implementers of OTLP receivers for signals other than Profiling should
        /// treat the presence of this value as a non-fatal issue.
        /// Log an error or warning indicating an unexpected field intended for the
[...]

PR #3529, opened 2026-06-01
refactor: leverage static str key when possible

tisonkun:
## Changes

TRIVIAL AS IS

When we are sure a Key is backed by stastic str, we can avoid allocations a bit.

Since they are all internal impls, we can do this and change it further if anything changes.

cc @cijothomas


opentelemetry-appender-log/src/lib.rs
```diff
@@ -680,7 +680,8 @@ mod any_value {
         ) -> Result<(), Self::Error> {
             let key = match key.serialize(ValueSerializer)? {
                 Some(AnyValue::String(key)) => Key::from(String::from(key)),
-                key => Key::from(format!("{key:?}")),
+                Some(key) => Key::from(format!("{key:?}")),
+                None => Key::from("None"),
```
tisonkun, line 684: Actually a proper solution is to return `AnyValue::Empty` on all the None branches here - https://github.com/open-telemetry/opentelemetry-rust/issues/3528

But that's a separated issue.

lalitb (approved):

tisonkun:
@cijothomas @lalitb shall we merge this patch now? I don't think it should take too long pending :D

cijothomas:
> @cijothomas @lalitb shall we merge this patch now? I don't think it should take too long pending :D

Sorry. Added to queue now.

PR #3530, opened 2026-06-03
feat(sdk): add otel.sdk.log.created self-observability metric

cijothomas:
## Changes

Adds the `otel.sdk.log.created` self-observability metric to the Logs SDK.
A `u64` counter is bound once per `SdkLogger` (no attributes) and incremented
on every `emit` call that passes the telemetry-suppression check.

* [x] CHANGELOG.md updated
* [x] Unit tests added / N/A
  - Behavioral conformance will be covered by Weaver Live Check
    (in-progress: open-telemetry/opentelemetry-rust-contrib#613).
    SDK control-flow paths (suppression early-return, post-shutdown
    no-op routing) are exercised by existing tests.
* [x] Public API changes added to `CHANGELOG.md` / N/A
* [x] Documentation updated / N/A

## Notes

* Gated behind the `experimental_metrics_bound_instruments` Cargo feature,
  matching the existing `otel.sdk.processor.log.processed` metric. The
  semconv definition is still `stability: development`.
* Cross-SDK alignment: matches the Java/JS/Python pattern of incrementing
  after the per-emit gates and before processor dispatch. Rust's
  telemetry-suppression early-return is SDK-specific; no other reference
  SDK has an equivalent concept.
* Semconv: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/otel/sdk-metrics.md#metric-otelsdklogcreated

### DIFF
--- a/opentelemetry-sdk/src/logs/logger.rs
+++ b/opentelemetry-sdk/src/logs/logger.rs
@@ -15,11 +15,30 @@ use opentelemetry::time::now;
 pub struct SdkLogger {
     scope: InstrumentationScope,
     provider: SdkLoggerProvider,
+
+    // Bound is not strictly needed (no attributes), but the semconv is still
+    // `development` so the metric must be feature-gated; reuse the same
+    // `experimental_metrics_bound_instruments` flag as the other SDK
+    // self-observability metrics for consistency.
+    #[cfg(feature = "experimental_metrics_bound_instruments")]
+    log_created_counter: opentelemetry::metrics::BoundCounter<u64>,
 }
 
 impl SdkLogger {
     pub(crate) fn new(scope: InstrumentationScope, provider: SdkLoggerProvider) -> Self {
-        SdkLogger { scope, provider }
+        #[cfg(feature = "experimental_metrics_bound_instruments")]
+        let log_created_counter = opentelemetry::global::meter("otel.sdk")
+            .u64_counter("otel.sdk.log.created")
+            .with_description("The number of logs submitted to enabled SDK Loggers.")
+            .with_unit("{log_record}")
+            .build()
+            .bind(&[]);
+        SdkLogger {
+            scope,
+            provider,
+            #[cfg(feature = "experimental_metrics_bound_instruments")]
+            log_created_counter,
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/logs/logger.rs
```diff
@@ -35,6 +54,10 @@ impl opentelemetry::logs::Logger for SdkLogger {
         if Context::is_current_telemetry_suppressed() {
             return;
         }
+
+        #[cfg(feature = "experimental_metrics_bound_instruments")]
+        self.log_created_counter.add(1);
```
lalitb, line 67: nit -   Could we increment this only when the logger is actually enabled, so no-op loggers don’t get counted?

lalitb (approved):

TommyCpp (approved):

cijothomas:
Need to make some tweaks before merging.

opentelemetry-sdk/src/logs/logger.rs
cijothomas: This is something that I have been thinking about for.a while. I think the best option for this metric is to count everything that is given to the SDK, irrespective of what happens to it afterwards. So this is truly measuring how much log was given to SDK. and an operator can use this in conjunction with processor metrics/exporter metrics etc. to paint the full picture. 

if we do not count situations where logger is Disabled(), filtered inside OTel, then we'd lose that visibility. Ofcourse we still lose visibility if the filtering occurs in `tracing` itself, but that is not something OTel can solve.

The sem. conventions wording is not very clear and implementations are inconsistent from what I can tell. Working on proposing a change to semantic convention in parallel.
This PR can be hold off until that convention change itself is approved.

cijothomas: I've opened a semantic-convention proposal to clarify this: open-telemetry/semantic-conventions#3939. It proposes counting every record submitted to an enabled Logger at intake, before per-record filtering, with the one exclusion being a Logger disabled via LoggerConfig.enabled=false (which behaves as a No-op and wouldn't be counted).

That disabled case doesn't apply to Rust today anyway: we don't implement LoggerConfig, so there's no disabled-Logger state (and no per-record severity/trace filtering). Every SdkLogger is enabled, so counting every submitted record here lines up with the proposal. If the proposal lands and Rust later adds Logger disabling, the counter would move behind that enabled check.

lalitb (approved):

PR #3538, opened 2026-06-06
fix(sdk): stop batch processors from panicking on worker thread failures

### LINKED ISSUE #3375: Shutdown paths panic on thread join failures instead of returning errors

bryantbiggs:
Fixes #3375

## Changes

`BatchSpanProcessor` and `BatchLogProcessor` panicked during **shutdown** via `.unwrap()` on thread operations, and the shutdown path also mishandled two non-panic cases. This fixes the shutdown path **without any breaking API change** (signatures are unchanged; `shutdown()` already returns `OTelSdkResult`).

`shutdown_with_timeout` now matches on `recv_timeout` explicitly, with one behavior per outcome:

- **`Ok(result)` — worker exported the final batch and replied:** join the worker, then return the worker's own export result. Previously the result was discarded with `|_|`, so a failed final shutdown export was silently reported as `Ok(())`.
- **`Disconnected` — worker dropped the reply sender without responding (it panicked mid-shutdown, e.g. during the final export or the exporter's own shutdown):** join the worker so the original panic message is surfaced as `OTelSdkError::InternalFailure`, instead of a masking *second* panic or an opaque "disconnected channel" error.
- **`Timeout` — worker still running:** return `OTelSdkError::Timeout` without joining (joining could block past the timeout). The worker-leak / data-loss on this arm is fundamental — the worker is parked in an uninterruptible `block_on(export())` that cannot be cancelled — and is tracked, together with the rest of the export-cancellation class, in #3542.

Supporting changes:

- A poisoned handle mutex is recovered via `into_inner()` rather than `unwrap()`, extracted into a `take_worker_handle()` helper used on both the clean and disconnected arms, and the lock is released before the blocking `join()`.
- The shared `join_worker_thread()` helper (in `util.rs`) converts a worker-thread panic payload into `OTelSdkError::InternalFailure`, preserving the original panic message.
- New tests per processor: `shutdown_returns_final_export_error` (a failing final export propagates) and `shutdown_surfaces_worker_panic_during_export` (a worker panic is surfaced as an error, with no second panic), alongside the existing `join_worker_thread` helper tests.

### Constructor spawn failure: intentionally still fail-fast

The thread-spawn `.expect()` in the constructors is **kept as a panic**, per [review discussion](https://github.com/open-telemetry/opentelemetry-rust/pull/3538#discussion_r3367961873). A spawn failure at construction is an unrecoverable setup-time error — without the worker the processor can never export anything — and the [OTel error-handling spec](https://opentelemetry.io/docs/specs/otel/error-handling/) permits failing fast on initialization. The panic now surfaces the underlying OS error (usually resource exhaustion) instead of a bare `"Thread spawn failed."`.

Returning a `Result` from construction is the preferred long-term fix, but it is a breaking change to the stable Logs/Metrics SDK; it is tracked separately for the OTLP exporter builder / v2.0 via #2690 and #3462.

* [x] Changes in public API reviewed (if applicable) — no public API changes; all signatures unchanged

### DIFF
--- a/opentelemetry-sdk/src/logs/batch_log_processor.rs
+++ b/opentelemetry-sdk/src/logs/batch_log_processor.rs
@@ -171,6 +171,26 @@ impl Debug for BatchLogProcessor {
     }
 }
 
+/// Joins the batch log processor's background worker thread during shutdown,
+/// converting a panic in that thread into an [`OTelSdkError`] instead of
+/// propagating it as a second panic that would mask the original failure.
+///
+/// Returns `Ok(())` when there is no worker to join — for example when the
+/// thread could not be spawned at construction time (see
+/// [`BatchLogProcessor::new`]).
+fn join_worker_thread(handle: Option<thread::JoinHandle<()>>) -> OTelSdkResult {
+    let Some(handle) = handle else {
+        return Ok(());
+    };
+    handle.join().map_err(|payload| {
+        // Surface the original panic message rather than losing it.
+        OTelSdkError::InternalFailure(format!(
+            "the batch log processor worker thread panicked during shutdown: {}",
+            crate::util::panic_message(&*payload)
+        ))
+    })
+}
+
 impl LogProcessor for BatchLogProcessor {
     fn emit(&self, record: &mut SdkLogRecord, instrumentation: &InstrumentationScope) {
         let result = self
@@ -309,12 +329,18 @@ impl LogProcessor for BatchLogProcessor {
                 receiver
                     .recv_timeout(timeout)
                     .map(|_| {
-                        // join the background thread after receiving back the
-                        // shutdown signal
-                        if let Some(handle) = self.handle.lock().unwrap().take() {
-                            handle.join().unwrap();
-                        }
-                        OTelSdkResult::Ok(())
+                        // Take the worker handle out and release the lock before
+                        // the (blocking) join. This lock is only ever held here
+                        // across an infallible `take()`, so poisoning is
+                        // practically unreachable; recover the guard defensively
+                        // anyway, since a poisoned lock still guards a valid
+                        // handle and shutdown should not fail over it.
+                        let handle = self
+                            .handle
+                            .lock()
+                            .unwrap_or_else(|poisoned| poisoned.into_inner())
+                            .take();
+                        join_worker_thread(handle)
                     })
                     .map_err(|err| match err {
                         RecvTimeoutError::Timeout => {
@@ -381,6 +407,16 @@ impl BatchLogProcessor {
         let current_batch_size = Arc::new(AtomicUsize::new(0));
         let current_batch_size_for_thread = current_batch_size.clone();
 
+        // Spawn the background worker thread. A spawn failure (e.g. the OS
+        // refusing to create the thread under resource exhaustion) is an
+        // unrecoverable setup-time error: without the worker this processor can
+        // never export anything. Fail fast — the spec permits failing on
+        // initialization — and surface the underlying OS error so the cause is
+        // diagnosable instead of a bare "Thread spawn failed.".
+        //
+        // TODO(#3375): return `Result<Self, OTelSdkError>` from this
+        // constructor at the next breaking change so callers can react to a
+        // spawn failure (see also #2690 / #3462).
         let handle = thread::Builder::new()
             .name("OpenTelemetry.Logs.BatchProcessor".to_string())
             .spawn(move || {
@@ -528,7 +564,9 @@ impl BatchLogProcessor {
                     name: "BatchLogProcessor.ThreadStopped"
                 );
             })
-            .expect("Thread spawn failed."); //TODO: Handle thread spawn failure
+            .unwrap_or_else(|err| {
+                panic!("OpenTelemetry batch log processor failed to spawn its worker thread: {err}")
+            });
 
         // Self-diagnostics: create the otel.sdk.processor.log.processed counter
         #[cfg(feature = "experimental_metrics_bound_instruments")]
@@ -850,6 +888,31 @@ mod tests {
     use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
     use std::sync::{Arc, Mutex};
     use std::time::Duration;
+
+    #[test]
+    fn join_worker_thread_none_is_ok() {
+        assert!(super::join_worker_thread(None).is_ok());
+    }
+
+    #[test]
+    fn join_worker_thread_joins_clean_thread() {
+        let handle = std::thread::spawn(|| {});
+        assert!(super::join_worker_thread(Some(handle)).is_ok());
+    }
+
+    #[test]
+    fn join_worker_thread_converts_panic_to_error() {
+        // A panicked worker must become an error during shutdown, never a
+        // second panic, and the original panic message must be preserved.
+        let handle = std::thread::spawn(|| panic!("boom in worker"));
+        match super::join_worker_thread(Some(handle)) {
+            Err(crate::error::OTelSdkError::InternalFailure(msg)) => {
+                assert!(msg.contains("panicked during shutdown"), "got: {msg}");
+                assert!(msg.contains("boom in worker"), "got: {msg}");
+            }
+            other => panic!("expected InternalFailure, got {other:?}"),
+        }
+    }
 
     #[test]
     fn test_default_const_values() {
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -527,8 +592,18 @@ impl BatchLogProcessor {
                 otel_debug!(
                     name: "BatchLogProcessor.ThreadStopped"
                 );
-            })
-            .expect("Thread spawn failed."); //TODO: Handle thread spawn failure
+            }) {
+            Ok(handle) => Some(handle),
```
cijothomas: This may be undesirable than the current panic. While OTel does not allow panic/exceptions, there are actually allowed (and encouraged) in startup/setup code where fail-fast is better than continue to run - There is no recovery - if thread is not started, then this processor will not be able to export anything ever.
I'd keep the current panic. (Ideally the builder should have returned Result, but since this is part of stable API, we can't change return type until next major version)

cijothomas: @bryantbiggs Lets remove *this* aspect from this PR, and rest should be reviewable/mergeable.
I am also open to any other ideas of how to handle the spawn failure?

bryantbiggs: Makes sense!

I've dropped the spawn-failure handling and its born-dead bookkeeping. What's left is just the shutdown fix — the actual bug in #3375 — where `join().unwrap()` fired a *second* panic that masked the worker's original one; that path now returns `OTelSdkError` (and recovers a poisoned lock instead of unwrapping), with no API change.

One small thing: I kept the panic but made it include the underlying OS error instead of the bare `"Thread spawn failed."`, since it's almost always resource exhaustion and that detail is worth surfacing. Happy to revert that too if you'd rather leave it untouched.

opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -309,12 +329,18 @@ impl LogProcessor for BatchLogProcessor {
...
+                        // across an infallible `take()`, so poisoning is
+                        // practically unreachable; recover the guard defensively
+                        // anyway, since a poisoned lock still guards a valid
+                        // handle and shutdown should not fail over it.
+                        let handle = self
+                            .handle
+                            .lock()
+                            .unwrap_or_else(|poisoned| poisoned.into_inner())
+                            .take();
+                        join_worker_thread(handle)
```
lalitb: This only joins the worker after it sends the shutdown response. If the worker panics before that, like during the final export or exporter shutdown, `recv_timeout` will return `Disconnected` and we never join the thread. That means we still lose the real panic message. Can we also join the worker in the disconnected case before returning the shutdown error?

lalitb: Same pattern for span processor too.

opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -309,12 +329,18 @@ impl LogProcessor for BatchLogProcessor {
                 receiver
                     .recv_timeout(timeout)
                     .map(|_| {
```
lalitb: This is pre-existing, but since this PR is already changing the shutdown error path: this still ignores the result sent back by the worker. If the final export during shutdown returns `Err(...)`, we receive it here but drop it because the closure uses `|_|`. Then if the worker thread joins cleanly, shutdown can still return `Ok(())`

Can we keep the received result and return it after joining the worker, so shutdown still reports final export failures?

Similar for span processor.

opentelemetry-sdk/src/logs/batch_log_processor.rs
bryantbiggs: Done in https://github.com/open-telemetry/opentelemetry-rust/pull/3538/commits/93592fae2235cdbfafb5714ebf17c7996bc7dc0e. `shutdown_with_timeout` now matches on `recv_timeout` explicitly, and the `Disconnected` arm joins the worker via `join_worker_thread(self.take_worker_handle())` — which converts the worker's panic payload into `OTelSdkError::InternalFailure` with the original message, instead of returning the opaque "disconnected channel" error. Applied to both `BatchLogProcessor` and `BatchSpanProcessor`. New `shutdown_surfaces_worker_panic_during_export` tests in each module assert the real panic message is surfaced and that there is no second panic.

opentelemetry-sdk/src/logs/batch_log_processor.rs
bryantbiggs: Done in https://github.com/open-telemetry/opentelemetry-rust/pull/3538/commits/93592fae2235cdbfafb5714ebf17c7996bc7dc0e. The `Ok(result)` arm now joins the worker and then returns the worker's own result, so a failed final shutdown export propagates instead of being dropped by the previous `|_|`. `shutdown_returns_final_export_error` in both processors asserts `shutdown()` returns the export error rather than `Ok(())`. The `Timeout` arm still returns without joining since the worker may still be running; the worker-leak on that arm is the uninterruptible-`block_on` case tracked in #3542.

scottgerring:
Nice work! I've applied the same panic-surfacing idea to the async-runtime processors in #3413 - the join-based shutdown design there means there's no separate Disconnected case, but the `JoinError::Panic` payload was being silently dropped in the same way. 👍

PR #3540, opened 2026-06-07
feat(api): add StringValue::from_static_str

juchiast:
## Changes

Add `StringValue::from_static_str`, make it possible to create a StringValue at compile time.

### DIFF
--- a/opentelemetry/src/common.rs
+++ b/opentelemetry/src/common.rs
@@ -259,6 +259,11 @@ impl AsRef<str> for StringValue {
 }
 
 impl StringValue {
+    /// Create a new const `StringValue`.
+    pub const fn from_static_str(value: &'static str) -> Self {
+        StringValue(OtelString::Static(value))
+    }
+
     /// Returns a string slice to this value
     pub fn as_str(&self) -> &str {
         self.0.as_str()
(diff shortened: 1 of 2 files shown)


opentelemetry/src/common.rs
```diff
@@ -259,6 +259,11 @@ impl AsRef<str> for StringValue {
 }
 
 impl StringValue {
+    /// Create a new const `StringValue`.
+    pub const fn from_static_str(value: &'static str) -> Self {
```
lalitb, line 272: Could we add a small doctest showing the actual const use case here?

lalitb (approved):

opentelemetry/src/common.rs
juchiast: added

juchiast: Unfortunately, we can't make a KeyValue at compile time with the current initializer. I changed the example.

opentelemetry/src/common.rs
```diff
@@ -259,6 +259,20 @@ impl AsRef<str> for StringValue {
 }
 
 impl StringValue {
+    /// Create a new const `StringValue`.
+    ///
+    /// This is useful for making static label values:
```
cijothomas, line 264: let labels = &[KeyValue::new("key", "value")];

^ couldn't we already this today?

PR #3541, opened 2026-06-07
fix(metrics): export canonicalized attribute order from ValueMap

shuheiktgw:
## Changes

This PR addresses the issue originally described in quickwit-oss/quickwit#6465. In short, asynchronous counters that use delta temporality may currently export incorrect values.

For asynchronous counters using delta temporality, the SDK computes deltas by looking up the previously exported value by the exported attribute set. That delta calculation happens in [`precomputed_sum`](https://github.com/open-telemetry/opentelemetry-rust/blob/7cf7d697626a591394d75c1a7efddcb55198caa6/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs#L105-L115). The problematic path is [`ValueMap::drain_and_reset`](https://github.com/open-telemetry/opentelemetry-rust/blob/7cf7d697626a591394d75c1a7efddcb55198caa6/opentelemetry-sdk/src/metrics/internal/mod.rs#L341-L376), which drains tracker entries for asynchronous instruments. If the same logical metric stream can be exported with alternating attribute orders, the previous value is not found and the delta calculation falls back to zero, so the full cumulative value can be exported as the delta.

For example, consider an asynchronous counter with two attributes, such as `quickwit_indexing_processed_bytes` with labels `index` and `docs_processed_status`. The caller can record the attributes as `[index, docs_processed_status]`, while the canonical sorted order is `[docs_processed_status, index]`. If both orders are present as keys for the same tracker, one collection may export value `100` with `[index, docs_processed_status]` and store that as the previously reported value. A later collection may see the same tracker as `[docs_processed_status, index]` with current value `120`. Since the previous value was stored under the other attribute order, the lookup misses and the SDK exports delta `120` instead of the expected `20`.

This PR makes `TrackerMaps` keep canonical attribute keys as the source of truth for collection/export, while retaining raw attribute-order aliases only as lookup fast paths. `Key` now accepts the caller-provided attributes and lazily computes the canonical key only when needed. As a result, `ValueMap` exports a stable canonical attribute order while preserving fast lookups for repeated measurements in the original order.

### Performance

This PR keeps the measurement lookup path equivalent to the previous implementation: it first checks the caller-provided attribute order, and only on a miss does it compute and look up the canonicalized key. Since canonicalization remains lazy and the common raw-key hit path is unchanged, I do not expect a performance regression there.

The latest [`compare benchmarks (PR)` run](https://github.com/open-telemetry/opentelemetry-rust/actions/runs/27499152614/job/81278686821?pr=3541) completed successfully for `d0764a5` and did not flag a performance regression; the metrics counter/add benchmarks relevant to this change stayed within benchmark noise.

One thing worth calling out is memory usage. This PR introduces a new `canonical` map, so memory consumption can increase. However, the canonical map does not duplicate attribute arrays or tracker state: canonical keys and tracker entries are shared through `Arc`. The incremental cost is therefore an additional `HashMap` entry per logical attribute set, plus the canonical map’s bucket/control storage, rather than a full duplicate of the aggregation state.

* [x] Changes in public API reviewed (if applicable; no public API changes)


scottgerring:
Hey @shuheiktgw thanks for consolidating this! 
Quickly re-labelling and triggering the benchmark runs.

shuheiktgw:
@cijothomas @scottgerring I’ve cleaned up the code, added a few more tests, and confirmed through the benchmarks that there’s no performance regression. Could you take a look when you get a chance? Thanks!

cijothomas:
@shuheiktgw Thanks. The issue is clear now. I'll get to full review shortly.

shuheiktgw:
Thank you, @cijothomas!

opentelemetry-sdk/src/metrics/internal/tracker_maps.rs
```diff
@@ -0,0 +1,384 @@
+use std::cell::OnceCell;
```
cijothomas: @shuheiktgw Sharing the comment here while going through this file... I am not sure if we need such big changes. IIUC, the issue is specifically when SDK has to do Cumulative -> Delta conversion and for that it has to diff the values from current iteration vs previous. The look-up for that is done using attributes, which could be in sorted or incoming order, and hence causes incorrect calculation.
Could we do a attributes.sort()+dedup() in the pre-computed-sum's delta() method where it is doing the lookup? That should fix the issue.

I think you have a solid unit test to prove the issue - lets try a simple fix (see above), and see if the test passes.

cijothomas (changes requested):
Thanks for catching the bug! This was indeed a edge we missed.
Requesting changes as I think this can be fixed with a much smaller change. See
https://github.com/open-telemetry/opentelemetry-rust/pull/3541/changes#r3480078693 for details.

shuheiktgw:
Thank you for the feedback, @cijothomas ! I’ve updated the PR. Could you take another look?

opentelemetry-sdk/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
   `queue_full` and `already_shutdown`, enabling operators to distinguish
   successful processing from drops due to full queue or post-shutdown emits.
   ([#3514](https://github.com/open-telemetry/opentelemetry-rust/pull/3514))
+- Fixed delta `ObservableCounter` streams treating the same logical attribute
+  set in a different order as a new stream and re-exporting the cumulative value
+  as the delta.
```
cijothomas: - Fixed asynchronous counters (`ObservableCounter`, `ObservableUpDownCounter`) using delta temporality reporting incorrect deltas when observed attributes were recorded in an unsorted key order.

^ suggested phrasing for end user facing changelog.

cijothomas (approved):
LGTM, suggested to rephrase changelog to make it more easy to follow for end users.

shuheiktgw:
Thank you, @cijothomas. I’ve updated the note so could you take another look when you have a chance?

PR #3545, opened 2026-06-09
fix(sdk): normalize metric attribute keys

shuheiktgw:
## Changes

This is a second iteration of https://github.com/open-telemetry/opentelemetry-rust/pull/3541 with a smaller change set. For the detailed background, please see the original PR.

The issue is that `ValueMap::measure` can store attributes in the `trackers` HashMap before they have been sorted and deduplicated. For precomputed sums, delta collection later compares the current value with the previously reported value using `reported.get(&attributes)` in [`precomputed_sum.rs`](https://github.com/open-telemetry/opentelemetry-rust/blob/25717768b2b94822b5e54b7950555bbbd632cbf2/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs#L109). If the stored attributes are not canonicalized consistently, that lookup can miss the previous value and report an incorrect delta for `ObservableCounter` instruments using delta temporality.

This version keeps the fix simple by sorting and deduplicating all non-empty attributes before looking up or inserting a tracker, and by using the same canonical key for bound-instrument tracker insertion. If keeping the previous optimization is important, I would appreciate guidance on what performance measurements would be most useful.

* [x] Changes in public API reviewed (not applicable; no public API changes)

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/mod.rs
+++ b/opentelemetry-sdk/src/metrics/internal/mod.rs
@@ -133,19 +133,12 @@ where
             return;
         }
 
+        let sorted_attrs = sort_and_dedup(attributes);
+
         let Ok(trackers) = self.trackers.read() else {
             return;
         };
 
-        // Try to retrieve and update the tracker with the attributes in the provided order first
-        if let Some(tracker) = trackers.get(attributes) {
-            tracker.aggregator.update(value);
-            tracker.has_been_updated.store(true, Ordering::Release);
-            return;
-        }
-
-        // Try to retrieve and update the tracker with the attributes sorted.
-        let sorted_attrs = sort_and_dedup(attributes);
         if let Some(tracker) = trackers.get(sorted_attrs.as_slice()) {
             tracker.aggregator.update(value);
             tracker.has_been_updated.store(true, Ordering::Release);
@@ -159,12 +152,9 @@ where
             return;
         };
 
-        // Recheck both the provided and sorted orders after acquiring the write lock
-        // in case another thread has pushed an update in the meantime.
-        if let Some(tracker) = trackers.get(attributes) {
-            tracker.aggregator.update(value);
-            tracker.has_been_updated.store(true, Ordering::Release);
-        } else if let Some(tracker) = trackers.get(sorted_attrs.as_slice()) {
+        // Recheck after acquiring the write lock in case another thread has
+        // pushed an update in the meantime.
+        if let Some(tracker) = trackers.get(sorted_attrs.as_slice()) {
             tracker.aggregator.update(value);
             tracker.has_been_updated.store(true, Ordering::Release);
         } else if self.is_under_cardinality_limit() {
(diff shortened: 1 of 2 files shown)


opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -133,19 +133,12 @@ where
             return;
         }
 
+        let sorted_attrs = sort_and_dedup(attributes);
```
Mallets: I don't think there is the need of moving `sorted_attrs` here before the `read()`.

opentelemetry-sdk/src/metrics/internal/mod.rs
```diff
@@ -137,15 +137,8 @@ where
             return;
         };
 
-        // Try to retrieve and update the tracker with the attributes in the provided order first
```
cijothomas, line 140: this is intentional to do lookup in the provided order to avoid sort/allocation in the hot path. We should not change this and regress perf.

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-rust/pull/3545/changes#r3381437889 
We cannot regress the perf on hot path. 

Can you send a minimal repro/test-case demonstrating the issue, so we can do a targeted fix, as needed, and avoid regressing entire Metrics hot path.

shuheiktgw:
Hi @carlosalberto 👋 Thanks for the comment!

I created [a draft PR](https://github.com/open-telemetry/opentelemetry-rust/pull/3546) to demonstrate the issue. Please try running the test code in that PR. It fails probabilistically. Also, in [my previous PR](https://github.com/open-telemetry/opentelemetry-rust/pull/3541), I actually tried to preserve the fast path, so please take a look if you’re interested.

Thanks!

scottgerring:
> Hi @carlosalberto 👋 Thanks for the comment!
> 
> I created [a draft PR](https://github.com/open-telemetry/opentelemetry-rust/pull/3546) to demonstrate the issue. Please try running the test code in that PR. It fails probabilistically. Also, in [my previous PR](https://github.com/open-telemetry/opentelemetry-rust/pull/3541), I actually tried to preserve the fast path, so please take a look if you’re interested.
> 
> Thanks!

Hey @shuheiktgw , if you have a test that reproduces the issue can you include it in this PR so the PR shows that it is resolved? Likewise if you have a variant that doesn't regress performance perhaps you can highlight that variant in the open PR. This'll make it easier for us to reason about the proposal and its impact!

shuheiktgw:
Hi @scottgerring, thanks for the guidance! I’ve added the test to the PR and included a performance section to highlight the relevant considerations in the description. Please take a look when you get a chance 🙂

https://github.com/open-telemetry/opentelemetry-rust/pull/3541

cijothomas:
> Hi @scottgerring, thanks for the guidance! I’ve added the test to the PR and included a performance section to highlight the relevant considerations in the description. Please take a look when you get a chance 🙂
> 
> #3541

I don't see a test added in this PR? Can you ensure all commits are pushed so we can take a look.

PR #3503, opened 2026-05-07
feat: Implement scope_info_enabled

ArthurSens:
Related to #3474, but entirely focused on the implementation of Instrumentation Scope.

Here I'm doing quite a few things:
* I'm removing the exposition of the `otel_scope_info` metric, which was removed from the spec long ago.
* I'm switching the function signature from `without_scope_info` to `scope_info_enabled`, following the latest discussions we had at https://github.com/open-telemetry/opentelemetry-specification/pull/5056
* I'm handling the transformation of the instrumentation scope to regular labels

Disclosure: Although I've done Rust before, I'm not great at it. I do rely on LLMs to write most of the code.

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #3548, opened 2026-06-10
perf: Do not allocate for scope attributes that collide with reserved ones

ArthurSens:
Follow up from https://github.com/open-telemetry/opentelemetry-rust/pull/3503#discussion_r3388988776

## Changes

Prevent allocation in `format!` for labels that we know will collide with reserved ones.


opentelemetry-prometheus/src/lib.rs
ArthurSens, line 432: I considered removing this, but if we get something like `schema-url`, it will become `schema_url` after sanitization, and we need to drop it as well.

Is there a way that we can avoid allocation when they match a regex? I assume not, since regex is usually more resource-intensive than an exact match...?

cijothomas: agreed. this PR is good as-is.

cijothomas (approved):

PR #3549, opened 2026-06-10
fix(prometheus-exporter): Concatenate metric attributes with Scope during collisions

ArthurSens:
Follow up from https://github.com/open-telemetry/opentelemetry-rust/pull/3503#discussion_r3389029168

## Changes

Small change so `extra` labels always have preference over metric attributes in `get_attributes`. From my understanding, this will make sure Scope Name/Version/SchemaURL and attributes will always superseed metric attributes

### DIFF
--- a/opentelemetry-prometheus/src/lib.rs
+++ b/opentelemetry-prometheus/src/lib.rs
@@ -377,28 +377,64 @@ impl prometheus::core::Collector for Collector {
 /// It sanitizes invalid characters and handles duplicate keys (due to
 /// sanitization) by sorting and concatenating the values following the spec.
 fn get_attrs(kvs: &mut dyn Iterator<Item = (&Key, &Value)>, extra: &[LabelPair]) -> Vec<LabelPair> {
-    let mut keys_map = BTreeMap::<String, Vec<String>>::new();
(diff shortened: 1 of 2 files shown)


ArthurSens:
Do we need a changelog entry if the bug wasn't released 🤔 ?

opentelemetry-prometheus/src/lib.rs
```diff
@@ -380,6 +380,10 @@ fn get_attrs(kvs: &mut dyn Iterator<Item = (&Key, &Value)>, extra: &[LabelPair])
     let mut keys_map = BTreeMap::<String, Vec<String>>::new();
     for (key, value) in kvs {
         let key = utils::sanitize_prom_kv(key.as_str());
+        if extra.iter().any(|label| label.name() == key) {
```
cijothomas: The spec says this must be concatenated, not overridden.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-attributes


> In such cases, the values MUST be concatenated together, separated by ;, and ordered by the lexicographical order of the original keys.

ArthurSens: whoops, my bad

cijothomas (approved):
LGTM. no changelog needed

PR #3556, opened 2026-06-14
docs: clarify code-based config takes precedence over env vars

cijothomas:
Updates CONTRIBUTING.md to reflect the decision in #1225: when a setting is provided both via `with_xxx` and an environment variable, the code-based value wins. Users who want env-var-based runtime configuration should leave the corresponding `with_xxx` calls out of their code.

The existing wording still listed env vars as higher priority, which is the opposite of the agreed direction.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -284,10 +284,7 @@ For users that want to implement their own exporters. It's RECOMMENDED to wrap a
 
 ### Priority of configurations
 
-OpenTelemetry supports multiple ways to configure the API, SDK and other components. The priority of configurations is as follows:
-
-- Environment variables
-- Compiling time configurations provided in the source code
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -284,10 +284,9 @@ For users that want to implement their own exporters. It's RECOMMENDED to wrap a
 
 ### Priority of configurations
 
-OpenTelemetry supports multiple ways to configure the API, SDK and other components. The priority of configurations is as follows:
+OpenTelemetry supports multiple ways to configure the API, SDK and other components. When the same setting is provided through more than one mechanism, code-based configuration (e.g. `with_xxx` builder methods) takes precedence over environment variables. Users who want runtime configuration via environment variables should leave the corresponding `with_xxx` calls out of their code.
 
-- Environment variables
-- Compiling time configurations provided in the source code
+See [#1225](https://github.com/open-telemetry/opentelemetry-rust/issues/1225) for the discussion that led to this decision.
```
utpilla: Same comments as in https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/654

cijothomas: addressed in latest. PTAL

utpilla (approved):

PR #3558, opened 2026-06-18
fix(sdk): count batch items before enqueue to fix flush/shutdown race

### LINKED ISSUE #3453: [Bug]: BatchSpanProcessor` and BatchLogProcessor can miss already-enqueued items during force_flush() / shutdown()
## Description

There is a pre-existing race in the thread-based `BatchSpanProcessor` and `BatchLogProcessor` around `force_flush()` and `shutdown()`.

On the producer side, items are enqueued into the data channel first and only then reflected in `current_batch_size`. The consumer side uses a snapshot of `current_batch_size` to decide whether there is work to drain.

That leaves a window where:
1. `try_send(...)` has already succeeded, so the item is in the channel
2. `current_batch_size` has not been incremented yet
3. `force_flush()` or `shutdown()` runs concurrently
4. the worker snapshots `current_batch_size == 0`
5. the worker drains nothing and returns success

In that case:
- `force_flush()` may return before all already-enqueued items are exported
- `shutdown()` may miss an already-enqueued item and then exit, potentially dropping it

## Why this matters

The current batch export helpers use `current_batch_size` as the source of truth for whether work is pending, but the real source of truth is split across:
- the channel, which contains the actual queued items
- the counter, which is only updated after enqueue

Because enqueue and counting are not atomic as a pair, `force_flush()` / `shutdown()` do not currently guarantee draining all items that were already successfully enqueued before the flush/shutdown call races with the producer.

## Current status

PR #3441 fixes a different bug: overdraining past the snapshotted target, underflowing `current_batch_size`, and later spinning on the wrapped counter. That fix looks correct, but it does not address this pre-existing flush/shutdown gap.

## Reproduction sketch

A simplified sequence:

1. Producer thread successfully does `try_send(item)`
2. Before it executes `current_batch_size.fetch_add(1, ...)`, another thread calls `force_flush()` or `shutdown()`
3. Worker snapshots `current_batch_size == 0`
4. Worker decides there is nothing to export and returns `Ok(())`
5. The already-enqueued item remains in the channel
6. In the `shutdown()` case, the worker may exit immediately afterward, dropping that item

## Expected behavior

- `force_flush()` should not return success until all items that were already successfully enqueued are exported
- `shutdown()` should not exit before exporting all items that were already successfully enqueued

## Possible directions

Some possible approaches:
- add explicit retry/wait logic in `force_flush()` / `shutdown()` when the counter snapshot is zero but the queue may still contain newly enqueued items
- change the synchronization/design so flush/shutdown can observe both enqueue completion and pending-item accounting consistently
- avoid relying solely on `current_batch_size` as the decision point for whether drain work exists

## Affected code

- `opentelemetry-sdk/src/trace/span_processor.rs`
- `opentelemetry-sdk/src/logs/batch_log_processor.rs`

## Notes

This is specifically about the thread-based batch processors, not the async-runtime variants.

0.31.1

0.31.1

_No response_

### Relevant log output

```shell
-
```

### Tip

Wassbdr:
Fixes #3453

## Problem

In the thread-based `BatchSpanProcessor` and `BatchLogProcessor`, the producer (`on_end()` / `emit()`) does `try_send(item)` into the data channel **first** and only then `current_batch_size.fetch_add(1)`. The worker uses a snapshot of `current_batch_size` (`load(Acquire)`) as the drain *target* for `force_flush()` / `shutdown()`.

If a flush/shutdown drain snapshots the counter in the window **between** a successful `try_send` and the `fetch_add`, it under-counts: the worker drains nothing (or too little) and returns success. As a result:

- `force_flush()` can return before an already-enqueued item is exported.
- `shutdown()` can miss an already-enqueued item and then exit, dropping it.

## Fix

Increment `current_batch_size` **before** enqueueing, and revert it (`fetch_sub`) when the send fails (`Full` or `Disconnected`). Producer-side only; the drain logic is untouched.

This flips the counter's invariant from "may under-count the channel" (lossy) to **"counter is always ≥ channel depth"** (an upper bound). A flush/shutdown drain can therefore no longer under-count items already in the queue.

## Why the drain logic doesn't need to change

The drain already tolerates a *transient over-count* (an item counted but not yet enqueued), thanks to the #3441 fix:

- `batch_limit = max_export_batch_size.min(target - total_exported)`
- break out of the loop on an empty chunk
- `fetch_sub` only the count actually exported

So when the worker snapshots a `target` that includes a not-yet-enqueued item, it exports what's in the channel, breaks on the empty chunk, and leaves that item's `+1` in the counter to be drained on a later cycle. No underflow, no spin. **#3441's drain logic and its regression test are left unchanged.**

### Race interleavings considered

1. **Original bug** (sent but not counted) — eliminated by construction (count precedes send).
2. **Symmetric race** (counted but not yet sent) during a flush — drain exports what's present, breaks on empty chunk, the phantom `+1` survives and is drained next cycle. The item wasn't "enqueued before the flush" in any happens-before sense, so missing it is within contract.
3. **Shutdown** — if `try_send` Ok happens-before the `shutdown()` call, the `fetch_add` (sequenced-before the send) is visible to the worker via the mutex-backed control channel → item is both counted and in the channel → exported.
4. **Queue full / disconnected** — `fetch_add` then send fails then `fetch_sub`: transient over-count of 1, handled like case 2; dropped-count accounting unchanged.

## Tests

- `batchspanprocessor_drain_handles_counted_but_not_yet_enqueued_spans` — drives the drain helper with counter=2 / channel depth=1, asserts 1 exported, counter stays at 1 (no underflow), and the late item is exported on the next cycle.
- `batchspanprocessor_on_end_reverts_count_when_queue_full` / `test_batch_log_processor_emit_reverts_count_when_queue_full` — overflow the queue while the exporter is blocked, assert dropped items don't remain counted as pending and the counter settles to 0 after flush. (Verified these fail without the queue-full revert.)
- `batchspanprocessor_all_spans_accounted_for` — spans analog of the existing logs stress test (4 threads × 100k), asserting `received + dropped == emitted`.

`cargo test -p opentelemetry_sdk --all-features --lib`, `cargo clippy --all-targets --all-features -- -Dwarnings`, and `cargo fmt --all -- --check` all pass.

### DIFF
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -1524,7 +1524,7 @@ mod tests {
 
         let processor = BatchSpanProcessor::new(exporter, config);
 
-        let total_spans_per_thread = 100_000;
+        let total_spans_per_thread = 10_000;
         let num_threads = 4;
         let total_spans_to_emit = total_spans_per_thread * num_threads;
 
--- a/opentelemetry-sdk/src/trace/span_processor.rs
+++ b/opentelemetry-sdk/src/trace/span_processor.rs
@@ -583,17 +583,20 @@ impl SpanProcessor for BatchSpanProcessor {
 
     /// Handles span end.
     fn on_end(&self, span: SpanData) {
+        // Count the span before enqueueing it so that a concurrent
+        // force_flush()/shutdown() drain never observes an
+        // enqueued-but-uncounted span and misses it (issue #3453). If the
+        // send fails, the increment is reverted in the error arms below.
+        let previous_batch_size = self.current_batch_size.fetch_add(1, Ordering::AcqRel);
         let result = self.span_sender.try_send(span);
 
         // match for result and handle each separately
         match result {
             Ok(_) => {
                 // Successfully sent the span to the data channel.
-                // Increment the current batch size and check if it has reached
-                // the max export batch size.
-                if self.current_batch_size.fetch_add(1, Ordering::AcqRel) + 1
-                    >= self.max_export_batch_size
-                {
+                // Check if the current batch size has reached the max export
+                // batch size.
+                if previous_batch_size + 1 >= self.max_export_batch_size {
                     // Check if the a control message for exporting spans is
                     // already sent to the worker thread. If not, send a control
                     // message to export spans. `export_span_message_sent` is set
@@ -630,6 +633,8 @@ impl SpanProcessor for BatchSpanProcessor {
                 }
             }
             Err(std::sync::mpsc::TrySendError::Full(_)) => {
+                // The span never entered the channel; revert the increment.
+                self.current_batch_size.fetch_sub(1, Ordering::AcqRel);
                 // Increment dropped spans count. The first time we have to drop
                 // a span, emit a warning.
                 if self.dropped_spans_count.fetch_add(1, Ordering::Relaxed) == 0 {
@@ -638,6 +643,8 @@ impl SpanProcessor for BatchSpanProcessor {
                 }
             }
             Err(std::sync::mpsc::TrySendError::Disconnected(_)) => {
+                // The span never entered the channel; revert the increment.
+                self.current_batch_size.fetch_sub(1, Ordering::AcqRel);
                 // Given background thread is the only receiver, and it's
                 // disconnected, it indicates the thread is shutdown
                 otel_warn!(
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -1335,6 +1342,216 @@ mod tests {
...
+
+        let config = BatchConfigBuilder::default()
+            .with_max_queue_size(2048)
+            .with_max_export_batch_size(512)
+            .with_scheduled_delay(Duration::from_millis(5))
+            .build();
+
+        let processor = BatchSpanProcessor::new(exporter, config);
+
+        let total_spans_per_thread = 100_000;
```
lalitb: Can we reduce this to something like `10_000` per thread? That still gives 40k spans total, but avoids adding a 400k span stress test to the normal unit suite. The targeted tests above already cover the race more directly.

lalitb (approved):
LGTM with nit comment. Thanks.

Wassbdr:
Thanks @lalitb! Reduced `batchspanprocessor_all_spans_accounted_for` to 10k spans per thread (40k total) in ee817378 — still exercises the received+dropped==emitted invariant without the 400k-span cost. The targeted race tests cover the race directly.

cijothomas:
@Wassbdr can you fix the changelog conflict and this is good to merge.

Wassbdr:
Done — rebased on latest main and resolved the CHANGELOG conflict. Note the force-push cancelled auto-merge; could you re-enable it (or merge) once checks are green? Thanks!

PR #3562, opened 2026-06-23
chore: Release opentelemetry-semantic-conventions 0.32.1

cijothomas:
Releases `opentelemetry-semantic-conventions` 0.32.1 picking up the v1.42.0 spec update from PR #3561. The previous release tracked v1.34.0; cutting a patch release now (rather than 0.33.0) so contrib instrumentations can pick up the newer conventions without waiting for the next coordinated workspace release and without a minor-version bump.


lalitb (approved):

opentelemetry-semantic-conventions/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## vNext
 
+## 0.33.0
```
utpilla: Could we release `0.32.1` instead?

cijothomas: done

utpilla (approved):

Issue #3553, opened 2026-06-11
ci: weaver live-check gate for SDK self-observability events


cijothomas:
Wire `weaver registry live-check` into CI so every SDK-emitted self-observability event declared in an in-tree semconv registry is exercised end-to-end on every PR. Fail the job if a declared event receives zero samples or has event-level violations. SDK-side companion to [open-telemetry/semantic-conventions#3723](https://github.com/open-telemetry/semantic-conventions/pull/3723) (`otel.sdk.component.shutdown` is the first event).

## Reference implementations

- otel-arrow MVP: [#3047](https://github.com/open-telemetry/otel-arrow/pull/3047)
- otel-arrow refactor onto composite actions: [#3175](https://github.com/open-telemetry/otel-arrow/pull/3175)
- Weaver composite actions: [open-telemetry/weaver#1448](https://github.com/open-telemetry/weaver/pull/1448)

## Sketch

- In-tree registry (e.g. `semconv/`) with `manifest.yaml` pulling upstream semconv and one `groups/event.<name>.yaml` per declared event.
- A runnable that emits the events (dedicated example preferred over reusing `examples/basic-otlp`).
- Workflow using `weaver-live-check-{start,stop}` + a registry-driven assert step (zero-sample / violation gate, copied from otel-arrow). `fail-on: none` until weaver lands per-event coverage ([#1474](https://github.com/open-telemetry/weaver/issues/1474)) and severity-scoped exit codes ([#1473](https://github.com/open-telemetry/weaver/issues/1473)).
- Not required-status initially; promote later.

PR #3565, opened 2026-06-23
ci: add weaver live-check workflow for SDK self-observability metrics

cijothomas:
Towards #3553.

Adds a `weaver registry live-check` CI workflow that validates the SDK self-observability metric `otel.sdk.processor.log.processed` against the upstream semantic conventions registry when relevant SDK, OTLP exporter, CI app, or workflow files change.

## Changes

- Adds `.github/workflows/sdk-self-observability.yml` using the upstream `weaver-live-check-{start,stop}` composite actions (v0.24.0) and semconv v1.42.0, with `fail-on: violation`.
- Adds `.github/ci-apps/self-obs-live-check/`, a purpose-built CI binary that creates a batch log processor backed by an in-memory log exporter, emits log records, and exports the resulting SDK self-observability metric to Weaver over OTLP/gRPC.
- Adds the CI apps directory to the Cargo workspace.

## Scope

This is an initial step toward #3553. It validates one SDK self-observability metric directly against the published upstream registry; broader event coverage and zero-sample assertions remain follow-up work.

The workflow runs on pull requests, pushes to `main`, merge groups, and manual dispatch, but is not yet a required status check. It can be promoted after the gate has proven stable.


jensholdgaard:
As a downstream data point in favour of this: we run essentially this exact pattern in [Ourios](https://github.com/jensholdgaard/ourios) (a Parquet + DataFusion log-storage backend built on the OTel stack), and it's been solid enough that we promoted it to a **required** CI gate.

Our `live-check (weaver)` job boots the real `ourios-server`, drives a little traffic, and runs `weaver registry live-check --future` over the emitted self-observability metrics — the same shape as this PR (dedicated telemetry source → live-check against the registry, no separate assertion harness). Two differences from your setup, neither material to the approach:

- We validate against our own `semconv/registry/` (which *extends* the upstream registry with `ourios.*` metrics) rather than upstream directly — but it's the same `weaver registry live-check` mechanism on the same semconv v1.42.0 baseline.
- We emit from the production server binary + real traffic rather than a purpose-built example. Both work fine.

On the "not yet a required check / promote once proven stable" note — it graduates cleanly. It's in our `ci-success` required set and has been stable (no flakiness). The pairing that's worked well for us is the *static* `weaver registry check` (a no-diff gate on the generated constants) **plus** this *runtime* live-check — together they keep the declared registry and the actually-emitted telemetry honest with each other, which neither catches alone.

So the staged-rollout endpoint you're describing is reachable in practice. +1 to landing this — self-obs metrics silently drifting from semconv is exactly the failure it prevents, and it's nicer to catch in CI than in a consumer's dashboard. Happy to share our workflow as a reference if useful.

Issue #3374, opened 2026-02-20
Tracking: spec-required environment variables not yet supported


bryantbiggs:
## Summary

Cross-referencing the [SDK environment variable specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/) and [OTLP exporter specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/) against the current implementation reveals several spec-required environment variables that are not yet supported and have no PR in flight. This issue tracks them in one place.

The recently merged #3358 documents what **is** supported today. This issue covers the gaps.

## Environment variables with active PRs (for reference)

These are already being worked on and are **not** blocked:

- [x] `OTEL_SDK_DISABLED` — #1936 / PR #3088
- [x] `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` — PR #3090
- [x] `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` / `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL` / `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL` — PR #3363
- [x] `OTEL_EXPORTER_OTLP_*_CERTIFICATE` / `OTEL_EXPORTER_OTLP_*_CLIENT_CERTIFICATE` / `OTEL_EXPORTER_OTLP_*_CLIENT_KEY` — PR #3364
- [x] `OTEL_EXPORTER_OTLP_*_INSECURE` — PR #3365

### SDK — General / Attribute Limits

- [ ] `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` — global fallback for all signals; per-span variant is in PR #3090 but the global limit is not addressed
- [ ] `OTEL_ATTRIBUTE_COUNT_LIMIT` — global fallback for all signals; not implemented at all

### SDK — Propagators

- [ ] `OTEL_PROPAGATORS` — #3127 exists but has no PR; also related to #3285

### SDK — Log Record Limits

- [ ] `OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT` — no issue or PR
- [ ] `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` — no issue or PR

### SDK — Metrics / Exemplars

[...]

PR #3568, opened 2026-06-24
feat(sdk): support OTEL_ATTRIBUTE_COUNT_LIMIT env var

Wassbdr:
Part of #3374

## Problem

`OTEL_ATTRIBUTE_COUNT_LIMIT` is the spec-defined **general fallback** for the maximum span attribute count ([SDK configuration spec](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#attribute-limits)). The SDK only read the span-specific `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, so setting the general variable had no effect. This is one of the gaps tracked in #3374.

## Change

In `Config::default()`, read `OTEL_ATTRIBUTE_COUNT_LIMIT` and apply it to `max_attributes_per_span`, then let `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT` override it when both are set (per the spec's precedence). Code-based configuration (e.g. `with_max_attributes_per_span`) continues to take precedence over both, since the builder starts from `Config::default()`.

The general limit deliberately does **not** touch the per-event / per-link attribute count limits — those have their own spec defaults (128) and are not governed by `OTEL_ATTRIBUTE_COUNT_LIMIT`.

## Tests

- `otel_attribute_count_limit_sets_span_attribute_limit` — general var applies to span attributes and leaves per-event / per-link limits at their defaults.
- `span_attribute_count_limit_overrides_general_limit` — span-specific var wins when both are set.
- `default_span_attribute_limit_when_unset` — default of 128 when neither is set.

Also documented the variable and its precedence in the crate-level env-var table and added a CHANGELOG entry.

`cargo test -p opentelemetry_sdk --all-features --lib trace::config`, `cargo clippy --all-targets --all-features -- -Dwarnings`, and `cargo fmt --all -- --check` all pass.

### DIFF
--- a/opentelemetry-sdk/src/lib.rs
+++ b/opentelemetry-sdk/src/lib.rs
@@ -94,11 +94,17 @@
 //! | `OTEL_TRACES_SAMPLER` | Sampler to use. Valid values: `always_on`, `always_off`, `traceidratio`, `parentbased_always_on`, `parentbased_always_off`, `parentbased_traceidratio`. | `parentbased_always_on` |
 //! | `OTEL_TRACES_SAMPLER_ARG` | Argument for the sampler. Used when `OTEL_TRACES_SAMPLER` is `traceidratio` or `parentbased_traceidratio`. Must be a float between 0.0 and 1.0. | `1.0` |
 //!
+//! ### General Attribute Limits
+//!
+//! | Variable | Description | Default |
+//! |---|---|---|
+//! | `OTEL_ATTRIBUTE_COUNT_LIMIT` | Maximum number of attributes allowed on a span. Acts as the fallback for `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, which takes precedence when both are set. | `128` |
+//!
 //! ### Trace: Span Limits
 //!
 //! | Variable | Description | Default |
 //! |---|---|---|
-//! | `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT` | Maximum number of attributes allowed on a span. | `128` |
+//! | `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT` | Maximum number of attributes allowed on a span. Overrides `OTEL_ATTRIBUTE_COUNT_LIMIT` when set. | `128` |
 //! | `OTEL_SPAN_EVENT_COUNT_LIMIT` | Maximum number of events allowed on a span. | `128` |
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/lib.rs
```diff
@@ -94,11 +94,17 @@
 //! | `OTEL_TRACES_SAMPLER` | Sampler to use. Valid values: `always_on`, `always_off`, `traceidratio`, `parentbased_always_on`, `parentbased_always_off`, `parentbased_traceidratio`. | `parentbased_always_on` |
 //! | `OTEL_TRACES_SAMPLER_ARG` | Argument for the sampler. Used when `OTEL_TRACES_SAMPLER` is `traceidratio` or `parentbased_traceidratio`. Must be a float between 0.0 and 1.0. | `1.0` |
 //!
+//! ### General Attribute Limits
+//!
+//! | Variable | Description | Default |
+//! |---|---|---|
+//! | `OTEL_ATTRIBUTE_COUNT_LIMIT` | Maximum number of attributes allowed on a span. Acts as the fallback for `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, which takes precedence when both are set. | `128` |
```
cijothomas, line 101: OTEL_ATTRIBUTE_COUNT_LIMIT -> is this supposed to affect spans and logs too?

cijothomas: I think we should do this only after Logs also support this, else this may misled users to believing that setting this ENV variable will enforce limits, but in reality - Logs don't enforce it.

cijothomas (changes requested):
Thanks for your contribution. Please see https://github.com/open-telemetry/opentelemetry-rust/pull/3568/changes#r3471676694

This is better done after Logging fixes.

Wassbdr:
Good point — thanks. The logs SDK doesn't enforce any attribute limit today, so I've opened #3571 to add that first (`OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` + `with_max_attributes_per_log`, with `OTEL_ATTRIBUTE_COUNT_LIMIT` as fallback, enforced in `SdkLogger::emit`).

Once #3571 lands, I'll update this PR so `OTEL_ATTRIBUTE_COUNT_LIMIT` consistently acts as the general fallback for both spans and logs. Happy to keep this one open and rebase on top of #3571, or close it and fold the general-fallback wording into a follow-up — whichever you prefer.

PR #3570, opened 2026-06-25
feat(trace): Span::record_error: Emit `exception.stacktrace`

9999years:
## Changes

Rust's `std::error::Error`s expose a chain of sources with the `Error::source` method. Let's use those to emit an `exception.stacktrace` attribute in accordance with the OpenTelemetry Semantic Conventions.

See: <https://opentelemetry.io/docs/specs/semconv/registry/attributes/exception/#exception-stacktrace>

### DIFF
--- a/opentelemetry/src/trace/span.rs
+++ b/opentelemetry/src/trace/span.rs
@@ -71,7 +71,12 @@ pub trait Span {
     /// If this span is not being recorded then this method does nothing.
     fn record_error(&mut self, err: &dyn Error) {
         if self.is_recording() {
-            let attributes = vec![KeyValue::new("exception.message", err.to_string())];
+            let message = err.to_string();
+            let mut attributes = Vec::with_capacity(2);
+            if let Some(stacktrace) = format_error_sources(err, &message) {
+                attributes.push(KeyValue::new("exception.stacktrace", stacktrace));
+            }
+            attributes.push(KeyValue::new("exception.message", message));
             self.add_event("exception", attributes);
         }
     }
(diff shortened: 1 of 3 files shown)


opentelemetry/src/trace/span.rs
```diff
@@ -317,13 +322,139 @@ impl Status {
     }
 }
 
+/// Format [`Error::source`]s into an 'error source chain' for use in the `exception.stacktrace`
+/// field of an exception event.
+///
+/// See: <https://opentelemetry.io/docs/specs/semconv/registry/attributes/exception/#exception-stacktrace>
+///
+/// Error messages are output one per line, like this:
```
cijothomas, line 330: that can't be really considered stack chain..

9999years: Do you mean that it's not a stack trace because it doesn't have line numbers? I agree. Do you think we should try to extract a backtrace? I'm not sure if we could get one reliably from an arbitrary `Error` value, though.

9999years:
Not stale, waiting for review

9999years:
it's fine to take a while to review a pr but closing it because you didn't review it just feels so pointless?

9999years:
> please feel free to reopen this PR

btw i don't even have permissions to do this

cijothomas:
> it's fine to take a while to review a pr but closing it because you didn't review it just feels so pointless?

The current stale automation is a very simple bot and cannot distinguish whether a PR is waiting for the author or for maintainer review. A more advanced bot is being rolled out across OpenTelemetry that explicitly marks PRs as waiting for review, waiting for the author, and so on, but it has not reached this repository yet. Sorry that this PR was closed while it was waiting on us. Will reopen and request review now.

PR #3571, opened 2026-06-26
feat(sdk): enforce log record attribute count limit

Wassbdr:
Part of #3374. Prerequisite for #3568.

## Context

In #3568 I added `OTEL_ATTRIBUTE_COUNT_LIMIT` for spans. @cijothomas [pointed out](https://github.com/open-telemetry/opentelemetry-rust/pull/3568#discussion_r3471676694) that the general variable is meant to apply to logs too, and that the logs SDK does not enforce any attribute limit today — so exposing the general variable before logs support exists would mislead users. This PR adds that missing logs support first.

## Change

`SdkLoggerProvider` now enforces a maximum attribute count per log record (default `128`), dropping attributes added beyond the limit when the record is emitted. Configurable, in precedence order, via:

1. `LoggerProviderBuilder::with_max_attributes_per_log` (code-based)
2. `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT`
3. `OTEL_ATTRIBUTE_COUNT_LIMIT` (general fallback)
4. default `128`

Enforcement is done once in `SdkLogger::emit`, before the record is handed to processors, using a new `GrowableArray::truncate`. The limit is resolved at provider build time and stored on `LoggerProviderInner`.

This mirrors the span behavior (`max_attributes_per_span` + `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`) and keeps the hot path cheap: a single truncate per emitted record, a no-op when within the limit.

## Notes / scope

- Truncation drops the **last-added** attributes (keeping the first), matching the span limit semantics.
- This PR does **not** surface a `dropped_attributes_count` on `SdkLogRecord` or wire it through OTLP (the proto transform currently sends `0`). That's a separate, larger change and can follow; happy to do it if preferred.

## Tests

- `GrowableArray::truncate` unit test (inline region, overflow region, boundary, no-op, zero).
- Logger provider tests: default limit truncates; within-limit kept; builder value wins; `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` over `OTEL_ATTRIBUTE_COUNT_LIMIT`; general var as fallback.

`cargo test -p opentelemetry_sdk --all-features --lib`, `cargo clippy --all-targets --all-features -- -Dwarnings`, and `cargo fmt --all -- --check` all pass.

### DIFF
--- a/opentelemetry-sdk/src/logs/logger.rs
+++ b/opentelemetry-sdk/src/logs/logger.rs
@@ -27,7 +27,9 @@ impl opentelemetry::logs::Logger for SdkLogger {
     type LogRecord = SdkLogRecord;
 
     fn create_log_record(&self) -> Self::LogRecord {
-        SdkLogRecord::new()
+        // Stamp the record with the configured attribute limit so that
+        // `add_attribute` can enforce it eagerly as attributes are added.
+        SdkLogRecord::new_with_limit(self.provider.max_attributes_per_log())
     }
 
     /// Emit a `LogRecord`.
--- a/opentelemetry-sdk/src/logs/logger_provider.rs
+++ b/opentelemetry-sdk/src/logs/logger_provider.rs
@@ -1,3 +1,4 @@
+use super::record::DEFAULT_MAX_ATTRIBUTES_PER_LOG;
 use super::{BatchLogProcessor, LogProcessor, SdkLogger, SimpleLogProcessor};
 use crate::error::{OTelSdkError, OTelSdkResult};
 use crate::logs::LogExporter;
@@ -22,8 +23,16 @@ fn noop_logger_provider() -> &'static SdkLoggerProvider {
         inner: Arc::new(LoggerProviderInner {
             processors: Vec::new(),
             is_shutdown: AtomicBool::new(true),
+            max_attributes_per_log: DEFAULT_MAX_ATTRIBUTES_PER_LOG,
         }),
     })
+}
+
+#[inline]
+fn read_u32_env(key: &str) -> Option<u32> {
+    std::env::var(key)
+        .ok()
+        .and_then(|value| value.parse::<u32>().ok())
 }
 
 #[derive(Debug, Clone)]
@@ -82,6 +91,12 @@ impl SdkLoggerProvider {
         &self.inner.processors
     }
 
+    /// Maximum number of attributes retained per log record. Attributes beyond
+    /// this limit are dropped when the record is emitted.
+    pub(crate) fn max_attributes_per_log(&self) -> u32 {
+        self.inner.max_attributes_per_log
+    }
+
     /// Force flush all remaining logs in log processors and return results.
     pub fn force_flush(&self) -> OTelSdkResult {
         let result: Vec<_> = self
@@ -135,6 +150,7 @@ impl SdkLoggerProvider {
 struct LoggerProviderInner {
     processors: Vec<Box<dyn LogProcessor>>,
     is_shutdown: AtomicBool,
+    max_attributes_per_log: u32,
 }
 
 impl LoggerProviderInner {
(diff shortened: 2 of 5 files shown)


opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -1007,4 +1052,98 @@ mod tests {
...
+
+    fn emit_record_with_attributes(provider: &SdkLoggerProvider, count: usize) -> usize {
+        let logger = provider.logger("test");
+        let mut record = logger.create_log_record();
+        for i in 0..count {
+            record.add_attribute(Key::new(format!("key{i}")), AnyValue::Int(i as i64));
+        }
+        logger.emit(record);
+        provider.force_flush().unwrap();
+        count
```
cijothomas: nit: no need to return the count here.

opentelemetry-sdk/src/logs/logger.rs
```diff
@@ -51,6 +51,10 @@ impl opentelemetry::logs::Logger for SdkLogger {
             record.observed_timestamp = Some(now());
         }
 
+        // Enforce the configured attribute count limit before handing the
+        // record to processors, dropping the attributes added last.
+        record.truncate_attributes(provider.max_attributes_per_log() as usize);
```
cijothomas: if the original intend behind enforcing this limit is to protect the sdk from taking up too much memory when a mistaken code adds too much attributes, then we should prevent it from being added in the first place. Truncating afterwards won't help.
Could you try eagerly checking the limit in each add_attribute() call? That'd also make it relatively easy to track the dropped count.
Please do benchmark and see if we regress measurably.

opentelemetry-sdk/src/logs/logger_provider.rs
```diff
@@ -287,6 +306,20 @@ impl LoggerProviderBuilder {
         LoggerProviderBuilder { resource, ..self }
     }
 
+    /// Specify the maximum number of attributes retained per log record.
+    ///
+    /// Attributes added beyond this limit are dropped when the record is
+    /// emitted. When not set, this is read from the
+    /// `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` environment variable, which falls
+    /// back to `OTEL_ATTRIBUTE_COUNT_LIMIT`, and otherwise defaults to `128`.
```
cijothomas, line 312: nit: after this is merged, we should follow up immediately with your other PR so this variable will be consistently applied in spans too.

Wassbdr:
Thanks for the review @cijothomas — addressed all three:

**Enforce eagerly in `add_attribute` + track dropped count.** Reworked: enforcement now happens in `SdkLogRecord::add_attribute`, which rejects attributes once the limit is reached (the storage never grows past the limit) and increments a `dropped_attributes_count`, exposed via `SdkLogRecord::dropped_attributes_count()`. The limit is stamped onto the record at `create_log_record` time. Removed the emit-time truncation (and the now-unused `GrowableArray::truncate`).

**Benchmark.** Ran `cargo bench --bench log-attributes` (appender-tracing, 0–12 attributes) as an A/B of this branch vs `main` on the same machine. No regression on any case — criterion reports either "No change" or "improved". Representative:

| case | change vs main (p) |
|---|---|
| otel_0_attributes | No change (p=0.61) |
| otel_2_attributes | No change (p=0.15) |
| otel_5_attributes | improved (p=0.00) |
| otel_10_attributes | improved (p=0.00) |

(The machine is a noisy WSL2 box, so the "improved" deltas are noise; the point is nothing regressed. The added cost is a single `len() < max` comparison per attribute.)

**nit: helper return value / follow-up PR.** Removed the unused return from `emit_record_with_attributes`. And yes — once this lands I'll rebase #3568 so `OTEL_ATTRIBUTE_COUNT_LIMIT` is applied consistently to spans too.

One scope note: this tracks `dropped_attributes_count` on the record but does not yet wire it through the OTLP proto transform (still sends `0`). Happy to do that here or as a fast follow — let me know your preference.

PR #3582, opened 2026-07-03
fix(appender-tracing): always normalize event metadata when `experimental_metadata_attributes` is enabled

### LINKED ISSUE #3543: [Bug]: Appender-tracing creates log records with the wrong metadata under `experimental_span_attributes` feature
### What happened?

When using `opentelemetry-appender-tracing` in conjunction with `tracing-log` to generate log records, the exported log records carry the `target` of the original log record. The `tracing-log` crate [statically reports](https://docs.rs/tracing-log/latest/src/tracing_log/lib.rs.html#284) this (its `Metadata::target`) as `"log"` for all of its events. However, under the `experimental_span_attributes` feature flag it looks like this was supposed to be uplifted from `log.target` through [`NormalizeEvent`](https://docs.rs/tracing-log/latest/tracing_log/trait.NormalizeEvent.html) - yet it isn't, because the target is evaluated first. This leads to the instrumentation scope name also being wrong, as that is filled in from the `target`.

I've tried to reconstruct the relevant history to the best of my ability:
- In https://github.com/open-telemetry/opentelemetry-rust/pull/1380, normalization was added. Because the semantic conventions are still unstable, this happened under an experimental feature flag (cf. https://github.com/open-telemetry/opentelemetry-rust/pull/1380#discussion_r1397635566)
- https://github.com/open-telemetry/opentelemetry-rust/issues/1683 then asked for filling in the instrumentation scope from `target`. 
- This was implemented in https://github.com/open-telemetry/opentelemetry-rust/pull/1869, at the time respecting the feature flag because normalization happened before anything else if the feature was enabled.
- In https://github.com/open-telemetry/opentelemetry-rust/pull/2668, extraction of metadata severity and target is moved to the top of the event handler to allow dynamically checking if the event should be enabled, fixing https://github.com/open-telemetry/opentelemetry-rust/issues/1682.
  - This breaks the feature flag, since the two attributes are not recomputed after the metadata is normalized.
- To make this check more dynamic (https://github.com/open-telemetry/opentelemetry-rust/pull/2756), https://github.com/open-telemetry/opentelemetry-rust/pull/2756 hoists `name` up to the top as well.
  -  This completely decouples the fields set on `log_record` from the normalized metadata, as even the severity text (that is not pre-computed) now refers back to the original metadata. Only the visitor for extracting the `code.` now uses the normalized metadata.

Independently of whether the `code.` attributes are wanted, normalizing the event seems appropriate. While the `NormalizeEvent` trait seems to have been primarily introduced [to allow printing the correct name in `tracing-fmt`](https://github.com/tokio-rs/tracing/issues/185), the target / scope of the original `log::log!` (or `info!` or `error!` etc.) invocation is by definition more correct than just `log`, as it reports the actual location of the log event in the originating crate. I'm not sure if it needs to be tied to the same feature flag as the experimental attributes, which are uplifted through the dedicated visitor, but at the very least when the event metadata is normalized I'd expect the resulting target to be post-norm. 

0.32.0

0.32.0

OTLP

### Relevant log output

```shell

```

### Tip

dquirl-R1:
Fixes #3543
~~Design discussion issue (if applicable) #~~

## Changes

Consistently applies `NormalizeEvent` if the `experimental_metadata_attributes` feature is active to ensure that the generated log records that are emitted forward the original `log.target` as the log event target instead of defaulting to `"log"`. 

The change impacts all extracted metadata, not just the target. It also applies to 
- the `severity`, `target` and `name` used to check if `self.logger.event_enabled` (cf. #2668, #2756)
- the `severity` number set on the log record itself
- the severity text set on the log record itself, which is derived from the event metadata `level`

It does **not** affect the behaviour of `EnvFilter`, because of the interaction between `tracing-log` normalizing through `AsTrace` (a _different_ normalization) before calling `enabled` on the dispatcher in `dispatch_record` and the way `EnvFilter` caches `FilterState` and doesn't use `event_enabled` (see also https://github.com/open-telemetry/opentelemetry-rust/issues/3543#issuecomment-4659241874).

Note that this reverts #2799 as the optimization is incompatible with applying normalization before setting the target.

### DIFF
--- a/opentelemetry-appender-tracing/CHANGELOG.md
+++ b/opentelemetry-appender-tracing/CHANGELOG.md
@@ -6,6 +6,13 @@
   `OpenTelemetryTracingBridge::builder_with_scope_attributes(..)`.
   [3415](https://github.com/open-telemetry/opentelemetry-rust/issues/3415)
 
+- Fixes [#3543](https://github.com/open-telemetry/opentelemetry-rust/issues/3543):
+  Event metadata is now always normalized when `experimental_metadata_attributes` 
+  is enabled. In particular, when using the feature the underlying OTEL logger
+  will receive the value of `log.target` as its log event `target` for events 
+  generated from the `log` crate (e.g., via `tracing-log`) instead of the static
+  target value `"log"`.
+
 ## 0.32.0
 
 Released 2026-May-08
--- a/opentelemetry-appender-tracing/src/layer.rs
+++ b/opentelemetry-appender-tracing/src/layer.rs
@@ -477,7 +477,14 @@ where
         #[cfg_attr(not(feature = "experimental_span_attributes"), allow(unused_variables))]
         ctx: tracing_subscriber::layer::Context<'_, S>,
     ) {
+        #[cfg(feature = "experimental_metadata_attributes")]
+        let normalized_meta = event.normalized_metadata();
+        #[cfg(feature = "experimental_metadata_attributes")]
+        let metadata = normalized_meta.as_ref().unwrap_or_else(|| event.metadata());
+
+        #[cfg(not(feature = "experimental_metadata_attributes"))]
         let metadata = event.metadata();
+
         let severity = severity_of_level(metadata.level());
         let target = metadata.target();
         let name = metadata.name();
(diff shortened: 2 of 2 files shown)


dquirl-R1:
Sorry, this is our first contribution to a project that requires a CLA, so it took a while to jump through all the hoops. I wasn't sure if I was actually able to reopen the PR, so I created a new PR at #3639 that is also rebased on current `main`.

PR #3585, opened 2026-07-07
feat: add context event observer API

yannham:
## Motivation

[OTEP: Thread Context: Sharing Thread-Level Information with external readers](https://github.com/open-telemetry/opentelemetry-specification/pull/4947) is on the verge of being merged as an official OTEP. This proposal defines a protocol for an instrumented application to publish thread-level context somewhere that is discoverable by an external, potentially out-of-process reader (typically the eBPF profiler).

In order to properly implement this, any telemetry crate building upon opentelemetry-rust needs to be notified about context switches. This PR adds such a capability/API in a very simple way, currently gated behind an experimental feature.

## Changes

Add an API for downstream crates to subscribe to context switch events through the `ContextObserver` trait. The initial implementation is minimal and has only two events, `on_context_enter` and `on_context_exit`. We currently only allow one global observer.

Additionally, for any implementation of the OTEP mentioned above to be remotely performant, we'll also need to store alongside opentelemetry-rust's `Context` the `ThreadContextRecord` view of this context defined in the OTEP, which bears similar information but in a different shape. This PR adds a generic free-form field to `Context`, the observer's context view, which represents an immutable, set-once view of the associated context for the observer.

The view is behind a `OnceLock` so that:

- there's an obvious `Default` implementation
- the lifecycle design questions are entirely offloaded to the observer. Whether to initialize the view immediately or at first `on_context_enter`, etc.

An alternative would be to have an `Option<_>`, but that would need to be initialized at context creation, which would need an additional context event (e.g. `on_context_created` or `make_view`). Another possibility would be to require `Default` on `ObserverContextView`, but that would make it not object-safe which is out of the question.

All in all the lock is simple, minimize changes to this repo, doesn't prevent future optimization (having more events for optimized view creations, like when deriving a context from an existing one). To only price is the lock's additional size, (I suppose at most a usize).

### DIFF
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -19,12 +19,17 @@ use std::fmt;
 use std::hash::{BuildHasherDefault, Hasher};
 use std::marker::PhantomData;
 use std::sync::Arc;
+#[cfg(feature = "experimental_context_observer")]
+use std::sync::OnceLock;
 
 #[cfg(feature = "futures")]
 mod future_ext;
 
 #[cfg(feature = "futures")]
 pub use future_ext::{FutureExt, WithContext};
+
+#[cfg(feature = "experimental_context_observer")]
+pub use context_observer::*;
 
 thread_local! {
     static CURRENT_CONTEXT: RefCell<ContextStack> = RefCell::new(ContextStack::default());
@@ -98,7 +103,25 @@ pub struct Context {
     pub(crate) span: Option<Arc<SynchronizedSpan>>,
     entries: Option<Arc<EntryMap>>,
     suppress_telemetry: bool,
+    /// A context's observer view of this [Context].
+    ///
+    /// The main application for observers is to share information from the current context with
+    /// external readers (e.g. a full host eBPF profiler). In that case, the observer will attach
+    /// and detach some data on context events using its own mechanism. The shared data has no
+    /// reason to match our internal representation ([Context]).
+    ///
+    /// [observer_state] makes it possible to store the observer's view of the context alongside a
+    /// [Context] value, such that it can be attached and detached on context events.
+    #[cfg(feature = "experimental_context_observer")]
+    observer_cx_view: OnceLock<Arc<dyn ObserverContextView>>,
 }
+
+/// A context's observer view of a [Context]. This is basically an arbitrary data structure.
+///
+/// The `Any` supertrait lets an observer recover the concrete type it stored in the
+/// `observer_cx_view` slot (via trait upcasting to `&dyn Any` and `downcast_ref`).
+#[cfg(feature = "experimental_context_observer")]
+pub trait ObserverContextView: Any + Send + Sync {}
 
 type EntryMap = HashMap<TypeId, Arc<dyn Any + Sync + Send>, BuildHasherDefault<IdHasher>>;
 
@@ -264,6 +287,8 @@ impl Context {
             #[cfg(feature = "trace")]
             span: self.span.clone(),
             suppress_telemetry: self.suppress_telemetry,
+            #[cfg(feature = "experimental_context_observer")]
+            observer_cx_view: OnceLock::new(),
         }
     }
 
@@ -363,6 +388,8 @@ impl Context {
             #[cfg(feature = "trace")]
             span: self.span.clone(),
             suppress_telemetry: true,
+            #[cfg(feature = "experimental_context_observer")]
+            observer_cx_view: OnceLock::new(),
         }
     }
 
@@ -426,12 +453,21 @@ impl Context {
         Self::map_current(|cx| cx.is_telemetry_suppressed())
     }
 
+    /// Returns the observer's view of this context, if any.
+    #[cfg(feature = "experimental_context_observer")]
+    #[inline]
+    pub fn observer_cx_view(&self) -> &OnceLock<Arc<dyn ObserverContextView>> {
+        &self.observer_cx_view
+    }
+
     #[cfg(feature = "trace")]
     pub(crate) fn current_with_synchronized_span(value: SynchronizedSpan) -> Self {
         Self::map_current(|cx| Context {
             span: Some(Arc::new(value)),
             entries: cx.entries.clone(),
             suppress_telemetry: cx.suppress_telemetry,
+            #[cfg(feature = "experimental_context_observer")]
+            observer_cx_view: OnceLock::new(),
         })
     }
 
@@ -441,6 +477,8 @@ impl Context {
             span: Some(Arc::new(value)),
             entries: self.entries.clone(),
             suppress_telemetry: self.suppress_telemetry,
+            #[cfg(feature = "experimental_context_observer")]
+            observer_cx_view: OnceLock::new(),
         }
     }
 }
@@ -557,6 +595,11 @@ impl ContextStack {
         // top of the [`ContextStack`] as the `current_cx`.
         let next_id = self.stack.len() + 1;
         if next_id < ContextStack::MAX_POS.into() {
+            #[cfg(feature = "experimental_context_observer")]
+            if let Some(observer) = GlobalContextObserver::get() {
+                observer.on_context_enter(&self.current_cx, &cx);
+            }
+
             let current_cx = std::mem::replace(&mut self.current_cx, cx);
             self.stack.push(Some(current_cx));
             next_id as u16
@@ -602,6 +645,11 @@ impl ContextStack {
             // empty context is always at the bottom of the stack if the
             // [`ContextStack`] is not empty.
             if let Some(Some(next_cx)) = self.stack.pop() {
+                #[cfg(feature = "experimental_context_observer")]
+                if let Some(observer) = GlobalContextObserver::get() {
+                    observer.on_context_exit(&self.current_cx, &next_cx);
+                }
+
                 // Extract and return only the span to avoid cloning the entire Context
                 #[cfg(feature = "trace")]
                 {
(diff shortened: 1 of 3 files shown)


opentelemetry/src/context.rs
```diff
@@ -98,8 +103,23 @@ pub struct Context {
...
+    ///
+    /// The main application for observers is to share information from the current context with
+    /// external readers (e.g. a full host eBPF profiler). In that case, the observer will attach
+    /// and detach some data on context events using its own mechanism. The shared data has no
+    /// reason to match our internal representation ([Context]).
+    ///
+    /// [observer_state] makes it possible to store the observer's view of the context alongside a
+    /// [Context] value, such that it can be attached and detached on context events.
+    #[cfg(feature = "experimental_context_observer")]
+    observer_cx_view: OnceLock<Arc<dyn ObserverContextView>>,
```
scottgerring: This s a clever way of doing this! I was imagining something with lazily-invoked callbacks but this feels nicer.

opentelemetry/src/context.rs
scottgerring, line 365: We take a mutable borrow on the context here ....

opentelemetry/src/context.rs
```diff
@@ -557,6 +592,11 @@ impl ContextStack {
         // top of the [`ContextStack`] as the `current_cx`.
         let next_id = self.stack.len() + 1;
         if next_id < ContextStack::MAX_POS.into() {
+            #[cfg(feature = "experimental_context_observer")]
+            if let Some(observer) = GlobalContextObserver::get() {
+                observer.on_context_enter(&self.current_cx, &cx);
```
scottgerring, line 617: ... and invoke the observer here. I think this means that Bad Things will happen if the observer does things like calling `Context::current()`, tries to attach another context, or emits telemetry probably even - as this'd lead to trying to take another mutable borrow on the RefCell? 

I think a typical observer shouldnt need to do these things but perhaps we should make that explicit in the API docs?

opentelemetry/src/context.rs
scottgerring: Because we have quite a low MSRV (1.75), I suspect that users can't rely on trait upcasting to do things with their view. Because its methodless, they'll be able to store a view but not recover its concrete type. I think we should be able to add a helper to facilitate this, and probably we should add a test to validate and show the general usage shape too

scottgerring (changes requested):

opentelemetry/src/context.rs
yannham: Ah, very good point, that would panic :thinking: we should definitively warn about it. I wonder if there's a way to prevent that but this is going to be difficult, since `Context::current()` is basically a freely accessible global, I don't see an easy and ergonomic way to prevent `on_context_enter` specifically to access it (without making the life of everyone else harder).

I wonder if the simplest might not be to move the call to `on_context_enter` out of the stack borrow entirely. One possibility is after the push: we can look at the return value of `push` and if it's not `MAX`, then we did entered the context. But since it's owned by the stack,we will need to borrow the stack to provide a reference to the context still, even if only immutably. Which sounds back to square one.

If we do it _before_, it should work though. We can check that there's space in the stack, then call `on_context_enter`, and then do the push. Wdyt? The cost is one additional load/store to the `RefCell`'s flag I suppose, but it's touched anyway by the `borrow_mut` so should be in cache

yannham: (since it's thread-local and there's no .await point nothing else can borrow in the meantime I assume)

scottgerring: That sounds reasonable - want to try? The cost sounds reasonable to me especially considering with expect good cache locality; it would also be trivial to regress with the benchmark suite to quantify that in ns overhead.

Alternatively, how gross is just documenting it? I mean - if we say its forbidden, and someone writes code that does this, they will find out _very quickly_ that it is verboten - its not going to hide some heisenbug. 

It is probably illustrative to think about whether or not someone might reasonably want to _do something_ that either produces telemetry or needs to poke the context directly in a hook. On the telemetry front I could imagine logging, at least in debug scenarios, might come up

yannham: > Alternatively, how gross is just documenting it? I mean - if we say its forbidden, and someone writes code that does this, they will find out very quickly that it is verboten - its not going to hide some heisenbug.

Yes, at least it panics and isn't just silently wrong/corrupted/UB. Still, I'm not sure how easy it is to relate the error (that will say multiple borrows at some random context _usage_ site, unrelated to the observer) to the the observer.

> It is probably illustrative to think about whether or not someone might reasonably want to do something that either produces telemetry or needs to poke the context directly in a hook. On the telemetry front I could imagine logging, at least in debug scenarios, might come up

I agree this is the important question to answer. Technically, both directions are sound/acceptable but maybe taking multiple borrows just shouldn't happen in an observer? At least that was my initial assumption, but as you mention logging might actually be use-case. Though I expect observers to be really "low-level" and maybe not log through the same pipeline as the instrumented application.

Since the observer implementation is almost part of opentelemetry-rust (to me, it is to be thought of as an extension mechanism, and maybe at some point the thread-level context publication code will actually be moved from downstream libraries to here), I still feel like it _should not_ affect the contexts. I'll update the documentation for now and keep the code identical. If someone comes to complain about a legitimate use-case through an issue, we can always switch to the non-locking implementation.

scottgerring:
@yannham give me a shout when you are ready for us to look again!

yannham:
@scottgerring it is ready for another look :slightly_smiling_face:

opentelemetry/src/context.rs
```diff
@@ -98,6 +103,17 @@ pub struct Context {
...
     entries: Option<Arc<EntryMap>>,
     suppress_telemetry: bool,
+    /// A context's observer view of this [Context].
+    ///
+    /// The main application for observers is to share information from the current context with
+    /// external readers (e.g. a full host eBPF profiler). In that case, the observer will attach
+    /// and detach some data on context events using its own mechanism. The shared data has no
+    /// reason to match our internal representation ([Context]).
+    ///
+    /// [observer_state] makes it possible to store the observer's view of the context alongside a
```
scottgerring: I think this should be: 

```suggestion
    /// [observer_view] makes it possible to store the observer's view of the context alongside a
```

scottgerring (commented):
LGTM! One doc nit. 
I note that the OTEP for TLS context has landed, so it would be lovely to get this in (and then perhaps even a OTel-native context observer on top? :D) 

https://github.com/open-telemetry/opentelemetry-specification/pull/4947

scottgerring (approved):

yannham:
It seems the CI failure is unrelated and affects a pre-existing example (`basic-otlp`)?

opentelemetry/src/context.rs
```diff
@@ -98,6 +103,17 @@ pub struct Context {
...
+    ///
+    /// The main application for observers is to share information from the current context with
+    /// external readers (e.g. a full host eBPF profiler). In that case, the observer will attach
+    /// and detach some data on context events using its own mechanism. The shared data has no
+    /// reason to match our internal representation ([Context]).
+    ///
+    /// [observer_view] makes it possible to store the observer's view of the context alongside a
+    /// [Context] value, such that it can be attached and detached on context events.
+    #[cfg(feature = "experimental_context_observer")]
+    observer_view: OnceLock<Arc<dyn ObserverContextView>>,
```
lalitb: I think the view cell may need to be shared across `Context` clones. Right now, if a `Context` is cloned before the view is initialized, each clone gets its own empty `OnceLock`. That seems risky for async use, because `WithContext` clones the stored context on every poll, so the observer may create a new view each time instead of reusing one for the same logical context. Would it make sense to store this as something like `Arc<OnceLock<...>>`?

lalitb:
Overall I’m fine with this being added behind the experimental feature flag. Since it is experimental, we still have room to change or remove it later if the spec/API shape changes.

That said, this does add new public surface area, so I’d like to hear what other @open-telemetry/rust-maintainers think before we merge.

opentelemetry/src/context.rs
yannham: This is a good point, and something I actually considered. The current solution does incur a fixed size overhead per context (from `OnceLock`, I assume it's something like an atomic `usize` for the state). The inner view is still shared, but you do use additional space for the lock on each context copy.

On the other hand, it has one advantage: there's no heap allocation as long as the view isn't initialized. It's closer to `Option<Arc<_>>` in that respect, though of course `Option<Arc<_>>` doesn't even have a size overhead .

I suppose both approaches are reasonable, and it depends on whether we expect contexts to never have a view during their lifetime (does that really happen in practice? A context that is created but never activated?). There's the third option of `Option` (no pun intended) as I mentioned in the description, but at the cost of adding more code and requiring the view initialization to be eager.

Happy to get more opinions/directions on this. I don't have a strong opinion myself.

scottgerring:
@lalitb @cijothomas if no-one else has strong opinions against this, are we happy to merge it in behind the feature flag? AFAIK we're happy with the review and have just left open a window for any dissent!

opentelemetry/src/context.rs
bantonsson: The observer is handed all the contexts it should ever need. I think that clearly stating that this is forbidden in the API is good enough.

opentelemetry/src/context.rs
bantonsson: Thinking back and forth on this, I think that the concern raised by @lalitb is a very common case and especially if one uses https://github.com/tokio-rs/tracing-opentelemetry to do the integration. Every async function annotated with `#[tracing::instrument]` would recreate the view on every poll.

Just thinking out loud here. Maybe an API with an `Option<Arc<_>>` and an `onCreate` for the observer that is not called on pure `clone` would be a better option.

yannham: > Just thinking out loud here. Maybe an API with an Option<Arc<_>> and an onCreate for the observer that is not called on pure clone would be a better option.

It's also a reasonable option, yes. If the observer still wants to be lazy, they can put the `OnceLock<_>` layer in their view type. It costs one more check in this case, but it's cheaper for observers that don't need the lazyness, who don't have to pay the `OnceLock` price.

yannham: And I think I totally missed your original point @lalitb, which was not that you pay a bit of space for an empty `OnceLock`, but that you lose sharing if you clone _before_ the view is initialized. You're entirely right and we should either go with `Option<Arc<_>>` and the `on_create` API or use `Arc<OnceLock<_>>`.

yannham:
I've implemented @bantonsson 's [suggestion](https://github.com/open-telemetry/opentelemetry-rust/pull/3585#discussion_r3759877108) in the latest version. The added complexity is rather small, and it solves both the sharing issue and the once lock's overhead at once. If an implementation really needs lazyness, it can always include a `OnceLock` or any kind of interior mutability within the `ObserverContextView` object.

scottgerring:
Cool! I think we should be happy to put this in particularly because of the experimental flag.
@lalitb wdyt?

bantonsson (approved):

PR #3586, opened 2026-07-07
fix(resource): percent-decode attribute values

### LINKED ISSUE #857: Decode values from `OTEL_RESOURCE_ATTRIBUTES`
While using the Go SDK I noticed an inconsistency in how `OTEL_RESOURCE_ATTRIBUTES` baggage values are encoded and decoded. This was due to a difference in the version of the W3C Baggage spec used for [baggage propagation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#propagation) and the [`OTEL_RESOURCE_ATTRIBUTES` decoding](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable).

https://github.com/open-telemetry/opentelemetry-specification/pull/2670 updates the W3C Baggage spec version used by the Resource spec to require percent-decoding values from  `OTEL_RESOURCE_ATTRIBUTES`.

9999years:
Fixes #857

## Changes

`$OTEL_RESOURCE_ATTRIBUTES` values are now percent-decoded in line with the spec.

### DIFF
--- a/opentelemetry-sdk/src/resource/env.rs
+++ b/opentelemetry-sdk/src/resource/env.rs
@@ -3,7 +3,8 @@
 //! Implementation of `ResourceDetector` to extract a `Resource` from environment
 //! variables.
 use crate::resource::{Resource, ResourceDetector};
-use opentelemetry::{Key, KeyValue, Value};
+use opentelemetry::{otel_warn, Key, KeyValue, Value};
+use percent_encoding::percent_decode_str;
 use std::env;
 
 const OTEL_RESOURCE_ATTRIBUTES: &str = "OTEL_RESOURCE_ATTRIBUTES";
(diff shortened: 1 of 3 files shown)


opentelemetry-sdk/src/resource/env.rs
```diff
@@ -50,9 +51,13 @@ fn construct_otel_resources(s: String) -> Resource {
                 None => return None,
             };
             let key = parts.0.trim();
```
lalitb: I believe we should also percent-decode `keys`. As per [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable):

``
The OTEL_RESOURCE_ATTRIBUTES environment variable will contain of a list of key value pairs, represented as key1=value1,key2=value2. All attribute values MUST be considered strings. The , and = characters in keys and values MUST be percent encoded. Other characters MAY be [percent-encoded](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1), e.g. values outside the ANSI characters set.
``

opentelemetry-sdk/src/resource/env.rs
```diff
@@ -50,9 +51,13 @@ fn construct_otel_resources(s: String) -> Resource {
...
-            let value = parts.1.trim();
+            // Trim OWS *before* percent-decoding so that any encoded whitespace
+            // (e.g. `%20`) in the value is preserved. Per the Resource SDK spec
+            // the value MUST be percent-decoded; keys are baggage tokens and are
+            // not encoded.
+            let value = percent_decode_str(parts.1.trim()).decode_utf8_lossy();
```
lalitb: `decode_utf8_lossy()` accepts invalid encoded UTF-8 by inserting replacement characters. The Resource SDK spec says decoding failures should discard the entire env var and report an error. The existing baggage propagator already uses `decode_utf8()` and logs invalid UTF-8 instead of doing lossy decoding, so this new path should probably follow that shape.

9999years:
Done, PTAL

PR #1753, opened 2024-05-13
Enable integration test for Span validation

lalitb:
- Add custom serialization for i64 as [int64, fixed64, uint64 should be serialized as string in json](https://protobuf.dev/programming-guides/proto3/#json)
- Custom serialization for u64 for timestamp was added in #1666 
- Enable integration test for Span validtion.

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [x] Changes in public API reviewed (if applicable)

PR #3595, opened 2026-07-15
fix(proto): accept empty AnyValue in OTLP/JSON

bruno-espino:
## Changes

The [`AnyValue` protobuf definition](https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/common/v1/common.proto#L25-L50) explicitly permits its `oneof` to be unset and defines that state as empty. The [OTLP/JSON encoding rules](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding) also require receivers to ignore unknown fields and unmarshal the message as if they were absent.

The custom `AnyValue` visitor introduced in [#1753](https://github.com/open-telemetry/opentelemetry-rust/pull/1753) instead returns an error when it sees no recognized value field. This rejects both `{}` and unknown-only objects and can reject the entire OTLP request.

This occurs in real exporter traffic: [Cloudflare Workers native OpenTelemetry export](https://developers.cloudflare.com/workers/observability/exporting-opentelemetry-data/) encodes a JavaScript log field containing `null` as an empty OTLP `AnyValue` (`{"key":"value","value":{}}`). A receiver using this decoder currently rejects the whole batch.

Deserialize those objects as the generated empty representation (`AnyValue { value: None }`). This also handles newer value fields safely for receivers that have not learned them yet. Regression coverage includes direct values, span attributes, and nested log fields.

* [x] Appropriate `CHANGELOG.md` updated
* [x] No public API changes

### DIFF
--- a/opentelemetry-proto/src/proto.rs
+++ b/opentelemetry-proto/src/proto.rs
@@ -151,13 +151,10 @@ pub(crate) mod serializers {
                     }
                 }
 
-                if let Some(v) = value {
-                    Ok(Some(v))
-                } else {
-                    Err(de::Error::custom(
-                        "Invalid data for Value, no known keys found",
-                    ))
-                }
+                // An AnyValue with no selected field is valid and considered empty.
+                // Unknown fields are ignored according to the OTLP/JSON specification,
+                // so an empty or unknown-only object has the same representation.
+                Ok(value)
             }
         }
 
--- a/opentelemetry-proto/CHANGELOG.md
+++ b/opentelemetry-proto/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- **Bug fix**: Accept empty `AnyValue` objects in OTLP/JSON payloads instead of rejecting the entire request.
+
 ## 0.32.0
 
 Released 2026-May-08
--- a/opentelemetry-proto/tests/json_serde.rs
+++ b/opentelemetry-proto/tests/json_serde.rs
@@ -637,6 +637,78 @@ mod json_serde {
         }
     }
 
+    mod empty_any_value {
+        use super::*;
+
+        #[test]
+        fn deserialize_empty_and_unknown_only_values() {
+            for json in [r#"{}"#, r#"{"futureValue":"ignored"}"#] {
+                let actual: AnyValue =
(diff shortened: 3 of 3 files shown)


bruno-espino:
@lalitb @TommyCpp this adjusts the custom `AnyValue` visitor introduced in #1753. Since you authored and approved that change respectively, a review would be appreciated if you have time.

opentelemetry-proto/src/proto.rs
```diff
@@ -151,13 +151,10 @@ pub(crate) mod serializers {
...
-                    Ok(Some(v))
-                } else {
-                    Err(de::Error::custom(
-                        "Invalid data for Value, no known keys found",
-                    ))
-                }
+                // An AnyValue with no selected field is valid and considered empty.
+                // Unknown fields are ignored according to the OTLP/JSON specification,
+                // so an empty or unknown-only object has the same representation.
+                Ok(value)
```
lalitb, line 157: I don’t think we should treat unknown-only objects the same as `{}` here. ProtoJSON rejects unknown fields by default unless an ignore-unknown option is enabled, so an object like `{"futureValue": ...}` should still fail.

Can we make this return `Ok(None)` only for a truly empty object, and keep the existing error when keys were present but none were known?


CORRECTION -  I was looking at the default ProtoJSON behavior, but OTLP/JSON has a specific deviation here: receivers MUST ignore unknown fields and unmarshal as if they were absent. Given that, treating an unknown-only object the same as `{}` looks valid for this OTLP/JSON path. Sorry for the noise.

lalitb (approved):

jensholdgaard:
Nice fix — and thanks for the thorough real-world motivation. Reducing the tail to `Ok(value)` is the right call: the OTLP/JSON rule to ignore unknown fields means an unknown-only object and an empty `{}` should both decode to the unset `oneof`, which this does correctly.

Small heads-up on parallel work: I hit the same bug independently and filed #3598 (the serializer's own `{}` output not round-tripping) plus a narrower PR #3599 that only accepts the empty case. Yours is the better fix — it also handles the unknown-only case, which mine rejects — so I'll close #3599 in favour of this. You might consider adding `closes #3598` so the canonical issue gets tied off when this merges.

PR #3597, opened 2026-07-17
chore: update Jaeger to v2

thompson-tomo:
Fixes #

## Changes

The jaegar tracing image has been switched as existing image is eol jaegertracing/jaeger#6321

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -144,7 +144,7 @@
 //! Jaeger natively supports the OTLP protocol, making it easy to send traces directly:
 //!
 //! ```shell
-//! $ docker run -p 16686:16686 -p 4317:4317 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:latest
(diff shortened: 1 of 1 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -144,7 +144,7 @@
 //! Jaeger natively supports the OTLP protocol, making it easy to send traces directly:
 //!
 //! ```shell
-//! $ docker run -p 16686:16686 -p 4317:4317 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:latest
+//! $ docker run -p 16686:16686 -p 4317:4317 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/jaeger:latest
```
lalitb: Since this now uses Jaeger v2, can we drop `-e COLLECTOR_OTLP_ENABLED=true`? Jaeger v2 all-in-one already exposes OTLP on 4317 by default, and v2 config docs say it no longer recognizes Jaeger v1-style env vars the same way. Keeping this flag makes the example look like the old v1 setup is still required.

thompson-tomo: Done

lalitb (approved):

PR #3599, opened 2026-07-19
fix(proto): deserialize `{}` as an empty AnyValue in with-serde

### LINKED ISSUE #3598: opentelemetry-proto with-serde: `{}` (unset AnyValue, valid proto3 JSON) fails to deserialize — serializer's own output doesn't round-trip
### What happened?

`opentelemetry-proto`'s `with-serde` deserializer rejects `{}` as an `AnyValue` — but `{}` is proto3 JSON's encoding of a message with no fields set, which for `AnyValue` is the unset `oneof` that the proto itself documents as valid: *"It is valid for all values to be unspecified in which case this AnyValue is considered to be 'empty'."*

Two concrete consequences:

1. **Real wire data fails to parse.** The OpenTelemetry Collector's file exporter emits `"body": {}` for body-less log records — e.g. the OTel event records the current `opentelemetry-demo` (`main`) produces (`shipping.feature_flag.evaluated`: `eventName` + attributes, body unset). Any consumer parsing that OTLP/JSON through `with-serde` fails at the first such record. Minimal reproducer:

```rust
use opentelemetry_proto::tonic::common::v1::AnyValue;
// fails: "Invalid data for Value, no known keys found"
let v: AnyValue = serde_json::from_str("{}").unwrap();
```

2. **The serializer's own output doesn't round-trip.** `AnyValue { value: None }` serializes to `{}` (the `serde(flatten)` of an unset oneof emits no fields), which the deserializer then rejects:

```rust
let empty = AnyValue { value: None };
let json = serde_json::to_string(&empty).unwrap();   // "{}"
let back: AnyValue = serde_json::from_str(&json).unwrap();  // fails
```

The cause is `deserialize_from_value`'s `ValueVisitor` in `opentelemetry-proto/src/proto.rs`: after draining the map it returns an error whenever no known key produced a value — conflating "object with only unknown keys" (arguably an error) with "object with **no** keys" (the valid unset encoding).

We hit this in production-adjacent code (an OTLP/JSON ingest path) against a real Collector capture; evidence trail with the captured record: https://github.com/jensholdgaard/ourios/issues/549. I have a minimal fix (track whether any key was seen; empty object → `Ok(None)`; unknown-keys-only objects still error) with a regression test — PR follows.

0.32.0 (bug confirmed present at current `main`)

n/a (opentelemetry-proto 0.32.0, `with-serde` feature)

N/A

### Relevant log output

```
Invalid data for Value, no known keys found
```

jensholdgaard:
Fixes #3598

## Changes

`with-serde`'s `ValueVisitor` (`deserialize_from_value`) rejected **every** `AnyValue` object that produced no known-key value — conflating two cases:

- an object with **only unknown keys** → still an error (unchanged), and
- an object with **no keys at all** (`{}`) → proto3 JSON's encoding of a message with no fields set, which for `AnyValue` is the unset `oneof` the proto itself documents as valid (*"It is valid for all values to be unspecified in which case this AnyValue is considered to be 'empty'"*).

The empty object is not hypothetical wire data: the OpenTelemetry Collector's file exporter emits `"body": {}` for body-less log records (e.g. the event records the current `opentelemetry-demo` produces), and it is also **exactly what this crate's own serializer emits** for `AnyValue { value: None }` (the `serde(flatten)` of an unset oneof serializes no fields) — so the serializer's output previously failed to round-trip through its own deserializer.

The fix tracks whether the visitor saw any key: a genuinely empty object deserializes to the unset value (`Ok(None)`); unknown-keys-only objects keep the existing `"Invalid data for Value, no known keys found"` error. New regression test (`json_serde::value::empty`) covers serialize + deserialize of the empty `AnyValue`; it fails without the `proto.rs` change.

Found via a real Collector capture in a downstream OTLP/JSON ingest path; evidence trail: https://github.com/jensholdgaard/ourios/issues/549.

**Disclosure**: this fix was developed with AI assistance (Claude, driving a coding session under my direction); I've reviewed and verified the change, the mutation check (test fails without the patch), and the test run (`cargo test -p opentelemetry-proto --features gen-tonic-messages,logs,trace,metrics,zpages,with-serde` — all pass except `grpc_build`, which requires a local `protoc` and was not run in my environment; no generated files are touched by this PR).

* [ ] Changes in public API reviewed (if applicable) — no public API change (behavioral fix inside the existing deserializer)

### DIFF
--- a/opentelemetry-proto/CHANGELOG.md
+++ b/opentelemetry-proto/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- **Bug fix**: `with-serde` now deserializes `{}` as an empty `AnyValue` (`value: None`). Proto3 JSON encodes a message with no fields set as `{}` — the unset `oneof` the proto documents as a valid "empty" `AnyValue` — and it is what the serializer itself emits for `AnyValue { value: None }`, as well as what the OpenTelemetry Collector's file exporter writes for body-less log records. Previously this failed with "Invalid data for Value, no known keys found"; objects containing only unknown keys still fail. ([#3598](https://github.com/open-telemetry/opentelemetry-rust/issues/3598))
+
 ## 0.32.0
 
 Released 2026-May-08
--- a/opentelemetry-proto/CHANGELOG.md
+++ b/opentelemetry-proto/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## vNext
 
-- **Bug fix**: `with-serde` now deserializes `{}` as an empty `AnyValue` (`value: None`). Proto3 JSON encodes a message with no fields set as `{}` — the unset `oneof` the proto documents as a valid "empty" `AnyValue` — and it is what the serializer itself emits for `AnyValue { value: None }`, as well as what the OpenTelemetry Collector's file exporter writes for body-less log records. Previously this failed with "Invalid data for Value, no known keys found"; objects containing only unknown keys still fail. ([#3598](https://github.com/open-telemetry/opentelemetry-rust/issues/3598))
+- **Bug fix**: `with-serde` now deserializes `{}` as an empty `AnyValue` (`value: None`). Previously this failed with "Invalid data for Value, no known keys found". ([#3598](https://github.com/open-telemetry/opentelemetry-rust/issues/3598))
 
 ## 0.32.0
 
--- a/opentelemetry-proto/src/proto.rs
+++ b/opentelemetry-proto/src/proto.rs
@@ -110,8 +110,10 @@ pub(crate) mod serializers {
                 V: de::MapAccess<'de>,
             {
                 let mut value: Option<any_value::Value> = None;
+                let mut saw_key = false;
 
                 while let Some(key) = map.next_key::<String>()? {
+                    saw_key = true;
                     let key_str = key.as_str();
                     match key_str {
                         "stringValue" => {
@@ -153,6 +155,18 @@ pub(crate) mod serializers {
 
                 if let Some(v) = value {
                     Ok(Some(v))
+                } else if !saw_key {
+                    // An empty JSON object is proto3 JSON's encoding of a
+                    // message with no fields set — for `AnyValue`, the unset
+                    // oneof ("It is valid for all values to be unspecified in
+                    // which case this AnyValue is considered to be 'empty'").
+                    // It is also what this module's own serializer emits for
+                    // `AnyValue { value: None }` (the `flatten` of an unset
+                    // oneof serializes no fields), and what real producers
+                    // emit on the wire: the OpenTelemetry Collector's file
+                    // exporter writes `"body": {}` for body-less event
+                    // records.
+                    Ok(None)
                 } else {
                     Err(de::Error::custom(
                         "Invalid data for Value, no known keys found",
--- a/opentelemetry-proto/tests/json_serde.rs
+++ b/opentelemetry-proto/tests/json_serde.rs
@@ -473,6 +473,32 @@ mod json_serde {
         use super::*;
 
         #[test]
+        fn empty() {
+            // proto3 JSON encodes a message with no fields set as `{}` —
+            // for `AnyValue` that is the unset oneof, which the proto
+            // itself documents as valid ("It is valid for all values to
+            // be unspecified in which case this AnyValue is considered
+            // to be 'empty'"). Real producers emit it on the wire (the
+            // Collector's file exporter writes `"body": {}` for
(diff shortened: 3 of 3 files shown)


jensholdgaard:
The `msrv (windows-latest)` failure is unrelated to this PR — it will fail for any fresh CI run since 2026-07-17:

- The job fails while `cargo-msrv` verifies `opentelemetry-otlp/examples/basic-otlp` against Rust 1.75, with the error inside **tokio**: `` no method named `wait` found for struct `OnceLock` `` at `tokio-1.53.0/src/signal/windows/sys.rs:112`.
- **tokio 1.53.0 was published 2026-07-17** (crates.io) — after this repo's most recent green runs — and fresh dependency resolution now picks it up. It declares `rust-version = "1.71"` but `OnceLock::wait` was stabilized in Rust 1.86, so 1.53.0 violates its own declared MSRV on Windows targets (the `cfg(windows)` signal path presumably isn't covered by tokio's MSRV CI). Filed upstream with tokio (link to follow in this thread).
- This PR touches only `opentelemetry-proto`'s serde module, a test, and its changelog — no dependencies or manifests — so re-running the job cannot help until either tokio yanks/fixes 1.53.0 or the example pins `tokio <1.53` for the MSRV check.

jensholdgaard:
Filed with tokio: tokio-rs/tokio#8299 (1.53.0 declares `rust-version = 1.71` but uses `OnceLock::wait`, stabilized in Rust 1.86, in the `cfg(windows)` signal path).

jensholdgaard:
Upstream fix submitted: tokio-rs/tokio#8300 (removes the `OnceLock::wait`, MSRV-verified with 1.71 on the windows-msvc target). Once tokio ships it (presumably 1.53.1), this repo's `msrv (windows-latest)` job recovers on fresh resolution with no change needed here.

jensholdgaard:
CI status: with tokio 1.53.1 released (tokio-rs/tokio#8300), both `msrv` jobs now pass — the earlier `msrv (windows-latest)` failure was the tokio 1.53.0 MSRV regression tracked in #3600, resolved upstream.

The one remaining red, `coverage`, is a Codecov-infrastructure flake unrelated to this change: the uploader failed to fetch Codecov's GPG public key ("gpg: no valid OpenPGP data found" → "Can't check signature: No public key") and exited before any upload. A job re-run should clear it.

EasyCLA and all other checks pass — ready for review.

opentelemetry-proto/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- **Bug fix**: `with-serde` now deserializes `{}` as an empty `AnyValue` (`value: None`). Proto3 JSON encodes a message with no fields set as `{}` — the unset `oneof` the proto documents as a valid "empty" `AnyValue` — and it is what the serializer itself emits for `AnyValue { value: None }`, as well as what the OpenTelemetry Collector's file exporter writes for body-less log records. Previously this failed with "Invalid data for Value, no known keys found"; objects containing only unknown keys still fail. ([#3598](https://github.com/open-telemetry/opentelemetry-rust/issues/3598))
```
scottgerring: I reckon we can tighten this up a bit - it should be easy to skim over these: 

```suggestion
- **Bug fix**: `with-serde` now deserializes `{}` as an empty `AnyValue` (`value: None`). Previously this failed with "Invalid data for Value, no known keys found". ([#3598](https://github.com/open-telemetry/opentelemetry-rust/issues/3598))
```

scottgerring (changes requested):
LGTM! Tinny nit on the changelog but otherwise 🚢 
Thanks for turning this one around quickly @jensholdgaard

opentelemetry-proto/CHANGELOG.md
jensholdgaard: Applied verbatim in ea8d247 — thanks for the quick review!

lalitb (approved):
Thanks.

lalitb:
@jensholdgaard  -  Thanks for fixing this. This overlaps with #3595, which fixes the same empty `AnyValue` issue and also handles the OTLP/JSON unknown-field behavior.

Since we should merge only one of these, would you be okay if we proceed with #3595 and close this PR?

jensholdgaard:
Closing in favour of #3595, which fixes the same bug (#3598) more completely: it accepts both the empty `{}` case this PR handles *and* unknown-only objects (`{"futureValue":"ignored"}`), matching the OTLP/JSON "ignore unknown fields" rule — where this PR's `saw_key` approach would still reject the unknown-only case. #3595 is also already CI-green (MSRV included) and maintainer-approved.

PR #3605, opened 2026-07-24
fix(sdk): spawn periodic reader worker at reader build time

### LINKED ISSUE #3601: [Bug]: async-runtime PeriodicReader spawns its worker during SdkMeterProvider build, not PeriodicReader::build(), capturing the wrong tokio runtime (deadlock/panic)
### What happened?

`periodic_reader_with_async_runtime::PeriodicReader` takes a `runtime` argument, but does not spawn its worker task in `build()`. Instead it stashes the spawn closure and runs it lazily inside `MetricReader::register_pipeline`, which executes later, during `SdkMeterProvider::builder().build()`:

- `build()` stores the closure as `ProducerOrWorker::Worker` ([periodic_reader_with_async_runtime.rs#L145](https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry_sdk-0.32.1/opentelemetry-sdk/src/metrics/periodic_reader_with_async_runtime.rs#L145))
- `register_pipeline()` invokes it ([periodic_reader_with_async_runtime.rs#L337](https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry_sdk-0.32.1/opentelemetry-sdk/src/metrics/periodic_reader_with_async_runtime.rs#L337)), and `runtime::Tokio::spawn` is a plain `tokio::spawn` ([runtime.rs#L81](https://github.com/open-telemetry/opentelemetry-rust/blob/opentelemetry_sdk-0.32.1/opentelemetry-sdk/src/runtime.rs#L81))

So the tokio runtime the worker lands on is whatever is ambient at **provider** build time, not at **reader** build time — even though the reader is the component that takes the `runtime` argument. This is inconsistent with the async-runtime `BatchSpanProcessor` and `BatchLogProcessor`, which spawn their workers eagerly at `build()` time, and it makes the worker's runtime depend on an unrelated later call site. Two concrete consequences below, both reproducible with only `opentelemetry`, `opentelemetry_sdk`, and `tokio` (no exporter/network involved).

I'd expect the worker to be spawned in `PeriodicReader::build()`, matching the span/log processors, so that the runtime context in which the reader is constructed is the one that matters.

**Shared `Cargo.toml` for both examples:**

```toml
[package]
name = "otel-mre"
version = "0.1.0"
edition = "2021"

[dependencies]
opentelemetry = "0.32"
opentelemetry_sdk = { version = "0.32", features = [
    "experimental_metrics_periodicreader_with_async_runtime",
    "rt-tokio",
    "testing",
] }
tokio = { version = "1", features = ["full"] }
```

#### Example 1: `shutdown()` deadlocks on a current-thread runtime

When everything runs on one current-thread runtime — i.e. what any default `#[tokio::test]` or `#[tokio::main(flavor = "current_thread")]` does — the worker is spawned onto the application's own runtime, and `shutdown()` deadlocks: `shutdown_with_timeout` blocks the calling thread with `futures_executor::block_on` while the worker it is waiting on is parked on that same (only) thread.

```rust
use opentelemetry::metrics::MeterProvider;
use opentelemetry_sdk::metrics::{
    InMemoryMetricExporter, SdkMeterProvider, periodic_reader_with_async_runtime::PeriodicReader,
};
use opentelemetry_sdk::runtime;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    // watchdog so the MRE terminates instead of hanging forever
    std::thread::spawn(|| {
        std::thread::sleep(std::time::Duration::from_secs(10));
        eprintln!("DEADLOCK: shutdown() did not return within 10s");
        std::process::exit(1);
    });

    let exporter = InMemoryMetricExporter::default();
    let provider = SdkMeterProvider::builder().with_reader(reader).build();

    provider.meter("app").u64_counter("my_counter").build().add(1, &[]);

    println!("shutting down...");
    provider.shutdown().unwrap();
    println!("shutdown complete"); // never reached
}
```

I realize blocking `shutdown()` inside an async context is itself questionable, but the deadlock can't be avoided by constructing the reader somewhere better: no runtime context at reader-build time is honored, as the next example shows.

#### Example 2: the runtime entered when building the reader is ignored; provider build panics outside a runtime

Building the reader inside a dedicated runtime's context (e.g. to isolate exporter work from the application, in the spirit of #3084) has no effect — the spawn happens at provider build, which panics if no runtime is ambient there:

```rust
use opentelemetry_sdk::metrics::{
    InMemoryMetricExporter, SdkMeterProvider, periodic_reader_with_async_runtime::PeriodicReader,
};
use opentelemetry_sdk::runtime;

fn main() {
    let rt = tokio::runtime::Builder::new_multi_thread()
        .worker_threads(1)
        .enable_all()
        .build()
        .unwrap();

    // Build the reader inside a dedicated runtime's context...
    let reader = {
        let _guard = rt.enter();
        PeriodicReader::builder(InMemoryMetricExporter::default(), runtime::Tokio).build()
    };

    // ...but the worker is actually spawned HERE, on the ambient runtime — of
    // which there is none:
    let provider = SdkMeterProvider::builder().with_reader(reader).build();
    // panics: "there is no reactor running, must be called from the context of
    // a Tokio 1.x runtime"

    provider.shutdown().unwrap();
}
```

#### Downstream context

We hit this in the `logfire` SDK (pydantic/logfire-rust#141): the worker landing on the application's runtime instead of our dedicated suppressed export runtime meant (a) `shutdown()` hung for any current-thread-runtime app, and (b) the metric export's own hyper/reqwest/mio `tracing` events were captured and exported as telemetry (13 noise log records per export in our measurement). Our workaround (pydantic/logfire-rust#142) is a `MetricReader` wrapper whose `register_pipeline` enters the desired runtime's handle before delegating — which works, but relies on knowing this implementation detail.

#### Suggested fix

Spawn the worker in `PeriodicReader::build()`, matching `BatchSpanProcessor`/`BatchLogProcessor` with async runtimes. (If lazy spawning must be kept, capturing a `tokio::runtime::Handle` at build time and using it in `register_pipeline` would also fix the surprise.)

0.32

0.32.1

N/A (reproduces with `InMemoryMetricExporter`; originally observed with OTLP)

### Relevant log output

```shell
$ cargo run --bin hang
shutting down...
DEADLOCK: shutdown() did not return within 10s

$ cargo run --bin panic
thread 'main' panicked at .../opentelemetry_sdk-0.32.1/src/runtime.rs:81:17:
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

karan-vk:
Fixes #3601

## Changes

`periodic_reader_with_async_runtime::PeriodicReader` stored its spawn closure and only ran it inside `register_pipeline`, so the worker task landed on whatever tokio runtime was ambient when the meter provider was built rather than when the reader, which takes the `runtime` argument, was built. Building the reader and the provider in different contexts could deadlock or panic.

The worker is now spawned eagerly in `build()`, matching the async-runtime `BatchSpanProcessor` and `BatchLogProcessor`, and `ProducerOrWorker` is reduced to a plain `Option` producer. Added a regression test that builds the reader inside a runtime and the provider outside of one; it panics without this fix.

Behavior note: a reader that is built but never attached to a provider now runs its interval ticker and logs collect errors, where before the worker was never spawned. This matches the existing behavior of the async-runtime batch processors.

* [ ] Changes in public API reviewed (if applicable) - no public API changes

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## vNext
 
+- `periodic_reader_with_async_runtime::PeriodicReader` now spawns its background
+  worker when the reader is built instead of when the meter provider is built.
+  The worker therefore runs on the async runtime that is ambient at reader
+  construction, where the `runtime` argument is supplied, instead of capturing
+  whichever runtime is ambient at provider construction, which could deadlock or
+  panic when the two contexts differed.
+  ([#3601](https://github.com/open-telemetry/opentelemetry-rust/issues/3601))
+
 - Made `futures-channel`, `futures-executor`, `futures-util`, and `thiserror`
   optional, enabling a minimal SDK build. With `default-features = false`, the
   SDK's only dependency is the `opentelemetry` API crate.
--- a/opentelemetry-sdk/src/metrics/periodic_reader_with_async_runtime.rs
+++ b/opentelemetry-sdk/src/metrics/periodic_reader_with_async_runtime.rs
@@ -1,5 +1,5 @@
 use std::{
-    env, fmt, mem,
+    env, fmt,
     sync::{Arc, Mutex, Weak},
     time::Duration,
 };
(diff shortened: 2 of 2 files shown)


scottgerring:
I think #3413 should help out here also and we should aim to get that in soon to clear up all of these _weird unexpected deadlocks_ issues at once

karan-vk:
Took a proper look through #3413 and I think it's a really nice piece of work. Consolidating `Tokio`/`TokioCurrentThread` into a single runtime that detects the flavor at spawn time (and treating a single-worker multi_thread runtime as starvation-prone too, which is what any 1-vCPU container gets by default) removes a whole class of footguns, and the `SpawnHandle` associated type keeps the trait open for custom runtimes pinned to a specific `Handle`. Shutdown joining the worker directly and surfacing the final export error, and even worker panics, is a big step up from the oneshot signalling, and it's great to see the #2056 test un-ignored. Fully agree it should go in soon. Happy to help there, whether that's review, testing it against the #3601 repros, or rebase gruntwork.

For this PR specifically, I think the two compose rather than overlap: #3413 fixes how and where the spawn executes and how shutdown waits, while this one fixes when the worker is captured (at reader build, where the `runtime` argument is supplied, matching the async-runtime batch processors). As #3413 stands, the worker closure is still invoked inside `register_pipeline`, so the flavor detection runs against whatever context is ambient at provider build. The case in pydantic/logfire-rust#141 (reader built inside a dedicated runtime to keep hyper/reqwest's own `tracing` events out of the exported telemetry) would still spawn onto the app runtime whenever that one is a healthy multi-thread runtime.

A nice side effect of landing this one first is that the #3413 rebase of this file gets simpler: with the spawn in `build()`, the handle exists at construction, so `worker_handle` can start out as `Running(handle)`, and the `WorkerState::Complete(Ok(()))` placeholder, the placeholder worker closure, and `ProducerOrWorker` itself all fall away. It's a small change with no public API impact and it's rebased on latest main. Happy to go whichever way you prefer, though.

scottgerring:
Hey @karan-vk , thanks for looking more into this!
I've looked again and I agree. Almost, neither my clanker nor my actual human review could find anything to fault; thanks for the minimal, clean improvement. LGTM. 

@cijothomas would be great to merge this when you have a moment!

scottgerring (approved):

PR #3610, opened 2026-07-30
test: move self-observability tests into self_obs modules and auto-discover them in test.sh

cijothomas:
Self-observability tests each need their own process (set-once global MeterProvider), so `scripts/test.sh` listed one line per test. Every new metric PR appended to that block, causing repeated merge conflicts.

This moves those tests into `self_obs` modules and replaces the hardcoded list with a discovery loop that finds ignored tests by the `::self_obs::` path marker. New metric tests are picked up automatically.

No behavior change; test infrastructure only.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -45,12 +45,19 @@ cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features trace::ru
 cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features trace::runtime_tests::test_set_provider_single_thread_tokio_shutdown -- --ignored --exact
 
 echo "Running ignored tests for opentelemetry-sdk package (self-diagnostics, requires global MeterProvider)"
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::batch_log_processor::tests::self_diagnostics_counter_records_success -- --ignored --exact
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::batch_log_processor::tests::self_diagnostics_counter_records_queue_full_drops -- --ignored --exact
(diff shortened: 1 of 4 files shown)


scottgerring:
Semi-related: maybe we should move to nexttest? This will fix this part:

> Self-observability tests each need their own process (set-once global MeterProvider), so scripts/test.sh listed one line per test. Every new metric PR appended to that block, causing repeated merge conflicts.

scripts/test.sh
```diff
@@ -45,18 +45,19 @@ cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features trace::ru
...
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::batch_log_processor::tests::self_diagnostics_counter_records_queue_full_drops -- --ignored --exact
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::batch_log_processor::tests::self_diagnostics_counter_records_already_shutdown_drops -- --ignored --exact
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::batch_log_processor::tests::self_diagnostics_queue_capacity -- --ignored --exact
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::simple_log_processor::tests::self_diagnostics_counter_records_success -- --ignored --exact
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::simple_log_processor::tests::self_diagnostics_counter_records_already_shutdown_drops -- --ignored --exact
-cargo test --manifest-path=opentelemetry-sdk/Cargo.toml --all-features logs::logger::tests::log_created_counts_intake_without_processors -- --ignored --exact
+# Self-observability tests live in `self_obs` modules and each needs its own
```
scottgerring, line 48: if we wanted to move to nextest, a great deal of this would disappear

scottgerring (approved):
seems sensible; might be worth moving to nextest in the long run

PR #3611, opened 2026-07-31
feat(sdk): add log processor queue capacity metric

cijothomas:
Adds `otel.sdk.processor.log.queue.capacity` for `BatchLogProcessor` as an observable up-down counter, gated behind `experimental_metrics_bound_instruments`.


opentelemetry-sdk/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- Added SDK self-observability metrics `otel.sdk.processor.log.queue.size` and
```
scottgerring: nit: missing PR link

scottgerring:
Grepping through the code, should we also update this design doc TODO?

https://github.com/open-telemetry/opentelemetry-rust/blob/53a828af3a75b4f4bd62d3459abbf8ad518a5d39/docs/design/observability.md?plain=1#L92-L93

opentelemetry-sdk/src/logs/batch_log_processor.rs
```diff
@@ -404,6 +404,33 @@ impl BatchLogProcessor {
...
+                .with_description(
+                    "The maximum number of log records the queue of a given instance of \
+                     an SDK log processor can hold.",
+                )
+                .with_unit("{log_record}")
+                .with_callback(move |observer| {
+                    // The capacity value is constant; the Weak upgrade is only a
+                    // liveness guard so a dropped processor stops emitting this
+                    // otherwise-unregisterable callback.
+                    if capacity_state.upgrade().is_some() {
```
utpilla, line 428: The capacity callback only needs a liveness check. It observes a constant and never reads through the `Weak`. `upgrade()` does more than needed (CAS to build an Arc, then drop it); a plain `strong_count()` load expresses the intent directly. Not a perf concern (since it only runs at collection time) but it's just clearer with the intention.


```suggestion
                    // otherwise-unregisterable callback.  We just need to know the
                    // processor is still alive, so check `strong_count()` rather
                    // than materializing an Arc via `upgrade()`.
                    if capacity_state.strong_count() > 0 {
```

utpilla (approved):

PR #3618, opened 2026-08-02
fix(opentelemetry): test with `--no-default-features` & fix OOB panic

MggMuggins:
Cargo's feature unification (rfcs#3692) means that opentelemetry_sdk implicitly enables the default features of opentelemetry, so `--no-default-features` actually builds with default features enabled.

This masked a bug introduced in 7087f65c when building without `trace`. Prefer ignoring a few doctests to silently reducing test coverage for non-default feature combinations via feature unification.

Fixes: 7087f65c00a1fed22f29bbc67b0dffb38b30d73c
Fixes: 627f252d23ef3e2ee598855dd19ec63a28fcc794

I found this while packaging a few otel crates for Debian; our CI pipeline runs the tests for each feature from crates.io, so feature unification doesn't happen in our environment.

## Changes

Mostly trivial disabling tests to fix the tests with various feature combinations.

I'd like to flag that **I don't fully understand** the high-level purpose & semantics of `pop_id`; I believe the changes that I've made here align with the semantics of the function when `trace` is enabled, but I'd appreciate a careful review of that part of the PR.

* [ ] ~~Unit tests added/updated (if applicable)~~
* [ ] ~~Changes in public API reviewed (if applicable)~~

### DIFF
--- a/opentelemetry/src/context.rs
+++ b/opentelemetry/src/context.rs
@@ -590,6 +590,8 @@ impl ContextStack {
             );
             #[cfg(feature = "trace")]
             return None;
+            #[cfg(not(feature = "trace"))]
+            return;
         }
         let len: u16 = self.stack.len() as u16;
         // Are we at the top of the [`ContextStack`]?
@@ -615,6 +617,8 @@ impl ContextStack {
             }
             #[cfg(feature = "trace")]
             return None;
+            #[cfg(not(feature = "trace"))]
+            return;
         } else {
             // This is an out of order pop.
             if pos >= len {
@@ -627,6 +631,8 @@ impl ContextStack {
                 );
                 #[cfg(feature = "trace")]
                 return None;
+                #[cfg(not(feature = "trace"))]
+                return;
             }
             // Clear out the entry at the given id and extract its span
             #[cfg(feature = "trace")]
@@ -657,7 +663,9 @@ impl Default for ContextStack {
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[cfg(feature = "futures")]
     use std::time::Duration;
+    #[cfg(feature = "futures")]
     use tokio::time::sleep;
 
     #[derive(Debug, PartialEq)]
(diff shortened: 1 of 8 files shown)


MggMuggins:
Unfortunately I need to check with my employer regarding the CLA; my apologies. I'll try to get back soon.

opentelemetry/src/logs/logger.rs
```diff
@@ -31,7 +31,7 @@ pub trait LoggerProvider {
     ///
     /// # Examples
     ///
-    /// ```
+    /// ```ignore
```
scottgerring, line 34: It seems like the fundamental tension here is that these doctests are trying to show how to use opentelemetry traits, but using the SDK types to do so. It feels like the cleanest thing to do would be to write these in a fashion that they don't require the SDK types, but I don't know how illuminating the resulting doc tests would be ... 

I see two useful options:
**ignore them**
in which case, can we also remove the dev dep in cargo.toml completely? 

**feature gate them**
If we feature-gated these tests too (like you already do in lib.rs and trace_context.rs), then the dev-dep could stay with default-features = false and the doctests would still run in workspace CI where opentelemetry_sdk gets its features via workspace unification. But then this would fail:

```sh
cargo test -p opentelemetry --all-features --doc 
```


I'm pretty happy to just ignore them and fix the build issue at hand to be honest. If we want to clean this up later, we can have a go at rewriting them in abstract terms and see how they look.

opentelemetry/src/context.rs
```diff
@@ -590,6 +590,8 @@ impl ContextStack {
             );
             #[cfg(feature = "trace")]
             return None;
+            #[cfg(not(feature = "trace"))]
```
scottgerring, line 593: 👍

scottgerring (approved):
This is a bit fiddly eh!
IMHO ignoring them is the pragmatic response here, unless @cijothomas or @lalitb or whoever goes to click the merge button sees a better option.

cijothomas:
@MggMuggins Could you sign the EasyCLA part? That is a pre-req before we can consider any contributions.

MggMuggins:
I really do apologize for the CLA delay; I'm following up again with my employer.

scottgerring:
@MggMuggins could you sign the CLA or rather not?

MggMuggins:
I'm still working to get permission from my employer and/or be covered by the corporate CLA; so not right now :slightly_frowning_face: . I'll follow up as soon as I know more/have a course of action. Apologies for the mess.

PR #3624, opened 2026-08-06
feat(sdk): collect exemplars for explicit-bucket histograms

### LINKED ISSUE #3369: Implement exemplar collection for metrics (required for stable)
## Summary

While reviewing the metrics SDK implementation, I noticed that exemplar data structures exist but no actual collection or sampling logic is implemented. Every aggregation code path unconditionally passes `exemplars: vec![]`, meaning exemplars are never populated regardless of configuration.

The OpenTelemetry specification marks exemplar support as **Stable**, making this a requirement for a conformant 1.0 release. Both the Go and Java SDKs fully implement exemplars today.

## Current State

**What exists:**
- `Exemplar<T>` struct in `opentelemetry-sdk/src/metrics/data/mod.rs` with all the right fields (`filtered_attributes`, `time`, `value`, `span_id`, `trace_id`)
- `exemplars: Vec<Exemplar<T>>` field on every data point type (`GaugeDataPoint`, `SumDataPoint`, `HistogramDataPoint`, `ExponentialHistogramDataPoint`)
- Accessor methods like `.exemplars()` on each data point type
- Proto/OTLP `Exemplar` type for serialization

**What is missing (all hollow):**
- Every aggregation unconditionally sets `exemplars: vec![]` — there are **21 occurrences** across `sum.rs`, `histogram.rs`, `exponential_histogram.rs`, `last_value.rs`, `precomputed_sum.rs`, and `aggregate.rs`
- No `ExemplarFilter` trait or implementations (`AlwaysOn`, `AlwaysOff`, `TraceBased`)
- No `ExemplarReservoir` trait or implementations (`SimpleFixedSizeExemplarReservoir`, `AlignedHistogramBucketExemplarReservoir`)
- No integration of exemplar sampling into the measurement recording pipeline
- No support for the `OTEL_METRICS_EXEMPLAR_FILTER` environment variable

## What Needs to Be Implemented

Per the [OpenTelemetry Metrics SDK specification](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplar):

1. **`ExemplarFilter`** — determines which measurements are eligible to become exemplars
   - `AlwaysOn` — every measurement is eligible
   - `AlwaysOff` — no measurements are eligible
   - `TraceBased` (default) — only measurements recorded within a sampled span are eligible

2. **`ExemplarReservoir`** — stores and samples exemplars per time series
   - `SimpleFixedSizeExemplarReservoir` — uniform random sampling across a fixed number of slots
   - `AlignedHistogramBucketExemplarReservoir` — one exemplar per histogram bucket (default for explicit bucket histograms)

3. **Pipeline integration** — wire exemplar sampling into `record()` / `measure()` paths in all aggregators so that exemplars are actually collected and emitted in exported data points

4. **`OTEL_METRICS_EXEMPLAR_FILTER` environment variable** — allow users to configure the exemplar filter via env var (`always_on`, `always_off`, `trace_based`)

5. **View-level configuration** — allow configuring both the filter and reservoir through the View API

6. **Context propagation** — extract `span_id` and `trace_id` from the current context when recording an exemplar, and capture filtered attributes

## Spec Requirements to Note

- Exemplar sampling SHOULD be turned on by default
- If exemplar sampling is off, the SDK MUST NOT have overhead related to exemplar sampling
- The default filter is `TraceBased`
- Custom reservoir implementations must be supported
- Each reservoir instance is per-metric-stream (per unique attribute set)

## References

- [OTel Metrics SDK Spec: Exemplar](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplar)
- [OTel Metrics SDK Spec: ExemplarFilter](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplarfilter)
- [OTel Metrics SDK Spec: ExemplarReservoir](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exemplarreservoir)
- [Go SDK implementation](https://github.com/open-telemetry/opentelemetry-go/tree/main/sdk/metric/internal/exemplar)
- [Java SDK implementation](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/exemplar)

ormeilu:
Closes #3369

## What this does

`Exemplar<T>` and the `exemplars` fields on data points already existed in the SDK, and the read accessors were already public, but nothing ever populated them. This adds the collection half for the **explicit-bucket histogram**.

A measurement recorded inside a sampled span is retained alongside the histogram data point together with its trace id and span id, so a backend can link from a latency bucket straight to a representative trace.

```rust
use opentelemetry_sdk::metrics::{ExemplarFilter, SdkMeterProvider};

    .with_exemplar_filter(ExemplarFilter::TraceBased) // the default
    .with_reader(reader)
    .build();
```

Everything is behind a new `spec_unstable_metrics_exemplars` feature, following the precedent of `spec_unstable_metrics_views`.

## Spec conformance

- **`ExemplarFilter`**: `AlwaysOn`, `AlwaysOff`, `TraceBased` (default). `TraceBased` checks the **sampled** flag, not merely the presence of a span.
- **`AlignedHistogramBucketExemplarReservoir`**: at most one exemplar per bucket, uniformly weighted (reservoir sampling, k=1).
- **Exemplar contents**: value, time, trace id and span id of the span active when recording, and the attributes a view's attribute filter dropped. This includes recordings through **bound instruments**, which resolve their dropped attributes once at bind time.
- Reservoirs are drained on every collection in both temporalities, so a stale trace id is never re-exported.

## Design: the reservoir decides first

Nothing is allocated, stamped or cloned for a measurement the reservoir does not keep.

1. `ExemplarSampler::offer` applies the filter. If the measurement is eligible it copies the 24 bytes of trace id and span id into a **stack-only** `ExemplarOffer` that merely borrows the measurement's attributes. There is no `Box`, no clock read, and no `SpanContext` clone.
2. The aggregator receives one nullable pointer to that offer inside its `PreComputedValue`.
3. Under the `Mutex<Buckets<T>>` the histogram already takes, the reservoir makes its 1/n decision. **Only if it keeps the measurement** does it read the clock, collect the filtered attributes and build the `Exemplar`.

Consequences:

- **`Aggregator::PreComputedValue` becomes a GAT** (`PreComputedValue<'a>`), so the histogram's value can borrow the offer. The other aggregators change by one line (`type PreComputedValue<'a> = T;`). This was the only way I found to get "no allocation unless stored" without `unsafe`. Happy to take another route if you'd rather not touch the trait.
- With the feature **off**, the offer reference is a ZST, so `PreComputedValue` is `(T, usize)`-sized, exactly as on `main`.
- The reservoir allocates its slots lazily on the first kept exemplar, so an attribute set that never records inside a sampled span costs no memory.
- The timestamp is read under the lock, so an exemplar's time can never precede the start of the interval it is exported with when a collection races the measurement.
- The reservoir draw uses a thread-local `SmallRng`. It needs uniformity, not unpredictability.
- **Tradeoff:** for a *kept* measurement the view's attribute filter runs again under the per-series lock to produce the dropped attributes. That is about one measurement per bucket per interval at steady state, versus an allocation per eligible measurement before.

## Performance

`cargo bench --bench metrics_histogram`, feature on, same machine. "Before" is the earlier revision of this PR, which boxed the offer.

| Benchmark | Before | Now |
|---|---|---|
| `Histogram_Record_Exemplar_AlwaysOff` | 63.2 ns | 65.0 ns (within noise) |
| `Histogram_Record_Exemplar_TraceBased` (no span) | 68.7 ns | 67.1 ns |
| `Histogram_Record_Exemplar_AlwaysOn` | 119.3 ns | **73.4 ns** |
| `Histogram_Record_Exemplar_TraceBased_SampledSpan` (new) | 253.8 ns | **71.3 ns** |
| `Histogram_RecordAndCollect_SampledSpan_TraceBased` (new) | 552–755 ns | 499–666 ns |

`Histogram_Record` with the feature **off**: 163.7 ns. `AlwaysOff` returns before touching thread-local storage.

### Why it is faster now

The earlier revision built the whole exemplar for **every eligible measurement**, before the reservoir had decided anything:

| Work per eligible measurement | Before | Now |
|---|---|---|
| heap-allocate and free a `Box` for the offer | always | never |
| read the clock | always | only when kept |
| clone the `SpanContext`, which deep-copies its `TraceState` (a `VecDeque` plus its `String`s) | always | never, it is borrowed |
| draw from the ChaCha thread rng under the lock | always | `SmallRng` |

Reservoir sampling with k=1 keeps the n-th measurement of a bucket with probability 1/n, so in a busy bucket nearly every offer is discarded. The earlier revision paid ~190 ns per measurement to build something it then threw away; now a discarded offer costs ~8 ns over `AlwaysOff`.

That is what the two new benchmarks separate:

- `TraceBased_SampledSpan` records inside a sampled span carrying a `tracestate` and never collects. It is the steady state of a busy bucket, where offers are discarded, and it is where the −71% comes from. The earlier benchmarks never recorded inside a span, so they never exercised this path at all.
- `RecordAndCollect_SampledSpan` collects after every recording, so **every** recording is one the reservoir keeps. Collection dominates the number and it is noisy (two runs shown), but it bounds the other extreme: a kept exemplar costs about what it did before. The gain is large for a busy histogram and small for a quiet one.

## Privacy note

As the spec requires, an exemplar retains the attributes a view's attribute filter removed. So a view is **not** a redaction mechanism once exemplars are collected. This is called out in the docs of `ExemplarFilter` and `with_exemplar_filter`, and in the CHANGELOG.

## Scope, and what is deliberately left out

- **Sum and gauge**: lock-free atomics today. A reservoir needs a mutex, so this deserves its own PR and benchmark.
- **Exponential histogram**: needs `SimpleFixedSizeExemplarReservoir`. Mechanical follow-up.
- **Asynchronous instruments**: no meaningful ambient context in a callback. They are forced to `AlwaysOff` and emit no exemplars.
- **Known gap, cardinality overflow**: an exemplar on the `otel.metric.overflow` data point carries only the view-dropped attributes, not the measurement's own attributes that overflow discarded. Fixing it means threading the overflow decision out of `ValueMap::measure`. I'd rather do that separately.

## Checks

- `cargo fmt --all` clean
- `cargo clippy -p opentelemetry_sdk --all-targets --all-features -- -Dwarnings` clean
- clippy `--lib -Dwarnings` clean for: `metrics`, `metrics + exemplars`, `metrics + bound_instruments`, and `metrics + exemplars + bound_instruments`
- `cargo test -p opentelemetry_sdk --all-features --lib` passes

## Tests

- Filter behaviour for all three variants.
- Trace and span ids match the active sampled span; nothing is recorded under an unsampled span or with no span; `AlwaysOn` keeps an unsampled span's ids.
- At most one exemplar per bucket, and every measurement in a bucket is equally likely to survive (statistical test, bounds >7σ wide).
- Concurrent recordings from different spans never pair a value with another span's ids.
- No leak across collection cycles, in delta and cumulative.
- View-dropped attributes are retained, through both the unbound **and the bound** path.
- A bound handle captures the span active at record time, not at bind time.
- The reservoir allocates nothing until a measurement is eligible; an exemplar is stamped when it is kept.
- End-to-end tests through `with_exemplar_filter`.


scottgerring:
Hey @ormeilu , thanks for the PR!
I've chucked a label on so the benchmarks will run.
I will have a skim over this, but I think @cijothomas will need to be involved as metrics is an area of depth of his

opentelemetry-sdk/src/metrics/internal/exemplar/enabled.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub(crate) fn offer(&self) -> Option<Box<ExemplarOffer>> {
+        match self.filter {
+            ExemplarFilter::AlwaysOff => None,
+            ExemplarFilter::AlwaysOn => Some(Box::new(Context::map_current(|cx| {
+                let mut offer = ExemplarOffer::at(now());
+                // An unsampled or absent span leaves the ids zeroed, which is
+                // what the spec asks for: the measurement is still eligible.
+                if cx.has_active_span() {
+                    let span_cx = cx.span().span_context().clone();
+                    if span_cx.is_sampled() {
```
scottgerring: Is this too restrictive? It's not clear to me from the spec that we should be dropping span and trace ID just because they are unsampled

opentelemetry-sdk/src/metrics/pipeline.rs
```diff
@@ -416,10 +418,16 @@ where
             let cardinality_limit = stream
                 .cardinality_limit
                 .unwrap_or(DEFAULT_CARDINALITY_LIMIT);
+            #[cfg(feature = "spec_unstable_metrics_exemplars")]
+            let exemplars = ExemplarSampler::new(self.pipeline.exemplar_filter);
+            #[cfg(not(feature = "spec_unstable_metrics_exemplars"))]
+            let exemplars = ExemplarSampler::new();
```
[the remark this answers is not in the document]
ormeilu: Good, I'll use AlwaysOff for observable instrument kinds and add a regression test

opentelemetry-sdk/src/metrics/internal/histogram.rs
```diff
@@ -51,23 +62,25 @@ where
...
     /// returns buckets with `n` bins.
     fn new(n: usize) -> Buckets<T> {
         Buckets {
             counts: vec![0; n],
+            count: 0,
+            total: T::default(),
             min: T::max(),
             max: T::min(),
-            ..Default::default()
+            exemplars: AlignedHistogramBucketReservoir::new(n),
```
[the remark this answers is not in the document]
ormeilu: Also good,  I'll preserve that representation but allocate at least one reservoir slot

opentelemetry-sdk/src/metrics/internal/exemplar/enabled.rs
ormeilu: Yes, good catch
I've updated AlwaysOn to retain IDs from any valid active span and leave them zero only when no valid span is present

ormeilu:
Benchmark results across fresh runs on the same machine:

| Build | `Histogram_Record` | vs main |
|---|---:|---:|
| `main` (baseline) | 147.16 ns | — |
| branch, feature **off** | 147.43 ns | +0.18% |
| branch, feature **on**, default `TraceBased`, no active span | 154.95 ns | **+5.29%** |

Feature-on filter isolation across two alternating runs:

| Filter | Time | Delta vs `AlwaysOff` |
|---|---:|---:|
| `AlwaysOff` | 60.49 ns | — |
| `TraceBased` (no active span) | 62.41 ns | +1.92 ns |
| `AlwaysOn` | 107.49 ns | +47.00 ns |

`AlwaysOn` is more expensive because it creates an offer and reads the clock for every measurement.

Commands used toggled only `spec_unstable_metrics_exemplars`, with `experimental_metrics_custom_reader` enabled in both builds.

ormeilu:
@scottgerring hey! Is there anything else I should do?

scottgerring:
> @scottgerring hey! Is there anything else I should do?

hey @ormeilu I've just asked @cijothomas to take a look; he's spent a great deal of time on metrics and will have good feedback. Sorry for the delay!

ormeilu:
Bumping this PR

cijothomas:
@ormeilu Sorry for the delay. I am currently focusing on OTLP stabilization as we plan to declare it stable with a candidate release this week. I'll get back to this PR as soon as the OTLP release is out, by end of this week.

opentelemetry-sdk/src/metrics/exemplar.rs
```diff
@@ -0,0 +1,38 @@
...
+///     .with_exemplar_filter(ExemplarFilter::AlwaysOn)
+///     .build();
+/// ```
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum ExemplarFilter {
+    /// Every measurement is eligible to become an exemplar.
+    ///
+    /// Measurements recorded outside an active span still produce an exemplar,
+    /// with an all-zero trace id and span id. An active span's ids are retained
```
cijothomas: An active span's ids are retained
    /// regardless of its sampling decision.

^ Unsure what this statement means.

ormeilu: it meant: under `AlwaysOn`, a measurement recorded inside an unsampled span still produces an exemplar, and that exemplar keeps the unsampled span's real trace id and span id. They don't get zeroed. That's the difference from `TraceBased`, which drops those measurements. The all-zero ids only show up when there's no active span at all

anyway, I've reworded the doc, thank you!

opentelemetry-sdk/src/metrics/internal/exemplar/enabled.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /// Records the attributes that aggregation is about to discard.
+    ///
+    /// The spec requires an exemplar to retain any attribute from the
+    /// measurement that the aggregated time series does not preserve. Without a
+    /// view filter `filtered` is the same slice as `attrs` and there is nothing
+    /// to record, which is the overwhelmingly common case — so the length check
+    /// short-circuits before any allocation.
+    pub(crate) fn set_filtered_attributes(&mut self, attrs: &[KeyValue], filtered: &[KeyValue]) {
```
cijothomas: How would this work in case of Bound instruments? Bind check View based filters at bind time. The recording time - there is no attributes passed at all. So we need to store the filtered attributes at bind time itself so it can be used in Exemplars.

opentelemetry-sdk/src/metrics/internal/exemplar/enabled.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Boxed so that the ineligible case — every measurement outside a sampled
+    /// span, which is the common one — costs a null pointer to carry down to
+    /// the aggregator rather than a 64-byte struct. The allocation happens only
+    /// for measurements that are actually going to be sampled, which are
+    /// allocating an `Exemplar` anyway.
+    #[inline]
+    pub(crate) fn offer(&self) -> Option<Box<ExemplarOffer>> {
+        match self.filter {
+            ExemplarFilter::AlwaysOff => None,
+            ExemplarFilter::AlwaysOn => Some(Box::new(Context::map_current(|cx| {
```
cijothomas: This boxing causes heap allocation, which should be avoided. At this stage, the Reservoir hasn't decided if it'll store the exemplar or not. So allocation here is not acceptable. it may be okay if the reservoir chose to store the measurement as an exemplar .

opentelemetry-sdk/src/metrics/internal/exemplar/enabled.rs
ormeilu: yep, missed the bound path. fixed in 754b0a5 - dropped attrs are saved at bind time now, test added

opentelemetry-sdk/src/metrics/internal/exemplar/enabled.rs
ormeilu: you're right, fixed in 754b0a5 - nothing is allocated now unless the reservoir keeps the exemplar

had to make `PreComputedValue` a GAT for that (so histogram can borrow the offer), not sure if you're ok with it?

bench inside a sampled span went 253ns -> 71ns btw

ormeilu:
@cijothomas fixed both, also updated the description with new numbers

the GAT change on `PreComputedValue` is the only thing I'm unsure about, wdyt?

no rush, I know you're busy with OTLP

PR #3621, opened 2026-08-05
feat: support OTLP retry for sync exporters & stabilise feature flags 

scottgerring:
Retry with exponential backoff is a spec MUST for OTLP exporters, but until now it only worked behind `experimental-grpc-retry` / `experimental-http-retry` feature flags. Those flags existed because the retry implementation depended on the SDK's experimental `Runtime` trait just to sleep between attempts.

This PR removes that dependency entirely:

- Replaces `Runtime::delay()` with a small internal `sleep_for()` helper that uses `tokio::time::sleep` when inside a Tokio runtime, or `std::thread::sleep` on bare OS threads
- Adds a time budget (deadline = exporter timeout) to cap total retry duration
- Caps server-provided throttle delays (Retry-After / RetryInfo) at 30s
- Removes the `experimental-grpc-retry` and `experimental-http-retry` feature flags

The default `RetryPolicy` has `max_retries: 0` (no retry), so this is not a behaviour change out of the box. Users opt in with `.with_retry_policy(RetryPolicy::recommended())` or a custom policy. We can flip the default in a couple versions, but for now we're aiming to hit the feature requirements of the OTLP exporters with a safe change.

The adaptive sleep means both paths work correctly:
- Default batch processors (dedicated OS thread, no Tokio reactor) → `std::thread::sleep`
- Async processors / gRPC with tonic (Tokio runtime present) → cooperative `tokio::time::sleep`

This is overall a simplification - no new public traits, feature flags, or deps, net loss of about 150 LoC. 

Relates to #3061

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-rust/blob/main/CONTRIBUTING.md) guidelines followed
* [ ] Unit tests added/updated (if applicable)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
* [ ] Changes in public API reviewed (if applicable)

PR #3627, opened 2026-08-13
fix(http): preserve error responses for retry classification

cijothomas:
## Changes

- return HTTP 4xx/5xx responses from built-in reqwest and hyper clients instead of converting them into transport errors
- preserve status codes and headers such as `Retry-After` for OTLP retry classification
- add regression coverage for status and header preservation with the blocking reqwest client

## Why

PR #3621 relies on the HTTP client returning non-success responses so the exporter can distinguish retryable, non-retryable, and throttled failures. Previously, the built-in clients called `error_for_status`, causing HTTP 400 to be retried and HTTP 429 to lose its `Retry-After` hint.

## Validation

- `cargo test -p opentelemetry-http --all-features --lib`
- `cargo clippy -p opentelemetry-http --all-targets --all-features -- -Dwarnings`
- external OTLP HTTP failure harness covering 400, 429, 503, deadline, exhaustion, and connection-refusal scenarios

### DIFF
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Return HTTP error responses from the built-in reqwest and hyper clients instead
   of converting 4xx and 5xx statuses into transport errors. This preserves the
   response status and headers for exporter retry classification.
+  If your code relied on `send_bytes` returning `Err` for non-success statuses,
+  call `ResponseExt::error_for_status()` on the response instead.
 - **Breaking** Removed `reqwest-rustls-webpki-roots` feature. The `webpki-roots` cargo feature was
   removed from `reqwest` in v0.13.0. Use `reqwest-rustls` instead, which now correctly enables
   `reqwest/rustls` (platform native trust roots). To use Mozilla's embedded CA bundle, construct a
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## vNext
 
+- Return HTTP error responses from the built-in reqwest and hyper clients instead
+  of converting 4xx and 5xx statuses into transport errors. This preserves the
+  response status and headers for exporter retry classification.
 - **Breaking** Removed `reqwest-rustls-webpki-roots` feature. The `webpki-roots` cargo feature was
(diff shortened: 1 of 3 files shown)


opentelemetry-http/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## vNext
 
+- Return HTTP error responses from the built-in reqwest and hyper clients instead
+  of converting 4xx and 5xx statuses into transport errors. This preserves the
+  response status and headers for exporter retry classification.
```
scottgerring, line 7:
```suggestion
- Return HTTP error responses from the built-in reqwest and hyper clients instead
  of converting 4xx and 5xx statuses into transport errors. This preserves the
  response status and headers for exporter retry classification. 
  If your code relied on `send_bytes` returning `Err` for non-success
  statuses, call `ResponseExt::error_for_status()` on the response instead.
```

... seems unlikely, but in case anyone was doing this

scottgerring:
Good catch, and a bit disappointed we had this in there since the experimental features to be honest. I've spent some more time with this this morning, and added more test coverage focussing on the seam around the exporters in #3629

scottgerring (approved):

opentelemetry-http/src/lib.rs
```diff
@@ -309,6 +323,85 @@ mod tests {
...
+            let mut request = [0; 1024];
+            let _ = stream.read(&mut request).unwrap();
+            stream
+                .write_all(
+                    b"HTTP/1.1 429 Too Many Requests\r\n\
+                      Retry-After: 7\r\n\
+                      Content-Length: 0\r\n\
+                      Connection: close\r\n\r\n",
+                )
+                .unwrap();
```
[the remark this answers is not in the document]
cijothomas: Fixed in ed073706: the mock response headers now start at column zero, so no indentation whitespace is written on the wire. Re-ran the async reqwest, blocking reqwest, hyper, all-features tests, and clippy.

opentelemetry-http/CHANGELOG.md
cijothomas: Applied in ed073706. The changelog now directs callers that relied on non-success responses returning Err to call ResponseExt::error_for_status() explicitly.

PR #3628, opened 2026-08-13
fix(otlp): preserve sub-second gRPC retry delays

cijothomas:
## Changes

- preserve the complete gRPC `RetryInfo` duration instead of converting it to whole seconds
- retain the existing maximum server-delay cap
- update regression coverage to verify a 500 ms delay remains 500 ms

## Why

`Duration::as_secs()` truncates fractional seconds. A server-provided retry delay of 500 ms therefore became zero, causing the exporter to retry immediately instead of honoring the throttle hint.

## Validation

- `cargo test -p opentelemetry-otlp --all-features retry_classification::tests::grpc_tests`
- `cargo clippy -p opentelemetry-otlp --all-targets --all-features -- -Dwarnings`

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- Preserve sub-second gRPC `RetryInfo` delays instead of truncating them to
+  whole seconds.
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
LGTM

PR #3630, opened 2026-08-14
test(otlp): add real HTTP retry integration tests

cijothomas:
## Changes

- Add an in-process scripted OTLP/HTTP endpoint for integration tests.
- Exercise the real async reqwest exporter through local SDK providers.
- Emit traces, logs, and metrics using their public APIs.
- Cover:
  - \`503\` retry followed by success
  - \`400\` without retry
  - \`429 Retry-After\` throttling followed by success
- Assert request counts, retry timing, OTLP paths, content types, and decoded protobuf payloads.

## Why

Existing retry tests mock \`HttpClient\`, while the Collector integration tests exercise successful ingestion. This adds coverage across the complete user-facing path:

\`telemetry API → SDK provider → processor/reader → exporter → reqwest → HTTP response → retry classification → retry\`

This guards against transport-level behavior that mocked clients cannot detect, such as losing HTTP status codes or \`Retry-After\` headers.

## Scope

This intentionally starts with OTLP traces, logs, and metrics over HTTP/protobuf using the async reqwest client. Additional HTTP clients, gRPC, and failure scenarios can build on the fake endpoint in follow-up PRs.

### DIFF
--- a/opentelemetry-otlp/tests/integration_test/src/fake_otlp_http.rs
+++ b/opentelemetry-otlp/tests/integration_test/src/fake_otlp_http.rs
@@ -81,6 +81,8 @@ impl FakeOtlpHttpEndpoint {
 
                 let response = responses
                     .pop_front()
+                    // Fall back to 500 if more requests arrive than scripted responses;
+                    // this makes unexpected extra requests fail visibly in assertions.
                     .unwrap_or_else(|| ScriptedHttpResponse::new(500));
                 if let Err(error) = write_response(&mut stream, response).await {
                     server_errors.lock().unwrap().push(error.to_string());
--- a/opentelemetry-otlp/tests/integration_test/src/fake_otlp_http.rs
+++ b/opentelemetry-otlp/tests/integration_test/src/fake_otlp_http.rs
@@ -1,0 +1,231 @@
+use std::collections::{HashMap, VecDeque};
+use std::io;
+use std::net::SocketAddr;
+use std::sync::{Arc, Mutex};
+use std::time::Instant;
+
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::{TcpListener, TcpStream};
+use tokio::task::JoinHandle;
+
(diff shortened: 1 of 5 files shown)


opentelemetry-otlp/tests/integration_test/src/fake_otlp_http.rs
```diff
@@ -0,0 +1,231 @@
...
+                    Err(error) => {
+                        server_errors.lock().unwrap().push(error.to_string());
+                        continue;
+                    }
+                };
+                captured_requests.lock().unwrap().push(request);
+
+                let response = responses
+                    .pop_front()
+                    .unwrap_or_else(|| ScriptedHttpResponse::new(500));
```
scottgerring, line 86: so it's easier to follow for future readers:

   ```suggestion
                       // Fall back to 500 if more requests arrive than scripted responses;
                       // this makes unexpected extra requests fail visibly in assertions.
                       .unwrap_or_else(|| ScriptedHttpResponse::new(500));
   ```

opentelemetry-otlp/tests/integration_test/src/fake_otlp_http.rs
```diff
@@ -0,0 +1,231 @@
...
+#[derive(Clone, Debug)]
+pub struct CapturedHttpRequest {
+    pub request_line: String,
+    pub headers: HashMap<String, String>,
+    pub body: Vec<u8>,
+    pub received_at: Instant,
+}
+
+#[derive(Debug)]
+pub struct FakeOtlpHttpEndpoint {
```
scottgerring, line 42: I think  [wiremock](https://lib.rs/crates/wiremock) could make things a bit easier here

cijothomas: I took a brief look at it. For now, the requirements are simple enough and we can precisly control timestamps to check if retry-delay is honored easrily. If this outgrows our simple setup, lets re-consider.

scottgerring (approved):
🚢

PR #3632, opened 2026-08-18
fix(otlp): align HTTP retry classification with spec

cijothomas:
## Summary

Align OTLP/HTTP retry classification with the OTLP specification and honor server throttling instructions on `503 Service Unavailable` responses.

## Before

- `429 Too Many Requests` was retried and honored `Retry-After`.
- Every `5xx` response, including `500`, `501`, and `505`-`599`, was retried.
- `503` was retried but its `Retry-After` header was ignored in favor of normal exponential backoff.
- Transport and network failures were retried.

## After

- Only the OTLP-defined retryable response codes are retried: `429`, `502`, `503`, and `504`.
- Both `429` and `503` honor a valid `Retry-After` header.
- Other HTTP error responses, including `500`, are treated as non-retryable.
- Transport and network failures remain retryable.

The change adds classifier coverage for the complete status-code behavior and exporter-level coverage proving that `500` is not retried and `503 Retry-After` delays the next attempt.

### DIFF
--- a/opentelemetry-otlp/src/retry_classification.rs
+++ b/opentelemetry-otlp/src/retry_classification.rs
@@ -31,22 +31,21 @@ pub mod http {
         retry_after_header: Option<&str>,
     ) -> RetryErrorType {
         match status_code {
-            // 429 Too Many Requests - check for Retry-After
-            429 => {
+            // OTLP/HTTP throttling responses may provide an explicit retry delay.
+            429 | 503 => {
                 if let Some(retry_after) = retry_after_header {
                     if let Some(duration) = parse_retry_after(retry_after) {
                         return RetryErrorType::Throttled(duration);
                     }
                 }
-                // Fallback to retryable if no valid Retry-After
                 RetryErrorType::Retryable
             }
-            // 5xx Server errors - retryable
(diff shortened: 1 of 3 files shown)


opentelemetry-otlp/src/retry_classification.rs
```diff
@@ -171,90 +170,103 @@ mod tests {
         use std::time::Duration;
 
         #[test]
-        fn test_http_429_with_retry_after_seconds() {
-            let result = classify_http_error(429, Some("30"));
-            assert_eq!(result, RetryErrorType::Throttled(Duration::from_secs(30)));
+        fn test_http_throttling_responses_with_retry_after_seconds() {
+            for status_code in [429, 503] {
```
scottgerring, line 174: nit: we could use `rstest` here - we've got this in opentelemetry-sdk already. So we'd have something like ...

```rust
#[rstest]
#[case(429)]
#[case(503)]
fn test_http_thorrling_responses_with_retry_after_seconds(#[case] status_code: u16) { 
```

scottgerring (approved):
Good catch - this is indeed in the spec! I've not known `Retry-After`'s to be a 503 thing, but it does seem like it can happen and it is easy enough to supprot

PR #3633, opened 2026-08-18
feat(otlp): limit HTTP request body size

scottgerring:
## Summary

Implements the [OTLP/HTTP request body size requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request). This is a `SHOULD`, but it is straightforward enough so let's do it in pursuit of our OTLP stabilisation. There is a similar requirement for gRPC but we will look at that separately. 

OTLP/HTTP request bodies are now limited to 64 MiB by default. The limit is checked before compression and, when compression is enabled, against the compressed body as well - this might seem academic, but compression can in some cases make payloads _larger_, so we may as well be thorough. Users can configure it with `WithHttpConfig::with_max_request_body_size`.

## Oversized request behavior

An oversized request causes the export operation to fail before any HTTP request is made. It is not sent or retried, and the error explains that the telemetry was discarded and suggests reducing the batch size or telemetry item size, or configuring a larger limit.

This hard failure is intentional. The OTLP specification says that, when the limit is exceeded, the client **MUST NOT** make the request and **SHOULD** record that it was discarded. The exporter does not attempt to split, truncate, or re-batch the request. Re-batching at this layer would add complex signal-specific behavior and is not required by the specification. Users should instead adjust their telemetry processor or batch configuration, reduce unusually large telemetry items, or set a limit compatible with their collector.

## Changes

- Add a 64 MiB default OTLP/HTTP request body limit.
- Validate the serialized body before compression and the resulting body after compression.
- Add `WithHttpConfig::with_max_request_body_size` for programmatic configuration.
- Ensure oversized requests do not enter the HTTP send or retry path.
- Preserve actionable serialization, compression, and size-limit errors for metrics exports.
- Document the behavior and public API in the crate documentation and changelog.

## Public API

Adding `WithHttpConfig::with_max_request_body_size` is a breaking change for external implementations of the public `WithHttpConfig` trait. Those implementations must add the new required method.

- [x] Unit tests added/updated
- [x] Appropriate `CHANGELOG.md` updated
- [x] Changes in public API reviewed

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- **Breaking** Add the required `WithHttpConfig::with_max_request_body_size`
+  method. External implementations of `WithHttpConfig` must implement it.
+  OTLP/HTTP request bodies are now limited to 64 MiB by default, before and
+  after compression; oversized requests are discarded without being sent or
+  retried.
 - Fix OTLP/HTTP retry classification to retry only status codes 429, 502, 503,
   and 504, as required by the OTLP specification. The exporter now also honors
   `Retry-After` on 503 responses. Other HTTP error responses, including 500,
--- a/opentelemetry-otlp/src/exporter/http/metrics.rs
+++ b/opentelemetry-otlp/src/exporter/http/metrics.rs
@@ -10,14 +10,12 @@ use super::OtlpHttpClient;
 
 impl MetricsClient for OtlpHttpClient {
     async fn export(&self, metrics: &ResourceMetrics) -> OTelSdkResult {
-        let build_body_wrapper = |client: &OtlpHttpClient, metrics: &ResourceMetrics| {
-            client
-                .build_metrics_export_body(metrics)
-                .ok_or_else(|| "Failed to serialize metrics".to_string())
-        };
-
         let response_body = self
-            .export_http_with_retry(metrics, build_body_wrapper, "HttpMetricsClient.Export")
+            .export_http_with_retry(
+                metrics,
+                OtlpHttpClient::build_metrics_export_body,
+                "HttpMetricsClient.Export",
+            )
             .await?;
 
         handle_partial_success(&response_body, self.protocol);
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -28,6 +28,10 @@ use std::time::Duration;
 use crate::retry::{RetryErrorType, RetryPolicy};
 use crate::retry_classification::http::classify_http_error;
 
+// Recommended by the OTLP/HTTP specification:
+// https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-request
+const DEFAULT_MAX_REQUEST_BODY_SIZE: usize = 64 * 1024 * 1024;
+
 // Shared HTTP retry functionality
 /// HTTP-specific error wrapper for retry classification
 #[derive(Debug)]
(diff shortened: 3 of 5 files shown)


opentelemetry-otlp/src/exporter/http/metrics.rs
```diff
@@ -10,14 +10,12 @@ use super::OtlpHttpClient;
...
     async fn export(&self, metrics: &ResourceMetrics) -> OTelSdkResult {
-        let build_body_wrapper = |client: &OtlpHttpClient, metrics: &ResourceMetrics| {
-            client
-                .build_metrics_export_body(metrics)
-                .ok_or_else(|| "Failed to serialize metrics".to_string())
-        };
-
         let response_body = self
-            .export_http_with_retry(metrics, build_body_wrapper, "HttpMetricsClient.Export")
+            .export_http_with_retry(
```
scottgerring, line 14: This removes the Option wrapper because request-size validation now happens while building the body. Returning Result preserves the specific oversized-body or compression error instead of reporting the misleading generic “Failed to serialize metrics” message. It also aligns metrics with the trace and log body builders.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -617,7 +656,7 @@ impl OtlpHttpClient {
     fn build_metrics_export_body(
         &self,
         metrics: &ResourceMetrics,
-    ) -> Option<(Vec<u8>, &'static str, Option<&'static str>)> {
+    ) -> Result<(Vec<u8>, &'static str, Option<&'static str>), String> {
```
scottgerring, line 663: This changes metrics body construction from Option to Result so serialization, compression, and request-size failures retain their actionable error message instead of becoming the generic “Failed to serialize metrics” error. Traces and logs already use this shape

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -275,15 +282,19 @@ impl HttpExporterBuilder {
...
             protocol,
             timeout,
             compression,
             self.http_config.retry_policy.take(),
-        ))
+        );
+        if let Some(max_request_body_size) = self.http_config.max_request_body_size {
+            client.max_request_body_size = max_request_body_size;
+        }
+        Ok(client)
```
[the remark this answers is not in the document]
scottgerring: This is OTT. Pattern is fine.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -518,23 +532,47 @@ impl OtlpHttpClient {
...
+    }
+
+    fn validate_request_body_size(
+        &self,
+        body: &[u8],
+        representation: &'static str,
+    ) -> Result<(), String> {
+        if body.len() > self.max_request_body_size {
+            return Err(format!(
+                "OTLP HTTP {representation} request body is {} bytes, exceeding the configured limit of {} bytes; request will not be sent and telemetry data will be discarded; reduce the batch size, telemetry item size, or configure a larger limit",
```
cijothomas, line 571: ❤️ This is well-written - clear and actionable to user. Could we use the same for metrics path too, if not already done?

scottgerring: Cheers! It does already - `build_metrics_export_body` calls `process_body`, which in turn calls `validate_request_body_size` 

Just to be clear, this is just for HTTP; the gRPC side of this for all signals we can tackle in another PR

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## vNext
 
+- **Breaking** Add the required `WithHttpConfig::with_max_request_body_size`
```
cijothomas, line 5: suggested rephrase to avoid giving heart attack to end users when they keep seeing more breaking changes 🤣 

- Limit OTLP/HTTP request bodies to 64 MiB by default, enforced before and after
  compression. Oversized requests are discarded without being sent or retried.
  Configurable via `WithHttpConfig::with_max_request_body_size`.
  **Breaking only for custom `WithHttpConfig` implementations:** the new method
  is required and external impls must add it. Code calling the exporter builders
  is unaffected.

cijothomas (approved):

PR #3634, opened 2026-08-18
fix(otlp): honor gRPC RetryInfo for unavailable

cijothomas:
## Summary

Honor server-provided gRPC `RetryInfo` delays on OTLP `Unavailable` responses, as specified by the OTLP throttling protocol.

## Before

- `Unavailable` was classified as retryable regardless of attached error details.
- A server-provided `RetryInfo.retry_delay` was ignored.
- Retries therefore used the exporter's normal exponential-backoff delay instead of the server-requested delay.

## After

- `Unavailable` with a positive `RetryInfo.retry_delay` is classified as throttled and honors the server delay.
- `Unavailable` without `RetryInfo`, or with a zero delay, continues to use normal exponential backoff.
- Fractional delays are preserved and large delays retain the existing ten-minute classifier cap.
- Existing `ResourceExhausted` classification remains unchanged.

The change adds focused coverage for positive, fractional, capped, missing, and zero-delay `RetryInfo` behavior.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -7,6 +7,9 @@
   OTLP/HTTP request bodies are now limited to 64 MiB by default, before and
   after compression; oversized requests are discarded without being sent or
   retried.
+- Honor positive gRPC `RetryInfo` delays returned with `Unavailable` responses.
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
LGTM

Issue #2757, opened 2025-03-05
Review, test and ensure TraceState,Baggage parsing is defensive


cijothomas:
OpenTelemetry instrumentations rely on Propagators to parse incoming headers and extract Context (TraceContext, Baggage). Opening an issue to make sure these paths are sufficiently covered by tests.

Otel .NET published this advisory earlier today, where malformed `tracestate` header could be used by a bad actor and cause service to go down.
https://github.com/open-telemetry/opentelemetry-dotnet/security/advisories/GHSA-8785-wc3w-h8q6

PR #3635, opened 2026-08-18
fix: cap tracestate at the 32 list-members the spec allows

serhiy-bzhezytskyy:
Part of #2757

## Changes

The W3C spec allows "a maximum of 32 list-members in a list" (section 3.3.1.1). This crate enforced that
nowhere. `from_str` kept every comma-separated member of an inbound `tracestate`, and `from_key_value` and
`insert` would build a larger one locally: 64 inserts produced an 875-character header on inject, past the 512
characters section 3.3.1.5 asks vendors to propagate at minimum, and a parser holding to the limit dropped half
of those members at the next hop.

All three entry points now keep at most 32 members and drop from the end of the list, which is the direction
section 3.3.1.5 prescribes for truncation. `insert` keeps the new member and evicts the oldest, as
`opentelemetry-go`'s `Insert` does. Parsing stops at the 32nd member, so an oversized inbound
header is neither scanned nor allocated past the limit.

Two behaviour changes for callers. A `tracestate` with more than 32 members now yields the first 32 instead of
all of them, and a malformed member past the 32nd no longer fails the whole header, because
`TraceContextPropagator::extract` maps an `Err` to `TraceState::default()` and erroring would discard 32 legal
members because of a 33rd. A malformed member within the limit still returns `Err`, unchanged. Building past
the limit no longer grows the state: `from_key_value` takes the first 32 pairs, so a pair past the 32nd is not
validated either, and `insert` evicts.

Six tests, three per direction. Parsing: the cap, an invalid member beyond the limit being ignored, an invalid
member within the limit still failing. Building: `from_key_value` keeping the first 32, `insert` keeping the
newest 32, `insert` of an existing key evicting nothing. Reverting a direction's cap fails its first two; the
third guards behaviour the change must not alter.

One caller changes behaviour with it: `opentelemetry-jaeger-propagator` builds a `TraceState` from every
`uberctx-` carrier entry through `from_key_value`, so a carrier with 200 of them now yields 32 members instead
of 200. That crate is deprecated, but the change is user-visible, so its CHANGELOG has a line too.

`insert` could have returned an `Err` at capacity instead, since it already returns a `Result`. I did not take
that route: the spec puts the newest member at the front and truncates from the end, so refusing the caller's
own entry is the one outcome it argues against, and none of Go, C++, Python or Ruby errors here either.
`opentelemetry-js` is adding its mutation half in #6964. `baggage.rs` in this crate already caps entries and
total length.

Validated locally with the full `scripts/precommit.sh` that AGENTS.md requires for changes under
`opentelemetry/src/`, and with `cargo semver-checks check-release -p opentelemetry`, which reports no semver
update required.

Assisted-By: Claude Fable 5

### DIFF
--- a/opentelemetry/src/trace/span_context.rs
+++ b/opentelemetry/src/trace/span_context.rs
@@ -14,6 +14,8 @@ use thiserror::Error;
 #[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
 pub struct TraceState(Option<VecDeque<(String, String)>>);
 
+const MAX_LIST_MEMBERS: usize = 32;
+
 impl TraceState {
     /// The default `TraceState`, as a constant
     pub const NONE: TraceState = TraceState(None);
@@ -61,7 +63,10 @@ impl TraceState {
         !(value.contains(',') || value.contains('='))
     }
 
-    /// Creates a new `TraceState` from the given key-value collection.
+    /// Creates a new `TraceState` from the given key-value collection, keeping at most the 32
+    /// list-members the [W3C specification] allows.
+    ///
+    /// [W3C specification]: https://www.w3.org/TR/trace-context/#tracestate-header-field-values
     ///
     /// # Examples
     ///
@@ -82,6 +87,7 @@ impl TraceState {
     {
         let ordered_data = trace_state
             .into_iter()
+            .take(MAX_LIST_MEMBERS)
             .map(|(key, value)| {
                 let (key, value) = (key.to_string(), value.to_string());
                 if !TraceState::valid_key(key.as_str()) {
@@ -120,6 +126,9 @@ impl TraceState {
     /// invalid per the [W3 Spec] an `Err` is returned, else a new `TraceState` with the
     /// updated key/value is returned.
     ///
+    /// A `TraceState` holds at most the 32 list-members the [W3 Spec] allows. Inserting into a full
+    /// `TraceState` keeps the inserted pair and drops the last list-member.
+    ///
     /// [W3 Spec]: https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field
     pub fn insert<K, V>(&self, key: K, value: V) -> TraceStateResult<TraceState>
     where
@@ -136,6 +145,9 @@ impl TraceState {
 
         let mut trace_state = self.delete_from_deque(&key);
         let kvs = trace_state.0.get_or_insert(VecDeque::with_capacity(1));
+        if kvs.len() >= MAX_LIST_MEMBERS {
+            kvs.truncate(MAX_LIST_MEMBERS - 1);
+        }
 
         kvs.push_front((key, value));
 
@@ -193,10 +205,9 @@ impl FromStr for TraceState {
     type Err = TraceStateError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let list_members: Vec<&str> = s.split_terminator(',').collect();
-        let mut key_value_pairs: Vec<(String, String)> = Vec::with_capacity(list_members.len());
+        let mut key_value_pairs: Vec<(String, String)> = Vec::new();
 
-        for list_member in list_members {
+        for list_member in s.split_terminator(',').take(MAX_LIST_MEMBERS) {
             match list_member.find('=') {
                 None => return Err(TraceStateError::List(list_member.to_string())),
                 Some(separator_index) => {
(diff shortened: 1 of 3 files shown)


opentelemetry/src/trace/span_context.rs
```diff
@@ -82,6 +87,7 @@ impl TraceState {
     {
         let ordered_data = trace_state
             .into_iter()
+            .take(MAX_LIST_MEMBERS)
```
lalitb, line 92: Could we mention in the `from_key_value` rustdoc that pairs after the first 32 are ignored without being validated? For example, an invalid 33rd pair previously returned `Err`, but now returns a valid `TraceState` containing the first 32. That behavior is easy to miss from “keeping at most 32” alone.

opentelemetry/src/trace/span_context.rs
```diff
@@ -61,7 +63,10 @@ impl TraceState {
         !(value.contains(',') || value.contains('='))
     }
 
-    /// Creates a new `TraceState` from the given key-value collection.
+    /// Creates a new `TraceState` from the given key-value collection, keeping at most the 32
+    /// list-members the [W3C specification] allows.
+    ///
+    /// [W3C specification]: https://www.w3.org/TR/trace-context/#tracestate-header-field-values
```
lalitb: nit - The OTel propagator [specification](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#w3c-trace-context-requirements) requires W3C Trace Context Level 2. Could this new link point to https://www.w3.org/TR/trace-context-2/#tracestate-header? The 32-member rule is unchanged, but this would direct readers to the version OTel currently follows.

opentelemetry/src/trace/span_context.rs
```diff
@@ -496,4 +507,72 @@ mod tests {
...
         assert_eq!(iter.len(), 2);
     }
+
+    #[test]
+    fn test_tracestate_from_str_keeps_at_most_32_list_members() {
+        let header = (0..64)
+            .map(|i| format!("key{i}=value{i}"))
+            .collect::<Vec<_>>()
+            .join(",");
+        let trace_state = TraceState::from_str(&header).unwrap();
```
lalitb, line 519: Could this also assert the generated header() contains only the retained members? The motivating regression was the oversized value emitted during injection, so checking the serialized result would cover that user-visible behavior directly.

lalitb (approved):
LGTM. Left a few non-blocking suggestions around docs and test coverage.

serhiy-bzhezytskyy:
Thanks — pushed a commit for all three. Linked docs and error messages to Trace Context Level 2 (which OTel's propagator spec requires, and which states the cap more precisely — the right-most member is dropped on overflow, matching `insert()`'s eviction order). `from_key_value`'s doc now says pairs past the 32nd are dropped without validation. `header()` is asserted in the `from_str` cap test.

TommyCpp (approved):

opentelemetry/src/trace/span_context.rs
```diff
@@ -193,10 +207,9 @@ impl FromStr for TraceState {
     type Err = TraceStateError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let list_members: Vec<&str> = s.split_terminator(',').collect();
-        let mut key_value_pairs: Vec<(String, String)> = Vec::with_capacity(list_members.len());
+        let mut key_value_pairs: Vec<(String, String)> = Vec::new();
 
-        for list_member in list_members {
+        for list_member in s.split_terminator(',').take(MAX_LIST_MEMBERS) {
```
cijothomas, line 212: if incoming triacetate from >32 members, shouldn't we reject the entire trace-state, instead of trying to take the 1st 32 members?

serhiy-bzhezytskyy: The spec explicitly prescribes truncation for this case (W3C Trace Context Level 2, §tracestate-header-field-values): "If adding an entry would cause the tracestate list to contain more than 32 list-members the right-most list-member should be removed from the list." That's about count only — nothing in the tracestate section requires discarding the whole header the way Baggage's "MUST NOT propagate any partial list-members" does (that MUST is Baggage-specific).

Precedent is mixed, not uniform: opentelemetry-go's ParseTraceState rejects the whole tracestate on >32 (returns an error the caller discards to an empty TraceState — trace/tracestate.go:212-214). opentelemetry-js truncates to the first 32, same as here (tracestate-impl.ts:86-92).

I went with truncation to avoid throwing away otherwise-valid vendor data over a chattier-than-expected upstream, and it matches JS's precedent. Happy to switch to reject-the-whole-header if you'd rather match Go's stricter reading — it's a small change either way.

serhiy-bzhezytskyy: Leaving the truncation, on the spec's tracestate wording and the js precedent above.

cijothomas (approved):

PR #3637, opened 2026-08-19
fix(otlp): back off repeated throttled retries

scottgerring:
## Summary

After a server-provided `RetryInfo` or `Retry-After` delay, subsequent export failures now continue exponential backoff from the server-requested delay rather than reusing a fixed value or dropping back to the policy initial.

## Before

- Repeated throttling reused the same server delay indefinitely.
- A retryable error following a throttle fell back to the 100ms policy initial.
- No jitter was applied to throttled retries.

## After

- Server delay seeds the exponential backoff base via `delay.max(server_delay)`.
- Repeated or mixed failures double from there, capped by the effective maximum.
- Jitter applies uniformly to all retry sleeps.
- A server delay never reduces an already-grown backoff.
- The overall exporter timeout remains the final bound.

Aligns with the `RetryInfo` guidance:
> If retrying requests also fail, clients should use an exponential backoff
> scheme to increase gradually the delay between retries based on `retry_delay`.

## Cross-SDK comparison

As the exact mechanism is somewhat underspecified in the spec, i've had a look at the golang and dotnet SDKs for reference. All three use the the server-provided delay as a minimum, and subsequent failures back off from there.

| Aspect | Go (`cenkalti/backoff`) | .NET | This PR |
|---|---|---|---|
| Throttle: current sleep | `max(backoff, throttle)` | Exact server delay | `max(backoff, throttle)` + jitter |
| Throttle: effect on next retry | None (backoff state independent) | `next_base = throttle * 1.5` | `delay = max(delay, throttle); delay *= 2` |
| Server delay < current backoff | Backoff wins (independent sequence) | Sleeps shorter server delay | Backoff wins (never reduces) |
| Multiplier | 1.5 | 1.5 | 2 |
| Jitter on throttled retries | Yes (built into backoff lib) | No (exact server delay) | Yes (additive) |

Go keeps its exponential backoff sequence independent of throttle events and uses `max` per attempt. .NET seeds the next backoff base from the throttle but sleeps the exact server delay on the current attempt (can go below current
backoff). This PR merges the throttle into the backoff state and never reduces it, which is more conservative but equally spec-compliant since the spec requires waiting "at least `retry_delay`."

Relates to #3061

### DIFF
--- a/opentelemetry-otlp/src/retry.rs
+++ b/opentelemetry-otlp/src/retry.rs
@@ -148,7 +148,12 @@ where
 {
     let start = Instant::now();
     let mut attempt = 0;
-    let mut delay = policy.initial_delay_ms;
+    let mut delay = Duration::from_millis(policy.initial_delay_ms);
+    // The delay cap may be raised beyond the configured max if a server-provided
+    // throttle delay exceeds it, since retrying sooner than the server requested
+    // would defeat the purpose of backpressure signalling. The overall exporter
+    // timeout remains the final bound.
+    let mut delay_cap = Duration::from_millis(policy.max_delay_ms);
 
     loop {
         match operation().await {
@@ -178,9 +183,8 @@ where
                     }
                     RetryErrorType::Retryable if attempt < policy.max_retries => {
                         attempt += 1;
-                        let jitter = generate_jitter(policy.jitter_ms);
-                        let delay_ms = std::cmp::min(delay + jitter, policy.max_delay_ms);
-                        let sleep_duration = Duration::from_millis(delay_ms);
+                        let jitter = Duration::from_millis(generate_jitter(policy.jitter_ms));
+                        let sleep_duration = delay.saturating_add(jitter).min(delay_cap);
 
                         // If the backoff delay would consume all remaining budget, fail now
                         let remaining = deadline.saturating_sub(start.elapsed());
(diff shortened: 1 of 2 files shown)


opentelemetry-otlp/src/retry.rs
```diff
@@ -416,9 +429,130 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_throttle_after_normal_backoff_does_not_reduce_delay() {
+        let policy = RetryPolicy {
+            max_retries: 3,
+            initial_delay_ms: 50,
+            max_delay_ms: 500,
+            jitter_ms: 0,
+        };
+        let attempts = AtomicUsize::new(0);
+        let start = Instant::now();
```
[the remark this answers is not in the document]
scottgerring: OTT; leaving alone.

opentelemetry-otlp/src/retry.rs
```diff
@@ -416,9 +429,130 @@ mod tests {
         let elapsed = start.elapsed();
         assert_eq!(result, Ok("success"));
         assert_eq!(attempts.load(Ordering::SeqCst), 2);
-        // Should have waited ~50ms (the server delay), not 1000ms (the default)
+        // Should have waited ~50ms (the server delay raises it from the 10ms default)
         assert!(elapsed >= Duration::from_millis(50));
-        assert!(elapsed < Duration::from_millis(500));
+        assert!(elapsed < Duration::from_millis(200));
```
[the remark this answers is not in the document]
scottgerring: OTT; leaving alone.

opentelemetry-otlp/src/retry.rs
```diff
@@ -200,16 +204,25 @@ where
...
                     }
                     RetryErrorType::Throttled(server_delay) if attempt < policy.max_retries => {
                         attempt += 1;
-                        // Cap server-provided delay to prevent excessive blocking
-                        let capped_delay = server_delay.min(MAX_THROTTLE_DURATION);
+                        let capped_server_delay = server_delay.min(MAX_THROTTLE_DURATION);
+
+                        // RetryInfo is the base for subsequent exponential backoff. A later
+                        // RetryInfo may extend, but must not shorten, the current backoff.
+                        delay_cap = delay_cap.max(capped_server_delay);
```
lalitb: This seems incorrect to me, as it does not increase the backoff when the server delay is higher than `max_delay_ms`.

For example, the recommended maximum is 1.6s. If the server asks us to wait 5s, `delay_cap` becomes 5s. After the first sleep we double the delay, but immediately cap it back to 5s. The retry sequence is therefore 5s, 5s, 5s instead of 5s, 10s, 20s. Jitter is also lost because it is capped back to 5s.

When a server delay raises the configured cap, can we allow the backoff to grow up to `MAX_THROTTLE_DURATION`? Please also add a test where the server delay is greater than `max_delay_ms`, including a normal retryable error after the throttle.

opentelemetry-otlp/src/retry.rs
lalitb: >     /// This overrides exponential backoff timing.


nit -   This comment is no longer correct. The server delay now becomes the starting point for exponential backoff instead of overriding the backoff for one retry.

Can we update this comment to explain the new behavior? Please also check the `retry_with_backoff` documentation in this file so both places describe the same policy.

opentelemetry-otlp/src/retry.rs
scottgerring: Good catch! I've spent too long staring at this it seems. Pushed a fix, documented, captured in tests; we had one that was close but didn't cover this edge case.

scottgerring:
@lalitb addressed feedback - thanks for looking!

TommyCpp (approved):

Issue #3427, opened 2026-03-18
[Bug]: No programmatic API to set a base endpoint with automatic signal path appending for HTTP exporters


aswiniverse:
## Problem

The Rust OTLP HTTP exporters have no programmatic way to provide a **base URL** and have the SDK automatically append the signal path (`/v1/logs`, `/v1/traces`, `/v1/metrics`).

`.with_endpoint()` treats its argument as a full URL and uses it verbatim — there is no programmatic equivalent of `OTEL_EXPORTER_OTLP_ENDPOINT` (which does append the signal path).

## Comparison with Go SDK

The Go SDK distinguishes between two use cases via two separate methods:

```go
// Base URL → SDK appends /v1/logs automatically ✅
otlploghttp.WithEndpoint("localhost:4318")

// Full URL → used as-is ✅
otlploghttp.WithEndpointURL("http://localhost:4318/v1/logs")
```

The Rust SDK only has one method:

```rust
// Always treated as full URL → signal path never appended ✗
.with_endpoint("http://localhost:4318")
```

There is no Rust equivalent of Go's `WithEndpoint(host:port)`.

## Root cause in `exporter/http/mod.rs`

`resolve_http_endpoint` (L705–735) has three code paths:

```rust
// 1. Programmatic .with_endpoint() → returned as-is, no path appended ✗
if let Some(provider_endpoint) = provided_endpoint.filter(|s| !s.is_empty()) {
    return provider_endpoint.parse()
}

// 2. Signal-specific env var (e.g. OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) → returned as-is ✗
// (comment: "per signal env var is not modified")

// 3. Generic OTEL_EXPORTER_OTLP_ENDPOINT → /v1/logs appended ✓
build_endpoint_uri(&s, signal_endpoint_path)
```

Signal path appending is only reachable via the generic env var. There is no programmatic equivalent.

## Impact

Applications that configure the OTLP endpoint programmatically (e.g. reading from a config file) cannot provide a base URL — they must either:
- Hardcode the full signal path (e.g. `http://localhost:4318/v1/logs`), or
- Fall back to setting `OTEL_EXPORTER_OTLP_ENDPOINT` as an environment variable at runtime

This is a real footgun for library consumers such as [Codex CLI](https://github.com/openai/codex), which reads an endpoint from user config and calls `.with_endpoint()` — requiring users to manually include `/v1/logs` in their config.

## Suggested fix

[...]

PR #3646, opened 2026-08-25
docs: fix OTLP HTTP with_endpoint example (endpoint used verbatim)

dcdeniz:
Towards #3427

## Changes

The HTTP transport example in the `opentelemetry-otlp` crate-level docs stated that `.with_endpoint()` appends the signal path (`/v1/traces`, etc.) automatically. It doesn't: `resolve_http_endpoint` uses the provided endpoint verbatim, and only the generic `OTEL_EXPORTER_OTLP_ENDPOINT` env var appends the signal path (confirmed by the crate's own tests, e.g. `.with_endpoint(".../v1/tracesbutnotreally")` staying verbatim).

This corrects the doc comment and the example so users include the signal path themselves, and points to `OTEL_EXPORTER_OTLP_ENDPOINT` for automatic appending. The gRPC example and the env-var table were already correct and are untouched.

Docs-only change; no code or public API changes.

* [ ] Unit tests added/updated (if applicable) — N/A (docs only)
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes — N/A (docs only)
* [ ] Changes in public API reviewed (if applicable) — N/A

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -486,9 +486,10 @@
 //!
 //! let exporter = opentelemetry_otlp::SpanExporter::builder()
 //!     .with_http()
-//!     // Target URL, used verbatim: the signal path is NOT appended, so include it
-//!     // yourself (e.g. /v1/traces). Defaults to http://localhost:4318/v1/traces.
-//!     // Set OTEL_EXPORTER_OTLP_ENDPOINT to have the signal path appended automatically.
+//!     // Target URL, used verbatim: include the signal path, e.g. /v1/traces.
+//!     // OTEL_EXPORTER_OTLP_TRACES_ENDPOINT is also used verbatim.
+//!     // OTEL_EXPORTER_OTLP_ENDPOINT is a base URL and appends /v1/traces automatically.
+//!     // Defaults to http://localhost:4318/v1/traces.
 //!     .with_endpoint("http://my-collector:4318/v1/traces")
 //!     // Per-export timeout. Defaults to 10 s.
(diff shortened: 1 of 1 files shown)


scottgerring:
Hey @dcdeniz , thanks for contributing!
Could you  take a look at the CLA and consider signing it if you are happy with it so we can have a look at the PR?

dcdeniz:
Hey @scottgerring, just signed it. Thanks for reminding me

Made this PR as my first intro into this repo, so I thought some routine hygiene would be a good kick-off point

cijothomas:
The updated example correctly documents that `.with_endpoint()` is used verbatim, but it removes the nearby mention of `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, which has the same important behavior. Could we retain that distinction explicitly? For example:

```rust
// Target URL, used verbatim: include the signal path, e.g. /v1/traces.
// OTEL_EXPORTER_OTLP_TRACES_ENDPOINT is also used verbatim.
// OTEL_EXPORTER_OTLP_ENDPOINT is a base URL and appends /v1/traces automatically.
// Defaults to http://localhost:4318/v1/traces.
.with_endpoint("http://my-collector:4318/v1/traces")
```

dcdeniz:
Hey @cijothomas, makes sense thanks for the advice, still pretty new here. pushed an update that restores the `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` distinction so all three cases are explicit now: verbatim `with_endpoint`, verbatim signal env var, and the base-url `OTEL_EXPORTER_OTLP_ENDPOINT` that appends. lmk if you want the wording tweaked

cijothomas (approved):

PR #3647, opened 2026-08-25
feat(sdk): publicly export OTEL_* env var and default constants

### LINKED ISSUE #3623: [Feature]: Pulicly export `OTEL_*_DEFAULT` constants
### Related Problems?

In [Arti](https://arti.torproject.org/), we would like to inherit various default configuration values from OpenTelemetry, which still allowing the user to configure them via our configuration system. This is difficult, as we don't have any programmatic access to these defaults.

Make `OTEL_*_DEFAULT` constants `pub`, rather than `pub(crate)`

### Considered Alternatives

_No response_

### Additional Context

_No response_

### Tip

darkraider01:
Constants like `OTEL_BSP_SCHEDULE_DELAY_DEFAULT`, `OTEL_BLRP_MAX_QUEUE_SIZE_DEFAULT`, etc. were `pub(crate)`, so there was no way to read the SDK's own defaults from outside the crate. Arti (Tor) hit this directly - their config system needs to fall back to OTel's defaults when a user leaves a value unset, and right now that means hardcoding a second copy of every number instead of reading it from us.

This promotes them to `pub` and re-exports under each signal's module:

- `opentelemetry_sdk::trace` - `OTEL_BSP_*` / `OTEL_BSP_*_DEFAULT`
- `opentelemetry_sdk::logs` - `OTEL_BLRP_*` / `OTEL_BLRP_*_DEFAULT`
- `opentelemetry_sdk::metrics` - `OTEL_METRIC_EXPORT_INTERVAL{,_DEFAULT}`

Metrics needed a rename too - `DEFAULT_INTERVAL` and `METRIC_EXPORT_INTERVAL_NAME` didn't follow the `OTEL_*`/`OTEL_*_DEFAULT` pairing everything else uses, so I renamed them to `OTEL_METRIC_EXPORT_INTERVAL_DEFAULT` and `OTEL_METRIC_EXPORT_INTERVAL` per @scottgerring's comment on the issue.

Two things I went back and forth on and would rather ask than guess:

1. `OTEL_BSP_EXPORT_TIMEOUT` and `OTEL_BSP_MAX_CONCURRENT_EXPORTS` - the thread-based `BatchSpanProcessor` ignores both (there's already a test for this), only the experimental async-runtime one honors them. I exported them anyway since the name/default is still accurate info, and added a doc note saying which processor actually uses it. Open to gating them behind the same feature flag instead if that's cleaner.
2. `OTEL_BLRP_EXPORT_TIMEOUT`/`_DEFAULT` are still behind `experimental_logs_batch_log_processor_with_async_runtime` since that's where the existing `pub(crate)` constants already were - didn't want to widen availability beyond what was asked.

Also noticed `periodic_reader_with_async_runtime.rs` has its own private copy of the interval constants that's now out of sync naming-wise with the public ones. Left it alone since it's a separate file and wasn't part of the ask, but flagging it in case you'd rather fold it in here.

## Changes

- Made `OTEL_BSP_*`/`OTEL_BLRP_*`/`OTEL_METRIC_EXPORT_INTERVAL*` constants `pub` and re-exported them from `trace`, `logs`, and `metrics` module roots
- Renamed metrics' `DEFAULT_INTERVAL`/`METRIC_EXPORT_INTERVAL_NAME` to match the `OTEL_*_DEFAULT` naming convention
- Added a doc caveat on the two BSP constants the sync processor doesn't actually use
- Added an integration test (`opentelemetry-sdk/tests/otel_default_constants.rs`) so the re-exports are checked from outside the crate, not just internally
- CHANGELOG updated

Fixes #3623


opentelemetry-sdk/src/metrics/periodic_reader.rs
```diff
@@ -21,9 +21,12 @@ use super::{
     Temporality,
 };
 
-const DEFAULT_INTERVAL: Duration = Duration::from_secs(60);
+/// Default delay interval between two consecutive exports for [PeriodicReader].
+pub const OTEL_METRIC_EXPORT_INTERVAL_DEFAULT: Duration = Duration::from_secs(60);
```
scottgerring: extremely nitty, but could we swap this and `OTEL_METRIC_EXPORT_INTERVAL` around to follow the pattern

scottgerring:
Overall this looks great - clean, well-scoped, good tests and docs. A couple of thoughts on the timeout constants:

> `OTEL_BLRP_EXPORT_TIMEOUT`/`_DEFAULT` are still behind `experimental_logs_batch_log_processor_with_async_runtime` since that's where the existing `pub(crate)` constants already were — didn't want to widen availability beyond what was asked.

I'd reckon we should remove the feature gate here now as they're spec defined consts. They only ended up behind that feature because that's where the code that used them happened to live. But for the use case in #3623, gating a spec-defined constant behind an experimental feature means consumers need to opt into an unrelated implementation detail just to read a number. We could move them alongside the other BLRP constants  ungated and re-exported from `opentelemetry_sdk::logs`.

> `OTEL_BSP_EXPORT_TIMEOUT` and `OTEL_BSP_MAX_CONCURRENT_EXPORTS` - the thread-based `BatchSpanProcessor` ignores both (there's already a test for this), only the experimental async-runtime one honors them. I exported them anyway since the name/default is still accurate info, and added a doc note saying which processor actually uses it. Open to gating them behind the same feature flag instead if that's cleaner.

Exporting them unconditionally seems right, but the doc caveat on `OTEL_BSP_EXPORT_TIMEOUT` is just excusing a spec compliance gap - the processor should honour this. Do you mind opening a follow-up issue along the lines of "thread-based BatchSpanProcessor/BatchLogProcessor should honour export timeout"? The doc note is fine for now, but it would be nice to clean this up and probably not scope-creep this PR!

darkraider01:
Went to open the follow-up and turns out we already have one - #2576 "Enforce timeouts" covers this exact thing, down to already noting OTEL_BSP_EXPORT_TIMEOUT/OTEL_BLRP_EXPORT_TIMEOUT/OTEL_METRIC_EXPORT_TIMEOUT aren't honored by default. It also frames it better than I would have - the intended fix is on the exporter side, not the processor (tonic already enforces its own timeout, reqwest-blocking doesn't). Didn't want to file something that just duplicates it, so I linked this PR over there instead. Let me know if you'd still rather have a separate one scoped just to the two processors.

Also pushed the OTEL_BLRP_EXPORT_TIMEOUT change - ungated now, re-exported with the rest of the OTEL_BLRP_* constants, same doc caveat as OTEL_BSP_EXPORT_TIMEOUT.

scottgerring (approved):
LGTM, thanks @darkraider01 !

opentelemetry-sdk/src/trace/span_processor.rs
```diff
@@ -51,25 +51,46 @@ use std::sync::atomic::AtomicBool;
...
+/// See [`OTEL_BSP_EXPORT_TIMEOUT`] for which processors honor this value.
+pub const OTEL_BSP_EXPORT_TIMEOUT_DEFAULT: Duration = Duration::from_millis(30_000);
+/// Environment variable for configuring the maximum number of concurrent
+/// exports.
+///
+/// This value is honored by
+/// `span_processor_with_async_runtime::BatchSpanProcessor`, where it limits the
+/// number of concurrent export tasks. The thread-based [`BatchSpanProcessor`]
+/// exports serially and ignores this setting.
+pub const OTEL_BSP_MAX_CONCURRENT_EXPORTS: &str = "OTEL_BSP_MAX_CONCURRENT_EXPORTS";
```
cijothomas: I'd leave these one out for now. They are not spec supported, and there is work-in-progress in spec for making soething for it officially.
This is only added in Spans here, and also only work in the non-default processors.

darkraider01:
Good call, pushed - reverted OTEL_BSP_MAX_CONCURRENT_EXPORTS/_DEFAULT back to pub(crate). Left OTEL_BSP_EXPORT_TIMEOUT public since that one is spec defined, unlike concurrent-exports.

PR #3650, opened 2026-08-27
fix(otlp): reject invalid HTTP endpoint environment values

cijothomas:
## Summary

Fail OTLP/HTTP exporter construction when an explicitly configured endpoint environment variable is invalid, instead of silently falling through to another endpoint or localhost.

## Before

- An invalid signal-specific endpoint, such as `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`, was silently ignored.
- Resolution continued to `OTEL_EXPORTER_OTLP_ENDPOINT` or the default localhost endpoint.
- An invalid generic endpoint was also silently ignored in favor of localhost.
- Empty endpoint environment variables were not consistently treated as unset.

This could make exporter construction succeed while telemetry was sent to a destination different from the one the operator configured.

## After

- A present, non-empty signal-specific endpoint must parse successfully or exporter construction fails.
- A present, non-empty generic endpoint must produce a valid signal endpoint or exporter construction fails.
- Errors identify the offending environment variable and value.
- Empty endpoint environment variables are treated as unset, per the OpenTelemetry environment-variable specification.
- Programmatic configuration remains the highest-priority source.

The change includes coverage for signal-specific and generic invalid values, empty values, programmatic invalid endpoints, precedence, and non-Unicode environment values on Unix.

### DIFF
--- a/opentelemetry-otlp/src/exporter/http/mod.rs
+++ b/opentelemetry-otlp/src/exporter/http/mod.rs
@@ -733,6 +733,29 @@ fn build_endpoint_uri(endpoint: &str, path: &str) -> Result<Uri, ExporterBuildEr
     })
 }
 
+fn endpoint_from_env(variable: &str) -> Result<Option<String>, ExporterBuildError> {
+    match env::var(variable) {
+        Ok(value) if value.is_empty() => Ok(None),
+        Ok(value) => Ok(Some(value)),
+        Err(env::VarError::NotPresent) => Ok(None),
+        Err(env::VarError::NotUnicode(_)) => Err(ExporterBuildError::InvalidConfig {
+            name: variable.to_string(),
+            reason: "environment variable value is not valid Unicode".to_string(),
+        }),
+    }
+}
+
+fn invalid_endpoint_env(
+    variable: &str,
+    value: &str,
+    error: ExporterBuildError,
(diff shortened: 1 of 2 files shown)


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -746,18 +769,18 @@ fn resolve_http_endpoint(
...
-    } else if let Some(endpoint) = env::var(OTEL_EXPORTER_OTLP_ENDPOINT)
-        .ok()
-        .and_then(|s| build_endpoint_uri(&s, signal_endpoint_path).ok())
-    {
+        endpoint
+            .parse()
+            .map_err(|er: http::uri::InvalidUri| {
+                ExporterBuildError::InvalidUri(endpoint.clone(), er.to_string())
+            })
+            .map_err(|error| invalid_endpoint_env(signal_endpoint_var, &endpoint, error))
```
scottgerring, line 783: will this double nest errors? - i think it might look like:

```
"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: invalid endpoint '<value>': invalid URI <value>. Reason <parse error>" 
```

scottgerring (approved):
LGTM!

PR #3651, opened 2026-08-27
fix(otlp): reject invalid gRPC endpoint environment values

scottgerring:
## Summary

gRPC counterpart to #3650. Applies the same `endpoint_from_env` pattern to the tonic exporter: empty endpoint env vars are treated as unset, and non-Unicode values fail exporter construction instead of being silently ignored.

Unlike the HTTP side, invalid URI *syntax* already failed at `Endpoint::from_shared` — this change only adds empty-as-unset semantics and non-Unicode detection.

### DIFF
--- a/opentelemetry-otlp/src/exporter/tonic/mod.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/mod.rs
@@ -268,7 +268,7 @@ impl TonicExporterBuilder {
         let config = self.exporter_config;
 
         let endpoint_str = apply_insecure_scheme(
-            Self::resolve_endpoint(signal_endpoint_var, config.endpoint),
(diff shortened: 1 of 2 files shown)


opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -552,6 +555,18 @@ pub(crate) fn render_source_chain(err: &(dyn std::error::Error + 'static)) -> St
 /// path/query/fragment delimiter, so `unix:///tmp/x` is treated as schemed
 /// while `collector:4317/p://q` is not.
 ///
+fn endpoint_from_env(variable: &str) -> Result<Option<String>, ExporterBuildError> {
```
cijothomas, line 558: this method is misplaced. can you move this above the comments for insecure_shceme.

PR #3652, opened 2026-08-28
perf(proto): avoid cloning every span when grouping by scope

namse:
Fixes #

## Changes

`group_spans_by_resource_and_scope` takes `Vec<SpanData>` by value, but groups the spans by *reference* into a `HashMap<&InstrumentationScope, Vec<&SpanData>>` and then finishes with `span_data.clone().into()`. Since the function already owns the batch, that clone duplicates every span on every export, and the copy is dropped immediately after the proto conversion reads it.

This groups the spans while consuming the batch, so each one is moved into `impl From<SpanData> for Span` instead of being cloned first. The `InstrumentationScope` is cloned once per distinct scope rather than once per span.

One observable change: the order of `ScopeSpans` in the result is now insertion order instead of `HashMap` iteration order, so the output is deterministic for a given batch. The existing tests do not depend on the order and pass unchanged.

## Benchmarks

Apple M-series, `--release`, 200 iterations per row, spans built to look like typical HTTP server spans. Times are per batch; `transform` is `group_spans_by_resource_and_scope` alone, `encode` is `prost::Message::encode` of the resulting `ExportTraceServiceRequest` (unchanged by this PR, shown for scale). Encoded payload bytes were asserted identical between the two implementations.

| batch | transform (before) | transform (after) | encode | transform + encode |
|---|---|---|---|---|
| 512 spans, 6 attrs | 529 µs | **290 µs** (-45%) | 174 µs | 702 µs → **464 µs** (-34%) |
| 512 spans, 11 attrs | 843 µs | **442 µs** (-48%) | 254 µs | 1.097 ms → **696 µs** (-37%) |
| 512 spans, 11 attrs + 5 events | 1.150 ms | **545 µs** (-53%) | 406 µs | 1.556 ms → **951 µs** (-39%) |
| 2048 spans, 11 attrs | 3.440 ms | **1.818 ms** (-47%) | 1.040 ms | 4.480 ms → **2.857 ms** (-36%) |

The allocation churn goes away too: the 512-span batch above encodes to 269 KiB, and the clone was building an equivalent amount of owned `String`/`Vec` data per export just to drop it.

<details>
<summary>Benchmark harness</summary>

A standalone binary depending on `opentelemetry`, `opentelemetry_sdk`, `opentelemetry-proto` (`gen-tonic-messages`, `trace`) and `prost`. It builds a `Vec<SpanData>` once, then per iteration clones the batch (outside the timed region, since the function consumes it), times the grouping call, and times the encode:

```rust
for _ in 0..rounds {
    let cloned = batch.clone();

    let transform_start = Instant::now();
    let resource_spans = group_spans_by_resource_and_scope(cloned, &resource_proto);
    transform_total += transform_start.elapsed();

    let request = ExportTraceServiceRequest { resource_spans };
    let mut buffer = Vec::with_capacity(request.encoded_len());

    let encode_start = Instant::now();
    request.encode(&mut buffer).unwrap();
    encode_total += encode_start.elapsed();

    std::hint::black_box(&buffer);
}
```

The "after" column runs the same loop against a local copy of the new implementation, and asserts `buffer.len()` matches the "before" run.

</details>

* [x] Unit tests added/updated (if applicable) — the existing `group_spans_by_resource_and_scope` tests cover single-scope, multi-scope and schema-url behaviour and pass unchanged
* [x] Changes in public API reviewed (if applicable) — no signature change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SXikH9N5aBgd4iPKz9Su3r

### DIFF
--- a/opentelemetry-proto/CHANGELOG.md
+++ b/opentelemetry-proto/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## vNext
 
 - **Bug fix**: Accept empty `AnyValue` objects in OTLP/JSON payloads instead of rejecting the entire request.
+- **Performance**: `group_spans_by_resource_and_scope` no longer clones every `SpanData`. The batch is owned by the function, so spans are now moved into the proto conversion instead of being grouped by reference and cloned. This cuts the transform time of a 512-span batch roughly in half.
 
 ## 0.32.0
 
--- a/opentelemetry-proto/CHANGELOG.md
+++ b/opentelemetry-proto/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## vNext
 
 - **Bug fix**: Accept empty `AnyValue` objects in OTLP/JSON payloads instead of rejecting the entire request.
-- **Performance**: `group_spans_by_resource_and_scope` no longer clones every `SpanData`. The batch is owned by the function, so spans are now moved into the proto conversion instead of being grouped by reference and cloned. This cuts the transform time of a 512-span batch roughly in half.
+- **Performance**: `group_spans_by_resource_and_scope` no longer clones every `SpanData`. The batch is owned by the function, so spans are now moved into the proto conversion instead of being grouped by reference and cloned. This cuts the transform time of a 512-span batch roughly in half. ([#3652](https://github.com/open-telemetry/opentelemetry-rust/pull/3652))
 
 ## 0.32.0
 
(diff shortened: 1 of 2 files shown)


scottgerring:
@bantonsson reckon you could take a look at this one when you've got a moment? You've spent the most time of anyone thinking about perf in tracing!

bantonsson (approved):

bantonsson:
This is ready to be merged @cijothomas

cijothomas (approved):

PR #3655, opened 2026-08-29
fix(context)!: replace over-broad FutureExt with precise extension traits

### LINKED ISSUE #3291: [Feature]: Deprecate FutureExt trait for having a far too general blanket implementation
### Related Problems?

The `opentelemetry::context::future_ext::FutureExt` extension trait has a blanket implementation that is overly general:

```rust
impl<T: Sized> FutureExt for T {}
```

This is a code smell of utmost degree. `10i32.with_current_context()` should not work.

This blanket-impl was likely made because `WithContext` implements `Future`, `Stream` and `Sink` and the original author was ignorant of how to achieve a narrower extension trait.

Introduce three separate traits providing a `with_context` extension method for `Future`, `Stream`, and `Sink` separately.

### Considered Alternatives

Don't import `FutureExt` and just make my own extension trait that isn't poorly coded.

### Additional Context

_No response_

### Tip

dcdeniz:
Fixes #3291

The blanket `impl<T> FutureExt for T` puts `with_context()` on literally every type, so it collides with `anyhow::Context::with_context` and you get an inference error at any call site where both traits are in scope. This swaps the blanket impl for three precise traits.

This is breaking, and it has to be. The blanket impl is the collision, there's no way to fix #3291 while keeping it.

Trait shape is lifted from @Kile-Asmussen's #3292, which went stale. Credit there, I just picked it up and finished it off.

## Changes

- removed `impl<T> FutureExt for T`
- added `FutureContextExt`, `StreamContextExt`, `SinkContextExt<I>`. Same `with_context` / `with_current_context` methods, but implemented only for `Future` / `Stream` / `Sink` respectively
- `FutureExt` is deprecated and kept as an inert shell with no impl, so existing `use ...FutureExt` statements still resolve instead of hard-erroring
- threaded the new traits through the re-export chain (`context` → `trace::context` → `trace`)
- migrated the doc examples and the `tracing-http-propagator` example over
- CHANGELOG: Breaking (blanket impl removed), Added (new traits), plus a short migration note

On @cijothomas's ambiguity question: with the blanket impl gone `FutureExt` has no impl at all, so `.with_context()` on a future resolves to `FutureContextExt` unambiguously, even under `use trace::*`. I checked this compiles.

The tradeoff, to be upfront about it: a type that implements more than one of Future/Stream/Sink now needs to disambiguate at the call site. That's a real cost but it's narrow, and it's the price of not shadowing every other crate's `with_context`.

## Testing

Added a stream/sink test covering both new traits. Existing future tests still pass unchanged apart from the import.

- [x] Unit tests added/updated
- [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes
- [x] Changes in public API reviewed — yes, and intentional. Breaking, documented in the changelog with migration guidance.

@cijothomas @scottgerring if you'd rather keep the blanket impl and leave #3291 open, say so and I'll close this. Otherwise I think this is the fix.

### DIFF
--- a/opentelemetry/CHANGELOG.md
+++ b/opentelemetry/CHANGELOG.md
@@ -17,6 +17,8 @@
   Gated behind the `experimental_context_observer` feature flag.
 - `otel_info!`, `otel_warn!`, `otel_debug!`, and `otel_error!` macros now accept quoted-key fields
   (e.g. `"otel.component.type" = "value"`) for dotted attribute names.
+- **Added** `FutureContextExt`, `StreamContextExt`, and `SinkContextExt` traits, providing `with_context` / `with_current_context` for futures, streams, and sinks respectively (#3291).
+- **Breaking** Removed the blanket `impl<T> FutureExt for T` implementation, which exposed `with_context` on *every* type and could collide with similarly named methods from other crates (e.g. `anyhow::Context::with_context`). `FutureExt` is now `#[deprecated]` and has no implementation; migrate to the new traits above (`FutureContextExt` for futures). A type implementing more than one of `Future`/`Stream`/`Sink` must now disambiguate the call (e.g. `FutureContextExt::with_context(value, cx)`).
 - **Added** `BoundGauge<T>` and `BoundUpDownCounter<T>` types (and the
   corresponding `Gauge::bind()` / `UpDownCounter::bind()` methods), completing
   the experimental bound-instrument API across all sync instruments
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -6,7 +6,7 @@ use opentelemetry::{
     global::{self, BoxedTracer},
     logs::LogRecord,
     propagation::TextMapCompositePropagator,
-    trace::{FutureExt, Span, SpanKind, TraceContextExt, Tracer},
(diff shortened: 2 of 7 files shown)


opentelemetry/CHANGELOG.md
```diff
@@ -1,9 +1,8 @@
 # Changelog
 
 ## vNext
-
-- `otel_info!`, `otel_warn!`, `otel_debug!`, and `otel_error!` macros now accept quoted-key fields
-  (e.g. `"otel.component.type" = "value"`) for dotted attribute names.
+- **Added** `FutureContextExt`, `StreamContextExt`, and `SinkContextExt` traits, providing `with_context` / `with_current_context` for futures, streams, and sinks respectively (#3291).
+- **Breaking** Removed the blanket `impl<T> FutureExt for T` implementation, which exposed `with_context` on *every* type and could collide with similarly named methods from other crates (e.g. `anyhow::Context::with_context`). `FutureExt` is now `#[deprecated]` and has no implementation; migrate to the new traits above (`FutureContextExt` for futures). A type implementing more than one of `Future`/`Stream`/`Sink` must now disambiguate the call (e.g. `FutureContextExt::with_context(value, cx)`).
```
TommyCpp, line 5: We should keep 
`otel_info!`, `otel_warn!`, `otel_debug!`, and `otel_error!` macros now accept quoted-key fields`

PR #3658, opened 2026-08-31
fix(otlp): reject invalid gRPC endpoint environment values

scottgerring:
## Summary

gRPC counterpart to #3650. Replaces #3651 after some GH CLI issues. Applies the same `endpoint_from_env` pattern to the tonic exporter: empty endpoint env vars are treated as unset, and non-Unicode values fail exporter construction instead of being silently ignored.

Unlike the HTTP side, invalid URI *syntax* already failed at `Endpoint::from_shared` — this change only adds empty-as-unset semantics and non-Unicode detection.


scottgerring:
@cijothomas , this replaces #3651 addressing your comment. Apologies for the noise!

cijothomas (approved):

PR #3659, opened 2026-08-31
refactor(sdk): SpanProcessor::on_end takes &mut FinishedSpan instead of owned SpanData

### LINKED ISSUE #2726: Tracing SDK - Processors get clone of SpanData which is incorrect

darkraider01:
## Summary

This revives #3410 (originally by @paullegranddc, refined by @bryantbiggs). It was fully
reviewed and approved, then got closed stale waiting on a release-batching window that's
long since passed. See #2726 for the original discussion.

`SpanProcessor::on_end` currently takes `SpanData` by value. With more than one processor
configured, `Span::ensure_ended_and_exported` clones the full `SpanData` once per
processor, even if a processor only looks at a couple fields.

This changes the signature to `fn on_end(&self, span: &mut FinishedSpan)`:

- `ReadableSpan` gives read-only getters (`name()`, `attributes()`, `span_kind()`, etc.),
  so a processor that only reads never triggers a clone.
- `FinishedSpan::consume()` lets the last processor in the chain take real ownership via
  `Option::take()` (zero-copy). Earlier processors only get a clone if they actually call
  `consume()`.
- This keeps one processor's handling from affecting what later processors see, which is
  what the spec requires and what killed the very first attempt at this (#2895, which
  used a plain `&mut SpanData`).

The included benchmark shows that adding read-only processors does not trigger eager
`SpanData` clones. A clone is performed only when a non-final processor explicitly calls
`consume()`; the final registered processor can take ownership without a clone.

## Scope and follow-up

This PR intentionally keeps the `&mut FinishedSpan` / `ReadableSpan` design that was
previously reviewed. It removes eager `SpanData` cloning for read-only processors and
preserves zero-copy ownership transfer to the last registered processor that calls
`consume()`.

A follow-up PR will simplify that mechanism by changing `on_end` to take
`FinishedSpan<'_>` by value, replacing the mutable consumption state machine with Rust
ownership, and moving clone-free live-span reads to inherent `Span` methods. That larger
public API redesign is deliberately out of scope here so this revived PR can remain
focused.

## What's different from #3410

#3410 was written before three things landed in `BatchSpanProcessor::on_end`:

- #3609, the `otel.sdk.processor.span.processed` self-observability counter
- #3558, the enqueue-count race fix
- #3610, which reorganized the self-obs tests into `self_obs` modules

I cherry-picked #3410's 5 commits as-is (keeping authorship), then added one fixup
commit that:
- Keeps the current `BatchSpanProcessor::on_end` body (counting + metrics) untouched and
  just adds `let Some(span) = span.consume() else { return };` at the top, before the
  `current_batch_size` increment.
- Updates the test call sites added since #3410 (race-fix and self-obs tests) to build
  `&mut FinishedSpan::new(...)`.

The core ownership model remains the reviewed `&mut FinishedSpan` design. Subsequent
review fixes and tests are included in this PR; the by-value `FinishedSpan<'_>` redesign
is reserved for the follow-up described above.

## Test plan

- [x] `cargo build -p opentelemetry-sdk --all-features` and `--no-default-features`
- [x] `cargo test -p opentelemetry-sdk --all-features`: 401 unit tests + all doctests pass
- [x] All 5 `trace::span_processor::...::self_obs` tests pass individually (they touch
      the global `MeterProvider`, so they need isolation, same as `test.sh` does)
- [x] `cargo bench --bench span_processor_api --features testing`: cost stays flat
      (~520-720ns) across 0/1/2/4 processors
- [x] `cargo build --workspace`: stress, examples, and everything else still compiles
- [x] `cargo clippy --workspace --all-features --all-targets` and `cargo fmt --check` clean

Closes #2726.

### DIFF
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -15,12 +15,17 @@ use opentelemetry_sdk::{
     error::OTelSdkResult,
     logs::{LogProcessor, SdkLogRecord, SdkLoggerProvider},
     propagation::{BaggagePropagator, TraceContextPropagator},
-    trace::{SdkTracerProvider, SpanProcessor},
+    trace::{FinishedSpan, ReadableSpan, SdkTracerProvider, SpanProcessor},
 };
 use opentelemetry_semantic_conventions::trace;
 use opentelemetry_stdout::{LogExporter, SpanExporter};
 use std::time::Duration;
-use std::{convert::Infallible, net::SocketAddr, sync::OnceLock};
+use std::{
+    collections::HashMap,
+    convert::Infallible,
+    net::SocketAddr,
+    sync::{Mutex, OnceLock},
+};
 use tokio::net::TcpListener;
 use tracing::info;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
@@ -84,6 +89,7 @@ async fn router(
         let span = tracer
             .span_builder("router")
             .with_kind(SpanKind::Server)
+            .with_attributes([KeyValue::new("http.route", req.uri().path().to_string())])
             .start_with_context(tracer, &parent_cx);
 
         info!(name = "router", message = "Dispatching request");
@@ -103,6 +109,66 @@ async fn router(
     };
 
     response
+}
+
+#[derive(Debug, Default)]
+/// A custom span processor that counts concurrent requests for each route (indentified by the http.route
+/// attribute) and adds that information to the span attributes.
+struct RouteConcurrencyCounterSpanProcessor(Mutex<HashMap<opentelemetry::Key, usize>>);
+
+impl SpanProcessor for RouteConcurrencyCounterSpanProcessor {
+    fn force_flush(&self) -> OTelSdkResult {
+        Ok(())
+    }
+
+    fn shutdown_with_timeout(&self, _timeout: Duration) -> crate::OTelSdkResult {
+        Ok(())
+    }
+
+    fn on_start(&self, span: &mut opentelemetry_sdk::trace::Span, _cx: &Context) {
+        if !matches!(span.span_kind(), SpanKind::Server) {
+            return;
+        }
+        let Some(route) = span
+            .attributes()
+            .iter()
+            .find(|kv| kv.key.as_str() == "http.route")
+        else {
+            return;
+        };
+        let Ok(mut counts) = self.0.lock() else {
+            return;
+        };
+        let count = counts.entry(route.key.clone()).or_default();
+        *count += 1;
+        span.set_attribute(KeyValue::new(
+            "http.route.concurrent_requests",
+            *count as i64,
+        ));
+    }
+
+    fn on_end(&self, span: &mut FinishedSpan) {
+        if !matches!(span.span_kind(), SpanKind::Server) {
+            return;
+        }
+        let Some(route) = span
+            .attributes()
+            .iter()
+            .find(|kv| kv.key.as_str() == "http.route")
+        else {
+            return;
+        };
+        let Ok(mut counts) = self.0.lock() else {
+            return;
+        };
+        let Some(count) = counts.get_mut(&route.key) else {
+            return;
+        };
+        *count -= 1;
+        if *count == 0 {
+            counts.remove(&route.key);
+        }
+    }
 }
 
 /// A custom log processor that enriches LogRecords with baggage attributes.
@@ -146,7 +212,7 @@ impl SpanProcessor for EnrichWithBaggageSpanProcessor {
         }
     }
 
-    fn on_end(&self, _span: opentelemetry_sdk::trace::SpanData) {}
+    fn on_end(&self, _span: &mut opentelemetry_sdk::trace::FinishedSpan) {}
 }
 
 fn init_tracer() -> SdkTracerProvider {
@@ -162,6 +228,7 @@ fn init_tracer() -> SdkTracerProvider {
     // Setup tracerprovider with stdout exporter
     // that prints the spans to stdout.
     let provider = SdkTracerProvider::builder()
+        .with_span_processor(RouteConcurrencyCounterSpanProcessor::default())
         .with_span_processor(EnrichWithBaggageSpanProcessor)
         .with_simple_exporter(SpanExporter::default())
         .build();
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -112,7 +112,7 @@ async fn router(
 }
 
 #[derive(Debug, Default)]
-/// A custom span processor that counts concurrent requests for each route (indentified by the http.route
+/// A custom span processor that counts concurrent requests for each route (identified by the http.route
 /// attribute) and adds that information to the span attributes.
 struct RouteConcurrencyCounterSpanProcessor(Mutex<HashMap<opentelemetry::Key, usize>>);
 
--- a/examples/tracing-http-propagator/src/server.rs
+++ b/examples/tracing-http-propagator/src/server.rs
@@ -111,10 +111,21 @@ async fn router(
     response
 }
 
+/// Returns the `http.route` value for server spans, `None` otherwise.
+fn route_of(span: &impl ReadableSpan) -> Option<String> {
+    if !matches!(span.span_kind(), SpanKind::Server) {
+        return None;
+    }
+    span.attributes()
+        .iter()
+        .find(|kv| kv.key.as_str() == "http.route")
+        .map(|kv| kv.value.to_string())
+}
+
 #[derive(Debug, Default)]
 /// A custom span processor that counts concurrent requests for each route (identified by the http.route
 /// attribute) and adds that information to the span attributes.
-struct RouteConcurrencyCounterSpanProcessor(Mutex<HashMap<opentelemetry::Key, usize>>);
+struct RouteConcurrencyCounterSpanProcessor(Mutex<HashMap<String, usize>>);
 
 impl SpanProcessor for RouteConcurrencyCounterSpanProcessor {
     fn force_flush(&self) -> OTelSdkResult {
@@ -126,20 +137,13 @@ impl SpanProcessor for RouteConcurrencyCounterSpanProcessor {
     }
 
     fn on_start(&self, span: &mut opentelemetry_sdk::trace::Span, _cx: &Context) {
-        if !matches!(span.span_kind(), SpanKind::Server) {
-            return;
-        }
-        let Some(route) = span
-            .attributes()
-            .iter()
-            .find(|kv| kv.key.as_str() == "http.route")
-        else {
+        let Some(route) = route_of(span) else {
             return;
         };
         let Ok(mut counts) = self.0.lock() else {
             return;
         };
-        let count = counts.entry(route.key.clone()).or_default();
+        let count = counts.entry(route).or_default();
         *count += 1;
         span.set_attribute(KeyValue::new(
             "http.route.concurrent_requests",
@@ -148,25 +152,18 @@ impl SpanProcessor for RouteConcurrencyCounterSpanProcessor {
     }
 
     fn on_end(&self, span: &mut FinishedSpan) {
-        if !matches!(span.span_kind(), SpanKind::Server) {
-            return;
-        }
-        let Some(route) = span
-            .attributes()
-            .iter()
-            .find(|kv| kv.key.as_str() == "http.route")
-        else {
+        let Some(route) = route_of(span) else {
             return;
         };
         let Ok(mut counts) = self.0.lock() else {
             return;
         };
-        let Some(count) = counts.get_mut(&route.key) else {
+        let Some(count) = counts.get_mut(&route) else {
             return;
         };
         *count -= 1;
         if *count == 0 {
-            counts.remove(&route.key);
+            counts.remove(&route);
         }
     }
 }
--- a/opentelemetry-sdk/src/trace/span.rs
+++ b/opentelemetry-sdk/src/trace/span.rs
@@ -37,6 +37,8 @@ pub(crate) struct SpanData {
     pub(crate) start_time: SystemTime,
     /// Span end time
     pub(crate) end_time: SystemTime,
+    /// Whether end_time was explicitly set via `end_with_timestamp`
+    pub(crate) end_time_set: bool,
     /// Span attributes
     pub(crate) attributes: Vec<KeyValue>,
     /// The number of attributes that were above the configured limit, and thus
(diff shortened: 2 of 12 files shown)


scottgerring:
Hey @paullegranddc do you have time to look at this again?

darkraider01:
Was re-reading my own diff and spotted something I'm not sure about.

In end_and_export the original PR moves the span context out instead of cloning it:


let span_context = std::mem::replace(&mut self.span_context, SpanContext::empty_context());
main just does self.span_context.clone() there. I get why it was changed, it saves a TraceState clone, but the side effect is that once a span ends, span_context() on it comes back all zeroes:


BEFORE: trace_id=93240633251a59c9f50c083000de8de4 span_id=25c8b0a0397bb568
AFTER : trace_id=00000000000000000000000000000000 span_id=0000000000000000
That's a public method on the Span trait, so anyone who ends a span and then grabs the trace id for log correlation would silently get nothing.

Took me a while to notice because allows_to_get_span_context_after_end passes either way. create_span() in that test already starts with empty_context(), so the assert holds regardless of what end() does to it.

I'm inclined to just put the clone back and make that test use a real tracer span instead, the saving seems small next to changing public behavior. Fine either way, or we keep the move and call it out as breaking if the perf is worth it.

examples/tracing-http-propagator/src/server.rs
```diff
@@ -105,6 +111,66 @@ async fn router(
     response
 }
 
+#[derive(Debug, Default)]
+/// A custom span processor that counts concurrent requests for each route (indentified by the http.route
```
bantonsson:
```suggestion
/// A custom span processor that counts concurrent requests for each route (identified by the http.route
```

bantonsson (changes requested):
Thanks for reviving this @darkraider01 

> I'm inclined to just put the clone back and make that test use a real tracer span instead, the saving seems small next to changing public behavior. Fine either way, or we keep the move and call it out as breaking if the perf is worth it.

I do think that the perf is worth it, so let's fix the test and call it out in the release notes. A quick local check with 1 and 32 entries in the TraceState shows:

| TraceState | clone | `mem::replace` | cost of the clone | cost as % of span lifecycle |
|---|---|---|---|---|
| 1 entry (realistic) | 245 ns | 212 ns | ~33 ns | ~13% |
| 32 entries (W3C max) | 2.02 µs | 1.11 µs | ~900 ns | ~45% |

darkraider01:
Pushed both of those, thanks for the numbers, that settles it. Renamed the test to span_context_is_cleared_after_end and it now starts from a real tracer span, so it actually checks the context is valid before end() and empty after. Also added a breaking note in the changelog explaining what to do instead.

While I was in there I ran into a few other things. All minor, none of them blocking, just flagging them so you can say what's worth doing.

1. We build the export data before checking how many processors there are.


let mut finished_span =
    FinishedSpan::new(build_export_data(data, span_context, &self.tracer));

let span_processors = provider.span_processors();
for (i, processor) in span_processors.iter().enumerate() {
main short circuits on [] => {}, so with no processors registered it skips all of it. Now we build the whole SpanData, clone the instrumentation scope, and drop it on the floor. I suspect that's why 0_processors benches slower than 1_processors in the numbers up top, which looked odd to me at the time. Fix is just an early return when the slice is empty.

2. FinishedSpan::new is public but reset isn't.

new always sets is_last_processor: true and reset is the only way to change it. So if someone writes their own composite or fan out processor outside the SDK, the first inner processor takes the data and the rest get None. Their only way around it is building a fresh FinishedSpan per processor, which is exactly the cloning this change is trying to avoid. Either reset wants to be public, or new should be pub(crate) with a separate testing constructor.

3. Registration order now affects allocations.

Since the last processor gets the move, [Batch, Logging] and [Logging, Batch] behave differently. The trait docs explain the mechanism but don't say the useful part, which is to register whichever processor needs ownership last. Probably just a sentence in the docs.

Happy to fold any of these into this PR, or keep it focused and do them separately, whichever you prefer. 
@bantonsson

bantonsson:
Hey @darkraider01 I'm sorry for not reading the OTel spec before asking you to do changes. The spec says that the span context returned *must* be valid after `end`.

https://opentelemetry.io/docs/specs/otel/trace/api/#get-context

Let's just go back to the clone for this PR, and then the follow up things, including putting the `TraceState` `VecDeque` inside an `Arc`, can be in a separate PR.

bantonsson (approved):

scottgerring:
I think we should be able to finally merge this one!

examples/tracing-http-propagator/src/server.rs
```diff
@@ -105,6 +111,66 @@ async fn router(
...
+            .attributes()
+            .iter()
+            .find(|kv| kv.key.as_str() == "http.route")
+        else {
+            return;
+        };
+        let Ok(mut counts) = self.0.lock() else {
+            return;
+        };
+        let count = counts.entry(route.key.clone()).or_default();
```
TommyCpp: The `route.key` here is always `http.route` right? If the goal is to count request for different route(`/a`, `/b`, etc), shouldn't we use `route.value` here?

examples/tracing-http-propagator/src/server.rs
```diff
@@ -105,6 +111,66 @@ async fn router(
...
+            .attributes()
+            .iter()
+            .find(|kv| kv.key.as_str() == "http.route")
+        else {
+            return;
+        };
+        let Ok(mut counts) = self.0.lock() else {
+            return;
+        };
+        let Some(count) = counts.get_mut(&route.key) else {
```
TommyCpp: ditto

examples/tracing-http-propagator/src/server.rs
```diff
@@ -105,6 +111,66 @@ async fn router(
...
+        if !matches!(span.span_kind(), SpanKind::Server) {
+            return;
+        }
+        let Some(route) = span
+            .attributes()
+            .iter()
+            .find(|kv| kv.key.as_str() == "http.route")
+        else {
+            return;
+        };
```
TommyCpp: This can be abstract away and make read only 

```rust
    /// Returns the `http.route` value for server spans, `None` otherwise.
    fn route_of(span: &impl ReadableSpan) -> Option<String> {
        if !matches!(span.span_kind(), SpanKind::Server) {
            return None;
        }
        span.attributes()
            .iter()
            .find(|kv| kv.key.as_str() == "http.route")
            .map(|kv| kv.value.to_string())
    }
```

scottgerring (approved):

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -194,60 +196,402 @@ impl opentelemetry::trace::Span for Span {
...
+    pub fn consume(&mut self) -> Option<crate::trace::SpanData> {
+        if self.is_consumed {
+            opentelemetry::otel_error!(name: "FinishedSpan.ConsumeTwice", message = "consume called twice on FinishedSpan in the same span processor");
+            return None;
+        }
+        self.is_consumed = true;
+        if self.is_last_processor {
+            self.span.take()
+        } else {
+            self.span.clone()
```
utpilla, line 430: For a non-last processor, `consume()` sets `is_consumed` but leaves `self.span` populated because it returns a clone. The `ReadableSpan` getters inspect only `self.span`, so they continue returning the original values. For the last processor, `take()` removes the data and the same getters return defaults.

This makes the documented post-`consume()` behavior depend on processor position. Could the getters also treat `is_consumed` as unavailable, with `reset()` restoring access for the next processor? A test covering `consume()` after `reset(false)` would catch this case.

darkraider01: All ReadableSpan getters on FinishedSpan now inspect an internal helper that yields None whenever is_consumed is true, ensuring consistent default values post-consume() regardless of processor position in the chain. Calling reset() clears is_consumed, restoring access for subsequent processors. Added unit and pipeline tests covering consume() after reset(false) and verifying access restoration on reset().

opentelemetry-sdk/src/trace/span.rs
```diff
@@ -194,60 +196,406 @@ impl opentelemetry::trace::Span for Span {
...
+/// ```
+pub struct FinishedSpan {
+    span: Option<crate::trace::SpanData>,
+    is_last_processor: bool,
+    is_consumed: bool,
+}
+
+impl FinishedSpan {
+    /// Creates a new `FinishedSpan` with the given span data.
+    pub fn new(span_data: crate::trace::SpanData) -> Self {
```
utpilla: Could we mark this method `pub(crate)`?

Since `FinishedSpan::new` is public and `on_end` receives `&mut FinishedSpan`, a processor can bypass the read-only contract by replacing the wrapper:

```rust
*span = FinishedSpan::new(replacement);
```
Subsequent processors would then observe replacement data. Could this constructor be `pub(crate)` so the mutable reference is limited to `consume()` bookkeeping rather than allowing the completed span to be replaced?

```suggestion
    pub(crate) fn new(span_data: crate::trace::SpanData) -> Self {
```

utpilla:
I found the earlier `Cow` discussion in #2726. It appears it was dismissed because the span must remain available when an earlier processor does not take ownership. I think that concern is addressed if the SDK creates a separate borrowed-or-owned value for each callback rather than reusing one mutable value across callbacks.

Conceptually, the dispatch can borrow the completed `SpanData` to every preceding processor and move it into the final processor:
```rust
let Some((last, preceding)) = processors.split_last() else {
    return;
};
for processor in preceding {
    processor.on_end(FinishedSpan::borrowed(&span_data));
}
last.on_end(FinishedSpan::owned(span_data));
```

Internally, `FinishedSpan` could be a thin wrapper around `Cow`:
```rust
pub struct FinishedSpan<'a>(Cow<'a, SpanData>);
impl<'a> FinishedSpan<'a> {
    pub fn span_data(&self) -> &SpanData {
        self.0.as_ref()
    }
    pub fn into_owned(self) -> SpanData {
        self.0.into_owned()
    }
    pub fn reborrow(&self) -> FinishedSpan<'_> {
        FinishedSpan(Cow::Borrowed(self.0.as_ref()))
    }
}
```

The `borrowed` and `owned` constructors could remain crate-private. The processor contract would take the wrapper by value:
```rust
fn on_end(&self, span: FinishedSpan<'_>);
```

A read-only processor performs no clone:
```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    if span.span_data().attributes.iter().any(should_count) {
        // ...
    }
}
```

An ownership-taking processor calls `into_owned()`:
```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    self.sender.try_send(span.into_owned());
}
```

This gives the intended behavior:
- A preceding ownership-taking processor clones the borrowed `SpanData`.
- The final ownership-taking processor moves the original `SpanData`.
- A processor that only reads does not clone.

A filtering processor can forward the same value:
```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    if should_export(span.span_data()) {
        self.next.on_end(span);
    }
}
```

A composite processor can efficiently fan out by reborrowing for preceding children and forwarding its original value to the final child:
```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    let Some((last, preceding)) = self.processors.split_last() else {
        return;
    };
    for processor in preceding {
        processor.on_end(span.reborrow());
    }
    last.on_end(span);
}
```

With this approach, I also do not think `ReadableSpan` is needed for `on_end`. `FinishedSpan::span_data()` returns an immutable `&SpanData`, whose existing public fields already expose the completed-span representation without duplicating its API as a getter trait.

`ReadableSpan` also makes live `Span` state readable during `on_start`, but that is a separate public API decision and is not required to eliminate completed-span clones. If that capability is desirable, (for example, as part of future [`OnEnding`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onending) support) it could be designed independently rather than coupled to this ownership change.

Compared with the current implementation, this preserves the same minimum `N - 1` clones for `N` ownership-taking processors, including zero clones for a sole or final batch processor, while removing:
- `is_last_processor`
- `is_consumed`
- `reset()`
- post-consumption default values
- the new `ReadableSpan` trait and its duplicated getters
- double-consumption handling

Taking the wrapper by value also lets Rust ownership enforce consumption: after `into_owned()`, the processor cannot continue reading or consume it again.

scottgerring (commented):
.

bantonsson:
Hi @utpilla, I implemented your `Cow` variant to test it. A few points:

1. The original PR #2962 clearly states that it is trying to solve both the read in `on_start` and the clone in `on_end`. The reason is that both of them are performance critical for our work in the [Datadog OpenTelemetry pipeline](https://github.com/DataDog/dd-trace-rs). Separately, the current `on_start` also does not meet the spec: [`onStart`](https://opentelemetry.io/docs/specs/otel/trace/sdk/#onstart) hands the processor a readable and writable span, while our `on_start` is write-only, and the only read (`Span::exported_data()`) clones the whole span. So moving out the `ReadableSpan` fix is not really an option.

2. While the `Cow` solution looks identical, it does incur a constant overhead once at least one processor is registered. I implemented it locally, and the overhead on my machine is 12 ns:

   - `0_processors`: 137.5 ns before, 138.9 ns after, no change (p = 0.78)
   - `1_processors`: 154.6 ns before, 166.6 ns after, +12.0 ns
   - `2_processors`: 155.2 ns before, 168.6 ns after, +13.4 ns
   - `4_processors`: 155.2 ns before, 171.4 ns after, +16.2 ns

   The cost is a constant 12 ns that appears with the first processor, plus about 1.6 ns per additional processor. The memcpy moves the whole wrapper, which for the owned case contains the full `SpanData`, into the callee's stack frame.

3. Having the `ReadableSpan` implementation for the `FinishedSpan` is just nice symmetry between `on_start` and `on_end`, but not really important if it has to be removed. The `on_start` half is the one we need.

What needs to be decided for this PR is if the constant overhead of the memcpy for the owned `Cow` which does simplify the code a bit, is worth it for that change.

utpilla:
@bantonsson 

Thanks, both points make sense. I agree that:

1. The live `Span` needs a borrowed read API so `on_start` can inspect it without calling `exported_data()` and cloning the entire span.
2. Passing `Cow<'_, SpanData>` directly is not ideal if its owned variant moves the complete `SpanData` through the callback.

I think we can preserve both goals while simplifying the `on_end` ownership contract.

Instead of storing `SpanData` inline, `FinishedSpan` could be a small by-value wrapper whose variants both contain references:

```rust
pub struct FinishedSpan<'a> {
    inner: FinishedSpanInner<'a>,
}

enum FinishedSpanInner<'a> {
    Borrowed(&'a SpanData),
    Owned(&'a mut Option<SpanData>),
}
```

The public API could be limited to:

```rust
impl FinishedSpan<'_> {
    /// Inspect the completed span without cloning.
    pub fn span_data(&self) -> &SpanData {
        match &self.inner {
            FinishedSpanInner::Borrowed(data) => data,
            FinishedSpanInner::Owned(slot) => {
                slot.as_ref().expect("span data must be available")
            }
        }
    }

    /// Obtain ownership, cloning a borrowed span or taking the final span.
    pub fn into_owned(self) -> SpanData {
        match self.inner {
            FinishedSpanInner::Borrowed(data) => data.clone(),
            FinishedSpanInner::Owned(slot) => {
                slot.take().expect("span data must be available")
            }
        }
    }

    /// Create a borrowed view for a delegated processor.
    pub fn reborrow(&self) -> FinishedSpan<'_> {
        FinishedSpan::borrowed(self.span_data())
    }
}
```

The `borrowed` and `owned` constructors would remain crate-private. The processor contract would take the wrapper by value:

```rust
fn on_end(&self, span: FinishedSpan<'_>);
```

The SDK would retain the actual `SpanData` in one location, borrow it to preceding processors, and allow the final processor to take it:

```rust
let mut span_data = Some(build_export_data(...));
let Some((last, preceding)) = processors.split_last() else {
    return;
};

for processor in preceding {
    processor.on_end(FinishedSpan::borrowed(
        span_data.as_ref().expect("span data must be available"),
    ));
}

last.on_end(FinishedSpan::owned(&mut span_data));
```

This keeps the call-site UX straightforward.

A read-only processor uses the existing completed-span representation:

```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    let span = span.span_data();

    if span.attributes.iter().any(should_count) {
        // ...
    }
}
```

An ownership-taking processor consumes the wrapper:

```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    self.sender.try_send(span.into_owned());
}
```

A filtering processor can forward the same value:

```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    if should_export(span.span_data()) {
        self.next.on_end(span);
    }
}
```

A composite processor can reborrow for preceding children and forward its original ownership opportunity to the final child:

```rust
fn on_end(&self, span: FinishedSpan<'_>) {
    let Some((last, preceding)) = self.processors.split_last() else {
        return;
    };

    for processor in preceding {
        processor.on_end(span.reborrow());
    }

    last.on_end(span);
}
```

This preserves the intended ownership behavior:

- A preceding ownership-taking processor clones the borrowed `SpanData`.
- The final ownership-taking processor takes the original `SpanData`.
- A processor that only reads does not clone.
- A composite processor can recursively preserve the same behavior.

For the live `Span`, I agree that some public borrowed read API is needed for `on_start`. However, that does not necessarily require introducing a separate `ReadableSpan` trait.

If completed spans are inspected through `FinishedSpan::span_data() -> &SpanData`, `ReadableSpan` would primarily have one implementation: the live `Span`. Could the borrowed getters instead be inherent methods on `Span`?

```rust
impl Span {
    pub fn name(&self) -> Option<&str> {
        // ...
    }

    pub fn attributes(&self) -> &[KeyValue] {
        // ...
    }

    pub fn events(&self) -> &[Event] {
        // ...
    }

    pub fn status(&self) -> &Status {
        // ...
    }
}
```

The `on_start` UX would remain read/write:

```rust
fn on_start(&self, span: &mut Span, _cx: &Context) {
    if span.attributes().iter().any(should_count) {
        // Read without cloning.
    }

    span.set_attribute(...);
}
```

The same getters would also work if an `OnEnding` hook is added later, since that hook receives the live `Span` as well.

A `ReadableSpan` trait could still be useful if we specifically want generic functions over multiple span representations. But if completed spans use the existing `SpanData` representation, inherent getters on `Span` seem to provide the required capability with one less public abstraction.

Compared with the current reusable `&mut FinishedSpan`, this approach removes the need for:

- `is_last_processor`
- `is_consumed`
- `reset()`
- post-consumption default values
- double-consumption handling
- a `ReadableSpan` implementation that duplicates the `SpanData` API

Taking `FinishedSpan` by value also lets normal Rust ownership express the callback lifecycle:

- `into_owned(self)` can only be called once.
- A processor cannot continue reading after taking ownership.
- The borrowed-versus-final-owned distinction remains private.
- `SpanData` itself is not moved through the callback unless ownership is requested.

This separates the two concerns cleanly:

- `Span` provides borrowed read/write access during `on_start` and potentially `OnEnding`.
- `SpanData` remains the existing completed-span representation.
- `FinishedSpan<'_>` only controls efficient ownership transfer during `on_end`.

What do you think?

bantonsson:
@utpilla I think that your last proposal is fine.

It's quite a big change to add to this PR. Would you be up for doing it @darkraider01 , or should I or @utpilla do it?

darkraider01:
yea i'd be happy to do it, maybe would be better to do it in a follow up pr since i agree it's a big change to be added

darkraider01:
Superseded by #3741, which retains the clone-avoidance goal while simplifying the completed-span ownership architecture.

PR #3660, opened 2026-08-31
fix(otlp): enable retries by default

cijothomas:
## Summary

- enable retries by default for transient OTLP/HTTP and OTLP/gRPC export failures, using the recommended exponential-backoff policy with up to 3 retries (4 attempts total)
- add `RetryPolicy::disabled()` as a clear single-attempt opt-out while preserving custom policies through `.with_retry_policy(...)`
- consolidate the unreleased retry changelog and scope feature-flag migration to users who enabled the old experimental flags

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -28,8 +28,8 @@
   async-runtime processors.
   The default `RetryPolicy` follows the OTLP specification's retry requirement,
   using exponential backoff with up to 3 retries (4 attempts total). Configure
-  `.with_retry_policy(...)` explicitly to customize the policy or disable
-  retries.
+  `.with_retry_policy(...)` explicitly to customize the policy, or use
+  `.with_retry_policy(RetryPolicy::disabled())` to disable retries.
   **Migration**: Remove the experimental retry feature flags from your `Cargo.toml`.
   [#3621](https://github.com/open-telemetry/opentelemetry-rust/pull/3621)
 - Add support for INSECURE environment variables for gRPC (env-var-only, no builder method, per spec):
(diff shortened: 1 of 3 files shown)


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,31 @@
 
 ## vNext
 
+### Retry
+
+- Retries are now enabled by default for OTLP/HTTP and OTLP/gRPC. The default
```
scottgerring, line 7: should we mention that this is the case for both the regular and unstable/async processors?

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,31 @@
...
+  policy uses exponential backoff and jitter with up to 3 retries (4 attempts
+  total). Use `.with_retry_policy(RetryPolicy::disabled())` to disable retries,
+  or provide a custom `RetryPolicy` to change the behavior.
+- **Migration for users of the experimental retry features:** If your
+  `Cargo.toml` enables `grpc-tonic-with-retry`, `http-proto-with-retry`, or
```
scottgerring, line 12: just noticed this 

```suggestion
  `Cargo.toml` enables `experimental-grpc-retry`, `experimental-http-retry`, or
```

opentelemetry-otlp/CHANGELOG.md
scottgerring:
```suggestion
- Retries are now enabled by default for OTLP/HTTP and OTLP/gRPC exporters when used with the default thread-based processors as well as the experimental async-runtime variants. The default
```

scottgerring (approved):
lgtm! couple doc nits

PR #3672, opened 2026-09-01
refactor(otlp): stabilize retry policy API

cijothomas:
## Summary

Prepare the OTLP retry API for stabilization by defining a small, evolvable public configuration surface and keeping transport-specific execution details internal.

- retain `RetryPolicy` at the crate root as the only supported retry API
- make the retry engine, error type, protocol classifiers, and implementation modules crate-private
- replace public policy fields with `Duration`-based fluent configuration methods
- separate public policy configuration from the retry execution engine
- tighten feature gates so retry internals compile only for transport and signal combinations that use them
- make jitter generation safe for maximum `Duration` values
- update examples and migration guidance for the stabilized API shape

## API impact

This removes the previously public retry implementation modules and prevents construction of `RetryPolicy` with struct literals. Users should import `opentelemetry_otlp::RetryPolicy` and configure custom policies through its `with_*` methods.

The resulting public surface can evolve without exposing classifier semantics, transport dependency types, engine signatures, or policy storage details as semver commitments.

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -658,12 +658,15 @@ mod metric;
 #[cfg(any(feature = "http-proto", feature = "http-json", feature = "grpc-tonic"))]
 mod span;
 
-#[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
-pub mod retry_classification;
+#[cfg(all(
+    any(feature = "trace", feature = "metrics", feature = "logs"),
+    any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json")
+))]
+mod retry_classification;
 
 /// Retry logic for exporting telemetry data.
 #[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
-pub mod retry;
+mod retry;
 
 pub use crate::exporter::Compression;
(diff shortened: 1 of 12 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -658,12 +658,28 @@ mod metric;
...
+        feature = "grpc-tonic",
+        any(feature = "trace", feature = "metrics", feature = "logs")
+    )
+))]
+mod retry;
 
-/// Retry logic for exporting telemetry data.
 #[cfg(any(feature = "grpc-tonic", feature = "http-proto", feature = "http-json"))]
-pub mod retry;
+mod retry_policy;
```
scottgerring, line 688: nit: is it worth pointing out that this is gated more broadly than `retry_classification` because users need to be able to configure this in order to configure any transport? 

I've been burnt enough times by the interaction of feature flags that any hints in the gating feel helpful!

scottgerring (approved):
makes sense!

opentelemetry-otlp/src/lib.rs
cijothomas: Added a comment explaining why `RetryPolicy` uses the broader transport-only gate: transport builders expose policy configuration even when no signal is enabled, while the retry execution modules only compile when they have a signal consumer.

PR #3679, opened 2026-09-02
fix(sdk): drop NaN/Infinity measurements in Histogram, Sum, LastValue, PrecomputedSum

### LINKED ISSUE #3656: [Bug]: Explicit bucket histogram `sum` permanently corrupted by NaN/Infinity measurements
### What happened?

Recording a NaN (or +/-Infinity) value into a `Histogram` instrument permanently corrupts
that data point's `sum`. Because `x + NaN == NaN` for any `x`, once `Buckets::total`
becomes NaN it stays NaN on every subsequent update. With Cumulative temporality (the
SDK default), the metric stream reports `sum: NaN` on every export from that point on,
for the rest of the process's lifetime , there is no self-healing. The bad value is also
silently misclassified into bucket index 0 (`bounds.partition_point(|&x| x < f)` treats
`x < NaN` as always false), rather than being rejected.

Root cause:
- `Aggregator::update` for `Mutex<Buckets<T>>` in
  opentelemetry-sdk/src/metrics/internal/histogram.rs does
  `buckets.total += value` with no finiteness check.
- `Histogram::call` in the same file computes the bucket index and calls `update`
  unconditionally, with no NaN/Infinity guard.

For comparison, the sibling exponential-histogram aggregator already handles this
correctly: opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs explicitly
checks `!measurement.into_float().is_finite()` and drops non-finite measurements before
recording. The explicit-bucket `Histogram` has no equivalent guard.

What I expected: recording a NaN/Infinite value should be ignored (matching the
exponential histogram's existing behavior), not silently poison `sum` forever.

Minimal repro:

    use opentelemetry::global;
    use opentelemetry_sdk::metrics::SdkMeterProvider;

    let provider = SdkMeterProvider::builder().build();
    global::set_meter_provider(provider.clone());
    let meter = global::meter("repro");
    let histogram = meter.f64_histogram("test").build();

    histogram.record(1.0, &[]);
    histogram.record(f64::NAN, &[]); // <- poisons `sum`
    histogram.record(2.0, &[]);

    // Every subsequent export of this metric stream reports `sum: NaN`

0.32.0

0.32.1

N/A

### Relevant log output

```shell

```

### Tip

darkraider01:
Fixes #3656.

Turns out the NaN/Infinity poisoning wasn't just a histogram sum problem. Once I checked the other aggregators (as discussed in the issue comments), they all had the same gap except ExponentialHistogram, which already guards for it.

What was broken:
- Histogram: total += value with no finiteness check, so one NaN poisons sum forever. Also NaN gets miscounted into bucket 0 since partition_point treats x < NaN as always false.
- Sum: same additive corruption, current + NaN is NaN forever after that.
- LastValue: just overwrites on the next call, so it recovers, but still exports one bad value in between.
- PrecomputedSum: also overwrite based, but under delta temporality the cached NaN poisons the delta for the next cycle too, so a single bad value corrupts two exports.
- Every bound handle (behind experimental_metrics_bound_instruments) has the exact same unguarded path as its unbound counterpart.

Fix is just copying ExponentialHistogram's existing check (measurement not finite -> drop it) into all four aggregators, at both the unbound call and the bound handle call, so it happens before any lock or map work.

Added one test per aggregator recording NaN/Inf alongside valid values and checking the bad ones get dropped. Ran the full metrics internal test suite plus clippy and fmt, all clean. Updated the changelog too.

### DIFF
--- a/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
@@ -30,6 +30,12 @@ struct BoundPrecomputedSumHandle<T: Number> {
 #[cfg(feature = "experimental_metrics_bound_instruments")]
 impl<T: Number> BoundMeasure<T> for BoundPrecomputedSumHandle<T> {
     fn call(&self, measurement: T) {
+        // Mirror unbound PrecomputedSum::call: ignore NaN and infinity so a bad
+        // value can't get cached in `reported` and poison the delta computed
+        // against the next valid measurement.
+        if !measurement.into_float().is_finite() {
+            return;
+        }
         self.tracker.aggregator.update(measurement);
         self.tracker.has_been_updated.store(true, Ordering::Release);
     }
@@ -171,6 +177,12 @@ where
     T: Number,
 {
     fn call(&self, measurement: T, attrs: &[KeyValue]) {
+        // Ignore NaN and infinity: a cached NaN in `reported` poisons the delta
+        // computed against the *next* valid measurement too
+        // (`valid_val - NaN == NaN`), so a single bad value corrupts two cycles.
+        if !measurement.into_float().is_finite() {
+            return;
+        }
         self.filter.apply(attrs, |filtered| {
             self.value_map.measure(measurement, filtered);
         })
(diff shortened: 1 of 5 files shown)


opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
```diff
@@ -171,6 +177,12 @@ where
     T: Number,
 {
     fn call(&self, measurement: T, attrs: &[KeyValue]) {
+        // Ignore NaN and infinity: a cached NaN in `reported` poisons the delta
+        // computed against the *next* valid measurement too
+        // (`valid_val - NaN == NaN`), so a single bad value corrupts two cycles.
+        if !measurement.into_float().is_finite() {
```
lalitb, line 183: Could we add a delta test with `100 -> collect -> NaN -> collect -> 110 -> collect`? Dropping `NaN` leaves `new_reported` empty and clears the baseline, so the next collection exports `110`, counting the earlier `100` again.

This also happens today for missing observations, so I'm fine handling recovery separately. Should rejected observations preserve the last valid baseline? We'd need to keep stale-attribute eviction working too, since retaining every missing entry could grow `reported` indefinitely.

lalitb (approved):
LGTM. The guards look good. My comment on delta recovery is non-blocking; we can track that separately since it also affects missing observations today.

darkraider01:
> LGTM. The guards look good. My comment on delta recovery is non-blocking; we can track that separately since it also affects missing observations today.

Since dropping NaN works just like a skipped observation right now, the baseline gets wiped and causes the re-count.

I'll add the test case (100 -> NaN -> 110) to lock down this behavior.
Agree that saving the previous baseline should be handled separately, I'll open a follow-up issue for that so we can figure out how to keep the old value without leaking memory on old attributes.

scottgerring:
I think this is ready to merge baring resolution of the conflict 
@darkraider01 do you mind cleaning it up?

darkraider01:
done! thank you!

PR #3683, opened 2026-09-02
fix(proto): accept omitted OTLP JSON collector fields

### LINKED ISSUE #3678: [Bug]: OTLP/JSON rejects empty collector requests with omitted resource arrays
### What happened?

The OTLP collector request schemas define their resource arrays as [`repeated` fields](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/collector/logs/v1/logs_service.proto#L36-L47). The [ProtoJSON mapping](https://protobuf.dev/programming-guides/json/#presence-and-default-values) omits empty repeated fields by default, and the [OTLP specification](https://opentelemetry.io/docs/specs/otlp/#full-success-1) says receivers should respond successfully to empty requests.

The OTLP specification also describes [empty telemetry envelopes](https://opentelemetry.io/docs/specs/otlp/#empty-telemetry-envelopes) as a condition that can occur, for example after filtering removes all contained telemetry.

The generated collector request types currently require the resource array during deserialization. For example, deserializing `{}` as `ExportLogsServiceRequest` fails with `missing field resourceLogs` instead of producing an empty request.

Expected behavior: omitted resource arrays decode to empty vectors for trace, log, metric, and profile export requests.

A patch and regression tests are available in #3683.

N/A. This affects `opentelemetry-proto` 0.32.0 and current `main` as of 2026-09-02.

N/A.

OTLP

### Relevant log output

missing field `resourceLogs` at line 1 column 2

bruno-espino:
Fixes #3678

## Changes

The OTLP collector request schemas define their top-level resource arrays as `repeated` fields for [traces](https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/collector/trace/v1/trace_service.proto#L36-L43), [logs](https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/collector/logs/v1/logs_service.proto#L36-L43), [metrics](https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/collector/metrics/v1/metrics_service.proto#L36-L43), and [profiles](https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto#L33-L42). The [ProtoJSON mapping](https://protobuf.dev/programming-guides/json/#presence-and-default-values) omits empty repeated fields by default, and the [OTLP specification](https://opentelemetry.io/docs/specs/otlp/#full-success-1) says receivers should respond successfully to empty requests.

The generated collector request types instead require their resource arrays during deserialization. For example, an empty logs request (`{}`) fails with `missing field resourceLogs` instead of producing an empty request.

Add serde defaults for the trace, log, metric, and profile request roots, and configure the generator to preserve them. Omitted resource arrays then decode to empty vectors without changing non-empty lowerCamelCase payloads.

This intentionally does not accept original proto field names such as `resource_logs`: [OTLP/JSON requires lowerCamelCase keys and says original proto field names are invalid](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding).

Regression tests cover omitted resource arrays for all four request types.

* [x] Appropriate `CHANGELOG.md` updated
* [x] No public API changes

### DIFF
--- a/opentelemetry-proto/CHANGELOG.md
+++ b/opentelemetry-proto/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## vNext
 
 - **Bug fix**: Accept empty `AnyValue` objects in OTLP/JSON payloads instead of rejecting the entire request.
+- **Bug fix**: Accept omitted resource fields in empty OTLP/JSON collector requests.
 
 ## 0.32.0
 
--- a/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.collector.logs.v1.rs
+++ b/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.collector.logs.v1.rs
@@ -2,6 +2,7 @@
 #[cfg_attr(feature = "with-schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "with-serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(feature = "with-serde", serde(default))]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ExportLogsServiceRequest {
     /// An array of ResourceLogs.
--- a/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.collector.metrics.v1.rs
+++ b/opentelemetry-proto/src/proto/tonic/opentelemetry.proto.collector.metrics.v1.rs
@@ -2,6 +2,7 @@
 #[cfg_attr(feature = "with-schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "with-serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "with-serde", serde(rename_all = "camelCase"))]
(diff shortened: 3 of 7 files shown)


lalitb (approved):
LGTM, Thanks. 

There's a similar issue in the `Export*PartialSuccess` types: `{"partialSuccess": {"errorMessage":"backend warning"}}` fails to deserialize because the rejection count is missing, so the exporter drops the warning. Could we track defaults for those types in a separate follow-up? This doesn't block the fix here.

Also, the CI follow-up applies to both this PR and #3682. Adding `cargo test -p opentelemetry-proto --all-features --test json_serde `in either PR would ensure these regression tests run.

bruno-espino:
Thanks @lalitb! Reproduced the missing rejection-count error and opened #3713 to track defaults for the partial-success types separately. The explicit `json_serde` CI invocation is now included in #3682, so it will cover these regression tests as well once both PRs are merged.

PR #3685, opened 2026-09-03
feat(sdk): allow configuring all 5 ParentBased sampler branches

### LINKED ISSUE #3657: [Feature]: Allow configuring all 5 ParentBased sampler branches per the trace spec
### Related Problems?

The OpenTelemetry trace spec defines `ParentBased` as a composite sampler with 5
independently configurable delegate samplers:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#parentbased

  1. root                       — no parent exists
  2. remote_parent_sampled      — remote parent, sampled     (spec default: AlwaysOn)
  3. remote_parent_not_sampled  — remote parent, not sampled (spec default: AlwaysOff)
  4. local_parent_sampled       — local parent, sampled      (spec default: AlwaysOn)
  5. local_parent_not_sampled   — local parent, not sampled  (spec default: AlwaysOff)

Today, `Sampler::ParentBased(Box<dyn ShouldSample>)` in
opentelemetry-sdk/src/trace/sampler.rs only lets users configure case 1 (`root`).
Cases 2-5 are hardcoded to the spec defaults and cannot be overridden.

This blocks common sampling strategies, e.g.:
- Always keep spans under a sampled remote parent, but downsample local child spans
  to cut volume in a chatty internal call graph.
- Sample a low percentage of unsampled remote traffic, while always keeping
  everything once a trace is already flowing sampled through the service.

Not aware of an existing tracking issue in this repo for this gap.

I'd like the ParentBased sampler to expose all 5 branches as independently
configurable delegate samplers, e.g. via a builder:

    let sampler = Sampler::parent_based(root_sampler)
        .with_remote_parent_sampled(some_sampler)
        .with_remote_parent_not_sampled(some_sampler)
        .with_local_parent_sampled(some_sampler)
        .with_local_parent_not_sampled(some_sampler);

Any branch left unset should fall back to the current spec-mandated default
(AlwaysOn for *_sampled, AlwaysOff for *_not_sampled), so this is purely additive
and doesn't change behavior for existing `Sampler::ParentBased(...)` callers.

### Considered Alternatives

- Leave as-is and have users write a custom `ShouldSample` impl that reimplements
  all of ParentBased's parent/remote/sampled branching themselves just to override
  one branch — works, but means reimplementing spec logic that already exists in
  the SDK, and is easy to get subtly wrong (e.g. mishandling the local-vs-remote
  distinction).
- Only add configurability for the two "sampled" branches (the more commonly
  tuned pair) instead of all 4 — smaller surface area, but doesn't fully satisfy
  the spec and would need revisiting later if someone needs the "not_sampled"
  branches too.

### Additional Context

Other OTel SDKs (e.g. Java, Go, Python) already expose all 5 branches via their
ParentBased builders, so this would bring the Rust SDK to parity.

### Tip

darkraider01:
Fixes #3657.

ParentBased only let you configure the root case (no parent), the other 4 branches from the spec (remote/local parent, sampled/not sampled) were hardcoded to AlwaysOn/AlwaysOff with no way to override them.

Added Sampler::parent_based(root), which returns a ParentBasedSampler you can chain with with_remote_parent_sampled, with_remote_parent_not_sampled, with_local_parent_sampled and with_local_parent_not_sampled. Any branch you don't set falls back to the same spec default as before, so existing Sampler::ParentBased(...) usage is untouched, this is purely additive.

Didn't need a new Sampler enum variant for this. with_sampler already accepts anything implementing ShouldSample, so ParentBasedSampler just implements that trait directly and can be passed straight in.

Added two tests, one checking the defaults still match spec when nothing is overridden, one checking each of the 4 branches actually gets used when overridden. Ran the full sampler test suite plus clippy and fmt, all clean. Changelog updated.

### DIFF
--- a/opentelemetry-sdk/src/trace/sampler.rs
+++ b/opentelemetry-sdk/src/trace/sampler.rs
@@ -191,6 +191,125 @@ impl Sampler {
     {
         JaegerRemoteSamplerBuilder::new(runtime, http_client, default_sampler, service_name)
     }
+
+    /// Create a [`ParentBasedSampler`] with `root` as the sampler used for spans with
+    /// no parent.
+    ///
+    /// `Sampler::ParentBased` only lets you configure that one root case; the other 4
+    /// branches defined by the [ParentBased sampler spec] (remote/local parent,
+    /// sampled/not-sampled) are fixed to the spec defaults. Use this instead when you
+    /// need to override any of those, via `ParentBasedSampler`'s `with_*` methods.
+    ///
+    /// [ParentBased sampler spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#parentbased
+    pub fn parent_based(root: impl Into<Box<dyn ShouldSample>>) -> ParentBasedSampler {
+        ParentBasedSampler {
+            root: root.into(),
+            remote_parent_sampled: None,
+            remote_parent_not_sampled: None,
+            local_parent_sampled: None,
+            local_parent_not_sampled: None,
+        }
+    }
+}
+
+/// A [`ShouldSample`] sampler that lets each of the 5 branches defined by the
+/// [ParentBased sampler spec] be configured independently: the root sampler (no
+/// parent), and a sampler for each combination of remote/local parent and
+/// sampled/not-sampled. Any branch left unset falls back to the spec default
+/// (`Sampler::AlwaysOn` for the two `*_sampled` branches, `Sampler::AlwaysOff` for
+/// the two `*_not_sampled` branches).
+///
+/// Build one with [`Sampler::parent_based`].
+///
+/// [ParentBased sampler spec]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#parentbased
+#[derive(Clone, Debug)]
+pub struct ParentBasedSampler {
+    root: Box<dyn ShouldSample>,
+    remote_parent_sampled: Option<Box<dyn ShouldSample>>,
+    remote_parent_not_sampled: Option<Box<dyn ShouldSample>>,
+    local_parent_sampled: Option<Box<dyn ShouldSample>>,
+    local_parent_not_sampled: Option<Box<dyn ShouldSample>>,
+}
+
+impl ParentBasedSampler {
+    /// Sampler used when the parent's `SpanContext` is remote and sampled.
+    /// Defaults to `Sampler::AlwaysOn`.
+    pub fn with_remote_parent_sampled(mut self, sampler: impl Into<Box<dyn ShouldSample>>) -> Self {
+        self.remote_parent_sampled = Some(sampler.into());
+        self
+    }
+
+    /// Sampler used when the parent's `SpanContext` is remote and not sampled.
+    /// Defaults to `Sampler::AlwaysOff`.
+    pub fn with_remote_parent_not_sampled(
+        mut self,
+        sampler: impl Into<Box<dyn ShouldSample>>,
+    ) -> Self {
+        self.remote_parent_not_sampled = Some(sampler.into());
+        self
+    }
+
+    /// Sampler used when the parent's `SpanContext` is local and sampled.
+    /// Defaults to `Sampler::AlwaysOn`.
+    pub fn with_local_parent_sampled(mut self, sampler: impl Into<Box<dyn ShouldSample>>) -> Self {
+        self.local_parent_sampled = Some(sampler.into());
+        self
+    }
+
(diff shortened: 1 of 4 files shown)


opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -191,6 +191,125 @@ impl Sampler {
...
+        name: &str,
+        span_kind: &SpanKind,
+        attributes: &[KeyValue],
+        links: &[Link],
+    ) -> SamplingResult {
+        let delegate = |sampler: &dyn ShouldSample| {
+            sampler.should_sample(parent_context, trace_id, name, span_kind, attributes, links)
+        };
+
+        match parent_context.filter(|cx| cx.has_active_span()) {
```
TommyCpp:
```suggestion
match parent_context.filter(|cx| cx.span().span_context().is_valid()) {
```
We need to handle invalid but active span too here

darkraider01: thanks! Updated to filter on cx.span().span_context().is_valid() , and added unit tests covering both invalid active spans and empty contexts .

lalitb: Could we also fix `SdkTracer::build_with_context`? It still copies the invalid parent's trace ID, so the root fallback with `AlwaysOn` creates a sampled span with an all-zero trace ID. Please generate a valid ID before sampling and add a tracer-level test covering this.

opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -191,6 +191,138 @@ impl Sampler {
...
+}
+
+/// A [`ShouldSample`] sampler that lets each of the 5 branches defined by the
+/// [ParentBased sampler spec] be configured independently: the root sampler (no
+/// parent or an invalid parent), and a sampler for each combination of remote/local parent and
+/// sampled/not-sampled. Any branch left unset falls back to the spec default
+/// (`Sampler::AlwaysOn` for the two `*_sampled` branches, `Sampler::AlwaysOff` for
+/// the two `*_not_sampled` branches).
+///
+/// Build one with [`ParentBasedSampler::new`] or [`Sampler::parent_based`].
```
lalitb, line 222: Could we mention that this type preserves the delegate's attributes and modified trace state, while `Sampler::ParentBased` keeps only its decision? That difference matters when switching an existing custom root sampler to this API.

opentelemetry-sdk/src/trace/sampler.rs
```diff
@@ -466,4 +601,171 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn parent_based_sampler_respects_branch_overrides() {
+        let sampler = Sampler::parent_based(Sampler::AlwaysOff)
+            .with_remote_parent_sampled(Sampler::AlwaysOff)
+            .with_remote_parent_not_sampled(Sampler::AlwaysOn)
+            .with_local_parent_sampled(Sampler::AlwaysOff)
+            .with_local_parent_not_sampled(Sampler::AlwaysOn);
```
lalitb: Could we use distinguishable custom delegates here? Remote and local branches return identical decisions, so swapping them would still pass. A delegate that returns a unique attribute and trace state would let us check both branch selection and full result forwarding. Please also check that the root delegate receives the original parent context.

opentelemetry-sdk/src/trace/sampler.rs
darkraider01: Fixed `build_with_context` now filters the parent span on `is_valid()` before using it for `psc`/`trace_id`, so an invalid parent (e.g. from an unparsable traceparent) no longer gets its all-zero trace ID copied; the sampler now sees no parent and a fresh trace ID is generated. Added a tracer-level test (`invalid_parent_starts_a_new_trace`) covering this.

opentelemetry-sdk/src/trace/sampler.rs
darkraider01: Good point, added a note to the ParentBasedSampler doc comment that it forwards the delegate's SamplingResult unchanged (attributes + modified trace state), unlike Sampler::ParentBased which keeps only the decision.

opentelemetry-sdk/src/trace/sampler.rs
darkraider01: Done  replaced the AlwaysOn/AlwaysOff delegates with a TaggedSampler that tags its result with its own branch name and the parent span id it was handed, so remote/local mixups now fail. Also asserts the forwarded trace state and that the root delegate receives the original parent context.

lalitb (approved):
LGTM. Thanks.

PR #3688, opened 2026-09-03
refactor(otlp): seal exporter configuration traits

cijothomas:
## Summary

- seal `WithExportConfig`, `WithHttpConfig`, and `WithTonicConfig` with private supertraits
- keep all three traits public for importing and calling their existing configuration methods
- clarify the trait contract in rustdoc
- consolidate the changelog so it no longer instructs external `WithHttpConfig` implementations to add a method

These traits provide extension methods for OTLP-owned builders. External implementations have no access to the private builder configuration they are designed to modify, but leaving the traits open would make every future method addition a breaking change.

Normal exporter builder usage is unchanged. Only implementing these traits for downstream-owned types is no longer supported.


scottgerring (approved):
seems reasonable

PR #3689, opened 2026-09-03
refactor(otlp): internalize transport builders

cijothomas:
## Summary

- make `HttpExporterBuilder` and `TonicExporterBuilder` crate-private
- remove their root-level re-exports
- keep the public signal builders and `.with_http()` / `.with_tonic()` configuration paths unchanged
- document direct migrations for span, log, and metric exporters

The raw transport builders duplicate the supported signal-first API and expose transport implementation structure as part of the semver contract. After this change, exporters are consistently created through `SpanExporter::builder()`, `MetricExporter::builder()`, or `LogExporter::builder()`.

## Migration

| Previous API | Replacement |
| --- | --- |
| `HttpExporterBuilder::default().build_span_exporter()` | `SpanExporter::builder().with_http().build()` |
| `HttpExporterBuilder::default().build_log_exporter()` | `LogExporter::builder().with_http().build()` |
| `HttpExporterBuilder::default().build_metrics_exporter(temporality)` | `MetricExporter::builder().with_http().with_temporality(temporality).build()` |
| `TonicExporterBuilder::default()` | The corresponding signal builder followed by `.with_tonic()` |

Transport configuration continues to chain after selecting the transport. For example, custom HTTP client configuration migrates as follows:

```rust
// Before
let exporter = HttpExporterBuilder::default()
    .with_http_client(client)
    .with_endpoint(endpoint)
    .build_span_exporter()?;

// After
    .with_http()
    .with_http_client(client)
    .with_endpoint(endpoint)
    .build()?;
```

Headers, TLS configuration, metadata, interceptors, compression, retry policy, request body limits, endpoints, and timeouts remain available through the signal builders after `.with_http()` or `.with_tonic()`.

### Reusing shared HTTP configuration

Code that previously returned a concrete `HttpExporterBuilder` can instead use a generic helper to configure any HTTP-selected signal builder:

```rust
fn configure_http<B>(builder: B) -> B
where
    B: WithExportConfig + WithHttpConfig,
{
    builder
        .with_timeout(Duration::from_secs(5))
        .with_compression(Compression::Gzip)
}

let span_exporter = configure_http(SpanExporter::builder().with_http()).build()?;
let log_exporter = configure_http(LogExporter::builder().with_http()).build()?;
```

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -714,8 +714,12 @@ pub use crate::logs::{
 };
 
 #[cfg(any(feature = "http-proto", feature = "http-json"))]
+use crate::exporter::http::HttpExporterBuilder;
+#[cfg(any(feature = "http-proto", feature = "http-json"))]
 pub use crate::exporter::http::WithHttpConfig;
 
+#[cfg(feature = "grpc-tonic")]
+use crate::exporter::tonic::TonicExporterBuilder;
 #[cfg(feature = "grpc-tonic")]
 pub use crate::exporter::tonic::WithTonicConfig;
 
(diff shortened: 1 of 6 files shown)


scottgerring:
Previously, users could centralise common programmatic configuration in a function returning the concrete transport builder:

```rust
fn common_builder() -> HttpExporterBuilder {
    HttpExporterBuilder::default()
        .with_timeout(Duration::from_secs(5))
        .with_compression(Compression::Gzip)
}
```

The equivalent remains possible after this change using a generic configuration helper:

```rust
fn configure_http<B>(builder: B) -> B
where
    B: WithExportConfig + WithHttpConfig,
{
    builder
        .with_timeout(Duration::from_secs(5))
        .with_compression(Compression::Gzip)
}
```

It might be worth including this pattern in the migration guide. I've seen concrete builder factory functions like this in downstream projects, and their migration is less obvious than the inline builder examples.

opentelemetry-otlp/src/lib.rs
```diff
@@ -737,28 +741,18 @@ pub use retry_policy::RetryPolicy;
...
 pub struct NoExporterBuilderSet;
 
-/// Type to hold the [TonicExporterBuilder] and indicate it has been set.
-///
-/// Allowing access to [TonicExporterBuilder] specific configuration methods.
+/// Type indicating that the tonic transport has been selected.
 #[cfg(feature = "grpc-tonic")]
 // This is for clippy to work with only the grpc-tonic feature enabled
 #[allow(unused)]
 #[derive(Debug, Default)]
```
scottgerring, line 748: Since we're reducing the pre-stabilisation surface, should the transport marker types remain publicly Default? They've gotta remain nameable for the typestate API, but removing Default or hiding them from documentation would better steer users through .with_http() and .with_tonic()

lalitb (approved):
Thanks.

opentelemetry-otlp/src/lib.rs
cijothomas: yep, #3716 took care of this — both transport marker types no longer implement `Default`.

cijothomas:
yeah, good call. felt like a bit much for the changelog, so i added it to the PR description as an extra migration example.

PR #3692, opened 2026-09-04
refactor(otlp): finalize exporter build error API

### LINKED ISSUE #3691: Stabilization: finalize the OTLP exporter build error API
## Problem

`ExporterBuildError` is part of the public `opentelemetry-otlp` API, but its source still says that it needs to be refined and polished before stabilization.

Although the enum is `#[non_exhaustive]`, its existing variants and payload shapes remain part of the compatibility contract. Downstream crates can currently construct and match implementation-specific variants such as `InternalFailure`, `InvalidUri`, and `NoHttpClient`.

This prevents us from freely reorganizing exporter initialization errors after declaring the OTLP crate stable.

## Proposed direction

Before stabilization, deliberately choose one of these API models:

1. An opaque public error type backed by private implementation details; or
2. A finalized, stable taxonomy of programmatically matchable error kinds.

The resulting API should distinguish configuration errors that users can act on without exposing transport implementation details or relying on error message strings.

## Completion criteria

- Remove the existing refine-and-polish TODO.
- Document which error information is stable for programmatic use.
- Avoid exposing dependency-specific or implementation-specific error types.
- Provide migration guidance if existing public variants become inaccessible.

cijothomas:
Configuration parsing uses the opaque `ParseConfigError` introduced in #3742, which is now merged. This PR focuses on exporter-construction errors.

Exporter builders currently expose implementation-specific error variants that would become long-term public API commitments. This change replaces them with two exhaustive categories: `InvalidConfiguration(String)` for settings users can correct, and `InternalFailure(String)` for internal initialization failures.

Normal exporter builder usage is unchanged. The breaking changes affect code matching or constructing removed error variants.

Invalid endpoints, transport/protocol mismatches, missing HTTP clients, unavailable programmatic compression, and invalid TLS configuration map to `InvalidConfiguration`. HTTP-client initialization failures map to `InternalFailure`. Tonic endpoint diagnostics identify the originating setting, and tonic/TLS source chains are included in diagnostic text.

Environment enum parsing and reqwest client-construction handling are already on `main` through #3730 and #3724. This PR preserves those behaviors and the private tonic compression conversion from #3737.

### API decision

Following `docs/adr/001_error_handling.md`, the error set is intentionally closed and exhaustive. New construction failures should map to the existing categories; adding a category would be a breaking API change. Diagnostic strings are for troubleshooting and are not a stable programmatic interface. Runtime export failures continue to use `OTelSdkError`.

### Migration

| Previous variant | Replacement |
| --- | --- |
| `InvalidUri`, `InvalidConfig`, `NoHttpClient` | `InvalidConfiguration` |
| `UnsupportedCompressionAlgorithm` | `InvalidConfiguration` when building |
| `FeatureRequiredForCompressionAlgorithm` | `InvalidConfiguration` for programmatic configuration |
| `ThreadSpawnFailed` | `InternalFailure` |

Match the two categories directly:

```rust
match error {
    ExporterBuildError::InvalidConfiguration(message) => {
        eprintln!("fix the exporter configuration: {message}");
    }
    ExporterBuildError::InternalFailure(message) => {
        eprintln!("exporter initialization failed: {message}");
    }
}
```

Code that propagates build errors with `?` without inspecting their variants needs no changes. Invalid or unavailable enum environment values continue to warn and fall back rather than return a build error.

Closes #3691

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -125,6 +125,8 @@ release:
   default. Failures while constructing a default reqwest client now return
   `InternalFailure` instead of silently falling back to a differently
   configured client.
+  Tonic endpoint errors identify the originating environment variable when
+  validating the URI or reporting endpoint-related TLS setup failures.
   [#3691](https://github.com/open-telemetry/opentelemetry-rust/issues/3691)
 - Return an exporter build error for invalid OTLP/HTTP endpoint environment
   variables instead of silently falling back to another endpoint or localhost.
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -87,6 +87,14 @@ release:
   `.with_http()` or `.with_tonic()`.
 - **Breaking** Removed the deprecated `tls` feature alias. Replace `tls` with
   `tls-ring`, or select `tls-aws-lc` or `tls-provider-agnostic` explicitly.
+- **Breaking:** Simplified `ExporterBuildError` to the exhaustive
+  `InvalidConfiguration(String)` and `InternalFailure(String)` variants.
+  Configuration errors such as invalid endpoints, missing HTTP clients,
+  unsupported compression, and missing compression features now use
+  `InvalidConfiguration`. Match one of these categories instead of the
+  previous implementation-specific variants. `Compression::from_str` now
+  returns `ParseCompressionError`.
+  [#3691](https://github.com/open-telemetry/opentelemetry-rust/issues/3691)
 - Return an exporter build error for invalid OTLP/HTTP endpoint environment
   variables instead of silently falling back to another endpoint or localhost.
   Empty endpoint environment variables are now treated as unset.
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -89,11 +89,13 @@ release:
   `tls-ring`, or select `tls-aws-lc` or `tls-provider-agnostic` explicitly.
 - **Breaking:** Simplified `ExporterBuildError` to the exhaustive
   `InvalidConfiguration(String)` and `InternalFailure(String)` variants.
+  The enum is no longer marked `#[non_exhaustive]`.
   Configuration errors such as invalid endpoints, missing HTTP clients,
   unsupported compression, and missing compression features now use
-  `InvalidConfiguration`. Match one of these categories instead of the
-  previous implementation-specific variants. `Compression::from_str` now
-  returns `ParseCompressionError`.
+  `InvalidConfiguration`. Invalid protocol environment variables now return
+  `InvalidConfiguration` instead of falling back to another protocol. Match
+  one of these categories instead of the previous implementation-specific
+  variants. `Compression::from_str` now returns `ParseCompressionError`.
   [#3691](https://github.com/open-telemetry/opentelemetry-rust/issues/3691)
 - Return an exporter build error for invalid OTLP/HTTP endpoint environment
   variables instead of silently falling back to another endpoint or localhost.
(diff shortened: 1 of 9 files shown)


opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -90,66 +90,67 @@ pub(crate) struct ExportConfig {
...
     provided_protocol: Option<Protocol>,
-) -> Protocol {
+) -> Result<Protocol, ExporterBuildError> {
     if let Some(protocol) = provided_protocol {
-        return protocol;
+        return Ok(protocol);
     }
-    if let Some(protocol) = Protocol::parse_from_env_var(signal_protocol_var) {
-        return protocol;
+    if let Some(protocol) = Protocol::parse_from_env_var(signal_protocol_var)? {
```
scottgerring, line 97: This shouldn't error out when it is unable to parse the environment variable. [The spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#enum) says:

> For sources accepting an enum value, if the user provides a value the implementation does not recognize, the implementation MUST generate a warning and gracefully ignore the setting.

Adding `?` here, an invalid signal-specific value prevents resolution from continuing to the generic protocol value or feature default. 

The same section also says enum values SHOULD be interpreted case-insensitively, which does not appear to happen here, and i've just noticed we're not doing that in Protocol::parse_from_env_var

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -664,28 +719,73 @@ mod tests {
...
+    #[test]
+    fn test_invalid_signal_protocol_does_not_fall_back_to_generic() {
+        run_env_test(
+            vec![
+                (crate::OTEL_EXPORTER_OTLP_TRACES_PROTOCOL, "invalid"),
+                (crate::OTEL_EXPORTER_OTLP_PROTOCOL, "invalid-generic"),
+            ],
+            || {
+                let error =
+                    super::resolve_protocol(crate::OTEL_EXPORTER_OTLP_TRACES_PROTOCOL, None)
```
scottgerring, line 764: Could we test this user-visible behaviour through the public exporter builder rather than calling the private `resolve_protocol` helper directly? For example, set the environment variables and call `SpanExporter::builder().build()`, then assert the intended fallback behaviour - that would verify the complete path through signal-specific precedence and transport selection, rather than only the helper implementation.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -34,6 +34,16 @@ release:
 
 ### Other changes
 
+- **Breaking:** Simplified `ExporterBuildError` to the exhaustive
+  `InvalidConfiguration(String)` and `InternalFailure(String)` variants.
+  The enum is no longer marked `#[non_exhaustive]`.
+  Configuration errors such as invalid endpoints, missing HTTP clients,
+  unsupported compression, and missing compression features now use
+  `InvalidConfiguration`. Invalid protocol environment variables now return
+  `InvalidConfiguration` instead of falling back to another protocol. Match
```
scottgerring, line 43:
```suggestion
Match
```

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -34,6 +34,16 @@ release:
 
 ### Other changes
 
+- **Breaking:** Simplified `ExporterBuildError` to the exhaustive
```
scottgerring, line 37: Could we include a short before/after example of matching `ExporterBuildError` here? Showing an old match and its replacement with `InvalidConfiguration` and `InternalFailure` would make the changelog migration self-contained.

scottgerring (approved):
LGTM once out of draft

scottgerring (approved):
LGTM; thanks for addressing feedback

PR #3699, opened 2026-09-04
refactor(http): remove unused hyper body type

scottgerring:
Remove the unused public `opentelemetry_http::hyper::Body` wrapper and use `http_body_util::Full<Bytes>` directly. This also removes its unsafe pin projection. I can't see any usages of this via the github search and it's unnecessarily available through our public API. 

Validated with all-feature and Hyper-only `opentelemetry-http` tests, clippy, and the OTLP Hyper client check.


cijothomas (approved):

cijothomas (approved):
Looks good. The wrapper adds no behavior over `Full<Bytes>`, and removing it keeps the public HTTP surface smaller before stabilization.

lalitb (approved):

scottgerring:
fixed conflict; good to merge

PR #3701, opened 2026-09-04

scottgerring:
## Changes

Seal the `ResponseExt` trait in `opentelemetry-http` so it cannot be implemented by downstream crates.

The trait provides a single convenience method (`error_for_status`) with a blanket impl for `http::Response<T>`. There is no reason for external code to implement this trait -- they only need to call its method. Sealing it gives us freedom to evolve the trait (add methods with default impls, change signatures) without breaking downstream.

### Implementation

### Impact

- **Calling code is unaffected** -- `response.error_for_status()` continues to work.
- **No one can `impl ResponseExt for MyType`** any more (this was never useful since the blanket impl already covers all `Response<T>`).

## Merge requirement

None -- this is a non-breaking API refinement (restricting implementors of a trait that only had a blanket impl).

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -301,8 +301,15 @@ pub mod hyper {
     }
 }
 
+mod private {
+    pub trait Sealed {}
+    impl<T> Sealed for http::Response<T> {}
+}
+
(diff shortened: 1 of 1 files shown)


scottgerring:
Closing in favour of a new PR with the updated branch (force-push blocked by branch protection).

PR #3704, opened 2026-09-04
fix(otlp): use spec-compliant values for Protocol and Compression serde

scottgerring:
## Changes

The `serialize` feature's derived `Serialize`/`Deserialize` implementations for `Protocol` and `Compression` were using Rust variant names instead of the spec-defined values.

**Before**
| Variant | Serialised as |
|---|---|
| `Protocol::Grpc` | `"Grpc"` |
| `Protocol::HttpBinary` | `"HttpBinary"` |
| `Protocol::HttpJson` | `"HttpJson"` |
| `Compression::Gzip` | `"Gzip"` |
| `Compression::Zstd` | `"Zstd"` |

**After**

| Variant | Serialised as |
|---|---|
| `Protocol::Grpc` | `"grpc"` |
| `Protocol::HttpBinary` | `"http/protobuf"` |
| `Protocol::HttpJson` | `"http/json"` |
| `Compression::Gzip` | `"gzip"` |
| `Compression::Zstd` | `"zstd"` |

These now match the values used in `OTEL_EXPORTER_OTLP_PROTOCOL` and `OTEL_EXPORTER_OTLP_COMPRESSION` environment variables, as well as the existing `Display`/`FromStr` implementations.

## Implementation

- Add `#[serde(rename = ...)]` to each `Protocol` variant (per-variant, since the spec values aren't a simple case transformation)
- Add `#[serde(rename_all = "lowercase")]` to `Compression`
- All serde attributes are gated behind `#[cfg_attr(feature = "serialize", ...)]`

## Breaking change

This is a **breaking change for the `serialize` feature only**. Anyone deserialising `Protocol` or `Compression` from previously-serialised data using the old Rust variant names will need to migrate their stored data. The `serialize` feature is not widely used and the old values were clearly incorrect, so the migration burden should be minimal.

## Tests

Added 5 roundtrip tests covering all variants of both enums.

### DIFF
--- a/opentelemetry-otlp/src/exporter/mod.rs
+++ b/opentelemetry-otlp/src/exporter/mod.rs
@@ -152,6 +152,7 @@ pub enum ExporterBuildError {
 
 /// The compression algorithm to use when sending data.
 #[cfg_attr(feature = "serialize", derive(Deserialize, Serialize))]
+#[cfg_attr(feature = "serialize", serde(rename_all = "lowercase"))]
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum Compression {
     /// Compresses data using gzip.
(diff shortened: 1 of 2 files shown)


scottgerring:
Superseded — removing the feature entirely instead of fixing the serde representations. See replacement PR.

PR #3706, opened 2026-09-04
chore(otlp): remove Default from typestate marker types

scottgerring:
## Changes

Remove derived `Default` implementations from the typestate marker types
`TonicExporterBuilderSet` and `HttpExporterBuilderSet`, and remove the
entirely unused `NoExporterConfig` type.

### `Default` removal

These types are typestate markers that enforce the exporter builder flow:

```
builder() → with_tonic() / with_http() → build()
```

Having `Default` on them means a user could bypass the intended API:
```rust
// Before: this compiles, skipping the with_tonic() step
let exporter = SpanExporterBuilder::<TonicExporterBuilderSet>::default().build()?;
```

After this change, users must go through the intended builder path.

`NoExporterBuilderSet` **keeps** its `Default` since it is the initial state
and the exporter builders (`SpanExporterBuilder`, `MetricExporterBuilder`,
`LogExporterBuilder`) derive `Default` on the generic struct, which requires
the type parameter to be `Default`.

### `NoExporterConfig` removal

`NoExporterConfig` was `#[doc(hidden)]` and not referenced anywhere in the
crate. Removed as dead code.

### DIFF
--- a/opentelemetry-otlp/src/lib.rs
+++ b/opentelemetry-otlp/src/lib.rs
@@ -743,14 +743,14 @@ pub struct NoExporterBuilderSet;
 #[cfg(feature = "grpc-tonic")]
 // This is for clippy to work with only the grpc-tonic feature enabled
 #[allow(unused)]
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct TonicExporterBuilderSet(TonicExporterBuilder);
 
 /// Type to hold the [HttpExporterBuilder] and indicate it has been set.
 ///
 /// Allowing access to [HttpExporterBuilder] specific configuration methods.
 #[cfg(any(feature = "http-proto", feature = "http-json"))]
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct HttpExporterBuilderSet(HttpExporterBuilder);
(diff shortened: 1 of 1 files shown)


opentelemetry-otlp/src/lib.rs
```diff
@@ -743,14 +743,14 @@ pub struct NoExporterBuilderSet;
 #[cfg(feature = "grpc-tonic")]
 // This is for clippy to work with only the grpc-tonic feature enabled
 #[allow(unused)]
-#[derive(Debug, Default)]
+#[derive(Debug)]
```
lalitb, line 746: Could we add a breaking-change entry to the OTLP changelog? Removing these derives also removes `Default` from the transport-selected span, log, and metric builders. The PR description already has the migration example, so that could go into the changelog along with the `NoExporterConfig` removal.

lalitb (approved):
Thanks for the cleanup. LGTM once the changelog is updated for breaking changes.

scottgerring:
Closing in favour of #3716 - cannot force-push to branches on origin. The new PR includes the breaking changelog entry requested by @lalitb.

PR #3707, opened 2026-09-04

scottgerring:
## Changes

**1. Seal the `ResponseExt` trait** in `opentelemetry-http` so it cannot be implemented by downstream crates.

The trait provides a single convenience method (`error_for_status`) with a blanket impl for `http::Response<T>`. There is no reason for external code to implement this trait -- they only need to call its method. Sealing it gives us freedom to evolve the trait (add methods with default impls, change signatures) without breaking downstream.

**Impact:**
- Calling code is unaffected -- `response.error_for_status()` continues to work.
- No one can `impl ResponseExt for MyType` any more (this was never useful since the blanket impl already covers all `Response<T>`).

**2. Fix pre-existing flaky test** in `opentelemetry-otlp`.

`metric::build_tests::build_with_default_transport` could fail when the parallel `invalid_env_var_returns_error` test set `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` to `"invalid"`. Fixed by wrapping the test body with `temp_env::with_var_unset` to isolate it from env var pollution.

Supersedes #3701 (branch protection blocked force-push).

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -301,8 +301,15 @@ pub mod hyper {
     }
 }
 
+mod private {
+    pub trait Sealed {}
+    impl<T> Sealed for http::Response<T> {}
+}
+
 /// Methods to make working with responses from the [`HttpClient`] trait easier.
-pub trait ResponseExt: Sized {
+///
+/// This trait is sealed and cannot be implemented outside of this crate.
+pub trait ResponseExt: private::Sealed + Sized {
     /// Turn a response into an error if the HTTP status does not indicate success (200 - 299).
     fn error_for_status(self) -> Result<Self, HttpError>;
 }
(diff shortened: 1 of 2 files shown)


opentelemetry-http/src/lib.rs
```diff
@@ -301,8 +301,15 @@ pub mod hyper {
...
+mod private {
+    pub trait Sealed {}
+    impl<T> Sealed for http::Response<T> {}
+}
+
 /// Methods to make working with responses from the [`HttpClient`] trait easier.
-pub trait ResponseExt: Sized {
+///
+/// This trait is sealed and cannot be implemented outside of this crate.
+pub trait ResponseExt: private::Sealed + Sized {
```
lalitb, line 312: Could we add it as a **Breaking** change in changelog? Downstream implementations such as `impl ResponseExt for MyResponse` will stop compiling after this change. Calls to `error_for_status()` on `http::Response<T>` remain unchanged.

lalitb (approved):
LGTM once the changelog entry is added. Thanks.

scottgerring:
Closing in favour of #3715 - cannot force-push to branches on origin. The new PR includes the changelog entry requested by @lalitb.

PR #3708, opened 2026-09-04
refactor(otlp)!: remove the `serialize` feature

scottgerring:
## Changes

### What's removed

## Breaking change

Users who depend on `opentelemetry-otlp` with `features = ["serialize"]` will get a compilation error. The migration is to remove the feature flag and, if serde support is needed, implement it in application code using the existing `Display`/`FromStr` impls.

Supersedes #3704.


scottgerring:
Reopening with updated branch that includes changelog entry.

PR #3709, opened 2026-09-04
refactor(otlp)!: remove the `serialize` feature

scottgerring:
## Changes

### What's removed

## Breaking change

Users who depend on `opentelemetry-otlp` with `features = ["serialize"]` will get a compilation error. The migration is to remove the feature flag and, if serde support is needed, implement it in application code using the existing `Display`/`FromStr` impls.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -52,6 +52,15 @@ release:
   Endpoints with an explicit scheme (e.g., `http://`, `https://`, `unix://`) are unaffected.
   [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
   [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
+- **Breaking** Removed the `serialize` feature flag. This feature gated
+  `Serialize`/`Deserialize` derives on `Protocol` and `Compression`, but the
+  derived representations were incorrect (Rust variant names instead of spec
+  values) and the feature had no known consumers. The equivalent feature was
+  removed from the core `opentelemetry` crate in 2022.
+  **Migration**: Remove `serialize` from your feature list. If you need serde
+  support for these types, implement it in your own code using the existing
+  `Display`/`FromStr` impls.
+  [#3708](https://github.com/open-telemetry/opentelemetry-rust/pull/3708)
 - **Breaking** Removed `reqwest-rustls-webpki-roots` feature. The `webpki-roots` cargo feature was
   removed from `reqwest` in v0.13.0, making this feature broken for anyone resolving `reqwest >= 0.13.0`.
   **Migration**: Use `reqwest-rustls` instead (now correctly uses `reqwest/rustls` with platform native
--- a/opentelemetry-otlp/Cargo.toml
+++ b/opentelemetry-otlp/Cargo.toml
@@ -40,7 +40,6 @@ tokio = { workspace = true, features = ["sync", "rt", "time"], optional = true }
 reqwest = { workspace = true, optional = true }
 http = { workspace = true, optional = true }
 httpdate = { version = "1.0.3", optional = true }
-serde = { workspace = true, features = ["derive"], optional = true }
 thiserror = { workspace = true }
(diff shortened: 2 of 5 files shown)


opentelemetry-otlp/allowed-external-types.toml
```diff
@@ -5,12 +5,6 @@
 allowed_external_types = [
     "opentelemetry_http::*",
     "opentelemetry_sdk::*",
-    # serde
```
cijothomas, line 8: While reviewing your other PR, I was asking why would we need this! thanks  for another cleanup!

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -52,6 +52,15 @@ release:
   Endpoints with an explicit scheme (e.g., `http://`, `https://`, `unix://`) are unaffected.
   [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
   [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
+- **Breaking** Removed the `serialize` feature flag. This feature gated
+  `Serialize`/`Deserialize` derives on `Protocol` and `Compression`, but the
+  derived representations were incorrect (Rust variant names instead of spec
+  values) and the feature had no known consumers. The equivalent feature was
+  removed from the core `opentelemetry` crate in 2022.
+  **Migration**: Remove `serialize` from your feature list. If you need serde
+  support for these types, implement it in your own code using the existing
```
cijothomas, line 61: I think this migration bit needs a tweak. Users can’t implement serde’s traits directly for `Protocol` or `Compression` because they don’t own either the traits or the types, and `Protocol` doesn’t currently implement `Display` or `FromStr` anyway.

Maybe say something like: “Remove `serialize` from your feature list. If these values are part of serializable app config, define a local config enum or wrapper and convert it to `Protocol` or `Compression` when building the exporter.” The declarative config work in `opentelemetry-config` also feels like the natural future home for the serde-facing types.

The PR body has the same migration wording, and this link should be #3709 rather than the superseded #3708.

opentelemetry-otlp/Cargo.toml
```diff
@@ -40,7 +40,6 @@ tokio = { workspace = true, features = ["sync", "rt", "time"], optional = true }
 reqwest = { workspace = true, optional = true }
 http = { workspace = true, optional = true }
 httpdate = { version = "1.0.3", optional = true }
-serde = { workspace = true, features = ["derive"], optional = true }
 thiserror = { workspace = true }
```
cijothomas, line 44: Tiny thing: this also removes the implicit `serde` feature, not just `serialize`. Probably worth calling that out in the changelog too, so anyone using either name knows to remove both.

opentelemetry-otlp/CHANGELOG.md
```diff
@@ -52,6 +52,15 @@ release:
   Endpoints with an explicit scheme (e.g., `http://`, `https://`, `unix://`) are unaffected.
   [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
   [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
+- **Breaking** Removed the `serialize` feature flag. This feature gated
+  `Serialize`/`Deserialize` derives on `Protocol` and `Compression`, but the
+  derived representations were incorrect (Rust variant names instead of spec
+  values) and the feature had no known consumers. The equivalent feature was
```
cijothomas, line 58: “No known consumers” feels a little stronger than we can really know here. Maybe just drop that bit and say the feature only covered these two enums?

opentelemetry-otlp/CHANGELOG.md
scottgerring: > because they don’t own either the traits or the types 

Ah yes 🤦 
as i accidentally pushed the original PR to `origin` and not my fork I can no longer add changes to it (I wonder if we can fix this one day? it seems to be some branch protection issue), so I have addressed this and re-opened in #3711 .

scottgerring:
@cijothomas  Replaced with #3711 and addressed issues there as I can no longer push to this accidentally `origin` pushed branch

PR #3711, opened 2026-09-07
refactor(otlp)!: remove the serialize feature

scottgerring:
> [!NOTE]
> Supersedes #3709 which was mistakenly pushed to origin and can therefore no longer be updated 

## Changes

### What's removed

## Breaking change

Users who depend on `opentelemetry-otlp` with `features = ["serialize"]` will get a compilation error.

**Migration**: Remove `serialize` (and `serde`, if listed) from your feature list. If these values are part of serialisable app config, define a local config enum or wrapper and convert it to `Protocol` or `Compression` when building the exporter.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -72,15 +72,17 @@ release:
   Endpoints with an explicit scheme (e.g., `http://`, `https://`, `unix://`) are unaffected.
   [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
   [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
-- **Breaking** Removed the `serialize` feature flag. This feature gated
-  `Serialize`/`Deserialize` derives on `Protocol` and `Compression`, but the
-  derived representations were incorrect (Rust variant names instead of spec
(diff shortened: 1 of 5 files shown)


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -72,6 +72,17 @@ release:
   Endpoints with an explicit scheme (e.g., `http://`, `https://`, `unix://`) are unaffected.
   [#774](https://github.com/open-telemetry/opentelemetry-rust/issues/774)
   [#984](https://github.com/open-telemetry/opentelemetry-rust/issues/984)
+- **Breaking** Removed the `serialize` feature flag and its implicit `serde`
```
lalitb, line 75: nit - add PR number.

opentelemetry-otlp/src/lib.rs
```diff
@@ -268,7 +268,6 @@
 //! * `logs`: Includes the logs exporters.
 //!
 //! The following feature flags generate additional code and types:
```
lalitb: nit - remove this line too ?

lalitb (approved):
LGTM. Thanks. Please also copy the corrected migration guidance from the changelog into the PR description, since `Protocol` doesn’t implement `Display` or `FromStr`.

PR #3715, opened 2026-09-08

scottgerring:
> [!NOTE]
> This replaces #3707 (approved by @lalitb) because we cannot update branches pushed to origin.

## Changes

**1. Seal the `ResponseExt` trait** in `opentelemetry-http` so it cannot be implemented by downstream crates.

The trait provides a single convenience method (`error_for_status`) with a blanket impl for `http::Response<T>`. There is no reason for external code to implement this trait - they only need to call its method. Sealing it gives us freedom to evolve the trait (add methods with default impls, change signatures) without breaking downstream.

**Impact:**
- Calling code is unaffected - `response.error_for_status()` continues to work.
- No one can `impl ResponseExt for MyType` any more (this was never useful since the blanket impl already covers all `Response<T>`).

**2. Fix pre-existing flaky test** in `opentelemetry-otlp`.

`metric::build_tests::build_with_default_transport` could fail when the parallel `invalid_env_var_returns_error` test set `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` to `"invalid"`. Fixed by wrapping the test body with `temp_env::with_var_unset` to isolate it from env var pollution.

Supersedes #3701 (branch protection blocked force-push).

### DIFF
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## vNext
 
+- **Breaking** Sealed the `ResponseExt` trait so it can no longer be implemented by
+  downstream crates. The trait provides a blanket implementation for all
+  `http::Response<T>` types, so calling code is unaffected -- only
+  `impl ResponseExt for MyType` will stop compiling. If you have a custom
+  implementation, remove it and rely on the blanket impl instead.
+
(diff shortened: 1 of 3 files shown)


scottgerring:
@lalitb this replaces #3707 which you approved yesterday; i've updated changelog as requested. Forgot to push a few of these branches to fork instead of origin ...

cijothomas (approved):

PR #3717, opened 2026-09-08
docs(otlp): fix stale with_protocol docs and complete gRPC trait summary

scottgerring:
> [!NOTE]
> This replaces #3703 (approved by @lalitb) because we cannot update branches pushed to origin.

## Changes

**`WithExportConfig::with_protocol` doc** (exporter/mod.rs):

The doc comment claimed "All exporters in this crate only support one protocol, thus choosing the protocol is a no-op at the moment." This has been stale since the typestate builder was introduced: protocol selection now matters for HTTP sub-protocol choice (`HttpBinary` vs `HttpJson`) and is validated against the chosen transport (setting `Grpc` on an HTTP builder returns an error).

Replaced with accurate documentation that explains:
- The primary use case (HTTP sub-protocol selection)
- What happens on transport mismatch (build error)
- How default protocol depends on enabled cargo features
- That `with_protocol()` is only available after transport selection, and is not involved in the auto-select builder path
- Precedence over environment variables

**gRPC config trait summary** (lib.rs):

The `WithTonicConfig` method listing in the "gRPC (tonic) - all configuration options" section was missing `with_retry_policy`, which was added as part of the retry stabilisation (#3672). Added it to keep the summary complete.

### DIFF
--- a/opentelemetry-otlp/src/exporter/mod.rs
+++ b/opentelemetry-otlp/src/exporter/mod.rs
@@ -291,13 +291,25 @@ pub trait WithExportConfig {
     ///
     /// Note: Programmatically setting this will override any value set via the environment variable.
     fn with_endpoint<T: Into<String>>(self, endpoint: T) -> Self;
-    /// Set the protocol to use when communicating with the collector.
+    /// Set the transport protocol to use when communicating with the collector.
     ///
-    /// Note that protocols that are not supported by exporters will be ignored. The exporter
-    /// will use default protocol in this case.
+    /// This is mainly useful on the HTTP transport to choose between
+    /// [`Protocol::HttpBinary`] (protobuf) and [`Protocol::HttpJson`].
+    /// Which of these is the default depends on which cargo features are
+    /// enabled (priority: `http-json` > `http-proto` > `grpc-tonic`).
+    /// Setting a protocol that conflicts with the chosen transport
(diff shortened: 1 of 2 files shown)


opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -291,13 +291,25 @@ pub trait WithExportConfig {
...
     fn with_endpoint<T: Into<String>>(self, endpoint: T) -> Self;
-    /// Set the protocol to use when communicating with the collector.
+    /// Set the transport protocol to use when communicating with the collector.
     ///
-    /// Note that protocols that are not supported by exporters will be ignored. The exporter
-    /// will use default protocol in this case.
+    /// This is mainly useful on the HTTP transport to choose between
+    /// [`Protocol::HttpBinary`] (protobuf) and [`Protocol::HttpJson`].
+    /// Which of these is the default depends on which cargo features are
+    /// enabled (priority: `http-json` > `http-proto` > `grpc-tonic`).
```
cijothomas: this priority ordering could be confusing.. I think its best to leave it out  of the `with_protocol` method documentation.

scottgerring: fair call; addressed

cijothomas (approved):

PR #3718, opened 2026-09-09
feat(trace): support the W3C random trace ID flag

### LINKED ISSUE #3270: [Feature]: Support the new W3C random flag.
### Related Problems?

Have SpanContext support the new [W3C random flag](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#traceid-randomness).

Have SpanContext support the new [W3C random flag](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#traceid-randomness).

### Considered Alternatives

_No response_

### Additional Context

_No response_

### Tip

ayush-that:
Fixes #3270

## Changes

Random trace IDs can now wear their little badge.

Adds RANDOM helpers, generator opt-in, SDK flag handling, and W3C propagation. Custom generators default to false; RandomIdGenerator opts in. Children retain the flag even when dropped.

B3's private deferred bit moves out of the way. Local drops clear it, and OTLP conversions filter private bits while preserving W3C flags.

Tests, Clippy, formatting, docs, and all precommit stages passed.

- [x] Unit tests added/updated (if applicable)
- [x] Appropriate CHANGELOG.md files updated for non-trivial, user-facing changes
- [ ] Changes in public API reviewed (if applicable)

### DIFF
--- a/opentelemetry-sdk/src/trace/id_generator/mod.rs
+++ b/opentelemetry-sdk/src/trace/id_generator/mod.rs
@@ -10,6 +10,21 @@ pub trait IdGenerator: Send + Sync + fmt::Debug {
 
     /// Generate a new `SpanId`
     fn new_span_id(&self) -> SpanId;
+
+    /// Returns `true` if every trace ID returned by [`new_trace_id`] has at least its
+    /// right-most 7 bytes generated randomly or pseudo-randomly with uniform
+    /// distribution, as required by
+    /// [W3C Trace Context Level 2](https://www.w3.org/TR/trace-context-2/#random-trace-id-flag).
+    ///
+    /// When this returns `true`, the SDK sets [`TraceFlags::RANDOM`] on the span
+    /// context of every root span. The flag is a positive assertion about the
+    /// trace ID, so this defaults to `false` and implementations must opt in explicitly.
+    ///
+    /// [`new_trace_id`]: IdGenerator::new_trace_id
+    /// [`TraceFlags::RANDOM`]: opentelemetry::trace::TraceFlags::RANDOM
+    fn is_random(&self) -> bool {
+        false
+    }
 }
 
 /// Default [`IdGenerator`] implementation.
@@ -27,6 +42,38 @@ impl IdGenerator for RandomIdGenerator {
 
     fn new_span_id(&self) -> SpanId {
         CURRENT_RNG.with(|rng| SpanId::from(rng.borrow_mut().random::<u64>()))
+    }
+
+    fn is_random(&self) -> bool {
+        true
+    }
+}
+
+#[cfg(all(test, feature = "testing", feature = "trace"))]
+mod tests {
+    use super::*;
+
+    #[derive(Debug)]
+    struct CounterIdGenerator;
+
+    impl IdGenerator for CounterIdGenerator {
+        fn new_trace_id(&self) -> TraceId {
+            TraceId::from(1)
+        }
+
+        fn new_span_id(&self) -> SpanId {
+            SpanId::from(1)
+        }
+    }
+
+    #[test]
+    fn id_generator_is_random_defaults_to_false() {
+        assert!(!CounterIdGenerator.is_random());
+    }
+
+    #[test]
+    fn random_id_generator_is_random() {
+        assert!(RandomIdGenerator::default().is_random());
     }
 }
 
--- a/opentelemetry-sdk/src/trace/tracer.rs
+++ b/opentelemetry-sdk/src/trace/tracer.rs
@@ -187,6 +187,7 @@ impl opentelemetry::trace::Tracer for SdkTracer {
         let config = provider.config();
         let span_id = config.id_generator.new_span_id();
         let trace_id;
+        let trace_flags;
         let mut psc = &SpanContext::empty_context();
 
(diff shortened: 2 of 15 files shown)


opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -250,8 +252,13 @@ impl opentelemetry::trace::Tracer for SdkTracer {
                 )
             }
             SamplingDecision::Drop => {
-                let span_context =
-                    SpanContext::new(trace_id, span_id, TraceFlags::default(), false, trace_state);
+                let span_context = SpanContext::new(
+                    trace_id,
+                    span_id,
+                    trace_flags.with_sampled(false),
```
lalitb: This also preserves B3's deferred marker after the sampler returns `Drop`. In `opentelemetry-zipkin/src/propagator/mod.rs`, `inject_context()` sees that marker and omits the sampling decision. Previously, this branch cleared the marker and B3 sent sampled `0`.

Could we keep `RANDOM` while ensuring B3 still sends `0` after a local drop? Please add a test covering a deferred B3 parent -> SDK span with `AlwaysOff` -> B3 injection.

opentelemetry-zipkin/src/propagator/mod.rs
```diff
@@ -31,7 +31,7 @@ const B3_SPAN_ID_HEADER: &str = "x-b3-spanid";
 const B3_SAMPLED_HEADER: &str = "x-b3-sampled";
 const B3_PARENT_SPAN_ID_HEADER: &str = "x-b3-parentspanid";
 
-const TRACE_FLAG_DEFERRED: TraceFlags = TraceFlags::new(0x02);
+const TRACE_FLAG_DEFERRED: TraceFlags = TraceFlags::new(0x80);
```
lalitb, line 34: This marker is private to B3, but the OTLP conversions in `transform/logs.rs` and `transform/trace.rs` copy trace flags without filtering it out. Since dropped spans now keep the marker, logs emitted under those spans and links to them can include `0x80` in OTLP.

Could we filter out B3-only bits during OTLP conversion, while keeping them in the context for B3 propagation? Please add log and link tests for this.

opentelemetry-sdk/src/trace/tracer.rs
```diff
@@ -195,12 +196,14 @@ impl opentelemetry::trace::Tracer for SdkTracer {
...
 
-        // Build context for sampling decision
+        // Inherit flags with the trace ID, or derive them from its generator.
         if let Some(sc) = parent_span.as_ref().map(|parent| parent.span_context()) {
             trace_id = sc.trace_id();
+            trace_flags = sc.trace_flags();
             psc = sc;
         } else {
             trace_id = config.id_generator.new_trace_id();
+            trace_flags = TraceFlags::default().with_random(config.id_generator.is_random());
```
lalitb, line 203: #3685 now fixes invalid-parent handling here. When combining these changes, please keep that fix and check that an invalid parent starts a new trace with `RANDOM` set when the ID generator returns `true`.

opentelemetry-zipkin/src/propagator/mod.rs
ayush-that: OTLP conversions now strip private bits without changing the context.
I have added span, link, and log tests.

opentelemetry-sdk/src/trace/tracer.rs
ayush-that: I've kept the validity check from #3685 and added a test for both generator settings. It will check the fresh trace ID, RANDOM flag, and cleared parent span ID.

opentelemetry-sdk/src/trace/tracer.rs
ayush-that: Local drops now keep only RANDOM.
I've added the required tests.

lalitb (commented):
LGTM. Although the SDK randomness extension is still marked Development in the spec, Python and Java already expose it in their stable SDKs, so I’m comfortable following the same approach here, unless other @open-telemetry/rust-maintainers / @open-telemetry/rust-approvers  have any concerns.

lalitb (approved):

opentelemetry-sdk/src/trace/id_generator/mod.rs
```diff
@@ -10,6 +10,21 @@ pub trait IdGenerator: Send + Sync + fmt::Debug {
...
+    /// [W3C Trace Context Level 2](https://www.w3.org/TR/trace-context-2/#random-trace-id-flag).
+    ///
+    /// When this returns `true`, the SDK sets [`TraceFlags::RANDOM`] on the span
+    /// context of every root span. The flag is a positive assertion about the
+    /// trace ID, so this defaults to `false` and implementations must opt in explicitly.
+    ///
+    /// [`new_trace_id`]: IdGenerator::new_trace_id
+    /// [`TraceFlags::RANDOM`]: opentelemetry::trace::TraceFlags::RANDOM
+    fn is_random(&self) -> bool {
+        false
```
cijothomas: lets not provide impln. let each generator implement this and return true/false.

ayush-that: done. each generator now has to implement it explicitly.

PR #3719, opened 2026-09-09
docs: explain filtering options and processor tradeoffs

cijothomas:
## Changes

Explain the tradeoffs between filtering in the logging library, an SDK processor, and the Collector or telemetry pipeline, including `tracing-subscriber`'s filtering capabilities. Present processor filtering as an option for applications with a reason to use it, with guidance on wrapping downstream processors and forwarding lifecycle methods.

Extend `logs-advanced` with a filtering processor composed with the existing enrichment processor, and add a compiling `SpanProcessor::on_end` example. Both use attributes for illustration; filtering conditions can use other available information or processor state.

Warn that filtering individual spans can produce incomplete or broken traces. Recommend tail sampling in the Collector or pipeline for coordinated decisions based on completed spans, and explain that it cannot recover spans already discarded by the SDK.

* [x] Unit tests added/updated (not applicable; the runnable example and doctest cover the documentation)
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial, user-facing changes (not applicable; documentation and examples only)
* [x] Changes in public API reviewed (not applicable; no public API changes)

### DIFF
--- a/docs/logs.md
+++ b/docs/logs.md
@@ -59,6 +59,34 @@ records reach that bridge.
 
 [issue #2877]: https://github.com/open-telemetry/opentelemetry-rust/issues/2877
 
+## Filtering log records
+
+Use logging-framework filters when possible: they reject events before an
+OpenTelemetry log record is created. A custom [`LogProcessor`] is useful when
+the decision requires the completed `SdkLogRecord`, its instrumentation scope,
+runtime configuration, or other processor state.
+
+A filtering processor should wrap the processor that exports or batches
+records. In its `emit` method, evaluate the filtering condition and call the
+wrapped processor only when the record should be kept. Register only the
+wrapper with `SdkLoggerProvider`. Registering both processors separately does
+not form a pipeline; each registered processor receives the record
+independently.
+
+The runnable [logs-advanced example] uses an attribute value as its filtering
+condition. A processor could instead filter by severity, event name, scope,
+body, external configuration, or any other information available to it. The
+example also demonstrates forwarding processor lifecycle methods and composing
+the filter with `SimpleLogProcessor`; a production application can wrap
+`BatchLogProcessor` in the same way.
+
+`LogProcessor::event_enabled` can reject records earlier, but it receives only
+severity, target, and event name. Conditions that require any other record data
+must be evaluated in `emit`.
+
+[`LogProcessor`]: https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/logs/trait.LogProcessor.html
+[logs-advanced example]: ../examples/logs-advanced/
+
 ## OpenTelemetry Log Bridge API
 
 Do **not** use the OpenTelemetry Log Bridge API (part of the `opentelemetry`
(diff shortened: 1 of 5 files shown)


docs/logs.md
```diff
@@ -59,6 +59,51 @@ records reach that bridge.
 
 [issue #2877]: https://github.com/open-telemetry/opentelemetry-rust/issues/2877
 
+## Filtering log records
+
+Filtering in a processor is one option. The appropriate place depends on what
```
utpilla, line 64: Could we structure this section around the situations that would drive someone to choose each filtering layer, before describing the benefits and costs?

A useful general rule might be to filter at the earliest layer that has both the information and policy control needed for the decision. Some concrete motivations would make the distinctions clearer:

- **Logging library:** Prefer this when the condition is available at the callsite, such as severity, target/module, event name, or other logging metadata. This avoids record construction, bridge conversion, SDK processing, batching, serialization, and transport.
**- SDK processor:** This becomes useful when the condition requires the normalized SdkLogRecord, instrumentation scope, or application-local state; when records from multiple logging libraries, bridges, or third-party instrumentation need one OpenTelemetry-level policy; or when different export destinations require different rules. Record creation and conversion have already happened, but rejected records can still be prevented from reaching batching and export.
- **Collector or telemetry pipeline:** This is appropriate when policy must be centrally managed across services, changed without application redeployment, or evaluated using pipeline-added information such as Kubernetes enrichment. It does not avoid application-side processing or transport to the Collector.

It may also be worth noting that downstream filtering is not the right boundary for sensitive data that must not leave the process or host.

Framing the section this way would explain what might require each choice; the current advantages and tradeoffs could then follow naturally.

docs/logs.md
```diff
@@ -59,6 +59,51 @@ records reach that bridge.
...
+  occurred, and the processor implementation is maintained in the application.
+- **Collector or telemetry pipeline:** Filtering downstream can centralize
+  policy across applications and use data added by pipeline processors. It
+  still incurs the application-side work and transport cost of sending records
+  to that point in the pipeline.
+
+If you have a reason to filter in an SDK processor, the following approach
+shows how to compose it with other processors.
+
+A filtering processor should wrap the processor that exports or batches
```
utpilla, line 85: Could we explain more explicitly why wrapping is required, perhaps with a small pipeline diagram?

`SdkLoggerProvider` delivers records independently to every processor registered with it. Therefore, registering a filtering processor beside a batch or simple processor does not allow the filter to prevent that processor from receiving the record, regardless of registration order.

The intended composition is:

```
SdkLoggerProvider
└── FilteringProcessor
    └── BatchLogProcessor
        └── Exporter
```

The filtering processor is the outermost processor and is invoked first. When it rejects a record, it does not call the wrapped processor, so the record is not queued, batched, or exported.

In contrast, this does not create a filtering pipeline:

```
SdkLoggerProvider
├── FilteringProcessor
└── BatchLogProcessor
    └── Exporter
```

Both processors receive the record independently. It would also help to clarify that the filter controls only the processor it wraps; any sibling processor remains unaffected, so each export branch requiring filtering must be wrapped separately.

utpilla (approved):
Left some suggestions. Overall, this is a useful improvement as it is.

PR #3720, opened 2026-09-10
fix(metrics): retain delta baseline across missed collection in PrecomputedSum

### LINKED ISSUE #3714: [Bug]: PrecomputedSum in Delta temporality resets baseline on missed/rejected observations, double-counting values
### What happened?

In `PrecomputedSum::delta`, collection drains and resets the `value_map`. Any attribute set that is not observed in the current cycle (e.g. an async callback temporarily misses an observation, errors, or an invalid measurement like NaN is dropped) is omitted from `new_reported`.
When `*reported = new_reported` executes, all unobserved baselines are cleared.
When the instrument subsequently observes a new measurement, `reported.get(...)` returns `None` and defaults to `0`. This causes the delta to be computed against `0` rather than the last valid baseline, re-counting previously exported values.

#### Reproduction:
1. Cycle 1: Observe `100` -> Collect: exports delta `100` (baseline cached as `100`).
2. Cycle 2: Missed observation (or dropped `NaN`) -> Collect: exports nothing, but clears `reported`.
3. Cycle 3: Observe `110` -> Collect: exports delta `110` (`110 - 0`) instead of `10` (`110 - 100`), double-counting the initial `100`.

### What did you expect to happen?
Missed or rejected observations should not immediately discard the baseline. The next valid observation should compute the delta relative to the last reported value.

0.32.0 (and main)

0.32.1 (and main)

_No response_

### Relevant log output

```shell

```

### Tip

darkraider01:
Fixes #3714

## Changes

- **Baseline retention & eviction**: Replaced the raw cached value in `PrecomputedSum::reported` with `BaselineEntry<T>` tracking `last_value` and consecutive `missed_count`.
  - Increments `missed_count` at the start of each collection cycle.
  - Resets `missed_count` to 0 whenever an attribute set has an accepted observation.
  - Retains the baseline through exactly one missed collection cycle so temporary gaps do not reset delta calculation.
  - Evicts the baseline after two consecutive missed collection cycles to keep memory bounded.
- **Non-finite rejection**: Filter out `NaN` and infinite values upfront in `Measure::call` and `BoundPrecomputedSumHandle::call`, preventing invalid measurements from being recorded and treating those cycles as missed collections.
- **Tests**:
  - Updated `observable_counter_delta_attribute_set_reappears_after_gap` to expect delta `30` (`80 - 50`) on reappearance after 1 missed cycle.
  - Added unit test `delta_rejects_nan_and_retains_baseline` for `f64` NaN rejection and baseline preservation.
  - Added unit test `delta_evicts_baseline_after_two_consecutive_missed_collections`.
  - Added integration tests `observable_counter_delta_stale_eviction` and `observable_counter_delta_f64_nan_rejection`.
- **Documentation**: Added entry in `opentelemetry-sdk/CHANGELOG.md` under `## vNext`.

### DIFF
--- a/opentelemetry-sdk/CHANGELOG.md
+++ b/opentelemetry-sdk/CHANGELOG.md
@@ -44,6 +44,11 @@
 - Fixed asynchronous counters (`ObservableCounter`, `ObservableUpDownCounter`)
   using delta temporality reporting incorrect deltas when observed attributes
   were recorded in an unsorted key order.
+- Fixed asynchronous counters (`ObservableCounter`, `ObservableUpDownCounter`)
+  with delta temporality losing their baseline after a single missed collection
+  cycle or rejected non-finite measurement, retaining the baseline through one
+  missed cycle and evicting after two consecutive missed cycles.
+  ([#3714](https://github.com/open-telemetry/opentelemetry-rust/issues/3714))
 
 ## 0.32.1
 
--- a/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
+++ b/opentelemetry-sdk/src/metrics/internal/precomputed_sum.rs
@@ -14,7 +14,10 @@ use super::{last_value::Assign, sort_and_dedup, AtomicTracker, Number, ValueMap}
 #[cfg(feature = "experimental_metrics_bound_instruments")]
 use super::{BoundMeasure, NoopBoundMeasure, TrackerEntry};
 use super::{ComputeAggregation, Measure};
-use std::{collections::HashMap, sync::Mutex};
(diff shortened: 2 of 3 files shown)


cijothomas (changes requested):
I’d prefer to wait for clarification in [specification(https://github.com/open-telemetry/opentelemetry-specification/issues/4861) before changing this behavior. The proposed one-cycle retention policy is not currently defined by the specification and its better to get a spec blessing to ensure consistency.

darkraider01:
Alright ! I'll fold it into draft PR until then

PR #3721, opened 2026-09-10
ci(http): check external API types

cijothomas:
## Changes

Add `opentelemetry-http` to the external-types CI matrix and define its intentional external API dependencies:

- `bytes` and `http` types used by the HTTP client contract and propagation adapters
- OpenTelemetry propagation traits implemented by the public adapters
- `hyper-util` connector types and bounds exposed by `HyperClient`

This ensures future changes to these external API dependencies require explicit review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
   external-types:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
LGTM once you're out of draft. good idea to add these

PR #3722, opened 2026-09-10
docs(http): mark status handling change as breaking

cijothomas:
Marks the built-in client status-handling change as breaking. The changelog retains the migration guidance for callers that want non-2xx responses converted into errors and clarifies that transport failures and timeouts still return `Err`.

### DIFF
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -17,9 +17,12 @@
   client request bodies.
 - Limit HTTP response body reads to 4 MiB in built-in HTTP clients (`reqwest` async/blocking and `hyper`) by enforcing the cap while streaming response chunks and reads that exceed the limit are aborted to prevent unbounded memory allocation.
 
-- Return HTTP error responses from the built-in reqwest and hyper clients instead
-  of converting 4xx and 5xx statuses into transport errors. This preserves the
-  response status and headers for exporter retry classification.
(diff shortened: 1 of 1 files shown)


opentelemetry-http/CHANGELOG.md
```diff
@@ -17,9 +17,10 @@
   client request bodies.
 - Limit HTTP response body reads to 4 MiB in built-in HTTP clients (`reqwest` async/blocking and `hyper`) by enforcing the cap while streaming response chunks and reads that exceed the limit are aborted to prevent unbounded memory allocation.
 
-- Return HTTP error responses from the built-in reqwest and hyper clients instead
-  of converting 4xx and 5xx statuses into transport errors. This preserves the
-  response status and headers for exporter retry classification.
+- **Breaking** Built-in reqwest and hyper clients now return HTTP 4xx and 5xx
+  responses as `Ok(Response<Bytes>)` instead of `Err(HttpError)`. This preserves
+  the response status and headers for exporter retry classification. Transport
+  failures and timeouts continue to return `Err`.
```
scottgerring: I think it would be worth mentioning briefly that the transport layer Result is telling us if we succeeded in making a request and getting a result, even if the result indicated a (logical) failure. 

When I first read this my brain immediately thought "why is a failure returning a success?" and this would clarify that for other folks.

scottgerring (approved):
approved with nitty-nit

PR #3723, opened 2026-09-10
docs(http): document client API and features

cijothomas:
Adds crate-level documentation covering propagation adapters, available HTTP-client features, custom `HttpClient` implementations, response semantics, the response-body limit, and Hyper runtime and constructor behavior.

Includes a compiling custom-client example and enables the crate’s first doctest.

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -1,3 +1,66 @@
+//! HTTP types and client adapters shared by OpenTelemetry components.
+//!
+//! [`HeaderInjector`] and [`HeaderExtractor`] adapt an [`http::HeaderMap`] to
+//! OpenTelemetry's text-map propagation interfaces. [`HttpClient`] is the
+//! transport abstraction used by exporters and other components that issue
+//! HTTP requests.
+//!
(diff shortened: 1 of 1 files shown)


opentelemetry-http/src/lib.rs
```diff
@@ -60,22 +123,26 @@ impl Extractor for HeaderExtractor<'_> {
...
 pub type HttpError = Box<dyn std::error::Error + Send + Sync + 'static>;
 
-/// A minimal interface necessary for sending requests over HTTP.
-/// Used primarily for exporting telemetry over HTTP. Also used for fetching
-/// sampling strategies for JaegerRemoteSampler
+/// A minimal interface for sending byte-oriented HTTP requests.
+///
+/// This is primarily used for exporting telemetry and for fetching remote
+/// sampling strategies. Implementations are responsible for enforcing any
+/// required request timeout, including while reading the response body.
```
scottgerring, line 133: Noting this is true once #3725 lands

scottgerring (approved):
Nice docs! I think we are getting to the tail of things; I note here we're not making interface changes.

PR #3724, opened 2026-09-10
fix(otlp): return reqwest client build errors

cijothomas:
## Summary

The OTLP HTTP exporter previously discarded errors from constructing its built-in reqwest client and silently substituted a default client. That fallback did not preserve the exporter-configured timeout, so exporter construction could succeed with behavior different from the requested configuration.

This change returns `ExporterBuildError::InternalFailure` when construction of either the async or blocking reqwest client fails.

For the blocking client, it also replaces panic-prone thread handling:

- A failure to spawn the client setup thread returns the existing `ExporterBuildError::ThreadSpawnFailed`.
- A panic in the setup thread returns `ExporterBuildError::InternalFailure`.

Custom HTTP clients and the Hyper client are unchanged. This PR is intentionally limited to `opentelemetry-otlp`; the deprecated Zipkin exporter is unchanged.

### DIFF
--- a/opentelemetry-otlp/CHANGELOG.md
+++ b/opentelemetry-otlp/CHANGELOG.md
@@ -34,6 +34,11 @@ release:
 
 ### Other changes
 
+- Return an exporter build error when construction of a built-in reqwest HTTP
+  client fails instead of silently falling back to a client without the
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
LGTM with a nit: do we want a test to cover the new error path?

PR #3725, opened 2026-09-10
fix(http): apply Hyper timeout to response body

cijothomas:
Extends `HyperClient`’s configured timeout across request dispatch, response headers, and complete bounded response-body collection.

Adds a regression test where the server sends headers and then stalls the response body.

### DIFF
--- a/opentelemetry-http/CHANGELOG.md
+++ b/opentelemetry-http/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## vNext
 
+- Apply `HyperClient`'s configured timeout to the complete response body, not
+  only request dispatch and response headers.
+
 - **Breaking** Sealed the `ResponseExt` trait so it can no longer be implemented by
   downstream crates. The trait provides a blanket implementation for all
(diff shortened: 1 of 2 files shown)


scottgerring (approved):
nice catch. how did you notice this?

cijothomas:
> nice catch. how did you notice this?

2 things
1. I've been thinking about how to enforce timeouts in the BatchProcessor (the default Thread based on do not enforce timeouts today) as we currently rely on export clients to enforce timeouts so a rogue exporter can stall entire pipeline. This is something we've discussed a lot couple of years ago too!
2. I've previously spend a lot of time thinking/documenting what duration is covered by a client/server Span - specifically - does the span ends when then first byte is written? or entire result? These are things which are not clearly covered in many instrumentations. We did document the behavior for .NET (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/README.md#activity-duration-and-httpserverrequestduration-metric-calculation) - with these context given, copilot did the rest easily.

PR #3726, opened 2026-09-11
fix(otlp): enforce complete export deadline

cijothomas:
Makes the configured OTLP timeout a total export budget covering serialization, every retry attempt, backoff, HTTP response bodies, and gRPC response decoding and trailers.

Adds `HttpClientTimeout` for communicating the remaining per-attempt budget to HTTP clients. Built-in reqwest and Hyper clients honor it, while tonic applies the remaining budget around the complete generated RPC future and sends `grpc-timeout`.

Includes dedicated-thread late-success coverage, decreasing retry-budget tests, response-body stall tests for every built-in HTTP client, and tonic response-body stall coverage.

Depends on #3725.

The full precommit script reaches its workspace feature-matrix phase, then encounters the existing cargo-hack integration-test manifest failure involving `opentelemetry-appender-tracing`.

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -233,31 +233,34 @@ pub mod hyper {
                     .headers_mut()
                     .insert(http::header::AUTHORIZATION, authorization.clone());
             }
-            let mut response = time::timeout(self.timeout, self.inner.request(request)).await??;
-            let capacity = response
-                .body()
-                .size_hint()
-                .upper()
-                .unwrap_or(0)
-                .min(MAX_RESPONSE_BODY_BYTES as u64) as usize;
-            let mut body_bytes = bytes::BytesMut::with_capacity(capacity);
-            let status = response.status();
-            let headers = std::mem::take(response.headers_mut());
-            let mut body = response.into_body();
-            while let Some(frame) = body.frame().await {
-                let frame = frame?;
-                if let Ok(chunk) = frame.into_data() {
-                    if body_bytes.len() + chunk.len() > MAX_RESPONSE_BODY_BYTES {
-                        return Err(Box::new(ResponseBodyTooLarge));
+            time::timeout(self.timeout, async {
+                let mut response = self.inner.request(request).await?;
+                let capacity = response
+                    .body()
+                    .size_hint()
+                    .upper()
+                    .unwrap_or(0)
+                    .min(MAX_RESPONSE_BODY_BYTES as u64) as usize;
+                let mut body_bytes = bytes::BytesMut::with_capacity(capacity);
+                let status = response.status();
+                let headers = std::mem::take(response.headers_mut());
+                let mut body = response.into_body();
+                while let Some(frame) = body.frame().await {
+                    let frame = frame?;
+                    if let Ok(chunk) = frame.into_data() {
+                        if body_bytes.len() + chunk.len() > MAX_RESPONSE_BODY_BYTES {
+                            return Err(Box::new(ResponseBodyTooLarge) as HttpError);
+                        }
+                        body_bytes.extend_from_slice(&chunk);
                     }
-                    body_bytes.extend_from_slice(&chunk);
                 }
-            }
-            let mut http_response = Response::builder()
-                .status(status)
-                .body(body_bytes.freeze())?;
-            *http_response.headers_mut() = headers;
-            Ok(http_response)
+                let mut http_response = Response::builder()
+                    .status(status)
+                    .body(body_bytes.freeze())?;
+                *http_response.headers_mut() = headers;
+                Ok(http_response)
+            })
+            .await?
         }
     }
 }
@@ -518,6 +521,25 @@ Connection: close\r\n\r\n",
             addr
         }
 
+        #[cfg(feature = "hyper")]
+        async fn start_stalled_body_server() -> SocketAddr {
+            let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+            let addr = listener.local_addr().unwrap();
+            tokio::spawn(async move {
+                if let Ok((mut socket, _)) = listener.accept().await {
+                    let mut buf = [0u8; 1024];
+                    let _ = socket.read(&mut buf).await;
+                    let _ = socket
+                        .write_all(
+                            b"HTTP/1.1 200 OK\r\nContent-Length: 1\r\nConnection: close\r\n\r\n",
+                        )
+                        .await;
+                    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                }
+            });
+            addr
+        }
+
         async fn assert_within_limit(client: &dyn HttpClient, addr: SocketAddr) {
             let request = Request::builder()
                 .method("POST")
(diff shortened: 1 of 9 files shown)


opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -441,10 +441,34 @@ async fn tonic_retry_with_backoff<F, Fut, T>(
...
+    any(feature = "trace", feature = "metrics", feature = "logs")
+))]
+async fn tonic_request_with_timeout<F, T>(
+    timeout: std::time::Duration,
+    request: F,
+) -> Result<T, tonic::Status>
+where
+    F: Future<Output = Result<T, tonic::Status>>,
+{
+    tokio::time::timeout(timeout, request)
```
scottgerring, line 469: It looks like this path is included in non-tokio based export paths, for instance the threaded processors, and will panic for lack of a tokio runtime. We need a proper runtime independent way to race a future against a timeout. In `retry.rs`, we have `sleep_for` which establishes the general pattern we've used to get retry going across these, but it does not handle the future side. 

I think in retry.rs we could add something like this:

```rust

   /// Polls `future` until it completes or `duration` expires.
   ///
   /// Dropping the request future after timeout provides cancellation where the
   /// underlying operation supports cancellation.
   pub(crate) async fn race_with_timeout<F>(
       duration: Duration,
       future: F,
   ) -> Result<F::Output, DeadlineElapsed>
   where
       F: Future,
   {
       let mut future = pin!(future);
       let mut deadline = pin!(futures_timer::Delay::new(duration));

       poll_fn(|cx| {
           // Poll the deadline first so expiration wins once the budget has elapsed.
           if deadline.as_mut().poll(cx).is_ready() {
               return Poll::Ready(Err(DeadlineElapsed));
           }

           future.as_mut().poll(cx).map(Ok)
       })
       .await
   }

```

This adds a dep on [futures_timer](https://crates.io/crates/futures-timer) which we could avoid if we really wanted to, but it is broadly used and appears quite stable.

opentelemetry-http/src/lib.rs
```diff
@@ -76,6 +96,9 @@ pub trait HttpClient: Debug + Send + Sync {
     ///
     /// Returns an error if it can't connect to the server or the request could not be completed,
     /// e.g. because of a timeout, infinite redirects, or a loss of connection.
+    ///
+    /// Implementations should honor [`HttpClientTimeout`] in the request's
+    /// extensions and apply it to the complete response body.
```
scottgerring, line 101: Should we clarify the custom `HttpClient` contract and reflect it in the changelog? `HttpClient::send_bytes` currently says implementations "should" honour `HttpClientTimeout`, while `with_http_client` says they "must" honour it for the exporter timeout to be a hard bound. Existing custom clients will silently ignore the new extension and can exceed the deadline. Probably we should either scope the hard-bound guarantee explicitly to the built-in clients, or make this consistently a MUST and include it in the changelog as migration guidance

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -422,24 +422,39 @@ impl OtlpHttpClient {
...
+            return Err(opentelemetry_sdk::error::OTelSdkError::InternalFailure(
+                "OTLP export deadline exceeded".to_string(),
+            ));
+        }
 
         let retry_data = Arc::new(HttpRetryData {
             body,
             headers: self.headers.clone(),
             endpoint: self.collector_endpoint.to_string(),
         });
```
scottgerring, line 440: This header and endpoint cloning happens after retry_budget is calculated, so it is excluded from the total export budget. Could we construct retry_data before calculating the remaining budget?

opentelemetry-otlp/src/exporter/tonic/mod.rs
```diff
@@ -821,6 +845,20 @@ mod tests {
     use http::{HeaderMap, HeaderName, HeaderValue};
     use tonic::metadata::{MetadataMap, MetadataValue};
 
+    #[tokio::test]
+    #[cfg(any(feature = "trace", feature = "metrics", feature = "logs"))]
+    async fn tonic_response_body_stall_is_bounded_by_timeout() {
```
scottgerring, line 850: This shows that the timeout helper bounds an arbitrary pending future, but it doesn't exercise a tonic response body or the exporter path. Could we add a regression test through at least one real tonic exporter/client where the server sends response headers and then stalls the body or trailers? That would lock in the integration between the generated RPC future, `Request::set_timeout`, and the signal client’s use of the helper.

opentelemetry-http/src/lib.rs
```diff
@@ -101,14 +124,19 @@ mod reqwest {
...
 
     #[async_trait]
     impl HttpClient for reqwest::Client {
         async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError> {
             otel_debug!(name: "ReqwestClient.Send");
-            let request = request.try_into()?;
+            let timeout = request.extensions().get::<HttpClientTimeout>().copied();
+            let mut request: reqwest::Request = request.try_into()?;
+            if let Some(timeout) = timeout {
+                *request.timeout_mut() = Some(timeout.duration());
```
scottgerring, line 138: I believe this will override a built-in timeout the user may have set on their customised client. If the user specifies a _shorter_ timeout than our remaining budget, I think increasing it to our budget is incorrect. I think we should use the shorter of the two.

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -422,24 +422,39 @@ impl OtlpHttpClient {
     {
         use crate::retry::retry_with_backoff;
 
+        let export_start = std::time::Instant::now();
         // Build request body once before retry loop
         let (body, content_type, content_encoding) = build_body_fn(self, data)
```
scottgerring, line 427: we account for this, but we can't bound it. If someone does something particularly heinous on a short deadline we might timeout during this.

I think this is weird edge case stuff and would be fine with just documenting it

opentelemetry-otlp/src/retry.rs
```diff
@@ -122,8 +120,23 @@ where
...
+            return Err(deadline_error());
+        }
+
+        match operation(remaining).await {
+            Ok(result) => {
+                if start.elapsed() >= deadline {
+                    otel_warn!(name: "Export.Failed.DeadlineExceeded",
+                        operation = operation_name,
+                        retries = attempt,
+                        message = "OTLP export completed after its deadline - telemetry data will be lost"
```
scottgerring, line 134: Should we provide some advice here?

scottgerring (commented):
(noting this is a draft, but keen to get the OTLP stuff tucked away).

 Beyond the inline comments, I wonder if we could simplify and reduce the risk here by:

 - retaining the existing total retry budget, but passing its remaining duration to each built-in transport attempt
 - treating serialisation and request preparation as time deducted from that budget, rather than claiming they are strictly interruptible
 - scoping the hard timeout guarantee to built-in HTTP clients, leaving the extension advisory for custom clients
 - handling tonic’s complete-response timeout separately using a runtime-independent mechanism.

 I reckon this would address the important failure mode - that we can have substantial runovers on total duration with slow retries while reducing the scope a bit

PR #3728, opened 2026-09-11
fix: redact exporter credentials from Debug output

cijothomas:
This is preventive hardening of `Debug` output, not a fix for a known credential leak in the SDK’s internal logging, and is not a stability or release blocker.

Applications may print an exporter or builder while troubleshooting configuration. Derived `Debug` output can include configured credentials even when the application only intended to inspect the surrounding configuration. This change makes that operation safer by summarizing configuration and marking HTTP headers and gRPC metadata as sensitive, following the existing Rust HTTP ecosystem pattern.

Credentials remain accessible and are sent unchanged; this does not prevent applications from explicitly printing them. Collector endpoints remain visible. Regression tests check that credentials remain usable while being omitted from debug output.

### DIFF
--- a/opentelemetry-http/src/lib.rs
+++ b/opentelemetry-http/src/lib.rs
@@ -76,6 +76,9 @@ pub trait HttpClient: Debug + Send + Sync {
     ///
     /// Returns an error if it can't connect to the server or the request could not be completed,
     /// e.g. because of a timeout, infinite redirects, or a loss of connection.
+    ///
+    /// Implementations must ensure their [`Debug`] output does not expose
+    /// authentication headers or other credentials.
     async fn send_bytes(&self, request: Request<Bytes>) -> Result<Response<Bytes>, HttpError>;
 }
 
@@ -183,7 +186,7 @@ pub mod hyper {
     use std::time::Duration;
     use tokio::time;
 
-    #[derive(Debug, Clone)]
+    #[derive(Clone)]
     pub struct HyperClient<C = HttpConnector>
     where
         C: Connect + Clone + Send + Sync + 'static,
@@ -200,6 +203,10 @@ pub mod hyper {
         pub fn new(connector: C, timeout: Duration, authorization: Option<HeaderValue>) -> Self {
             // TODO - support custom executor
             let inner = Client::builder(hyper_util::rt::TokioExecutor::new()).build(connector);
+            let authorization = authorization.map(|mut value| {
+                value.set_sensitive(true);
+                value
+            });
             Self {
                 inner,
                 timeout,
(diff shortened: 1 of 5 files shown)


opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -804,6 +833,12 @@ fn add_header_from_string(input: &str, headers: &mut HashMap<HeaderName, HeaderV
     }));
 }
 
+fn mark_header_values_sensitive(headers: &mut HashMap<HeaderName, HeaderValue>) {
+    for value in headers.values_mut() {
+        value.set_sensitive(true);
```
scottgerring, line 838: I wondered if there are any downsides to marking _everything_ sensitive. It looks like this might reduce compressibility of the HTTP stream --> 

[set_sensitive](https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.is_sensitive) API mentions:
> HPACK encoders used by HTTP/2.0 implementations can choose not to compress them

HPACK is a mechanism to de-dupe headers over multiple requests, so by saying _everything_ is sensitive in some situations we'll end up recalling the complete header set every time rather than just pointing at previous HPACK records. 

Is marking every configured value sensitive an intentional conservative trade-off, or should we constrain it to headers likely to contain credentials? If the former probably we should document this

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -404,6 +418,21 @@ pub(crate) struct OtlpHttpClient {
     resource: opentelemetry_proto::transform::common::tonic::ResourceAttributesWithSchema,
 }
 
+impl Debug for OtlpHttpClient {
+    fn fmt(&self, formatter: &mut Formatter<'_>) -> std::fmt::Result {
+        formatter
+            .debug_struct("OtlpHttpClient")
+            .field("collector_endpoint", &self.collector_endpoint)
```
scottgerring, line 425: I know you mention this in the PR description, but it is not unheard of to include API keys in the query string. Given the otherwise conservative nature of the PR (treating all headers as secret) - this jumps out.  Not sure what to suggest ... maybe we can just drop all query parameters for debug?

opentelemetry-otlp/src/exporter/http/mod.rs
```diff
@@ -900,6 +935,72 @@ mod tests {
...
+        let debug = format!("{builder:?}");
+        assert!(!debug.contains(SECRET));
+        assert!(debug.contains("client_configured"));
+        assert!(debug.contains("header_count"));
+
+        let mut headers = std::collections::HashMap::from([(
+            http::HeaderName::from_static("authorization"),
+            http::HeaderValue::from_static(SECRET),
+        )]);
+        super::mark_header_values_sensitive(&mut headers);
```
scottgerring, line 977: Could we test this through the regular exporter path? If the production builder stopped calling the helper, this test would still pass

opentelemetry-http/src/lib.rs
```diff
@@ -134,6 +134,9 @@ pub type HttpError = Box<dyn std::error::Error + Send + Sync + 'static>;
 ///
 /// HTTP clients may depend on a particular async runtime. This trait allows
 /// users to supply an implementation suitable for their runtime.
+///
+/// Implementations must ensure their [`Debug`] output does not expose
+/// credentials or other sensitive configuration.
```
scottgerring, line 139: noting that this is effectively unenforceable, but probably that's just how it has to be. Perhaps pragmatically we should in the changelog clarify that this is a requirement for them? Something like:

>  Custom HttpClient implementations remain responsible for ensuring their own Debug output does not expose credentials or other sensitive configuration. OTLP exporter Debug output no longer formats configured client internals, but directly formatting a custom client may still expose whatever its implementation includes.

scottgerring (commented):
i note this is draft, but some comments inline!

PR #3732, opened 2026-09-14
perf(sdk): reduce span processor export allocations

utpilla:
## Changes

Updates the Trace SDK exporter interface to use a borrowed `SpanBatch`, similar to the `LogBatch` change in #2057.

This avoids unnecessary allocations in the span processors:

- `SimpleSpanProcessor` uses a stack-allocated single-span batch instead of `vec![span]`.
- `BatchSpanProcessor` clears and reuses its batch buffer instead of allocating through `split_off(0)`.

Exporters that require ownership, including OTLP exporters, continue to copy the spans at their existing ownership boundary. OTLP serialization and transport behavior are otherwise unchanged.

## API change

`SpanExporter::export` now accepts `SpanBatch<'_>` instead of `Vec<SpanData>`.

`SpanBatch` exposes only `iter()` to keep the new public API minimal.

### DIFF
--- a/opentelemetry-otlp/src/exporter/tonic/trace.rs
+++ b/opentelemetry-otlp/src/exporter/tonic/trace.rs
@@ -9,7 +9,7 @@ use opentelemetry_proto::transform::trace::tonic::group_spans_by_resource_and_sc
 use opentelemetry_sdk::error::OTelSdkError;
 use opentelemetry_sdk::{
     error::OTelSdkResult,
-    trace::{SpanData, SpanExporter},
+    trace::{SpanBatch, SpanData, SpanExporter},
 };
 use tonic::{codegen::CompressionEncoding, service::Interceptor, transport::Channel, Request};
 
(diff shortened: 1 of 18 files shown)


opentelemetry-sdk/src/trace/export.rs
```diff
@@ -7,6 +7,26 @@ use std::borrow::Cow;
...
+///
+/// The batch borrows span data from the processor for the duration of the
+/// export operation.
+#[derive(Debug)]
+pub struct SpanBatch<'a> {
+    spans: &'a [SpanData],
+}
+
+impl<'a> SpanBatch<'a> {
+    pub(crate) fn new(spans: &'a [SpanData]) -> Self {
```
lalitb, line 20: Could we provide a public way to construct `SpanBatch`, either through this constructor or `From<&[SpanData]>`? Custom processors that manage their own batching need it to call `SpanExporter::export`.

This is also public for `LogBatch`.

opentelemetry-otlp/src/exporter/tonic/trace.rs
```diff
@@ -67,7 +67,8 @@ impl TonicTracesClient {
 }
 
 impl SpanExporter for TonicTracesClient {
-    async fn export(&self, batch: Vec<SpanData>) -> OTelSdkResult {
+    async fn export(&self, batch: SpanBatch<'_>) -> OTelSdkResult {
+        let batch: Vec<SpanData> = batch.iter().cloned().collect();
```
lalitb, line 71: This recreates the vector allocation removed from the processor and  additionally clones every span. Can we use the pattern used for logs. 

Same for http/trace.rs

scottgerring:
Added `#performance` label so we can see the managed benchmark results

PR #3734, opened 2026-09-14
perf(sdk): borrow completed spans across processors

utpilla:
## Summary

Change `SpanProcessor::on_end` to receive a borrowed `&SpanData` instead of taking ownership.

The SDK now builds the completed `SpanData` once and passes the same immutable reference to every registered processor. Processors clone only when they need to retain the span after `on_end` returns, such as when enqueueing it for background processing.

This is independent of the `SpanBatch` exporter change in #3732.

## Processor convention

- `on_start(&mut Span, &Context)` remains the mutation hook while the span is recording.
- `on_end(&SpanData)` is for inspecting or forwarding finalized span data.
- Processors that retain completed spans must clone them explicitly.

This follows the ownership convention introduced for Logs in [#1726](https://github.com/open-telemetry/opentelemetry-rust/pull/1726), where processors receive a borrowed log record and asynchronous processors clone only when they require ownership. Trace uses an immutable borrow because `SpanData` represents an already completed span, while Logs use a mutable borrow so processors can enrich a record before forwarding it.

### Filtering processors

A filtering processor can inspect and forward the same completed span without cloning:

```rust
fn on_end(&self, span: &SpanData) {
    if should_export(span) {
        self.next.on_end(span);
    }
}
```

The existing filtering example and processor documentation have been updated to show this convention.

## Benchmark

Added `span-creation-simple/two-processors`, using two no-op processors to isolate the cost of passing completed spans through the processor chain.

| Version | Time per span |
|---|---:|
| Owned `SpanData` | 799.55 ns |
| Borrowed `&SpanData` | 407.67 ns |

This is approximately **49.0% faster** for the measured two-processor scenario with five span attributes.

## API change

Processor implementations change from:

```rust
fn on_end(&self, span: SpanData)
```

to:

```rust
fn on_end(&self, span: &SpanData)
```

Processors that send spans to another thread or otherwise retain them must call `span.clone()`.


scottgerring:
Hey @utpilla , we have had a number of folks working on this in #3659 for quite some time; is there something missing there?

utpilla:
> Hey @utpilla , we have had a number of folks working on this in #3659 for quite some time; is there something missing there?

Thanks for pointing that @scottgerring! I'll close this PR in favor of that. I have left a comment there about the API design: https://github.com/open-telemetry/opentelemetry-rust/pull/3659#issuecomment-5672724637

PR #3742, opened 2026-09-16
refactor(otlp)!: separate configuration parsing errors

cijothomas:
`Compression::from_str` currently returns `ExporterBuildError`, even though parsing a configuration value does not build an exporter. Return a dedicated, opaque `ParseConfigError` instead, with `Display` and `std::error::Error` implementations. Its fields remain private, and diagnostic wording is not a stable programmatic interface.

Exporter builder usage, accepted compression strings, and environment configuration behavior are unchanged. Invalid environment values still warn and fall back. This PR leaves `ExporterBuildError` unchanged and separates the parsing contract from the build-error consolidation in #3692. It adds no other public parsers.

### Why `ParseConfigError`

Compression is currently the only setting with a public string parser. The general name is intentional: invalid configuration strings have the same handling needs, so future public parsers for other OTLP settings could reuse this opaque error rather than introduce a separate type for each setting. This does not commit us to adding those parsers or exposing setting-specific error variants, fields, or accessors.

The distinction is between parsing a configuration value (`ParseConfigError`) and constructing an exporter (`ExporterBuildError`). `ParseCompressionError` would describe today's only producer more narrowly, but would prevent naturally sharing the type if another configuration parser is added.

### Migration

This is breaking for code that expects compression parsing to return `ExporterBuildError`, including explicit result types, variant matches, and propagation into functions returning that error type. Update those parsing paths to use `ParseConfigError` or map it into the application's own error type.

```rust
// Before:
let result: Result<Compression, ExporterBuildError> = value.parse();

// After:
let result: Result<Compression, ParseConfigError> = value.parse();
```

Normal `.with_compression(Compression::Gzip)` and `.build()` calls need no changes.

### Alternative considered

Remove the public `FromStr` implementation for `Compression` and keep parsing internal to exporter configuration. This would avoid introducing `ParseConfigError`, but would break existing `.parse::<Compression>()` / `Compression::from_str(...)` calls and integrations requiring `Compression: FromStr`. Typed builder calls and exporter environment configuration would continue to work.

Future declarative configuration support in Rust would likely reduce the need for applications to parse these strings themselves: the configuration layer can validate values and pass typed settings to exporter builders without requiring a public `FromStr` implementation. This could reduce the inconvenience of removing the parser, but would not eliminate the immediate migration cost or cover every application's custom configuration needs. This PR currently retains `FromStr`; removal is an alternative for reviewers to consider before stabilization.


opentelemetry-otlp/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
 ## vNext
 
+- Exporter builder usage and environment configuration are unchanged.
+  **Breaking for callers parsing compression strings:** `Compression::from_str`
+  (including `.parse::<Compression>()`) now returns the opaque `ParseConfigError`
+  instead of `ExporterBuildError`. Update explicit result types and error handling
+  that expects `ExporterBuildError::UnsupportedCompressionAlgorithm`. The new error
+  implements `Display` and `std::error::Error`; its message is for diagnostics.
+  Accepted strings and parsing behavior are unchanged.
+
```
scottgerring, line 12: can we add a brief example?

opentelemetry-otlp/src/exporter/mod.rs
```diff
@@ -224,16 +224,26 @@ impl Display for Compression {
     }
 }
 
+/// An error parsing an OTLP configuration value.
+///
+/// The error message provides diagnostic context. Its wording is not a stable
+/// interface and should not be used for programmatic decisions.
+#[derive(Debug, Error)]
+#[error("{message}")]
+pub struct ParseConfigError {
```
scottgerring, line 233: Should it be ... 

```suggestion
pub struct ParseCompressionError {
```

? The name implies it is broader than it is

scottgerring (approved):
Couple of nits, otherwise LGTM.

I'm not totally convinced we shouldn't just make this private and use it only from the env var parsing, but as we've already got a lot of changes in this release and I don't feel particularly strongly about it happy to err on the side of not accidentally breaking folks and address it in this fashion!

opentelemetry-otlp/src/exporter/mod.rs
cijothomas: We picked the broader name so future parsers for other settings could reuse it, without needing an error type per setting. Compression is the only one today. Added that reasoning to the PR description.

PR #3743, opened 2026-09-17
fix: defer context value drops until after borrow release

### LINKED ISSUE #3738: [Bug]: ContextGuard::drop panics when a stored context value calls Context::current()
### What happened?

Dropping a context guard can panic if a value stored with `Context::with_value()` calls `Context::current()`  from its `Drop` implementation.

Min repro:

`Cargo.toml` :
```rust
[package]
name = "context-drop-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
opentelemetry = { version = "=0.32.0" }
```

`src/main.rs`:
```rust
use opentelemetry::Context;

struct ReadContextOnDrop;

impl Drop for ReadContextOnDrop {
    fn drop(&mut self) {
        Context::current();
    }
}

fn main() {
    drop(Context::new().with_value(ReadContextOnDrop).attach());
}
```

0.32

0.32

_No response_

### Relevant log output

```shell
thread 'main' (1108606) panicked at /Users/<>/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/opentelemetry-0.32.0/src/context.rs:164:38:
RefCell already mutably borrowed
```

### Tip

uasmirnov:
Fixes #3738

## Changes

Context values are now dropped after the internal borrow is released.
This prevents a panic if their Drop implementation calls Context::current().

I added tests for normal and out of order guard drops, benchmarks showed up to about 4% overhead.


scottgerring:
Oh, this is interesting. I haven't reviewed the PR, but because the context stuff is often hot-path-fiddly I figured i'd have a quick look at the benchmarks!

TL;DR on my M3 Max using rustc 1.98.0 I actually see an improvement - all of the `context_attach` cases measured between 20 and 30% faster 🫨 Picking into the binaries, it seems like in my local build setup (i.e. i doubt this is consistently reproducible) ,  this caused LLVM to inline the LocalKey::with stuff into ContextGuard::drop, saving us an extra branch and the function call overhead. This isn't really because we're doing less work, its just that the PR happened to change its inlining. 

As our benchmarking infra here is currently broken, i've raised #3744  to fix it so we can see what our formal system thinks about the perf changes. It would also be great if @bantonsson could have a look when he has a chance as he is the most familiar with the context hot path work.

bantonsson (approved):
Thank you @uasmirnov for the fix.

I didn't get the _magic_ performance increase that @scottgerring did, but a performance decrease of 2.5-4% which I think is totally acceptable for this fix.

scottgerring:
Nothing notable in [CI benchmark run](https://github.com/open-telemetry/opentelemetry-rust/actions/runs/35318677080/job/105515966570?pr=3743) 🥳

lalitb:
The overflow [branch](https://github.com/uasmirnov/opentelemetry-rust/blob/6c5fd583b2805166393e59b48a827c3a44345a82/opentelemetry/src/context.rs#L623) in `ContextStack::push` still drops the rejected context while the mutable borrow is held, so a stored value calling `Context::current()` during cleanup can panic there. Could we track that in a follow-up issue? It predates this PR and doesn't need to block this fix.

lalitb:
LGTM. I also ran the context attach/drop benchmarks on my MacBook M4 Pro with Rust 1.98.0. My results were closer to @scottgerring's: the PR was consistently 25–37% faster across all existing nine cases in two runs per revision. No regression observed in this setup.

lalitb (approved):

uasmirnov:
> The overflow [branch](https://github.com/uasmirnov/opentelemetry-rust/blob/6c5fd583b2805166393e59b48a827c3a44345a82/opentelemetry/src/context.rs#L623) in `ContextStack::push` still drops the rejected context while the mutable borrow is held, so a stored value calling `Context::current()` during cleanup can panic there. Could we track that in a follow-up issue? It predates this PR and doesn't need to block this fix.

@lalitb I couldn’t find an issue for this; i will open issue and try to fix it while the code is still fresh.

Issue #3127, opened 2025-08-14
[Feature]: add support for `OTEL_PROPAGATORS` environment variable


jan-xyz:
### Related Problems?

We use this environment var via a provided configmap from the platform to have a consistent propagator in our Kubernetes cluster. I would be nice to also have the support for it in the rust ecosystem.

### Describe the solution you'd like:

Automatically configuring the Propagator based on the `OTEL_PROPAGATORS` environment variable instead of a noop propagator.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

taisho6339:
I've identified a critical architectural challenge when implementing this feature. We could fix [this part](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/global/propagation.rs#L19-L23
) to load the env var and set the specified propagators automatically, but it's impossible. 
The reason is that Propagator implementations (TraceContextPropagator, BaggagePropagator) are located in opentelemetry-sdk, whereas the propagation file is in the opentelemetry crate. This separation creates a dependency on opentelemetry-sdk from opentelemetry.

## Two Possible Approaches

### Option 1: Move propagators to opentelemetry (Breaking Change)

Move TraceContextPropagator and BaggagePropagator from opentelemetry-sdk to opentelemetry. This enables implicit environment variable loading at initialization.

✅ Enables automatic configuration like Python SDK
⚠️ Breaking change: Users must update import paths

### Option 2: Provide explicit helper function (Non-Breaking)

Add set_global_text_map_propagator_from_env() in opentelemetry-sdk. Users must explicitly call this function at application startup.

✅ No breaking changes
❌ Not automatic/implicit as per specification
❌ Requires user action (doesn't match Python/Java behavior)

Thus, I'd be happy to gather your opinions.

scottgerring:
Having another look at this since @darkraider01 expressed interest in working on it.

I agree with @taisho6339 that the first option is a non-starter. The second option, adding an explicit helper in `opentelemetry-sdk`, is clearly a net improvement, and I think we should start there.

[...]

PR #3745, opened 2026-09-17
feat(sdk): add set_global_text_map_propagator_from_env helper

darkraider01:
Part of #3127

## Changes

Adds an explicit helper function in `opentelemetry-sdk` to configure the process-global text map propagator from the `OTEL_PROPAGATORS` environment variable:
`opentelemetry_sdk::propagation::set_global_text_map_propagator_from_env()`

Key details:
- Respects the OpenTelemetry specification default: `"tracecontext,baggage"`.
- Supports built-in propagators: `tracecontext`, `baggage`, and `none`.
- Handles case-insensitivity, surrounding whitespace, and deduplication of propagator entries.
- Unrecognized or unsupported values (such as `b3` or `jaeger`) emit internal `otel_warn!` diagnostics and are gracefully ignored.
- Pure parsing and construction logic is separated from global mutation to enable isolated unit testing without global state mutation.
- Automatic integration with `SdkTracerProvider` / builder initialization is intentionally deferred to avoid global side effects and multiple-provider issues.


opentelemetry-sdk/src/propagation/env.rs
```diff
@@ -0,0 +1,409 @@
...
+                        propagators.push(Box::new(TraceContextPropagator::new()));
+                    }
+                    "baggage" => {
+                        propagators.push(Box::new(BaggagePropagator::new()));
+                    }
+                    _ => {
+                        otel_warn!(
+                            name: "TextMapPropagator.Config.UnsupportedPropagator",
+                            message = format!(
+                                "Propagator '{name}' is not supported by this SDK helper and was ignored. Supported values are: tracecontext, baggage, none."
```
cijothomas: when we say "ignored", does it mean we proceed as-if the ENV did not exist, and fallback to the default?

darkraider01: Ah, no  I meant we just skip that unsupported value. We don’t treat the whole env var as missing or fall back to the default. I’ll reword this to make that clearer.
(and add coverage if needed)

opentelemetry-sdk/src/propagation/env.rs
```diff
@@ -0,0 +1,416 @@
...
+        PropagatorConfig::None => TextMapCompositePropagator::new(vec![]),
+        PropagatorConfig::MixedNone => {
+            otel_warn!(
+                name: "TextMapPropagator.Config.InvalidCombination",
+                message = format!(
+                    "OTEL_PROPAGATORS contains 'none' combined with other propagators ('{raw}'). The 'none' propagator must not be combined with other propagators. Falling back to an empty text map propagator."
+                ),
+                otel_propagators = raw,
+            );
+            TextMapCompositePropagator::new(vec![])
```
scottgerring: The `MixedNone` arm falls back to an empty composite, which discards any valid propagators the user also listed (e.g. `"none,tracecontext"` produces no propagation at all). The spec's [general Enum guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#enum) is to warn on unrecognised/invalid entries and skip them, keeping valid ones. A symmetrical treatment here would be: warn that `none` cannot be combined with others, skip the `none`, and apply the remaining valid propagators - consistent with how `b3` or `jaeger` entries are handled.

opentelemetry-sdk/src/propagation/env.rs
scottgerring: Changed wording here LGTM

scottgerring (changes requested):
One comment inline, and i think you need to resolve conflicts!

opentelemetry-sdk/src/propagation/env.rs
darkraider01: mixed none configurations now warn and skip none while preserving the remaining valid propagators!

PR #3748, opened 2026-09-18
feat: extract B3 propagator into a standalone crate

cijothomas:
The Zipkin exporter is deprecated, but B3 propagation remains required as a core package by the [OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#propagators-distribution). Add `opentelemetry-propagator-b3` so applications can keep using B3 when the Zipkin exporter is removed.

Copy the existing B3 implementation and tests into the new crate, preserving its API and propagation behavior, including the multiple-header default. Users migrate by changing their dependency and imports to `opentelemetry_propagator_b3::{B3Encoding, Propagator}`.

Includes migration guidance, the new crate's changelog, and CI/publishing configuration. The Zipkin crate is unchanged; its removal will be handled separately.

Validation: formatting, Clippy, and all 16 B3 and Zipkin unit tests pass locally. The new crate's doctest, public-type check, and packaging dry run also pass.


lalitb (approved):
LGTM. Thanks.

